PR #1875, opened 2018-12-13
Support R4 label for binary trainers

### LINKED ISSUE #1874: Our binary learners estimators work only on top of boolean label
https://github.com/dotnet/machinelearning/blob/dfe9f3ad36cf382516276ad902a5f89a5f21c7e7/test/Microsoft.ML.Benchmarks/KMeansAndLogisticRegressionBench.cs
Fail currently because R4 is not supported.

Ivanidzo4ka:
fix #1874

### DIFF
--- a/src/Microsoft.ML.Data/Training/TrainerEstimatorBase.cs
+++ b/src/Microsoft.ML.Data/Training/TrainerEstimatorBase.cs
@@ -2,11 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Collections.Generic;
-using System.Linq;
 using Microsoft.ML.Core.Data;
 using Microsoft.ML.Data;
 using Microsoft.ML.Runtime.Data;
+using System.Linq;
 
 namespace Microsoft.ML.Runtime.Training
 {
@@ -167,7 +166,7 @@ namespace Microsoft.ML.Runtime.Training
                 SchemaShape.Column label,
                 SchemaShape.Column weight = default,
                 SchemaShape.Column groupId = default)
-            :base(host, feature, label, weight)
+            : base(host, feature, label, weight)
         {
             GroupIdColumn = groupId;
         }
--- a/src/Microsoft.ML.FastTree/FastTreeClassification.cs
+++ b/src/Microsoft.ML.FastTree/FastTreeClassification.cs
@@ -227,6 +227,19 @@ namespace Microsoft.ML.Trainers.FastTree
             _trainSetLabels = GetClassificationLabelsFromRatings(TrainSet).ToArray(TrainSet.NumDocs);
             //Here we set regression labels to what is in bin file if the values were not overriden with floats
         }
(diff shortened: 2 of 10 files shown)


sfilipi:
>         public readonly SchemaShape.Column LabelColumn;

My problem with this is that we allow those types in the check, but the SchemaShape column contains only one ColumnType/RawType.... and that gets defined in the ctor of the trainers as a ...boolean.

---
Refers to: src/Microsoft.ML.Data/Training/TrainerEstimatorBase.cs:36 in ca674d3. [](commit_id = ca674d384cf6c3dcdfc5350a0e56e01a82ed2c0b, deletion_comment = False)

sfilipi:
>                   TrainerUtils.MakeBoolScalarLabel(args.LabelColumn))

Like here, this defines it as a bool.. 

---
Refers to: src/Microsoft.ML.HalLearners/SymSgdClassificationTrainer.cs:184 in ca674d3. [](commit_id = ca674d384cf6c3dcdfc5350a0e56e01a82ed2c0b, deletion_comment = False)

Ivanidzo4ka:
>                   TrainerUtils.MakeBoolScalarLabel(args.LabelColumn))

I don't like this either, but I can't come up with better solution. 
Array of supported columns a bit weird.
Don't pass anything for binary classification, is also lame.

---
In reply to: [447171216](https://github.com/dotnet/machinelearning/pull/1875#issuecomment-447171216) [](ancestors = 447171216)

---
Refers to: src/Microsoft.ML.HalLearners/SymSgdClassificationTrainer.cs:184 in ca674d3. [](commit_id = ca674d384cf6c3dcdfc5350a0e56e01a82ed2c0b, deletion_comment = False)

sfilipi:
> > ```
> >               TrainerUtils.MakeBoolScalarLabel(args.LabelColumn))
> > ```
> 
> I don't like this either, but I can't come up with better solution.
> Array of supported columns a bit weird.
> Don't pass anything for binary classification, is also lame.
> 
> In reply to: [447171216](https://github.com/dotnet/machinelearning/pull/1875#issuecomment-447171216) [](ancestors = 447171216)
> 
> Refers to: src/Microsoft.ML.HalLearners/SymSgdClassificationTrainer.cs:184 in [ca674d3](https://github.com/dotnet/machinelearning/commit/ca674d384cf6c3dcdfc5350a0e56e01a82ed2c0b). [](commit_id = [ca674d3](https://github.com/dotnet/machinelearning/commit/ca674d384cf6c3dcdfc5350a0e56e01a82ed2c0b), deletion_comment = False)

Why not support just boolean? The users can always convert it.

Anipik:
will it also enable R4 support for multiclass classfication ?

TomFinley:
> Why not support just boolean? The users can always convert it.

This is the correct policy. Indeed have we codified this in the case of [`IDataView` implementation specifically](https://github.com/dotnet/machinelearning/blob/master/docs/code/IDataViewImplementation.md#getters-must-fail-for-invalid-types), but it also applies here.

`IDataView`s are composable structures. Also we are writing an API, and an API with too many "smarts" baked in never winds up being such a smart API.

Consider closing this PR. Thanks.

Issue #561, opened 2018-07-19
API to create a TextLoader from class metadata


eerhardt:
In the `LearningPipeline` API, we have the ability to create a TextLoader object using metadata applied to a regular C# class:

```C#
pipeline.Add(new TextLoader(dataPath).CreateFrom<HousePriceData>(useHeader: true, separator: ','));

public class HousePriceData
{
    [Column(ordinal: "0")]
    public string Id;
    [Column(ordinal: "1")]
    public string Date;
    [Column(ordinal: "2", name: "Label")]
    public float Price;
    [Column(ordinal: "3")]
    public float Bedrooms;
    [Column(ordinal: "4")]
    public float Bathrooms;
    [Column(ordinal: "5")]
    public float SqftLiving;
    [Column(ordinal: "6")]
    public float SqftLot;
    ...
```

I find it more intuitive to have a class decorated with metadata to use to load data instead of imperatively building up a schema in code, like the following:

```C#
private static TextLoader.Column ScalarCol(string name, int ordinal, DataKind kind = DataKind.Num)
    => new TextLoader.Column() { Name = name, Type = kind, Source = new[] { new TextLoader.Range() { Min = ordinal, Max = ordinal } } };

var loader = new TextLoader(env, new TextLoader.Arguments()
{
    HasHeader = true,
    SeparatorChars = new char[] { ',' },
    // These column declarations are meant to mirror those that appear in HousePriceData.
    Column = new[] {
        ScalarCol("Id", 0, DataKind.Text),
        ScalarCol("Date", 1, DataKind.Text),
        ScalarCol("Label", 2),
        ScalarCol("Bedrooms", 3),
        ScalarCol("Bathrooms", 4),
        ScalarCol("SqftLiving", 5),
        ScalarCol("SqftLot", 6),
}, new MultiFileSource(dataPath));
```

This issue is being opened to ensure we preserve this behavior with the new direct access API design proposed in #371 (possibly using a different API design, but preserving the functionality).

/cc @ericstj @TomFinley @Zruty0 @terrajobst

Zruty0:
I definitely think we should be able to do both. 

Declarative schema is shorter, but imperative schema allows to read the data without knowing its schema at compile time.

eerhardt:
Agreed.  I wasn't proposing to remove the imperative API to build the columns up using code. That is definitely necessary.  I was proposing building a "convenience" API on top of it, which built up the objects from metadata.

glebuk:
Done per Senja

Issue #1515, opened 2018-11-02
Be able to create a reader by just specifying an schema class instead of all the columns


CESARDELATORRE:
In current version (v0.6 and v0.7) when creating a TextLoader we need to provide all the columns explicetely like the following:

_loader = mlContext.Data.TextReader(new TextLoader.Arguments()
                                                {
                                                    Separator = ",",
                                                    HasHeader = true,
                                                    Column = new[]
                                                                {
                                                                new TextLoader.Column("Season", DataKind.R4, 2),
                                                                new TextLoader.Column("Year", DataKind.R4, 3),
                                                                new TextLoader.Column("Month", DataKind.R4, 4),
                                                                new TextLoader.Column("Hour", DataKind.R4, 5),
                                                                new TextLoader.Column("Holiday", DataKind.R4, 6),
                                                                new TextLoader.Column("Weekday", DataKind.R4, 7),
                                                                new TextLoader.Column("WorkingDay", DataKind.R4, 8),
                                                                new TextLoader.Column("Weather", DataKind.R4, 9),
                                                                new TextLoader.Column("Temperature", DataKind.R4, 10),
                                                                new TextLoader.Column("NormalizedTemperature", DataKind.R4, 11),
                                                                new TextLoader.Column("Humidity", DataKind.R4, 12),
                                                                new TextLoader.Column("Windspeed", DataKind.R4, 13),
                                                                new TextLoader.Column("Count", DataKind.R4, 16)
                                                                }
                                                }

Since you usually also have a schema class for the observations, like the following:

    public class DemandObservation
    {
        public float Season;
        public float Year;
        public float Month;
        public float Hour;
        public float Holiday;
        public float Weekday;
        public float WorkingDay;
        public float Weather;
        public float Temperature;
        public float NormalizedTemperature;
        public float Humidity;
        public float Windspeed;
        public float Count;   // This is the observed count, to be used a "label" to predict
    }

It would be very convenient to be able to create a TextReader by just providing the class, like this:

mlContext.Data.TextReader(DemandObservation);

Zruty0:
Unfortunately, such a simple interface will not work, because the order of fields in the class is not defined, so we cannot expect `Season` to be column number 0, and `Month` to be number 2. We would still need some attribute to tell TextLoader where to read the column from:

```csharp
public class DemandObservation
{
    [LoadColumn(0)]   
    public float Season;
    [LoadColumn(1)]   
    public float Year;
    [LoadColumn(2)]   
[...]

PR #1878, opened 2018-12-14
Schema based text loader

sfilipi:
Initial work to address #561. 

It moves the current reflection-based logic outside of the legacy project, into the TextLoader catalog, creating an API for it. 
Decouples the LoadColumn and Column attributes.


test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -582,10 +582,10 @@ private int GetMaxIndexForOnePrediction(MNISTPrediction onePrediction)
 
         public class MNISTData
         {
-            [Column("0")]
+            [LoadColumn("0")]
             public long Label;
 
-            [Column(ordinal: "1-784")]
+            [LoadColumn(range: "1-784")]
```
Ivanidzo4ka: >LoadColumn [](start = 13, length = 10)

I think we need some good word smith here.
It's probably slightly better than just column, but still confusing, especially next to that ColumnName attribute few lines below.

Can it be TextLoader.MarkingColumn? TextLoader.MappingColumn, something related to textloader since it's used only for Textloader (right?)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -31,9 +35,109 @@ public static class TextLoaderSaverCatalog
...
+        /// <param name="supportSparse"></param>
+        /// <param name="trimWhitespace"></param>
+        public static TextLoader CreateTextReader<TInput>(this DataOperations catalog,
+            bool hasHeader = TextLoader.DefaultArguments.HasHeader,
+            char separator = TextLoader.DefaultArguments.Separator,
+            bool allowQuotedStrings = TextLoader.DefaultArguments.AllowQuoting,
+            bool supportSparse = TextLoader.DefaultArguments.AllowSparse,
+            bool trimWhitespace = TextLoader.DefaultArguments.TrimWhitespace)
+        {
+            var userType = typeof(TInput);
```
wschin: Can you describe what this function is doing? I'd like to see something `ComputeOutputSchema` in this [PR](https://github.com/dotnet/machinelearning/pull/1879/files) where high-level description is added to header and low-level descriptions are inline to explain high-level concept in detail?

sfilipi:
Reading the description of both #561 and #1515, i prototyped around having more attribute parameters for the LoadColumn attribute, to get closed to the TextLoader.Range. 
In the process, discovered that the members for the Attribute derived class cannot be of the more advanced types: like classes, nullable primitives (ints), tuples. 

I tried to eliminate the need for the existing Ordinal string by adding a` start/end string` (can't do int? ), `int[] columns` and `int[]inverseColumns` thinking they would suffice to more strongly type and support most of the  scenarios. And with them we can do the 
1-*  - auto end
1-4 - defined ranges 
1, 2, 3 - defined columns, or inverse columns

but the scenario that we can't do is multiple ranges. i.e. 1-4, 5-6

One idea on how to convert that would be to provide an array where the users  would specify the indexes with the convention that they should be treated in groups; example: new [] { 1, 3, 5, 7} would translate to 1-3, 5-7 but that seems UGLY. 

Another idea is to let the user specify multiple attributes for the range, and have the column name specified separately. Example:

```
 [LoadColumn(start:"0", end: "8")]
 [LoadColumn(start:"10", end: "18")]
 [ColumnName( "Features")]
 public float SepalLength;
```

But, IMHO, the range regex beats this too from the prospective of simplicity and ease of use. 

**So to summarize, i propose**:
1-  We keep the old ordinal, as string, rename to range
2-  We add a boolean `LoadInverseRange` to determine whether we add the columns defined in range or they should be the one not loaded (AllOther in TextLoader.Range)  

The construction of the TextLoader columns from the attributes is not final. I will complete after getting a final word here. 

@eerhardt, @TomFinley, @Zruty0, @CESARDELATORRE

sfilipi:
> Reading the description of both #561 and #1515, i prototyped around having more attribute parameters for the LoadColumn attribute, to get closed to the TextLoader.Range.
> In the process, discovered that the members for the Attribute derived class cannot be of the more advanced types: like classes, nullable primitives (ints), tuples.
> 
> I tried to eliminate the need for the existing Ordinal string by adding a` start/end string` (can't do int? ), `int[] columns` and `int[]inverseColumns` thinking they would suffice to more strongly type and support most of the scenarios. And with them we can do the
> 1-* - auto end
> 1-4 - defined ranges
> 1, 2, 3 - defined columns, or inverse columns
> 
> but the scenario that we can't do is multiple ranges. i.e. 1-4, 5-6
> 
> One idea on how to convert that would be to provide an array where the users would specify the indexes with the convention that they should be treated in groups; example: new [] { 1, 3, 5, 7} would translate to 1-3, 5-7 but that seems UGLY.
> 
> Another idea is to let the user specify multiple attributes for the range, and have the column name specified separately. Example:
> 
> ```
>  [LoadColumn(start:"0", end: "8")]
>  [LoadColumn(start:"10", end: "18")]
>  [ColumnName( "Features")]
>  public float SepalLength;
> ```
> But, IMHO, the range regex beats this too from the prospective of simplicity and ease of use.
> 
> **So to summarize, i propose**:
> 1- We keep the old ordinal, as string, rename to range
> 2- We add a boolean `LoadInverseRange` to determine whether we add the columns defined in range or they should be the one not loaded (AllOther in TextLoader.Range)
> 
> The construction of the TextLoader columns from the attributes is not final. I will complete after getting a final word here.
> 
> @eerhardt, @TomFinley, @Zruty0, @CESARDELATORRE

Decision is to break down the range/ordinal. If more granularity is needed, the users can use the text loader columns. #RESOLVED

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -68,30 +68,110 @@ public VectorTypeAttribute(params int[] dims)
     /// column encapsulates.
     /// </summary>
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property, AllowMultiple = false, Inherited = true)]
-    public sealed class ColumnAttribute : Attribute
+    public sealed class LoadColumnAttribute : Attribute
     {
-        public ColumnAttribute(string ordinal, string name = null)
+
+        public LoadColumnAttribute(int ordinal, string name = null, bool loadAllOthers = false)
```
sfilipi: >public [](start = 8, length = 6)

XMl docs on all ctors and samples for all ctors

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
sfilipi: It is called LoadColumn, if you are referring to the new attribute. 

---
In reply to: [241852931](https://github.com/dotnet/machinelearning/pull/1878#discussion_r241852931) [](ancestors = 241852931)

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -82,16 +93,47 @@ public ColumnAttribute(string ordinal, string name = null)
...
+        /// <summary>
+        /// If this is set to true, the columns defined in the range will be
+        /// excluded from loading, and all the other ones will be the ones loaded.
+        /// </summary>
+        public bool LoadInverseRange { get; }
+
+        public string Start { get; }
+        public string End { get; }
+        public int[] ColumnsIndexes { get; }
+        public int[] InverseColumnsIndexes { get; }
```
sfilipi: not needed, remove

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -68,30 +68,95 @@ public VectorTypeAttribute(params int[] dims)
     /// column encapsulates.
     /// </summary>
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property, AllowMultiple = false, Inherited = true)]
-    public sealed class ColumnAttribute : Attribute
+    public sealed class LoadColumnAttribute : Attribute
```
Zruty0: >LoadColumnAttribute [](start = 24, length = 19)

This attribute should go next to `TextLoader`, not as part of this class.

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -68,30 +68,95 @@ public VectorTypeAttribute(params int[] dims)
...
     {
-        public ColumnAttribute(string ordinal, string name = null)
+        /// <summary>
+        /// Initializes new instance of <see cref="LoadColumnAttribute"/>.
+        /// </summary>
+        /// <param name="ordinal">The index of the column in the text file.</param>
+        /// <param name="name">The optional name of the column, if it should be different from the name of the field or property where the attribute is positioned.</param>
+        /// <param name="loadAllOthers">Wheather to load all columns, besides the one specified in the ordinal.</param>
+        public LoadColumnAttribute(int ordinal, string name = null, bool loadAllOthers = false)
```
Zruty0: >ordinal [](start = 39, length = 7)

how about `index`, or `columnIndex`? The word `ordinal` does sort-of mean the same thing, but it's the only place in the codebase that we use the word

Zruty0: >name [](start = 55, length = 4)

can we remove `name`? This duplicates the functionality of `ColumnName` for no good reason

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -68,30 +68,95 @@ public VectorTypeAttribute(params int[] dims)
...
+        }
+
+        /// <summary>
+        /// Initializes new instance of <see cref="LoadColumnAttribute"/>.
+        /// </summary>
+        /// <param name="start">The starting column index, for the range.</param>
+        /// <param name="end">The ending column index, for the range.</param>
+        /// <param name="name">The optional name of the column, if it should be different from the name of the field or property where the attribute is positioned.</param>
+        /// <param name="columnIndexes">Distinct text file column indces to load as part of this column.</param>
+        public LoadColumnAttribute(string start, string end = null, string name = null, int[] columnIndexes = null)
```
Zruty0: >string [](start = 35, length = 6)

start and end should be integers here.

Zruty0: >columnIndexes [](start = 94, length = 13)

I think this parameter is more misleading than it is helpful. Can you remove it?

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -68,30 +68,95 @@ public VectorTypeAttribute(params int[] dims)
...
+                    Sources.Add(new TextLoader.Range(col));
+            }
+        }
+
+        /// <summary>
+        /// Initializes new instance of <see cref="LoadColumnAttribute"/>.
+        /// </summary>
+        /// <param name="columnIndexes">Distinct text file column indces to load as part of this column.</param>
+        /// <param name="name">The optional name of the column, if it should be different from the name of the field or property where the attribute is positioned.</param>
+        public LoadColumnAttribute(int[] columnIndexes, string name = null)
```
Zruty0: >columnIndexes [](start = 41, length = 13)

`columnIndices`

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -68,30 +68,95 @@ public VectorTypeAttribute(params int[] dims)
...
+            : this(columnIndexes[0].ToString(), name)
         {
+            Sources = new List<TextLoader.Range>();
+            foreach (var col in columnIndexes)
+                Sources.Add(new TextLoader.Range(col));
+        }
+
+#pragma warning disable 618
```
Zruty0: >pragma [](start = 1, length = 6)

which warning are we disabling?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -16,35 +20,126 @@ public static class TextLoaderSaverCatalog
...
+
+            var columns = new TextLoader.Column[memberInfos.Length];
+
+            for (int index = 0; index < memberInfos.Length; index++)
+            {
+                var memberInfo = memberInfos[index];
+                var mappingAttr = memberInfo.GetCustomAttribute<LoadColumnAttribute>();
+                var mptr = memberInfo.GetCustomAttributes();
+
+                Contracts.Assert(mappingAttr != null, $"Field or property {memberInfo.Name} is missing the LoadColumn attribute");
```
Zruty0: >LoadColumn [](start = 107, length = 10)

maybe `nameof` ?

Zruty0: >Assert [](start = 26, length = 6)

it should be `Check` if the user can hit it

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -16,35 +20,126 @@ public static class TextLoaderSaverCatalog
...
+        /// <param name="allowQuotedStrings">Whether the input may include quoted values,
+        /// which can contain separator characters, colons,
+        /// and distinguish empty values from missing values. When true, consecutive separators
+        /// denote a missing value and an empty value is denoted by \"\".
+        /// When false, consecutive separators denote an empty value.</param>
+        /// <param name="supportSparse">Whether the input may include sparse representations for example,
+        /// if one of the row contains "5 2:6 4:3" that's mean there are 5 columns all zero
+        /// except for 3rd and 5th columns which have values 6 and 3</param>
+        /// <param name="trimWhitespace">Remove trailing whitespace from lines</param>
+        public static TextLoader CreateTextReader<TInput>(this DataOperations catalog,
```
Zruty0: >CreateTextReader [](start = 33, length = 16)

This entire code should move out of the catalog and into a static method of `TextLoader`. The catalogs in general should not contain any meaningful code, they are there to expose the existing public interface in a consistent 'folder-like' structure.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -16,35 +20,126 @@ public static class TextLoaderSaverCatalog
...
         /// Read a data view from a text file using <see cref="TextLoader"/>.
         /// </summary>
-        /// <param name="catalog">The catalog.</param>
+        /// <param name="catalog">The <see cref="DataOperations"/> catalog.</param>
         /// <param name="columns">The columns of the schema.</param>
         /// <param name="hasHeader">Whether the file has a header.</param>
         /// <param name="separatorChar">The character used as separator between data points in a row. By default the tab character is used as separator.</param>
         /// <param name="path">The path to the file.</param>
         /// <returns>The data view.</returns>
         public static IDataView ReadFromTextFile(this DataOperations catalog,
```
Zruty0: >ReadFromTextFile [](start = 32, length = 16)

we should also add a templated version of `ReadFromTextFile`

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
sfilipi: can't have nullable ints. we can't leave 'end' as Auto, if i flip to ints. that ok?

---
In reply to: [242669241](https://github.com/dotnet/machinelearning/pull/1878#discussion_r242669241) [](ancestors = 242669241)

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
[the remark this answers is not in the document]
sfilipi: Do you think this will be useful when the user wants to load all features into one column, and the label in another. 

we'd spare them having to look up their cols cardinality, and a concat. 

---
In reply to: [242673339](https://github.com/dotnet/machinelearning/pull/1878#discussion_r242673339) [](ancestors = 242673339)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
sfilipi: Using this comment to ask a meta quesiton: does it make more sense to continue that to throw? Maybe not all the properties of the data model will be decorated with LoadColumn. @Zruty0, @tomfinly

---
In reply to: [242670646](https://github.com/dotnet/machinelearning/pull/1878#discussion_r242670646) [](ancestors = 242670646)

docs/code/MlNetCookBook.md
```diff
@@ -131,6 +131,39 @@ var reader = mlContext.Data.CreateTextReader(new[] {
...
+	[LoadColumn(1)]
+	public string Workclass { get; set; }
+
+	[LoadColumn(2)]
+	public string Education { get; set; }
+
+	[LoadColumn(3)]
+	public string MaritalStatus { get; set; }
+
+	public string[] AllFeatures { get; set; }
```
Zruty0: >AllFeatures [](start = 17, length = 11)

will this work? I am suspecting that reading this will throw

docs/code/MlNetCookBook.md
```diff
@@ -131,6 +131,39 @@ var reader = mlContext.Data.CreateTextReader(new[] {
...
+
+// Create a new context for ML.NET operations. It can be used for exception tracking and logging, 
+// as a catalog of available operations and as the source of randomness.
+var mlContext = new MLContext();
+
+// Read the data into a data view.
+var data = mlContext.Data.ReadFromTextFile<InspectedRow>(dataPath,
+	// First line of the file is a header, not a data row.
+	hasHeader: true
+)
```
Zruty0: >	 [](start = 1, length = 1)

please avoid tabs in this file

docs/code/MlNetCookBook.md
sfilipi: it does work because i changed the logic. See my questions about whether we should process only the annotated members/field, to make the data models more usable. 

---
In reply to: [243104636](https://github.com/dotnet/machinelearning/pull/1878#discussion_r243104636) [](ancestors = 243104636)

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
Zruty0: If we want to preserve the '10-*' logic, I would go with a different attribute for it (`[LoadAllColumns(start: 10)]` or something like that).

---
In reply to: [242741038](https://github.com/dotnet/machinelearning/pull/1878#discussion_r242741038) [](ancestors = 242741038,242669241)

Zruty0: I don't think the '*' is going to be as useful in API as it is in cmdline.

---
In reply to: [243120230](https://github.com/dotnet/machinelearning/pull/1878#discussion_r243120230) [](ancestors = 243120230,242741038,242669241)

src/Microsoft.ML.Data/DataLoadSave/Text/LoadColumnAttribute.cs
```diff
@@ -0,0 +1,65 @@
...
+            Sources.Add(new TextLoader.Range(columnIndex));
+        }
+
+        /// <summary>
+        /// Initializes new instance of <see cref="LoadColumnAttribute"/>.
+        /// </summary>
+        /// <param name="start">The starting column index, for the range.</param>
+        /// <param name="end">The ending column index, for the range.</param>
+        public LoadColumnAttribute(int start, int end)
+             : this(start.ToString()) //REVIEW  this is incorrect, but it is just temporary there, until the Legacy API's TextLoader gets deleted.
```
Zruty0: >//REVIEW  this is incorrect, but it is just temporary there, until the Legacy API's TextLoader gets deleted. [](start = 38, length = 108)

`// REVIEW: `
Also place it somewhere at the beginning of the line

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1352,6 +1362,76 @@ public void Save(ModelSaveContext ctx)
...
+
+            var columns = new List<Column>();
+
+            for (int index = 0; index < memberInfos.Length; index++)
+            {
+                var memberInfo = memberInfos[index];
+                var mappingAttr = memberInfo.GetCustomAttribute<LoadColumnAttribute>();
+
+                if(mappingAttr == null)
+                    continue;
```
Zruty0: hmm, no longer throwing? I'm not sure I'm a fan of it.

test/Microsoft.ML.Tests/Scenarios/Api/Estimators/TrainWithInitialPredictor.cs
```diff
@@ -22,7 +22,7 @@ public void New_TrainWithInitialPredictor()
 
             var ml = new MLContext(seed: 1, conc: 1);
 
-            var data = ml.Data.CreateTextReader(TestDatasets.Sentiment.GetLoaderColumns(), hasHeader: true).Read(GetDataPath(TestDatasets.Sentiment.trainFilename));
```
Zruty0: >GetLoaderColumns [](start = 71, length = 16)

do we still use `GetLoaderColumns()` anywhere? Maybe we can delete the method?

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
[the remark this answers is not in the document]
Zruty0: still not enough reason, in my opinion

---
In reply to: [242741558](https://github.com/dotnet/machinelearning/pull/1878#discussion_r242741558) [](ancestors = 242741558,242673339)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
sfilipi: @tomfinley @glebuk what do you think?

---
In reply to: [243120594](https://github.com/dotnet/machinelearning/pull/1878#discussion_r243120594) [](ancestors = 243120594)

test/Microsoft.ML.Tests/Scenarios/Api/Estimators/TrainWithInitialPredictor.cs
sfilipi: we do use it in a few tests. 

---
In reply to: [243121856](https://github.com/dotnet/machinelearning/pull/1878#discussion_r243121856) [](ancestors = 243121856)

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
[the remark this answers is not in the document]
sfilipi: removed

---
In reply to: [243121936](https://github.com/dotnet/machinelearning/pull/1878#discussion_r243121936) [](ancestors = 243121936,242741558,242673339)

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
sfilipi: remove

---
In reply to: [243120291](https://github.com/dotnet/machinelearning/pull/1878#discussion_r243120291) [](ancestors = 243120291,243120230,242741038,242669241)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
sfilipi: Resolving the questions, since the discussion has moved to the other comment. 

---
In reply to: [243040913](https://github.com/dotnet/machinelearning/pull/1878#discussion_r243040913) [](ancestors = 243040913,242670646)

src/Microsoft.ML.Data/DataLoadSave/Text/LoadColumnAttribute.cs
```diff
@@ -0,0 +1,71 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.ML.Runtime.Data;
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.ML.Data
+{
+#pragma warning disable 618
+    /// <summary>
```
singlis: Why is the warning disable needed? I would document why its disabled.

src/Microsoft.ML.Data/DataLoadSave/Text/LoadColumnAttribute.cs
```diff
@@ -0,0 +1,71 @@
...
+             : this(start.ToString())
+        {
+            Sources.Add(new TextLoader.Range(start, end));
+        }
+
+        /// <summary>
+        /// Initializes new instance of <see cref="LoadColumnAttribute"/>.
+        /// </summary>
+        /// <param name="columnIndexes">Distinct text file column indices to load as part of this column.</param>
+        // REVIEW: Calling the private constructor with just the columnIndexes[0] parameter, is incorrect,
```
singlis: Can all the "REVIEW" items be tagged with a github issue? Just thinking that maybe easier to track when the Legacy  API's TextLoader is deleted.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -343,28 +344,28 @@ public class ArgumentsCore
...
+            public bool AllowSparse = DefaultArguments.AllowSparse;
 
             [Argument(ArgumentType.AtMostOnce,
                 HelpText = "Number of source columns in the text data. Default is that sparse rows contain their size information.",
                 ShortName = "size")]
             public int? InputSize;
 
             [Argument(ArgumentType.AtMostOnce, Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly, HelpText = "Source column separator. Options: tab, space, comma, single character", ShortName = "sep")]
-            public string Separator = "tab";
+            public string Separator = "tab"; //DefaultArguments.Separator
```
singlis: >DefaultArguments [](start = 47, length = 16)

Is this comment needed?

singlis: or maybe just two default arguments Separator and SeparatorChars? Two different params doesnt guarantee things to be in sync. Im assuming since these are argruments that they can be set by the user -- so what happens if a user specifies Separator to be space, but doesnt set the separator chars?

---
In reply to: [243411408](https://github.com/dotnet/machinelearning/pull/1878#discussion_r243411408) [](ancestors = 243411408)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
singlis: I agree with you Senja, is there some benefit with throwing?

---
In reply to: [243358643](https://github.com/dotnet/machinelearning/pull/1878#discussion_r243358643) [](ancestors = 243358643,243120594)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -16,35 +14,66 @@ public static class TextLoaderSaverCatalog
...
+        /// <param name="catalog">The <see cref="DataOperations"/> catalog.</param>
         /// <param name="columns">The columns of the schema.</param>
         /// <param name="hasHeader">Whether the file has a header.</param>
         /// <param name="separatorChar">The character used as separator between data points in a row. By default the tab character is used as separator.</param>
         /// <param name="dataSample">The optional location of a data sample.</param>
         public static TextLoader CreateTextReader(this DataOperations catalog,
-            Column[] columns, bool hasHeader = false, char separatorChar = '\t', IMultiStreamSource dataSample = null)
+            TextLoader.Column[] columns,
+            bool hasHeader = false,
+            char separatorChar = '\t',
```
singlis: separatorChar = TextLoader.DefaultArguments.Separator?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
TomFinley: > If we throw, we restrict usage by expecting the class to be dedicated to just loading the data, whereas if we don't the type can be used for other thing, and the decorated fields can be used for loading.

Interesting @sfilipi. So, it strikes me that there are two possible dangers we must guard against.

1. A user attempts to use a class to define a schema, *without* proper attribution by mistake. We know this happens a lot.
2. A user attempts to use a class to define a schema, *with* proper attribution, but wants to have some additional "extra" properties free for no particular reason.

Like I said I know 1 comes up, and throwing has helped people to guard against misuse. 2 has been the state for a while now. In fact, if we consider all prior usage of the API (where we follow this pattern), and all prior usage of the text loader, and I think of all the questions that we've gotten *about* them over the years, both outside of Microsoft in ML.NET and inside Microsoft with our internal tool, I don't recall that ever happening. (Have we received requests for this? I am finding it difficult to imagine a concrete scenario.)

Even if the scenario **were** somehow to come up, I would prefer to solve it by introducing another attribute to indicate that this property is not to be considered by the loader, rather than to have this behavior. So, something like, `[NoLoader]` or whatnot. (I don't insist on the name. I also don't want us to add it now, until there's a concrete request for it, since my suspicion is that it will never come up.)

For this reason, I disagree @sfilipi and @singlis and agree with @Zruty0. It must throw.

Issue #1501, opened 2018-11-01
ISchema removal


Zruty0:
Remove `ISchema` interface altogether. Remove all obsolete methods of `Schema` class.

This should really be done after #1500

Zruty0:
@wschin Awesome work! Finally the crusade against the old-school metadata is complete!

PR #1879, opened 2018-12-14
Bindings in ChooseColumnsByIndexTransform not ISchema

wschin:
This PR is a part of #1501. We refactorize the `Binding` in `ChooseColumnsByIndexTransform` by making it not an `ISchema` but still maintaining necessary functionalities for connecting input and output. For the functionalities remained, please see the non-private member functions of `Bindings`. Some comments are added for a better readability.


src/Microsoft.ML.Data/Dirty/ChooseColumnsByIndexTransform.cs
```diff
@@ -191,9 +139,9 @@ private static VersionInfo GetVersionInfo()
         {
             return new VersionInfo(
                 modelSignature: "CHSCOLIF",
-                verWrittenCur: 0x00010001, // Initial
-                verReadableCur: 0x00010001,
-                verWeCanReadBack: 0x00010001,
+                verWrittenCur: 0x00010002,
```
Ivanidzo4ka: >0x00010002 [](start = 31, length = 10)

I would say, if you can preserve old saving mechanism, try to preserve it.
Do you have any particular reason to switch to new format?

wschin: This format has been changed previously. You can see that the old doc string describing doesn't match what previously stored.

---
In reply to: [241838591](https://github.com/dotnet/machinelearning/pull/1879#discussion_r241838591) [](ancestors = 241838591)

Ivanidzo4ka: I've check internal repo, and format is same as in code you changing.

---
In reply to: [241843452](https://github.com/dotnet/machinelearning/pull/1879#discussion_r241843452) [](ancestors = 241843452,241838591)

src/Microsoft.ML.Data/Dirty/ChooseColumnsByIndexTransform.cs
```diff
@@ -31,158 +30,115 @@ public sealed class Arguments
...
+                Contracts.AssertValue(sourceSchema);
 
                 // *** Binary format ***
-                // bool(as byte): whether the indicated source columns are columns to keep, or drop
-                // int: number of source column indices
-                // int[]: source column indices
+                // bool (as byte): operation mode
+                // int[]: selected source column indices
+                _drop = ctx.Reader.ReadBoolByte();
+                _sources = ctx.Reader.ReadIntArray();
```
Ivanidzo4ka: This look suspicious.

So imagine I create transform with params [0,2,3] and drop:false
And we do it for two Schemas, one with 4 columns, one with 5.

result for both of them would be schema with 3 columns [0,2,3].

now let's work with [0,2,3] and drop:true

I pass schema with 4 columns to this transform and save it.
we will save [1] in source.

Now I load this transform and pass schema with 5 columns.
result would be same schema with [1] column.

But if I start on 5 columns schema I will have [1,4] and if I pass that saved transform schema with 4 columns, i'm screwed.


So please, preserve old logic and respect drop parameter.

Ivanidzo4ka: Also would be nice if you throw some test coverage for this transform.

```
  [TestMethod, TestCategory("DataPipeSerialization")]
        public void SavePipeChooseColumnsByIndex()
        {
            TestCore(null, true,
                new[] {
                    "loader=Text{",
                    "  col=Label:U1[0-1]:0",
                    "  col=Features:U2:1-*",
                    "  col=A:U1[1-5]:1",
                    "  col=B:U1[3-8]:2",
                    "}",
                    "xf=ChooseColumnsByIndex{ind=2 ind=0}"
                });

            Done();
        }

        [TestMethod, TestCategory("DataPipeSerialization")]
        public void SavePipeChooseColumnsByIndexDrop()
        {
            TestCore(null, true,
                new[] {
                    "loader=Text{",
                    "  col=Label:U1[0-1]:0",
                    "  col=Features:U2:1-*",
                    "  col=A:U1[1-5]:1",
                    "  col=B:U1[3-8]:2",
                    "}",
                    "xf=ChooseColumnsByIndex{ind=3 ind=0 drop}"
                });

            Done();
        }

```

---
In reply to: [241891989](https://github.com/dotnet/machinelearning/pull/1879#discussion_r241891989) [](ancestors = 241891989)

wschin: What is `true`? There is no `TestCore` with prototype `TestCore(string, bool, ...)`.

---
In reply to: [241892330](https://github.com/dotnet/machinelearning/pull/1879#discussion_r241892330) [](ancestors = 241892330,241891989)

wschin: Tests are added and old logic is back.

---
In reply to: [241903270](https://github.com/dotnet/machinelearning/pull/1879#discussion_r241903270) [](ancestors = 241903270,241892330,241891989)

PR #1847, opened 2018-12-07
Update of FeatureContributionCalculation to new API

artidoro:
Fixes #1791.

As this is a WIP PR, I am still completing the work, but I would really appreciate your feedback on what I have done so far.

In this PR:
- [x]  I created a new ITransformer for FeatureContributionCalculation (previously known as WhatTheFeature) by converting previous ISchemaBoundRowMapper to two separate classes: a IRowMapper and a ISchemaBoundMapper.
- [x] I created a new IEstimator that produces the transformer.
- [x] I added tests for the Transformer and Estimator. (Need to add a few more)
- [x] I added MlContext extensions.
- [x] I added documentation, and checks for arguments and types.

Will do in a separate PR:
- Add static extensions for the estimators.
- I allowed pipelines, and not just IPredictor, to be passed to the IEstimator.

PR #1887, opened 2018-12-16
Remove ISchema in FeatureContributionCalculationTransform

wschin:
As title. It's a part of #1501. This change is not difficult because the output schema only contains one column, which is per feature contribution.

### DIFF
--- a/src/Microsoft.ML.Data/Scorers/FeatureContributionCalculationTransform.cs
+++ b/src/Microsoft.ML.Data/Scorers/FeatureContributionCalculationTransform.cs
@@ -364,9 +364,9 @@ namespace Microsoft.ML.Runtime.Data
                 }
                 else
                 {
-                    _outputSchema = Schema.Create(new FeatureContributionSchema(_env, DefaultColumnNames.FeatureContributions,
-                        new VectorType(NumberType.R4, schema.Feature.Type as VectorType),
-                        InputSchema, InputRoleMappedSchema.Feature.Index));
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.Data/Scorers/FeatureContributionCalculationTransform.cs
```diff
@@ -420,77 +420,5 @@ public Row GetRow(Row input, Func<int, bool> active)
...
-                _ectx.CheckParam(0 <= col && col < ColumnCount, nameof(col));
-                return _types[col];
-            }
-
-            public IEnumerable<KeyValuePair<string, ColumnType>> GetMetadataTypes(int col)
-            {
-                _ectx.CheckParam(col == 0, nameof(col));
-                if (_hasSlotNames)
-                    yield return MetadataUtils.GetSlotNamesPair(_featureVectorSize);
-            }
```
Ivanidzo4ka: that stuff would propagate slotnames for new added column from feature column.
I don't think you do that in new output schema.

wschin: [This PR](https://github.com/dotnet/machinelearning/pull/1847/files) is using the same strategy, so there is no reason to maintain that behavior.

wschin:
Close because https://github.com/dotnet/machinelearning/pull/1847/files also removes ISchema.

PR #1892, opened 2018-12-17
Fix hash join transform.

### LINKED ISSUE #1891: Hash join fails with bad image exception.
Add this test:
```
        [Fact]
        public void EntryPointHashJoinCountTable()
        {
            TestEntryPointPipelineRoutine(GetDataPath("breast-cancer.txt"), "col=Text:TX:1-9 col=Label:0",
                new[]
                {
                    "Transforms.HashConverter",
                },
                new[]
                {
                    @"'Column': [
                      {
                        'Name': 'Temp',
                        'Src': 'Text',
                        'CustomSlotMap': '0,1;2,3,4,5'
                      }
                      ]",
                });
        }
```
you get following exception.
```
Message: System.InvalidOperationException : Splitter/consolidator worker encountered exception while consuming source data---- 
System.BadImageFormatException : An attempt was made to load a program with an incorrect format. (Exception from HRESULT: 0x8007000B)
```

Ivanidzo4ka:
Fixes #1891

### DIFF
--- a/src/Microsoft.ML.Transforms/HashJoiningTransform.cs
+++ b/src/Microsoft.ML.Transforms/HashJoiningTransform.cs
@@ -540,23 +540,23 @@ namespace Microsoft.ML.Transforms.Conversions
             var mask = (1U << _exes[iinfo].HashBits) - 1;
             var hashSeed = _exes[iinfo].HashSeed;
             bool ordered = _exes[iinfo].Ordered;
+            var denseSource = default(VBuffer<TSrc>);
             return
                 (ref VBuffer<uint> dst) =>
                 {
                     getSrc(ref src);
                     Host.Check(src.Length == expectedSrcLength);
                     var hashes = VBufferEditor.Create(ref dst, n);
-
+                    src.CopyToDense(ref denseSource);
                     for (int i = 0; i < n; i++)
                     {
                         uint hash = hashSeed;
-
                         foreach (var srcSlot in slotMap[i])
                         {
                             // REVIEW: some legacy code hashes 0 for srcSlot in ord- case, do we need to preserve this behavior?
                             if (ordered)
                                 hash = Hashing.MurmurRound(hash, (uint)srcSlot);
-                            hash = hashFunction(src.GetItemOrDefault(srcSlot), hash);
+                            hash = hashFunction(denseSource.GetItemOrDefault(srcSlot), hash);
                         }
 
                         hashes.Values[i] = (Hashing.MixHash(hash) & mask) + 1; // +1 to offset from zero, which has special meaning for KeyType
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Transforms/HashJoiningTransform.cs
```diff
@@ -541,28 +540,11 @@ private ValueGetter<VBuffer<uint>> ComposeGetterVecToVec<TSrc>(Row input, int ii
             var mask = (1U << _exes[iinfo].HashBits) - 1;
             var hashSeed = _exes[iinfo].HashSeed;
             bool ordered = _exes[iinfo].Ordered;
-            TSrc[] denseValues = null;
             return
                 (ref VBuffer<uint> dst) =>
                 {
                     getSrc(ref src);
                     Host.Check(src.Length == expectedSrcLength);
-                    ReadOnlySpan<TSrc> values;
```
Ivanidzo4ka: >ReadOnlySpan<TSrc> values; [](start = 20, length = 26)

I have no idea why, but as soon as I add ReadOnlySpan inside delegate, I start getting BadImage exception.
So I access elements directly instead of that strange copying mechanism.

Zruty0: ouch. Is this something .NET colleagues should look at? It kind of smells of a .NET issue to me

---
In reply to: [242317331](https://github.com/dotnet/machinelearning/pull/1892#discussion_r242317331) [](ancestors = 242317331)

test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -4037,6 +4037,33 @@ public void EntryPointSlidingWindow()
                       'L': '1'"
                 });
         }
+
+        [Fact]
+        public void EntryPointHashJoinCountTable()
```
sfilipi: >EntryPointHashJoinCountTable [](start = 20, length = 28)

does it have to be an entry point test?

test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -4037,6 +4037,33 @@ public void EntryPointSlidingWindow()
...
+                        'Src': 'Text'
+                      },
+                      {
+                        'Name': 'Temp2',
+                        'Src': 'Text',
+                        'CustomSlotMap': '0,1;2,3,4,5'
+                      }
+
+                      ]"
+                });
```
sfilipi: does it need to call  Done();

src/Microsoft.ML.Transforms/HashJoiningTransform.cs
Ivanidzo4ka: @danmosemsft do you know who can take a look on this?

Issue #1500, opened 2018-11-01
Final interface of Schema


Zruty0:
As we have agreed on the API reviews, we will make the following modifications to `Schema`:
- Make it an `ICollection<Schema.Column>`
- Expand `Column` class to contain 2 fields: `IsHidden` and `Index`.
- Remove all methods except those exposed by collections, `this` accessors and `GetColumnOrNull`.
- Make constructors internal. Create a `SchemaBuilder` (in a different namespace `Microsoft.ML.Schema`) that would be responsible to building the schema. 
  - This is to make it more obvious that columns are NOT reused by different schemas.
- Move `Schema.Metadata.Builder` into a separate class `MetadataBuilder` in namespace `Microsoft.ML.Schema`.
- Finally, probably rename to `DataSchema`

TomFinley:
Pursuant to [API review notes](https://github.com/dotnet/apireviews/pull/74/files#diff-8b410ccce45654bbddc866aafa3eec50R10), this `Column` structure should probably be a struct.

sfilipi:
The only part that remains in this, is cleaning up usages of TryGetColumnIndex; which is not user-facing anymore; and renaming Schema to DataSchema.

sfilipi:
All the work for this is done. 
The only thing left atm, the renaming, is covered in #2297.

PR #1893, opened 2018-12-17
Internalizing MetadataUtils

Zruty0:
Contributes to #1500 

Moved most of MetadataUtils to internal, except some public methods to access common metadata.
Removed some methods that are now trivial.

### DIFF
--- a/src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
+++ b/src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
@@ -2,18 +2,17 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using Float = System.Single;
-
-using System;
+using Microsoft.ML.Data;
 using Microsoft.ML.Runtime;
 using Microsoft.ML.Runtime.Data;
-using Microsoft.ML.Runtime.Model;
 using Microsoft.ML.Runtime.Internal.Utilities;
+using Microsoft.ML.Runtime.Model;
 using Microsoft.ML.Runtime.Model.Onnx;
-using Newtonsoft.Json.Linq;
 using Microsoft.ML.Runtime.Model.Pfa;
-using System.Collections.Generic;
-using Microsoft.ML.Data;
+using Newtonsoft.Json.Linq;
+using System;
+
+using Float = System.Single;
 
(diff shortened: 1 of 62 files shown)


Ivanidzo4ka:
>     public static class MetadataUtils

If you internalize MetadataUtils, why it's still public :) ?

---
Refers to: src/Microsoft.ML.Core/Data/MetadataUtils.cs:21 in 612075c. [](commit_id = 612075ca5143d917bdd47eef731999cd01da283e, deletion_comment = False)

src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
```diff
@@ -13,6 +13,7 @@
 using Newtonsoft.Json.Linq;
 using Microsoft.ML.Runtime.Model.Pfa;
 using System.Collections.Generic;
+using Microsoft.ML.Data;
```
Zruty0: >using [](start = 0, length = 5)

I'm sorting all the usings now, don't do the Latin counting again!

Ivanidzo4ka:
> 

Can you kill it and sort?

---
Refers to: src/Microsoft.ML.Data/Depricated/Instances/HeaderSchema.cs:8 in 612075c. [](commit_id = 612075ca5143d917bdd47eef731999cd01da283e, deletion_comment = False)

src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
Ivanidzo4ka: I can do Japanese! Ichi, ni, san, shi ...

---
In reply to: [242338122](https://github.com/dotnet/machinelearning/pull/1893#discussion_r242338122) [](ancestors = 242338122)

Zruty0:
Please look at MetadataUtils first: all the other files are automatically changed to continue compiling with the changes in MetadataUtils.

TomFinley:
>     public static class MetadataUtils

Parts of it. I suspect though that just having this list of string constants doesn't really do much of anyone any good -- if we want conveniences on top of `Schema.Metadata` that's great, but this by itself we might consider just an intermediate stage. But, as an intermeidate stage I think it's fine, we definitely need to change the namespace anyway...

---
In reply to: [448022694](https://github.com/dotnet/machinelearning/pull/1893#issuecomment-448022694) [](ancestors = 448022694,448020032)

---
Refers to: src/Microsoft.ML.Core/Data/MetadataUtils.cs:21 in 612075c. [](commit_id = 612075ca5143d917bdd47eef731999cd01da283e, deletion_comment = False)

PR #1894, opened 2018-12-17
Making Schema implement ISchema explicitly

Zruty0:
Addresses #1500 

Made Schema implement ISchema explicitly. Removed all calls to the ISchema inteface when Schema is accessed, with the exception of TryGetColumnIndex, which was made internal.

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/FeatureContributionCalculationTransform.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/FeatureContributionCalculationTransform.cs
@@ -85,7 +85,7 @@ namespace Microsoft.ML.Samples.Dynamic
                 var value = row.Features[featureOfInterest];
                 var contribution = row.FeatureContributions[featureOfInterest];
                 var percentContribution = 100 * contribution / row.Score;
-                var name = data.Schema.GetColumnName(featureOfInterest + 1);
+                var name = data.Schema[(int) (featureOfInterest + 1)].Name;
                 var weight = weights.GetValues()[featureOfInterest];
 
                 Console.WriteLine("{0:0.00}\t{1:0.00}\t{2}\t{3:0.00}\t{4:0.00}\t{5:0.00}\t{6:0.00}",
--- a/src/Microsoft.ML.Core/Data/Schema.cs
+++ b/src/Microsoft.ML.Core/Data/Schema.cs
@@ -300,6 +300,17 @@ namespace Microsoft.ML.Data
             return getter;
         }
 
+        /// <summary>
+        /// Legacy method to get the column index.
+        /// DO NOT USE: use <see cref="GetColumnOrNull"/> instead.
+        /// </summary>
+        [BestFriend]
+        internal bool TryGetColumnIndex(string name, out int col)
+        {
+            col = GetColumnOrNull(name)?.Index ?? -1;
+            return col >= 0;
+        }
+
         #region Legacy schema API to be removed
         /// <summary>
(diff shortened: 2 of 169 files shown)


docs/samples/Microsoft.ML.Samples/Dynamic/FeatureContributionCalculationTransform.cs
```diff
@@ -85,7 +85,7 @@ public static void FeatureContributionCalculationTransform_Regression()
                 var value = row.Features[featureOfInterest];
                 var contribution = row.FeatureContributions[featureOfInterest];
                 var percentContribution = 100 * contribution / row.Score;
-                var name = data.Schema.GetColumnName(featureOfInterest + 1);
+                var name = data.Schema[(int) (featureOfInterest + 1)].Name;
```
Ivanidzo4ka: >(int) [](start = 39, length = 5)

why cast to int if it's already int?
To catch overflow?

Zruty0: I have no idea, this is Resharper doing the work for me. I believe you are right, it's for the overflow.

---
In reply to: [242359826](https://github.com/dotnet/machinelearning/pull/1894#discussion_r242359826) [](ancestors = 242359826)

Zruty0:
You can start looking now. Again, biggest changes are in `Schema.cs`, and the remaining files have been updated to compile.

src/Microsoft.ML.Core/Data/Schema.cs
```diff
@@ -310,38 +300,37 @@ private static Delegate GetMetadataGetterDelegate<TValue>(ISchema schema, int co
             return getter;
         }
 
+        /// <summary>
+        /// Legacy method to get the column index.
+        /// DO NOT USE: use <see cref="GetColumnOrNull"/> instead.
```
TomFinley: >DO NOT USE [](start = 12, length = 10)

I'm guessing using actual `Obsolete` on this and other code is undesirable?

TomFinley:
>     public sealed class Schema : ISchema, IReadOnlyList<Schema.Column>

How close are we to making `ISchema` itself an internal interface, if not yet deleting it outright?

---
Refers to: src/Microsoft.ML.Core/Data/Schema.cs:25 in 5b17cc4. [](commit_id = 5b17cc42e776e4a0a512af3e89118ee10cf31d3e, deletion_comment = False)

Zruty0:
>     public sealed class Schema : ISchema, IReadOnlyList<Schema.Column>

I will know today. The 2 major blockers was Schema and ColumnBindingsBase.

---
In reply to: [448109260](https://github.com/dotnet/machinelearning/pull/1894#issuecomment-448109260) [](ancestors = 448109260)

---
Refers to: src/Microsoft.ML.Core/Data/Schema.cs:25 in 5b17cc4. [](commit_id = 5b17cc42e776e4a0a512af3e89118ee10cf31d3e, deletion_comment = False)

src/Microsoft.ML.Core/Data/Schema.cs
Zruty0: It's 450+ uses, and they cannot be all fixed mechanically with Resharper. So I opted to internal-best-friend it. The rest of the `ISchema` will be removed, but this one, I think, may stay for longer.

Actually obsolete-ing is going to be quite disruptive to the code base either...

---
In reply to: [242417023](https://github.com/dotnet/machinelearning/pull/1894#discussion_r242417023) [](ancestors = 242417023)

PR #1897, opened 2018-12-17
Fix copyright for Microsoft.ML.TensorFlow.Redist

abgoswam:
Deleted extra space in copyright


Ivanidzo4ka:
I think it deserve issue in form of 4 page essay, about forgiveness and misfortune of human race, and how small thing can lead to major achievements!

PR #1898, opened 2018-12-17
Remove Legacy dependency from Microsoft.ML.EntryPoints project

yaeldekel:
The TrainTest, CV and OVA macros use the pipeline API. This PR removes the dependency on pipeline API by creating the EntryPointNodes directly.


montebhoover:
>     public static class CrossValidationBinaryMacro

Should we delete `CrossValidationBinaryMacro` and `TrainTestBinaryMacro` since they aren't used?  Or is there a case where we want the ability to call the binary version of these that the normal version doesn't cover?

---
Refers to: src/Microsoft.ML.EntryPoints/CrossValidationBinaryMacro.cs:23 in eee4e6b. [](commit_id = eee4e6b61fdbeba1636e79a58f60fcf21776c930, deletion_comment = False)

src/Microsoft.ML.EntryPoints/TrainTestMacro.cs
```diff
@@ -84,10 +82,6 @@ public sealed class Output
                 "provided as the Input.TransformModel.", SortOrder = 1)]
             public PredictorModel PredictorModel;
 
-            [TlcModule.Output(Desc = "The final model including the trained predictor model and the model from the transforms, " +
-                "provided as the Input.TransformModel.", SortOrder = 2)]
-            public TransformModel TransformModel;
```
sfilipi: this is not needed anymore? I believe we added it because there was no other way to save the final transforms model by itself.

yaeldekel: It was only used in the legacy Pipeline API scenario, since the models created by the pipelines were TransformModels (they included the scorer), so they are not needed anymore.

---
In reply to: [242420400](https://github.com/dotnet/machinelearning/pull/1898#discussion_r242420400) [](ancestors = 242420400)

TomFinley: Very interesting! The sort of ridiculous web you had to weave to get this API to do anything with itself was quite something. 😄

PR #1902, opened 2018-12-18
Internalize RoleMappedSchema and implications thereof

TomFinley:
Internalization of everything related to role mapped schema, which includes schema-bindable mappers, much of the internal infrasturcture of evaluators, and so forth. As usual the commits are structured in such a way as only one "group" of code gets internalized at a time, with only the current last commit internalizing `RoleMappedSchema` itself.


src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs
```diff
@@ -324,7 +325,7 @@ private List<CoefficientStatistics> GetUnorderedCoefficientStatistics(LinearBina
         /// <summary>
         /// Gets the coefficient statistics as an object.
         /// </summary>
-        public CoefficientStatistics[] GetCoefficientStatistics(LinearBinaryModelParameters parent, RoleMappedSchema schema, int paramCountCap)
+        internal CoefficientStatistics[] GetCoefficientStatistics(LinearBinaryModelParameters parent, RoleMappedSchema schema, int paramCountCap)
```
sfilipi: >internal [](start = 8, length = 8)

can this stay public? There is no other way to get the coefficient stats.

sfilipi: we should have a workitem to have this method public, if this gets checked in as is, and address that in 0.9 

---
In reply to: [242656194](https://github.com/dotnet/machinelearning/pull/1902#discussion_r242656194) [](ancestors = 242656194)

TomFinley: We may need to build a public surface to it, but it can't be this specifically. We'll have to figure something else out.

PR #1905, opened 2018-12-18
Fix LDA bug that ignores per-column arguments

### LINKED ISSUE #1904: LDA transform ignores per-column arguments
The constructor that instantiates the transform from the Arguments object ignores the arguments defined per column and just uses the global values defined in the args.

yaeldekel:
Fixes #1904 .


src/Microsoft.ML.Transforms/Text/LdaTransform.cs
```diff
@@ -235,8 +235,17 @@ public sealed class ColumnInfo
 
             internal ColumnInfo(Column item, Arguments args) :
                 this(item.Source, item.Name,
```
Ivanidzo4ka: >item.Source [](start = 21, length = 11)

item.Source??item.Name

if you don't mind to change this one.

PR #1922, opened 2018-12-19
Onnxtransform - api changes for GPU support

### LINKED ISSUE #1834: OnnxTransform -- upgrade to support Linux, Mac and CUDA GPU
Add support for Linux X64
Add support for Mac X64
Add support for CUDA GPU for Windows and Linux (no Mac yet)

jignparm:
Fixes #1834 

Add CUDA 10.0 GPU execution support

Add Linux support

Add Mac support

### DIFF
--- a/build/Dependencies.props
+++ b/build/Dependencies.props
@@ -15,7 +15,7 @@
   <PropertyGroup>
     <GoogleProtobufPackageVersion>3.5.1</GoogleProtobufPackageVersion>
     <LightGBMPackageVersion>2.2.1.1</LightGBMPackageVersion>
-    <MicrosoftMLOnnxRuntimePackageVersion>0.1.5</MicrosoftMLOnnxRuntimePackageVersion>
+    <MicrosoftMLOnnxRuntimeGpuPackageVersion>0.1.5</MicrosoftMLOnnxRuntimeGpuPackageVersion>
     <MlNetMklDepsPackageVersion>0.0.0.7</MlNetMklDepsPackageVersion>
     <ParquetDotNetPackageVersion>2.1.3</ParquetDotNetPackageVersion>
     <SystemDrawingCommonPackageVersion>4.5.0</SystemDrawingCommonPackageVersion>
--- a/build/Dependencies.props
+++ b/build/Dependencies.props
@@ -15,7 +15,7 @@
   <PropertyGroup>
     <GoogleProtobufPackageVersion>3.5.1</GoogleProtobufPackageVersion>
     <LightGBMPackageVersion>2.2.1.1</LightGBMPackageVersion>
-    <MicrosoftMLOnnxRuntimeGpuPackageVersion>0.1.5</MicrosoftMLOnnxRuntimeGpuPackageVersion>
+    <MicrosoftMLOnnxRuntimeGpuPackageVersion>0.1.5-dev-11e91b61</MicrosoftMLOnnxRuntimeGpuPackageVersion>
     <MlNetMklDepsPackageVersion>0.0.0.7</MlNetMklDepsPackageVersion>
     <ParquetDotNetPackageVersion>2.1.3</ParquetDotNetPackageVersion>
     <SystemDrawingCommonPackageVersion>4.5.0</SystemDrawingCommonPackageVersion>
--- a/build/Dependencies.props
+++ b/build/Dependencies.props
@@ -15,7 +15,7 @@
   <PropertyGroup>
     <GoogleProtobufPackageVersion>3.5.1</GoogleProtobufPackageVersion>
     <LightGBMPackageVersion>2.2.1.1</LightGBMPackageVersion>
-    <MicrosoftMLOnnxRuntimeGpuPackageVersion>0.1.5-dev-11e91b61</MicrosoftMLOnnxRuntimeGpuPackageVersion>
+    <MicrosoftMLOnnxRuntimeGpuPackageVersion>0.1.5</MicrosoftMLOnnxRuntimeGpuPackageVersion>
     <MlNetMklDepsPackageVersion>0.0.0.7</MlNetMklDepsPackageVersion>
     <ParquetDotNetPackageVersion>2.1.3</ParquetDotNetPackageVersion>
     <SystemDrawingCommonPackageVersion>4.5.0</SystemDrawingCommonPackageVersion>
--- a/src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
+++ b/src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
@@ -48,6 +48,13 @@ namespace Microsoft.ML.Transforms
     /// <p>Supports inferencing of models in 1.2 and 1.3 format, using the
     /// <a href='https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime/'>Microsoft.ML.OnnxRuntime</a> library
     /// </p>
+    /// <p>Supports optional GPU execution via the CUDA 10.0. Requires
+    /// <a href='https://developer.nvidia.com/cuda-downloads'>CUDA 10.0 Toolkit</a>
+    /// and
+    /// <a href='https://developer.nvidia.com/cudnn'>cuDNN</a>
+    /// libraries need to be installed models can be scored in GPU. By default models are run on CPU. To execute on GPU,
+    /// set the parameter 'gpuDeviceID' to a valid non-negative number.
+    /// </p>
     /// <p>The inputs and outputs of the onnx models must of of Tensors. Sequence and Maps are not yet supported.</p>
     /// <p>Visit https://github.com/onnx/models to see a list of readily available models to get started with.</p>
     /// <p>Refer to http://onnx.ai' for more information about ONNX.</p>
@@ -64,6 +71,12 @@ namespace Microsoft.ML.Transforms
 
             [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "Name of the output column.", SortOrder = 2)]
             public string[] OutputColumns;
+
+            [Argument(ArgumentType.AtMostOnce | ArgumentType.Required, HelpText = "GPU device id to run on (e.g. 0,1,..). -1 runs on CPU. Requires CUDA 10.0.", SortOrder = 3)]
+            public int GpuDeviceId = -1;
+
+            [Argument(ArgumentType.AtMostOnce | ArgumentType.Required, HelpText = "If true, resumes execution on CPU upon GPU error. If false, will raise the GPU execption.", SortOrder = 4)]
+            public bool FallbackToCpu = false;
         }
 
         private readonly Arguments _args;
@@ -91,15 +104,27 @@ namespace Microsoft.ML.Transforms
             loaderAssemblyName: typeof(OnnxTransform).Assembly.FullName);
         }
 
-        public static IDataTransform Create(IHostEnvironment env, IDataView input, string modelFile)
+        public static IDataTransform Create(IHostEnvironment env, IDataView input, string modelFile, int gpuDeviceId = -1, bool fallbackToCpu = false)
         {
-            var args = new Arguments { ModelFile = modelFile, InputColumns = new string[] { }, OutputColumns = new string[] { } };
+            var args = new Arguments {
+                ModelFile = modelFile,
+                InputColumns = new string[] { },
+                OutputColumns = new string[] { },
+                GpuDeviceId = gpuDeviceId,
+                FallbackToCpu = fallbackToCpu };
+
             return Create(env, args, input);
         }
 
-        public static IDataTransform Create(IHostEnvironment env, IDataView input, string modelFile, string[] inputColumns, string[] outputColumns)
+        public static IDataTransform Create(IHostEnvironment env, IDataView input, string modelFile, string[] inputColumns, string[] outputColumns, int gpuDeviceId = -1, bool fallbackToCpu = false)
         {
-            var args = new Arguments { ModelFile = modelFile, InputColumns = inputColumns, OutputColumns = outputColumns };
+            var args = new Arguments {
+                ModelFile = modelFile,
+                InputColumns = inputColumns,
+                OutputColumns = outputColumns,
+                GpuDeviceId = gpuDeviceId,
+                FallbackToCpu = fallbackToCpu };
+
             return Create(env, args, input);
         }
 
@@ -163,10 +188,10 @@ namespace Microsoft.ML.Transforms
             {
                 Host.CheckNonWhiteSpace(args.ModelFile, nameof(args.ModelFile));
                 Host.CheckUserArg(File.Exists(args.ModelFile), nameof(args.ModelFile));
-                Model = new OnnxModel(args.ModelFile);
+                Model = new OnnxModel(args.ModelFile, args.GpuDeviceId, args.FallbackToCpu);
             }
             else
-                Model = OnnxModel.CreateFromBytes(modelBytes);
+                Model = OnnxModel.CreateFromBytes(modelBytes, args.GpuDeviceId, args.FallbackToCpu);
 
             var modelInfo = Model.ModelInfo;
             Inputs = (args.InputColumns.Count() == 0 ) ? Model.InputNames.ToArray() : args.InputColumns;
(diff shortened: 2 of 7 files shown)


build/Dependencies.props
```diff
@@ -15,7 +15,7 @@
   <PropertyGroup>
     <GoogleProtobufPackageVersion>3.5.1</GoogleProtobufPackageVersion>
     <LightGBMPackageVersion>2.2.1.1</LightGBMPackageVersion>
-    <MicrosoftMLOnnxRuntimePackageVersion>0.1.5</MicrosoftMLOnnxRuntimePackageVersion>
+    <MicrosoftMLOnnxRuntimeGpuPackageVersion>0.1.5-dev-11e91b61</MicrosoftMLOnnxRuntimeGpuPackageVersion>
```
sfilipi: >>0.1.5-dev-11e91b61< [](start = 44, length = 20)

double-checking that it is desired to go with a dev version for 0.9

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
```diff
@@ -48,6 +48,13 @@ namespace Microsoft.ML.Transforms
     /// <p>Supports inferencing of models in 1.2 and 1.3 format, using the
     /// <a href='https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime/'>Microsoft.ML.OnnxRuntime</a> library
     /// </p>
+    /// <p>Supports optional GPU execution via the CUDA 10.0. Requires
+    /// <a href='https://developer.nvidia.com/cuda-downloads'>CUDA 10.0 Toolkit</a>
+    /// and
+    /// <a href='https://developer.nvidia.com/cudnn'>cuDNN</a>
+    /// libraries need to be installed models can be scored in GPU. By default models are run on CPU. To execute on GPU,
```
sfilipi: >need [](start = 18, length = 4)

remove, since the sentence starts with Requires.

sfilipi: >installed models can be scored in GPU [](start = 29, length = 37)

installed so models

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
```diff
@@ -48,6 +48,13 @@ namespace Microsoft.ML.Transforms
...
+    /// set the parameter 'gpuDeviceID' to a valid non-negative number.
```
sfilipi: >'gpuDeviceID' [](start = 26, length = 13)

<see cref ="OnnxTransform.Arguments.GpuDeviceId"/>

sfilipi: >valid non-negative number [](start = 45, length = 25)

what does valid-non negative mean? Are the GPUs available in the system referenced by some sort of load index?

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
```diff
@@ -188,17 +213,68 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Sch
         }
 
         public OnnxTransform(IHostEnvironment env, string modelFile)
-            : this(env, new Arguments() { ModelFile = modelFile, InputColumns = new string[] { }, OutputColumns = new string[] { } })
+            : this(env, modelFile, -1, false)
+        {
+
+        }
+
+        public OnnxTransform(IHostEnvironment env, string modelFile, int gpuDeviceId = -1, bool fallbackToCpu = false)
```
sfilipi: >OnnxTransform [](start = 15, length = 13)

xml docs on anything public.

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
```diff
@@ -91,15 +104,27 @@ private static VersionInfo GetVersionInfo()
...
+                InputColumns = new string[] { },
+                OutputColumns = new string[] { },
+                GpuDeviceId = gpuDeviceId,
+                FallbackToCpu = fallbackToCpu };
+
             return Create(env, args, input);
         }
 
-        public static IDataTransform Create(IHostEnvironment env, IDataView input, string modelFile, string[] inputColumns, string[] outputColumns)
+        public static IDataTransform Create(IHostEnvironment env, IDataView input, string modelFile, string[] inputColumns, string[] outputColumns, int gpuDeviceId = -1, bool fallbackToCpu = false)
```
sfilipi: >public [](start = 8, length = 6)

can all Create method be made non-public, preferably private?

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
```diff
@@ -188,17 +213,68 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Sch
...
+                ModelFile = modelFile,
+                InputColumns = new[] { inputColumn },
+                OutputColumns = new[] { outputColumn },
+                GpuDeviceId = gpuDeviceId,
+                FallbackToCpu = fallbackToCpu
+            })
         {
         }
 
         public OnnxTransform(IHostEnvironment env, string modelFile, string[] inputColumns, string[] outputColumns)
```
sfilipi: >  public OnnxTransform(IHostEnvironment env, string modelFile, string[] inputColumns, string[] outputColumns) [](start = 6, length = 109)

to me this ctor and the next one seems like the same thing,

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
```diff
@@ -188,17 +213,68 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Sch
...
+                ModelFile = modelFile,
+                InputColumns = new string[] {},
+                OutputColumns = new string[] {},
+                GpuDeviceId = gpuDeviceId,
+                FallbackToCpu = fallbackToCpu
+            })
         {
         }
 
         public OnnxTransform(IHostEnvironment env, string modelFile, string inputColumn, string outputColumn)
```
sfilipi: > public OnnxTransform(IHostEnvironment env, string modelFile, string inputColumn, string outputColumn) [](start = 7, length = 102)

i think the next ctor might be enough.

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
sfilipi: > public OnnxTransform(IHostEnvironment env, string modelFile, int gpuDeviceId = -1, bool fallbackToCpu = false) [](start = 7, length = 111)

maybe remove the previous ctor. Calling this without providing the optional params will fail with "ambiguous call" messages.

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
```diff
@@ -73,10 +73,44 @@ public OnnxNodeInfo(string name, OnnxShape shape, System.Type type)
...
+        {
+        }
+
+        /// <summary>
+        /// Constructs OnnxModel object from file.
+        /// </summary>
+        /// <param name="modelFile">Model file path</param>
+        /// <param name="gpuDeviceId">Gpu device ID to execute on. -1 for CPU</param>
+        /// <param name="fallbackToCpu">If true, resumes CPU execution quitely upon GPU error</param>
+        public OnnxModel(string modelFile, int gpuDeviceId = -1, bool fallbackToCpu = false)
```
sfilipi: >  public OnnxModel(string modelFile, int gpuDeviceId = -1, bool fallbackToCpu = false) [](start = 6, length = 86)

same comment about ctor redundance

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
```diff
@@ -73,10 +73,44 @@ public OnnxNodeInfo(string name, OnnxShape shape, System.Type type)
...
         {
             _modelFile = modelFile;
-            _session = new InferenceSession(modelFile);
+
+            if (gpuDeviceId >= 0)
+            {
+                try
+                {
+                    _session = new InferenceSession(modelFile, SessionOptions.MakeSessionOptionWithCudaProvider(gpuDeviceId));
+                }
```
sfilipi: is there a way to assert that this is a valid ID before calling on the method?

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
```diff
@@ -73,10 +73,44 @@ public OnnxNodeInfo(string name, OnnxShape shape, System.Type type)
...
+                catch (OnnxRuntimeException e)
+                {
+                    if (fallbackToCpu)
+                        _session = new InferenceSession(modelFile);
+                    else
+                        throw e;
```
sfilipi: > throw e; [](start = 23, length = 9)

should use the channel to throw a meaningful message

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
```diff
@@ -73,10 +73,44 @@ public OnnxNodeInfo(string name, OnnxShape shape, System.Type type)
...
+
+        /// <summary>
+        /// Constructs OnnxModel object from file.
+        /// </summary>
+        /// <param name="modelFile">Model file path</param>
+        /// <param name="gpuDeviceId">Gpu device ID to execute on. -1 for CPU</param>
+        /// <param name="fallbackToCpu">If true, resumes CPU execution quitely upon GPU error</param>
+        public OnnxModel(string modelFile, int gpuDeviceId = -1, bool fallbackToCpu = false)
         {
             _modelFile = modelFile;
```
singlis: What about checking the file path? Should this throw a FileNotFoundException if the file is not found?

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
singlis: +1, the default arguments should get the same result of OnnxModel(string modelFile).

---
In reply to: [243430419](https://github.com/dotnet/machinelearning/pull/1922#discussion_r243430419) [](ancestors = 243430419)

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
```diff
@@ -73,10 +73,44 @@ public OnnxNodeInfo(string name, OnnxShape shape, System.Type type)
...
-            _session = new InferenceSession(modelFile);
+
+            if (gpuDeviceId >= 0)
+            {
+                try
+                {
+                    _session = new InferenceSession(modelFile, SessionOptions.MakeSessionOptionWithCudaProvider(gpuDeviceId));
+                }
+                catch (OnnxRuntimeException e)
```
singlis: >OnnxRuntimeException [](start = 23, length = 20)

Are we guaranteed to always get an OnnxRuntimeException thrown when the InferenceSession fails?

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
```diff
@@ -48,6 +48,13 @@ namespace Microsoft.ML.Transforms
     /// <p>Supports inferencing of models in 1.2 and 1.3 format, using the
     /// <a href='https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime/'>Microsoft.ML.OnnxRuntime</a> library
```
Ivanidzo4ka: ><a href='https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime/'>Microsoft.ML.OnnxRuntime</a> [](start = 8, length = 95)

not anymore, right?

build/Dependencies.props
jignparm: This is not the final  package to use. The merge with master will contain the official onnxruntime package. This package is used for testing the CI build only.

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
jignparm: Removed redundant ctor.

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
jignparm: We don't need a separate file check. Onnxruntime checks for this already.

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
jignparm: If the ID is invalid, the user gets a fairly descriptive string error message saying "inordinal device ID".

The usage is straightforward... just specify an integer (0, 1, 2 etc.).  MLNet doesn't have a GPU build to put an example for this, otherwise we could introduce a conditional fact to showcase an example.

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
jignparm: Yes. There will always be an OnnxRuntimeException upon any failure.

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
jignparm: Updated with a comment -- this exception is caught  by OnnxTransform and rethrown as a Host.Except().

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
```diff
@@ -46,9 +46,15 @@ namespace Microsoft.ML.Transforms
...
+    /// <a href='https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime/'>Microsoft.ML.OnnxRuntime</a> library.
     /// </p>
-    /// <p>The inputs and outputs of the onnx models must of of Tensors. Sequence and Maps are not yet supported.</p>
+    /// <p>Models are scored on CPU by default. If GPU execution is needed (optional), install
+    /// <a href='https://developer.nvidia.com/cuda-downloads'>CUDA 10.0 Toolkit</a>
+    /// and
+    /// <a href='https://developer.nvidia.com/cudnn'>cuDNN</a>
+    /// , and set the parameter 'gpuDeviceId' to a valid non-negative integer. Typical device ID values are 0 or 1.
+    /// </p>
+    /// <p>The inputs and outputs of the ONNX models must of Tensor type. Sequence and Maps are not yet supported.</p>
```
sfilipi: >must of Tensor  [](start = 53, length = 15)

must be of Tensor

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
sfilipi: I believe the tendency is to use the constructors to instantiate, and leave the static Create methods for the reflection-based component catalog approach if you don't have a private constructor with just environment and arguments (matching the signature. )

Wondering if you even need the Create methods. 

---
In reply to: [243467291](https://github.com/dotnet/machinelearning/pull/1922#discussion_r243467291) [](ancestors = 243467291)

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
```diff
@@ -187,18 +218,68 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Sch
...
         }
 
-        public OnnxTransform(IHostEnvironment env, string modelFile)
-            : this(env, new Arguments() { ModelFile = modelFile, InputColumns = new string[] { }, OutputColumns = new string[] { } })
+        /// <summary>
+        /// Transform for scoring ONNX models. Input data column names/types must exactly match
+        /// all model input names. All possible output columns are generated, with names/types
+        /// specified by model.
+        /// </summary>
+        /// <param name="env"></param>
```
sfilipi: ><param name="env"></param> [](start = 12, length = 26)

add text: the environment to use

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
```diff
@@ -187,18 +218,68 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Sch
...
+        /// <param name="modelFile">Model file path</param>
+        /// <param name="gpuDeviceId">Optional device ID to run execution on</param>
+        /// <param name="fallbackToCpu">If GPU error, raise exception or fallback to CPU</param>
```
sfilipi: I think we tend to place the '.' at the end of the sentences, since they will appear in the user facing documentation website.

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
```diff
@@ -64,6 +71,12 @@ public sealed class Arguments : TransformInputBase
 
             [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "Name of the output column.", SortOrder = 2)]
             public string[] OutputColumns;
+
+            [Argument(ArgumentType.AtMostOnce | ArgumentType.Required, HelpText = "GPU device id to run on (e.g. 0,1,..). -1 runs on CPU. Requires CUDA 10.0.", SortOrder = 3)]
+            public int GpuDeviceId = -1;
```
shmoradims: >public int GpuDeviceId = -1; [](start = 12, length = 28)

I suggest we use nullable<int> (int?) instead of int with a special -1 value. Int with special -1 as null, is a C/C++ usage because there's no built-in nullable type in C/C++. In C#, that type is supported, so we should use it. That way we don't have this magic -1 value floating all over the code. I suggest we change all C# methods that take GpuDeviceId to use int? instead of int. The value should be changed to int when OnnxRuntime is called at the very end.

Issue #433, opened 2018-06-27
Direct API: Auto-normalization


TomFinley:
One of the details of training that happen after a loader/transform pipeline is created, but before the cache. We've typically automatically done this for users. While usage in the API is very distinct in that people tend to like implicit behavior in tools but dislike implicit behavior in APIs, at least offering a convenience for normalization is appropriate.

## Existing Method

Some familiar with this codebase are aware of this existing method, in the `TrainUtils` utility class, that serves a similar function.

https://github.com/dotnet/machinelearning/blob/2501049f5cb60ed2c9ec191d2937cab7b59824da/src/Microsoft.ML.Data/Commands/TrainCommand.cs#L492

The goal of *that* method is not to provide a convenient API, so much as to factor out code common to the various commands that train models (e.g., train, traintest, cross-validation, some transforms like train-and-score). The same is true of many methods in that `TrainUtils` class. This as we see in the first few lines:

https://github.com/dotnet/machinelearning/blob/2501049f5cb60ed2c9ec191d2937cab7b59824da/src/Microsoft.ML.Data/Commands/TrainCommand.cs#L499-L506

While beneficial in providing consistent behavior across all of these things from a command-line perspective, the condition where it just exits would be inappropriate to have in an ML.NET API -- you might imagine someone designing a method with a parameter `bool doNothing` where the first thing is, if it's true, the method returns without doing anything. Again, appropriate from the point of view of factoring out common code, but not appropriate for an API. Also the method of communicating important information to the user is via the console, which again is not the most helpful option for an API.

## Proposed API Helpers

Nonetheless, this function has several things that are helpful to do: it detects if a trainer wants normalization, if data is normalized, and if appropriate and necessary applies normalization.

This would probably take the form of a static method on the `NormalizerTransform` class, perhaps following this signature:

```csharp
public static bool CreateIfNeeded(IHostEnvironment env, ref RoleMappedData data, ITrainer trainer)
```

We could also have two additional methods to provide key information.

```csharp
public static bool FeatureVectorIsNormalized(RoleMappedData data)
public static bool NeedsNormalization(this ITrainer trainer)
```

TomFinley:
Incidnetally, followup on #371.

TomFinley:
[...]

PR #1924, opened 2018-12-19
Replace ColumnInfo usage with Schema.Column, remove ColumnInfo

### LINKED ISSUE #1923: ColumnInfo associated with RoleMappedSchema should be replaced with Schema.Column
We have a fun class called `RoleMappedSchema`. It operates over a class called `ColumnInfo` that looks like this:

https://github.com/dotnet/machinelearning/blob/98163f9c481ceda7887c9ba6abdf74608aa4a3b9/src/Microsoft.ML.Core/Data/RoleMappedSchema.cs#L18-L22

This structure is quite useful -- oftentimes we want to have names and types and indices of columns, all bundled together. We used this hundreds of places. In fact it was so useful we decided to build something `Schema.Column` that looked a little something like this:

https://github.com/dotnet/machinelearning/blob/98163f9c481ceda7887c9ba6abdf74608aa4a3b9/src/Microsoft.ML.Core/Data/Schema.cs#L86-L106

Clearly this new more fundamental structure and the older legacy structure have some things in common. In fact, the newer structure is what we are going with for our public API, so we should probably remove that older legacy structure.

We will shift usage of `ColumnInfo` to use `Schema.Column` instead (nullables where appropriate, where things can be null), and then remove `ColumnInfo`.

TomFinley:
Fixes #1923. As usual commits are structured in such a way as to make the code approachable. Since the first usage was against `RoleMappedSchema` and that was used in many, *many* places, I'm sorry to say that is probably the biggest change.

### DIFF
--- a/src/Microsoft.ML.Core/Data/RoleMappedSchema.cs
+++ b/src/Microsoft.ML.Core/Data/RoleMappedSchema.cs
@@ -9,57 +9,6 @@ using Microsoft.ML.Runtime.Internal.Utilities;
 namespace Microsoft.ML.Runtime.Data
 {
     /// <summary>
-    /// This contains information about a column in an <see cref="IDataView"/>. It is essentially a convenience cache
-    /// containing the name, column index, and column type for the column. The intended usage is that users of <see cref="RoleMappedSchema"/>
-    /// will have a convenient method of getting the index and type without having to separately query it through the <see cref="Schema"/>,
-    /// since practically the first thing a consumer of a <see cref="RoleMappedSchema"/> will want to do once they get a mappping is get
-    /// the type and index of the corresponding column.
-    /// </summary>
-    public sealed class ColumnInfo
-    {
-        public readonly string Name;
-        public readonly int Index;
-        public readonly ColumnType Type;
-
-        private ColumnInfo(string name, int index, ColumnType type)
-        {
-            Name = name;
-            Index = index;
-            Type = type;
-        }
-
-        /// <summary>
-        /// Tries to create a ColumnInfo for the column with the given name in the given schema. Returns
-        /// false if the name doesn't map to a column.
-        /// </summary>
-        public static bool TryCreateFromName(Schema schema, string name, out ColumnInfo colInfo)
-        {
-            Contracts.CheckValue(schema, nameof(schema));
-            Contracts.CheckNonEmpty(name, nameof(name));
-
-            colInfo = null;
-            if (!schema.TryGetColumnIndex(name, out int index))
-                return false;
-
-            colInfo = new ColumnInfo(name, index, schema[index].Type);
-            return true;
-        }
-
-        /// <summary>
-        /// Creates a ColumnInfo for the column with the given column index. Note that the name
-        /// of the column might actually map to a different column, so this should be used with care
-        /// and rarely.
-        /// </summary>
-        public static ColumnInfo CreateFromIndex(Schema schema, int index)
-        {
-            Contracts.CheckValue(schema, nameof(schema));
-            Contracts.CheckParam(0 <= index && index < schema.Count, nameof(index));
-
-            return new ColumnInfo(schema[index].Name, index, schema[index].Type);
-        }
-    }
-
-    /// <summary>
     /// Encapsulates an <see cref="Schema"/> plus column role mapping information. The purpose of role mappings is to
     /// provide information on what the intended usage is for. That is: while a given data view may have a column named
     /// "Features", by itself that is insufficient: the trainer must be fed a role mapping that says that the role
--- a/src/Microsoft.ML.Core/Data/RoleMappedSchema.cs
+++ b/src/Microsoft.ML.Core/Data/RoleMappedSchema.cs
@@ -29,18 +29,6 @@ namespace Microsoft.ML.Runtime.Data
         }
 
         /// <summary>
-        /// Create a ColumnInfo for the column with the given name in the given schema. Throws if the name
-        /// doesn't map to a column.
-        /// </summary>
-        public static ColumnInfo CreateFromName(Schema schema, string name, string descName)
-        {
-            if (!TryCreateFromName(schema, name, out var colInfo))
-                throw Contracts.ExceptParam(nameof(name), $"{descName} column '{name}' not found");
-
-            return colInfo;
-        }
-
-        /// <summary>
         /// Tries to create a ColumnInfo for the column with the given name in the given schema. Returns
         /// false if the name doesn't map to a column.
         /// </summary>
@@ -192,32 +180,32 @@ namespace Microsoft.ML.Runtime.Data
         /// <summary>
         /// The <see cref="ColumnRole.Feature"/> column, when there is exactly one (null otherwise).
         /// </summary>
-        public ColumnInfo Feature { get; }
+        public Schema.Column? Feature { get; }
 
         /// <summary>
         /// The <see cref="ColumnRole.Label"/> column, when there is exactly one (null otherwise).
         /// </summary>
-        public ColumnInfo Label { get; }
+        public Schema.Column? Label { get; }
 
         /// <summary>
         /// The <see cref="ColumnRole.Group"/> column, when there is exactly one (null otherwise).
         /// </summary>
-        public ColumnInfo Group { get; }
+        public Schema.Column? Group { get; }
 
         /// <summary>
         /// The <see cref="ColumnRole.Weight"/> column, when there is exactly one (null otherwise).
         /// </summary>
-        public ColumnInfo Weight { get; }
+        public Schema.Column? Weight { get; }
 
         /// <summary>
         /// The <see cref="ColumnRole.Name"/> column, when there is exactly one (null otherwise).
         /// </summary>
-        public ColumnInfo Name { get; }
+        public Schema.Column? Name { get; }
 
         // Maps from role to the associated column infos.
-        private readonly Dictionary<string, IReadOnlyList<ColumnInfo>> _map;
+        private readonly Dictionary<string, IReadOnlyList<Schema.Column>> _map;
 
-        private RoleMappedSchema(Schema schema, Dictionary<string, IReadOnlyList<ColumnInfo>> map)
+        private RoleMappedSchema(Schema schema, Dictionary<string, IReadOnlyList<Schema.Column>> map)
         {
             Contracts.AssertValue(schema);
             Contracts.AssertValue(map);
@@ -256,42 +244,40 @@ namespace Microsoft.ML.Runtime.Data
             }
         }
 
-        private RoleMappedSchema(Schema schema, Dictionary<string, List<ColumnInfo>> map)
+        private RoleMappedSchema(Schema schema, Dictionary<string, List<Schema.Column>> map)
             : this(schema, Copy(map))
         {
         }
 
-        private static void Add(Dictionary<string, List<ColumnInfo>> map, ColumnRole role, ColumnInfo info)
+        private static void Add(Dictionary<string, List<Schema.Column>> map, ColumnRole role, Schema.Column column)
         {
             Contracts.AssertValue(map);
             Contracts.AssertNonEmpty(role.Value);
-            Contracts.AssertValue(info);
 
             if (!map.TryGetValue(role.Value, out var list))
             {
-                list = new List<ColumnInfo>();
+                list = new List<Schema.Column>();
                 map.Add(role.Value, list);
             }
-            list.Add(info);
+            list.Add(column);
         }
 
-        private static Dictionary<string, List<ColumnInfo>> MapFromNames(Schema schema, IEnumerable<KeyValuePair<ColumnRole, string>> roles, bool opt = false)
+        private static Dictionary<string, List<Schema.Column>> MapFromNames(Schema schema, IEnumerable<KeyValuePair<ColumnRole, string>> roles, bool opt = false)
         {
             Contracts.AssertValue(schema);
             Contracts.AssertValue(roles);
 
-            var map = new Dictionary<string, List<ColumnInfo>>();
+            var map = new Dictionary<string, List<Schema.Column>>();
             foreach (var kvp in roles)
             {
                 Contracts.AssertNonEmpty(kvp.Key.Value);
                 if (string.IsNullOrEmpty(kvp.Value))
                     continue;
-                ColumnInfo info;
-                if (!opt)
-                    info = ColumnInfo.CreateFromName(schema, kvp.Value, kvp.Key.Value);
-                else if (!ColumnInfo.TryCreateFromName(schema, kvp.Value, out info))
-                    continue;
-                Add(map, kvp.Key.Value, info);
+                var info = schema.GetColumnOrNull(kvp.Value);
+                if (info.HasValue)
+                    Add(map, kvp.Key.Value, info.Value);
+                else if (!opt)
+                    throw Contracts.ExceptParam(nameof(schema), $"{kvp.Value} column '{kvp.Key.Value}' not found");
             }
             return map;
         }
@@ -318,18 +304,18 @@ namespace Microsoft.ML.Runtime.Data
         /// If there are columns of the given role, this returns the infos as a readonly list. Otherwise,
         /// it returns null.
         /// </summary>
-        public IReadOnlyList<ColumnInfo> GetColumns(ColumnRole role)
+        public IReadOnlyList<Schema.Column> GetColumns(ColumnRole role)
             => _map.TryGetValue(role.Value, out var list) ? list : null;
 
         /// <summary>
         /// An enumerable over all role-column associations within this object.
         /// </summary>
-        public IEnumerable<KeyValuePair<ColumnRole, ColumnInfo>> GetColumnRoles()
+        public IEnumerable<KeyValuePair<ColumnRole, Schema.Column>> GetColumnRoles()
         {
             foreach (var roleAndList in _map)
             {
                 foreach (var info in roleAndList.Value)
-                    yield return new KeyValuePair<ColumnRole, ColumnInfo>(roleAndList.Key, info);
+                    yield return new KeyValuePair<ColumnRole, Schema.Column>(roleAndList.Key, info);
             }
         }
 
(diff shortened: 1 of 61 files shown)


src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
```diff
@@ -172,13 +172,13 @@ private protected override Aggregator GetAggregatorCore(RoleMappedSchema schema,
         private ReadOnlyMemory<char>[] GetClassNames(RoleMappedSchema schema)
         {
             // Get the label names if they exist, or use the default names.
-            ColumnType type;
             var labelNames = default(VBuffer<ReadOnlyMemory<char>>);
-            if (schema.Label.Type.IsKey &&
-                (type = schema.Schema[schema.Label.Index].Metadata.Schema.GetColumnOrNull(MetadataUtils.Kinds.KeyValues)?.Type) != null &&
-                type.ItemType.IsKnownSizeVector && type.ItemType.IsText)
```
TomFinley: While most of the changes in the PR are intended to be over identical functionality, this was an exception since it was clearly a bug... `type.ItemType.IsKnownSizeVector` must always be false, since an item type could never be a vector type of any sort, known size or not!

src/Microsoft.ML.Data/Evaluators/MamlEvaluator.cs
```diff
@@ -239,23 +239,23 @@ private IDataView WrapPerInstance(RoleMappedData perInst)
                 colsToKeep.Add(MetricKinds.ColumnNames.FoldIndex);
 
             // Maml always outputs a name column, if it doesn't exist add a GenerateNumberTransform.
-            if (perInst.Schema.Name == null)
+            if (perInst.Schema.Name?.Name is string nameName)
```
TomFinley: This is sort of a fun application of pattern matching. If that evaluated to `null`, then the pattern would not match. Consider this small program:

```csharp
private static void TestString(string s)
{
    if (s is string ss)
        Console.WriteLine($"Yup {ss}");
    else
        Console.WriteLine("Nope");
}

private static void Main(string[] args)
{
    TestString(null);
    TestString("hi");
}
```

The output is:

```
Nope
Yup hi
```

Very good. Yet, let's imagine I change to this:

```csharp
    if (s is var ss)
```

Now, that `ss` is still of `string` type, but now the outcome becomes this:

```
Yup
Yup hi
```

I don't fully appreciate in what way this `var` pattern matching matches the type, yet evinces such very different behavior! No doubt one of the donut people knows more about this.

src/Microsoft.ML.Core/Data/RoleMappedSchema.cs
```diff
@@ -256,42 +193,40 @@ private RoleMappedSchema(Schema schema, Dictionary<string, IReadOnlyList<ColumnI
...
-                if (!opt)
-                    info = ColumnInfo.CreateFromName(schema, kvp.Value, kvp.Key.Value);
-                else if (!ColumnInfo.TryCreateFromName(schema, kvp.Value, out info))
-                    continue;
-                Add(map, kvp.Key.Value, info);
+                var info = schema.GetColumnOrNull(kvp.Value);
+                if (info.HasValue)
+                    Add(map, kvp.Key.Value, info.Value);
+                else if (!opt)
+                    throw Contracts.ExceptParam(nameof(schema), $"{kvp.Value} column '{kvp.Key.Value}' not found");
```
Zruty0: >ExceptParam [](start = 36, length = 11)

do you think `ExceptSchemaMismatch` is a good fit here?

src/Microsoft.ML.Data/Evaluators/QuantileRegressionEvaluator.cs
```diff
@@ -62,10 +62,10 @@ private protected override void CheckScoreAndLabelTypes(RoleMappedSchema schema)
                 throw Host.Except(
                     "Score column '{0}' has type '{1}' but must be a known length vector of type R4 or R8", score.Name, t);
             }
-            Host.Check(schema.Label != null, "Could not find the label column");
-            t = schema.Label.Type;
+            Host.CheckParam(schema.Label.HasValue, nameof(schema), "Must contain a label column");
+            t = schema.Label.Value.Type;
             if (t != NumberType.R4)
-                throw Host.Except("Label column '{0}' has type '{1}' but must be R4", schema.Label.Name, t);
+                throw Host.Except("Label column '{0}' has type '{1}' but must be R4", schema.Label.Value.Name, t);
```
Zruty0: >Except [](start = 27, length = 6)

again maybe `ExceptSchemaMismatch`?

Zruty0:
>         private protected override void CheckScoreAndLabelTypes(RoleMappedSchema schema)

the comment for `ExceptSchemaMismatch` applies for all of these overrides. I can see this is not the reason for the PR, so I won't insist on you doing it now, but, I assume, you touch it you fix it :)

---
Refers to: src/Microsoft.ML.Data/Evaluators/RegressionEvaluator.cs:55 in d2f1818. [](commit_id = d2f18189fe0fc3e4a20ac857a5b69e9159edc359, deletion_comment = False)

src/Microsoft.ML.StandardLearners/FactorizationMachine/FactorizationMachineTrainer.cs
```diff
@@ -323,7 +321,12 @@ private static float CalculateLossSlope(float label, float modelResponse)
                 var validFeatureColumns = data.Schema.GetColumns(RoleMappedSchema.ColumnRole.Feature);
                 Host.Assert(fieldCount == validFeatureColumns.Count);
                 for (int f = 0; f < fieldCount; f++)
-                    Host.Assert(featureColumns[f] == validFeatureColumns[f]);
+                {
+                    var featCol = featureColumns[f];
+                    var validFeatCol = validFeatureColumns[f];
+                    Host.Assert(featCol.Name == validFeatCol.Name);
+                    Host.Assert(featCol.Type == validFeatCol.Type);
```
Zruty0: >== [](start = 45, length = 2)

Is it reference-equal or `Equals` ?

src/Microsoft.ML.TimeSeries/AdaptiveSingularSpectrumSequenceModeler.cs
```diff
@@ -1225,12 +1225,14 @@ internal override void Train(FixedSizeQueue<Single> data)
...
         internal override void Train(RoleMappedData data)
         {
-            _host.CheckParam(data != null, nameof(data), "The input series for training cannot be null.");
-            if (data.Schema.Feature.Type != NumberType.Float)
-                throw _host.ExceptUserArg(nameof(data.Schema.Feature.Name), "The feature column has  type '{0}', but must be a float.", data.Schema.Feature.Type);
+            _host.CheckValue(data, nameof(data));
+            _host.CheckParam(data.Schema.Feature.HasValue, nameof(data), "Must have features column.");
+            var featureCol = data.Schema.Feature.Value;
+            if (featureCol.Type != NumberType.Float)
+                throw _host.ExceptParam(nameof(data), "The feature column has  type '{0}', but must be a float.", featureCol.Type);
```
Zruty0: >ExceptParam [](start = 28, length = 11)

another case for `ExceptSchemaMismatch`

src/Microsoft.ML.StandardLearners/FactorizationMachine/FactorizationMachineTrainer.cs
TomFinley: We don't have such a thing as reference equals on `Schema.Column`, so the closest I could do is equals on the fields (which in the case of type is a reference equals).

src/Microsoft.ML.Core/Data/RoleMappedSchema.cs
TomFinley: Maybe. TBH I was never quite sure what `ExceptSchemaMismatch` was for exactly.

src/Microsoft.ML.TimeSeries/AdaptiveSingularSpectrumSequenceModeler.cs
TomFinley: All right. I see this on 1232, but the case of 1229 (could not find *any* column with that role, no matter its name), does not seem to be served by this class of helper. (Which is fine.)

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
```diff
@@ -129,11 +129,11 @@ private protected override void CheckScoreAndLabelTypes(RoleMappedSchema schema)
...
             if (t.IsVector || t.ItemType != NumberType.Float)
-                throw host.SchemaSensitive().Except("Score column '{0}' has type '{1}' but must be R4", score, t);
-            host.Check(schema.Label != null, "Could not find the label column");
-            t = schema.Label.Type;
+                throw host.ExceptSchemaMismatch(nameof(schema), "score", score.Name, "R4", t.ToString());
+            host.Check(schema.Label.HasValue, "Could not find the label column");
+            t = schema.Label.Value.Type;
             if (t != NumberType.R4 && t != NumberType.R8 && t != BoolType.Instance && t.KeyCount != 2)
-                throw host.SchemaSensitive().Except("Label column '{0}' has type '{1}' but must be R4, R8, BL or a 2-value key", schema.Label.Name, t);
+                throw host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "R4, R8, BL or a 2-value key", t.ToString());
```
justinormont: Does a Binary learner currently support R4 or R8?

src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
```diff
@@ -97,15 +97,15 @@ private protected override void CheckScoreAndLabelTypes(RoleMappedSchema schema)
             var t = score.Type;
             if (t != NumberType.Float)
                 throw Host.Except("Score column '{0}' has type '{1}' but must be R4", score, t).MarkSensitive(MessageSensitivity.Schema);
-            Host.Check(schema.Label != null, "Could not find the label column");
-            t = schema.Label.Type;
+            Host.Check(schema.Label.HasValue, "Could not find the label column");
+            t = schema.Label.Value.Type;
             if (t != NumberType.Float && t.KeyCount != 2)
-                throw Host.Except("Label column '{0}' has type '{1}' but must be R4 or a 2-value key", schema.Label.Name, t).MarkSensitive(MessageSensitivity.Schema);
+                throw Host.Except("Label column '{0}' has type '{1}' but must be R4 or a 2-value key", schema.Label.Value.Name, t).MarkSensitive(MessageSensitivity.Schema);
```
sfilipi: >R4  [](start = 81, length = 3)

sidenote, does it make sense to start using 'float' etc. in the user errors, instead of the internal types?

sfilipi: nvm, I'll log an issue and we can discuss there, rather than comments on  this PR. 

---
In reply to: [243381184](https://github.com/dotnet/machinelearning/pull/1924#discussion_r243381184) [](ancestors = 243381184)

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
TomFinley: No idea. But the evaluator here claims to support both.

src/Microsoft.ML.Data/Evaluators/QuantileRegressionEvaluator.cs
```diff
@@ -50,22 +50,19 @@ private protected override IRowMapper CreatePerInstanceRowMapper(RoleMappedSchem
...
-                    "Score column '{0}' has type '{1}' but must be a known length vector of type R4 or R8", score.Name, t);
-            }
-            Host.Check(schema.Label != null, "Could not find the label column");
-            t = schema.Label.Type;
+                throw Host.ExceptSchemaMismatch(nameof(schema), "score", score.Name, "vector of type R4 or R8", t.ToString());
+            Host.CheckParam(schema.Label.HasValue, nameof(schema), "Must contain a label column");
+            t = schema.Label.Value.Type;
             if (t != NumberType.R4)
-                throw Host.Except("Label column '{0}' has type '{1}' but must be R4", schema.Label.Name, t);
+                throw Host.ExceptSchemaMismatch(nameof(schema), "score", schema.Label.Value.Name, "R4", t.ToString());
```
sfilipi: >score [](start = 65, length = 5)

label

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
```diff
@@ -171,14 +172,14 @@ private static bool CheckScoreColumnKind(Schema schema, int col)
 
             if (!string.IsNullOrWhiteSpace(name))
             {
-                ColumnInfo info;
 #pragma warning disable MSML_ContractsNameUsesNameof
-                if (!ColumnInfo.TryCreateFromName(schema, name, out info))
+                var col = schema.GetColumnOrNull(name);
+                if (!col.HasValue)
```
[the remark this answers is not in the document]
TomFinley: Sorry, not sure what's is being asked here @sfilipi , could you clarify? `schema` is checked for null, the `name` has been checked for null, and as you see below the nullable struct is being checked for a value and throwing an appropriate exception. As far as I can tell no other entities are involved in the line you're commenting on?

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
```diff
@@ -115,51 +115,52 @@ private static bool CheckScoreColumnKind(Schema schema, int col)
...
+                col.Metadata.GetValue(MetadataUtils.Kinds.ScoreColumnKind, ref tmp);
                 ectx.Assert(ReadOnlyMemoryUtils.EqualsStr(kind, tmp));
 #endif
                 // REVIEW: What should this do about hidden columns? Currently we ignore them.
-                if (schema[col].IsHidden)
+                if (col.IsHidden)
                     continue;
-                if (schema.TryGetMetadata(TextType.Instance, MetadataUtils.Kinds.ScoreValueKind, col, ref tmp) &&
-                    ReadOnlyMemoryUtils.EqualsStr(valueKind, tmp))
+                if (col.Metadata.Schema.GetColumnOrNull(MetadataUtils.Kinds.ScoreValueKind)?.Type == TextType.Instance)
```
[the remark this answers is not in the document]
TomFinley: It absolutely is. I think you're suggesting that we don't *really* need to check for the type, is that so?

`IDataView` is an open system, and metadata is optional -- but the flipside of being optional and open is that we must accomodate ourselves (without error) that the type of metadata might not quite be what we expect.

I've written about this in the past, particularly relevant is the first paragraph of [this section of my `IDataView` implementation document](https://github.com/dotnet/machinelearning/blob/master/docs/code/IDataViewImplementation.md#metadata), titled "Metadata."

src/Microsoft.ML.Data/Evaluators/QuantileRegressionEvaluator.cs
TomFinley: Ah good one thanks @sfilipi !

src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
TomFinley: It may. In fact I think we've discussed this as a desirable outcome, but if we do it we should do it *everywhere* as part of a deliberate policy. So your plan to open an issue seems good to me, @sfilipi.

src/Microsoft.ML.Data/Transforms/NormalizeUtils.cs
```diff
@@ -70,27 +70,6 @@ internal interface IColumnFunction : ICanSaveModel
         NormalizingTransformer.NormalizerModelParametersBase GetNormalizerModelParams();
     }
 
-    internal static class NormalizeUtils
```
sfilipi: >internal static class NormalizeUtils [](start = 4, length = 36)

double-checking that removing this is intentional.

src/Microsoft.ML.StandardLearners/Standard/SdcaBinary.cs
```diff
@@ -120,13 +120,12 @@ private protected RoleMappedData PrepareDataFromTrainingExamples(IChannel ch, Ro
...
             var examplesToFeedTrain = new RoleMappedData(idvToFeedTrain, roles);
 
-            ch.AssertValue(examplesToFeedTrain.Schema.Label);
-            ch.AssertValue(examplesToFeedTrain.Schema.Feature);
-            if (examples.Schema.Weight != null)
-                ch.AssertValue(examplesToFeedTrain.Schema.Weight);
+            ch.Assert(examplesToFeedTrain.Schema.Label.HasValue);
+            ch.Assert(examplesToFeedTrain.Schema.Feature.HasValue);
+            if (examples.Schema.Weight.HasValue)
+                ch.Assert(examplesToFeedTrain.Schema.Weight.HasValue);
```
sfilipi: did you mean to assert the Value :

ch.Assert(examplesToFeedTrain.Schema.Weight.Value);

src/Microsoft.ML.Data/Transforms/NormalizeUtils.cs
TomFinley: It is. It is an artifact of a time back in summer, back when we were first formulating ideas of what the public API would look like, and in fact the ideas around this were fairly ill formed. (This particular code was introduced in #433.) Clearly our thinking has moved on since then. This *particular* method (which as you see used to operate over `ColumnInfo`) is now directly taken care of by the `bool IsNormalized(this Schema.Column column)` extension method. And, because `RoleMappedSchema` now directly returns `Schema.Column` instead of `ColumnInfo` (which is the point of this PR, after all), there was no longer any reason to have this method. But thanks for checking @sfilipi.

---
In reply to: [243412114](https://github.com/dotnet/machinelearning/pull/1924#discussion_r243412114) [](ancestors = 243412114)

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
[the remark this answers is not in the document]
sfilipi: I mean col can be null, since it comes from GetColumnOrNull. 

Shall it check for null before accessing members on it: 

Next line: col?.HasValue

---
In reply to: [243391027](https://github.com/dotnet/machinelearning/pull/1924#discussion_r243391027) [](ancestors = 243391027)

src/Microsoft.ML.StandardLearners/Standard/SdcaBinary.cs
TomFinley: I absolutely did not mean to check on the value. First of all, I cannot, since `Nullable<T>` is a struct and `Schema.Column` is a struct, we cannot call `AssertValue` on it. Also, even if hypothetically I could check on `Value`, if `HasValue` were false then accessing that `Value` property would itself throw an exception, which is not what I want (see [documentation of `Nullable<T> here](https://docs.microsoft.com/en-us/dotnet/api/system.nullable-1.value?view=netstandard-2.0)).

If you're asking if these should be `Check` and not `Assert`, it may not be obvious but `examplesToFeedTrain` is derived from `examples` (see the forematter of this, all that stuff about shuffling and whatnot), and the roles of `examples` were already checked or otherwise confirmed to be present (see, for example, lines 102 for label, 103 for float, and line 125 in this block for weights). So asserts are the appropriate choice here, since the intention of this code is to communicate to the reader what we *expect* to be true, versus input validation, which would be more properly handled via checks or conditions (as in fact it is).

PR #1926, opened 2018-12-19
Update PFI tests so they configure sensitive learners to use a single thread.

### LINKED ISSUE #1925: Spurious Failures on TestPfiClusteringOnDenseFeatures test
TestPfiClusteringOnDenseFeatures has, I believe introduced a week ago in PR #1832, has been responsible for some seemingly spurious test failures that are blocking builds. See, e.g., this build here from @Zruty0 for his PR #1920 , where PFI (which is totally unrelated to @Zruty0 's change) is blocking checking it in.

My best guess as to why this is happening is that for this clustering test the number of threads was not set to 1. (That obviously will cause issues, but whether it's the only issue is more than I know right now.) Sometimes you can *kinda* get away with that (e.g., sometimes results from logistic regression can resemble each other in different configuration settings since it is solving a convex problem), but this is not true of clustering. For that reason, the test `TestPfiClusteringOnDenseFeatures` has been responsible for multiple spurious test failures since its introduction last week.

We will have a PR to set the threads to 1, at least as a first try to get the situation somewhat under control. Note that this will  not be necessary in FastTree since it is engineered in such a way that it gets the same result no matter how many or few threads are used.

/cc @rogancarr @shmoradims @artidoro

TomFinley:
Fixes #1925 .

While I was at it I also set the number of threads to logistic regression to 1 as a better practice, since even though it's less likely to trigger a failure here, better safe than sorry. As mentioned in the issue since FastTree's results don't depend on number of threads I did not set it there.

Also incidentally started using `Assert.Equals(a, b)` instead of `Assert.True(b == a)`, so as to provide slightly more helpful error messages for when test failures happen, but that should not affect the running of the test.

As with most spurious test failures, due to the indeterminacy of the effects I am not actually 100% sure this will solve the issue, but it was just one obvious mistake I saw when reviewing the test code, so I figured this might help improve the spurious failures on account of this issue.

### DIFF
--- a/test/Microsoft.ML.Tests/PermutationFeatureImportanceTests.cs
+++ b/test/Microsoft.ML.Tests/PermutationFeatureImportanceTests.cs
@@ -40,18 +40,18 @@ namespace Microsoft.ML.Tests
             // X4Rand: 3
 
             // For the following metrics lower is better, so maximum delta means more important feature, and vice versa
-            Assert.True(MinDeltaIndex(pfi, m => m.L1) == 3);
-            Assert.True(MaxDeltaIndex(pfi, m => m.L1) == 1);
+            Assert.Equal(3, MinDeltaIndex(pfi, m => m.L1));
+            Assert.Equal(1, MaxDeltaIndex(pfi, m => m.L1));
 
-            Assert.True(MinDeltaIndex(pfi, m => m.L2) == 3);
-            Assert.True(MaxDeltaIndex(pfi, m => m.L2) == 1);
+            Assert.Equal(3, MinDeltaIndex(pfi, m => m.L2));
+            Assert.Equal(1, MaxDeltaIndex(pfi, m => m.L2));
 
-            Assert.True(MinDeltaIndex(pfi, m => m.Rms) == 3);
-            Assert.True(MaxDeltaIndex(pfi, m => m.Rms) == 1);
+            Assert.Equal(3, MinDeltaIndex(pfi, m => m.Rms));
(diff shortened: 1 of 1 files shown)


test/Microsoft.ML.Tests/PermutationFeatureImportanceTests.cs
```diff
@@ -325,8 +325,8 @@ public void TestPfiClusteringOnDenseFeatures()
 
             var preview = data.Preview();
 
-            var model = ML.Clustering.Trainers.KMeans("Features", clustersCount: 5, 
-                advancedSettings: args =>{ args.NormalizeFeatures = NormalizeOption.No;})
+            var model = ML.Clustering.Trainers.KMeans("Features", clustersCount: 5,
+                advancedSettings: args =>{ args.NormalizeFeatures = NormalizeOption.No; args.NumThreads = 1; })
```
TomFinley: This is what I am hoping will solve the issue.

TomFinley: I am also not sure what is this `NormalizeOption.No`. I do not think `Fit` will auto-normalize in any event anyway?

Zruty0: Yep, I think this argument is only use by the entry point.

---
In reply to: [242977200](https://github.com/dotnet/machinelearning/pull/1926#discussion_r242977200) [](ancestors = 242977200)

PR #1928, opened 2018-12-19
BinaryLoader cursor doesn't set state to "Done"

### LINKED ISSUE #1927: The BinaryLoader cursor doesn't always set its state to "Done" when it's done
If instead of the usual while loop:

while (cursor.MoveNext())
{
// do stuff
}

we have this loop:

while (cursor.State != CursorState.Done)
{
// do stuff
cursor.MoveNext();
}

we get an exception. This happens in the PerGroupTransformBase class, that uses an auxiliary cursor that moves ahead to check where the current group ends.

yaeldekel:
Fixes #1927 .


src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs
```diff
@@ -1366,7 +1366,10 @@ public Cursor(BinaryLoader parent, Func<int, bool> predicate, Random rand)
             protected override void Dispose(bool disposing)
             {
                 if (_disposed)
```
TomFinley: >_disposed [](start = 20, length = 9)

Hi @yaeldekel, thanks for looking at this...

This strikes me as odd. Certainly it seems a violation of the classic dispose pattern.

This only gets called if `_disposed` is true. I would have expected that in normal instances of the pattern, if dispose has already has been called, that it will immediately exit. To do otherwise suggests that there is *different* behavior when you call the dispose method multiple times, which is a clear sign something is seriously wrong!

Also we're calling the base dispose method at the *start* of the method call, but this is incorrect -- just as constructors are called inside out, the corresponding disposers should be called outside in -- that is, the base method should be the *last* thing called (which is, indeed, what happens).

I imagine that there's some actual bug that's happening at some level, but whatever is going on, I find it hard to believe that this is the solution?

src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs
```diff
@@ -2037,7 +2044,7 @@ protected override bool MoveNextCore()
                         // threads will exit if all potentially blocking operations are
                         // waiting on the same cancellation token that we catch here.
                         Contracts.Assert(ex.CancellationToken == _exMarshaller.Token);
-                        _disposed = true;
```
TomFinley: >_disposed = true; [](start = 24, length = 17)

I notice we've removed `_disposed` but not replaced it with a `_done`. I *think* it should also be a `_done` here, since that flag controls whether or not the dispose method will try to clean up the threads, and the comment above seems to indicate that in such a case as this it is not necessary to clean up the threads, so we should still set the `_done` flag that suppresses the thread cleanup? Is it not so?

yaeldekel: Right. Thanks, I wasn't sure about this case :-).

---
In reply to: [243345481](https://github.com/dotnet/machinelearning/pull/1928#discussion_r243345481) [](ancestors = 243345481)

PR #1929, opened 2018-12-19
Implement ICanSaveInIniFormat interface for GamPredictor

shmoradims:
Add support for saving GAM models as ini format. Parity results with maml are included in the tests.


src/Microsoft.ML.FastTree/FastTree.cs
```diff
@@ -3404,4 +3377,47 @@ public TreeNode(Dictionary<string, object> keyValues)
             public Dictionary<string, object> KeyValues { get; }
         }
     }
+    internal static class FastTreeIniFormatUtils
```
rogancarr: Can you add this as a separate file in `FastTree/Utils`. I'm not sure we need it in `FastTree.cs`

src/Microsoft.ML.FastTree/FastTree.cs
```diff
@@ -2988,37 +2988,10 @@ void ICanSaveInIniFormat.SaveAsIni(TextWriter writer, RoleMappedSchema schema, I
             Host.CheckValue(writer, nameof(writer));
             Host.CheckValue(schema, nameof(schema));
```
rogancarr: Don't need to check here as you don't use it in this function.

src/Microsoft.ML.FastTree/GamTrainer.cs
```diff
@@ -1028,6 +1030,118 @@ void ICanSaveSummary.SaveSummary(TextWriter writer, RoleMappedSchema schema)
...
+            // It's eaiser to remove them here, than to modify the FastTree code.
+            var goodLines = ini.Split(new[] { '\n' }).Where(line => !line.StartsWith("SplitGain="));
+            ini = string.Join("\n", goodLines);
+            writer.WriteLine(ini);
+        }
+
+        // GAM bins should be converted to balanced trees / binary search trees
+        // so that scoring takes O(log(n)) instead of O(n). The following utility
+        // creates a balanced tree.
+        private (float[], int[], int[]) CreateBalancedTree(int numInternalNodes, double[] binThresholds)
```
rogancarr: Does returning a list limit what C# or .NET versions we can target? If so, we can switch this to an `out` pattern.

src/Microsoft.ML.FastTree/GamTrainer.cs
```diff
@@ -1028,6 +1030,118 @@ void ICanSaveSummary.SaveSummary(TextWriter writer, RoleMappedSchema schema)
...
+
+                var splitFeatures = Enumerable.Repeat(featureIndex, numInternalNodes).ToArray();
+                var (treeThresholds, lteChild, gtChild) = CreateBalancedTree(numInternalNodes, binThresholds);
+                var tree = CreateRegressionTree(numLeaves, splitFeatures, treeThresholds, lteChild, gtChild, effects);
+                ensemble.AddTree(tree);
+            }
+
+            // Tried adding the intercept as the bias term in the final ini aggregator,
+            // but that didn't seem to have any effects during testing.
+            // Adding the intercept as a dummy tree with the output values being the model intercept,
```
rogancarr: This is a good comment for review, but for the final check-in, just keep this line.

test/Microsoft.ML.Predictor.Tests/TestIniModels.cs
```diff
@@ -504,4 +511,95 @@ public ProcessDebugInformation RunCommandLine(string commandLine, string dir)
...
+            string modelIniPath = GetOutputPath(FullTestName + "-model.ini");
+            using (Stream iniStream = File.Create(modelIniPath))
+            using (StreamWriter iniWriter = Utils.OpenWriter(iniStream))
+                model.LastTransformer.Model.SaveAsIni(iniWriter, roleMappedSchema);
+
+            var results = mlContext.Regression.Evaluate(data);
+
+            // Getting parity results from maml.exe:
+            // maml.exe ini ini=model.ini out=model_ini.zip data=breast-cancer.txt  loader=TextLoader{col=Label:R4:0 col=Features:R4:1-9} xf=NAHandleTransform{col=Features slot=- ind=-} kind=Regression
+            Assert.Equal(0.093256807643323947, results.L1);
```
rogancarr: I'm not sure about this test. On one hand, it's good to make sure that it can produce the score. On the other hand, it's a new baseline test for GAMs. I'd say remove it from ML.NET.

src/Microsoft.ML.FastTree/GamTrainer.cs
```diff
@@ -1028,6 +1030,118 @@ void ICanSaveSummary.SaveSummary(TextWriter writer, RoleMappedSchema schema)
...
+        }
+
+        private int CreateBalancedTreeRecursive(int lower, int upper,
+            List<int> internalNodeIndices, List<int> lteChild, List<int> gtChild, ref int internalNodeId)
+        {
+            if (lower > upper)
+            {
+                // Base case: we've reached a leaf node
+                Host.Assert(lower == upper + 1);
+                return ~lower;
```
singlis: Why negate lower?

src/Microsoft.ML.FastTree/GamTrainer.cs
```diff
@@ -1028,6 +1030,118 @@ void ICanSaveSummary.SaveSummary(TextWriter writer, RoleMappedSchema schema)
...
+            }
+            else
+            {
+                // This is postorder traversal algorithm and populating the internalNodeIndices/lte/gt lists in reverse.
+                // Preorder is the only option, because we need the results of both left/right recursions for populating the lists.
+                // As a result, lists are populated in reverse, because the root node should be the first item on the lists.
+                var mid = (lower + upper) / 2;
```
singlis: So I think this is good information - but for me its missing a key point, which is you are building a balanced tree by dividing the array in half, then taking each half and dividing that half where each midway point becomes a node in the tree.

src/Microsoft.ML.FastTree/GamTrainer.cs
```diff
@@ -1028,6 +1030,118 @@ void ICanSaveSummary.SaveSummary(TextWriter writer, RoleMappedSchema schema)
...
+                var left = CreateBalancedTreeRecursive(
+                    lower, mid - 1, internalNodeIndices, lteChild, gtChild, ref internalNodeId);
+                var right = CreateBalancedTreeRecursive(
+                    mid + 1, upper, internalNodeIndices, lteChild, gtChild, ref internalNodeId);
+                internalNodeIndices.Insert(0, mid);
+                lteChild.Insert(0, left);
```
rogancarr: This looks like we have one depth too many. Won't the final Right and Left children be the same?

src/Microsoft.ML.FastTree/FastTree.cs
shmoradims: Removed.

---
In reply to: [243139197](https://github.com/dotnet/machinelearning/pull/1929#discussion_r243139197) [](ancestors = 243139197)

src/Microsoft.ML.FastTree/FastTree.cs
shmoradims: Moved it to utils

---
In reply to: [243139101](https://github.com/dotnet/machinelearning/pull/1929#discussion_r243139101) [](ancestors = 243139101)

src/Microsoft.ML.FastTree/GamTrainer.cs
shmoradims: Tuples are heavily used as return type in Static API, so using it here doesn't create more version restrictions than what we already have in the rest of the code. I think this way it fine.

---
In reply to: [243140927](https://github.com/dotnet/machinelearning/pull/1929#discussion_r243140927) [](ancestors = 243140927)

test/Microsoft.ML.Predictor.Tests/TestIniModels.cs
shmoradims: So, do you want me to remove both TestGamRegressionIni and TestGamBinaryClassificationIni tests?

---
In reply to: [243141824](https://github.com/dotnet/machinelearning/pull/1929#discussion_r243141824) [](ancestors = 243141824)

src/Microsoft.ML.FastTree/GamTrainer.cs
shmoradims: The leaf values are negated in the ini format. The base case happens when we reach leaf nodes.

---
In reply to: [243443992](https://github.com/dotnet/machinelearning/pull/1929#discussion_r243443992) [](ancestors = 243443992)

src/Microsoft.ML.FastTree/GamTrainer.cs
shmoradims: RegressionTree has internal nodes and leaf nodes. That extra depth ends up in the leaf nodes which ends the recursion. The final left and right nodes will be two leaf nodes, OR 1 leaf node and 1 internal node (which will subsequently end up with 2 leaf nodes).

---
In reply to: [243455772](https://github.com/dotnet/machinelearning/pull/1929#discussion_r243455772) [](ancestors = 243455772)

artidoro:
> using System;

Can you add the license declaration?

---
Refers to: src/Microsoft.ML.FastTree/Utils/FastTreeIniFileUtils.cs:1 in 8022e33. [](commit_id = 8022e33d4ed2bc208f64377719d68f7da592f859, deletion_comment = False)

src/Microsoft.ML.FastTree/GamTrainer.cs
```diff
@@ -1,13 +1,13 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
-
 using System;
```
artidoro: If you can add back the line after the declaration it would be better.

test/Microsoft.ML.Predictor.Tests/TestIniModels.cs
artidoro: We definitely need a test that checks that the new Ini file produces the same results as the GAM. It would be ideal if you could compare the the outputs with a full file compare instead of using Assert.Equal. Is it possible to do that?

---
In reply to: [244387146](https://github.com/dotnet/machinelearning/pull/1929#discussion_r244387146) [](ancestors = 244387146,243141824)

shmoradims:
> using System;

done

---
In reply to: [450932989](https://github.com/dotnet/machinelearning/pull/1929#issuecomment-450932989) [](ancestors = 450932989)

---
Refers to: src/Microsoft.ML.FastTree/Utils/FastTreeIniFileUtils.cs:1 in 8022e33. [](commit_id = 8022e33d4ed2bc208f64377719d68f7da592f859, deletion_comment = False)

src/Microsoft.ML.FastTree/GamTrainer.cs
shmoradims: done

---
In reply to: [244804207](https://github.com/dotnet/machinelearning/pull/1929#discussion_r244804207) [](ancestors = 244804207)

test/Microsoft.ML.Predictor.Tests/TestIniModels.cs
shmoradims: Yes, it's better to keep these parity tests and also prevent regressions cause by refactoring in fasttree code.

The problem with ini file comparison is the same as testing baseline comparison, which needs to extract numbers in the file and compare them with some tolerance, because verbatim comparison will fail due to changes in the higher decimal places. We already have those facilities but they're fitted for train-test / CV type tasks that are common in ML. I didn't want to modify those files for this one-off ini-file comparison. Also half of this testing is done in TLC, b/c the ini parsing code is only in TLC. The baseline metrics actually come from TLC.

More importantly, the ini-format is not an open-source format and is Microsoft IP. If we want to check-in an ini file for baseline, we should have LCA clear that. 

For those reasons, I think this simple test is good enough for now. 

---
In reply to: [244822621](https://github.com/dotnet/machinelearning/pull/1929#discussion_r244822621) [](ancestors = 244822621,244387146,243141824)

Issue #1424, opened 2018-10-29
Incorrect model returned by LightGBM calibration for multiclass with softmax


mjmckp:
https://github.com/dotnet/machinelearning/blob/5123aeedb91a435236e8b94ba0f17a6d15a20db5/src/Microsoft.ML.LightGBM/LightGbmMulticlassTrainer.cs#L104

The model returned here is only correct if the objective type is `multiclassova`, since the `OvaPredictor` predictor type created here always just normalises the `_numClass` probabilities to sum to one:
https://github.com/dotnet/machinelearning/blob/5123aeedb91a435236e8b94ba0f17a6d15a20db5/src/Microsoft.ML.StandardLearners/Standard/MultiClass/Ova.cs#L575

However, when the objective type is `multiclass` (i.e., softmax), what should happen is that the `OvaPredictor` type should be given an array of predictors that return the _raw_ scores from each of the trees (i.e., the result of `CreateBinaryPredictor` instead of the `FeatureWeightsCalibratedPredictor`), and instead of normalising the outputs of these predictors, it should be computing the softmax function of the raw scores, as is done in the LightGBM implementation:
https://github.com/Microsoft/LightGBM/blob/087d30623aef4b50592cf528eb50c1c55baeaab8/src/objective/multiclass_objective.hpp#L115

These kind of bugs would be easily detected if there were unit tests comparing the results of evaluating the managed ML.NET GBDTs against the original native implementation (i.e., calling `LGBM_BoosterPredictForMat` on the Booster object used to calibrate the model).

Zruty0:
So there's 2 issues here. 
1. Write a couple tests that ensure that ML.NET prediction output matches native LightGBM. These will now fail because of the outlined problem.
2. Fix the code so that the tests pass.

wschin:
@mjmckp, I don't think ML.NET lightGBM supports those regression formulations because we have
```csharp
        private protected override void CheckAndUpdateParametersBeforeTraining(IChannel ch, RoleMappedData data, float[] labels, int[] groups)
        {
            Options["objective"] = "regression";
            // Add default metric.
            if (!Options.ContainsKey("metric"))
                Options["metric"] = "l2";
        }
```

mjmckp:
It should be straightforward to support those kinds of regressions as well, right?

Btw, two other things:
* The `Str2DoubleArray` of the `Booster` object cannot parse inf and nan values generated by LightGBM.  The modified method below will, though:
```
[...]

PR #1934, opened 2018-12-20
Loads LightGBM inf/nan properly

wschin:
Follow a suggestion mentioned in #1424 to load inf/nan from trained LightGBM model.

### DIFF
--- a/src/Microsoft.ML.LightGBM/WrappedLightGbmBooster.cs
+++ b/src/Microsoft.ML.LightGBM/WrappedLightGbmBooster.cs
@@ -96,10 +96,10 @@ namespace Microsoft.ML.LightGBM
 
         private static double[] Str2DoubleArray(string str, char delimiter)
         {
-            return str.Split(delimiter).Select(
-                x => { double t; double.TryParse(x, out t); return t; }
-            ).ToArray();
-
+            return str.Split(delimiter)
+                      .Select(s => double.TryParse(s.Replace("inf", "∞"), out double rslt) ? rslt :
+                                    (s.Contains("nan") ? double.NaN : throw new Exception($"Cannot parse as double: {s}")))
+                      .ToArray();
         }
 
         private static int[] Str2IntArray(string str, char delimiter)
--- a/src/Microsoft.ML.LightGBM/WrappedLightGbmBooster.cs
+++ b/src/Microsoft.ML.LightGBM/WrappedLightGbmBooster.cs
@@ -96,10 +96,24 @@ namespace Microsoft.ML.LightGBM
 
         private static double[] Str2DoubleArray(string str, char delimiter)
         {
-            return str.Split(delimiter)
-                      .Select(s => double.TryParse(s.Replace("inf", "∞"), out double rslt) ? rslt :
-                                    (s.Contains("nan") ? double.NaN : throw new Exception($"Cannot parse as double: {s}")))
-                      .ToArray();
+            var values = new List<double>();
+            foreach (var token in str.Split(delimiter))
+            {
+                var trimmed = token.Trim();
+
+                if (trimmed.Equals("inf", StringComparison.OrdinalIgnoreCase))
+                    values.Add(double.PositiveInfinity);
+                else if (trimmed.Equals("-inf", StringComparison.OrdinalIgnoreCase))
+                    values.Add(double.NegativeInfinity);
+                else if (trimmed.Contains("nan"))
+                    values.Add(double.NaN);
+                else
+                    // The value carried in the trimmed string is not inf, -inf, or nan.
+                    // Therefore, double.Parse should be able to generate a valid number from it.
+                    // If parsing fails, an exception will be thrown.
+                    values.Add(double.Parse(trimmed));
+            }
+            return values.ToArray();
         }
 
         private static int[] Str2IntArray(string str, char delimiter)


src/Microsoft.ML.LightGBM/WrappedLightGbmBooster.cs
```diff
@@ -96,10 +96,10 @@ internal unsafe string GetModelString()
 
         private static double[] Str2DoubleArray(string str, char delimiter)
         {
-            return str.Split(delimiter).Select(
-                x => { double t; double.TryParse(x, out t); return t; }
-            ).ToArray();
-
+            return str.Split(delimiter)
+                      .Select(s => double.TryParse(s.Replace("inf", "∞"), out double rslt) ? rslt :
```
singlis: .Replace and .Contains are case sensitive - should these be insensitive?
I know the old code did not do this - but do we need to check for empty string? Or is it assumed that check has been done before this call?

wschin: `Select` could filter out empty cases, I believe. That strings are LightGBM format, so let's leave them as they are.

src/Microsoft.ML.LightGBM/WrappedLightGbmBooster.cs
```diff
@@ -96,10 +96,10 @@ internal unsafe string GetModelString()
...
+                                    (s.Contains("nan") ? double.NaN : throw new Exception($"Cannot parse as double: {s}")))
```
Ivanidzo4ka: >throw new Exception($"Cannot parse as double: {s}")) [](start = 70, length = 52)

please, never throw unmarked exception in your code.

Ivanidzo4ka: I would probably either skip it completely or just return default value for double as we had before.

---
In reply to: [243433489](https://github.com/dotnet/machinelearning/pull/1934#discussion_r243433489) [](ancestors = 243433489)

src/Microsoft.ML.LightGBM/WrappedLightGbmBooster.cs
tannergooding: Why is this replacing `inf` with `∞` rather than with the culture-insensitive string (which is `Infinity`)

wschin: What's the difference?

---
In reply to: [243438498](https://github.com/dotnet/machinelearning/pull/1934#discussion_r243438498) [](ancestors = 243438498)

src/Microsoft.ML.LightGBM/WrappedLightGbmBooster.cs
wschin: Question: do we want to fully mimic LightGBM behavior? If yes, we need this change. No, we can just close this PR.

---
In reply to: [243433959](https://github.com/dotnet/machinelearning/pull/1934#discussion_r243433959) [](ancestors = 243433959,243433489)

src/Microsoft.ML.LightGBM/WrappedLightGbmBooster.cs
tannergooding: `∞` will only parse successfully for cultures where `NumberFormatInfo.PositiveInfinitySymbol == "∞"`; the latter is the invariant case and is stable/won't change. You should probably be doing: `double.TryParse(value, out double result, CultureInfo.InvariantCulture)` for anything that you aren't prepared to support culture specific inputs on.

tannergooding: Also, for this particular case, it is probably better to just have the following (or something very similar, depending on the exact inputs expected):
```csharp
string sTrim = s.Trim();
if (sTrim.Equals("inf", StringComparison.OrdinalIgnoreCase))
{
    return double.PositiveInfinity;
}
else if (sTrim.Equals("-inf", StringComparison.OrdinalIgnoreCase))
{
    return double.NegativeInfinity;
}
```

`double.Parse` and `double.TryParse` require that the entire string match, so doing a `string.Replace` is relatively pointless and just incurs additional allocations/overhead

src/Microsoft.ML.LightGBM/WrappedLightGbmBooster.cs
Ivanidzo4ka: I'm not against your PR, I'm against throwing unmarked exception.
Sorry for my inability to express my thoughts in proper understandable way.
>I would probably either skip it completely or just return default value for double as we had before

I don't mean your whole new code, I mean case where we can't parse double.
We have two options for that case:
- Throw MARKED exception like `throw env.Exception(message)`
- Use default value for double.

tannergooding: If it is expected that this throws, you should probably explicitly check the strings for `inf/-inf/nan` (https://github.com/dotnet/machinelearning/pull/1934#discussion_r243441276) and fallback to just calling `double.Parse` (which will throw the appropriate format exception)

src/Microsoft.ML.LightGBM/WrappedLightGbmBooster.cs
```diff
@@ -96,10 +96,24 @@ internal unsafe string GetModelString()
...
         {
-            return str.Split(delimiter).Select(
-                x => { double t; double.TryParse(x, out t); return t; }
-            ).ToArray();
+            var values = new List<double>();
+            foreach (var token in str.Split(delimiter))
+            {
+                var trimmed = token.Trim();
 
+                if (trimmed.Equals("inf", StringComparison.OrdinalIgnoreCase))
```
mjmckp: Does this really work?  The strings output by LightGBM are +nan.0, +inf.0, and -inf.0.  How are you handling the trailing .0 bit?

PR #1937, opened 2018-12-20
Remove "Runtime" from all namespace reference and move public facing classes used in Samples into Microsoft.ML namespace.

### LINKED ISSUE #1697: Remove "Runtime" from all namespace reference, sort namespaces as per .net convention and enforce the same.
Remove "Runtime" from all namespace reference, sort namespaces as per .net convention listed below and enforce the same.

**SA1210:** Using directives must be ordered alphabetically by namespace.
**SA1208:** System using directives must be placed before other using directives.

codemzs:
fixes #1697


TomFinley:
Hi @codemzs thanks for looking at this.

One thing you'll note is that you've changed 917 files. This means that both codeflow and the git web review are overwhelmed, since codeflow can only display so many files at a time. The git web tool is an inadequate tool at the best of times.

So basically, I'm unsure how to review this. This PR seems way, way too large. What would you recommend?

Ivanidzo4ka:
I've see some changes like:
Microsoft.ML.Runtime.Internal.Utilities -> Microsoft.ML.Internal.Utilities.
In Issue you referencing, I don't see any mentions about removing Runtime from namespace names.
Is it described in another issue? Can you link it?

TomFinley:
Hi @codemzs , given #1956 is in, this seems to be obsolete. Want to close?

codemzs:
@TomFinley Yes, I have closed it.

PR #1938, opened 2018-12-20
Remove IRowCursorConsolidator.

### LINKED ISSUE #1867: Remove IRowCursorConsolidator
Consider this method:

https://github.com/dotnet/machinelearning/blob/9067a1be58434999e37db7504d6b4e06eca8bbf4/src/Microsoft.ML.Core/Data/IDataView.cs#L114-L115

This returns an instance of the mysterious `IRowCursorConsolidator` interface. Why an interface? In retrospect I'm not quite sure.

Having an interface allows for different implementations, but we've never actually really exploited that capability. Nor, even if we were of such a mind to do so, would it be clear how we could. What would they even do differently? The semantics around `Batch` and whatnot are sufficiently clear and simple as to make only one implementation obvious, and even if we did have different implementations since the resulting cursors result most often from transformers (that is, components downstream from the set creation), they couldn't really do anything radically different anyway, since to do anything implementation specific on any cursor would be to break the composability at the core of what makes `IDataView` work at all.

So: get rid of this interface, and replace all usage of it with a simple utility method *somewhere* that can be called to do the reconciliation. It needn't even be a public utility method, but there may be reasons to do so.

TomFinley:
Fixes #1867.

### DIFF
--- a/src/Microsoft.ML.Core/Data/IDataView.cs
+++ b/src/Microsoft.ML.Core/Data/IDataView.cs
@@ -63,7 +63,7 @@ namespace Microsoft.ML.Runtime.Data
 
     /// <summary>
     /// The input and output of Query Operators (Transforms). This is the fundamental data pipeline
-    /// type, comparable to IEnumerable for LINQ.
+    /// type, comparable to <see cref="IEnumerable{T}"/> for LINQ.
     /// </summary>
     public interface IDataView
     {
@@ -92,7 +92,7 @@ namespace Microsoft.ML.Runtime.Data
         RowCursor GetRowCursor(Func<int, bool> needCol, Random rand = null);
 
         /// <summary>
-        /// This constructs a set of parallel batch cursors. The value n is a recommended limit
+        /// This constructs a set of parallel batch cursors. The value <paramref name="n"/> is a recommended limit
         /// on cardinality. If <paramref name="n"/> is non-positive, this indicates that the caller
         /// has no recommendation, and the implementation should have some default behavior to cover
         /// this case. Note that this is strictly a recommendation: it is entirely possible that
@@ -104,16 +104,18 @@ namespace Microsoft.ML.Runtime.Data
         /// but all rows should be returned by exactly one of the cursors returned from this cursor.
         /// The cursors can have their values reconciled downstream through the use of the
         /// <see cref="Row.Batch"/> property.
+        ///
+        /// The typical usage pattern is that a set of cursors is requested, each of them is then
+        /// given to a set of working threads that consume from them independently while, ultimately,
+        /// the results are finally collated in the end by exploiting the ordering of the <see cref="Row.Batch"/>
+        /// property described above. More typical scenarios will be content with pulling from the single
+        /// serial cursor of <see cref="GetRowCursor(Func{int, bool}, Random)"/>.
         /// </summary>
-        /// <param name="consolidator">This is an object that can be used to reconcile the
-        /// returned array of cursors. When the array of cursors is of length 1, it is legal,
-        /// indeed expected, that this parameter should be null.</param>
         /// <param name="needCol">The predicate, where a column is active if this returns true.</param>
         /// <param name="n">The suggested degree of parallelism.</param>
         /// <param name="rand">An instance </param>
         /// <returns></returns>
-        RowCursor[] GetRowCursorSet(out IRowCursorConsolidator consolidator,
-            Func<int, bool> needCol, int n, Random rand = null);
+        RowCursor[] GetRowCursorSet(Func<int, bool> needCol, int n, Random rand = null);
 
         /// <summary>
         /// Gets an instance of Schema.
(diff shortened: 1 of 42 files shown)


src/Microsoft.ML.Core/Data/IDataView.cs
```diff
@@ -122,20 +124,8 @@ public interface IDataView
...
-        /// <summary>
-        /// Create a consolidated cursor from the given parallel cursor set.
-        /// </summary>
-        RowCursor CreateCursor(IChannelProvider provider, RowCursor[] inputs);
-    }
-
-    /// <summary>
-    /// Delegate type to get a value. This can used for efficient access to data in an IRow
-    /// or IRowCursor.
+    /// Delegate type to get a value. This can used for efficient access to data in a <see cref="Row"/>
```
singlis: minor - "can be used"?

src/Microsoft.ML.Core/Data/IDataView.cs
```diff
@@ -146,43 +136,50 @@ public interface IRowCursorConsolidator
...
         public abstract long Position { get; }
 
         /// <summary>
-        /// This provides a means for reconciling multiple streams of counted things. Generally, in each stream,
-        /// batch numbers should be non-decreasing. Furthermore, any given batch number should only appear in one
-        /// of the streams. Order is determined by batch number. The reconciler ensures that each stream (that is
-        /// still active) has at least one item available, then takes the item with the smallest batch number.
+        /// This provides a means for reconciling multiple rows that have been produced generally from
+        /// <see cref="IDataView.GetRowCursorSet(Func{int, bool}, int, Random)"/>. When getting a set, there is a need
+        /// to, while allowing parallel processing to proceed, always have an aim thatthe original order should be
```
singlis: minor - thatthe

src/Microsoft.ML.Core/Data/IDataView.cs
```diff
@@ -146,43 +136,50 @@ public interface IRowCursorConsolidator
...
+        /// reconverable. So: for any cursor implementation, batch numbers should be non-decreasing. Furthermore, any
```
singlis: recoverable?

singlis: Are you saying that the original order should always be maintained?

TomFinley: Given a cursor set, you should be (in principle) able to recover the order that would have been returned from a similarly configured call to the serial `GetRowCursor` by exploiting the `Batch` property, yes. I will perhaps try to make this more clear in the docs.

src/Microsoft.ML.Core/Data/IDataView.cs
TomFinley: Thanks @singlis !

src/Microsoft.ML.Core/Data/IDataView.cs
TomFinley: Yup, thank you!

src/Microsoft.ML.Core/Data/IDataView.cs
TomFinley: Whether I succeeded in clarifying though is another matter, if you could read over it and see if my few additional explanatory sentences help or hurt, I'd appreciate it, thanks @singlis !

src/Microsoft.ML.Data/Data/DataViewUtils.cs
```diff
@@ -299,214 +294,198 @@ private Splitter(Schema schema)
...
                 }
+                int idIdx = activeToCol.Length + (int)ExtraIndex.Id;
+                outPipes[idIdx] = OutPipe.Create(NumberType.UG, GetPool(NumberType.UG, ourPools, idIdx));
 
-                private static RowCursor ConsolidateCore(IChannelProvider provider, RowCursor[] inputs, ref object[] ourPools, IChannel ch)
+                // Create the structures to synchronize between the workers and the consumer.
+                const int toConsumeBound = 4;
+                var toConsume = new BlockingCollection<Batch>(toConsumeBound);
+                var batchColumnPool = new MadeObjectPool<BatchColumn[]>(() => new BatchColumn[outPipes.Length]);
+                Thread[] workers = new Thread[inputs.Length];
```
singlis: So threads are used here, can the cursor set be large? Like in the hundreds for example? If so, would tasks be a better alternative?

singlis:
Looks good to me - I didnt quite follow the "Scott document review".

src/Microsoft.ML.Data/Data/DataViewUtils.cs
```diff
@@ -129,15 +129,13 @@ public static int GetThreadCount(IHost host, int num = 0, bool preferOne = false
...
-            var inputs = view.GetRowCursorSet(out consolidator, predicate, cthd, rand);
+            var inputs = view.GetRowCursorSet(predicate, cthd, rand);
             host.Check(Utils.Size(inputs) > 0);
-            host.Check(inputs.Length == 1 || consolidator != null);
 
             if (inputs.Length == 1)
                 curs = inputs[0];
             else
-                curs = consolidator.CreateCursor(host, inputs);
+                curs = DataViewUtils.ConsolidateGeneric(host, inputs, 64);
```
sfilipi: >64 [](start = 70, length = 2)

maybe have a comment for why 64.

PR #1944, opened 2018-12-20
ONNX conversion should once a step is not convertible to ONNX

### LINKED ISSUE #1396: An internal SaveAsOnnx test fails
The test
```
        public void OnnxNormalize()
        {
            // Some normalize types are not supported. These will be ignored.
            // If they are supported in the future, then this baseline will
            // need to be updated.
            InitOnnxPath(out string onnxName, out string onnxPath, out string jsonPath, out string onnxArg);
            string pathData = GetDataPath(@"..\UCI\iris", "iris.data");
            MainForTest("saveonnx odrop=A domain=ONNX loader=text{col=A:0-3 sep=comma} " +
                "xf=minmax{col=MMZ:A zero-} xf=minmax{col=MM:A} " +
                "xf=meanvar{col=MV:A} xf=lognormal{col=LN:A} xf=bin{col=BIN:A} " +
                $"data={{{pathData}}}" +
                onnxArg);

            CheckEqualityNormalized(OutRoot, onnxName);
            OnnxSmokeTest(onnxPath);
            Done();
        }
```
fails because a null string `variableName`, is fed into
```
ctx.CreateNode("Identity", variableName, trueVariableName, ctx.GetNodeName("Identity"), "");
```
as its input name.

wschin:
Fixes #1396.


src/Microsoft.ML.Onnx/SaveOnnxCommand.cs
```diff
@@ -243,6 +243,9 @@ private void Run(IChannel ch)
                     continue;
 
                 var variableName = ctx.TryGetVariableName(idataviewColumnName);
+                // Null variable name occurs when an unsupported transform produces an output and a downsteam step consumes that output.
+                // The reason is that unsupported transforms are not able to produce outputs in ONNX graph.
+                ch.Check(variableName != null, "The targeted pipeline can not be fully converted into ONNX format.");
```
Ivanidzo4ka: Why we doing this here?
We have pipeline and we go through it step by step and call CanSaveAsOnnx on it, right?
If one of steps can't be converted we should stop.
How you pass line 228 to get here?
Can we do this in line 228 instead of this?

wschin: In its current design, user can drop some input columns so that the conversion of some transforms will fail because of missing input columns.

Ivanidzo4ka: Wait, we have CanSaveOnnx and SaveAsOnnx functions.
I always though first one returns you true or false based on can it actually be saved or not, without checking any content of context and what variable available or not.
Second one can throw you exception if something not found.
Am I wrong?
If whole purpose of you PR just to notify user what one step in pipeline is not convertible, why not do that in line 282?

wschin: ICanSaveOnnx returning true doesn't mean we will convert that transform. We convert a transform if two conditions are satisfied: (1) input columns are not dropped by user (2) the considered transform is convertible. That means, `something not found` sometime is asked by the user so we can't throw until the very end of the conversion.

Ivanidzo4ka: And what is reason why we don't throw exception during SaveAsOnnx?
>Simply went to all SaveAsOnnx and SaveAsOnnxCore and add throwing mechanism if proper.
From your description for this PR I expected what that would be code, but instead you checking this after execution artifacts.

wschin: That was true before rebasing. I realized that those checks don't fit the existing framework's assumption. User can discard some transforms but we don't know if the pruned graph is legal when converting individual transform. There are two reasons of SaveAsOnnx returning false: (1) caused by users so we can't throw because the resulted graph can still be legal (2) caused by unsupported transform.

That's a disadvantage of mixing `conversion` and `parsing` in the same phase. In Onnxmltools, we have parsing stage followed by conversion stage, so conversion will not happen if parsing algorithm doesn't produce a legal graph.

wschin:
@Ivanidzo4ka, is the new error message better?

PR #1946, opened 2018-12-20
Throwing exception if the underlying rsp fails with an exception

Anipik:
The benchmarks involving rsps don't fail even when the underlying rsp has thrown an exception

### DIFF
--- a/test/Microsoft.ML.Benchmarks/Helpers/ExecuteMaml.cs
+++ b/test/Microsoft.ML.Benchmarks/Helpers/ExecuteMaml.cs
@@ -1,0 +1,18 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.ML.Tools;
+using System;
+
+namespace Microsoft.ML.Benchmarks
+{
+    internal static class ExecuteMaml
+    {
+        public static void ExecuteMamlCommand(this string command, MLContext environment)
+        {
+            if (Maml.MainCore(environment, command, alwaysPrintStacktrace: false) < 0)
+                throw new Exception($"Command {command} returned negative error code");
+        }
+    }
+}
--- a/test/Microsoft.ML.Benchmarks/Numeric/Ranking.cs
+++ b/test/Microsoft.ML.Benchmarks/Numeric/Ranking.cs
@@ -2,14 +2,13 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.IO;
 using BenchmarkDotNet.Attributes;
 using Microsoft.ML.Data;
 using Microsoft.ML.LightGBM;
 using Microsoft.ML.RunTests;
-using Microsoft.ML.Tools;
 using Microsoft.ML.Trainers.FastTree;
 using Microsoft.ML.Transforms.Conversions;
+using System.IO;
 
 namespace Microsoft.ML.Benchmarks
 {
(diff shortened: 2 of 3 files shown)


Ivanidzo4ka:
Out of curiosity,  why we even use rsp?

Anipik:
>  Out of curiosity, why we even use rsp?

Because  when we introduced benchmarking, we wanted to have them asap. And converting them to api required someone on the ml team and it was not possible at that time

Anipik:
cc @danmosemsft @justinormont @adamsitnik

TomFinley:
Hi @Anipik thanks for writing this.

Minor point... so, I see this appears to be part of something called a `Dispose` method. Could we name it something else? I think naming it this when it's not actually a dispose method violates some sort of .NET design guideline (despite the fact that this does not implement `IDisposable`). It's at least needlessly confusing.

While we're at it, is there a reason to not make this base class abstract?

test/Microsoft.ML.Benchmarks/Helpers/BaseBenchClass.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Microsoft.ML.Benchmarks
+{
+   public class BaseBenchClass
+    {
+        public int _passed;
+
+        [IterationCleanup]
+        public void Dispose()
+        {
+            if (_passed < 0)
```
justinormont: FYI to any readers...

The negative return value from `Maml.MainCore()` is used due to AEther expecting a negative value:
https://github.com/dotnet/machinelearning/blob/d885ebc49c1fff04209714bb2757166a6f2681c9/src/Microsoft.ML.Maml/MAML.cs#L209-L210

Anipik: is AEther a tool ?

Anipik:
@justinormont @Ivanidzo4ka can you review this one ?

test/Microsoft.ML.Benchmarks/Helpers/BaseBenchClass.cs
```diff
@@ -0,0 +1,23 @@
...
+    public abstract class BaseBenchClass
+    {
+        public int _passed;
+
+        [IterationCleanup]
+        public void CheckForInnerException()
+        {
+            if (_passed < 0)
+            {
+                throw new Exception($"The benchmark iteration threw an unhandled exception");
```
adamsitnik: can we get some more info here? any details about the exception? or make Maml just throw instead of returning error codes?

test/Microsoft.ML.Benchmarks/Helpers/BaseBenchClass.cs
justinormont: Correct, it's a tool calling this code which drove the odd return values from `Maml.MainCore()`.

test/Microsoft.ML.Benchmarks/Helpers/BaseBenchClass.cs
Anipik: @TomFinley is it okay to change it to  maml throwing instead of benchmarks ?

TomFinley: It is for various reasons it may not be best to describe here not acceptable for maml itself to throw.

test/Microsoft.ML.Benchmarks/Helpers/ExecuteMaml.cs
```diff
@@ -0,0 +1,18 @@
...
+using Microsoft.ML.Tools;
+using System;
+
+namespace Microsoft.ML.Benchmarks
+{
+    internal static class ExecuteMaml
+    {
+        public static void ExecuteMamlCommand(this string command, MLContext environment)
+        {
+            if (Maml.MainCore(environment, command, alwaysPrintStacktrace: false) < 0)
```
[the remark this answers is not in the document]
Anipik: yes 
if no exception is thrown :- 
https://github.com/dotnet/machinelearning/blob/a570da14a41f2870eb8f61d84496a58422398253/src/Microsoft.ML.Maml/MAML.cs#L142
if exception is thrown :- https://github.com/dotnet/machinelearning/blob/a570da14a41f2870eb8f61d84496a58422398253/src/Microsoft.ML.Maml/MAML.cs#L209

These are the only 2 places where we are setting the return value

Issue #1104, opened 2018-10-01
How to use EntryPoint saved .zip files with the "new API"?


eerhardt:
### System information

- **OS version/distro**: all
- **.NET Version (eg., dotnet --info)**:  
```
> dotnet --info
.NET Core SDK (reflecting any global.json):
 Version:   3.0.100-alpha1-009622
 Commit:    5a14412a49

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17134
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.0.100-alpha1-009622\

Host (useful for support):
  Version: 3.0.0-preview1-26924-01
  Commit:  4b8641b848
```

### Issue

- **What did you do?**
Attempt to load a .zip file that was created by EntryPoints using `TransformerChain.LoadFrom`.

- **What happened?**
```
An unhandled exception was thrown by the application.
Microsoft.Azure.WebJobs.Host.FunctionInvocationException: Exception while executing function: Function1 ---> System.FormatException: Corrupt model file
   at Microsoft.ML.Runtime.Model.ModelLoadContext.LoadModel[TRes,TSig](IHostEnvironment env, TRes& result, RepositoryReader rep, String dir, Object[] extra)
   at Microsoft.ML.Runtime.Data.TransformerChain.LoadFrom(IHostEnvironment env, Stream stream)
   at MLFunction.Predictor.PredictAsync(GitHubIssue issue, TraceWriter log) in F:\git\MLFunction\Predictor.cs:line 69
```

- **What did you expect?**
I expected to be able to use a .zip file that was created using the EntryPoints API with the "new API".

### Notes

If we are keeping the EntryPoints API, users will be creating model files using EntryPoints. We need to have an API that allows C#/.NET developers to use those model files.

GalOshri:
Let's investigate to see if one of the new APIs solve this.

Zruty0:
As far as I know, our model formats slightly diverge at this point. We plan to fix that soon.

Zruty0:
@Ivanidzo4ka could you quickly look into that? If this is a 4-hour fix, we should make it before the upcoming release.

Ivanidzo4ka:
Can we get clear what we want?
Entry points right now can still use functionality which didn't get converted to IEstimator/ITransformer. 
So it's impossible to always save entry points model as transformer chain (although it's possible sometimes)
[...]

PR #993, opened 2018-09-22
Extended contexts to regression and multiclass, added FFM pigstension

Zruty0:
Part of #754, extends #949.
Added MulticlassClassification context and RegressionContext, with corresponding Evaluate methods. Also added FFM binary trainer to the context extensions.

PR #1951, opened 2018-12-21
Load entry point models

Ivanidzo4ka:
Address https://github.com/dotnet/machinelearning/issues/1104
As long as NimbusML uses old version ml.net (pre 0.9) I don't have other way around other than this hacky way of loading it.


src/Microsoft.ML.Data/DataLoadSave/TransformerChain.cs
```diff
@@ -252,8 +252,47 @@ public static TransformerChain<ITransformer> LoadFrom(IHostEnvironment env, Stre
...
-                ModelLoadContext.LoadModel<TransformerChain<ITransformer>, SignatureLoadModel>(env, out var transformerChain, rep, LoaderSignature);
-                return transformerChain;
+                try
+                {
+                    ModelLoadContext.LoadModel<TransformerChain<ITransformer>, SignatureLoadModel>(env, out var transformerChain, rep, LoaderSignature);
+                    return transformerChain;
+                }
+                catch
+                {
+                    var chain = ModelFileUtils.LoadPipeline(env, stream, new MultiFileSource(null), extractInnerPipe: false);
```
TomFinley: Usually when you do something that has a bad code smell (e.g., some sort of global `try/catch` of literally any type of exception ***whatsoever***) it is either best to either (1) explain why it must be so or (2) try not to do it. :) Either is fine with me.

src/Microsoft.ML.Data/DataView/RowToRowMapperTransform.cs
```diff
@@ -44,6 +45,11 @@ internal interface IRowMapper : ICanSaveModel
         /// Returns information about the output columns, including their name, type and any metadata information.
         /// </summary>
         Schema.DetachedColumn[] GetOutputColumns();
+
+        /// <summary>
+        /// Returns parent transfomer which uses this mapper.
+        /// </summary>
+        ITransformer GetTransformer();
```
TomFinley: So, this one strikes me as a little mysterious. The reason why `IRowMapper` exists *at all* is to make constructing a single `ITransformer` more convenient -- it is analogous in some respects, in this way, to a more serious `IValueMapper`, and a more specific mapper than `IRowToRowMapper`. So why does *this* specific thing need it? I suspect that it doesn't really belong here.

Ivanidzo4ka: I'm just trying to achieve my goal of loading models in old format in less invasive way.
Right now I can load CompositeLoader which contains list of RowToRowMapperTransforms, (I don't even know is it Normalize, Pca, or any other transform, just RowToRowMapperTransform) if transform was converted, or just instance of IDataTransform.
If it's RowToRowMapperTransform, I need a way to get real ITransformer out of it, and mapper (which is IRowMapper) is only thing which I can use.

So basically since Mapper is pretty much all the time know it's parent (ITransformer) I'm just adding this method to it in order to get it back during walk through CompositeLoader transforms.



---
In reply to: [244916610](https://github.com/dotnet/machinelearning/pull/1951#discussion_r244916610) [](ancestors = 244916610)

TomFinley: Sorry @ivanidzo4ka if I have missed something -- that's entirely possible -- but what does that have to do with this interface? All that tells me is that by using `CompositeDataLoader` we used the wrong class, or *it* is using the wrong class, or something. I mean, if I had a structure that took `IEnumerable`, but then I realized I wanted in fact the `IList` or whatever, I would think, "ok, this structure should take an `IList` instead!" I wouldn't think, "gee, can't possibly change the structure, I should change `IEnumerable` to produce the `IList` which it really is, since practically that's going to be the source of the `IEnumerable` all the time." I mean, that just seems backwards.

In addition to that, it seems contrary to our plans for how we've previously discussed how to load old models. The plan, as I recall, was the following: things that had been saved using the old (now disfavored) `IDataTransform` interface models, we will attempt to load as the new, shiny, `ITransformer` implementations. So: a normalizer (for instance) saved as `IDataTransform` in old version, we will read that form as an `ITransformer`. If it is something we have deliberately chosen *not* to convert to an `ITransform` (e.g., a filter), then it is dropped. (Pursuant to #993.) We would load them as `ITransformer` or nothing. We wouldn't have some "liminal" state where we use the old structures (which are being retired!) before converting them to the "real" structures we actually want. We would just use the new structures directly.

Now, from what I see, we are not doing that. Instead, I see we are *keeping* the `IDataTransform` instances and all the other baggage associated with them, and continue to rely on their presence. I am in fact going to keep using not only `IDataTransform`, but this old `IDataLoader` interface implementation `CompositeDataLoader`. But, because what we really wanted to do was load `ITransform` instances, we are now providing a bunch of backdoors through these created objects to recover the original `ITransform` instances upon which they are now based, instead of just using `ITransformer` directly and being done with it. I mean, why is `IDataTransform` or `IDataLoader` entering the picture *at all*? Doesn't that just complicate our ultimate desire to do away with them? (Pursuant to #1995.)

I get that you wanted to get away with just re-using the old `CompositeDataLoader`, and so had to make a bunch of old compromises because of that choice, but the correct thing to do was not use it, is it not so?

This seems architecturally a little backwards to me.

---
In reply to: [245146290](https://github.com/dotnet/machinelearning/pull/1951#discussion_r245146290) [](ancestors = 245146290,244916610)

TomFinley: Ah OK, so this is intended to be a temporary situation. Could we somehow clarify this on `IRowMapper` interface? It just seems so odd, and the comment is not describing its "temporary nature" too well at the moment.

Zruty0:
Is this work being abandoned? I viewed this as quite important, since Nimbus models were not loadable with ML.NET. Has that been updated?

src/Microsoft.ML.Data/DataView/RowToRowMapperTransform.cs
Ivanidzo4ka: I change description and modified comment to state what this is temporary solution.

sfilipi: @ivmatan@microsoft.com the world makes more sense to me after reading this comment. Shall we save this explanation as code comments in the TransformerChain.LoadFrom?

---
In reply to: [245194301](https://github.com/dotnet/machinelearning/pull/1951#discussion_r245194301) [](ancestors = 245194301)

sfilipi:
@ivmatan@microsoft.com i see that you added a bunch of models, in the test/data/backcompat . Did you intend to use them in tests?

PR #1953, opened 2018-12-22
Add an example for static pipeline with in-memory data and show how to get class probabilities

### LINKED ISSUE #1881: Multi classification - Probability
Hello,
I have an application that use using ml.net multiclassification trainer to predict a category. However, it seems as though we remove the TryGetScoreLabelNames() method from the library. Basically, the application would like to output as well how confident (in percentage) is the predicted label. How can I achieve that in ML.NET 0.8?

### LINKED ISSUE #1947: Need Example of Generating DataView and Feeding It to Static Pipeline
All of example (static) pipelines start with a text file, which is not very true when deploying models to apps. We need to create examples using in-memory C# data structures.

wschin:
Fixes #1947 and fixes #1881.

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
+++ b/docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
@@ -1,5 +1,6 @@
 ﻿using Microsoft.ML.Data;
 using Microsoft.ML.LightGBM.StaticPipe;
+using Microsoft.ML.SamplesUtils;
 using Microsoft.ML.StaticPipe;
 using System;
 using System.Collections.Generic;
@@ -9,71 +10,6 @@ namespace Microsoft.ML.Samples.Static
 {
     class LightGBMMulticlassWithInMemoryData
     {
-        /// <summary>
-        /// Number of features per example used in <see cref="MultiClassLightGbmStaticPipelineWithInMemoryData"/>.
-        /// </summary>
-        private const int _featureVectorLength = 10;
-
-        /// <summary>
-        /// Data point used in <see cref="MultiClassLightGbmStaticPipelineWithInMemoryData"/>. A data set there 
-        /// is a collection of <see cref="NativeExample"/>.
-        /// </summary>
-        private class NativeExample
-        {
-            [VectorType(_featureVectorLength)]
-            public float[] Features;
-            [ColumnName("Label")]
-            public string Label;
-            public uint LabelIndex;
-            public uint PredictedLabelIndex;
-            [VectorType(4)]
-            // The probabilities of being "AA", "BB", "CC", and "DD".
-            public float[] Scores;
-
-            public NativeExample()
-            {
-                Features = new float[_featureVectorLength];
-            }
-        }
-
-        /// <summary>
-        /// Helper function used to generate <see cref="NativeExample"/>s.
-        /// </summary>
-        private static List<NativeExample> GenerateRandomExamples(int count)
-        {
-            var examples = new List<NativeExample>();
-            var rnd = new Random(0);
-            for (int i = 0; i < count; ++i)
-            {
-                var example = new NativeExample();
-                var res = i % 4;
-                // Generate random float feature values.
-                for (int j = 0; j < _featureVectorLength; ++j)
-                {
-                    var value = (float)rnd.NextDouble() + res * 0.2f;
-                    example.Features[j] = value;
-                }
-
-                // Generate label based on feature sum.
-                if (res == 0)
-                    example.Label = "AA";
-                else if (res == 1)
-                    example.Label = "BB";
-                else if (res == 2)
-                    example.Label = "CC";
-                else
-                    example.Label = "DD";
-
-                // The following three attributes are just placeholder for storing prediction results.
-                example.LabelIndex = default;
-                example.PredictedLabelIndex = default;
-                example.Scores = new float[4];
-
-                examples.Add(example);
-            }
-            return examples;
-        }
-
         public void MultiClassLightGbmStaticPipelineWithInMemoryData()
         {
             // Create a general context for ML.NET operations. It can be used for exception tracking and logging,
@@ -81,7 +17,7 @@ namespace Microsoft.ML.Samples.Static
             var mlContext = new MLContext();
 
             // Create in-memory examples as C# native class.
-            var examples = GenerateRandomExamples(1000);
+            var examples = DatasetUtils.GenerateRandomMulticlassClassificationExamples(1000);
 
             // Convert native C# class to IDataView, a consumble format to ML.NET functions.
             var dataView = ComponentCreation.CreateDataView(mlContext, examples);
@@ -132,7 +68,7 @@ namespace Microsoft.ML.Samples.Static
             Console.WriteLine ("Macro accuracy: {0}, Micro accuracy: {1}.", 0.863482146891263, 0.86309523809523814);
 
             // Convert prediction in ML.NET format to native C# class.
-            var nativePredictions = new List<NativeExample>(prediction.AsDynamic.AsEnumerable<NativeExample>(mlContext, false));
+            var nativePredictions = new List<DatasetUtils.MulticlassClassificationExample>(prediction.AsDynamic.AsEnumerable<DatasetUtils.MulticlassClassificationExample>(mlContext, false));
 
             // Get schema object out of the prediction. It contains metadata such as the mapping from predicted label index
             // (e.g., 1) to its actual label (e.g., "AA"). The call to "AsDynamic" converts our statically-typed pipeline into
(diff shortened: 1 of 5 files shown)


TomFinley:
This may be worthwhile as an example, but it would be better if we could create a static pipeline out of a type *directly*, since the entire purpose of the static pipeline idiom is type safety and compile time checks, which *should* be achievable, and yet which this does not do. So perhaps a note in the sample that it is not intended to be a permanent solution?

wschin:
@TomFinley, I added your comment to an [inline comment](https://github.com/dotnet/machinelearning/pull/1953/commits/31d226913ccdfdfa96fd04201ded9b838f921947). Do you think it's ok now?

docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
```diff
@@ -0,0 +1,163 @@
...
+            public NativeExample()
+            {
+                Features = new float[_featureVectorLength];
+            }
+        }
+
+        /// <summary>
+        /// Helper function used to generate <see cref="NativeExample"/>s.
+        /// </summary>
+        private static List<NativeExample> GenerateRandomExamples(int count)
```
sfilipi: >   private static List<NativeExample> GenerateRandomExamples(int count) [](start = 5, length = 71)

I'd move this in the SampleUtils project, and reference it from the test, and from here.

docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
```diff
@@ -0,0 +1,163 @@
...
+                examples.Add(example);
+            }
+            return examples;
+        }
+
+        public void MultiClassLightGbmStaticPipelineWithInMemoryData()
+        {
+            // Create a general context for ML.NET operations. It can be used for exception tracking and logging,
+            // as a catalog of available operations and as the source of randomness.
+            var mlContext = new MLContext(seed: 1, conc: 1);
```
sfilipi: >seed: 1, conc: 1 [](start = 42, length = 16)

let's omit those in samples, and go with defaults as much as we can.

docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
```diff
@@ -0,0 +1,163 @@
...
+
+            // Context for calling static classifiers. It contains constructors of classifiers and evaluation utilities.
+            var ctx = new MulticlassClassificationContext(mlContext);
```
sfilipi: remove this, and use mlContext.MulticlassClassificationContext further down

docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
```diff
@@ -0,0 +1,163 @@
...
+            var schema = prediction.AsDynamic.Schema;
+
+            // Retrieve the mapping from labels to label indexes.
+            var labelBuffer = new VBuffer<ReadOnlyMemory<char>>(); 
+            schema[nameof(NativeExample.PredictedLabelIndex)].Metadata.GetValue("KeyValues", ref labelBuffer);
+            var nativeLabels = labelBuffer.DenseValues().ToList(); // nativeLabels[nativePrediction.PredictedLabelIndex-1] is the original label indexed by nativePrediction.PredictedLabelIndex.
+
+            // Show prediction result for the 3rd example.
+            var nativePrediction = nativePredictions[2];
+            Console.WriteLine("Our predicted label to this example is {0} with probability {1}",
```
sfilipi: > Console.WriteLine(" [](start = 11, length = 20)

have a comment for the WriteLines showing here directly what the output would  be.

docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
```diff
@@ -0,0 +1,163 @@
...
+                example.LabelIndex = default;
+                example.PredictedLabelIndex = default;
+                example.Scores = new float[4];
+
+                examples.Add(example);
+            }
+            return examples;
+        }
+
+        public void MultiClassLightGbmStaticPipelineWithInMemoryData()
```
sfilipi: >MultiClassLightGbmStaticPipelineWithInMemoryData [](start = 20, length = 48)

i think most of the emphasis in 1.0 is on the dynamic API. Would this example add value as a dynamic sample?

docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
```diff
@@ -0,0 +1,163 @@
+﻿using Microsoft.ML.Data;
+using Microsoft.ML.LightGBM.StaticPipe;
+using Microsoft.ML.StaticPipe;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.ML.Samples.Static
+{
+    class LightGBMMulticlassWithInMemoryData
```
sfilipi: >LightGBMMulticlassWithInMemoryData [](start = 10, length = 34)

this needs to be referenced from some other file, otherwise it won't display in the documentation. 

Maybe from the LightGBMStatics catalog, if we have one already?

docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
wschin: Maybe in another PR if really needed.

---
In reply to: [244876215](https://github.com/dotnet/machinelearning/pull/1953#discussion_r244876215) [](ancestors = 244876215)

docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
wschin: Sure.

---
In reply to: [244875689](https://github.com/dotnet/machinelearning/pull/1953#discussion_r244875689) [](ancestors = 244875689)

docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
wschin: No problem. Will do the same for Line 159 below.

---
In reply to: [244876086](https://github.com/dotnet/machinelearning/pull/1953#discussion_r244876086) [](ancestors = 244876086)

docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
wschin: I am not quite following. Do you have any example? In addition, could we reference a sample from our main library? I thought sample project and ML.NET are independent?

---
In reply to: [244876344](https://github.com/dotnet/machinelearning/pull/1953#discussion_r244876344) [](ancestors = 244876344)

docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
wschin: If we move this to SamplesUtils, then user will not able to see a full example in doc.

---
In reply to: [244875572](https://github.com/dotnet/machinelearning/pull/1953#discussion_r244875572) [](ancestors = 244875572)

TomFinley:
> using Microsoft.ML.Data;

Header?

---
Refers to: docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs:1 in 5aa58c3. [](commit_id = 5aa58c35930b894d839091875a56ae1192bc71da, deletion_comment = False)

TomFinley:
Yeah, that'll do.

---
In reply to: [450924472](https://github.com/dotnet/machinelearning/pull/1953#issuecomment-450924472) [](ancestors = 450924472)

docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
sfilipi: i believe the main purpose of those examples in the docs are to get copy/pasted in a VS project, add the necessary references and study it there, otherwise int he web page many things don't make sense. 

There is also guideline to keep those examples small, and too the point, otherwise their length is an immediate  turnoff for average user. 

---
In reply to: [245743767](https://github.com/dotnet/machinelearning/pull/1953#discussion_r245743767) [](ancestors = 245743767,244875572)

wschin:
> using Microsoft.ML.Data;

Example files usually do contain a header because they could be encapsulated in another class' doc string. 

---
In reply to: [452028645](https://github.com/dotnet/machinelearning/pull/1953#issuecomment-452028645) [](ancestors = 452028645)

---
Refers to: docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs:1 in 5aa58c3. [](commit_id = 5aa58c35930b894d839091875a56ae1192bc71da, deletion_comment = False)

docs/samples/Microsoft.ML.Samples/Static/LightGBMMulticlassWithInMemoryData.cs
wschin: Ok. Will move NativeExample and its generator to SamplesUtils.

---
In reply to: [245796925](https://github.com/dotnet/machinelearning/pull/1953#discussion_r245796925) [](ancestors = 245796925,245743767,244875572)

PR #1963, opened 2018-12-24
Sort namespaces as per stylecop rules/.net convention and remove unused namespaces from source files.

### LINKED ISSUE #1961: Sort namespaces in source files as per Microsoft .NET convention
Parent issue #1697 

StyleCop rules:
**SA1210:** Using directives must be ordered alphabetically by namespace.
**SA1208:** System using directives must be placed before other using directives.

### LINKED ISSUE #1962: Remove unused namespaces in source files.

codemzs:
Sorts namespaces as per stylecop rules/.net convention and removes unused namespaces from source files. Standardizes namespace to be always sorted as per .NET convention in source files.

My impressions after doing this change:
1) There is decent amount of code under IF DEFs that seems dead and should be removed, example: 
In IntArray.cs
#if USE_SINGLE_PRECISION
    using FloatType = System.Single;
#else
    using FloatType = System.Double;
#endif

USE_SINGLE_PRECISION is not defined anywhere and FloatType is not at all used anywhere in the file. We should consider opening an issue to do this sort of clean up. There are many such examples.

2) Consider removing Float and use float directly.   

CC: @TomFinley 

fixes #1961
fixes #1962


src/Microsoft.ML.Core/Utilities/MathUtils.cs
```diff
@@ -2,11 +2,9 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using Float = System.Single;
-
 using System;
 using System.Collections.Generic;
-using System.Linq;
+using Float = System.Single;
```
justinormont: Should we rename `Float` in the file to `Single`? It seems odd having our own alias of `float`/`Single`.

justinormont: Seems we do this in many of the file... (not just this one file)

codemzs: May be we should but does it make sense to be it owns change so that its easier on the reviewers? Apparently there are 54 files where Float = System.Single is being used so I suspect removing them all and changing all their references to float might pollute this change. Lets see what @TomFinley feels? I'm fine either ways.

codemzs:
@sharwell This is great feedback and I was thinking the same how we could standardize, thanks. Do you by any chance have an example of a repo that contains such standardization?

codemzs:
@sharwell Thanks for reaching out on IM today evening and helping me with standardizing stylecop. I found some errors using it that I had earlier missed, not sure how even though I used VS's power tools extensions to sort namespaces and remove unused ones.

src/Microsoft.ML.FastTree/SumupPerformanceCommand.cs
```diff
@@ -15,16 +15,16 @@
...
 using Microsoft.ML.Trainers.FastTree.Internal;
-using Microsoft.ML.Internal.Utilities;
 
 [assembly: LoadableClass(typeof(SumupPerformanceCommand), typeof(SumupPerformanceCommand.Arguments), typeof(SignatureCommand),
     "", "FastTreeSumupPerformance", "ftsumup")]
 
 namespace Microsoft.ML.Trainers.FastTree
 {
-    using Stopwatch = System.Diagnostics.Stopwatch;
+using Stopwatch = System.Diagnostics.Stopwatch;
```
sharwell: :confused: not sure how this indentation change happened.

src/Microsoft.ML.Core/Utilities/MathUtils.cs
TomFinley: It's unclear to me why we're bringing that up in this PR. There is no urgency to that change. We've been doing it opportunistically. Let's continue that policy. Doing it all at once would cause a great deal of conflicts with existing changes for no particular reason whatsoever. (These changes being at the *header* of the file are at least easier to manage.)

TomFinley:
Hi @codemzs, may I ask how we automated this change? My reasons for wondering are, I feel like in a PR like this it would be more helpful to review *how* we automated, rather than review every single file.

codemzs:
Hi @TomFinley , I used VS’s powertools extension and ran it on every solution, compiled it, fixed extra lines errors and repeated the process. Lastly I  enforced namespaces be sorted via editor configuration using the method @sharwell suggested, this also verified the namespaces are sorted as per .NET standards outlined in this PR’s description.

codemzs:
Thanks @Anipik for reviewing!

src/Microsoft.ML.Core/Utilities/MathUtils.cs
justinormont: Agreed. 

I initially didn't notice `using Float = System.Single;` was used in many files. If it was only in one, I'd ask for the change. It's a larger tasks which belongs in its own PR.

Issue #1529, opened 2018-11-05
IDataView Cleanup: Predicates from int to Column


TomFinley:
As seen in #1500, schema is being changed so that schemas contain columns.

For various reasons it may be easiest once `IDataView` is a class.

Let us consider the use of predicates in the `IDataView` system, e.g., when getting a cursor:

https://github.com/dotnet/machinelearning/blob/f9202628fbfac9e599e8c63dc5ed26eae77afbee/src/Microsoft.ML.Core/Data/IDataView.cs#L103

or elsewhere when forming a mapper, and getting dependencies:

https://github.com/dotnet/machinelearning/blob/f9202628fbfac9e599e8c63dc5ed26eae77afbee/src/Microsoft.ML.Core/Data/ISchemaBindableMapper.cs#L77

The use of integer indices here has sometimes led to confusion or even bugs. With the change of #1500 under consideration, this suggests a possibly better way.

It may be worth considering whether the columns in the new scheme suggested in #1500 should have a backref to the original schema (even as an internal field that is checked by the data-view abstract class), so as to enable an easy way to check whether that column in fact came from that schema, or, even without that backref, to check whether the columns exist.

We could also consider this dependency be expressed not as a *delegate*, but instead just some sort of collection of columns, since that would also make this easier to explain.

Note that while this makes the *interface* to `IDataView` easier, it makes the *implementation* harder, at least, if we suppose that all dataviews are possible for handling these inputs correctly and verifying that there aren't any shenaningans going on with input columns being from a different schema (which we can and so almost certainly should do under this new scheme). This suggests a change to `IDataView`, possibly done once `IDataView` is a class, so that the utility mapping from these column objects back down to indices (which must still happen internally) is handled by common code. It would also enable if these column objects have some sort of internal backreference to the schema, the ability to check that the input schemas are in fact correct. (This we obviously cannot do today with indices!)

/cc @Zruty0 @shauheen @terrajobst

eerhardt:
@TomFinley - does this also apply to these functions?

https://github.com/dotnet/machinelearning/blob/c90fa51fff86a5b44ff9184d3c22356d6f3fd440/src/Microsoft.Data.DataView/IDataView.cs#L134-L144

If not, and we want to keep these functions taking an `int`, we should rename the parameters to `columnIndex` instead of `col`.

sfilipi:
@eerhardt thanks for the note. I'll make the change on the PR that will close this issue.

TomFinley:
[...]

PR #1975, opened 2018-12-28
Initial refactor to swap the predicates for an IEnumerable<Schema.Columns> in the GetRowCursor and GetRowCursorSet

sfilipi:
Addresses #1529 by swapping the predicates for IEnumerable<Schema.Columns> in the GetRowCursor and GetRowCursorSet

### DIFF
--- a/src/Microsoft.ML.Core/Data/IDataView.cs
+++ b/src/Microsoft.ML.Core/Data/IDataView.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 
 namespace Microsoft.ML.Data
@@ -32,8 +33,8 @@ namespace Microsoft.ML.Data
         /// <summary>
         /// Get a row cursor. The active column indices are those for which needCol(col) returns true.
         /// The schema of the returned cursor will be the same as the schema of the IDataView, but getting
-        /// a getter for an inactive columns will throw. The <paramref name="colsNeeded"/> predicate must be
-        /// non-null.If set to <langword>null</langword> no column is active.
+        /// a getter for inactive columns will throw. The <paramref name="colsNeeded"/> indicate the columns that are needed
+        /// to iterate over.If set to an empty <see cref="IEnumerable"/> no column is requested.
         /// </summary>
         RowCursor GetRowCursor(IEnumerable<Schema.Column> colsNeeded, Random rand = null);
 
@@ -55,9 +56,9 @@ namespace Microsoft.ML.Data
         /// scenarios will be content with pulling from the single serial cursor of
         /// <see cref="GetRowCursor(IEnumerable{Schema.Column}, Random)"/>.
         /// </summary>
-        /// <param name="colsNeeded">The active columns needed. If set to <langword>null</langword> no column is active.</param>
+        /// <param name="colsNeeded">The active columns needed. If passed an empty <see cref="IEnumerable"/> no column is requested.</param>
         /// <param name="n">The suggested degree of parallelism.</param>
-        /// <param name="rand">An instance </param>
+        /// <param name="rand">An instance of <see cref="Random"/> to seed randomizing the access.</param>
         /// <returns></returns>
         RowCursor[] GetRowCursorSet(IEnumerable<Schema.Column> colsNeeded, int n, Random rand = null);
 
--- a/src/Microsoft.ML.Core/Data/IDataView.cs
+++ b/src/Microsoft.ML.Core/Data/IDataView.cs
@@ -33,10 +33,10 @@ namespace Microsoft.ML.Data
         /// <summary>
         /// Get a row cursor. The active column indices are those for which needCol(col) returns true.
         /// The schema of the returned cursor will be the same as the schema of the IDataView, but getting
-        /// a getter for inactive columns will throw. The <paramref name="colsNeeded"/> indicate the columns that are needed
+        /// a getter for inactive columns will throw. The <paramref name="columnsNeeded"/> indicate the columns that are needed
         /// to iterate over.If set to an empty <see cref="IEnumerable"/> no column is requested.
         /// </summary>
-        RowCursor GetRowCursor(IEnumerable<Schema.Column> colsNeeded, Random rand = null);
+        RowCursor GetRowCursor(IEnumerable<Schema.Column> columnsNeeded, Random rand = null);
 
         /// <summary>
         /// This constructs a set of parallel batch cursors. The value <paramref name="n"/> is a recommended limit on
@@ -56,11 +56,11 @@ namespace Microsoft.ML.Data
         /// scenarios will be content with pulling from the single serial cursor of
         /// <see cref="GetRowCursor(IEnumerable{Schema.Column}, Random)"/>.
         /// </summary>
-        /// <param name="colsNeeded">The active columns needed. If passed an empty <see cref="IEnumerable"/> no column is requested.</param>
+        /// <param name="columnsNeeded">The active columns needed. If passed an empty <see cref="IEnumerable"/> no column is requested.</param>
         /// <param name="n">The suggested degree of parallelism.</param>
         /// <param name="rand">An instance of <see cref="Random"/> to seed randomizing the access.</param>
         /// <returns></returns>
-        RowCursor[] GetRowCursorSet(IEnumerable<Schema.Column> colsNeeded, int n, Random rand = null);
+        RowCursor[] GetRowCursorSet(IEnumerable<Schema.Column> columnsNeeded, int n, Random rand = null);
 
         /// <summary>
         /// Gets an instance of Schema.
--- a/src/Microsoft.ML.Core/Data/IDataView.cs
+++ b/src/Microsoft.ML.Core/Data/IDataView.cs
@@ -32,10 +32,10 @@ namespace Microsoft.ML.Data
         /// <summary>
         /// Get a row cursor. The active column indices are those for which needCol(col) returns true.
         /// The schema of the returned cursor will be the same as the schema of the IDataView, but getting
-        /// a getter for an inactive columns will throw. The <paramref name="needCol"/> predicate must be
-        /// non-null. To activate all columns, pass "col => true".
+        /// a getter for an inactive columns will throw. The <paramref name="colsNeeded"/> predicate must be
+        /// non-null.If set to <langword>null</langword> no column is active.
         /// </summary>
-        RowCursor GetRowCursor(Func<int, bool> needCol, Random rand = null);
+        RowCursor GetRowCursor(IEnumerable<Schema.Column> colsNeeded, Random rand = null);
 
         /// <summary>
         /// This constructs a set of parallel batch cursors. The value <paramref name="n"/> is a recommended limit on
@@ -44,7 +44,7 @@ namespace Microsoft.ML.Data
         /// recommendation: it is entirely possible that an implementation can return a different number of cursors.
         ///
         /// The cursors should return the same data as returned through
-        /// <see cref="GetRowCursor(Func{int, bool}, Random)"/>, except partitioned: no two cursors should return the
+        /// <see cref="GetRowCursor(IEnumerable{Schema.Column}, Random)"/>, except partitioned: no two cursors should return the
         /// "same" row as would have been returned through the regular serial cursor, but all rows should be returned by
         /// exactly one of the cursors returned from this cursor. The cursors can have their values reconciled
         /// downstream through the use of the <see cref="Row.Batch"/> property.
@@ -53,13 +53,13 @@ namespace Microsoft.ML.Data
         /// working threads that consume from them independently while, ultimately, the results are finally collated in
         /// the end by exploiting the ordering of the <see cref="Row.Batch"/> property described above. More typical
         /// scenarios will be content with pulling from the single serial cursor of
-        /// <see cref="GetRowCursor(Func{int, bool}, Random)"/>.
+        /// <see cref="GetRowCursor(IEnumerable{Schema.Column}, Random)"/>.
         /// </summary>
-        /// <param name="needCol">The predicate, where a column is active if this returns true.</param>
+        /// <param name="colsNeeded">The active columns needed. If set to <langword>null</langword> no column is active.</param>
         /// <param name="n">The suggested degree of parallelism.</param>
         /// <param name="rand">An instance </param>
         /// <returns></returns>
-        RowCursor[] GetRowCursorSet(Func<int, bool> needCol, int n, Random rand = null);
+        RowCursor[] GetRowCursorSet(IEnumerable<Schema.Column> colsNeeded, int n, Random rand = null);
 
         /// <summary>
         /// Gets an instance of Schema.
@@ -93,15 +93,15 @@ namespace Microsoft.ML.Data
 
         /// <summary>
         /// This provides a means for reconciling multiple rows that have been produced generally from
-        /// <see cref="IDataView.GetRowCursorSet(Func{int, bool}, int, Random)"/>. When getting a set, there is a need
+        /// <see cref="IDataView.GetRowCursorSet(IEnumerable{Schema.Column}, int, Random)"/>. When getting a set, there is a need
         /// to, while allowing parallel processing to proceed, always have an aim that the original order should be
         /// reconverable. Note, whether or not a user cares about that original order in ones specific application is
         /// another story altogether (most callers of this as a practical matter do not, otherwise they would not call
         /// it), but at least in principle it should be possible to reconstruct the original order one would get from an
-        /// identically configured <see cref="IDataView.GetRowCursor(Func{int, bool}, Random)"/>. So: for any cursor
+        /// identically configured <see cref="IDataView.GetRowCursor(IEnumerable{Schema.Column}, Random)"/>. So: for any cursor
         /// implementation, batch numbers should be non-decreasing. Furthermore, any given batch number should only
         /// appear in one of the cursors as returned by
-        /// <see cref="IDataView.GetRowCursorSet(Func{int, bool}, int, Random)"/>. In this way, order is determined by
+        /// <see cref="IDataView.GetRowCursorSet(IEnumerable{Schema.Column}, int, Random)"/>. In this way, order is determined by
         /// batch number. An operation that reconciles these cursors to produce a consistent single cursoring, could do
         /// so by drawing from the single cursor, among all cursors in the set, that has the smallest batch number
         /// available.
--- a/src/Microsoft.ML.Core/Data/Schema.cs
+++ b/src/Microsoft.ML.Core/Data/Schema.cs
@@ -13,11 +13,7 @@ namespace Microsoft.ML.Data
 {
     /// <summary>
     /// This class represents the <see cref="Schema"/> of an object like, for interstance, an <see cref="IDataView"/> or an <see cref="Row"/>.
-    /// On the high level, the schema is a collection of 'columns'. Each column has the following properties:
-    /// - Column name.
-    /// - Column type.
-    /// - Metadata. The metadata itself is a 'single-row dataset' (namely, an instance of <see cref="Metadata"/>), that contains its own schema
-    /// and values.
+    /// On the high level, the schema is a collection of <see cref="Schema.Column"/>.
     /// </summary>
     [System.Diagnostics.DebuggerTypeProxy(typeof(SchemaDebuggerProxy))]
     public sealed class Schema : IReadOnlyList<Schema.Column>
@@ -67,18 +63,6 @@ namespace Microsoft.ML.Data
             if (_nameMap.TryGetValue(name, out int col))
                 return _columns[col];
             return null;
-        }
-
-        public IEnumerable<Column> GetColumns(params string[] names)
-        {
-            foreach (var name in names)
-                yield return this[name];
-        }
-
-        public IEnumerable<Column> GetColumns(params int[] indices)
-        {
-            foreach (var index in indices)
-                yield return this[index];
         }
 
         public IEnumerator<Column> GetEnumerator() => ((IEnumerable<Column>)_columns).GetEnumerator();
--- a/src/Microsoft.ML.Core/Data/Schema.cs
+++ b/src/Microsoft.ML.Core/Data/Schema.cs
@@ -69,14 +69,23 @@ namespace Microsoft.ML.Data
             return null;
         }
 
+        public IEnumerable<Column> GetColumns(params string[] names)
+        {
+            foreach (var name in names)
+                yield return this[name];
+        }
+
+        public IEnumerable<Column> GetColumns(params int[] indices)
+        {
+            foreach (var index in indices)
+                yield return this[index];
+        }
+
         public IEnumerator<Column> GetEnumerator() => ((IEnumerable<Column>)_columns).GetEnumerator();
 
         IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
 
-        public override string ToString()
-        {
-            return $"{Count} columns";
-        }
+        public override string ToString() => $"{Count} columns";
 
         /// <summary>
         /// This class describes one column in the particular schema.
--- a/src/Microsoft.ML.Core/Utilities/Utils.cs
+++ b/src/Microsoft.ML.Core/Utilities/Utils.cs
@@ -6,10 +6,12 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
+using Microsoft.ML.Data;
 
 namespace Microsoft.ML.Internal.Utilities
 {
@@ -706,6 +708,16 @@ namespace Microsoft.ML.Internal.Utilities
             return result;
         }
 
+        public static bool[] BuildArray(int length, IEnumerable<Schema.Column> colsNeeded)
+        {
+            Contracts.CheckParam(length >= 0, nameof(length));
+
+            var result = new bool[length];
+            for (int i = 0; i < result.Length; i++)
+                result[i] = colsNeeded == null ? false : colsNeeded.Any(x => x.Index == i);
+            return result;
+        }
+
         public static T[] BuildArray<T>(int length, Func<int, T> func)
         {
             Contracts.CheckParam(length >= 0, nameof(length));
@@ -741,6 +753,41 @@ namespace Microsoft.ML.Internal.Utilities
             for (int c = 0; c < lim; ++c)
             {
                 if (!pred(c))
+                {
+                    invMap[c] = -1;
+                    continue;
+                }
+                invMap[c] = mapList.Count;
+                mapList.Add(c);
+            }
+            map = mapList.ToArray();
+        }
+
+        /// <summary>
+        /// Given a predicate, over a range of values defined by a limit calculate
+        /// first the values for which that predicate was true, and second an inverse
+        /// map.
+        /// </summary>
+        /// <param name="lim">Indicates the exclusive upper bound on the tested values</param>
+        /// <param name="colsNeeded">The set of columns the calling component operates on.</param>
+        /// <param name="map">An ascending array of values from 0 inclusive
+        /// to <paramref name="lim"/> exclusive, holding all values for which
+        /// <paramref name="colsNeeded"/> are present.
+        /// (The respective index appears in the <paramref name="colsNeeded"/> collection).</param>
+        /// <param name="invMap">Forms an inverse mapping of <paramref name="map"/>,
+        /// so that <c><paramref name="invMap"/>[<paramref name="map"/>[i]] == i</c>,
+        /// and for other entries not appearing in <paramref name="map"/>,
+        /// <c><paramref name="invMap"/>[i] == -1</c></param>
+        public static void BuildSubsetMaps(int lim, IEnumerable<Schema.Column> colsNeeded, out int[] map, out int[] invMap)
+        {
+            Contracts.CheckParam(lim >= 0, nameof(lim));
+
+            // REVIEW: Better names?
+            List<int> mapList = new List<int>();
+            invMap = new int[lim];
+            for (int c = 0; c < lim; ++c)
+            {
+                if (colsNeeded == null || !colsNeeded.Any(x => x.Index == c))
                 {
                     invMap[c] = -1;
                     continue;
--- a/src/Microsoft.ML.Core/Utilities/Utils.cs
+++ b/src/Microsoft.ML.Core/Utilities/Utils.cs
@@ -794,8 +794,8 @@ namespace Microsoft.ML.Internal.Utilities
             foreach (var col in colsNeeded)
             {
                 Contracts.Check(col.Index < lim);
-                mapList.Add(col.Index);
                 invMap[col.Index] = mapList.Count;
+                mapList.Add(col.Index);
             }
 
             map = mapList.ToArray();
--- a/src/Microsoft.ML.Core/Utilities/Utils.cs
+++ b/src/Microsoft.ML.Core/Utilities/Utils.cs
@@ -713,8 +713,12 @@ namespace Microsoft.ML.Internal.Utilities
             Contracts.CheckParam(length >= 0, nameof(length));
 
             var result = new bool[length];
-            for (int i = 0; i < result.Length; i++)
-                result[i] = colsNeeded == null ? false : colsNeeded.Any(x => x.Index == i);
+            foreach (var col in colsNeeded)
+            {
+                if(col.Index < result.Length)
+                    result[col.Index] = true;
+            }
+
             return result;
         }
 
@@ -764,8 +768,8 @@ namespace Microsoft.ML.Internal.Utilities
         }
 
         /// <summary>
-        /// Given a predicate, over a range of values defined by a limit calculate
-        /// first the values for which that predicate was true, and second an inverse
+        /// Given the columns needed, over a range of values defined by a limit calculate
+        /// first the values for which the column is present was true, and second an inverse
         /// map.
         /// </summary>
         /// <param name="lim">Indicates the exclusive upper bound on the tested values</param>
@@ -781,20 +785,19 @@ namespace Microsoft.ML.Internal.Utilities
         public static void BuildSubsetMaps(int lim, IEnumerable<Schema.Column> colsNeeded, out int[] map, out int[] invMap)
         {
             Contracts.CheckParam(lim >= 0, nameof(lim));
+            Contracts.CheckValue(colsNeeded, nameof(colsNeeded));
 
             // REVIEW: Better names?
             List<int> mapList = new List<int>();
-            invMap = new int[lim];
-            for (int c = 0; c < lim; ++c)
+            invMap = invMap = Enumerable.Repeat(-1, lim).ToArray<int>();
+
+            foreach (var col in colsNeeded)
             {
-                if (colsNeeded == null || !colsNeeded.Any(x => x.Index == c))
-                {
-                    invMap[c] = -1;
-                    continue;
-                }
-                invMap[c] = mapList.Count;
-                mapList.Add(c);
+                Contracts.Check(col.Index < lim);
+                mapList.Add(col.Index);
+                invMap[col.Index] = mapList.Count;
             }
+
             map = mapList.ToArray();
         }
 
--- a/src/Microsoft.ML.Core/Utilities/Utils.cs
+++ b/src/Microsoft.ML.Core/Utilities/Utils.cs
@@ -708,12 +708,12 @@ namespace Microsoft.ML.Internal.Utilities
             return result;
         }
 
-        public static bool[] BuildArray(int length, IEnumerable<Schema.Column> colsNeeded)
+        public static bool[] BuildArray(int length, IEnumerable<Schema.Column> columnsNeeded)
         {
             Contracts.CheckParam(length >= 0, nameof(length));
 
             var result = new bool[length];
-            foreach (var col in colsNeeded)
+            foreach (var col in columnsNeeded)
             {
                 if(col.Index < result.Length)
                     result[col.Index] = true;
@@ -773,25 +773,25 @@ namespace Microsoft.ML.Internal.Utilities
         /// map.
         /// </summary>
         /// <param name="lim">Indicates the exclusive upper bound on the tested values</param>
-        /// <param name="colsNeeded">The set of columns the calling component operates on.</param>
+        /// <param name="columnsNeeded">The set of columns the calling component operates on.</param>
         /// <param name="map">An ascending array of values from 0 inclusive
         /// to <paramref name="lim"/> exclusive, holding all values for which
-        /// <paramref name="colsNeeded"/> are present.
-        /// (The respective index appears in the <paramref name="colsNeeded"/> collection).</param>
+        /// <paramref name="columnsNeeded"/> are present.
+        /// (The respective index appears in the <paramref name="columnsNeeded"/> collection).</param>
         /// <param name="invMap">Forms an inverse mapping of <paramref name="map"/>,
         /// so that <c><paramref name="invMap"/>[<paramref name="map"/>[i]] == i</c>,
         /// and for other entries not appearing in <paramref name="map"/>,
         /// <c><paramref name="invMap"/>[i] == -1</c></param>
-        public static void BuildSubsetMaps(int lim, IEnumerable<Schema.Column> colsNeeded, out int[] map, out int[] invMap)
+        public static void BuildSubsetMaps(int lim, IEnumerable<Schema.Column> columnsNeeded, out int[] map, out int[] invMap)
         {
             Contracts.CheckParam(lim >= 0, nameof(lim));
-            Contracts.CheckValue(colsNeeded, nameof(colsNeeded));
+            Contracts.CheckValue(columnsNeeded, nameof(columnsNeeded));
 
             // REVIEW: Better names?
             List<int> mapList = new List<int>();
             invMap = invMap = Enumerable.Repeat(-1, lim).ToArray<int>();
 
-            foreach (var col in colsNeeded)
+            foreach (var col in columnsNeeded)
             {
                 Contracts.Check(col.Index < lim);
                 invMap[col.Index] = mapList.Count;
--- a/src/Microsoft.ML.Data/Data/DataViewUtils.cs
+++ b/src/Microsoft.ML.Data/Data/DataViewUtils.cs
@@ -122,7 +122,7 @@ namespace Microsoft.ML.Data
 
             int cthd = GetThreadCount(host);
             host.Assert(cthd > 0);
-            if (cthd == 1 || !AllCachable(view.Schema, colsNeeded))
+            if (cthd == 1 || !AllCacheable(view.Schema, colsNeeded))
             {
                 curs = null;
                 return false;
@@ -159,7 +159,7 @@ namespace Microsoft.ML.Data
                 return new RowCursor[1] { input };
 
             // If any active columns are not cachable, we can't split.
-            if (!AllCachable(input.Schema, input.IsColumnActive))
+            if (!AllCacheable(input.Schema, input.IsColumnActive))
                 return new RowCursor[1] { input };
 
             // REVIEW: Should we limit the cardinality to some reasonable size?
@@ -177,7 +177,7 @@ namespace Microsoft.ML.Data
         /// Return whether all the active columns, as determined by the predicate, are
         /// cachable - either primitive types or vector types.
         /// </summary>
-        public static bool AllCachable(Schema schema, Func<int, bool> predicate)
+        public static bool AllCacheable(Schema schema, Func<int, bool> predicate)
         {
             Contracts.CheckValue(schema, nameof(schema));
             Contracts.CheckValue(predicate, nameof(predicate));
@@ -187,7 +187,7 @@ namespace Microsoft.ML.Data
                 if (!predicate(col))
                     continue;
                 var type = schema[col].Type;
-                if (!IsCachable(type))
+                if (!IsCacheable(type))
                     return false;
             }
 
@@ -198,7 +198,7 @@ namespace Microsoft.ML.Data
         /// Return whether all the active columns, as determined by the predicate, are
         /// cachable - either primitive types or vector types.
         /// </summary>
-        public static bool AllCachable(Schema schema, IEnumerable<Schema.Column> colsNeeded)
+        public static bool AllCacheable(Schema schema, IEnumerable<Schema.Column> colsNeeded)
         {
             Contracts.CheckValue(schema, nameof(schema));
 
@@ -210,7 +210,7 @@ namespace Microsoft.ML.Data
                 if (!colsNeeded.Any(x => x.Index == col))
                     continue;
                 var type = schema[col].Type;
-                if (!IsCachable(type))
+                if (!IsCacheable(type))
                     return false;
             }
 
@@ -220,7 +220,7 @@ namespace Microsoft.ML.Data
         /// <summary>
         /// Determine whether the given type is cachable - either a primitive type or a vector type.
         /// </summary>
-        public static bool IsCachable(this ColumnType type)
+        public static bool IsCacheable(this ColumnType type)
         {
             return type != null && (type is PrimitiveType || type is VectorType);
         }
@@ -341,7 +341,7 @@ namespace Microsoft.ML.Data
 
                 RowCursor cursor = inputs[0];
                 var schema = cursor.Schema;
-                ch.CheckParam(AllCachable(schema, cursor.IsColumnActive), nameof(inputs), "Inputs had some uncachable input columns");
+                ch.CheckParam(AllCacheable(schema, cursor.IsColumnActive), nameof(inputs), "Inputs had some uncachable input columns");
 
                 int[] activeToCol;
                 int[] colToActive;
@@ -518,7 +518,7 @@ namespace Microsoft.ML.Data
                 ch.AssertValue(input);
                 ch.Assert(input.Schema == _schema);
                 ch.Assert(cthd >= 2);
-                ch.Assert(AllCachable(_schema, input.IsColumnActive));
+                ch.Assert(AllCacheable(_schema, input.IsColumnActive));
 
                 // REVIEW: Should the following be configurable?
                 // How would we even expose these sorts of parameters to a user?
@@ -547,7 +547,7 @@ namespace Microsoft.ML.Data
                     ch.Assert(c == 0 || activeToCol[c - 1] < activeToCol[c]);
                     ch.Assert(input.IsColumnActive(activeToCol[c]));
                     var type = input.Schema[activeToCol[c]].Type;
-                    ch.Assert(type.IsCachable());
+                    ch.Assert(type.IsCacheable());
                     arguments[1] = activeToCol[c];
                     var inPipe = inPipes[c] =
                         (InPipe)inGenMethod.MakeGenericMethod(type.RawType).Invoke(this, arguments);
--- a/src/Microsoft.ML.Data/Data/DataViewUtils.cs
+++ b/src/Microsoft.ML.Data/Data/DataViewUtils.cs
@@ -122,7 +122,7 @@ namespace Microsoft.ML.Data
 
             int cthd = GetThreadCount(host);
             host.Assert(cthd > 0);
-            if (cthd == 1 || !AllCacheable(view.Schema, colsNeeded))
+            if (cthd == 1 || !AllCacheable(colsNeeded))
             {
                 curs = null;
                 return false;
@@ -198,21 +198,16 @@ namespace Microsoft.ML.Data
         /// Return whether all the active columns, as determined by the predicate, are
         /// cachable - either primitive types or vector types.
         /// </summary>
-        public static bool AllCacheable(Schema schema, IEnumerable<Schema.Column> colsNeeded)
+        public static bool AllCacheable(IEnumerable<Schema.Column> colsNeeded)
         {
-            Contracts.CheckValue(schema, nameof(schema));
+            Contracts.CheckValue(colsNeeded, nameof(colsNeeded));
 
             if (colsNeeded == null)
                 return false;
 
-            for (int col = 0; col < schema.Count; col++)
-            {
-                if (!colsNeeded.Any(x => x.Index == col))
-                    continue;
-                var type = schema[col].Type;
-                if (!IsCacheable(type))
+            foreach (var col in colsNeeded)
+                if (!IsCacheable(col.Type))
                     return false;
-            }
 
             return true;
         }
@@ -221,9 +216,7 @@ namespace Microsoft.ML.Data
         /// Determine whether the given type is cachable - either a primitive type or a vector type.
         /// </summary>
         public static bool IsCacheable(this ColumnType type)
-        {
-            return type != null && (type is PrimitiveType || type is VectorType);
-        }
+            => type != null && (type is PrimitiveType || type is VectorType);
 
         /// <summary>
         /// Tests whether the cursors are mutually compatible for consolidation,
--- a/src/Microsoft.ML.Data/Data/DataViewUtils.cs
+++ b/src/Microsoft.ML.Data/Data/DataViewUtils.cs
@@ -82,7 +82,7 @@ namespace Microsoft.ML.Data
             if (countNullable != null)
                 return countNullable.Value;
             long count = 0;
-            using (var cursor = view.GetRowCursor(null))
+            using (var cursor = view.GetRowCursor())
             {
                 while (cursor.MoveNext())
                     count++;
--- a/src/Microsoft.ML.Data/Data/DataViewUtils.cs
+++ b/src/Microsoft.ML.Data/Data/DataViewUtils.cs
@@ -115,20 +115,20 @@ namespace Microsoft.ML.Data
         /// the target cardinality of the cursor set.
         /// </summary>
         public static bool TryCreateConsolidatingCursor(out RowCursor curs,
-            IDataView view, IEnumerable<Schema.Column> colsNeeded, IHost host, Random rand)
+            IDataView view, IEnumerable<Schema.Column> columnsNeeded, IHost host, Random rand)
         {
             Contracts.CheckValue(host, nameof(host));
             host.CheckValue(view, nameof(view));
 
             int cthd = GetThreadCount(host);
             host.Assert(cthd > 0);
-            if (cthd == 1 || !AllCacheable(colsNeeded))
+            if (cthd == 1 || !AllCacheable(columnsNeeded))
             {
                 curs = null;
                 return false;
             }
 
-            var inputs = view.GetRowCursorSet(colsNeeded, cthd, rand);
+            var inputs = view.GetRowCursorSet(columnsNeeded, cthd, rand);
             host.Check(Utils.Size(inputs) > 0);
 
             if (inputs.Length == 1)
@@ -198,14 +198,14 @@ namespace Microsoft.ML.Data
         /// Return whether all the active columns, as determined by the predicate, are
         /// cachable - either primitive types or vector types.
         /// </summary>
-        public static bool AllCacheable(IEnumerable<Schema.Column> colsNeeded)
+        public static bool AllCacheable(IEnumerable<Schema.Column> columnsNeeded)
         {
-            Contracts.CheckValue(colsNeeded, nameof(colsNeeded));
+            Contracts.CheckValue(columnsNeeded, nameof(columnsNeeded));
 
-            if (colsNeeded == null)
+            if (columnsNeeded == null)
                 return false;
 
-            foreach (var col in colsNeeded)
+            foreach (var col in columnsNeeded)
                 if (!IsCacheable(col.Type))
                     return false;
 
--- a/src/Microsoft.ML.Data/Data/DataViewUtils.cs
+++ b/src/Microsoft.ML.Data/Data/DataViewUtils.cs
@@ -82,7 +82,7 @@ namespace Microsoft.ML.Data
             if (countNullable != null)
                 return countNullable.Value;
             long count = 0;
-            using (var cursor = view.GetRowCursor(col => false))
+            using (var cursor = view.GetRowCursor(null))
             {
                 while (cursor.MoveNext())
                     count++;
@@ -115,21 +115,20 @@ namespace Microsoft.ML.Data
         /// the target cardinality of the cursor set.
         /// </summary>
         public static bool TryCreateConsolidatingCursor(out RowCursor curs,
-            IDataView view, Func<int, bool> predicate, IHost host, Random rand)
+            IDataView view, IEnumerable<Schema.Column> colsNeeded, IHost host, Random rand)
         {
             Contracts.CheckValue(host, nameof(host));
             host.CheckValue(view, nameof(view));
-            host.CheckValue(predicate, nameof(predicate));
 
             int cthd = GetThreadCount(host);
             host.Assert(cthd > 0);
-            if (cthd == 1 || !AllCachable(view.Schema, predicate))
+            if (cthd == 1 || !AllCachable(view.Schema, colsNeeded))
             {
                 curs = null;
                 return false;
             }
 
-            var inputs = view.GetRowCursorSet(predicate, cthd, rand);
+            var inputs = view.GetRowCursorSet(colsNeeded, cthd, rand);
             host.Check(Utils.Size(inputs) > 0);
 
             if (inputs.Length == 1)
@@ -186,6 +185,29 @@ namespace Microsoft.ML.Data
             for (int col = 0; col < schema.Count; col++)
             {
                 if (!predicate(col))
+                    continue;
+                var type = schema[col].Type;
+                if (!IsCachable(type))
+                    return false;
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Return whether all the active columns, as determined by the predicate, are
+        /// cachable - either primitive types or vector types.
+        /// </summary>
+        public static bool AllCachable(Schema schema, IEnumerable<Schema.Column> colsNeeded)
+        {
+            Contracts.CheckValue(schema, nameof(schema));
+
+            if (colsNeeded == null)
+                return false;
+
+            for (int col = 0; col < schema.Count; col++)
+            {
+                if (!colsNeeded.Any(x => x.Index == col))
                     continue;
                 var type = schema[col].Type;
                 if (!IsCachable(type))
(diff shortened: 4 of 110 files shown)


sfilipi:
@TomFinley there are a lot of tests failing, so I anticipate quite a bit more changes. 
Before finalizing i wanted to check the impact of the changes on the perf, and see if I can extend removing the predicates on the Cursor ctors. 
Removing them from GetDependancies etc. Can come on a later PR, IMO. 

Let me know if you spot any systematic errors.

TomFinley:
> Before finalizing i wanted to check the impact of the changes on the perf,

That seems wise, but is there a reason to suspect so? This would only be during cursor or row creation, I must expect, where we have historically not observed optimization playing a significant role?

src/Microsoft.ML.Data/Transforms/ColumnSelecting.cs
```diff
@@ -625,31 +625,33 @@ public SelectColumnsDataTransform(IHostEnvironment env, ColumnSelectingTransform
...
+                Func<int, bool> predicate = c => colsNeeded == null ? false : colsNeeded.Any(x => x.Index == c);
                 _host.AssertValueOrNull(rand);
 
                 // Build out the active state for the input
-                var inputPred = GetDependencies(needCol);
-                var inputRowCursor = Source.GetRowCursor(inputPred, rand);
+                var inputPred = GetDependencies(predicate);
 
+                var inputRowCursor = Source.GetRowCursor(colsNeeded, rand);
+                var inputCols = Source.Schema.Where(x => inputPred(x.Index));
```
sfilipi: swap, and use inputCols in GetRowCursor

src/Microsoft.ML.Data/Training/TrainerUtils.cs
```diff
@@ -214,26 +214,22 @@ public static void CheckOptGroup(this RoleMappedData data)
...
         {
             Contracts.AssertValue(data);
             Contracts.AssertValueOrNull(extraCols);
 
-            var cols = new HashSet<int>();
+            var columns = data.Data.Schema.Where(x => extraCols.Contains(x.Index));
+
             if ((opt & CursOpt.Label) != 0)
-                AddOpt(cols, data.Schema.Label);
+                columns.Append(data.Schema.Label.Value);
```
sfilipi: >columns.Append [](start = 16, length = 14)

fix

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -254,12 +254,13 @@ private ColumnType GetOutputType(Schema inputSchema, ColumnInfo column)
             var types = new ColumnType[_columns.Length];
             List<int> invertIinfos = null;
             List<int> invertHashMaxCounts = null;
-            HashSet<int> sourceColumnsForInvertHash = new HashSet<int>();
+            var sourceColumnsForInvertHash = new List<Schema.Column>();
             for (int i = 0; i < _columns.Length; i++)
             {
-                if (!input.Schema.TryGetColumnIndex(ColumnPairs[i].input, out int srcCol))
+                Schema.Column? srcCol = input.Schema.GetColumnOrNull(ColumnPairs[i].input);
+                if (srcCol != null)
```
sfilipi: >!= [](start = 27, length = 2)

nope

src/Microsoft.ML.Data/DataView/RowToRowMapperTransform.cs
```diff
@@ -183,22 +183,29 @@ private bool[] GetActive(Func<int, bool> predicate, out Func<int, bool> predicat
...
-            Host.CheckValue(predicate, nameof(predicate));
             Host.CheckValueOrNull(rand);
 
+            Func<int, bool> predicate = c => colsNeeded == null ? false : colsNeeded.Any(x => x.Index == c);
+
             Func<int, bool> predicateInput;
             var active = GetActive(predicate, out predicateInput);
 
-            var inputs = Source.GetRowCursorSet(predicateInput, n, rand);
+            var inputCols = colsNeeded.Where(x => predicateInput(x.Index)).ToList().AsReadOnly();
```
sfilipi: >.ToList().AsReadOnly() [](start = 74, length = 22)

remove

src/Microsoft.ML.Data/DataLoadSave/PartitionedFileLoader.cs
```diff
@@ -365,29 +365,35 @@ private sealed class Cursor : RootCursorBase
...
                 _parent = parent;
+                _colsNeeded = colsNeeded;
 
-                _active = Utils.BuildArray(Schema.Count, predicate);
+                _active = Utils.BuildArray(Schema.Count, colsNeeded);
                 _subActive = _active.Take(SubColumnCount).ToArray();
                 _colValues = new ReadOnlyMemory<char>[Schema.Count - SubColumnCount];
 
+                // sefilipi: Test this
+                _subActiveColsNeeded = Schema.Where((x, i) => _subActive[i]);
```
sfilipi: > _subActive[i] [](start = 61, length = 14)

check  for out of bound, here and everywhere

src/Microsoft.ML.Transforms/PermutationFeatureImportance.cs
```diff
@@ -139,7 +139,7 @@ internal static class PermutationFeatureImportance<TMetric, TResult> where TResu
                 var valuesRowCount = 0;
                 // REVIEW: Seems like if the labels are NaN, so that all metrics are NaN, this command will be useless.
                 // In which case probably erroring out is probably the most useful thing.
-                using (var cursor = data.GetRowCursor(col => col == featuresColumnIndex))
+                using (var cursor = data.GetRowCursor(data.Schema.Where(x =>  x.Index == featuresColumnIndex)))
```
TomFinley: >data.Schema.Where(x =>  x.Index == featuresColumnIndex [](start = 54, length = 54)

Much better than this of course would be `new[] { data.Schema[featuresColumnIndex] }`. :) But I would actually suggest actually something else, I suspect that it would be best actually to have an extension method to `IDataView` somewhere in our public API with the signature `RowCursor GetRowCursor(params Schema.Column[] columns)` to handle this case, which I must expect would be very common.

TomFinley:
Having read through your changes a bit, I think that you might benefit from a utility at the level where you go from the externally visible `IEnumerable<Schema.Column>` to the internal `Func<int, bool>` that is still being exploited internally? By doing it in this decentralized fashion instead of having an internal utility, this both makes your code really complicated (since you're re-writing what amounts to similar code again and again), and also makes it slow since your current mechanism appears to be to exploit LINQ -- LINQ is a fine abstraction and convenience, but not necessarily something we want in production code.

So for example, let's imagine you had some utility that tool `Schema` and `IEnumerable<Schema.Column>`, and returned `Func<int, bool>`? Maybe `Utils.CreateActivePredictate` or something. In that case, that function could be shared among all implementations, and be optimized to handle special common cases... e.g., if it's an enumerable covering *all* columns, it returns `col => true` with no further fuss. If empty, return `col => false`. If there are 1000 columns but only 2 are selected, it can just compose a delegate with an exact test instead of working through LINQ. Etc. Etc. What you think?

src/Microsoft.ML.Data/DataView/CacheDataView.cs
```diff
@@ -203,10 +203,12 @@ public int MapInputToCacheColumnIndex(int inputIndex)
             return _rowCount;
         }
 
-        public RowCursor GetRowCursor(Func<int, bool> predicate, Random rand = null)
+        public RowCursor GetRowCursor(IEnumerable<Schema.Column> colsNeeded, Random rand = null)
         {
-            _host.CheckValue(predicate, nameof(predicate));
             _host.CheckValueOrNull(rand);
+
+            Func<int, bool> predicate = c => colsNeeded == null ? false : colsNeeded.Any(x => x.Index == c);
```
TomFinley: >Func<int, bool> predicate = c => colsNeeded == null ? false : colsNeeded.Any(x => x.Index == c); [](start = 12, length = 96)

Here's a place where the utility would be used to simplify code.

src/Microsoft.ML.Transforms/PermutationFeatureImportance.cs
sfilipi: I wasn't 100% sure that x.Index and the position of the columns in the Shema matches, hence this checking on the Index everywhere. 

After I get all the tests passing, I'll swap this out everywhere.

---
In reply to: [244382663](https://github.com/dotnet/machinelearning/pull/1975#discussion_r244382663) [](ancestors = 244382663)

sfilipi:
Makes sense, thanks for the suggestion! 

---
In reply to: [450407562](https://github.com/dotnet/machinelearning/pull/1975#issuecomment-450407562) [](ancestors = 450407562)

src/Microsoft.ML.Transforms/PermutationFeatureImportance.cs
TomFinley: OK! Also w.r.t. the extension, if you see that the single column selection is common, might even be good to have extension overloads for `Schema.Column`, and `Schema.Column, Schema.Column` (for two accessors, again if you observe it being common), in addition to the suggested `params` extension.

---
In reply to: [244384709](https://github.com/dotnet/machinelearning/pull/1975#discussion_r244384709) [](ancestors = 244384709,244382663)

src/Microsoft.ML.Data/DataView/Transposer.cs
```diff
@@ -439,7 +435,7 @@ protected override ValueGetter<VBuffer<T>> GetGetterCore()
                         Ch.Check(IsGood, "Cannot get values in the cursor's current state");
                         if (!valid)
                         {
-                            using (var cursor = _view.GetRowCursor(c => c == _col))
+                            using (var cursor = _view.GetRowCursor(_view.Schema))
```
sfilipi: >(_view.Schema [](start = 66, length = 13)

fix

src/Microsoft.ML.Data/DataView/Transposer.cs
```diff
@@ -857,21 +853,26 @@ private void OutputColumnToSplitterIndices(int col, out int splitInd, out int sp
...
-            public RowCursor GetRowCursor(Func<int, bool> predicate, Random rand = null)
+            public RowCursor GetRowCursor(IEnumerable<Schema.Column> colsNeeded, Random rand = null)
             {
-                _host.CheckValue(predicate, nameof(predicate));
+                Func<int, bool> predicate = c => colsNeeded == null ? false : colsNeeded.Any(x => x.Index == c);
+
                 bool[] activeSplitters;
                 var srcPred = CreateInputPredicate(predicate, out activeSplitters);
-                return new Cursor(_host, this, _input.GetRowCursor(srcPred, rand), predicate, activeSplitters);
+                return new Cursor(_host, this, _input.GetRowCursor(colsNeeded, rand), predicate, activeSplitters);
```
sfilipi: >colsNeeded [](start = 67, length = 10)

columns from srcPred

src/Microsoft.ML.Transforms/GroupTransform.cs
```diff
@@ -582,7 +582,7 @@ public Cursor(GroupTransform parent, Func<int, bool> predicate)
                     if (_active[i + _groupCount])
                         srcActiveTrailing[schema.KeepIds[i]] = true;
                 }
-                _trailingCursor = parent.Source.GetRowCursor(x => srcActiveTrailing[x]);
+                _trailingCursor = parent.Source.GetRowCursor(activeCols);
```
sfilipi: >activeCols [](start = 61, length = 10)

this should be a diff set, created based on srcActiveTrailing.

sfilipi:
@tfinley@gmail.com checking on one more thing: are you suggesting I avoid LINQ, and turn to traditional loops etc?

---
In reply to: [450410147](https://github.com/dotnet/machinelearning/pull/1975#issuecomment-450410147) [](ancestors = 450410147,450407562)

src/Microsoft.ML.Core/Data/Schema.cs
```diff
@@ -69,14 +69,23 @@ public sealed class Schema : IReadOnlyList<Schema.Column>
             return null;
         }
 
+        public IEnumerable<Column> GetColumns(params string[] names)
+        {
```
TomFinley: I really would rather not. What is the point of adding such a thing to `Schema`? If we wanted to have some internal extension method, that's fine.

src/Microsoft.ML.Data/DataView/DataViewExtensions.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.ML.Data
+{
+    public static class DataViewExtensions
```
TomFinley: >DataViewExtensions [](start = 24, length = 18)

Whoops, this class was made public.

src/Microsoft.ML.Data/DataView/DataViewExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        /// <summary>
+        /// Get a row cursor. The active column indices are those for which needCol(col) returns true.
+        /// The schema of the returned cursor will be the same as the schema of the IDataView, but getting
+        /// a getter for an inactive columns will throw. The <paramref name="namesOfcolsNeeded"/> predicate must be
+        /// non-null.If set to <langword>null</langword> no column is active.
```
TomFinley: >langword [](start = 32, length = 8)

Is `<langword>` a thing? In VS at least while I can use its facilities to autocomplete to `<see langword` (which is what I've seen used elsewhere), I do not see that `<langword>` just itself is a thing. Though, you have more experience with the doc system than I do perhaps, so maybe I just somehow missed it? Do you have a reference?

src/Microsoft.ML.Data/DataView/DataViewExtensions.cs
TomFinley: Did you add this because you thought this is what I meant by a utility? Or was there some other reason? If it was for my benefit, I actually meant (1) something that works purely over `Schema.Column`. In any case, I'm not sure this is a good idea. I think operating over `string` and `int` directly is dangerous and gives up a lot of error checking capability that is roughly 75% of my own motivation for wanting this change. Also whatever "shims" we have, even if necessary, ought to be completely temporary, and ought to work in the *opposite* direction. (That is, I accept that *internally* we might have some things that work directly over `int` indices, so mapping from `Schema.Column` to `int` based structures might be fine. Doing the opposite though, is going in the wrong direction.)

---
In reply to: [245080660](https://github.com/dotnet/machinelearning/pull/1975#discussion_r245080660) [](ancestors = 245080660)

src/Microsoft.ML.Transforms/PermutationFeatureImportance.cs
TomFinley: This was marked as resolved, could you tell me @sfilipi if you added these extensions anywhere? The closest thing I saw that is new since my last review was `DataViewExtensions`, but that doesn't operate over `Schema.Column` as it should, but over `string` and `int`, which is very dangerous.

---
In reply to: [244385783](https://github.com/dotnet/machinelearning/pull/1975#discussion_r244385783) [](ancestors = 244385783,244384709,244382663)

src/Microsoft.ML.Data/Utilities/ColumnCursor.cs
```diff
@@ -99,7 +100,7 @@ private static IEnumerable<T> GetColumnDirect<T>(IDataView data, int col)
             Contracts.AssertValue(data);
             Contracts.Assert(0 <= col && col < data.Schema.Count);
 
-            using (var cursor = data.GetRowCursor(col.Equals))
+            using (var cursor = data.GetRowCursor(col))
```
TomFinley: >col [](start = 50, length = 3)

Once converted to be over schema, something like this would be `data.Schema[col]`, if we are to do this.

src/Microsoft.ML.Data/Data/RowCursorUtils.cs
```diff
@@ -488,6 +488,25 @@ public static IDataView RowAsDataView(IHostEnvironment env, Row row)
...
+        /// Given a collection of <see cref="Schema.Column"/>, that is a subset of the Schema of the data, create a predicate,
+        /// that when passed a column index, will return <langword>true</langword> or <langword>false</langword>, based on whether
+        /// the column with the given <see cref="Schema.Column.Index"/> is part of the <paramref name="colsNeeded"/>.
+        /// </summary>
+        /// <param name="colsNeeded">The subset of columns from the <see cref="Schema"/> that are needed from this <see cref="RowCursor"/>.</param>
+        /// <param name="schemaColsCardinality">The total number of <see cref="Schema.Column"/> in the <see cref="Schema"/>.</param>
+        [BestFriend]
+        internal static Func<int, bool> FromColumnsToPredicate(IEnumerable<Schema.Column> colsNeeded, int schemaColsCardinality)
+        {
+            if (colsNeeded == null)
```
TomFinley: >if (colsNeeded == null) [](start = 12, length = 23)

Do we intend to support `null` as a valid input? We never supported `null` as a valid input for a predicate.

src/Microsoft.ML.Data/Data/RowCursorUtils.cs
```diff
@@ -488,6 +488,25 @@ public static IDataView RowAsDataView(IHostEnvironment env, Row row)
...
+                return c => false;
+
+            if (colsNeeded.Count() == schemaColsCardinality)
```
TomFinley: >colsNeeded.Count() == schemaColsCardinality [](start = 16, length = 43)

That's not right, is it? If I have a dataview with 3 columns, and have `schema[0], schema[0], schema[2]`, you're hosed.

src/Microsoft.ML.Data/Data/RowCursorUtils.cs
```diff
@@ -488,6 +488,25 @@ public static IDataView RowAsDataView(IHostEnvironment env, Row row)
...
 
+        /// <summary>
+        /// Given a collection of <see cref="Schema.Column"/>, that is a subset of the Schema of the data, create a predicate,
+        /// that when passed a column index, will return <langword>true</langword> or <langword>false</langword>, based on whether
+        /// the column with the given <see cref="Schema.Column.Index"/> is part of the <paramref name="colsNeeded"/>.
+        /// </summary>
+        /// <param name="colsNeeded">The subset of columns from the <see cref="Schema"/> that are needed from this <see cref="RowCursor"/>.</param>
+        /// <param name="schemaColsCardinality">The total number of <see cref="Schema.Column"/> in the <see cref="Schema"/>.</param>
+        [BestFriend]
+        internal static Func<int, bool> FromColumnsToPredicate(IEnumerable<Schema.Column> colsNeeded, int schemaColsCardinality)
```
TomFinley: >int schemaColsCardinality [](start = 102, length = 25)

This utility should have included `Schema`, not just the cardinality. If it included the schema, there's a *lot* more validation and checking that we could do.

src/Microsoft.ML.Data/DataView/DataViewExtensions.cs
sfilipi: thanks for the catch! should switch to <see langword 

---
In reply to: [245081549](https://github.com/dotnet/machinelearning/pull/1975#discussion_r245081549) [](ancestors = 245081549)

src/Microsoft.ML.Data/DataView/DataViewExtensions.cs
sfilipi: you did suggest that its signature involves the Schema.Column and looks like this:

<code>GetRowCursor(IEnumerable<Schema.Column> colsNeeded)</code>

But since in most places the code looked like this:

1- Getting the indices of the columns needed, 
2- subset those indices from the schema
3- pass those columns to this method

i compacted the above by creating those methods that take in directly column names or indices. 

To enable that validation that you are suggesting, and encapsulating the columns sub setting from the schema here, i think the following should work, let me know if something is a miss:

1- Internalize this class, and its methods. 
2- keep the method signatures as they are, to centralize the way the schema gets accessed, and the validation
3- before accessing the columns on the schema, here, validate they exist and throw. 


---
In reply to: [245082390](https://github.com/dotnet/machinelearning/pull/1975#discussion_r245082390) [](ancestors = 245082390,245080660)

src/Microsoft.ML.Data/Data/RowCursorUtils.cs
sfilipi: We used to pass false when none of the columns was needed. 

I substituted that with null. Would you suggest something else? An empty collection?

---
In reply to: [245084779](https://github.com/dotnet/machinelearning/pull/1975#discussion_r245084779) [](ancestors = 245084779)

src/Microsoft.ML.Transforms/PermutationFeatureImportance.cs
sfilipi: Added now, will show up in the next PR update soon. 

---
In reply to: [245083396](https://github.com/dotnet/machinelearning/pull/1975#discussion_r245083396) [](ancestors = 245083396,244385783,244384709,244382663)

test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -4078,7 +4078,7 @@ public void TestSimpleTrainExperiment()
             var schema = data.Schema;
             var b = schema.TryGetColumnIndex("AUC", out int aucCol);
             Assert.True(b);
-            using (var cursor = data.GetRowCursor(col => col == aucCol))
+            using (var cursor = data.GetRowCursor(data.Schema.Where(col => col.Index == aucCol)))
```
sfilipi: >data.Schema.Where(col => col.Index == aucCol) [](start = 50, length = 45)

replace with data.Schema[aucCol] here and everywhere.

src/Microsoft.ML.Core/Data/IDataView.cs
```diff
@@ -85,10 +85,10 @@ public interface IDataView
         /// <summary>
         /// Get a row cursor. The active column indices are those for which needCol(col) returns true.
         /// The schema of the returned cursor will be the same as the schema of the IDataView, but getting
-        /// a getter for an inactive columns will throw. The <paramref name="needCol"/> predicate must be
-        /// non-null. To activate all columns, pass "col => true".
+        /// a getter for an inactive columns will throw. The <paramref name="colsNeeded"/> predicate must be
```
sfilipi: >predicate  [](start = 91, length = 10)

fix

src/Microsoft.ML.Core/Data/IDataView.cs
```diff
@@ -85,10 +85,10 @@ public interface IDataView
...
+        /// non-null.If set to <langword>null</langword> no column is active.
```
sfilipi: >If set to <langword>null</langword> no column is active. [](start = 21, length = 56)

fix

src/Microsoft.ML.Core/Data/IDataView.cs
```diff
@@ -106,13 +106,13 @@ public interface IDataView
         /// working threads that consume from them independently while, ultimately, the results are finally collated in
         /// the end by exploiting the ordering of the <see cref="Row.Batch"/> property described above. More typical
         /// scenarios will be content with pulling from the single serial cursor of
-        /// <see cref="GetRowCursor(Func{int, bool}, Random)"/>.
+        /// <see cref="GetRowCursor(IEnumerable{Schema.Column}, Random)"/>.
         /// </summary>
-        /// <param name="needCol">The predicate, where a column is active if this returns true.</param>
+        /// <param name="colsNeeded">The active columns needed. If set to <langword>null</langword> no column is active.</param>
```
sfilipi: ><langword>null</langword> [](start = 74, length = 25)

fix

src/Microsoft.ML.Data/DataView/Transposer.cs
```diff
@@ -438,7 +434,7 @@ protected override ValueGetter<VBuffer<T>> GetGetterCore()
                         Ch.Check(IsGood, "Cannot get values in the cursor's current state");
                         if (!valid)
                         {
-                            using (var cursor = _view.GetRowCursor(c => c == _col))
+                            using (var cursor = _view.GetRowCursor(new[] { _view.Schema[_col] }))
```
sfilipi: >new[] { _view.Schema[_col] } [](start = 67, length = 28)

remove array creation

src/Microsoft.ML.Data/DataView/Transposer.cs
```diff
@@ -588,7 +584,7 @@ private void EnsureValid()
                 // is having its values loaded into _indices/_values/_counts while the current column is being
                 // served up to the consumer through _cbuff.
 
-                using (var cursor = _view.GetRowCursor(c => c == _colCurr))
+                using (var cursor = _view.GetRowCursor(new[] { _view.Schema[_colCurr] }))
```
sfilipi: >new[] { _view.Schema[_colCurr]  [](start = 55, length = 31)

remove array

src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
```diff
@@ -678,8 +678,8 @@ private protected override void PrintFoldResultsCore(IChannel ch, Dictionary<str
             bool hasStratVals = overall.Schema.TryGetColumnIndex(MetricKinds.ColumnNames.StratVal, out stratVal);
             Contracts.Assert(hasStrat == hasStratVals);
             long numAnomalies = 0;
-            using (var cursor = overall.GetRowCursor(col => col == numAnomIndex ||
-                (hasStrat && col == stratCol)))
+            using (var cursor = overall.GetRowCursor(overall.Schema.Where(col => col.Name.Equals(AnomalyDetectionEvaluator.OverallMetrics.NumAnomalies)||
+                (hasStrat && col.Name.Equals(MetricKinds.ColumnNames.StratCol)))))
```
sfilipi: no linq

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -483,8 +483,8 @@ private void CompareSelectedR4VectorColumns(string leftColumnName, string rightC
             var leftColumnIndex = left.Schema[leftColumnName].Index;
             var rightColumnIndex = right.Schema[rightColumnName].Index;
 
-            using (var expectedCursor = left.GetRowCursor(columnIndex => leftColumnIndex == columnIndex))
-            using (var actualCursor = right.GetRowCursor(columnIndex => rightColumnIndex == columnIndex))
+            using (var expectedCursor = left.GetRowCursor(left.Schema.Where(columnIndex => leftColumnIndex == columnIndex.Index)))
```
sfilipi: >eft.Schema.Where(columnIndex => leftColumnIndex == columnIndex.Index))) [](start = 59, length = 71)

no linq needed, her eand below.

src/Microsoft.ML.Transforms/StatefulFilterTransform.cs
```diff
@@ -177,14 +178,7 @@ public Cursor(StatefulFilterTransform<TSrc, TDst, TState> parent, RowCursor<TSrc
...
 
-                Func<int, bool> appendedPredicate =
-                    col =>
-                    {
-                        col = _parent._bindings.AddedColumnIndices[col];
-                        return predicate(col);
-                    };
-
-                _appendedRow = appendedDataView.GetRowCursor(appendedPredicate);
+                _appendedRow = appendedDataView.GetRowCursor(_parent._bindings.Schema);
```
sfilipi: >_parent._bindings.Schema [](start = 61, length = 24)

fix to contains only columns for predicate true

sfilipi: Double-check this is right. 

---
In reply to: [246108558](https://github.com/dotnet/machinelearning/pull/1975#discussion_r246108558) [](ancestors = 246108558)

sfilipi:
@TomFinley @Ivanidzo4ka @yaeldekel, when you have a moment. Thanks.

src/Microsoft.ML.PCA/PcaTransform.cs
```diff
@@ -453,7 +453,8 @@ private void Project(IDataView trainingData, float[][] mean, float[][][] omega,
                     activeColumns[sInfo.WeightColumnIndex] = true;
             }
 
-            using (var cursor = trainingData.GetRowCursor(col => activeColumns[col]))
+            var inputCols = trainingData.Schema.Where(x => activeColumns[x.Index]);
```
Ivanidzo4ka: >activeColumns [](start = 59, length = 13)

Is it possible to replace activeColumns with ColumnPairs->Input + WeightColumnIndex? Or worth it?

src/Microsoft.ML.HalLearners/OlsLinearRegression.cs
```diff
@@ -155,7 +155,11 @@ private protected override OlsLinearRegressionModelParameters TrainModelCore(Tra
                 if (typeFeat.ItemType != NumberType.Float)
                     throw ch.Except("Incompatible feature column type {0}, must be vector of {1}", typeFeat, NumberType.Float);
 
-                var cursorFactory = new FloatLabelCursor.Factory(examples, CursOpt.Label | CursOpt.Features);
+                CursOpt cursorOpt = CursOpt.Label | CursOpt.Features;
+                if (examples.Schema.Weight.HasValue)
+                    cursorOpt |= CursOpt.Weight;
+
+                var cursorFactory = new FloatLabelCursor.Factory(examples, cursorOpt);//sefilipi: double-check on the change.
```
Ivanidzo4ka: >//sefilipi: double-check on the change [](start = 86, length = 38)

do we need it?

src/Microsoft.ML.Data/DataView/DataViewExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// Get a row cursor. The <paramref name="colNeeded"/> is the active column.
+        /// The schema of the returned cursor will be the same as the schema of the IDataView, but getting
+        /// a getter for the other, inactive columns will throw.
+        /// </summary>
+        /// <param name="colNeeded">The column requested by this <see cref="RowCursor"/>, or as otherwise called, the active column.</param>
+        /// <param name="dv">The <see cref="IDataView"/> containing the <paramref name="colNeeded"/>.</param>
+        [BestFriend]
+        internal static RowCursor GetRowCursor(this IDataView dv, Schema.Column colNeeded)
```
Ivanidzo4ka: >Schema.Column colNeeded [](start = 66, length = 23)

I was under impression what `params something[]` let you pass one entity of something and you don't need cases for one value.

sfilipi: True, but Tom suggested we have those because they are more performant than calling the signature with params.

src/Microsoft.ML.Data/Data/RowCursorUtils.cs
```diff
@@ -464,6 +464,30 @@ public static IDataView RowAsDataView(IHostEnvironment env, Row row)
...
+        internal static Func<int, bool> FromColumnsToPredicate(IEnumerable<Schema.Column> colsNeeded, Schema sourceSchema)
+        {
+            Contracts.CheckValue(colsNeeded, nameof(colsNeeded));
+            Contracts.CheckValue(sourceSchema, nameof(sourceSchema));
+
+            bool[] indicesRequested = new bool[sourceSchema.Count];
+
+            foreach (var col in colsNeeded)
+            {
+                Contracts.Assert(col.Index < indicesRequested.Length, $"The requested column: {col} is not part of the {nameof(sourceSchema)}");
```
Ivanidzo4ka: >Assert [](start = 26, length = 6)

it's a bit strange to have mix of Check and Assert in code

src/Microsoft.ML.Data/Data/DataViewUtils.cs
```diff
@@ -194,6 +193,29 @@ public static bool AllCachable(Schema schema, Func<int, bool> predicate)
             return true;
         }
 
+        /// <summary>
+        /// Return whether all the active columns, as determined by the predicate, are
+        /// cachable - either primitive types or vector types.
+        /// </summary>
+        public static bool AllCachable(Schema schema, IEnumerable<Schema.Column> colsNeeded)
```
Ivanidzo4ka: >Cachable [](start = 30, length = 8)

Please ignore my comment, unless you have some free time:
[https://lists.w3.org/Archives/Public/ietf-http-wg-old/1997SepDec/0003.html](https://lists.w3.org/Archives/Public/ietf-http-wg-old/1997SepDec/0003.html)

I also look in google index and "Cachable" appears 148 000 times while "Cacheable" 876 000.

src/Microsoft.ML.Data/DataView/DataViewExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+            list.Add(colNeeded);
+
+            return dv.GetRowCursor(list);
+        }
+
+        /// <summary>
+        /// Get a row cursor. No colums are needed by this <see cref="RowCursor"/>.
+        /// </summary>
+        [BestFriend]
+        internal static RowCursor GetRowCursor(this IDataView dv) =>  dv.GetRowCursor(new List<Schema.Column>());
```
Ivanidzo4ka: Any chance you want to have GetRowCursorForAll, or GetFullRowCursor, or someother name which would return cursor on all columns?
From what I see in code, it's quite popular usage.

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -751,6 +763,41 @@ public static void BuildSubsetMaps(int lim, Func<int, bool> pred, out int[] map,
...
+            for (int c = 0; c < lim; ++c)
+            {
+                if (colsNeeded == null || !colsNeeded.Any(x => x.Index == c))
+                {
+                    invMap[c] = -1;
+                    continue;
+                }
+                invMap[c] = mapList.Count;
+                mapList.Add(c);
+            }
```
Ivanidzo4ka: Can you have foreach loop over colsNeeded instead of this. No ones guarantee you cached in memory IEnumerable, it can be anything including some over network connection collection and with trigger to start nuclear war if you access it multiple times within short period of time. You don't want to start nuclear war, right?

src/Microsoft.ML.Data/DataLoadSave/PartitionedFileLoader.cs
```diff
@@ -364,29 +364,35 @@ private sealed class Cursor : RootCursorBase
...
                 _parent = parent;
+                _colsNeeded = colsNeeded;
 
-                _active = Utils.BuildArray(Schema.Count, predicate);
+                _active = Utils.BuildArray(Schema.Count, colsNeeded);
                 _subActive = _active.Take(SubColumnCount).ToArray();
                 _colValues = new ReadOnlyMemory<char>[Schema.Count - SubColumnCount];
 
+                // sefilipi: Test this
+                _subActiveColsNeeded = Schema.Where(x => (_subActive?.Length > x.Index) && _subActive[x.Index]);
```
Ivanidzo4ka: >_subActiveColsNeeded [](start = 16, length = 20)

is this same as `Utils.BuildArray(SubColumCount, colsNeeded);` ?

src/Microsoft.ML.Data/DataView/AppendRowsDataView.cs
```diff
@@ -209,17 +209,15 @@ private sealed class Cursor : CursorBase
...
                 : base(parent)
             {
-                Ch.AssertValue(needCol);
-
                 _currentSourceIndex = 0;
-                _currentCursor = Sources[_currentSourceIndex].GetRowCursor(needCol);
+                _currentCursor = Sources[_currentSourceIndex].GetRowCursor(colsNeeded);
                 _currentIdGetter = _currentCursor.GetIdGetter();
                 for (int c = 0; c < Getters.Length; c++)
                 {
```
Ivanidzo4ka: foreach over colsNeeded please

src/Microsoft.ML.Data/DataView/AppendRowsDataView.cs
```diff
@@ -314,13 +313,13 @@ public RandCursor(AppendRowsDataView parent, Func<int, bool> needCol, Random ran
                 for (int i = 0; i < counts.Length; i++)
                 {
                     Ch.Assert(counts[i] >= 0);
-                    _cursorSet[i] = parent._sources[i].GetRowCursor(needCol, RandomUtils.Create(_rand));
+                    _cursorSet[i] = parent._sources[i].GetRowCursor(colsNeeded, RandomUtils.Create(_rand));
                 }
                 _sampler = new MultinomialWithoutReplacementSampler(Ch, counts, rand);
                 _currentSourceIndex = -1;
                 for (int c = 0; c < Getters.Length; c++)
```
Ivanidzo4ka: >for (int c = 0; c < Getters.Length; c++) [](start = 16, length = 40)

foreach over colsNeeded please.

src/Microsoft.ML.Data/DataView/DataViewExtensions.cs
sfilipi: True, but Tom suggested we have those because they are more performant than calling the signature with params.

---
In reply to: [246949003](https://github.com/dotnet/machinelearning/pull/1975#discussion_r246949003) [](ancestors = 246949003)

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -706,6 +708,16 @@ public static T[] CreateArray<T>(int length, T value)
             return result;
         }
 
+        public static bool[] BuildArray(int length, IEnumerable<Schema.Column> colsNeeded)
+        {
+            Contracts.CheckParam(length >= 0, nameof(length));
+
+            var result = new bool[length];
+            for (int i = 0; i < result.Length; i++)
+                result[i] = colsNeeded == null ? false : colsNeeded.Any(x => x.Index == i);
```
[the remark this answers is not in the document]
sfilipi: ah, good catch! I am doing that somewhere else!

---
In reply to: [246979144](https://github.com/dotnet/machinelearning/pull/1975#discussion_r246979144) [](ancestors = 246979144)

src/Microsoft.ML.Data/Data/RowCursorUtils.cs
TomFinley: Let's review what the three types of things are for:

1. Asserts are basically code documentation. You rarely, to the point of almost never, want string messages associated with them. That's not what they're for.
2. Checks are for things that are actually intended to be runtime compiletime checked -- that is, something that might be checked. But, because C# like nearly all languages does not do lazy evaluation of arguments, I have forbidden passing anything *except* a string literal, since the message is formatted, strings allocated, and everything, whether you actually wind up wanting the error message *or not*.
3. You will note throughout the codebase the use of `if (b) throw Contracts.Except` or something similar in the case. This is primarily for exactly this case where we want to throw a complex message, but we only want to pay the cost of formatting the message in the case where it is actually needed. So this is what you should use here.

So please throw one of the excepts. Assert is wrong since it is for code documentation, documenting invariants, and so on. (Is `ExceptSchemaMismatch` inappropriate here? Maybe it throws the wrong type of message.)

We could be and probably should be a *lot* more aggressive with our checks FYI. Not only should the index be *in range* but the fields or properties of the `Schema.Column` at that index should be identical.

---
In reply to: [247006403](https://github.com/dotnet/machinelearning/pull/1975#discussion_r247006403) [](ancestors = 247006403,246950288)

src/Microsoft.ML.Data/DataView/DataViewExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <summary>
+        /// Get a row cursor. The <paramref name="colsNeeded"/> are the active columns.
+        /// The schema of the returned cursor will be the same as the schema of the IDataView, but getting
+        /// a getter for an inactive columns will throw.
+        /// </summary>
+        /// <param name="colsNeeded">The columns requested by this <see cref="RowCursor"/>, or as otherwise called, the active columns.
+        /// An empty collection indicates that no column is needed.</param>
+        /// <param name="dv">The <see cref="IDataView"/> containing the <paramref name="colsNeeded"/>.</param>
+        [BestFriend]
+        internal static RowCursor GetRowCursor(this IDataView dv, params Schema.Column[] colsNeeded)
```
TomFinley: This class appears to already be internal. So, can we simplify this by just getting rid of the `[BestFriend]` on the methods, and making the methods public? That's what we do everywhere else I believe.

src/Microsoft.ML.Data/DataView/DataViewExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <param name="colNeeded">The column requested by this <see cref="RowCursor"/>, or as otherwise called, the active column.</param>
+        /// <param name="dv">The <see cref="IDataView"/> containing the <paramref name="colNeeded"/>.</param>
+        [BestFriend]
+        internal static RowCursor GetRowCursor(this IDataView dv, Schema.Column colNeeded)
+        {
+            Contracts.Assert(dv.Schema[colNeeded.Index].Equals(colNeeded), $"The requested column named: {colNeeded.Name}, with index: {colNeeded.Index} and type: {colNeeded.Type}" +
+                   $" is not present in the {nameof(IDataView)} where the {nameof(RowCursor)} is being requested.");
+
+            var list = new List<Schema.Column>();
+            list.Add(colNeeded);
```
TomFinley: So, the point of introducing this method was to avoid the allocation of the array via a `params`. But, instead what we're doing is *far worse*, we are allocating a list, then adding to it, which will wind up being far slower. There is no point in having this overload in that case, since that is *far* worse than the params array.

src/Microsoft.ML.Data/DataView/DataViewExtensions.cs
TomFinley: Well, it can be, under the appropriate conditions. :)

---
In reply to: [247005512](https://github.com/dotnet/machinelearning/pull/1975#discussion_r247005512) [](ancestors = 247005512,246949003)

src/Microsoft.ML.Data/DataView/DataViewExtensions.cs
sfilipi: This also answers my question on Eric's previous PR :)

---
In reply to: [247007778](https://github.com/dotnet/machinelearning/pull/1975#discussion_r247007778) [](ancestors = 247007778)

src/Microsoft.ML.Data/DataLoadSave/PartitionedFileLoader.cs
sfilipi: that returns an array, this is an IEnumerable<Schema.Column>

---
In reply to: [246980849](https://github.com/dotnet/machinelearning/pull/1975#discussion_r246980849) [](ancestors = 246980849)

src/Microsoft.ML.Data/Data/DataViewUtils.cs
sfilipi: since you spent time looking this up, I'll rename it!

---
In reply to: [246952100](https://github.com/dotnet/machinelearning/pull/1975#discussion_r246952100) [](ancestors = 246952100)

Ivanidzo4ka: Please, don't this is not only place in code where we call it cachable, and i don't want you to spend your time on such minutiae.

---
In reply to: [247017070](https://github.com/dotnet/machinelearning/pull/1975#discussion_r247017070) [](ancestors = 247017070,246952100)

Ivanidzo4ka: Only reason why I left comment is link with description of rule and exception out of it for using or not using "e".

---
In reply to: [247017629](https://github.com/dotnet/machinelearning/pull/1975#discussion_r247017629) [](ancestors = 247017629,247017070,246952100)

sfilipi: Rename does it for us, it is just 2 clicks and typing one letter !

---
In reply to: [247017629](https://github.com/dotnet/machinelearning/pull/1975#discussion_r247017629) [](ancestors = 247017629,247017070,246952100)

src/Microsoft.ML.Core/Utilities/Utils.cs
sfilipi: Hummmm, idk if i want to start a nuclear war. I'll go literally sleep on it :)

---
In reply to: [246979630](https://github.com/dotnet/machinelearning/pull/1975#discussion_r246979630) [](ancestors = 246979630)

src/Microsoft.ML.Data/Data/DataViewUtils.cs
```diff
@@ -187,7 +186,30 @@ public static bool AllCachable(Schema schema, Func<int, bool> predicate)
...
+        /// cachable - either primitive types or vector types.
+        /// </summary>
+        public static bool AllCacheable(Schema schema, IEnumerable<Schema.Column> colsNeeded)
+        {
+            Contracts.CheckValue(schema, nameof(schema));
+
+            if (colsNeeded == null)
+                return false;
+
+            for (int col = 0; col < schema.Count; col++)
```
Ivanidzo4ka: >for (int col = 0; col < schema.Count; col++) [](start = 12, length = 44)

can you use foreach over colsNeeded?

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -822,7 +822,9 @@ public static ICalibrator TrainCalibrator(IHostEnvironment env, IChannel ch, ICa
                 if (data.Schema.Weight?.Name is string weightName && scored.Schema.GetColumnOrNull(weightName)?.Index is int weightIdx)
                     weightCol = weightIdx;
                 ch.Info("Training calibrator.");
-                using (var cursor = scored.GetRowCursor(col => col == labelCol || col == scoreCol || col == weightCol))
+
+                var cols = weightCol > -1 ? new Schema.Column[] { scored.Schema[labelCol], scored.Schema[scoreCol], scored.Schema[weightCol] } : new Schema.Column[] { scored.Schema[labelCol], scored.Schema[scoreCol] };
```
Ivanidzo4ka: >var cols = weightCol > -1 ? new Schema.Column[] { scored.Schema[labelCol], scored.Schema[scoreCol], scored.Schema[weightCol] } : new Schema.Column[] { scored.Schema[labelCol], scored.Schema[scoreCol] }; [](start = 16, length = 202)

some formatting would be nice here.

src/Microsoft.ML.TimeSeries/AdaptiveSingularSpectrumSequenceModeler.cs
```diff
@@ -1232,13 +1232,14 @@ internal override void Train(RoleMappedData data)
                 throw _host.ExceptSchemaMismatch(nameof(data), "feature", featureCol.Name, "R4", featureCol.Type.ToString());
 
             Single[] dataArray = new Single[_trainSize];
-            int col = featureCol.Index;
+            var col = new List<Schema.Column>();
+            col.Add(featureCol);
```
Ivanidzo4ka: why you need collection if you put only one element into it?

src/Microsoft.ML.Transforms/StatefulFilterTransform.cs
```diff
@@ -178,13 +179,13 @@ public Cursor(StatefulFilterTransform<TSrc, TDst, TState> parent, RowCursor<TSrc
...
-                        return predicate(col);
-                    };
+                col =>
+                {
+                    col = _parent._bindings.AddedColumnIndices[col];
+                    return predicate(col);
+                };
 
-                _appendedRow = appendedDataView.GetRowCursor(appendedPredicate);
+                _appendedRow = appendedDataView.GetRowCursorForAllColumns();
```
Ivanidzo4ka: >GetRowCursorForAllColumns [](start = 48, length = 25)

it doesn't look the same as before.
And why we even need that appendedPredicate one line above?

src/Microsoft.ML.Transforms/Text/NgramHashingTransformer.cs
```diff
@@ -401,11 +401,11 @@ internal void Save(ModelSaveContext ctx)
                 var active = new bool[1];
                 string[][] friendlyNames = _columns.Select(c => c.FriendlyNames).ToArray();
                 // We will create invert hash helper class, which would store in itself all original ngrams and their mapping into hash values.
-                var helper = new InvertHashHelper(this, input.Schema, friendlyNames, sourceColumnsForInvertHash.Contains, invertHashMaxCounts);
+                var helper = new InvertHashHelper(this, input.Schema, friendlyNames, c => sourceColumnsForInvertHash.Any(col => col.Index == c), invertHashMaxCounts);
```
Ivanidzo4ka: >c => sourceColumnsForInvertHash.Any(col => col.Index == c) [](start = 85, length = 58)

or you can just modify InvertHashHelper and accept list of columns instead of predicate. It's get used once.

test/Microsoft.ML.Benchmarks/CacheDataViewBench.cs
```diff
@@ -59,17 +60,19 @@ public void Setup()
...
-            _col = _cacheDataView.Schema.GetColumnOrNull("A").Value;
+            _col = _cacheDataView.Schema["A"];
             _seeker = ((IRowSeekable)_cacheDataView).GetSeeker(colIndex => colIndex == _col.Index);
             _seekerGetter = _seeker.GetGetter<int>(_col.Index);
         }
 
         [Benchmark]
         public void CacheWithCursor()
         {
+            var cols = new List<Schema.Column>();
```
Ivanidzo4ka: >cols [](start = 16, length = 4)

Why we create collection for one element?

test/Microsoft.ML.OnnxTransformTest/DnnImageFeaturizerTest.cs
```diff
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
```
Ivanidzo4ka: Where this dependency comes from?

src/Microsoft.ML.Data/Transforms/NAFilter.cs
```diff
@@ -204,25 +205,27 @@ private static bool TestType(ColumnType type)
...
         }
 
-        public override RowCursor[] GetRowCursorSet(Func<int, bool> predicate, int n, Random rand = null)
+        public override RowCursor[] GetRowCursorSet(IEnumerable<Schema.Column> colsNeeded, int n, Random rand = null)
         {
-            Host.CheckValue(predicate, nameof(predicate));
             Host.CheckValueOrNull(rand);
+            var predicate = RowCursorUtils.FromColumnsToPredicate(colsNeeded, OutputSchema);
+
+            Func<int, bool> inputPred = GetActive(predicate, out bool[] active);
```
Ivanidzo4ka: So I see two reasons why you have this `FromColumnsToPredicate` method.
One is because in IRowToRowMapper we have 
`Func<int, bool> GetDependencies(Func<int, bool> predicate);` and `Row GetRow(Row input, Func<int, bool> active);`
And since you haven't change it, you need to create predicate.
I hope in later PRs you will change all predicates to list of columns, but I understand what you have to do it in chunks.

Second is case like this where you have `GetActive` function which accepts predicate and returns predicate. 
It's not part of any interface as far as I can see, and it's just a patter we have in our code.
I don't think such private local `GetActive` functions should continue to work with predicates.
If end goal to get rid of predicates in general, I don't see in what other PRs you will touch them again. Maybe when you decide to kill `FromColumnsToPredicate` method. Of maybe if you search for `Func<int,bool>` and notice their presence.

I'm fine with this code if you actually plan to remove all predicates you will eventually change `GetActive` method. Just make sure you take care of them.

sfilipi: Thanks for writing this up, @ivanidzo4ka. I started cleaning up the predicated everywhere, and backed up because 1- the PR would be even bigger. 2- Testing it would be harder. 

Plan is to check this in, and start removing the predicates from GetDependencies, GetRow which is part of the public IRowToRowMapper. With those remove the utility function. 

---
In reply to: [247212899](https://github.com/dotnet/machinelearning/pull/1975#discussion_r247212899) [](ancestors = 247212899)

test/Microsoft.ML.Benchmarks/CacheDataViewBench.cs
sfilipi: leftover from before the utilities. thanks for the catch, i'll review it all again. 

---
In reply to: [247208960](https://github.com/dotnet/machinelearning/pull/1975#discussion_r247208960) [](ancestors = 247208960)

src/Microsoft.ML.Transforms/StatefulFilterTransform.cs
sfilipi: Should go. removed it, than added it back for debugging.. thanks. 

---
In reply to: [247208022](https://github.com/dotnet/machinelearning/pull/1975#discussion_r247208022) [](ancestors = 247208022)

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -751,6 +767,40 @@ public static void BuildSubsetMaps(int lim, Func<int, bool> pred, out int[] map,
...
+
+            // REVIEW: Better names?
+            List<int> mapList = new List<int>();
+            invMap = invMap = Enumerable.Repeat(-1, lim).ToArray<int>();
+
+            foreach (var col in colsNeeded)
+            {
+                Contracts.Check(col.Index < lim);
+                mapList.Add(col.Index);
+                invMap[col.Index] = mapList.Count;
```
sfilipi: swap order.

src/Microsoft.ML.PCA/PcaTransform.cs
sfilipi: looking into. I'll update in another iteration. 

---
In reply to: [246938070](https://github.com/dotnet/machinelearning/pull/1975#discussion_r246938070) [](ancestors = 246938070)

src/Microsoft.ML.Data/Data/RowCursorUtils.cs
```diff
@@ -469,6 +469,35 @@ public static IDataView RowAsDataView(IHostEnvironment env, Row row)
...
+
+            foreach (var col in colsNeeded)
+            {
+                if(sourceSchema.GetColumnOrNull(col.Name) == null)
+                    throw Contracts.Except($"The requested column: {col} is not part of the {nameof(sourceSchema)}");
+
+                indicesRequested.Add(col.Index);
+            }
+
+            return c => indicesRequested.Contains(c);
```
sfilipi: the implementation on iteration 7 - 9 was different; bool[] for the indices. Switched to this because noticed that the indices of the colsNeeded in the appendedDataView of the StatefulFilterTransform, did not match the index of the same column name and type on the appended data view. Thinking back, the problem is/was in the StatefulFilterTransform.  Revert this back.

Anipik:
@sfilipi I am not aware if we have any hard rules of not merging the pr due to some regression. but it would be nice to find the reason behind the regression using some profiler

src/Microsoft.ML.PCA/PcaTransform.cs
sfilipi: I deem it not worth it. To figure out which of the  trainingData.Schema column, we'd still need to iterate/LINQ over all the sInfo and select all indexes, find those indexes in the trainingData.Schema columns. 

Feels similar, as far as complexity, to me. 

---
In reply to: [247287016](https://github.com/dotnet/machinelearning/pull/1975#discussion_r247287016) [](ancestors = 247287016,246938070)

sfilipi:
I am running the profiler for the MakeIrisPredictions benchmarks, that has the most change before and after, and the numbers gotten for various runs on the same branch vary more than the numbers on the before/after change comparison:

running time 1 -  142226 
running time 2 -  119337 

I compared the profiling cpu unit measurements, and % time spent in the before and after profiler trace and there were no meaningful changes:

Before the change:
Function Name Total CPU [unit, %] Self CPU [unit, %] Module
| - Microsoft.ML.Training.TrainerUtils::CreateRowCursor 84906 (55.18%) 7 (0.00%) Microsoft.ML.Data.dll
| - Microsoft.ML.Data.TransformBase::GetRowCursor 84773 (55.10%) 9 (0.01%) Microsoft.ML.Data.dll

After the change:
Function Name Total CPU [unit, %] Self CPU [unit, %] Module
| - Microsoft.ML.Training.TrainerUtils::CreateRowCursor 65433 (54.83%) 6 (0.01%) Microsoft.ML.Data.dll
| - Microsoft.ML.Data.TransformBase::GetRowCursor 65330 (54.74%) 15 (0.01%) Microsoft.ML.Data.dll



---
In reply to: [454118319](https://github.com/dotnet/machinelearning/pull/1975#issuecomment-454118319) [](ancestors = 454118319)

src/Microsoft.ML.Core/Data/IDataView.cs
```diff
@@ -85,10 +86,10 @@ public interface IDataView
         /// <summary>
         /// Get a row cursor. The active column indices are those for which needCol(col) returns true.
         /// The schema of the returned cursor will be the same as the schema of the IDataView, but getting
-        /// a getter for an inactive columns will throw. The <paramref name="needCol"/> predicate must be
-        /// non-null. To activate all columns, pass "col => true".
+        /// a getter for inactive columns will throw. The <paramref name="colsNeeded"/> indicate the columns that are needed
+        /// to iterate over.If set to an empty <see cref="IEnumerable"/> no column is requested.
         /// </summary>
-        RowCursor GetRowCursor(Func<int, bool> needCol, Random rand = null);
+        RowCursor GetRowCursor(IEnumerable<Schema.Column> colsNeeded, Random rand = null);
```
eerhardt: Can you spell out the parameter name? `colsNeeded` => `columnsNeeded`.

[https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions)

X DO NOT use abbreviations or contractions as part of identifier names.

For example, use GetWindow rather than GetWin.

src/Microsoft.ML.Transforms/StatefulFilterTransform.cs
```diff
@@ -171,20 +172,14 @@ public Cursor(StatefulFilterTransform<TSrc, TDst, TState> parent, RowCursor<TSrc
                 CursorChannelAttribute.TrySetCursorChannel(_parent.Host, _dst, Ch);
                 CursorChannelAttribute.TrySetCursorChannel(_parent.Host, _state, Ch);
 
-                if (parent._initStateAction != null)
-                    parent._initStateAction(_state);
+                parent._initStateAction?.Invoke(_state);
 
                 var appendedDataView = new DataViewConstructionUtils.SingleRowLoopDataView<TDst>(parent.Host, _parent._addedSchema);
                 appendedDataView.SetCurrentRowObject(_dst);
+                var appendedIndices = _parent._bindings.AddedColumnIndices;
```
sfilipi: >appendedIndices [](start = 20, length = 15)

remove

src/Microsoft.ML.Transforms/StatefulFilterTransform.cs
```diff
@@ -171,20 +172,14 @@ public Cursor(StatefulFilterTransform<TSrc, TDst, TState> parent, RowCursor<TSrc
...
-                Func<int, bool> appendedPredicate =
-                    col =>
-                    {
-                        col = _parent._bindings.AddedColumnIndices[col];
-                        return predicate(col);
-                    };
-
-                _appendedRow = appendedDataView.GetRowCursor(appendedPredicate);
+                var cols = colsNeeded.Select(c => c.Name);
+                _appendedRow = appendedDataView.GetRowCursor(appendedDataView.Schema.Where(c => cols.Contains(c.Name)));
```
TomFinley: >cols.Contains(c.Name) [](start = 96, length = 21)

Consider also paying attention only to non-hidden columns, just to make things a bit more restrictive? Does this make sense?

src/Microsoft.ML.Transforms/StatefulFilterTransform.cs
```diff
@@ -154,11 +155,11 @@ private sealed class Cursor : RootCursorBase
 
             public override long Batch => _input.Batch;
 
-            public Cursor(StatefulFilterTransform<TSrc, TDst, TState> parent, RowCursor<TSrc> input, Func<int, bool> predicate)
+            public Cursor(StatefulFilterTransform<TSrc, TDst, TState> parent, RowCursor<TSrc> input, IEnumerable<Schema.Column> colsNeeded)
```
sfilipi: >colsNeeded [](start = 128, length = 10)

columnsNeeded

src/Microsoft.ML.Transforms/StatefulFilterTransform.cs
TomFinley: Actually we can simplify this is a lot by just having `cols.Select(c => appendedDataView.Schema[c.Name])`. Which is an `IEnumerable` and far more clear about your intentions, I believe.

---
In reply to: [248452752](https://github.com/dotnet/machinelearning/pull/1975#discussion_r248452752) [](ancestors = 248452752)

TomFinley:
> cc @Anipik, @TomFinley @eerhardt @Ivanidzo4ka what do you think; is there more investigation needed in your opinion?

Nothing leaps out to me, except the `PredictIris` things. I kind of feel like if it is somehow attributable to this, given that only the public surface has not changed and not the underlying infrastructure, I'd like to understand why, since that doesn't make much sense to me.

I'd probably re-run those, see if it isn't something spurious (like, say, Windows update picked that moment to start a background install, or whatever), and then profile and see if it can be traced to the new predicates. If neither of those, then, I'd probably say screw it and move on.

I think that work (nailing the perf of the underlying predicate building) can be done independently of this work (getting public API surface right), so I'd be inclined to check this in first, then fix the perf problem later, since if there is a problem it seems incredibly specialized.

Those three specifically are valuable benchmarks though. In fact for this change, considering that they are very "cursor creation heavy" I might say they're really among the only benchmarks of interest. (The end to end benchmarks I consider more or less worthless to judge your change.)

src/Microsoft.ML.Transforms/StatefulFilterTransform.cs
sfilipi: Keeping it to Where, to check for IsHidden. Can't do that with Select. 

---
In reply to: [248453731](https://github.com/dotnet/machinelearning/pull/1975#discussion_r248453731) [](ancestors = 248453731,248452752)

sfilipi:
Re-basing this is a bit of work, I'd love to close on it, if someone else has time to take a second look.

src/Microsoft.ML.Data/DataView/DataViewExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        /// <summary>
+        /// Get a row cursor. The <paramref name="colNeeded"/> is the active column.
+        /// The schema of the returned cursor will be the same as the schema of the IDataView, but getting
+        /// a getter for the other, inactive columns will throw.
+        /// </summary>
+        /// <param name="colNeeded">The column requested by this <see cref="RowCursor"/>, or as otherwise called, the active column.</param>
+        /// <param name="dv">The <see cref="IDataView"/> containing the <paramref name="colNeeded"/>.</param>
+        public static RowCursor GetRowCursor(this IDataView dv, Schema.Column colNeeded)
```
Ivanidzo4ka: >colNeeded [](start = 78, length = 9)

columnNeed :)

test/Microsoft.ML.Tests/Scenarios/Api/TestApi.cs
```diff
@@ -73,7 +73,7 @@ public void CursorChannelExposedInMapTransform()
                     Assert.NotNull(input.Channel);
                     return false;
                 }, null);
-            filter1.GetRowCursor(col => true).MoveNext();
+            filter1.GetRowCursor(filter1.Schema).MoveNext();
```
Ivanidzo4ka: >GetRowCursor [](start = 20, length = 12)

GetRowCursorForAll ?

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -741,7 +742,7 @@ public void TensorFlowTransformCifar()
             IDataView trans = new TensorFlowTransformer(mlContext, tensorFlowModel, "Input", "Output").Transform(pixels);
 
             trans.Schema.TryGetColumnIndex("Output", out int output);
-            using (var cursor = trans.GetRowCursor(col => col == output))
+            using (var cursor = trans.GetRowCursorForAllColumns())
```
Ivanidzo4ka: >GetRowCursorForAllColumns [](start = 38, length = 25)

GetRowCursor(trans.Schema["Output"]) and you can kill output variable above.

src/Microsoft.ML.Data/DataView/DataViewExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+            Contracts.Assert(dv.Schema[colNeeded.Index].Equals(colNeeded), $"The requested column named: {colNeeded.Name}, with index: {colNeeded.Index} and type: {colNeeded.Type}" +
+                   $" is not present in the {nameof(IDataView)} where the {nameof(RowCursor)} is being requested.");
+
+            return dv.GetRowCursor(Enumerable.Repeat(colNeeded,1));
+        }
+
+        /// <summary>
+        /// Get a row cursor. No colums are needed by this <see cref="RowCursor"/>.
+        /// </summary>
+        public static RowCursor GetRowCursor(this IDataView dv) =>  dv.GetRowCursor(new HashSet<Schema.Column>());
```
Ivanidzo4ka: >new HashSet [](start = 84, length = 11)

Enumerable.Empty<Schema.Column> probably more readable than HashSet

src/Microsoft.ML.Transforms/StatefulFilterTransform.cs
TomFinley: Hi @sfilipi , bringiing this up again... Absolutely you can. The string indexer can *never* return a hidden column. That's what hidden means!

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
sfilipi: Uh.. thanks. 

---
In reply to: [249123646](https://github.com/dotnet/machinelearning/pull/1975#discussion_r249123646) [](ancestors = 249123646)

src/Microsoft.ML.Transforms/StatefulFilterTransform.cs
sfilipi: Keeping it as is for a different reason: the columns need to be selected from the appendedDataview, and not from cols needed, because the indexes are different... 

---
In reply to: [249126662](https://github.com/dotnet/machinelearning/pull/1975#discussion_r249126662) [](ancestors = 249126662)

PR #1979, opened 2018-12-28
making GetCoefficientStatistics public

### LINKED ISSUE #1977: GetCoefficientStatistics in Microsoft.ML.Learners.LinearModelStatistics  needs to be public
[GetCoefficientStatistics](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs#L327) in Microsoft.ML.Learners.LinearModelStatistics  needs to be publicly accessible. 

The statistics calculated provide no value, if they can't be accessed.

sfilipi:
Fixes #1977 by making method GetCoefficientStatistics  public.

### DIFF
--- a/src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs
+++ b/src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs
@@ -27,12 +27,12 @@ namespace Microsoft.ML.Learners
     public readonly struct CoefficientStatistics
     {
         public readonly string Name;
-        public readonly Single Estimate;
-        public readonly Single StandardError;
-        public readonly Single ZScore;
-        public readonly Single PValue;
+        public readonly float Estimate;
+        public readonly float StandardError;
+        public readonly float ZScore;
+        public readonly float PValue;
 
-        public CoefficientStatistics(string name, Single estimate, Single stdError, Single zScore, Single pValue)
+        public CoefficientStatistics(string name, float estimate, float stdError, float zScore, float pValue)
         {
             Contracts.AssertNonEmpty(name);
             Name = name;
@@ -69,10 +69,10 @@ namespace Microsoft.ML.Learners
         private readonly long _trainingExampleCount;
 
         // The deviance of this model.
-        private readonly Single _deviance;
+        private readonly float _deviance;
 
         // The deviance of the null hypothesis.
-        private readonly Single _nullDeviance;
+        private readonly float _nullDeviance;
 
         // Total count of parameters.
         private readonly int _paramCount;
@@ -82,17 +82,17 @@ namespace Microsoft.ML.Learners
         // It could be null when there are too many non-zero weights so that
         // the memory is insufficient to hold the Hessian matrix necessary for the computation
         // of the variance-covariance matrix.
-        private readonly VBuffer<Single>? _coeffStdError;
+        private readonly VBuffer<float>? _coeffStdError;
 
         public long TrainingExampleCount => _trainingExampleCount;
 
-        public Single Deviance => _deviance;
+        public float Deviance => _deviance;
 
-        public Single NullDeviance => _nullDeviance;
+        public float NullDeviance => _nullDeviance;
 
         public int ParametersCount => _paramCount;
 
-        internal LinearModelStatistics(IHostEnvironment env, long trainingExampleCount, int paramCount, Single deviance, Single nullDeviance)
+        internal LinearModelStatistics(IHostEnvironment env, long trainingExampleCount, int paramCount, float deviance, float nullDeviance)
         {
             Contracts.AssertValue(env);
             env.Assert(trainingExampleCount > 0);
@@ -104,7 +104,7 @@ namespace Microsoft.ML.Learners
             _nullDeviance = nullDeviance;
         }
 
-        internal LinearModelStatistics(IHostEnvironment env, long trainingExampleCount, int paramCount, Single deviance, Single nullDeviance, in VBuffer<Single> coeffStdError)
+        internal LinearModelStatistics(IHostEnvironment env, long trainingExampleCount, int paramCount, float deviance, float nullDeviance, in VBuffer<float> coeffStdError)
             : this(env, trainingExampleCount, paramCount, deviance, nullDeviance)
         {
             _env.Assert(coeffStdError.GetValues().Length == _paramCount);
@@ -120,10 +120,10 @@ namespace Microsoft.ML.Learners
             // *** Binary Format ***
             // int: count of parameters
             // long: count of training examples
-            // Single: deviance
-            // Single: null deviance
+            // float: deviance
+            // float: null deviance
             // bool: whether standard error is included
-            // (Conditional) Single[_paramCount]: values of std errors of coefficients
+            // (Conditional) float[_paramCount]: values of std errors of coefficients
             // (Conditional) int: length of std errors of coefficients
             // (Conditional) int[_paramCount]: indices of std errors of coefficients
 
@@ -143,18 +143,18 @@ namespace Microsoft.ML.Learners
                 return;
             }
 
-            Single[] stdErrorValues = ctx.Reader.ReadFloatArray(_paramCount);
+            float[] stdErrorValues = ctx.Reader.ReadFloatArray(_paramCount);
             int length = ctx.Reader.ReadInt32();
             _env.CheckDecode(length >= _paramCount);
             if (length == _paramCount)
             {
-                _coeffStdError = new VBuffer<Single>(length, stdErrorValues);
+                _coeffStdError = new VBuffer<float>(length, stdErrorValues);
                 return;
             }
 
             _env.Assert(length > _paramCount);
             int[] stdErrorIndices = ctx.Reader.ReadIntArray(_paramCount);
-            _coeffStdError = new VBuffer<Single>(length, _paramCount, stdErrorValues, stdErrorIndices);
+            _coeffStdError = new VBuffer<float>(length, _paramCount, stdErrorValues, stdErrorIndices);
         }
 
         internal static LinearModelStatistics Create(IHostEnvironment env, ModelLoadContext ctx)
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs
```diff
@@ -27,12 +27,12 @@ namespace Microsoft.ML.Learners
     public readonly struct CoefficientStatistics
     {
         public readonly string Name;
```
TomFinley: >public readonly string Name; [](start = 8, length = 28)

As discussed, if this is to be part of our public API we should probably as part of this PR make this entire structure a bit less goofy. If I were to trace back the root of what makes this structure a confusing structure, I think this choice here -- having the name of the slot instead of just the slot index, is the original sin.

This should have just been `The fact that this is a `Name` instead of just a `SlotIndex` is the cause for an enormous amount of confusion in this structure. This requires that we link it with the schema column for features, but from an API perspective this is just something that could have easily (in fact, more easily) be done by the user themselves. As it is, the structure is offering "convenience" but in a way that inevitably leads to a more complicated API.

Once you do that, restructuring the odd API below (which I'll comment more on in the test file) should become more approachable.

test/Microsoft.ML.Tests/TrainerEstimators/LbfgsTests.cs
```diff
@@ -80,14 +81,24 @@ public void TestLogisticRegressionStats_MKL()
                 s.StdComputer = new ComputeLRTrainingStdThroughHal();
             }));
 
-            var transformerChain = pipe.Fit(dataView) as TransformerChain<BinaryPredictionTransformer<ParameterMixingCalibratedPredictor>>;
+            var transformer = pipe.Fit(dataView) as TransformerChain<BinaryPredictionTransformer<ParameterMixingCalibratedPredictor>>;
 
-            var linearModel = transformerChain.LastTransformer.Model.SubPredictor as LinearBinaryModelParameters;
+            var linearModel = transformer.LastTransformer.Model.SubPredictor as LinearBinaryModelParameters;
             var stats = linearModel.Statistics;
             LinearModelStatistics.TryGetBiasStatistics(stats, 2, out float stdError, out float zScore, out float pValue);
```
TomFinley: >LinearModelStatistics.TryGetBiasStatistics(stats, 2, out float stdError, out float zScore, out float pValue); [](start = 12, length = 109)

This is the first thing that does not make much sense to me. It seems like `linearModel.Statistics` should offer similar accessors through an instance method. (Why `out` parameters? Why not just properties on the original structure?) This is just really complicated.

test/Microsoft.ML.Tests/TrainerEstimators/LbfgsTests.cs
```diff
@@ -80,14 +81,24 @@ public void TestLogisticRegressionStats_MKL()
...
 
             CompareNumbersWithTolerance(stdError, 0.250672936);
             CompareNumbersWithTolerance(zScore, 7.97852373);
+
+            var scoredData = transformer.Transform(dataView);
+
+            var coeffcients  = stats.GetCoefficientStatistics(linearModel, scoredData.Schema["Features"], 100);
```
TomFinley: >stats.GetCoefficientStatistics(linearModel, scoredData.Schema["Features"], 100); [](start = 31, length = 80)

This is the second thing that does not make sense to me. Again should just be an instance method. As discussed also, the complication that this code itself is responsible for extracting out the slot names was a mistake. Anticipating some people's reaction, if the reaction from some people would be "well it's more convenient," then the correct remediation is to just make slot names easier to access (somehow), definitely not to "solve" the problem here as is done.

test/Microsoft.ML.Tests/TrainerEstimators/LbfgsTests.cs
```diff
@@ -80,14 +81,24 @@ public void TestLogisticRegressionStats_MKL()
...
+
+            Assert.Equal(19, coeffcients.Length);
```
TomFinley: >coeffcients [](start = 29, length = 11)

Not a big deal at all since it's just a test, but while you're at it, this is a typo. `coeffcients` => `coefficients`.

src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs
TomFinley: This is not to say that it might be convenient to have an internal utility method that makes saving as text more convenient for the `LinearModelParameters` and whatnot, but confusing the needs of the text exporter (which is not part of our public API) with what we actually do in the public API was a misstep.

---
In reply to: [244374085](https://github.com/dotnet/machinelearning/pull/1979#discussion_r244374085) [](ancestors = 244374085)

test/Microsoft.ML.Tests/TrainerEstimators/LbfgsTests.cs
shmoradims: +1

As Tom pointed out, the API for getting statistics needs some cleanup.  linearModel.Statistics or some other simple 1-liner should be the only code a user needs.

---
In reply to: [244374354](https://github.com/dotnet/machinelearning/pull/1979#discussion_r244374354) [](ancestors = 244374354)

sfilipi:
@TomFinley @shmoradims I am in the process of doing the suggested refactoring, and figured that i should go a bit further: break down this class in tow parts: one that gives basic stats - used by MultiClass and LR in the case when the extended cals are not required, and another class that offers the extended stats in top of the previous. 

This will be a bit more complex of a change, since it will require loading modes backwards etc; i am not confident that if i publish the changes today/tomorrow the next round of review will be final and  also it will be a larger change to cherry pick. 

Are you guys ok with checking in this PR, since `GetCoefficientStatistics` needs to be public, and getting the rest of the changes on the next PR, but in 0.10?

sfilipi:
thanks Tom. Created #2010 with full description of the other issues with it

test/Microsoft.ML.Tests/TrainerEstimators/LbfgsTests.cs
```diff
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Linq;
```
Ivanidzo4ka: Why we need Linq here?

src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs
```diff
@@ -324,33 +326,31 @@ private List<CoefficientStatistics> GetUnorderedCoefficientStatistics(LinearBina
         /// <summary>
         /// Gets the coefficient statistics as an object.
         /// </summary>
```
Ivanidzo4ka: Need better summary than this one.

src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs
```diff
@@ -324,33 +326,31 @@ private List<CoefficientStatistics> GetUnorderedCoefficientStatistics(LinearBina
...
-        internal CoefficientStatistics[] GetCoefficientStatistics(LinearBinaryModelParameters parent, RoleMappedSchema schema, int paramCountCap)
+        public CoefficientStatistics[] GetCoefficientStatistics(LinearBinaryModelParameters parent, Schema.Column featureColumn, int paramCountCap)
```
Ivanidzo4ka: >paramCountCap [](start = 133, length = 13)

this isn't great name. Param is such a generic term, I have no idea what it's capping.

src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs
```diff
@@ -82,17 +82,17 @@ private static VersionInfo GetVersionInfo()
         // It could be null when there are too many non-zero weights so that
         // the memory is insufficient to hold the Hessian matrix necessary for the computation
         // of the variance-covariance matrix.
-        private readonly VBuffer<Single>? _coeffStdError;
+        private readonly VBuffer<float>? _coeffStdError;
 
         public long TrainingExampleCount => _trainingExampleCount;
 
-        public Single Deviance => _deviance;
+        public float Deviance => _deviance;
```
Ivanidzo4ka: >_deviance [](start = 33, length = 9)

why we need _deviance?

src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs
```diff
@@ -82,17 +82,17 @@ private static VersionInfo GetVersionInfo()
...
 
-        public Single NullDeviance => _nullDeviance;
+        public float NullDeviance => _nullDeviance;
```
Ivanidzo4ka: >_nullDeviance [](start = 37, length = 13)

I don't see any reason for _nullDeviance to exist.

Ivanidzo4ka: >NullDeviance [](start = 21, length = 12)

Need proper comment

src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs
Ivanidzo4ka: >Deviance [](start = 21, length = 8)

need proper comment.

test/Microsoft.ML.Tests/TrainerEstimators/LbfgsTests.cs
Ivanidzo4ka: >GetCoefficientStatistics [](start = 37, length = 24)

I find it strange to have 
TryGetBiasStatistics which accept LinearModelStatistics object
and GetCoefficientStatistics which are part of  LinearModelStatistics object.
a) relationship to LinearModelStatistics object
b) why one is Try, and other is just Get?

sfilipi: Yep, ugly. I am fixing all of it now, working on #2010

---
In reply to: [245155368](https://github.com/dotnet/machinelearning/pull/1979#discussion_r245155368) [](ancestors = 245155368)

PR #1988, opened 2018-12-31
Support back compat for ngram hash

### LINKED ISSUE #1982: NGramHashingTransformer cannot read old models
The serialization version was recently incremented, but we should be able to load old models using SignatureLoadDataTransform.

Ivanidzo4ka:
fixes #1982


test/Microsoft.ML.Tests/Transformers/TextFeaturizerTests.cs
```diff
@@ -269,6 +270,22 @@ public void NgramWorkout()
...
+            string dropModelPath = GetDataPath("backcompat/ngram.zip");
+            string sentimentDataPath = GetDataPath("wikipedia-detox-250-line-data.tsv");
+            var data = TextLoader.CreateReader(ML, ctx => (
+                    Sentiment: ctx.LoadBool(0),
+                    SentimentText: ctx.LoadText(1)), hasHeader: true)
+                .Read(sentimentDataPath);
+            using (FileStream fs = File.OpenRead(dropModelPath))
+            {
+                var result = ModelFileUtils.LoadTransforms(Env, data.AsDynamic, fs);
+                var foundColumnFeature = result.Schema.TryGetColumnIndex("Features", out int featureIdx);
```
TomFinley: >foundColumnFeature [](start = 20, length = 18)

Were you intending to assert this somewhere? Also, could you please not use the `ISchema` derived functions, but the newer `Schema` ones? We're going to be getting rid of the `ISchema` ones sooner or later. In this case, preferred would be `GetColumnOrNull`.

[https://github.com/dotnet/machinelearning/blob/7e21afa4612ddca53bbf9d67b84df2b58d4636c2/src/Microsoft.ML.Core/Data/Schema.cs#L64](https://github.com/dotnet/machinelearning/blob/7e21afa4612ddca53bbf9d67b84df2b58d4636c2/src/Microsoft.ML.Core/Data/Schema.cs#L64)

test/Microsoft.ML.Tests/Transformers/TextFeaturizerTests.cs
```diff
@@ -269,6 +270,23 @@ public void NgramWorkout()
             Done();
         }
 
+        [Fact]
+        void TestNgramCompatColumns()
```
justinormont: Thanks for the new test! Should we be testing the specific output values too? (or is this somewhere I'm not seeing it?)

justinormont:
Not to nickle-and-dime, but is 273KB needed to add to the repo? Existing repo zips to 18.3MB. Seems a simple ngram-hash model should be tiny. 

Looks like space in your model is from an ngram dictionary & normalizer. Perhaps set `bits=10` & `max=100` to minify.

Command: _(now 18KB)_
`maml.exe Train loader=TextLoader{col=Sentiment:R4:0 col=SentimentText:TX:1 header=+} data=D:\src\fork-machinelearning\test\data\wikipedia-detox-250-line-data.tsv out=d:\tlc\1.model.zip xf=CopyColumns{col=Label:Sentiment} xf=WordBagTransform{col=Column:SentimentText tok=WordTokenizeTransform weighting=TfIdf max=100} xf=CategoricalTransform{col=Column kind=Key} xf=NgramHashTransform{col=Features1:Column all=- rehash=+ bits=10} xf=NgramTransform{col=Features2:Column max=100} xf=Concat{col=Features:Features1,Features2}`

src/Microsoft.ML.Transforms/Text/NgramHashingTransformer.cs
```diff
@@ -165,11 +165,13 @@ private static VersionInfo GetVersionInfo()
                 // verWrittenCur: 0x00010002, // Invert hash key values, hash fix
                 verWrittenCur: 0x00010003, // Get rid of writing float size in model context and change saving format
                 verReadableCur: 0x00010003,
-                verWeCanReadBack: 0x00010003,
+                verWeCanReadBack: 0x00010002,
                 loaderSignature: LoaderSignature,
                 loaderAssemblyName: typeof(NgramHashingTransformer).Assembly.FullName);
         }
 
+        private const int VersionTransformer = 0x00010003;
```
sfilipi: >VersionTransformer [](start = 26, length = 18)

it would be more intuitite, i think, if called: CurrentVersion.

Issue #1984, opened 2018-12-29
LinearSupportVectorMachines trainer not working


daholste:
### System information

- **OS version/distro**: Windows 10
- **.NET Version (eg., dotnet --info)**: .NET Core 2.1

### Issue

**Setup code**
```
const string trainDataPath = @"C:\data\sample_train2.csv";
const string validationDataPath = @"C:\data\sample_valid2.csv";
const string testDataPath = @"C:\data\sample_test2.csv";

var mlContext = new MLContext();

// auto-infer text loader args
var textLoaderArgs = RecipeInference.MyAutoMlInferTextLoaderArguments(mlContext, trainDataPath, "Label");

// load data
var textLoader = new TextLoader(mlContext,
	new TextLoader.Arguments()
	{
		Separator = ",",
		HasHeader = true,
		Column = new[]
		{
			new TextLoader.Column("Age", DataKind.R4, 0),
			new TextLoader.Column("Workclass", DataKind.TX, 1),
			new TextLoader.Column("Fnlwgt", DataKind.R4, 2),
			new TextLoader.Column("Education", DataKind.TX, 3),
			new TextLoader.Column("EducationNum", DataKind.R4, 4),
			new TextLoader.Column("MaritalStatus", DataKind.TX, 5),
			new TextLoader.Column("Occupation", DataKind.TX, 6),
			new TextLoader.Column("Relationship", DataKind.TX, 7),
			new TextLoader.Column("Race", DataKind.TX, 8),
			new TextLoader.Column("Sex", DataKind.TX, 9),
			new TextLoader.Column("CapitalGain", DataKind.R4, 10),
			new TextLoader.Column("CapitalLoss", DataKind.R4, 11),
			new TextLoader.Column("HoursPerWeek", DataKind.R4, 12),
			new TextLoader.Column("NativeCountry", DataKind.TX, 13),
			new TextLoader.Column("Label", DataKind.Bool, 14),
		}
	});

var trainData = textLoader.Read(trainDataPath);
var validationData = textLoader.Read(validationDataPath);
var testData = textLoader.Read(testDataPath);

// preprocess
var preprocessorEstimator = mlContext.Transforms.Categorical.OneHotEncoding("Workclass", "Workclass")
	.Append(mlContext.Transforms.Categorical.OneHotEncoding("Education", "Education"))
	.Append(mlContext.Transforms.Categorical.OneHotEncoding("MaritalStatus", "MaritalStatus"))
[...]

PR #1998, opened 2019-01-02
Add test for linear svm and polish it's code

Ivanidzo4ka:
Related to https://github.com/dotnet/machinelearning/issues/1984.


src/Microsoft.ML.StandardLearners/Standard/Online/LinearSvm.cs
```diff
@@ -278,11 +276,6 @@ private protected override TrainStateBase MakeState(IChannel ch, int numFeatures
             return new TrainState(ch, numFeatures, predictor, this);
```
sfilipi: nit and only if you want to: change to =>

PR #2001, opened 2019-01-02
Make CompositeSchema not an ISchema

wschin:
Continue removing ISchema one-by-one. It's a part of #1501.

### DIFF
--- a/src/Microsoft.ML.Data/DataView/CompositeSchema.cs
+++ b/src/Microsoft.ML.Data/DataView/CompositeSchema.cs
@@ -13,16 +13,16 @@ namespace Microsoft.ML.Data
     /// A convenience class for concatenating several schemas together.
     /// This would be necessary when combining IDataViews through any type of combining operation, for example, zip.
     /// </summary>
-    internal sealed class CompositeSchema : ISchema
+    internal sealed class ZipBinding
     {
         private readonly Schema[] _sources;
 
-        public Schema AsSchema { get; }
+        public Schema OutputSchema { get; }
 
         // Zero followed by cumulative column counts. Zero being used for the empty case.
         private readonly int[] _cumulativeColCounts;
 
-        public CompositeSchema(Schema[] sources)
+        public ZipBinding(Schema[] sources)
         {
             Contracts.AssertNonEmpty(sources);
             _sources = sources;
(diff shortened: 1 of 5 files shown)


src/Microsoft.ML.Data/DataView/ZipDataView.cs
```diff
@@ -106,7 +106,7 @@ public RowCursor[] GetRowCursorSet(Func<int, bool> predicate, int n, Random rand
         private sealed class Cursor : RootCursorBase
         {
             private readonly RowCursor[] _cursors;
-            private readonly CompositeSchema _compositeSchema;
+            private readonly ZipBinding _compositeSchema;
```
Ivanidzo4ka: >_compositeSchema [](start = 40, length = 16)

can you change name of variable?

src/Microsoft.ML.Data/DataView/CompositeSchema.cs
```diff
@@ -13,16 +13,16 @@ namespace Microsoft.ML.Data
     /// A convenience class for concatenating several schemas together.
     /// This would be necessary when combining IDataViews through any type of combining operation, for example, zip.
     /// </summary>
-    internal sealed class CompositeSchema : ISchema
+    internal sealed class ZipBinding
```
Ivanidzo4ka: Shall we rename file as well?

src/Microsoft.ML.Data/DataView/CompositeSchema.cs
```diff
@@ -13,16 +13,16 @@ namespace Microsoft.ML.Data
     /// A convenience class for concatenating several schemas together.
     /// This would be necessary when combining IDataViews through any type of combining operation, for example, zip.
     /// </summary>
```
artidoro: Maybe we need to change this comment, as it seems to be suggesting that zip is not the only use?

src/Microsoft.ML.Data/DataView/CompositeSchema.cs
wschin: Done.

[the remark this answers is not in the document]
wschin: It's not a Schema. It takes several input schemas and produce an output schema.

src/Microsoft.ML.Data/DataView/ZipDataView.cs
wschin: Sure.

Issue #1604, opened 2018-11-12
Clean up our auto-caching


Zruty0:
Currently, some of our trainers cache the data prior to training, without a possibility to disable that. 

I believe the good incremental step would be to disable all auto-caching, and rely on user to call `AppendCacheCheckpoint` prior to multi-pass training. 

This is not really ideal, since the default setup for multi-pass trainers will train slower. I still think it is better to have a consistent story about our 'smarts' (that is, we have no auto-normalization, no auto-caching and no auto-calibration), and use extensive documentation (and tooling, in the future) to cover these pitfalls.

cc @GalOshri @TomFinley @eerhardt

justinormont:
I would prefer we make a good model by default for the user w/ auto-normalization & auto-calibration. 

The auto-caching is simply a matter of speed, not final model quality, but is quite related to the overall user happiness.

Zruty0:
I think this is not the first time we have this argument.
Again, the reasons we don't want to make these auto-smarts part of the core API is because sometimes they are making mistakes, and sometimes costly mistakes:

- For auto-caching, we may blow up the machine's memory by caching the training data, where a non-cached training would've succeeded (maybe slower). 
- Auto-caching assumes that the original training data is slow to access, but if it's a memory-backed dataset (or another cache), this is not true. So auto-caching may make training slower, and ALSO consume more memory.
- Auto-calibration happens on the training set. We assume that most of the time it's OK, given that we only learn 2 parameters, but we already saw cases where model quality degrades because of that.
- Auto-normalization may normalize data that is otherwise pretty regular, potentially making model larger, and training slower.
- Auto-normalization has a potential for user confusion: they thought that they merely trained a linear model, but in reality they train a pair of models.

Because of the above, we don't want any of these smarts to be part of the core ML.NET API. We need to expose the API to normalize, cache and calibrate at user's request. Our existing smarts can be converted into tooling (VS code analyzer warnings etc.).

sfilipi:
Checking the impl mechanism: so we'll have configs for norm/calibration/cache (or we already have them through TrainContext/TrainInfo) and default normalization, calibration and caching will be no?

Zruty0:
No, we just remove all auto-caching, calibration and normalization, period.

The users will be responsible to normalize the data if needed (via `mlContext.Normalize`), cache in memory if desired (via `mlContext.Data.Cache` or `pipeline.AppendCacheCheckpoint`), and calibrate if desired (after #1622 is done).

[...]

PR #2008, opened 2019-01-03
Update tests (using Iris dataset) to use new API

abgoswam:
- Update a couple of tests (using Iris dataset) to use the new API.

### DIFF
--- a/test/Microsoft.ML.Tests/Microsoft.ML.Tests.csproj
+++ b/test/Microsoft.ML.Tests/Microsoft.ML.Tests.csproj
@@ -22,7 +22,6 @@
     <ProjectReference Include="..\..\src\Microsoft.ML.StaticPipe\Microsoft.ML.StaticPipe.csproj" />
     <ProjectReference Include="..\..\src\Microsoft.ML.TensorFlow.StaticPipe\Microsoft.ML.TensorFlow.StaticPipe.csproj" />
     <ProjectReference Include="..\..\src\Microsoft.ML.TensorFlow\Microsoft.ML.TensorFlow.csproj" />
-    <ProjectReference Include="..\..\src\Microsoft.ML.Legacy\Microsoft.ML.Legacy.csproj" />
     <ProjectReference Include="..\..\src\Microsoft.ML.TimeSeries\Microsoft.ML.TimeSeries.csproj" />
     <ProjectReference Include="..\Microsoft.ML.Predictor.Tests\Microsoft.ML.Predictor.Tests.csproj" />
     <ProjectReference Include="..\Microsoft.ML.TestFramework\Microsoft.ML.TestFramework.csproj" />
--- a/test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
+++ b/test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
@@ -3,33 +3,45 @@
 // See the LICENSE file in the project root for more information.
 
 using Microsoft.ML.Data;
-using Microsoft.ML.Legacy.Models;
-using Microsoft.ML.Legacy.Trainers;
-using Microsoft.ML.Legacy.Transforms;
+using Microsoft.ML.RunTests;
 using Xunit;
-using TextLoader = Microsoft.ML.Legacy.Data.TextLoader;
 
 namespace Microsoft.ML.Scenarios
 {
-#pragma warning disable 612, 618
     public partial class ScenariosTests
     {
         [Fact]
-        public void TrainAndPredictIrisModelTest()
+        public void New_TrainAndPredictIrisModelTest()
         {
-            string dataPath = GetDataPath("iris.txt");
+            var mlContext = new MLContext(seed: 1, conc: 1);
 
-            var pipeline = new Legacy.LearningPipeline(seed: 1, conc: 1);
+            var reader = mlContext.Data.CreateTextReader(columns: new[]
+                {
+                    new TextLoader.Column("Label", DataKind.R4, 0),
+                    new TextLoader.Column("SepalLength", DataKind.R4, 1),
+                    new TextLoader.Column("SepalWidth", DataKind.R4, 2),
+                    new TextLoader.Column("PetalLength", DataKind.R4, 3),
+                    new TextLoader.Column("PetalWidth", DataKind.R4, 4)
+                }
+            );
 
-            pipeline.Add(new TextLoader(dataPath).CreateFrom<IrisData>(useHeader: false));
-            pipeline.Add(new ColumnConcatenator(outputColumn: "Features",
-                "SepalLength", "SepalWidth", "PetalLength", "PetalWidth"));
+            var pipe = mlContext.Transforms.Concatenate("Features", "SepalLength", "SepalWidth", "PetalLength", "PetalWidth")
+                            .Append(mlContext.Transforms.Normalize("Features"))
+                            .AppendCacheCheckpoint(mlContext)
+                            .Append(mlContext.MulticlassClassification.Trainers.StochasticDualCoordinateAscent("Label", "Features", advancedSettings: s => s.NumThreads = 1));
 
-            pipeline.Add(new StochasticDualCoordinateAscentClassifier());
+            // Read training and test data sets
+            string dataPath = GetDataPath(TestDatasets.iris.trainFilename);
+            string testDataPath = dataPath;
+            var trainData = reader.Read(dataPath);
+            var testData = reader.Read(testDataPath);
 
-            Legacy.PredictionModel<IrisData, IrisPrediction> model = pipeline.Train<IrisData, IrisPrediction>();
+            // Train the pipeline
+            var trainedModel = pipe.Fit(trainData);
 
-            IrisPrediction prediction = model.Predict(new IrisData()
+            // Make predictions
+            var predictFunction = trainedModel.CreatePredictionEngine<IrisData, IrisPrediction>(mlContext);
+            IrisPrediction prediction = predictFunction.Predict(new IrisData()
             {
                 SepalLength = 5.1f,
                 SepalWidth = 3.3f,
@@ -41,7 +53,7 @@ namespace Microsoft.ML.Scenarios
             Assert.Equal(0, prediction.PredictedLabels[1], 2);
             Assert.Equal(0, prediction.PredictedLabels[2], 2);
 
-            prediction = model.Predict(new IrisData()
+            prediction = predictFunction.Predict(new IrisData()
             {
                 SepalLength = 6.4f,
                 SepalWidth = 3.1f,
@@ -53,7 +65,7 @@ namespace Microsoft.ML.Scenarios
             Assert.Equal(0, prediction.PredictedLabels[1], 2);
             Assert.Equal(1, prediction.PredictedLabels[2], 2);
 
-            prediction = model.Predict(new IrisData()
+            prediction = predictFunction.Predict(new IrisData()
             {
                 SepalLength = 4.4f,
                 SepalWidth = 3.1f,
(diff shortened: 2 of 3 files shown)


test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
```diff
@@ -3,33 +3,45 @@
...
-using TextLoader = Microsoft.ML.Legacy.Data.TextLoader;
 
 namespace Microsoft.ML.Scenarios
 {
-#pragma warning disable 612, 618
     public partial class ScenariosTests
     {
         [Fact]
-        public void TrainAndPredictIrisModelTest()
+        public void New_TrainAndPredictIrisModelTest()
```
codemzs: >New_ [](start = 20, length = 4)

I believe "New" prefix was used because we had two APIs but now the old API is gone and there is just one API, do you feel we still need to prefix test with "New"?

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
```diff
@@ -3,33 +3,45 @@
...
+                    new TextLoader.Column("SepalLength", DataKind.R4, 1),
+                    new TextLoader.Column("SepalWidth", DataKind.R4, 2),
+                    new TextLoader.Column("PetalLength", DataKind.R4, 3),
+                    new TextLoader.Column("PetalWidth", DataKind.R4, 4)
+                }
+            );
+
+            var pipe = mlContext.Transforms.Concatenate("Features", "SepalLength", "SepalWidth", "PetalLength", "PetalWidth")
+                            .Append(mlContext.Transforms.Normalize("Features"))
+                            .AppendCacheCheckpoint(mlContext)
```
codemzs: Why are we adding normalizer transform? just curious.

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
```diff
@@ -3,33 +3,45 @@
...
+                            .Append(mlContext.MulticlassClassification.Trainers.StochasticDualCoordinateAscent("Label", "Features", advancedSettings: s => s.NumThreads = 1));
+
+            // Read training and test data sets
+            string dataPath = GetDataPath(TestDatasets.iris.trainFilename);
+            string testDataPath = dataPath;
```
codemzs: >string testDataPath = dataPath; [](start = 12, length = 31)

Why is this needed? We are using prediction engine to test.

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
```diff
@@ -3,33 +3,45 @@
...
+            var trainData = reader.Read(dataPath);
+            var testData = reader.Read(testDataPath);
```
codemzs: Why is this needed?

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
abgoswam: I added it during development when I had both tests in this file.  Forgot yo drop the 'New_' prefix after deleting the old test.  Will fix.

---
In reply to: [245111276](https://github.com/dotnet/machinelearning/pull/2008#discussion_r245111276) [](ancestors = 245111276)

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
```diff
@@ -65,53 +77,20 @@ public void TrainAndPredictIrisModelTest()
...
-            Assert.Equal(48, matrix["1", "1"]);
-            Assert.Equal(2, matrix[1, 2]);
-            Assert.Equal(2, matrix["1", "2"]);
-
-            Assert.Equal(0, matrix[2, 0]);
-            Assert.Equal(0, matrix["2", "0"]);
-            Assert.Equal(1, matrix[2, 1]);
-            Assert.Equal(1, matrix["2", "1"]);
-            Assert.Equal(49, matrix[2, 2]);
-            Assert.Equal(49, matrix["2", "2"]);
```
codemzs: Do we not have confusion matrix in metrics?

test/Microsoft.ML.Tests/Microsoft.ML.Tests.csproj
```diff
@@ -22,7 +22,6 @@
     <ProjectReference Include="..\..\src\Microsoft.ML.StaticPipe\Microsoft.ML.StaticPipe.csproj" />
     <ProjectReference Include="..\..\src\Microsoft.ML.TensorFlow.StaticPipe\Microsoft.ML.TensorFlow.StaticPipe.csproj" />
     <ProjectReference Include="..\..\src\Microsoft.ML.TensorFlow\Microsoft.ML.TensorFlow.csproj" />
-    <ProjectReference Include="..\..\src\Microsoft.ML.Legacy\Microsoft.ML.Legacy.csproj" />
```
codemzs: ><ProjectReference Include="..\..\src\Microsoft.ML.Legacy\Microsoft.ML.Legacy.csproj" /> [](start = 4, length = 87)

Nice!!

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
abgoswam: I think in the new API  the paradigm is to be explicit  about whether are using  Normalizers  in the pipeline .  

(In the old Legacy API normalizers used to be auto-inserted based on the algorithm)

Hence decided to be explicit about normalizers.  Please see #1604  which talks about  removing 'smarts' (i.e.  no auto-normalization, no auto-caching and no auto-calibration) in the new API


---
In reply to: [245111952](https://github.com/dotnet/machinelearning/pull/2008#discussion_r245111952) [](ancestors = 245111952)

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
abgoswam: I do not see it inside `metrics`.  Who would be the best person to confirm  whether we support `ConfusionMatrix` in the new API ?

If we do, is there an example I can follow ? I do not see it in the CookBook either

---
In reply to: [245114114](https://github.com/dotnet/machinelearning/pull/2008#discussion_r245114114) [](ancestors = 245114114)

Ivanidzo4ka:
@abgoswam Can we just delete this files instead of modifying them? We have test which do evaluation, we have tests which do predictions (with new API), what's the point of having this ones?

codemzs:
@Ivanidzo4ka We did that and it results in code coverage drop. Check out my PR that removes these files. It’s possible removing some tests might be ok but which ones will require tests by tests code coverage analysis. It might just be easier to convert all the test to new API. Even though these tests are doing training, evaluation, etc but they may be using different transforms, trainers, settings, etc.

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
```diff
@@ -65,53 +77,20 @@ public void TrainAndPredictIrisModelTest()
...
-            evaluator.OutputTopKAcc = 3;
-            ClassificationMetrics metrics = evaluator.Evaluate(model, testData);
+            // Evaluate the trained pipeline
+            var predicted = trainedModel.Transform(testData);
+            var metrics = mlContext.MulticlassClassification.Evaluate(predicted, topK:3);
 
             Assert.Equal(.98, metrics.AccuracyMacro);
             Assert.Equal(.98, metrics.AccuracyMicro, 2);
-            Assert.Equal(.06, metrics.LogLoss, 2);
+            Assert.InRange(metrics.LogLoss, .05, .06);
```
wschin: Could we use Assert.Equal(…, …, 2)?

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationWithStringLabelTests.cs
```diff
@@ -3,126 +3,100 @@
...
-            evaluator.OutputTopKAcc = 3;
-            ClassificationMetrics metrics = evaluator.Evaluate(model, testData);
+            // Evaluate the trained pipeline
+            var predicted = trainedModel.Transform(testData);
+            var metrics = mlContext.MulticlassClassification.Evaluate(predicted, topK: 3);
 
             Assert.Equal(.98, metrics.AccuracyMacro);
             Assert.Equal(.98, metrics.AccuracyMicro, 2);
-            Assert.Equal(.06, metrics.LogLoss, 2);
+            Assert.InRange(metrics.LogLoss, .05, .06);
```
wschin: Assert.Equal(metrics.Logloss, 0.055, 3) maybe?

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationWithStringLabelTests.cs
```diff
@@ -139,9 +113,17 @@ public class IrisDataWithStringLabel
             [LoadColumn(3)]
             public float PetalWidth;
 
-            [LoadColumn(4), ColumnName("Label")]
+            [LoadColumn(4)]
             public string IrisPlantType;
         }
+
+        public class IrisPredictionWithStringLabel
```
wschin: Are those data structures used outside this file?

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
codemzs: I will also prefer Equal

Ivanidzo4ka:
I've run it with tests and without tests, and only dll for which code coverage went down is Legacy, which is expected. Everything else either state same, or went up by 0.01% :)


---
In reply to: [451270185](https://github.com/dotnet/machinelearning/pull/2008#issuecomment-451270185) [](ancestors = 451270185)

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
abgoswam: Created issue #2009  to bring back support for `ConfusionMatrix` in the new API 

---
In reply to: [245117612](https://github.com/dotnet/machinelearning/pull/2008#discussion_r245117612) [](ancestors = 245117612,245114114)

codemzs:
I have explained to you in person the issue with running code coverage per test in the context of parallelizing test conversion effort. Say you have two tests, each assigned to two devs, these two tests are exercising same code paths, if both devs in parallel run code coverage for the test assigned to them then both will end up removing both tests because of this race condition and then we lose code coverage. I do not see this a safe and quick way to prevent regression in code coverage due to the removal Legacy API binary. The fastest and safest way is to convert all the test to the new API. Correct me if I'm wrong.

---
In reply to: [451278660](https://github.com/dotnet/machinelearning/pull/2008#issuecomment-451278660) [](ancestors = 451278660,451270185)

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationWithStringLabelTests.cs
abgoswam: fixed

---
In reply to: [245133430](https://github.com/dotnet/machinelearning/pull/2008#discussion_r245133430) [](ancestors = 245133430)

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
abgoswam: fixed

---
In reply to: [245113327](https://github.com/dotnet/machinelearning/pull/2008#discussion_r245113327) [](ancestors = 245113327,245111276)

test/Microsoft.ML.Tests/Scenarios/IrisPlantClassificationTests.cs
abgoswam: fixed

---
In reply to: [245133023](https://github.com/dotnet/machinelearning/pull/2008#discussion_r245133023) [](ancestors = 245133023)

PR #2011, opened 2019-01-03
Removing Iteration Setup from Benchmarks

Anipik:
Specifying IterationSetup benchmark causes the benchmark to run just only once in an iteration. This is a problem if the benchmarks takes less than 100ms to execute.

As it is being executed just once per iteration and number of iterations is 20, no jitting is done on netcore3.0 (requires 30 calls).

The possible solutions were to increase the length + increase the warmCount to 30 (this could be a problem for other benchmarks)

Another solution is just removing the iterationSetup. In CacheWithSeeker we are not doing anything that is specific to an iteration and we can directly move the code to globalSetup.

for CacheWithCursor I moved the implementation to the actual benchmark as the iteration setup itseld was taking a small time. But it will good if we could move the iteration setup to globalSetup. @TomFinley any suggetions will be helpful.

Also Earlier this was reported as regression for netcore3.0 but after testing the benchmark with both the solutions, I found there is no regression in this benchmark.

### DIFF
--- a/test/Microsoft.ML.Benchmarks/CacheDataViewBench.cs
+++ b/test/Microsoft.ML.Benchmarks/CacheDataViewBench.cs
@@ -14,9 +14,10 @@ namespace Microsoft.ML.Benchmarks
 
         // Global.
         private IDataView _cacheDataView;
-        // Per iteration.
         private RowCursor _cursor;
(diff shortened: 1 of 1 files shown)


Anipik:
Thanks for the info @TomFinley

PR #2013, opened 2019-01-03
New ONNX converter interface and some tests

wschin:
This PR create a ONNX converter targeting ITransformer. A full example is added as a test; we train a pipeline (normalize + SDCA) using new APIs, convert the trained model, and call ONNXRuntime to evaluate it. In addition, we recreate existing ONNX conversion tests using new APIs to ensure at least identical test coverage after removing ONNX tests using legacy APIs.

Working item summary:
1. Create new ONNX converter interface which is able to handle ITransformer.
2. Remove OnnxTest.cs
3. Recreate ONNX conversions in OnnxTest.cs using new APIs.
4. Add extra end-to-end tests. We train ML.NET model, convert it to ONNX format, and call ONNX runtime to execute it. The results produced by ONNX runtime and ML.NET's original model are checked.


src/Microsoft.ML.Onnx/TransformerChainOnnxConverter.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System.Collections.Generic;
+using Microsoft.ML.Core.Data;
+using Microsoft.ML.Data;
+using Microsoft.ML.UniversalModelFormat.Onnx;
+
+namespace Microsoft.ML.Model.Onnx
+{
+    public class TransformerChainOnnxConverter
```
TomFinley: >public class [](start = 4, length = 12)

static class

src/Microsoft.ML.Onnx/TransformerChainOnnxConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        public static ModelProto Convert<T>(TransformerChain<T> chain, IDataView inputData, HashSet<string> inputColumnNamesToDrop=null, HashSet<string> outputColumnNamesToDrop=null) where T : class, ITransformer
```
TomFinley: >TransformerChain<T> chain [](start = 44, length = 25)

Is there a reason to do this? If you wanted specialized behavior for the transformer chain, would it be sufficient to just check if it is a transformer chain in the other method, and not make this public?

The presence of a specialized overload for transformer chain specifically makes me suspect that you're having different behavior for that. But this makes me wonder, since a `TransformerChain<T>` is also an `ITransformer`, does this mean that if I were to just feed it in as such below, that it would not work, or have a flaw of some kind? In which case, the method below we'd say is buggy. (Since it purports to work over any `ITransformer`. But if it *does* work, then why have a separate method anyway? So I'd put that logic (if indeed any must exist) on the second method, and *not* expose whatever implementation specific type polymorphism to the user. Since they shouldn't have to care about that.

codemzs:
Thanks, @wschin. The one main issue that I see is you are testing for just Windows and not other platforms.  You are also not checking if the produced ONNX JSON file is matching the baseline JSON file, by not doing this we will lose existing code coverage that we have with current tests. Is there a way we can use existing SaveONNXCommand by somehow creating a wrapper around it? The reason I ask this is because it can convert an IDataView and Model zip file to ONNX but your current implementation just converts the in memory pipeline and does it even allow to convert to JSON representation. 

@TomFinley, What are your thoughts?

src/Microsoft.ML.Onnx/SaveOnnxCommand.cs
```diff
@@ -134,7 +137,51 @@ private void GetPipe(OnnxContextImpl ctx, IChannel ch, IDataView end, out IDataV
                 transform = (source = transform.Source) as IDataTransform;
             }
 
-            Host.AssertValue(source);
+            Contracts.AssertValue(source);
+        }
+
+        [BestFriend]
+        internal static ModelProto ConvertTransformListToOnnxModel(OnnxContextImpl ctx, IDataView inputData, IDataView outputData,
```
TomFinley: Why `[BestFriend]`? I don't see that you're exposing this elsewhere.

src/Microsoft.ML.Onnx/TransformerChainOnnxConverter.cs
TomFinley: >ModelProto [](start = 22, length = 10)

Our typical approach for new capabilities like this is that we'd, as much as possible, make operations like this. How do we current expose saving an `ITransformer`? Through `ModelOperationsCatalog`. This should have likewise been an extension method on `ModelOperationsCatalog`. (And, like all other things like this, in namespace `Microsoft.ML`.) That was this functionality gets discoverable through `MLContext`, which is our strategy for things like this so far.

TomFinley: >HashSet<string> inputColumnNamesToDrop=null, HashSet<string> outputColumnNamesToDrop [](start = 92, length = 84)

While I know why we have these from the *command's* perspective (since that kind of has to be a one-stop-shop) to do it from the *API*'s perspective is incorrect, since it complicates the call enormously. If someone wants to drop a column on `inputData`, they should just do that as they do literally everywhere else in the API... by dropping that column. Similarly, it seems that if someone does not want to include transform outputs produced, they ought to add something to the transform chain (again, dropping columns) to limit the output to only those columns they care about. At least, so I think.

wschin:
Those tests have no json file which can be reused because (1) no such a test before or (2) json file gets changed because random seed mechanisms are different in new and old contexts. If you really want json files, we will have to (1) accept those tests first, (2) produce json files from here, and (3) make other tests for comparing those text outputs. Honestly, comparing text outputs do not make a lot sense for me because the correctness of a conversion can only determined by its output values. 

This is a wrapper around SaveOnnxCommand.

I don't understand the need of JSON. We create protobuf object so as long as protobuf package is available, user can convert the output model to text naturally.


---
In reply to: [451359788](https://github.com/dotnet/machinelearning/pull/2013#issuecomment-451359788) [](ancestors = 451359788)

src/Microsoft.ML.Onnx/SaveOnnxCommand.cs
wschin: I will remove it. Thanks.

---
In reply to: [245215299](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245215299) [](ancestors = 245215299)

codemzs:
I will keep this simple and outline the goals for this work:
1) We do NOT want regress code coverage.
2) We want our tests to run on Windows, MacOS and Linux just as previous ONNX conversion test ran. 
3) You are not exercising code that exports to json format for a model that is done at    [https://github.com/dotnet/machinelearning/blob/9a6711b7ba926164b3255be3a03615d90723aaeb/test/Microsoft.ML.Tests/OnnxTests.cs#L150](https://github.com/dotnet/machinelearning/blob/9a6711b7ba926164b3255be3a03615d90723aaeb/test/Microsoft.ML.Tests/OnnxTests.cs#L150)

I can tell with a lot of confidence this PR regresses the current code coverage and I will be happy to show you tomorrow with code coverage tool. 


---
In reply to: [451365439](https://github.com/dotnet/machinelearning/pull/2013#issuecomment-451365439) [](ancestors = 451365439,451359788)

wschin:
My argument doesn't conflict with yours. It's ok to add json but I will have to check in those tests before doing any json. Those json files need to be tested by those windows-only tests.

---
In reply to: [451365439](https://github.com/dotnet/machinelearning/pull/2013#issuecomment-451365439) [](ancestors = 451365439,451359788)

src/Microsoft.ML.Onnx/TransformerChainOnnxConverter.cs
wschin: I will delete the version for `TransformerChain<T>`.

---
In reply to: [245214766](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245214766) [](ancestors = 245214766)

wschin: Sounds good. I will make those parameter not accessible to users. On the other hand, I am not sure if I can reproduce some old tests without the two argument, so having an internal version with the two arguments might be a backup plan.

---
In reply to: [245216200](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245216200) [](ancestors = 245216200)

TomFinley:
> My argument doesn't conflict with yours. It's ok to add json but I will have to check in those tests before doing any json. Those models need to be tested by those windows-only tests before being saved as jsons. That also means, we will have at least two batchs of PRs. Notice that old tests are not reproducible.

Hi @wschin and @codemzs, there's something I don't fully appreciate here, maybe you can help me. As I understand it, we are producing a `ModelProto` object as part of the test. Is it not so? And, we would expect the result of this to be deterministic, I hope? And the simplest way we can detect whether something changes is to just take this `ModelProto`, export it as a JSON or YAML or some other easily interpreted textual fformat (which I believe protobuf provides facilities for), and baseline that, so that if something changes, we can see at a glance without having to do a lot of work, *what* has changed. (Either due to a failure of our expectations, a change in the ONNX library and protobuf defs, or some other issue.) So, with that plan, is there something stopping us from doing that right now?

I often think we use baseline testing where it really isn't appropriate, but in this case, where we have this other complex piece of machinery we're trying to interop with, and we have this complex object we're producing many parts of which can change in ways invisible to us, a baseline test makes me feel less nervous.

We should probably not have export as JSON in the API itself though, since we are writing an API and the most natural way to export as JSON or whatever is to use existing facilities for handling this through protobuf.

src/Microsoft.ML.Onnx/TransformerChainOnnxConverter.cs
wschin: Done.

---
In reply to: [245213997](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245213997) [](ancestors = 245213997)

src/Microsoft.ML.Onnx/TransformerChainOnnxConverter.cs
wschin: A fix will be pushed soon.

---
In reply to: [245215676](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245215676) [](ancestors = 245215676)

wschin:
Text comparison added.

---
In reply to: [451377212](https://github.com/dotnet/machinelearning/pull/2013#issuecomment-451377212) [](ancestors = 451377212,451365439,451359788)

src/Microsoft.ML.Onnx/ProtabilityCatalog.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Collections.Generic;
+using Microsoft.ML.Core.Data;
+using Microsoft.ML.Data;
+using Microsoft.ML.Model.Onnx;
+using Microsoft.ML.UniversalModelFormat.Onnx;
+
+namespace Microsoft.ML
+{
+    public static class ProtabilityCatalog
```
TomFinley: >ProtabilityCatalog [](start = 24, length = 18)

Should probably be named `OnnxExportExtensions` or somesuch like this.

src/Microsoft.ML.Onnx/ProtabilityCatalog.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>
+        /// Convert the specified <see cref="ITransformer"/> to ONNX format. Note that ONNX uses Google's Protobuf so the returned value is a Protobuf object.
+        /// </summary>
+        /// <param name="catalog">A field in <see cref="MLContext"/> which this function associated with.</param>
+        /// <param name="transform">The <see cref="ITransformer"/> that will be converted into ONNX format.</param>
+        /// <param name="inputData">The input of the specified transform.</param>
+        /// <returns></returns>
+        public static ModelProto ConvertToOnnx(this ModelOperationsCatalog.PortabilityTransforms catalog, ITransformer transform, IDataView inputData)
+        {
+            var env = new MLContext(seed: 1);
```
TomFinley: >var env = new MLContext(seed: 1); [](start = 12, length = 33)

Sorry, why are we creating the environment, instead of getting it from the catalog?

src/Microsoft.ML.Onnx/ProtabilityCatalog.cs
```diff
@@ -0,0 +1,32 @@
...
+            var ctx = new OnnxContextImpl(env, "model", "ML.NET", "0", 0, "com.microsoft", OnnxVersion.Stable);
+            var outputData = transform.Transform(inputData);
+            IDataView root = null;
+            IDataView sink = null;
+            LinkedList<ITransformCanSaveOnnx> transforms = null;
+            using (var ch = (env as IChannelProvider).Start("ONNX conversion"))
+                SaveOnnxCommand.GetPipe(ctx, ch, outputData, out root, out sink, out transforms);
+
+            return SaveOnnxCommand.ConvertTransformListToOnnxModel(ctx, root, sink, transforms, null, null);
```
TomFinley: >return SaveOnnxCommand.ConvertTransformListToOnnxModel(ctx, root, sink, transforms, null, null); [](start = 12, length = 96)

Would it be inappropriate to make this part of the `using` block? It just seems a little odd to consider that `ConvertTransformListToOnnxModel` is *separate* from a channel quite explicitly identified as being "ONNX Conversion", since this call, right here, is precisely where that conversion is happening? So why is it not in the using block of that channel? Just seems a bit off.

Once you do that, you can also get rid of your `root` and `sink`, and change `out sink, out transforms` to `out var sink, out var transforms` to clean up your code a bit...

src/Microsoft.ML.Onnx/ProtabilityCatalog.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// Convert the specified <see cref="ITransformer"/> to ONNX format. Note that ONNX uses Google's Protobuf so the returned value is a Protobuf object.
+        /// </summary>
+        /// <param name="catalog">A field in <see cref="MLContext"/> which this function associated with.</param>
+        /// <param name="transform">The <see cref="ITransformer"/> that will be converted into ONNX format.</param>
+        /// <param name="inputData">The input of the specified transform.</param>
+        /// <returns></returns>
+        public static ModelProto ConvertToOnnx(this ModelOperationsCatalog.PortabilityTransforms catalog, ITransformer transform, IDataView inputData)
+        {
+            var env = new MLContext(seed: 1);
+            var ctx = new OnnxContextImpl(env, "model", "ML.NET", "0", 0, "com.microsoft", OnnxVersion.Stable);
```
TomFinley: >com.microsoft [](start = 75, length = 13)

I am not an expert on ONNX, but is this some sort of namespace? When this was functionality that was meant for internal Microsoft consumption, having this be a default may have been appropriate, but when shipped as ML.NET (which involves, of course, many non-Microsoft), should it not be something the user specifies? Or is its role something I don't fully appreciate? (Could be that `com.microsoft` is entirely possible here, I'd just like to understand why if so, because that seems odd. If it is appropriate, surely a comment to that effect would be desirable.)

src/Microsoft.ML.Onnx/SaveOnnxCommand.cs
```diff
@@ -134,7 +136,50 @@ private void GetPipe(OnnxContextImpl ctx, IChannel ch, IDataView end, out IDataV
                 transform = (source = transform.Source) as IDataTransform;
             }
 
-            Host.AssertValue(source);
+            Contracts.AssertValue(source);
+        }
+
+        internal static ModelProto ConvertTransformListToOnnxModel(OnnxContextImpl ctx, IDataView inputData, IDataView outputData,
+            LinkedList<ITransformCanSaveOnnx> transforms, HashSet<string> inputColumnNamesToDrop=null, HashSet<string> outputColumnNamesToDrop=null)
```
TomFinley: >=null [](start = 142, length = 5)

Maybe ctrl-k-d to make VS format the file? What you think?

src/Microsoft.ML.Onnx/SaveOnnxCommand.cs
```diff
@@ -113,9 +114,10 @@ public override void Run()
             }
         }
 
-        private void GetPipe(OnnxContextImpl ctx, IChannel ch, IDataView end, out IDataView source, out IDataView trueEnd, out LinkedList<ITransformCanSaveOnnx> transforms)
+        internal static void GetPipe(OnnxContextImpl ctx, IChannel ch, IDataView end, out IDataView source, out IDataView trueEnd, out LinkedList<ITransformCanSaveOnnx> transforms)
         {
-            Host.AssertValue(end);
+            Contracts.AssertValue(end);
```
TomFinley: >Contracts [](start = 12, length = 9)

Why not `ch`?

src/Microsoft.ML.Onnx/SaveOnnxCommand.cs
```diff
@@ -134,7 +136,50 @@ private void GetPipe(OnnxContextImpl ctx, IChannel ch, IDataView end, out IDataV
                 transform = (source = transform.Source) as IDataTransform;
             }
 
-            Host.AssertValue(source);
+            Contracts.AssertValue(source);
```
TomFinley: >Contracts [](start = 12, length = 9)

Similar note here.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -60,6 +62,17 @@ internal ExplainabilityTransforms(ModelOperationsCatalog owner) : base(owner)
             }
         }
 
+        /// <summary>
+        /// The catalog of model protability operations. Member function of this classes are able to convert the associated object to a protable format,
+        /// so that the fitted pipeline can easily be depolyed to other platforms. Currently, the only supported format is ONNX (https://github.com/onnx/onnx).
+        /// </summary>
+        public sealed class PortabilityTransforms : SubCatalogBase
```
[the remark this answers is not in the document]
wschin: Two changes are made.
(1) Remove PortabilityTransforms
(2) Move SaveToOnnx to a direct extension to ModelOperationsCatelog

---
In reply to: [245425207](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245425207) [](ancestors = 245425207)

src/Microsoft.ML.Onnx/ProtabilityCatalog.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Collections.Generic;
+using Microsoft.ML.Core.Data;
```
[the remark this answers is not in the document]
wschin: No problem.

---
In reply to: [245425377](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245425377) [](ancestors = 245425377)

src/Microsoft.ML.Onnx/ProtabilityCatalog.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.ML.UniversalModelFormat.Onnx;
+
+namespace Microsoft.ML
+{
+    public static class ProtabilityCatalog
+    {
+        /// <summary>
+        /// Convert the specified <see cref="ITransformer"/> to ONNX format. Note that ONNX uses Google's Protobuf so the returned value is a Protobuf object.
+        /// </summary>
+        /// <param name="catalog">A field in <see cref="MLContext"/> which this function associated with.</param>
```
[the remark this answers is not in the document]
wschin: We use `The class that <see cref="ConvertToOnnx(ModelOperationsCatalog, ITransformer, IDataView)"/> attached to` now.

---
In reply to: [245425596](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245425596) [](ancestors = 245425596)

src/Microsoft.ML.Onnx/ProtabilityCatalog.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <param name="transform">The <see cref="ITransformer"/> that will be converted into ONNX format.</param>
+        /// <param name="inputData">The input of the specified transform.</param>
+        /// <returns></returns>
```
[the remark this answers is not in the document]
wschin: `An ONNX model equivalent to the converted ML.NET model.` will be added.

---
In reply to: [245425735](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245425735) [](ancestors = 245425735)

src/Microsoft.ML.Onnx/ProtabilityCatalog.cs
```diff
@@ -0,0 +1,32 @@
...
+        public static ModelProto ConvertToOnnx(this ModelOperationsCatalog.PortabilityTransforms catalog, ITransformer transform, IDataView inputData)
```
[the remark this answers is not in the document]
wschin: No problem.

---
In reply to: [245425439](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245425439) [](ancestors = 245425439)

src/Microsoft.ML.Onnx/ProtabilityCatalog.cs
wschin: I will make `ModelOperationsCatalog.Environment` a BestFriend and access it from here (another assembly).

---
In reply to: [245427576](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245427576) [](ancestors = 245427576)

src/Microsoft.ML.Onnx/ProtabilityCatalog.cs
wschin: It should be "ai.onnx.ml". Sorry about causing confusion. Currently, we don't have anything Microsoft-specific.

---
In reply to: [245428777](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245428777) [](ancestors = 245428777)

src/Microsoft.ML.Onnx/SaveOnnxCommand.cs
wschin: No problem.

---
In reply to: [245428925](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245428925) [](ancestors = 245428925)

src/Microsoft.ML.Onnx/SaveOnnxCommand.cs
wschin: Sure. Thank you.

---
In reply to: [245429058](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245429058) [](ancestors = 245429058)

src/Microsoft.ML.Onnx/SaveOnnxCommand.cs
wschin: Formatted ModelOperationsCatalog.cs, OnnxConversionTest.cs, and ProtabilityConversion.cs.

---
In reply to: [245428880](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245428880) [](ancestors = 245428880)

src/Microsoft.ML.Onnx/ProtabilityCatalog.cs
wschin: No problem. File name will be changed accordingly.

---
In reply to: [245427449](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245427449) [](ancestors = 245427449)

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -0,0 +1,225 @@
+﻿using System;
+using System.IO;
```
TomFinley: Don't forget your header!

src/Microsoft.ML.Onnx/ProtabilityCatalog.cs
TomFinley: No idea, since I have no idea what it is for. 😄 But perhaps there are experts elsewhere we can ask to decide on the right answer.

wschin: Because ML.NET is not Microsoft-specific, I just change it to `machinelearning.dotnet` so that `machinelearning.dotnet` + `ML.NET converter version` + `model name` can be unique IDs of models. 

---
In reply to: [245446247](https://github.com/dotnet/machinelearning/pull/2013#discussion_r245446247) [](ancestors = 245446247)

wschin:
Thank you Zeeshan.

Issue #1933, opened 2018-12-20
Enable running tests against .NET Framework


danmosemsft:
Right now the tests are only built and run against `netcoreapp2.1`. We want to be able to run our unit tests on .NET Framework as well, as we will support that when we release 1.0

1. The build part is because of [here](https://github.com/dotnet/machinelearning/blob/master/test/Directory.Build.props#L5-L8). The tests either need to be built against `netstandard2.0` or multitargeted. Building against `netstandard2.0` (modifying the line above) fails because the tests depend on the package `xunit.runner.visualstudio 2.4.0` which it happens does not have support for `netstandard2.0`. So probably this line needs to be changed to something like `<DefaultTestTargetFramework>net461,netstandard2.0</DefaultTestTargetFramework>` (comma separated) so that it produces assets for both. Not sure which version of .NET Framework to use here. Whichever it is, the developer will need reference assemblies installed so perhaps we can pick whatever the default is for those.

2. The running part I did not investigate. Maybe the above is sufficient and it's just another flag to pass to run against .NET Framework.

danmosemsft:
cc @ericstj @Anipik

Issue #1072, opened 2018-09-27
The new API does not work on .NET Framework < 4.7.1


yaeldekel:
It relies on the extension method in System.Linq "Append" that was added only in 4.7.1.

eerhardt:
Unfortunately there are a few methods in .NET Standard 2.0 which aren't available in .NET Framework 4.6.1, but 4.6.1 "claims" it supports .NET Standard 2.0.

Here is the list:  https://github.com/dotnet/standard/blob/205332796a9a0f869ff698402c3496f14cc71f50/netstandard/src/ApiCompatBaseline.net461.txt

/cc @ericstj @terrajobst @weshaggard

markusweimer:
Does this mean we'd have to raise the min version of .NET Framework required for ML.NET 0.6?

eerhardt:
> Does this mean we'd have to raise the min version of .NET Framework required for ML.NET 0.6?

If we don't fix this issue for 0.6, Yes. If someone tries running the new API on an older .NET Framework, they will an exception.

markusweimer:
@TomFinley how strong is our dependency on the method mentioned by Yael? Can we code around this? Maybe for 0.7?

Zruty0:
umm, I think we could just add our implementation of `Append` to TLC until we upgrade the .NET version?

`Append(x)` is the same as `Concat(new[] {x})`

markusweimer:
> umm, I think we could just add our implementation of Append to TLC until we upgrade the .NET version?

Sounds good. Can we make it conditional on the version of .NET a user has?

Zruty0:
Also, I'm not too strongly attached to `Append`. I only used it because my C# was rusty, and I forgot that 'there is no Append in LINQ'. And sure enough, it appeared while I was not looking.

markusweimer:
@Zruty0, is this fixable between now and 0.6?

yaeldekel:
I can create a PR today.

yaeldekel:
I looked at the list that @eerhardt posted, and Append looks like the only API that we use in there.

markusweimer:
Cool! Unless you have a *bug* to fix for 0.6, it would be great to get this done before the release.

PR #2018, opened 2019-01-04
Adding a new CI leg for netframework

Anipik:
Fixes https://github.com/dotnet/machinelearning/issues/1933
We are currently not running or building our tests on netframework.
However we are going to support netframework when we release 1.0 , so just adding a new ci leg to consistently build and test stuff on netframework


test/Microsoft.ML.CodeAnalyzer.Tests/Code/BestFriendTest.cs
```diff
@@ -66,8 +66,9 @@ public void BestFriend()
                 peRef = MetadataReference.CreateFromImage(peImage);
             }
 
+            refs.ToList().Add(peRef);
             var comp = projB.GetCompilationAsync().Result
-                .WithReferences(refs.Append(peRef).ToArray());
+                .WithReferences(refs.ToArray());
```
danmosemsft: Part of the PR?

Anipik: yes append function is not available on .netframwork(This project was earlier targeting the .netcoreapp2.1)

artidoro:
Since you are working on this could you order the CI legs alphabetically.

build/ci/phase-template.yml
```diff
@@ -4,13 +4,15 @@ parameters:
   buildScript: ''
   queue: {}
   customMatrixes: ''
+  defaultTestTargetFramework: 'netcoreapp2.1'
```
safern: nit, no need to use quotes.

test/Directory.Build.props
```diff
@@ -2,7 +2,6 @@
   <Import Project="..\Directory.Build.props" />
 
   <PropertyGroup>
-    <DefaultTestTargetFramework>netcoreapp2.1</DefaultTestTargetFramework>
```
safern: What if I want to just run one test project with dotnet test? I think DefaultTestTargetFramework should be defaulted to netcoreapp2.1 whenever it is not set.

```xml
<DefaultTestTargetFramework Condition="'$(DefaultTestTargetFramework)' == ''">netcoreapp2.1</DefaultTestTargetFramework>
<DefaultTestTargetFramework Condition="'$(UseIntrinsics)' == 'true' AND '$(DefaultTestTargetFramework)' == ''">netcoreapp3.0</DefaultTestTargetFramework>
```

Something like that maybe?

test/Microsoft.ML.TestFramework/Microsoft.ML.TestFramework.csproj
```diff
@@ -1,6 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <DefineConstants>CORECLR</DefineConstants>
+    <DefineConstants Condition="'$(DefaultTestTargetFramework)' == 'net462'">CORECLR;NETFX</DefineConstants>
```
safern: You might just want to append the DefineConstants and that will bring all the common constants for both frameworks, i.e, in this case CORECLR.
```suggestion
    <DefineConstants Condition="'$(DefaultTestTargetFramework)' == 'net462'">$(DefineConstants);NETFX</DefineConstants>
```

build/ci/phase-template.yml
```diff
@@ -4,13 +4,15 @@ parameters:
   buildScript: ''
   queue: {}
   customMatrixes: ''
+  defaultTestTargetFramework: netcoreapp2.1
```
eerhardt: I don't think we should have this default here - it is duplicate of the default that is in the msbuild `.props` file: [https://github.com/dotnet/machinelearning/blob/master/test/Directory.Build.props#L5](https://github.com/dotnet/machinelearning/blob/master/test/Directory.Build.props#L5).

It would be great if the default is only put in a single place, so we only need to update a single place.

config.json
```diff
@@ -6,6 +6,13 @@
       "values": [ "Debug", "Release", "Debug-Intrinsics", "Release-Intrinsics" ],
       "defaultValue": "Debug"
     },
+    "DefaultTestTargetFramework" : 
+    {
+      "description": "Sets the Target Framework for the tests",
+      "valueType": "property",
+      "values": [ "netcoreapp2.1", "net462"],
```
eerhardt: Is it possible to not have the list here? Can it just accept any string?

config.json
```diff
@@ -6,6 +6,13 @@
...
+      "defaultValue": "netcoreapp2.1"
```
eerhardt: Again, hard-coding this default outside of MSBuild is less than ideal.

Directory.Build.props
```diff
@@ -5,6 +5,7 @@
   <Import Project="build/Dependencies.props" />
 
   <PropertyGroup>
+    <DefaultTestTargetFramework Condition="'$(DefaultTestTargetFramework)' == ''">netcoreapp2.1</DefaultTestTargetFramework>
```
eerhardt: This doesn't belong here. It is a "test" only property. It should be in the `test` directory (and it already is [https://github.com/dotnet/machinelearning/blob/master/test/Directory.Build.props#L5](https://github.com/dotnet/machinelearning/blob/master/test/Directory.Build.props#L5)).

docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>$(DefaultTestTargetFramework)</TargetFramework>
```
eerhardt: This doesn't make sense.  The `docs` shouldn't depend on a Test specific property.

src/Microsoft.ML.KMeansClustering/Microsoft.ML.KMeansClustering.csproj
```diff
@@ -1,8 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
     <IncludeInPackage>Microsoft.ML</IncludeInPackage>
+    <TargetFrameworks>netstandard2.0;$(DefaultTestTargetFramework)</TargetFrameworks>
```
eerhardt: This should be reverted.

src/Microsoft.ML.Console/Microsoft.ML.Console.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>$(DefaultTestTargetFramework)</TargetFramework>
```
eerhardt: Again - we should have `src` projects depend on a `test` property.

test/Microsoft.ML.CodeAnalyzer.Tests/Code/BestFriendTest.cs
```diff
@@ -66,8 +66,11 @@ public void BestFriend()
                 peRef = MetadataReference.CreateFromImage(peImage);
             }
 
+            var refList = refs.ToList();
```
eerhardt: Might as well just make `refs` a List to start.

test/Microsoft.ML.FSharp.Tests/Microsoft.ML.FSharp.Tests.fsproj
```diff
@@ -4,7 +4,7 @@
     <!-- Empty TargetFramework because test\Directory.Build.props sets TFM, but FSharp.Tests is multi-targeted -->
     <TargetFramework></TargetFramework>
     <TargetFrameworks>$(DefaultTestTargetFramework)</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Unix'">$(TargetFrameworks); net461</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' != 'Unix'">$(TargetFrameworks)</TargetFrameworks>
```
eerhardt: This whole line can be removed.

test/Microsoft.ML.TestFramework/Microsoft.ML.TestFramework.csproj
eerhardt: You don't need to define your own constant here. They are generated automatically by the build system.

See [https://docs.microsoft.com/en-us/dotnet/standard/frameworks](https://docs.microsoft.com/en-us/dotnet/standard/frameworks)

public class MyClass
{
    static void Main()
    {
#if NET40
        Console.WriteLine("Target framework: .NET Framework 4.0");
#elif NET45  
        Console.WriteLine("Target framework: .NET Framework 4.5");
#else
        Console.WriteLine("Target framework: .NET Standard 1.4");
#endif
    }
}

---
In reply to: [246516157](https://github.com/dotnet/machinelearning/pull/2018#discussion_r246516157) [](ancestors = 246516157)

test/Directory.Build.props
```diff
@@ -2,10 +2,9 @@
...
     <DefaultTestTargetFramework Condition="'$(UseIntrinsics)' == 'true'">netcoreapp3.0</DefaultTestTargetFramework>
-    
-    <TargetFramework>$(DefaultTestTargetFramework)</TargetFramework>
+    <TargetFrameworks>$(DefaultTestTargetFramework)</TargetFrameworks>
```
eerhardt: We shouldn't be setting `TargetFrameworks` - this should be `TargetFramework`

test/Microsoft.ML.Tests/Transformers/NormalizerTests.cs
```diff
@@ -324,7 +324,7 @@ public void WhiteningWorkout()
             Done();
         }
 
-        [Fact]
+        [ConditionalFact(typeof(BaseTestBaseline), nameof(BaseTestBaseline.NotFullFramework))]
```
eerhardt: we should be investigating why these tests are failing. That is part of why we are doing this exercise.

Anipik: The failure on framework is a native dependency issue, I will try to resolve it before merging this one.

Anipik: i opened the issue https://github.com/dotnet/machinelearning/issues/2104
This is the similar issue i.e while using maml we are not sometimes able to load some modules

config.json
```diff
@@ -112,6 +112,18 @@
             "Configuration": "Release-Intrinsics"
           }
         },
+        "debug-netfx": {
+          "description": "Sets optimization level to debug for managed build configuration and builds against fullframework. (/p:Configuration=Debug-netfx)",
+          "settings": {
+            "Configuration": "Debug-netfx"
```
safern: I wouldn't use netfx. I would use the actual target group so that this can scale. If we ever want to run them against another target group we can just add a new configuration. For example the configuration would be `Debug-net461` -- then in the msbuild props I would check if Configuration contains -net, then strip down the framework from what is after the `-` and set it as the `DefaultTestTargetFramework`. @eerhardt what do you think?

Anipik: I choose this name because I though if in future we want to upgrade the version of netfx. we will just need to update  ```<DefaultTestTargetFramework Condition="$(Configuration.Contains('-netfx'))">net462</DefaultTestTargetFramework>```

But yeah if there is possibility that we want to run for different versions of netfx at the same time or in ci, then we may have to change some things.

safern: Agree, to upgrade the current version it is easier the way yo have it, but if we end up choosing multiple versions then we will have to do what I suggested. Disregard then.

test/Directory.Build.props
```diff
@@ -3,9 +3,10 @@
 
   <PropertyGroup>
     <DefaultTestTargetFramework>netcoreapp2.1</DefaultTestTargetFramework>
+    <DefaultTestTargetFramework Condition="$(Configuration.Contains('-netfx'))">net462</DefaultTestTargetFramework>
```
safern: Use, .EndsWith rather than Contains.
```suggestion
    <DefaultTestTargetFramework Condition="$(Configuration.EndsWith('-netfx'))">net462</DefaultTestTargetFramework>
```

.vsts-dotnet-ci.yml
```diff
@@ -18,36 +32,36 @@ phases:
...
     customMatrixes:
-      Build_Debug_Intrinsics:
-        _configuration: Debug-Intrinsics
-        _config_short: DI
-      Build_Release_Intrinsics:
-        _configuration: Release-Intrinsics
-        _config_short: RI
+      Build_Debug_netfx:
+        _configuration: Debug
+        _config_short: DF
```
safern: DFX? F doesn't tell me much.

.vsts-dotnet-ci.yml
```diff
@@ -18,36 +32,36 @@ phases:
...
     buildScript: build.cmd
     customMatrixes:
-      Build_Debug_Intrinsics:
-        _configuration: Debug-Intrinsics
-        _config_short: DI
-      Build_Release_Intrinsics:
-        _configuration: Release-Intrinsics
-        _config_short: RI
+      Build_Debug_netfx:
+        _configuration: Debug
```
safern: I just realized this is wrong. This configurations should be have -netfx at the end. You're building for netcoreapp2.1 currently. 
```suggestion
        _configuration: Debug-netfx
```

.vsts-dotnet-ci.yml
```diff
@@ -18,36 +32,36 @@ phases:
...
+        _config_short: DF
+      Build_Release_netfx:
+        _configuration: Release
```
safern:
```suggestion
        _configuration: Release-netfx
```

Anipik:
> This looks way better and it also enables people building for netfx from VS. Did you test that choosing that configuration worked fine? Intellisense and all the VS features?

The build from VS is working fine. the test explorer is giving some trouble. I will try to give it another try. (could be becuase of my VS as VS2019 is not working smoothly for me)

test/Microsoft.ML.CodeAnalyzer.Tests/Code/BestFriendTest.cs
eerhardt: The `Append` method is one of those rare APIs that is in `netstandard2.0` but we didn't ship in `net461`. We had a similar bug in the product - see https://github.com/dotnet/machinelearning/issues/1072

test/Microsoft.ML.FSharp.Tests/Microsoft.ML.FSharp.Tests.fsproj
```diff
@@ -4,7 +4,6 @@
     <!-- Empty TargetFramework because test\Directory.Build.props sets TFM, but FSharp.Tests is multi-targeted -->
```
eerhardt: Actually, this whole chunk can be removed now. F# tests will get netfx coverage through the netfx leg. There is no reason for this project to be multi-targeted.

Remove this comment and the 2 lines below:

```
    <TargetFramework></TargetFramework>
    <TargetFrameworks>$(DefaultTestTargetFramework)</TargetFrameworks>
```

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -41,7 +41,7 @@ public OnnxConversionTest(ITestOutputHelper output) : base(output)
         /// call <see cref="OnnxScoringEstimator"/> to evaluate that file. The outputs of <see cref="OnnxScoringEstimator"/> are checked against the original
         /// ML.NET model's outputs.
         /// </summary>
-        [Fact]
+        [ConditionalFact(typeof(BaseTestBaseline), nameof(BaseTestBaseline.NotFullFramework))]
```
eerhardt: Can you add a link to [https://github.com/dotnet/machinelearning/issues/2106](https://github.com/dotnet/machinelearning/issues/2106) here?

Anipik:
Thanks to @safern  and @eerhardt for both  feedback and help

shauheen:
#2104 #2106 need to be addressed.

PR #2020, opened 2019-01-04
Convert CSharpAPI based tests to EntryPoint based test.

codemzs:
Converts [CSharpAPI tests](https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Core.Tests/UnitTests/TestCSharpApi.cs) to entrypoints based test. CSharpAPI converts the pipeline to a json graph which is converted by entrypoint infrastructure to MLNET pipeline via entrypoints. This change converts the CSharp based pipeline to Json graph and invokes the graph runner, hence preserving code coverage of entry points infrastructure and ML.NET transforms/trainers, etc.

Code coverage seems to drop by 0.09%, bulk of it seems to be coming from microsoft.ml.core.tests.dll - 0.55% where TestCSharpAPI.cs file was deleted, followed by microsoft.ml.core.dll - 0.24%, microsoft.ml.maml.dll - 0.13%, microsoft.ml.transforms.dll - 0.02%.


test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -3868,6 +3868,1794 @@ public void EntryPointHashJoinCountTable()
                 });
         }
 
+        [Fact]
+        public void TestSimpleExperiment()
+        {
+            var dataPath = GetDataPath("adult.tiny.with-schema.txt");
+            string inputGraph = @"{
+                'Nodes':
+                [{
```
abgoswam: am curious, how does one construct the json graph like the ones you used in these tests ?

codemzs: json is my second language in case you didn't know :) jokes aside, the old CSharpAPI basically constructs a json graph underneath and then passes it to the graph runner that executes the pipeline by calling into entrypoints. I stepped into the debugger and got the graph and replaced the pipeline with it while keeping rest of the test the same. I used graph runner to run the graph and get the idataview object that the test was using.

PR #1971, opened 2018-12-27
WIP End of an era - Delete Microsoft.ML.Legacy and related tests.

codemzs:
I still need to remove all those baseline files that are no longer used and then document all the tests that may need to be re-written using the new API such as ONNX converter test and FSharp test. In the end I want to compare how much the code coverage has changed between this change and master. 

CC: @TomFinley, @sfilipi

PR #2021, opened 2019-01-04
Remove Legacy dependency in ML.FSharp.Tests

artidoro:
Part of the effort described in #1971.

This PR removes Legacy code in ML.TestFramework and ML.FSharp.Tests, and replaces it with the new Estimator, Transformer API. Specifically, it changes tests that were using ModelHelper.cs (although those tests were skipped anyway as the dataset is missing). It also changes the F# code in the F# smoke tests to use the new API.  

I checked the code coverage effect of this PR, and it increases the coverage by 0.02% (possibly for deleting test utility code that was not used).

EDIT:
We decided to remove the tests relying on the housing dataset in PR #2024. I am therefore reverting my changes on those tests and letting that PR remove all references to Legacy in the TestFramework DLL.


test/Microsoft.ML.Tests/Scenarios/HousePricePredictionTests.cs
```diff
@@ -16,16 +16,17 @@ public partial class ScenariosTests : BaseTestClass
          The application will let users submit information about their house, and see a price they could expect if they put the house for sale.
          Because real estate transaction data is public, Contoso has historical data they intend to use to train Machine Learning prediction engine. 
         */
-#pragma warning disable 612
         [Fact(Skip = "Missing data set. See https://github.com/dotnet/machinelearning/issues/3")]
-        public async void PredictHousePriceModelTest()
```
abgoswam: In my upcoming PR, I am deleting this test .

This test is being skipped so not not adding to code coverage.  Also, I believe this test was  testing the model  load and prediction path.   

In your PR,  you are using a model  generated using doing .Fit()  but there is already an existing test for it `HousePriceTrainAndPredictionTests.cs`

@codemzs .  Thoughts ?

artidoro: As discussed in person, I am reverting my changes and letting @abgoswam remove all tests that use the housing data set. See PR #2024 for more details.

test/Microsoft.ML.FSharp.Tests/SmokeTests.fs
```diff
@@ -77,51 +74,23 @@ module SmokeTest1 =
...
-                NumLeaves = 5, 
-                NumTrees = 5, 
-                MinDocumentsInLeafs = 2
-            ))
-
-        let model = pipeline.Train<SentimentData, SentimentPrediction>()
+        let ml = MLContext(seed = new System.Nullable<int>(1), conc = 1)
+        let data = ml.Data.ReadFromTextFile<SentimentData>(testDataPath, hasHeader = true)
+
+        let pipeline = ml.Transforms.Text.FeaturizeText("SentimentText", "Features")
```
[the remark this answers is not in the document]
artidoro: By default it is set to 10. I am not sure how to create an Action<Arguments> object in F#, and I believe this will not change the code coverage. We have tests that are specific to FastTree for binary classification that set the MinDocumentsInLeafs to something different.

PR #2029, opened 2019-01-04
Added more documentation and extra CPU path test

jignparm:
Updated XML docs for public estimator methods.
Update XML to point to new GPU package
Added tests for extra CPU paths (gpuDeviceId=null, fallBackToCPU=true)

### DIFF
--- a/src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
+++ b/src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
@@ -52,7 +52,7 @@ namespace Microsoft.ML.Transforms
     /// , and set the parameter 'gpuDeviceId' to a valid non-negative integer. Typical device ID values are 0 or 1.
     /// </p>
     /// <p>The inputs and outputs of the ONNX models must be Tensor type. Sequence and Maps are not yet supported.</p>
-    /// <p>OnnxRuntime currently works on Windows x64 bit platforms only. Linux and OSX to be supported soon.</p>
+    /// <p>OnnxRuntime currently works on Windows 64-bit platforms only. Linux and OSX to be supported soon.</p>
     /// <p>Visit https://github.com/onnx/models to see a list of readily available models to get started with.</p>
     /// <p>Refer to http://onnx.ai' for more information about ONNX.</p>
     /// </remarks>
--- a/src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
+++ b/src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
@@ -42,8 +42,8 @@ namespace Microsoft.ML.Transforms
     /// </format>
     /// </summary>
     /// <remarks>
-    /// <p>Supports inferencing of models in 1.2 and 1.3 format, using the
-    /// <a href='https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime/'>Microsoft.ML.OnnxRuntime</a> library.
+    /// <p>Supports inferencing of models in ONNX 1.2 and 1.3 format (opset 7, 8 and 9), using the
+    /// <a href='https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu/'>Microsoft.ML.OnnxRuntime.Gpu</a> library.
     /// </p>
     /// <p>Models are scored on CPU by default. If GPU execution is needed (optional), install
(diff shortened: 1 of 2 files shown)


test/Microsoft.ML.OnnxTransformTest/OnnxTransformTests.cs
```diff
@@ -126,8 +126,10 @@ void TestSimpleCase()
             catch (InvalidOperationException) { }
         }
 
-        [ConditionalFact(typeof(Environment), nameof(Environment.Is64BitProcess))] // x86 fails with "An attempt was made to load a program with an incorrect format."
-        void TestOldSavingAndLoading()
+        [ConditionalTheory(typeof(Environment), nameof(Environment.Is64BitProcess))] // x86 fails with "An attempt was made to load a program with an incorrect format."
+        [InlineData(null, false)]
```
sfilipi: >InlineData [](start = 9, length = 10)

haven't seem this before. What is it for?

jignparm: The InlineData are values passed into the method (so gpuDeviceId=Null, and fallBackToCpu=false for the first run)

Each InlineData creates 2 extra tests.

This only works with Theory methods (not Fact methods)

---
In reply to: [245449923](https://github.com/dotnet/machinelearning/pull/2029#discussion_r245449923) [](ancestors = 245449923)

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
```diff
@@ -531,13 +532,32 @@ public NamedOnnxValue GetNamedOnnxValue()
...
+        /// <summary>
+        /// Transform for scoring ONNX models. Input data column names/types must exactly match
+        /// all model input names. All possible output columns are generated, with names/types
+        /// specified by model.
+        /// </summary>
+        /// <param name="env">The environment to use.</param>
+        /// <param name="modelFile">Model file path.</param>
+        /// <param name="gpuDeviceId">Optional GPU device ID to run execution on. Null for CPU.</param>
+        /// <param name="fallbackToCpu">If GPU error, raise exception or fallback to CPU.</param>
+        public OnnxScoringEstimator(IHostEnvironment env, string modelFile, int? gpuDeviceId = null, bool fallbackToCpu = false)
```
justinormont: I think it's better for the user to default `fallbackToCpu` to true. This lets the user set GPU and should it fail, let the run continue (hopefully w/ a warning).

jignparm: If a user specifies a GPU id , then the expectation is to run on that device only, and error out if that's not the case.

Setting it to True will incur more 'silent failures', and therefore slower results in general. 



---
In reply to: [245454861](https://github.com/dotnet/machinelearning/pull/2029#discussion_r245454861) [](ancestors = 245454861)

justinormont: Yes, though I'm always advocating for a better user experience. Some image model can be slow enough on a CPU to be a disincentive for users due to runtime. 

I prefer: (1) try to init GPU, (2) use if available; user can override and say use CPU. This causes the happy path to be fast and lets the tyranny of defaults to fall in the user's favor.

jignparm: The 'always-try-gpu-first' may need some thoughts on the following questions:

1> Which deviceID to run on if  multiple are available? Only idle-devices, or in-use-devices are OK as well? 

   -  if in-use-devices are Ok (because they are only lightly loaded), then what are good thresholds to use (e.g. use only if at least 50% GPU memory is free).   It's possible that loading 1 more process onto that device will cause out-of-memory exceptions.

   - If only-idle devices are OK, then should we check that they are idle 'now', or idle in last n-seconds? It is OK if a device has been idle for only 1 micro-second, or should check that it's been idle for 5 minutes?

2> GPU may not be the fastest execution provider. For tiny models, CPU execution may be way faster -- can we tell by model size which one would benefit from GPU? 

3> Also, would/should the user know that this app will grab his GPU by default? Is it OK to do so without their knowledge?

Leaving this value as-is for now -- can update later if needed.


 

---
In reply to: [245460863](https://github.com/dotnet/machinelearning/pull/2029#discussion_r245460863) [](ancestors = 245460863)

PR #2030, opened 2019-01-04
Fix build break

abgoswam:
Fix to mitigate the build break.


abgoswam:
Thanks Zeeshan for merging!

PR #2043, opened 2019-01-06
End of an era - Delete Microsoft.ML.Legacy

codemzs:
Deletes Microsoft.ML.Legacy project and related references/files including BatchPredictionEngine.

replaces #1971


src/Microsoft.ML.StandardLearners/Standard/MultiClass/doc.xml
```diff
@@ -14,10 +14,6 @@
         This multi-class trainer accepts binary feature values of type float, i.e., feature values are either true or false.
         Specifically a feature value greater than zero is treated as true.
       </remarks>
-      <seealso cref='Microsoft.ML.Legacy.Trainers.LogisticRegressionClassifier'/>
-      <seealso cref='Microsoft.ML.Legacy.Trainers.LightGbmClassifier'/>
-      <seealso cref='Microsoft.ML.Legacy.Trainers.StochasticDualCoordinateAscentClassifier'/>
-      <seealso cref='Microsoft.ML.Legacy.Models.OneVersusAll'/>
```
artidoro: Could we keep a reference to the trainers here? The trainers with the new API

src/Microsoft.ML.StandardLearners/Standard/MultiClass/doc.xml
```diff
@@ -47,10 +43,6 @@
         request caching, as it will be performing multiple passes over the data set.
         These learner will request normalization from the data pipeline if the classifier indicates it would benefit from it.</para>
       </remarks>
-      <seealso cref='Microsoft.ML.Legacy.Trainers.LogisticRegressionClassifier'/>
-      <seealso cref='Microsoft.ML.Legacy.Trainers.LightGbmClassifier'/>
-      <seealso cref='Microsoft.ML.Legacy.Trainers.StochasticDualCoordinateAscentClassifier'/>
-      <seealso cref='Microsoft.ML.Legacy.Trainers.NaiveBayesClassifier'/>
```
artidoro: Same here.

src/Microsoft.ML.Data/Data/RowCursorUtils.cs
```diff
@@ -325,30 +325,6 @@ private static Func<bool> GetIsNewGroupDelegateCore<T>(Row cursor, int col)
             };
         }
 
-        [Obsolete("The usages of this appear to be based on a total misunderstanding of what Batch actually is. It is a mechanism " +
-            "to enable sharding and recovery of parallelized data, and has nothing to do with actual data.")]
-        [BestFriend]
-        internal static Func<bool> GetIsNewBatchDelegate(Row cursor, int batchSize)
```
yaeldekel: >GetIsNewBatchDelegate [](start = 35, length = 21)

This is used in our internal repo. it is used for scoring models that score a whole batch of examples at a time. It may be useful here in the future as well, for example, TF has the capability to score multiple examples at a time as well (may give speedups when using GPUs). 

By the comment on the ObsoleteAttribute above, it seems that the name of this method might be misleading - perhaps we could keep the method but find a more appropriate name for it?

codemzs: Thanks, @yaeldekel! Shouldn’t this be in the internal repo then? Otherwise this is just dead code sitting here with an assumption it may be used in future.

TomFinley: Hi @yaeldekel and @codemzs, I wrote this obsolete comment. Whoever wrote this method, as I said, completely misunderstood what `Batch` on a row cursor was for. Using `Batch` in this fashion is completely pointless -- the batch is meant as a reconcilation mechanism for parallel cursors, as described in [its documentation](https://github.com/dotnet/machinelearning/blob/faffd179c961f120ec4e6babb06bbfb2cca6a6ea/src/Microsoft.ML.Core/Data/IDataView.cs#L169), further [code documentation](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Core/Data/ICursor.md#batch), literally every single usage of `Batch` throughout the codebase.

@codemzs's removal of it is appropriate. The internal code may be using it, but, as explained, that's because someone committed a blunder.

src/Microsoft.ML.StandardLearners/Standard/MultiClass/doc.xml
codemzs: @sfilipi had asked to swap these names out in the previous PR.

src/Microsoft.ML.StandardLearners/Standard/MultiClass/doc.xml
codemzs: @sfilipi had asked to swap these names out in the previous PR.

test/Microsoft.ML.Tests/TextLoaderTests.cs
```diff
@@ -133,10 +132,11 @@ public void TestTextLoaderInvalidLongMax()
 
     public class TextLoaderTests : BaseTestClass
     {
+        ConsoleEnvironment env;
```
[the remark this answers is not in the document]
codemzs: ConsoleEnvironment  seems to be used as the global environment type in tests.

codemzs:
Thank you @TomFinley and @artidoro for reviewing.

Issue #1798, opened 2018-11-30
Creation of components through MLContext: advanced options and other feedback


TomFinley:
In our estimators and other similar components often have advanced settings, because, sometimes people have unusual circumstances. At the same time, there is a 95% or 99% scenario for "simple" usage that most people will be happy with. For this reason we have often made a distinction between common and advanced settings, as we see here.

https://github.com/dotnet/machinelearning/blob/cb37c7e7f1e1b29b5608a2755db793c5435d10b1/src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs#L29-L38

There are some possible things that excited feedback:

1. Echoing feedback seen in #1690, these things where we're making something should have the prefix `Create`, even in situations where this a catalog where we are *always* creating. Note: `Create` preferred to `Make`.

2. The worth of ASP.NET style configuration was questioned (seen above as `Action<FastTreeRegressionTrainer.Arguments> advancedSettings`), e.g., there may not be much purpose in having a delegate. The older style where it just takes the `Arguments` *period* was preferred.

3. Having this `Arguments` object as a nested class the component being created was viewed as positive, but this would be more idiomatically called `Options` -- `Arguments` was a holdover name from when these were exclusively for command line arguments, but for the API this is not a great name. So while keeping the general structure of how they are placed currently, they should probably be renamed to `Options`.

4. It is good to have the convenience for the simple arguments, however, if we have both simple and advanced settings, we should not mix them but have instead two distinct constructors/extension methods. (E.g., in the above, we would have two methods, one that took the advanced options.) To do otherwise is to invite confusion about which "wins" if we have the setting set in both.

    * Note that phase setting "set in both," which suggests that these settings object should retain the "simpler" settings in them. This reinforces feedback elsewhere as seen [here](https://github.com/dotnet/machinelearning/pull/1352#discussion_r228319239).

5. If the simple arguments are totally sufficient, then there is no need to expose this `Arguments` class in hte public API. (For practical reasons relating to command line and entry-point usage, we still need to always have these `Arguments` objects, but if they serve no purpose for the API the class can simply be made internal.)

/cc @KrzysztofCwalina, @terrajobst , on whose feedback this list is primarily based, and who can correct me and provide clarification in case I misspoke.

TomFinley:
[...]

PR #2047, opened 2019-01-07
Modify API for advanced settings. (FastTree,  RandomForest)

### LINKED ISSUE #2019: FastTree learning rate in advanced settings does not override the default value
These lines in the BoostingFastTreeTrainerBase ctor reset it to the value specified directly (or the default value):

            if (Args.LearningRates != learningRate)
            {
                using (var ch = Host.Start($"Setting learning rate to: {learningRate} as supplied in the direct arguments."))
                    Args.LearningRates = learningRate;
            }

(Reported in #1983 .)

abgoswam:
Towards #1798 . Also fixes #2019 

This PR addresses **only** the  FastTree and RandomForest algos.  For the other remaining components, we will have separate PRs.

The following 3 changes have been made:

1. Two distinct constructors/extension methods, one for simple arguments and the other for advanced options
2. Rename `Arguments` to `Options`
3. Pass `Options`  objects as arguments instead of `Action` delegate
4. Added Help text as the XML comment of `Options`
5. Rename `Options` objects  as `options`  (instead of `args` or `advancedSettings` used so far)


sfilipi:
>             public bool UnbalancedSets = false;

Might be good to put the Help text as the XML comment of the arguments. We didn't do it on the previous swap because there was half a thought for the Args classes to go away and get substituted with something else, like direct fields. 

---
Refers to: src/Microsoft.ML.FastTree/FastTreeArguments.cs:31 in b255eb3. [](commit_id = b255eb3e7b9bdb8f0fe481bbadd1cec94d724538, deletion_comment = False)

src/Microsoft.ML.FastTree/FastTreeRegression.cs
```diff
@@ -62,24 +62,39 @@ public sealed partial class FastTreeRegressionTrainer
...
+        /// </summary>
+        /// <param name="env">The private instance of <see cref="IHostEnvironment"/>.</param>
+        /// <param name="labelColumn">The name of the label column.</param>
+        /// <param name="featureColumn">The name of the feature column.</param>
+        /// <param name="weightColumn">The name for the column containing the initial weight.</param>
+        /// <param name="advancedSettings">Advanced arguments to the algorithm.</param>
+        public FastTreeRegressionTrainer(IHostEnvironment env,
+            string labelColumn,
+            string featureColumn,
+            string weightColumn,
```
artidoro: These fields are also present in the Options object I believe. I think we should not have them as arguments in the constructor to eliminate the risk of them being set in two different places.

artidoro:
@TomFinley @abgoswam 
When inspecting the Options class for FastTreeRegressionTrainer, I see that it contains the following field: `IDataView TrainData`

It seems that this could lead to confusion, as we don't set training data in the arguments. Instead we pass it to the Fit() method. Correct me if I am wrong, but I think that this field is used by the entrypoints. How would we go about removing it from the Options class then?

src/Microsoft.ML.FastTree/BoostingFastTree.cs
```diff
@@ -40,6 +39,16 @@ protected BoostingFastTreeTrainerBase(IHostEnvironment env, TArgs args, SchemaSh
             }
         }
 
+        protected BoostingFastTreeTrainerBase(IHostEnvironment env,
+            SchemaShape.Column label,
+            string featureColumn,
+            string weightColumn,
+            string groupIdColumn,
+            TArgs advancedSettings)
```
yaeldekel: >TArgs advancedSettings [](start = 12, length = 22)

Hi @Abhishek Goswami, thanks for looking into this issue. Your change will also fix the problem described in issue #2019 ([https://github.com/dotnet/machinelearning/issues/2019](https://github.com/dotnet/machinelearning/issues/2019)).

yaeldekel:
>                     Args.LearningRates = learningRate;

This doesn't need to be inside the "if" anymore, since the base ctor does not change the value of Args.LearningRates. Reporting it in a channel is also not needed.

---
Refers to: src/Microsoft.ML.FastTree/BoostingFastTree.cs:38 in b255eb3. [](commit_id = b255eb3e7b9bdb8f0fe481bbadd1cec94d724538, deletion_comment = False)

abgoswam:
The observation makes sense.  We should probably create a separate issue for this, so we investigate  why we have `IDataView TrainData`  inside `Options`,  usages etc. 

I would keep it outside the purview of this PR, because it might have side-effects which I do not want to handle in this PR.

---
In reply to: [451888937](https://github.com/dotnet/machinelearning/pull/2047#issuecomment-451888937) [](ancestors = 451888937)

src/Microsoft.ML.FastTree/BoostingFastTree.cs
abgoswam: Thanks for the info.  Will reference #2019 in my PR  so it gets closed upon merging this PR

---
In reply to: [245743197](https://github.com/dotnet/machinelearning/pull/2047#discussion_r245743197) [](ancestors = 245743197)

src/Microsoft.ML.FastTree/FastTreeRegression.cs
abgoswam: Thanks for pointing this out.  I have updated the API signature to not have these.

---
In reply to: [245607972](https://github.com/dotnet/machinelearning/pull/2047#discussion_r245607972) [](ancestors = 245607972,245603313)

abgoswam:
>             public bool UnbalancedSets = false;

fixed. 

---
In reply to: [451848307](https://github.com/dotnet/machinelearning/pull/2047#issuecomment-451848307) [](ancestors = 451848307)

---
Refers to: src/Microsoft.ML.FastTree/FastTreeArguments.cs:31 in b255eb3. [](commit_id = b255eb3e7b9bdb8f0fe481bbadd1cec94d724538, deletion_comment = False)

abgoswam:
>                     Args.LearningRates = learningRate;

fixed

---
In reply to: [452024015](https://github.com/dotnet/machinelearning/pull/2047#issuecomment-452024015) [](ancestors = 452024015)

---
Refers to: src/Microsoft.ML.FastTree/BoostingFastTree.cs:38 in b255eb3. [](commit_id = b255eb3e7b9bdb8f0fe481bbadd1cec94d724538, deletion_comment = False)

src/Microsoft.ML.FastTree/FastTreeArguments.cs
```diff
@@ -24,8 +24,11 @@ internal interface IFastTreeTrainerFactory : IComponentFactory<ITrainer>
     public sealed partial class FastTreeBinaryClassificationTrainer
     {
         [TlcModule.Component(Name = LoadNameValue, FriendlyName = UserNameValue, Desc = Summary)]
-        public sealed class Arguments : BoostedTreeArgs, IFastTreeTrainerFactory
+        public sealed class Options : BoostedTreeArgs, IFastTreeTrainerFactory
         {
+            /// <summary>
+            /// Should we use derivatives optimized for unbalanced sets?
+            /// </summary>
```
artidoro: Maybe for a summary description of an argument, it would be better to have a declarative sentence instead of a question. Could you change this to a declarative sentence?

src/Microsoft.ML.FastTree/FastTreeArguments.cs
```diff
@@ -333,72 +447,130 @@ public abstract class BoostedTreeArgs : TreeArgs
         // REVIEW: TLC FR likes to call it bestStepRegressionTrees which might be more appropriate.
         //Use the second derivative for split gains (not just outputs). Use MaxTreeOutput to "clip" cases where the second derivative is too close to zero.
         //Turning BSR on makes larger steps in initial stages and converges to better results with fewer trees (though in the end, it asymptotes to the same results).
+        /// <summary>
+        /// Use best regression step trees?
+        /// </summary>
```
artidoro: Same here if you could make this in a declarative sentence that would be better I think.

src/Microsoft.ML.FastTree/FastTreeArguments.cs
```diff
@@ -333,72 +447,130 @@ public abstract class BoostedTreeArgs : TreeArgs
...
+        /// The learning rate.
+        /// </summary>
         [Argument(ArgumentType.LastOccurenceWins, HelpText = "The learning rate", ShortName = "lr", SortOrder = 4)]
         [TGUI(Label = "Learning Rate", SuggestedSweeps = "0.025-0.4;log")]
         [TlcModule.SweepableFloatParamAttribute("LearningRates", 0.025f, 0.4f, isLogScale: true)]
         public Double LearningRates = Defaults.LearningRates;
 
+        /// <summary>
+        /// Shrinkage.
+        /// </summary>
```
artidoro: Would be great to have more than just a single word to describe this. Maybe on the tlc documentation website there is more information on this parameter.

src/Microsoft.ML.FastTree/FastTreeClassification.cs
```diff
@@ -127,26 +127,24 @@ public sealed partial class FastTreeBinaryClassificationTrainer :
...
+            double learningRate = Defaults.LearningRates)
+            : base(env, TrainerUtils.MakeBoolScalarLabel(labelColumn), featureColumn, weightColumn, null, numLeaves, numTrees, minDatapointsInLeaves, learningRate)
         {
             // Set the sigmoid parameter to the 2 * learning rate, for traditional FastTreeClassification loss
             _sigmoidParameter = 2.0 * Args.LearningRates;
         }
 
         /// <summary>
-        /// Initializes a new instance of <see cref="FastTreeBinaryClassificationTrainer"/> by using the legacy <see cref="Arguments"/> class.
+        /// Initializes a new instance of <see cref="FastTreeBinaryClassificationTrainer"/> by using the legacy <see cref="Options"/> class.
```
artidoro: >legacy [](start = 105, length = 6)

We need to change the comment.

src/Microsoft.ML.FastTree/FastTreeRanking.cs
```diff
@@ -80,17 +79,16 @@ public sealed partial class FastTreeRankingTrainer
...
-            : base(env, TrainerUtils.MakeR4ScalarColumn(labelColumn), featureColumn, weightColumn, groupIdColumn, numLeaves, numTrees, minDatapointsInLeaves, learningRate, advancedSettings)
+            double learningRate = Defaults.LearningRates)
+            : base(env, TrainerUtils.MakeR4ScalarColumn(labelColumn), featureColumn, weightColumn, groupIdColumn, numLeaves, numTrees, minDatapointsInLeaves, learningRate)
         {
             Host.CheckNonEmpty(groupIdColumn, nameof(groupIdColumn));
         }
 
         /// <summary>
-        /// Initializes a new instance of <see cref="FastTreeRankingTrainer"/> by using the legacy <see cref="Arguments"/> class.
+        /// Initializes a new instance of <see cref="FastTreeRankingTrainer"/> by using the legacy <see cref="Options"/> class.
```
artidoro: >legacy [](start = 92, length = 6)

Same here, this is not legacy now.

src/Microsoft.ML.FastTree/FastTreeRegression.cs
```diff
@@ -62,24 +62,22 @@ public sealed partial class FastTreeRegressionTrainer
...
-            Action<Arguments> advancedSettings = null)
-            : base(env, TrainerUtils.MakeR4ScalarColumn(labelColumn), featureColumn, weightColumn, null, numLeaves, numTrees, minDatapointsInLeaves, learningRate, advancedSettings)
+            double learningRate = Defaults.LearningRates)
+            : base(env, TrainerUtils.MakeR4ScalarColumn(labelColumn), featureColumn, weightColumn, null, numLeaves, numTrees, minDatapointsInLeaves, learningRate)
         {
         }
 
         /// <summary>
-        /// Initializes a new instance of <see cref="FastTreeRegressionTrainer"/> by using the legacy <see cref="Arguments"/> class.
+        /// Initializes a new instance of <see cref="FastTreeRegressionTrainer"/> by using the legacy <see cref="Options"/> class.
```
artidoro: >legacy [](start = 95, length = 6)

Not legacy.

artidoro:
In order to distinguish between the two constructors, would be great to add in summary documentation of the one that takes Options a few words explaining that this is the complete set of arguments including advanced settings. Something like: "Includes advanced options" (maybe use parameters, or settings instead of options).

src/Microsoft.ML.FastTree/FastTreeClassification.cs
```diff
@@ -127,26 +127,24 @@ public sealed partial class FastTreeBinaryClassificationTrainer :
...
         /// </summary>
-        internal FastTreeBinaryClassificationTrainer(IHostEnvironment env, Arguments args)
+        public FastTreeBinaryClassificationTrainer(IHostEnvironment env, Options args)
```
artidoro: >args [](start = 81, length = 4)

Since we are renaming the Arguments class to Options, I think we should also change the parameter name to something more in line with this change. Maybe "options", or "settings". Maybe good idea to run this through Tom, or someone else to check.

src/Microsoft.ML.FastTree/FastTreeTweedie.cs
```diff
@@ -78,9 +76,9 @@ public sealed partial class FastTreeTweedieTrainer
         }
 
         /// <summary>
-        /// Initializes a new instance of <see cref="FastTreeTweedieTrainer"/> by using the legacy <see cref="Arguments"/> class.
+        /// Initializes a new instance of <see cref="FastTreeTweedieTrainer"/> by using the legacy <see cref="Options"/> class.
```
artidoro: >legacy [](start = 92, length = 6)

Change legacy.

src/Microsoft.ML.FastTree/RandomForestClassification.cs
```diff
@@ -145,26 +145,24 @@ public sealed class Arguments : FastForestArgumentsBase
...
+            double learningRate = Defaults.LearningRates)
+            : base(env, TrainerUtils.MakeBoolScalarLabel(labelColumn), featureColumn, weightColumn, null, numLeaves, numTrees, minDatapointsInLeaves, learningRate)
         {
             Host.CheckNonEmpty(labelColumn, nameof(labelColumn));
             Host.CheckNonEmpty(featureColumn, nameof(featureColumn));
         }
 
         /// <summary>
-        /// Initializes a new instance of <see cref="FastForestClassification"/> by using the legacy <see cref="Arguments"/> class.
+        /// Initializes a new instance of <see cref="FastForestClassification"/> by using the legacy <see cref="Options"/> class.
```
artidoro: >legacy [](start = 94, length = 6)

Remove legacy.

src/Microsoft.ML.FastTree/RandomForestRegression.cs
```diff
@@ -163,26 +163,24 @@ public sealed class Arguments : FastForestArgumentsBase
...
+            double learningRate = Defaults.LearningRates)
+            : base(env, TrainerUtils.MakeR4ScalarColumn(labelColumn), featureColumn, weightColumn, null, numLeaves, numTrees, minDatapointsInLeaves, learningRate)
         {
             Host.CheckNonEmpty(labelColumn, nameof(labelColumn));
             Host.CheckNonEmpty(featureColumn, nameof(featureColumn));
         }
 
         /// <summary>
-        /// Initializes a new instance of <see cref="FastForestRegression"/> by using the legacy <see cref="Arguments"/> class.
+        /// Initializes a new instance of <see cref="FastForestRegression"/> by using the legacy <see cref="Options"/> class.
```
artidoro: >legacy [](start = 90, length = 6)

Change legacy.

src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
```diff
@@ -24,20 +24,31 @@ public static class TreeExtensions
...
+            return new FastTreeRegressionTrainer(env, labelColumn, featureColumn, weights, numLeaves, numTrees, minDatapointsInLeaves, learningRate);
+        }
+
+        /// <summary>
+        /// Predict a target using a decision tree regression model trained with the <see cref="FastTreeRegressionTrainer"/>.
+        /// </summary>
+        /// <param name="ctx">The <see cref="RegressionContext"/>.</param>
+        /// <param name="advancedSettings">Algorithm advanced settings.</param>
+        public static FastTreeRegressionTrainer FastTree(this RegressionContext.RegressionTrainers ctx,
+            FastTreeRegressionTrainer.Options advancedSettings)
```
artidoro: >advancedSettings [](start = 46, length = 16)

I agree that we should name this something different from arguments. However, now it does not only contain advanced settings, it also contains the names of the input columns, and some basic settings. Would be great if this name matched the name we give to the Options parameter in the public constructor of the trainers.

artidoro:
Sounds good, this may be outside the scope of the PR. I suspect that it is easier than we think. Maybe it simply requires that field to be made internal in the new Options class. If that's the case, maybe we can still have that as part of this PR. I will ask Tom and update this thread. 

---
In reply to: [452110380](https://github.com/dotnet/machinelearning/pull/2047#issuecomment-452110380) [](ancestors = 452110380,451888937)

src/Microsoft.ML.StaticPipe/TreeTrainersStatic.cs
```diff
@@ -45,16 +44,56 @@ public static class TreeRegressionExtensions
...
+        /// <returns>The Score output column indicating the predicted value.</returns>
+        /// <example>
+        /// <format type="text/markdown">
+        /// <![CDATA[
+        ///  [!code-csharp[FastTree](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Static/FastTreeRegression.cs)]
+        /// ]]></format>
+        /// </example>
+        public static Scalar<float> FastTree(this RegressionContext.RegressionTrainers ctx,
+            Scalar<float> label, Vector<float> features, Scalar<float> weights,
+            FastTreeRegressionTrainer.Options advancedSettings,
```
artidoro: Same for static extensions, if we could rename the advancedSettings to something else it would be better I think.

artidoro: Is it possible to remove these fields?

src/Microsoft.ML.FastTree/FastTreeArguments.cs
abgoswam: i guess its one of those known terms in Gradient Boosted trees. do u have any suggestions ?

---
In reply to: [245964690](https://github.com/dotnet/machinelearning/pull/2047#discussion_r245964690) [](ancestors = 245964690)

artidoro:
From talking with Tom, it seems that this is a problem that will require a separate PR!

---
In reply to: [452273298](https://github.com/dotnet/machinelearning/pull/2047#issuecomment-452273298) [](ancestors = 452273298,452110380,451888937)

src/Microsoft.ML.FastTree/FastTreeArguments.cs
abgoswam: fixed

---
In reply to: [245963403](https://github.com/dotnet/machinelearning/pull/2047#discussion_r245963403) [](ancestors = 245963403)

src/Microsoft.ML.FastTree/FastTreeArguments.cs
abgoswam: fixed

---
In reply to: [245964319](https://github.com/dotnet/machinelearning/pull/2047#discussion_r245964319) [](ancestors = 245964319)

src/Microsoft.ML.FastTree/FastTreeClassification.cs
abgoswam: fixed

---
In reply to: [245965546](https://github.com/dotnet/machinelearning/pull/2047#discussion_r245965546) [](ancestors = 245965546)

src/Microsoft.ML.FastTree/FastTreeRanking.cs
abgoswam: fixed

---
In reply to: [245966171](https://github.com/dotnet/machinelearning/pull/2047#discussion_r245966171) [](ancestors = 245966171)

src/Microsoft.ML.FastTree/FastTreeRegression.cs
abgoswam: fixed

---
In reply to: [245966414](https://github.com/dotnet/machinelearning/pull/2047#discussion_r245966414) [](ancestors = 245966414)

src/Microsoft.ML.FastTree/FastTreeTweedie.cs
abgoswam: fixed

---
In reply to: [245968568](https://github.com/dotnet/machinelearning/pull/2047#discussion_r245968568) [](ancestors = 245968568)

src/Microsoft.ML.FastTree/RandomForestClassification.cs
abgoswam: fixed

---
In reply to: [245968731](https://github.com/dotnet/machinelearning/pull/2047#discussion_r245968731) [](ancestors = 245968731)

src/Microsoft.ML.FastTree/RandomForestRegression.cs
abgoswam: fixed

---
In reply to: [245968841](https://github.com/dotnet/machinelearning/pull/2047#discussion_r245968841) [](ancestors = 245968841)

abgoswam:
Currently the  summary documentation for  the constructor  taking `Options` argument reads  : ` Initializes a new instance of <see cref="FastTreeBinaryClassificationTrainer"/> by using the <see cref="Options"/> class.`
The `Options` summary reads  `Algorithm advanced settings`

I feel this is sufficient. Isn't it  ?  


---
In reply to: [452270462](https://github.com/dotnet/machinelearning/pull/2047#issuecomment-452270462) [](ancestors = 452270462)

TomFinley:
> @TomFinley @abgoswam
> When inspecting the Options class for FastTreeRegressionTrainer, I see that it contains the following field: `IDataView TrainData`
> 
> It seems that this could lead to confusion, as we don't set training data in the arguments. Instead we pass it to the Fit() method. Correct me if I am wrong, but I think that this field is used by the entrypoints. How would we go about removing it from the Options class then? #Closed

*Anything* relating to entry-points specifically should be `internal`, but we must do so carefully, since currently entry-points and command line processing relies on processing of fields which is currently done on *public* fields only, whereas it should be over public or internal fields. Once that is done, any fields that are relevant to entry-points only should be internal.

(This is a problem for *all* the settings objects, that they are exposing things that should not be in the public API in some cases. (Indeed entry-points will need a lot of work to make them work for estimators and transformers. Estimator graphs are sort of a "lightweight" estimator chain as they stand right now that performs absolutely no validation. But for now, anything relating to entry-points should just be hidden.)

src/Microsoft.ML.StaticPipe/TreeTrainersStatic.cs
abgoswam: @Tom.  

In the API which accepts the `Options` argument,  how do we handle the  arguments  `label`, `features`, `weights`  in the Static API ?

In the corresponding Dynamic API,  we were able to remove  the parameters  for `labelColumn`, `featureColumn` and `weightColumn`  because  these were already part of the `Options` class  


---
In reply to: [245972452](https://github.com/dotnet/machinelearning/pull/2047#discussion_r245972452) [](ancestors = 245972452)

TomFinley: You obviously cannot remove the parameters for these *here*. The options object just accepts a `string`. The entire point of the statically typed API is that someone can define a pipeline in a way that is type safe. So, no, you are not removing these parameters.

---
In reply to: [246165575](https://github.com/dotnet/machinelearning/pull/2047#discussion_r246165575) [](ancestors = 246165575,245972452)

src/Microsoft.ML.FastTree/FastTreeClassification.cs
```diff
@@ -127,27 +127,27 @@ public sealed partial class FastTreeBinaryClassificationTrainer :
...
 
         /// <summary>
-        /// Initializes a new instance of <see cref="FastTreeBinaryClassificationTrainer"/> by using the legacy <see cref="Arguments"/> class.
+        /// Initializes a new instance of <see cref="FastTreeBinaryClassificationTrainer"/> by using the <see cref="Options"/> class.
         /// </summary>
-        internal FastTreeBinaryClassificationTrainer(IHostEnvironment env, Arguments args)
-            : base(env, args, TrainerUtils.MakeBoolScalarLabel(args.LabelColumn))
+        /// <param name="env">The instance of <see cref="IHostEnvironment"/>.</param>
+        /// <param name="options">Algorithm advanced settings.</param>
+        public FastTreeBinaryClassificationTrainer(IHostEnvironment env, Options options)
```
TomFinley: >public FastTreeBinaryClassificationTrainer(IHostEnvironment env, Options options) [](start = 8, length = 81)

So, why add **two constructors**? Doesn't it suffice to just have the two overloads be on the `MLContext` extension methods? Shouldn't there just be *one* constructor? (Possibly an internal one, in which case you could avoid changing this altogether, except for the rename to options?)

src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
```diff
@@ -24,20 +24,31 @@ public static class TreeExtensions
...
             int numTrees = Defaults.NumTrees,
             int minDatapointsInLeaves = Defaults.MinDocumentsInLeaves,
-            double learningRate = Defaults.LearningRates,
-            Action<FastTreeRegressionTrainer.Arguments> advancedSettings = null)
+            double learningRate = Defaults.LearningRates)
         {
             Contracts.CheckValue(ctx, nameof(ctx));
             var env = CatalogUtils.GetEnvironment(ctx);
-            return new FastTreeRegressionTrainer(env, labelColumn, featureColumn, weights, numLeaves, numTrees, minDatapointsInLeaves, learningRate, advancedSettings);
+            return new FastTreeRegressionTrainer(env, labelColumn, featureColumn, weights, numLeaves, numTrees, minDatapointsInLeaves, learningRate);
```
TomFinley: >FastTreeRegressionTrainer(env, labelColumn, featureColumn, weights, numLeaves, numTrees, minDatapointsInLeaves, learningRate); [](start = 23, length = 126)

So, this is kind of what I mean. While these two extension methods are good, I might have preferred that we just have a *single* constructor (always taking `Options`), and we just keep the flat constructor as simple as possible (e.g., take only the `Options` class, at least publicly). If there are "N" extension methods on this (which we consider the premier public surface!) there don't have to be a corresponding "N" constructors. This could have easily (and indeed more cleanly) acheived with exactly one constructor (again possibly internal) on the underlying object that just took an `Options` object, simply. As it is we're making the underlying object considerably more complex, I think, than it really has to be.

src/Microsoft.ML.StaticPipe/TreeTrainersStatic.cs
abgoswam: Thanks Tom for the clarification.

---
In reply to: [246170641](https://github.com/dotnet/machinelearning/pull/2047#discussion_r246170641) [](ancestors = 246170641,246165575,245972452)

abgoswam:
Thanks Tom for the explanation.  I will create a separate issue for this.  

---
In reply to: [452460937](https://github.com/dotnet/machinelearning/pull/2047#issuecomment-452460937) [](ancestors = 452460937)

src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
abgoswam: Created issue #2100 to  track this. 

For now I have made the constructors internal.

---
In reply to: [246172562](https://github.com/dotnet/machinelearning/pull/2047#discussion_r246172562) [](ancestors = 246172562)

src/Microsoft.ML.FastTree/FastTreeClassification.cs
abgoswam: Created issue #2100 to  track this. 

For now I have made both the constructors internal.

---
In reply to: [246171294](https://github.com/dotnet/machinelearning/pull/2047#discussion_r246171294) [](ancestors = 246171294)

src/Microsoft.ML.FastTree/FastTreeArguments.cs
artidoro: Maybe "The shrinkage". But yes I looked it up and you are right, so not necessary.

---
In reply to: [246105587](https://github.com/dotnet/machinelearning/pull/2047#discussion_r246105587) [](ancestors = 246105587,245964690)

PR #2048, opened 2019-01-07
Cleanup the statistics usage API

### LINKED ISSUE #2010: LinearModelStatistics API usage cleanup
There are several issues with the usage of the [LinearModelStatistics ](https://github.com/dotnet/machinelearning/blob/25f3287036fd5956d3726a9dc7c510883eb608d4/src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs). 
Some of them are captured on the discussion of PR #1979. 

Another issue with this class as is, is its two constructors. 
[One of them](https://github.com/dotnet/machinelearning/blob/25f3287036fd5956d3726a9dc7c510883eb608d4/src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs#L95) initializes just the trainingExampleCount, deviance, nullDeviance. 
[The other one](https://github.com/dotnet/machinelearning/blob/25f3287036fd5956d3726a9dc7c510883eb608d4/src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs#L107) adds to it the bias and weights, so parameter specific stats can be calculated. 

The uncertain existance of the weights, bias makes all the methods of this class do work on the first batch of stats (trainingExampleCount, deviance, nullDeviance) than check for the existence of the weights and bias std, and  exit if this is missing, or continue to work.  See lines [199](https://github.com/dotnet/machinelearning/blob/25f3287036fd5956d3726a9dc7c510883eb608d4/src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs#L199), [222](https://github.com/dotnet/machinelearning/blob/25f3287036fd5956d3726a9dc7c510883eb608d4/src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs#L222), [238](https://github.com/dotnet/machinelearning/blob/25f3287036fd5956d3726a9dc7c510883eb608d4/src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs#L238); almost every method has its half-way out. 

The implementation should clearly be broken down, potentially hierarchically, all methods should be instance methods and whatever field/property they need from the ModelParams class that hosts them, should be passed down to the constructor.

sfilipi:
Resolves #2010 by breaking down the existing LinearModelStatistics into two classes: ModelStatistics and LinearParameterModelStatistics. 

The only reason why the two classes are hierarchical, is backwards compatibility, so we can load either one of them in the Statistics field that exists in [MulticlassLogisticRegressionModelParameters ](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.StandardLearners/Standard/LogisticRegression/MulticlassLogisticRegression.cs#L374)and [LinearBinaryModelParameters](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.StandardLearners/Standard/LinearModelParameters.cs#L410). 

If we don't want to worry about backwards compatibility, ModelStatistics and LinearParameterModelStatstics, could each be a property in the LinearBinaryModelParameters, the artifact of training LR. As far as MLR, ModelStatistics would replace the current LinearModelStatistics.


src/Microsoft.ML.Data/Model/ModelLoading.cs
```diff
@@ -134,8 +179,6 @@ public bool ContainsModel(string name)
                     return true;
             }
 
-            // TryLoadModelCore should rewind on failure.
-            Contracts.Assert(fp == ent.Stream.Position);
```
sfilipi: >Contracts.Assert(fp == ent.Stream.Position); [](start = 12, length = 44)

The first thing to look at on this PR is whether the changes on removing throwing for the case when the header length/body length of the expected signature, and the actual signature in the TryLoadModel is acceptable. 

I feel like it can be ok to not throw, since this is a Try-type method, but i know that this area of the codebase has been hashed pretty well in the past :)

If we are not ok with deleting this assert, and the ones below in TryLoadModelCore, and we also want to preserve backwards compatibility, I'll unwind the changes to have all the stats in one class. (and with that have methods like GetBais, or GetCoefficients exit when the std of the coefficients is not provided.. ) 

If we don't care about backwards compatibility, we can keep those asserts.

src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs
```diff
@@ -15,24 +15,29 @@
...
 
 namespace Microsoft.ML.Learners
 {
     /// <summary>
     /// Represents a coefficient statistics object.
     /// </summary>
-    public readonly struct CoefficientStatistics
+    public sealed class CoefficientStatistics
     {
         public readonly string Name;
```
sfilipi: >Name [](start = 31, length = 4)

@tomfinley,  I haven't addressed your comment about removing the name from here, and operating on index. 
Since this is user facing, i was thinking that names might be easier to interpret? 

Do you think it would be enough to modify the API in line 373, that currently takes the FeatureColumn, to instead take the slot names of the metadata? The user would be in charge of extracting the slot names, and passing them to the API.

test/Microsoft.ML.Tests/TrainerEstimators/LbfgsTests.cs
```diff
@@ -84,21 +88,66 @@ public void TestLogisticRegressionWithStats()
...
+            Assert.Equal(19, coefficients.Length);
 
-            foreach(var coefficient in coeffcients)
+            foreach(var coefficient in coefficients)
                 Assert.True(coefficient.StandardError < 1.0);
 
         }
+
+        [Fact]
+        public void TestLRWithStatsBackCompatibility()
```
sfilipi: >TestLRWithStatsBackCompatibility [](start = 20, length = 32)

if this PR is acceptable, add a test about LR backwards compatibility for model with stats saved.

TomFinley:
> The only reason why the two classes are hierarchical, is backwards compatibility, so we can load either one of them in the Statistics field that exists in [MulticlassLogisticRegressionModelParameters ](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.StandardLearners/Standard/LogisticRegression/MulticlassLogisticRegression.cs#L374)and [LinearBinaryModelParameters](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.StandardLearners/Standard/LinearModelParameters.cs#L410).

Hi @sfilipi , thanks for working on this. If the goal is backwards compatibility, like with the `IDataTransform` to `ITransformer` work, it is unclear to me that the structure of a *binary* format needs to be in any way reflected in the C# type system representing the result of that deserialization. Even if they need to share deserialization logic, I could imagine just simply changing this method to be a create method returning an `object` if I like, with one or the other being returned. (That if they actually need to share deserialization logic, which is not entirely clear to me.) Then for calling from `MulticlassLogisticRegressionModelParameters ` we cast to one, and in the other case the other. Is it possible?

I have not had the chance to study your change this in detail, but in other contexts when we've thought that somehow the binary format "requires" some sub-ideal class relationship, it has never quite been true yet, so my suspicion is that it is not true here either, maybe?

src/Microsoft.ML.StandardLearners/Standard/ModelStatistics.cs
TomFinley: But the names are not easier to interpret, because everywhere else we index a feature, we do so by name.

sfilipi: I think I don't fully understand your point. Happy to make the change, just would like to understand the problem. 

The feature "names" are already saved in the annotations/metadata of the feature column. Since that is done, why should we not leave it to it, rather than convert to indexes where the users would have to know that bias is the first weight, etc. 

---
In reply to: [246904367](https://github.com/dotnet/machinelearning/pull/2048#discussion_r246904367) [](ancestors = 246904367)

sfilipi:
I think it makes sense to leave them hierarchical as the second one, LinearModelParameterStatistics is really a superset of the ModelStatisticsBase. 

---
In reply to: [452845784](https://github.com/dotnet/machinelearning/pull/2048#issuecomment-452845784) [](ancestors = 452845784)

test/BaselineOutput/Common/EntryPoints/ensemble-summary.txt
```diff
@@ -68,16 +68,16 @@ AIC:                       	114.1969
...
-Features.bare_nuclei	3.16846   	0.4579377 	6.918975  	0 ***
-Features.thickness	2.143086  	0.4306555 	4.976335  	6.556511E-07 ***
-Features.uniform_shape	1.711214  	0.4222687 	4.05243   	5.072355E-05 ***
-Features.uniform_size	1.494185  	0.4099987 	3.644366  	0.0002681017 ***
-Features.normal_nucleoli	1.519048  	0.4728626 	3.212452  	0.00131619 **
-Features.bland_chromatin	1.386847  	0.4451247 	3.115636  	0.001835644 **
-Features.adhesion	1.383187  	0.4832917 	2.862014  	0.004209697 **
-Features.epit_size	0.8832003 	0.4570503 	1.932392  	0.05331099 .
-Cat.1          	-0.8226236	0.3383799 	-2.431065 	0.01505452 *
+-4.860323 	0.7128119 	-6.818521 	0 ***
```
Ivanidzo4ka: it used to be 5 columns, now it's only 4, why?

src/Microsoft.ML.StandardLearners/Standard/LinearModelParameters.cs
```diff
@@ -455,9 +453,19 @@ private LinearBinaryModelParameters(IHostEnvironment env, ModelLoadContext ctx)
...
             // *** Binary format ***
             // (Base class)
-            // LinearModelStatistics: model statistics (optional, in a separate stream)
-
-            ctx.LoadModelOrNull<LinearModelStatistics, SignatureLoadModel>(Host, out _stats, ModelStatsSubModelFilename);
+            // LinearModelParameterStatistics: model statistics (optional, in a separate stream)
+            try
+            {
+                LinearModelParameterStatistics stats;
+                ctx.LoadModelOrNull<LinearModelParameterStatistics, SignatureLoadModel>(Host, out stats, ModelStatsSubModelFilename);
```
yaeldekel: >LoadModelOrNull [](start = 20, length = 15)

Does this throw on failure, or just return null?

test/BaselineOutput/Common/EntryPoints/ensemble-summary.txt
```diff
@@ -20,16 +20,16 @@ AIC:                       	118.2943
...
-Features.thickness	2.353567  	0.4267568 	5.515007  	5.960464E-08 ***
-Features.bare_nuclei	2.435889  	0.451504  	5.395056  	5.960464E-08 ***
-Features.uniform_shape	1.944249  	0.4137097 	4.699549  	2.622604E-06 ***
-Features.uniform_size	1.786538  	0.4204066 	4.249547  	2.145767E-05 ***
-Features.bland_chromatin	1.611417  	0.4605175 	3.499144  	0.0004668236 ***
-Features.adhesion	1.38072   	0.4821554 	2.863641  	0.00418818 **
-Features.normal_nucleoli	1.345759  	0.4784136 	2.812962  	0.00490886 **
-Features.epit_size	1.083109  	0.4566914 	2.371643  	0.0177092 *
-Cat.1          	-0.7715381	0.3420695 	-2.255501 	0.02410179 *
+-5.120674 	0.6998186 	-7.317145 	0 ***
```
yaeldekel: Did we lose the slot names here?

test/BaselineOutput/Common/EntryPoints/ensemble-summary.txt
sfilipi: Tom suggested we remove the name from the CoefficientStatistics class. Users should use the index in the array to map to the feature vector. 

Not a fan of the change given it reduces the readability of those summary files. 

---
In reply to: [264870766](https://github.com/dotnet/machinelearning/pull/2048#discussion_r264870766) [](ancestors = 264870766)

src/Microsoft.ML.StandardLearners/Standard/LinearModelParameters.cs
sfilipi: throw if the signature supplied is not correct. 

---
In reply to: [264870883](https://github.com/dotnet/machinelearning/pull/2048#discussion_r264870883) [](ancestors = 264870883)

test/BaselineOutput/Common/EntryPoints/ensemble-summary.txt
sfilipi: yes as suggested here.. 

after seeing the summary files change, i think we should get the back. @tfinley@gmail.com 
[https://github.com/dotnet/machinelearning/pull/2048#discussion_r245563884](https://github.com/dotnet/machinelearning/pull/2048#discussion_r245563884)


---
In reply to: [264872767](https://github.com/dotnet/machinelearning/pull/2048#discussion_r264872767) [](ancestors = 264872767)

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
```diff
@@ -16,249 +16,343 @@
...
         }
     }
 
-    // REVIEW: Make this class a loadable class and implement ICanSaveModel.
     // REVIEW: Reconcile with the stats in OLS learner.
     /// <summary>
     /// The statistics for linear predictor.
     /// </summary>
-    public sealed class LinearModelStatistics : ICanSaveModel
+    public class ModelStatisticsBase : ICanSaveModel
```
TomFinley: >public class [](start = 4, length = 12)

Is this intended to be an abstract class?

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
```diff
@@ -16,249 +16,343 @@
...
-                modelSignature: "LMODSTAT",
-                verWrittenCur: 0x00010001, // Initial
-                verReadableCur: 0x00010001,
-                verWeCanReadBack: 0x00010001,
-                loaderSignature: LoaderSignature,
-                loaderAssemblyName: typeof(LinearModelStatistics).Assembly.FullName);
-        }
-
-        private readonly IHostEnvironment _env;
+        protected IHostEnvironment Env;
```
TomFinley: >protected [](start = 8, length = 9)

Protected members are still part of the public surface. `private protected` please.

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
sfilipi: no, i still need to instantiate it.

---
In reply to: [265254412](https://github.com/dotnet/machinelearning/pull/2048#discussion_r265254412) [](ancestors = 265254412)

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
```diff
@@ -284,100 +378,99 @@ public static bool TryGetBiasStatistics(LinearModelStatistics stats, float bias,
...
+            Contracts.AssertValue(Env);
 
-            var weights = parent.Weights as IReadOnlyList<float>;
-            _env.Assert(_paramCount == 1 || weights != null);
-            _env.Assert(_coeffStdError.Value.Length == weights.Count + 1);
+            Env.Assert(_coeffStdError.Length == _weights.Length + 1);
 
             var names = default(VBuffer<ReadOnlyMemory<char>>);
 
             featureColumn.Annotations.GetValue(AnnotationUtils.Kinds.SlotNames, ref names);
```
sfilipi: remove

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
```diff
@@ -387,64 +480,49 @@ internal void SaveText(TextWriter writer, LinearBinaryModelParameters parent, Da
...
                 return;
+            int index = 0;
+            resultCollection.Add(new KeyValuePair<string, object>(
+                   (index++).ToString(),
+                   new float[] { biasStats.Estimate, biasStats.StandardError, biasStats.ZScore, biasStats.PValue }));
 
             foreach (var coeffStat in coeffStats)
             {
                 resultCollection.Add(new KeyValuePair<string, object>(
-                    coeffStat.Name,
```
TomFinley: >coeffStat.Name [](start = 20, length = 14)

You'll note that this is where the name used to be written to the summary. Instead of just removing this, we should have exploited the metadata on the `DataViewSchema.Column featureColumn` parameter to this method.

test/BaselineOutput/Common/EntryPoints/ensemble-summary.txt
TomFinley: I think we settled this offline, but just to update the thread on that discussion.

So, let's begin by observing that the problem of producing summaries that somehow involve feature names or in the case of multiclass lable names or other information from the schema is absolutely not something somehow unique to this statistics class. Every time we write a summary (as here), or export to code, or save as text, or INI, or whatever, we usually incorporate the feature names as part of that export feature. Let's also observe that *no-where else* in the code do we duplicate the information already in the schema as part of the saved model, because we don't need to. That information is just part of the `DataViewSchema`. Clearly, there is a way for this to work by not duplicating data in every conceivable location. :)

The way in which this happens is that when you save a summary, this works via this interface.

[https://github.com/dotnet/machinelearning/blob/b455174cb848be054f8a800dbe76e7deb28c699e/src/Microsoft.ML.Data/Dirty/PredictorInterfaces.cs#L98-L101](https://github.com/dotnet/machinelearning/blob/b455174cb848be054f8a800dbe76e7deb28c699e/src/Microsoft.ML.Data/Dirty/PredictorInterfaces.cs#L98-L101)

or alternately depending on the situation via this interface:

[https://github.com/dotnet/machinelearning/blob/b455174cb848be054f8a800dbe76e7deb28c699e/src/Microsoft.ML.Data/Dirty/PredictorInterfaces.cs#L109-L115](https://github.com/dotnet/machinelearning/blob/b455174cb848be054f8a800dbe76e7deb28c699e/src/Microsoft.ML.Data/Dirty/PredictorInterfaces.cs#L109-L115)

Etc. etc., lots of similar interfaces in that file. Note that `RoleMappedSchema schema`! ***That's really important.*** That is the mechanism by how when we save a summary, we get access to things like feature names, label names for multiclass, or any other column annotation information we may want access to! And exploiting this is how this code used to work. Now, it worked in a strange three step process: it would extract the slots names, *then* store them in this side structure for no reason, *then* the summary writing would take the names that were unnecessarily stored. But just because I'm saying "let's not store this unnecessary information" doesn't mean I'm saying we can't have the slot names. The slot names are literally *right there*.

Should we have names in the output report? Yes. Is the right way to do it by storing the names with the structure, thereby producing needless duplication of data that is already in the model, bloating them unnecessarily? Absolutely not! The right way is to take advantage of the presence of this schema, which we *absolutely do have access to*, as we see above.

---
In reply to: [264902026](https://github.com/dotnet/machinelearning/pull/2048#discussion_r264902026) [](ancestors = 264902026,264870766)

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
TomFinley: Who instantiates it?

---
In reply to: [265259755](https://github.com/dotnet/machinelearning/pull/2048#discussion_r265259755) [](ancestors = 265259755,265254412)

test/BaselineOutput/Common/EntryPoints/ensemble-summary.txt
TomFinley: See other comment.

---
In reply to: [264903719](https://github.com/dotnet/machinelearning/pull/2048#discussion_r264903719) [](ancestors = 264903719,264872767)

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
```diff
@@ -16,249 +16,343 @@
...
+            writer.WriteLine("Count of training examples:\t{0}", TrainingExampleCount);
+            writer.WriteLine("Residual Deviance:         \t{0}", Deviance);
+            writer.WriteLine("Null Deviance:             \t{0}", NullDeviance);
+            writer.WriteLine("AIC:                       \t{0}", 2 * ParametersCount + Deviance);
+        }
+
+        /// <summary>
+        /// Support method for linear models and <see cref="ICanGetSummaryInKeyValuePairs"/>.
+        /// </summary>
+        internal virtual void SaveSummaryInKeyValuePairs(DataViewSchema.Column featureColumn, int paramCountCap, List<KeyValuePair<string, object>> resultCollection)
```
TomFinley: >featureColumn [](start = 79, length = 13)

Note this `featureColumn` parameter. The method `featureColumn.GetSlotNames` may be useful to get the feature names for expanding the report.

test/BaselineOutput/Common/EntryPoints/ensemble-summary.txt
sfilipi: Thanks for the info and the suggestion Tom. 

As i was working to re-establish the feature names in the summaries, I noticed that the API to get the statistics, see  current ModelStatistics.cs file, line 425,  

`GetWeightsCoefficientStatistics(DataViewSchema.Column featureColumn, int paramCountCap)`

It has the `paramCountCap`  argument indicating how many TOP coefficients to return.  We are ordering those coefficient statistics objects by Zscores, and effectively losing the mapping between the indexes in this array and the indexes in the weights array. This is why the duplication of the feature name existed inside the structure. 

and keeping the index seems like a better solution than snapshotting the indexes before reordering in yet another data structure. 

I kept the index of the features together with the coefficient statistics, for space efficiency, but for this use case might as well revert back to the name. Let me know if I am missing anything. 

---
In reply to: [265277049](https://github.com/dotnet/machinelearning/pull/2048#discussion_r265277049) [](ancestors = 265277049,264902026,264870766)

TomFinley: > It has the `paramCountCap` argument indicating how many TOP coefficients to return. We are ordering those coefficient statistics objects by Zscores, and effectively losing the mapping between the indexes in this array and the indexes in the weights array.

Sorry, what?

sfilipi: Yes, I added that to preserve the original feature slot index, and I am using it. 

In the comment i explained why i had to use it. 

---
In reply to: [265673248](https://github.com/dotnet/machinelearning/pull/2048#discussion_r265673248) [](ancestors = 265673248)

TomFinley: Ah OK. You were still talking about "might as well revert back to the name" which suggested you still thought having the name around was a good idea.

test/Microsoft.ML.Tests/TrainerEstimators/LbfgsTests.cs
```diff
@@ -7,6 +7,9 @@
 using Microsoft.ML.Data;
 using Microsoft.ML.Trainers;
 using Xunit;
+using System.IO;
+using Microsoft.ML.Model;
+using System;
```
TomFinley: Not a big deal, just curious what this `.Model` is for.

test/BaselineOutput/Common/EntryPoints/ensemble-summary.txt
```diff
@@ -68,7 +68,7 @@ AIC:                       	114.1969
 
 Coefficients statistics:
 Coefficient    	Estimate	Std. Error	z value  	Pr(>|z|)
-(Bias)         	-4.860323 	0.7128119 	-6.818521 	0 ***
+(Bias)	-4.860323 	0.7128119 	-6.818521 	0 ***
 Features.bare_nuclei	3.16846   	0.4579377 	6.918975  	0 ***
```
TomFinley: >Features.bare_nuclei [](start = 0, length = 20)

Yay they're back. :)

sfilipi:
@yaeldekel @Ivanidzo4ka when you get a chance, this needs a second set of eyes. Your comments are addressed.

test/Microsoft.ML.Tests/TrainerEstimators/LbfgsTests.cs
```diff
@@ -53,23 +56,27 @@ public void TestEstimatorPoissonRegression()
...
-            Assert.Equal(0f, zScore);
+            Assert.Null(stats2);
+
+            Done();
         }
 
+
         [Fact]
-        public void TestLogisticRegressionWithStats()
+        public void TestLRWithStats()
```
sfilipi: test failing in CI definitions making use of Intrinsics.

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/MulticlassLogisticRegression.cs
```diff
@@ -589,7 +594,7 @@ private protected override void SaveCore(ModelSaveContext ctx)
             // float[]: non-zero weights
             // bool: whether label names are present
             // int[]: Id of label names (optional, in a separate stream)
-            // LinearModelStatistics: model statistics (optional, in a separate stream)
+            // LinearModelParameterStatistics: model statistics (optional, in a separate stream)
```
wschin: >LinearModelParameterStatistics [](start = 15, length = 30)

ModelStatisticsBase?

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
```diff
@@ -16,249 +16,345 @@
...
             StandardError = stdError;
             ZScore = zScore;
             PValue = pValue;
         }
     }
 
-    // REVIEW: Make this class a loadable class and implement ICanSaveModel.
     // REVIEW: Reconcile with the stats in OLS learner.
     /// <summary>
     /// The statistics for linear predictor.
```
wschin: >statistics for linear predictor [](start = 12, length = 31)

Is `ModelStatisticsBase` linear model specific?

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
wschin: Agree with Tom.

---
In reply to: [265277153](https://github.com/dotnet/machinelearning/pull/2048#discussion_r265277153) [](ancestors = 265277153,265259755,265254412)

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
```diff
@@ -284,100 +380,124 @@ public static bool TryGetBiasStatistics(LinearModelStatistics stats, float bias,
...
+
+            return featureNames.ToArray();
+
         }
 
         /// <summary>
         /// Gets the coefficient statistics as an object.
         /// </summary>
-        public CoefficientStatistics[] GetCoefficientStatistics(LinearBinaryModelParameters parent, DataViewSchema.Column featureColumn, int paramCountCap)
+        public CoefficientStatistics[] GetWeightsCoefficientStatistics(int paramCountCap)
```
wschin: >CoefficientStatistics[] [](start = 15, length = 23)

IReadOnlyList<CoefficientStatistics>

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
```diff
@@ -16,249 +16,345 @@
...
+            var pValue = 1.0f - (float)ProbabilityFunctions.Erf(Math.Abs(zScore / sqrt2));
+
+            //int feature index, float estimate, float stdError, float zScore, float pValue
+            return new CoefficientStatistics(0, bias, stdError, zScore, pValue);
         }
 
-        private static void GetUnorderedCoefficientStatistics(LinearModelStatistics stats, in VBuffer<float> weights, in VBuffer<ReadOnlyMemory<char>> names,
+        // sefilipi: do we want to offer this? Do we want to offer setting the weights?
+        public CoefficientStatistics GetBiasStatistics() => GetBiasStatisticsForValue(_bias);
+
```
wschin: Doc string please.

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
```diff
@@ -16,249 +16,345 @@
...
+                ctx.Writer.WriteIntsNoCount(_coeffStdError.GetIndices());
         }
 
         /// <summary>
-        /// Computes the standart deviation, Z-Score and p-Value.
+        /// Computes the standart deviation, Z-Score and p-Value for the value being passed as the bias.
         /// </summary>
-        public static bool TryGetBiasStatistics(LinearModelStatistics stats, float bias, out float stdError, out float zScore, out float pValue)
+        public CoefficientStatistics GetBiasStatisticsForValue(float bias)
         {
```
wschin: Is it necessary to users? Could we internalize it?

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
```diff
@@ -16,249 +16,345 @@
...
     /// </summary>
-    public readonly struct CoefficientStatistics
+    public sealed class CoefficientStatistics
     {
-        public readonly string Name;
         public readonly float Estimate;
         public readonly float StandardError;
         public readonly float ZScore;
         public readonly float PValue;
+        public readonly int FeatureIndex;
```
wschin: >FeatureIndex [](start = 28, length = 12)

Maybe we can call it `Index` and add a doc string saying that it's a feature index?
```
var coefficientStat = new ...
coefficientStat.FeatureIndex // Coefficient and Feature carry similar information and this `index` is this coefficient index.
coefficientStat.Index // looks like it's clear enough?
```

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/MulticlassLogisticRegression.cs
```diff
@@ -377,7 +377,7 @@ private static VersionInfo GetVersionInfo()
 
         private readonly float[] _biases;
         private readonly VBuffer<float>[] _weights;
-        private readonly LinearModelStatistics _stats;
+        public readonly ModelStatisticsBase Statistics;
```
wschin: Can this really be `Base`? What if use want to access statistics of the underlying linear function? I don't feel casting is a good idea.

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
sfilipi: it has current external usage, should keep it. I'll add a functional test too. 

---
In reply to: [266657218](https://github.com/dotnet/machinelearning/pull/2048#discussion_r266657218) [](ancestors = 266657218)

src/Microsoft.ML.StandardTrainers/Standard/ModelStatistics.cs
sfilipi: Yes, and LR, MLR are the only classes that set it. You got me thinking: I should derive another class from LinearModelParams and have the stats object only there. LRModelParams. 
I'll log an issue and do that on a separate PR. 

---
In reply to: [266648417](https://github.com/dotnet/machinelearning/pull/2048#discussion_r266648417) [](ancestors = 266648417)

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/MulticlassLogisticRegression.cs
sfilipi: we are not casting anywhere in this file. Are you referring to the other one?

---
In reply to: [266660951](https://github.com/dotnet/machinelearning/pull/2048#discussion_r266660951) [](ancestors = 266660951)

test/Microsoft.ML.Tests/TrainerEstimators/LbfgsTests.cs
sfilipi: Might go away after rebase. It is in there for IPredictorWithFeatureWeights and ModelFileUtils. 
I think ModelFileUtils got moved. Not sure about IPredictorWithFeatureWeights .

---
In reply to: [265696171](https://github.com/dotnet/machinelearning/pull/2048#discussion_r265696171) [](ancestors = 265696171)

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/MulticlassLogisticRegression.cs
wschin: They happen outside this file. We do `ModelStatisticsBase`  as `LinearModelStatisticsBase` somewhere.

sfilipi:
close/reopen to see if the merge button wakes up.

test/Microsoft.ML.TestFramework/Attributes/LessThanNetCore30OrNotNetCoreAndX64FactAttribute.cs
```diff
@@ -20,4 +20,20 @@ protected override bool IsEnvironmentSupported()
             return Environment.Is64BitProcess && AppDomain.CurrentDomain.GetData("FX_PRODUCT_VERSION") == null;
         }
     }
+
+    /// <summary>
+    /// A fact for tests requiring x64 environment and either .NET Core version lower than 3.0 or framework other than .NET Core.
+    /// </summary>
+    public sealed class LessThanNetCore30OrNotNetCore : EnvironmentSpecificFactAttribute
```
eerhardt: (nit) - Attribute classes should have the `Attribute` suffix in their name.

Issue #1860, opened 2018-12-11
Extract IDataView into its own assembly and NuGet package


eerhardt:
IDataView is a very flexible, efficient way of describing tabular data (columns and rows) in a read-only manner.

https://github.com/dotnet/machinelearning/blob/master/docs/code/IDataViewDesignPrinciples.md

At its heart are 2 key concepts:
- Schema (describing the columns)
- Cursoring (how to read the rows of data)

It has other capabilities that I won’t enumerate here, the above link describes them in more detail.

IDataView is very useful as an abstraction for tabular data that will allow users to pass data between two independent libraries.

For example, ML.NET is able to both consume and produce IDataView instances. Say there was a .NET library for Apache Arrow. If the Arrow .NET data type implements IDataView, the Apache Arrow data can be passed directly into ML.NET without having to copy it into a format that ML.NET consumes.

Another example is: say we had a visualization/graphing/plotting library in .NET that could consume data using IDataView. Then we could take data that was produced by ML.NET, or Apache Arrow, and feed it directly into the graphing library. There would be no need to copy, or change the shape of the data at all. And there is no need for this graphing library to know anything about ML.NET or Apache Arrow.

In my mind, we can use IDataView in a similar manner to what [OData](https://www.odata.org/) was promised to be: An exchange format which allows producers and consumers of data to communicate in a standardized way. (Although, OData has more capabilities such as filtering, sorting, updating data, etc which I am not proposing we add to IDataView. I was just using it as an analogy.)

/cc @TomFinley @Zruty0 @markusweimer @danmosemsft @stephentoub

eerhardt:
In order to extract these types, the following work will need to be performed:

- [ ] [Remove ISchema](https://github.com/dotnet/machinelearning/issues/1501)
- [ ] [Rename UInt128](https://github.com/dotnet/machinelearning/issues/1534)
- [ ] Finish the ColumnType cleanup work started with https://github.com/dotnet/machinelearning/issues/1533
- [ ] IRowCursorConsolidator needs to break [its dependency on `IChannelProvider`](https://github.com/dotnet/machinelearning/blob/1f73f720e132e920fab9fe5af49dcf595c2276c1/src/Microsoft.ML.Core/Data/IDataView.cs#L132). We cannot bring `IChannelProvider` into the new `DataView` NuGet/assembly.

/cc @singlis

TomFinley:
[...]

Issue #1533, opened 2018-11-05
IDataView Cleanup: ColumnType cleanup


TomFinley:
The `IDataView` type system is extensible (as we see with [`ImageType`](https://github.com/dotnet/machinelearning/blob/f9202628fbfac9e599e8c63dc5ed26eae77afbee/src/Microsoft.ML.ImageAnalytics/ImageType.cs#L12)).

This is fine, but there is something confusing about `ColumnType` as well, since there are lots of methods and properties on the base class `ColumnType` that are specific for derived types. For example, `.IsVector`, `.KeyCount`, and other such things, that are only really relevant if the type *is* either vector, key, or whatever.

## Why clean up?

There are lots of things on `ColumnType` that are unappealing. There are things like `AsVector` and `IsVector` which, as the documentation states, are equivalent to `as VectorType` and `is VectorType`. I mean, *why*? You save a few characters here and there, but at the cost of complicating one of the most central classes in the API.

Some things are just plain old silly. Why `IsTimeSpan`? How useful is that, really? Some things are like this.

There's also `DataKind`. This is so strange. This has already caused a fair amount of confusion among some people: they see this, and they think, "oh the types are just from this `enum`." No, they're not.

The reality is, these things are *conveniences*, but they're conveniences I think that confuse people (multiple smart people have thought their presence meant the type system was *not* extensible), so maybe we ought not to expose them, at least, not in their current form.

## Why not clean up?

In a sense, forming an analogue between the IDV and .NET type systems, there is some precedent for this sort of thing: if we consider `System.Type`. This has the property `IsArray`, with the methods `GetArrayRank`, which is only sensible to use. However in *our* case, `ColumnType` is an abstract class, and while `System.Type` is abstract, its inheritance structure does not capture specific types of values in this sense. If, hypothetically, `.GetType` of an `int[]` returned some type `System.ArrayType` that descended from `System.Type`, then we might equally hypothetically imagine that the method to get the array rank would be on that derived `ArrayType`, rather than on `Type` directly.

There is also a practical consideration. The reality though is that some types are definitely more important and more heavily used than others.

Let's imagine that we kept the `ColumnType` inheritance structure as it is now, but removed any properties relevant only to any derived type, specifically. What would hypothetically happen? I picked this usage more or less randomly from our codebase.

[...]

PR #2057, opened 2019-01-07
Remove "KeyType" specific members on ColumnType.

eerhardt:
Remove the following members from ColumnType:

- IsKey
- KeyCount
- KeyCountCore


src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
```diff
@@ -17,5 +17,10 @@ internal static class ColumnTypeExtensions
         public static bool IsStandardScalar(this ColumnType columnType) =>
             (columnType is NumberType) || (columnType is TextType) || (columnType is BoolType) ||
             (columnType is TimeSpanType) || (columnType is DateTimeType) || (columnType is DateTimeOffsetType);
+
+        /// <summary>
+        /// Zero return means either it's not a key type or the cardinality is unknown.
+        /// </summary>
+        public static int KeyCount(this ColumnType columnType) => (columnType as KeyType)?.Count ?? 0;
```
eerhardt: Thoughts on whether this should be `GetKeyCount()` instead? It's an internal API, so we can rename it at any time...

src/Microsoft.ML.Data/Data/RowCursorUtils.cs
```diff
@@ -407,9 +409,12 @@ public static ValueGetter<VBuffer<Single>> GetLabelGetter(SlotCursor cursor)
                 return cursor.GetGetter<Single>();
             if (type == NumberType.R8 || type is BoolType)
                 return GetVecGetterAs<Single>(NumberType.R4, cursor);
-            Contracts.Check(type.IsKey, "Only floating point number, boolean, and key type values can be used as label.");
+            if (!(type is KeyType keyType))
+            {
+                throw Contracts.Except("Only floating point number, boolean, and key type values can be used as label.");
+            }
```
artidoro: Minor comment: maybe we can do this in one line.

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
```diff
@@ -17,5 +17,10 @@ internal static class ColumnTypeExtensions
...
     }
```
artidoro: Why do we keep this extension? You very nicely cast to KeyType and check inline with (type is KeyType key) and then access the Count field directly. Isn't that a more correct access the Count field?

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
```diff
@@ -17,5 +17,10 @@ internal static class ColumnTypeExtensions
         public static bool IsStandardScalar(this ColumnType columnType) =>
```
sfilipi: >public [](start = 8, length = 6)

Checking, those are public, rather than internal to spare the [BestFriend] decoration?

src/Microsoft.ML.Data/Transforms/NormalizeColumn.cs
```diff
@@ -746,14 +746,14 @@ private ValueGetter<int> GetLabelGetter(Row row, int col, out int labelCardinali
...
-                if (type.IsKey)
+                if (type is KeyType keyType)
                 {
-                    Host.Assert(type.KeyCount > 0);
-                    labelCardinality = type.KeyCount;
+                    Host.Assert(keyType.Count > 0);
+                    labelCardinality = keyType.Count;
 
-                    int size = type.KeyCount;
+                    int size = keyType.Count;
```
sfilipi: > int size = keyType.Count; [](start = 19, length = 26)

funny this gets declared, instead of using labelCardinality

src/Microsoft.ML.Recommender/RecommenderUtils.cs
```diff
@@ -32,8 +32,7 @@ public static void CheckAndGetMatrixIndexColumns(RoleMappedData data, out Schema
         private static bool TryMarshalGoodRowColumnType(ColumnType type, out KeyType keyType)
         {
             keyType = type as KeyType;
-            return type.KeyCount > 0 && type.RawKind == DataKind.U4 &&
-                   keyType != null;
+            return keyType != null && keyType.Count > 0 && type.RawKind == DataKind.U4;
```
sfilipi: >keyType != null && keyType.Count [](start = 19, length = 32)

can those be compacted to keyType?.Count

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
TomFinley: They're definitely internal. The class is internal and `[BestFriend]`ed, so the methods of course don't have to be.

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
TomFinley: Since you solicited an opinion a verbified name would probably be best. Even for internal APIs, it might be best to set good practices, since I notice bad practices have a way of spreading. 😄 But given that your PR is ready *right now* to check in, I'd be inclined to not bother...

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
eerhardt: It's just a convenience/helper for the cases where we would have cast to KeyType anyway. This just consolidates the code, so we don't have to have the `as KeyType` and `?? 0` everywhere. This line of code would need to be duplicated 59 times if we didn't keep it.

src/Microsoft.ML.Data/Data/RowCursorUtils.cs
eerhardt: Unfortunately we can't use `Contracts.Check` anymore because the C# compiler doesn't know that `Contracts.Check` will throw if the parameter evaluates to `false`. So it complains if you use `keyType` after the `Contracts.Check` line.

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
eerhardt: I'll fix it - it is a simple "Rename" in VS, so very little work.

PR #2058, opened 2019-01-07
MLContext.Data.ReadFromBinary to accept IMultiStreamSource rather than Stream

### LINKED ISSUE #2052: BinaryLoader created from IMultiStreamSource not Stream
Whenever a binary loader is created, it can be crated from either a path or a stream. It is inappropriate for the public API to intimate that it is possible for the binary loader to be creatable from a stream. Even though it can be, this is a historical accident as the format (and its reader) predates `IMultiStreamSource`. Rather it, like all readers of this shape, should take `IMultiStreamSource`. That *internally* it operates over a single stream is an implementation detail that should not be visible externally.

najeeb-kazmi:
Fixes #2052

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoaderSaverCatalog.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoaderSaverCatalog.cs
@@ -14,14 +14,14 @@ namespace Microsoft.ML
         /// Read a data view from a Stream on a binary file using <see cref="BinaryLoader"/>.
         /// </summary>
         /// <param name="catalog">The catalog.</param>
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoaderSaverCatalog.cs
```diff
@@ -14,14 +14,14 @@ public static class BinaryLoaderSaverCatalog
         /// Read a data view from a Stream on a binary file using <see cref="BinaryLoader"/>.
```
Ivanidzo4ka: >Stream [](start = 36, length = 6)

need to be updated as well.

PR #2059, opened 2019-01-07
Make TextLoader.ArgumentsCore.Separator internal

### LINKED ISSUE #2041: TextLoader Separator/SeparatorChars
Consider these two settings:

https://github.com/dotnet/machinelearning/blob/496e185fdaf8969ffd856c65c42dd259368a242d/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L356-L360

They both exist for the same purpose but serve very different endpoints. One exists for the benefit of the command line, one for the benefit of the API. The API one (currently named `SeparatorChars`) should instead become `Separators` (plural pursuant to #2040). The other, currently named `Separator`, can remain named that, but should be invisible to API users since it exists purely for the benefit of the command line users. So, it should become `internal`.

/cc @stephentoub

mareklinka:
Since `TextLoader.ArgumentsCore.Separator` is only used for the command line interface, #2041 aims to make the field internal so that it doesn't show for API users. In order to facilitate this hiding, the `CmdParser.GetArgumentInfo` had to be updated to reflect on both public and private fields of the provided type.

At the same time, the API-serving `SeparatorChars` field was renamed to simply `Separators`.

Work in progress. More fields might be turned `internal`, depending on discussion in the issue.

Fixes #2041


src/Microsoft.ML.Core/CommandLine/CmdParser.cs
```diff
@@ -466,7 +466,7 @@ private static ArgumentInfo GetArgumentInfo(Type type, object defaults)
             var map = new Dictionary<string, Argument>();
             Argument def = null;
 
-            foreach (FieldInfo field in type.GetFields())
+            foreach (FieldInfo field in type.GetFields(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance))
```
TomFinley: >BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance [](start = 55, length = 68)

Cool. As mentioned in the associated issue, we might need to do something like this in one or two more places.

mareklinka:
As mentioned, I refactored the `LightGbmArguments.UpdateParameters` method to pay attention to the `ArgumentAttribute`. I wrote three simple unit tests as a safety net for the refactoring, these should be removed before merging as they don't provide any real value (and increase maintenance costs later on).

I was unsure whether to do any additional checks on the attribute (e.g. check `Visibility`, similar to what `CmdParser.GetArgumentInfo` does) - let me know if anything is necessary here, I'll add it.

cc @TomFinley

Ivanidzo4ka:
Can you also touch https://github.com/dotnet/machinelearning/blob/0c62e30b4d9eabb60322b2a3e75bc90e20007889/src/Microsoft.ML.Data/EntryPoints/InputBuilder.cs#L61
?

src/Microsoft.ML.LightGBM/LightGbmArguments.cs
```diff
@@ -56,9 +56,23 @@ protected BoosterParameter(TArgs args)
...
-                FieldInfo[] fields = Args.GetType().GetFields();
+                FieldInfo[] fields = Args.GetType().GetFields(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);
                 foreach (var field in fields)
+                {
+                    var attribute = GetAttribute(field);
+
+                    if (attribute == null)
+                    {
+                        continue;
+                    }
```
sfilipi: >} [](start = 20, length = 1)

stylistic nit: no { } for one line instructions

src/Microsoft.ML.LightGBM/LightGbmArguments.cs
```diff
@@ -56,9 +56,23 @@ protected BoosterParameter(TArgs args)
...
+
                     res[GetArgName(field.Name)] = field.GetValue(Args);
+                }
+            }
+
+            private ArgumentAttribute GetAttribute(FieldInfo field)
+            {
+                return field.GetCustomAttribute<ArgumentAttribute>(false);
```
sfilipi: nit: favor => syntax

test/Microsoft.ML.Predictor.Tests/SmokeTests.cs
```diff
@@ -0,0 +1,45 @@
+using System.Collections.Generic;
+using Microsoft.ML.LightGBM;
+using Xunit;
+
+namespace Microsoft.ML.RunTests
+{
+    public class SmokeTests
```
sfilipi: > public class SmokeTests [](start = 3, length = 24)

were you planning to remove those?

sfilipi:
thanks for working on this @mareklinka. It looks good to me, and i'll be good to sign off as soon as the SmokeTests.cs is gone. 

---
In reply to: [454704447](https://github.com/dotnet/machinelearning/pull/2059#issuecomment-454704447) [](ancestors = 454704447)

src/Microsoft.ML.LightGBM/LightGbmArguments.cs
mareklinka: Right, force of habit. Fixed.

test/Microsoft.ML.Predictor.Tests/SmokeTests.cs
mareklinka: Yup. Done.

PR #2073, opened 2019-01-08
Added AggressiveInlining attribute to hashing and updating the sdk version for netcorapp 3.0

Anipik:
Related to https://github.com/dotnet/coreclr/issues/21583

There has been a 25 percent improvement in the HashVectorString benchmark. 
The Time reduced from 34.61 to 26.16 for netcoreapp2.1
The Time reduced from 34.62 to 27 for netcoreapp3.0 preview

Note - Earlier this test was showing regression(14%) on netcoreapp3.0 but doesnot seem like a  regression now


src/Microsoft.ML.Core/Utilities/Hashing.cs
```diff
@@ -113,6 +114,7 @@ public static uint MurmurRound(uint hash, uint chunk)
         /// * 0x0800 to 0xFFFF : 1110xxxx 10xxxxxx 10xxxxxx
         /// NOTE: This MUST match the StringBuilder version below.
         /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
jkotas: Are you sure that the AggressiveInlining is needed / profitable on all these methods? E.g. this method is very large - I doubt that it is beneficial to aggresive inline it. 

The original issue was just about the Rotate.

Anipik: I tried removing AggressiveInlining attribute from murmur hash and the time for the benchmarks rises by 1.2 seconds for both netcoreapp2.1 and netcoreapp3.0

In the email thread @kouvel mentioned to add the attributes  to both the functions. Although the code that I added for Murmur hash in the related issue was little bit smaller version of the actual code

kouvel: Sorry I meant to say MurmurRound not MurmurHash. The last MurmurHash overload may be small enough to be worth inlining but the others look too large.

jkotas: Right, `AggressiveInlining` on method with loops is very questionable.

justinormont:
Would it be reasonable to modify DotNet's heuristics on when to inline? Aka, instead of making the change here, modify DotNet itself so other packages also get the gains.

kouvel:
It's difficult for inlining heuristics to be perfect, there will always be cases where forcing it would be necessary in perf-critical cases. It may be easier with improved heuristics or having an API that can be treated as an intrinsic to catch `Rotate` but catching `MurmurRound` would be more difficult because its code becomes smaller and faster after `Rotate` is inlined.

PR #2085, opened 2019-01-09
OnnxTransform -- Update to OnnxRuntime 0.2.0

### LINKED ISSUE #2056: Add Linux support for OnnxRuntime and OnnxTransform
Require scoring support for Onnx models on Linux

jignparm:
Add Linux support for  OnnxTransform.

Fixes #2056


Directory.Build.props
```diff
@@ -18,6 +18,7 @@
       https://api.nuget.org/v3/index.json;
       https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json;
       https://dotnet.myget.org/F/dotnet-core/api/v3/index.json;
+      https://www.myget.org/F/onnxruntime-dev/api/v3/index.json
```
eerhardt: Will the package be added to `nuget.org` soon? When it is, can this be removed?

pkg/Microsoft.ML.OnnxTransform/Microsoft.ML.OnnxTransform.nupkgproj
```diff
@@ -7,7 +7,7 @@
 
   <ItemGroup>
     <ProjectReference Include="../Microsoft.ML/Microsoft.ML.nupkgproj" />
-    <PackageReference Include="Microsoft.ML.OnnxRuntime.Gpu" Version="$(MicrosoftMLOnnxRuntimeGpuPackageVersion)"/>
+    <PackageReference Include="Microsoft.ML.OnnxRuntime" Version="$(MicrosoftMLOnnxRuntimePackageVersion)"/>
```
eerhardt: Why aren't we using the GPU package anymore?

test/Microsoft.ML.OnnxTransformTest/DnnImageFeaturizerTest.cs
```diff
@@ -60,6 +60,8 @@ public DnnImageFeaturizerTests(ITestOutputHelper helper) : base(helper)
         [ConditionalFact(typeof(Environment), nameof(Environment.Is64BitProcess))]
         void TestDnnImageFeaturizer()
         {
+            // Onnxruntime supports Ubuntu 16.04, but not CentOS
```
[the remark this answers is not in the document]
eerhardt: Can this be put into the `ConditionalFact` instead?

---
In reply to: [251941977](https://github.com/dotnet/machinelearning/pull/2085#discussion_r251941977) [](ancestors = 251941977)

test/Microsoft.ML.OnnxTransformTest/OnnxTransformTests.cs
```diff
@@ -234,17 +240,21 @@ public void OnnxStatic()
...
         void TestCommandLine()
         {
+            // Onnxruntime supports Ubuntu 16.04, but not CentOS
+            // Do not execute on CentOS image
             if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                 return;
 
             var env = new MLContext();
-            var x = Maml.Main(new[] { @"showschema loader=Text{col=data_0:R4:0-150527} xf=Onnx{InputColumns={data_0} OutputColumns={softmaxout_1} model={squeezenet/00000001/model.onnx} GpuDeviceId=0 FallbackToCpu=+}" });
+            var x = Maml.Main(new[] { @"showschema loader=Text{col=data_0:R4:0-150527} xf=Onnx{InputColumns={data_0} OutputColumns={softmaxout_1} model={squeezenet/00000001/model.onnx} GpuDeviceId=" + Int32.MinValue + " FallbackToCpu=+}" });
```
eerhardt: Does `Int32.MinValue` mean something specific?

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
```diff
@@ -81,11 +81,13 @@ public OnnxModel(string modelFile, int? gpuDeviceId = null, bool fallbackToCpu =
         {
             _modelFile = modelFile;
 
-            if (gpuDeviceId.HasValue)
+            // Reserve Int32.MinValue for test purpose.
```
eerhardt: Why can't we use `null` instead?

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
```diff
@@ -43,16 +43,19 @@ namespace Microsoft.ML.Transforms
...
+    /// <a href='https://developer.nvidia.com/cuda-downloads'>CUDA 9.1 Toolkit</a>
     /// and
-    /// <a href='https://developer.nvidia.com/cudnn'>cuDNN</a>
-    /// , and set the parameter 'gpuDeviceId' to a valid non-negative integer. Typical device ID values are 0 or 1.
+    /// <a href='https://developer.nvidia.com/cudnn'>cuDNN</a>.
+    ///  Set parameter 'gpuDeviceId' to a valid non-negative integer. Typical device ID values are 0 or 1.
     /// </p>
     /// <p>The inputs and outputs of the ONNX models must be Tensor type. Sequence and Maps are not yet supported.</p>
-    /// <p>OnnxRuntime currently works on Windows 64-bit platforms only. Linux and OSX to be supported soon.</p>
+    /// <p>OnnxRuntime currently works on Windows and Ubuntu 16.04 Linux 64-bit platforms. OSX to be supported soon.</p>
```
eerhardt: What about `macOS`?

test/Microsoft.ML.OnnxTransformTest/OnnxTransformTests.cs
shmoradims: If it is something specific, please define it as a const (in OnnxUtils, OnnxModel, or wherever makes sense) with the proper descriptive name. Magic numbers are confusing.

---
In reply to: [251942412](https://github.com/dotnet/machinelearning/pull/2085#discussion_r251942412) [](ancestors = 251942412)

test/Microsoft.ML.OnnxTransformTest/DnnImageFeaturizerTest.cs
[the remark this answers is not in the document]
jignparm: What's the best way to distinguish using ConditionalFact + Environment? Environment.OSVersion.PlatformID contains the following enums -- doesn't look suitable to add to ConditionalFact, but maybe another variable does.

        Win32S = 0,
        Win32Windows = 1,
        Win32NT = 2,
        WinCE = 3,
        Unix = 4,
        Xbox = 5,
        MacOSX = 6

---
In reply to: [251942092](https://github.com/dotnet/machinelearning/pull/2085#discussion_r251942092) [](ancestors = 251942092,251941977)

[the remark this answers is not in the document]
eerhardt: Unfortunately the `Environment.OSVersion` API is broken, please don't use it.

Instead, use https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation?view=netstandard-2.0

I don't agree with the way we are currently doing our ConditionalFacts in the repo.

https://github.com/dotnet/machinelearning/blob/8cdfa613c363b45de4b61abef8790943243ba818/test/Microsoft.ML.TestFramework/BaseTestBaseline.cs#L29-L33

Instead, I think we should have an `[OnnxFact]` and `[TensorFlowFact]` etc. And inside of that attribute, we have the rules for when Onnx can run, or when TensorFlow can run. Does that make sense?

Can you add a new attribute that says "this test can run on Windows x64 and macOS x64" and name it something like `[OnnxFact]` (or a better name)?  Here's an example: https://github.com/dotnet/corefx/blob/a10890f4ffe0fadf090c922578ba0e606ebdd16c/src/System.Data.Odbc/tests/TestCommon/CheckConnStrSetupFactAttribute.cs#L9

test/Microsoft.ML.OnnxTransformTest/OnnxTransformTests.cs
jignparm: Defined "public const int NullGpuID = int.MinValue" in OnnxTransformer class. OnnxModel and OnnxUtils class was made "internal sealed" in previous round, since we don't want to expose externally.

---
In reply to: [252010420](https://github.com/dotnet/machinelearning/pull/2085#discussion_r252010420) [](ancestors = 252010420,251942412)

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
jignparm: changed to null.

---
In reply to: [251942603](https://github.com/dotnet/machinelearning/pull/2085#discussion_r251942603) [](ancestors = 251942603)

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
jignparm: MacOs is to-be-supported. The runtimes are not tested yet.

---
In reply to: [251942701](https://github.com/dotnet/machinelearning/pull/2085#discussion_r251942701) [](ancestors = 251942701)

pkg/Microsoft.ML.OnnxTransform/Microsoft.ML.OnnxTransform.nupkgproj
jignparm: The GPU package will be available on Nuget.Org.  In particular for Linux, CPU execution is blocked if the libraries are not available.
Using the CPU package will enable Linux testing as well, if an Ubuntu CI leg is available in the future.

---
In reply to: [251941810](https://github.com/dotnet/machinelearning/pull/2085#discussion_r251941810) [](ancestors = 251941810)

test/Microsoft.ML.OnnxTransformTest/DnnImageFeaturizerTest.cs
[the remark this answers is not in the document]
jignparm: Thanks for the pointers. Updated to use [OnnxFact] instead, which essentially runs on Windows 64 bit. Mac OS is not supported, but there's an option of extending to Linux Ubuntu if a new leg is introduced.

pkg/Microsoft.ML.OnnxTransform/Microsoft.ML.OnnxTransform.nupkgproj
eerhardt: I’m not understanding this. Will ML.NET use the GPU package or not? Why did we switch to the GPU package before, only to switch back off the GPU package now?

jignparm: The primary reason is that OnnxRuntime GPU package won't run on Linux  without the CUDA libraries are installed, even if a user wants to just run on CPU. To avoid confusion, we should go with a default of a CPU package, which will run cross-platform (Ubuntu at least) without requiring any extra software. And then, to use the GPU package, request users to install CUDA and the GPU OnnxRuntime package. The out-of-box package should run everywhere without extra software. When we add MacOS support, the runtimes will be available only in the CPU package (there's no GPU runtime for Mac), so long term using the CPU package seems like a better option.

Most other frameworks have 2 separate packages.

See the  MXNet GPU support instructions here --> [https://mxnet.incubator.apache.org/versions/master/install/windows_setup.html#install-with-gpus](https://mxnet.incubator.apache.org/versions/master/install/windows_setup.html#install-with-gpus). 
And Theano as well --> [http://deeplearning.net/software/theano/install_windows.html](http://deeplearning.net/software/theano/install_windows.html).



---
In reply to: [252102461](https://github.com/dotnet/machinelearning/pull/2085#discussion_r252102461) [](ancestors = 252102461)

Directory.Build.props
jignparm: removed.

pkg/Microsoft.ML.OnnxTransform/Microsoft.ML.OnnxTransform.nupkgproj
eerhardt: OK, this works for now. I agree it is best to have our "default" work in as many places as possible.

> documentation currently says to use the Microsoft.ML.OnnxRuntime.Gpu package if users want to run on GPU. Currently they'll need to rebuild MLNET from source to do this. 

We typically don't take this route in .NET libraries*. Instead we typically provide all options in binary form, i.e. add an optional OnnxTransform-Gpu separate package/transform. 

(*) Reasons for this:
1. It means the files are no longer signed by Microsoft.
2. It messes with dependencies, because now if someone wanted to ship a package that depended on the OnnxTransform.Gpu package, they would need to redistribute it themselves. And if 2 people wanted to do it, they would have separate packages.

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
eerhardt: I meant, why do we need to have `OnnxTransformer.NullGpuID`.  Why can't we use the literal `null` for the `int?` parameter?

---
In reply to: [252049662](https://github.com/dotnet/machinelearning/pull/2085#discussion_r252049662) [](ancestors = 252049662,251942603)

test/Microsoft.ML.OnnxTransformTest/DnnImageFeaturizerTest.cs
```diff
@@ -61,9 +61,11 @@ public DnnImageFeaturizerTests(ITestOutputHelper helper) : base(helper)
...
```
eerhardt: Can you use the `[OnnxFact]` here instead? And remove the

```
            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                return;
```

below?

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
jignparm: The command line parser test (below), requires values for all input parameters, even for nullable parameters (not sure if this is by design or not). However, for the nullable input parameters, there's doesn't seem to be a way to pass a value which represents 'null'.  Since we can't pass in a null value in the particular test, I needed a non-null value which would not actually try to run on Gpu.  Let me know if there's a way to pass in a 'null' value in the command line parser test -- that would obviate the need for this variable.

(https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.OnnxTransformTest/OnnxTransformTests.cs#L243)

eerhardt: It's because the property has `ArgumentType.Required` on its attribute:

https://github.com/dotnet/machinelearning/blob/0c385eee4b2bc532ae90b9d4fe913d18547f96ac/src/Microsoft.ML.OnnxTransform/OnnxTransform.cs#L73-L74

jignparm: Update the attribute to [Argument(ArgumentType.AtMostOnce | *ArgumentType.AtMostOnce*,...]
And removed the NullGpu now.

test/Microsoft.ML.OnnxTransformTest/OnnxTransformTests.cs
```diff
@@ -21,9 +21,26 @@
...
+
+    /// <summary>
+    /// A Fact attribute for Onnx unit tests. Onnxruntime only supported
+    /// on Windows, Linux (Ubuntu 16.04) and 64-bit platforms.
+    /// </summary>
+    public class OnnxFact : FactAttribute
     {
+        public OnnxFact()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ||
```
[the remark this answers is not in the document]
jignparm: The logic is correct. The skip is because MLNet is on CentOS image. The transform runs fine on Ubuntu (locally). We cannot test on Linux image in MLNet

src/Microsoft.ML.OnnxTransform/OnnxTransform.cs
```diff
@@ -70,10 +73,10 @@ public sealed class Arguments : TransformInputBase
             [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "Name of the output column.", SortOrder = 2)]
             public string[] OutputColumns;
 
-            [Argument(ArgumentType.AtMostOnce | ArgumentType.Required, HelpText = "GPU device id to run on (e.g. 0,1,..). Null for CPU. Requires CUDA 10.0.", SortOrder = 3)]
+            [Argument(ArgumentType.AtMostOnce | ArgumentType.AtMostOnce, HelpText = "GPU device id to run on (e.g. 0,1,..). Null for CPU. Requires CUDA 9.1.", SortOrder = 3)]
```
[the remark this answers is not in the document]
jignparm: Good catch. removed.

test/Microsoft.ML.OnnxTransformTest/DnnImageFeaturizerTest.cs
jignparm: Leaving it as-is for now. Switching to OnnxFact is causing some other odd issues.

src/Microsoft.ML.OnnxTransform/OnnxUtils.cs
```diff
@@ -82,11 +82,13 @@ public OnnxModel(string modelFile, int? gpuDeviceId = null, bool fallbackToCpu =
...
```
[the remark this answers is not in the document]
jignparm: removed.

Ivanidzo4ka:
>minor change to kick off build
You don't have to do it, you can always restart or retry build via UI.

PR #2088, opened 2019-01-09
Replacing TryGetColumnIndex with invocations of GetColumnOrNull 

sfilipi:
Work towards #1500 by removing references of TryGetColumnIndex, and replacing them with GetColumnOrNull in a few projects.

### DIFF
--- a/src/Microsoft.ML.EntryPoints/FeatureCombiner.cs
+++ b/src/Microsoft.ML.EntryPoints/FeatureCombiner.cs
@@ -233,7 +233,7 @@ namespace Microsoft.ML.EntryPoints
 
             var labelCol = input.Data.Schema.GetColumnOrNull(input.LabelColumn);
             if (!labelCol.HasValue)
-                throw host.Except($"Column '{input.LabelColumn}' not found.");
+                throw host.ExceptSchemaMismatch(nameof(input), "Label", input.LabelColumn);
 
             var labelType = labelCol.Value.Type;
             if (labelType.IsKey || labelType is BoolType)
@@ -269,7 +269,7 @@ namespace Microsoft.ML.EntryPoints
 
             var predictedLabelCol = input.Data.Schema.GetColumnOrNull(input.PredictedLabelColumn);
             if (!predictedLabelCol.HasValue)
-                throw host.Except($"Column '{input.PredictedLabelColumn}' not found.");
+                throw host.ExceptSchemaMismatch(nameof(input), "PredictedLabel",input.PredictedLabelColumn);
             var predictedLabelType = predictedLabelCol.Value.Type;
             if (predictedLabelType is NumberType || predictedLabelType is BoolType)
             {
--- a/src/Microsoft.ML.EntryPoints/FeatureCombiner.cs
+++ b/src/Microsoft.ML.EntryPoints/FeatureCombiner.cs
@@ -115,15 +115,15 @@ namespace Microsoft.ML.EntryPoints
         {
             Contracts.AssertValue(data);
             Contracts.AssertNonWhiteSpace(colName);
-            int col;
             var schema = data.Schema;
-            if (!schema.TryGetColumnIndex(colName, out col))
+            var col = schema.GetColumnOrNull(colName);
+            if (!col.HasValue)
                 return null;
-            var type = schema[col].Metadata.Schema.GetColumnOrNull(MetadataUtils.Kinds.KeyValues)?.Type;
+            var type = col.Value.Metadata.Schema.GetColumnOrNull(MetadataUtils.Kinds.KeyValues)?.Type;
             if (type == null || !type.IsKnownSizeVector || !(type.ItemType is TextType))
                 return null;
             var metadata = default(VBuffer<ReadOnlyMemory<char>>);
-            schema[col].Metadata.GetValue(MetadataUtils.Kinds.KeyValues, ref metadata);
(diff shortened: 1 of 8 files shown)


src/Microsoft.ML.FastTree/GamTrainer.cs
```diff
@@ -1285,8 +1285,9 @@ public Context(IChannel ch, GamModelParametersBase pred, RoleMappedData data, IE
                             new RoleMappedSchema.ColumnRole(MetadataUtils.Const.ScoreValueKind.Score).Bind(DefaultColumnNames.Score));
                     }
 
-                    _data.Schema.Schema.TryGetColumnIndex(DefaultColumnNames.Features, out int featureIndex);
-                    MetadataUtils.TryGetCategoricalFeatureIndices(_data.Schema.Schema, featureIndex, out _catsMap);
+                    var featureCol = _data.Schema.Schema.GetColumnOrNull(DefaultColumnNames.Features);
+                    ch.Assert(featureCol.HasValue);
+                    MetadataUtils.TryGetCategoricalFeatureIndices(_data.Schema.Schema, featureCol.Value.Index, out _catsMap);
```
TomFinley: In this particular case, you are asserting that it has a value, which is fine, but if the purpose is to just to clarify this, you could *probably* just get away with just doing `.Value` at the end and operating over that, since that also clarifies, "yes, I definitely expect this to have a value."

src/Microsoft.ML.EntryPoints/FeatureCombiner.cs
```diff
@@ -266,10 +267,10 @@ public static CommonOutputs.TransformOutput ConvertPredictedLabel(IHostEnvironme
             host.CheckValue(input, nameof(input));
             EntryPointUtils.CheckInputArgs(host, input);
 
-            int predictedLabelCol;
-            if (!input.Data.Schema.TryGetColumnIndex(input.PredictedLabelColumn, out predictedLabelCol))
+            var predictedLabelCol = input.Data.Schema.GetColumnOrNull(input.PredictedLabelColumn);
+            if (!predictedLabelCol.HasValue)
                 throw host.Except($"Column '{input.PredictedLabelColumn}' not found.");
```
TomFinley: >Except [](start = 27, length = 6)

I'm not sure if you want to do this at the same time, but in the general movement to start using `ExceptSchemaMismatch` to ensure greater uniformity between error messages, you might consider doing that transition at the same time. But, not sure if you want to.

src/Microsoft.ML.HalLearners/VectorWhitening.cs
```diff
@@ -381,9 +381,12 @@ private static void GetColTypesAndIndex(IHostEnvironment env, IDataView inputDat
 
             for (int i = 0; i < columns.Length; i++)
             {
-                if (!inputSchema.TryGetColumnIndex(columns[i].Input, out cols[i]))
+                var col = inputSchema.GetColumnOrNull(columns[i].Input);
+                if (!col.HasValue)
                     throw env.ExceptSchemaMismatch(nameof(inputSchema), "input", columns[i].Input);
```
TomFinley: >ExceptSchemaMismatch [](start = 30, length = 20)

This would be an example of a current  existing usage of `ExceptSchemaMismatch`, perhaps a pattern to follow in the other places where we are currently using `Except`.

src/Microsoft.ML.LightGBM/LightGbmTrainerBase.cs
```diff
@@ -285,8 +285,9 @@ private CategoricalMetaData GetCategoricalMetaData(IChannel ch, RoleMappedData t
                 ch.Info("Auto-tuning parameters: " + nameof(Args.UseCat) + " = " + useCat);
             if (useCat)
             {
-                trainData.Schema.Schema.TryGetColumnIndex(DefaultColumnNames.Features, out int featureIndex);
-                MetadataUtils.TryGetCategoricalFeatureIndices(trainData.Schema.Schema, featureIndex, out categoricalFeatures);
+                var featureCol = trainData.Schema.Schema.GetColumnOrNull(DefaultColumnNames.Features);
+                ch.Assert(featureCol.HasValue);
```
TomFinley: >ch.Assert(featureCol.HasValue); [](start = 16, length = 31)

Similar note, can probably just use `.Value` in the line above to clarify "yes I expect this to have a value." But, as you like.

TomFinley:
Looks more or less good @sfilipi ! One thing though, you seem to be when you definitely *know* the column is there, still using this `OrNull` facility. Would it not be more appropriate, as we do elsewhere, to use this facility?

https://github.com/dotnet/machinelearning/blob/bdc9a9e1348abfdd2287d43e2633089e9eba36c7/src/Microsoft.ML.Core/Data/Schema.cs#L38

This would be simpler, and possibly more idiomatic with much of our usage elsewhere. What do you think?

sfilipi:
Good point! thanks for suggesting it!

---
In reply to: [452842662](https://github.com/dotnet/machinelearning/pull/2088#issuecomment-452842662) [](ancestors = 452842662)

src/Microsoft.ML.HalLearners/VectorWhitening.cs
sfilipi: Cool!

---
In reply to: [246524124](https://github.com/dotnet/machinelearning/pull/2088#discussion_r246524124) [](ancestors = 246524124)

PR #2091, opened 2019-01-09
remove unnecessary dependency

### LINKED ISSUE #2090: Microsoft.ML.Transforms.csproj should not have a dependency on mkl
Apparently [Microsoft.Ml.Transforms](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Transforms/Microsoft.ML.Transforms.csproj#L51) contains a dependency in the mlnetmkldeps package. 

No project that gets packaged under the main Microsoft.ML NuGet should depend on MKL. Components that need MKL should get moved to the HalLearners project.

sfilipi:
Resolves #2090 by removing the mlnetmkldeps package dependency.


src/Microsoft.ML.Transforms/Microsoft.ML.Transforms.csproj
```diff
@@ -48,7 +48,6 @@
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.ML.Core\Microsoft.ML.Core.csproj" />
     <ProjectReference Include="..\Microsoft.ML.Data\Microsoft.ML.Data.csproj" />
-    <PackageReference Include="MlNetMklDeps" Version="$(MlNetMklDepsPackageVersion)" />
```
Ivanidzo4ka: I think you need to add this line to HalLearners project. Otherwise we don't fetch MKL at all and SymSGD can't be build.

sfilipi: thanks Ivan. 

---
In reply to: [246568063](https://github.com/dotnet/machinelearning/pull/2091#discussion_r246568063) [](ancestors = 246568063)

PR #2101, opened 2019-01-10
Adding a test to show supported data types in TensorFlowTransform

### LINKED ISSUE #745: Handle additional types as input and output in TF models
Initially only floats/doubles will be handled, and we should investigate which other types are commonly used in TF models.

zeahmed:
This PR adds addition test to fix #745.


test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -70,6 +70,62 @@ public void TensorFlowTransformMatrixMultiplicationTest()
...
+        /// Test to ensure the support for int and int64 types.
+        /// </summary>
+        [ConditionalFact(typeof(Environment), nameof(Environment.Is64BitProcess))] // TensorFlow is 64-bit only
+        public void TensorFlowTransformAdditionOfIntAndInt64Test()
+        {
+            var model_location = "model_matadd_int_int64";
+            var mlContext = new MLContext(seed: 1, conc: 1);
+            // Pipeline
+            var loader = ComponentCreation.CreateDataView(mlContext,
+                    new List<TestDataInt>(new TestDataInt[] { new TestDataInt() { a = new[] { 1L, 2,
```
wschin: Not aligned?

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -70,6 +70,62 @@ public void TensorFlowTransformMatrixMultiplicationTest()
...
+                cgetter(ref c);
+
+                var cValues = c.GetValues();
+                Assert.Equal(1 + 1, cValues[0]);
+                Assert.Equal(2 + 2, cValues[1]);
+                Assert.Equal(3 + 3, cValues[2]);
+                Assert.Equal(4 + 4, cValues[3]);
+
+                Assert.True(cursor.MoveNext());
+                c = default;
```
yaeldekel: >c = default; [](start = 16, length = 12)

This line is not needed.

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -70,6 +70,62 @@ public void TensorFlowTransformMatrixMultiplicationTest()
...
+            var trans = TensorFlowTransform.Create(mlContext, loader, model_location, new[] { "c" }, new[] { "a", "b" });
+
+            using (var cursor = trans.GetRowCursor(a => true))
+            {
+                var cgetter = cursor.GetGetter<VBuffer<long>>(2);
+                Assert.True(cursor.MoveNext());
+                VBuffer<long> c = default;
+                cgetter(ref c);
+
+                var cValues = c.GetValues();
```
yaeldekel: >cValues [](start = 20, length = 7)

Add an assert for the length of this.

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -70,6 +70,62 @@ public void TensorFlowTransformMatrixMultiplicationTest()
...
+                Assert.Equal(1 + 1, cValues[0]);
+                Assert.Equal(2 + 2, cValues[1]);
+                Assert.Equal(3 + 3, cValues[2]);
+                Assert.Equal(4 + 4, cValues[3]);
+
+                Assert.True(cursor.MoveNext());
+                c = default;
+                cgetter(ref c);
+
+                cValues = c.GetValues();
```
yaeldekel: >cValues [](start = 16, length = 7)

Add an assert for the length of this.

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -70,6 +70,62 @@ public void TensorFlowTransformMatrixMultiplicationTest()
...
+            public long[] a;
+            [VectorType(4)]
+            public int[] b;
+        }
+
+        /// <summary>
+        /// Test to ensure the support for int and int64 types.
+        /// </summary>
+        [ConditionalFact(typeof(Environment), nameof(Environment.Is64BitProcess))] // TensorFlow is 64-bit only
+        public void TensorFlowTransformAdditionOfIntAndInt64Test()
```
singlis: >TensorFlowTransformAdditionOfIntAndInt64Test [](start = 20, length = 44)

Hi Zeeshan, is there supposed to be a test for the 32-bit model? I thought the other tests were 32-bit models, but It looks like the other tests either are skipped or disable unless its a 64-bit process.

zeahmed: TensorFlow works only on 64-bit platform. So, i think none of the TF tests are 32-bit.

---
In reply to: [246982838](https://github.com/dotnet/machinelearning/pull/2101#discussion_r246982838) [](ancestors = 246982838)

singlis: ah! sorry I misread, you are handling int64 and int32 types...so then do the other tests use models that are using int32s?


---
In reply to: [246982838](https://github.com/dotnet/machinelearning/pull/2101#discussion_r246982838) [](ancestors = 246982838)

singlis: actually I see -- you have a TestDataInt class that has both long and int....

---
In reply to: [246983378](https://github.com/dotnet/machinelearning/pull/2101#discussion_r246983378) [](ancestors = 246983378,246982838)

Ivanidzo4ka:
Tests are good, but original issue is not about tests, and more about, what types we can actually support? 
Can we support bool? Can we support short?
Do we have list of supported types in our documentation?
Do we have any exceptions in case if user passes type which we don't support? Not the cryptic one, but one which would say: "Current type "A" is unsupported please use following types: float, int, int64"?

wschin:
Can Tensorflow transform work with key types? They are integers as well.

zeahmed:
Updated the test to show how the supported types can be used.

---
In reply to: [453611413](https://github.com/dotnet/machinelearning/pull/2101#issuecomment-453611413) [](ancestors = 453611413,453388399)

zeahmed:
No, if they can be represented as any of the these types (double, float, long, int, short, sbyte, ulong, uint, ushort, byte and bool) then yes.

---
In reply to: [453671331](https://github.com/dotnet/machinelearning/pull/2101#issuecomment-453671331) [](ancestors = 453671331)

PR #2108, opened 2019-01-10
Added <Link> to CommonPackage.props for native DLLs

### LINKED ISSUE #1851: Native assemblies from NuGet aren't copied correctly for packages.config
### System information

- **OS version/distro**: Windows
- **.NET Version (eg., dotnet --info)**: .NET Framework 4.7.1

### Issue

- **What did you do?**
Create an "old style" .csproj using packages.config and added Microsoft.ML nuget package and built.

When creating the project, if you uncheck the "Create directory for solution" checkbox, then the .sln and .csproj are in the same folder. When you restore NuGet packages, the nuget files get put in a `packages` folder in the same folder as the .csproj.

- **What happened?**
The native assemblies (CpuMathNative, FastTreeNative, LdaNative, etc) were all copied into subdirectories of my output path instead of directly in my output path.

From looking at a binlog of the build, it appears that the `AssignTargetPath` is getting confused at our `Content` items since they appear to be part of the project (since they are under the same folder as the .csproj).

- **What did you expect?**
The native assemblies should be copied directly to the output folder.

### Note

To fix this, we should put the `<Link>` metadata on our content items:

https://github.com/dotnet/machinelearning/blob/14c7a472579afa3ce98bad2e3495e4c0524471b9/pkg/common/CommonPackage.props#L10-L20

See
* https://github.com/dotnet/roslyn/issues/15137#issuecomment-260470890
* https://github.com/Microsoft/onnxruntime/pull/127#discussion_r239845745

For other places that had this same bug.

najeeb-kazmi:
Fixes #1851 

I built the nugets locally and tested in a new solution to verify that the issue described in #1851 was resolved.

### DIFF
--- a/pkg/common/CommonPackage.props
+++ b/pkg/common/CommonPackage.props
@@ -11,13 +11,13 @@
              Condition="'$(PlatformTarget)' == 'x64'">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Visible>false</Visible>
-	  <Link>%(Filename)%(Extension)</Link>
+      <Link>%(Filename)%(Extension)</Link>
     </Content>
     <Content Include="$(MSBuildThisFileDirectory)\..\..\runtimes\win-x86\native\*.dll"
              Condition="'$(PlatformTarget)' == 'x86'">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Visible>false</Visible>
-	  <Link>%(Filename)%(Extension)</Link>
+      <Link>%(Filename)%(Extension)</Link>
     </Content>
   </ItemGroup>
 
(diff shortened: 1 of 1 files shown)


pkg/common/CommonPackage.props
```diff
@@ -11,11 +11,13 @@
              Condition="'$(PlatformTarget)' == 'x64'">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Visible>false</Visible>
+	  <Link>%(Filename)%(Extension)</Link>
```
eerhardt: (nit) can you fix the whitespace here? The file is currently using spaces, and you've inserted tabs.

najeeb-kazmi: Fixed

Issue #2002, opened 2019-01-02
Running a Dry Run for the benchmarks


Anipik:
Currently we are not running benchmarks on our CI. So they keep on breaking when we change the behavior. I propose running a dry job for all the benchmarks i.e. running the benchmarks for atleast one iteration.
eg.
Kmeans was broken and fixed by https://github.com/dotnet/machinelearning/pull/1883/files
Similarly before that stoachastic Gradient benchmarks were broken

cc @danmosemsft @eerhardt @adamsitnik

justinormont:
It would be useful to setup nightly runs of the benchmarks to look for regressions in perf. The benchmark outputs would be sent to a remote service for storage & visualization. We could go further and run on each PR before merging and gate on perf.

PR #2109, opened 2019-01-10
Added Dry job for benchmarks CI

Anipik:
Fixes https://github.com/dotnet/machinelearning/issues/2002

This PR will run all the benchmarks in the ci atleast one. This will ensure that the commits have not broken any of the benchmarks

### DIFF
--- a/test/Microsoft.ML.Benchmarks.Tests/BenchmarksTest.cs
+++ b/test/Microsoft.ML.Benchmarks.Tests/BenchmarksTest.cs
@@ -41,7 +41,7 @@ namespace Microsoft.ML.Benchmarks.Tests
         public static TheoryData<Type> GetBenchmarks()
         {
             TheoryData<Type> benchmarks = new TheoryData<Type>();
-            Assembly asm = Assembly.LoadFrom("Microsoft.ML.Benchmarks.dll");
+            Assembly asm = typeof(StochasticDualCoordinateAscentClassifierBench).Assembly;
 
             var types = from type in asm.GetTypes()
                         where Attribute.IsDefined(type, typeof(CIBenchmark))
--- a/test/Microsoft.ML.Benchmarks/Numeric/Ranking.cs
+++ b/test/Microsoft.ML.Benchmarks/Numeric/Ranking.cs
@@ -6,6 +6,7 @@ using BenchmarkDotNet.Attributes;
 using Microsoft.ML.Data;
 using Microsoft.ML.LightGBM;
 using Microsoft.ML.RunTests;
+using Microsoft.ML.TestFramework;
 using Microsoft.ML.Trainers.FastTree;
 using Microsoft.ML.Transforms.Conversions;
 using System.IO;
@@ -21,8 +22,8 @@ namespace Microsoft.ML.Benchmarks
         [GlobalSetup]
         public void SetupTrainingSpeedTests()
         {
-            _mslrWeb10k_Validate = TestFramework.BaseTestClass.GetDataPath(TestDatasets.MSLRWeb.validFilename);
-            _mslrWeb10k_Train = TestFramework.BaseTestClass.GetDataPath(TestDatasets.MSLRWeb.trainFilename);
+            _mslrWeb10k_Validate = BaseTestClass.GetDataPath(TestDatasets.MSLRWeb.validFilename);
+            _mslrWeb10k_Train = BaseTestClass.GetDataPath(TestDatasets.MSLRWeb.trainFilename);
 
             if (!File.Exists(_mslrWeb10k_Validate))
                 throw new FileNotFoundException(string.Format(Errors.DatasetNotFound, _mslrWeb10k_Validate));
@@ -71,9 +72,9 @@ namespace Microsoft.ML.Benchmarks
         [GlobalSetup]
         public void SetupScoringSpeedTests()
         {
-            _mslrWeb10k_Test = TestFramework.BaseTestClass.GetDataPath(TestDatasets.MSLRWeb.testFilename);
-            _mslrWeb10k_Validate = TestFramework.BaseTestClass.GetDataPath(TestDatasets.MSLRWeb.validFilename);
-            _mslrWeb10k_Train = TestFramework.BaseTestClass.GetDataPath(TestDatasets.MSLRWeb.trainFilename);
+            _mslrWeb10k_Test = BaseTestClass.GetDataPath(TestDatasets.MSLRWeb.testFilename);
+            _mslrWeb10k_Validate = BaseTestClass.GetDataPath(TestDatasets.MSLRWeb.validFilename);
+            _mslrWeb10k_Train = BaseTestClass.GetDataPath(TestDatasets.MSLRWeb.trainFilename);
 
             if (!File.Exists(_mslrWeb10k_Test))
                 throw new FileNotFoundException(string.Format(Errors.DatasetNotFound, _mslrWeb10k_Test));
--- a/test/Microsoft.ML.Benchmarks/Helpers/CIBenchmark.cs
+++ b/test/Microsoft.ML.Benchmarks/Helpers/CIBenchmark.cs
@@ -6,6 +6,10 @@ using System;
 
 namespace Microsoft.ML.Benchmarks.Harness
 {
+    /// <summary>
+    /// This attribute is used to identify the benchmarks 
+    /// which we want to run on the CI.
+    /// </summary>
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
     public class CIBenchmark : Attribute
     {
--- a/test/Microsoft.ML.Benchmarks/KMeansAndLogisticRegressionBench.cs
+++ b/test/Microsoft.ML.Benchmarks/KMeansAndLogisticRegressionBench.cs
@@ -6,13 +6,14 @@ using BenchmarkDotNet.Attributes;
 using Microsoft.ML.Benchmarks.Harness;
 using Microsoft.ML.Data;
 using Microsoft.ML.Internal.Calibration;
+using Microsoft.ML.TestFramework;
 
 namespace Microsoft.ML.Benchmarks
 {
     [CIBenchmark]
     public class KMeansAndLogisticRegressionBench
     {
-        private readonly string _dataPath = TestFramework.BaseTestClass.GetDataPath("adult.tiny.with-schema.txt");
+        private readonly string _dataPath = BaseTestClass.GetDataPath("adult.tiny.with-schema.txt");
 
         [Benchmark]
         public ParameterMixingCalibratedPredictor TrainKMeansAndLR()
--- a/test/Microsoft.ML.Benchmarks/PredictionEngineBench.cs
+++ b/test/Microsoft.ML.Benchmarks/PredictionEngineBench.cs
@@ -5,6 +5,7 @@
 using BenchmarkDotNet.Attributes;
 using Microsoft.ML.Benchmarks.Harness;
 using Microsoft.ML.Data;
+using Microsoft.ML.TestFramework;
 using Microsoft.ML.Trainers;
 using Microsoft.ML.Transforms;
 using Microsoft.ML.Transforms.Text;
@@ -34,7 +35,7 @@ namespace Microsoft.ML.Benchmarks
                 PetalWidth = 5.1f,
             };
 
-            string _irisDataPath = TestFramework.BaseTestClass.GetDataPath("iris.txt");
+            string _irisDataPath = BaseTestClass.GetDataPath("iris.txt");
 
             var env = new MLContext(seed: 1, conc: 1);
             var reader = new TextLoader(env,
@@ -67,7 +68,7 @@ namespace Microsoft.ML.Benchmarks
                 SentimentText = "Not a big fan of this."
             };
 
-            string _sentimentDataPath = TestFramework.BaseTestClass.GetDataPath("wikipedia-detox-250-line-data.tsv");
+            string _sentimentDataPath = BaseTestClass.GetDataPath("wikipedia-detox-250-line-data.tsv");
 
             var env = new MLContext(seed: 1, conc: 1);
             var reader = new TextLoader(env, columns: new[]
@@ -96,7 +97,7 @@ namespace Microsoft.ML.Benchmarks
                 Features = new[] { 5f, 1f, 1f, 1f, 2f, 1f, 3f, 1f, 1f }
             };
 
-            string _breastCancerDataPath = TestFramework.BaseTestClass.GetDataPath("breast-cancer.txt");
+            string _breastCancerDataPath = BaseTestClass.GetDataPath("breast-cancer.txt");
 
             var env = new MLContext(seed: 1, conc: 1);
             var reader = new TextLoader(env, columns: new[]
--- a/test/Microsoft.ML.Benchmarks/RffTransform.cs
+++ b/test/Microsoft.ML.Benchmarks/RffTransform.cs
@@ -7,6 +7,7 @@ using BenchmarkDotNet.Attributes;
 using Microsoft.ML.Benchmarks.Harness;
 using Microsoft.ML.Data;
 using Microsoft.ML.RunTests;
+using Microsoft.ML.TestFramework;
 using Microsoft.ML.Transforms.Conversions;
 
 namespace Microsoft.ML.Benchmarks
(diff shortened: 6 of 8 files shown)


test/Microsoft.ML.Benchmarks/Helpers/BenchmarkHelpers.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System.IO;
+
+namespace Microsoft.ML.Benchmarks
+{
+    internal class BenchmarkHelpers
+    {
+        public static string RootPath = Directory.GetParent(Directory.GetCurrentDirectory()).Parent.Parent.Parent.Parent.Parent.Parent.Parent.FullName;
+        public static string TestDataPath =  Path.Combine(RootPath, @"test/data");
+    }
```
adamsitnik: since we have the helper class, we could put the logic for building the path here:

```cs
internal static string GetPath(string fileName) => Path.Combine(TestDataPath, fileName);
```

and every benchmark would just call it providing a file name + those fields would be private and when we change the logic for building the path we can do it in one place with single change

test/Microsoft.ML.Benchmarks/Helpers/BenchmarkHelpers.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.IO;
+
+namespace Microsoft.ML.Benchmarks
+{
+    internal class BenchmarkHelpers
```
adamsitnik: this name is too general. What this class does? can you descibe it using simple words? these words should be the name of the class

test/Microsoft.ML.Benchmarks.Tests/BenchmarksTest.cs
```diff
@@ -39,14 +42,25 @@ public class BenchmarksTest
 
         private ITestOutputHelper Output { get; }
 
-#if DEBUG
-        [Fact(Skip = SkipTheDebug)]
+#if DEBUG || NET462
+        [ConditionalTheory(Skip = SkipTheDebugAndFullFramework)]
```
adamsitnik: this could be simplified: we are using `if` and `ConditionalTheory` here while the `ConditionalTheory` could just point to a method encapsulating the logic. For example:

```cs
        public bool CanExecute =>
#if DEBUG
            false; // BenchmarkDotNet does not allow running the benchmarks in Debug, so this test is disabled for DEBUG
#elif NET462
            false; // We are currently not running Benchmarks for FullFramework
#else
            Environment.Is64BitProcess; // we don't support 32 bit yet
#endif

[ConditionalTheory(typeof(BenchmarksTest), nameof(CanExecute))]
```

test/Microsoft.ML.Benchmarks.Tests/BenchmarksTest.cs
```diff
@@ -39,14 +42,25 @@ public class BenchmarksTest
...
 #else
-        [ConditionalFact(typeof(Environment), nameof(Environment.Is64BitProcess))]
+        [ConditionalTheory(typeof(Environment), nameof(Environment.Is64BitProcess))]
 #endif
-        public void BenchmarksProjectIsNotBroken()
+        // [InlineData(typeof(RankingTrain))]
+        // [InlineData(typeof(RankingTest))]
+        // [InlineData(typeof(MultiClassClassificationTrain))]
+        // [InlineData(typeof(MultiClassClassificationTest))]
```
adamsitnik: dead code (commented lines): you should either add comments explaining why those benchmarks are not executed or remove the code

adamsitnik:
@Anipik thanks for fixing all my comments. LGTM! :shipit:

build/ci/phase-template.yml
```diff
@@ -19,15 +19,19 @@ phases:
           Build_Debug:
             _configuration: Debug
             _config_short: D
+            _includeBenchmarkData: false
           Build_Release:
             _configuration: Release
             _config_short: R
+            _includeBenchmarkData: true
```
singlis: >_includeBenchmarkData [](start = 12, length = 21)

Does this always enable the benchmarking when building on release builds? Main concern is that I dont want to run benchmarks when I do a release build on my machine.

test/Microsoft.ML.Benchmarks/Helpers/DataPathProvider.cs
```diff
@@ -0,0 +1,17 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.IO;
+
+namespace Microsoft.ML.Benchmarks
+{
+    internal class DataPathProvider
+    {
+        private static string RootPath = Directory.GetParent(Directory.GetCurrentDirectory()).Parent.Parent.Parent.Parent.Parent.Parent.Parent.FullName;
```
singlis: Thats a lot of parents. Could this be a maintenance hassle if we change our directory structure? 

What is the runtime directory for this?

build/ci/phase-template.yml
Anipik: This downloads the data required for some of the benchmarks. It’s nothing has to do with running benchmarks.
The version of the benchmarks that we have enabled for release builds is a dry job version. Dry job version refers running these benchmarks just once. 
Currently we have just enabled dry job for benchmarks which doesn’t take more couple of seconds to run
If you want to actually run the benchmarks, then you will have to go into the benchmarks project and run and they won’t be run automatically

test/Microsoft.ML.Benchmarks/Helpers/DataPathProvider.cs
Anipik: we are already using the similar structure
https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.TestFramework/BaseTestClass.cs#L40

yes if we change build structure then we will have to change everything.

The current Runtime Directory is 
```C#
C:\git\Mlnet\bin\AnyCPU.Release-netfx\Microsoft.ML.Benchmarks.Tests\netcoreapp2.1\
```

eerhardt: These values are duplicated with [https://github.com/dotnet/machinelearning/blob/64b60d2805ba4ef1fd694c3211d170fdf0699c3b/test/Microsoft.ML.TestFramework/BaseTestClass.cs#L33-L44](https://github.com/dotnet/machinelearning/blob/64b60d2805ba4ef1fd694c3211d170fdf0699c3b/test/Microsoft.ML.TestFramework/BaseTestClass.cs#L33-L44).

Can we refactor this code so it is contained in 1 spot?

eerhardt:
>   <Import Project="$(RepoRoot)\build\ExternalBenchmarkDataFiles.props" /> 

I don't think you need this import anymore, since you are no longer using the data in the file.

---
Refers to: test/Microsoft.ML.Benchmarks/Microsoft.ML.Benchmarks.csproj:2 in 1adb502. [](commit_id = 1adb50244178597ece5e274fe528a0d9f80aa126, deletion_comment = False)

eerhardt:
>     <Content Include="..\data\iris.txt" Link="Input\iris.txt">

It's odd to me that some of the data files are copied to the output, but some aren't. Can we be consistent?

---
Refers to: test/Microsoft.ML.Benchmarks/Microsoft.ML.Benchmarks.csproj:30 in 1adb502. [](commit_id = 1adb50244178597ece5e274fe528a0d9f80aa126, deletion_comment = False)

adamsitnik:
> It's odd to me that some of the data files are copied to the output, but some aren't. Can we be consistent?

@eerhardt some of them are 6.5 GB big. I guess that this is why we have the difference

.vsts-dotnet-ci.yml
```diff
@@ -16,9 +16,11 @@ phases:
       Build_Debug_Intrinsics:
         _configuration: Debug-Intrinsics
         _config_short: DI
+        _includeBenchmarkData: false
```
eerhardt: Why are we setting these parameters here and in the phase-template?  I think they only need to be set in the phase-template, right?

Anipik: This parameter needs to be passed for each configuration.
In phase-template we are definining the structure of different phases and some default values eg. everything for release and debug configuration has been defined there.
But for netfx and netcoreapp3.0, we have a different build configuration than debug and release,
hence has to pass those build confuguration specific variable

Anipik:
> It's odd to me that some of the data files are copied to the output, but some aren't. Can we be consistent?

As adam said this is due to the fact that these are some large datasets and we are downloading them fom azure. although i can make the consistent for all the datasets i.e not copying the dataset to output folder.

build/ci/phase-template.yml
singlis: OK thanks for the explanation.

---
In reply to: [246990441](https://github.com/dotnet/machinelearning/pull/2109#discussion_r246990441) [](ancestors = 246990441)

singlis: Actually -- I just saw that the size of the datasets are 6.5 GB that we download from azure. In total how big is the download from Azure and how long does the download take? Does it always download or does it skip if the files are present?

---
In reply to: [247217818](https://github.com/dotnet/machinelearning/pull/2109#discussion_r247217818) [](ancestors = 247217818,246990441)

Anipik: It skips the files if they are already present.
It takes about 1.30 minutes to download the datasets on CI

test/Microsoft.ML.Benchmarks/Helpers/DataPathProvider.cs
Anipik: This is not possible as the currentAssembly locations are different for benchmark tests and other tests,
eg for all other tests, the currentAssemblyLocation = 
```"C:\\git\\machinelearning\\bin\\AnyCPU.Release\\Microsoft.ML.Benchmarks\\netcoreapp2.1\\Microsoft.ML.TestFramework.dll"```
for benchmarks it is equal to 
```"C:\\git\\machinelearning\\bin\\AnyCPU.Release\\Microsoft.ML.Benchmarks\\netcoreapp2.1\\0c2f3461-1faf-408d-9864-57e4a29ec2f4\\bin\\Release\\netcoreapp2.1\\Microsoft.ML.TestFramework.dll"```

**Comment by Anipik** [comment]
@eerhardt can you take a look at this ?

File: test/Microsoft.ML.Benchmarks/Helpers/DataPathProvider.cs (hunk 6 above)
Comment (eerhardt) [reply test/Microsoft.ML.Benchmarks/Helpers/DataPathProvider.cs]: Maybe a better approach here would be to navigate up until we hit the "repo root" directory. Then navigate down to `test\data`. Instead of hard-coding how many parent directories until the root.

See 
https://github.com/dotnet/cli/blob/6da8a92712d114af8131c2683c0c0f78436f6d80/test/Microsoft.DotNet.Tools.Tests.Utilities/RepoDirectoriesProvider.cs#L24-L57
or
https://github.com/dotnet/sdk/blob/67275f2a66b71e480fcc36b17a5b4b0e06f8e5da/src/Tests/Microsoft.NET.TestFramework/TestContext.cs#L156-L170 

for examples of how to do this.

File: test/Microsoft.ML.TestFramework/BaseTestClass.cs [hunk 8]
```diff
@@ -66,33 +85,34 @@ protected virtual void Cleanup()
         {
         }
 
-        protected string RootDir => _rootDir;
-        protected string OutDir => _outDir;
+        protected static string RootDir { get; }
```
eerhardt: I would name this `RepoRoot` or `RepoRootDir`. Just "RootDir" isn't very descriptive.

test/Microsoft.ML.Benchmarks.Tests/BenchmarksTest.cs
```diff
@@ -28,25 +28,30 @@ public class BenchmarkTouchingNativeDependency
...
-        public void BenchmarksProjectIsNotBroken()
+
+        [ConditionalTheory(typeof(BenchmarksTest), nameof(CanExecute))]
+        [InlineData(typeof(BenchmarkTouchingNativeDependency))]
+        [InlineData(typeof(CacheDataViewBench))]
+        [InlineData(typeof(HashBench))]
+        [InlineData(typeof(KMeansAndLogisticRegressionBench))]
+        [InlineData(typeof(PredictionEngineBench))]
+        [InlineData(typeof(RffTransformTrain))]
+        [InlineData(typeof(StochasticDualCoordinateAscentClassifierBench))]
```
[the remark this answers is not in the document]
Anipik: > Would it be better to load the assembly and look for classes?

Some of the benchmarks takes a long time to execute and we are not running them as a part of CI. We can introduce a attribute similar to the ```[OuterLoop]``` to indicate which benchmarks not to run on CI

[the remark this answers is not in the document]
eerhardt: > and we are not running them as a part of CI. 

I assume we are "OK" with breaking those tests from time-to-time then?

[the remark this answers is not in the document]
Anipik: its a trade off, if we include them we will have to increase our CI time from 45 to 60-75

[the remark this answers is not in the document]
eerhardt: 45?? Right now each leg takes around 20 min or less.

https://dev.azure.com/dnceng/public/_build/results?buildId=72181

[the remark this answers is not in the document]
Anipik: Yes our timeout limit is 45 minutes and its not enough with benchmarks, we will have to increase that limit to 75

test/Microsoft.ML.Benchmarks.Tests/BenchmarksTest.cs
```diff
@@ -41,14 +38,24 @@ public class BenchmarksTest
             Environment.Is64BitProcess; // we don't support 32 bit yet
 #endif
 
+        public static TheoryData<Type> GetBenchmarks()
+        {
+            TheoryData<Type> benchmarks = new TheoryData<Type>();
+            Assembly asm = Assembly.LoadFrom("Microsoft.ML.Benchmarks.dll");
```
eerhardt: Can you use `typeof(StochasticDualCoordinateAscentClassifierBench).Assembly` instead?

test/Microsoft.ML.Benchmarks/Harness/CIBenchmark.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Microsoft.ML.Benchmarks.Harness
+{
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    public class CIBenchmark : Attribute
```
eerhardt: It might be a good idea to add a `<summary>` comment to this attribute explaining its purpose.

test/Microsoft.ML.Benchmarks/Numeric/Ranking.cs
```diff
@@ -21,8 +21,8 @@ public class RankingTrain
         [GlobalSetup]
         public void SetupTrainingSpeedTests()
         {
-            _mslrWeb10k_Validate = Path.GetFullPath(TestDatasets.MSLRWeb.validFilename);
-            _mslrWeb10k_Train = Path.GetFullPath(TestDatasets.MSLRWeb.trainFilename);
+            _mslrWeb10k_Validate = TestFramework.BaseTestClass.GetDataPath(TestDatasets.MSLRWeb.validFilename);
```
eerhardt: Can't this just be `BaseTestClass.GetDataPath`? Why do you need to qualify it with `TestFramwork.`?

test/Microsoft.ML.Benchmarks.Tests/BenchmarksTest.cs
Anipik: yes sure i will do that but is there any specific reason behind this ?

test/Microsoft.ML.Benchmarks/Numeric/Ranking.cs
Anipik: It was done to just avoid adding a reference to ```using Microsoft.ML.Runtests``` statement.
i dont have a preference. so should i go with ```BaseTestClass.GetDataPath``` ?

test/Microsoft.ML.Benchmarks.Tests/BenchmarksTest.cs
eerhardt: The reason is because the assembly is already loaded, since we have a reference to it. Using a strongly-typed dependency is better because it explicitly shows the dependency, and if someone removed the reference (or renamed the assembly), the code would no longer compile. Also, it is more performant because Assembly.LoadFrom has to interpret the string, read the file, etc. to decide that the assembly is already loaded.

PR #2110, opened 2019-01-10
fixing the AdultData LoadColumn, and IrisWithAllFeatures class.

sfilipi:
Fixing errors regarding data loading,  spotted in the cookbook.

### DIFF
--- a/docs/code/MlNetCookBook.md
+++ b/docs/code/MlNetCookBook.md
@@ -285,11 +285,11 @@ Or by creating a data model for it:
 ```csharp
 private class AdultData
 {
-    [LoadColumn("0", "10"), ColumnName("Features")]
-    public float FeatureVector { get; }
+    [LoadColumn(0, 10), ColumnName("Features")]
+    public float FeatureVector { get; set;}
 
     [LoadColumn(11)]
-    public float Target { get; }
+    public float Target { get; set;}
 }
 
 // Create a new context for ML.NET operations. It can be used for exception tracking and logging, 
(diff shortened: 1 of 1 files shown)


docs/code/MlNetCookBook.md
```diff
@@ -285,11 +285,11 @@ Or by creating a data model for it:
 ```csharp
 private class AdultData
 {
-    [LoadColumn("0", "10"), ColumnName("Features")]
-    public float FeatureVector { get; }
+    [LoadColumn(0, 10), ColumnName("Features")]
+    public float FeatureVector { get; set;}
```
wschin: Maybe just
```suggestion
    public float FeatureVector;
```
?

docs/code/MlNetCookBook.md
```diff
@@ -285,11 +285,11 @@ Or by creating a data model for it:
...
 
     [LoadColumn(11)]
-    public float Target { get; }
+    public float Target { get; set;}
```
wschin: Same here. Not sure if having `get` and `set` is a convention but I'd remove them for shorter bug equivalent code.
```suggestion
    public float Target;
```

docs/code/MlNetCookBook.md
sfilipi: I'll double-check. I believe it has to be a property, not a field.

shauheen: >float [](start = 11, length = 5)

I don't think this would work.

sfilipi:
merging this to make the doc at pair with the code in the tests, and taking on another pr to refactor the cookbook and verify the examples

PR #2111, opened 2019-01-10
Say good-bye to transpose ISchema

wschin:
We remove `ITransposeSchema` and move `GetSlotType` to `TransposeSlotTypes` field in `ITransposeDataView`. Several derived classes' `ISchema` members are removed because they are not used at all.

Party of #1501 continues.

### DIFF
--- a/src/Microsoft.ML.Data/Data/ITransposeDataView.cs
+++ b/src/Microsoft.ML.Data/Data/ITransposeDataView.cs
@@ -20,18 +20,17 @@ namespace Microsoft.ML.Data
     /// <see cref="IDataView"/>.
     ///
     /// The interface only advertises that columns may be accessible in slot-wise fashion. A column
-    /// is accessible in this fashion iff <see cref="TransposeSchema"/>'s
-    /// <see cref="ITransposeSchema.GetSlotType"/> returns a non-null value.
+    /// is accessible in this fashion iff <see cref="TransposeSlotTypeHolder"/>'s
+    /// <see cref="ITransposeSlotTypeHolder.GetSlotType"/> returns a non-null value.
     /// </summary>
     [BestFriend]
     internal interface ITransposeDataView : IDataView
     {
         /// <summary>
         /// An enhanced schema, containing information on the transposition properties, if any,
-        /// of each column. Note that there is no contract or suggestion that this property
-        /// should be equal to <see cref="IDataView.Schema"/>.
+        /// of each column.
         /// </summary>
-        ITransposeSchema TransposeSchema { get; }
+        ITransposeSlotTypeHolder TransposeSlotTypeHolder { get; }
 
         /// <summary>
         /// Presents a cursor over the slots of a transposable column, or throws if the column
@@ -44,15 +43,15 @@ namespace Microsoft.ML.Data
     /// The transpose schema returns the schema information of the view we have transposed.
     /// </summary>
     [BestFriend]
-    internal interface ITransposeSchema : ISchema
+    internal interface ITransposeSlotTypeHolder
     {
         /// <summary>
-        /// Analogous to <see cref="ISchema.GetColumnType"/>, except instead of returning the type of value
-        /// accessible through the <see cref="RowCursor"/>, returns the item type of value accessible
-        /// through the <see cref="SlotCursor"/>. This will return <c>null</c> iff this particular
-        /// column is not transposable, that is, it cannot be viewed in a slotwise fashion. Observe from
-        /// the return type that this will always be a vector type. This vector type should be of fixed
-        /// size and one dimension.
+        /// Analogous to <see cref="ColumnType"/> in <see cref="Schema"/>, except instead of returning
+        /// the type of value accessible through the <see cref="RowCursor"/>, returns the item type of
+        /// value accessible through the <see cref="SlotCursor"/>. This will return <c>null</c> iff
+        /// this particular column is not transposable, that is, it cannot be viewed in a slotwise fashion.
+        /// Observe from the return type that this will always be a vector type. This vector type should
+        /// be of fixed size and one dimension.
         /// </summary>
         VectorType GetSlotType(int col);
     }
--- a/src/Microsoft.ML.Data/Data/ITransposeDataView.cs
+++ b/src/Microsoft.ML.Data/Data/ITransposeDataView.cs
@@ -20,8 +20,7 @@ namespace Microsoft.ML.Data
     /// way of accessing the data stored in a <see cref="IDataView"/>.
     ///
     /// The interface only advertises that columns may be accessible in slot-wise fashion. The i-th column
-    /// is accessible in this fashion iff <see cref="TransposeSlotTypes"/>[i] is not <see langword="null"/>
-    /// and <see cref="TransposeSlotTypes"/>[i] is not <see langword="null"/>.
+    /// is accessible in this fashion iff <see cref="GetSlotType"/> with col=i doesn't return <see langword="null"/>.
     /// </summary>
     [BestFriend]
     internal interface ITransposeDataView : IDataView
@@ -33,11 +32,12 @@ namespace Microsoft.ML.Data
         SlotCursor GetSlotCursor(int col);
 
         /// <summary>
-        /// <see cref="TransposeSlotTypes"/>[i] specifies the type of all values at the i-th column of <see cref="IDataView"/>.
-        /// For example, if <see cref="IDataView.Schema"/>[i] is a scalar float column, then <see cref="TransposeSlotTypes"/>[i]
-        /// may return a <see cref="VectorType"/> whose <see cref="VectorType.ItemType"/> field is <see cref="NumberType.R4"/>.
-        /// If the i-th column can't be iterated column-wisely, <see cref="TransposeSlotTypes"/>[i] may be <see langword="null"/>.
+        /// <see cref="GetSlotType"/> (input argument is named col) specifies the type of all values at the col-th column of
+        /// <see cref="IDataView"/>.  For example, if <see cref="IDataView.Schema"/>[i] is a scalar float column, then
+        /// <see cref="GetSlotType"/> with col=i may return a <see cref="VectorType"/> whose <see cref="VectorType.ItemType"/>
+        /// field is <see cref="NumberType.R4"/>. If the i-th column can't be iterated column-wisely, this function may
+        /// return <see langword="null"/>.
         /// </summary>
-        VectorType[] TransposeSlotTypes { get; }
+        VectorType GetSlotType(int col);
     }
 }
--- a/src/Microsoft.ML.Data/Data/ITransposeDataView.cs
+++ b/src/Microsoft.ML.Data/Data/ITransposeDataView.cs
@@ -14,45 +14,30 @@ namespace Microsoft.ML.Data
     /// <summary>
     /// A view of data where columns can optionally be accessed slot by slot, as opposed to row
     /// by row in a typical dataview. A slot-accessible column can be accessed with a slot-by-slot
-    /// cursor via an <see cref="SlotCursor"/> (naturally, as opposed to row-by-row through an
-    /// <see cref="RowCursor"/>). This interface is intended to be implemented by classes that
-    /// want to provide an option for an alternate way of accessing the data stored in a
-    /// <see cref="IDataView"/>.
+    /// cursor via an <see cref="SlotCursor"/> returned by <see cref="GetSlotCursor(int)"/>
+    /// (naturally, as opposed to row-by-row through an <see cref="RowCursor"/>). This interface
+    /// is intended to be implemented by classes that want to provide an option for an alternate
+    /// way of accessing the data stored in a <see cref="IDataView"/>.
     ///
-    /// The interface only advertises that columns may be accessible in slot-wise fashion. A column
-    /// is accessible in this fashion iff <see cref="TransposeSlotTypeHolder"/>'s
-    /// <see cref="ITransposeSlotTypeHolder.GetSlotType"/> returns a non-null value.
+    /// The interface only advertises that columns may be accessible in slot-wise fashion. The i-th column
+    /// is accessible in this fashion iff <see cref="TransposeSlotTypes"/>[i] is not <see langword="null"/>
+    /// and <see cref="TransposeSlotTypes"/>[i] is not <see langword="null"/>.
     /// </summary>
     [BestFriend]
     internal interface ITransposeDataView : IDataView
     {
         /// <summary>
-        /// An enhanced schema, containing information on the transposition properties, if any,
-        /// of each column.
-        /// </summary>
-        ITransposeSlotTypeHolder TransposeSlotTypeHolder { get; }
-
-        /// <summary>
         /// Presents a cursor over the slots of a transposable column, or throws if the column
         /// is not transposable.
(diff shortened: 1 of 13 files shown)


src/Microsoft.ML.Data/DataView/Transposer.cs
```diff
@@ -1559,58 +1534,26 @@ private void GetId(ref RowId id)
...
-            public ColumnType GetMetadataTypeOrNull(string kind, int col)
-            {
-                return _schema[col].Metadata.Schema.GetColumnOrNull(kind)?.Type;
-            }
-
-            public void GetMetadata<TValue>(string kind, int col, ref TValue value)
-            {
-                _schema[col].Metadata.GetValue(kind, ref value);
-            }
         }
```
Ivanidzo4ka: This one is odd one.
Why I need schema object if all I care is schema.count?
Can you fix summary, it's no longer related to ISchema.
Maybe call it EmptySlotTypeHolder?

wschin: Fixed.

src/Microsoft.ML.Data/Data/ITransposeDataView.cs
```diff
@@ -44,15 +43,15 @@ internal interface ITransposeDataView : IDataView
     /// The transpose schema returns the schema information of the view we have transposed.
     /// </summary>
     [BestFriend]
-    internal interface ITransposeSchema : ISchema
+    internal interface ITransposeSlotTypeHolder
```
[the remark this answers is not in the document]
TomFinley: Putting this method `GetSlotType` directly on the `ITransposeDataView` itself, and destroying the former `ITransposeSchema` seems a good plan to me. For the purpose of a public API, we may need to think more carefully about this, but we'll have to think very carefully about that anyway.

---
In reply to: [246979206](https://github.com/dotnet/machinelearning/pull/2111#discussion_r246979206) [](ancestors = 246979206)

TomFinley:
>     public abstract class SlotCursor : IDisposable

Not your change, but just something I note... Seems like so much else of the transpose data view is internal, I have to imagine that if we `[BestFriend] internal` this nothing bad would happen. Is it not so? If so, we should probably do so.

---
Refers to: src/Microsoft.ML.Data/Data/SlotCursor.cs:13 in 01c0598. [](commit_id = 01c059842b8da85a36b5d769daf5b81c91665806, deletion_comment = False)

eerhardt:
@wschin - a way to make a public class implement an internal interface and not make the methods public, is to explicitly implement the interface. 
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation

wschin:
@eerhardt, I have used that pattern on `TransposeSlotTypes` in `ITransposeDataView`. Are you saying I should do the same to make `SlotCursor` not externally-accessible?

test/Microsoft.ML.Predictor.Tests/TestTransposer.cs
```diff
@@ -199,9 +199,9 @@ public void TransposerTest()
             using (Transposer trans = Transposer.Create(Env, view, true, 3, 5, 4))
             {
                 // Check to see that A, B, and C were not transposed somehow.
-                Assert.Null(trans.TransposeSchema.GetSlotType(0));
-                Assert.Null(trans.TransposeSchema.GetSlotType(1));
-                Assert.Null(trans.TransposeSchema.GetSlotType(2));
+                Assert.Null((trans as ITransposeDataView).TransposeSlotTypes[0]);
```
eerhardt: Can you just do the cast 1 time instead of 3 times in a row?

eerhardt:
> Are you saying I should do the same to make SlotCursor not externally-accessible?

Yes. If we want to make `SlotCursor` internal, you can use explicitly implemented interfaces so all the classes that implement `ITransposeDataView` can be public, but don't need to expose a `public SlotCursor GetSlotCursor(int col)` method. So if you make `SlotCursor` internal, and then explicitly implement `ITransposeDataView.GetSlotCursor` on all the implementations, you should be able to compile successfully.

src/Microsoft.ML.Data/Data/ITransposeDataView.cs
```diff
@@ -14,46 +14,30 @@ namespace Microsoft.ML.Data
...
-        /// column is not transposable, that is, it cannot be viewed in a slotwise fashion. Observe from
-        /// the return type that this will always be a vector type. This vector type should be of fixed
-        /// size and one dimension.
+        /// <see cref="TransposeSlotTypes"/>[i] specifies the type of all values at the i-th column of <see cref="IDataView"/>.
+        /// For example, if <see cref="IDataView.Schema"/>[i] is a scalar float column, then <see cref="TransposeSlotTypes"/>[i]
+        /// may return a <see cref="VectorType"/> whose <see cref="VectorType.ItemType"/> field is <see cref="NumberType.R4"/>.
+        /// If the i-th column can't be iterated column-wisely, <see cref="TransposeSlotTypes"/>[i] may be <see langword="null"/>.
         /// </summary>
-        VectorType GetSlotType(int col);
+        VectorType[] TransposeSlotTypes { get; }
```
TomFinley: >VectorType[] [](start = 8, length = 12)

I would prefer the old method, if possible. However, if you insist on replacing it with a collection, I'd prefer that it be something not mutable. Like, say, a read only list, or an `System.Collections.Immutable` array, or something like this. Having what is essentially part of the schema be a mutable structure is not an acceptable state even for something internal.

TomFinley:
>     public abstract class SlotCursor : IDisposable

Remember explicit interface implementations.

---
In reply to: [455007710](https://github.com/dotnet/machinelearning/pull/2111#issuecomment-455007710) [](ancestors = 455007710,454099677)

---
Refers to: src/Microsoft.ML.Data/Data/SlotCursor.cs:13 in d693ad9. [](commit_id = d693ad958e16691bd1bd0309c721f8d6200c0f67, deletion_comment = False)

src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
```diff
@@ -582,10 +586,10 @@ public RowCursor[] GetRowCursorSet(Func<int, bool> predicate, int n, Random rand
             return View.GetRowCursorSet(predicate, n, rand);
         }
 
-        public SlotCursor GetSlotCursor(int col)
+        SlotCursor ITransposeDataView.GetSlotCursor(int col)
```
TomFinley: >SlotCursor ITransposeDataView.GetSlotCursor [](start = 8, length = 43)

Ah, I see you figured out the explicit interface implementations after all. :) Good work.

src/Microsoft.ML.Data/Data/ITransposeDataView.cs
```diff
@@ -14,46 +14,30 @@ namespace Microsoft.ML.Data
...
-        /// Analogous to <see cref="ISchema.GetColumnType"/>, except instead of returning the type of value
-        /// accessible through the <see cref="RowCursor"/>, returns the item type of value accessible
-        /// through the <see cref="SlotCursor"/>. This will return <c>null</c> iff this particular
-        /// column is not transposable, that is, it cannot be viewed in a slotwise fashion. Observe from
-        /// the return type that this will always be a vector type. This vector type should be of fixed
-        /// size and one dimension.
+        /// <see cref="TransposeSlotTypes"/>[i] specifies the type of all values at the i-th column of <see cref="IDataView"/>.
+        /// For example, if <see cref="IDataView.Schema"/>[i] is a scalar float column, then <see cref="TransposeSlotTypes"/>[i]
+        /// may return a <see cref="VectorType"/> whose <see cref="VectorType.ItemType"/> field is <see cref="NumberType.R4"/>.
+        /// If the i-th column can't be iterated column-wisely, <see cref="TransposeSlotTypes"/>[i] may be <see langword="null"/>.
```
TomFinley: >cref="TransposeSlotTypes"/>[i] may be <see langword="null"/> [](start = 69, length = 60)

Also apparently, from what I see of the usage, even the array *itself* might be `null`. The more I read of the new code the more I think we got it right the first time with a method, not exposing some collection that we have to test whether it is null or not before we even think about querying... yeah.

test/Microsoft.ML.Predictor.Tests/TestTransposer.cs
```diff
@@ -199,9 +199,10 @@ public void TransposerTest()
             using (Transposer trans = Transposer.Create(Env, view, true, 3, 5, 4))
             {
                 // Check to see that A, B, and C were not transposed somehow.
-                Assert.Null(trans.TransposeSchema.GetSlotType(0));
-                Assert.Null(trans.TransposeSchema.GetSlotType(1));
-                Assert.Null(trans.TransposeSchema.GetSlotType(2));
+                var itdv = trans as ITransposeDataView;
```
TomFinley: >trans as ITransposeDataView [](start = 27, length = 27)

When we are *certain* that something implements an interface, please just use a `()` style cast. I notice you are not testing it, so, you are not sure.

Look at it this way... If you happen to be wrong with `()`, you will get a cast error. If you happen to be wrong with `as`, you will get a null exception... which could also happen because apparently it is legal for `.TransposeSlotTypes` to be `null` (to judge from other code), and you did not test for that. So it is more direct.

src/Microsoft.ML.FastTree/GamTrainer.cs
```diff
@@ -268,7 +268,8 @@ private bool UseTranspose(bool? useTranspose, RoleMappedData data)
 
             if (useTranspose.HasValue)
                 return useTranspose.Value;
-            return data.Data is ITransposeDataView td && td.TransposeSchema.GetSlotType(data.Schema.Feature.Value.Index) != null;
+            return data.Data is ITransposeDataView td && td.TransposeSlotTypes != null
+                && td.TransposeSlotTypes[data.Schema.Feature.Value.Index] != null;
```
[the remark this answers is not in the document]
wschin: You're right. Will change that array back to method in the next iteration.

---
In reply to: [248788605](https://github.com/dotnet/machinelearning/pull/2111#discussion_r248788605) [](ancestors = 248788605)

test/Microsoft.ML.Predictor.Tests/TestTransposer.cs
wschin: Ok. I also checked some other places by finding references to `ITransposeDataView` in VS.

---
In reply to: [248781623](https://github.com/dotnet/machinelearning/pull/2111#discussion_r248781623) [](ancestors = 248781623)

src/Microsoft.ML.Data/Data/ITransposeDataView.cs
wschin: Just replaced this with a function, `GetSlotType(int col)`. Many thanks.

---
In reply to: [248780062](https://github.com/dotnet/machinelearning/pull/2111#discussion_r248780062) [](ancestors = 248780062)

src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
wschin: Thanks to your and Eric's comments.

---
In reply to: [248778138](https://github.com/dotnet/machinelearning/pull/2111#discussion_r248778138) [](ancestors = 248778138)

src/Microsoft.ML.Data/Data/ITransposeDataView.cs
wschin: Your wish is granted.

---
In reply to: [248777208](https://github.com/dotnet/machinelearning/pull/2111#discussion_r248777208) [](ancestors = 248777208)

test/Microsoft.ML.Predictor.Tests/TestTransposer.cs
wschin: We have only one `(ITransposeDataView)trans` now.

---
In reply to: [248704226](https://github.com/dotnet/machinelearning/pull/2111#discussion_r248704226) [](ancestors = 248704226)

wschin:
>     public abstract class SlotCursor : IDisposable

No problem!

---
In reply to: [455268001](https://github.com/dotnet/machinelearning/pull/2111#issuecomment-455268001) [](ancestors = 455268001,455007710,454099677)

---
Refers to: src/Microsoft.ML.Data/Data/SlotCursor.cs:13 in d693ad9. [](commit_id = d693ad958e16691bd1bd0309c721f8d6200c0f67, deletion_comment = False)

wschin:
Thanks to your guidance. `SlotCursor` is invisible from outside now.

---
In reply to: [455205205](https://github.com/dotnet/machinelearning/pull/2111#issuecomment-455205205) [](ancestors = 455205205)

src/Microsoft.ML.Data/Data/ITransposeDataView.cs
TomFinley: Cool, I have two left I think!

---
In reply to: [248815747](https://github.com/dotnet/machinelearning/pull/2111#discussion_r248815747) [](ancestors = 248815747,248777208)

src/Microsoft.ML.Data/DataLoadSave/Transpose/TransposeLoader.cs
```diff
@@ -378,6 +364,17 @@ private bool HasRowData
...
+        internal const string Summary = "Loads a binary transposed data file.";
+        internal const string LoadName = "TransposeLoader";
+
+        // We return the schema view's schema, because we don't necessarily want
+        // something that can be cast to a transpose schema, and also because the
+        // transpose schema is defined after the entries have been read, which
+        // inspect the schema. We also want to ensure that the useful property that
+        // a cursor and view's schemas are the same, is preserved, which allows us
+        // to use the cursors from the schema view if convenient to do so.
+        public Schema Schema => _schemaEntry.GetView().Schema;
```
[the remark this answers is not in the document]
wschin: It should be very minor because most of our time is spent on training something.

src/Microsoft.ML.Data/DataLoadSave/Transpose/TransposeLoader.cs
```diff
@@ -845,7 +793,7 @@ private void Init(int col)
                 Ch.Assert(0 <= col && col < Schema.Count);
                 Ch.Assert(_colToActivesIndex[col] >= 0);
                 var type = Schema[col].Type;
-                Ch.Assert(((ITransposeDataView)_parent).TransposeSchema.GetSlotType(col).GetValueCount() == _parent._header.RowCount);
+                Ch.Assert((type as VectorType).GetValueCount() == _parent._header.RowCount);
```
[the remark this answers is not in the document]
wschin: Fixed. Thanks.

test/Microsoft.ML.Core.Tests/UnitTests/TestHosts.cs
```diff
@@ -83,7 +83,7 @@ public void LogEventProcessesMessages()
             env.Log += (sender, e) => messages.Add(e.Message);
 
             // create a dummy text reader to trigger log messages
-            env.Data.CreateTextReader(
+            env.Data.CreateTextLoader(
```
eerhardt: We should get this change in soon since the build is currently broken. If we can't check this in right away, let's get a separate PR out with just fixing the build.

/cc @najeeb-kazmi

PR #2115, opened 2019-01-10
Adds string support for ValueMappingEstimator

### LINKED ISSUE #2114: ValueMappingTransform does not handle type string for keys and values
ValueMappingTransform does handle ReadOnlyMemory<char> for values and keys, it does not handle strings. 

This should be updated as users will naturally want to use strings for the data type. With this support, the following should work:
```
            var educationKeys = new List<string>()
            {
                "0-5yrs",
                "6-11yrs",
                "12+yrs"
            };

            var educationValues = new List<string>()
            {
                "Cat1",
                "Cat2", 
                "Cat3"
            };

            var pipeline = new ValueMappingEstimator<string, string>(ml, educationKeys, educationValues, ("Education", "EducationCategory"));
```

singlis:
This adds support for strings to be used for keys and values for the ValueMappingEstimator (as opposed to ReadOnlyMemory<char>).

This also converts all tests to use string types as well as adds an
additional test for string vector values.

Fixes #2114


src/Microsoft.ML.Data/Transforms/ValueMappingTransformer.cs
```diff
@@ -165,8 +182,9 @@ private static ValueGetter<VBuffer<ReadOnlyMemory<char>>> GetKeyValueGetter<TKey
             var keyType = GetPrimitiveType(typeof(TKey), out bool isKeyVectorType);
             var valueType = GetPrimitiveType(typeof(TValue), out bool isValueVectorType);
             var dataViewBuilder = new ArrayDataViewBuilder(env);
-            dataViewBuilder.AddColumn(keyColumnName, keyType, keys.ToArray());
-            dataViewBuilder.AddColumn(valueColumnName, valueType, values.ToArray());
+            AddColumnWrapper(dataViewBuilder, keyColumnName, keyType, keys.ToArray());
+            AddColumnWrapper(dataViewBuilder, valueColumnName, valueType, values.ToArray());
+            //dataViewBuilder.AddColumn(valueColumnName, valueType, values.ToArray());
```
Ivanidzo4ka: >//dataViewBuilder.AddColumn(valueColumnName, valueType, values.ToArray()); [](start = 12, length = 74)

can be removed.

src/Microsoft.ML.Data/Transforms/ValueMappingTransformer.cs
```diff
@@ -165,8 +182,9 @@ private static ValueGetter<VBuffer<ReadOnlyMemory<char>>> GetKeyValueGetter<TKey
             var keyType = GetPrimitiveType(typeof(TKey), out bool isKeyVectorType);
             var valueType = GetPrimitiveType(typeof(TValue), out bool isValueVectorType);
             var dataViewBuilder = new ArrayDataViewBuilder(env);
-            dataViewBuilder.AddColumn(keyColumnName, keyType, keys.ToArray());
```
Ivanidzo4ka: >keyType [](start = 53, length = 7)

in old code if you pass TKey as string, keyType for it would be ReadOnlyMemory<char>, why you need what wrapping around?

src/Microsoft.ML.Data/Transforms/ValueMappingTransformer.cs
```diff
@@ -153,6 +153,23 @@ private static ValueGetter<VBuffer<ReadOnlyMemory<char>>> GetKeyValueGetter<TKey
                 };
         }
 
+        internal static void AddColumnWrapper<T>(ArrayDataViewBuilder builder, string columnName, PrimitiveType primitiveType, T[] values)
+        {
+            if (typeof(T) == typeof(string))
+                builder.AddColumn(columnName, values.Select(x=>x.ToString()).ToArray());
```
Ivanidzo4ka: >values.Select(x=>x.ToString() [](start = 46, length = 29)

 `T[] values`
`typeof(T) == typeof(string)`

why you need to call ToString() on string?

src/Microsoft.ML.Data/Transforms/ValueMappingTransformer.cs
```diff
@@ -153,6 +153,23 @@ private static ValueGetter<VBuffer<ReadOnlyMemory<char>>> GetKeyValueGetter<TKey
...
+            else
+                builder.AddColumn(columnName, primitiveType, values);
+        }
+
+        internal static void AddColumnWrapper<T>(ArrayDataViewBuilder builder, string columnName, PrimitiveType primitiveType, T[][] values)
+        {
+            if (typeof(T) == typeof(string))
+
+                builder.AddColumn(columnName, primitiveType, values.Select(x => x.Select( y => y.ToString().AsMemory() ).ToArray()).ToArray());
```
wschin: >x => x.Select( y => y.ToString().AsMemory() ).ToArray()) [](start = 75, length = 56)

Please try avoiding recursive structure in one line.

src/Microsoft.ML.Data/Transforms/ValueMappingTransformer.cs
```diff
@@ -153,6 +153,23 @@ private static ValueGetter<VBuffer<ReadOnlyMemory<char>>> GetKeyValueGetter<TKey
                 };
         }
 
+        internal static void AddColumnWrapper<T>(ArrayDataViewBuilder builder, string columnName, PrimitiveType primitiveType, T[] values)
+        {
```
wschin: Doc string please.

src/Microsoft.ML.Data/Transforms/ValueMappingTransformer.cs
```diff
@@ -153,6 +153,23 @@ private static ValueGetter<VBuffer<ReadOnlyMemory<char>>> GetKeyValueGetter<TKey
...
+            if (typeof(T) == typeof(string))
+                builder.AddColumn(columnName, values.Select(x=>x.ToString()).ToArray());
+            else
+                builder.AddColumn(columnName, primitiveType, values);
+        }
+
+        internal static void AddColumnWrapper<T>(ArrayDataViewBuilder builder, string columnName, PrimitiveType primitiveType, T[][] values)
+        {
```
wschin: Doc string!

src/Microsoft.ML.Data/Transforms/ValueMappingTransformer.cs
```diff
@@ -153,6 +153,23 @@ private static ValueGetter<VBuffer<ReadOnlyMemory<char>>> GetKeyValueGetter<TKey
                 };
         }
 
+        internal static void AddColumnWrapper<T>(ArrayDataViewBuilder builder, string columnName, PrimitiveType primitiveType, T[] values)
+        {
+            if (typeof(T) == typeof(string))
```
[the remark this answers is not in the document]
singlis: Yes - ValueMappingEstimator will continue to work with ReadOnlyMemory<char> with these changes.

---
In reply to: [247216613](https://github.com/dotnet/machinelearning/pull/2115#discussion_r247216613) [](ancestors = 247216613)

src/Microsoft.ML.Data/Transforms/ValueMappingTransformer.cs
singlis: Because I need to convert to a memory stream with AsMemory, which  needs to operate on a string type. .ToString ensures that whatever type T is, it will be done on a string.

---
In reply to: [246978062](https://github.com/dotnet/machinelearning/pull/2115#discussion_r246978062) [](ancestors = 246978062)

src/Microsoft.ML.Data/Transforms/ValueMappingTransformer.cs
singlis: Unfortunately the old code did not work if TKey was of string type. 

So here is what happens:
ArrayDataViewBuilder.AddColumn<T> will fail if T is of type string. The reason for this is for two reasons:
1) The PrimtiveType.RawType is a ReadonlyMemory<char> which does not match type of string
2) The string values are not converted to ReadOnlyMemory. 

ArrayDataViewBuilder.AddColumn does have an implementation that takes in a string, but that is not called when you are calling it from a function that is using a generic argument. So I added this wrapping that will check if the type is string, if so, convert to ReadOnlyMemoryChar and pass that to the builder.


---
In reply to: [246977957](https://github.com/dotnet/machinelearning/pull/2115#discussion_r246977957) [](ancestors = 246977957)

src/Microsoft.ML.Data/Transforms/ValueMappingTransformer.cs
singlis: Its now expanded into a few more lines...

---
In reply to: [247217068](https://github.com/dotnet/machinelearning/pull/2115#discussion_r247217068) [](ancestors = 247217068)

src/Microsoft.ML.Data/Transforms/ValueMappingTransformer.cs
sfilipi: but this condition will get hit only if typeof(T) == typeof(string); so you know for sure values is a string.. 

---
In reply to: [247232387](https://github.com/dotnet/machinelearning/pull/2115#discussion_r247232387) [](ancestors = 247232387,246978062)

singlis:
@Ivanidzo4ka and @sfilipi the code has been updated and there are a few open issues that I have responded to. Can you take a look when you have a chance?

src/Microsoft.ML.Data/Transforms/ValueMappingTransformer.cs
sfilipi: i think that the if clause, will not execute, if T is not a string, it will fall to else.. 

---
In reply to: [247605115](https://github.com/dotnet/machinelearning/pull/2115#discussion_r247605115) [](ancestors = 247605115,247304740,247232387,246978062)

singlis: I don't quite follow if you are saying that falling to the else is a problem or if the compiler should deduce that T is of type string.

Overall I am wanting to handle the case where T is of type string. Everything else should fall to the else clause as we would handle it as-is. In the case when T is of type string, we still have to convert T to a string because I am not able to call .AsMemory on a T object.  Therefore I am ensuring it is a string type by calling toString (hopefully calling toString on a string object is a no-op -- and this will only happen on strings because of the if-check). The other option is that I can cast T to a string if that is more clear.

Does that address your concern? Or am I misunderstanding?


---
In reply to: [249960836](https://github.com/dotnet/machinelearning/pull/2115#discussion_r249960836) [](ancestors = 249960836,247605115,247304740,247232387,246978062)

PR #2116, opened 2019-01-10
WIP Code coverage support

codemzs:
This is a WIP that adds code coverage support. This change contains code that generates code coverage files. In the next iteration I will add code to upload code coverage files to CodeCov once I know builds pass.

### DIFF
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -31,6 +31,7 @@
     <PackageReference Include="xunit" Version="2.4.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
     <PackageReference Include="Microsoft.DotNet.XUnitExtensions" Version="2.4.0-prerelease-63213-02" />
+    <PackageReference Include="coverlet.msbuild" Version="2.5.0" />
   </ItemGroup>
 
 </Project>
--- a/test/Directory.Build.targets
+++ b/test/Directory.Build.targets
@@ -1,5 +1,13 @@
 <Project>
 
   <Import Project="..\Directory.Build.targets" />
-
+  
+  <PropertyGroup Condition="'$(Coverage)' == 'true'">
+      <CollectCoverage>true</CollectCoverage>
+      <CoverletOutputFormat>opencover</CoverletOutputFormat>
+      <CoverletOutput>$(MSBuildThisFileDirectory)..\bin\coverage\$(AssemblyName).xml</CoverletOutput>
+      <Include>[Microsoft.ML.*]*</Include>
+      <ExcludeByFile>$(MSBuildThisFileDirectory)**\*.*</ExcludeByFile>
+  </PropertyGroup>
+  
 </Project>
--- a/test/run-tests.proj
+++ b/test/run-tests.proj
@@ -10,7 +10,7 @@
     <Message Importance="High" Text="Running tests ..." />
     <MSBuild Targets="VSTest"
              Projects="@(Project)"
-             Properties="VSTestNoBuild=true" />
+             Properties="VSTestNoBuild=true;Coverage=true" />
   </Target>
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), Directory.Build.targets))\Directory.Build.targets" />


test/run-tests.proj
```diff
@@ -10,7 +10,7 @@
     <Message Importance="High" Text="Running tests ..." />
     <MSBuild Targets="VSTest"
              Projects="@(Project)"
-             Properties="VSTestNoBuild=true" />
+             Properties="VSTestNoBuild=true;Coverage=true;CopyLocalLockFileAssemblies=true" />
```
sharwell: > CopyLocalLockFileAssemblies=true

Is this not a duplicate of the value in **Directory.Build.targets** set whenever `Coverage` is true?

test/Directory.Build.targets
```diff
@@ -1,5 +1,14 @@
...
   <Import Project="..\Directory.Build.targets" />
-
+  
+  <PropertyGroup Condition="'$(Coverage)' == 'true'">
+      <CollectCoverage>true</CollectCoverage>
+      <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+      <CoverletOutputFormat>opencover</CoverletOutputFormat>
+      <CoverletOutput>$(MSBuildThisFileDirectory)..\bin\coverage\$(AssemblyName).xml</CoverletOutput>
+      <Include>[Microsoft.ML.*]*</Include>
+      <ExcludeByFile>$(MSBuildThisFileDirectory)**\*.*</ExcludeByFile>
```
sharwell: 💡 Coverage in test files is typically a good thing, for the same overall reasons why coverage in production code is good.

```suggestion
```

test/run-tests.proj
```diff
@@ -10,7 +10,7 @@
     <Message Importance="High" Text="Running tests ..." />
     <MSBuild Targets="VSTest"
              Projects="@(Project)"
-             Properties="VSTestNoBuild=true" />
+             Properties="VSTestNoBuild=true;Coverage=true;" />
```
eerhardt: There should be a way to shut this off locally, if devs don't want it.

test/Directory.Build.targets
```diff
@@ -1,5 +1,13 @@
 <Project>
 
   <Import Project="..\Directory.Build.targets" />
-
+  
+  <PropertyGroup Condition="'$(Coverage)' == 'true' and '$(Configuration)'=='Release' and '$(OS)' == 'Windows_NT' and '$(TargetArchitecture)' == 'x64'">
```
eerhardt: A few questions about this -

1. Why is this being set in a `.targets` file? It should go into the `.props` file instead.
2. Why only on Windows and Release builds?  (I assume the x64 is a limitation of the coverage tool. If it isn't - then why the condition on it?)

test/Directory.Build.targets
```diff
@@ -1,5 +1,13 @@
...
+      <CollectCoverage>true</CollectCoverage>
+      <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+      <CoverletOutputFormat>opencover</CoverletOutputFormat>
+      <CoverletOutput>$(MSBuildThisFileDirectory)..\bin\coverage\$(AssemblyName).xml</CoverletOutput>
```
eerhardt: Use `$(BaseOutputPath)` instead of explicitly pathing to it. We have "common" properties for these directories [https://github.com/dotnet/machinelearning/blob/d26510f5efc6a22f0d1a73663e22095f60a4bcf6/Directory.Build.props#L30-L40](https://github.com/dotnet/machinelearning/blob/d26510f5efc6a22f0d1a73663e22095f60a4bcf6/Directory.Build.props#L30-L40)

Shouldn't we be putting these files under `$(BaseOutputPath)$(PlatformConfig)` ? That way I can run debug vs. release, and the files don't overwrite each other.

test/Directory.Build.targets
```diff
@@ -1,5 +1,13 @@
 <Project>
 
   <Import Project="..\Directory.Build.targets" />
-
+  
+  <PropertyGroup Condition="'$(Coverage)' == 'true' and '$(Configuration)'=='Release' and '$(OS)' == 'Windows_NT' and '$(TargetArchitecture)' == 'x64'">
+      <CollectCoverage>true</CollectCoverage>
+      <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
```
eerhardt: Why is this necessary? Does the coverage tool require it? If so, can you link to information on it?

test/Directory.Build.targets
ViktorHofer: It belongs into the targets file as it relies on AssemblyName which isn't available before the targets file. I saw that this was now changed to MSBuildProjectName which isn't correct.

test/run-tests.proj
ViktorHofer: Correct, that option needs to be added to the run tool and removed here.

test/Directory.Build.targets
codemzs: Why isn’t it correct?

ViktorHofer: You can override an assembly's name with the `<AssemblyName>` property. If you build a project the produced assembly will use that name and not the project name. Hence the coverage report filename should also rely on `AssemblyName`.

eerhardt: In ML.NET, the `$(MSBuildProjectName)` will always equal `$(AssemblyName)`. If someone wants to stray from that pattern, they can set the `CoverletOutput` property as well. But it would be highly discouraged to not keep the project name match the assembly name.

ViktorHofer: Ok but what's the *reason* why you wouldn't use AssemblyName here?

eerhardt: The reason is so these properties can be set in a `.props` file. Setting properties in `.targets` files is discouraged (for me anyway) because then the `.csproj` can't override them without dropping out of the `<Project Sdk="Microsoft.NET.Sdk">` syntax.

eerhardt: Another question to be asking here is: Why doesn't Coverlet default this for us? Why are we (and every other consumer) responsible for setting this?

ViktorHofer: I know that you have a lot of experience in this area but I wouldn't encourage others to follow this principle. In this case it works as there's a msbuild property that could be used before the project is loaded but usually that's not the case. Also by conditioning these properties in a targets file with `'$(Property)' == ''` allows you to overwrite them in the project without dropping the sdk attribute.

cc @ericstj for opinion.

ViktorHofer: > Another question to be asking here is: Why doesn't Coverlet default this for us? Why are we (and every other consumer) responsible for setting this?

I believe @codemzs mentioned that otherwise the file would just be dropped in the project directory.

sharwell: > Another question to be asking here is: Why doesn't Coverlet default this for us? Why are we (and every other consumer) responsible for setting this?

It does set a default: `$(MSBuildProjectDirectory)coverage\`

test/Directory.Build.props
```diff
@@ -31,6 +31,15 @@
     <PackageReference Include="xunit" Version="2.4.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
     <PackageReference Include="Microsoft.DotNet.XUnitExtensions" Version="2.4.0-prerelease-63213-02" />
+    <PackageReference Include="coverlet.msbuild" Version="2.5.0" />
+    <PackageReference Include="Codecov" Version="1.1.0" />
   </ItemGroup>
-
+  <PropertyGroup Condition="'$(Coverage)' == 'true' and !$(Configuration.EndsWith('-netfx'))">
```
sharwell: 💡 We can allow the `Coverage` property to control this.

```suggestion
  <PropertyGroup Condition="'$(Coverage)' == 'true'">
```

ViktorHofer: Correct. Also the Coverage option should be moved to https://github.com/dotnet/machinelearning/blob/master/config.json and passed in in the yaml/groovy scripts.

test/Directory.Build.targets
eerhardt: Who wants that default? Outputs shouldn't be dropped in the source directory.... They should be going to the output directory.

Looking at the code, it also doesn't output a file based on the `$(AssemblyName)` either - https://github.com/tonerdo/coverlet/blob/28fb5af1947c7240a7c36ebc4072daa0c80d26d5/src/coverlet.msbuild.tasks/CoverageResultTask.cs#L95.  What would be wrong with outputting the file to `bin\AnyCPU.Debug\Microsoft.ML.Core.Tests\netcoreapp2.1\coverage.opencover.xml`?
In order to do that, you can just set `<CoverletOutput>$(OutputPath)</>`

test/Directory.Build.props
sharwell: > Also the Coverage option … should be moved to …

This part is covered under a separate conversation above: https://github.com/dotnet/machinelearning/pull/2116#discussion_r247177985

test/Directory.Build.targets
ViktorHofer: > Who wants that default? Outputs shouldn't be dropped in the source directory.... They should be going to the output directory.

Please open an issue for that in https://github.com/tonerdo/coverlet/.

> What would be wrong with outputting the file to 

+1.

test/Directory.Build.props
ViktorHofer: right but that conversation is already flagged as "Outdated" and easy to be overlooked. I shamelessly used this convo to raise awareness ;)

test/Directory.Build.targets
eerhardt: @ViktorHofer - we do the same thing in corefx (assuming `MSBuildProjectName == AssemblyName`). It would be very confusing if people didn't follow this convention:

https://github.com/dotnet/corefx/blob/77096755443730962810662501f82550fdb4656d/Directory.Build.props#L478
https://github.com/dotnet/corefx/blob/77096755443730962810662501f82550fdb4656d/Directory.Build.props#L356
https://github.com/dotnet/corefx/blob/8f2afd786b04ba95b13d5a50b6f1e566b16cf5ac/eng/ReferenceAssemblies.props#L11-L12
https://github.com/dotnet/corefx/blob/8735f858c09443b52a25a48c6f5b9aaf01e78043/src/shims/manual/Directory.Build.props#L29

eerhardt: > Please open an issue for that in https://github.com/tonerdo/coverlet/.

There already is an issue https://github.com/tonerdo/coverlet/issues/152, but it was closed without a reason/justification. I've asked why it was closed.

TomFinley:
OK. Just easier to get into good habits if we never practice bad ones. Even when practicing. :)

codemzs:
@TomFinley Agreed.

sharwell:
@TomFinley I don't think I've ever seen a code coverage implementation work on the first try in a new infrastructure. The current experiment completely aligns with my expectations given my experiences with this subject in the past.

TomFinley:
Yeah that's fine @sharwell . Mostly it's the merge from *master* into a *feature branch* that sort of turns my stomach.

test/Directory.Build.props
```diff
@@ -31,6 +31,16 @@
     <PackageReference Include="xunit" Version="2.4.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
     <PackageReference Include="Microsoft.DotNet.XUnitExtensions" Version="2.4.0-prerelease-63213-02" />
+    <PackageReference Include="coverlet.msbuild" Version="2.5.1" />
+    <PackageReference Include="Codecov" Version=$(CodecovVersion) />
```
sharwell:
```suggestion
    <PackageReference Include="Codecov" Version="$(CodecovVersion)" />
```

build/Codecov.proj
```diff
@@ -0,0 +1,35 @@
...
+      <_CodecovArgs Include="--flag;$(Configuration)" Condition="'$(Configuration)' != ''" />
+    </ItemGroup>
+
+
+    <!-- <Exec Condition="'$(UseCodecov)' == 'true'"
+          Command="&quot;$(_CodecovPath)&quot; @(_CodecovArgs, ' ')" /> -->
+    <Message Importance="high" Text="&quot;$(_CodecovPath)&quot; @(_CodecovArgs, ' ')" />
+  </Target>
+
+  <Import Project="..\Packages.targets" />
```
sharwell:
```suggestion
```

build/Codecov.proj
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <PropertyGroup>
+    <RepoRoot>$(MSBuildThisFileDirectory)..\</RepoRoot>
+  </PropertyGroup>
```
sharwell: 📝 This group should be unnecessary

build/Codecov.proj
```diff
@@ -0,0 +1,35 @@
...
+
```
sharwell: 💡 

```suggestion

  <Import Project="Dependencies.props" />
```

build/Codecov.proj
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <Include Project="NuGet.props" />
```
sharwell:
```suggestion
  <Import Project="NuGet.props" />
```

PR #2123, opened 2019-01-11
Remove ISchema in TensorflowUtils.cs and polish its shape translation

wschin:
As title. Party member of #1501.


src/Microsoft.ML.TensorFlow/TensorFlow/TensorflowUtils.cs
```diff
@@ -20,51 +20,53 @@ public static class TensorFlowUtils
...
-                    inputOpsGetter = (int col, ref VBuffer<ReadOnlyMemory<char>> dst) =>
-                        dst = new VBuffer<ReadOnlyMemory<char>>(op.NumInputs, inputOps);
-                }
-                inputOpsGetters.Add(inputOpsGetter);
-
-                MetadataUtils.MetadataGetter<ReadOnlyMemory<char>> opTypeGetter =
-                    (int col, ref ReadOnlyMemory<char> dst) => dst = new ReadOnlyMemory<char>(op.OpType.ToArray());
-                opTypeGetters.Add(opTypeGetter);
-
-                var columnType = Utils.Size(shapeArray) == 1 && shapeArray[0] <= 0 ? new VectorType(mlType) :
```
wschin: Note: we should not use recursive if-else-like at all. It's not readable. In fact, this line is equivalent to
```charp
                var columnType = new VectorType(mlType);
                if (Utils.Size(shapeArray) > 0 && shapeArray.Skip(1).All(x => x > 0))
                    columnType = new VectorType(mlType, shapeArray[0] > 0 ? shapeArray : shapeArray.Skip(1).ToArray());
```

src/Microsoft.ML.TensorFlow/TensorFlow/TensorflowUtils.cs
```diff
@@ -15,65 +15,71 @@ namespace Microsoft.ML.Transforms.TensorFlow
...
-                inputOpsLengths.Add(op.NumInputs);
-                MetadataUtils.MetadataGetter<VBuffer<ReadOnlyMemory<char>>> inputOpsGetter = null;
+                // Construct the final ML.NET type of a Tensorflow variable.
+                var shapeArray = graph.GetTensorShape(op[0]).ToIntArray();
+                var columnType = new VectorType(mlType);
+                if (Utils.Size(shapeArray) > 0 && shapeArray.Skip(1).All(x => x > 0))
+                    columnType = new VectorType(mlType, shapeArray[0] > 0 ? shapeArray : shapeArray.Skip(1).ToArray());
+
+                // There can be at most two metadata fields.
+                //  1. The first field always presents. Its value is this operator's ID. For example,
```
yaeldekel: >operator's ID [](start = 74, length = 13)

This is inaccurate. The ID is in op.Name. op.OpType is a description of what the operator does (some examples of OpTypes are: Const, Placeholder, Softmax, Add, MatMul, etc.).

src/Microsoft.ML.TensorFlow/TensorFlow/TensorflowUtils.cs
```diff
@@ -15,65 +15,71 @@ namespace Microsoft.ML.Transforms.TensorFlow
...
-                    Utils.Size(shapeArray) > 0 && shapeArray.Skip(1).All(x => x > 0) ?
-                        new VectorType(mlType, shapeArray[0] > 0 ? shapeArray : shapeArray.Skip(1).ToArray())
-                        : new VectorType(mlType);
-                res.Add(new KeyValuePair<string, ColumnType>(op.Name, columnType));
+                    metadataBuilder.Add(TensorflowUpstreamOperatorsKind, new VectorType(TextType.Instance, op.NumInputs),
+                        (ref VBuffer<ReadOnlyMemory<char>> value) =>
+                            {
+                                var bufferEditor = VBufferEditor.Create(ref value, op.NumInputs);
+                                for (int i = 0; i < op.NumInputs; ++i)
+                                    bufferEditor.Values[i] = op.GetInput(i).Operation.Name.AsMemory();
```
yaeldekel: I think you should not put the foreach variable inside a lambda expression.

wschin: Any suggestion for copying array to vbuffer?

src/Microsoft.ML.TensorFlow/TensorFlow/TensorflowUtils.cs
wschin: Nice catch!

src/Microsoft.ML.TensorFlow/TensorFlow/TensorflowUtils.cs
```diff
@@ -15,65 +15,71 @@ namespace Microsoft.ML.Transforms.TensorFlow
...
-                var shape = graph.GetTensorShape(op[0]);
-                var shapeArray = shape.ToIntArray();
-
-                inputOpsLengths.Add(op.NumInputs);
-                MetadataUtils.MetadataGetter<VBuffer<ReadOnlyMemory<char>>> inputOpsGetter = null;
+                // Construct the final ML.NET type of a Tensorflow variable.
+                var shapeArray = graph.GetTensorShape(op[0]).ToIntArray();
+                var columnType = new VectorType(mlType);
+                if (Utils.Size(shapeArray) > 0 && shapeArray.Skip(1).All(x => x > 0))
+                    columnType = new VectorType(mlType, shapeArray[0] > 0 ? shapeArray : shapeArray.Skip(1).ToArray());
```
zeahmed: >columnType [](start = 20, length = 10)

VectorType constructor throws exception when length of dimension is 0. I think this could happen when shapeArray = {1}. I see in previous code there is a special check for shapeArray.length of `1` and `0`. Can you please make sure if code works in this case.

wschin: I don't see an exception in its original version. This code is equivalent to the old setting. If old code throws, it would throw as well.

zeahmed: I see that you added a check `!(Utils.Size(shapeArray) == 1 && shapeArray[0] <= 0)` in commit 5. This will make `columnType == null` in case where shapeArray = {0} correct? 

What I meant to say that 
``` csharp
var columnType = Utils.Size(shapeArray) == 1 && shapeArray[0] == -1 ? new VectorType(mlType) :
                    Utils.Size(shapeArray) > 0 && shapeArray.Skip(1).All(x => x > 0) ?
                        new VectorType(mlType, shapeArray[0] > 0 ? shapeArray : shapeArray.Skip(1).ToArray())
                        : new VectorType(mlType);
```

is not equivalent to 

``` csharp
if (!(Utils.Size(shapeArray) == 1 && shapeArray[0] <= 0) &&
                    (Utils.Size(shapeArray) > 0 && shapeArray.Skip(1).All(x => x > 0)))
                    columnType = new VectorType(mlType, shapeArray[0] > 0 ? shapeArray : shapeArray.Skip(1).ToArray());
```

Test might be passing but there are cases when it will fail.

---
In reply to: [247230900](https://github.com/dotnet/machinelearning/pull/2123#discussion_r247230900) [](ancestors = 247230900)

wschin: They are equivalent... That's why recursive if-else condition shouldn't be used anywhere. The only situation which is not `VectorType(mlType)` happens when the two conditions are both false. For all other cases, we get `VectorType(mlType)`.

---
In reply to: [247239512](https://github.com/dotnet/machinelearning/pull/2123#discussion_r247239512) [](ancestors = 247239512,247230900)

zeahmed: OK, I see columtype is created at line 48 `var columnType = new VectorType(mlType);` with that its worsk now after 4th commit. Thanks for the explanation.

---
In reply to: [247259950](https://github.com/dotnet/machinelearning/pull/2123#discussion_r247259950) [](ancestors = 247259950,247239512,247230900)

PR #2124, opened 2019-01-11
Make array argument names plural

### LINKED ISSUE #2040: Pluralize names of settings/arguments field values to plural when array
Consider this content of concatentating transformer's `Argument`'s class:

https://github.com/dotnet/machinelearning/blob/496e185fdaf8969ffd856c65c42dd259368a242d/src/Microsoft.ML.Data/Transforms/ColumnConcatenatingTransformer.cs#L116-L117

Or this similar entry:

https://github.com/dotnet/machinelearning/blob/496e185fdaf8969ffd856c65c42dd259368a242d/src/Microsoft.ML.Data/Transforms/ColumnConcatenatingTransformer.cs#L122-L123

You will notice something curious. This is an array of `Column[]`, yet is named `Column`, not `Columns`. The reason for this is somewhat odd: this settings object *originally* arose out of the need to provide a command line settable object back when this code was supporting a tool (and not an API!). And, the presence of an array indicate that this setting can be set multiple times... so one could say, for instance, `column=Foo column=Bar column=Biz`, and so wind up with this field populated with 3 items. In this setting the name "column" is the most natural name.

Now we are trying to ship an API. What we did for the convenience of the command line now causes confusion for an API user, because they see an array (which can clearly handle multiple items), yet it has a singular name!

Fortunately, we can have our cake and eat it too: the `ArgumentAttribute` has something called `Name`, to allow the field name to be distinguished from the command line name. So, this:

```csharp
[Argument(ArgumentType.Multiple | ArgumentType.Required,
    HelpText = "New column definition(s) (optional form: name:srcs)",
    ShortName = "col", SortOrder = 1)]
public Column[] Column;
```

could become this:

```csharp
[Argument(ArgumentType.Multiple | ArgumentType.Required,
    HelpText = "New column definition(s) (optional form: name:srcs)",
    Name = "Column", ShortName = "col", SortOrder = 1)]
public Column[] Columns;
```

and remain equivalent from the command line perspective, while not being confusing from the API perspective.

As a general rule any field with an `ArgumentAttribute` attribute on it that is also an array, we have often not made plural, for the aforementioned reasons. If it is singular:

1. Make the field name itself plural and

2. Retain the old singular command line as this `Name` parameter, so as to retain the benefit of why we named it this way in the first place.

However, please, not thoughtlessly! Think about whether that change makes sense *in context*.

/cc @stephentoub

najeeb-kazmi:
Fixes #2040 

Makes array argument names plural where they were singular.

Fixes `CmdParser` to check for Attribute name in addition to field Name.

Fixes Entry Points `InputBuilder` to check for both Attribute and Field names.


src/Microsoft.ML.Data/Commands/CrossValidationCommand.cs
```diff
@@ -57,8 +57,9 @@ public sealed class Arguments : DataCommand.ArgumentsBase
             [Argument(ArgumentType.LastOccurenceWins, HelpText = "Column to use for stratification", ShortName = "strat", SortOrder = 7)]
             public string StratificationColumn;
 
-            [Argument(ArgumentType.LastOccurenceWins, HelpText = "Columns with custom kinds declared through key assignments, for example, col[Kind]=Name to assign column named 'Name' kind 'Kind'", ShortName = "col", SortOrder = 10)]
-            public KeyValuePair<string, string>[] CustomColumn;
+            [Argument(ArgumentType.LastOccurenceWins, HelpText = "Columns with custom kinds declared through key assignments, for example, col[Kind]=Name to assign column named 'Name' kind 'Kind'",
+                Name = "CustomColumn", ShortName = "col", SortOrder = 10)]
```
abgoswam: >CustomColumn [](start = 24, length = 12)

So should this be 'CustomColumns' ?  

(here, and everywhere else)

najeeb-kazmi: No. This is so that the behavior remains the same for the command line. Similarly for 'Column'. See #2040 for details.

abgoswam:
For the 3 places you mention, where we used `string Terms`  the  convention proposed in the PR seems confusing to me.

In particular `public string TermsList` looks odd. Don't you think ?

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs
```diff
@@ -107,11 +107,11 @@ public abstract class ArgumentsBase : TransformInputBase
             [Argument(ArgumentType.AtMostOnce, HelpText = "Maximum number of terms to keep per column when auto-training", ShortName = "max", SortOrder = 5)]
             public int MaxNumTerms = ValueToKeyMappingEstimator.Defaults.MaxNumTerms;
 
-            [Argument(ArgumentType.AtMostOnce, HelpText = "Comma separated list of terms", SortOrder = 105, Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly)]
-            public string Terms;
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Comma separated list of terms", Name = "Terms", SortOrder = 105, Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly)]
+            public string TermsList;
```
abgoswam: >public string TermsList; [](start = 12, length = 24)

This looks  rather odd.  We are using suffix `List`  but its type is a `string`

src/Microsoft.ML.Data/Commands/CrossValidationCommand.cs
abgoswam: Ah ok. Thanks for the clarification.

---
In reply to: [247224979](https://github.com/dotnet/machinelearning/pull/2124#discussion_r247224979) [](ancestors = 247224979)

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs
sfilipi: i think that is an overload for the below. 
Does it make sense to log an issue about making things like this internal? TextLoader has a few of those too. 

---
In reply to: [247227162](https://github.com/dotnet/machinelearning/pull/2124#discussion_r247227162) [](ancestors = 247227162)

sfilipi: I opened this: [https://github.com/dotnet/machinelearning/issues/2133](https://github.com/dotnet/machinelearning/issues/2133)

---
In reply to: [247279013](https://github.com/dotnet/machinelearning/pull/2124#discussion_r247279013) [](ancestors = 247279013,247227162)

src/Microsoft.ML.EntryPoints/CrossValidationMacro.cs
```diff
@@ -95,7 +95,7 @@ public sealed class Output
         {
             [TlcModule.Output(Desc = "The final model including the trained predictor model and the model from the transforms, " +
                 "provided as the Input.TransformModel.", SortOrder = 1)]
-            public PredictorModel[] PredictorModel;
+            public PredictorModel[] PredictorModels;
```
sfilipi: >PredictorModels [](start = 36, length = 15)

i think this might have side-effects in the entry points; but that is probably ok. 

@yaeldekel might have an opinion too.

src/Microsoft.ML.EntryPoints/MacroUtils.cs
```diff
@@ -75,21 +75,21 @@ public static EvaluateInputBase GetEvaluatorArgs(TrainerKinds kind, out string e
         public sealed class ArrayIPredictorModelInput
         {
             [Argument(ArgumentType.Required, HelpText = "The models", SortOrder = 1)]
-            public PredictorModel[] Model;
+            public PredictorModel[] Models;
```
sfilipi: >Models [](start = 36, length = 6)

Same note on tis file, about entry points. 

Also, the EntryPoint project is not directly user facing.. idk if we still want to do it for consistency.

src/Microsoft.ML.EntryPoints/CrossValidationMacro.cs
najeeb-kazmi: You're right, it did break some entry point tests. I realize this one is not an `ArgumentAttribute` to begin with, so I'm reverting it in my next change (I haven't pushed yet).

Many other entry point tests also broke because of renaming `Column` to `Columns` and setting `Name = "Column"`. Basically, the graph runner ctor checks for validity of node names, but it only checks against `ArgumentAttribute.Name` and the `ArgumentAttribute.Aliases`, and since `Name` was set to "Column" and Aliases were the `ShortName` list, all of those checks failed. 

It wouldn't have been correct to change `ShortName` to "col,Columns` as that is visible on the command line, and "Columns" doesn't make sense on the command line. I ended up modifying the check in `InputBuilder`.

src/Microsoft.ML.EntryPoints/MacroUtils.cs
najeeb-kazmi: This one does't break the tests, but I do agree we might want to rethink touching entry point arguments right now. The effects might propagate to NimbusML for instance.

@TomFinley what do you think?

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs
najeeb-kazmi: It does look odd. And its not an overload for the one below. They are named `string Terms` and `string[] Term` originally :/. It complains if I name the string array "Terms" as well.

Also, it is breaking some command line tests, which complain about not finding "Terms" despite me adding `Name = "Terms"` to `string TermsList` in my branch. Reverting the names fixes it. Would you suggest not touching these in this PR and taking it up in #2133 ?

cc: @TomFinley

najeeb-kazmi: Alternatively, I could rename `string[] Term` to `string[] TermArray` and keep `string Terms`, and decide about making one internal in @2133. Thoughts?

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs
```diff
@@ -49,11 +49,11 @@ public abstract class ColumnBase : OneToOneColumn
             [Argument(ArgumentType.AtMostOnce, HelpText = "Maximum number of terms to keep when auto-training", ShortName = "max")]
             public int? MaxNumTerms;
 
-            [Argument(ArgumentType.AtMostOnce, HelpText = "Comma separated list of terms", Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly)]
-            public string Terms;
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Comma separated list of terms", Name = "Terms", Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly)]
+            public string Term;
```
najeeb-kazmi: >public [](start = 12, length = 6)

To be made internal in @2133, after #2059 gets merged

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs
```diff
@@ -107,11 +107,11 @@ public abstract class ArgumentsBase : TransformInputBase
             [Argument(ArgumentType.AtMostOnce, HelpText = "Maximum number of terms to keep per column when auto-training", ShortName = "max", SortOrder = 5)]
             public int MaxNumTerms = ValueToKeyMappingEstimator.Defaults.MaxNumTerms;
 
-            [Argument(ArgumentType.AtMostOnce, HelpText = "Comma separated list of terms", SortOrder = 105, Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly)]
-            public string Terms;
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Comma separated list of terms", Name = "Terms", SortOrder = 105, Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly)]
+            public string Term;
```
najeeb-kazmi: >public [](start = 12, length = 6)

To be made internal in #2133, after #2059 gets merged

src/Microsoft.ML.Transforms/Text/StopWordsRemovingTransformer.cs
```diff
@@ -616,11 +616,11 @@ public bool TryUnparse(StringBuilder sb)
 
         public abstract class ArgumentsBase
         {
-            [Argument(ArgumentType.AtMostOnce, HelpText = "Comma separated list of stopwords", Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly)]
-            public string Stopwords;
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Comma separated list of stopwords", Name = "Stopwords", Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly)]
+            public string Stopword;
```
najeeb-kazmi: >public [](start = 12, length = 6)

To be made internal in #2133, after #2059 gets merged

src/Microsoft.ML.Core/CommandLine/CmdParser.cs
```diff
@@ -477,6 +477,11 @@ private static ArgumentInfo GetArgumentInfo(Type type, object defaults)
                         throw Contracts.Except("Duplicate default argument '{0}' vs '{1}'", def.LongName, field.Name);
 
                     string name = ArgCase(field.Name);
+
+                    string attrName = null;
+                    if (attr.Name != null && attr.Name.ToLowerInvariant() != name.ToLowerInvariant())
```
singlis: Do you care about empty strings too? If so, I would use  String.IsEmptyOrNull. 
Also you can do a case insensitive compare with String.Equals(a,b,StringComparison.OrginalIgnoreCase)

src/Microsoft.ML.Core/CommandLine/CmdParser.cs
```diff
@@ -495,6 +500,11 @@ private static ArgumentInfo GetArgumentInfo(Type type, object defaults)
 
                     if (map.ContainsKey(name.ToLowerInvariant()))
                         throw Contracts.Except("Duplicate name '{0}' in argument type '{1}'", name, type.Name);
+                    if (attrName != null)
```
singlis: >attrName [](start = 24, length = 8)

Same comment here about empty string

src/Microsoft.ML.Core/CommandLine/CmdParser.cs
```diff
@@ -495,6 +500,11 @@ private static ArgumentInfo GetArgumentInfo(Type type, object defaults)
...
+                    {
+                        if (map.ContainsKey(attrName.ToLowerInvariant()))
```
singlis: You may also construct the map with a case insensitive string compare, that way you dont have to do a lower variant. 

See here:  [https://stackoverflow.com/questions/13988643/case-insensitive-dictionary-with-string-key-type-in-c-sharp](https://stackoverflow.com/questions/13988643/case-insensitive-dictionary-with-string-key-type-in-c-sharp)

src/Microsoft.ML.Core/CommandLine/CmdParser.cs
```diff
@@ -495,6 +500,11 @@ private static ArgumentInfo GetArgumentInfo(Type type, object defaults)
...
+                            throw Contracts.Except("Duplicate name '{0}' in argument type '{1}'", attrName, type.Name);
```
singlis: Please convert these to be interpolated.

src/Microsoft.ML.Core/CommandLine/CmdParser.cs
```diff
@@ -508,6 +518,8 @@ private static ArgumentInfo GetArgumentInfo(Type type, object defaults)
 
                     // Note that we put the default arg in the map to ensure that no other args use the same name.
                     map.Add(name.ToLowerInvariant(), arg);
+                    if (attrName != null)
+                        map.Add(attrName.ToLowerInvariant(), arg);
```
singlis: >ToLowerInvariant [](start = 41, length = 16)

If you can do a case insensitive comparer for the map the toLowerInvariant is not needed. However if you want to convert to lower invariant for other reasons, I would suggest doing the conversion at the top of the function and using that through the rest of the code. That way you do the conversion once.

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformerImpl.cs
```diff
@@ -219,7 +219,7 @@ public override void ParseAddTermArg(ref ReadOnlyMemory<char> terms, IChannel ch
                 }
 
                 if (Count == 0)
-                    throw ch.ExceptUserArg(nameof(Arguments.Terms), "Nothing parsed as '{0}'", ItemType);
+                    throw ch.ExceptUserArg(nameof(Arguments.Term), "Nothing parsed as '{0}'", ItemType);
```
singlis: Since you are chaning these strings, can you interpolate them?

src/Microsoft.ML.StandardLearners/Standard/Online/AveragedLinear.cs
```diff
@@ -72,13 +71,13 @@ public abstract class AveragedLinearTrainer<TTransformer, TModel> : OnlineLinear
...
+            protected float Gain;
 
             protected int NumNoUpdates;
 
             // For computing averaged weights and bias (if needed)
-            protected VBuffer<Float> TotalWeights;
-            protected Float TotalBias;
+            protected VBuffer<float> TotalWeights;
+            protected float TotalBias;
             protected Double NumWeightUpdates;
```
singlis: >Double [](start = 22, length = 6)

Should these be double instead of Double?

src/Microsoft.ML.Transforms/MutualInformationFeatureSelection.cs
```diff
@@ -41,9 +41,11 @@ public static class Defaults
 
         public sealed class Arguments : TransformInputBase
         {
-            [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "Columns to use for feature selection", ShortName = "col",
+            [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "Columns to use for feature selection",
+                Name = "Column",
+                ShortName = "col",
                 SortOrder = 1)]
```
sfilipi: maybe keep in one line.

src/Microsoft.ML.StandardLearners/Standard/Online/AveragedLinear.cs
```diff
@@ -11,7 +11,6 @@
 using Microsoft.ML.Internal.Utilities;
 using Microsoft.ML.Learners;
 using Microsoft.ML.Numeric;
-using Float = System.Single;
```
sfilipi: was this change inserted to chase the review boredom :)

src/Microsoft.ML.ImageAnalytics/ImageLoaderTransform.cs
```diff
@@ -56,8 +56,8 @@ public bool TryUnparse(StringBuilder sb)
         public sealed class Arguments : TransformInputBase
         {
             [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "New column definition(s) (optional form: name:src)",
-                ShortName = "col", SortOrder = 1)]
-            public Column[] Column;
+                Name = "Column", ShortName = "col", SortOrder = 1)]
```
sfilipi: >    N [](start = 12, length = 5)

alignment

src/Microsoft.ML.Core/CommandLine/CmdParser.cs
glebuk: (!string.IsNullorEmpty(attr.Name) && 
attr.Name.Equals(name, StringComparison.InvariantCultureIgnoreCase)


---
In reply to: [248905796](https://github.com/dotnet/machinelearning/pull/2124#discussion_r248905796) [](ancestors = 248905796)

src/Microsoft.ML.Data/Commands/CrossValidationCommand.cs
glebuk: >col [](start = 52, length = 3)

still, need to add customColumns as valid command name:
Name = "CustomColumn", ShortName = "col, CustomColumns"

src/Microsoft.ML.Data/Commands/DataCommand.cs
```diff
@@ -53,8 +53,9 @@ public abstract class ArgumentsBase
                 HelpText = "Desired degree of parallelism in the data pipeline", ShortName = "n")]
             public int? Parallel;
 
-            [Argument(ArgumentType.Multiple, Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly, HelpText = "Transform", ShortName = "xf", SignatureType = typeof(SignatureDataTransform))]
-            public KeyValuePair<string, IComponentFactory<IDataView, IDataTransform>>[] Transform;
+            [Argument(ArgumentType.Multiple, Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly,
+                HelpText = "Transform", Name ="Transform", ShortName = "xf", SignatureType = typeof(SignatureDataTransform))]
```
glebuk: >Transform [](start = 47, length = 9)

also add Transforms to aliases via shortname = "xf, Transforms"

src/Microsoft.ML.Data/Dirty/ChooseColumnsByIndexTransform.cs
```diff
@@ -22,8 +22,8 @@ public sealed class ChooseColumnsByIndexTransform : RowToRowTransformBase
     {
         public sealed class Arguments
         {
-            [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "Column index to select", ShortName = "ind")]
-            public int[] Index;
+            [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "Column index to select", Name = "Index", ShortName = "ind")]
```
glebuk: >Column index to select [](start = 81, length = 22)

Indices?

glebuk:
>                     // Drop columns indexed by args.Index

indices?

---
Refers to: src/Microsoft.ML.Data/Dirty/ChooseColumnsByIndexTransform.cs:86 in df10cb7. [](commit_id = df10cb78a43e3bc3456534c6ea30c29f7642edee, deletion_comment = False)

glebuk:
>         [TlcModule.EntryPoint(Name = "Transforms.ScoreColumnSelector", Desc = "Selects only the last score columns and the extra columns specified in the arguments.", UserName = "Choose Columns By Index")]

indices?

---
Refers to: src/Microsoft.ML.Data/EntryPoints/ScoreColumnSelector.cs:22 in df10cb7. [](commit_id = df10cb78a43e3bc3456534c6ea30c29f7642edee, deletion_comment = False)

src/Microsoft.ML.Data/Evaluators/EvaluatorBase.cs
```diff
@@ -367,7 +367,7 @@ private protected AggregatorDictionaryBase(RoleMappedSchema schema, string strat
 
                 if (stratType.GetKeyCount() == 0 && !(stratType is TextType))
                 {
-                    throw Contracts.ExceptUserArg(nameof(MamlEvaluatorBase.ArgumentsBase.StratColumn),
+                    throw Contracts.ExceptUserArg(nameof(MamlEvaluatorBase.ArgumentsBase.StratColumns),
                         "Stratification column '{0}' has type '{1}', but must be a known count key or text", stratCol, stratType);
```
glebuk: >column [](start = 40, length = 6)

columns

src/Microsoft.ML.Data/Commands/TrainTestCommand.cs
```diff
@@ -52,8 +52,9 @@ public sealed class Arguments : DataCommand.ArgumentsBase
             [Argument(ArgumentType.AtMostOnce, HelpText = "Name column name", ShortName = "name", SortOrder = 6)]
             public string NameColumn = DefaultColumnNames.Name;
 
-            [Argument(ArgumentType.LastOccurenceWins, HelpText = "Columns with custom kinds declared through key assignments, for example, col[Kind]=Name to assign column named 'Name' kind 'Kind'", ShortName = "col", SortOrder = 10)]
-            public KeyValuePair<string, string>[] CustomColumn;
+            [Argument(ArgumentType.LastOccurenceWins, HelpText = "Columns with custom kinds declared through key assignments, for example, col[Kind]=Name to assign column named 'Name' kind 'Kind'",
+                Name = "CustomColumn", ShortName = "col", SortOrder = 10)]
+            public KeyValuePair<string, string>[] CustomColumns;
```
glebuk: >CustomColumns [](start = 50, length = 13)

Please add CustomColumns to the list of shortname aliases.  This comment applies for all the changes. where the new name is not listed in the [argument

src/Microsoft.ML.Core/CommandLine/CmdParser.cs
```diff
@@ -477,6 +477,11 @@ private static ArgumentInfo GetArgumentInfo(Type type, object defaults)
                         throw Contracts.Except("Duplicate default argument '{0}' vs '{1}'", def.LongName, field.Name);
 
                     string name = ArgCase(field.Name);
+
+                    string attrName = null;
+                    if (attr.Name != null && attr.Name.ToLowerInvariant() != name.ToLowerInvariant())
+                        attrName = attr.Name;
```
TomFinley: My expectation is that the logic should be the same as for entry-points... if `Name` *is* set, then we would not use the field name from the user's perspective at all.

src/Microsoft.ML.Transforms/Text/StopWordsRemovingTransformer.cs
najeeb-kazmi: Name is an alias (with my changes to the cmd line runner). I'm adding the Name attribute so that behavior of existing command lines doesn't get affected with renaming of the field name.

src/Microsoft.ML.Transforms/MutualInformationFeatureSelection.cs
najeeb-kazmi: Done

src/Microsoft.ML.StandardLearners/Standard/Online/AveragedLinear.cs
najeeb-kazmi: :)

src/Microsoft.ML.StandardLearners/Standard/Online/AveragedLinear.cs
najeeb-kazmi: Changed to double.

src/Microsoft.ML.Core/CommandLine/CmdParser.cs
najeeb-kazmi: Not applicable here, since this is attrName, not attr.Name. It's going to be either null or non-empty.

src/Microsoft.ML.Core/CommandLine/CmdParser.cs
najeeb-kazmi: Done

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformerImpl.cs
najeeb-kazmi: Done

src/Microsoft.ML.Core/CommandLine/CmdParser.cs
najeeb-kazmi: Done

src/Microsoft.ML.Data/Dirty/ChooseColumnsByIndexTransform.cs
najeeb-kazmi: Done

src/Microsoft.ML.Core/CommandLine/CmdParser.cs
TomFinley: See the impleemntation of `ArgumentAttribute.Name`; it already guards against this. If you tried to set something to be empty, or even all whitespace, the property value would be null.

TomFinley:
Hi @najeeb-kazmi or others, is this being worked on?

najeeb-kazmi:
> Hi @najeeb-kazmi or others, is this being worked on?

@TomFinley Yes, I have modified to CmdParser to override field name with with attribute name if it is present. The comamnd line tests that were failing are passing now. Currently, I'm looking into the EntryPoints tests that fail when I override the field name with the attribute name.

artidoro:
>                 string[] term = null,

Does it make sense to update it here as well?

---
Refers to: src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs:187 in ac7ee5a. [](commit_id = ac7ee5a2dd0695ddbcf4b1d37a17123d1b7e1776, deletion_comment = False)

najeeb-kazmi:
Thanks @artidoro for your help in getting the conflicts merged!

Issue #581, opened 2018-07-25
Three major concepts: Estimators, Transformers and Data


Zruty0:
This is still an incomplete proposal, but I played for a bit with what I had, and it looks promising to me so far.

The general idea is that we narrow our 'zoo' of components (transforms, predictors, scorers, loaders etc) down to three kinds:

- The **data**. An `IDataView` with schema, like before.
- The **transformer**. This is an object that can transform data and output data.
```C#
    public interface IDataTransformer
    {
        IDataView Transform(IDataView input);
        ISchema GetOutputSchema(ISchema inputSchema);
    }
```
- The **estimator**. This is the 'trainer'. The object that can 'train' a transformer using data.
```C#
    public interface IDataEstimator
    {
        IDataTransformer Fit(IDataView input);
        SchemaShape GetOutputSchema(SchemaShape inputSchema);
    }
```

Obviously, a chain of transformers can itself behave as a transformer, and a chain of estimators can behave like estimators.

We also introduce a 'data reader' (and its estimator), responsible for bringing the data 'from outside' (think loaders):
```C#
    public interface IDataReader<TIn>
    {
        IDataView Read(TIn input);
        ISchema GetOutputSchema();
    }

    public interface IDataReaderEstimator<TIn>
    {
        IDataReader<TIn> Fit(TIn input);
        SchemaShape GetOutputSchema();
    }
```

| Old component | New component |
| ----------------- | ------------------- |
| Data | Data |
| Transform | Transformer |
| Trainable transform (before it is trained) |  Estimator |
| Trainable transform (after it is trained) |  Transformer |
| Trainer | Estimator |
| Predictor | **not sure yet**. I'm thinking like 'a field of the scoring transformer?' |
| Scorer | Transformer |
| Untrainable loader | Data reader |
| Trainable loader | Estimator of data reader |

I have gone through the motions of creating a 'pipeline estimator' and 'pipeline transformer' objects, which then allows me to write this code to train and test:

```C#
            var env = new TlcEnvironment();

            var pipeline = new EstimatorPipe<IMultiStreamSource>(new MyTextLoader(env, MakeTextLoaderArgs()));
            pipeline.Append(new MyConcatTransformer(env, "Features", "SepalLength", "SepalWidth", "PetalLength", "PetalWidth"))
                    .Append(new MyNormalizer(env, "Features"))
                    .Append(new MySdca(env));

            var model = pipeline.Fit(new MultiFileSource(@"e:\data\iris.txt"));

[...]

PR #2125, opened 2019-01-11
Rename CreateTextReader to CreateTextLoader

### LINKED ISSUE #1690: Rename mlContext.Data.TextReader() to mlContext.Data.CreateTextLoader()
Since the object being created is a "TextLoader", I'd recommend the method's name to say so, instead of "TextReader" which looks like a different type. 

I mean code like the following feels confusing:

```
            TextLoader textLoader = mlContext.Data.TextReader(new TextLoader.Arguments()
                                                    {
                                                        Separator = "tab",
                                                        HasHeader = true,
                                                        Column = new[]
                                                                    {
                                                                    new TextLoader.Column("Label", DataKind.Bool, 0),
                                                                    new TextLoader.Column("Text", DataKind.Text, 1)
                                                                    }
                                                    });
```

In addition to that, methods should have a verb as part of the name, so if what it is doing is to create a TextReader, let's say so, as:

mlContext.Data.CreateTextLoader()

We're not being consistent now when having some methods with a verb but other methods just a noun:

image

It is also against C# conventions (and most languages). A method's name should have a verb describing the action being performed by that method:

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-methods

najeeb-kazmi:
Fixes #1690 

Rename `CreateTextReader` to `CreateTextLoader` to conform to the return type:

```csharp
TextLoader textLoader = mlContext.Data.CreateTextLoader(new TextLoader.Arguments() ...
```

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
@@ -11,14 +11,14 @@ namespace Microsoft.ML
     public static class TextLoaderSaverCatalog
     {
         /// <summary>
-        /// Create a text reader <see cref="TextLoader"/>.
+        /// Create a text loader <see cref="TextLoader"/>.
         /// </summary>
         /// <param name="catalog">The <see cref="DataOperations"/> catalog.</param>
         /// <param name="columns">The columns of the schema.</param>
         /// <param name="hasHeader">Whether the file has a header.</param>
         /// <param name="separatorChar">The character used as separator between data points in a row. By default the tab character is used as separator.</param>
         /// <param name="dataSample">The optional location of a data sample.</param>
-        public static TextLoader CreateTextReader(this DataOperations catalog,
+        public static TextLoader CreateTextLoader(this DataOperations catalog,
             TextLoader.Column[] columns,
             bool hasHeader = TextLoader.DefaultArguments.HasHeader,
             char separatorChar = TextLoader.DefaultArguments.Separator,
@@ -26,18 +26,18 @@ namespace Microsoft.ML
             => new TextLoader(CatalogUtils.GetEnvironment(catalog), columns, hasHeader, separatorChar, dataSample);
 
         /// <summary>
-        /// Create a text reader <see cref="TextLoader"/>.
+        /// Create a text loader <see cref="TextLoader"/>.
         /// </summary>
         /// <param name="catalog">The <see cref="DataOperations"/> catalog.</param>
         /// <param name="args">Defines the settings of the load operation.</param>
         /// <param name="dataSample">Allows to expose items that can be used for reading.</param>
-        public static TextLoader CreateTextReader(this DataOperations catalog,
+        public static TextLoader CreateTextLoader(this DataOperations catalog,
             TextLoader.Arguments args,
             IMultiStreamSource dataSample = null)
             => new TextLoader(CatalogUtils.GetEnvironment(catalog), args, dataSample);
 
         /// <summary>
-        /// Create a text reader <see cref="TextLoader"/> by inferencing the dataset schema from a data model type.
+        /// Create a text loader <see cref="TextLoader"/> by inferencing the dataset schema from a data model type.
         /// </summary>
         /// <param name="catalog">The <see cref="DataOperations"/> catalog.</param>
         /// <param name="hasHeader">Does the file contains header?</param>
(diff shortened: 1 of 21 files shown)


src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -11,33 +11,33 @@ namespace Microsoft.ML
     public static class TextLoaderSaverCatalog
     {
         /// <summary>
-        /// Create a text reader <see cref="TextLoader"/>.
+        /// Create a text loader <see cref="TextLoader"/>.
         /// </summary>
         /// <param name="catalog">The <see cref="DataOperations"/> catalog.</param>
```
wschin: The doc string to `catalog` is not clear. Could you make a better one? It doesn't describe its relation to other operators and the text loader being created.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -11,33 +11,33 @@ namespace Microsoft.ML
...
         /// <param name="columns">The columns of the schema.</param>
```
wschin: `columns` doc string is not very useful too. Can I feed arbitrary columns into this function?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -11,33 +11,33 @@ namespace Microsoft.ML
...
         /// <param name="hasHeader">Whether the file has a header.</param>
         /// <param name="separatorChar">The character used as separator between data points in a row. By default the tab character is used as separator.</param>
         /// <param name="dataSample">The optional location of a data sample.</param>
```
wschin: Humm, what is `dataSample`? I know it was not you, but we need to gradually increase the quality of our doc.

sfilipi: >/// <param name="dataSample">The optional location of a data sample.</para [](start = 8, length = 74)

?Do you want to also create a small sample, for the apis of this file, and link it from here?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -11,33 +11,33 @@ namespace Microsoft.ML
...
-        public static TextLoader CreateTextReader(this DataOperations catalog,
+        public static TextLoader CreateTextLoader(this DataOperations catalog,
```
sfilipi: >CreateTextLoader [](start = 33, length = 16)

My personal take on this is: TextReader is a much better name than TextLoader. I would rather rename the  TextLoader class to TextReader.. is that ok for the cmd usage.

TomFinley: This is correct, we are replacing the `IDataLoader` idiom with the `IDataReader` idiom, as part of #581 and much subsequent work.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
najeeb-kazmi: This is actually an extension method to MLContext. The doc string for catalog is not visible to the user.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
najeeb-kazmi: `TextReader` is conflated with `System.IO.TextReader` and should be avoided. I've filed #2144 for renaming the class. I propose making the method name and return type consistent in this PR and revisiting the name when we rename the class.

cc: @glebuk @eerhardt

TomFinley: Ah. Good point,. Maybe we should actually yname `IDataReader` to `IDataLoader`? And call the existing legacy `IDataLoader` interface something like `IDataLoaderLegacy`? What we all think?

TomFinley: Similar renames to `IDataReaderEstimator` to I suppose `IDataLoaderEstimator`.

CESARDELATORRE: My point is to be consistent between method's name and returned type, so either on of the following would be good for me: 👍 

public static Text**Loader** CreateText**Loader**()

public static Text**Reader** CreateText**Reader**()

PR #2128, opened 2019-01-11
Publish test trx files on failure

eerhardt:
There are times when the tests fail, but we don't know why. Publish the trx files as well during test failures.

### DIFF
--- a/build/ci/phase-template.yml
+++ b/build/ci/phase-template.yml
@@ -55,7 +55,9 @@ phases:
       condition: not(succeeded())
       inputs:
         sourceFolder: $(Build.SourcesDirectory)/bin
-        contents: '**/TestOutput/**/*'
+        contents: |
+          **/TestOutput/**/*
+          **/*.trx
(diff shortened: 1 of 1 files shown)


safern:
You might want to force a test failure to make sure this works. There are some special cases where it doesn't work quite well with paths and wildcards. That is why the .log files had to be published on a separate task.

eerhardt:
> You might want to force a test failure to make sure this works

What's the odds that my PR passes on first run? 😃

eerhardt:
> What's the odds that my PR passes on first run?

I guess the answer was "close to 0". https://dev.azure.com/dnceng/public/_build/results?buildId=71361 has the `.trx` files and the `TestOutput` files in the artifacts.

safern:
> I guess the answer was "close to 0"

LOL. So it works, then you're good to merge I guess.

shauheen:
> > You might want to force a test failure to make sure this works
> 
> What's the odds that my PR passes on first run? 😃

oh man!

PR #2131, opened 2019-01-11
Remove "VectorType" specific members on ColumnType.

eerhardt:
Remove the following members from ColumnType:

- IsVector
- ItemType
- IsKnownSizeVector
- VectorSize
- ValueCount


src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
```diff
@@ -22,5 +22,44 @@ internal static class ColumnTypeExtensions
         /// Zero return means either it's not a key type or the cardinality is unknown.
         /// </summary>
         public static int GetKeyCount(this ColumnType columnType) => (columnType as KeyType)?.Count ?? 0;
+
+        public static ColumnType ItemType(this ColumnType columnType) => (columnType as VectorType)?.ItemType ?? columnType;
```
Ivanidzo4ka: it deserves comment.

src/Microsoft.ML.Core/Data/IEstimator.cs
```diff
@@ -146,17 +146,28 @@ public SchemaShape(IEnumerable<Column> columns)
             out ColumnType itemType,
             out bool isKey)
         {
-            if (type.IsKnownSizeVector)
-                vecKind = Column.VectorKind.Vector;
-            else if (type.IsVector)
-                vecKind = Column.VectorKind.VariableVector;
+            if (type is VectorType vectorType)
```
Ivanidzo4ka: >type is VectorType vectorType [](start = 16, length = 29)

did you touch this code before you wrote extensions, or what's the reason for not using them?

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
eerhardt: Agreed. I'll add it. Plus I need to make these methods have "verb" names.

eerhardt:
OK - this change is ready for review. Sorry for how big it grew to.

src/Microsoft.ML.Transforms/MissingValueReplacing.cs
```diff
@@ -267,30 +268,31 @@ private MissingValueReplacingTransformer(IHost host, ModelLoadContext ctx)
                 if (!saver.TryLoadTypeAndValue(ctx.Reader.BaseStream, out ColumnType savedType, out object repValue))
                     throw Host.ExceptDecode();
                 _replaceTypes[i] = savedType;
-                if (savedType.IsVector)
+                if (savedType is VectorType savedVectorType)
                 {
                     // REVIEW: The current implementation takes the serialized VBuffer, densifies it, and stores the values array.
                     // It might be of value to consider storing the VBUffer in order to possibly benefit from sparsity. However, this would
                     // necessitate a reimplementation of the FillValues code to accomodate sparse VBuffers.
                     object[] args = new object[] { repValue, _replaceTypes[i], i };
```
Ivanidzo4ka: >_replaceTypes [](start = 61, length = 13)

Is it worth to cast _replaceType as VectorType and modify `GetValuesArray`  to accept `VectorType` other than `ColumnType`

src/Microsoft.ML.Transforms/MissingValueReplacing.cs
```diff
@@ -595,20 +597,20 @@ public Mapper(MissingValueReplacingTransformer parent, Schema inputSchema)
...
+                    if (!type.GetItemType().Equals(repType.GetItemType()))
                         throw Host.ExceptParam(nameof(InputSchema), "Column '{0}' item type '{1}' does not match expected ColumnType of '{2}'",
-                            _infos[i].Source, _parent._replaceTypes[i].ItemType.ToString(), _infos[i].TypeSrc);
+                            _infos[i].Source, _parent._replaceTypes[i].GetItemType().ToString(), _infos[i].TypeSrc);
                     // If type is a vector and the value is not either a scalar or a vector of the same size, throw an error.
-                    if (repType.IsVector)
+                    if (repType is VectorType repVectorType)
                     {
-                        if (!type.IsVector)
+                        if (!(type is VectorType vectorType2))
```
Ivanidzo4ka: >vectorType2 [](start = 49, length = 11)

you already casted type to VectorType, why you have this vectorType2?

src/Microsoft.ML.Data/Transforms/NormalizeColumnSng.cs
```diff
@@ -605,10 +605,10 @@ public ImplVec(IHost host, TFloat[] scale, TFloat[] offset, int[] indicesNonZero
                     {
                     }
 
-                    public static new ImplVec Create(ModelLoadContext ctx, IHost host, ColumnType typeSrc)
+                    public static ImplVec Create(ModelLoadContext ctx, IHost host, VectorType typeSrc)
```
Ivanidzo4ka: >public static ImplVec Create(ModelLoadContext ctx, IHost host, VectorType typeSrc) [](start = 20, length = 82)

`public delegate void SignatureLoadColumnFunction(ModelLoadContext ctx, IHost host, ColumnType typeSrc);`
Is it works because VectorType is based on ColumnType, or we just don't have tests which covers loading of normalizer?

Ivanidzo4ka: I guess it works because it calls create method in NormalizeColumn.cs in which we call these methods.

---
In reply to: [247765979](https://github.com/dotnet/machinelearning/pull/2131#discussion_r247765979) [](ancestors = 247765979)

src/Microsoft.ML.Data/Transforms/NormalizeColumn.cs
```diff
@@ -180,11 +180,12 @@ public abstract class ArgumentsBase : TransformInputBase
 
             public string TestType(ColumnType type)
             {
-                if (type.ItemType != NumberType.R4 && type.ItemType != NumberType.R8)
+                ColumnType itemType = type.GetItemType();
```
Ivanidzo4ka: >GetItemType [](start = 43, length = 11)

In most of the cases intent of this routine is to execute both if statements and return null. I would cast to vector here, feel free to switch statements.

src/Microsoft.ML.Data/Transforms/KeyToVector.cs
```diff
@@ -263,10 +263,11 @@ public Mapper(KeyToVectorMappingTransformer parent, Schema inputSchema)
                 _types = new VectorType[_parent.ColumnPairs.Length];
                 for (int i = 0; i < _parent.ColumnPairs.Length; i++)
                 {
-                    if (_parent._columns[i].Bag || _infos[i].TypeSrc.ValueCount == 1)
-                        _types[i] = new VectorType(NumberType.Float, _infos[i].TypeSrc.ItemType.GetKeyCount());
+                    int valueCount = _infos[i].TypeSrc.GetValueCount();
```
Ivanidzo4ka: >_infos[i].TypeSrc [](start = 37, length = 17)

do you want to cast `_info[i].TypeSrc` to vector first? or code becomes to hard to read?

src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
```diff
@@ -63,7 +63,7 @@ private static ISchemaBoundMapper WrapIfNeeded(IHostEnvironment env, ISchemaBoun
 
             if (trainSchema?.Label == null)
                 return mapper; // We don't even have a label identified in a training schema.
-            var keyType = trainSchema.Label.Value.Metadata.Schema.GetColumnOrNull(MetadataUtils.Kinds.KeyValues)?.Type;
+            var keyType = trainSchema.Label.Value.Metadata.Schema.GetColumnOrNull(MetadataUtils.Kinds.KeyValues)?.Type as VectorType;
```
Ivanidzo4ka: >VectorType [](start = 122, length = 10)

why?

src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
```diff
@@ -290,7 +289,7 @@ private static ColumnType GetPredColType(ColumnType scoreType, ISchemaBoundRowMa
             var labelNameBindableMapper = mapper.Bindable as MultiClassClassifierScorer.LabelNameBindableMapper;
             if (labelNameBindableMapper == null)
                 return BoolType.Instance;
-            return new KeyType(DataKind.U4, 0, labelNameBindableMapper.Type.VectorSize);
+            return new KeyType(DataKind.U4, 0, labelNameBindableMapper.Type.Size);
```
Ivanidzo4ka: why not GetVectorSize?

src/Microsoft.ML.Data/Scorers/MultiClassClassifierScorer.cs
```diff
@@ -379,7 +379,7 @@ private static ISchemaBoundMapper WrapIfNeeded(IHostEnvironment env, ISchemaBoun
 
             if (trainSchema?.Label == null)
                 return mapper; // We don't even have a label identified in a training schema.
-            var keyType = trainSchema.Label.Value.Metadata.Schema.GetColumnOrNull(MetadataUtils.Kinds.KeyValues)?.Type;
+            var keyType = trainSchema.Label.Value.Metadata.Schema.GetColumnOrNull(MetadataUtils.Kinds.KeyValues)?.Type as VectorType;
```
Ivanidzo4ka: >as VectorType; [](start = 119, length = 14)

Can you point on a reason for this new constraint?

src/Microsoft.ML.Core/Data/IEstimator.cs
```diff
@@ -146,17 +146,28 @@ public SchemaShape(IEnumerable<Column> columns)
...
+            {
+                if (vectorType.IsKnownSize)
+                {
+                    vecKind = Column.VectorKind.Vector;
+                }
+                else
+                {
```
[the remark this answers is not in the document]
eerhardt: In general, I try to always brace `if` conditions. Our [coding guidelines](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md) allow for either. See https://github.com/dotnet/corefx/issues/381 for the reasons why this guideline was written as it is.

src/Microsoft.ML.Transforms/MissingValueReplacing.cs
eerhardt: Good call. Fixed.

src/Microsoft.ML.Data/DataView/DataViewConstructionUtils.cs
```diff
@@ -855,16 +857,20 @@ public MetadataInfo(string kind, T value, ColumnType metadataType = null)
             else
             {
                 // Make sure that the types are compatible with the declared type, including whether it is a vector type.
-                if (isVector != metadataType.IsVector)
+                VectorType metdataVectorType = metadataType as VectorType;
+                bool metdataIsVector = metdataVectorType != null;
```
Ivanidzo4ka: can you spell them `metadataSomething` and not `metdataSomething` I don't see any reason to short it.

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
```diff
@@ -1100,14 +1100,14 @@ private void CheckInputColumnTypes(Schema schema)
                 throw Host.Except("Label column '{0}' has type '{1}' but must be R4, R8, BL or a 2-value key", LabelCol, t);
 
             t = schema[ScoreIndex].Type;
-            if (t.IsVector || t.ItemType != NumberType.Float)
```
Ivanidzo4ka: >t.IsVector [](start = 16, length = 10)

Why we no longer check is type a vector?

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
```diff
@@ -22,5 +22,48 @@ internal static class ColumnTypeExtensions
...
+        /// <summary>
+        /// For non-vector types, this returns the column type itself (i.e., return <paramref name="columnType"/>).
+        /// For vector types, this returns the type of the items stored as values in vector.
+        /// </summary>
+        public static ColumnType GetItemType(this ColumnType columnType) => (columnType as VectorType)?.ItemType ?? columnType;
+
+        /// <summary>
+        /// Zero return means either it's not a vector or the size is unknown.
+        /// </summary>
+        public static int GetVectorSize(this ColumnType columnType) => (columnType as VectorType)?.Size ?? 0;
```
TomFinley: Just something I found interesting. When I saw this code, I thought to myself that maybe something like an `is` would actually work faster. That is, instead of `(columnType as VectorType)?.Size ?? 0` prefer `columnType is VectorType vecType ? vecType.Size : 0`. But, I benchmarked an analogous situation, and I cannot see that one is faster than the other. So, this code is fine, but I found the result surprising. 😄

src/Microsoft.ML.Data/Transforms/DropSlotsTransform.cs
```diff
@@ -503,22 +508,22 @@ private static bool IsValidColumnType(ColumnType type)
 
                 categoricalRanges = null;
                 var typeSrc = _srcTypes[iinfo];
-                if (!typeSrc.IsVector)
+                if (!(typeSrc is VectorType vectorType))
```
TomFinley: Not necessary to change, but whenever this pattern happens it always confuses me... `if (!(a is B b) {} else {... usage of b ...}`. 😄

src/Microsoft.ML.Data/Transforms/InvertHashUtils.cs
```diff
@@ -390,9 +391,10 @@ private static void Save(IChannel ch, ModelSaveContext ctx, CodecFactory factory
             IValueCodec codec;
             var result = factory.TryGetCodec(new VectorType(TextType.Instance), out codec);
             ch.Assert(result);
-            ch.Assert(codec.Type.IsVector);
-            ch.Assert(codec.Type.VectorSize == 0);
-            ch.Assert(codec.Type.ItemType.RawType == typeof(ReadOnlyMemory<char>));
+            VectorType vectorType = codec.Type as VectorType;
+            ch.Assert(vectorType != null);
+            ch.Assert(vectorType.Size == 0);
+            ch.Assert(vectorType.ItemType.RawType == typeof(ReadOnlyMemory<char>));
```
TomFinley: This *works* I guess, but would `ch.Assert(vectorType.ItemType == TextType.Instance);` be a bit more appropriate and in line with our patterns elsewhere? The expectation is that this is not just *any* `ColumnType` that happens to have read only memory of char as the underlying .NET type, but quite specifically the `TextType` itself, right? Which is a stronger statement. (In addition to being more concisely and clearly expressed.)

I know that wasn't how the original code was written -- indeed I'm pretty sure I'm the one that wrote it considering what file this is -- but, that was goofy of me. 😄 Not a big deal, just so long as you're fixing stuff anyway for @Ivanidzo4ka .

src/Microsoft.ML.Data/Transforms/NormalizeColumnSng.cs
TomFinley: I think @eerhardt 's code here represents an improvement, since it is more restrictive about when this was called, whereas the old code was, frankly, a bit too freewheeling about whether it was handling vector or scalar types... 😛

src/Microsoft.ML.Transforms/KeyToVectorMapping.cs
```diff
@@ -478,7 +480,7 @@ public override SchemaShape GetOutputSchema(SchemaShape inputSchema)
             {
                 if (!inputSchema.TryFindColumn(colInfo.Input, out var col))
                     throw Host.ExceptSchemaMismatch(nameof(inputSchema), "input", colInfo.Input);
-                if ((col.ItemType.ItemType.RawKind == default) || !(col.ItemType.IsVector || col.ItemType is PrimitiveType))
+                if ((col.ItemType.GetItemType().RawKind == default) || !(col.ItemType is VectorType || col.ItemType is PrimitiveType))
                     throw Host.ExceptSchemaMismatch(nameof(inputSchema), "input", colInfo.Input);
 
                 var metadata = new List<SchemaShape.Column>();
```
TomFinley: Ugh. I thought I was getting close to done and then I see no, github "abbreviated" the remaining files which I have to open one at a time. La ti da... 😛 😆

src/Microsoft.ML.Transforms/MissingValueReplacing.cs
```diff
@@ -151,13 +151,14 @@ internal static string TestType(ColumnType type)
         {
             // Item type must have an NA value that exists and is not equal to its default value.
             Func<ColumnType, string> func = TestType<int>;
-            return Utils.MarshalInvoke(func, type.ItemType.RawType, type.ItemType);
+            var itemType = type.GetItemType();
```
TomFinley: Probably harmless to keep it part of the statement, but no big deal.

src/Microsoft.ML.Transforms/MissingValueReplacing.cs
TomFinley: I don't think the test on line 599 is available any more, since that `vectorType` would only have scope inside the conditional? And anyway, that `type` is now potentially reassigned to something else, from what I see?

src/Microsoft.ML.Transforms/UngroupTransform.cs
```diff
@@ -327,11 +327,11 @@ public UngroupBinding(IExceptionContext ectx, Schema inputSchema, UngroupMode mo
                     int col;
                     if (!inputSchema.TryGetColumnIndex(name, out col))
                         throw ectx.ExceptUserArg(nameof(Arguments.Column), "Pivot column '{0}' is not found", name);
-                    var colType = inputSchema[col].Type;
-                    if (!colType.IsVector || !(colType.ItemType is PrimitiveType))
+                    var colType = inputSchema[col].Type as VectorType;
+                    if (colType == null || !(colType.ItemType is PrimitiveType))
```
TomFinley: Could probably just be `if (!(colType?.ItemType is PrimitiveType))`? Just a small bit cleaner.

src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
TomFinley: Better to use the actual real property on `VectorType` directly if we can.

src/Microsoft.ML.Data/DataView/DataViewConstructionUtils.cs
eerhardt: Oops. Type-o.  Fixed.

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
eerhardt: I guess I didn't dig into the IL generated before - but my assumption was that the IL would be pretty similar.

My benchmark showed they are basically identical:

```C#
    public class TypeChecking
    {
        private static ColumnType t1 = BoolType.Instance;
        private static ColumnType t2 = NumberType.R4;
        private static ColumnType t3 = new VectorType(BoolType.Instance, 5);

        [Benchmark]
        public int AsTrue()
        {
            return (t3 as VectorType)?.Size ?? 0;
        }

        [Benchmark]
        public int IsTrue()
        {
            return t3 is VectorType vectorType ? vectorType.Size : 0;
        }

        [Benchmark]
        public int AsFalse()
        {
            return (t2 as VectorType)?.Size ?? 0;
        }

        [Benchmark]
        public int IsFalse()
        {
            return t2 is VectorType vectorType ? vectorType.Size : 0;
        }
    }
```

```
 Method |     Mean |     Error |    StdDev |
-------- |---------:|----------:|----------:|
  AsTrue | 1.701 ns | 0.0603 ns | 0.0564 ns |
  IsTrue | 1.703 ns | 0.0359 ns | 0.0336 ns |
 AsFalse | 1.248 ns | 0.0271 ns | 0.0254 ns |
 IsFalse | 1.245 ns | 0.0386 ns | 0.0342 ns |
```

And here's the generated IL:

```IL
    .method public hidebysig instance int32 AsTrue () cil managed 
    {
        .custom instance void [BenchmarkDotNet]BenchmarkDotNet.Attributes.BenchmarkAttribute::.ctor() = (
            01 00 00 00
        )
        IL_0000: ldsfld class [Microsoft.ML.Core]Microsoft.ML.Runtime.Data.ColumnType RandomBenchmarks.TypeChecking::t3
        IL_0005: isinst [Microsoft.ML.Core]Microsoft.ML.Runtime.Data.VectorType
        IL_000a: dup
        IL_000b: brtrue.s IL_0010

        IL_000d: pop
        IL_000e: ldc.i4.0
        IL_000f: ret

        IL_0010: call instance int32 [Microsoft.ML.Core]Microsoft.ML.Runtime.Data.VectorType::get_Size()
        IL_0015: ret
    }

   .method public hidebysig instance int32 IsTrue () cil managed 
    {
        .custom instance void [BenchmarkDotNet]BenchmarkDotNet.Attributes.BenchmarkAttribute::.ctor() = (
            01 00 00 00
        )
        .locals init (
            [0] class [Microsoft.ML.Core]Microsoft.ML.Runtime.Data.VectorType V_0
        )

        IL_0000: ldsfld class [Microsoft.ML.Core]Microsoft.ML.Runtime.Data.ColumnType RandomBenchmarks.TypeChecking::t3
        IL_0005: isinst [Microsoft.ML.Core]Microsoft.ML.Runtime.Data.VectorType
        IL_000a: dup
        IL_000b: stloc.0
        IL_000c: brtrue.s IL_0010

        IL_000e: ldc.i4.0
        IL_000f: ret

        IL_0010: ldloc.0
        IL_0011: callvirt instance int32 [Microsoft.ML.Core]Microsoft.ML.Runtime.Data.VectorType::get_Size()
        IL_0016: ret
    }
```

src/Microsoft.ML.Data/Transforms/DropSlotsTransform.cs
eerhardt: I agree I don't like the pattern either... but I didn't want to change the logic too much. And in the situations where you want to `return` or `throw` when the object isn't a type, this is probably the best way to do it. So it felt useless to try to avoid it.

src/Microsoft.ML.Data/Transforms/DropSlotsTransform.cs
```diff
@@ -761,20 +767,21 @@ private Delegate MakeVecGetter(Row input, int iinfo)
             {
                 Host.AssertValue(input);
                 Host.Assert(0 <= iinfo && iinfo < _parent.ColumnPairs.Length);
-                Host.Assert(_srcTypes[iinfo].IsVector);
+                VectorType vectorType = _srcTypes[iinfo] as VectorType;
```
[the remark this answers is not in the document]
eerhardt: Agreed. Will fix.

src/Microsoft.ML.Data/Transforms/InvertHashUtils.cs
eerhardt: Good call. Will fix.

src/Microsoft.ML.Transforms/MissingValueReplacing.cs
eerhardt: I just figured it was easier to capture the value than call the same method twice...

---
In reply to: [248020618](https://github.com/dotnet/machinelearning/pull/2131#discussion_r248020618) [](ancestors = 248020618)

src/Microsoft.ML.Data/Transforms/NormalizeColumn.cs
eerhardt: I switched them around and only cast to `VectorType` one time.

---
In reply to: [247766966](https://github.com/dotnet/machinelearning/pull/2131#discussion_r247766966) [](ancestors = 247766966)

src/Microsoft.ML.Data/Transforms/KeyToVector.cs
eerhardt: Yes, I think the code is hard to read in that case. It is unfortunate that we have 2 casts here, but I don't think it is a huge deal, do you?

```C#
                for (int i = 0; i < _parent.ColumnPairs.Length; i++)
                {
                    int valueCount;
                    ColumnType srcItemType;
                    if (_infos[i].TypeSrc is VectorType vectorType)
                    {
                        valueCount = vectorType.Size;
                        srcItemType = vectorType.ItemType;
                    }
                    else
                    {
                        valueCount = 1;
                        srcItemType = _infos[i].TypeSrc;
                    }

                    if (_parent._columns[i].Bag || valueCount == 1)
                        _types[i] = new VectorType(NumberType.Float, srcItemType.GetKeyCount());
                    else
                        _types[i] = new VectorType(NumberType.Float, valueCount, srcItemType.GetKeyCount());
                }
```

---
In reply to: [247767549](https://github.com/dotnet/machinelearning/pull/2131#discussion_r247767549) [](ancestors = 247767549)

src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
eerhardt: `CanWrap` checks whether this column is a `VectorType`, so this is "safe". Doing the cast was necessary for line 71's usage of `keyType.ItemType`.

---
In reply to: [247769095](https://github.com/dotnet/machinelearning/pull/2131#discussion_r247769095) [](ancestors = 247769095)

src/Microsoft.ML.Data/Scorers/MultiClassClassifierScorer.cs
eerhardt: Same reasoning as for BinaryClassifierScorer.cs - 

`CanWrap` checks whether this column is a `VectorType`, so this is "safe". Doing the cast was necessary for line 71's usage of `keyType.ItemType`.

---
In reply to: [247769535](https://github.com/dotnet/machinelearning/pull/2131#discussion_r247769535) [](ancestors = 247769535)

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
TomFinley: Yup, I observed the same thing. The IL for `foo?.Bar ?? biz` is a lot smarter than I thought it might be.

src/Microsoft.ML.Transforms/UngroupTransform.cs
TomFinley: Actually, now that I'm thinking about it a little bit more, given that `VectorType.ItemType` *is* in fact a `PrimitiveType`, is this test here equivalent to `colType == null`?

eerhardt: Good call. Fixed.

PR #2136, opened 2019-01-14
Revert tests changes in tensorflow to unblock builds

Ivanidzo4ka:
I can't find proper way to handle tests right now for case where we load same model twice. So i'm removing that part of test

### DIFF
--- a/src/Microsoft.ML.TensorFlow/TensorFlowModelInfo.cs
+++ b/src/Microsoft.ML.TensorFlow/TensorFlowModelInfo.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using Microsoft.ML.Data;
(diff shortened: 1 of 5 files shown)


src/Microsoft.ML.TensorFlow/TensorFlowModelInfo.cs
```diff
@@ -55,5 +56,10 @@ public Schema GetInputSchema()
         {
             return TensorFlowUtils.GetModelSchema(_env, Session.Graph, "Placeholder");
         }
+
+        public void Dispose()
+        {
+            Session.Dispose();
+        }
```
TomFinley: Hi @ivanidzo4ka, I think you should probably follow the [advice of the pattern](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose). What you think?

Also, this is an extensible class (not sealed), so it should follow advice for extensible classes, which complicates the story somewhat. (Unless it not being sealed was a simple mistake, in which case that should be fixed.)

PR #2139, opened 2019-01-14
Remove ISchema in BinaryLoader

wschin:
As title. Parade of #1501 continues today.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs
@@ -562,6 +562,7 @@ namespace Microsoft.ML.Data.IO
 
                 public override Delegate GetGetter()
                 {
+                    EnsureValue();
                     ValueGetter<T> getter = (ref T value) => value = Value;
                     return getter;
                 }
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs
```diff
@@ -472,6 +472,11 @@ private abstract class MetadataTableOfContentsEntry
 
             protected readonly BinaryLoader Parent;
 
+            /// <summary>
+            /// Retuen the getter to the stored entry value as <see cref="Delegate"/>.
```
eerhardt: type-o `Retuen`

wschin: Thank you. Fixed.

src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs
```diff
@@ -757,7 +730,7 @@ public void GetMetadata<TValue>(string kind, int col, ref TValue value)
         /// </summary>
         private const ulong ReaderFirstVersion = 0x0001000100010002;
 
-        public Schema Schema { get { return _schema; } }
+        public Schema Schema { get { return _outputSchema; } }
```
TomFinley: >public Schema Schema { get { return _outputSchema; } } [](start = 8, length = 54)

Cool thanks @wshin. If perchance you happen to need to do some more commits, no harm C# 6/7-ifying these properties. But no hard feelings if you don't of course, just if quite convenient.

PR #2140, opened 2019-01-14
Remove ISchema in TextLoader.cs and TextLoaderCursor.cs

wschin:
As title. This might be the tail of #1501.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -520,26 +520,25 @@ namespace Microsoft.ML.Data
             }
         }
 
-        private sealed class Bindings : ISchema
+        private sealed class Bindings
         {
+            /// <summary>
+            /// <see cref="Infos"/>[i] stores the i-th column's name type loaded from the input text file.
+            /// </summary>
             public readonly ColInfo[] Infos;
-            public readonly Dictionary<string, int> NameToInfoIndex;
+            /// <summary>
+            /// <see cref="Infos"/>[i] stores the i-th column's metadata, slot names.
+            /// </summary>
             private readonly VBuffer<ReadOnlyMemory<char>>[] _slotNames;
-            // Empty iff either header+ not set in args, or if no header present, or upon load
-            // there was no header stored in the model.
+            /// <summary>
+            /// Empty iff either header+ not set in args, or if no header present, or upon load
+            /// there was no header stored in the model.
+            /// </summary>
             private readonly ReadOnlyMemory<char> _header;
 
-            private readonly MetadataUtils.MetadataGetter<VBuffer<ReadOnlyMemory<char>>> _getSlotNames;
-
-            public Schema AsSchema { get; }
-
-            private Bindings()
-            {
-                _getSlotNames = GetSlotNames;
-            }
+            public Schema OutputSchema { get; }
 
             public Bindings(TextLoader parent, Column[] cols, IMultiStreamSource headerFile, IMultiStreamSource dataSample)
-                : this()
             {
                 Contracts.AssertNonEmpty(cols);
                 Contracts.AssertValueOrNull(headerFile);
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -520,26 +520,25 @@ public static ColInfo Create(string name, PrimitiveType itemType, Segment[] segs
             }
         }
 
-        private sealed class Bindings : ISchema
+        private sealed class Bindings
         {
+            /// <summary>
+            /// <see cref="Infos"/>[i] stores the i-th column's name type loaded from the input text file.
```
TomFinley: >name type [](start = 64, length = 9)

Not a big deal, but what is a "name type"?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -520,26 +520,25 @@ public static ColInfo Create(string name, PrimitiveType itemType, Segment[] segs
...
             public readonly ColInfo[] Infos;
-            public readonly Dictionary<string, int> NameToInfoIndex;
+            /// <summary>
+            /// <see cref="Infos"/>[i] stores the i-th column's metadata, slot names.
+            /// </summary>
             private readonly VBuffer<ReadOnlyMemory<char>>[] _slotNames;
-            // Empty iff either header+ not set in args, or if no header present, or upon load
-            // there was no header stored in the model.
+            /// <summary>
+            /// Empty iff either header+ not set in args, or if no header present, or upon load
```
TomFinley: >either header+ not set in args [](start = 26, length = 30)

I might prefer something like an actual reference using a `<see` tag to the `Arguments.HasHeader` field, given that we're now making this an XML comment. Not a big deal though.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
wschin: `name type` is changed to `name and type`. Thanks.

---
In reply to: [247681984](https://github.com/dotnet/machinelearning/pull/2140#discussion_r247681984) [](ancestors = 247681984)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
wschin: Now we have
```charp
            /// <summary>
            /// Empty if <see cref="ArgumentsCore.HasHeader"/> is false, no header presents, or upon load
            /// there was no header stored in the model.
            /// </summary>
```

---
In reply to: [247682367](https://github.com/dotnet/machinelearning/pull/2140#discussion_r247682367) [](ancestors = 247682367)

wschin:
@TomFinley, Thank you! Let me open an issue for cleaning `TextLoader.ColInfo` because some utility functions rely on that type so removing may incur a mid-scale re-factorization.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -520,26 +520,26 @@ public static ColInfo Create(string name, PrimitiveType itemType, Segment[] segs
...
-            public readonly Dictionary<string, int> NameToInfoIndex;
+            /// <summary>
+            /// <see cref="Infos"/>[i] stores the i-th column's metadata, named <see cref="MetadataUtils.Kinds.SlotNames"/>
+            /// in <see cref="Schema.Metadata"/>.
+            /// </summary>
             private readonly VBuffer<ReadOnlyMemory<char>>[] _slotNames;
-            // Empty iff either header+ not set in args, or if no header present, or upon load
-            // there was no header stored in the model.
+            /// <summary>
+            /// Empty if <see cref="ArgumentsCore.HasHeader"/> is false, no header presents, or upon load
```
sfilipi: >false [](start = 66, length = 5)

<see langward="false" />

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -520,26 +520,26 @@ public static ColInfo Create(string name, PrimitiveType itemType, Segment[] segs
...
 
-            private readonly MetadataUtils.MetadataGetter<VBuffer<ReadOnlyMemory<char>>> _getSlotNames;
-
-            public Schema AsSchema { get; }
-
-            private Bindings()
-            {
-                _getSlotNames = GetSlotNames;
-            }
+            public Schema OutputSchema { get; }
```
sfilipi: >OutputSchema [](start = 26, length = 12)

we'd want comments on anything public.

PR #2142, opened 2019-01-14
Refactor GAM Predictor to be Public-ly Creatable

### LINKED ISSUE #1948: Public constructors for GamModelParameters
Part of #1698 

Right now, in `BinaryClassificationGamModelParameters` and `RegressionGamModelParameters` we have constructors that look like the following;

```csharp
internal BinaryClassificationGamModelParameters(IHostEnvironment env, int inputLength, Dataset trainset,
            double meanEffect, double[][] binEffects, int[] featureMap)
            : base(env, LoaderSignature, inputLength, trainset, meanEffect, binEffects, featureMap) { }
```

We want to make public constructors for all `ModelParameters` (previously `predictors`). However, for GAMs, this requires the user to pass in a training set, which is a `Dataset` object in `Microsoft.ML.FastTree.Internal`. This is not something we should expect the end user of the API to be aware of, so the constructor was not made public. Currently, `Dataset` is needed in `GamModelParametersBase` to map features to flocks and get bin upper bounds. We should rewrite GAMs so that this information is passed on from the Trainer rather than be computed when constructing the `GamModelParameters`.

rogancarr:
This PR refactors the GAM trainer and predictor such that all training information remains in the trainer, and the GAM predictor is for any generic (binned) GAM.

Fixes #1948

**Update**:  Two small bug fixes included in this PR
1. Fixed a serialization error when reading in GAM models written prior to 0.6. Now we issue a warning with a workaround.
2. Fixed a scoring error for sparse datasets.


src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -0,0 +1,909 @@
...
+using Microsoft.ML.Trainers.FastTree;
+using Microsoft.ML.Trainers.FastTree.Internal;
+using Microsoft.ML.Training;
+
+[assembly: LoadableClass(typeof(GamModelParametersBase.VisualizationCommand), typeof(GamModelParametersBase.VisualizationCommand.Arguments), typeof(SignatureCommand),
+    "GAM Vizualization Command", GamModelParametersBase.VisualizationCommand.LoadName, "gamviz", DocName = "command/GamViz.md")]
+
+namespace Microsoft.ML.Trainers.FastTree
+{
+    public abstract class GamModelParametersBase : ModelParametersBase<float>, IValueMapper, ICalculateFeatureContribution,
```
sfilipi: >public [](start = 4, length = 6)

a line of XML in anything public..

rogancarr: Done.

src/Microsoft.ML.FastTree/GamClassification.cs
```diff
@@ -173,8 +175,10 @@ private static VersionInfo GetVersionInfo()
         {
             return new VersionInfo(
                 modelSignature: "GAM BINP",
-                verWrittenCur: 0x00010001,
-                verReadableCur: 0x00010001,
+                // verWrittenCur: 0x00010001, // Initial
+                // verWrittenCur: 0x00010001, // Added Intercept but collided from release 0.6-0.9
+                verWrittenCur: 0x00020001,    // Added Intercept (version revved to address collisions)
```
Ivanidzo4ka: >0x00020001 [](start = 31, length = 10)

I think in general we upgrade version by last digit, i.e. `0x0010002` instead of `0x00020001`

src/Microsoft.ML.FastTree/GamClassification.cs
```diff
@@ -157,14 +156,17 @@ protected override SchemaShape.Column[] GetOutputColumnsCore(SchemaShape inputSc
...
+    /// </summary>
+    public sealed class BinaryClassificationGamModelParameters : GamModelParametersBase, IPredictorProducing<float>
     {
         internal const string LoaderSignature = "BinaryClassGamPredictor";
         public override PredictionKind PredictionKind => PredictionKind.BinaryClassification;
 
-        internal BinaryClassificationGamModelParameters(IHostEnvironment env, int inputLength, Dataset trainset,
-            double meanEffect, double[][] binEffects, int[] featureMap)
-            : base(env, LoaderSignature, inputLength, trainset, meanEffect, binEffects, featureMap) { }
+        public BinaryClassificationGamModelParameters(IHostEnvironment env,
```
artidoro: >BinaryClassificationGamModelParameters [](start = 15, length = 38)

Since you are making this public, could you also add XML documentation of what it does and what the parameters are?

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -0,0 +1,911 @@
...
+
+        ColumnType IValueMapper.InputType => _inputType;
+        ColumnType IValueMapper.OutputType => _outputType;
+
+        /// <summary>
+        /// Used to determine the contribution of each feature to the score of an example by <see cref="FeatureContributionCalculatingTransformer"/>.
+        /// For Generalized Additive Models (GAM), the contribution of a feature is equal to the shape function for the given feature evaluated at
+        /// the feature value.
+        /// </summary>
+        public FeatureContributionCalculator FeatureContributionClaculator => new FeatureContributionCalculator(this);
```
Ivanidzo4ka: >FeatureContributionCalculator [](start = 15, length = 29)

this class has one internal property and one internal method.
I advice to make it internal instead of public (as class in general and as property of Gam predictor)

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -0,0 +1,911 @@
...
+        }
+
+        /// <summary>
+        /// Get the bin upper bounds for each feature.
+        /// </summary>
+        /// <param name="featureIndex">The index of the feature (in the training vector) to get.</param>
+        /// <returns>The bin upper bounds. May be null if this feature has no bins.</returns>
+        public double[] GetFeatureBinUpperBounds(int featureIndex)
+        {
+            Contracts.Assert(0 <= featureIndex && featureIndex < _numFeatures);
```
Ivanidzo4ka: >Assert [](start = 22, length = 6)

use Check for public methods.
Also give it proper message.

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -0,0 +1,911 @@
...
+    /// <summary>
+    /// The base class for GAM Model Parameters.
+    /// </summary>
+    public abstract class GamModelParametersBase : ModelParametersBase<float>, IValueMapper, ICalculateFeatureContribution,
+        IFeatureContributionMapper, ICanSaveInTextFormat, ICanSaveSummary, ICanSaveInIniFormat
+    {
+        private readonly double[][] _binUpperBounds;
+        private readonly double[][] _binEffects;
+        public readonly double Intercept;
+        private readonly int _numFeatures;
```
Ivanidzo4ka: >_numFeatures [](start = 29, length = 12)

I don't see any way for user to access this, but in same time it's used as constrain in public method.
Can we expose it with proper description?

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -0,0 +1,911 @@
...
+        }
+
+        /// <summary>
+        /// Get the binned weights for each feature.
+        /// </summary>
+        /// <param name="featureIndex">The index of the feature (in the training vector) to get.</param>
+        /// <returns>The binned weights for each feature.</returns>
+        public double[] GetFeatureWeights(int featureIndex)
+        {
+            Contracts.Assert(0 <= featureIndex && featureIndex < _numFeatures);
```
Ivanidzo4ka: >Assert [](start = 22, length = 6)

check and proper message.

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -0,0 +1,911 @@
...
+            binIndex = Algorithms.FindFirstGE(_binUpperBounds[featureIndex], featureValue);
+            return _binEffects[featureIndex][binIndex];
+        }
+
+        /// <summary>
+        /// Get the bin upper bounds for each feature.
+        /// </summary>
+        /// <param name="featureIndex">The index of the feature (in the training vector) to get.</param>
+        /// <returns>The bin upper bounds. May be null if this feature has no bins.</returns>
+        public double[] GetFeatureBinUpperBounds(int featureIndex)
```
artidoro: >GetFeatureBinUpperBounds [](start = 24, length = 24)

Is there a reason why we don't want to return the upper bounds for all features with an array of arrays?

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -0,0 +1,911 @@
...
+        /// </summary>
+        /// <param name="featureIndex">The index of the feature (in the training vector) to get.</param>
+        /// <returns>The binned weights for each feature.</returns>
+        public double[] GetFeatureWeights(int featureIndex)
+        {
+            Contracts.Assert(0 <= featureIndex && featureIndex < _numFeatures);
+            double[] featureWeights;
+            if (_inputFeatureToDatasetFeatureMap.TryGetValue(featureIndex, out int j))
+            {
+                featureWeights = new double[_binUpperBounds[j].Length];
```
artidoro: >_binUpperBounds [](start = 44, length = 15)

Just checking that this is the same length as _binEffects[j]

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -0,0 +1,911 @@
...
+                leafValues: leafValues,
+                // Ignored arguments
+                splitGain: new double[numInternalNodes],
+                defaultValueForMissing: new float[numInternalNodes],
+                categoricalSplitFeatures: new int[numInternalNodes][],
+                categoricalSplit: new bool[numInternalNodes]);
+        }
+
+        /// <summary>
+        /// The GAM model visualization command. Because the data access commands must access private members of
```
Ivanidzo4ka: >must access private members [](start = 82, length = 27)

I think it's a sign what we haven't properly design predictor.
From what I see it needs:
`_inputType.ValueCount` which is same as `_numFeatures` which I think we need to expose.
`_inputFeatureToDatasetFeatureMap` , `_binEffects` and `_featureMap` First two I think you use in `GetFeatureBinUpperBounds` and it can probably be rewritten. last one can be made as internal.

I found it's strange what we make command line thing which has more functionality than user can achieve by himself.

src/Microsoft.ML.FastTree/GamModelParameters.cs
Ivanidzo4ka: >Contracts [](start = 12, length = 9)

You have Host.

src/Microsoft.ML.FastTree/GamModelParameters.cs
Ivanidzo4ka: >Contracts [](start = 12, length = 9)

Please use Host.

Ivanidzo4ka:
> using System;

no header???!!!11

---
Refers to: src/Microsoft.ML.FastTree/GamModelParameters.cs:1 in add0cc8. [](commit_id = add0cc8b83eed03581562ebb6dd8855a82ffecd1, deletion_comment = False)

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -0,0 +1,911 @@
...
+                        // Add the value and subtract the value at zero that was previously accounted for
+                        value += GetBinEffect(j, featuresValues[i], out bins[j]) - GetBinEffect(j, 0);
+                }
+            }
+            return value;
+        }
+
+        private double GetBinEffect(int featureIndex, double featureValue)
+        {
+            Contracts.Assert(0 <= featureIndex && featureIndex < _numFeatures);
```
Ivanidzo4ka: >Contracts [](start = 12, length = 9)

Please change `Contracts` to `Host` where it possible.

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -0,0 +1,911 @@
...
+                var left = CreateBalancedTreeRecursive(
+                    lower, mid - 1, internalNodeIndices, lteChild, gtChild, ref internalNodeId);
+                var right = CreateBalancedTreeRecursive(
+                    mid + 1, upper, internalNodeIndices, lteChild, gtChild, ref internalNodeId);
+                internalNodeIndices.Insert(0, mid);
+                lteChild.Insert(0, left);
+                gtChild.Insert(0, right);
+                return --internalNodeId;
+            }
+        }
```
Ivanidzo4ka: new line

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Can you clarify what you mean here?

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: No real reason. Let's add this as a separate issue.

src/Microsoft.ML.FastTree/GamModelParameters.cs
Ivanidzo4ka: This thing gonna be visible to any user. In same time if you look on content of this class, it has only internal objects.
Basically during debugging user will see FeatureContributionCalculator, but it will have zero public methods and properties.
We have that thing for a reason, but I doubt it should be public.

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Good catch! Should be the same and checked when defined, but not good practice. Changing to use _binEffects[j].Length.

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Point of confusion. Currently, there can be fewer shape functions built than input features because of flocking (I think this is a bug and will file one separately.)

`_inputType.ValueCount` is _different_ than `_numFeatures`. I've rename `numFeatures` to `NumShapeFunctions` to clarify this.

`inputFeaturesToDatasetFeatureMap` has been renamed to `_inputFeatureToShapeFunctionMap`

`featureMap` has been renamed to `shapeToInputMap`. (and is `private`).

> I found it's strange what we make command line thing which has more functionality than user can achieve by himself.

 You mean the visualization command? With the new public interfaces, the user should be able to hand-edit models and create versions with the modified shapes.

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Done

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Done.

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Done.

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Done.

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Done.

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Done.

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Done.

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: @artidoro Does this method need to be `public` to work?

src/Microsoft.ML.FastTree/GamClassification.cs
rogancarr: Done.

src/Microsoft.ML.FastTree/GamClassification.cs
rogancarr: Done.

src/Microsoft.ML.FastTree/GamModelParameters.cs
Ivanidzo4ka: >_inputType.ValueCount is different than _numFeatures.

If `featureToInputMap` is null, I think they should be same, otherwise it's a size of `featureToInputMap`. Sorry, didn't properly check constructor of predictor.

>You mean the visualization command? With the new public interfaces, the user should be able to hand-edit models and create versions with the modified shapes.
I just find it strange what we have nested internal command in predictor. And only reason why it's nested is because we using some internal guts of predictor. I understand what you can do hand-edits, but can user visualize model same way how our visualize command work? If not, it's probably sign what we didn't expose enough.

rogancarr: Ah, I see what you are saying. The visualization command is for an internal tool that may not be released.

TomFinley:
Hi @rogancarr I'm having trouble getting started reading this, since I can't quite find where the predictors have been changed to use generics?

rogancarr:
> Hi @rogancarr I'm having trouble getting started reading this, since I can't quite find where the predictors have been changed to use generics?

Hi @TomFinley! Apologies for the confusion. I renamed this PR to make more sense. This PR changes the GAM predictor s.t. it can be created with the output of any GAM training (with binned results). This is part of the move to create public constructors for our `ModelParameters` classes.

It does not have anything to do with `generics` in the programming language sense.

Issue #1198, opened 2018-10-09
Revise type mappings in MLNET-to-ONNX conversion


justinormont:
Currently, for ONNX, we are mapping a `U4` datatype (an unsigned 32-bit integer) to an `Int64`.

Should we be instead mapping the `U4` datatype to `Uint32` in ONNX? Or is there no support for a `Uint32`, and we're storing in an `Int64`?

https://github.com/dotnet/machinelearning/blob/96439752cbaad099093ca0e8c770576bf3a53ad4/src/Microsoft.ML.Onnx/OnnxUtils.cs#L329-L331

In the [above code](https://github.com/dotnet/machinelearning/blob/96439752cbaad099093ca0e8c770576bf3a53ad4/src/Microsoft.ML.Onnx/OnnxUtils.cs#L329-L331), you'll notice the mapping is currently:
* `BL` to `Float` 
* `TX` to `String`
* `I1` to `Int8`
* `U1` to `Uint8`
* `I2` to `Int16`
* `U2` to `Uint16`
* `I4` to `Int32`
* `U4` to `Int64`   <- **This one is odd**
* `I8` to `Int64`
* `U8` to `Uint64`
* `R4` to `Float`
* `R8` to `Double`

The `BL` to `Float` & `U4` to `Int64` seem odd.

@wschin noted we have been mapping `U4` to `Int64` for the last two releases of WinML: https://github.com/dotnet/machinelearning/pull/947#discussion_r223443413

TomFinley:
Indeed there are now 4 and 8 byte unsigned int types in ONNX, see:

https://github.com/dotnet/machinelearning/blob/96439752cbaad099093ca0e8c770576bf3a53ad4/src/Microsoft.ML.Onnx/OnnxMl.cs#L2969-L2970

We should use them, now that they exist.

justinormont:
@TomFinley : Should we also map `BL` to `Bool`, or is there a reason behind `Float`?

`Bool` is available:
https://github.com/dotnet/machinelearning/blob/96439752cbaad099093ca0e8c770576bf3a53ad4/src/Microsoft.ML.Onnx/OnnxMl.cs#L2962-L2964

Current `BL` mapping is to `Float`:
https://github.com/dotnet/machinelearning/blob/96439752cbaad099093ca0e8c770576bf3a53ad4/src/Microsoft.ML.Onnx/OnnxUtils.cs#L308-L310

wschin:
[...]

PR #2143, opened 2019-01-15
Remove ColumnType.RawKind usages Round 1.

eerhardt:
Remove all usages of RawKind that are outside of ML.Core and ML.Data assemblies. The next round will completely remove ColumnType.RawKind.

### DIFF
--- a/src/Microsoft.ML.Onnx/OnnxUtils.cs
+++ b/src/Microsoft.ML.Onnx/OnnxUtils.cs
@@ -327,7 +327,7 @@ namespace Microsoft.ML.Model.Onnx
                 dataType = TensorProto.Types.DataType.Double;
             else
             {
-                string msg = "Unsupported type: " + rawType.ToString();
+                string msg = "Unsupported type: " + type.ToString();
                 Contracts.Check(false, msg);
             }
 
--- a/src/Microsoft.ML.Onnx/OnnxUtils.cs
+++ b/src/Microsoft.ML.Onnx/OnnxUtils.cs
@@ -295,56 +295,40 @@ namespace Microsoft.ML.Model.Onnx
             Contracts.CheckNonEmpty(colName, nameof(colName));
 
             TensorProto.Types.DataType dataType = TensorProto.Types.DataType.Undefined;
-            DataKind rawKind;
+            Type rawType;
             if (type is VectorType vectorType)
-                rawKind = vectorType.ItemType.RawKind;
-            else if (type is KeyType keyType)
-                rawKind = keyType.RawKind;
+                rawType = vectorType.ItemType.RawType;
             else
-                rawKind = type.RawKind;
+                rawType = type.RawType;
 
-            switch (rawKind)
+            if (rawType == typeof(bool))
+                dataType = TensorProto.Types.DataType.Float;
+            else if (rawType == typeof(ReadOnlyMemory<char>))
+                dataType = TensorProto.Types.DataType.String;
+            else if (rawType == typeof(sbyte))
+                dataType = TensorProto.Types.DataType.Int8;
+            else if (rawType == typeof(byte))
+                dataType = TensorProto.Types.DataType.Uint8;
+            else if (rawType == typeof(short))
+                dataType = TensorProto.Types.DataType.Int16;
+            else if (rawType == typeof(ushort))
+                dataType = TensorProto.Types.DataType.Uint16;
+            else if (rawType == typeof(int))
+                dataType = TensorProto.Types.DataType.Int32;
+            else if (rawType == typeof(uint))
+                dataType = TensorProto.Types.DataType.Int64;
+            else if (rawType == typeof(long))
+                dataType = TensorProto.Types.DataType.Int64;
+            else if (rawType == typeof(ulong))
+                dataType = TensorProto.Types.DataType.Uint64;
+            else if (rawType == typeof(float))
+                dataType = TensorProto.Types.DataType.Float;
+            else if (rawType == typeof(double))
+                dataType = TensorProto.Types.DataType.Double;
+            else
             {
-                case DataKind.BL:
-                    dataType = TensorProto.Types.DataType.Float;
-                    break;
-                case DataKind.TX:
-                    dataType = TensorProto.Types.DataType.String;
-                    break;
-                case DataKind.I1:
-                    dataType = TensorProto.Types.DataType.Int8;
-                    break;
-                case DataKind.U1:
-                    dataType = TensorProto.Types.DataType.Uint8;
-                    break;
-                case DataKind.I2:
-                    dataType = TensorProto.Types.DataType.Int16;
-                    break;
-                case DataKind.U2:
-                    dataType = TensorProto.Types.DataType.Uint16;
-                    break;
-                case DataKind.I4:
-                    dataType = TensorProto.Types.DataType.Int32;
-                    break;
-                case DataKind.U4:
-                    dataType = TensorProto.Types.DataType.Int64;
-                    break;
-                case DataKind.I8:
-                    dataType = TensorProto.Types.DataType.Int64;
-                    break;
-                case DataKind.U8:
-                    dataType = TensorProto.Types.DataType.Uint64;
-                    break;
-                case DataKind.R4:
-                    dataType = TensorProto.Types.DataType.Float;
-                    break;
-                case DataKind.R8:
-                    dataType = TensorProto.Types.DataType.Double;
-                    break;
-                default:
-                    string msg = "Unsupported type: DataKind " + rawKind.ToString();
-                    Contracts.Check(false, msg);
-                    break;
+                string msg = "Unsupported type: " + rawType.ToString();
+                Contracts.Check(false, msg);
             }
 
             string name = colName;
--- a/src/Microsoft.ML.StaticPipe/StaticSchemaShape.cs
+++ b/src/Microsoft.ML.StaticPipe/StaticSchemaShape.cs
@@ -134,7 +134,7 @@ namespace Microsoft.ML.StaticPipe.Runtime
 
             if (col.IsKey)
             {
-                Type physType = StaticKind(col.ItemType.RawKind);
+                Type physType = GetPhysicalType(col.ItemType);
                 Contracts.Assert(physType == typeof(byte) || physType == typeof(ushort)
                     || physType == typeof(uint) || physType == typeof(ulong));
                 var keyType = typeof(Key<>).MakeGenericType(physType);
@@ -158,7 +158,7 @@ namespace Microsoft.ML.StaticPipe.Runtime
 
             if (col.ItemType is PrimitiveType pt)
             {
-                Type physType = StaticKind(pt.RawKind);
+                Type physType = GetPhysicalType(pt);
                 // Though I am unaware of any existing instances, it is theoretically possible for a
                 // primitive type to exist, have the same data kind as one of the existing types, and yet
                 // not be one of the built in types. (For example, an outside analogy to the key types.) For this
@@ -266,7 +266,7 @@ namespace Microsoft.ML.StaticPipe.Runtime
 
             if (t is KeyType kt)
             {
-                Type physType = StaticKind(kt.RawKind);
+                Type physType = GetPhysicalType(kt);
                 Contracts.Assert(physType == typeof(byte) || physType == typeof(ushort)
                     || physType == typeof(uint) || physType == typeof(ulong));
                 var keyType = kt.Count > 0 ? typeof(Key<>) : typeof(VarKey<>);
@@ -302,7 +302,7 @@ namespace Microsoft.ML.StaticPipe.Runtime
 
             if (t is PrimitiveType pt)
             {
-                Type physType = StaticKind(pt.RawKind);
+                Type physType = GetPhysicalType(pt);
                 // Though I am unaware of any existing instances, it is theoretically possible for a
                 // primitive type to exist, have the same data kind as one of the existing types, and yet
                 // not be one of the built in types. (For example, an outside analogy to the key types.) For this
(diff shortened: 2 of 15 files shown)


src/Microsoft.ML.Onnx/OnnxUtils.cs
```diff
@@ -295,56 +295,40 @@ public ModelArgs(string name, TensorProto.Types.DataType dataType, List<long> di
...
             if (type is VectorType vectorType)
-                rawKind = vectorType.ItemType.RawKind;
-            else if (type is KeyType keyType)
-                rawKind = keyType.RawKind;
+                rawType = vectorType.ItemType.RawType;
+            else
+                rawType = type.RawType;
+
+            if (rawType == typeof(bool))
+                dataType = TensorProto.Types.DataType.Float;
```
sfilipi: >Float [](start = 54, length = 5)

bool?

src/Microsoft.ML.Transforms/Text/NgramUtils.cs
```diff
@@ -200,4 +201,18 @@ private bool ProcessSkipNgrams(int icol, int i, int skips)
...
+    {
+        public static bool IsValidNgramRawType(Type rawType)
+        {
+            // Can only accept key types that can be converted to U4 (uint).
+            return rawType == typeof(sbyte)
+                || rawType == typeof(byte)
+                || rawType == typeof(short)
+                || rawType == typeof(ushort)
+                || rawType == typeof(int)
+                || rawType == typeof(uint);
```
sfilipi: >|| rawType == typeof(uint); [](start = 16, length = 27)

since the condition was: type.ItemType.RawKind > DataKind.U4, should it accept string?

src/Microsoft.ML.StaticPipe/StaticSchemaShape.cs
```diff
@@ -327,34 +327,21 @@ private static Type GetTypeOrNull(Schema.Column col)
...
+                case DateTimeType dateTimeType:
+                case DateTimeOffsetType dateTimeOffsetType:
+                case BoolType boolType:
+                    return columnType.RawType;
+                case TextType textType:
+                    return typeof(string);
 
                 default:
-                    throw Contracts.ExceptParam(nameof(kind), $"Unrecognized type '{kind}'");
+                    return null;
```
sfilipi: >return null; [](start = 20, length = 12)

curious, why change to return null, rather than throw? If an invalid type gets passed, would it be better to throw?

src/Microsoft.ML.Onnx/OnnxUtils.cs
eerhardt: I was wondering the same thing when I made the change...

src/Microsoft.ML.Transforms/Text/NgramUtils.cs
eerhardt: I inverted the condition. Before, the code was checking for `RawKind > DataKind.U4` and returning `false` Now, this check is checking for "valid" types - ones that are `<= DataKind.U4`. (Although, I did make a change that `0` and negative were "valid" before, but they are no longer.)

src/Microsoft.ML.StaticPipe/StaticSchemaShape.cs
eerhardt: I was trying to preserve the first line's behavior:

```
case default(DataKind): return null;
```

When DataKind is `default`, it usually means it is a "custom" or "structured" type - like `VectorType` or `ImageType`. So I was preserving that behavior here - when you see an unrecognized `ColumnType` - return `null`.

I assume the existing `throw` condition was in place for handling invalid `enum` values - like `29` and `-451`.

---
In reply to: [247782795](https://github.com/dotnet/machinelearning/pull/2143#discussion_r247782795) [](ancestors = 247782795)

src/Microsoft.ML.Onnx/OnnxUtils.cs
```diff
@@ -295,56 +295,40 @@ public ModelArgs(string name, TensorProto.Types.DataType dataType, List<long> di
...
             TensorProto.Types.DataType dataType = TensorProto.Types.DataType.Undefined;
-            DataKind rawKind;
+            Type rawType;
             if (type is VectorType vectorType)
-                rawKind = vectorType.ItemType.RawKind;
-            else if (type is KeyType keyType)
-                rawKind = keyType.RawKind;
+                rawType = vectorType.ItemType.RawType;
+            else
+                rawType = type.RawType;
```
TomFinley: >RawType [](start = 31, length = 7)

I am not sure operating over `RawType` is an appropriate alternative to `RawKind`, in this case. I might prefer that we operate over the exact `ColumnType` implementations, since that is more descriptive of what we're trying to do I think.

To give a concrete example, I see the code below is conflating `uint` as being identical, even thought the same .NET type is used for both raw `uint`s and actual key types... same .NET type, very different information being encoded. This might be a sign that the entire function is screwed up (I see @sfilipi found what appears to be a bug below with booleans, so that's well within the realm of possibility), but I'm a little concerned that this is the first thing I see.

src/Microsoft.ML.StaticPipe/StaticSchemaShape.cs
```diff
@@ -327,34 +327,22 @@ private static Type GetTypeOrNull(Schema.Column col)
...
-                case DataKind.DateTimeZone: return typeof(DateTimeOffset);
+                case NumberType numberType:
+                case KeyType keyType:
+                case TimeSpanType timeSpanType:
+                case DateTimeType dateTimeType:
+                case DateTimeOffsetType dateTimeOffsetType:
+                case BoolType boolType:
+                    return columnType.RawType;
+                case TextType textType:
+                    return typeof(string);
```
TomFinley: FYI this style of change you have here is more what I'd be more comfortable with in the the `OnnxUtils` case. (Now, in that other case, it would also probably benefit from some `when` statements to further refine.)

src/Microsoft.ML.Transforms/KeyToVectorMapping.cs
```diff
@@ -478,7 +478,7 @@ public override SchemaShape GetOutputSchema(SchemaShape inputSchema)
             {
                 if (!inputSchema.TryFindColumn(colInfo.Input, out var col))
                     throw Host.ExceptSchemaMismatch(nameof(inputSchema), "input", colInfo.Input);
-                if ((col.ItemType.ItemType.RawKind == default) || !(col.ItemType.IsVector || col.ItemType is PrimitiveType))
+                if (!(col.ItemType.IsVector || col.ItemType is PrimitiveType))
```
TomFinley: >ItemType [](start = 51, length = 8)

I think this is probably going to change anyway with your other change on vectors, but just one thing to keep in mind: While `ItemType` on `ColumnType` itself was a `ColumnType` (which is correct), `ItemType` on the vector type itself is and always has been a `PrimitiveType` anyway. So this test here seems at least odd

src/Microsoft.ML.Transforms/MissingValueReplacing.cs
```diff
@@ -896,16 +896,14 @@ public void SaveAsOnnx(OnnxContext ctx)
...
             private bool SaveAsOnnxCore(OnnxContext ctx, int iinfo, ColInfo info, string srcVariableName, string dstVariableName)
             {
-                DataKind rawKind;
+                Type rawType;
                 var type = _infos[iinfo].TypeSrc;
                 if (type is VectorType vectorType)
-                    rawKind = vectorType.ItemType.RawKind;
-                else if (type is KeyType keyType)
-                    rawKind = keyType.RawKind;
+                    rawType = vectorType.ItemType.RawType;
```
TomFinley: >rawType = vectorType.ItemType.RawType; [](start = 20, length = 38)

Placeholder note, once your other PR goes in you could probably benefit from your new `GetItemType` extension.

src/Microsoft.ML.Transforms/Text/NgramTransform.cs
```diff
@@ -843,7 +843,7 @@ internal static bool IsColumnTypeValid(ColumnType type)
             if (!(type.ItemType is KeyType itemKeyType))
                 return false;
             // Can only accept key types that can be converted to U4.
-            if (itemKeyType.Count == 0 && type.ItemType.RawKind > DataKind.U4)
+            if (itemKeyType.Count == 0 && !NgramUtils.IsValidNgramRawType(itemKeyType.RawType))
```
TomFinley: >IsValidNgramRawType(itemKeyType.RawType [](start = 54, length = 39)

If I read the comment, it says, `Can only accept key types that can be converted to U4,` in every place that this `IsValidNgramRawType` is being called. That basically means U1, U2, and U4 are acceptable, and only U8 is not. So I'd have expected, *based on this comment*, that this would be a test against whether the raw type is a `ulong` or not.

*However*, if I then go and read this `IsValidNgramRawType`, it is doing something else! It is checking against all sorts of types that have nothing to do with key-types *at all*, including `sbyte`, `short`, `int`... which makes no sense to me, given the comment, the fact that these are *definitely* meant to operate over keytypes, etc.

I would have expected just making sure it isn't a `ulong` would have been sufficient, but...

test/Microsoft.ML.Core.Tests/UnitTests/ScoreSchemaTest.cs
```diff
@@ -47,7 +47,7 @@ public void SequencePredictorSchemaTest()
             Assert.True(scoreColumn.Type is KeyType);
             Assert.Equal((scoreColumnType as KeyType).Min, (scoreColumn.Type as KeyType).Min);
             Assert.Equal((scoreColumnType as KeyType).Count, (scoreColumn.Type as KeyType).Count);
-            Assert.Equal((scoreColumnType as KeyType).RawKind, (scoreColumn.Type as KeyType).RawKind);
+            Assert.Equal((scoreColumnType as KeyType).RawType, (scoreColumn.Type as KeyType).RawType);
```
TomFinley: >scoreColumn.Type as KeyType [](start = 64, length = 27)

This was not your code, and given that it's in a test code we might consider it lower priority, but this was so offensive that I have to comment on it. And you're touching it. :D :D

This code has a lot of nonsense like this:

```csharp
Assert.Equal((scoreColumnType as KeyType).Min, (scoreColumn.Type as KeyType).Min);
```

That makes no sense. The only conceivable reason to use an `as` style cast is if you're anticipating that it *might* not be this case. But then the code turns around and *directly accesses a property* without a `?.`, so that if it was in fact not that type we'd throw with a null reference exception. Which is really silly. I don't know if you want to fix all this code so that it is less offensive, but maybe at least the things you're touching directly could be written to be a bit more sane.

Not blocking, if you don't feel like doing it, just, the code in its current form makes my skin crawl.

src/Microsoft.ML.Onnx/OnnxUtils.cs
```diff
@@ -295,56 +295,40 @@ public ModelArgs(string name, TensorProto.Types.DataType dataType, List<long> di
...
+            else if (rawType == typeof(byte))
+                dataType = TensorProto.Types.DataType.Uint8;
+            else if (rawType == typeof(short))
+                dataType = TensorProto.Types.DataType.Int16;
+            else if (rawType == typeof(ushort))
+                dataType = TensorProto.Types.DataType.Uint16;
+            else if (rawType == typeof(int))
+                dataType = TensorProto.Types.DataType.Int32;
+            else if (rawType == typeof(uint))
+                dataType = TensorProto.Types.DataType.Int64;
```
eerhardt: @wschin - this one looks incorrect as well.

eerhardt: It appears both BL and U4 mapped these values when this repo was created - [https://github.com/dotnet/machinelearning/blob/f0e639af5ffdc839aae8e65d19b5a9a1f0db634a/src/Microsoft.ML.Data/Model/Onnx/OnnxUtils.cs#L296-L328](https://github.com/dotnet/machinelearning/blob/f0e639af5ffdc839aae8e65d19b5a9a1f0db634a/src/Microsoft.ML.Data/Model/Onnx/OnnxUtils.cs#L296-L328)

---
In reply to: [248108591](https://github.com/dotnet/machinelearning/pull/2143#discussion_r248108591) [](ancestors = 248108591)

src/Microsoft.ML.Onnx/OnnxUtils.cs
eerhardt: > I am not sure operating over RawType is an appropriate alternative to RawKind, in this case

I don't think I agree. In this case we are mapping from ML.NET's `ColumnType` to Onnx's `TensorProto.Types.DataType`. It doesn't really matter if we are using a `KeyType` or `NumberType.U4`. In either case, we need to get a `TensorProto.Types.DataType.UInt32`. Actually - what would we return for `KeyType`? We wouldn't want to hard-code it to just `TensorProto.Types.DataType.UInt32`, would we? What if it was a KeyType that was using `UInt16`?


---
In reply to: [248033212](https://github.com/dotnet/machinelearning/pull/2143#discussion_r248033212) [](ancestors = 248033212)

src/Microsoft.ML.Onnx/OnnxUtils.cs
```diff
@@ -295,56 +295,40 @@ public ModelArgs(string name, TensorProto.Types.DataType dataType, List<long> di
...
-                    dataType = TensorProto.Types.DataType.Float;
-                    break;
-                case DataKind.R8:
-                    dataType = TensorProto.Types.DataType.Double;
-                    break;
-                default:
-                    string msg = "Unsupported type: DataKind " + rawKind.ToString();
-                    Contracts.Check(false, msg);
-                    break;
+                string msg = "Unsupported type: " + rawType.ToString();
```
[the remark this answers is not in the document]
eerhardt: Agreed. Fixed.

---
In reply to: [248032092](https://github.com/dotnet/machinelearning/pull/2143#discussion_r248032092) [](ancestors = 248032092)

src/Microsoft.ML.Transforms/Text/NgramTransform.cs
eerhardt: I was just converting the existing logic: `type.ItemType.RawKind > DataKind.U4`.

I can change that method to assume that `KeyType`s can only be `byte`, `ushort`, `uint`, or `ulong`, and only make it check for `ulong`.

---
In reply to: [248037058](https://github.com/dotnet/machinelearning/pull/2143#discussion_r248037058) [](ancestors = 248037058)

src/Microsoft.ML.Transforms/KeyToVectorMapping.cs
eerhardt: Actually this is a bug (or at least confusion) in our API in my opinion.  `col` is of type `SchemaShape.Column`. It has a member on it - `public readonly ColumnType ItemType;`. So `col.ItemType` isn't actually the `VectorType.ItemType`. It could be _any_ `ColumnType` (for example, it could be an `ImageType`).

We then split out the "vector"-ness to a different `public readonly VectorKind Kind;` property.

---
In reply to: [248034844](https://github.com/dotnet/machinelearning/pull/2143#discussion_r248034844) [](ancestors = 248034844)

src/Microsoft.ML.Onnx/OnnxUtils.cs
eerhardt: also /cc @zeahmed and @jignparm - in case they understand both of these type mappings.

test/Microsoft.ML.Core.Tests/UnitTests/ScoreSchemaTest.cs
eerhardt: Fixed.

---
In reply to: [248038416](https://github.com/dotnet/machinelearning/pull/2143#discussion_r248038416) [](ancestors = 248038416)

src/Microsoft.ML.Onnx/OnnxUtils.cs
wschin: Since we always translate uint to ONNX Int64, this should be fine.

---
In reply to: [248112875](https://github.com/dotnet/machinelearning/pull/2143#discussion_r248112875) [](ancestors = 248112875)

src/Microsoft.ML.Onnx/OnnxUtils.cs
wschin: This probably is caused by the limited types accepted by old ONNX operators implemented in RS4. As ONNXRuntime supports much more ops than dark-age implementations, we should revise our type mappings and create more tests. This is a reasonable setting but it can be improved in another PR.

---
In reply to: [247951442](https://github.com/dotnet/machinelearning/pull/2143#discussion_r247951442) [](ancestors = 247951442)

eerhardt: Looks like we have an existing bug to make this better - https://github.com/dotnet/machinelearning/issues/1198

PR #2146, opened 2019-01-15
KeyType Simplification

### LINKED ISSUE #1540: IDataView Cleanup: KeyType Simplification
We have by now multiple years of experience with `KeyType`. In the documentation of the `IDatView` type system, [key types](https://github.com/dotnet/machinelearning/blob/master/docs/code/IDataViewTypeSystem.md#key-types) are described as having the following properties:

All of our current transformers produce key-types that are (1) known size and (2) have a minimum value of `0`, and are (3) contiguous, with absolutely no exceptions. Also if we consider how key-types are actually used (always, even in the hash transformer, considered enumerations into a set), then there is no particular rason for this. I therefore suggest that we simplify the key types by removing the properties describing this, and just make it a rule.

This does represent a loss of capability, but it is a capability that AFAIK has never been exercised beyond unit tests and small toy demonstrations to clarify to new developers, "by the way, our key-types have this really, really weird capability..."

On the flip side, I feel like having a *count* have the maximum of `int.MaxValue` is arguably too limiting. The reason for this is that it was anticipated that the only real usage of key values would be to project into a vector, but this is not true. I would like to replace it with a `ulong` maximum value instead.

artidoro:
Fixes #1540.

In this PR I remove the Min and Contiguous fields of KeyType. The reason for doing so is elaborated in the issue #1540.

Before this PR, a valid range of values for a key could be 1000 to 4000. This was represented by a key with Min 1000 and Count 4001. Having a key that starts at a value that is not 1 only very rarely happened in practice.

Now by default a valid range will start at 1, up to Count, with 0 being used for missing values.

As part of this PR I also removed the Min and Contiguous field of KeyRange (KeyRange serves the purpose of representing a valid range of values for a KeyType). It is useful to note that the max of a KeyRange will be equal to Count - 1 of the associated KeyType.

Also fixed the comments and the [documentation ](https://github.com/dotnet/machinelearning/blob/master/docs/code/IDataViewTypeSystem.md#key-types) on the type system to make sure they reflect the change.


artidoro:
Need to update the documentation on the datatypes [https://github.com/dotnet/machinelearning/blob/master/docs/code/IDataViewTypeSystem.md#key-types](https://github.com/dotnet/machinelearning/blob/master/docs/code/IDataViewTypeSystem.md#key-types).

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -554,43 +554,47 @@ public override bool Equals(ColumnType other)
...
         {
         }
 
         [BestFriend]
-        internal KeyType(DataKind kind, ulong min, int count, bool contiguous = true)
-            : this(ToRawType(kind), kind, min, count, contiguous)
+        internal KeyType(DataKind kind, int count)
+            : this(ToRawType(kind), kind, (ulong)count)
         {
+            Contracts.Assert(0 <= count);
```
yaeldekel: >Assert [](start = 22, length = 6)

Should this be a check?

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
```diff
@@ -21,6 +21,32 @@ internal static class ColumnTypeExtensions
...
+            ectx.Assert(count <= int.MaxValue, "KeyType range exceeds int.MaxValue.");
+            return (int)count;
+        }
+    }
+
+    /// <summary>
+    /// Extnesion methods related to the KeyType class.
+    /// </summary>
+    [BestFriend]
+    internal static class KeyTypExtensions
```
yaeldekel: Add an 'e' :-).

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -640,23 +644,14 @@ public static bool IsValidDataType(Type type)
         }
 
         /// <summary>
-        /// This is the Min of the key type for display purposes and conversion to/from text. The values
-        /// actually stored always start at 1 (for the smallest legal value), with zero being reserved
-        /// for "not there"/"none". Typical Min values are 0 or 1, but can be any value >= 0.
+        /// If this key type has a known cardinality, Count is that cardinality. Otherwise, this returns
```
yaeldekel: >Otherwise [](start = 81, length = 9)

I am wondering if there is an advantage to make the distinction between unknown cardinality and Count=ulong.MaxValue. In the Parse methods it seems like there doesn't need to be a distinction. Are there scenarios where the KeyType would behave differently for Count=0 vs. Count=ulong.MaxValue?

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
```diff
@@ -21,6 +21,32 @@ internal static class ColumnTypeExtensions
...
+    {
+        /// <summary>
+        /// Sometimes it is necessary to cast the Count to an int. This performs overflow check.
+        /// </summary>
+        public static int CheckRangeReturnCount(this KeyType key, IExceptionContext ectx = null)
+        {
+            ectx.Assert(key.Count <= int.MaxValue, "KeyType range exceeds int.MaxValue.");
```
TomFinley: >ectx.Assert [](start = 12, length = 11)

The method says `check`, but you're using an assert here. Should be an assert.

TomFinley: >range [](start = 60, length = 5)

`Count`, isn't it? Also just for the sake of making me happy maybe use something like `nameof(KeyType) + "." + nameof(KeyType.Count)` instead of `"KeyType range"` just to preserve whatever little is left of my sanity.

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
```diff
@@ -21,6 +21,32 @@ internal static class ColumnTypeExtensions
...
-        public static int GetKeyCount(this ColumnType columnType) => (columnType as KeyType)?.Count ?? 0;
+        public static ulong GetKeyCount(this ColumnType columnType) => (columnType as KeyType)?.Count ?? 0;
+
+        /// <summary>
+        /// Sometimes it is necessary to cast the Count to an int. This performs overflow check.
+        /// </summary>
+        public static int CheckRangeReturnCount(this ColumnType columnType, IExceptionContext ectx = null)
+        {
+            ulong count = columnType.GetKeyCount();
+            ectx.Assert(count <= int.MaxValue, "KeyType range exceeds int.MaxValue.");
```
TomFinley: >Assert [](start = 17, length = 6)

Another assert.

src/Microsoft.ML.Core/Data/MetadataUtils.cs
```diff
@@ -175,7 +175,7 @@ internal static KeyType ScoreColumnSetIdType
             {
                 if (_scoreColumnSetIdType == null)
                 {
-                    var type = new KeyType(DataKind.U4, 0, 0);
+                    var type = new KeyType(DataKind.U4, 0);
```
TomFinley: >KeyType(DataKind.U4, 0); [](start = 35, length = 24)

Interesting. I think we may have to do something else here. Maybe the count becomes, say, `int.MaxValue - 1` or something.

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -554,43 +554,47 @@ public override bool Equals(ColumnType other)
...
 
-            Contiguous = contiguous;
-            Min = min;
-            Count = count;
+        [BestFriend]
+        internal KeyType(Type type, int count)
+            : this(type, CheckRefRawType(type), (ulong)count)
+        {
+            Contracts.Assert(0 <= count);
         }
```
TomFinley: So, just an FYI, when @eerhardt factors out IDataView, I think he *might* also be moving the canonical types. (Though I am not certain if `KeyType` is part of this. It *might* be.) If that is so, we cannot rely upon this. Indeed much of his work as we see in #2131 and #2143 and many other things already checked in deals with reversing this. So rather than introducing more internal constructors, we should I hope just use the public one? Is that unreasonable?

src/Microsoft.ML.Core/Data/ColumnType.cs
TomFinley: Yes, we probably should not even have it at all though.

---
In reply to: [248102911](https://github.com/dotnet/machinelearning/pull/2146#discussion_r248102911) [](ancestors = 248102911)

src/Microsoft.ML.Core/Data/ColumnType.cs
TomFinley: Or we could have a separate public constructor since `int` is a *really* common case. But as @yaeldekel says, that should be a proper check.

---
In reply to: [248115787](https://github.com/dotnet/machinelearning/pull/2146#discussion_r248115787) [](ancestors = 248115787)

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -554,43 +554,47 @@ public override bool Equals(ColumnType other)
     /// class numbers, etc. For example, in multi-class classification, the label is typically
     /// a class number which is naturally a KeyType.
     ///
-    /// KeyTypes can be contiguous (the class number example), in which case they can have
-    /// a cardinality/Count. For non-contiguous KeyTypes the Count property returns zero.
-    /// Any KeyType (contiguous or not) can have a Min value. The Min value is always >= 0.
+    /// KeyTypes with a known range of values have a cardinality/Count. If the range is
```
TomFinley: >known range of values [](start = 24, length = 21)

All values are now of known range.

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -554,43 +554,47 @@ public override bool Equals(ColumnType other)
...
+    /// unknown the Count property returns zero.
```
TomFinley: >unknown the Count property returns zero. [](start = 8, length = 40)

There is no longer any such concept. Key types always have known count in the new world.

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -554,43 +554,47 @@ public override bool Equals(ColumnType other)
...
+        internal KeyType(Type type, int count)
+            : this(type, CheckRefRawType(type), (ulong)count)
+        {
+            Contracts.Assert(0 <= count);
         }
 
-        public KeyType(Type type, ulong min, int count, bool contiguous = true)
-            : this(type, CheckRefRawType(type), min, count, contiguous)
+        [BestFriend]
+        internal KeyType(DataKind kind, ulong count)
```
TomFinley: >ulong count [](start = 40, length = 11)

So, previously `Count = 0` indicated that the key type was unknown. You are still allowing count value of 0, while changing the behavior rather drastically.

Now we have a choice:

1. We retain the old behavior, that key-type of count = 0 while being of known sized is simply *not a thing*.  
2. Key-types of known count, and that count *might* be 0, is now a thing. (Previously this was impossible.)

While mathematically 2 makes a certain amount of sense (why would we fail to capture the case of the empty set?), I might favor 1 until the case for 2 becomes clear, especially since we know that we've gotten along for quite some time just fine with 1. And, it will mean less work for you.

src/Microsoft.ML.Core/Data/ColumnType.cs
TomFinley: One of the things we see in the original issue is, we want to get rid of the concept of the "unknown cardinality" altogether.

src/Microsoft.ML.Core/Data/ColumnType.cs
artidoro: Thanks for pointing out the difference between Assert and Check, I was ignoring it and makes much more sense like this. 

---
In reply to: [248116058](https://github.com/dotnet/machinelearning/pull/2146#discussion_r248116058) [](ancestors = 248116058,248102911)

src/Microsoft.ML.Core/Data/ColumnType.cs
artidoro: I will keep a public constructor for `int count` with a proper check since, as you mentioned, it is really the most common case. 

Regarding the other constructor which takes `DataKind`, when possible I switched to the public constructors, but in 7 occurrences it would require a lot more work. So I decided to keep it and wait for Eric's changes to remove `DataKind`. After that it will be trivial to remove this constructor. 

---
In reply to: [248116250](https://github.com/dotnet/machinelearning/pull/2146#discussion_r248116250) [](ancestors = 248116250,248115787)

docs/samples/Microsoft.ML.Samples/Dynamic/MatrixFactorization.cs
```diff
@@ -20,13 +20,11 @@ internal class MatrixElement
         {
             // Matrix column index starts from _synthesizedMatrixFirstColumnIndex and is at most
             // _synthesizedMatrixFirstColumnIndex + _synthesizedMatrixColumnCount - 1.
```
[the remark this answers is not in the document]
artidoro: Yes you are right, I am removing that.

---
In reply to: [248153257](https://github.com/dotnet/machinelearning/pull/2146#discussion_r248153257) [](ancestors = 248153257)

src/Microsoft.ML.Core/Data/ColumnType.cs
artidoro: Would you enforce this in the constructor?

---
In reply to: [248122337](https://github.com/dotnet/machinelearning/pull/2146#discussion_r248122337) [](ancestors = 248122337)

src/Microsoft.ML.Core/Data/ColumnType.cs
artidoro: As you and Tom pointed out, I am going to remove the option of an unknown cardinality for KeyType. I will replace that scenario with the maximum available value for count given the datatype used.

---
In reply to: [248106490](https://github.com/dotnet/machinelearning/pull/2146#discussion_r248106490) [](ancestors = 248106490)

artidoro: Ok that makes sense I will remove that option!

---
In reply to: [248367868](https://github.com/dotnet/machinelearning/pull/2146#discussion_r248367868) [](ancestors = 248367868)

src/Microsoft.ML.Core/Data/ColumnType.cs
TomFinley: Sure, where else could it possibly be enforced?

artidoro: Yes I figured :) I have a check in the first constructor.

---
In reply to: [248901904](https://github.com/dotnet/machinelearning/pull/2146#discussion_r248901904) [](ancestors = 248901904)

src/Microsoft.ML.Core/Data/ColumnType.cs
TomFinley: Sounds good thank you.

---
In reply to: [248483878](https://github.com/dotnet/machinelearning/pull/2146#discussion_r248483878) [](ancestors = 248483878,248116250,248115787)

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -478,42 +478,38 @@ public override bool Equals(ColumnType other)
...
-            : this(type, CheckRefRawType(type), min, count, contiguous)
+        public KeyType(Type type, ulong count)
+            : this(type, CheckRefRawType(type), count)
+        {
+        }
+
+        public KeyType(Type type, int count)
+            : this(type, CheckRefRawType(type), (ulong)count)
         {
+            Contracts.CheckParam(0 < count, nameof(count));
```
TomFinley: >Contracts.CheckParam(0 < count, nameof(count)); [](start = 12, length = 47)

Could we have a message on this, sort of like we used to have? Given that this is part of what we'd call "core IDataView" it is somewhat to our advantage to have it be correct.

src/Microsoft.ML.Data/DataLoadSave/Binary/Codecs.cs
```diff
@@ -1218,18 +1218,14 @@ public int WriteParameterization(Stream stream)
...
                     writer.Write(_type.Count);
-                    total += sizeof(int);
+                    total += sizeof(ulong);
                 }
                 return total;
             }
 
             // REVIEW: There is something a little bit troubling here. If someone, say,
-            // produces a column on KeyType(I4, 0, 4, true) and then returns 10 as a value in
+            // produces a column on KeyType(I4, 4, true) and then returns 10 as a value in
```
TomFinley: >true [](start = 51, length = 4)

Is this `true` relevant any longer?

src/Microsoft.ML.Data/DataLoadSave/Binary/CodecFactory.cs
```diff
@@ -68,7 +68,8 @@ public CodecFactory(IHostEnvironment env, MemoryStreamPool memPool = null)
             RegisterOtherCodec("DvTimeSpan", new UnsafeTypeCodec<TimeSpan>(this).GetCodec);
 
             RegisterOtherCodec("VBuffer", GetVBufferCodec);
-            RegisterOtherCodec("Key", GetKeyCodec);
+            RegisterOtherCodec("KeyNew", GetKeyCodec);
```
TomFinley: >KeyNew [](start = 32, length = 6)

"New" is relative. Everything here is old but once was new... Maybe Key2? Otherwise if a new version comes along, we'd have "Key", "KeyNew", and then, maybe, "KeySuperNew".

src/Microsoft.ML.Data/DataLoadSave/Binary/Codecs.cs
```diff
@@ -1276,7 +1272,112 @@ private bool GetKeyCodec(Stream definitionStream, out IValueCodec codec)
...
+            // Construct the key type.
+            var itemType = innerCodec.Type as PrimitiveType;
+            Contracts.CheckDecode(itemType != null);
+            Contracts.CheckDecode(KeyType.IsValidDataKind(itemType.RawKind));
+            KeyType type;
+            using (BinaryReader reader = OpenBinaryReader(definitionStream))
+            {
+                ulong count = reader.ReadUInt64();
+
+                Contracts.CheckDecode(0 <= count);
```
TomFinley: >CheckDecode(0 <= count); [](start = 26, length = 24)

This should now be 1, I believe.

src/Microsoft.ML.Data/DataLoadSave/Binary/Codecs.cs
```diff
@@ -1276,7 +1272,112 @@ private bool GetKeyCodec(Stream definitionStream, out IValueCodec codec)
                 Contracts.CheckDecode((ulong)count <= itemType.RawKind.ToMaxInt());
                 Contracts.CheckDecode(contiguous || count == 0);
 
-                type = new KeyType(itemType.RawKind, min, count, contiguous);
+                type = new KeyType(itemType.RawType, count);
```
TomFinley: >count [](start = 53, length = 5)

Let's imagine, hypothetically, that the count were 0, that is, it is of unknown cardinality. In such an instance as that, I would not like this to fail. It seems like the only reasonable course would be that we map it to the greatest possible value... so if U1, we put in `byte.MaxValue`, if U2, then `ushort.MaxValue`, etc.? This would be a suitable enough mapping. If we get rid of the concept of "unknown count," then I'd rather not fail, I'd rather just say, "you know what? I know I know nothing" and allow it to be everything. :) Which seems less harmful.

TomFinley: >KeyType [](start = 27, length = 7)

I'd feel better with some tests of this. Maybe, *in the old version*, prepare a completely empty binary IDV file (empty in the sense of no rows), but with a schema where you exercise all these strange conditions. You check in that IDV file. (It's a binary file, but a relatively small one.)

So, for instance, I might create something that looks like this:

    dotnet MML.dll savedata loader=text{col=A:U4[0-2]:0 col=B:U4[2-5]:0 col=C:U8[5-10]:0 col=D:U2[0-*]:0} dout=bubba.idv

If I were subsequently to do this:

    dotnet MML.dll showschema data=bubba.idv

Then what I see is this:

```
4 columns:
  A: Key<U4, 0-2>
  B: Key<U4, 2-5>
  C: Key<U8, 5-10>
  D: Key<U2, 0-*>
```

whereas what I might see *after* your changes is something more like this:

```
4 columns:
  A: Key<U4, 0-2>
  B: Key<U4, 0-3>
  C: Key<U8, 0-5>
  D: Key<U2, 0-65534>
```

But... like I say might be nice to test it. (I'm doing this through command line, but that's just for demonstration purposes. The test itself should probably be precise and programmatic.)

src/Microsoft.ML.Data/DataLoadSave/Binary/CodecFactory.cs
artidoro: Well I don't think this protocol was very well designed to begin with as it does not allow for a version number. I was trying to reflect the practice in the rest of the file where there seems to be a "Old" version of some codecs which says "old" in their name. 

I agree however that your suggestion of numbering would be the least harm here.

src/Microsoft.ML.Data/DataLoadSave/Binary/Codecs.cs
artidoro: Yes you are right!

src/Microsoft.ML.Data/DataLoadSave/Binary/Codecs.cs
TomFinley: BTW probably have to be more comprehensive than I was, that was just a small example.

---
In reply to: [249132474](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249132474) [](ancestors = 249132474)

src/Microsoft.ML.Core/Data/ColumnType.cs
artidoro: That is a really good point! Thank you for the suggestion.

---
In reply to: [249127901](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249127901) [](ancestors = 249127901)

src/Microsoft.ML.Data/DataLoadSave/Binary/Codecs.cs
artidoro: I agree with adding a test, and good catch I had forgotten about that!! 

---
In reply to: [249132474](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249132474) [](ancestors = 249132474)

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -478,42 +478,39 @@ public override bool Equals(ColumnType other)
...
-
-            Contracts.CheckParam(min >= 0, nameof(min));
-            Contracts.CheckParam(count >= 0, nameof(count), "Must be non-negative.");
-            Contracts.CheckParam((ulong)count <= ulong.MaxValue - min, nameof(count));
-            Contracts.CheckParam((ulong)count <= kind.ToMaxInt(), nameof(count));
-            Contracts.CheckParam(contiguous || count == 0, nameof(count), "Must be 0 for non-contiguous");
-
-            Contiguous = contiguous;
-            Min = min;
+            Contracts.CheckParam(0 < count && count <= kind.ToMaxInt(), nameof(count),
```
yaeldekel: ><= [](start = 52, length = 2)

Shouldn't this inequality be strict?

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
```diff
@@ -19,9 +19,19 @@ internal static class ColumnTypeExtensions
...
-        public static int GetKeyCount(this ColumnType columnType) => (columnType as KeyType)?.Count ?? 0;
+        public static ulong GetKeyCount(this ColumnType columnType) => (columnType as KeyType)?.Count ?? 0;
+
+        /// <summary>
+        /// Sometimes it is necessary to cast the Count to an int. This performs overflow check.
+        /// </summary>
+        public static int AssertRangeReturnCount(this ColumnType columnType, IExceptionContext ectx = null)
+        {
+            ulong count = columnType.GetKeyCount();
+            ectx.Assert(count <= int.MaxValue, nameof(KeyType) + "." + nameof(KeyType.Count) + " exceeds int.MaxValue.");
```
yaeldekel: >Assert [](start = 17, length = 6)

Why is this an assert? Isn't it intended so that callers of this method don't need to do a Check?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -776,19 +774,25 @@ public Bindings(ModelLoadContext ctx, TextLoader parent)
...
+                        ulong count;
                         Contracts.CheckDecode(KeyType.IsValidDataKind(kind));
-
-                        bool isContig = ctx.Reader.ReadBoolByte();
-                        ulong min = ctx.Reader.ReadUInt64();
-                        Contracts.CheckDecode(min >= 0);
-                        int count = ctx.Reader.ReadInt32();
+                        if (ctx.Header.ModelVerWritten < 0x0001000C)
+                        {
+                            bool isContig = ctx.Reader.ReadBoolByte();
```
yaeldekel: >isContig [](start = 33, length = 8)

You can check that this is true and throw if it isn't.

src/Microsoft.ML.Data/DataLoadSave/Text/TextSaver.cs
```diff
@@ -491,14 +491,12 @@ private TextLoader.Column GetColumn(string name, ColumnType type, int? start)
             ColumnType itemType = vectorType?.ItemType ?? type;
             if (itemType is KeyType key)
             {
-                if (!key.Contiguous)
-                    keyRange = new KeyRange(key.Min, contiguous: false);
-                else if (key.Count == 0)
-                    keyRange = new KeyRange(key.Min);
+                if (key.Count == 0)
```
yaeldekel: >key.Count == 0 [](start = 20, length = 14)

Is this allowed?

src/Microsoft.ML.Data/Scorers/ClusteringScorer.cs
```diff
@@ -101,7 +101,9 @@ protected override Delegate GetPredictedLabelGetter(Row output, out Delegate sco
 
             long cachedPosition = -1;
             VBuffer<Float> score = default(VBuffer<Float>);
-            int scoreLength = Bindings.PredColType.GetKeyCount();
+            ulong keyCount = Bindings.PredColType is KeyType key ? key.Count : 0;
+            Contracts.Assert(keyCount <= int.MaxValue);
```
yaeldekel: Can this be replaced with AssertRangeReturnKeyCount here as well?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
artidoro: I this case I translate the old KeyType into a new one that is a superset of the old one. 

---
In reply to: [249206893](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249206893) [](ancestors = 249206893)

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -478,42 +478,39 @@ public override bool Equals(ColumnType other)
...
+                "The cardinality of a " + nameof(KeyType) + " must not exceed " + nameof(type) + ".MaxValue and must be strictly positive.");
```
TomFinley: >nameof(type) [](start = 82, length = 12)

This will be `type.MaxValue`, but I think you really meant `type.Name`. Note that `nameof(type)` is literally going to just be the string `"type"`.

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
TomFinley: Yes, this *looks* like a check to me, based on how it is used. FYI if we meant an assert then there is no need for an error message, since asserts are just for code documentation.

---
In reply to: [249203459](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249203459) [](ancestors = 249203459)

src/Microsoft.ML.Data/EntryPoints/PredictorModelImpl.cs
```diff
@@ -125,7 +125,7 @@ internal override string[] GetLabelInfo(IHostEnvironment env, out ColumnType lab
             if (trainRms.Label != null)
             {
                 labelType = trainRms.Label.Value.Type;
-                if (labelType is KeyType keyType && trainRms.Label.Value.HasKeyValues(keyType.Count))
```
TomFinley: >labelType is KeyType keyType [](start = 20, length = 28)

Simplifying this may be acceptable if we also simultaneously fix the bug in `HasKeyValues` that I see we have (at least in `master` right now) that `HasKeyValues` will succeed even if the input column is not a key type. :)

src/Microsoft.ML.Data/Scorers/ClusteringScorer.cs
artidoro: Yes! Fixed.

---
In reply to: [249208483](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249208483) [](ancestors = 249208483)

src/Microsoft.ML.Data/EntryPoints/PredictorModelImpl.cs
artidoro: It will return false in current version because GetKeyCount() returns 0 when type is not KeyType.

---
In reply to: [249213625](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249213625) [](ancestors = 249213625)

src/Microsoft.ML.Core/Data/ColumnType.cs
artidoro: I think that it should not be strict. What `Count` expresses is the **largest** possible value for a key. The bottleneck is the key underlying `type`. At most a key can take the largest representable value in that type, which is kind.ToMaxInt().  

---
In reply to: [249202365](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249202365) [](ancestors = 249202365)

src/Microsoft.ML.StaticPipe/StaticSchemaShape.cs
```diff
@@ -280,7 +280,8 @@ private static Type GetTypeOrNull(Schema.Column col)
                     {
                         var kvcol = meta.Schema[kvcolIndex];
                         var kvType = kvcol.Type;
-                        if (kvType is VectorType kvVecType && kvVecType.Size == kt.Count)
+                        Contracts.Assert(kt.Count <= int.MaxValue);
```
artidoro: >Assert [](start = 34, length = 6)

Is it fine to have an Assert here?

src/Microsoft.ML.Data/EntryPoints/PredictorModelImpl.cs
TomFinley: Ah that makes sense. Thanks @artidoro .

src/Microsoft.ML.StaticPipe/StaticSchemaShape.cs
TomFinley: This assert is fine. `kvVecType.Size` is integer, and above you check whether `kt.Count` equals it. So, it must be in the range of an `int`. So, what you are doing here is you're asserting to make clear to the reader that your `(int)` cast below is fine, which seems like a perfectly fine use of an assert for code documentation @artidoro .

src/Microsoft.ML.Data/Utilities/TypeParsingUtils.cs
```diff
@@ -57,23 +57,21 @@ public static KeyType ConstructKeyType(DataKind? type, KeyRange range)
 
             DataKind kind;
             KeyType keyType;
-            kind = type ?? DataKind.U4;
+            kind = type ?? DataKind.U8;
             Contracts.CheckUserArg(KeyType.IsValidDataKind(kind), nameof(TextLoader.Column.Type), "Bad item type for Key");
-            Contracts.CheckUserArg(range.Min >= 0, nameof(range.Min), "min must be non-negative");
```
TomFinley: >Min [](start = 41, length = 3)

Just thought of this... what is happening to `KeyRange.Min`? We should probably be asserting that `Min` is 0, if we are to keep it at all. (Of course, in the absence of support for min values we could just change the format rather entirely so that what is currently `U4[0-100]` could be parsed as `U4[100]`, but maybe we want to hold off on that for now.)

TomFinley:
> 

Since unknown length keys are no longer supported, it seems like this condition should return *false*, should it not?

---
Refers to: src/Microsoft.ML.Data/Utilities/TypeParsingUtils.cs:124 in 1b220f6. [](commit_id = 1b220f6deee9a56aeefadec2ce0ae87cd19653dd, deletion_comment = False)

src/Microsoft.ML.Data/Utilities/TypeParsingUtils.cs
```diff
@@ -116,19 +106,18 @@ private bool TryParse(string str)
...
             {
-                if (!ulong.TryParse(str, out Min))
+                if (!ulong.TryParse(str, out min))
                     return false;
-                Contiguous = false;
                 return true;
             }
 
-            if (!ulong.TryParse(str.Substring(0, ich), out Min))
+            if (!ulong.TryParse(str.Substring(0, ich), out min))
```
TomFinley: >min [](start = 59, length = 3)

What if `min` is non-zero, since we no longer support that? What then?

TomFinley: We need to think more carefully about what the format here should be. Just leaving it unchanged is clearly wrong (since it is meant to support some concepts that no longer exist), but what woudl be right, I am less certain about. Maybe @yaeldekel has some thoughts here.

---
In reply to: [249262458](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249262458) [](ancestors = 249262458)

artidoro: I initially thought that this would come with a subsequent PR. But I think it can be easily done as part of this PR. 

Currently the format looks like: "TYPE[MIN-COUNT]". I think the format should change to something that looks like "TYPE[COUNT]", so something like this: U4[100].

This removes the Min, and we can also enforce the Count to be required (so that "*" is not acceptable for Count).
It should be pretty easy to do this change. 

---
In reply to: [249262473](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249262473) [](ancestors = 249262473,249262458)

src/Microsoft.ML.Data/Utilities/TypeParsingUtils.cs
artidoro: I decided to remove the Min from KeyRange as KeyType and KeyRange should map to each other. I answered in another of your comments that I agree we should match the string format of KeyRange not to have the Min and to have a determined Count.

---
In reply to: [249262433](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249262433) [](ancestors = 249262433)

src/Microsoft.ML.Data/Utilities/TypeParsingUtils.cs
TomFinley: Maybe parts of it can be done in a separate PR, but I think those parts of the syntax that are meant to support features that no longer exist should be made to throw now, at least. So, `0-5` is fine, but `2-5` or `0-*` or `0-` is not for `KeyRange`, would be the idea, since where used, those things, even though we never used them, had a definite meaning and we need to be clear that's no longer supported.

...Not that anything other than our tests really used it, but, you know, still. 😄

TomFinley: Also when you say "removes the Min", I think you also mean changing the format so that instead of `U4[0-10]` or something it would be just `U4[10]`. While this is fine, I think it is important that at least for the short term we continue to support `U4[0-10]` as an *input* format at least, so that older text files from prior versions should be readable, for at least some versions, until we determine that support is no longer necessary.

---
In reply to: [249625047](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249625047) [](ancestors = 249625047)

docs/code/IDataViewTypeSystem.md
```diff
@@ -383,82 +383,53 @@ Key types are instances of the sealed class `KeyType`, which derives from
...
 
-* A count value, between `0` and `int.MaxValue`, inclusive
+* A count value, between `1` and `ulong.MaxValue`, inclusive
 
-* A "minimum" value, between `0` and `ulong.MaxValue`, inclusive
-
-* A Boolean value indicating whether the values of the key type are contiguous
-
-Regardless of the minimum and count values, the representational value zero
+Regardless of the count value, the representational value zero
```
wschin: Can we use `ulong.MaxValue` to denote NaN? It can further simplify the conversion between integers and keys. Note that conversion happens in type casting and reading C# native class as `IDataView`.
```suggestion
Regardless of the count value, the representational value zero
```

docs/code/IDataViewTypeSystem.md
```diff
@@ -383,82 +383,53 @@ Key types are instances of the sealed class `KeyType`, which derives from
...
+* The `Count` property returns the count, or cardinality of the key type. This 
+  is of type `ulong`. The legal representation values are from one up to 
+  and including `Count`. The `Count` is required to be representable in the 
+  underlying type, so, for example, the `Count` value of a key type based
+  on `System.Byte` must not exceed `255`. As an example of the usefulness 
+  of the `Count` property, consider the `KeyToVector` transform implemented
+  as part of ML.NET. It maps from a key type value to an indicator vector.
+  The length of the vector is the `Count` of the key type, which is required
+  to be positive. For a key value of `k`, with `1 ≤ k ≤ Count`, the resulting
+  vector has a value of one in the (`k-1`)th slot, and zero in all other slots.
```
wschin: Maybe
```suggestion
  vector has a value of one in the (`k`)th slot, and zero in all other slots.
```
? The 1st slot is indexed by key value 1, isn't it?

docs/code/IDataViewTypeSystem.md
```diff
@@ -383,82 +383,53 @@ Key types are instances of the sealed class `KeyType`, which derives from
 
 In addition to its underlying type, a key type specifies:
 
-* A count value, between `0` and `int.MaxValue`, inclusive
+* A count value, between `1` and `ulong.MaxValue`, inclusive
```
wschin: At my first glance, I didn't know what does `inclusive` mean but then I realize how stupid I was after reading through the following paragraph. Maybe
```suggestion
* A count value, from `1` to `ulong.MaxValue`.
```
can help other naive users?

docs/samples/Microsoft.ML.Samples/Dynamic/MatrixFactorization.cs
```diff
@@ -48,8 +39,8 @@ public static void MatrixFactorizationInMemoryData()
         {
             // Create an in-memory matrix as a list of tuples (column index, row index, value).
             var dataMatrix = new List<MatrixElement>();
-            for (uint i = _synthesizedMatrixFirstColumnIndex; i < _synthesizedMatrixFirstColumnIndex + _synthesizedMatrixColumnCount; ++i)
-                for (uint j = _synthesizedMatrixFirstRowIndex; j < _synthesizedMatrixFirstRowIndex + _synthesizedMatrixRowCount; ++j)
+            for (uint i = 0; i < _synthesizedMatrixColumnCount; ++i)
```
wschin: This change doesn't look correct. This example aims at testing user data set where indexes starting from `_synthesizedMatrixFirstColumnIndex=1`.
```suggestion
            for (uint i = 0; i < _synthesizedMatrixColumnCount; ++i)
```

docs/code/IDataViewTypeSystem.md
TomFinley: I do not like the suggested change as much as I might like. Chaning "between" to "from" is good. (Between suggests exclusion.) Removing "inclusive" is bad. What you think?

docs/code/IDataViewTypeSystem.md
TomFinley: Depends on whether for teh sake of this discussion we're considering the "first" slot to be literally 1st or 0th.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
TomFinley: I think we should throw as @yaeldekel suggests. The reason is that we are now suddenly loading things as a very, very different type now. If anyone *happened* to be making use of this functionality (which, frankly, I doubt, but let's just suppose they were) your suggested change does not strike me as necessarily one that they would welcome. Indeed in some cases it would not even be possible... `U1[5000-5002]` used to be legal and sensible, now if it maps to `U1[0-50002]`, what then? So I agree with @yaeldekel that throwing here is best policy.

src/Microsoft.ML.Data/DataLoadSave/Binary/Codecs.cs
```diff
@@ -1276,7 +1272,116 @@ private bool GetKeyCodec(Stream definitionStream, out IValueCodec codec)
...
+                else
+                    type = new KeyType(itemType.RawType, count);
+            }
+            // Next create the key codec.
+            Type codecType = typeof(KeyCodecOld<>).MakeGenericType(itemType.RawType);
+            codec = (IValueCodec)Activator.CreateInstance(codecType, this, type, innerCodec);
+            return true;
+        }
+
+        private bool GetKeyCodecOld(ColumnType type, out IValueCodec codec)
```
wschin: Does having `GetKeyCodecOld` mean we still use old key type attribute somewhere?

src/Microsoft.ML.Data/DataLoadSave/Binary/Codecs.cs
```diff
@@ -1276,7 +1272,116 @@ private bool GetKeyCodec(Stream definitionStream, out IValueCodec codec)
...
+            }
+        }
+
+        private bool GetKeyCodec(Stream definitionStream, out IValueCodec codec)
+        {
+            // The first value in the definition stream will be the internal codec.
+            IValueCodec innerCodec;
+            if (!TryReadCodec(definitionStream, out innerCodec))
+            {
+                codec = default(IValueCodec);
```
wschin: Maybe this can work.
```suggestion
                codec = default;
```

src/Microsoft.ML.Data/Utilities/TypeParsingUtils.cs
artidoro: From discussing with Tom and Yael, we decided to do the following:

Change the format to `U4[100]`, or `U4[]` and maintain backwards compatibility with `U4[0-100]` and `U4[0-*]` throwing if `Min`  is not 0. 

---
In reply to: [249626204](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249626204) [](ancestors = 249626204,249625047)

src/Microsoft.ML.Data/DataLoadSave/Binary/Codecs.cs
artidoro: This is necessary to allow backwards compatibility when reading files that were generated before this change. Take a look at the test that I added in BinaryLoaderSaverTests.cs in ML.Tests. Without maintaining the old codec, it would be impossible to read that file.

---
In reply to: [249909180](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249909180) [](ancestors = 249909180)

docs/code/IDataViewTypeSystem.md
wschin: I see. `k-1` is fine, so let me close this comment.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
artidoro: Ok sounds good! I will update this then.

---
In reply to: [249908025](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249908025) [](ancestors = 249908025)

docs/code/IDataViewTypeSystem.md
wschin: My original idea comes from having same value for key value and its underlying representation. This way we can simplify the loading/writing value from/to C# data structure. But I got your point, NA'd better to be unique.

src/Microsoft.ML.Data/DataLoadSave/Binary/Codecs.cs
wschin: Why do we need BC? Key type is changed to a new form.

---
In reply to: [249912390](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249912390) [](ancestors = 249912390,249909180)

src/Microsoft.ML.Data/DataLoadSave/Text/TextSaver.cs
```diff
@@ -491,15 +491,7 @@ private TextLoader.Column GetColumn(string name, ColumnType type, int? start)
...
-                if (!key.Contiguous)
-                    keyRange = new KeyRange(key.Min, contiguous: false);
-                else if (key.Count == 0)
-                    keyRange = new KeyRange(key.Min);
-                else
-                {
-                    Contracts.Assert(key.Count >= 1);
-                    keyRange = new KeyRange(key.Min, key.Min + (ulong)(key.Count - 1));
-                }
+                keyRange = new KeyRange(key.Count - 1);
```
wschin: >KeyRange [](start = 31, length = 8)

Why do we still need `KeyRange`?

test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
```diff
@@ -146,21 +146,21 @@ internal class MatrixElement
...
+            [KeyType(Count = _synthesizedMatrixRowCount)]
             public uint MatrixRowIndex;
             // The value at the MatrixColumnIndex-th column and the MatrixRowIndex-th row in the considered matrix.
             public float Value;
         }
 
         internal class MatrixElementForScore
         {
-            [KeyType(Contiguous = true, Count = _synthesizedMatrixColumnCount, Min = _synthesizedMatrixFirstColumnIndex)]
+            [KeyType(Count = _synthesizedMatrixColumnCount)]
```
wschin: >_synthesizedMatrixColumnCount [](start = 29, length = 29)

This should be `_synthesizedMatrixColumnCount +_synthesizedMatrixFirstColumnIndex` because `Count` stands for maximum key index allowed now. Do you think we could rename `Count` to `Max`?

src/Microsoft.ML.Data/DataLoadSave/Binary/Codecs.cs
TomFinley: There is no reason to break backwards compatibility. We have I hope succeeded in never doing so. The entire reason for this change is that we never *used* these old capabilities. So if we have never used them, and we save data with this format (which is entirely possible), why would we break them? It is correct to maintain BC.

I do not view this as KeyType being changed to a new form at all. Rather, we are keeping only those parts of the KeyType have actually proved useful, and getting rid of the rest that just led to codebloat. I would not welcome any change that would break backcompat.

test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
```diff
@@ -233,11 +231,11 @@ internal class MatrixElementZeroBased
         {
             // Matrix column index starts from 0 and is at most _synthesizedMatrixColumnCount-1.
             // Contieuous=true means that all values from 0 to _synthesizedMatrixColumnCount-1 are allowed keys.
```
wschin: >Contieuous [](start = 15, length = 10)

No `Contiguous` anymore.

test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
```diff
@@ -233,11 +231,11 @@ internal class MatrixElementZeroBased
...
-            [KeyType(Contiguous = true, Count = _synthesizedMatrixColumnCount, Min = 0)]
+            [KeyType(Count = _synthesizedMatrixColumnCount)]
             public uint MatrixColumnIndex;
             // Matrix row index starts from 0 and is at most _synthesizedMatrixRowCount-1.
             // Contieuous=true means that all values from 0 to _synthesizedMatrixRowCount-1 are allowed keys.
```
wschin: >Contieuous [](start = 15, length = 10)

No `Contiguous` too.

wschin:
>         // The variable _synthesizedMatrixFirstRowIndex indicates the integer that would be mapped to the first row index. If user data uses

Because in ML.NET key type's minimal value is zero, the first row index is always zero in C# data structure (e.g., MatrixColumnIndex=0 and MatrixRowIndex=0 in MatrixElement below specifies the value at the upper-left corner in the training matrix). If user's row index starts with 1, their row index 1 would be mapped to the 2nd row in matrix factorization module and their first row may contain no values. This behavior is also true to column index.

---
Refers to: test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs:137 in 1b220f6. [](commit_id = 1b220f6deee9a56aeefadec2ce0ae87cd19653dd, deletion_comment = False)

wschin:
>         // 0-based indices for rows, _synthesizedMatrixFirstRowIndex can be set to 0. Similarly, for 1-based indices, _synthesizedMatrixFirstRowIndex

Removed.

---
Refers to: test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs:138 in 1b220f6. [](commit_id = 1b220f6deee9a56aeefadec2ce0ae87cd19653dd, deletion_comment = False)

test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
```diff
@@ -146,21 +146,21 @@ internal class MatrixElement
         {
             // Matrix column index starts from 1 and is at most _synthesizedMatrixColumnCount.
             // Contieuous=true means that all values from 1 to _synthesizedMatrixColumnCount are allowed keys.
```
wschin: >Contieuous [](start = 15, length = 10)

No `Contiguous`.

test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
```diff
@@ -146,21 +146,21 @@ internal class MatrixElement
...
-            [KeyType(Contiguous = true, Count = _synthesizedMatrixColumnCount, Min = _synthesizedMatrixFirstColumnIndex)]
+            [KeyType(Count = _synthesizedMatrixColumnCount)]
             public uint MatrixColumnIndex;
             // Matrix row index starts from 1 and is at most _synthesizedMatrixRowCount.
             // Contieuous=true means that all values from 1 to _synthesizedMatrixRowCount are allowed keys.
```
wschin: >Contieuous [](start = 15, length = 10)

No `Contiguous`.

test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
```diff
@@ -146,21 +146,21 @@ internal class MatrixElement
         {
             // Matrix column index starts from 1 and is at most _synthesizedMatrixColumnCount.
```
wschin: >1 [](start = 47, length = 1)

0

test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
```diff
@@ -146,21 +146,21 @@ internal class MatrixElement
...
             // Contieuous=true means that all values from 1 to _synthesizedMatrixColumnCount are allowed keys.
-            [KeyType(Contiguous = true, Count = _synthesizedMatrixColumnCount, Min = _synthesizedMatrixFirstColumnIndex)]
+            [KeyType(Count = _synthesizedMatrixColumnCount)]
             public uint MatrixColumnIndex;
             // Matrix row index starts from 1 and is at most _synthesizedMatrixRowCount.
```
wschin: >1 [](start = 44, length = 1)

0

src/Microsoft.ML.Data/DataLoadSave/Text/TextSaver.cs
artidoro: `KeyRange` is used mainly in the arguments, for example in the TextLoader arguments. It takes care of parsing, and unparsing into the form U4[100] (and the other format that we discussed in the PR). It's therefore mainly a utility class for parsing and for now I think it makes sense to keep it.

---
In reply to: [249920350](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249920350) [](ancestors = 249920350)

test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
TomFinley: I'd rather we talk about the renaming as a separate issue, since that would be a very disruptive change. That is, we'd have this one (comparatively) simple change to remove the old capability, then the renaming as a separate PR. Even if we thought this was a good idea (and we very well may) I would still prefer to have that in two commits against master.

src/Microsoft.ML.Core/Data/ColumnType.cs
TomFinley: Right, so, if I have a `U1`, values 1 through 255 would represent the range of possible keys, and correspond to a count of 255, which equals I think `kind.ToMaxInt`. So I think as written it is correct. Though I agree it is slightly confusing and you have to think about it a while before realizing it is correct -- something about it somehow just tickles my brain every time I see it, not quite sure what.

---
In reply to: [249229341](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249229341) [](ancestors = 249229341,249202365)

test/Microsoft.ML.Tests/BinaryLoaderSaverTests.cs
```diff
@@ -0,0 +1,39 @@
...
+        public BinaryLoaderSaverTests(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        [Fact]
+        public void OldKeyTypeCodecTest()
+        {
+            // Checks that we can load IDataViews defined with unknown cardinality KeyType.
+            // schema-codec-test.idv was generated with the following command before simplifying the KeyType:
+            // dotnet MML.dll savedata loader=text{col=A:U4[0-2]:0 col=B:U4[2-5]:0 col=C:U1[5-10]:0 col=D:U2[0-*]:0 col=E:U4[4-*]:0 col=F:U8[5-*]:0} dout=codectest.idv
```
TomFinley: >col=A:U4[0-2]:0 col=B:U4[2-5]:0 col=C:U1[5-10]:0 col=D:U2[0-*]:0 col=E:U4[4-*]:0 col=F:U8[5-* [](start = 51, length = 93)

So right now this correspond to this.

```
col=A:U4[2]:0
col=B:U4[3]:1
col=C:U1[5]:2
col=D:U2[65534]:3
col=E:U4[4294967294]:4
col=F:U8[18446744073709551614]:5
```

Now here I am wondering something. We had previously that if a specified a range, the range was inclusive -- that is, `0-2` being 0,1,2, has a count of 3. I wonder if it would be more clear if in the "single number" case, whether this indicates a count or a range. Because if I see `U4[2]` I might think this is a key value with 2 values, but in fact, there are 3, because that 2 is merely an upper bound. This at least strikes me as confusing. So we might read `U4[0-2]` but it is equivalent to `U4[3]`. Not sure. Maybe ask opinion of @yaeldekel and @wschin.

TomFinley: Actually yeah the more I think about it the more confused I get... so in the new scheme, `U4[2]` would correspond to a count of *three* where the actual physical values corresponding to valid values are `1`, `2`, and `3`. So I think in this "single value" case it should just be count.

---
In reply to: [249994613](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249994613) [](ancestors = 249994613)

artidoro: As you wrote in your message, there is a fundamental difference: are we using the `Count` as a max (for a range) or as a count of values. In this change, I use Count as a the upper bound of the range of valid values, and the upper bound is inclusive.

In fact, Wei-Sheng suggested to rename Count to Max.

---
In reply to: [249996684](https://github.com/dotnet/machinelearning/pull/2146#discussion_r249996684) [](ancestors = 249996684,249994613)

artidoro: I have initially worked on removing `KeyRange`. I succeeded in removing it completely, but I found one scenario that is not allowed without `KeyRange`. 

Both for TextLoader.Column and for TypeConvertingTransformer.ColumnInfo we used to define a `KeyRange`. We used to check if the Column defines a key by checking whether the `KeyRange `is null. If it is defined but no `Max` is provided we would map to `kind.ToMaxInt()` when constructing a `KeyType `from the `KeyRange`.

In my change I had replaced `KeyRange` with `ulong? KeyCount`. This simplified nicely some of the code since KeyRange used to define a `Max` which was essentially acting as `KeyCount - 1`. However, we do not distinguish the case `KeyRange == null` and `KeyRange != null && KeyRange.Max == null`. In other words `ulong? KeyCount == null` could mean that the current column does not define a key, or that the current column defines a key, but that its `keyCount `is not defined and should be mapped to `kind.ToMaxInt()`.

I therefore decided for a compromise. I transformed `KeyRange` into `KeyCount`, and implemented the logic that makes the `Count` of `KeyCount` match `KeyType.Count`.

wschin: Could we use `Count=0` to denote `kind.ToMaxInt()`?

---
In reply to: [250095679](https://github.com/dotnet/machinelearning/pull/2146#discussion_r250095679) [](ancestors = 250095679)

test/BaselineOutput/Common/BinaryLoaderSaver/OldKeyTypeCodecTest.txt
```diff
@@ -0,0 +1,9 @@
+#@ TextLoader{
+#@   sep=tab
+#@   col=A:U4[3]:0
+#@   col=B:U4[6]:1
+#@   col=C:U1[11]:2
+#@   col=D:U2[65535]:3
+#@   col=E:U4[4294967295]:4
+#@   col=F:U8[18446744073709551615]:5
```
TomFinley: >18446744073709551615 [](start = 14, length = 20)

Not a big deal at all, and probably this is the right thing, but when I read this I kind of wonder if rendering this as `U8[]` would be a better choice. :) Though it's probably best to be explicit, even if it is a bit verbose.

artidoro: I agree that it would look better. However, if conceptually we decide to require a defined Count for KeyType, for consistency I would argue in favor of keeping it in the string representation of the KeyType.

---
In reply to: [250352178](https://github.com/dotnet/machinelearning/pull/2146#discussion_r250352178) [](ancestors = 250352178)

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -456,79 +456,29 @@ public override bool Equals(ColumnType other)
     /// class numbers, etc. For example, in multi-class classification, the label is typically
     /// a class number which is naturally a KeyType.
     ///
-    /// KeyTypes can be contiguous (the class number example), in which case they can have
-    /// a cardinality/Count. For non-contiguous KeyTypes the Count property returns zero.
-    /// Any KeyType (contiguous or not) can have a Min value. The Min value is always >= 0.
+    /// KeyTypes have a cardinality/Count that is strictly positive.
```
wschin:
```suggestion
    /// KeyTypes have a cardinality (i.e., Count) that is strictly positive.
```

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -456,79 +456,29 @@ public override bool Equals(ColumnType other)
...
     ///
     /// Note that the representation value does not necessarily match the logical value.
```
wschin:
```suggestion
    /// Note that the underlying representation value does not necessarily match the logical value.
```

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -456,79 +456,29 @@ public override bool Equals(ColumnType other)
...
-    /// For example, if a KeyType has range 1000-5000, then it has a Min of 1000, Count
-    /// of 4001, but the representational values are 1-4001. The representation value zero
-    /// is reserved to mean none/invalid.
+    /// For example, if a KeyType has range 0-5000, then it has a Count of 5001, but
+    /// the representational values are 1-5001. The representation value zero is reserved
+    /// to mean none/invalid.
```
wschin: A missing is a valid value, I guess.
```suggestion
    /// to mean a missing value (similar to NaN).
```

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -456,79 +456,29 @@ public override bool Equals(ColumnType other)
...
-            Contracts.CheckParam(count >= 0, nameof(count), "Must be non-negative.");
-            Contracts.CheckParam((ulong)count <= ulong.MaxValue - min, nameof(count));
-            Contracts.CheckParam((ulong)count <= kind.ToMaxInt(), nameof(count));
-            Contracts.CheckParam(contiguous || count == 0, nameof(count), "Must be 0 for non-contiguous");
-
-            Contiguous = contiguous;
-            Min = min;
+            if (count == 0 || type.ToMaxInt() < count)
+                throw Contracts.ExceptParam(nameof(count), "The cardinality of a {0} must not exceed {1}.MaxValue" +
+                    " and must be strictly positive.", nameof(KeyType), type.Name);
```
wschin:
```suggestion
                    " and must be strictly positive but got {2}.", nameof(KeyType), type.Name, count);
```

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -456,79 +456,29 @@ public override bool Equals(ColumnType other)
...
-            {
-                case DataKind.U1:
-                case DataKind.U2:
-                case DataKind.U4:
-                case DataKind.U8:
-                    return true;
-                default:
-                    return false;
-            }
+            Contracts.CheckParam(0 < count, nameof(count), "The cardinality of a " + nameof(KeyType) + " must be strictly positive.");
```
wschin: Do you need to check the two conditions in Line 472?

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -542,23 +492,13 @@ public static bool IsValidDataType(Type type)
...
-        /// for "not there"/"none". Typical Min values are 0 or 1, but can be any value >= 0.
-        /// </summary>
-        public ulong Min { get; }
-
-        /// <summary>
-        /// If this key type has contiguous values and a known cardinality, Count is that cardinality.
-        /// Otherwise, this returns zero. Note that such a key type can be converted to a bit vector
-        /// representation by mapping to a vector of length Count, with "id" mapped to a vector with
-        /// 1 in slot (id - 1) and 0 in all other slots. This is the standard "indicator"
+        /// Count is the cardinality of the KeyType. Note that such a key type can be converted to a
```
wschin: nit: Count and KeyType can be replaced with references to the actual objects.

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -542,23 +492,13 @@ public static bool IsValidDataType(Type type)
...
+        /// bit vector representation by mapping to a vector of length Count, with "id" mapped to a
+        /// vector with 1 in slot (id - 1) and 0 in all other slots. This is the standard "indicator"
         /// representation. Note that an id of 0 is used to represent the notion "none", which is
         /// typically mapped to a vector of all zeros (of length Count).
```
wschin:
```suggestion
        /// typically mapped, by for example, one-hot encoding, to a vector of all zeros (of length Count).
```

src/Microsoft.ML.Core/Data/ColumnType.cs
artidoro: This is checking that the input count is not negative (which would be nonsense when casting). The other two checks are run anyways as this constructor will call the first.

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
```diff
@@ -77,4 +87,20 @@ public static bool SameSizeAndItemType(this ColumnType columnType, ColumnType ot
             return otherVectorType.Size == otherVectorType.Size;
         }
     }
+
+    /// <summary>
+    /// Extension methods related to the KeyType class.
+    /// </summary>
+    [BestFriend]
+    internal static class KeyTypeExtensions
```
eerhardt: We should be putting new classes in their own files, with the file name matching the class name. It gets too hard to navigate when files contain random classes.

src/Microsoft.ML.Transforms/Text/NgramTransform.cs
```diff
@@ -597,12 +597,12 @@ private void GetSlotNames(int iinfo, int size, ref VBuffer<ReadOnlyMemory<char>>
...
                 Host.Assert(0 <= iinfo && iinfo < _parent.ColumnPairs.Length);
 
-                var keyCount = _srcTypes[iinfo].GetItemType().GetKeyCount();
-                Host.Assert(InputSchema[_srcCols[iinfo]].HasKeyValues(keyCount));
+                Host.Assert(InputSchema[_srcCols[iinfo]].HasKeyValues(_srcTypes[iinfo].GetItemType()));
 
                 var unigramNames = new VBuffer<ReadOnlyMemory<char>>();
 
                 // Get the key values of the unigrams.
+                var keyCount = _srcTypes[iinfo].GetItemType().ChecktRangeReturnCount(Host);
```
eerhardt: (nit) why not capture `_srcTypes[iinfo].GetItemType()` in a variable so it doesn't need to be called twice?

src/Microsoft.ML.Data/Transforms/TypeConverting.cs
```diff
@@ -156,7 +155,8 @@ private static VersionInfo GetVersionInfo()
                 modelSignature: "CONVERTF",
                 // verWrittenCur: 0x00010001, // Initial
                 // verWrittenCur: 0x00010002, // Added support for keyRange
-                verWrittenCur: 0x00010003, // Change to transformer leads to change of saving objects.
+                //verWrittenCur: 0x00010003, // Change to transformer leads to change of saving objects.
+                verWrittenCur: 0x00010004, // Removed Min and Contiguous from KeyCount.
```
eerhardt: Do you have a test that loads a model with the old version to ensure backwards compat?

src/Microsoft.ML.Data/Transforms/TypeConverting.cs
```diff
@@ -274,29 +270,36 @@ private TypeConvertingTransformer(IHost host, ModelLoadContext ctx)
...
                 byte b = ctx.Reader.ReadByte();
                 var kind = (DataKind)(b & 0x7F);
                 Host.CheckDecode(Enum.IsDefined(typeof(DataKind), kind));
-                KeyRange range = null;
+                KeyCount keyCount = null;
                 if ((b & 0x80) != 0)
                 {
-                    range = new KeyRange();
-                    range.Min = ctx.Reader.ReadUInt64();
+                    if (ctx.Header.ModelVerWritten < 0x00010004)
```
eerhardt: It would be better to declare a constant for `0x00010004` rather than hard-coding that number everywhere.

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
```diff
@@ -19,9 +19,19 @@ internal static class ColumnTypeExtensions
...
-        /// Zero return means either it's not a key type or the cardinality is unknown.
+        /// Zero return means it's not a key type.
         /// </summary>
-        public static int GetKeyCount(this ColumnType columnType) => (columnType as KeyType)?.Count ?? 0;
+        public static ulong GetKeyCount(this ColumnType columnType) => (columnType as KeyType)?.Count ?? 0;
+
+        /// <summary>
+        /// Sometimes it is necessary to cast the Count to an int. This performs overflow check.
+        /// </summary>
+        public static int ChecktRangeReturnCount(this ColumnType columnType, IExceptionContext ectx = null)
```
eerhardt: Wait - why do we have this here, and one for `KeyType`?

Also, type-o `ChecktRangeReturnCount` there seems to be an errant `t` after `Check`.

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
```diff
@@ -77,4 +87,20 @@ public static bool SameSizeAndItemType(this ColumnType columnType, ColumnType ot
...
+    /// <summary>
+    /// Extension methods related to the KeyType class.
+    /// </summary>
+    [BestFriend]
+    internal static class KeyTypeExtensions
+    {
+        /// <summary>
+        /// Sometimes it is necessary to cast the Count to an int. This performs overflow check.
+        /// </summary>
+        public static int CheckRangeReturnCount(this KeyType key, IExceptionContext ectx = null)
```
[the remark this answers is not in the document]
artidoro: I have renamed the method to `GetKeyCountAsInt32`. Thank you for the suggestion!

---
In reply to: [250651853](https://github.com/dotnet/machinelearning/pull/2146#discussion_r250651853) [](ancestors = 250651853)

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
artidoro: Makes sense! I moved this to a separate file.

---
In reply to: [250648529](https://github.com/dotnet/machinelearning/pull/2146#discussion_r250648529) [](ancestors = 250648529)

src/Microsoft.ML.Data/Transforms/TypeConverting.cs
artidoro: I added one, and did the same for MatrixFactorization.

src/Microsoft.ML.Transforms/Text/NgramTransform.cs
artidoro: Done!

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -905,6 +906,7 @@ private Schema ComputeOutputSchema()
         public const string LoaderSignature = "TextLoader";
 
         private const uint VerForceVectorSupported = 0x0001000A;
+        private const uint VersionNoMinCount = 0x00010004;
```
Ivanidzo4ka: >0x00010004 [](start = 47, length = 10)

Why it's 10004 and not 1000B ?

Issue #2099, opened 2019-01-09
[BUG] Possible Bug when training causes a "never-ending-training" when VS-F5 Debugging/attached but works OK without debugger attached


CESARDELATORRE:
This was working properly in ML.NET v0.8, so looks like a bug appearing in v0.9?

**CONTEXT:**
I just migrated the **Iris-MultiClassClassification** sample to ML.NET **v0.9**.

Code is here, ONLY in **branch 09migration**:

https://github.com/dotnet/machinelearning-samples/tree/09migration/samples/csharp/getting-started/MulticlassClassification_Iris 

**PROBLEM:**
If I run it without attaching to debugger such as with Ctrl+F5, it works/trains properly:

![image](https://user-images.githubusercontent.com/1712635/50934719-60f8c000-141e-11e9-83ca-02928601d8b8.png)

However, if I run the console app with F5 (Debugging with VS attached to the process), then the app is "training" the model "forever", meaining that instead of 20 seconds I've been waiting for more than 30 minutes and it "never ends"..

![image](https://user-images.githubusercontent.com/1712635/50934520-b7193380-141d-11e9-9cfb-7066652189c3.png)

I talked to @danmosemsft and we think it can be related to any issue with too many threads being created?

In the debug window we see a bunch of threads being created, not sure if that's the cause? Too many threads being created and that's causing the issue to Visual Studio?

The thread 0x3ab4 has exited with code 0 (0x0).
The thread 0x5038 has exited with code 0 (0x0).
The thread 0x2994 has exited with code 0 (0x0).
The thread 0x4b18 has exited with code 0 (0x0).
The thread 0xf3c has exited with code 0 (0x0).
The thread 0x4e8 has exited with code 0 (0x0).
The thread 0x3cd0 has exited with code 0 (0x0).
The thread 0x28d4 has exited with code 0 (0x0).
The thread 0x56cc has exited with code 0 (0x0).
The thread 0x38d0 has exited with code 0 (0x0).
The thread 0x658 has exited with code 0 (0x0).
The thread 0x17d4 has exited with code 0 (0x0).
The thread 0x3ef0 has exited with code 0 (0x0).
The thread 0x6284 has exited with code 0 (0x0).
The thread 0x1664 has exited with code 0 (0x0).
The thread 0x2c5c has exited with code 0 (0x0).
The thread 0x536c has exited with code 0 (0x0).
The thread 0x1e04 has exited with code 0 (0x0).
[...]

PR #2152, opened 2019-01-15
Reuse threads in ThreadUtils

stephentoub:
ThreadUtils.CreateBackgroundThread is being truthful to its name and creating a new thread for every call.  However, even for very simple ML.NET usage, such as the MulticlassClassification_Iris demo app, this method is being called ~110,000 times, resulting in ~110,000 threads getting created and destroyed.  That adds measurable overhead in normal runs, but when the debugger is attached it makes the solution effectively unusable, as every thread creation/destruction is tracked by Visual Studio, leading to significant overheads that make an F5 execution last "forever" (== I gave up waiting).

The right solution is for the higher-level algorithms and architecture to be better about its need for threads, creating them only when necessary and otherwise relying on the .NET ThreadPool for execution, via either ThreadPool.QueueUserWorkItem or via Task.Run or the like.

However, as an immediate stop-gap that significantly helps the situation, this commit allows the created threads to be reused for a period of time such that not every call ends up creating a new thread.  In my runs:
- The same demo that app that created ~110K threads now creates only ~32.
- With ctrl-F5 (e.g. no debugger attached), it previously took ~13 seconds, and with this change now takes ~6.
- With F5 (debugger attached), execution previously took "forever", now takes ~190 seconds (still way too high, but vastly improved).

The CreateBackgroundThread method is renamed to StartBackgroundThread, and returns a Task instead of a Thread, such that callers may synchronize with that instead with Thread.Join.  In several cases, this avoids additional threads from being consumed, where callers were blocking a thread pool thread doing a synchronous wait for all tasks, and where now that's entirely avoided via Task.WhenAll.  Eventually, more call sites can be fixed as well, as more of the code base is moved to async/await; for now this just handles the obvious ones that don't require any significant restructuring.

Contributes to https://github.com/dotnet/machinelearning/issues/2099


stephentoub:
ps If we could instead address the number of LineReaders getting created and spinning up threads (https://github.com/dotnet/machinelearning/issues/2099#issuecomment-454514191), maybe this won't be needed.

TomFinley:
While that would probably be a good idea independently, your code seems like genuinely good code just by itself? Do you disagree?

---
In reply to: [454516407](https://github.com/dotnet/machinelearning/pull/2152#issuecomment-454516407) [](ancestors = 454516407)

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
```diff
@@ -4,44 +4,154 @@
...
     internal static partial class Utils
     {
-        public static Thread CreateBackgroundThread(ParameterizedThreadStart start)
-        {
-            return new Thread(start)
-            {
-                IsBackground = true
-            };
-        }
+        public static Task StartBackgroundThread(Action start) =>
```
TomFinley: >StartBackgroundThread [](start = 27, length = 21)

Is the fact that this thing is returning a task, yet the name of the thing still has thread it in, potentially confusing? This is not part of our public API, but might cause some confusion.

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
```diff
@@ -4,44 +4,154 @@
...
-            };
-        }
+        public static Thread CreateForegroundThread(ParameterizedThreadStart start) =>
+            new Thread(start) { IsBackground = false };
 
-        public static Thread CreateForegroundThread(ThreadStart start)
+        /// <summary>
+        /// Naive thread pool focused on reducing the latency to execution of chunky work items as much as possible.
+        /// If a thread is ready to process a work item the moment a work item is queued, it's used, otherwise
+        /// a new thread is created.  This is meant as a stop-gap measure for workloads that would otherwise be
```
TomFinley: >   [](start = 36, length = 2)

So, not a huge blocker, but if you read our comments you'll see that we generally follow the standard advice of one space, rather than two, after sentences. (E.g., [as described here](https://slate.com/technology/2011/01/two-spaces-after-a-period-why-you-should-never-ever-do-it.html).) I would perhaps put it a bit less insultingly than the author of that article, while of course agreeing with the general advice.

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
```diff
@@ -4,44 +4,154 @@
...
+                    // Create a new background thread to run the work.
+                    var t = new Thread(() =>
+                    {
+                        // Repeatedly get the next item and invoke it, setting its TCS when we're done.
+                        // This will wait for up to the idle time before giving up and exiting.
+                        while (TryDequeue(out (Delegate, object, TaskCompletionSource<bool>) item))
+                        {
+                            try
+                            {
+                                if (item.Item1 is Action<object> pts)
```
TomFinley: >Item1 [](start = 41, length = 5)

I don't object to value tuples per se, but I wonder if the `Item1`, `Item2`, `Item3`, impedes code clarity.

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
```diff
@@ -4,44 +4,154 @@
...
+                    });
+                    t.IsBackground = true;
+                    t.Start();
+                }
+
+                bool Enqueue((Delegate, object, TaskCompletionSource<bool>) item)
+                {
+                    // Enqueue the work.  If there are currently fewer threads waiting
+                    // for work than there are work items in the queue, create another
+                    // thread.  This is a heuristic, in that we might end up creating
```
danmosemsft: "create another                     // thread." -> "return false to indicate a new thread should be created"
but perhaps it would be clearer to create the thread within Enqueue?

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
```diff
@@ -4,44 +4,154 @@
...
+                    // more threads than are truly needed, but this whole type is being
+                    // used to replace a previous solution where every work item created
+                    // its own thread, so this is an improvement regardless of any
+                    // such inefficiencies.
+                    lock (_queue)
```
danmosemsft: I'm confused. While a thread is waiting for work, it holds the lock on `_queue`. How can Enqueue take the lock such that it can provide work?

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
```diff
@@ -4,44 +4,154 @@
...
+                return tcs.Task;
+
+                void CreateThread()
+                {
+                    // Create a new background thread to run the work.
+                    var t = new Thread(() =>
+                    {
+                        // Repeatedly get the next item and invoke it, setting its TCS when we're done.
+                        // This will wait for up to the idle time before giving up and exiting.
+                        while (TryDequeue(out (Delegate, object, TaskCompletionSource<bool>) item))
```
danmosemsft: nit, it's easier to read if you name these so you can use eg `tcs` instead of `Item3`

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
danmosemsft: Oh never mind, I forgot that Monitor.Wait releases the lock.

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
```diff
@@ -4,44 +4,154 @@
...
+                    // decreased after.  Each time it waits, it'll wait for at most
+                    // the idle timeout before giving up.
+                    lock (_queue)
+                    {
+                        _availableThreads++;
+                        try
+                        {
+                            while (_queue.Count == 0)
+                            {
+                                if (!Monitor.Wait(_queue, IdleMilliseconds))
```
TomFinley: >Monitor.Wait(_queue [](start = 37, length = 19)

So, I have not actually used `Monitor` directly too often, so see if I understand. You above `lock (_queue)`. Then here, you call `Monitor.Wait`, which when called, *releases* the thread, but then once it returns, the lock is re-established. And then you return from this, whereupon the `finally` block runs and `_availableThreads` is decremented still in the critical section. But if the exception is thrown (in the case where the lock was not reacquired at all), we *still* enter the finally section to decrement, even though we must be out  of the critical section in that specific exception-thrown scenario? Is that right and intentional?

danmosemsft: Can you clarify which exception? I don't think there should be an exception thrown here?

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
danmosemsft: It is possible to name them eg `while (TryDequeue(out (Delegate myDelegate, object myObject, TaskCompletionSource<bool> myTcs) item))` and then use those names instead of `Item1` etc.

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
TomFinley: Sure. So, let me just clarify, that in the absence of direct experience with this method I am going by [the documentation](https://docs.microsoft.com/en-us/dotnet/api/system.threading.monitor.wait?view=netstandard-2.0). So maybe I am way off base.

Nonetheless, that documentation describes two scenarios where it returns `true` and `false` (in both cases the lock is re-established) and a third possibility which is that it does not return at all, and an exception is thrown. It looks to me like the increment/decrement of this counter relies upon the lock being re-established in order to be valid.

Perhaps we have the attitude that, if an exception is thrown, we're totally screwed anyway. I might accept that logic, but, if that *was* in fact the thinking, I and others might benefit from it being made explicit. (But then why bother decrementing at all?)

danmosemsft: I think an exception would only occur if there was a bug in the code, we don't _usually_ code defensively around our own code failing for our own bugs (?). If you're asking because there's a `finally` - I agree, I don't know why there's a finally, since if there is an exception here, it will not be handled, and the process will terminate moments after...

TomFinley: OK. I just wasn't sure if the only conceivable reason for this was a bug in the code being called. Is that true? That does sort of bring up the question though, what happens *if* an exception is thrown.

Sometimes we throw exceptions due to reasons that have nothing whatever to do with bugs in our code -- e.g., the user invokes something with bad input data, we might conceivably throw in one of these workers. Have we considered that this is possible?

I'd be less concerned if this wasn't an API, but it is.

stephentoub:
> While that would probably be a good idea independently, your code seems like genuinely good code just by itself? Do you disagree?

Maybe.  I don't know enough about the usage patterns here to know whether it would be beneficial or not.  If there are other cases where we're expecting to be calling this method lots of times in short succession to do work, then yeah, we should keep this.  If, however, this situation rarely arises once the LineReader thing is fixed, then this is probably just adding unnecessary complexity.

My suggestion would be to take the fix, then work on the LineReader thing, and then re-evaluate whether this can be removed and just replaced with `new Thread(...) { IsBackground = true }.Start()` again.

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
stephentoub: Ok, I'll rename this to RunOnBackgroundThread or something like that.

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
stephentoub: Ok, will change.

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
stephentoub: I can name them.

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
stephentoub: Ok

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
stephentoub: Ok

src/Microsoft.ML.Core/Utilities/ThreadUtils.cs
stephentoub: > So, I have not actually used Monitor directly too often, so see if I understand. You above lock (_queue). Then here, you call Monitor.Wait, which when called, releases the _queue lock, but then once it returns (if it returns!), the lock is re-established.

Yes.

> And then you return from this, whereupon the finally block runs and _availableThreads is decremented still in the critical section.

Correct.

> But if the exception is thrown (in the case where the lock was not reacquired at all), we still enter the finally section to decrement, even though we must be out of the critical section in that specific exception-thrown scenario? Is that right and intentional?

The lock will always be reacquired as part of Wait returning/throwing.  You see it documented otherwise somewhere?

> Can you clarify which exception? I don't think there should be an exception thrown here?

Since this is being used correctly (e.g. the lock is held by this thread when the method is called, the argument object isn't null, the timeout is valid, etc.), the only exceptions that might be thrown are a ThreadAbortException or a ThreadInterruptedException.  The former won't be thrown on core, and in practice neither of them would really be thrown since no one is calling .Abort() or .Interrupt() on the Thread object (the former could be thrown as well if this were all in an AppDomain that was getting torn down, again not on core).  So in reality the try/finally is unlikely to matter, I'm just in the habit of it for things like this.  I can remove it if you prefer.

TomFinley: That's good enough for me thanks. The odd specificity of the documentation "when it returns" is what made me suspicious that something else might happen if it were somehow interrupted. But if it's not so that's great.

justinormont:
This is hitting the 45min timeout for the tests running. 

Closing/rubbing some cheetah blood on it/reopening to re-run the CI test.

stephentoub:
> This is hitting the 45min timeout for the tests running.

Is that likely an indication that I've got a bug here, or is that commonly seen on other PRs?

justinormont:
@stephentoub: I'm seeing other PRs pass tests w/ extra time to spare. So I think it's a good indication that this PR is slowing down the tests. There are benchmarks which you can run on your new code (and the old) to compare runtimes.

stephentoub:
> There are benchmarks which you can run on your new code (and the old) to compare runtimes.

Thanks.  Can you point me to directions on how to run those?

justinormont:
@Anipik: Do we have instructions on how to run the benchmarks?

Anipik:
yes, https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Benchmarks/README.md

The sdk version is mentioned here https://github.com/dotnet/machinelearning/blob/master/DotnetCLIVersion.netcoreapp.latest.txt

currently we are using preview one for netcore3.0

stephentoub:
The issue was just a functional bug, which I fixed (I was missing a check for whether the queue was empty or not).

stephentoub:
I ran the perf tests.  Most of them look like they're within the noise, but the TrainIris test dropped from 10.38s before the change to 4.54s after, and a couple others (e.g. TrainSentiment) look like they might have a 15-20% improvement.  I'm not sure what's different about those that highlights the impact of this.

justinormont:
@stephentoub : Is this ready to be merged in?

@TomFinley : Thanks for the analysis. It offers a good explanation for why Iris & SDCA in particular are doing better. We should likely add caching to the benchmarks (when they were created we had auto-caching); this assumes users know about how to utilize benmarking and therefore we should keep the benchmark representative of user pipelines -- cc: @Anipik

TomFinley:
> @TomFinley : Thanks for the analysis. It offers a good explanation for why Iris & SDCA in particular are doing better. We should likely add caching to the benchmarks (when they were created we had auto-caching); this assumes users know about how to utilize benmarking and therefore we should keep the benchmark representative of user pipelines -- cc: @Anipik

Caching to the benchmarks sounds good. I actually rather like that the test itself does not have caching -- I can think of few better stress tests on our infrastructure than running a dataset as tiny as this through the loader/transformation pipe on something as brutal with the pipeline as SDCA. 😆

That big green button is looking awfully tempting @stephentoub ... I just want to press it.

stephentoub:
> That big green button is looking awfully tempting @stephentoub ... I just want to press it.

Hah :smile: Understood.  I'm re-running the tests one more time, just for a sanity check.

TomFinley:
Like @justinormont I'm not too concerned... so @stephentoub it is not the test that is flaky, but neither would I blame your change. I think it is a real issue in the existing code. We've observed on the test machines, when reading data, on the first pass that read "ends" early. This test is the Iris dataset of 150 lines, with the 3 classes appearing in sequence in groups of 50 lines. In this case it has read 50 or fewer lines when trying to detect the number of classes, and complains "hey wait a second, there's only one class here, I can't do anything with this." As we see below in your log:

> ```
> 2019-01-20T15:11:50.3696166Z  System.ArgumentOutOfRangeException : Must be at least 2.
> 2019-01-20T15:11:50.3696444Z Parameter name: numClasses
> ```

I don't recall seeing this particular test failing too often before, but I consider this failure a manifestation of that same bug. (If only because it seems to be consistent with our hypothesis of what the problem is. The more common case seems to be in SDCA complaining, "wait a second, in my first pass I observed only X instances, suddenly I've seen in a subsequent pass an X+1th instance, what are you trying to pull on me?"

Still we have trouble: attempts to repro locally are elusive. As in, I don't think anyone has succeeded in doing so despite many attempts. So even if your change was somehow making the flaw in our code more obvious, we'd obviously welcome it, since then maybe we could *finally* get a handle on this thing.

I could easily believe your change might make the bug more obvious: for various reasons that wouldn't be terribly interesting to review, even though we don't have solid information on the bug, indirect information suggests a race condition: text loader, the column-set consolidator, or the cache data view, is inappropriately reporting that there is no more data from its source when in fact there is more data. Delays in multithreaded computation have a way of making race conditions less obvious; you got rid of a lot of delays, as we see from your benchmarks, so maybe suddenly we'll start seeing this more often. Maybe even, if we're extra lucky, even in a debugger where we can actually see what the heck is going on, finally. 

But that's just my hope. Who knows.

stephentoub:
Thanks for the solid explanation.  Good enough for me.

stephentoub:
> Just wondering: @stephentoub can you merge, or do you not have that privilege in this repo? 

@TomFinley, I don't have the privilege.  Enjoy :smile:

PR #2153, opened 2019-01-15
Adding packages to the solution

### LINKED ISSUE #1168: Solution file don't have references for all package projects
https://github.com/dotnet/machinelearning/tree/master/pkg
we have quite a lot of packages, but solution file has reference only for 3 of them
https://github.com/dotnet/machinelearning/blob/d5175899b604d852dab08107e583c78cbe5d6948/Microsoft.ML.sln#L52

Should we update solution and have policy to add new packages into it, or should we just remove them from solution?

Any thoughts?

shauheen:
Does not effect users, however would be nice to have all packages in the solution. Fixes #1168


Ivanidzo4ka:
> 		pkg\Microsoft.ML\build\netstandard2.0\Microsoft.ML.props = pkg\Microsoft.ML\build\netstandard2.0\Microsoft.ML.props

we no longer have this file, can you remove it?

---
Refers to: Microsoft.ML.sln:70 in 0848156. [](commit_id = 08481563c246dc28b0cb710779cb2979a0be22c8, deletion_comment = False)

Issue #1945, opened 2018-12-20
lib_lightgbm.dll is not getting loaded while running benchmarks on .NetFramework


Anipik:
when we try to run this https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Benchmarks/Numeric/Ranking.cs#L50
and 
https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Benchmarks/Text/MultiClassClassification.cs#L47

on .netFramework, the benchmark is not able to load the lib_lightgbm.dll
```
Unexpected exception: Unable to load DLL 'lib_lightgbm': The specified module could not be found. (Exception from HRESULT: 0x8007007E), 'System.DllNotFoundException'
```

The benchmarks run properly on .NetCore

cc @danmosemsft @eerhardt @adamsitnik

danmosemsft:
@TomFinley who is the owner of the lightgbm package who can look at this?

codemzs:
Are you getting this error on other non windows platform? Based on very quick search seems like windows upgrade error. This dll comes from a nuget that is published by @guolinke

Anipik:
I am getting this error in windows platform. I have not checked it for non-windows platform

eerhardt:
@Anipik - can you try adding `<RuntimeIdentifier>win-x64</RuntimeIdentifier>` to your .csproj? By default, .NET Framework apps run as 32-bit applications, and LightGBM doesn't support the x86 architecture. It only supports `win-x64`.

Anipik:
@eerhardt I tried running it by specifying 64 bit  bit its still not able to find the lighgbm dll

artemiusgreat:
TL.TR - another way to fix this issue is to install Microsoft Visual C++ Redistributable x64 
https://support.microsoft.com/en-ae/help/2977003/the-latest-supported-visual-c-downloads 

This is how it happened. 

1. Local computer works fine - Windows Pro x64 
2. OVH VPS server works fine - Windows Server 2016 x64 
3. Vultr VPS doesn't load LightGbm DLL - Windows Pro x64 

[...]

PR #2157, opened 2019-01-16
Support running benchmarks on netfx

Anipik:
Fixes https://github.com/dotnet/machinelearning/issues/1945

This PR makes changes in order to enable developers to run ML benchmarks against .NetFramework

### DIFF
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -2,10 +2,9 @@
   <Import Project="..\Directory.Build.props" />
 
   <PropertyGroup>
-    <DefaultTestTargetFramework>netcoreapp2.1</DefaultTestTargetFramework>
-    <DefaultTestTargetFramework Condition="$(Configuration.EndsWith('-netfx'))">net462</DefaultTestTargetFramework>
-    <DefaultTestTargetFramework Condition="'$(UseIntrinsics)' == 'true'">netcoreapp3.0</DefaultTestTargetFramework>
-    <TargetFramework>$(DefaultTestTargetFramework)</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework Condition="$(Configuration.EndsWith('-netfx'))">net462</TargetFramework>
+    <TargetFramework Condition="'$(UseIntrinsics)' == 'true'">netcoreapp3.0</TargetFramework>
     <RuntimeIdentifier Condition="'$(TargetFramework)' == 'net462'">win-x64</RuntimeIdentifier>
     <IsPackable>false</IsPackable>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -3,9 +3,9 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
-    <TargetFramework Condition="$(Configuration.EndsWith('-netfx'))">net462</TargetFramework>
+    <TargetFramework Condition="$(Configuration.EndsWith('-netfx'))">net461</TargetFramework>
     <TargetFramework Condition="'$(UseIntrinsics)' == 'true'">netcoreapp3.0</TargetFramework>
-    <RuntimeIdentifier Condition="'$(TargetFramework)' == 'net462'">win-x64</RuntimeIdentifier>
+    <RuntimeIdentifier Condition="'$(TargetFramework)' == 'net461'">win-x64</RuntimeIdentifier>
     <IsPackable>false</IsPackable>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
 
--- a/test/Microsoft.ML.Benchmarks/Harness/Configs.cs
+++ b/test/Microsoft.ML.Benchmarks/Harness/Configs.cs
@@ -38,8 +38,8 @@ namespace Microsoft.ML.Benchmarks
         /// </summary>
         private IToolchain CreateToolchain()
         {
-#if NET462
-            var tfm = "net462";
+#if NET461
+            var tfm = "net461";
             var csProj = CsProjClassicNetToolchain.Net462;
 #else
             var tfm = NetCoreAppSettings.Current.Value.TargetFrameworkMoniker;
@@ -56,7 +56,7 @@ namespace Microsoft.ML.Benchmarks
         {
 #if NETCOREAPP3_0
             return "Release-Intrinsics";
-#elif NET462
+#elif NET461
             return "Release-netfx";
 #else
             return "Release";
--- a/test/Microsoft.ML.Benchmarks/Harness/Configs.cs
+++ b/test/Microsoft.ML.Benchmarks/Harness/Configs.cs
@@ -14,7 +14,7 @@ namespace Microsoft.ML.Benchmarks
 {
     public class RecommendedConfig : ManualConfig
     {
-        protected static List<MsBuildArgument> msbuildArguments = new List<MsBuildArgument>() { new MsBuildArgument($"/p:Configuration={GetBuildConfigurationName()}") };
+        protected static readonly IReadOnlyList<MsBuildArgument> msbuildArguments = new List<MsBuildArgument>() { new MsBuildArgument($"/p:Configuration={GetBuildConfigurationName()}") };
 
         public RecommendedConfig()
         {
@@ -40,7 +40,7 @@ namespace Microsoft.ML.Benchmarks
         {
 #if NET461
             var tfm = "net461";
-            var csProj = CsProjClassicNetToolchain.Net462;
+            var csProj = CsProjClassicNetToolchain.Net461;
 #else
             var tfm = NetCoreAppSettings.Current.Value.TargetFrameworkMoniker;
             var csProj = CsProjCoreToolchain.Current.Value;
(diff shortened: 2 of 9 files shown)


test/Directory.Build.props
```diff
@@ -2,10 +2,9 @@
   <Import Project="..\Directory.Build.props" />
 
   <PropertyGroup>
-    <DefaultTestTargetFramework>netcoreapp2.1</DefaultTestTargetFramework>
-    <DefaultTestTargetFramework Condition="$(Configuration.EndsWith('-netfx'))">net462</DefaultTestTargetFramework>
-    <DefaultTestTargetFramework Condition="'$(UseIntrinsics)' == 'true'">netcoreapp3.0</DefaultTestTargetFramework>
-    <TargetFramework>$(DefaultTestTargetFramework)</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework Condition="$(Configuration.EndsWith('-netfx'))">net462</TargetFramework>
```
eerhardt: (not part of this change) - do you remember why we choose `net462` instead of `net461`?

test/Microsoft.ML.Benchmarks/Harness/Configs.cs
```diff
@@ -22,6 +25,7 @@ public RecommendedConfig()
 
         protected virtual Job GetJobDefinition()
             => Job.Default
+                .With(msbuildArguments)
```
eerhardt: Isn't this a duplicate of line 20 above?

```
                .WithCustomBuildConfiguration(GetBuildConfigurationName())
```

Why do we need both?

test/Microsoft.ML.Benchmarks/Harness/Configs.cs
```diff
@@ -55,6 +65,7 @@ public class TrainConfig : RecommendedConfig
     {
         protected override Job GetJobDefinition()
             => Job.Dry // the "Dry" job runs the benchmark exactly once, without any warmup to mimic real-world scenario
+                  .With(msbuildArguments)
```
eerhardt: Shouldn't this be

```
.WithCustomBuildConfiguration(GetBuildConfigurationName())
```
?

eerhardt:
> using System.Collections.Generic;

Can you add a copyright header while you are in here?

---
Refers to: test/Microsoft.ML.Benchmarks/Harness/Configs.cs:1 in 42b36d7. [](commit_id = 42b36d7e91439554bb7a27afd3ea228a671964c8, deletion_comment = False)

test/Directory.Build.props
Anipik: I remember trying it with net471 and santi suggested to try with net462 to just make the build successfull. It was a while back. and i didnt try building with any other versions

danmosemsft: If net461 works we should probably use that everywhere, since that supports NET Standard 2.0 it seems like the natural choice.

test/Microsoft.ML.Benchmarks/Harness/ProjectGenerator.cs
```diff
@@ -35,7 +42,7 @@ protected override void GenerateProject(BuildPartition buildPartition, Artifacts
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <OutputPath>bin\{buildPartition.BuildConfiguration}</OutputPath>
-    <TargetFramework>{TargetFrameworkMoniker}</TargetFramework>
+    <TargetFramework>{TargetFrameworkMoniker}</TargetFramework>{rid}
```
adamsitnik: I think that MsBuild ignores empty `<RuntimeIdentifier></RuntimeIdentifier>` so we can remove the condition and the need for extra variable

```suggestion
    <TargetFramework>{TargetFrameworkMoniker}</TargetFramework>
    <RuntimeIdentifier>{runtimeIdentifier}</RuntimeIdentifier>
```

test/Microsoft.ML.Benchmarks/Harness/ProjectGenerator.cs
```diff
@@ -53,6 +60,9 @@ protected override void GenerateProject(BuildPartition buildPartition, Artifacts
   </ItemGroup>
 </Project>");
 
+        protected override string GetBinariesDirectoryPath(string buildArtifactsDirectoryPath, string configuration)
```
adamsitnik: it would be good to add a comment explaining why we need that

test/Microsoft.ML.Benchmarks/Numeric/Ranking.cs
```diff
@@ -85,8 +85,11 @@ public void SetupScoringSpeedTests()
             if (!File.Exists(_mslrWeb10k_Train))
                 throw new FileNotFoundException(string.Format(Errors.DatasetNotFound, _mslrWeb10k_Train));
 
+#if !NET461
+            _modelPath_MSLR = Path.Combine(Directory.GetParent(typeof(RankingTest).Assembly.CodeBase.Substring(8)).FullName, @"FastTreeRankingModel.zip");
```
adamsitnik: `Substring(8)` is very fragile. Can find a way to avoid the substring? Sth like:

```suggestion
            _modelPath_MSLR = Path.Combine(Path.GetDirectoryName(typeof(RankingTest).Assembly.Location), "FastTreeRankingModel.zip");
```

test/Microsoft.ML.Benchmarks/Text/MultiClassClassification.cs
```diff
@@ -104,7 +104,11 @@ public void SetupScoringSpeedTests()
             if (!File.Exists(_dataPath_Wiki))
                 throw new FileNotFoundException(string.Format(Errors.DatasetNotFound, _dataPath_Wiki));
 
+#if NET461
+            _modelPath_Wiki = Path.Combine(Directory.GetParent(typeof(MultiClassClassificationTest).Assembly.CodeBase.Substring(8)).FullName, @"WikiModel.zip");
```
adamsitnik: again: substring

test/Microsoft.ML.Benchmarks/Harness/ProjectGenerator.cs
```diff
@@ -25,8 +25,15 @@ namespace Microsoft.ML.Benchmarks.Harness
     /// </summary>
     public class ProjectGenerator : CsProjGenerator
     {
+        private string rid = string.Empty;
+        private string ridValue = string.Empty;
```
adamsitnik:
```suggestion
```

test/Microsoft.ML.Benchmarks/Numeric/Ranking.cs
```diff
@@ -85,7 +85,7 @@ public void SetupScoringSpeedTests()
             if (!File.Exists(_mslrWeb10k_Train))
                 throw new FileNotFoundException(string.Format(Errors.DatasetNotFound, _mslrWeb10k_Train));
 
-            _modelPath_MSLR = Path.Combine(Directory.GetCurrentDirectory(), @"FastTreeRankingModel.zip");
+            _modelPath_MSLR = Path.Combine(Path.GetDirectoryName(typeof(RankingTest).Assembly.Location), "FastTreeRankingModel.zip");
```
eerhardt: Why is this change necessary?

Anipik: otherwise on .netfx it creates this new model file in the root directory

eerhardt: This confuses me. What do you mean by "the root directory"? Why would the current netfx have its current directory set to that?

also, the code is still using Directory.GetCurrentDirectory() in `Test_Multiclass_WikiDetox_BigramsAndTrichar_OVAAveragedPerceptron`. WHy doesn't that need to change?

Anipik: > Why would the current netfx have its current directory set to that?

I don't have nay concrete explaination for why it is setting current directory to ```C:\git\Mlnet\test\Microsoft.ML.Benchmarks```

eerhardt: @adamsitnik - do you happen to know? Is this something BenchmarkDotNet is controlling?

adamsitnik: @eerhardt  Yes BDN sets the working directory to the folder with built dlls for .NET Core and to null for FX.  The problem is that I dont remember why and I would prefer not to change it if I dont have to ;) It most probably has to deal with the fact that we run executable for FX and do `dotnet name.dll` for Core

https://github.com/dotnet/BenchmarkDotNet/blob/573566b70f0ff27572b9268002fe9748be4de6ea/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliExecutor.cs#L72

https://github.com/dotnet/BenchmarkDotNet/blob/573566b70f0ff27572b9268002fe9748be4de6ea/src/BenchmarkDotNet/Toolchains/Executor.cs#L35

Anipik:
@eerhardt is there anything else that needs to be done here ?

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -34,7 +34,7 @@ public abstract partial class BaseTestBaseline : BaseTestClass
 
         protected BaseTestBaseline(ITestOutputHelper output) : base(output)
         {
-#if NET462
+#if NET461
```
eerhardt: NOTE: I just learned about `VersionlessImplicitFrameworkDefine` in the .NET Core SDK - https://github.com/dotnet/sdk/pull/2073. It appears this can be changed to `#if NETFRAMEWORK`

Anipik: I will add this change in my next PR

PR #2161, opened 2019-01-16
Changing LicenseURL to License in Nugets

### LINKED ISSUE #2095: nuget metadata needs to be updated
`<licenseUrl>` element will be deprecated, please consider switching to specifying the license in the package. [Learn more](https://aka.ms/deprecateLicenseUrl).

shauheen:
Fixes #2095 

Nuget.org requires updating the metadata to use License instead. This PR changes that.


pkg/Microsoft.ML.TensorFlow.Redist/Microsoft.ML.TensorFlow.Redist.nupkgproj
```diff
@@ -4,7 +4,7 @@
     <Authors>The TensorFlow Authors</Authors>
     <TargetFramework>netstandard2.0</TargetFramework>
     <PackageDescription>$(MSBuildProjectName) contains the TensorFlow C library version $(TensorFlowVersion) redistributed as a NuGet package.</PackageDescription>
-    <PackageLicenseUrl>https://github.com/tensorflow/tensorflow/blob/master/LICENSE</PackageLicenseUrl>
+    <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
```
eerhardt: NOTE: a difference I see between

https://github.com/aspnet/AspNetCore/blob/master/LICENSE.txt
and
https://github.com/tensorflow/tensorflow/blob/master/LICENSE

is that the TensorFlow one has a Copyright at the top

```Copyright 2018 The TensorFlow Authors.  All rights reserved.```

We do download the actual TF license (see line 22 in this file). Maybe we should just be including that as the `PackageLicenseFile`, and not use this `PackageLicenseExpression`, since the license may actually be different.

NOTE: I think there's an issue on how we are downloading that license file. The way we currently have it, we are downloading the GitHub HTML page with the license information. But instead, we should be getting the content of the file.
Basically, we should change
https://github.com/dotnet/machinelearning/blob/1b6b3c3ab22a79e593913fbc9397cef93256aa9c/src/Redist/Microsoft.ML.TensorFlow.Redist/Microsoft.ML.TensorFlow.Redist.proj#L13

to use ```https://raw.githubusercontent.com/tensorflow/tensorflow/master/LICENSE``` instead.

File: pkg/Directory.Build.props [hunk 2]
```diff
@@ -21,7 +21,7 @@
   <!-- nuspec properties -->
   <PropertyGroup>
     <Authors>Microsoft</Authors>
-    <PackageLicenseUrl>https://github.com/dotnet/machinelearning/blob/master/LICENSE</PackageLicenseUrl>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
```
eerhardt: Similar comment here - our LICENSE file in the repo has a copyright in it. I think we should just be packaging that file - `$(RepoRoot)\LICENSE`

shauheen:
locally the NuGets are building fine, however I see `Enabling license acceptance requires a license url.` on the build system. The [documentation](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-a-license-expression-or-a-license-file) doesn't seem to have anything for this. @eerhardt do you have any info on this?

pkg/Microsoft.ML.TensorFlow.Redist/Microsoft.ML.TensorFlow.Redist.nupkgproj
```diff
@@ -4,7 +4,7 @@
     <Authors>The TensorFlow Authors</Authors>
     <TargetFramework>netstandard2.0</TargetFramework>
     <PackageDescription>$(MSBuildProjectName) contains the TensorFlow C library version $(TensorFlowVersion) redistributed as a NuGet package.</PackageDescription>
-    <PackageLicenseUrl>https://github.com/tensorflow/tensorflow/blob/master/LICENSE</PackageLicenseUrl>
+    <PackageLicenseFile>LICENSE.txt</PackageLicenseFile>
```
abgoswam: does it get downloaded with .txt extension ?

(in the .proj file I do not see the .txt extension)

eerhardt: The `.txt` extension gets added when the file gets copied:

https://github.com/dotnet/machinelearning/blob/b89ce70824885d8c3a57dc62d6c88cbae9fd8a60/src/Redist/Microsoft.ML.TensorFlow.Redist/Microsoft.ML.TensorFlow.Redist.proj#L106-L108

abgoswam: awesome. thanks for the info.

---
In reply to: [248750410](https://github.com/dotnet/machinelearning/pull/2161#discussion_r248750410) [](ancestors = 248750410)

PR #2169, opened 2019-01-17
Rename types inside MLContext as Catalogs

### LINKED ISSUE #1796: Rename types inside MLContext as Catalogs
We had an idea in #949 to have "context" objects that enabled the easy and convenient creation of training algorithms. We eventually came to like this idea so much that it expanded considerably, until we came to think it actually a good idea that *most* components (whether trainers or not) could work through this object, to the point where eventually in in #1098, there was a "master" context, of which the formerly independent contexts became properties of that master context.

The newer things that were added were then called "catalogs," for example:

https://github.com/dotnet/machinelearning/blob/cb37c7e7f1e1b29b5608a2755db793c5435d10b1/src/Microsoft.ML.Data/MLContext.cs#L47

https://github.com/dotnet/machinelearning/blob/cb37c7e7f1e1b29b5608a2755db793c5435d10b1/src/Microsoft.ML.Data/MLContext.cs#L52

However the old objects retained the type names they had back when they were independent.

https://github.com/dotnet/machinelearning/blob/cb37c7e7f1e1b29b5608a2755db793c5435d10b1/src/Microsoft.ML.Data/MLContext.cs#L25

We ought to probably standardize the name of these things as being "catalogs." E.g., things in here with the type name suffix `Context` should change that to `Catalog`, things that are catalogs but aren't named that yet (`DataOperations`) should be standardized as well.

najeeb-kazmi:
Fixes #1796 

Renames `XyzContext` types inside `MLContext` to `XyzCatalog`.

Renames `DataOperations` to `DataOperationsCatalog`.


src/Microsoft.ML.Core/Prediction/TrainCatalog.cs
```diff
@@ -8,11 +8,11 @@ namespace Microsoft.ML
 {
     /// <summary>
     /// Holds information relevant to trainers. Instances of this class are meant to be constructed and passed
-    /// into <see cref="ITrainer{TPredictor}.Train(TrainContext)"/> or <see cref="ITrainer.Train(TrainContext)"/>.
+    /// into <see cref="ITrainer{TPredictor}.Train(TrainCatalog)"/> or <see cref="ITrainer.Train(TrainCatalog)"/>.
     /// This holds at least a training set, as well as optioonally a predictor.
     /// </summary>
     [BestFriend]
-    internal sealed class TrainContext
+    internal sealed class TrainCatalog
```
sfilipi: >TrainCatalog [](start = 26, length = 12)

I don't think this is to be named a catalog. we overload the term context. Context is fine in this case..

src/Microsoft.ML.Data/Training/TrainerBase.cs
```diff
@@ -29,11 +29,11 @@ private protected TrainerBase(IHostEnvironment env, string name)
             Host = env.Register(name);
         }
 
-        IPredictor ITrainer.Train(TrainContext context) => Train(context);
+        IPredictor ITrainer.Train(TrainCatalog context) => Train(context);
 
-        TPredictor ITrainer<TPredictor>.Train(TrainContext context) => Train(context);
+        TPredictor ITrainer<TPredictor>.Train(TrainCatalog context) => Train(context);
```
TomFinley: >TrainCatalog [](start = 46, length = 12)

Hmmm... reading through this, it seems like the majority of changes I've seen at least so far have been due to changing `TrainContext`, which as @sfilipi pointed out should not be done. I'll try to re-review again once that's undone, if that's OK?

najeeb-kazmi: @TomFinley undone!

TomFinley: Thanks, starting again now

src/Microsoft.ML.StaticPipe/KMeansStatic.cs
```diff
@@ -16,9 +16,9 @@ namespace Microsoft.ML.StaticPipe
     public static class KMeansClusteringExtensions
     {
         /// <summary>
-        /// KMeans <see cref="ClusteringContext"/> extension method.
+        /// KMeans <see cref="ClusteringCatalog"/> extension method.
         /// </summary>
-        /// <param name="ctx">The clustering context trainer object.</param>
+        /// <param name="catalog">The clustering catalog trainer object.</param>
```
TomFinley: >catalog [](start = 25, length = 7)

Great, thanks for fixing these as well, I'd kind of forgotten that we'd want to change the parameter names and descriptors here... and there sure are a lot of them. ;)

test/Microsoft.ML.StaticPipelineTesting/Training.cs
```diff
@@ -38,7 +38,7 @@ public void SdcaRegression()
             var dataPath = GetDataPath(TestDatasets.generatedRegressionDataset.trainFilename);
             var dataSource = new MultiFileSource(dataPath);
 
-            var ctx = new RegressionContext(env);
+            var catalog = new RegressionCatalog(env);
```
TomFinley: >new RegressionCatalog(env); [](start = 26, length = 27)

Hmmmm interesting. This guy is a bit of an odd duck, I guess it was done before we "unified" the contexts under `MLContext`. We'll probably want to bring these more up to date at some point lest they mislead people. (Not part of this PR, to be clear.)

fwaris:
advanced settings was removed from the API for 
 ctx.BinaryClassification.Trainers.FastTree

Is there a simple way of specifying  "unbalancedSets" as before?

PR #2176, opened 2019-01-17
Remove ColumnType.RawKind usages Round 2.

eerhardt:
Removes the "easy" usages of ColumnType.RawKind.

There are still about 38 usages of ColumnType.RawKind left, but they are a little more involved, and needs refactoring. I will fully remove them in the next PR.

### DIFF
--- a/src/Microsoft.ML.Core/Data/ColumnType.cs
+++ b/src/Microsoft.ML.Core/Data/ColumnType.cs
@@ -120,6 +120,22 @@ namespace Microsoft.ML.Data
                 return DateTimeOffsetType.Instance;
             return NumberType.FromKind(kind);
         }
+
+        [BestFriend]
+        internal static PrimitiveType FromType(Type type)
+        {
+            if (type == typeof(ReadOnlyMemory<char>) || type == typeof(string))
+                return TextType.Instance;
+            if (type == typeof(bool))
+                return BoolType.Instance;
+            if (type == typeof(TimeSpan))
+                return TimeSpanType.Instance;
+            if (type == typeof(DateTime))
+                return DateTimeType.Instance;
+            if (type == typeof(DateTimeOffset))
+                return DateTimeOffsetType.Instance;
+            return NumberType.FromType(type);
+        }
     }
 
     /// <summary>
@@ -325,7 +341,7 @@ namespace Microsoft.ML.Data
         }
 
         [BestFriend]
-        internal static NumberType FromType(Type type)
+        internal static new NumberType FromType(Type type)
         {
             DataKind kind;
             if (type.TryGetDataKind(out kind))
@@ -339,7 +355,7 @@ namespace Microsoft.ML.Data
         {
             if (other == this)
                 return true;
-            Contracts.Assert(other == null || !(other is NumberType) || other.RawKind != RawKind);
+            Contracts.Assert(other == null || !(other is NumberType) || other.RawType != RawType);
             return false;
         }
 
@@ -589,9 +605,8 @@ namespace Microsoft.ML.Data
 
             if (!(other is KeyType tmp))
                 return false;
-            if (RawKind != tmp.RawKind)
+            if (RawType != tmp.RawType)
                 return false;
-            Contracts.Assert(RawType == tmp.RawType);
             if (Contiguous != tmp.Contiguous)
                 return false;
             if (Min != tmp.Min)
--- a/src/Microsoft.ML.Core/Data/ColumnType.cs
+++ b/src/Microsoft.ML.Core/Data/ColumnType.cs
@@ -124,7 +124,7 @@ namespace Microsoft.ML.Data
         [BestFriend]
         internal static PrimitiveType FromType(Type type)
         {
-            if (type == typeof(ReadOnlyMemory<char>) || type == typeof(string))
+            if (type == typeof(ReadOnlyMemory<char>))
                 return TextType.Instance;
             if (type == typeof(bool))
                 return BoolType.Instance;
--- a/src/Microsoft.ML.Core/Data/DataKind.cs
+++ b/src/Microsoft.ML.Core/Data/DataKind.cs
@@ -227,6 +227,41 @@ namespace Microsoft.ML.Data
         }
 
         /// <summary>
+        /// Returns true if type is a valid DataKind type. Otherwise, false.
+        /// </summary>
+        public static bool IsValidDataKindType(this Type type)
+        {
+            Contracts.CheckValueOrNull(type);
+
+            return type == typeof(sbyte)
+                || type == typeof(byte)
+                || type == typeof(short)
+                || type == typeof(ushort)
+                || type == typeof(int)
+                || type == typeof(uint)
+                || type == typeof(long)
+                || type == typeof(ulong)
+                || type == typeof(float)
+                || type == typeof(double)
+                || type == typeof(ReadOnlyMemory<char>) || type == typeof(string)
+                || type == typeof(bool)
+                || type == typeof(TimeSpan)
+                || type == typeof(DateTime)
+                || type == typeof(DateTimeOffset)
+                || type == typeof(RowId);
+        }
+
+        /// <summary>
+        /// Returns true if the types are compatible with each other.
+        /// </summary>
+        public static bool AreDataKindCompatibleTypes(Type left, Type right)
+        {
+            return left == right
+                || (left == typeof(ReadOnlyMemory<char>) && right == typeof(string))
+                || (left == typeof(string) && right == typeof(ReadOnlyMemory<char>));
+        }
+
+        /// <summary>
         /// Get the canonical string for a DataKind. Note that using DataKind.ToString() is not stable
         /// and is also slow, so use this instead.
         /// </summary>
--- a/src/Microsoft.ML.Core/Data/DataKind.cs
+++ b/src/Microsoft.ML.Core/Data/DataKind.cs
@@ -227,41 +227,6 @@ namespace Microsoft.ML.Data
         }
 
         /// <summary>
-        /// Returns true if type is a valid DataKind type. Otherwise, false.
-        /// </summary>
-        public static bool IsValidDataKindType(this Type type)
-        {
-            Contracts.CheckValueOrNull(type);
-
-            return type == typeof(sbyte)
-                || type == typeof(byte)
-                || type == typeof(short)
-                || type == typeof(ushort)
-                || type == typeof(int)
-                || type == typeof(uint)
-                || type == typeof(long)
-                || type == typeof(ulong)
-                || type == typeof(float)
-                || type == typeof(double)
-                || type == typeof(ReadOnlyMemory<char>) || type == typeof(string)
-                || type == typeof(bool)
-                || type == typeof(TimeSpan)
-                || type == typeof(DateTime)
-                || type == typeof(DateTimeOffset)
-                || type == typeof(RowId);
-        }
-
-        /// <summary>
-        /// Returns true if the types are compatible with each other.
-        /// </summary>
-        public static bool AreDataKindCompatibleTypes(Type left, Type right)
-        {
-            return left == right
-                || (left == typeof(ReadOnlyMemory<char>) && right == typeof(string))
-                || (left == typeof(string) && right == typeof(ReadOnlyMemory<char>));
-        }
-
-        /// <summary>
         /// Get the canonical string for a DataKind. Note that using DataKind.ToString() is not stable
         /// and is also slow, so use this instead.
         /// </summary>
--- a/src/Microsoft.ML.Core/Data/MetadataUtils.cs
+++ b/src/Microsoft.ML.Core/Data/MetadataUtils.cs
@@ -238,7 +238,7 @@ namespace Microsoft.ML.Data
             for (int col = 0; col < schema.Count; col++)
             {
                 var columnType = schema[col].Metadata.Schema.GetColumnOrNull(metadataKind)?.Type;
-                if (columnType == null || !(columnType is KeyType) || columnType.RawKind != DataKind.U4)
+                if (columnType == null || !(columnType is KeyType) || columnType.RawType != typeof(uint))
                     continue;
                 if (filterFunc != null && !filterFunc(schema, col))
                     continue;
@@ -263,7 +263,7 @@ namespace Microsoft.ML.Data
             for (int col = 0; col < schema.Count; col++)
             {
                 var columnType = schema[col].Metadata.Schema.GetColumnOrNull(metadataKind)?.Type;
-                if (columnType != null && columnType is KeyType && columnType.RawKind == DataKind.U4)
+                if (columnType != null && columnType is KeyType && columnType.RawType == typeof(uint))
                 {
                     uint val = 0;
                     schema[col].Metadata.GetValue(metadataKind, ref val);
--- a/src/Microsoft.ML.Core/Data/MetadataUtils.cs
+++ b/src/Microsoft.ML.Core/Data/MetadataUtils.cs
@@ -238,7 +238,7 @@ namespace Microsoft.ML.Data
             for (int col = 0; col < schema.Count; col++)
             {
                 var columnType = schema[col].Metadata.Schema.GetColumnOrNull(metadataKind)?.Type;
-                if (columnType == null || !(columnType is KeyType) || columnType.RawType != typeof(uint))
+                if (!(columnType is KeyType) || columnType.RawType != typeof(uint))
                     continue;
                 if (filterFunc != null && !filterFunc(schema, col))
                     continue;
@@ -263,7 +263,7 @@ namespace Microsoft.ML.Data
             for (int col = 0; col < schema.Count; col++)
             {
                 var columnType = schema[col].Metadata.Schema.GetColumnOrNull(metadataKind)?.Type;
-                if (columnType != null && columnType is KeyType && columnType.RawType == typeof(uint))
+                if (columnType is KeyType && columnType.RawType == typeof(uint))
                 {
                     uint val = 0;
                     schema[col].Metadata.GetValue(metadataKind, ref val);
@@ -283,7 +283,7 @@ namespace Microsoft.ML.Data
             for (int col = 0; col < schema.Count; col++)
             {
                 var columnType = schema[col].Metadata.Schema.GetColumnOrNull(metadataKind)?.Type;
-                if (columnType != null && columnType is TextType)
+                if (columnType is TextType)
                 {
                     ReadOnlyMemory<char> val = default;
                     schema[col].Metadata.GetValue(metadataKind, ref val);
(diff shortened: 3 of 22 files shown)


src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -409,102 +409,91 @@ private ValueGetter<uint> ComposeGetterOne(Row input, int iinfo, int srcCol, Col
...
-                case DataKind.U1:
-                    return MakeScalarHashGetter<byte, HashU1>(input, srcCol, seed, mask);
-                case DataKind.U2:
-                    return MakeScalarHashGetter<ushort, HashU2>(input, srcCol, seed, mask);
-                case DataKind.U4:
-                    return MakeScalarHashGetter<uint, HashU4>(input, srcCol, seed, mask);
-                case DataKind.U8:
-                    return MakeScalarHashGetter<ulong, HashU8>(input, srcCol, seed, mask);
-                case DataKind.U16:
+            if (srcType.RawType == typeof(byte))
```
TomFinley: >if (srcType.RawType == typeof(byte)) [](start = 12, length = 36)

It used to not matter as much since it was a case statement, but now that it's a chained if statement we probably should give some thought to what is the most common case. The most common case for hashing is text, so we should probably check that *first*, rather than letting it fall through as the absolute last case when all else fails.

Also probably this would probably be a good time to to checks against the `ColumnType` instances themselves -- the original author did things in fashion because they considered it to be more concise, but I'm not sure that reasoning holds any longer.

src/Microsoft.ML.Data/Transforms/KeyToVector.cs
```diff
@@ -767,7 +767,7 @@ public override SchemaShape GetOutputSchema(SchemaShape inputSchema)
             {
                 if (!inputSchema.TryFindColumn(colInfo.Input, out var col))
                     throw Host.ExceptSchemaMismatch(nameof(inputSchema), "input", colInfo.Input);
-                if ((col.ItemType.GetItemType().RawKind == default) || !(col.ItemType is VectorType || col.ItemType is PrimitiveType))
+                if (!col.ItemType.GetItemType().RawType.IsValidDataKindType() || !(col.ItemType is VectorType || col.ItemType is PrimitiveType))
```
TomFinley: >col.ItemType.GetItemType().RawType.IsValidDataKindType() || !(col.ItemType is VectorType || col.ItemType is PrimitiveType)) [](start = 21, length = 123)

Not your code, but this test seems completely wrong. As in, almost nothing about this test looks right, or even sensible. `col.ItemType` could *never* be a `VectorType` in any instance, according to its definition. In fact under the current type scheme, discounting any types a user may create themselves and the image type, I think this will succeed on everything??

Actually wait, the error message below is wrong too... it will just claim the column is missing (you see it is identical to the test below)... Many of the metadata tests seem off... And of course it claims that it always produces a vector of known size, but that's definitely not true. Wow. I was going to suggest you fix this when I thought it was just confined to one test, but it seems like literally half the lines in this are flat out wrong. OK. :) I'll file a separate issue...

Filed as #2185. Don't bother fixing of course. :D :D

src/Microsoft.ML.Data/Transforms/Normalizer.cs
```diff
@@ -352,7 +352,7 @@ internal static void SaveType(ModelSaveContext ctx, ColumnType type)
                 ctx.Writer.Write(vectorType?.Size ?? 0);
 
                 ColumnType itemType = vectorType?.ItemType ?? type;
-                var itemKind = itemType.RawKind;
+                itemType.RawType.TryGetDataKind(out DataKind itemKind);
                 Contracts.Assert(itemKind == DataKind.R4 || itemKind == DataKind.R8);
                 ctx.Writer.Write((byte)itemKind);
             }
```
TomFinley: What a very, very strange format... if there are only two choices, why not just write a bool?

Anyway, I think if we want to get rid of datakind, I think the best choice would be to write the bytes as constants... so I guess, `ctx.Writer.Write((byte)(itemType == NumberType.Double ? 10 : 9))` or something like that. I guess we'd need a similar test during the deserialization. Not sure if you're up for this in this PR...

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingEstimator.cs
```diff
@@ -60,7 +60,7 @@ public SchemaShape GetOutputSchema(SchemaShape inputSchema)
                 if (!inputSchema.TryFindColumn(colInfo.Input, out var col))
                     throw _host.ExceptSchemaMismatch(nameof(inputSchema), "input", colInfo.Input);
 
-                if ((col.ItemType.GetItemType().RawKind == default) || !(col.ItemType is VectorType || col.ItemType is PrimitiveType))
+                if (!col.ItemType.GetItemType().RawType.IsValidDataKindType() || !(col.ItemType is VectorType || col.ItemType is PrimitiveType))
```
TomFinley: >if (!col.ItemType.GetItemType().RawType.IsValidDataKindType() || !(col.ItemType is VectorType || col.ItemType is PrimitiveType)) [](start = 16, length = 128)

Oh boy. More similar problems... I'll make a note of it #2185.

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs
```diff
@@ -261,12 +261,12 @@ private static (string input, string output)[] GetColumnPairs(ColumnInfo[] colum
...
         }
 
-        internal string TestIsKnownDataKind(ColumnType type)
+        private string TestIsKnownDataKind(ColumnType type)
         {
             VectorType vectorType = type as VectorType;
             ColumnType itemType = vectorType?.ItemType ?? type;
 
-            if (itemType.RawKind != default && (vectorType != null || type is PrimitiveType))
+            if (itemType.RawType.IsValidDataKindType() && (vectorType != null || type is PrimitiveType))
```
TomFinley: > && (vectorType != null || type is PrimitiveType) [](start = 54, length = 49)

I could be wrong, but it seems like many of these `IsValidDataKindType` are actually trying to capture the logic of what you've already written in your `IsStandardScalar` extension method on `ColumnType`. That is, "does it have one of the known datakinds," which is functionally the same as "is it one of the known builtin types."

(Also, incidentally I do not believe, if this first test succeeds, that it is possible for the statement after the `&&` to be false.)

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -409,102 +409,91 @@ private ValueGetter<uint> ComposeGetterOne(Row input, int iinfo, int srcCol, Col
...
-                        Host.Assert(srcType.RawKind == DataKind.U8);
-                        return MakeScalarHashGetter<ulong, HashKey8>(input, srcCol, seed, mask);
-                }
+                if (srcType.RawType == typeof(byte))
+                    return MakeScalarHashGetter<byte, HashKey1>(input, srcCol, seed, mask);
+                else if (srcType.RawType == typeof(ushort))
+                    return MakeScalarHashGetter<ushort, HashKey2>(input, srcCol, seed, mask);
+                else if (srcType.RawType == typeof(uint))
+                    return MakeScalarHashGetter<uint, HashKey4>(input, srcCol, seed, mask);
+
```
TomFinley: By a similar token, I might want to check for `uint` first, since that is going to be the key type 99% of the time. :)

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
```diff
@@ -183,7 +183,7 @@ private static bool CheckScoreColumnKind(Schema schema, int col)
 
             // Get the score column set id from colScore.
             var type = schema[colScore].Metadata.Schema.GetColumnOrNull(MetadataUtils.Kinds.ScoreColumnSetId)?.Type;
-            if (type == null || !(type is KeyType) || type.RawKind != DataKind.U4)
+            if (type == null || !(type is KeyType) || type.RawType != typeof(uint))
```
TomFinley: >type == null || [](start = 16, length = 15)

Similar note, I think we can do without the null check given that the `is` test will catch the null case anyway, I think?

src/Microsoft.ML.Data/DataView/InternalSchemaDefinition.cs
```diff
@@ -159,50 +161,33 @@ public static void GetVectorAndKind(MemberInfo memberInfo, out bool isVector, ou
...
             {
-                rawItemType = rawType;
+                itemType = rawType;
                 isVector = false;
             }
 
-            // Get the data kind, and the item's column type.
-            if (rawItemType == typeof(string))
-                kind = DataKind.Text;
-            else if (!rawItemType.TryGetDataKind(out kind))
```
TomFinley: This was actually important -- we do not *want* `string` associated with the type `TextType`, because `TextType.Instance.RawType` is not `string`, it is `ReadOnlyMemory<char>`. We need to handle that type mapping "specially" just as we specially map arrays to `VBuffer` in other code here, and so on, and so on. It is not the job of `ColumnType` to be aware of the details of how in this convenience API we perform mappings from user types into `IDataView` land. So if that is why I see those tests against `string` that I commented on earlier as being odd, let's not do that.

src/Microsoft.ML.Core/Data/MetadataUtils.cs
```diff
@@ -263,7 +263,7 @@ internal static IEnumerable<int> GetColumnSet(this Schema schema, string metadat
             for (int col = 0; col < schema.Count; col++)
             {
                 var columnType = schema[col].Metadata.Schema.GetColumnOrNull(metadataKind)?.Type;
-                if (columnType != null && columnType is KeyType && columnType.RawKind == DataKind.U4)
+                if (columnType != null && columnType is KeyType && columnType.RawType == typeof(uint))
```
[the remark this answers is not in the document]
eerhardt: Fixed.

---
In reply to: [248938772](https://github.com/dotnet/machinelearning/pull/2176#discussion_r248938772) [](ancestors = 248938772)

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
eerhardt: Fixed.

---
In reply to: [248939217](https://github.com/dotnet/machinelearning/pull/2176#discussion_r248939217) [](ancestors = 248939217)

src/Microsoft.ML.Data/Transforms/Hashing.cs
eerhardt: > The most common case for hashing is text, so we should probably check that first, 

Good call. Fixing.

> Also probably this would probably be a good time to to checks against the ColumnType instances themselves

In my mind, there are 2 reasons why I like comparing the types here, and not against the ColumnType instances.

1. It is nice and easy to match up the `if (.... == typeof(byte))` with the next line `MakeScalarHashGetter<byte,....`. You can match the `byte` and `byte` and easily know you are using the same type.
2. It is reflective of the code above for `KeyType`, where you can't actually check for `NumberType.U4`, since it is a `KeyType`.

I don't feel strongly about these reasons, but they do seem to be nice. If you'd like it the other way, I can change it. Just let me know if you feel strongly for it to be changed.

---
In reply to: [248925374](https://github.com/dotnet/machinelearning/pull/2176#discussion_r248925374) [](ancestors = 248925374)

src/Microsoft.ML.Data/Transforms/Hashing.cs
eerhardt: Fixed.

---
In reply to: [248937303](https://github.com/dotnet/machinelearning/pull/2176#discussion_r248937303) [](ancestors = 248937303)

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs
eerhardt: > many of these IsValidDataKindType are actually trying to capture the logic of what you've already written in your IsStandardScalar extension method on ColumnType

That's a good call. I forgot about that method. I've removed `IsValidDataKindType` and am using `IsStandardScalar` instead.

>I do not believe, if this first test succeeds, that it is not possible for the statement after the && to be false.

Agreed. Removed the 2nd expression.

---
In reply to: [248936996](https://github.com/dotnet/machinelearning/pull/2176#discussion_r248936996) [](ancestors = 248936996)

src/Microsoft.ML.Core/Data/DataKind.cs
```diff
@@ -226,6 +226,41 @@ public static bool TryGetDataKind(this Type type, out DataKind kind)
...
+                || type == typeof(byte)
+                || type == typeof(short)
+                || type == typeof(ushort)
+                || type == typeof(int)
+                || type == typeof(uint)
+                || type == typeof(long)
+                || type == typeof(ulong)
+                || type == typeof(float)
+                || type == typeof(double)
+                || type == typeof(ReadOnlyMemory<char>) || type == typeof(string)
```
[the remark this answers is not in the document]
eerhardt: I've deleted this method.

---
In reply to: [248938496](https://github.com/dotnet/machinelearning/pull/2176#discussion_r248938496) [](ancestors = 248938496)

src/Microsoft.ML.Core/Data/ColumnType.cs
```diff
@@ -120,6 +120,22 @@ internal static PrimitiveType FromKind(DataKind kind)
                 return DateTimeOffsetType.Instance;
             return NumberType.FromKind(kind);
         }
+
+        [BestFriend]
+        internal static PrimitiveType FromType(Type type)
+        {
+            if (type == typeof(ReadOnlyMemory<char>) || type == typeof(string))
```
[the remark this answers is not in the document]
eerhardt: I've removed `|| type == typeof(string)` from this line. All the callers are guaranteed to do the `string` => `ROM<char>` type conversion before making this call.

---
In reply to: [248938623](https://github.com/dotnet/machinelearning/pull/2176#discussion_r248938623) [](ancestors = 248938623)

src/Microsoft.ML.Data/DataView/InternalSchemaDefinition.cs
eerhardt: Good call. I've added this back, by doing the `string` => `ROM<char>` type conversion here.

---
In reply to: [248940771](https://github.com/dotnet/machinelearning/pull/2176#discussion_r248940771) [](ancestors = 248940771)

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs
eerhardt: Actually this broke a test. There is a test that passes in a `Vec<Key<U4>>` into this method. Thus, `RawType.IsValidDataKindType` and `IsStandardScalar` are not equivalent because the first allows for KeyTypes.

---
In reply to: [249152590](https://github.com/dotnet/machinelearning/pull/2176#discussion_r249152590) [](ancestors = 249152590,248936996)

src/Microsoft.ML.Data/DataLoadSave/Binary/Codecs.cs
```diff
@@ -154,27 +154,6 @@ private sealed class UnsafeTypeCodec<T> : SimpleCodec<T> where T : struct
...
-                        case DataKind.I4:
-                            return typeof(int).Name;
-                        case DataKind.I8:
-                            return typeof(long).Name;
-                        case DataKind.TS:
-                            return typeof(TimeSpan).Name;
-                    }
-                    return base.LoadName;
-                }
-            }
```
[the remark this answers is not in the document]
sfilipi: cc @yaeldekel in case it breaks something in TLC

---
In reply to: [249104896](https://github.com/dotnet/machinelearning/pull/2176#discussion_r249104896) [](ancestors = 249104896)

src/Microsoft.ML.Data/Model/Pfa/PfaUtils.cs
```diff
@@ -180,36 +180,45 @@ private static JToken PfaTypeOrNullCore(ColumnType itemType)
...
-                        return Int;
-                    case DataKind.U4:
-                    case DataKind.I8:
-                    case DataKind.U8:
-                        return Long;
-                    case DataKind.R4:
-                    // REVIEW: This should really be float. But, for the
+                    return Int;
+                }
+                else if (rawType == typeof(uint)
```
sfilipi: >uint [](start = 43, length = 4)

not your change, but i wonder if uint belongs here, and not with the clause above, where typeof(int) is included. 
AFAIK int and uint have the same length. Are we going by numeric ranges?

src/Microsoft.ML.Data/Model/Pfa/PfaUtils.cs
```diff
@@ -180,36 +180,45 @@ private static JToken PfaTypeOrNullCore(ColumnType itemType)
...
+                else if (rawType == typeof(System.ReadOnlyMemory<char>)
+                    || rawType == typeof(string))
+                {
+                    return String;
                 }
+
+                return null;
             }
 
             public static JToken DefaultTokenOrNull(PrimitiveType itemType)
```
sfilipi: >DefaultTokenOrNull [](start = 33, length = 18)

can most of this be substituted with default(itemType)

src/Microsoft.ML.Data/Model/Pfa/PfaUtils.cs
eerhardt: I assume if you are using `uint`, you have the possibility of having values that are greater than the signed `int.MaxValue`. Thus, if we did use `Int`, it would overflow to a negative value, which I'm sure wouldn't produce the correct results.

---
In reply to: [249968334](https://github.com/dotnet/machinelearning/pull/2176#discussion_r249968334) [](ancestors = 249968334)

src/Microsoft.ML.Data/Model/Pfa/PfaUtils.cs
eerhardt: No. We could do that if this method was a generic method - `DefaultTokenOrNull<T>`. Then you could just return `default(T)`. But since the `System.Type` we are inspecting here isn't known at compile time, you can't use the C# `default` keyword.

---
In reply to: [249968942](https://github.com/dotnet/machinelearning/pull/2176#discussion_r249968942) [](ancestors = 249968942)

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs
TomFinley: Huh interesting. Anyway I like your solution to the problem I see below.

---
In reply to: [249206559](https://github.com/dotnet/machinelearning/pull/2176#discussion_r249206559) [](ancestors = 249206559,249152590,248936996)

src/Microsoft.ML.Data/Transforms/Hashing.cs
TomFinley: Fair enough, I think your reasoning works.

---
In reply to: [249119279](https://github.com/dotnet/machinelearning/pull/2176#discussion_r249119279) [](ancestors = 249119279,248925374)

PR #2188, opened 2019-01-18
Test GAM Public APIs

### LINKED ISSUE #2186: GAMs are missing tests on some methods
GAMs have end-to-end tests, but certain public APIs are not tested:
- Creating a model from new parameters
- GetFeatureWeights()
- GetBinUpperBounds()

### LINKED ISSUE #2187: GAMs only allow getting one feature at a time
As pointed out by @artidoro  on #2142, it is not possible to get the entire set of shape functions and bins in one go. Rather, we have ` GetFeatureWeights(int featureIndex)`. We need a `GetFeatureWeights()` that fetches the entire array.

rogancarr:
This PR adds new APIs to GAMs to retrieve all the shape function bins and effects (in addition to the current APIs to get only one shape at a time). This PR also adds tests to cover all the public APIs, including expected behavior for the public constructor.

Fixes #2186 
Fixes #2187


docs/samples/Microsoft.ML.Samples/Dynamic/GeneralizedAdditiveModels.cs
```diff
@@ -78,9 +78,9 @@ public static void RunExample()
             var studentTeacherRatioIndex = featureNames.ToList().FindIndex(str => str.Equals("TeacherRatio"));
 
             // Next, let's get the array of bin upper bounds from the model for this feature
```
glebuk: >bin upper bounds from the  [](start = 44, length = 26)

Please clarify meaning of word "bin" here.  Unclear if this is short for Binary or histogram bins or something else

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -300,12 +300,25 @@ public double[] GetFeatureBinUpperBounds(int featureIndex)
             return featureBins;
         }
 
+        /// <summary>
+        /// Get all the bin upper bounds.
+        /// </summary>
+        public double[][] GetBinUpperBounds()
```
glebuk: >GetBinUpperBounds [](start = 26, length = 17)

If this is a perf-critical method, then it would be quite slow given all the array alocations

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -300,12 +300,25 @@ public double[] GetFeatureBinUpperBounds(int featureIndex)
...
+        {
+            double[][] featureBins = new double[_binUpperBounds.Length][];
```
glebuk: >_binUpperBounds.Length [](start = 48, length = 22)

Why ctor uses _binUpperBounds.Length and loop featureBins.Length?  If they disagree this would crash. If they agree then always use the same...

glebuk:
>             }

You can simplify the code by inversing the "if" condition:

if (!_inputFeatureToShapeFunctionMap.TryGetValue(featureIndex, out int j))
     return new double[0];

featureBins = new double[_binUpperBounds[j].Length];
_binUpperBounds[j].CopyTo(featureBins, 0);



---
Refers to: src/Microsoft.ML.FastTree/GamModelParameters.cs:298 in fe4395a. [](commit_id = fe4395a40e0265c8e520aa74ae9f6b8d1908b5af, deletion_comment = False)

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -322,6 +335,19 @@ public double[] GetFeatureWeights(int featureIndex)
             return featureWeights;
         }
 
+        /// <summary>
+        /// Get all the binned weights.
```
glebuk: >weights [](start = 31, length = 7)

effects vs weights? Why naming in method vs comment? Try to harmonize.

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -300,12 +300,25 @@ public double[] GetFeatureBinUpperBounds(int featureIndex)
...
+            double[][] featureBins = new double[_binUpperBounds.Length][];
+            for (int i = 0; i < featureBins.Length; i++)
+            {
+                featureBins[i] = GetBinUpperBounds(i);
+            }
+            return featureBins;
+        }
+
         /// <summary>
         /// Get the binned weights for each feature.
```
glebuk: >weights [](start = 27, length = 7)

weights vs effects

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -322,6 +335,19 @@ public double[] GetFeatureWeights(int featureIndex)
             return featureWeights;
         }
 
+        /// <summary>
+        /// Get all the binned weights.
+        /// </summary>
+        public double[][] GetBinEffects()
+        {
+            double[][] featureWeights = new double[_binEffects.Length][];
```
glebuk: >[_binEffects.Length [](start = 50, length = 19)

Same comment as above - use same var for length in both places - ctor and the loop.  Suggest you store the length in a variable and use it throughout - that avoids calling Length property on every iteration.  This should apply to all similar methods in this file.

src/Microsoft.ML.FastTree/GamModelParameters.cs
```diff
@@ -322,6 +335,19 @@ public double[] GetFeatureWeights(int featureIndex)
...
+            for (int i = 0; i < featureWeights.Length; i++)
+            {
+                featureWeights[i] = GetBinEffects(i);
```
glebuk: >featureWeights [](start = 16, length = 14)

Would you be much better off by doing the _binEffects.CopyTo instead of calling the overload?

test/Microsoft.ML.Predictor.Tests/TestGamPublicInterfaces.cs
```diff
@@ -0,0 +1,74 @@
...
+namespace Microsoft.ML.RunTests
+{
+    public class TestGamPublicInterfaces
+    {
+        public TestGamPublicInterfaces()
+        {
+        }
+
+        [Fact]
+        [TestCategory("FastTree")]
```
glebuk: >FastTree [](start = 23, length = 8)

should there be a gam category?

test/Microsoft.ML.Predictor.Tests/TestGamPublicInterfaces.cs
```diff
@@ -0,0 +1,74 @@
...
+        public void TestGamDirectInstatiation()
+        {
+            var mlContext = new MLContext(seed: 1);
+
+            double intercept = 1;
+
+            var binUpperBounds = new double[2][];
+            binUpperBounds[0] = new double[] { 1, 2, 3 };
+            binUpperBounds[1] = new double[] { 4, 5, 6 };
```
glebuk: can rewrite initializer as:
var binUpperBounds = new double[][]
{
    new int[] {  1, 2, 3 },
    new int[] { 4, 5, 6 },
};

test/Microsoft.ML.Predictor.Tests/TestGamPublicInterfaces.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            var binEffects = new double[2][];
+            binEffects[0] = new double[] { 0, 1, 2 };
+            binEffects[1] = new double[] { 2, 1, 0 };
+
+            var gam = new RegressionGamModelParameters(mlContext, binUpperBounds, binEffects, intercept);
```
glebuk: >RegressionGamModelParameters [](start = 26, length = 28)

Add at least two invalid cases:
1. When the count of upper and lower bounds are different
2. When uppper bound is < lower bound.

docs/samples/Microsoft.ML.Samples/Dynamic/GeneralizedAdditiveModels.cs
rogancarr: Done.

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Done.

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Done.

test/Microsoft.ML.Predictor.Tests/TestGamPublicInterfaces.cs
```diff
@@ -0,0 +1,74 @@
...
+            Assert.Equal(array1.Length, array2.Length);
+            for (int i = 0; i < array1.Length; i++)
+                CheckArrayEquality(array1[i], array2[i]);
+        }
+
+        private void CheckArrayEquality(double[] array1, double[] array2)
+        {
+            Assert.Equal(array1.Length, array2.Length);
+            for (int i = 0; i < array1.Length; i++)
+                Assert.Equal(array1[i], array2[i], 6);
```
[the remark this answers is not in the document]
rogancarr: Done.

test/Microsoft.ML.Predictor.Tests/TestGamPublicInterfaces.cs
rogancarr: Great cases!

test/Microsoft.ML.Predictor.Tests/TestGamPublicInterfaces.cs
rogancarr: Done.

test/Microsoft.ML.Predictor.Tests/TestGamPublicInterfaces.cs
rogancarr: GAM uses the FastTree internals, so we keep it in `FastTree`. 🤷‍♂️

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Done.

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Done.

src/Microsoft.ML.FastTree/GamModelParameters.cs
rogancarr: Done.

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -651,6 +661,29 @@ public static bool IsSorted(int[] values)
...
+        public static bool IsSorted(double[] values)
+        {
+            if (Utils.Size(values) <= 1)
+                return true;
+
+            var prev = values[0];
+
+            for (int i = 1; i < values.Length; i++)
+            {
+                if (values[i] < prev)
```
glebuk: what happens in cases of .NAN,  .INF and other special values?  
I believe all comparisons of NAN return false.

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -651,6 +661,29 @@ public static bool IsSorted(int[] values)
             return true;
         }
 
+        /// <summary>
+        /// Checks that an input array is monotonically increasing
```
glebuk: >increasing [](start = 56, length = 10)

terminate all comments with "." .

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -651,6 +661,29 @@ public static bool IsSorted(int[] values)
...
+        /// <param name="values">An array of values</param>
+        /// <returns>True if the array is monotonically increasing; false otherwise.</returns>
+        public static bool IsSorted(double[] values)
+        {
+            if (Utils.Size(values) <= 1)
+                return true;
+
+            var prev = values[0];
+
+            for (int i = 1; i < values.Length; i++)
```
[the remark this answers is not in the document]
rogancarr: Fixed multiple accesses of `values[i]`.

Not sure about `<array>.Length`. I'm seeing that it's actually marginally slower to move the `<array>.Length` call. Looping over a `<array>.Length` accessor vs. a precalculated `var length = <array>.Length` is about 12% faster.

src/Microsoft.ML.Core/Utilities/Utils.cs
rogancarr: Special cases include: `Epsilon`, `MaxValue`, `MinValue`, `NaN`, `NegativeInfinity`, `PositiveInfinity`. Only `NaN` has ambiguous behavior. I've noted the expected behavior for this function in the docstring: Anything with a NaN is considered to not be monotonically increasing.

src/Microsoft.ML.Core/Utilities/Utils.cs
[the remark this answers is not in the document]
rogancarr: Here is the test I use:

```cs
double[] x = Enumerable.Range(0, 1000000).Select(i => (double) i).ToArray();

var watch = System.Diagnostics.Stopwatch.StartNew();

int l = 0;
var length = x.Length;
for (int i = 0; i < 1000000000; i++)
{
	//l += x.Length;
	l += length;
}

watch.Stop();

var elapsedSeconds = watch.ElapsedMilliseconds / 1000.0;
Console.WriteLine($"L={l} and took {elapsedSeconds} seconds");
```

Output:
length: `L=-1530494976 and took 1.946 seconds`
x.Length: `L=-1530494976 and took 1.669 seconds`
(The average over many runs is similar.)

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -615,37 +615,73 @@ public static string ExtractLettersAndNumbers(string value)
...
         public static bool IsSorted(IList<float> values)
         {
             if (Utils.Size(values) <= 1)
                 return true;
 
-            var prev = values[0];
-
-            for (int i = 1; i < values.Count; i++)
+            var previousValue = values[0];
+            var listLength = values.Count;
```
rogancarr: Note that this is faster than calculating at each iteration.

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -615,37 +615,73 @@ public static string ExtractLettersAndNumbers(string value)
...
+            for (int i = 1; i < listLength; i++)
             {
-                if (!(values[i] >= prev))
+                var currentValue = values[i];
+                if (currentValue < previousValue)
```
rogancarr: Switch check to match other implementations. `<` could be cheaper than `>=`, maybe.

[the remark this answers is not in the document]
rogancarr: I actually disagree about `Sort()`. I think that the presence of NaNs should result in a return of `false`. The assumption of ordered arrays in our codebase is that the numbers in the array are monotonically increasing. NaN is not a number, so it cannot be part of a monotonically increasing sequence.

And practically speaking, these methods are used as sanity checks on user-provided data, so we don't want to make them more permissive than they currently are.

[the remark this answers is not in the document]
rogancarr: Also, the original functions were set to fail out on NaNs. I do not want to change the behavior of these functions because I don't know what the downstream effects will be.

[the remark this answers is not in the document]
glebuk: Then let's rename the method to IsMonotonicallyIncreasing

---
In reply to: [250355488](https://github.com/dotnet/machinelearning/pull/2188#discussion_r250355488) [](ancestors = 250355488)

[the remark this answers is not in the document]
glebuk: Otherwise need to rename the method.

---
In reply to: [250036842](https://github.com/dotnet/machinelearning/pull/2188#discussion_r250036842) [](ancestors = 250036842)

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -534,15 +534,7 @@ public static int[] GetRandomPermutation(Random rand, int size)
...
-        public static void Shuffle<T>(Random rand, Span<T> rgv)
-        {
-            Contracts.AssertValue(rand);
-
-            for (int iv = 0; iv < rgv.Length; iv++)
-                Swap(ref rgv[iv], ref rgv[iv + rand.Next(rgv.Length - iv)]);
-        }
-
-        public static bool AreEqual(float[] arr1, float[] arr2)
+        public static bool AreEqual(Single[] arr1, Single[] arr2)
```
glebuk: >Single [](start = 36, length = 6)

why not float?

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -559,7 +551,7 @@ public static bool AreEqual(float[] arr1, float[] arr2)
             return true;
         }
 
-        public static bool AreEqual(double[] arr1, double[] arr2)
+        public static bool AreEqual(Double[] arr1, Double[] arr2)
```
glebuk: >Double [](start = 36, length = 6)

why not double? lowercase

src/Microsoft.ML.Core/Utilities/Utils.cs
rogancarr: Woah, good catch. Weird merge reversion.

src/Microsoft.ML.Core/Utilities/Utils.cs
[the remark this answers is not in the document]
rogancarr: Renamed.

src/Microsoft.ML.Core/Utilities/Utils.cs
rogancarr: Reverted reversion.

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -615,37 +615,75 @@ public static string ExtractLettersAndNumbers(string value)
             return Regex.Replace(value, "[^A-Za-z0-9]", "");
         }
 
-        public static bool IsSorted(IList<float> values)
+        /// <summary>
+        /// Checks that an input IList is monotonically increasing.
+        /// </summary>
+        /// <param name="values">An array of values</param>
+        /// <returns>True if the array is monotonically increasing; false otherwise.
```
Ivanidzo4ka: >if the array is monotonically increasing [](start = 26, length = 40)

maybe:
`if each element is greater or equal compare to previous elements` for simple minded people?

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -615,37 +615,75 @@ public static string ExtractLettersAndNumbers(string value)
...
+        /// ILists containing NaN values are considered to be not monotonically increasing.</returns>
+        public static bool IsMonotonicallyIncreasing(IList<float> values)
```
Ivanidzo4ka: >IsMonotonicallyIncreasing [](start = 27, length = 25)

```
        public static bool IsMonotonicallyIncreasing<T>(IList<T> values) where T : IComparable
        {
            if (values.Count <= 1)
                return true;

            var previousValue = values[0];
            var listLength = values.Count;
            for (int i = 1; i < listLength; i++)
            {
                var currentValue = values[i];
                // Inverted check for NaNs
                if (currentValue.CompareTo(previousValue) < 0)
                    return false;

                previousValue = currentValue;
            }

            return true;
        }
```
Hope I get CompareTo condition right.
That way you don't have to define second function
also I'm not sure regarding first if, does empty array is actually MonotonicallyIncreasing?

src/Microsoft.ML.Core/Utilities/Utils.cs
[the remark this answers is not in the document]
Ivanidzo4ka: Just curious are we talking about Release or Debug version of your code?
In Debug quite a lot of optimization on compiler level is turned off.

---
In reply to: [250033045](https://github.com/dotnet/machinelearning/pull/2188#discussion_r250033045) [](ancestors = 250033045,250018107)

src/Microsoft.ML.FastTree/GamModelParameters.cs
Ivanidzo4ka: It sill says weights

---
In reply to: [250005684](https://github.com/dotnet/machinelearning/pull/2188#discussion_r250005684) [](ancestors = 250005684)

test/Microsoft.ML.Predictor.Tests/TestGamPublicInterfaces.cs
```diff
@@ -0,0 +1,102 @@
...
+using Microsoft.ML.Trainers.FastTree;
+using Xunit;
+
+namespace Microsoft.ML.RunTests
+{
+    public class TestGamPublicInterfaces
+    {
+        public TestGamPublicInterfaces()
+        {
+        }
```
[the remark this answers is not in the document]
rogancarr: Not at all &ndash; removed

src/Microsoft.ML.Core/Utilities/Utils.cs
rogancarr: Good call. I added your language to the xml docs :)

src/Microsoft.ML.Core/Utilities/Utils.cs
Ivanidzo4ka: regarding 1.
Array is already IList, you don't need to have two implementations.
https://docs.microsoft.com/en-us/dotnet/api/system.array?view=netframework-4.7.2

2. unfortunately, you are right.
But you still can have only IList implementation!

rogancarr: You are totally right.

And for the original question, as to whether an empty array is monotonically increasing, I'm keeping the original implementation so we don't cause any downstream effects. I guess you could say that it's not not monotonically increasing....

rogancarr: Updating s.t. all contracts specify `IList` as that is a broader level of support, as you point out.

PR #2192, opened 2019-01-18
[Tiny Change] Internal-best-friend CacheDataView and IRowSeekable

### LINKED ISSUE #2028: Internalization of CacheDataView and IRowSeekable
One thing we may want to consider is that, while having caching is nice and necessary, and we should expose some ability to *perform* caching if we want (as we do with the, exposing `AppendCacheCheckpoint`, as seen here:

https://github.com/dotnet/machinelearning/blob/5c00980abeae594161d0b9ee748d7d0b959962b5/src/Microsoft.ML.Data/DataLoadSave/EstimatorChain.cs#L99

This itself *uses* `CacheDataView`, but does not expose it. I am somewhat uncomfortable exposing the cache data view as part of our public API in a way that makes it a first class citizen of the API, at least, in v1.0 of the API. Its particular interface is a bit odd (possibly for good reasons, but odd nonetheless), and most of the reasons for its oddness have to do with how components might want to behave "specially" with it, and component authoring scenarios are not a target for v1.0. So hiding it, and `IRowSeekable` (which is a useful but, again, somewhat odd interface), might be good.

wschin:
Fixes #2028.

### DIFF
--- a/src/Microsoft.ML.Data/Properties/AssemblyInfo.cs
+++ b/src/Microsoft.ML.Data/Properties/AssemblyInfo.cs
@@ -13,6 +13,7 @@ using Microsoft.ML;
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Predictor.Tests" + PublicKey.TestValue)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.TimeSeries.Tests" + PublicKey.TestValue)]
 
+[assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Benchmarks" + PublicKey.TestValue)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.EntryPoints" + PublicKey.Value)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Maml" + PublicKey.Value)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.ResultProcessor" + PublicKey.Value)]
--- a/src/Microsoft.ML.Data/Data/IRowSeekable.cs
+++ b/src/Microsoft.ML.Data/Data/IRowSeekable.cs
@@ -11,7 +11,8 @@ namespace Microsoft.ML.Data
     /// <summary>
     /// Represents a data view that supports random access to a specific row.
     /// </summary>
-    public interface IRowSeekable
+    [BestFriend]
+    internal interface IRowSeekable
     {
         RowSeeker GetSeeker(Func<int, bool> predicate);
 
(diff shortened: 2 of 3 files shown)


TomFinley:
>     public abstract class RowSeeker : Row

Hmmm, this seems to be in part of the same vein. Can it be internal as well?

---
Refers to: src/Microsoft.ML.Data/Data/IRowSeekable.cs:27 in d67cb51. [](commit_id = d67cb515cf8919a9bd73c056373141bf753f971a, deletion_comment = False)

src/Microsoft.ML.Data/Properties/AssemblyInfo.cs
```diff
@@ -13,6 +13,7 @@
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Predictor.Tests" + PublicKey.TestValue)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.TimeSeries.Tests" + PublicKey.TestValue)]
 
+[assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Benchmarks" + PublicKey.TestValue)]
```
TomFinley: >Benchmarks [](start = 58, length = 10)

Out of curiosity, how are the benchmarks using this? Direct testing of seekable, other...?

TomFinley:
>     public abstract class RowSeeker : Row

Looks pretty good @wschin, thanks for doing this -- I might want this guy internalized too while we're at it, since it seems to be something relating to the implementation of `IRowSeekable`, and so long as that is internal there is no point in this being external.

---
In reply to: [455735389](https://github.com/dotnet/machinelearning/pull/2192#issuecomment-455735389) [](ancestors = 455735389)

---
Refers to: src/Microsoft.ML.Data/Data/IRowSeekable.cs:27 in d67cb51. [](commit_id = d67cb515cf8919a9bd73c056373141bf753f971a, deletion_comment = False)

src/Microsoft.ML.Data/Properties/AssemblyInfo.cs
wschin: Yes a benchmark test `CacheDataViewBench`; specifically,
```csharp
            _col = _cacheDataView.Schema.GetColumnOrNull("A").Value;
            _seeker = ((IRowSeekable)_cacheDataView).GetSeeker(colIndex => colIndex == _col.Index);
            _seekerGetter = _seeker.GetGetter<int>(_col.Index);
```

---
In reply to: [249226867](https://github.com/dotnet/machinelearning/pull/2192#discussion_r249226867) [](ancestors = 249226867)

wschin:
>     public abstract class RowSeeker : Row

`RowSeeker` is our friend now.

---
In reply to: [455735543](https://github.com/dotnet/machinelearning/pull/2192#issuecomment-455735543) [](ancestors = 455735543,455735389)

---
Refers to: src/Microsoft.ML.Data/Data/IRowSeekable.cs:27 in d67cb51. [](commit_id = d67cb515cf8919a9bd73c056373141bf753f971a, deletion_comment = False)

src/Microsoft.ML.Data/Properties/AssemblyInfo.cs
TomFinley: Ah. Yes that's fair, we would want to test this directly.

---
In reply to: [249872817](https://github.com/dotnet/machinelearning/pull/2192#discussion_r249872817) [](ancestors = 249872817,249226867)

PR #2194, opened 2019-01-20
Automate code coverage report as part of PRs.

### LINKED ISSUE #2193: Code Coverage
Integrate coverlet to get coverage files and upload them to codecov.io. Enable code coverage for Windows x64 Debug and later enable for all debug legs when covertlet releases a more performant version. Display code coverage report for every commit in every PR.

codemzs:
fixes #2193.

After this change, every PR will get a code coverage report against master for every commit. It will inform the author the impact of their change on the code coverage. This should hopefully lead to more informed, faster and higher quality code review, in addition, it will hopefully lead to improved code quality.

Currently coverage files are generated using coverlet tool and uploaded to codecov.io. For now coverage reports are generated using Windows x64 Debug build but later it will be enabled for all debug legs when coverlet releases a more performant version.


codemzs:
@sharwell Done.

test/Directory.Build.props
```diff
@@ -31,6 +31,20 @@
     <PackageReference Include="xunit" Version="2.4.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
     <PackageReference Include="Microsoft.DotNet.XUnitExtensions" Version="2.4.0-prerelease-63213-02" />
+    <PackageReference Include="coverlet.msbuild" Version="$(CoverletVersion)" />
+    <PackageReference Include="Codecov" Version="$(CodecovVersion)" />
```
ViktorHofer: You don't want this PackageReference here as that means that every test assemblies references Codecov which probably doesn't even make sense. Make sure that you restore that tool as part of the repo bootstrapping.

build/Codecov.proj
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <Import Project="..\Directory.Build.props" />
+  
+  <Target Name="Codecov">
+    <PropertyGroup Condition="'$(UseCodecov)' == 'true'">
```
ViktorHofer: You don't need the `UseCodecov` conditions here. If you want to have a condition, put it on the target.

codemzs: Makes sense.

test/Directory.Build.props
codemzs: @safern Does adding reference to codecov in codecov.proj will load the tool before the execution?

safern: I think you would just need to add it into: https://github.com/dotnet/machinelearning/blob/master/init-tools.msbuild#L9 since that is the project that is used to bootstrap external tools for the repo. 

Or you could just make sure that your Codecov target depends on Restore: https://github.com/dotnet/machinelearning/pull/2194/files#diff-180e04fde140ade031905faef1939141R5

And add the package references to that project itself.

codemzs: Thanks!

test/Directory.Build.props
```diff
@@ -31,6 +31,19 @@
     <PackageReference Include="xunit" Version="2.4.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
     <PackageReference Include="Microsoft.DotNet.XUnitExtensions" Version="2.4.0-prerelease-63213-02" />
+    <PackageReference Include="coverlet.msbuild" Version="$(CoverletVersion)" />
```
safern: @ViktorHofer is this dependency OK in all test projects?

build/ci/phase-template.yml
```diff
@@ -34,9 +34,18 @@ phases:
       displayName: Download Benchmark Data
     - ${{ if eq(parameters.name, 'MacOS') }}:
       - script: brew update && brew install libomp mono-libgdiplus gettext && brew link gettext --force
-        displayName: Install runtime dependencies
+        displayName: Install runtime dependencies     
+    - script: $(_buildScript) -$(_configuration) -runtests -coverage=true
+      displayName: Run Tests with code coverage.
+      condition: and(eq(variables._phaseName, 'Windows_x64'), eq(variables._configuration, 'Debug'))
```
safern: Do we really want to run code coverage in all PRs in the required build? Wouldn't be better to have an extra PR leg?

test/Directory.Build.props
ViktorHofer: jep. the msbuild package is used for that reason. It adds targets that can be used on a project level: https://github.com/tonerdo/coverlet/blob/master/src/coverlet.msbuild/coverlet.msbuild.targets

build/ci/phase-template.yml
codemzs: Myself and @sharwell decided against it after making pref improvements to code coverage. By creating an extra leg you are also wasting resources. Code coverage just runs on one leg of the required build and it does not impact the merge, i.e if you regress code coverage it won't block.

safern: Sounds good! Btw, you could simplify that condition if you declared a variable in your matrix, but it is just a small not important thing I won't block the PR on.

config.json
```diff
@@ -78,6 +78,12 @@
       "values": [],
       "defaultValue": ""
     },
+    "Coverage": {
```
safern: Is this needed at all? I don't see you using it at all.

codemzs: I'm using it to turn on coverage from build.cmd

safern: Oh yeah missed that parameter since you where just calling dotnet msbuild directly to do the coverage report.

codecov.yml
```diff
@@ -0,0 +1,20 @@
+# https://docs.codecov.io/docs/codecov-yaml
```
eerhardt: Is there a way to put this file somewhere other than the root of the repo?  Either in the `test` or `build` folders?  Our root directory is too cluttered the way it is.

build/ci/phase-template.yml
eerhardt: I don't think this should be the condition. Instead, we should have a parameter specifically for "analyze code coverage". and then this is based on that parameter. That way we can set the parameter on any leg(s) we'd like.

build/ci/phase-template.yml
```diff
@@ -34,9 +34,18 @@ phases:
...
     - script: $(_buildScript) -$(_configuration) -runtests
-      displayName: Run Tests
+      displayName: Run Tests without code coverage.
+      condition: not(and(eq(variables._phaseName, 'Windows_x64'), eq(variables._configuration, 'Debug')))
+    - script: dotnet msbuild build/Codecov.proj /t:Codecov /p:UseCodecov=true
```
eerhardt: This shouldn't be `dotnet msbuild` - which will use any old `dotnet` on the $PATH. Instead we should be using the SDK that we downloaded to the `Tools` directory. That way no matter what the build machine has installed on it, we always use a consistent SDK.

build/ci/phase-template.yml
```diff
@@ -34,9 +34,18 @@ phases:
       displayName: Download Benchmark Data
     - ${{ if eq(parameters.name, 'MacOS') }}:
       - script: brew update && brew install libomp mono-libgdiplus gettext && brew link gettext --force
-        displayName: Install runtime dependencies
+        displayName: Install runtime dependencies
```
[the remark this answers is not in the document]
sharwell:
```suggestion
        displayName: Install runtime dependencies
```

build/ci/phase-template.yml
eerhardt: I see that as part of your PR, if the `upload` step fails, it blocks my PR from merging. This is not intended, is it? If the codecov service fails for some reason, my PR is going to be blocked from merging?

sharwell: Failure to upload will not cause a failure with this configuration. See the following option:

https://github.com/codecov/codecov-bash/blob/4678d212cce2078bbaaf5027af0c0dafaad6a095/codecov#L138

codecov.yml
codemzs: Unfortunately this is not possible at this time.

sharwell: @eerhardt This location is the standard convention. Deviating from it would only serve to confuse users familiar with the tools.

eerhardt: It's shocking how many tools assume that they are important enough to force people to put files in the root of a repository.

build/Codecov.proj
```diff
@@ -0,0 +1,33 @@
...
+      <_CodecovArgs Include="-f;@(_CoverageReports)" />
+      <_CodecovArgs Include="-r;$(BUILD_REPOSITORY_NAME)" Condition="'$(BUILD_REPOSITORY_NAME)' != ''" />
+      <_CodecovArgs Include="--pr;$(SYSTEM_PULLREQUEST_PULLREQUESTNUMBER)" Condition="'$(SYSTEM_PULLREQUEST_PULLREQUESTNUMBER)' != ''" />
+      <_CodecovArgs Include="-b;$(BUILD_BUILDNUMBER)" Condition="'$(BUILD_BUILDNUMBER)' != ''" />
+      <_CodecovArgs Include="--branch;$(_BranchName)" Condition="'$(_BranchName)' != ''" />
+      <_CodecovArgs Include="-c;$(BUILD_SOURCEVERSION)" Condition="'$(BUILD_SOURCEVERSION)' != ''" />
+      <_CodecovArgs Include="-n;$(BUILD_DEFINITIONNAME)" Condition="'$(BUILD_DEFINITIONNAME)' != ''" />
+      <_CodecovArgs Include="--flag;$(Configuration)" Condition="'$(Configuration)' != ''" />
+      <_CodecovArgs Include="-t;$(CODECOV_TOKEN)" Condition="'$(CODECOV_TOKEN)' != ''" />
+      <_CodecovArgs Include="-Z"/>
```
sharwell: ❗️ This will cause a failing code coverage upload to result in a build failure. My understanding of previous comments from other reviewers is this is not desired.

```suggestion
```

codemzs: Yikes. Reverting it now.

build/Codecoverage.proj
```diff
@@ -0,0 +1,38 @@
...
+      <_CodecovArgs Include="-b;$(BUILD_BUILDNUMBER)" Condition="'$(BUILD_BUILDNUMBER)' != ''" />
+      <_CodecovArgs Include="--branch;$(_BranchName)" Condition="'$(_BranchName)' != ''" />
+      <_CodecovArgs Include="-c;$(BUILD_SOURCEVERSION)" Condition="'$(BUILD_SOURCEVERSION)' != ''" />
+      <_CodecovArgs Include="-n;$(BUILD_DEFINITIONNAME)" Condition="'$(BUILD_DEFINITIONNAME)' != ''" />
+      <_CodecovArgs Include="--flag;$(Configuration)" Condition="'$(Configuration)' != ''" />
+      <_CodecovArgs Include="-t;$(CODECOV_TOKEN)" Condition="'$(CODECOV_TOKEN)' != ''" />
+    </ItemGroup>
+
+    <Message Importance="high" Text="&quot;$(_CodecovPath)&quot; @(_CodecovArgs, ' ')" />
+    <Exec Condition="'$(UseCodecov)' == 'true'"
```
eerhardt: `'$(UseCodecov)' == 'true'` will always be true here because the Task is gated on it.

build/ci/phase-template.yml
```diff
@@ -35,8 +37,17 @@ phases:
     - ${{ if eq(parameters.name, 'MacOS') }}:
       - script: brew update && brew install libomp mono-libgdiplus gettext && brew link gettext --force
         displayName: Install runtime dependencies
+    - script: $(_buildScript) -$(_configuration) -runtests -coverage=true
```
eerhardt: Why can't this be a single script?

`- script: $(_buildScript) -$(_configuration) -runtests -coverage=$(_codeCoverage)`

codemzs: Good idea.

build/codecoverage-ci.yml
```diff
@@ -0,0 +1,23 @@
+################################################################################
+# ML.NET's PR validation build
+################################################################################
+
+resources:
+  containers:
+  - container: LinuxContainer
```
eerhardt: We don't use this right? Since we are only running on Windows.

build/codecoverage-ci.yml
```diff
@@ -0,0 +1,23 @@
+################################################################################
+# ML.NET's PR validation build
```
eerhardt: Can you update this comment?

codemzs:
Thanks @eerhardt for the review. The build files look much cleaner now with your feedback!
Thanks @safern , @ViktorHofer and @sharwell for reviewing!

build/Codecoverage.proj
```diff
@@ -0,0 +1,37 @@
...
+      <_CodecovArgs Include="-n;$(BUILD_DEFINITIONNAME)" Condition="'$(BUILD_DEFINITIONNAME)' != ''" />
+      <_CodecovArgs Include="--flag;$(Configuration)" Condition="'$(Configuration)' != ''" />
+      <_CodecovArgs Include="-t;$(CodeCovToken)" Condition="'$(CodeCovToken)' != ''" />
+    </ItemGroup>
+
+    <Message Importance="high" Text="&quot;$(_CodecovPath)&quot; @(_CodecovArgs, ' ')" />
+    <Exec Command="&quot;$(_CodecovPath)&quot; @(_CodecovArgs, ' ')" />
+  </Target>
+
+</Project>
```
sharwell: 📝 Missing newline at end of file

```suggestion
</Project>
```

build/codecoverage-ci.yml
```diff
@@ -0,0 +1,17 @@
...
+    name: Windows_x64
+    buildScript: build.cmd
+    customMatrixes:
+      Build_Debug:
+        _configuration: Debug
+        _config_short: D
+        _includeBenchmarkData: false
+    codeCoverage: true
+    queue:
+      name: Hosted VS2017
```
sharwell: 📝 Missing newline at end of file

```suggestion
      name: Hosted VS2017
```

codecov.yml
```diff
@@ -0,0 +1,20 @@
...
+comment:
+  layout: "diff"
+
+flags:
+  production:
+    paths:
+      - src/
+  test:
+    paths:
+      - test/
```
sharwell: 📝 Missing newline at end of file

```suggestion
      - test/
```

PR #2195, opened 2019-01-20
Remove GetRootCursor/MoveMany/CursorState

### LINKED ISSUE #1532: IDataView Cleanup: Clear cutting the `IRowCursor` interface jungle
Right now we have a fairly elaborate set of interfaces surrounding `IRowCursor`. These include `ISchematized`, `ICounted`, `IRow`.

While `IRowCursor` and `IRow` are quite useful, it is not clear if separating out `ICursor` and `ICounted` as separate interfaces is actually helpful, and we think we ought to get rid of `ISchematized` anyway (see #1502). The goal would be to refactor this jungle so that there are, relating to row cursors, two types, both abstract ***classes***, corresponding to `IRow` and `IRowCursor`, but of course not named that since they won't be interfaces. (What would be a good name is up for debate. `RowCursor` seems probably fine, but just `Row` does not seem like a good name.)

It may be that we have the name `Row` for now until we imagine a better name, and treat that renaming as a separate issue from this issue which mostly treats on the proper refactoring of the code.

`ICursor` does have some implementations that are not also `IRowCursor`, but it is unclear whether this *has* to be this way, or if it is just that way just because it was there.

The collapse of `ICounted` into these hypothetical `Row` and `RowCursor` classes would be mostly into `Row`.

## While we are at it...

While refactoring this code it would be advantageous to at the same time deal with the refactorings of other things methods in cursors, as we will be restructuring them anyway.

* The utility of `State` on cursors has proven to not be terribly useful. People create and use cursors more or less like they do enumerators -- they create them, move-next over them, then when they stop the thing is disposed. It is valuable for the sake of 

* The `MoveMany` method *seems* attractive. In many cases, you can in fact move faster than just an exhaustive `MoveNext` to get to where you want to go. The trouble is the utility for such a mechanism has proven to be limited -- sure, a `Skip` filter could be more performant, but beyond that usages are few. We may as well remove it.

* `GetRootCursor`'s performs an essential job (bypass nested move-next so that nested cursors, which is to say *most* cursors, have a fast `MoveNext`). By moving this to an abstract class, we can see if this "bypass" mechanism can be achieved in a less obnoxious (possibly internal) way to cursors.

* Possibly not to be done here, it would be nice if instead of `IRowCursor` (or rather, its abstract class descendant) being disposable, that the `IRow` is disposable. This is desirable but may require additional considerations, and might be best achieved in a separate issue.

* The utility of the `IsActiveColumn` is not immediately clear. Can we get away with simply not providing it? It may be though that we decide to keep it.

TomFinley:
Fixes #1532, in particular the last notes about things to remove from cursors.

As noted in a reply on the issue, the one thing I gave up on `IsColumnActive` because I realized eventually that it its removal did more harm than good, and it represented a genuine lapse in capability. Some test code started to become impossible to write, and it struck me that some debugging/test scenarios simply could not be served with it, even if in our "perfectly" written code it seemed to me that it was never genuinely useful.

The commits are intended to be structured usefully, as per usual.

* For `RowCursor.GetRootCursor`, I replaced that functionality with some implementation specific logic in `SynchronizedCursorBase`, since that's the only really important place that becomes critical anyway.
* `MoveMany`. Slight inefficiencies on the very sparse samplers and the skip/take filter. That's about it.
* `CursorState`. Want to know if the cursor is in a bad state? Check whether `Position` is negative.


src/Microsoft.ML.Data/DataView/EmptyDataView.cs
```diff
@@ -59,12 +59,7 @@ public Cursor(IChannelProvider provider, Schema schema, Func<int, bool> needCol)
 
             public override ValueGetter<RowId> GetIdGetter()
             {
-                return
-                    (ref RowId val) =>
-                    {
-                        Ch.Assert(!IsGood);
-                        throw Ch.Except("Cannot call ID getter in current state");
-                    };
+                return (ref RowId val) => throw Ch.Except("Cannot call ID getter in current state");
```
Ivanidzo4ka: >"Cannot call ID getter in current state" [](start = 58, length = 40)

In DataViewConstructionUtils you change it to RowCursorUtils.FetchValueStateError, but remain message same here.

src/Microsoft.ML.Data/Transforms/SkipTakeFilter.cs
```diff
@@ -211,7 +211,8 @@ private sealed class Cursor : LinkedRowRootCursorBase
             private long _rowsTaken;
             private bool _started;
 
-            public override long Batch {
+            public override long Batch
+            {
                 // SkipTakeFilter does not support cursor sets, so the batch number can always be zero.
                 get { return 0; }
             }
```
Ivanidzo4ka: maybe change it to this:
```
            // SkipTakeFilter does not support cursor sets, so the batch number can always be zero.
            public override long Batch => 0;
```

src/Microsoft.ML.Data/Transforms/SkipTakeFilter.cs
```diff
@@ -233,40 +234,37 @@ public override ValueGetter<RowId> GetIdGetter()
...
             {
-                return MoveManyCore(1);
-            }
-
-            protected override bool MoveManyCore(long count)
-            {
-                Ch.Assert(count > 0);
-                Ch.Assert(State == CursorState.NotStarted || State == CursorState.Good);
-
                 // Exit if count + _rowsTaken will overflow.
```
Ivanidzo4ka: >count [](start = 27, length = 5)

we no longer have count.

src/Microsoft.ML.Data/DataView/EmptyDataView.cs
TomFinley: Yes, and I see the same on the getter below as well. A more thorough review of implementations of this specifically revealed a few more, thank you @Ivanidzo4ka.

src/Microsoft.ML.Data/Transforms/SkipTakeFilter.cs
TomFinley: Sure. XML commented it while I was at it...

src/Microsoft.ML.Data/Transforms/SkipTakeFilter.cs
TomFinley: Right, thank you.

Issue #2100, opened 2019-01-10
Refactoring of Constructors


abgoswam:
Currently in our codebase, we have two `constructors`  that are used for initialization of the underlying object.

**Example:**

A. 
https://github.com/dotnet/machinelearning/blob/bdc9a9e1348abfdd2287d43e2633089e9eba36c7/src/Microsoft.ML.FastTree/FastTreeRanking.cs#L75-L84

B.
https://github.com/dotnet/machinelearning/blob/bdc9a9e1348abfdd2287d43e2633089e9eba36c7/src/Microsoft.ML.FastTree/FastTreeRanking.cs#L93-L95

We need to bringing the public API surface to the desired shape.  As such, we are making both constructors `internal`, and fixing other issues with the public API as outlined in #1798. 

Additionally, constructor (B)  has all the details  for constructing the underlying object. As such, we can delete constructor (A)  altogether.

NOTE:  We will do this issue only after finishing  the work to bring public API to the desired shape .

Constructors that need closer look towards unification:

SdcaBinaryTrainer
SdcaMultiClassTrainer
SdcaRegressionTrainer
StochasticGradientDescentClassificationTrainer
FastTreeRankingTrainer
FastTreeRegressionTrainer
FastTreeBinaryClassificationTrainer
FastForestClassification
FastForestRegression
PoissonRegression
LogisticRegression
BinaryClassificationGamTrainer
LightGbmBinaryTrainer
LightGbmMulticlassTrainer
LightGbmRankingTrainer
LightGbmRegressorTrainer

@TomFinley @glebuk @shauheen @sfilipi @artidoro @rogancarr

TomFinley:
Good, so #1098 will be "the way" for the public API, I guess, for v1.

abgoswam:
Yeap. That is why as part of the work for #1798  I am making the constructors `internal`.

rosebyte:
Trainer constructors are now internal and the public API uses `MLContext` catalog methods.

PR #2199, opened 2019-01-21
Modify API for advanced settings (GAM learners)

abgoswam:
Towards #1798 .

This PR addresses the following algos 

- BinaryClassificationGamTrainer
- RegressionGamTrainer

The following changes have been made:

1.  Two `public` extension methods, one for simple arguments and the other for advanced options
2. Make constructors `internal` 
3. Pass Options objects as arguments instead of Action delegate
4. Rename Arguments to Options
5. Rename Options objects as options (instead of args or advancedSettings used so far)


src/Microsoft.ML.HalLearners/SymSgdClassificationTrainer.cs
```diff
@@ -183,12 +183,12 @@ private protected override TPredictor TrainModelCore(TrainContext context)
...
-            : base(Contracts.CheckRef(env, nameof(env)).Register(LoadNameValue), TrainerUtils.MakeR4VecFeature(args.FeatureColumn),
-                  TrainerUtils.MakeBoolScalarLabel(args.LabelColumn))
+        internal SymSgdClassificationTrainer(IHostEnvironment env, Options options)
+            : base(Contracts.CheckRef(env, nameof(env)).Register(LoadNameValue), TrainerUtils.MakeR4VecFeature(options.FeatureColumn),
+                  TrainerUtils.MakeBoolScalarLabel(options.LabelColumn))
         {
-            args.Check(Host);
-            _options = args;
+            options.Check(Host);
+            _options = options;
```
sfilipi: >  _options = options; [](start = 10, length = 21)

check

src/Microsoft.ML.FastTree/GamClassification.cs
```diff
@@ -50,8 +50,8 @@ public sealed class Arguments : ArgumentsBase
         /// <summary>
         /// Initializes a new instance of <see cref="BinaryClassificationGamTrainer"/>
         /// </summary>
-        internal BinaryClassificationGamTrainer(IHostEnvironment env, Arguments args)
-             : base(env, args, LoadNameValue, TrainerUtils.MakeBoolScalarLabel(args.LabelColumn))
+        internal BinaryClassificationGamTrainer(IHostEnvironment env, Options options)
```
rogancarr: One thing that I've noticed about this switch to `options` is that we have two APIs for each learner: One with `options`, and one with a subset of the `options` spelled out. Wouldn't it make sense to only have one constructor that took options, and have the Catalog interface handle the multiplicity of ways to construct it?

That is, only have 
````cs
new MyTrainer(env, options)
````

And then have the catalog do this:

```cs
public static BinaryClassificationGamTrainer GeneralizedAdditiveModels(this BinaryClassificationCatalog.BinaryClassificationTrainers catalog,
            string labelColumn = DefaultColumnNames.Label,
            string featureColumn = DefaultColumnNames.Features,
            string weights = null,
            int numIterations = GamDefaults.NumIterations,
            double learningRate = GamDefaults.LearningRates,
            int maxBins = GamDefaults.MaxBins)
        {
            var options = new BinaryClassificationGamTrainer.Options{
                xyz = etc
            }
            return new BinaryClassificationGamTrainer(env, options);
        }
```

abgoswam: Indeed.  That's  issue #2100

Recall, we had issues with the *constructor* unification for FastTreeRanking.

P0 is to get the public APIs  ready. 
#2100 is P1/P2  in that context.

---
In reply to: [249970882](https://github.com/dotnet/machinelearning/pull/2199#discussion_r249970882) [](ancestors = 249970882)

Issue #1156, opened 2018-10-04
Enable tests against .NET Core 3.0 SSE and software-only paths


danmosemsft:
Seems we have 4 possible code paths in cpumath:

1) sse.cpp  -- used when not on .NET Core 3.0
2) avxintrinsics.cs  -- on .NET Core 3.0 with AVX available
3) sseintrinsics.cs	 -- on .NET Core 3.0 with SSE but not AVX
4) software fallback in cpumathutils.netcoreapp.cs -- on .NET Core 3.0 with no SIMD (eg., on ARM). There seems to be no software fallback when not on .NET Core 3.0.

For tests we have
A) CpuMath.UnitTests.netcoreapp. Executes whichever of (2), (3), (4) applies at run time. Which will almost surely be (2) in all cases.
B) CpuMath.UnitTests.netstandard. Same tests as above, but against (1).
C) Microsoft.ML.CpuMath.PerformanceTests. Perf, not functional, tests for about 20 entry points in (1), (2), and (3) explicitly.

We need to have tests for anything we ship and support, which I think means we have a gap for (3) and (4) ie on older x86 machines and on non x86 machines.

@tannergooding how commonly would the non AVX path (3) be encountered on typical customer x86 machines? Am I correct that (4) is not relevant to x86?

cc @Anipik

danmosemsft:
cc @eerhardt

eerhardt:
The way to get (3) and (4) exercised is to set environment variables. However, it currently isn't possible (from what I understand) to set environment variables during test execution time. So we will need to add some script that sets the environment variables and then runs the tests.

danmosemsft:
Would using our RemoteExecutorTestBase (with EnvironmentVariables added to RemoteInvokeOptions) do the trick?

eerhardt:
It's possible that RemoteExecutorTestBase would help here.

In a perfect world, I would still want each test reported as a single test, not all ~72 tests run as one "test". That would mean that each test would need to create its own process, which means the tests are going to take longer than they currently do.  We could take that approach to get the coverage now though.

My hope was that we could use a `testsettings` file for something like this, but it currently appears this isn't possible.

Anipik:
@eerhardt @tannergooding @danmosemsft 
```c# 
        [Theory]
        [InlineData("0", null)]
        [InlineData("1", null)]
        [InlineData("0", "COMPlus_EnableAvx")]
        [InlineData("1", "COMPlus_EnableAvx")]
[...]

PR #2200, opened 2019-01-22
Tests for Sse and Software implementations

Anipik:
Fixes https://github.com/dotnet/machinelearning/issues/1156

This pr adds tests for sse and software implementation of cpumath functions

### DIFF
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -6,6 +6,7 @@
     <TargetFramework Condition="$(Configuration.EndsWith('-netfx'))">net461</TargetFramework>
     <TargetFramework Condition="'$(UseIntrinsics)' == 'true'">netcoreapp3.0</TargetFramework>
     <RuntimeIdentifier Condition="'$(TargetFramework)' == 'net461'">win-x64</RuntimeIdentifier>
+    <DefineConstants Condition="'$(TargetFramework)' == 'netcoreapp3.0'">NETCOREAPP30</DefineConstants>
     <IsPackable>false</IsPackable>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
 
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -6,6 +6,7 @@
     <TargetFramework Condition="$(Configuration.EndsWith('-netfx'))">net461</TargetFramework>
     <TargetFramework Condition="'$(UseIntrinsics)' == 'true'">netcoreapp3.0</TargetFramework>
     <RuntimeIdentifier Condition="'$(TargetFramework)' == 'net461'">win-x64</RuntimeIdentifier>
+    <!-- This is required in order to have follwing extension files in the output folder of the projects referencing other projects. -->
     <AllowedReferenceRelatedFileExtensions>$(AllowedReferenceRelatedFileExtensions);.runtimeconfig.json;.runtimeconfig.dev.json;.deps.json</AllowedReferenceRelatedFileExtensions>  
     <IsPackable>false</IsPackable>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -6,7 +6,7 @@
     <TargetFramework Condition="$(Configuration.EndsWith('-netfx'))">net461</TargetFramework>
     <TargetFramework Condition="'$(UseIntrinsics)' == 'true'">netcoreapp3.0</TargetFramework>
     <RuntimeIdentifier Condition="'$(TargetFramework)' == 'net461'">win-x64</RuntimeIdentifier>
-    <DefineConstants Condition="'$(TargetFramework)' == 'netcoreapp3.0'">NETCOREAPP30</DefineConstants>
+    <AllowedReferenceRelatedFileExtensions>$(AllowedReferenceRelatedFileExtensions);.runtimeconfig.json;.runtimeconfig.dev.json;.deps.json</AllowedReferenceRelatedFileExtensions>  
     <IsPackable>false</IsPackable>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
 
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -6,7 +6,7 @@
     <TargetFramework Condition="$(Configuration.EndsWith('-netfx'))">net461</TargetFramework>
     <TargetFramework Condition="'$(UseIntrinsics)' == 'true'">netcoreapp3.0</TargetFramework>
     <RuntimeIdentifier Condition="'$(TargetFramework)' == 'net461'">win-x64</RuntimeIdentifier>
-    <!-- This is required in order to have follwing extension files in the output folder of the projects referencing other projects. -->
+    <!-- This is required in order to have following extension files in the output folder of the projects referencing other projects. -->
     <AllowedReferenceRelatedFileExtensions>$(AllowedReferenceRelatedFileExtensions);.runtimeconfig.json;.runtimeconfig.dev.json;.deps.json</AllowedReferenceRelatedFileExtensions>  
     <IsPackable>false</IsPackable>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
--- a/test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
+++ b/test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
@@ -5,22 +5,26 @@
 using System;
 using System.Collections.Generic;
 using Microsoft.ML.Internal.CpuMath;
+using Microsoft.ML.TestFramework;
 using Xunit;
 
 namespace Microsoft.ML.CpuMath.UnitTests
 {
     public class CpuMathUtilsUnitTests
     {
-        private readonly float[][] _testArrays;
-        private readonly int[] _testIndexArray;
-        private readonly AlignedArray[] _testMatrices;
-        private readonly AlignedArray[] _testSrcVectors;
-        private readonly AlignedArray[] _testDstVectors;
-        private readonly int _vectorAlignment = CpuMathUtils.GetVectorAlignment();
-        private readonly FloatEqualityComparer _comparer;
-        private readonly FloatEqualityComparerForMatMul _matMulComparer;
+        private static readonly float[][] _testArrays;
+        private static readonly int[] _testIndexArray;
+        private static readonly AlignedArray[] _testMatrices;
+        private static readonly AlignedArray[] _testSrcVectors;
+        private static readonly AlignedArray[] _testDstVectors;
+        private static readonly int _vectorAlignment = CpuMathUtils.GetVectorAlignment();
+        private static readonly FloatEqualityComparer _comparer;
+        private static readonly FloatEqualityComparerForMatMul _matMulComparer;
+        private static readonly string defaultMode = "defaultMode";
+        private static readonly string disableAvx = "COMPlus_EnableAvx";
+        private static readonly string disableAvxAndSse = "COMPlus_EnableHWIntrinsic";
 
-        public CpuMathUtilsUnitTests()
+        static CpuMathUtilsUnitTests()
         {
             // Padded array whose length is a multiple of 4
             float[] testArray1 = new float[16] { 1.96f, -2.38f, -9.76f, 13.84f, -106.37f, -26.93f, 32.45f, 3.29f, 1.96f, -2.38f, -9.76f, 13.84f, -106.37f, -26.93f, 32.45f, 3.29f };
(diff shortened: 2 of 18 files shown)


test/RemoteExecutorConsoleApp/RemoteExecutorConsoleApp.runtimeconfig.json
```diff
@@ -0,0 +1,9 @@
+{
```
eerhardt: I don't think this file should be checked in.

eerhardt:
> using System;

copyright

---
Refers to: test/RemoteExecutorConsoleApp/RemoteExecutorConsoleApp.cs:1 in 8be2e3c. [](commit_id = 8be2e3cb0b326b308acf2dfad1e961fb1a25d9d7, deletion_comment = False)

test/RemoteExecutorConsoleApp/RemoteExecutorConsoleApp.cs
```diff
@@ -0,0 +1,112 @@
...
+namespace RemoteExecutorConsoleApp
+{
+    class Program
+    {
+        static int Main(string[] args)
+        {
+            if (args.Length < 4)
+            {
+                Console.Error.WriteLine("Usage: {0} assemblyName typeName methodName exceptionFile [additionalArgs]", typeof(Program).GetTypeInfo().Assembly.GetName().Name);
+                Environment.Exit(-1);
```
eerhardt: Why `Environment.Exit` and returning `-1`?

test/Directory.Build.props
```diff
@@ -5,6 +5,7 @@
     <DefaultTestTargetFramework>netcoreapp2.1</DefaultTestTargetFramework>
     <DefaultTestTargetFramework Condition="$(Configuration.EndsWith('-netfx'))">net462</DefaultTestTargetFramework>
     <DefaultTestTargetFramework Condition="'$(UseIntrinsics)' == 'true'">netcoreapp3.0</DefaultTestTargetFramework>
+    <DefineConstants Condition="'$(DefaultTestTargetFramework)' == 'netcoreapp3.0'">NETCOREAPP30</DefineConstants>
```
eerhardt: This isn't necessary - these DefineConstants are defined implicitly. 

Instead of `NETCOREAPP30` => `NETCOREAPP3_0`

test/Microsoft.ML.TestFramework/RemoteExecutorTestBase.cs
```diff
@@ -0,0 +1,243 @@
...
+using Xunit.Sdk;
+
+namespace Microsoft.ML.TestFramework
+{
+    /// <summary>Base class used for all tests that need to spawn a remote process.</summary>
+    public abstract partial class RemoteExecutorTestBase
+    {
+        /// <summary>The name of the test console app.</summary>
+        protected static readonly string TestConsoleApp = Path.GetFullPath(@"RemoteExecutorConsoleApp.dll");
+#if NET461
```
eerhardt: I commented on your other PR, but with [https://github.com/dotnet/sdk/pull/2073](https://github.com/dotnet/sdk/pull/2073), this can be changed to `#if NETFRAMEWORK`

test/Microsoft.ML.TestFramework/RemoteExecutorTestBase.cs
```diff
@@ -0,0 +1,243 @@
...
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Sdk;
+
+namespace Microsoft.ML.TestFramework
+{
+    /// <summary>Base class used for all tests that need to spawn a remote process.</summary>
+    public abstract partial class RemoteExecutorTestBase
```
eerhardt: Why is this `partial`?

test/Microsoft.ML.TestFramework/RemoteExecutorTestBase.cs
```diff
@@ -0,0 +1,243 @@
...
+    {
+        /// <summary>The name of the test console app.</summary>
+        protected static readonly string TestConsoleApp = Path.GetFullPath(@"RemoteExecutorConsoleApp.dll");
+#if NET461
+        protected static readonly string HostRunnerName = "xunit.console.exe";
```
eerhardt: Who uses `HostRunnerName`?

test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
```diff
@@ -77,431 +81,661 @@ public CpuMathUtilsUnitTests()
...
-            CpuMathUtils.MatrixTimesSource(false, mat, src, dst, dst.Size);
-            float[] actual = new float[dst.Size];
-            dst.CopyTo(actual, 0, dst.Size);
-            Assert.Equal(expected, actual, _matMulComparer);
+                CpuMathUtils.MatrixTimesSource(false, mat, src, dst, dst.Size);
+                float[] actual = new float[dst.Size];
+                dst.CopyTo(actual, 0, dst.Size);
+                Assert.Equal(expected, actual, _matMulComparer);
+                return RemoteExecutorTestBase.SuccessExitCode;
+            }, mode, matTest, srcTest, dstTest, new RemoteInvokeOptions() { mode = mode }).Dispose();
```
eerhardt: Can `RemoteInvoke` just call `Dispose()`? That way every call site doesn't have to call it.

test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
```diff
@@ -77,431 +81,661 @@ public CpuMathUtilsUnitTests()
...
+                Assert.True(System.Runtime.Intrinsics.X86.Sse.IsSupported);
+            }
+            else if (mode == disableAvx)
+            {
+                Assert.False(System.Runtime.Intrinsics.X86.Avx.IsSupported);
+                Assert.True(System.Runtime.Intrinsics.X86.Sse.IsSupported);
+            }
+            else if (mode == disableAvxAndSse)
+            {
+                Assert.True(System.Runtime.Intrinsics.X86.Avx.IsSupported);
```
eerhardt: I assume this deserves a comment and a link to an issue...

test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
```diff
@@ -77,431 +81,661 @@ public CpuMathUtilsUnitTests()
             _testDstVectors = new AlignedArray[] { testDstVectorAligned1, testDstVectorAligned2 };
         }
 
+        private static void CheckProperFlag(string mode)
+        {
+#if NETCOREAPP30
```
eerhardt: So if we are running tests on non-.NET Core 3.0 (either full fx or 2.1), does this mean we are running the same test 3 times in a row? Is it worth only running it one time in those environments?

test/Microsoft.ML.TestFramework/RemoteExecutorTestBase.cs
```diff
@@ -0,0 +1,243 @@
...
+        }
+    }
+
+    /// <summary>Options used with RemoteInvoke.</summary>
+    public sealed class RemoteInvokeOptions
+    {
+        public bool Start { get; set; } = true;
+        public ProcessStartInfo StartInfo { get; set; } = new ProcessStartInfo();
+        public bool CheckExitCode { get; set; } = true;
+        public string mode { get; set; }
```
eerhardt: `mode` => `Mode`

eerhardt: Actually, we should probably make this a little more general purpose - as it stands it is really specific to HW intrinsics.

Maybe instead of using `mode`, the should be setting the EnvironmentVariables on the StartInfo.

---
In reply to: [249941399](https://github.com/dotnet/machinelearning/pull/2200#discussion_r249941399) [](ancestors = 249941399)

test/RemoteExecutorConsoleApp/RemoteExecutorConsoleApp.cs
Anipik: I didn't change it due to reason mentioned here https://github.com/dotnet/machinelearning/pull/2200/files#diff-f3a47db0e2b37818c49fd22029730991R91

test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
tannergooding: Even if there was an issue, I would still expect this to be testing `Assert.False` and for us to have an issue linked to whatever workaround we are using instead (which would likely be explicitly setting these both to false).

test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
```diff
@@ -77,431 +81,661 @@ public CpuMathUtilsUnitTests()
...
+        [MemberData(nameof(AddData))]
+        public void SumTest(string test, string mode)
         {
-            float[] src = (float[])_testArrays[test].Clone();
-            var actual = CpuMathUtils.Sum(src);
-            Assert.Equal(expected, actual, 2);
+            RemoteExecutorTestBase.RemoteInvoke((arg0, arg1) =>
+            {
+                CheckProperFlag(arg0);
+                float[] src = (float[])_testArrays[Int32.Parse(arg1)].Clone();
```
tannergooding: nit: The convention is to use `int.Parse`

test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
Anipik: I tried with the recent version of the core-sdk and the issue is still there. i.e setting COMPlus_EnableHWIntrinsic=0  is not disabling the avx and sse

tannergooding: I'll check locally (on a CoreCLR test) and see if I can spot the problem.

test/Directory.Build.props
```diff
@@ -6,7 +6,7 @@
     <TargetFramework Condition="$(Configuration.EndsWith('-netfx'))">net461</TargetFramework>
     <TargetFramework Condition="'$(UseIntrinsics)' == 'true'">netcoreapp3.0</TargetFramework>
     <RuntimeIdentifier Condition="'$(TargetFramework)' == 'net461'">win-x64</RuntimeIdentifier>
-    <DefineConstants Condition="'$(TargetFramework)' == 'netcoreapp3.0'">NETCOREAPP30</DefineConstants>
+    <AllowedReferenceRelatedFileExtensions>$(AllowedReferenceRelatedFileExtensions);.runtimeconfig.json;.runtimeconfig.dev.json;.deps.json</AllowedReferenceRelatedFileExtensions>
```
Anipik: This is required in the projects that are referencing the required project

eerhardt: That would make a fine comment in the file.

tannergooding:
@Anipik, I think something is likely wrong with either the build you are using or how you are setting the environment variables.

Testing locally (on a release build of CoreCLR), I see the variables working as intended.
```
D:\repos\coreclr\bin\tests\Windows_NT.x64.Release\JIT\HardwareIntrinsics\X86\Sse\Sse_ro [master ≡]> D:\repos\coreclr\bin\tests\Windows_NT.x64.Release\Tests\Core_Root\CoreRun.exe .\Sse_ro.exe Add.Single
Supported ISAs:
  AES:       True
  AVX:       True
  AVX2:      True
  BMI1:      True
  BMI2:      True
  FMA:       True
  LZCNT:     True
  PCLMULQDQ: True
  POPCNT:    True
  SSE:       True
  SSE2:      True
  SSE3:      True
  SSE4.1:    True
  SSE4.2:    True
  SSSE3:     True

Beginning test case Add.Single at 1/22/2019 3:54:18 PM
Random seed: 20010415; set environment variable CORECLR_SEED to this value to repro

Beginning scenario: RunBasicScenario_UnsafeRead
Beginning scenario: RunBasicScenario_Load
Beginning scenario: RunBasicScenario_LoadAligned
Beginning scenario: RunReflectionScenario_UnsafeRead
Beginning scenario: RunReflectionScenario_Load
Beginning scenario: RunReflectionScenario_LoadAligned
Beginning scenario: RunClsVarScenario
Beginning scenario: RunLclVarScenario_UnsafeRead
Beginning scenario: RunLclVarScenario_Load
Beginning scenario: RunLclVarScenario_LoadAligned
Beginning scenario: RunClassLclFldScenario
Beginning scenario: RunClassFldScenario
Beginning scenario: RunStructLclFldScenario
Beginning scenario: RunStructFldScenario

Ending test case at 1/22/2019 3:54:18 PM
D:\repos\coreclr\bin\tests\Windows_NT.x64.Release\JIT\HardwareIntrinsics\X86\Sse\Sse_ro [master ≡]> $env:COMPlus_EnableHWIntrinsic=0
D:\repos\coreclr\bin\tests\Windows_NT.x64.Release\JIT\HardwareIntrinsics\X86\Sse\Sse_ro [master ≡]> D:\repos\coreclr\bin\tests\Windows_NT.x64.Release\Tests\Core_Root\CoreRun.exe .\Sse_ro.exe Add.Single
Supported ISAs:
  AES:       False
  AVX:       False
  AVX2:      False
  BMI1:      False
  BMI2:      False
  FMA:       False
  LZCNT:     False
  PCLMULQDQ: False
  POPCNT:    False
  SSE:       False
  SSE2:      False
  SSE3:      False
  SSE4.1:    False
  SSE4.2:    False
  SSSE3:     False

Beginning test case Add.Single at 1/22/2019 3:54:29 PM
Random seed: 20010415; set environment variable CORECLR_SEED to this value to repro

Beginning scenario: RunUnsupportedScenario
Beginning scenario: RunBasicScenario_UnsafeRead

Ending test case at 1/22/2019 3:54:29 PM
D:\repos\coreclr\bin\tests\Windows_NT.x64.Release\JIT\HardwareIntrinsics\X86\Sse\Sse_ro [master ≡]> $env:COMPlus_EnableHWIntrinsic=1; $env:COMPlus_EnableSse=0
D:\repos\coreclr\bin\tests\Windows_NT.x64.Release\JIT\HardwareIntrinsics\X86\Sse\Sse_ro [master ≡]> D:\repos\coreclr\bin\tests\Windows_NT.x64.Release\Tests\Core_Root\CoreRun.exe .\Sse_ro.exe Add.Single
Supported ISAs:
  AES:       False
  AVX:       False
  AVX2:      False
  BMI1:      False
  BMI2:      False
  FMA:       False
  LZCNT:     True
  PCLMULQDQ: False
  POPCNT:    False
  SSE:       False
  SSE2:      False
  SSE3:      False
  SSE4.1:    False
  SSE4.2:    False
  SSSE3:     False

Beginning test case Add.Single at 1/22/2019 3:55:03 PM
Random seed: 20010415; set environment variable CORECLR_SEED to this value to repro

Beginning scenario: RunUnsupportedScenario
Beginning scenario: RunBasicScenario_UnsafeRead

Ending test case at 1/22/2019 3:55:03 PM
D:\repos\coreclr\bin\tests\Windows_NT.x64.Release\JIT\HardwareIntrinsics\X86\Sse\Sse_ro [master ≡]> $env:COMPlus_EnableSse=1; $env:COMPlus_EnableAvx=0
D:\repos\coreclr\bin\tests\Windows_NT.x64.Release\JIT\HardwareIntrinsics\X86\Sse\Sse_ro [master ≡]> D:\repos\coreclr\bin\tests\Windows_NT.x64.Release\Tests\Core_Root\CoreRun.exe .\Sse_ro.exe Add.Single
Supported ISAs:
  AES:       True
  AVX:       False
  AVX2:      False
  BMI1:      False
  BMI2:      False
  FMA:       False
  LZCNT:     True
  PCLMULQDQ: True
  POPCNT:    True
  SSE:       True
  SSE2:      True
  SSE3:      True
  SSE4.1:    True
  SSE4.2:    True
  SSSE3:     True

Beginning test case Add.Single at 1/22/2019 3:55:20 PM
Random seed: 20010415; set environment variable CORECLR_SEED to this value to repro

Beginning scenario: RunBasicScenario_UnsafeRead
Beginning scenario: RunBasicScenario_Load
Beginning scenario: RunBasicScenario_LoadAligned
Beginning scenario: RunReflectionScenario_UnsafeRead
Beginning scenario: RunReflectionScenario_Load
Beginning scenario: RunReflectionScenario_LoadAligned
Beginning scenario: RunClsVarScenario
Beginning scenario: RunLclVarScenario_UnsafeRead
Beginning scenario: RunLclVarScenario_Load
Beginning scenario: RunLclVarScenario_LoadAligned
Beginning scenario: RunClassLclFldScenario
Beginning scenario: RunClassFldScenario
Beginning scenario: RunStructLclFldScenario
Beginning scenario: RunStructFldScenario

Ending test case at 1/22/2019 3:55:20 PM
D:\repos\coreclr\bin\tests\Windows_NT.x64.Release\JIT\HardwareIntrinsics\X86\Sse\Sse_ro [master ≡]>
```

test/Microsoft.ML.TestFramework/RemoteExecutor.cs
```diff
@@ -0,0 +1,201 @@
...
+using Xunit.Sdk;
+
+namespace Microsoft.ML.TestFramework
+{
+    /// <summary>Base class used for all tests that need to spawn a remote process.</summary>
+    public static class RemoteExecutor
+    {
+        /// <summary>The name of the test console app.</summary>
+        public static readonly string TestConsoleApp = Path.GetFullPath(@"RemoteExecutorConsoleApp.dll");
+#if NETFRAMEWORK
```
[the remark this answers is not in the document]
danmosemsft: Ah maybe it's from Microsoft.NET.Sdk.BeforeCommon.targets

test/Microsoft.ML.TestFramework/RemoteExecutor.cs
```diff
@@ -0,0 +1,201 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
```
danmosemsft: Probably should have a comment, here or somewhere nearby, indicating where this code comes from - since someone might want to take an update from corefx one day.

test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
```diff
@@ -75,433 +84,681 @@ public CpuMathUtilsUnitTests()
...
 
-            CpuMathUtils.MatrixTimesSource(false, mat, src, dst, dst.Size);
-            float[] actual = new float[dst.Size];
-            dst.CopyTo(actual, 0, dst.Size);
-            Assert.Equal(expected, actual, _matMulComparer);
+                float[] expected = new float[dst.Size];
+                for (int i = 0; i < dst.Size; i++)
+                {
+                    float dotProduct = 0;
+                    for (int j = 0; j < src.Size; j++)
```
danmosemsft: You changed to calculating the expected results at runtime - because it's cleaner?

test/Microsoft.ML.TestFramework/RemoteExecutor.cs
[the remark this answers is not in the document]
Anipik: yes its automatically defined in .net core sdk https://github.com/dotnet/sdk/pull/2073

test/Microsoft.ML.TestFramework/RemoteExecutorTestBase.cs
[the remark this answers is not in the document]
tannergooding: > we can easily run them on ci

The same way that CoreCLR is running them. You have a separate leg for each configuration you want to test.

> There is a separate test for each code path and a person easily find and debug if any path is misbehaving.

Yes, this is better at that regards. The downside is that each test now has to have a `RemoteExecuteTestBase` call and wrap the logic in a delegate.

We could instead take advantage of various extensibility features that xUnit has. You would just need to define a custom fact/theory-attribute/discoverer. A simple example of this is the `RetryFact`: https://github.com/xunit/samples.xunit/tree/master/RetryFactExample

You would then just configure the `RunAsync` method in the custom `*TestCase` to kick off a process with the appropriate environment variables (basically what `RemoteExecutorTestBase` is doing).

This would ultimately mean that the test code continues looking "normal" and all the environment set-up/remote execution would be abstracted away.

test/Directory.Build.props
```diff
@@ -6,6 +6,8 @@
     <TargetFramework Condition="$(Configuration.EndsWith('-netfx'))">net461</TargetFramework>
     <TargetFramework Condition="'$(UseIntrinsics)' == 'true'">netcoreapp3.0</TargetFramework>
     <RuntimeIdentifier Condition="'$(TargetFramework)' == 'net461'">win-x64</RuntimeIdentifier>
+    <!-- This is required in order to have follwing extension files in the output folder of the projects referencing other projects. -->
```
eerhardt: (nit) `follwing`

test/RemoteExecutorConsoleApp/RemoteExecutorConsoleApp.csproj
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <StartupObject>RemoteExecutorConsoleApp.Program</StartupObject>
```
eerhardt: ?? Is this necessary?

Anipik: yes otherwise it gives an error of having multiple entry points. (I was not able to find the reason behind it though)

test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
```diff
@@ -75,433 +84,681 @@ public CpuMathUtilsUnitTests()
...
+                    }                    
+                    expected[i] = dotProduct;
+                }
+
+                CpuMathUtils.MatrixTimesSource(false, mat, src, dst, dst.Size);
+                float[] actual = new float[dst.Size];
+                dst.CopyTo(actual, 0, dst.Size);
+                Assert.Equal(expected, actual, _matMulComparer);
+                return RemoteExecutor.SuccessExitCode;
+            }, mode, matTest, srcTest, dstTest, new RemoteInvokeOptions(environmentVariables));
```
eerhardt: IMO it is surprising that you are flipping the order of parameters here from the parameters of the method:

```
public void MatMulTest(string matTest, string srcTest, string dstTest, string mode,
```
vs.
```
}, mode, matTest, srcTest, dstTest,
```

Which means inside the delegate `arg0` is `mode` and `arg1` is `matTest`.  I would think these should be in the same order, right?

Anipik: i will make the mode as the first argument in the test defination

test/RemoteExecutorConsoleApp/RemoteExecutorConsoleApp.csproj
eerhardt: Oh, I know why. Because it is in the `test` folder, and we are automatically referencing the `Test` SDK.

https://github.com/dotnet/machinelearning/blob/082e3fcd1df0a9568981a6d12ab529c11007a6ac/test/Directory.Build.props#L29

We could maybe remove that reference in this .csproj, or turn off the Target that is generating the other `Main` method.

eerhardt: See

https://github.com/Microsoft/vstest/blob/069c5f8ef14caff221951d0807485df5d57f0f87/src/package/nuspec/netcoreapp/Microsoft.NET.Test.Sdk.targets#L29-L39

for the parameter that turns this off

test/Microsoft.ML.TestFramework/RemoteExecutorTestBase.cs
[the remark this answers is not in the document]
Anipik: i was able to add a retryFact conditionalTheory attribute.
The problem i am facing in how to Kickoff RunAsync method in a different process ?

here is the sample code https://github.com/Anipik/machinelearning/commit/be09268d30e10acf8b3a6168220333f6a848f5fe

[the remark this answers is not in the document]
danmosemsft: What Tanner suggests would be ideal - I didn't realize that there was an extensibility point that would allow something like this - and I assume @stephentoub didn't know about it either when he wrote the RemoteExec infrastructure originally. However, with async and cross process in the picture it could be a bit fiddly to get this right. I would rather not do that in this PR. I suggest to use the tried and tested RemoteExec pattern we use in CoreFX here. If we want to improve it, we might want to do that in CoreFX first and then copy it here.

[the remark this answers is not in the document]
danmosemsft: @eerhardt thoughts? Even if we do this it should be in a different PR, in my opinion.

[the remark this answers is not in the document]
eerhardt: Doing this in a separate PR sounds fine with me.

PR #2202, opened 2019-01-22
Remove ColumnType.RawKind Round 3 (and final)

### LINKED ISSUE #1533: IDataView Cleanup: ColumnType cleanup

eerhardt:
Completely removes `ColumnType.RawKind`.

Part of the work necessary for #1860. 

Fix #1533.

### DIFF
--- a/src/Microsoft.ML.Core/Data/ColumnType.cs
+++ b/src/Microsoft.ML.Core/Data/ColumnType.cs
@@ -25,24 +25,6 @@ namespace Microsoft.ML.Data
         {
             Contracts.CheckValue(rawType, nameof(rawType));
             RawType = rawType;
-            RawType.TryGetDataKind(out var rawKind);
-            RawKind = rawKind;
-        }
-
-        /// <summary>
-        /// Internal sub types can pass both the <paramref name="rawType"/> and <paramref name="rawKind"/> values.
-        /// This asserts that they are consistent.
-        /// </summary>
-        private protected ColumnType(Type rawType, DataKind rawKind)
-        {
-            Contracts.AssertValue(rawType);
-#if DEBUG
-            DataKind tmp;
-            rawType.TryGetDataKind(out tmp);
-            Contracts.Assert(tmp == rawKind);
-#endif
-            RawType = rawType;
-            RawKind = rawKind;
         }
 
         /// <summary>
@@ -54,20 +36,11 @@ namespace Microsoft.ML.Data
         /// </summary>
         public Type RawType { get; }
 
-        /// <summary>
-        /// The <see cref="DataKind"/> corresponding to <see cref="RawType"/>, if there is one (<c>default</c> otherwise).
-        /// It is equivalent to the result produced by <see cref="DataKindExtensions.TryGetDataKind(Type, out DataKind)"/>.
-        /// For external code it would be preferable to operate over <see cref="RawType"/>.
-        /// </summary>
-        [BestFriend]
-        internal DataKind RawKind { get; }
-
         // IEquatable<T> interface recommends also to override base class implementations of
         // Object.Equals(Object) and GetHashCode. In classes below where Equals(ColumnType other)
         // is effectively a referencial comparison, there is no need to override base class implementations
         // of Object.Equals(Object) (and GetHashCode) since its also a referencial comparison.
         public abstract bool Equals(ColumnType other);
-
     }
 
     /// <summary>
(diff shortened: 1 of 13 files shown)


zeahmed:
>         /// and also the associated data kind for this type. If a data kind could not

Need to update?

---
Refers to: src/Microsoft.ML.Data/DataView/InternalSchemaDefinition.cs:136 in 9d53945. [](commit_id = 9d539450c98483695f6033b168f9d464b80bbc11, deletion_comment = False)

zeahmed:
>         /// They must both be vectors or scalars, and the raw data kind should match.

Here as well?

---
Refers to: src/Microsoft.ML.Data/DataView/TypedCursor.cs:136 in 9d53945. [](commit_id = 9d539450c98483695f6033b168f9d464b80bbc11, deletion_comment = False)

eerhardt:
This PR is ready for review now that #2176 is merged.

Issue #2185, opened 2019-01-18
Conditions in KeyToVectorMappingEstimator.GetOutputSchema seem wrong


TomFinley:
While reviewing #2176, I incidentally read what I take to be some bugs in this method, the schema shape propagation for key values:

https://github.com/dotnet/machinelearning/blob/bb92c06f7207546377f63d969ffe063780ee4922/src/Microsoft.ML.Data/Transforms/KeyToVector.cs#L762

As the name suggests, this "key to vector" transformer this produces a vector valued column out of a key-valued column. That key-valued column can be either a scalar, vector, or unknown size vector. Much of the logic in this method did not seem to conform to my understanding of what the transformer is doing (and, in one case, *should* be doing but isn't). It may be that we need some more test coverage for this, in addition to fixing the bugs, if I am right about any of these.

## Bug 1: The Type Test

Let's review the type test:

https://github.com/dotnet/machinelearning/blob/bb92c06f7207546377f63d969ffe063780ee4922/src/Microsoft.ML.Data/Transforms/KeyToVector.cs#L770

I see some problems here:

* This is the "key to vector" transform, yet, there is no test to see if it is a key?

* Instead first we see that it fails if it's of an unknown `DataKind` (which seems totally besides the point).

* We also see that we have this totally unnecessary `ItemType.GetItemType()`, which considering the description is not necessary.

* Related: later on we see `col.ItemType is VectorType`, which according to the documentation here could never be the case anyway:

 
 https://github.com/dotnet/machinelearning/blob/bb92c06f7207546377f63d969ffe063780ee4922/src/Microsoft.ML.Core/Data/IEstimator.cs#L46-L49

* `col.ItemType is PrimitiveType` is totally redundant given that we already had the much stronger test that it was one of the "known" datakinds. (Which, again, is still completely wrong.)

So, I think practically, we might have a much stronger test if we dropped practically all this matter and just tested `col.IsKey`. *Maybe* also verify that the raw type is not `ulong` or something. But, the rest of this doesn't make a lot of sense to me.

## Bug 2: Error message

Now what does that test lead to... it leads to this exception being thrown:

https://github.com/dotnet/machinelearning/blob/bb92c06f7207546377f63d969ffe063780ee4922/src/Microsoft.ML.Data/Transforms/KeyToVector.cs#L771

If you trace through to what becomes the exception message, it becomes this:

[...]

PR #2217, opened 2019-01-23
Improve exception message and make consistent with ExceptSchemaMismatch

### LINKED ISSUE #2044: Improve TrainerEstimatorBase exception messages for types
Consider this error message:

https://github.com/dotnet/machinelearning/blob/faffd179c961f120ec4e6babb06bbfb2cca6a6ea/src/Microsoft.ML.Data/Training/TrainerEstimatorBase.cs#L98-L99

This is quite a charming message. Most of the time (not always, perhaps) a trainer expects a known size vector of float. What if the user feeds in something else? They get this, and are told that it is "not compatible." It is absolutely true. That's not compatible. They try something else, if it doesn't happen to be exactly right, they get the same error message. Not compatible. So, while absolutely true, it sort of reminds me of this game:

Calvin and Hobbes "Nope guess again"

This error message is in the `TrainerEstimatorBase`. Which is to say, nearly all trainers (except specialized trainers) will check the features column through this, which means that a user's likelihood of seeing this message is fairly high.

Consider the following far more helpful example. Note that we even have a convenience function for standard reporting and formatting of this type mismatch error, since it is so common!

https://github.com/dotnet/machinelearning/blob/faffd179c961f120ec4e6babb06bbfb2cca6a6ea/src/Microsoft.ML.Data/Evaluators/MultiClassClassifierEvaluator.cs#L82-L83

Of course, it may be that there are plenty of places that we are not being less descriptive with types than we should be, but since trainers (and especially the base class!) will be used by nearly everyone, we might identify this as deserving special prioritization.

We should probably change `TrainerEstimatorBase` so that it or its subclasses use these more descriptive helpers, throughout the type checks. Ideally we'd also scour the codebase for typical checks of this form to make sure they are likewise descriptive.

That is at least what we should do, but, in a *super* ideal world, we might be able to go one step further and even provide advice on *what* should be done to get it into the required form (that is, be not just descriptive, but prescriptive, to borrow a phrase from @stephentoub), but how to do that is unclear to me. (The best thing I can think of is, we, by which I mean a data scientist, probably ought to write a web article on typical strategies for featurization, and why it's important to give it some thought, but that's certainly not something you want to put in an exception message. Though it might be a nice thing to *link* from the exception, if possible.)

/cc @stephentoub

artidoro:
Fixes #2044.

I make the exception message consistent using `ExceptSchemaMismatch` when possible. 
In particular, as mentioned in the issue #2044, I did it for the base class for TrainerEstimators.

### DIFF
--- a/src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
+++ b/src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
@@ -100,7 +100,7 @@ namespace Microsoft.ML.Data
             Host.Check(schema.Label.HasValue, "Could not find the label column");
             t = schema.Label.Value.Type;
             if (t != NumberType.Float && t.GetKeyCount() != 2)
-                throw Host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "float or a 2-value KeyType", t.ToString());
+                throw Host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "float or a KeyType with cardinality 2", t.ToString());
         }
 
         private protected override Aggregator GetAggregatorCore(RoleMappedSchema schema, string stratName)
@@ -666,7 +666,7 @@ namespace Microsoft.ML.Data
 
             IDataView overall;
             if (!metrics.TryGetValue(MetricKinds.OverallMetrics, out overall))
-                throw Host.ExceptSchemaMismatch(nameof(overall.Schema), "overall metrics", MetricKinds.OverallMetrics);
+                throw Host.Except("No overall metrics found");
 
             // Find the number of anomalies, and the thresholds.
             int numAnomIndex;
--- a/src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
+++ b/src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
@@ -96,11 +96,11 @@ namespace Microsoft.ML.Data
             var score = schema.GetUniqueColumn(MetadataUtils.Const.ScoreValueKind.Score);
             var t = score.Type;
             if (t != NumberType.Float)
-                throw Host.Except("Score column '{0}' has type '{1}' but must be R4", score, t).MarkSensitive(MessageSensitivity.Schema);
+                throw Host.ExceptSchemaMismatch(nameof(schema), "score", score.Name, "float", t.ToString());
             Host.Check(schema.Label.HasValue, "Could not find the label column");
             t = schema.Label.Value.Type;
             if (t != NumberType.Float && t.GetKeyCount() != 2)
-                throw Host.Except("Label column '{0}' has type '{1}' but must be R4 or a 2-value key", schema.Label.Value.Name, t).MarkSensitive(MessageSensitivity.Schema);
+                throw Host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "float or a 2-value key", t.ToString());
         }
 
         private protected override Aggregator GetAggregatorCore(RoleMappedSchema schema, string stratName)
@@ -631,13 +631,13 @@ namespace Microsoft.ML.Data
             {
                 int index;
                 if (!top.Schema.TryGetColumnIndex(AnomalyDetectionEvaluator.TopKResultsColumns.Instance, out index))
-                    throw Host.Except("Data view does not contain the 'Instance' column");
+                    throw Host.ExceptSchemaMismatch(nameof(top.Schema), "instance", AnomalyDetectionEvaluator.TopKResultsColumns.Instance);
                 var instanceGetter = cursor.GetGetter<ReadOnlyMemory<char>>(index);
                 if (!top.Schema.TryGetColumnIndex(AnomalyDetectionEvaluator.TopKResultsColumns.AnomalyScore, out index))
-                    throw Host.Except("Data view does not contain the 'Anomaly Score' column");
+                    throw Host.ExceptSchemaMismatch(nameof(top.Schema), "anomaly score", AnomalyDetectionEvaluator.TopKResultsColumns.AnomalyScore);
                 var scoreGetter = cursor.GetGetter<Single>(index);
                 if (!top.Schema.TryGetColumnIndex(AnomalyDetectionEvaluator.TopKResultsColumns.Label, out index))
-                    throw Host.Except("Data view does not contain the 'Label' column");
+                    throw Host.ExceptSchemaMismatch(nameof(top.Schema), "label", AnomalyDetectionEvaluator.TopKResultsColumns.Label);
                 var labelGetter = cursor.GetGetter<Single>(index);
 
                 bool hasRows = false;
@@ -666,12 +666,12 @@ namespace Microsoft.ML.Data
 
             IDataView overall;
             if (!metrics.TryGetValue(MetricKinds.OverallMetrics, out overall))
-                throw Host.Except("No overall metrics found");
+                throw Host.ExceptSchemaMismatch(nameof(overall.Schema), "overall metrics", MetricKinds.OverallMetrics);
 
             // Find the number of anomalies, and the thresholds.
             int numAnomIndex;
             if (!overall.Schema.TryGetColumnIndex(AnomalyDetectionEvaluator.OverallMetrics.NumAnomalies, out numAnomIndex))
-                throw Host.Except("Could not find the 'NumAnomalies' column");
+                throw Host.ExceptSchemaMismatch(nameof(overall.Schema), "number of anomalies", AnomalyDetectionEvaluator.OverallMetrics.NumAnomalies);
 
             int stratCol;
             var hasStrat = overall.Schema.TryGetColumnIndex(MetricKinds.ColumnNames.StratCol, out stratCol);
--- a/src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
+++ b/src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
@@ -100,7 +100,7 @@ namespace Microsoft.ML.Data
             Host.Check(schema.Label.HasValue, "Could not find the label column");
             t = schema.Label.Value.Type;
             if (t != NumberType.Float && t.GetKeyCount() != 2)
-                throw Host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "float or a 2-value key", t.ToString());
+                throw Host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "float or a 2-value KeyType", t.ToString());
         }
 
         private protected override Aggregator GetAggregatorCore(RoleMappedSchema schema, string stratName)
--- a/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
+++ b/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
@@ -129,11 +129,11 @@ namespace Microsoft.ML.Data
             var host = Host.SchemaSensitive();
             var t = score.Type;
             if (t != NumberType.Float)
-                throw host.ExceptSchemaMismatch(nameof(schema), "score", score.Name, "R4", t.ToString());
+                throw host.ExceptSchemaMismatch(nameof(schema), "score", score.Name, "float", t.ToString());
             host.Check(schema.Label.HasValue, "Could not find the label column");
             t = schema.Label.Value.Type;
             if (t != NumberType.R4 && t != NumberType.R8 && t != BoolType.Instance && t.GetKeyCount() != 2)
-                throw host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "R4, R8, BL or a 2-value key", t.ToString());
+                throw host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "float, double, bool, or a 2-value key", t.ToString());
         }
 
         private protected override void CheckCustomColumnTypesCore(RoleMappedSchema schema)
@@ -145,7 +145,7 @@ namespace Microsoft.ML.Data
                 host.CheckParam(prob.Count == 1, nameof(schema), "Cannot have multiple probability columns");
                 var probType = prob[0].Type;
                 if (probType != NumberType.Float)
-                    throw host.ExceptSchemaMismatch(nameof(schema), "probability", prob[0].Name, "R4", probType.ToString());
+                    throw host.ExceptSchemaMismatch(nameof(schema), "probability", prob[0].Name, "float", probType.ToString());
             }
             else if (!_useRaw)
             {
--- a/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
+++ b/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
@@ -133,7 +133,7 @@ namespace Microsoft.ML.Data
             host.Check(schema.Label.HasValue, "Could not find the label column");
             t = schema.Label.Value.Type;
             if (t != NumberType.R4 && t != NumberType.R8 && t != BoolType.Instance && t.GetKeyCount() != 2)
-                throw host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "float, double, bool, or a 2-value key", t.ToString());
+                throw host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "float, double, bool, or a 2-value KeyType", t.ToString());
         }
 
         private protected override void CheckCustomColumnTypesCore(RoleMappedSchema schema)
@@ -1098,7 +1098,7 @@ namespace Microsoft.ML.Data
 
             var t = schema[(int)LabelIndex].Type;
             if (t != NumberType.R4 && t != NumberType.R8 && t != BoolType.Instance && t.GetKeyCount() != 2)
-                throw Host.ExceptSchemaMismatch(nameof(schema), "label", LabelCol, "float, double, bool or a 2-value key", t.ToString());
+                throw Host.ExceptSchemaMismatch(nameof(schema), "label", LabelCol, "float, double, bool or a 2-value KeyType", t.ToString());
 
             t = schema[ScoreIndex].Type;
             if (t != NumberType.Float)
--- a/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
+++ b/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
@@ -133,7 +133,7 @@ namespace Microsoft.ML.Data
             host.Check(schema.Label.HasValue, "Could not find the label column");
             t = schema.Label.Value.Type;
             if (t != NumberType.R4 && t != NumberType.R8 && t != BoolType.Instance && t.GetKeyCount() != 2)
-                throw host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "float, double, bool, or a 2-value KeyType", t.ToString());
+                throw host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "float, double, bool, or a KeyType with cardinality 2", t.ToString());
         }
 
         private protected override void CheckCustomColumnTypesCore(RoleMappedSchema schema)
@@ -1098,7 +1098,7 @@ namespace Microsoft.ML.Data
 
             var t = schema[(int)LabelIndex].Type;
             if (t != NumberType.R4 && t != NumberType.R8 && t != BoolType.Instance && t.GetKeyCount() != 2)
-                throw Host.ExceptSchemaMismatch(nameof(schema), "label", LabelCol, "float, double, bool or a 2-value KeyType", t.ToString());
+                throw Host.ExceptSchemaMismatch(nameof(schema), "label", LabelCol, "float, double, bool or a KeyType with cardinality 2", t.ToString());
 
             t = schema[ScoreIndex].Type;
             if (t != NumberType.Float)
--- a/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
+++ b/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
@@ -1098,18 +1098,18 @@ namespace Microsoft.ML.Data
 
             var t = schema[(int)LabelIndex].Type;
             if (t != NumberType.R4 && t != NumberType.R8 && t != BoolType.Instance && t.GetKeyCount() != 2)
-                throw Host.Except("Label column '{0}' has type '{1}' but must be R4, R8, BL or a 2-value key", LabelCol, t);
+                throw Host.ExceptSchemaMismatch(nameof(schema), "label", LabelCol, "float, double, bool or a 2-value key", t.ToString());
 
             t = schema[ScoreIndex].Type;
             if (t != NumberType.Float)
-                throw Host.Except("Score column '{0}' has type '{1}' but must be R4", ScoreCol, t);
+                throw Host.ExceptSchemaMismatch(nameof(schema), "score", ScoreCol, "float", t.ToString());
 
             if (_probIndex >= 0)
             {
                 Host.Assert(!string.IsNullOrEmpty(_probCol));
                 t = schema[_probIndex].Type;
                 if (t != NumberType.Float)
-                    throw Host.Except("Probability column '{0}' has type '{1}' but must be R4", _probCol, t);
+                    throw Host.ExceptSchemaMismatch(nameof(schema), "probability", _probCol, "float", t.ToString());
             }
             else if (!_useRaw)
                 throw Host.Except("Cannot compute the predicted label from the probability column because it does not exist");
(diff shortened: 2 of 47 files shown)


src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
```diff
@@ -630,13 +630,13 @@ private protected override void PrintFoldResultsCore(IChannel ch, Dictionary<str
             {
                 int index;
                 if (!top.Schema.TryGetColumnIndex(AnomalyDetectionEvaluator.TopKResultsColumns.Instance, out index))
-                    throw Host.Except("Data view does not contain the 'Instance' column");
+                    throw Host.ExceptSchemaMismatch(nameof(top.Schema), "instance", AnomalyDetectionEvaluator.TopKResultsColumns.Instance);
                 var instanceGetter = cursor.GetGetter<ReadOnlyMemory<char>>(index);
                 if (!top.Schema.TryGetColumnIndex(AnomalyDetectionEvaluator.TopKResultsColumns.AnomalyScore, out index))
-                    throw Host.Except("Data view does not contain the 'Anomaly Score' column");
```
sfilipi: >Anomaly Score' [](start = 71, length = 14)

some people would get picky about the capitalization :)

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
```diff
@@ -1097,18 +1097,18 @@ private void CheckInputColumnTypes(Schema schema)
 
             var t = schema[(int) LabelIndex].Type;
             if (t != NumberType.R4 && t != NumberType.R8 && t != BoolType.Instance && t.GetKeyCount() != 2)
-                throw Host.Except("Label column '{0}' has type '{1}' but must be R4, R8, BL or a 2-value key", LabelCol, t);
```
sfilipi: >'{1}' but must be R4, R8, BL [](start = 63, length = 28)

some of them don't apply anymore for ML.Net, because for the estimators we create the label column to be of a specific type, depending on the task.. I think only bool applies here; BUT the catch is that the message would be incorrect for the CDM users, if we trim it to just bool. 

@tomfinley, @glebuk for preferences.

src/Microsoft.ML.Data/Evaluators/MultiClassClassifierEvaluator.cs
```diff
@@ -814,10 +814,10 @@ private void CheckInputColumnTypes(Schema schema)
 
             var scoreType = schema[ScoreIndex].Type as VectorType;
             if (scoreType == null || scoreType.Size < 2 || scoreType.ItemType != NumberType.Float)
-                throw Host.Except("Score column '{0}' has type '{1}' but must be a vector of two or more items of type R4", ScoreCol, scoreType);
+                throw Host.Except(nameof(schema), "score", "vector of two or more items of type float", ScoreCol, scoreType.ToString());
             var labelType = schema[LabelIndex].Type;
             if (labelType != NumberType.Float && labelType.GetKeyCount() <= 0)
-                throw Host.Except("Label column '{0}' has type '{1}' but must be a float or a known-cardinality key", LabelCol, labelType);
```
sfilipi: >a known-cardinality  [](start = 92, length = 20)

did this get omitted due to the recent changes in the Key type?

src/Microsoft.ML.Data/Evaluators/MultiOutputRegressionEvaluator.cs
```diff
@@ -547,7 +547,7 @@ private protected override Delegate[] CreateGettersCore(Row input, Func<int, boo
 
             var t = schema[LabelIndex].Type as VectorType;
             if (t == null || !t.IsKnownSize || (t.ItemType != NumberType.R4 && t.ItemType != NumberType.R8))
-                throw Host.Except("Label column '{0}' has type '{1}' but must be a known-size vector of R4 or R8", LabelCol, t);
+                throw Host.Except(nameof(schema), "label", LabelCol, "known-size vector of float or double", t.ToString());
```
sfilipi: i'd keep the word "type":

"known-size vector with items of type float or double"

src/Microsoft.ML.Data/Evaluators/MultiClassClassifierEvaluator.cs
artidoro: Yes! there will be no difference between known and unknown cardinality keys.

src/Microsoft.ML.Data/Evaluators/MultiOutputRegressionEvaluator.cs
```diff
@@ -556,7 +556,7 @@ private protected override Delegate[] CreateGettersCore(Row input, Func<int, boo
 
             t = schema[ScoreIndex].Type as VectorType;
             if (t == null || !t.IsKnownSize || t.ItemType != NumberType.Float)
-                throw Host.Except("Score column '{0}' has type '{1}' but must be a known length vector of type R4", ScoreCol, t);
+                throw Host.Except(nameof(schema), "score", ScoreCol, "known-size vector of float", t.ToString());
```
zeahmed: >Except [](start = 27, length = 6)

Inconsistent: You are not using `ExceptSchemaMismatch` here as I see you used it in `QuantileRegressionEvaluator.cs` at line 453 for the same type of error.

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
```diff
@@ -413,24 +413,25 @@ public SchemaShape GetOutputSchema(SchemaShape inputSchema)
...
 
             // Check if label column is good.
             var labelColumn = new SchemaShape.Column(LabelName, SchemaShape.Column.VectorKind.Scalar, NumberType.R4, false);
-            CheckColumnsCompatible(labelColumn, LabelName);
+            CheckColumnsCompatible(labelColumn, "label");
 
             // Check if columns of matrix's row and column indexes are good. Note that column of IDataView and column of matrix are two different things.
             var matrixColumnIndexColumn = new SchemaShape.Column(MatrixColumnIndexName, SchemaShape.Column.VectorKind.Scalar, NumberType.U4, true);
             var matrixRowIndexColumn = new SchemaShape.Column(MatrixRowIndexName, SchemaShape.Column.VectorKind.Scalar, NumberType.U4, true);
-            CheckColumnsCompatible(matrixColumnIndexColumn, MatrixColumnIndexName);
```
zeahmed: >MatrixColumnIndexName [](start = 60, length = 21)

is `MatrixColumnIndexName` not a correct column name?

src/Microsoft.ML.StandardLearners/FactorizationMachine/FactorizationMachineTrainer.cs
```diff
@@ -519,25 +519,26 @@ public SchemaShape GetOutputSchema(SchemaShape inputSchema)
...
-                    throw Host.ExceptSchemaMismatch(nameof(col), defaultName, defaultName);
+                    throw Host.ExceptSchemaMismatch(nameof(inputSchema), columnRole, column.Name);
 
                 if (!column.IsCompatibleWith(col))
-                    throw Host.Except($"{defaultName} column '{column.Name}' is not compatible");
+                    throw Host.ExceptSchemaMismatch(nameof(inputSchema), columnRole, column.Name,
+                        column.GetTypeString(), col.GetTypeString());
             }
 
-            CheckColumnsCompatible(LabelColumn, DefaultColumnNames.Label);
```
zeahmed: >DefaultColumnNames.Label [](start = 48, length = 24)

Are we going to remove `DefaultColumnNames` ?

src/Microsoft.ML.Data/Evaluators/MultiOutputRegressionEvaluator.cs
artidoro: Thanks for the catch!

---
In reply to: [250753412](https://github.com/dotnet/machinelearning/pull/2217#discussion_r250753412) [](ancestors = 250753412)

src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
artidoro: I don't think it changes to lowercase.

However, the second argument of `ExceptSchemaMismatch` is `string columnRole`  and the message becomes:
                `return $"Could not find {columnRole} column '{columnName}'";`

So columnRole only serves the purpose of qualifying "column", while the parameter `columnName` will give the exact name. That's why I have made it lowercase everywhere.



---
In reply to: [250474268](https://github.com/dotnet/machinelearning/pull/2217#discussion_r250474268) [](ancestors = 250474268)

src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
```diff
@@ -665,12 +665,12 @@ private protected override void PrintFoldResultsCore(IChannel ch, Dictionary<str
...
             IDataView overall;
             if (!metrics.TryGetValue(MetricKinds.OverallMetrics, out overall))
-                throw Host.Except("No overall metrics found");
+                throw Host.ExceptSchemaMismatch(nameof(overall.Schema), "overall metrics", MetricKinds.OverallMetrics);
 
             // Find the number of anomalies, and the thresholds.
             int numAnomIndex;
             if (!overall.Schema.TryGetColumnIndex(AnomalyDetectionEvaluator.OverallMetrics.NumAnomalies, out numAnomIndex))
-                throw Host.Except("Could not find the 'NumAnomalies' column");
+                throw Host.ExceptSchemaMismatch(nameof(overall.Schema), "number of anomalies", AnomalyDetectionEvaluator.OverallMetrics.NumAnomalies);
```
singlis: >number [](start = 73, length = 6)

I would not change the name of the column, in fact, I would just use the constant AnomalyDetectionEvaluator.OverallMetrics.NumAnomalies.

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
```diff
@@ -128,11 +128,11 @@ private protected override void CheckScoreAndLabelTypes(RoleMappedSchema schema)
             var host = Host.SchemaSensitive();
             var t = score.Type;
             if (t != NumberType.Float)
-                throw host.ExceptSchemaMismatch(nameof(schema), "score", score.Name, "R4", t.ToString());
+                throw host.ExceptSchemaMismatch(nameof(schema), "score", score.Name, "float", t.ToString());
             host.Check(schema.Label.HasValue, "Could not find the label column");
             t = schema.Label.Value.Type;
             if (t != NumberType.R4 && t != NumberType.R8 && t != BoolType.Instance && t.GetKeyCount() != 2)
-                throw host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "R4, R8, BL or a 2-value key", t.ToString());
+                throw host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "float, double, bool, or a 2-value KeyType", t.ToString());
```
singlis: >label [](start = 65, length = 5)

Instead of having the literal string "label" in each of these SchemaMismatch exceptions, what about adding some common column names somewhere as const strings? Label, Score, Probability, and Features seem like some common ones to use.

singlis: also what is a 2-value KeyType?

---
In reply to: [251127632](https://github.com/dotnet/machinelearning/pull/2217#discussion_r251127632) [](ancestors = 251127632)

src/Microsoft.ML.Data/Evaluators/ClusteringEvaluator.cs
```diff
@@ -99,13 +99,13 @@ private protected override void CheckScoreAndLabelTypes(RoleMappedSchema schema)
             if (type != null && type != NumberType.Float && !(type is KeyType keyType && keyType.Count > 0))
             {
                 throw Host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name,
-                    "R4 or key of known cardinality", type.ToString());
+                    "float or key", type.ToString());
```
singlis: >key [](start = 30, length = 3)

Should this be key or KeyType?

src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
```diff
@@ -95,11 +95,11 @@ private protected override void CheckScoreAndLabelTypes(RoleMappedSchema schema)
             var score = schema.GetUniqueColumn(MetadataUtils.Const.ScoreValueKind.Score);
             var t = score.Type;
             if (t != NumberType.Float)
-                throw Host.Except("Score column '{0}' has type '{1}' but must be R4", score, t).MarkSensitive(MessageSensitivity.Schema);
+                throw Host.ExceptSchemaMismatch(nameof(schema), "score", score.Name, "float", t.ToString());
             Host.Check(schema.Label.HasValue, "Could not find the label column");
             t = schema.Label.Value.Type;
             if (t != NumberType.Float && t.GetKeyCount() != 2)
-                throw Host.Except("Label column '{0}' has type '{1}' but must be R4 or a 2-value key", schema.Label.Value.Name, t).MarkSensitive(MessageSensitivity.Schema);
+                throw Host.ExceptSchemaMismatch(nameof(schema), "label", schema.Label.Value.Name, "float or a 2-value KeyType", t.ToString());
```
Ivanidzo4ka: >2-value KeyType [](start = 110, length = 15)

KeyType with cardinality 2 ?

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
artidoro: The problem with DefaultColumnNames is twofold:

1. It contains "Column" in the name (which would lead to a repetition of column in the message)
2. It does not contain many column roles such the very common "input" case.

In general, my understanding from talking with Tom is that the intent of the `columnRole` parameter was to be a interpretable qualification of the column role. 

Quoting Tom: 

"
I think Pete's intention was that it would be a descriptive human readable name.
So I've been going with hard coded small things like "input" and "label" and stuff.
"

---
In reply to: [251130649](https://github.com/dotnet/machinelearning/pull/2217#discussion_r251130649) [](ancestors = 251130649,251127632)

src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
artidoro: The name of the column does not include a space, so the current parameter `columnRole` is already not matching the actual column name. As you see in my version, I include both the parameter `column role` which is a description of the role of the column "anomaly score", and the actual column name `AnomalyDetectionEvaluator.TopKResultsColumns.AnomalyScore` as part of the exception message. 

---
In reply to: [251125939](https://github.com/dotnet/machinelearning/pull/2217#discussion_r251125939) [](ancestors = 251125939,250823347,250474268)

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
artidoro: will fix the 2-value KeyType thanks! I'll check everywhere for that error.

---
In reply to: [251143821](https://github.com/dotnet/machinelearning/pull/2217#discussion_r251143821) [](ancestors = 251143821,251130649,251127632)

singlis: Sorry - when you say it contains Column, are you referring to the variable name DefaultColumnNames? Is the concern this would look weird from the coding perspective? To the user, they willl just see the column name like "Label". And I think its good to have one place where this string is used as this makes it easier to maintain. I can talk more about it offline if you want.

---
In reply to: [251143821](https://github.com/dotnet/machinelearning/pull/2217#discussion_r251143821) [](ancestors = 251143821,251130649,251127632)

src/Microsoft.ML.Data/Transforms/RangeFilter.cs
```diff
@@ -150,11 +150,11 @@ private RangeFilter(IHost host, ModelLoadContext ctx, IDataView input)
             var column = ctx.LoadNonEmptyString();
             var schema = Source.Schema;
             if (!schema.TryGetColumnIndex(column, out _index))
-                throw Host.Except("column", "Source column '{0}' not found", column);
+                throw Host.ExceptSchemaMismatch(nameof(schema), "source", column);
 
             _type = schema[_index].Type;
             if (_type != NumberType.R4 && _type != NumberType.R8 && _type.GetKeyCount() == 0)
-                throw Host.Except("column", "Column '{0}' does not have compatible type", column);
+                throw Host.ExceptSchemaMismatch(nameof(schema), "source", column, "flaot, double or KeyType", _type.ToString());
```
singlis: >flaot [](start = 83, length = 5)

float mispelled

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
TomFinley: My preference would be for the type description to be accurate w.r.t. the actual requirements. From the condition, it is clearly allowing for float and double inputs, so the message ought to reflect that.

If we wanted to form an actual restriction that this *must* be a `bool`, I would treat that as a separate discussion to improving the actual messages. There might be a case to be made for that. (Indeed, in the so-called "static" API, I made this a requirement.) But I'd want to discuss that issue separately. Right now I'd say it's most important that the message be correct.

src/Microsoft.ML.Data/Evaluators/AnomalyDetectionEvaluator.cs
TomFinley: Right, this is meant to be human readable. What @Zruty0 was doing was trying to unify the error messages, I believe. And they were things like "the label column named 'Bubba' should be like this" or "the score column named 'Bizblam' should be like that," and whatnot. So I think the lowercase names for these roles specifically is appropriate, since it is just meant to be a textual message read and understood by a human.

For that same reason, I'm not too worried about these little strings everywhere. Considering that we have hundreds, possibly thousands of hand-written error messages, this seems fairly minor. So I'm not sure the situation requires something as elaborate as yet-another-static-class-full-of-strings. (I feel differently about something like `MetadataUtils.Kinds` where correctness of algorithms is at stake if we get those wrong.) At a certain point these things lend less clarity, not more, I think?

src/Microsoft.ML.Data/Transforms/RangeFilter.cs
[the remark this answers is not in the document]
artidoro: ColumnType is one of the many possibilities so I would rather not special case it.

---
In reply to: [251181409](https://github.com/dotnet/machinelearning/pull/2217#discussion_r251181409) [](ancestors = 251181409)

src/Microsoft.ML.EntryPoints/FeatureCombiner.cs
```diff
@@ -269,7 +269,7 @@ public static CommonOutputs.TransformOutput ConvertPredictedLabel(IHostEnvironme
 
             var predictedLabelCol = input.Data.Schema.GetColumnOrNull(input.PredictedLabelColumn);
             if (!predictedLabelCol.HasValue)
-                throw host.ExceptSchemaMismatch(nameof(input), "PredictedLabel",input.PredictedLabelColumn);
+                throw host.ExceptSchemaMismatch(nameof(input), "label", input.PredictedLabelColumn);
```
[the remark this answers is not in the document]
artidoro: Fixed.

---
In reply to: [251181481](https://github.com/dotnet/machinelearning/pull/2217#discussion_r251181481) [](ancestors = 251181481)

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
artidoro: I used DefaultColumnNames when possible (for the Features, Score, GroupID, Label).

src/Microsoft.ML.Data/Transforms/RangeFilter.cs
artidoro: Thanks for the suggestion.

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
```diff
@@ -413,24 +413,25 @@ public SchemaShape GetOutputSchema(SchemaShape inputSchema)
...
                 if (!cachedColumn.IsCompatibleWith(col))
-                    throw Host.Except($"{expectedColumnName} column '{cachedColumn.Name}' is not compatible");
+                    throw Host.ExceptSchemaMismatch(nameof(inputSchema), columnRole, cachedColumn.Name,
+                        cachedColumn.GetTypeString(), col.GetTypeString());
             }
 
             // Check if label column is good.
             var labelColumn = new SchemaShape.Column(LabelName, SchemaShape.Column.VectorKind.Scalar, NumberType.R4, false);
-            CheckColumnsCompatible(labelColumn, LabelName);
+            CheckColumnsCompatible(labelColumn, "label");
```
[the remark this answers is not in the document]
artidoro: That will be accessed through the labelColumn, and it's the user provided name for the column. The second argument  is for the column role.

src/Microsoft.ML.Core/Utilities/Contracts.cs
```diff
@@ -454,14 +455,22 @@ public static Exception ExceptSchemaMismatch(this IExceptionContext ctx, string
             => Process(new ArgumentOutOfRangeException(paramName, MakeSchemaMismatchMsg(columnRole, columnName)), ctx);
         public static Exception ExceptSchemaMismatch(string paramName, string columnRole, string columnName, string expectedType, string actualType)
             => Process(new ArgumentOutOfRangeException(paramName, MakeSchemaMismatchMsg(columnRole, columnName, expectedType, actualType)));
+        public static Exception ExceptSchemaMismatch(string paramName, string columnRole, string columnName, Type expectedType, string actualType)
+            => Process(new ArgumentOutOfRangeException(paramName, MakeSchemaMismatchMsg(columnRole, columnName, expectedType.ToString(), actualType)));
+        public static Exception ExceptSchemaMismatch(string paramName, string columnRole, string columnName, Type[] expectedType, string actualType)
```
TomFinley: I do not like this very much. First of all it is using the wrong type... the `IDataView` type system is based on its own type system, which is distinct from the .NET type system. (E.g., key values and unsigned integers have the same .NET type, but different `ColumnType` instances.) Also one of the most common instances of this is something abstract like, "a known sized vector of float," which is not a concept expressible in this. Also it seems to be introducing a "convenience" for a very specialized case.

So, all this is to say, this last addition I am unenthusiastic about. Could we revert that please, and reconsider?

src/Microsoft.ML.Recommender/MatrixFactorizationPredictor.cs
```diff
@@ -455,9 +455,9 @@ public MatrixFactorizationPredictionTransformer(IHostEnvironment host, ModelLoad
         public override Schema GetOutputSchema(Schema inputSchema)
         {
             if (!inputSchema.TryGetColumnIndex(MatrixColumnIndexColumnName, out int xCol))
-                throw Host.ExceptSchemaMismatch(nameof(inputSchema), RecommenderUtils.MatrixColumnIndexKind.Value, MatrixColumnIndexColumnName);
+                throw Host.ExceptSchemaMismatch(nameof(inputSchema), "matrixColumnIndex", MatrixColumnIndexColumnName);
             if (!inputSchema.TryGetColumnIndex(MatrixRowIndexColumnName, out int yCol))
-                throw Host.ExceptSchemaMismatch(nameof(inputSchema), RecommenderUtils.MatrixRowIndexKind.Value, MatrixRowIndexColumnName);
+                throw Host.ExceptSchemaMismatch(nameof(inputSchema), "matrixRowIndex", MatrixRowIndexColumnName);
```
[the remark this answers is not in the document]
artidoro: Done.

src/Microsoft.ML.Core/Utilities/Contracts.cs
artidoro: That sounds good, will do. What do you think of the use of typeof(float).ToString() instead of using "float"?

artidoro: After discussing with Tom, I am reverting my changes only keeping the switch from "input" to "source" column in transformers. @singlis

TomFinley: Thanks @artidoro . I think I should probably explain my thinking here, since it may seem counterintuitive.

So generally I've found that having conveniences to make things more consistent is a good thing. However, one thing that I've observed, even in this PR, is that for ***exception messages specifically***, I am a little wary of it. I consider the problems of #2185. One of the problems is the error messages are wrong, but to a casual reader it is not *obvious* that the error messages are wrong, since it is buried behind layers of "convenience" to make the messages consistent. I don't think the original author had ill intent, but they just got confused about what "convenience" led to what message and what the meaning of the inputs is. (As in fact we also saw here, with the controversy about "label" vs. "Label", and whether something was supposed to be a column name or not!!)

I'm not sure how to solve that issue, but it seems to me like this `ExceptSchemaMismatch` has some problem that makes it badly conceived if its effect of "consistency" is that people are writing wrong error messages, and being fundamentally confused about it. I don't know how to solve that issue. But the change to introduce *more* layers of abstraction would only compound the problem that these "conveniences" are hard to use and people evidently find them very confusing.

This is not to say I think we should not use `ExceptSchemaMismatch`, but I think it needs to be clarified in some fashion. Right now to a reviewer it is really, really hard to tell what is a "right" vs. "wrong" message (as seen by all the discussions in this PR), and the suggestion to get out of that hole by continuing to dig was not one I liked very much. Which is why I had to disagree with some other suggestions in this PR, sorry.

PR #2220, opened 2019-01-24
Extract IDataView into its own assembly and NuGet package

### LINKED ISSUE #1860: Extract IDataView into its own assembly and NuGet package

eerhardt:
Moving IDataView and related types into a new library: `Microsoft.Data.DataView`

Fix #1860

I've split the changes into 3 commits for easy reviewing:

1. Copy the files out into the new folder with minimal/zero changes
2. Get the new assembly building
3 Get ML.NET building on the new assembly


yaeldekel:
> namespace Microsoft.ML.Data

Will the namespace still have "ML" in it?

---
Refers to: src/Microsoft.Data.DataView/IDataView.cs:9 in 33c8cec. [](commit_id = 33c8cecfc1c25413c4a723beb62ed9200cbd62bb, deletion_comment = False)

eerhardt:
> Will the namespace still have "ML" in it?

Sorry, I should have commented on this originally.  My initial implementation is just to move the types in the same namespace. Then the next change is to move namespaces. Since that is going to affect a huge number of files, I wanted to put that in its own PR for easy reviewing.

I am also considering supporting more TFMs in the new library (`net45` and `netstandard1.x`) so it can get as far of reach as possible. That will be a future PR as well.

src/Microsoft.Data.DataView/ColumnType.cs
```diff
@@ -0,0 +1,374 @@
...
+        }
+
+        private static volatile NumberType _instI1;
+        public static NumberType I1
+        {
+            get
+            {
+                if (_instI1 == null)
+                    Interlocked.CompareExchange(ref _instI1, new NumberType(typeof(sbyte), "I1"), null);
+                return _instI1;
```
stephentoub: It's unlikely to make a real difference, especially on x86/64, but this will result in two volatile reads instead of one, since you're reading from the volatile first for the null check and then for the value to return.  You can instead do something like:
```C#
NumberType result = _instI1;
if (result == null)
{
    Interlocked.CompareExchange(ref _instI1, new NumberType(typeof(sbyte), "I1"), null);
    result = _instI1;
}
return result;
```
or if you wanted to be fancy about it:
```C#
return
    _instI1 ??
    Interlocked.CompareExchange(ref _instI1, new NumberType(typeof(sbyte), "I1"), null) ??
    _instI1;
```

src/Microsoft.Data.DataView/ColumnType.cs
```diff
@@ -0,0 +1,374 @@
...
+// See the LICENSE file in the project root for more information.
+
+#pragma warning disable 420 // volatile with Interlocked.CompareExchange
+
+using System;
+using System.Diagnostics;
+using System.Reflection;
+using System.Threading;
+
+namespace Microsoft.ML.Data
```
stephentoub: It's a bit strange to have an assembly named Microsoft.Data.* that contains the namespace Microsoft.ML.Data, no?  I'd have expected things to be in Microsoft.Data.*.

eerhardt: The plan is to rename this namespace. But that is being done in a separate PR because it will touch almost every file in the repo.

src/Microsoft.Data.DataView/ColumnType.cs
```diff
@@ -0,0 +1,374 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#pragma warning disable 420 // volatile with Interlocked.CompareExchange
```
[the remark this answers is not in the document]
eerhardt: Good call.  Addressed in #2296.

src/Microsoft.Data.DataView/ColumnType.cs
eerhardt: Good call.  Addressed in #2296.

src/Microsoft.Data.DataView/ColumnType.cs
```diff
@@ -0,0 +1,374 @@
...
+        {
+            if (typeof(IDisposable).GetTypeInfo().IsAssignableFrom(RawType.GetTypeInfo()))
+                throw new ArgumentException("A " + nameof(PrimitiveType) + " cannot have a disposable " + nameof(RawType), nameof(rawType));
+        }
+    }
+
+    /// <summary>
+    /// The standard text type.
+    /// </summary>
+    public sealed class TextType : PrimitiveType
```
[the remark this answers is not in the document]
eerhardt: Logged #2297 for this and other naming issues with the types in this assembly.

PR #2230, opened 2019-01-24
Enable test coverage

sharwell:
🚧 Testing


shauheen:
I don't think adding the test folder is valuable for the coverage at this point.

PR #2234, opened 2019-01-24
Adding attribution info for the Iris dataset

najeeb-kazmi:
Adding attribution info for the Iris dataset


najeeb-kazmi:
@codemzs could you take a look at why code coverage would drop by 5%? I'm only adding to a readme.

PR #2241, opened 2019-01-25
Make sure seed works for stratification column in TrainTest and CrossValidate

### LINKED ISSUE #1635: TrainTestSplit random seed
I am repeatedly calling `TrainTestSplit` for a data set (for cross validation) and see that the resulting split is the same every call. In sklearn, the `train_test_split` function has the possibility of taking a seed for a random number generator as an input. Could this be added also in ML.NET?

Ivanidzo4ka:
I hope this is final nail into #1635.
Fixes #1635


src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -151,7 +151,12 @@ private void EnsureStratificationColumn(ref IDataView data, ref string stratific
                     // Generate a new column with the hashed stratification column.
                     while (data.Schema.TryGetColumnIndex(stratificationColumn, out tmp))
                         stratificationColumn = string.Format("{0}_{1:000}", origStratCol, ++inc);
-                    data = new HashingEstimator(Host, origStratCol, stratificationColumn, 30).Fit(data).Transform(data);
+                    HashingTransformer.ColumnInfo columnInfo;
+                    if (seed.HasValue)
+                        columnInfo = new HashingTransformer.ColumnInfo(origStratCol, stratificationColumn, 30, seed.Value);
```
rogancarr: I wish there was a way to use `??` but I don't see it.

test/Microsoft.ML.Tests/Scenarios/Api/TestApi.cs
```diff
@@ -289,5 +289,52 @@ private List<BreastCancerExample> ReadBreastCancerExamples()
...
+            var dataPath = GetDataPath("adult.tiny.with-schema.txt");
+            // Create the reader: define the data columns and where to find them in the text file.
+            var input = mlContext.Data.ReadFromTextFile(dataPath, new[] {
+                            new TextLoader.Column("Label", DataKind.BL, 0),
+                            new TextLoader.Column("Workclass", DataKind.TX, 1),
+                            new TextLoader.Column("Education", DataKind.TX,2),
+                            new TextLoader.Column("Age", DataKind.R4,9)
+            }, hasHeader: true);
+
+            var (simpleTrain, simpleTest) = mlContext.BinaryClassification.TrainTestSplit(input);
```
rogancarr: Are we supporting older versions of the C# language in our examples?

test/Microsoft.ML.Tests/Scenarios/Api/TestApi.cs
```diff
@@ -289,5 +289,52 @@ private List<BreastCancerExample> ReadBreastCancerExamples()
...
+            Func<IDataView, List<string>> getWorkclass = (IDataView view) =>
```
rogancarr: Can you give a comment here on why you are doing this?

test/Microsoft.ML.Tests/Scenarios/Api/TestApi.cs
```diff
@@ -289,5 +289,52 @@ private List<BreastCancerExample> ReadBreastCancerExamples()
...
+            {
+                return view.GetColumn<ReadOnlyMemory<char>>(mlContext, "Workclass").Select(x => x.ToString()).ToList();
+            };
+            var simpleTrainWorkclass = getWorkclass(simpleTrain);
+            var simpleTestWorkClass = getWorkclass(simpleTest);
+            var (simpleTrainWithSeed, simpleTestWithSeed) = mlContext.BinaryClassification.TrainTestSplit(input, seed: 10);
+            var simpleWithSeedTrainWorkClass = getWorkclass(simpleTrainWithSeed);
+            var simpleWithSeedTestWorkClass = getWorkclass(simpleTestWithSeed);
+            // Validate we get differnt test sets (they are smaller than TrainSet)
```
rogancarr: different

test/Microsoft.ML.Tests/Scenarios/Api/TestApi.cs
```diff
@@ -289,5 +289,52 @@ private List<BreastCancerExample> ReadBreastCancerExamples()
...
+            Assert.NotEqual(simpleTestWorkClass, simpleWithSeedTestWorkClass);
+
+            var (stratTrain, stratTest) = mlContext.BinaryClassification.TrainTestSplit(input, stratificationColumn: "Workclass");
```
rogancarr: Maybe mark these blocks somehow to show that they are testing different parameter combinations.

rogancarr:
This is a weird use of stratification. I see here that it's designed to make sure that `GroupIds` don't split across train and test sets, but usually we stratify to make sure that %s of some `label` are equal across splits.

src/Microsoft.ML.Data/TrainCatalog.cs
sfilipi: >30 [](start = 107, length = 2)

thinking out loud: should this be settable too?

Ivanidzo4ka: I could make seed nullable, and apply default value if it's null, but that would remove visibility for user regarding what number we use by default.

---
In reply to: [251147195](https://github.com/dotnet/machinelearning/pull/2241#discussion_r251147195) [](ancestors = 251147195)

PR #2243, opened 2019-01-25
Public Interface of RegressionTree and TreeEnsemble

### LINKED ISSUE #1960: Make FastTree/LightGBM learned model suitable for public consumption
Related to #1901. Regarding FastTree/LightGBM we may want to refine the public prediction/learned model surface a bit more.

The structures exposed here, including [`TreeEnsemble`](https://github.com/dotnet/machinelearning/blob/00577c06ff99ad07006ca8c58cf974943cb1971f/src/Microsoft.ML.FastTree/TreeEnsemble/TreeEnsemble.cs#L21) but especially [`RegressionTree`](https://github.com/dotnet/machinelearning/blob/00577c06ff99ad07006ca8c58cf974943cb1971f/src/Microsoft.ML.FastTree/TreeEnsemble/RegressionTree.cs#L22), are dual use structures in the sense that they are used for both training and prediction. This means it is irretreivably mutable (as it must be during training) even during prediction. It is polymorphic in its structure, in the sense that various structures are active and populated during training than are used during prediction -- worse, in some cases the *same* structures are used, but have distinct meanings. (E.g., the feature indices during training are distinct from those used during prediction, since if FastTree determines it can only use a subset of features, it will during training not even consider the unusable features to exist, until it is done with training and maps the structure back to the original feature space so it is usable by ML.NET.) There are other serious but (in context) comparatively minor notes, like not being either sealed or abstract, disobeying more .NET naming and implementation guidelines than I care to enumerate, etc.

Yet, we cannot avoid exposing *some* structure, since of course people want to inspect the trees they have learnt. And, we use this structure to represent the trained ensembles of both the FastTree and LightGBM learner, and, if we ever clean up our XGBoost wrapper so it can be open sourced (or, perhaps even, someone implements a fresh wrapper for us), maybe it will wind up using exactly that same structure. This will also serve them better since this class's structure is, for the same reasons enumerated above, incomprehensible, since about half the members on it are things people shouldn't use. (Since they are used as explained above exclusively during training.)

What should be done instead is the following: during training there is an internal class, in fact, probably more or less the same class that exists now, with a separate *immutable* class that is exposed as the model during training, created out of instances of this class. This also implies a separate training/prediction structure of what is currently called `TreeEnsemble`, and of course the quantile trees will have to be fashioned in some way. Much of the code supporting prediction and the standard ML.NET interfaces would move into *that* set of immutable classes.

wschin:
This PR proposes some changes to make RegressionTree and TreeEnsemble not mutable to users. Our strategy is

1. Create wrapper classes `RegressionTree` and `TreeEnsemble` over their internal relatives. Those wrapper classes are not mutable.
2. Internalize everything (around `InternalRegressionTree` and `InternalTreeEnsemble`) which should not be public.
3. Internalize public constructors such as `FastForestClassificationModelParameters`, `FastForestRegressionModelParameters`, etc.
4. Some cleaning. For example, changing `Float` to `float` and removing unused `using` statements.

Hopefully this will fix #1960.


src/Microsoft.ML.LightGBM/LightGbmRankingTrainer.cs
```diff
@@ -50,7 +50,8 @@ private static VersionInfo GetVersionInfo()
         protected override uint VerCategoricalSplitSerialized => 0x00010005;
         public override PredictionKind PredictionKind => PredictionKind.Ranking;
 
-        public LightGbmRankingModelParameters(IHostEnvironment env, TreeEnsemble trainedEnsemble, int featureCount, string innerArgs)
+        [BestFriend]
+        internal LightGbmRankingModelParameters(IHostEnvironment env, TreeEnsemble trainedEnsemble, int featureCount, string innerArgs)
```
TomFinley: Hi Wei-Sheng, it is not clear to me that outside code is using this, could you confirm? If it is not I might prefer to make this just `internal` and not add the `[BestFriend]` attribute. I have similar comments on other things that were made `internal` while also being `[BestFriend]`ed.

wschin: Sorry Tom. I supposed to pushing a re-doing iteration to hide those things before you seeing them... but got blocked by hiding more and more staff. This should look much better hopefully later today.

---
In reply to: [251181975](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251181975) [](ancestors = 251181975)

src/Microsoft.ML.FastTree/TreeEnsemble/RegressionTree.cs
```diff
@@ -14,25 +14,73 @@
...
     {
-        private double _maxOutput;
+        /// <summary>
+        /// <see cref="RegressionTreeView"/> is an immutable wrapper over <see cref="_tree"/> for exposing some tree's attribute to users.
+        /// </summary>
+        private readonly RegressionTree _tree;
+
+        // Immutable accessors to the underlying tree structure. They help users to inspect tree learned using, for example,
+        // gradient-boosting decision tree and random forest.
+        public ReadOnlySpan<int> LteChild => _tree.LteChild;
```
TomFinley: >_tree.LteChild [](start = 45, length = 14)

Could we check something about this? One thing I am concerned about is that even if a tree doesn't grow to "full size" (e.g., because of no candidate splits), the underlying tree used in training still has the full number of potential sizes allocated. So you ask for "I want 20 leaves," this will still be of size 19, even if ultimately only 10 splits are performed. One way to write a test for this "underbuilt" scenario is to use a small dataset (e.g., breast-cancer, which has under 700 leaves), use 100 for min docs per leaf, but ask for 10 leaves. (Which is impossible.) So I expect that it will produce fewer than that number of leaves.

The reason I believe this is if I view this code:

[https://github.com/dotnet/machinelearning/blob/8cdfa613c363b45de4b61abef8790943243ba818/src/Microsoft.ML.FastTree/TreeEnsemble/RegressionTree.cs#L82-L96](https://github.com/dotnet/machinelearning/blob/8cdfa613c363b45de4b61abef8790943243ba818/src/Microsoft.ML.FastTree/TreeEnsemble/RegressionTree.cs#L82-L96)

I see these arrays are allocated to the max size that *might* be needed, but I do not see that it is ever downgraded. So you might be presenting "working memory."

src/Microsoft.ML.FastTree/TreeEnsemble/RegressionTree.cs
```diff
@@ -14,25 +14,73 @@
...
+        public ReadOnlySpan<int> GtChild => _tree.GtChild;
+        public ReadOnlySpan<int> NumericalSplitFeatureIndexes => _tree.SplitFeatures;
+        public ReadOnlySpan<float> NumericalSplitThresholds => _tree.RawThresholds;
+        public ReadOnlySpan<bool> CategoricalSplitFlags => _tree.CategoricalSplit;
+        public IReadOnlyList<IReadOnlyList<int>> CategoricalSplitFeatures => _tree.CategoricalSplitFeatures;
+        public IReadOnlyList<IReadOnlyList<int>> CategoricalSplitFeatureRanges => _tree.CategoricalSplitFeatureRanges;
```
TomFinley: You're going to have to document this *really* carefully, and a proper implementation will be quite annoying. This sort of advanced structure is something that might be best achieved using an actual method, e.g., get features and ranges for this node, if applicable.

src/Microsoft.ML.FastTree/TreeEnsemble/RegressionTree.cs
```diff
@@ -14,25 +14,73 @@
...
+            LeafSampleWeights = null;
+            if (tree is QuantileRegressionTree)
+            {
+                LeafSamples = ((QuantileRegressionTree)tree).ExtractLeafSamples();
+                LeafSampleWeights = ((QuantileRegressionTree)tree).ExtractLeafSampleWeights();
+            }
+        }
+    }
+
+    public class RegressionTree
```
TomFinley: >public [](start = 4, length = 6)

Probably internal.

src/Microsoft.ML.FastTree/TreeEnsemble/RegressionTree.cs
```diff
@@ -14,25 +14,73 @@
 using Microsoft.ML.Model;
 using Microsoft.ML.Model.Pfa;
 using Newtonsoft.Json.Linq;
-using Float = System.Single;
 
 namespace Microsoft.ML.Trainers.FastTree.Internal
 {
-    public class RegressionTree
+    public class RegressionTreeView
```
TomFinley: >RegressionTreeView [](start = 17, length = 18)

From an external perspective this name is problematic. Sure you've architectured it as a wrapper, but the name "view" asks people to wonder, a view of what? If that regression tree below were public you might have a case, but that is something we want to hide. I don't think you're going to get out of this without changing the name of the internal thing.

src/Microsoft.ML.FastTree/Training/EnsembleCompression/LassoBasedEnsembleCompressor.cs
```diff
@@ -15,7 +15,8 @@ namespace Microsoft.ML.Trainers.FastTree.Internal
     /// https://www-stat.stanford.edu/~hastie/Papers/glmnet.pdf
     /// </summary>
     /// <remarks>Author was Yasser Ganjisaffar during his internship.</remarks>
-    public class LassoBasedEnsembleCompressor : IEnsembleCompressor<short>
+    [BestFriend]
```
TomFinley: >[BestFriend] [](start = 4, length = 12)

While this probably should be internal, not best friend probably?

src/Microsoft.ML.FastTree/TreeEnsemble/RegressionTree.cs
wschin: Thanks for this nice suggestion. You will see
```csharp
        /// <summary>
        /// Return the training labels falling into the specified leaf.
        /// </summary>
        /// <param name="leafIndex">The index of the specified leaf.</param>
        /// <returns>Training labels</returns>
        public ReadOnlySpan<double> GetLeafSamplesAt(int leafIndex) => new ReadOnlySpan<double>(_leafSamples[leafIndex]);

        /// <summary>
        /// Return the weights for training labels falling into the specified leaf. If <see cref="GetLeafSamplesAt"/> and <see cref="GetLeafSampleWeightsAt"/>
        /// use the same input, the i-th returned value of this function is the weight of the i-th label in <see cref="GetLeafSamplesAt"/>.
        /// </summary>
        /// <param name="leafIndex">The index of the specified leaf.</param>
        /// <returns>Training labels' weights</returns>
        public ReadOnlySpan<double> GetLeafSampleWeightsAt(int leafIndex) => new ReadOnlySpan<double>(_leafSampleWeights[leafIndex]);
```
in the next iteration.

---
In reply to: [251182480](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251182480) [](ancestors = 251182480)

src/Microsoft.ML.FastTree/TreeEnsemble/RegressionTree.cs
wschin: How about `DecisionTreeRegressor` for this public class (inspired by ONNX)?

---
In reply to: [251182801](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251182801) [](ancestors = 251182801)

src/Microsoft.ML.FastTree/TreeEnsemble/RegressionTree.cs
wschin: It will be also best-friend because of being used in tests.

---
In reply to: [251182509](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251182509) [](ancestors = 251182509)

src/Microsoft.ML.FastTree/TreeEnsemble/TreeEnsemble.cs
```diff
@@ -16,7 +16,37 @@
 
 namespace Microsoft.ML.Trainers.FastTree.Internal
 {
-    public class TreeEnsemble
+    public class TreeEnsembleModel
```
wschin: >TreeEnsembleModel [](start = 17, length = 17)

Would `TreeRegressorCollection` be a better name?

src/Microsoft.ML.FastTree/Training/EnsembleCompression/LassoBasedEnsembleCompressor.cs
wschin: Yes, you're right.

---
In reply to: [251186572](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251186572) [](ancestors = 251186572)

src/Microsoft.ML.FastTree/TreeEnsemble/RegressionTree.cs
wschin: Everything wrapper on pre-allocated array with size [maxLeaves -1] will be created via `new ReadOnlySpan<...>(..., 0, _tree.NumNodes)`; for example,
```csharp
        public ReadOnlySpan<int> LteChild => new ReadOnlySpan<int>(_tree.LteChild, 0, _tree.NumNodes); 
        public ReadOnlySpan<int> GtChild => new ReadOnlySpan<int>(_tree.GtChild, 0, _tree.NumNodes); 
        public ReadOnlySpan<int> NumericalSplitFeatureIndexes => new ReadOnlySpan<int>(_tree.SplitFeatures, 0, _tree.NumNodes); 
        public ReadOnlySpan<float> NumericalSplitThresholds => new ReadOnlySpan<float>(_tree.RawThresholds, 0, _tree.NumNodes);
```

---
In reply to: [251182399](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251182399) [](ancestors = 251182399)

src/Microsoft.ML.FastTree/FastTreeTweedie.cs
```diff
@@ -129,7 +129,7 @@ protected override ObjectiveFunctionBase ConstructObjFunc(IChannel ch)
             return new ObjectiveImpl(TrainSet, Args);
         }
 
-        protected override OptimizationAlgorithm ConstructOptimizationAlgorithm(IChannel ch)
+        internal override OptimizationAlgorithm ConstructOptimizationAlgorithm(IChannel ch)
```
TomFinley: >internal [](start = 8, length = 8)

`private protected` is made for cases such as this.

src/Microsoft.ML.FastTree/FastTreeRegression.cs
```diff
@@ -124,7 +124,7 @@ protected override ObjectiveFunctionBase ConstructObjFunc(IChannel ch)
             return new ObjectiveImpl(TrainSet, Args);
         }
 
-        protected override OptimizationAlgorithm ConstructOptimizationAlgorithm(IChannel ch)
+        internal override OptimizationAlgorithm ConstructOptimizationAlgorithm(IChannel ch)
```
TomFinley: >internal [](start = 8, length = 8)

Another candidate for `private protected`.

src/Microsoft.ML.FastTree/FastTreeRegression.cs
```diff
@@ -494,7 +494,7 @@ private static FastTreeRegressionModelParameters Create(IHostEnvironment env, Mo
         public override PredictionKind PredictionKind => PredictionKind.Regression;
     }
 
-    public static partial class FastTree
+    public static partial class FastTreeEntryPoint
```
TomFinley: >FastTreeEntryPoint [](start = 32, length = 18)

FYI I have a change to make this internal, so perhaps this will not be necessary.

wschin: Ok. I will handle some merge conflicts but let me keep this for now for building it on CI.

---
In reply to: [251652346](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251652346) [](ancestors = 251652346)

src/Microsoft.ML.FastTree/FastTreeRegression.cs
wschin: All `ConstructOptimizationAlgorithm` are `private protected` now because we have
```csharp
    public abstract class FastTreeTrainerBase<TArgs, TTransformer, TModel>
    ....
        private protected abstract OptimizationAlgorithm ConstructOptimizationAlgorithm(IChannel ch);
    ....
```

---
In reply to: [251652163](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251652163) [](ancestors = 251652163)

TomFinley:
Adding @sfilipi , @zeahmed, and @glebuk since they often have thoughts about APIs.

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// <see cref="TreeRegressor"/> is an immutable wrapper over <see cref="_tree"/> for exposing some tree's attribute to users.
+        /// </summary>
+        private readonly RegressionTree _tree;
+
+        /// <summary>
+        /// Sample labels from training data. <see cref="_leafSamples"/>[i] stores the labels falling into the i-th leaf.
+        /// </summary>
+        private readonly List<double[]> _leafSamples;
+        /// <summary>
+        /// Sample labels' weights from training data. <see cref="_leafSampleWeights"/>[i] stores the weights for labels falling into the i-th leaf.
```
glebuk: >falling [](start = 121, length = 7)

ensure comments fit on once screen (<80 chars/line)

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
```diff
@@ -0,0 +1,124 @@
...
+    {
+        /// <summary>
+        /// <see cref="TreeRegressor"/> is an immutable wrapper over <see cref="_tree"/> for exposing some tree's attribute to users.
+        /// </summary>
+        private readonly RegressionTree _tree;
+
+        /// <summary>
+        /// Sample labels from training data. <see cref="_leafSamples"/>[i] stores the labels falling into the i-th leaf.
+        /// </summary>
+        private readonly List<double[]> _leafSamples;
```
glebuk: >List<double[] [](start = 25, length = 13)

Items in List<double[]> is actually mutable a structure of mutable structures.
Individual list elements and array elements can be modified. Is there a read-only structure we can use here instead?

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
TomFinley: Summary comments we've typically tried to keep at around 120 or 150 actually, from other files. 80 is a bit too harsh. (If you tried to enforce this policy the result would look ridiculous.) For markdown though we do try to keep it properly formatted to a "console window" width.

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
wschin: If we change it to `ImmutableArray`, we will have some fields with type `ReadOnlySpan` and two functions returning `ImmutableArray`. This field is private. so maybe we can keep it as it for now?

---
In reply to: [251948660](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251948660) [](ancestors = 251948660)

glebuk: Perhaps turn it into double[][]?

---
In reply to: [251953838](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251953838) [](ancestors = 251953838,251948660)

src/Microsoft.ML.FastTree/Representation/TreeRegressorCollection.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    /// <summary>
+    /// A list of <see cref="TreeRegressor"/>. To compute the output value of a <see cref="TreeRegressorCollection"/>, we need to compute
+    /// the output values of all trees in <see cref="Trees"/>, scale those values via <see cref="TreeWeights"/>, and finally sum the scaled
+    /// values and <see cref="Bias"/> up.
+    /// </summary>
+    public class TreeRegressorCollection
+    {
+        [BestFriend]
+        internal readonly TreeEnsemble TreeEnsemble; // It's a best friend for being accessed from LightGBM.
```
glebuk: >It's a best friend for being accessed from LightGBM. [](start = 56, length = 52)

convert to xml comment?

src/Microsoft.ML.FastTree/Training/EnsembleCompression/LassoBasedEnsembleCompressor.cs
```diff
@@ -533,7 +533,7 @@ private unsafe void LoadTargets(double[] trainScores, int bestIteration)
             }
         }
 
-        public bool Compress(IChannel ch, TreeEnsemble ensemble, double[] trainScores, int bestIteration, int maxTreesAfterCompression)
+        bool IEnsembleCompressor<short>.Compress(IChannel ch, TreeEnsemble ensemble, double[] trainScores, int bestIteration, int maxTreesAfterCompression)
```
glebuk: >bool [](start = 8, length = 4)

add explicit private access modifier everywhere.

src/Microsoft.ML.FastTree/Training/OptimizationAlgorithms/OptimizationAlgorithm.cs
```diff
@@ -95,6 +94,7 @@ public ScoreTracker GetScoreTracker(string name, Dataset set, double[] initScore
 
         protected abstract ScoreTracker ConstructScoreTracker(string name, Dataset set, double[] initScores);
 
+        // Regularize a regression tree with smoothing paramter alpha
```
glebuk: >Regularize a regression tree with smoothing paramter alpha [](start = 11, length = 58)

convert to xml comment for intellisense.

src/Microsoft.ML.FastTree/Training/Parallel/SingleTrainer.cs
```diff
@@ -19,37 +19,37 @@ namespace Microsoft.ML.Trainers.FastTree
     using LeafSplitCandidates = Internal.LeastSquaresRegressionTreeLearner.LeafSplitCandidates;
     using SplitInfo = Internal.LeastSquaresRegressionTreeLearner.SplitInfo;
 
-    public sealed class SingleTrainer : IParallelTraining
+    internal sealed class SingleTrainer : IParallelTraining
     {
-        public void CacheHistogram(bool isSmallerLeaf, int featureIdx, int subfeature, SufficientStatsBase sufficientStatsBase, bool hasWeights)
+        void IParallelTraining.CacheHistogram(bool isSmallerLeaf, int featureIdx, int subfeature, SufficientStatsBase sufficientStatsBase, bool hasWeights)
```
glebuk: private void

src/Microsoft.ML.FastTree/Training/EnsembleCompression/LassoBasedEnsembleCompressor.cs
wschin: This will not work for internal interface.

---
In reply to: [251971388](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251971388) [](ancestors = 251971388)

src/Microsoft.ML.FastTree/Training/Parallel/SingleTrainer.cs
wschin: No. This is an explicit implementation of an internal interface.

---
In reply to: [251976714](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251976714) [](ancestors = 251976714)

src/Microsoft.ML.FastTree/Training/OptimizationAlgorithms/OptimizationAlgorithm.cs
wschin: Sure.

---
In reply to: [251971688](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251971688) [](ancestors = 251971688)

src/Microsoft.ML.FastTree/Representation/TreeRegressorCollection.cs
wschin: Sure.

---
In reply to: [251969703](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251969703) [](ancestors = 251969703)

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// <see cref="LeafValues"/>[i] is the learned value at the i-th leaf.
+        /// </summary>
+        public ReadOnlySpan<double> LeafValues => new ReadOnlySpan<double>(_tree.LeafValues, 0, _tree.NumLeaves);
+        /// <summary>
+        /// Return categorical thresholds used at node indexed by nodeIndex. If the considered input feature does NOT
+        /// matche any of values returned by <see cref="GetCategoricalSplitFeaturesAt(int)"/>, we call it a
+        /// less-than-threshold event and therefore <see cref="LteChild"/>[nodeIndex] is the child node that input
+        /// should go next. The returned value is valid only if <see cref="CategoricalSplitFlags"/>[nodeIndex] is true.
+        /// </summary>
+        public ReadOnlySpan<int> GetCategoricalSplitFeaturesAt(int nodeIndex) => new ReadOnlySpan<int>(_tree.CategoricalSplitFeatures[nodeIndex]);
```
codemzs: >public ReadOnlySpan<int> GetCategoricalSplitFeaturesAt(int nodeIndex) => new ReadOnlySpan<int>(_tree.CategoricalSplitFeatures[nodeIndex]); [](start = 8, length = 138)

Can we cache these values?

src/Microsoft.ML.FastTree/TreeEnsemble/QuantileRegressionTree.cs
```diff
@@ -91,6 +97,39 @@ public void SetLabelsDistribution(double[] labelsDistribution, double[] weights)
...
+        {
+            leafSamples = new double[NumLeaves][];
+            leafSampleWeights = new double[NumLeaves][];
+            // If there is no training labels stored, we view the i-th leaf value as the only label stored at the i-th leaf.
+            var sampleCountPerLeaf = _labelsDistribution != null ? _labelsDistribution.Length / NumLeaves : 1;
+            for (int i = 0; i < NumLeaves; ++i)
+            {
+                var samplesPerLeaf = new List<double>();
+                var sampleWeightsPerLeaf = new List<double>();
+                for (int j = 0; j < sampleCountPerLeaf; ++j)
```
codemzs: I think you can make them as an array and pre allocate before you enter the loop.

test/Microsoft.ML.StaticPipelineTesting/TreeRepresentation.cs
```diff
@@ -0,0 +1,89 @@
...
+                c => (label: c.LoadFloat(11), features: c.LoadFloat(0, 10)),
+                separator: ';', hasHeader: true);
+
+            FastTreeRegressionModelParameters pred = null;
+
+            var est = reader.MakeNewEstimator()
+                .Append(r => (r.label, score: catalog.Trainers.FastTree(r.label, r.features,
+                    numTrees: 10,
+                    numLeaves: 5,
+                    onFit: (p) => { pred = p; })));
```
zeahmed: My understanding is that we are no longer doing any development on Static API. Do we need test using Static API?

test/Microsoft.ML.StaticPipelineTesting/TreeRepresentation.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            // Values in LeafValues:
+            // [0]	40.159015006449692	double
+            // [1]	80.434805844435061	double
+            // [2]	57.072130551545513	double
+            // [3]	82.898710076162757	double
+            // [4]	104.17547955322266	double
+            Assert.Equal(5, trees[0].LeafValues.Length);
+            Assert.Equal(40.159015006449692, trees[0].LeafValues[0], 6);
+            Assert.Equal(104.17547955322266, trees[0].LeafValues[4], 6);
```
zeahmed: Every in this test you are testing for first and last value only. Any specific reason for doing that?

test/Microsoft.ML.StaticPipelineTesting/TreeRepresentation.cs
```diff
@@ -0,0 +1,89 @@
...
+            Assert.Null(pred);
+            var model = pipe.Fit(dataSource);
+            Assert.NotNull(pred);
+
+            var treeCollection = pred.TrainedTreeCollection;
+            Assert.Equal(0, treeCollection.Bias);
+            Assert.Equal(10, treeCollection.Trees.Count);
+            Assert.Equal(10, treeCollection.TreeWeights.Count);
+
+            var trees = treeCollection.Trees;
```
zeahmed: >trees [](start = 16, length = 5)

Please check the number of trees as well.

wschin: Checked above at line 45.

---
In reply to: [252029977](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252029977) [](ancestors = 252029977)

zeahmed:

I dont see any change in this file. Is codeflow making me fool?

---
Refers to: test/Microsoft.ML.Core.Tests/Microsoft.ML.Core.Tests.csproj:1 in feae8e6. [](commit_id = feae8e6ec45f520826e6bfe9e4da12a6a08b5093, deletion_comment = False)

test/Microsoft.ML.StaticPipelineTesting/TreeRepresentation.cs
wschin: It will become very crowded in this function.

---
In reply to: [252029276](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252029276) [](ancestors = 252029276)

test/Microsoft.ML.StaticPipelineTesting/TreeRepresentation.cs
wschin: This test is not for static API, it's for tree's public area.

---
In reply to: [252028640](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252028640) [](ancestors = 252028640)

src/Microsoft.ML.FastTree/TreeEnsemble/QuantileRegressionTree.cs
wschin: No problem.

---
In reply to: [252027300](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252027300) [](ancestors = 252027300)

src/Microsoft.ML.FastTree/TreeEnsemble/TreeEnsemble.cs
```diff
@@ -357,9 +358,25 @@ internal void GetFeatureContributions(in VBuffer<float> features, ref VBuffer<fl
 
             builder.GetResult(ref contribs);
         }
+
+        public int GetMaxFeatureIndex()
```
zeahmed: >GetMaxFeatureIndex [](start = 19, length = 18)

Need to have xml documentation for anything that is public.

src/Microsoft.ML.FastTree/TreeEnsemble/RegressionTree.cs
```diff
@@ -14,25 +14,25 @@
 using Microsoft.ML.Model;
 using Microsoft.ML.Model.Pfa;
 using Newtonsoft.Json.Linq;
-using Float = System.Single;
 
 namespace Microsoft.ML.Trainers.FastTree.Internal
 {
-    public class RegressionTree
+    /// Note that <see cref="RegressionTree"/> is shared between FastTree and LightGBM asemblies,
```
zeahmed: >asemblies [](start = 87, length = 9)

typo

src/Microsoft.ML.FastTree/TreeEnsemble/TreeEnsemble.cs
wschin: It's an internal class but of course, we should doc it.

---
In reply to: [252031398](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252031398) [](ancestors = 252031398)

wschin:

I don't know..

---
In reply to: [458719645](https://github.com/dotnet/machinelearning/pull/2243#issuecomment-458719645) [](ancestors = 458719645)

---
Refers to: test/Microsoft.ML.Core.Tests/Microsoft.ML.Core.Tests.csproj:1 in feae8e6. [](commit_id = feae8e6ec45f520826e6bfe9e4da12a6a08b5093, deletion_comment = False)

src/Microsoft.ML.FastTree/TreeEnsemble/QuantileRegressionTree.cs
```diff
@@ -2,19 +2,25 @@
...
 using Float = System.Single;
 
 namespace Microsoft.ML.Trainers.FastTree.Internal
 {
-    public class QuantileRegressionTree : RegressionTree
+    internal class QuantileRegressionTree : RegressionTree
     {
-        // Holds the labels of samped instances for this tree
+        /// <summary>
+        /// Holds the labels of samped instances for this tree. This value can be null when training, for example, random forest (FastForest).
```
zeahmed: >samped [](start = 32, length = 6)

typo

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// <see cref="CategoricalSplitFlags"/>[nodeIndex] is true.
+        /// </summary>
+        public ReadOnlySpan<int> GetCategoricalCategoricalSplitFeatureRangeAt(int nodeIndex) => new ReadOnlySpan<int>(_tree.CategoricalSplitFeatureRanges[nodeIndex]);
+
+        /// <summary>
+        /// Return the training labels falling into the specified leaf.
+        /// </summary>
+        /// <param name="leafIndex">The index of the specified leaf.</param>
+        /// <returns>Training labels</returns>
+        public ReadOnlySpan<double> GetLeafSamplesAt(int leafIndex) => new ReadOnlySpan<double>(_leafSamples[leafIndex]);
```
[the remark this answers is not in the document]
wschin: ReadOnlySpan can't be a field or auto-implemented property.

---
In reply to: [252024429](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252024429) [](ancestors = 252024429)

src/Microsoft.ML.FastTree/TreeEnsemble/QuantileRegressionTree.cs
```diff
@@ -91,6 +97,39 @@ public void SetLabelsDistribution(double[] labelsDistribution, double[] weights)
             _instanceWeights = weights;
         }
 
+        /// <summary>
+        /// Copy training examples' labels and their weights to external variables.
+        /// </summary>
+        /// <param name="leafSamples">List of label collections. The type of a collction is a double array. The i-th label collection contains training examples' labels falling into the i-th leaf.</param>
+        /// <param name="leafSampleWeights">List of labels' weight collections. The type of a collction is a double array. The i-th collection contains weights of labels falling into the i-th leaf.
```
zeahmed: >collction [](start = 94, length = 9)

typo. See line 103 as well.

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// should go next. The returned value is valid only if <see cref="CategoricalSplitFlags"/>[nodeIndex] is true.
+        /// </summary>
+        public ReadOnlySpan<int> GetCategoricalSplitFeaturesAt(int nodeIndex) => new ReadOnlySpan<int>(_tree.CategoricalSplitFeatures[nodeIndex]);
+        /// <summary>
+        /// Return categorical thresholds' range used at node indexed by nodeIndex. A categorical split at node indexed
+        /// by nodeIndex can consider multiple consecutive input features at one time; their range is specified by
+        /// <see cref="GetCategoricalCategoricalSplitFeatureRangeAt(int)"/>.  The returned value is valid only if
+        /// <see cref="CategoricalSplitFlags"/>[nodeIndex] is true.
+        /// </summary>
+        public ReadOnlySpan<int> GetCategoricalCategoricalSplitFeatureRangeAt(int nodeIndex) => new ReadOnlySpan<int>(_tree.CategoricalSplitFeatureRanges[nodeIndex]);
```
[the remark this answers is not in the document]
wschin: Sure.

---
In reply to: [252024142](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252024142) [](ancestors = 252024142)

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
wschin: I guess we can't do this for ReadOnlySpan.

---
In reply to: [252021978](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252021978) [](ancestors = 252021978)

test/Microsoft.ML.StaticPipelineTesting/TreeRepresentation.cs
zeahmed: ok.

---
In reply to: [252030449](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252030449) [](ancestors = 252030449,252029977)

src/Microsoft.ML.FastTree/TreeEnsemble/RegressionTree.cs
wschin: Thanks! Let's do assemblies.

---
In reply to: [252031769](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252031769) [](ancestors = 252031769)

src/Microsoft.ML.FastTree/TreeEnsemble/QuantileRegressionTree.cs
wschin: Ok.

---
In reply to: [252033149](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252033149) [](ancestors = 252033149)

test/Microsoft.ML.StaticPipelineTesting/TreeRepresentation.cs
zeahmed: Yes, but it used Static API and I think you should write test using dynamic API (using MLContext) instead of Static API. This is what we decided for going forward.

---
In reply to: [252031278](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252031278) [](ancestors = 252031278,252028640)

src/Microsoft.ML.FastTree/TreeEnsemble/QuantileRegressionTree.cs
wschin: Nice catch.

---
In reply to: [252034011](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252034011) [](ancestors = 252034011)

test/Microsoft.ML.StaticPipelineTesting/TreeRepresentation.cs
wschin: As long as they are testing the same thing, I don't see a reason to create one extra test.

---
In reply to: [252036849](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252036849) [](ancestors = 252036849,252031278,252028640)

zeahmed: I am not asking for write another test. I saying this test should use Dynamic API.

---
In reply to: [252040126](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252040126) [](ancestors = 252040126,252036849,252031278,252028640)

wschin: I don't think this is the right way to do the test here. We don't expose ModelParameters through dynamic in this specific case to users.

---
In reply to: [252041372](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252041372) [](ancestors = 252041372,252040126,252036849,252031278,252028640)

zeahmed: can you see this test
[https://github.com/dotnet/machinelearning/blob/e3830910531f00013c27391914233a085a1394a4/test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs#L28](https://github.com/dotnet/machinelearning/blob/e3830910531f00013c27391914233a085a1394a4/test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs#L28)

The only thing changes in your case is line 31-35

---
In reply to: [252055619](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252055619) [](ancestors = 252055619,252041372,252040126,252036849,252031278,252028640)

test/Microsoft.ML.StaticPipelineTesting/TreeRepresentation.cs
wschin: In Iteration 18, we test all elements using built-in to test enumerables if possible and loops otherwise.

---
In reply to: [252031029](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252031029) [](ancestors = 252031029,252029276)

test/Microsoft.ML.StaticPipelineTesting/TreeRepresentation.cs
zeahmed: We should then file a bug that its not possible do this scenario with dynamic API.

---
In reply to: [252060908](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252060908) [](ancestors = 252060908,252059592,252055619,252041372,252040126,252036849,252031278,252028640)

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
```diff
@@ -0,0 +1,136 @@
...
+using System.Linq;
+using Microsoft.ML.Trainers.FastTree.Internal;
+
+namespace Microsoft.ML.FastTree.Representation
+{
+    /// <summary>
+    /// A container class for exposing <see cref="RegressionTree"/>'s attributes to users.
+    /// This class should not be mutable, so it contains a lot of read-only members.
+    /// </summary>
+    public class TreeRegressor
```
Ivanidzo4ka: >TreeRegressor [](start = 17, length = 13)

Did anyone yet ask you about sample file in [https://github.com/dotnet/machinelearning/tree/master/docs/samples/Microsoft.ML.Samples/Dynamic](https://github.com/dotnet/machinelearning/tree/master/docs/samples/Microsoft.ML.Samples/Dynamic) ?
Would be nice to put  
```
/// <example>
        /// <format type="text/markdown">
        /// <![CDATA[
        ///  [!code-csharp[FastTree](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/TreeRegressor.cs)]
        /// ]]></format>
```
here so user can see how to use this class in documentation.

src/Microsoft.ML.FastTree/Representation/TreeRegressorCollection.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Microsoft.ML.Trainers.FastTree.Internal;
+
+namespace Microsoft.ML.FastTree.Representation
+{
+    /// <summary>
+    /// A list of <see cref="TreeRegressor"/>. To compute the output value of a <see cref="TreeRegressorCollection"/>, we need to compute
+    /// the output values of all trees in <see cref="Trees"/>, scale those values via <see cref="TreeWeights"/>, and finally sum the scaled
+    /// values and <see cref="Bias"/> up.
```
Ivanidzo4ka: I don't see such behavior in our code.
[https://github.com/dotnet/machinelearning/blob/b4c10662706589426a0a870304e3088dec8077ac/src/Microsoft.ML.FastTree/TreeEnsemble/TreeEnsemble.cs#L257](https://github.com/dotnet/machinelearning/blob/b4c10662706589426a0a870304e3088dec8077ac/src/Microsoft.ML.FastTree/TreeEnsemble/TreeEnsemble.cs#L257)
From what I see, we just sum up tree output.
And I don't see any usage of Bias or TreeWeight in code other than setter, or output as PFA or INI.

Not sure is it right or not.
@tfinley@gmail.com  Probably should knew more.
But if we don't use Weights and Bias, I concern we need need to expose them.

test/Microsoft.ML.StaticPipelineTesting/TreeRepresentation.cs
```diff
@@ -0,0 +1,89 @@
+﻿using Microsoft.ML.Data;
+using Microsoft.ML.RunTests;
+using Microsoft.ML.StaticPipe;
+using Microsoft.ML.Trainers.FastTree;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.ML.StaticPipelineTesting
+{
+    public sealed class TreeRepresentation : BaseTestClassWithConsole
```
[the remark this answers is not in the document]
wschin: I guess we can't have an example until issue 2319 is resolved ([https://github.com/dotnet/machinelearning/issues/2319](https://github.com/dotnet/machinelearning/issues/2319)).

---
In reply to: [252086597](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252086597) [](ancestors = 252086597,252085792)

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
wschin: Not yet until we have address [issue 2319](https://github.com/dotnet/machinelearning/issues/2319).

---
In reply to: [252087326](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252087326) [](ancestors = 252087326)

test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -3552,7 +3552,7 @@ public void EntryPointTreeLeafFeaturizer()
                 Column = new[] { new ColumnConcatenatingTransformer.Column { Name = "Features", Source = new[] { "Categories", "NumericFeatures" } } }
             });
 
-            var fastTree = FastTree.TrainBinary(Env, new FastTreeBinaryClassificationTrainer.Options
+            var fastTree = Trainers.FastTree.FastTree.TrainBinary(Env, new FastTreeBinaryClassificationTrainer.Options
```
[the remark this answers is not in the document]
wschin: If I remove it, code won't be compiled.

---
In reply to: [252085507](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252085507) [](ancestors = 252085507)

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
wschin: Done.

---
In reply to: [251959736](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251959736) [](ancestors = 251959736,251949011)

src/Microsoft.ML.FastTree/FastTreeRegression.cs
wschin: It'd be nice to make it general. Having the same name of class and namespace doesn't look very good to me.

---
In reply to: [251949357](https://github.com/dotnet/machinelearning/pull/2243#discussion_r251949357) [](ancestors = 251949357)

src/Microsoft.ML.FastTree/TreeEnsemble/RegressionTree.cs
TomFinley: But it doesn't perform regression, so calling it a `Regressor` makes no sense? How about just `DecisionTree`?

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
```diff
@@ -0,0 +1,169 @@
+﻿using System;
+using System.Linq;
+using Microsoft.ML.Trainers.FastTree.Internal;
+
+namespace Microsoft.ML.FastTree.Representation
```
TomFinley: >Microsoft.ML.FastTree.Representation [](start = 10, length = 36)

I do not like this namespace. Someone is just working with `FastTree`, why are they suddenly hit with this thing "representation?"

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
```diff
@@ -0,0 +1,169 @@
...
+{
+    /// <summary>
+    /// A container class for exposing <see cref="RegressionTree"/>'s attributes to users.
+    /// This class should not be mutable, so it contains a lot of read-only members.
+    /// </summary>
+    public class TreeRegressor
```
TomFinley: >public class TreeRegressor [](start = 4, length = 26)

Classes should either be sealed or abstract.

TomFinley: >TreeRegressor [](start = 17, length = 13)

As already covered, this is *not* a tree regressor for the simple reason that it does not perform regression. It is the *result* of performing regression. The name `RegressionTree` was appropriate. Please rename the internal class something else appropriate.

src/Microsoft.ML.FastTree/Representation/TreeRegressorCollection.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// </summary>
+    public class TreeRegressorCollection
```
TomFinley: >TreeRegressorCollection [](start = 17, length = 23)

I do not like this calling something a collection if it does not implement `ICollection`. I am not BTW suggesting that you name it collection, since it differs from a collection in many important respects. Rather let's name it what it has always been named... `TreeEnsemble`. Was anything wrong with that name? Why are we changing it?

src/Microsoft.ML.FastTree/Representation/TreeRegressorCollection.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// When doing prediction, this is a value added to the weighted sum of all <see cref="Trees"/>' outputs.
+        /// </summary>
+        public double Bias { get; }
+
+        /// <summary>
+        /// <see cref="TreeWeights"/>[i] is the i-th <see cref="TreeRegressor"/>'s weight in this <see cref="TreeRegressorCollection"/>.
+        /// </summary>
+        public IReadOnlyList<double> TreeWeights { get; }
```
TomFinley: >IReadOnlyList<double> [](start = 15, length = 21)

Here we are using `IReadOnlyList`. Sometimes we are using `ReadOnlySpan`. How about we pick *one*? (I have something of a preference for `IReadOnlyList` since spans are stack-only concepts and it's unclear to me that that is appropriate in this case.)

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// Sample labels from training data. <see cref="_leafSamples"/>[i] stores the labels falling into the
+        /// i-th leaf.
+        /// </summary>
+        private readonly double[][] _leafSamples;
+        /// <summary>
+        /// Sample labels' weights from training data. <see cref="_leafSampleWeights"/>[i] stores the weights for
+        /// labels falling into the i-th leaf. <see cref="_leafSampleWeights"/>[i][j] is the weight of
+        /// <see cref="_leafSamples"/>[i][j].
+        /// </summary>
+        private readonly double[][] _leafSampleWeights;
```
TomFinley: So is this an attempt to reflect the polymorphism present in what is currently the internal class `RegressionTree`?

TomFinley: Let's not fake polymorphism. I think you ought to have an abstract class, possibly, `RegressionTreeBase`. It works over regression tree. You have two subclasses: `RegressionTree` (whose implementation is almost entirely empty), and `QuantileRegressionTree`, the latter working over quantile regression trees. Then your `TreeEnsemble` is generic over the type of tree, which must descend from `RegressionTreeBase`. E.g.: `public sealed class TreeEnsemble<TTree> where TTree : RegressionTreeBase`. Then the specific model parameter objects have as something accessible either `TreeEnsemble<RegressionTree>` or `TreeEnsemble<QuantileRegressionTree>`.

Note that I am using the desirable names for these things, not this "regressor" or "collection" stuff.

---
In reply to: [252500060](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252500060) [](ancestors = 252500060)

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
```diff
@@ -0,0 +1,169 @@
...
+        public ReadOnlySpan<float> NumericalSplitThresholds => new ReadOnlySpan<float>(_tree.RawThresholds, 0, _tree.NumNodes);
+        /// <summary>
+        /// Determine the types of splitting function. If <see cref="CategoricalSplitFlags"/>[i] is true, the i-th
+        /// node's uses categorical splitting function. Otherwise, traditional numerical split is used.
+        /// </summary>
+        public ReadOnlySpan<bool> CategoricalSplitFlags => new ReadOnlySpan<bool>(_tree.CategoricalSplit, 0, _tree.NumNodes);
+        /// <summary>
+        /// <see cref="LeafValues"/>[i] is the learned value at the i-th leaf.
+        /// </summary>
+        public ReadOnlySpan<double> LeafValues => new ReadOnlySpan<double>(_tree.LeafValues, 0, _tree.NumLeaves);
```
TomFinley: >ReadOnlySpan<double> [](start = 15, length = 20)

I am somewhat reluctant to use spans since they are a stack-only construct, and it is unclear to me that this will be appropriate or helpful in all cases. If it is to be a view over something, what do we think about `IReadOnlyList` as you have done elsewhere?

src/Microsoft.ML.FastTree/TreeEnsemble/TreeEnsemble.cs
wschin: Sure. I will rename TreeEnsembleModel to TreeEnsemble and TreeRegressor to RegressionTree in iteration 21.

---
In reply to: [252498510](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252498510) [](ancestors = 252498510)

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
wschin: Let me remove the `Representation`.

---
In reply to: [252498988](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252498988) [](ancestors = 252498988)

src/Microsoft.ML.FastTree/Representation/TreeRegressorCollection.cs
wschin: It will be TreeEnsemble in iteration 21.

---
In reply to: [252499668](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252499668) [](ancestors = 252499668)

src/Microsoft.ML.FastTree/Representation/TreeRegressorCollection.cs
wschin: No problem. IReadOnlyList will be the universal way to access enumerable attributes in Tree and TreeEnsemble.

---
In reply to: [252499854](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252499854) [](ancestors = 252499854)

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
wschin: Sure. Let's do IReadOnlyList in iteration 21.

---
In reply to: [252500747](https://github.com/dotnet/machinelearning/pull/2243#discussion_r252500747) [](ancestors = 252500747)

src/Microsoft.ML.FastTree/Representation/TreeRegressor.cs
wschin: Changes for addressing this comment were proposed in [this commit](https://github.com/dotnet/machinelearning/pull/2243/commits/9ffc6c8a9e1304e3554a34e1bcd70b11da00cb11). Could you please take a look?

src/Microsoft.ML.FastTree/RegressionTree.cs
```diff
@@ -0,0 +1,229 @@
...
+        //     /  \
+        // leaf1 leaf2
+        // The index of leaf starts with 1 because interally we use "-1" as the 1st leaf's index, "-2" for the 2nd leaf's index, and so on.
+        public int NumNodes => _tree.NumNodes;
+
+        internal RegressionTreeBase(InternalRegressionTree tree)
+        {
+            _tree = tree;
+
+            _lteChild = ImmutableArray.Create(_tree.LteChild, 0, _tree.NumNodes);
```
TomFinley: >ImmutableArray [](start = 24, length = 14)

This can do for now, but FYI wrapping in an immutable array creates a *copy*. Not a deal breaker, just something you ought to keep in mind.

src/Microsoft.ML.FastTree/RegressionTree.cs
```diff
@@ -0,0 +1,229 @@
...
+        /// <summary>
+        /// <see cref="LteChild"/>[i] is the i-th node's child index used when
+        /// (1) the numerical feature indexed by <see cref="NumericalSplitFeatureIndexes"/>[i] is less than the
+        /// threshold <see cref="NumericalSplitThresholds"/>[i], or
+        /// (2) the categorical features indexed by <see cref="GetCategoricalCategoricalSplitFeatureRangeAt(int)"/>'s
+        /// returned value with nodeIndex=i is NOT a sub-set of <see cref="GetCategoricalSplitFeaturesAt(int)"/> with
+        /// nodeIndex=i.
+        /// Note that the case (1) happens only when <see cref="CategoricalSplitFlags"/>[i] is true and otherwise (2)
+        /// occurs. A non-negative returned value means a node (i.e., not a leaf); for example, 2 means the 3rd node in
+        /// the underlying <see cref="_tree"/>. A negative returned value means a leaf; for example, -1 stands for the
```
TomFinley: >negative [](start = 50, length = 8)

Might help to use the word bitwise inverse (e.g., `~`) just for the benefit of people that know what that is. But we can refine the documentation later.

wschin:
@TomFinley, Many thanks! I will refine the doc and make `private protected` in another PR today.

PR #2245, opened 2019-01-25
No loader on APIs for ValueToKey/OneHotEncoding

### LINKED ISSUE #2231: Remove IDataLoader from public API surfaces for specifying data
So in "command line" world, we have things that look like this:

https://github.com/dotnet/machinelearning/blob/eed91b988fe6850d13255d599c9191c9a9b80aa0/src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs#L116-L123

This makes sense, considering that when invoking a command line, you are not working in the context of an existing process but starting a new one, so the most plausible source for data is some file, which we have to specify how to load and so on and so on.

However, then we enter API land, and (understandably, to be clear) people just decided to do a direct translation, as we see below:

https://github.com/dotnet/machinelearning/blob/eed91b988fe6850d13255d599c9191c9a9b80aa0/src/Microsoft.ML.Data/Transforms/ValueToKeyMappingEstimator.cs#L41-L42

That the API might resemble command line as a first preference is understandable, but in this specific context of an API, a variance from this trend would make sense. We've invented what amounts to an entirely new API to load data from a source, when we already have mechanisms to do this.

If we wanted this to work over input `IDataView`s, which seems to be what the authors are really getting at, then it should just do so directly. This has a few advantages:

1. No new way of loading files distinct from existing API for that same task,
2. Simpler method signatures,
3. Hides `IDataLoader` transform, which is something relating to #1995 we need to do anyway. (This in particular is why we might consider this to have some greater urgency.)

/cc @Ivanidzo4ka @sfilipi

TomFinley:
Fixes #2231. We should work to scour this more thoroughly once the changes to allow non-public settings objects on `Arguments`/`Settings` classes goes in.

Also: I had thought another estimator (stopwords) did not have loader as part of its name as I had thought, but I still fixed the name.

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
+++ b/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
@@ -112,15 +112,16 @@ namespace Microsoft.ML
            => new ValueToKeyMappingEstimator(CatalogUtils.GetEnvironment(catalog), inputColumn, outputColumn, maxNumTerms, sort);
 
         /// <summary>
-        /// Converts value types into <see cref="KeyType"/>, optionally loading the keys to use from <paramref name="termData"/>.
+        /// Converts value types into <see cref="KeyType"/>, optionally loading the keys to use from <paramref name="keyData"/>.
         /// </summary>
         /// <param name="catalog">The categorical transform's catalog.</param>
         /// <param name="columns">The data columns to map to keys.</param>
-        /// <param name="termData">The data view containing the terms. If unspecified, they will be
-        /// determined from the input data upon fitting.</param>
+        /// <param name="keyData">The data view containing the terms. If specified, this should be a single column data
+        /// view, and the key-values will be taken from taht column. If unspecified, the key-values will be determined
+        /// from the input data upon fitting.</param>
         public static ValueToKeyMappingEstimator MapValueToKey(this TransformsCatalog.ConversionTransforms catalog,
-            ValueToKeyMappingTransformer.ColumnInfo[] columns, IDataView termData = null)
-            => new ValueToKeyMappingEstimator(CatalogUtils.GetEnvironment(catalog), columns, termData);
+            ValueToKeyMappingTransformer.ColumnInfo[] columns, IDataView keyData = null)
+            => new ValueToKeyMappingEstimator(CatalogUtils.GetEnvironment(catalog), columns, keyData);
 
         /// <summary>
         /// Maps specified keys to specified values
--- a/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
+++ b/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
@@ -112,19 +112,15 @@ namespace Microsoft.ML
            => new ValueToKeyMappingEstimator(CatalogUtils.GetEnvironment(catalog), inputColumn, outputColumn, maxNumTerms, sort);
 
         /// <summary>
-        /// Converts value types into <see cref="KeyType"/> loading the keys to use from <paramref name="file"/>.
+        /// Converts value types into <see cref="KeyType"/>, optionally loading the keys to use from <paramref name="termData"/>.
         /// </summary>
         /// <param name="catalog">The categorical transform's catalog.</param>
         /// <param name="columns">The data columns to map to keys.</param>
-        /// <param name="file">The path of the file containing the terms.</param>
-        /// <param name="termsColumn"></param>
-        /// <param name="loaderFactory"></param>
+        /// <param name="termData">The data view containing the terms. If unspecified, they will be
+        /// determined from the input data upon fitting.</param>
         public static ValueToKeyMappingEstimator MapValueToKey(this TransformsCatalog.ConversionTransforms catalog,
-            ValueToKeyMappingTransformer.ColumnInfo[] columns,
-            string file = null,
-            string termsColumn = null,
-            IComponentFactory<IMultiStreamSource, IDataLoader> loaderFactory = null)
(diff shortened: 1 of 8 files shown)


src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
```diff
@@ -112,19 +112,15 @@ public static KeyToValueMappingEstimator MapKeyToValue(this TransformsCatalog.Co
...
         /// <summary>
-        /// Converts value types into <see cref="KeyType"/> loading the keys to use from <paramref name="file"/>.
+        /// Converts value types into <see cref="KeyType"/>, optionally loading the keys to use from <paramref name="termData"/>.
         /// </summary>
         /// <param name="catalog">The categorical transform's catalog.</param>
         /// <param name="columns">The data columns to map to keys.</param>
-        /// <param name="file">The path of the file containing the terms.</param>
-        /// <param name="termsColumn"></param>
-        /// <param name="loaderFactory"></param>
+        /// <param name="termData">The data view containing the terms. If unspecified, they will be
```
Ivanidzo4ka: Is it worth to specify what that dataview should have only one column and content of that column would be used to to specify allowed keys?
Can we also call it keyData?
I don't think we define Terms anywhere other than in maxNumTerms and even there in description we fallback to term keys.

TomFinley: Good idea. I also did not like the name, but didn't quite know exactly to call it. Fixed this here and I hope everywhere, and also made another few opportunistic name changes that I realized no longer applied. Also enhanced the documentation as requested.

src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
```diff
@@ -112,19 +112,15 @@ public static KeyToValueMappingEstimator MapKeyToValue(this TransformsCatalog.Co
...
+        /// determined from the input data upon fitting.</param>
         public static ValueToKeyMappingEstimator MapValueToKey(this TransformsCatalog.ConversionTransforms catalog,
```
Ivanidzo4ka: >ValueToKeyMappingEstimator [](start = 22, length = 26)

I've tried to look it up, but github doesn't allow you to smoothly explore file changes if file get renamed.
Why we even have this public constructor?
We already have ValueMapping which accepts IEnumerable for keys and ZeeshanA adding another constructor which accepts IDataView.
Considering all this restriction required on IDataView, don't you think it's easier to use ValueMapping transform?

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingEstimator.cs
```diff
@@ -33,23 +31,28 @@ public static class Defaults
...
         public ValueToKeyMappingEstimator(IHostEnvironment env, string inputColumn, string outputColumn = null, int maxNumTerms = Defaults.MaxNumTerms, ValueToKeyMappingTransformer.SortOrder sort = Defaults.Sort) :
-           this(env, new [] { new ValueToKeyMappingTransformer.ColumnInfo(inputColumn, outputColumn ?? inputColumn, maxNumTerms, sort) })
+           this(env, new[] { new ValueToKeyMappingTransformer.ColumnInfo(inputColumn, outputColumn ?? inputColumn, maxNumTerms, sort) })
         {
         }
 
-        public ValueToKeyMappingEstimator(IHostEnvironment env, ValueToKeyMappingTransformer.ColumnInfo[] columns,
-            string file = null, string termsColumn = null,
-            IComponentFactory<IMultiStreamSource, IDataLoader> loaderFactory = null)
+        public ValueToKeyMappingEstimator(IHostEnvironment env, ValueToKeyMappingTransformer.ColumnInfo[] columns, IDataView termData = null)
```
Ivanidzo4ka: >ValueToKeyMappingTransformer.ColumnInfo[] columns [](start = 64, length = 49)

Can we put it termData before columns and then do `params [] columns`?
Or maybe I can convince you to make all constructors with termData internal?

Ivanidzo4ka: >public [](start = 8, length = 6)

not on you, but this public constructor lucks documentation :)

src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
TomFinley: Hi Ivan, I don't understand this question, could you clarify?

Are you asking why there is a public constructor on the estimators? I did not change the public/non-public nature of this constructor of the estimator. If you are pointing out that the estimator constructors should be non-public since we prefer things be created through `MLContext`, I agree, and I believe there is already an issue tracking this, #2100. However, the issue I am addressing here is orthogonal to that. I agree it should be done. I see no one is assigned to it.

Also I am not sure what you are talking about w.r.t. another constructor. Certainly I see no other constructor here. I am trying to address the issue I see that whoever did this original estimator API decided, for whatever reason, that using component factories and `IDataLoader` was a great idea, as discussed in the issue. The most obvious way to maintain the same *sort* of functionality (loading something from a file) is to provide the constructor I have provided. Are you disagreeing with that?

Also I did not rename the file?

But maybe I am misunderstanding your question. Perhaps you could clarify.

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingEstimator.cs
TomFinley: #2100 already tracks the need to make these internal. I don't view that as relevant to the issue I am attempting to solve here. They both have to do with constructors, but I might prefer that we keep it clean, if at all possible.

TomFinley: I think the idea is that these constructors will be made internal, per #2100.

PR #2246, opened 2019-01-25
Disable publishing to myget

### LINKED ISSUE #2244: Official builds are broken because myget is full
We are not getting official builds because the step that publishes the NuGet package to myget is failing because our feed is full. See https://github.com/dotnet/core-eng/issues/5070.

We should disable this step, for now, until the underlying myget feed is fixed.

/cc @shauheen @safern @Ivanidzo4ka

eerhardt:
Fix #2244


Ivanidzo4ka:
How much time it usually takes to solve such type of issue? If it's less than a day it can be not worth the hassle of turning it off/on.

eerhardt:
It's been 2 days so far, with no end in sight.

Issue #2233, opened 2019-01-24
Enable to get ranked-list of predicted labels when performing a multi-class classification prediction


CESARDELATORRE:
This is a needed scenario that can be pretty common from a business needs perspective.

When using ML.NET code implementing multi-class classification, we’re returning/predicting a single label… 

**Multi-class Classification**
- Label: Original Label of the example.
- Score: Its an array whose length is equal to number of classes and contains probability for each class.
**- PredictedLabel: Predicted class.** _<-- (*) There's a single PredictedLabel_

But in many business scenarios, a list of possible labels where something can be classified is also very useful.

•	For instance, a real product might be related to multiple product-categories in an eCommerce, not just one category.
•	Another example where this scenario is possible (and we actually have it working), is image classification in TensorFlow, you can get a list of the best labels with a score per each, from the TF model.

In fact, the class used to get a prediction is like:

   ```
 public class MyPrediction
    {
        public float[] Score;

        public string PredictedLabelValue;
    }
```

So, we actually get an array of Scores, like here:

![image](https://user-images.githubusercontent.com/1712635/51705169-bc56b080-1fcf-11e9-8bd2-1ba930359a4a.png)

**We need a way to map the array of float confidence/score levels back to each original class/label names, in addition to the “single best” PredictedLabelValue.**

Additional info provided by **Ivan Matantsev**: 

_Technically it should be accessible via metadata in Schema, but I don’t see any way to access it in transformer.
Theoretically if you load model from file, you can call GetOutputSchema, but if you just do prediction, you don’t have dataview to get Schema you just have example/prediction classes.
So we either need a way to Expose output Schema in prediction function or to have Transformer which would spit out specific metadata for certain columns._

CESARDELATORRE:
Closing this one (it is already possible) in favor of this issue: https://github.com/dotnet/machinelearning/issues/2278

PR #2250, opened 2019-01-25
Expose schema for prediction engine

Ivanidzo4ka:
towards https://github.com/dotnet/machinelearning/issues/2233
I don't think our current work with metadata is best thing we can offer to the users, but at least this changes can unblock whole possibility to lookup into metadata for prediction engine.
@TomFinley  @CESARDELATORRE


Ivanidzo4ka:
I need to improve my way of communication and be more clear.
@TomFinley I think providing `OutputSchema` in `PredicitionEngine` is a right way to go, and it should be part of it. My comment was about how we currently work with metadata object. Methods like `GetSlotNames` or `GetKeyValues` is not user friendly and require some internal knowledge regarding how to get information out of them, and what is content of this metadata. 
I added `GetKeyValues` method by the analogy of already present `GetSlotNames` method. I don't think these methods are best methods for users possible, but in same time, I think we should have these methods, at least as internal because they ease process of development. And since there is no better work with metadata in current moment, I made that method public. 
I was under assumption what we will review metadata object in general before v1.0 release until that is done, I think it brings value to have some possibility for user to get SlotNames and KeyValues.

test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs
```diff
@@ -0,0 +1,57 @@
...
+                .Append(ml.MulticlassClassification.Trainers.StochasticDualCoordinateAscent(
+                    new SdcaMultiClassTrainer.Options { MaxIterations = 100, Shuffle = true, NumThreads = 1, }));
+
+            var model = pipeline.Fit(data).GetModelFor(TransformerScope.Scoring);
+            var engine = model.CreatePredictionEngine<IrisDataNoLabel, IrisPredictionNotCasted>(ml);
+
+            var testLoader = ml.Data.ReadFromTextFile(dataPath, TestDatasets.irisData.GetLoaderColumns(), hasHeader: true, separatorChar: ',');
+            var testData = ml.CreateEnumerable<IrisData>(testLoader, false);
+            // Slot names on top of Score column represent original labels for i-th value in Score array.
+            VBuffer<ReadOnlyMemory<char>> slotNames = default;
```
CESARDELATORRE: I'd wish we could have a simpler way to access the info about "ranked list of predicted 'possible categories/lables' " without needing to use `VBuffer<ReadOnlyMemory<char>>` and other kind of "low level" types which are not familiar to .NET developers. For upcoming servions we should hide VBuffer so it looks a lot simpler and user friendly for developers.
​

test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs
```diff
@@ -0,0 +1,57 @@
...
+            engine.OutputSchema[nameof(IrisPrediction.Score)].GetSlotNames(ref slotNames);
+            // Key names represent original values for PredictedLabel column.
+            VBuffer<ReadOnlyMemory<char>> keys = default;
+            engine.OutputSchema[nameof(IrisPrediction.PredictedLabel)].GetKeyValues(ref keys);
+
+            Assert.True(slotNames.GetItemOrDefault(0).ToString() == "Iris-setosa");
+            Assert.True(slotNames.GetItemOrDefault(1).ToString() == "Iris-versicolor");
+            Assert.True(slotNames.GetItemOrDefault(2).ToString() == "Iris-virginica");
+
```
CESARDELATORRE: Please, add around here the code relating each of the predicted labels with their score.

PR #2254, opened 2019-01-26
Move IDataView to Microsoft.Data.DataView namespace.

eerhardt:
Renaming the namespace of the IDataView types.

Follow up to #1860 

I also added a few copyrights where they were missing.

### DIFF
--- a/src/Microsoft.ML.Parquet/ParquetLoader.cs
+++ b/src/Microsoft.ML.Parquet/ParquetLoader.cs
@@ -9,6 +9,7 @@ using System.IO;
 using System.Linq;
 using System.Numerics;
 using System.Text;
+using Microsoft.Data.DataView;
 using Microsoft.ML;
 using Microsoft.ML.CommandLine;
 using Microsoft.ML.Data;
@@ -18,6 +19,7 @@ using Microsoft.ML.Model;
 using Parquet;
 using Parquet.Data;
 using Parquet.File.Values.Primitives;
+using DataViewSchema = Microsoft.Data.DataView.Schema;
 
 [assembly: LoadableClass(ParquetLoader.Summary, typeof(ParquetLoader), typeof(ParquetLoader.Arguments), typeof(SignatureDataLoader),
     ParquetLoader.LoaderName, ParquetLoader.LoaderSignature, ParquetLoader.ShortName)]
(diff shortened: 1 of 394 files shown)


src/Microsoft.ML.Parquet/ParquetLoader.cs
```diff
@@ -18,6 +19,7 @@
 using Parquet;
 using Parquet.Data;
 using Parquet.File.Values.Primitives;
+using DataViewSchema = Microsoft.Data.DataView.Schema;
```
TomFinley: The idea was broached I know, but resolution never reached: should we rename? Or else perhaps make `Schema` a nested class of something appropriate, so as to give it definition?

I guess if something as useful as the Parquet format can get away with naming something "schema" without complaint, maybe we (being, I suppose, less universally beloved for now) can get away with it as well fairly easily.

eerhardt:
> may I know what tool you used?

The Error List Window in VS, and CTRL+. -> Add Using.

:(

src/Microsoft.ML.Parquet/ParquetLoader.cs
eerhardt: We could rename the class to DataViewSchema, or ViewSchema.

We’d probably want to rename everything then. Row is a pretty generic name. So is ColumnType.

test/Microsoft.ML.CodeAnalyzer.Tests/TypeIsSchemaShapeTest.cs
```diff
@@ -78,15 +78,15 @@ public void ReturnTypeIsSchemaShapeClass()
...
-                diagPc.CreateDiagnosticResult(39, 13, "Class11"),
+                diagF.CreateDiagnosticResult(33, 13, "Class4", "F1"),
+                diagC.CreateDiagnosticResult(34, 13, "Class5"),
+                diagC.CreateDiagnosticResult(35, 13, "Class6"),
+
+                diagPnas.CreateDiagnosticResult(36, 13, "Class7", "F1"),
+                diagPs.CreateDiagnosticResult(37, 13, "Class8", "F2"),
+                diagPg.CreateDiagnosticResult(38, 13, "Class9", "F2"),
+                diagPc.CreateDiagnosticResult(39, 13, "Class10"),
+                diagPc.CreateDiagnosticResult(40, 13, "Class11"),
```
justinormont: I was scrolling thru, getting carpal tunnel, about to conclude "oh look an easy PR approval". 

What are these changing in response to?

eerhardt: This is our C# diagnostics analyzer test, which ensures our code analyzer gives the right warnings/errors on the correct code lines.

Because the code being analyzed used an ‘IDataView’ variable, I needed to insert a new ‘using’ statement at the top of the code. This moved all the code down one line, causing the tests to now be off by one line.

PR #2259, opened 2019-01-28
Fix disabled code coverage for src folder.

### LINKED ISSUE #2258: Code coverage has been disabled for src directory
Pull request #2200 (unintentionally) caused the complete omission of code coverage information from the **src** directory.

The following line causes a problem in **Directory.Build.props**:

https://github.com/dotnet/machinelearning/blob/2f0c4755888e6a6a7f2f1206516479a8120a089b/test/Directory.Build.props#L9-L10

By placing this definition early in the build, the following does not take effect:

https://github.com/Microsoft/msbuild/blob/3a844ce00a873d8ca6ba0a8abc0c6b64f772d937/src/Tasks/Microsoft.Common.CurrentVersion.targets#L584-L594

Currently the test projects fail to copy assemblies and symbol files to the output directories, which completely breaks code coverage instrumentation for referenced projects. The property needs to be moved to **Directory.Build.targets** so it comes after the common definition.

codemzs:
Seems like #2200 broke code coverage for src folder, hence production coverage is not being generated and even the codecov badge on the main isn't displaying any coverage. 

fixes #2258

### DIFF
--- a/test/Directory.Build.targets
+++ b/test/Directory.Build.targets
@@ -3,7 +3,8 @@
   <Import Project="..\Directory.Build.targets" />
 
   <PropertyGroup>
-    <!-- This is required in order to have following extension files in the output folder of the projects referencing other projects. -->
+    <!-- This needs to be in Directory.Build.targets or the default value for AllowedReferenceRelatedFileExtensions will not be applied. 
+         This is required in order to have following extension files in the output folder of the projects referencing other projects. -->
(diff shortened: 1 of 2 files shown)


test/Directory.Build.targets
```diff
@@ -2,6 +2,11 @@
 
   <Import Project="..\Directory.Build.targets" />
 
+  <PropertyGroup>
+    <!-- This is required in order to have following extension files in the output folder of the projects referencing other projects. -->
```
sharwell: 💡 Would be good to have a comment explaining that this needs to be in **Directory.Build.targets** or the default value for AllowedReferenceRelatedFileExtensions will not be applied.

Issue #2210, opened 2019-01-23
Update the cookbook to contain just the dynamic examples


sfilipi:
A few people have commented that having both the dynamic API and the static API examples in the cookbook is confusing for new users. 

Separate them into two documents.

PR #2284, opened 2019-01-28
Creating the MLNetCookBook_StaticApi.md and updating the current cookbook

sfilipi:
Creating the MLNetCookBook_StaticApi.md. The document contains the same content and the MLNetCookBook.md, but the sample code uses the static API.

Updating the official cookbook to remove the static API samples, and leave just the dynamic API ones.

Fixes: #2210

### DIFF
--- a/docs/code/MlNetCookBook.md
+++ b/docs/code/MlNetCookBook.md
@@ -2,13 +2,9 @@
 
 This document is intended to provide essential samples for common usage patterns of ML.NET. 
 It is advisable to be at least minimally familiar with [high-level concepts of ML.NET](MlNetHighLevelConcepts.md), otherwise the terminology in this document may be foreign to you.
-The examples in this document make use of the dynamic API, currently the supported ML.Net api.
-In ML.Net there is also a static api, that operates on the schema of the data, strongly typing the data type. That version of the API is considered experimental.
-If you are loading an existing model from a stream, there's no need to use static types (and it's also pretty hard to do). 
-Also, if the data view's schema is only known at runtime, there is no way to use static types. 
-Using the static API gives you compiler support: it's more likely that, if your code compiles, it will also work as intended.
-A version of this cookbook containing the same samples in the static API samples can be found in the [MlNetCookBook_staticApi.md](experimental\MlNetCookBook_staticApi.md)
-We welcome feedback about it.  
+The examples in this document make use of the dynamic API, currently the supported ML.NET API.
+In ML.NET there is also a static API, that operates on the schema of the data, strongly typing the data type. 
+That version of the API is considered experimental., more about it in the [MlNetCookBook_staticApi.md](experimental\MlNetCookBook_staticApi.md).
 
 ## How to use this cookbook
 
@@ -298,10 +294,10 @@ var data = mlContext.Data.ReadFromTextFile<InspectedRow>(dataPath,
 
 // Start creating our processing pipeline. For now, let's just concatenate all the text columns
 // together into one.
-var dynamicPipeline = mlContext.Transforms.Concatenate("AllFeatures", "Education", "MaritalStatus");
+var pipeline = mlContext.Transforms.Concatenate("AllFeatures", "Education", "MaritalStatus");
 
 // Fit our data pipeline and transform data with it.
-var transformedData = dynamicPipeline.Fit(data).Transform(data);
+var transformedData = pipeline.Fit(data).Transform(data);
 
 // 'transformedData' is a 'promise' of data. Let's actually read it.
 var someRows = mlContext
@@ -350,13 +346,13 @@ var trainData = mlContext.Data.ReadFromTextFile<AdultData>(dataPath,
 // Sometime, caching data in-memory after its first access can save some loading time when the data is going to be used
 // several times somewhere. The caching mechanism is also lazy; it only caches things after being used.
 // User can replace all the subsequently uses of "trainData" with "cachedTrainData". We still use "trainData" because
-// a caching step, which provides the same caching function, will be inserted in the considered "dynamicPipeline."
+// a caching step, which provides the same caching function, will be inserted in the considered "pipeline."
 var cachedTrainData = mlContext.Data.Cache(trainData);
 
 // Step two: define the learning pipeline. 
 
 // We 'start' the pipeline with the output of the reader.
-var dynamicPipeline =
+var pipeline =
     // First 'normalize' the data (rescale to be
     // between -1 and 1 for all examples)
     mlContext.Transforms.Normalize("FeatureVector")
@@ -373,7 +369,7 @@ var dynamicPipeline =
     .Append(mlContext.Regression.Trainers.StochasticDualCoordinateAscent(label: "Target", features: "FeatureVector"));
 
 // Step three. Fit the pipeline to the training data.
-var model = dynamicPipeline.Fit(trainData);
+var model = pipeline.Fit(trainData);
 ```
 ## How do I verify the model quality?
 
@@ -442,7 +438,7 @@ var trainData = mlContext.Data.ReadFromTextFile<IrisInput>(irisDataPath,
 );
 
 // Build the training pipeline.
-var dynamicPipeline =
+var pipeline =
     // Concatenate all the features together into one column 'Features'.
     mlContext.Transforms.Concatenate("Features", "SepalLength", "SepalWidth", "PetalLength", "PetalWidth")
     // Note that the label is text, so it needs to be converted to key.
@@ -455,7 +451,7 @@ var dynamicPipeline =
     .Append(mlContext.Transforms.Conversion.MapKeyToValue(("PredictedLabel", "Data")));
 
 // Train the model.
-var model = dynamicPipeline.Fit(trainData);
+var model = pipeline.Fit(trainData);
 ```
 
 Now, in order to use [schema comprehension](SchemaComprehension.md) for prediction, we define a pair of classes like following:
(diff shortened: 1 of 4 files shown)


docs/code/MlNetCookBook.md
```diff
@@ -2,6 +2,13 @@
 
 This document is intended to provide essential samples for common usage patterns of ML.NET. 
 It is advisable to be at least minimally familiar with [high-level concepts of ML.NET](MlNetHighLevelConcepts.md), otherwise the terminology in this document may be foreign to you.
+The examples in this document make use of the dynamic API, currently the supported ML.Net api.
```
shmoradims: >api [](start = 90, length = 3)

uppercase

docs/code/MlNetCookBook.md
```diff
@@ -2,6 +2,13 @@
...
+In ML.Net there is also a static api, that operates on the schema of the data, strongly typing the data type. That version of the API is considered experimental.
```
shmoradims: >api [](start = 33, length = 3)

upper here and everywhere

docs/code/MlNetCookBook.md
```diff
@@ -2,6 +2,13 @@
...
+If you are loading an existing model from a stream, there's no need to use static types (and it's also pretty hard to do). 
+Also, if the data view's schema is only known at runtime, there is no way to use static types. 
+Using the static API gives you compiler support: it's more likely that, if your code compiles, it will also work as intended.
+A version of this cookbook containing the same samples in the static API samples can be found in the [MlNetCookBook_staticApi.md](experimental\MlNetCookBook_staticApi.md)
+We welcome feedback about it.
```
shmoradims: I suggest we don't elaborate much about the static API in this doc and only limit ourselves to 1 sentence about it with the link. And push all the details (what it is, when to use it, when not to use it, advantages, motivation, etc) to the static_api.md file.

docs/code/MlNetCookBook.md
```diff
@@ -827,71 +562,53 @@ In the static pipeline API, we provide a set of `onFit` delegates that allow int
...
-        // Note that the label is a text, so it needs to be converted to key using 'ToKey' estimator.
-        Predictions: mlContext.MulticlassClassification.Trainers.Sdca(r.Label.ToKey(), r.Features,
-            // When the model is trained, the below delegate is going to be called.
-            // We use that to memorize the predictor object.
-            onFit: p => predictor = p)));
-
-// Train the model. During this call our 'onFit' delegate will be invoked,
-// and our 'predictor' will be set.
-var model = learningPipeline.Fit(trainData);
+var dynamicPipeline =
```
shmoradims: >dynamicPipeline [](start = 4, length = 15)

let's drop 'dynamic' prefix, and use 'pipeline' or 'learningPipeline' throughout the file.

docs/code/experimental/MlNetCookBook_StaticApi.md
```diff
@@ -0,0 +1,1026 @@
+# ML.NET Cookbook
+
+This document is intended to provide essential samples for common usage patterns of ML.NET, using examples from the experimental static API.
+The examples in this document make use of the static API, which at the time of writting is considered experimental.
+The supported way of composing pipelines in ML.NET is through the dynamic API. 
+You can find the same guide and examples on how to get started with ML.Net, dynamic API on the [cookbook reference for the dynamic API](..\MlNetCookBook.md)
+It is advisable to be at least minimally familiar with [high-level concepts of ML.NET](..\MlNetHighLevelConcepts.md), otherwise the terminology in this document may be foreign to you.
+We welcome feedback about the static API.
```
shmoradims: >. [](start = 40, length = 1)

let's add the motivation behind static API and how to use it. The stuff that I mentioned to be moved from the other md to here.

docs/code/experimental/MlNetCookBook_StaticApi.md
```diff
@@ -0,0 +1,1026 @@
...
+// We apply our FastTree binary classifier to predict the 'HasChurned' label.
+
+// First, transition to the statically-typed data view.
+var staticData = trainData.AssertStatic(mlContext, c => (
+        HasChurned: c.Bool.Scalar,
+        DemographicCategory: c.Text.Scalar,
+        LastVisits: c.R4.Vector));
+
+// Build the pipeline, same as the one above.
+var staticLearningPipeline = staticData.MakeNewEstimator()
```
shmoradims: >staticLearningPipeline [](start = 4, length = 22)

let's be consistent and pick one throughout the md: staticLearningPipeline or staticPipeline. I suggest just prefixing the name you chose in the dynamic.md file with 'static', so that when users copy-paste they don't end up mixing the default (dynamic) examples with this one.

docs/code/MlNetCookBook.md
sfilipi: moved that info in the static doc. 





---
In reply to: [251604799](https://github.com/dotnet/machinelearning/pull/2284#discussion_r251604799) [](ancestors = 251604799)

docs/code/MlNetCookBook.md
```diff
@@ -45,25 +48,19 @@ This is because we rely on extension methods a lot, and they only become availab
...
 using Microsoft.ML.Trainers;
 using Microsoft.ML.Transforms;
 ```
 
-- *Why is there two ways of doing things? Which one is better, static or dynamic?*
-We don't know yet, that's why we have two ways! 
+- *What is the [MLContext](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.mlcontext?view=ml-dotnet)?*
+The MLContext is a starting point for all ML.NET operations. 
+It is instantiated by user, and provides mechanisms for logging, exception tracking and logging, setting the source of randomness.
+It is the entry points for training, prediction, model operations and also serves as a catalog of available operations.
```
artidoro: >entry points [](start = 10, length = 12)

Would use the singular (no 's'). Maybe even substitute with "starting point"

docs/code/MlNetCookBook.md
```diff
@@ -45,25 +48,19 @@ This is because we rely on extension methods a lot, and they only become availab
...
-In the static pipeline, we need to know the two 'schema' types: the input and the output to the pipeline.
-One of them is already known: typically, the output schema of `reader` (which is the same as the schema of `reader.Read()`) is also the input schema of the learning pipeline. 
-
-The call to `x.MakeNewEstimator` is only using the `x`'s *schema* to create an empty pipeline, it doesn't use anything else from `x`. So, the following three lines would create the exactly same (empty) pipeline:
 ```csharp
-var p1 = reader.MakeNewEstimator();
-var p2 = reader.Read(dataLocation).MakeNewEstimator();
-var p3 = p1.MakeNewEstimator();
+// as a catalog of available operations and as the source of randomness.
+var mlContext = new MLContext();
```
artidoro: Was this line supposed to be kept?

docs/code/MlNetCookBook.md
```diff
@@ -779,7 +530,7 @@ IEnumerable<CustomerChurnInfo> churnData = GetChurnInfo();
 // Turn the data into the ML.NET data view.
 // We can use CreateDataView or CreateStreamingDataView, depending on whether 'churnData' is an IList, 
 // or merely an IEnumerable.
-var trainData = mlContext.Data.ReadFromEnumerable(churnData);
+var trainData = mlContext.CreateStreamingDataView(churnData);
```
artidoro: >CreateStreamingDataView [](start = 26, length = 23)

I think now it's `mlContext.Data.ReadFromEnumerable`. We should update the comment as well.

artidoro:
> // We demonstrate both below.

Maybe let's not reference the static API here?

---
Refers to: docs/code/MlNetCookBook.md:537 in afaf889. [](commit_id = afaf889ce0065fd2515cadc588f7dddbea4f32bd, deletion_comment = False)

docs/code/MlNetCookBook.md
```diff
@@ -827,71 +558,53 @@ In the static pipeline API, we provide a set of `onFit` delegates that allow int
```
artidoro: Here as well, I don't know if I would reference the static API.

docs/code/MlNetCookBook.md
sfilipi: yes, since we are talking about the MLContext. 

---
In reply to: [251650454](https://github.com/dotnet/machinelearning/pull/2284#discussion_r251650454) [](ancestors = 251650454)

docs/code/MlNetCookBook.md
```diff
@@ -2,6 +2,12 @@
 
 This document is intended to provide essential samples for common usage patterns of ML.NET. 
 It is advisable to be at least minimally familiar with [high-level concepts of ML.NET](MlNetHighLevelConcepts.md), otherwise the terminology in this document may be foreign to you.
+The examples in this document make use of the dynamic API, currently the supported ML.NET API.
+In ML.NET there is also a static API, that operates on the schema of the data, strongly typing the data type. 
+That version of the API is considered experimental., more about it in the [Static API Cookbook](experimental/MlNetCookBook_staticApi.md)
```
artidoro: I think that this line should be removed at this point?

docs/code/MlNetCookBook.md
artidoro: It seems this was the second part of a two line comment, and it feels like it's missing something. Maybe I would add the first part of the sentence.

Could you capitalize otherwise?

---
In reply to: [251944224](https://github.com/dotnet/machinelearning/pull/2284#discussion_r251944224) [](ancestors = 251944224,251650454)

PR #2303, opened 2019-01-29
Removed python naming conventions from samples - closes #2155

### LINKED ISSUE #2155: C# samples should probably use C# and not Python naming conventions
I was reading a PR when suddenly I saw something a bit odd.

https://github.com/dotnet/machinelearning/blob/6b9f5893ea6830481a8779f81f2b723e9b5cd6cd/docs/samples/Microsoft.ML.Samples/Dynamic/KeyToValue_Term.cs#L42-L47

Note the underscores in the middle of a variable name. While in private we might elect to bend the rules a little bit, in our public API and also probably our samples, we probably ought to write them with [consideration of standard practices](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions). This does not seem to be a justifiable or explainable diversion from the recommended practice, and I'm guessing it was added by mistake.

Each individual renaming is probably not huge, but might span several files, since the example listed above is just one of many examples I see in our samples project. So not a "hard" change, except possibly in scale?

daniel-loudon:
Fixes #2155 - C# samples should use C# naming conventions and not python naming conventions. Second PR on this issue because last PR had issues

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/FeatureContributionCalculationTransform.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/FeatureContributionCalculationTransform.cs
@@ -3,9 +3,9 @@ using Microsoft.ML.Data;
 
 namespace Microsoft.ML.Samples.Dynamic
 {
-    public class FeatureContributionCalculationTransform_RegressionExample
+    public class FeatureContributionCalculationTransformRegressionExample
     {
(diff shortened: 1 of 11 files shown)


sfilipi:
thank you @daniel-loudon

daniel-loudon:
Yep. I have had issues with visual studio and fetching from this repository.

PR #2308, opened 2019-01-29
Upload coverage files only when all tests pass.

codemzs:
Sometimes not all tests pass and coverage files with missing data is uploaded to codecov and this changes the baseline of the master repo and causes all sorts of fluctuations in the coverage report both in PR and on the main page.


codemzs:
@sharwell Thanks but I believe you mean Coverlet and not CodeCov, right?

sharwell:
@codemzs Yes correct

Issue #1995, opened 2019-01-02
Internalize concepts of IDataTransform/Loader/TransformTemplate.


TomFinley:
Before we had the `IEstimator`/`ITransformer`/`IDataView` triad (see #581), we combined all three concepts inside `IDataTransform`. The initialization, the data, *and* the data model were all conflated. This was the source for years of constant confusion internally. We of course had to abandon the idea when we were trying to make an API. Here's the interface, for reference:

https://github.com/dotnet/machinelearning/blob/2e67134b45c9253797d0fa137054ec8da1d7ad9c/src/Microsoft.ML.Data/Data/IDataLoader.cs#L90

Similarly, before we had `IDataReaderEstimator`/`IDataReader`/`IDataView` triad, we combined all three concepts inside `IDataLoader`.

https://github.com/dotnet/machinelearning/blob/2e67134b45c9253797d0fa137054ec8da1d7ad9c/src/Microsoft.ML.Data/Data/IDataLoader.cs#L53

We also have the old concept of `ITransformTemplate`, which the `ITransformer` interface has likewise obviated.

We've introduced the new concepts over the course of the last half year or so, which is great, yet the artifacts of the old world remain. The interfaces should be made internal so people don't stumble into using them. Further, the objects used through `ITransform` and `IDataView` should (even when still backed by an `IDataTransform` object!) be hidden so it is not obvious these concepts are being conflated internally, even if "under the hood" there is just a small shim to present what is (internally) an `IDataTransform` as publicly an `ITransformer`/`IDataView`. (Always with the aim, of course, of having the architecture reflect the shape it presents as its public surface, but what we can refactor over some time we can at least hide now.)

TomFinley:
One example of this that I ran across is the `PredictionTransformerBase`:

https://github.com/dotnet/machinelearning/blob/46efb8ed90d90d6e444394c3e3f5fbe16214109a/src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs#L38-L40

This is a `Transformer`, which is fine, but it quite inappropriately takes as a generic parameter something that is `TScorer : RowToRowScorerBase`, which is something exposing the "old idioms." However there is no reason to make this detail of implementation public, but it must be as a generic parameter.

[...]

PR #2309, opened 2019-01-29
Internalize IDataLoader

Ivanidzo4ka:
Towards #https://github.com/dotnet/machinelearning/issues/1995


TomFinley:
Thank you Ivan. What of the BinaryLoader? Should that remain public as well?

TomFinley:
> Funny. I expect to get bunch of exceptions from any class which implements IDataLoader and public, but apparently that's not the case.
> 
> In reply to: [458766322](https://github.com/dotnet/machinelearning/pull/2309#issuecomment-458766322) [](ancestors = 458766322)

Right. So whether a method is public or not versus whether it adheres to the contract are, in some sense, independent. So you can have an internal interface, a public class can adhere to that interface with public methods. This FYI is why we often have the additional work, if the class must remain public, that its implementing methods become explicit, if that makes sense?

For now, it would be preferable if, as I think you have done, that the implementing classes *themselves* become internal, since we would want them to adhere not to the `IDataLoader` interface but its successor, the `IIDataReader` interface. (Though we may in the end wind up deleting `IDataLoader` and renaming what we now call `IDataReader`, `IDataLoader` in its place.)

src/Microsoft.ML.Data/DataLoadSave/PartitionedPathParser.cs
```diff
@@ -193,12 +196,13 @@ public IEnumerable<string> ParseValues(string path)
 
     [TlcModule.Component(Name = ParquetPartitionedPathParser.LoadName, FriendlyName = ParquetPartitionedPathParser.UserName,
         Desc = ParquetPartitionedPathParser.Summary, Alias = ParquetPartitionedPathParser.ShortName)]
-    public class ParquetPartitionedPathParserFactory : IPartitionedPathParserFactory
+    internal class ParquetPartitionedPathParserFactory : IPartitionedPathParserFactory
```
TomFinley: >ParquetPartitionedPathParserFactory [](start = 19, length = 35)

Huh... weird. Not your code I know, just curious, is it obvious why this isn't living off in Parquet land?

src/Microsoft.ML.Data/Utilities/ComponentCreation.cs
```diff
@@ -126,7 +126,8 @@ public static IDataView LoadTransforms(this IHostEnvironment env, Stream modelSt
         /// <summary>
         /// Creates a data loader from the arguments object.
         /// </summary>
-        public static IDataLoader CreateLoader<TArgs>(this IHostEnvironment env, TArgs arguments, IMultiStreamSource files)
+        [BestFriend]
+        internal static IDataLoader CreateLoader<TArgs>(this IHostEnvironment env, TArgs arguments, IMultiStreamSource files)
```
TomFinley: >CreateLoader [](start = 36, length = 12)

Thanks Ivan. I wouldn't necessarily do this now, but this entire assembly seems like a *lot* of stuff we do not want public. But not as part of this PR.

src/Microsoft.ML.Data/DataLoadSave/PartitionedPathParser.cs
Ivanidzo4ka: I don't know, and from your tone it looks like you want me to move them to Parquet project. Your wish is my command master.

---
In reply to: [252374825](https://github.com/dotnet/machinelearning/pull/2309#discussion_r252374825) [](ancestors = 252374825)

artidoro:
>     /// we could simply replace this with an array of IFileHandle.

Could you move this Review comment as it is a public summary?

---
Refers to: src/Microsoft.ML.Data/Data/IDataLoader.cs:14 in 66de22e. [](commit_id = 66de22e97f4c2f0862386d3907e46ba85d01c76d, deletion_comment = False)

artidoro:
>         /// </summary>

Here as well.

---
Refers to: src/Microsoft.ML.Data/Data/IDataLoader.cs:36 in 66de22e. [](commit_id = 66de22e97f4c2f0862386d3907e46ba85d01c76d, deletion_comment = False)

artidoro:
>         public abstract class ArgumentsBase : TransformInputBase

Why not making `ArgumentsBase` and `Arguements` internal? As you did already in `ValueMapping`.

---
Refers to: src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs:111 in 66de22e. [](commit_id = 66de22e97f4c2f0862386d3907e46ba85d01c76d, deletion_comment = False)

artidoro:
>         public abstract class ArgumentsBase : TransformInputBase

But we will make it internal anyways as part of the issue that cleans up the transformers.

---
In reply to: [459065845](https://github.com/dotnet/machinelearning/pull/2309#issuecomment-459065845) [](ancestors = 459065845)

---
Refers to: src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs:111 in 66de22e. [](commit_id = 66de22e97f4c2f0862386d3907e46ba85d01c76d, deletion_comment = False)

Ivanidzo4ka:
>         public abstract class ArgumentsBase : TransformInputBase

Yeah, I would prefer to not diverge from intent of this PR

---
In reply to: [459066115](https://github.com/dotnet/machinelearning/pull/2309#issuecomment-459066115) [](ancestors = 459066115,459065845)

---
Refers to: src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs:111 in 66de22e. [](commit_id = 66de22e97f4c2f0862386d3907e46ba85d01c76d, deletion_comment = False)

src/Microsoft.ML.Parquet/PartitionedPathParser.cs
```diff
@@ -276,8 +280,7 @@ public void Save(ModelSaveContext ctx)
                 ctx.SaveString(sb.ToString());
             };
         }
-
-        public IEnumerable<PartitionedFileLoader.Column> ParseColumns(string path)
+        IEnumerable<PartitionedFileLoader.Column> IPartitionedPathParser.ParseColumns(string path)
```
artidoro: >IEnumerable<PartitionedFileLoader.Column> IPartitionedPathParser.ParseColumns(string path) [](start = 8, length = 90)

Why this change if the outer class is internal? Also you didn't change `ParseColumns` above, which makes me thing it does not need to be changed.

src/Microsoft.ML.Data/Transforms/ValueMapping.cs
```diff
@@ -362,7 +362,7 @@ internal bool TryUnparse(StringBuilder sb)
             }
         }
 
-        public sealed class Arguments
+        internal sealed class Arguments
```
artidoro: >internal [](start = 8, length = 8)

For consistency, I would either make this public and only make specific field internal, or make the `Arguments` internal in the other transforms too.

src/Microsoft.ML.Parquet/PartitionedPathParser.cs
Ivanidzo4ka: I made this change first, and only after that made whole class internal. Thanks for the catch!

---
In reply to: [252397204](https://github.com/dotnet/machinelearning/pull/2309#discussion_r252397204) [](ancestors = 252397204)

src/Microsoft.ML.Data/Transforms/ValueMapping.cs
Ivanidzo4ka: In other places we use `Arguments` classes as EntryPoint arguments. I don't want to get into business of cleaning entrypoints in this PR.
For this transform we don't have entrypoint, so I prefer to just hide it completely. 

---
In reply to: [252398379](https://github.com/dotnet/machinelearning/pull/2309#discussion_r252398379) [](ancestors = 252398379)

artidoro: Ok sounds good!

---
In reply to: [252399795](https://github.com/dotnet/machinelearning/pull/2309#discussion_r252399795) [](ancestors = 252399795,252398379)

src/Microsoft.ML.Data/DataLoadSave/PartitionedPathParser.cs
TomFinley: Well, that would certainly be very nice, thank you @Ivanidzo4ka. 😄

src/Microsoft.ML.Data/Utilities/ComponentCreation.cs
TomFinley: Oh what I meant was the whole class. But like I said not part of this PR.

PR #2323, opened 2019-01-30
Removed Read methods taking string parameters from TextLoader (#1797)

### LINKED ISSUE #1797: Remove TextLoader.Read(string), promote IDataReader<MSS> extension
Pursuant to #1090 and #1041 there was agreement that we should be able to just pass a *string* into a data reader that reads from streams, rather than have to muck about with the creation of this `IMultiStreamSource` object.

As it happens, this one request was done in two separate ways:

1. #1252 where this instance method on `TextLoader` was created:
  
 https://github.com/dotnet/machinelearning/blob/cb37c7e7f1e1b29b5608a2755db793c5435d10b1/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L1340

2. #1281 where an extension method on top of `IDataReader<IMultiStreamSource>` was introduced.

Clearly we don't need both, and the second being more flexible and more broadly applicable was the one that should be checked in, so let us remove the class specific method introduced in #1252. The extension method should however be moved to a namespace where it is more likely to be picked up (probably `Microsoft.ML`) since right now it's a bit buried and generally people would never see this extension method.

PaulTFreedman:
Unsure about the naming of `LocalPathReader` but I kept it the same due to the requirement that it be namespaced to `Microsoft.ML` - `DataViewExtensions` already exists and is in the `Microsoft.ML.Data` namespace.

Fixes #1797.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1312,10 +1312,6 @@ namespace Microsoft.ML.Data
 
         public IDataView Read(IMultiStreamSource source) => new BoundLoader(this, source);
 
-        public IDataView Read(string path) => Read(new MultiFileSource(path));
-
-        public IDataView Read(params string[] path) => Read(new MultiFileSource(path));
-
         internal static TextLoader CreateTextReader<TInput>(IHostEnvironment host,
            bool hasHeader = DefaultArguments.HasHeader,
            char separator = DefaultArguments.Separator,
--- a/test/Microsoft.ML.Predictor.Tests/Microsoft.ML.Predictor.Tests.csproj
+++ b/test/Microsoft.ML.Predictor.Tests/Microsoft.ML.Predictor.Tests.csproj
@@ -11,7 +11,6 @@
     <ProjectReference Include="..\..\src\Microsoft.ML.LightGBM\Microsoft.ML.LightGBM.csproj" />
     <ProjectReference Include="..\..\src\Microsoft.ML.PCA\Microsoft.ML.PCA.csproj" />
     <ProjectReference Include="..\..\src\Microsoft.ML.StandardLearners\Microsoft.ML.StandardLearners.csproj" />
-    <ProjectReference Include="..\..\src\Microsoft.ML.StaticPipe\Microsoft.ML.StaticPipe.csproj" />
     <ProjectReference Include="..\..\src\Microsoft.ML.Transforms\Microsoft.ML.Transforms.csproj" />
     <ProjectReference Include="..\Microsoft.ML.TestFramework\Microsoft.ML.TestFramework.csproj" />
   </ItemGroup>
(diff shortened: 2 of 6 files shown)


test/Microsoft.ML.Predictor.Tests/Microsoft.ML.Predictor.Tests.csproj
```diff
@@ -11,6 +11,7 @@
     <ProjectReference Include="..\..\src\Microsoft.ML.LightGBM\Microsoft.ML.LightGBM.csproj" />
     <ProjectReference Include="..\..\src\Microsoft.ML.PCA\Microsoft.ML.PCA.csproj" />
     <ProjectReference Include="..\..\src\Microsoft.ML.StandardLearners\Microsoft.ML.StandardLearners.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.ML.StaticPipe\Microsoft.ML.StaticPipe.csproj" />
```
Ivanidzo4ka: >StaticPipe [](start = 78, length = 10)

This is not a good sign.
In current moment we prefer to separate our StaticPipe project from rest of the code.

I would suggest to split LocalPathReader into two classes. One should stay next to TextLoader.cs and have just Read method without <Shape>.
Second should remain in StaticPipe project, and have Read method with <Shape>.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1312,10 +1312,6 @@ public void Save(ModelSaveContext ctx)
 
         public IDataView Read(IMultiStreamSource source) => new BoundLoader(this, source);
 
-        public IDataView Read(string path) => Read(new MultiFileSource(path));
-
-        public IDataView Read(params string[] path) => Read(new MultiFileSource(path));
```
Ivanidzo4ka: I think we prefer to have two methods because params is a bit slow one, and for performance we prefer to have one method for single file case (which is like 99%) and one for case with multiple file paths.

Ivanidzo4ka:
@PaulTFreedman thank you for your contribution! I left few comments, can you take care of them?

test/Microsoft.ML.Predictor.Tests/Microsoft.ML.Predictor.Tests.csproj
PaulTFreedman: Just to be clear on your meaning, when you say with/without "." you're talking about calling something like `reader.Read(...)` vs just `Read(...)`?

Ivanidzo4ka: Sorry for confusion, I forgot what web interface doesn't gracefully handles tags, and you need to escape them all the time.
I was talking about `public static DataView<TShape> Read<TShape>` that thing should remain in StaticPipe project.

If I correctly understand you code and intention you removing `public IDataView Read(string path)` from TextLoader but since `LocalPathReader` contains extension method on `IDataReader<IMultiStreamSource> reader` all our code continue to work since we fall back to that extension.

This is great, it just that extension shouldn't be part of StaticPipe project. I would prefer it to remain in Microsoft.ML.Data project.
In same time second extension method which works on top of `DataReader<IMultiStreamSource, TShape>` should remain to be inside StaticPipe project.

Does my explanation make sense?

Ivanidzo4ka:
> using Microsoft.ML.Data;

please add header (see same comments for details)

---
Refers to: src/Microsoft.ML.StaticPipe/LocalPathReader.cs:1 in 0a4cb1f. [](commit_id = 0a4cb1f09b326b10f785d4ba4dbc645534fa6810, deletion_comment = False)

PaulTFreedman:
I'm not sure why it fails the code coverage part, it happens locally too but the output doesn't give me any useful error message. I can try to debug when I have time but if someone else has any ideas that would be very helpful.

Ivanidzo4ka:
@PaulTFreedman  it looks fine now. At least I can see updated post from coverage service, and all builds in passed state. I just need to find second person to review this change. I'll make sure someone will take a look during Monday.

artidoro:
Thank you!

PR #2331, opened 2019-01-30
Fix fake schema's creation and add a test

### LINKED ISSUE #2322: Error in code FakeSchemaFactory.Create(...) throws Exception: Index was outside the bounds of the array.
Exception: Index was outside the bounds of the array. In FakeSchemaFactory.Create(...) 

There is mistake in code :
`   for (int j = 0; j < partialMetadata.Count; ++j)
                {
var metaColumnType = MakeColumnType(partialMetadata[i]);`

There should be "j" instead of "i" variable.

wschin:
Fix #2322 and add a test.


Ivanidzo4ka:
> using System;

header please

---
Refers to: test/Microsoft.ML.Tests/FakeSchemaTest.cs:1 in cf213f7. [](commit_id = cf213f7c2144367c3afe4c7564d9fc6c9a878700, deletion_comment = False)

PR #2334, opened 2019-01-30
Internalize ITransformTemplate

Ivanidzo4ka:
Towards #1995


src/Microsoft.ML.Data/Data/IDataLoader.cs
```diff
@@ -97,7 +97,8 @@ public interface IDataTransform : IDataView, ICanSaveModel
     /// Data transforms need to be able to apply themselves to a different input IDataView.
     /// This interface allows them to implement custom rebinding logic.
```
TomFinley: >logic [](start = 65, length = 5)

Until we can delete this altogether, it may be worthwhile to add a note here to explain that the preferred idioms is that things work via the new `ITransformer` mechanism, which is the successor of this sort of idiom. I might not suggest adding the `[Obsolete]` tag -- that's a bit heaviweight -- just something in the XML docstring here like, "yeah, yeah, this is here, but we're trying to get rid of it, don't write new code with it."

src/Microsoft.ML.Data/Transforms/ColumnSelecting.cs
```diff
@@ -682,7 +682,7 @@ public Row GetRow(Row input, Func<int, bool> active)
                 return new RowImpl(input, _mapper);
             }
 
-            public IDataTransform ApplyToData(IHostEnvironment env, IDataView newSource)
+            IDataTransform ITransformTemplate.ApplyToData(IHostEnvironment env, IDataView newSource)
```
TomFinley: >ITransformTemplate [](start = 27, length = 18)

I agree with you that this is probably the best policy since it even discourages internal code from using this now deprecated idiom, but just in case you are curious, it is not necessary to do this here since this nested class (`SelectColumnsDataTransform`) is itself not public.

PR #2343, opened 2019-01-31
Remove XmlIncludes

### LINKED ISSUE #2325: Clean up the XmlInclude EntryPoint attribute annotations
The `XmlInclude` EntryPoint annotation attribute was created when we needed to simultaneously document the so called runtime classes, and the autogenerated CSharpApi.cs classes. 

Now that the auto-generated C#api is gone, there is no need to have the `XmlInclude` EntryPoint annotation attribute.

So searching for `XmlInclude ` and deleting it everywhere would consider this work item complete.

mareklinka:
Remove `XmlInclude`s from entry point attributes.

Fixes #2325


sfilipi:
Looks great. Thanks for the help @mareklinka!

Issue #2263, opened 2019-01-28
LightGbmMulticlassTrainer does not train correctly


Shadow992:
### System information

- **OS version/distro**:
Windows 10 64bit

- **.NET Version (eg., dotnet --info)**: 
4.7.2 .Net Framework
ML.Net v0.9
Compiling with x64 and Release/Debug (does not matter)

### Issue

- **What did you do?**
I tried to get a simple LightGbmMulticlassTrainer working. I had have a look at several tutorials as I am quite new to this library but not to Machine Learning in general. 

- **What happened?**
When I train the dataset with LightGbmMulticlassTrainer, it will always return an accuracy of 33,3%, because it always predicts "Setosa". However what is strange is, that in general my approach works. When I use 

mlContext.MulticlassClassification.Trainers.StochasticDualCoordinateAscent()

instead of my LightGBM trainer it gives me accuracy of 100% (which is not too suprising as my testset and trainingset is the same).

### Source code / logs

            TextLoader textLoader = mlContext.Data.CreateTextReader(new TextLoader.Arguments()
            {
                Separator = ",",
                HasHeader = false,
                Column = new[]
                {
                    new TextLoader.Column("SepalLength", DataKind.R4, 0),
                    new TextLoader.Column("SepalWidth", DataKind.R4, 1),
                    new TextLoader.Column("PetalLength", DataKind.R4, 2),
                    new TextLoader.Column("PetalWidth", DataKind.R4, 3),
                    new TextLoader.Column("Label", DataKind.Text, 4)
                }
            });

            IDataView dataView = textLoader.Read(_dataPath);

            string featuresColumnName = "Features";
            var pipeline =  mlContext.Transforms.Conversion.MapValueToKey("Label")
                .Append(mlContext.Transforms.Concatenate("Features", "SepalLength", "SepalWidth", "PetalLength", "PetalWidth"))
                .Append(new LightGbmMulticlassTrainer(mlContext, "Label","Features"))
                .Append(mlContext.Transforms.Conversion.MapKeyToValue("PredictedLabel"));

            // Train the model.
            var model = pipeline.Fit(dataView);

            // Compute quality metrics
            var metrics = mlContext.MulticlassClassification.Evaluate(model.Transform(dataView));
            Console.WriteLine(metrics.AccuracyMacro);

Ivanidzo4ka:
Hello @Shadow992 
I've tried to reproduce your problem, but I get proper metrics and predictions.
My project is in attachment:
[...]

PR #2345, opened 2019-01-31
Handle lightgbm output in different locales properly

Ivanidzo4ka:
fixes https://github.com/dotnet/machinelearning/issues/2263


Ivanidzo4ka:
@shauheen  I would like to make this PR part of 0.10. It's a nasty bug which prevents anyone with comma as decimal separator from using LightGBM.

PR #2347, opened 2019-01-31
Updating the netcoreapp sdk to latest version

Anipik:


DotnetCLIVersion.netcoreapp.latest.txt
```diff
@@ -1 +1 @@
-3.0.100-preview-009812
\ No newline at end of file
+3.0.100-preview-010195
```
eerhardt: https://github.com/dotnet/core/blob/master/release-notes/3.0/preview/3.0.0-preview2.md

Says that the preview2 version is:

https://download.visualstudio.microsoft.com/download/pr/918c5936-1c58-4a6a-b738-0a9766ec6192/4f17576e7efe45c447a088adda08b2f9/dotnet-sdk-3.0.100-preview-010184-win-x64.zip

I think we should be using that since it is a publicly released build.

PR #2360, opened 2019-02-01
Fix locale-sensitivity for float.Parse calls in CpuMathUtilsUnitTests

### LINKED ISSUE #2342: Tests in CpuMathUtilsUnitTests are culture sensitive
Hello all. While working on #2325 I noticed that tests in class `CpuMathUtilsUnitTests` seem to be locale-sensitive. My computer uses Slovak locale, which uses comma (`,`) as a decimal separator. Simply running these tests leads to failures with exception message `Input string was not in a correct format.` This is coming from `float.Parse` calls.

Setting `CultureInfo.CurrentCulture = CultureInfo.InvariantCulture;` in the class constructor resolves the failures.

This should probably be resolved to lower friction for non-English contributors.

mareklinka:
As discussed in #2342, the tests in `CpuMathUtilsUnitTests` are culture-sensitive due to `float.Parse` calls which do not specify format providers.

I updated the calls to be `float.Parse(x, CultureInfo.InvariantCulture)`. Now the tests pass even on my Slovak locale which uses comma (`,`) as a decimal separator.

Fixes #2342

### DIFF
--- a/test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
+++ b/test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
@@ -272,7 +272,7 @@ namespace Microsoft.ML.CpuMath.UnitTests
             RemoteExecutor.RemoteInvoke((arg0, arg1, arg2) =>
             {
                 CheckProperFlag(arg0);
-                float defaultScale = float.Parse(arg2, NumberStyles.Any, CultureInfo.InvariantCulture);
+                float defaultScale = float.Parse(arg2, CultureInfo.InvariantCulture);
                 float[] dst = (float[])_testArrays[int.Parse(arg1)].Clone();
                 float[] expected = (float[])dst.Clone();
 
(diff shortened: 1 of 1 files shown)


test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
```diff
@@ -271,7 +272,7 @@ public void AddScalarUTest(string mode, string test, string scale, Dictionary<st
             RemoteExecutor.RemoteInvoke((arg0, arg1, arg2) =>
             {
                 CheckProperFlag(arg0);
-                float defaultScale = float.Parse(arg2);
+                float defaultScale = float.Parse(arg2, NumberStyles.Any, CultureInfo.InvariantCulture);
```
eerhardt: does it work to just use

```C#
float.Parse(arg2, CultureInfo.InvariantCulture);
```

?

mareklinka: Ugh, missed that particular overload. Will update.

PR #2362, opened 2019-02-01
Add ResizingMode.Fill to ImageResizerTransformer

### LINKED ISSUE #2022: Convolutional net gives different results under ML.NET and Keras/TensorFlow
### System information

- **OS version/distro**: Windows 10 1809
.NET Core SDK (reflecting any global.json):
 Version:   2.2.100
 Commit:    b9f2fa0ca8

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17763
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.2.100\

Host (useful for support):
  Version: 2.2.0
  Commit:  1249f08fed

.NET Core SDKs installed:
  2.1.202 [C:\Program Files\dotnet\sdk]
  2.1.500 [C:\Program Files\dotnet\sdk]
  2.1.502 [C:\Program Files\dotnet\sdk]
  2.2.100 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.0 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.2.0 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.0.9 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.6 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.2.0 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]

### Issue

I'm facing an issue while trying to score my Keras-trained, TensorFlow-backed model using ML.NET 0.8.

#### The ML task
This is a rather small convolutional neural net that inputs an image and attempts to identify two areas of interest in the image. We are attempting to read some data off a hardware device.

#### The model

We have a trained Keras model (.5h) and we converted it into a frozen TensorFlow model (.pb). The net contains 4 convolutional layers, 1 dense layer (the output), and uses ReLU activations. The output of the model is an array of 8 numbers: [X1_1, Y1_1, X2_1, Y2_1, X2_1, Y2_1, X2_2, Y2_2]. These are the top-left and bottom-right corners of the two areas of interest.

Netron diagram:

current_model_best

This model works very well for our use case when we execute it using Keras.

#### the ML.NET scoring setup

Since our main application is written in C#, I tried to use ML.NET to access the predictions from C# code. The pipeline is setup like this:

``` csharp
var pipeline =
    new ImageLoadingEstimator(env, imagePath, (nameof(AreaInput.Path), "ImageData"))
    .Append(new ImageResizingEstimator(env, "ImageData", "ImageResized", Areas.ImageWidth,
        Areas.ImageHeight))
    .Append(new ImagePixelExtractingEstimator(env,
        new ImagePixelExtractorTransform.ColumnInfo(
            "ImageResized", Areas.InputLayerName,
            ImagePixelExtractorTransform.ColorBits.Rgb, interleave: true,
            scale: 1 / 255f)))
    .Append(new TensorFlowEstimator(env, modelPath, new[] {Areas.InputLayerName},
        new[] {Areas.OutputLayerName}));
```

On the surface, we do the same thing as with Keras:
1. Load an image from the file system
2. Resize the image to 400x400 (the model's expected input size)
3. Convert the image to a vector (dimmensions of (batch_size, 400, 400, 3) - we are using RGB mode)
4. Feed this vector into the TF model

Then we make a predictor function from this pipeline and use that to score examples.

#### The issue

There are actually two issues with this. One is a problem with interoperability - the `ImageResizingEstimator` is doing something very different than Keras' `load_img`:

* `load_img(x, target_size=(400,400), color_mode="rgb")` loads the image in RGB and resizes it to 400x400. The image ends up "squeezed" into a square
* `ImageResizingEstimator to 400x400` either pads or crops the source image to be 400x400. The result of this operation is therefore very different.
* Resizing the source images to 400x400 BEFORE feeding them into the ImageResizingEstimator actually leads to an `ArgumentException` - the resulting bitmap is invalid when it gets to the `ImagePixelExtractingEstimator` stage.

This is mainly an issue with documentation, I guess - it would be very helpful to declare this behavior explicitly, especially for interoperability scenarios such as this. At any rate, while realizing this, we went and resized our training images to 401x401 (to bypass the third bullet point issue) and retrained the model. 

Our theory was that will both allow the resizer to do its job without exception and will avoid the difference in behavior of the resize behavior - since resizing from 401x401 to 400x400 does not change the bitmap too much. So the expectation was that feeding the same 401x401 image to Keras and ML.NET should produce the same predictions.

This did not happen. The ML.NET code seems to be working and produces predictions, however these predictions are way off of the Keras results (the Keras/TF results are actually correct which we proved by drawing the resulting coordinates on the images).

So the problem here is: is our TF setup wrong? Or is this some underlying difference in behavior which we can't see? I'm attaching a repro code as well. The ML.NET part is a .NET Core 2.2 project. To run the Keras part, you'll need the whole Keras stack installed (Keras + Tensorflow). The models and some sample images are included as well.

Sample output demonstrating the issue:

ML.NET predictions for three images:
78,151,124,196,157,152,208,195
83,160,131,208,167,162,220,207
82,161,136,212,174,163,229,211

Keras output for the same three images:
[96, 182, 166, 247, 218, 179, 291, 247]
[97, 186, 169, 253, 222, 182, 297, 252]
[99, 252, 172, 314, 225, 249, 294, 311]

Some additional details:
We tested this both against TensorFlow and Tensorflow-GPU, the issue persists. Therefore it's not a problem of running on GPU vs CPU. One possible explanation would be that I simply froze the TF graph wrong by specifying an incorrect final node, but I'm pretty sure TensorBoard is telling me that `final_layer/BiasAdd` is my final node:
image

Any help with tracing this down would be greatly appreciated - operationalizing a TF model directly from C# is an awesome capability to have.

#### Source
[MLNetConvNetRepro.zip](https://github.com/dotnet/machinelearning/files/2727391/MLNetConvNetRepro.zip)

mareklinka:
This PR adds a new resizing mode to the `ImageResizerTransformer`. This mode is called `Fill` and when used, the transformer ignores aspect ratio and squeezes/stretches the source image into the target dimensions.

The idea for this mode came up in #2022 - when attempting to score a Keras-trained model it turned out that Keras resizes images in a way that ML.NET cannot do on its own. To increase interoperability, the Fill mode was proposed.

I also included a test for the new mode. This tests uses a specially prepared image, runs the resizing transformation, and then checks whether the result fulfills the specification for the resizing mode.

Fixes #2022


Ivanidzo4ka:
@mareklinka Can you please unskip locally RegenerateEntryPointCatalog test, run it, and push changes in core_ep-list.tsv?

test/Microsoft.ML.Tests/ImagesTests.cs
```diff
@@ -710,5 +711,59 @@ public void TestBackAndForthConversionWithoutAlphaNoInterleaveNoOffset()
                 Done();
             }
         }
+
+        [Fact]
+        public void ImageResizerTransform_ResizingModeFill()
```
Ivanidzo4ka: >ImageResizerTransform_ResizingModeFill [](start = 20, length = 38)

Some of us have strong feelings regarding Python_style_of_defining_names, can you forgive us, and change it to C# style of `ImageResizerTransformResizingModeFill` ?

mareklinka: No problem 😄

mareklinka:
> 
> 
> @mareklinka Can you please unskip locally RegenerateEntryPointCatalog test, run it, and push changes in core_ep-list.tsv?

Is it possible the test only changed `core_manifest.json`? I see no changes in `core_ep-list.tsv`.

test/Microsoft.ML.Tests/ImagesTests.cs
```diff
@@ -710,5 +711,59 @@ public void TestBackAndForthConversionWithoutAlphaNoInterleaveNoOffset()
...
+                var bottomRight = bitmap.GetPixel(bitmap.Width - 1, bitmap.Height - 1);
+                var middle = bitmap.GetPixel(bitmap.Width / 2, bitmap.Height / 2);
+
+                // these points must be red
+                var midTop = bitmap.GetPixel(bitmap.Width / 2, bitmap.Height / 3);
+                var midBottom = bitmap.GetPixel(bitmap.Width / 2, bitmap.Height / 3 * 2);
+                var leftMid = bitmap.GetPixel(bitmap.Width / 3, bitmap.Height / 2);
+                var rightMid = bitmap.GetPixel(bitmap.Width / 3 * 2, bitmap.Height / 2);
+
+                Assert.All(new[] { topLeft, topRight, bottomLeft, bottomRight, middle }, c =>
```
Ivanidzo4ka: >  Assert.All(new[] { topLeft, topRight, bottomLeft, bottomRight, middle }, c => [](start = 14, length = 79)

The joy of multi environment support. Mac OS and Linux fails because some of values become 254 instead of 255.

Ivanidzo4ka:
Sorry, sometimes I forget which one is which. If you have changes in arguments it's manifest if you have changes in class names it's list. 

---
In reply to: [459814339](https://github.com/dotnet/machinelearning/pull/2362#issuecomment-459814339) [](ancestors = 459814339)

test/Microsoft.ML.Tests/ImagesTests.cs
mareklinka: Hm, weird. Some kind of rounding error, I guess. I'll rewrite the test to be a little more permissive.

src/Microsoft.ML.ImageAnalytics/ImageResizerTransform.cs
```diff
@@ -44,7 +44,10 @@ public enum ResizingKind : byte
             IsoPad = 0,
 
             [TGUI(Label = "Isotropic with Cropping")]
-            IsoCrop = 1
+            IsoCrop = 1,
+
```
artidoro: Would be great if you could a summary tag above these options, this will make it show in the documentation online and in the intellisense for future users.

```suggestion

/// <summary>
/// Ignore aspect ratio and squeeze/stretch into target dimensions.
/// </summary>
```

src/Microsoft.ML.ImageAnalytics/ImageResizerTransform.cs
```diff
@@ -44,7 +44,10 @@ public enum ResizingKind : byte
             IsoPad = 0,
```
artidoro:
```suggestion

           /// <summary>
           /// Isotropic with cropping.
           /// </summary>
```

src/Microsoft.ML.ImageAnalytics/ImageResizerTransform.cs
```diff
@@ -44,7 +44,10 @@ public enum ResizingKind : byte
...
```
artidoro:
```suggestion
            /// <summary>
            /// Isotropic with padding.
            /// </summary>
            IsoPad = 0,
```

artidoro:
Thanks again!

PR #2369, opened 2019-02-01
Address minor comments in #2243

wschin:
This paper continues addressing some small comments left in #2243.

### DIFF
--- a/src/Microsoft.ML.FastTree/RegressionTree.cs
+++ b/src/Microsoft.ML.FastTree/RegressionTree.cs
@@ -56,7 +56,9 @@ namespace Microsoft.ML.FastTree
         /// Note that the case (1) happens only when <see cref="CategoricalSplitFlags"/>[i] is true and otherwise (2)
         /// occurs. A non-negative returned value means a node (i.e., not a leaf); for example, 2 means the 3rd node in
         /// the underlying <see cref="_tree"/>. A negative returned value means a leaf; for example, -1 stands for the
-        /// first leaf in the underlying <see cref="_tree"/>.
+        /// <see langword="~"/>(-1)-th leaf in the underlying <see cref="_tree"/>. Note that <see langword="~"/> is the
+        /// bitwise complement operator in C#; for details, see
+        /// https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-complement-operator.
         /// </summary>
         public IReadOnlyList<int> LteChild => _lteChild;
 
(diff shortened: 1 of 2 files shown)


shauheen:
> As title.

Thanks @wschin , even though this is a small PR, please edit the description and at least mention the other PR here so that they are actually linked.

wschin:
@shauheen, they are linked. If you go to #2243, there is a footprint.

src/Microsoft.ML.FastTree/RegressionTree.cs
```diff
@@ -56,7 +56,9 @@ public abstract class RegressionTreeBase
         /// Note that the case (1) happens only when <see cref="CategoricalSplitFlags"/>[i] is true and otherwise (2)
         /// occurs. A non-negative returned value means a node (i.e., not a leaf); for example, 2 means the 3rd node in
         /// the underlying <see cref="_tree"/>. A negative returned value means a leaf; for example, -1 stands for the
-        /// first leaf in the underlying <see cref="_tree"/>.
+        /// <see langword="~"/>(-1)-th leaf in the underlying <see cref="_tree"/>. Note that <see langword="~"/> is the
+        /// bitwise complement operator in C#; for details, see
+        /// https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-complement-operator.
```
sfilipi: see <a href="[https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-complement-operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-complement-operator)">the documentation about the bitwise complement operator </a>

Ivanidzo4ka:
>         public IReadOnlyList<double> TreeWeights { get; }

I think I asked this question in previous PR, but I can't find that any more, so I ask again.
I don't see them been used during prediction, why we expose them?

---
Refers to: src/Microsoft.ML.FastTree/TreeEnsemble.cs:25 in 2eb1ef2. [](commit_id = 2eb1ef29569c0b3d1043226852831c083b1dcd73, deletion_comment = False)

TomFinley:
> @TomFinley, do you have any idea about exposing trees' weights to users? They are extracted from `InternalRegressionTree` and stored in (new) `TreeEnsemble`.

Trees definitely do have weights attached to them. If some code was written that does not account for them, then we ought to address that. Now, *functionally* I expect this won't have an effect since the weights are under most situations 1, but, that's besides the point.

Here's an example where the weights seem to be taken into account. You can find other usages here and there.

https://github.com/dotnet/machinelearning/blob/40dc35bea851c415911726c36712321b66dafe8b/src/Microsoft.ML.FastTree/TreeEnsemble/InternalTreeEnsemble.cs#L180

I think having weights on the trees themselves is useful, and we should continue to have them.

PR #2382, opened 2019-02-02
MLContext catalog cleanup

shmoradims:
End-users will use ML.NET mainly through MLContext. As part of documentation cleanup, I started from MLContext and cleaned up the top level catalogs as follows:

* Internal catalog constructors: All catalog constructors are made internal, since the user shouldn't be creating those objects.
* `protected internal` for some fields: doc.microsoft includes `protected` fields/methods (like [Host field here](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.binaryclassificationcontext?view=ml-dotnet#fields)). Since all our catalogs are sealed, I changed some `protected` fields/method to `protected internal` so that they don't get mentioned in the docs at all. 
* Xml cleanup: Fixed some incorrect/missing texts along the way.

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/TransformsCatalog.cs
+++ b/src/Microsoft.ML.Data/Transforms/TransformsCatalog.cs
@@ -12,10 +12,29 @@ namespace Microsoft.ML
     {
         internal IHostEnvironment Environment { get; }
 
+        /// <summary>
+        /// The list of operations over categorical data.
+        /// </summary>
         public CategoricalTransforms Categorical { get; }
+
+        /// <summary>
+        /// The list of operations for rescaling data.
+        /// </summary>
         public ConversionTransforms Conversion { get; }
+
+        /// <summary>
+        /// The list of operations for processing text data.
+        /// </summary>
         public TextTransforms Text { get; }
+
+        /// <summary>
+        /// The list of operations for data projection.
+        /// </summary>
         public ProjectionTransforms Projection { get; }
+
+        /// <summary>
+        /// The list of operations for selecting features based on some criteria.
+        /// </summary>
         public FeatureSelectionTransforms FeatureSelection { get; }
 
         internal TransformsCatalog(IHostEnvironment env)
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.Data/Transforms/TransformsCatalog.cs
```diff
@@ -12,10 +12,29 @@ public sealed class TransformsCatalog
     {
         internal IHostEnvironment Environment { get; }
 
+        /// <summary>
+        /// The list of operations over categorical data.
+        /// </summary>
         public CategoricalTransforms Categorical { get; }
+
+        /// <summary>
+        /// The list of operations for rescaling data.
```
Ivanidzo4ka: >rescaling data [](start = 39, length = 14)

`data conversion` maybe `casting data` or `data casting`? Hashing, TypeConversion is hard to call rescaling operation.

shmoradims: Good catch. This is actually an incorrect description left-over from when normalizers were here. I copy-pasted it, from the catalog, assuming it was correct. Will update it.

---
In reply to: [253240931](https://github.com/dotnet/machinelearning/pull/2382#discussion_r253240931) [](ancestors = 253240931)

src/Microsoft.ML.Data/Transforms/TransformsCatalog.cs
```diff
@@ -12,10 +12,29 @@ public sealed class TransformsCatalog
...
+        /// </summary>
         public TextTransforms Text { get; }
+
+        /// <summary>
+        /// The list of operations for data projection.
+        /// </summary>
         public ProjectionTransforms Projection { get; }
+
+        /// <summary>
+        /// The list of operations for selecting features based on some criteria.
```
zeahmed: >based on some criteria [](start = 58, length = 22)

I feel its unnecessary here because for rest of the properties we don't say any details.

PR #2383, opened 2019-02-02
Internalize IDataTransform

Ivanidzo4ka:
towards #1995

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/ColumnSelecting.cs
+++ b/src/Microsoft.ML.Data/Transforms/ColumnSelecting.cs
@@ -394,13 +394,15 @@ namespace Microsoft.ML.Transforms
             return transform.Transform(input);
         }
 
-        public static IDataTransform CreateKeep(IHostEnvironment env, IDataView input, string[] keepColumns, bool keepHidden = false)
+        [BestFriend]
+        internal static IDataTransform CreateKeep(IHostEnvironment env, IDataView input, string[] keepColumns, bool keepHidden = false)
         {
             var transform = new ColumnSelectingTransformer(env, keepColumns, null, keepHidden);
             return new SelectColumnsDataTransform(env, transform, new Mapper(transform, input.Schema), input);
         }
 
-        public static IDataTransform CreateDrop(IHostEnvironment env, IDataView input, params string[] dropColumns)
+        [BestFriend]
+        internal static IDataTransform CreateDrop(IHostEnvironment env, IDataView input, params string[] dropColumns)
         {
             var transform = new ColumnSelectingTransformer(env, null, dropColumns);
             return new SelectColumnsDataTransform(env, transform, new Mapper(transform, input.Schema), input);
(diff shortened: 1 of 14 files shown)


src/Microsoft.ML.Data/Transforms/ColumnSelecting.cs
```diff
@@ -394,13 +394,15 @@ public static IDataView Create(IHostEnvironment env, ModelLoadContext ctx, IData
...
+        [BestFriend]
+        internal static IDataTransform CreateKeep(IHostEnvironment env, IDataView input, string[] keepColumns, bool keepHidden = false)
         {
             var transform = new ColumnSelectingTransformer(env, keepColumns, null, keepHidden);
             return new SelectColumnsDataTransform(env, transform, new Mapper(transform, input.Schema), input);
         }
 
-        public static IDataTransform CreateDrop(IHostEnvironment env, IDataView input, params string[] dropColumns)
+        [BestFriend]
+        internal static IDataTransform CreateDrop(IHostEnvironment env, IDataView input, params string[] dropColumns)
```
artidoro: In my PR, I am making these private :)

artidoro:
>     {

What are we doing with these two? 

---
Refers to: src/Microsoft.ML.Data/Transforms/SkipTakeFilter.cs:270 in 9e7e888. [](commit_id = 9e7e8883a3f56735dbf68649fd3ef6becf52aa5a, deletion_comment = False)

Ivanidzo4ka:
>     {

Well, we need to create extension methods for DataOperationCatalog so user can call mlContext.Data.Take() and mlContext.Data.Skip()
After that I think it's safe to kill them.

---
In reply to: [459914466](https://github.com/dotnet/machinelearning/pull/2383#issuecomment-459914466) [](ancestors = 459914466)

---
Refers to: src/Microsoft.ML.Data/Transforms/SkipTakeFilter.cs:270 in 9e7e888. [](commit_id = 9e7e8883a3f56735dbf68649fd3ef6becf52aa5a, deletion_comment = False)

src/Microsoft.ML.Data/Transforms/ColumnSelecting.cs
Ivanidzo4ka: I think @rogancarr wanted mlContext.SchemaOperations catalog, so I would definitely add them to that catalog. 
But I don't understand how you can make them private.

btw, I should change type from `IDataTransform` to `IDataView`.

---
In reply to: [253240652](https://github.com/dotnet/machinelearning/pull/2383#discussion_r253240652) [](ancestors = 253240652)

artidoro: (I replaced them with new Transformer().Transform())

---
In reply to: [253245274](https://github.com/dotnet/machinelearning/pull/2383#discussion_r253245274) [](ancestors = 253245274,253240652)

Issue #77, opened 2018-05-08
Where is SentimentData defined?


zwcloud:
### System information

- **OS version/distro**: Win10
- **.NET Version (eg., dotnet --info)**:  2.1.200

### Issue

- **What did you do?**
I'm try the [examples](https://github.com/dotnet/machinelearning#examples) in the readme.

- **What happened?**

I installed Microsoft.ML package to a newly created .Net core console app.

And in the `Program.cs` file,

```
using System;
using Microsoft.ML;
using Microsoft.ML.Trainers;
using Microsoft.ML.Transforms;

namespace MLTry
{
    class Program
    {
        static void Main(string[] args)
        {
            var pipeline = new LearningPipeline();
            pipeline.Add(new TextLoader<SentimentData>(dataPath, separator: ","));
//...
```

It doesn't compile because `SentimentData` is not defined. And I searched for `SentimentData` in the repo but found nothing related.

- **What did you expect?**
Found the type `SentimentData`.

nevoroman:
SentimentData is just an input dataset class. Create and use a class that fits your data.
You can find the full tutorial for this code sample at docs.microsoft.com (https://docs.microsoft.com/ru-ru/dotnet/machine-learning/tutorials/sentiment-analysis)

PR #76, opened 2018-05-08
Fix broken link to sentiment prediction scenario source code.

rnrneverdies:
Current link points to an unexisting file (Scenario3_SentimentPrediction.cs), so I used the most probable sample file. 

We are excited to review your PR.

So we can do the best job, please check:

- [x ] There's a descriptive title that will make sense to other developers some time from now. 
- [o] There's associated issues. All PR's should have issue(s) associated - unless a trivial self-evident change such as fixing a typo. You can use the format `Fixes #nnnn` in your description to cause GitHub to automatically close the issue(s) when your PR is merged.
- [x] Your change description explains what the change does, why you chose your approach, and anything else that reviewers should know.
- [ ] You have included any necessary tests in the same PR.

PR #2386, opened 2019-02-02
Fixed output schema produced by ValueMappingEstimator

### LINKED ISSUE #2385: Fix output schema produced by ValueMappingEstimator
There is a bug in `ValueMappingEstimator` output schema. 
https://github.com/dotnet/machinelearning/blob/0e80d6c07580cea5ce072fd7cba02d7171dc61f4/src/Microsoft.ML.Data/Transforms/ValueMapping.cs#L61

When input is a vector the output produced is a vector. However, the schema produced by the `ValueMappingEstimator` shows that the output is scalar. This throws an exception when `ValueMappingEstimator` is followed by another estimator which checks for schema.

This regression was caused by #2162

zeahmed:
This PR fixes #2385.

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/ValueMapping.cs
+++ b/src/Microsoft.ML.Data/Transforms/ValueMapping.cs
@@ -73,9 +73,13 @@ namespace Microsoft.ML.Transforms.Conversions
                 if (!inputSchema.TryFindColumn(inputColumnName, out var originalColumn))
                     throw Host.ExceptSchemaMismatch(nameof(inputSchema), "input", inputColumnName);
 
-                if ((originalColumn.Kind == SchemaShape.Column.VectorKind.VariableVector ||
-                   originalColumn.Kind == SchemaShape.Column.VectorKind.Vector) && Transformer.ValueColumnType is VectorType)
-                    throw Host.ExceptNotSupp("Column '{0}' cannot be mapped to values when the column and the map values are both vector type.", inputColumnName);
+                if (originalColumn.Kind == SchemaShape.Column.VectorKind.VariableVector ||
+                   originalColumn.Kind == SchemaShape.Column.VectorKind.Vector)
+                {
+                    if (Transformer.ValueColumnType is VectorType)
+                        throw Host.ExceptNotSupp("Column '{0}' cannot be mapped to values when the column and the map values are both vector type.", inputColumnName);
+                    vectorKind = originalColumn.Kind;
+                }
                 // Create the Value column
                 var col = new SchemaShape.Column(outputColumnName, vectorKind, columnType, isKey, metadataShape);
                 resultDic[outputColumnName] = col;
--- a/test/Microsoft.ML.Tests/Transformers/ValueMappingTests.cs
+++ b/test/Microsoft.ML.Tests/Transformers/ValueMappingTests.cs
@@ -89,6 +89,9 @@ namespace Microsoft.ML.Tests.Transformers
             var estimator = new WordTokenizingEstimator(Env, new[]{
                     new WordTokenizingTransformer.ColumnInfo("TokenizeA", "A")
                 }).Append(new ValueMappingEstimator<ReadOnlyMemory<char>, int>(Env, keys, values, new[] { ("VecD", "TokenizeA"), ("E", "B"), ("F", "C") }));
+            var schema = estimator.GetOutputSchema(SchemaShape.Create(dataView.Schema));
+            Assert.True(schema.TryFindColumn("VecD", out var originalColumn));
+            Assert.Equal(SchemaShape.Column.VectorKind.VariableVector, originalColumn.Kind);
             var t = estimator.Fit(dataView);
 
             var result = t.Transform(dataView);


src/Microsoft.ML.Data/Transforms/ValueMapping.cs
```diff
@@ -73,9 +73,13 @@ public override SchemaShape GetOutputSchema(SchemaShape inputSchema)
...
 
-                if ((originalColumn.Kind == SchemaShape.Column.VectorKind.VariableVector ||
-                   originalColumn.Kind == SchemaShape.Column.VectorKind.Vector) && Transformer.ValueColumnType is VectorType)
-                    throw Host.ExceptNotSupp("Column '{0}' cannot be mapped to values when the column and the map values are both vector type.", inputColumnName);
+                if (originalColumn.Kind == SchemaShape.Column.VectorKind.VariableVector ||
+                   originalColumn.Kind == SchemaShape.Column.VectorKind.Vector)
+                {
+                    if (Transformer.ValueColumnType is VectorType)
+                        throw Host.ExceptNotSupp("Column '{0}' cannot be mapped to values when the column and the map values are both vector type.", inputColumnName);
+                    vectorKind = originalColumn.Kind;
```
zeahmed: >vectorKind = originalColumn.Kind; [](start = 20, length = 33)

When input is a vector, output should always be a vector. Otherwise, output is of ValueType kind.

test/Microsoft.ML.Tests/Transformers/ValueMappingTests.cs
```diff
@@ -89,6 +89,9 @@ public void ValueMapInputIsVectorTest()
             var estimator = new WordTokenizingEstimator(Env, new[]{
                     new WordTokenizingTransformer.ColumnInfo("TokenizeA", "A")
                 }).Append(new ValueMappingEstimator<ReadOnlyMemory<char>, int>(Env, keys, values, new[] { ("VecD", "TokenizeA"), ("E", "B"), ("F", "C") }));
+            var schema = estimator.GetOutputSchema(SchemaShape.Create(dataView.Schema));
+            Assert.True(schema.TryFindColumn("VecD", out var originalColumn));
+            Assert.Equal(SchemaShape.Column.VectorKind.VariableVector, originalColumn.Kind);
```
abgoswam: >originalColumn [](start = 71, length = 14)

can u also add test for values for this vector.

i believe it should be <2, 3 1> ?

zeahmed: If you look at the following code, it does exactly what meant.

---
In reply to: [253622276](https://github.com/dotnet/machinelearning/pull/2386#discussion_r253622276) [](ancestors = 253622276)

src/Microsoft.ML.Data/Transforms/ValueMapping.cs
```diff
@@ -73,9 +73,13 @@ public override SchemaShape GetOutputSchema(SchemaShape inputSchema)
...
                     throw Host.ExceptSchemaMismatch(nameof(inputSchema), "input", inputColumnName);
 
-                if ((originalColumn.Kind == SchemaShape.Column.VectorKind.VariableVector ||
-                   originalColumn.Kind == SchemaShape.Column.VectorKind.Vector) && Transformer.ValueColumnType is VectorType)
-                    throw Host.ExceptNotSupp("Column '{0}' cannot be mapped to values when the column and the map values are both vector type.", inputColumnName);
+                if (originalColumn.Kind == SchemaShape.Column.VectorKind.VariableVector ||
+                   originalColumn.Kind == SchemaShape.Column.VectorKind.Vector)
+                {
+                    if (Transformer.ValueColumnType is VectorType)
+                        throw Host.ExceptNotSupp("Column '{0}' cannot be mapped to values when the column and the map values are both vector type.", inputColumnName);
```
abgoswam: >when the column and the map values are both vector type. [](start = 90, length = 56)

i am not sure i understand this . could you give an example where this condition would be hit ?

src/Microsoft.ML.Data/Transforms/ValueMapping.cs
```diff
@@ -73,9 +73,13 @@ public override SchemaShape GetOutputSchema(SchemaShape inputSchema)
                 if (!inputSchema.TryFindColumn(inputColumnName, out var originalColumn))
                     throw Host.ExceptSchemaMismatch(nameof(inputSchema), "input", inputColumnName);
 
-                if ((originalColumn.Kind == SchemaShape.Column.VectorKind.VariableVector ||
-                   originalColumn.Kind == SchemaShape.Column.VectorKind.Vector) && Transformer.ValueColumnType is VectorType)
-                    throw Host.ExceptNotSupp("Column '{0}' cannot be mapped to values when the column and the map values are both vector type.", inputColumnName);
+                if (originalColumn.Kind == SchemaShape.Column.VectorKind.VariableVector ||
+                   originalColumn.Kind == SchemaShape.Column.VectorKind.Vector)
```
abgoswam: it seems the bug is that we have not updated the `vectorKind`  when the `originalColumn.Kind == SchemaShape.Column.VectorKind.VariableVector`

As such , instead of changing this logic should we just update the vectorKind ? i.e. add the lines:

 if (originalColumn.Kind == SchemaShape.Column.VectorKind.VariableVector)
                    vectorKind = SchemaShape.Column.VectorKind.VariableVector;

src/Microsoft.ML.Data/Transforms/ValueMapping.cs
zeahmed: This is in a situation where column={"a", "b", "c"} and map={"a": [1,2,3], "b":[4,5], "c":[6]}
the resulting column will be output_col={[1,2,3], [4,5], [6]}


---
In reply to: [253629485](https://github.com/dotnet/machinelearning/pull/2386#discussion_r253629485) [](ancestors = 253629485)

src/Microsoft.ML.Data/Transforms/ValueMapping.cs
zeahmed: I think this what's happening right now both for fixed length and variable length vector.

---
In reply to: [253654474](https://github.com/dotnet/machinelearning/pull/2386#discussion_r253654474) [](ancestors = 253654474)

test/Microsoft.ML.Tests/Transformers/ValueMappingTests.cs
```diff
@@ -89,6 +89,9 @@ public void ValueMapInputIsVectorTest()
             var estimator = new WordTokenizingEstimator(Env, new[]{
                     new WordTokenizingTransformer.ColumnInfo("TokenizeA", "A")
                 }).Append(new ValueMappingEstimator<ReadOnlyMemory<char>, int>(Env, keys, values, new[] { ("VecD", "TokenizeA"), ("E", "B"), ("F", "C") }));
+            var schema = estimator.GetOutputSchema(SchemaShape.Create(dataView.Schema));
+            Assert.True(schema.TryFindColumn("VecD", out var originalColumn));
```
artidoro: Could you add a `TestEstimatorCore`, to your tests for `ValueMap`? Should be pretty quick and I think it would help make sure there are no further inconsistencies.

zeahmed: hmm... none of the test in this file use `TestEstimatorCore`. So, I would avoid that here too.
Also, checks made here are explicitly testing specific logic that I think cannot be tested with `TestEstimatorCore`.

---
In reply to: [253658781](https://github.com/dotnet/machinelearning/pull/2386#discussion_r253658781) [](ancestors = 253658781)

artidoro: The reason I think it's important to have `TestEstimatorCore` is that it checks that the schema in` GetOutpuSchema` matches the schema that is actually produced. Since you found a bug in the output schema, I think it would be good to have this special case of an vector input go through the `TestEstimatorCore`. There is a test that uses `TestEstimatorCore` bellow: `ValueMappingWorkout`. 

---
In reply to: [253661005](https://github.com/dotnet/machinelearning/pull/2386#discussion_r253661005) [](ancestors = 253661005,253658781)

src/Microsoft.ML.Data/Transforms/ValueMapping.cs
abgoswam: thanks for the example.

so we shouldn't support  a scenario  where Transformer.ValueColumnType is VectorType
i.e   map={"a": [1,2,3], "b":[4,5], "c":[6]}

but then i do not understand  why we need the  condition in line #76-#77 ?  

if input is not vector :   
e.g.  
column="a"  => output_col = {[1,2, 3]}
column="b" =>  output_col = {[4,5]}

maybe i am missing something ?

---
In reply to: [253656971](https://github.com/dotnet/machinelearning/pull/2386#discussion_r253656971) [](ancestors = 253656971,253629485)

zeahmed: We do support the scenario where Transformer.ValueColumnType is VectorType. However, we don't support scenario where both input and Transformer.ValueColumnType is VectorType are of vector type. 

---
In reply to: [253665022](https://github.com/dotnet/machinelearning/pull/2386#discussion_r253665022) [](ancestors = 253665022,253656971,253629485)

test/Microsoft.ML.Tests/Transformers/ValueMappingTests.cs
singlis: @zeahmed please see here:
https://github.com/dotnet/machinelearning/blob/2dfc51aa7172199c1b2145e48ecb9eb855bb87f1/test/Microsoft.ML.Tests/Transformers/ValueMappingTests.cs#L506

You can add to this test by adding a new DataView that produces an array. Then the TestEstimatorCore would validate as what @artidoro  is suggesting.

zeahmed: Thanks for the explanation @artidoro and @singlis, I think this is an end to end scenario where an input is a valid input and output is being tested against expected output. The semantics of `TestEstimatorCore` is different where it test the schema propagation for valid, invalid inputs for estimators and transforms. 

Adding that here in this test needlessly complicate the test and will not test the problem that is being tested between line 92-94.

I think it should be done as a separate test and that is what `ValueMappingWorkout` is doing. However, its not doing it for all the input scenarios. So, I am creating an issue to do it separately https://github.com/dotnet/machinelearning/issues/2410

zeahmed:
Thanks all for the useful feedback!

Issue #1967, opened 2018-12-24
ML.net build failing with Visual Studio 2019  Version 16.0


Anipik:
I just rebooted my machine. So while setting up the visual studio, I installed the latest Visual Studio 2019.
The visual studio version for this 16.0

So when I try to build the repo, I get an error as 
``` Error: Visual Studio 2015 or 2017 required.```

My guess is that the problem is here 
https://github.com/dotnet/machinelearning/blob/master/src/Native/build.cmd#L48

This check fails as the value of the version is 16.0

I changed this line to 
```if %VisualStudioVersion% geq 15.0 (```

and the build works fine then.

is this an appropriate fix and anything else also needs to be done here ?

cc @danmosemsft @safern @eerhardt @TomFinley

**Comment by codemzs** [comment]
You may want to do an explicit equality comparison like we do for VS 2015 and VS 2017, not just greater than equal to, what if another version gets released that needs to have different set of setup?, in addition to that you may want to consider changing/adding few more things, for example, change the error message here below:

:MissingVersion
:: Can't find VS 2015 or 2017
echo Error: Visual Studio 2015 or 2017 required
echo        Please see https://github.com/dotnet/machinelearning/tree/master/Documentation for build instructions.
exit /b 1

Just like how we have setup lines for VS2017, we probably will need something for VS 2019, no? 
:VS2017
:: Setup vars for VS2017
set __PlatformToolset=v141
set __VSVersion=15 2017
if NOT "%__BuildArch%" == "arm64" (
    :: Set the environment for the native build
    call "%VS150COMNTOOLS%..\..\VC\Auxiliary\Build\vcvarsall.bat" %__VCBuildArch%
)
goto :SetupDirs

All in all, we need to have explicit references for VS 2019 just like we have for VS 2017/2015.

**Comment by Anipik** [comment]
Actually i tried changing i to  ```set __VSVersion=16 2019```  for 2019 VS but cmake fails to recognise it and i get this as an error
```
EXEC : CMake error : Could not create named generator Visual Studio 16 2019 Win64 [C:\git\machinelearning\src\Native\build.proj]
```
i am not sure whts the correct set of values for vsversion and platformToolset for vs2019

rrmistry:
If you're trying to build using `build.cmd` on Windows (at the root level of this repository) then there is a much simpler fix.

I traced down the call sequence to:

1. `.\build.cmd`
2. `src\Native\build.cmd`
[...]

PR #2388, opened 2019-02-02
Support buld using VS2019

kant2002:
CMake shipped with VS does not support VS2019 yet, so run generator for VS 2017. It will still build using VS 2019

Closes: #1967

### DIFF
--- a/src/Native/build.cmd
+++ b/src/Native/build.cmd
@@ -47,17 +47,29 @@ set "VSCMD_START_DIR=%__currentScriptDir%"
 call "%_VSCOMNTOOLS%\VsDevCmd.bat"
 
 :RunVCVars
-if "%VisualStudioVersion%"=="15.0" (
+if "%VisualStudioVersion%"=="16.0" (
+    goto :VS2019
+) else if "%VisualStudioVersion%"=="15.0" (
     goto :VS2017
 ) else if "%VisualStudioVersion%"=="14.0" (
     goto :VS2015
 )
 
 :MissingVersion
-:: Can't find VS 2015 or 2017
-echo Error: Visual Studio 2015 or 2017 required
+:: Can't find VS 2015, 2017 or 2019
+echo Error: Visual Studio 2015, 2017 or 2019 required
 echo        Please see https://github.com/dotnet/machinelearning/tree/master/Documentation for build instructions.
 exit /b 1
+
+:VS2019
+:: Setup vars for VS2019
+set __PlatformToolset=v142
+set __VSVersion=15 2017
+if NOT "%__BuildArch%" == "arm64" (
+    :: Set the environment for the native build
+    call "%VS160COMNTOOLS%..\..\VC\Auxiliary\Build\vcvarsall.bat" %__VCBuildArch%
+)
+goto :SetupDirs
 
 :VS2017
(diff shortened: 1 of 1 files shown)


src/Native/build.cmd
```diff
@@ -47,18 +47,30 @@ set "VSCMD_START_DIR=%__currentScriptDir%"
...
-echo Error: Visual Studio 2015 or 2017 required
+:: Can't find VS 2015, 2017 or 2019
+echo Error: Visual Studio 2015, 2017 or 2019 required
 echo        Please see https://github.com/dotnet/machinelearning/tree/master/Documentation for build instructions.
 exit /b 1
 
+:VS2019
+:: Setup vars for VS2019
+set __PlatformToolset=v142
+set __VSVersion=15 2017
```
Ivanidzo4ka: >2017 [](start = 19, length = 4)

if you look into this commit:
[https://github.com/dotnet/corefx/pull/33969/files](https://github.com/dotnet/corefx/pull/33969/files)
they set VSVersion to vs2019 and platformToolsSet to v141
I'm not sure about toolset but vsversion as vs2019  make more sense.

kant2002: `__PlatformToolset=v142` is what shipped with VS2019 by default, so I choose this option. v141 could be installed there, but as well and v140 so I choose default for VS2019.
CMake 3.13 shipped with VS2019 does not support "Visual Studio 16 2019" generator, so that's why this one strange value for `__VSVersion`

shauheen:
Thanks @kant2002 for submitting this PR. Would you please file an issue and associate the issue with this PR?

src/Native/build.cmd
eerhardt: I think we should be following what `corefx` does, shouldn't we?

https://github.com/dotnet/corefx/blob/66985ca21f89e4d2230f5fcb99f4e8b3d618c9af/src/Native/build-native.cmd#L73-L79

eerhardt: @safern - thoughts?

kant2002:
@shauheen as @Ivanidzo4ka point out this is related to #1967 and I update description so that issue would be automatically closed.

src/Native/build.cmd
safern: I thought that the current Cmake version that is out doesn't yet support this because the VS2017 generator uses the v141 toolset and VS2019 Preview2 had a breaking change which changed the toolset version file name, so cmake wouldn't find the compiler. Then on preview3 that was fixed. And @kant2002 is correct, v142 is the toolset for VS2019.

In corefx we haven't been able to yet build with VS 2019 Preview2 or Preview3, until cmake 3.14 is out the door and points to the right toolset version, but this is the needed change and in corefx we will have to move the toolset to v142.

safern: Note that the __VSVersion string that we set in corefx in build-native.cmd we override it when we call cmake with the generator to use VS2017 generator for now. That will be fixed once cmake 3.14 is out of the door and supports VS2019 generator. https://github.com/dotnet/corefx/blob/master/src/Native/Windows/gen-buildsys-win.bat#L17

So basically the values that we're setting to in build-native.cmd are ignored.

artidoro:
Thank you very much @kant2002 !

PR #2367, opened 2019-02-01
Creation of components through MLContext and cleanup (Onnx, Tensorflow, SelectColumn, KeytoBinVec, ValueMap)

artidoro:
This PR is part of the work outlined in #1798, and focuses on the GcnNorm, LpNorm, RandomFourier, CustomStopWords, VectorWhiten, PCA transformers/estimators:

1. Internalize constructors of estimators and transformers
2. Keep ColumnInfo and move it to to the estimators #1760
3. Rename Arguments -> Options
4. Internalize Options only when they are not used by public constructor. For all other cases, retain Options as public #1758 

This PR is marked as WIP because I have to figure out how to make the DNN assemblies BestFriends with the ONNX assembly. I need to figure out how to do that.

PR #2392, opened 2019-02-03
Fix naming of Options argument in  TensorFlowTransform API

### LINKED ISSUE #2391: Incorrect naming of Options argument in TensorFlowTransform public api
In PR : #2312  we modified the public surface of the TensorFlowTransform  as per issue #1798 #2280 

There is a bug - we did not rename the 'args' parameter. It should be 'options'  (else it breaks consistency with rest of public surface)

https://github.com/dotnet/machinelearning/blob/5b98e0cd4cdd150029fddf3c33c6c6bd71abb119/src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs#L53-L54

@ganik

abgoswam:
Fixes #2391 

- Renamed `Options` argument of the TensorFlowTransform API to `options`


artidoro:
I think much of this work is already taken care of in #2367, let's double check and sync.

abgoswam:
Let me check this in since I am using search criteria "Options args"  for find+replace in my PRs

If you find other inconsistencies with TFTransform we can do those separately in your PR #2367.

---
In reply to: [460088646](https://github.com/dotnet/machinelearning/pull/2392#issuecomment-460088646) [](ancestors = 460088646)

PR #2395, opened 2019-02-03
Fix build

Ivanidzo4ka:
fix current build after merges


justinormont:
We could turn on the "Require branches to be up to date before merging". I think this would ensure all PRs are working after merging. The downside is (I think) this would mean only 1 PR per E2E-time-of-CI.

screen shot 2019-02-03 at 14 35 28

This is under Branch protection rules @ https://github.com/dotnet/machinelearning-automl/settings/branches

PR #2402, opened 2019-02-04
Replace ConditionalFact usages with custom facts

### LINKED ISSUE #1220: Clean up `[ConditionalFact(typeof(Environment), nameof(Environment.Is64BitProcess))]`
It was suggested that creating explicit attributes for the various cases (Such as Onnx, TensorFlow, LightGBM, etc) may be a better long term solution to manage these tests across the various platforms.

See https://github.com/dotnet/machinelearning/pull/1008#discussion_r221985876 for more details.

mareklinka:
As proposed in #1220, this PR replaces all usages of `ConditionalFact` in tests with custom-tailored, environment-specific facts.

These specific facts all inherit from a single base `EnvironmentSpecificFactAttribute`. Several of the facts check the same condition (e.g. TensorFlow, Onnx, and X64 - they all simply check for X64 process) but I decided not to merge them - they might check the same thing now but I expect they might branch out later on. I consider the small duplication of code not worth the mental overhead of inheritance or composition here.

Together with this I also removed some now-obsolete checking of platform in Onnx-related tests (they were all Windows-only anyway) and removed the old `OnnxFact`.

Fixes #1220.


mareklinka:
Ok, so it seems I stumbled upon an interesting problem here. It seems tests marked with `ConditionalAttribute` were not running at all under full framework. Now that I migrated them to custom attributes, they are being run and fail.

Specifically, the failing Onnx tests are looking for the model files in the wrong directory. Debugging into it, this is the path is returned in the `ResNet18` extension method:

`C:\Users\Marek\AppData\Local\Temp\4e765c63-f6ab-4125-b0da-31d31f556eb7\4e765c63-f6ab-4125-b0da-31d31f556eb7\assembly\dl3\dd76d9c2\39935ac9_c0bcd401\DnnImageModels`

This feels like some kind of shadow copy is going on here.

I'm still looking into the tensorflow and MacOS fail.

mareklinka:
The TensorFlow failure seems to be of the same root cause - it was not running under full framework, now it runs. However, the directory `Maml.MainWithProgress` looks at is wrong:

`C:\Users\Marek\AppData\Local\Temp\bd281e66-574e-4cfc-b712-1768a003416c\bd281e66-574e-4cfc-b712-1768a003416c\assembly\dl3\c6759ee3\bfbd336c_c7bcd401`

I have half a mind to make `EnvironmentSpecificFactAttribute` always return `false` `#if NETFRAMEWORK` to emulate the old behavior (not running these tests on full framework). Disabling shadow copying might be an option here but I'm unfamiliar with the build setup so I don't know where to look.

Could someone point my in the right direction, @tannergooding, @eerhardt?

eerhardt:
@Anipik has been looking in this same space. Adding him to comment.

test/Microsoft.ML.Predictor.Tests/TestPredictors.cs
```diff
@@ -162,7 +163,7 @@ public void EarlyStoppingTest()
         /// <summary>
         /// Multiclass Logistic Regression test.
         /// </summary>
-        [ConditionalFact(typeof(BaseTestBaseline), nameof(BaseTestBaseline.LessThanNetCore30OrNotNetCore))] // netcore3.0 output differs from Baseline
+        [LessThanNetCore30OrNotNetCoreFact]
```
Anipik: We should do a similar thing as you are doing with X64Fact (providing a reason for skip), Because this is an issue that we want to correct in the future and not a limitation of the library (eg onnx runtime is just for 64 bit).

TomFinley:
Yes. Also given the subject area (TensorFlow components) I would also invite @jignparm to possibly share his thoughts.

Anipik:
> I have half a mind to make EnvironmentSpecificFactAttribute always return false #if NETFRAMEWORK to emulate the old behavior (not running these tests on full framework)

I would prefer to disable these (failing onnx test) tests just on fullFramework, using notFullFrameworkFact or NotFullFrameworkAndx64. As most of the tests are passing using conditional Fact are passing on fullFramework.

Anipik:
I am able to fix all the resnet failures in https://github.com/dotnet/machinelearning/pull/2411

mareklinka:
Now the Core 3.0 Debug intrinsics build failed. Looking into it, it tried to run the benchmarks in DEBUG build, which should not have happened - the `BenchmarkTheoryAttribute` uses the same `#if` directives as the original - if should skip `#if DEBUG`. It looks like the DEBUG constant was not emitted during compilation of the test project?

test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -1313,7 +1314,7 @@ public void EntryPointMulticlassPipelineEnsemble()
             }
         }
 
-        [ConditionalFact(typeof(BaseTestBaseline), nameof(BaseTestBaseline.LessThanNetCore30OrNotNetCore))]
+        [LessThanNetCore30OrNotNetCoreFact("netcore3.0 output differs from Baseline")]
```
eerhardt: `netcore3.0` => `netcoreapp3.0`.  We should use the real TFM here, so we don't confuse people.

Anipik:
@eerhardt I tried to debug the benchmarkTheory failure. For some reason Debug Constant is not working in this PR. I checked the Binlog and it is getting defined.
```
Property reassignment: $(DefineConstants)=";DEBUG;" (previous value: ";DEBUG") at C:\Program Files\dotnet\sdk\3.0.100-preview-010184\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.CSharp.props (22,5)
```

I tried adding it to the project as well but it didn't work. any suggestions why ?

mareklinka:
> can you also remove this https://github.com/dotnet/machinelearning/blob/master/test/Directory.Build.props#L32 ?

The issue with this is that that particular nuget also contains the definition of `TestCategory`. I can get rid of that but we would need to replace the `TestCategory` with XUnit's own `Trait`:

``` csharp
[TestCategory(Cat)]
[Fact(Skip = "Need CoreTLC specific baseline update")]
public void CommandCrossValidation()
{ ... }
```

would become

``` csharp
[Trait("Category", Cat)]
[Fact(Skip = "Need CoreTLC specific baseline update")]
public void CommandCrossValidation()
{ ... }
```

Alternatively, we could use the [`XUnit.Categories` nuget](https://github.com/brendanconnolly/Xunit.Categories) and end up with something cleaner:

```
[Category(Cat)]
[Fact(Skip = "Need CoreTLC specific baseline update")]
public void CommandCrossValidation()
{ ... }
```

This nuget also contains some other useful attributes, like `Feature`, `Bug`, `Work Item` etc. What do you think, @Anipik, @eerhardt?

eerhardt:
> This nuget also contains some other useful attributes, like `Feature`, `Bug`, `Work Item` etc. What do you think, @Anipik, @eerhardt?

Let's keep using the `Microsoft.DotNet.XUnitExtensions` package. It still provides value.

eerhardt:
> I tried adding it to the project as well but it didn't work. any suggestions why ?

The `#if DEBUG` check is now happening in the `Microsoft.ML.TestFramework` project. Which means you have to look to see what defines are being used when building that project.

Looking at the `.csproj`, when we do things like this:

https://github.com/dotnet/machinelearning/blob/834e4715afa070f5e5c3628693682943b3f0bd6d/test/Microsoft.ML.TestFramework/Microsoft.ML.TestFramework.csproj#L3

That will completely override anything that is set earlier in a `.props` file. The correct line here is:

```xml
<DefineConstants>$(DefineConstants);CORECLR</DefineConstants>
```

But even better would be if we deleted this line from the `.csproj` all together and removed any dead code that is behind `#if !CORECLR` in this project.

mareklinka:
> But even better would be if we deleted this line from the `.csproj` all together and removed any dead code that is behind `#if !CORECLR` in this project.

Sounds good. I'll take a look at this over the weekend, see what falls out. To clarify, we are talking specifically the `Microsoft.ML.TestFramework` project here, right?

eerhardt:
> To clarify, we are talking specifically the `Microsoft.ML.TestFramework` project here, right?

Yes. The `DEBUG` define is not working in that project apparently. Just that project needs to be fixed now. I'm sure others have the same problems, but let's only do this one for now.

test/Microsoft.ML.TestFramework/DataPipe/TestDataPipeBase.cs
```diff
@@ -1096,23 +1096,6 @@ protected Func<bool> GetColumnComparer(Row r1, Row r2, int col, ColumnType type,
                     return GetComparerVec<RowId>(r1, r2, col, size, (x, y) => x.Equals(y));
             }
 
-#if !CORECLR // REVIEW: Port Picture type to CoreTLC.
-            if (type is PictureType)
```
mareklinka: I think this could have been ported from `PictureType` to `ImageType`, if I understand the problem correctly? Would require a reference to `ImageAnalytics`, though.

eerhardt: Currently we are only building the project with `CORECLR` defined. So this is dead code that can be deleted.

Let's just delete it for now. If @Ivanidzo4ka thinks this is important, we can open a new issue to add image support here. Thoughts @Ivanidzo4ka?

test/Microsoft.ML.TestFramework/Attributes/NotFullFrameworkFactAttribute.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.ML.TestFramework.Attributes
+{
+    /// <summary>
+    /// A fact for tests requiring full .NET Framework.
+    /// </summary>
+    public sealed class NotFullFrameworkFactAttribute : EnvironmentSpecificFactAttribute
```
eerhardt: Is this used?  I don't see any usages doing a simple search.

mareklinka:
> Just one comment on a class that doesn't appear to be used. Other than that, this looks good. Thanks for tackling this - I think it really cleans the tests up nicely.

Got rid of the unused attribute. Glad to be of help 😄

test/Microsoft.ML.TestFramework/DataPipe/TestDataPipeBase.cs
Ivanidzo4ka: Feel free to delete. Some legacy code which we don't care about anymore

Issue #2251, opened 2019-01-25
IPredictor and related parts hiding


TomFinley:
We have a few types inside `IPredictor.cs` that are perhaps useful for internal infrastructure, but are not directly useful from the point of view of the `IEstimator`/`ITransformer`/`MLContext` public API.

https://github.com/dotnet/machinelearning/blob/3e03fcef46e0bdd6961e0cccb7cc490ab535fbaf/src/Microsoft.ML.Core/Prediction/IPredictor.cs#L10

https://github.com/dotnet/machinelearning/blob/3e03fcef46e0bdd6961e0cccb7cc490ab535fbaf/src/Microsoft.ML.Core/Prediction/IPredictor.cs#L31

The thing that makes me the most nervous are things like `IPredictorProducing` and `IDistPredictorProducing`, which are definitely marker interfaaces that we don't want, but I would argue we don't really want any of this to be public.

https://github.com/dotnet/machinelearning/blob/3e03fcef46e0bdd6961e0cccb7cc490ab535fbaf/src/Microsoft.ML.Core/Prediction/IPredictor.cs#L43

https://github.com/dotnet/machinelearning/blob/3e03fcef46e0bdd6961e0cccb7cc490ab535fbaf/src/Microsoft.ML.Core/Prediction/IPredictor.cs#L67

There are a significant number of side effects from this, including the hiding of many other APIs that we want hidden anyway, since anything that is part of the `IEstimator`/`ITransformer`/`IDataView`/`MLConext` idiom does not take an `IPredictor` (it *produces* things that implement that interface, but generally we do not use that interface). So this will be a driver for hiding many other things.

PR #2404, opened 2019-02-04
Begin hiding the IPredictorProducing

### LINKED ISSUE #2378: State of CalibratorPredictorBase v1
There have been some issues concerning calibrator estimators (#1871 and #1622) but not calibrators themselves.

So, calibrated models are basically wrappers for model that have.

They are ultimately something akin to `CalibratedPredictorBase`. The trouble with `CalibratedPredictorBase` is this property:

So, consider this code.

https://github.com/dotnet/machinelearning/blob/578c188876cd7f158aca214d2e729500183b588c/docs/samples/Microsoft.ML.Samples/Dynamic/PermutationFeatureImportance/PFIRegressionExample.cs#L21-L29

Focus on the last part, where we're able to get the feature weights.

What is this `SubPredictor`? It is this:

https://github.com/dotnet/machinelearning/blob/578c188876cd7f158aca214d2e729500183b588c/src/Microsoft.ML.Data/Prediction/Calibrator.cs#L130

Great news: it has a definite type! Bad news: that is just a marker interface. As a mechanism for the API, it is as useless as if it were just, say, of type `object` (which, incidentally, I will have to do anyway as part of #2251). For that reason, we see lots of code that looks like this:

https://github.com/dotnet/machinelearning/blob/578c188876cd7f158aca214d2e729500183b588c/test/Microsoft.ML.Tests/TrainerEstimators/LbfgsTests.cs#L66

https://github.com/dotnet/machinelearning/blob/578c188876cd7f158aca214d2e729500183b588c/test/Microsoft.ML.Predictor.Tests/TestIniModels.cs#L584

This reason is, that object by itself is not useful: to get the actual model parameters, you have to do a "magical cast" to somehow get it into the right format. This sort of worked in command-line land or entry-point land, since everything was more or less dynamically typed anyway.

It might be desirable that, when training a logistic regression binary classification model, I should be able to, in a type safe fashion, be able to inspect the model weights without having to perform any "magical casts."

The most obvious solution to me is the following: calibrators becomes some sort of class which involves generics on both the "subpredictor" model parameters, as well as the calibrator. Then things like logistic regression would return instances of that generic class, or else some class derived from that generic class if we decide that must be an abstract class for some reason.7

The alternative is: we accept that "magical casts" are desirable, which I would not like since it is a little silly since I view the above "desirable" state is perfectly reasonable. But some people really hate generics.

I believe @yaeldekel had some thoughts on this, perhaps others do as well.

TomFinley:
Towards but not quite finishing #2251. Why not finish? At issue is that we must resolve #2378 in some fashion (TBD) before we continue, but as it is I think this is a move in the right direction. The public surface using this interface has now been reduced considerably, but of course not quite completely eliminated. So this is *not* a work in progress, but I didn't want to leave such a big change lying around.

First commit internalizes the (less used) `IDistPredictorProducing`, while the second does some but not all of the work of internalizing `IPredictorProducing`.


src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -55,19 +57,22 @@ public abstract class PredictionTransformerBase<TModel, TScorer> : IPredictionTr
 
         protected abstract TScorer Scorer { get; set; }
 
-        protected PredictionTransformerBase(IHost host, TModel model, Schema trainSchema)
+        [BestFriend]
```
Ivanidzo4ka: >[BestFriend] [](start = 8, length = 12)

Is it necessary?

src/Microsoft.ML.Ensemble/Selector/IDiversityMeasure.cs
```diff
@@ -11,33 +11,33 @@
...
-    public interface IDiversityMeasure<TOutput>
+    internal interface IDiversityMeasure<TOutput>
     {
-        List<ModelDiversityMetric<TOutput>> CalculateDiversityMeasure(IList<FeatureSubsetModel<IPredictorProducing<TOutput>>> models,
-            ConcurrentDictionary<FeatureSubsetModel<IPredictorProducing<TOutput>>, TOutput[]> predictions);
+        List<ModelDiversityMetric<TOutput>> CalculateDiversityMeasure(IList<FeatureSubsetModel<TOutput>> models,
+            ConcurrentDictionary<FeatureSubsetModel<TOutput>, TOutput[]> predictions);
     }
 
     public delegate void SignatureEnsembleDiversityMeasure();
```
Ivanidzo4ka: >public [](start = 4, length = 6)

maybe hide it as well?

Ivanidzo4ka:
>     using TScalarPredictor = IPredictorProducing<Single>;

you can remove this one

---
Refers to: src/Microsoft.ML.Ensemble/Trainer/EnsembleModelParameters.cs:22 in ba51e48. [](commit_id = ba51e48ee767b59a6bf3e9f48b88dad7dc55a729, deletion_comment = False)

src/Microsoft.ML.HalLearners/OlsLinearRegression.cs
```diff
@@ -35,7 +35,6 @@
 namespace Microsoft.ML.Trainers.HalLearners
 {
     /// <include file='doc.xml' path='doc/members/member[@name="OLS"]/*' />
-    [BestFriend]
```
Ivanidzo4ka: >BestFriend [](start = 5, length = 10)

Out of curiosity, can we modify analyzer to complain about usage of [BestFriend] on top of public class?
Unrelated to PR, just wondering.

test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamples.cs
```diff
@@ -662,7 +662,7 @@ private void MixMatch(string dataPath)
             var binaryTrainer = mlContext.BinaryClassification.Trainers.AveragedPerceptron("Label", "Features");
 
             // Append the OVA learner to the pipeline.
-            dynamicPipe = dynamicPipe.Append(new Ova(mlContext, binaryTrainer));
+            dynamicPipe = dynamicPipe.Append(mlContext.MulticlassClassification.Trainers.OneVersusAll(binaryTrainer));
```
Ivanidzo4ka: >mlContext [](start = 45, length = 9)

can you update [https://github.com/dotnet/machinelearning/blob/master/docs/code/MlNetCookBook.md](https://github.com/dotnet/machinelearning/blob/master/docs/code/MlNetCookBook.md) as well?

src/Microsoft.ML.Ensemble/Trainer/Multiclass/MulticlassDataPartitionEnsembleTrainer.cs
```diff
@@ -60,7 +61,18 @@ public Arguments()
...
+                            // present in this class. The options to the estimator have no way of being communicated to the `ITrainer`
+                            // implementation, so there is a fundamnetal disconnect if someone chooses to ever use the *estimator* with
+                            // non-default column names. Unfortuantely no method of resolving this temporary strikes me as being any
+                            // less laborious than the proper fix, which is that this "meta" component should itself be a trainer
+                            // estimator, as opposed to a regular trainer.
+                            var trainerEstimator = new MulticlassLogisticRegression(env, LabelColumn, FeatureColumn);
+                            return TrainerUtils.MapTrainerEstimatorToTrainer<MulticlassLogisticRegression,
+                                MulticlassLogisticRegressionModelParameters, MulticlassLogisticRegressionModelParameters>(env, trainerEstimator);
+                        })
+                        //env => new MulticlassLogisticRegression(env, LabelColumn, FeatureColumn))
```
wschin: >//env => new MulticlassLogisticRegression(env, LabelColumn, FeatureColumn)) [](start = 24, length = 75)

Is this line still needed?

src/Microsoft.ML.FastTree/BoostingFastTree.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.ML.Trainers.FastTree
     public abstract class BoostingFastTreeTrainerBase<TArgs, TTransformer, TModel> : FastTreeTrainerBase<TArgs, TTransformer, TModel>
         where TTransformer : ISingleFeaturePredictionTransformer<TModel>
         where TArgs : BoostedTreeArgs, new()
-        where TModel : IPredictorProducing<Float>
+        where TModel : class
     {
```
wschin: I guess I will have to replace this with a more restrictive new type in my PR.

src/Microsoft.ML.HalLearners/OlsLinearRegression.cs
TomFinley: We can indeed. Though as you suggest, extremely low priority. Entered anyway as #2415.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
TomFinley: Yes, there are a couple subclasses of this abstract class in other assemblies that call this attribute.

Indeed, as you work on #1995 for `IDataTransform` I'd expect you will need to lock down some of the abstract classes of dataview to make external subclassing impossible, which will require a similar approach, I expect?

test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamples.cs
TomFinley: Sure. I wish though that this was a bit less awkward to keep these up to date. Or that we could just link samples from the .md itself directly, somehow.

src/Microsoft.ML.FastTree/BoostingFastTree.cs
TomFinley: Since it is a marker interface, the type restriction is irrelevant. Also since all invocations must be internal to the code anyway, there is no particular need for any restriction.

Issue #2104, opened 2019-01-10
Not able to load Whitening Module in Maml on full framework


Anipik:
```
[xUnit.net 00:00:03.65]     Microsoft.ML.Tests.Transformers.NormalizerTests.TestWhiteningCommandLine [FAIL]
Failed   Microsoft.ML.Tests.Transformers.NormalizerTests.TestWhiteningCommandLine
Error Message:
 System.InvalidOperationException : Unknown loadable class: whitening
Stack Trace:
   at Microsoft.ML.Tools.Maml.MainCore(IHostEnvironment env, String args, Boolean alwaysPrintStacktrace) in C:\git\machinelearning\src\Microsoft.ML.Maml\MAML.cs:line 210
   at Microsoft.ML.Tools.Maml.MainWithProgress(String args) in C:\git\machinelearning\src\Microsoft.ML.Maml\MAML.cs:line 68
   at Microsoft.ML.Tools.Maml.MainAll(String args) in C:\git\machinelearning\src\Microsoft.ML.Maml\MAML.cs:line 45
   at Microsoft.ML.Tools.Maml.Main(String[] args) in C:\git\machinelearning\src\Microsoft.ML.Maml\MAML.cs:line 34
   at Microsoft.ML.Tests.Transformers.NormalizerTests.TestWhiteningCommandLine() in C:\git\machinelearning\test\Microsoft.ML.Tests\Transformers\NormalizerTests.cs:line 332
```

I am trying to run TestWhiteningCommandLine test on netfx. The test using Maml.Main and it is not able to load the whitening module.

Adding ``` Type t = typeof(VectorWhiteningTransformer);``` helps solve the issue but I think we may require more permanent solution

cc @danmosemsft @eerhardt @TomFinley

Ivanidzo4ka:
Does SymSgd, Ols learner tests works fine? 
`Type t = typeof(VectorWhiteningTransformer)`
this looks like indication what we just don't link HalLearner project properly to our tests.

Anipik:
> Does SymSgd, Ols learner tests works fine?

yes there is no similar failure.

PR #2409, opened 2019-02-04
Fix loading type problem for netfx

Anipik:
Fixes https://github.com/dotnet/machinelearning/issues/2104

A couple of types doesn't get loaded properly by Maml commands on netfx. So we have to register them manually.
We are doing a similar things for maml commands in our benchmarks.
These tests fail on netfx and are being enabled in https://github.com/dotnet/machinelearning/pull/2402

### DIFF
--- a/test/Microsoft.ML.Tests/TensorFlowEstimatorTests.cs
+++ b/test/Microsoft.ML.Tests/TensorFlowEstimatorTests.cs
@@ -135,9 +135,8 @@ namespace Microsoft.ML.Tests
         [ConditionalFact(typeof(Environment), nameof(Environment.Is64BitProcess))] // x86 output differs from Baseline
         void TestCommandLine()
         {
-            IHostEnvironment environment = new MLContext();
-            environment.ComponentCatalog.RegisterAssembly(typeof(TensorFlowTransformer).Assembly);
-
+            // typeof helps to load the TensorFlowTransformer type.
+            Type type = typeof(TensorFlowTransformer);
             Assert.Equal(Maml.Main(new[] { @"showschema loader=Text{col=a:R4:0-3 col=b:R4:0-3} xf=TFTransform{inputs=a inputs=b outputs=c modellocation={model_matmul/frozen_saved_model.pb}}" }), (int)0);
         }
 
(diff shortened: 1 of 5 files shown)


test/Microsoft.ML.Tests/TensorFlowEstimatorTests.cs
```diff
@@ -135,8 +135,7 @@ void TestOldSavingAndLoading()
         [ConditionalFact(typeof(Environment), nameof(Environment.Is64BitProcess))] // x86 output differs from Baseline
         void TestCommandLine()
         {
-            var env = new MLContext();
-            IHostEnvironment environment = env;
+            IHostEnvironment environment = new MLContext();
```
eerhardt: I'm not sure what this code is doing... we are creating an environment, but not doing anything with it.

Anipik: i am using the environment to register the assmebly that is being used by maml. The maml was not able to load this dependency by itself on netfx.

eerhardt: It doesn't make sense to `using the environment to register the assmebly` because you are not using that Environment in the Maml command.

If you just want to ensure the assembly is loaded, you can just reference the `typeof(blah)`. But I think we should be using that `environment` in the `Maml` command.

Anipik: I tried exposing and using Maml.MainCore function but then it resulted in adding an assembly Registrar https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Maml/MAML.cs#L61
(which is the only major diff between main and maincore)
so I ended up just adding the typeof (I was earlier worried about these statements being removed in the release mode. but its working)

Anipik:
@Ivanidzo4ka @sfilipi @justinormont can you please review this PR ?

test/Microsoft.ML.Tests/TensorFlowEstimatorTests.cs
```diff
@@ -135,7 +135,8 @@ void TestOldSavingAndLoading()
         [ConditionalFact(typeof(Environment), nameof(Environment.Is64BitProcess))] // x86 output differs from Baseline
         void TestCommandLine()
         {
-            var env = new MLContext();
+            // typeof helps to load the TensorFlowTransformer type.
+            Type type = typeof(TensorFlowTransformer);
```
justinormont: Any idea why `typeof` helps?

Anipik: So when we create an instance of a type in Maml, we check if the type\assembly is present in the loaded modules.
Typeof helps to load the assembly involving this type.

Issue #2373, opened 2019-02-01
DnnImageFeaturizingTests failing for netfx


Anipik:
These tests were not being tested due to the bug in conditionalFact attribute.
However when I tried to run these tests on my local machine they are failing.

The call stack is this
```C#
System.ArgumentOutOfRangeException : Specified argument was out of the range of valid values.
Parameter name: ModelFile
Stack Trace:
  C:\git\machinelearning\src\Microsoft.ML.OnnxTransform\OnnxTransform.cs(155,0): at Microsoft.ML.Transforms.OnnxTransformer..ctor(IHostEnvironment env, Arguments args, Byte[] modelBytes)
  C:\git\machinelearning\src\Microsoft.ML.DnnImageFeaturizer.ResNet18\ResNet18Extension.cs(43,0): at Microsoft.ML.Transforms.ResNet18Extension.ResNet18(DnnImageModelSelector dnnModelContext, IHostEnvironment env, String outputColumnName, String inputColumnName, String modelDir)
  C:\git\machinelearning\src\Microsoft.ML.DnnImageFeaturizer.ResNet18\ResNet18Extension.cs(25,0): at Microsoft.ML.Transforms.ResNet18Extension.ResNet18(DnnImageModelSelector dnnModelContext, IHostEnvironment env, String outputColumnName, String inputColumnName)
  C:\git\machinelearning\src\Microsoft.ML.OnnxTransform\DnnImageFeaturizerTransform.cs(66,0): at Microsoft.ML.Transforms.DnnImageFeaturizerEstimator..ctor(IHostEnvironment env, String outputColumnName, Func`2 modelFactory, String inputColumnName)
  C:\git\machinelearning\test\Microsoft.ML.OnnxTransformTest\DnnImageFeaturizerTest.cs(161,0): at Microsoft.ML.Tests.DnnImageFeaturizerTests.TestOldSavingAndLoading()
```

The tests failing due to this are 
TestOldSavingAndLoading
OnnxStatic
TestDnnImageFeaturizer

cc @jignparm @yaeldekel @danmosemsft @eerhardt

PR #2411, opened 2019-02-05
ResNet Tests corrected for netfx

Anipik:
Fixes https://github.com/dotnet/machinelearning/issues/2373

The behavior of GetExecutingAssembly is different on netFramework. It return a temp path on .netFramework

This problem also occurred while enabling tests for netfx.

### DIFF
--- a/src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AlexNetExtension.cs
+++ b/src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AlexNetExtension.cs
@@ -22,7 +22,9 @@ namespace Microsoft.ML.Transforms
         /// </summary>
         public static EstimatorChain<ColumnCopyingTransformer> AlexNet(this DnnImageModelSelector dnnModelContext, IHostEnvironment env, string outputColumnName, string inputColumnName)
         {
-            return AlexNet(dnnModelContext, env, outputColumnName, inputColumnName, Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "DnnImageModels"));
+            // Substring is  removing initial "file:///" from the codebase string.
+            string executingAssemblyLocation = Directory.GetParent(Assembly.GetExecutingAssembly().CodeBase.Substring(8)).FullName;
+            return AlexNet(dnnModelContext, env, outputColumnName, inputColumnName, Path.Combine(executingAssemblyLocation, "DnnImageModels"));
         }
 
         /// <summary>
--- a/src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AlexNetExtension.cs
+++ b/src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AlexNetExtension.cs
@@ -3,7 +3,6 @@
 // See the LICENSE file in the project root for more information.
 
 using System.IO;
-using System.Reflection;
 using Microsoft.ML.Data;
 
 namespace Microsoft.ML.Transforms
@@ -22,9 +21,7 @@ namespace Microsoft.ML.Transforms
         /// </summary>
         public static EstimatorChain<ColumnCopyingTransformer> AlexNet(this DnnImageModelSelector dnnModelContext, IHostEnvironment env, string outputColumnName, string inputColumnName)
         {
-            // Substring is  removing initial "file:///" from the codebase string.
-            string executingAssemblyLocation = Directory.GetParent(Assembly.GetExecutingAssembly().CodeBase.Substring(8)).FullName;
-            return AlexNet(dnnModelContext, env, outputColumnName, inputColumnName, Path.Combine(executingAssemblyLocation, "DnnImageModels"));
+            return AlexNet(dnnModelContext, env, outputColumnName, inputColumnName, Path.Combine(AssemblyPathHelpers.GetExecutingAssemblyLocation(), "DnnImageModels"));
         }
 
         /// <summary>
--- a/src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AssemblyPathHelpers.cs
+++ b/src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AssemblyPathHelpers.cs
@@ -1,0 +1,20 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.IO;
+using System.Reflection;
+
+namespace Microsoft.ML.Transforms
+{
+    internal static class AssemblyPathHelpers
+    {
+        public static string GetExecutingAssemblyLocation()
+        {
+            string codeBaseUri = Assembly.GetExecutingAssembly().CodeBase;
+            string path = new Uri(codeBaseUri).AbsolutePath;
+            return Directory.GetParent(path).FullName;
+        }
+    }
+}
--- a/src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AssemblyPathHelpers.cs
+++ b/src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AssemblyPathHelpers.cs
@@ -4,7 +4,6 @@
 
 using System;
 using System.IO;
-using System.Reflection;
 
 namespace Microsoft.ML.Transforms
 {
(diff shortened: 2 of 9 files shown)


src/Microsoft.ML.DnnImageFeaturizer.ResNet18/ResNet18Extension.cs
```diff
@@ -22,7 +22,14 @@ public static class ResNet18Extension
         /// </summary>
         public static EstimatorChain<ColumnCopyingTransformer> ResNet18(this DnnImageModelSelector dnnModelContext, IHostEnvironment env, string outputColumnName, string inputColumnName)
         {
-            return ResNet18(dnnModelContext, env, outputColumnName, inputColumnName, Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "DnnImageModels"));
+#if NETFRAMEWORK
```
eerhardt: I don't understand how this works/fixes anything. We aren't multi-targeting this assembly.

eerhardt: We should be able to do this using only `netstandard2.0` APIs. We shouldn't need to multi-target to make this work.

src/Microsoft.ML.DnnImageFeaturizer.ResNet18/ResNet18Extension.cs
```diff
@@ -22,7 +22,9 @@ public static class ResNet18Extension
         /// </summary>
         public static EstimatorChain<ColumnCopyingTransformer> ResNet18(this DnnImageModelSelector dnnModelContext, IHostEnvironment env, string outputColumnName, string inputColumnName)
         {
-            return ResNet18(dnnModelContext, env, outputColumnName, inputColumnName, Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "DnnImageModels"));
+            // Substring is  removing initial "file:///" from the codebase string.
+            string executingAssemblyLocation = Directory.GetParent(Assembly.GetExecutingAssembly().CodeBase.Substring(8)).FullName;
```
eerhardt: There are a few other places we are calling `GetExecutingAssembly`

https://github.com/dotnet/machinelearning/search?q=GetExecutingAssembly&unscoped_q=GetExecutingAssembly

Do those places need to be fixed as well? Especially the other `DnnImageFeaturizer` places.

Anipik: I corrected it for other ```DnnImageFeaturizer``` places.
We can change at other places when we add tests for them.

Anipik:
@Ivanidzo4ka @sfilipi @justinormont can you take a look at this PR ?

src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AlexNetExtension.cs
```diff
@@ -22,7 +22,9 @@ public static class AlexNetExtension
         /// </summary>
         public static EstimatorChain<ColumnCopyingTransformer> AlexNet(this DnnImageModelSelector dnnModelContext, IHostEnvironment env, string outputColumnName, string inputColumnName)
         {
-            return AlexNet(dnnModelContext, env, outputColumnName, inputColumnName, Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "DnnImageModels"));
+            // Substring is  removing initial "file:///" from the codebase string.
+            string executingAssemblyLocation = Directory.GetParent(Assembly.GetExecutingAssembly().CodeBase.Substring(8)).FullName;
```
eerhardt: It feels like we should have an internal helper method for this, since we are copying the code in 4 separate places.

Anipik: @eerhardt these 4 separate places are in 4 different projects. So in order to have this in a helper method , the options I can think of
- These all 4 projects have a dependency on Microsoft.ML.OnnxTransform.csproj. We can add a internal helper method in one of those and use internals visible to attribute.
- put the file in one of the projects and use compile item in csproj of other 3 projects. (But we are doing it anywhere in this repo)

which option should I opt for or do u have any other option ?

jignparm:
Can we enable the DNNImage tests in this PR to verify if the fix works in the the netfx leg? It doesn't look like the DNNImage tests are part of the netfx test run currently.

Anipik:
> Can we enable the DNNImage tests in this PR to verify if the fix works in the the netfx leg? It doesn't look like the DNNImage tests are part of the netfx test run currently.

There is an already open PR by a  contributor to remove all conditional attributes https://github.com/dotnet/machinelearning/pull/2402

I have tested them locally by just removing the conditional fact attribute, if there is any fall out, i will add the commit in #2402

src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AlexNetExtension.cs
justinormont: Is there a better method than `.Substring(8)`? I don't know .NET well enough, but most languages/frameworks have a helper function this.

eerhardt: We do the first option all over the place.

We are doing the 2nd option here:

https://github.com/dotnet/machinelearning/blob/d1a2962b85138c186d8b4cbbb77c4c458b3abc87/src/Microsoft.ML.CpuMath/Microsoft.ML.CpuMath.csproj#L13-L15

My preference would be for the 2nd option, as it allows for more control (you don't have this internal dependency between projects).

jignparm: // Substring is  removing initial "file:///" from the codebase string.
            string executingAssemblyLocation = 
    Directory.GetParent(Assembly.GetExecutingAssembly().CodeBase.Substring(8)).FullName;

Aside from consolidation into a helper function (good idea), the "file:///" handling seems like it may work for now, but not sure if it'll work in all future versions as well (windows server for example, produces slightly different paths than workstation). If an alternate API can be used instead of looking for the prefix of "file:///", it would seem more stable.

Anipik:
@eerhardt @jignparm @justinormont I addressed the required feedback

src/Microsoft.ML.DnnImageFeaturizer.ResNet101/Microsoft.ML.DnnImageFeaturizer.ResNet101.csproj
```diff
@@ -5,6 +5,11 @@
     <IncludeInPackage>Microsoft.ML.DnnImageFeaturizer.ResNet101</IncludeInPackage>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Include="..\Microsoft.ML.DnnImageFeaturizer.AlexNet\AssemblyPathHelpers.cs" />
+    <None Include="**/*.cs" />
```
eerhardt: Why?

src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AssemblyPathHelpers.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Reflection;
+
+namespace Microsoft.ML.Transforms
+{
+    internal static class AssemblyPathHelpers
+    {
+        public static string GetExecutingAssemblyLocation()
+        {
+            string codeBaseUri = Assembly.GetExecutingAssembly().CodeBase;
+            string path = new Uri(codeBaseUri).AbsolutePath;
```
eerhardt: Take a look at @JeremyKuhne's advice here about using `Uri`s for local file paths.

[https://github.com/dotnet/core-setup/issues/4969#issuecomment-456923839](https://github.com/dotnet/core-setup/issues/4969#issuecomment-456923839)

src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AssemblyPathHelpers.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.IO;
+using System.Reflection;
+
+namespace Microsoft.ML.Transforms
+{
+    internal static class AssemblyPathHelpers
+    {
+        public static string GetExecutingAssemblyLocation()
+        {
+            string codeBaseUri = Assembly.GetExecutingAssembly().CodeBase;
```
eerhardt: Instead of `Assembly.GetExecutingAssembly()`, would it be better to say

`typeof(AssemblyPathHelpers).Assembly`?

src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AssemblyPathHelpers.cs
Anipik: i think we should be okay with using uri here because here we are not constructing a uri against a path.
The codebase already returns a uri in form of a string. I am converting it back to uri to get the local path from the uri.

Issue #2410, opened 2019-02-04
ValueMapping: Write test for schema propagation for all types of inputs.


zeahmed:
There is a ValueMapping test that test for the schema propagation for `ValueMapperEstimator` and `ValueMapperTransformer`
https://github.com/dotnet/machinelearning/blob/2dfc51aa7172199c1b2145e48ecb9eb855bb87f1/test/Microsoft.ML.Tests/Transformers/ValueMappingTests.cs#L506

This test is not complete in a sense that it does not test all the possible inputs that can be provided to `ValueMapperEstimator`. It should test the ValueMapperEstimator on all possible data type the estimator can take as input and on all possible maps types.  

CC: @artidoro, @singlis

Ivanidzo4ka:
Is this one done or not? I can see two merged PRs. Can we close this one?

PR #2416, opened 2019-02-05
ValueMapper: added more tests with 'TestEstimatorCore' and fixed a bug in ValueType.

### LINKED ISSUE #2412: ValueMapper: Output Schema is not vector when Value type is vector.
This is the bug similar to #2385. 

When value type of the dictionary/map is a vector the output is vector. However, the ValueMappingEstimator GetOutputSchema fails at the following line
https://github.com/dotnet/machinelearning/blob/1706b3cfbf838ddd9c5e5a8eba80c4f1d2fd1aec/src/Microsoft.ML.Core/Data/IEstimator.cs#L67

CC: @singlis

zeahmed:
This PR fixes #2412 and #2410.

1. Added tests to check schema propagation in `ValueMapperEstimator` and `ValueMapperTransformer` with `TestEstimatorCore`.
2. Fixed a bug that was causing an exception when `ValueMapperEstimator.GetOutputSchema` is called; given that ValueType is a vector.
3. Properly returning output schema based on whether the ValueType  is fixed length or variable length vector.


zeahmed:
@singlis and @artidoro, I addressed your comments raised in #2386 in this PR.

src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
```diff
@@ -176,6 +176,33 @@ public static KeyToValueMappingEstimator MapKeyToValue(this TransformsCatalog.Co
             params (string outputColumnName, string inputColumnName)[] columns)
             => new ValueMappingEstimator<TInputType, TOutputType>(CatalogUtils.GetEnvironment(catalog), keys, values, columns);
 
+        /// <summary>
+        /// <see cref="ValueMappingEstimator"/>
+        /// </summary>
+        /// <typeparam name="TInputType">The key type.</typeparam>
+        /// <typeparam name="TOutputType">The value type.</typeparam>
+        /// <param name="catalog">The categorical transform's catalog</param>
```
artidoro: Can you use the comment from the estimator constructor, I believe that this is the type of the items here.

test/Microsoft.ML.Tests/Transformers/ValueMappingTests.cs
```diff
@@ -506,6 +514,60 @@ public void ValueMappingWorkout()
 
             // Workout on value mapping
             var est = ML.Transforms.Conversion.ValueMap(keys, values, new[] { ("D", "A"), ("E", "B"), ("F", "C") });
+            var schema = est.GetOutputSchema(SchemaShape.Create(dataView.Schema));
+            foreach (var name in new[] { "D", "E", "F" })
+            {
+                Assert.True(schema.TryFindColumn(name, out var originalColumn));
+                Assert.Equal(SchemaShape.Column.VectorKind.Scalar, originalColumn.Kind);
+                Assert.Equal(NumberType.I4, originalColumn.ItemType);
+            }
```
singlis: I dont think you need to retest this here. TestEstimatorCore will run call GetOutputSchema multiple times.

test/Microsoft.ML.Tests/Transformers/ValueMappingTests.cs
```diff
@@ -506,6 +514,60 @@ public void ValueMappingWorkout()
...
+
+            // Workout on value mapping
+            var est = ML.Transforms.Conversion.ValueMap(keys, values, new[] { ("D", "A"), ("E", "B"), ("F", "C") });
+            var schema = est.GetOutputSchema(SchemaShape.Create(dataView.Schema));
+            foreach (var name in new[] { "D", "E", "F" })
+            {
+                Assert.True(schema.TryFindColumn(name, out var originalColumn));
+                Assert.Equal(SchemaShape.Column.VectorKind.VariableVector, originalColumn.Kind);
+                Assert.Equal(NumberType.I4, originalColumn.ItemType);
+            }
```
singlis: same here

PR #2417, opened 2019-02-05
Textloader internalizations, documentation, and Arguments refactoring

### LINKED ISSUE #2046: TextLoader Arguments/ArgumentsCore classes
Consider this class, the settings object for `TextLoader`:

https://github.com/dotnet/machinelearning/blob/faffd179c961f120ec4e6babb06bbfb2cca6a6ea/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L382

This descends from this class.

https://github.com/dotnet/machinelearning/blob/faffd179c961f120ec4e6babb06bbfb2cca6a6ea/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L337

Why establish this class relationship? Well, because we want to distinguish between arguments that are "core" vs. not, and so that should be retained when we save the "header" of the text file, vs. those that might vary from iteration to iteration.

This class is used in two places for this purpose, in two places exactly.

https://github.com/dotnet/machinelearning/blob/faffd179c961f120ec4e6babb06bbfb2cca6a6ea/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L1193-L1195

https://github.com/dotnet/machinelearning/blob/faffd179c961f120ec4e6babb06bbfb2cca6a6ea/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L1230-L1236

Back when these classes were written and meant to support a command line and GUI tool only, it was acceptable to use class relationships for this purpsoe -- we did not expose this class to the users via an API. Now that we do expose it through an API, this little "trick" is no longer acceptable and causes confusion. There are only three "special" non-core arguments to account for, surely we can handle their presence through some mechanism other than this odd pollution of our type hierarchy (which is visible to users necessarily), and instead just handle it in the code for the saving/loading of the header itself. (That is, the load/save code could just account for the three arguments directly, instead of working in this strange way through the command line processor.)

The end result of this should be there should be only one class, `Arguments`, containing everything that is now in these two classes. It is also essential that the arguments presently occurring *only* in the `Arguments` class at present be excluded from the header and header parsing code.

There are several ways we could imagine doing this.

1. The most obvious is to just special case this code in the `TextLoader` code itself.

2. Another possibility is we add another attribute to the command line processing code itself, to capture those arguments that are meant to capture purely runtime and not behavioral considerations. Indeed, this happens in other contexts: those components that benefit from GPU acceleration might, naturally, have the GPU device ID as a configuration parameter, but if we were to ask this component to describe its configuration, behaviorally we might want it to *exclude* that configuration, since that is not portable from one computer or platform to another. (Which is the purpose of the current arrangement.) Rather than special casing this, as suggested above, we could have another (internal!!) attribute to flag such arguments as these.

I give two options because O am not too particular as to how. I might favor 1 until we gain more experience in scenario 2 so as to justify a more general solution. Though, perhaps we have already reached that point, since I know scenario 2 that I have described has already come up, though in situations less central and important than the text loader.

/cc @stephentoub

artidoro:
Fixes #2046.
Related to #1798.

This PR does the following:

1. Addresses the problem raised by #2046 by creating a single `Arguments` class from the original two classes `Arguments` and `ArgumentsCore`. The issue suggested the first approach out of two possible ways to do this, which is the one I followed in this PR.
2. Renames `Arguments` to `Options`.
3. Internalizes the constructors and other fields that should not be in the public API.
4. Adds comments for all public fields and methods.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/DataReaderExtensions.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/DataReaderExtensions.cs
@@ -9,14 +9,12 @@ namespace Microsoft.ML
 {
     public static class DataReaderExtensions
     {
-        public static IDataView Read(this IDataReader<IMultiStreamSource> reader, string path)
-        {
-            return reader.Read(new MultiFileSource(path));
-        }
-
+        /// <summary>
+        /// Reads data from one or more file <paramref name="path"/> into an <see cref="IDataView"/>.
+        /// </summary>
+        /// <param name="reader">The reader to use.</param>
+        /// <param name="path">One or more paths from which to load data.</param>
         public static IDataView Read(this IDataReader<IMultiStreamSource> reader, params string[] path)
-        {
-            return reader.Read(new MultiFileSource(path));
-        }
+            => reader.Read(new MultiFileSource(path));
     }
 }
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1413,6 +1413,10 @@ namespace Microsoft.ML.Data
 
         public DataViewSchema GetOutputSchema() => _bindings.OutputSchema;
 
+        /// <summary>
+        /// Reads data from <paramref name="source"/> into an <see cref="IDataView"/>.
+        /// </summary>
+        /// <param name="source">The source from which to load data.</param>
         public IDataView Read(IMultiStreamSource source) => new BoundLoader(this, source);
 
         internal static TextLoader CreateTextReader<TInput>(IHostEnvironment host,
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -31,18 +31,16 @@ namespace Microsoft.ML.Data
         /// <summary>
         /// Describes how an input column should be mapped to an <see cref="IDataView"/> column.
         /// </summary>
-        /// <example>
-        /// Scalar column of <seealso cref="DataKind"/> I4 sourced from 2nd column
-        ///      col=ColumnName:I4:1
-        ///
-        /// Vector column of <seealso cref="DataKind"/> I4 that contains values from columns 1, 3 to 10
-        ///     col=ColumnName:I4:1,3-10
-        ///
-        /// Key range column of KeyType with underlying storage type U4 that contains values from columns 1, 3 to 10, that can go from 1 to 100 (0 reserved for out of range)
-        ///     col=ColumnName:U4[100]:1,3-10
-        /// </example>
         public sealed class Column
         {
+            // Examples of how a column is defined in command line API:
+            // Scalar column of <seealso cref="DataKind"/> I4 sourced from 2nd column
+            //      col=ColumnName:I4:1
+            // Vector column of <seealso cref="DataKind"/> I4 that contains values from columns 1, 3 to 10
+            //     col=ColumnName:I4:1,3-10
+            // Key range column of KeyType with underlying storage type U4 that contains values from columns 1, 3 to 10, that can go from 1 to 100 (0 reserved for out of range)
+            //     col=ColumnName:U4[100]:1,3-10
+
             /// <summary>
             /// Describes how an input column should be mapped to an <see cref="IDataView"/> column.
             /// </summary>
@@ -52,7 +50,7 @@ namespace Microsoft.ML.Data
             /// Describes how an input column should be mapped to an <see cref="IDataView"/> column.
             /// </summary>
             /// <param name="name">Name of the column.</param>
-            /// <param name="type">Type of the items in the column.</param>
+            /// <param name="type"><see cref="DataKind"/> of the items in the column. If <see langword="null"/> defaults to a float.</param>
             /// <param name="index">Index of the column.</param>
             public Column(string name, DataKind? type, int index)
                : this(name, type, new[] { new Range(index) }) { }
@@ -61,7 +59,7 @@ namespace Microsoft.ML.Data
             /// Describes how an input column should be mapped to an <see cref="IDataView"/> column.
             /// </summary>
             /// <param name="name">Name of the column.</param>
-            /// <param name="type">Type of the items in the column.</param>
+            /// <param name="type"><see cref="DataKind"/> of the items in the column. If <see langword="null"/> defaults to a float.</param>
             /// <param name="minIndex">The minimum inclusive index of the column.</param>
             /// <param name="maxIndex">The maximum-inclusive index of the column.</param>
             public Column(string name, DataKind? type, int minIndex, int maxIndex)
@@ -73,7 +71,7 @@ namespace Microsoft.ML.Data
             /// Describes how an input column should be mapped to an <see cref="IDataView"/> column.
             /// </summary>
             /// <param name="name">Name of the column.</param>
-            /// <param name="type">Type of the items in the column.</param>
+            /// <param name="type"><see cref="DataKind"/> of the items in the column. If <see langword="null"/> defaults to a float.</param>
             /// <param name="source">Source index range(s) of the column.</param>
             /// <param name="keyCount">For a key column, this defines the range of values.</param>
             public Column(string name, DataKind? type, Range[] source, KeyCount keyCount = null)
@@ -94,7 +92,7 @@ namespace Microsoft.ML.Data
             public string Name;
 
             /// <summary>
-            /// Type of the items in the column.
+            /// <see cref="DataKind"/> of the items in the column. If <see langword="null"/> defaults to a float.
             /// </summary>
             [Argument(ArgumentType.AtMostOnce, HelpText = "Type of the items in the column")]
             public DataKind? Type;
@@ -275,13 +273,13 @@ namespace Microsoft.ML.Data
             }
 
             /// <summary>
-            /// The minimum inclusive index of the column.
+            ///  The minimum index of the column, inclusive.
             /// </summary>
             [Argument(ArgumentType.Required, HelpText = "First index in the range")]
             public int Min;
 
             /// <summary>
-            /// The maximum-inclusive index of the column. If <see langword="null"/>
+            /// The maximum index of the column, inclusive. If <see langword="null"/>
             /// indicates that the <see cref="TextLoader"/> should auto-detect the legnth
             /// of the lines, and read untill the end.
             /// If max is specified, the fields <see cref="AutoEnd"/> and <see cref="VariableEnd"/> are ignored.
@@ -1301,6 +1299,21 @@ namespace Microsoft.ML.Data
             LDone:
                 return !error;
             }
+        }
+
+        /// <summary>
+        /// Checks whether the source contains the valid TextLoader.Arguments depiction.
+        /// </summary>
+        internal static bool FileContainsValidSchema(IHostEnvironment env, IMultiStreamSource files, out Options options)
+        {
+            Contracts.CheckValue(env, nameof(env));
+            var h = env.Register(RegistrationName);
+            h.CheckValue(files, nameof(files));
+            options = new Options();
+            Column[] cols;
+            bool error;
+            bool found = TryParseSchema(h, files, ref options, out cols, out error);
+            return found && !error && options.IsValid();
         }
 
         /// <summary>
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1223,9 +1223,17 @@ namespace Microsoft.ML.Data
 #pragma warning restore 649 // never assigned
         }
 
-        // See if we can extract valid arguments from the first data file.
-        // If so, update options and set cols to the combined set of columns.
-        // If not, set error to true if there was an error condition.
+        /// <summary>
+        /// See if we can extract valid arguments from the first data file. If so, update options and set cols to the combined set of columns.
+        /// If not, set error to true if there was an error condition.
+        /// </summary>
+        /// <remarks>
+        /// Not all arguments are extracted from the data file. There are three arguments that can vary from iteration to iteration and that are set
+        /// directly by the user in the options class. These three arguments are:
+        /// <see cref="Options.UseThreads"/>,
+        /// <see cref="Options.HeaderFile"/>,
+        /// <see cref="Options.MaxRows"/>
+        /// </remarks>
         private static bool TryParseSchema(IHost host, IMultiStreamSource files,
             ref Options options, out Column[] cols, out bool error)
         {
@@ -1237,10 +1245,11 @@ namespace Microsoft.ML.Data
             error = false;
 
             // Verify that the current schema-defining arguments are default.
-            // Get settings just for core arguments, not everything.
+            // Get a string representation of the settings for all the fields of the Options class besides the following three
+            // UseThreads, HeaderFile, MaxRows, which are set by the user directly.
             string tmp = CmdParser.GetSettings(host, options, new Options()
             {
-                // It's fine if the user sets the following three arguments that are not considered core arguments.
+                // It's fine if the user sets the following three arguments, as they are instance specific.
                 // Setting the defaults to the user provided values will avoid these in the output of the call CmdParser.GetSettings.
                 UseThreads = options.UseThreads,
                 HeaderFile = options.HeaderFile,
@@ -1281,11 +1290,11 @@ namespace Microsoft.ML.Data
                     goto LDone;
 
                 var optionsNew = new Options();
-                // Copy the core arguments to the new options.
+                // Set the fields of optionsNew to the arguments parsed from the file.
                 if (!CmdParser.ParseArguments(host, loader.GetSettingsString(), optionsNew, typeof(Options), msg => ch.Error(msg)))
                     goto LDone;
 
-                // Set the non core arguments in the new options.
+                // Overwrite the three arguments that vary from iteration to iteration with the values specified by the user in the options class.
                 optionsNew.UseThreads = options.UseThreads;
                 optionsNew.HeaderFile = options.HeaderFile;
                 optionsNew.MaxRows = options.MaxRows;
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1237,18 +1237,16 @@ namespace Microsoft.ML.Data
             cols = null;
             error = false;
 
-            // It's fine if the user sets the following three arguments that are not considered core arguments.
-            // Setting the defaults to the user provided values will avoid these in the output of the call CmdParser.GetSettings.
-            var optionsCoreDefault = new Options()
+            // Verify that the current schema-defining arguments are default.
+            // Get settings just for core arguments, not everything.
+            string tmp = CmdParser.GetSettings(host, options, new Options()
             {
+                // It's fine if the user sets the following three arguments that are not considered core arguments.
+                // Setting the defaults to the user provided values will avoid these in the output of the call CmdParser.GetSettings.
                 UseThreads = options.UseThreads,
                 HeaderFile = options.HeaderFile,
                 MaxRows = options.MaxRows
-            };
-
-            // Verify that the current schema-defining arguments are default.
-            // Get settings just for core arguments, not everything.
-            string tmp = CmdParser.GetSettings(host, options, optionsCoreDefault);
+            });
 
             // Try to get the schema information from the file.
             string str = Cursor.GetEmbeddedArgs(files);
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1237,9 +1237,18 @@ namespace Microsoft.ML.Data
             cols = null;
             error = false;
 
+            // It's fine if the user sets the following three arguments that are not considered core arguments.
+            // Setting the defaults to the user provided values will avoid these in the output of the call CmdParser.GetSettings.
+            var optionsCoreDefault = new Options()
+            {
+                UseThreads = options.UseThreads,
+                HeaderFile = options.HeaderFile,
+                MaxRows = options.MaxRows
+            };
+
             // Verify that the current schema-defining arguments are default.
             // Get settings just for core arguments, not everything.
-            string tmp = CmdParser.GetSettings(host, options, new Options());
+            string tmp = CmdParser.GetSettings(host, options, optionsCoreDefault);
 
             // Try to get the schema information from the file.
             string str = Cursor.GetEmbeddedArgs(files);
@@ -1249,7 +1258,7 @@ namespace Microsoft.ML.Data
             // Parse the extracted information.
             using (var ch = host.Start("Parsing options from file"))
             {
-                // If tmp is not empty, this means the user specified some additional arguments in the command line,
+                // If tmp is not empty, this means the user specified some additional arguments in the options or command line,
                 // such as quote- or sparse-. Warn them about it, since this means that the columns will not be read from the file.
                 if (!string.IsNullOrWhiteSpace(tmp))
                 {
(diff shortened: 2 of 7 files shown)


src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -41,16 +44,39 @@ public sealed partial class TextLoader : IDataReader<IMultiStreamSource>, ICanSa
...
+            /// <summary>
+            /// Describes how an input column should be mapped to an <see cref="IDataView"/> column.
+            /// </summary>
             public Column() { }
 
+            /// <summary>
+            /// Describes how an input column should be mapped to an <see cref="IDataView"/> column.
+            /// </summary>
+            /// <param name="name">Name of the column.</param>
+            /// <param name="type">Type of the items in the column.</param>
```
sfilipi: >Type [](start = 35, length = 4)

<see cref="DataKind"

sfilipi:
>         ///     col=ColumnName:U4[100]:1,3-10

those are cmd examples. Maybe move them in regular comments above the column, but not XML comments. 

---
Refers to: src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs:43 in f75b504. [](commit_id = f75b504fe60cdbc85475011b64c8194f9e922dee, deletion_comment = False)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -62,15 +88,27 @@ public Column(string name, DataKind? type, Range[] source, KeyCount keyCount = n
...
             }
 
+            /// <summary>
+            /// Name of the column.
+            /// </summary>
             [Argument(ArgumentType.AtMostOnce, HelpText = "Name of the column")]
             public string Name;
 
+            /// <summary>
+            /// Type of the items in the column.
```
sfilipi: >Type [](start = 16, length = 4)

<see cref="DataKind"

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -199,6 +237,9 @@ internal bool IsValid()
             }
         }
 
+        /// <summary>
+        /// Specifies the range of indices of input columns that should be mapped to an output column.
+        /// </summary>
         public sealed class Range
         {
             public Range() { }
```
sfilipi: > public Range() { } [](start = 11, length = 19)

is this used?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -234,27 +275,49 @@ public Range(int min, int? max)
                 AutoEnd = max == null;
             }
 
+            /// <summary>
+            /// The minimum inclusive index of the column.
```
sfilipi: >The minimum inclusive index of the column. [](start = 16, length = 42)

would phrasing it as follows sound better?

The minimum index of the column, inclusive.

test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -465,7 +465,7 @@ public void EntryPointCreateEnsemble()
                      MetadataUtils.Const.ScoreValueKind.Score)
                     ).Transform(individualScores[i]);
 
-                individualScores[i] = ColumnSelectingTransformer.CreateDrop(Env, individualScores[i], MetadataUtils.Const.ScoreValueKind.Score);
+                individualScores[i] = new ColumnSelectingTransformer(Env, null, new[] { MetadataUtils.Const.ScoreValueKind.Score }).Transform(individualScores[i]);
```
sfilipi: >null [](start = 74, length = 4)

this looks funny, but i guess the ColumnSelectingTransformer is internal.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1308,16 +1397,24 @@ public void Save(ModelSaveContext ctx)
             _bindings.Save(ctx);
         }
 
+        /// <summary>
+        /// Returns the <see cref="SchemaShape"/> of the schema which will be produced by the transformer.
```
sfilipi: >transformer [](start = 94, length = 11)

<see cref="TextLoader"

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1177,44 +1279,31 @@ private sealed class LoaderHolder
...
-            Contracts.CheckValue(env, nameof(env));
-            var h = env.Register(RegistrationName);
-            h.CheckValue(files, nameof(files));
-            args = new Arguments();
-            Column[] cols;
-            bool error;
-            bool found = TryParseSchema(h, files, ref args, out cols, out error);
-            return found && !error && args.IsValid();
-        }
-
```
sfilipi: is this deletion intentional?

artidoro: Well I noticed we never call it, so I decided to remove it, but maybe would be better to make it internal?

---
In reply to: [254402153](https://github.com/dotnet/machinelearning/pull/2417#discussion_r254402153) [](ancestors = 254402153)

artidoro: Since this might be used in TLC, I am just going to make it internal and keep it.

---
In reply to: [254512295](https://github.com/dotnet/machinelearning/pull/2417#discussion_r254512295) [](ancestors = 254512295,254402153)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
artidoro: It is! And I think you are right it looks a bit strange. After discussing with Tom, I am going to open an issue about reshaping the public surface of TextLoader and fix it separately. Let's keep this PR only for internalization and refactoring of Arguments.

---
In reply to: [254397852](https://github.com/dotnet/machinelearning/pull/2417#discussion_r254397852) [](ancestors = 254397852)

src/Microsoft.ML.Data/DataLoadSave/DataReaderExtensions.cs
```diff
@@ -10,14 +10,12 @@ namespace Microsoft.ML
 {
     public static class DataReaderExtensions
     {
-        public static IDataView Read(this IDataReader<IMultiStreamSource> reader, string path)
```
Ivanidzo4ka: >string path [](start = 82, length = 11)

[https://stackoverflow.com/questions/38679180/will-params-in-c-sharp-always-cause-a-new-array-to-be-allocated-on-every-call](https://stackoverflow.com/questions/38679180/will-params-in-c-sharp-always-cause-a-new-array-to-be-allocated-on-every-call)
Can you please remain this method?

src/Microsoft.ML.StaticPipe/LocalPathReader.cs
```diff
@@ -2,20 +2,20 @@
...
 // See the LICENSE file in the project root for more information.
 
+using System.Data;
 using Microsoft.ML.Data;
 
 namespace Microsoft.ML.StaticPipe
 {
     public static class LocalPathReader
     {
-        public static DataView<TShape> Read<TShape>(this DataReader<IMultiStreamSource, TShape> reader, string path)
```
Ivanidzo4ka: >path [](start = 111, length = 4)

see my comment on DataReaderExtensions.cs

Ivanidzo4ka:
>[@artidoro](/artidoro) [artidoro](/artidoro) [force-pushed](/dotnet/machinelearning/compare/7ff9724e67fed8ca46bca67f663dd17d48a6cc63..e3b17b9a652d33341b0b2a87705642e8d7cf7c49) the artidoro:textloader  branch from [`7ff9724`](/dotnet/machinelearning/commit/7ff9724e67fed8ca46bca67f663dd17d48a6cc63) to [`e3b17b9`](/dotnet/machinelearning/commit/e3b17b9a652d33341b0b2a87705642e8d7cf7c49) [2 days ago](#event-2123110343)

I'm glad what you awere of git fu, but rebase and force push screw up codeflow, it get;s hard to track comments. Can you just merge and push?

On other note, do you plan to update this PR?

justinormont:
Related issue: https://github.com/dotnet/machinelearning/issues/2472 "Any reason for inconsistency in params for TextLoader?"

src/Microsoft.ML.Data/DataLoadSave/DataReaderExtensions.cs
artidoro: Ok I understand your point, was not aware of that! That's a good thing to know :)

---
In reply to: [254525235](https://github.com/dotnet/machinelearning/pull/2417#discussion_r254525235) [](ancestors = 254525235)

artidoro: I have the impression that in many places in our code we simply have one overload with `params`.

---
In reply to: [255284570](https://github.com/dotnet/machinelearning/pull/2417#discussion_r255284570) [](ancestors = 255284570,254525235)

artidoro:
I am still trying to learn from master Tom about git magic, but I understand the problem with rebase once the PR is out. 

---
In reply to: [461882216](https://github.com/dotnet/machinelearning/pull/2417#issuecomment-461882216) [](ancestors = 461882216)

artidoro:
Absolutely, I will address that and further concerns regarding TextLoader in another PR. This one is just to reduce the current public surface to what is essential. In that issue, and in a future PR I think we should rethink what we expose and possibly simplify it.

---
In reply to: [461952978](https://github.com/dotnet/machinelearning/pull/2417#issuecomment-461952978) [](ancestors = 461952978)

artidoro:
I have resolved Ivan's comments, so if @Ivanidzo4ka and @sfilipi  could take another look it would be great.

src/Microsoft.ML.Data/DataLoadSave/DataReaderExtensions.cs
Ivanidzo4ka: Are you sure? If we talk about transforms we usually have simple case with input/output pair which is usually enough

---
In reply to: [255284582](https://github.com/dotnet/machinelearning/pull/2417#discussion_r255284582) [](ancestors = 255284582,255284570,254525235)

src/Microsoft.ML.Data/DataLoadSave/DataReaderExtensions.cs
```diff
@@ -10,14 +10,20 @@ namespace Microsoft.ML
 {
     public static class DataReaderExtensions
     {
+        /// <summary>
+        /// Reads data from a file <paramref name="path"/> into an <see cref="IDataView"/>.
+        /// </summary>
+        /// <param name="reader">The reader to use.</param>
+        /// <param name="path">Ppaths from which to load data.</param>
```
Ivanidzo4ka: >Ppaths [](start = 31, length = 6)

Just path

src/Microsoft.ML.Data/DataLoadSave/DataReaderExtensions.cs
```diff
@@ -10,14 +10,20 @@ namespace Microsoft.ML
...
-        {
-            return reader.Read(new MultiFileSource(path));
-        }
+            =>  reader.Read(new MultiFileSource(path));
 
+        /// <summary>
+        /// Reads data from one or more file <paramref name="path"/> into an <see cref="IDataView"/>.
+        /// </summary>
+        /// <param name="reader">The reader to use.</param>
+        /// <param name="path">One or more paths from which to load data.</param>
```
Ivanidzo4ka: >One or more paths [](start = 31, length = 17)

Paths, case with one path would be caught by method above

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -25,32 +25,56 @@
...
+            /// <summary>
+            /// Describes how an input column should be mapped to an <see cref="IDataView"/> column.
+            /// </summary>
             public Column() { }
 
+            /// <summary>
+            /// Describes how an input column should be mapped to an <see cref="IDataView"/> column.
+            /// </summary>
+            /// <param name="name">Name of the column.</param>
+            /// <param name="type"><see cref="DataKind"/> of the items in the column. If <see langword="null"/> defaults to a float.</param>
```
Ivanidzo4ka: >float [](start = 126, length = 5)

well, R4, it's datakind, not type.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -220,7 +259,7 @@ public Range(int index)
             /// <param name="min">The minimum inclusive index of the column.</param>
             /// <param name="max">The maximum-inclusive index of the column. If <c>null</c>
             /// indicates that the <see cref="TextLoader"/> should auto-detect the legnth
-            /// of the lines, and read till the end.</param>
+            /// of the lines, and read untill the end.</param>
```
singlis: >untill [](start = 39, length = 6)

until

test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
singlis: I preferred the CreateDrop/CreateKeep and make this constructor private to avoid confusion. 

---
In reply to: [254400947](https://github.com/dotnet/machinelearning/pull/2417#discussion_r254400947) [](ancestors = 254400947)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -25,32 +25,56 @@
...
+            /// <summary>
+            /// Describes how an input column should be mapped to an <see cref="IDataView"/> column.
+            /// </summary>
             public Column() { }
 
+            /// <summary>
+            /// Describes how an input column should be mapped to an <see cref="IDataView"/> column.
+            /// </summary>
+            /// <param name="name">Name of the column.</param>
+            /// <param name="type"><see cref="DataKind"/> of the items in the column. If <see langword="null"/> defaults to a <see cref="DataKind.R4"/>.</param>
```
Ivanidzo4ka: >If <see langword="null"/> [](start = 86, length = 25)

Just a rant.
Only reason why we have it is CommandLine in which you can say something like `col=ColA:0" and expect it to work.
On programmatic level idea of "Hey its a nullable field but if you set it null we will treat it as float" is weird.
If you want to change it, one tricky place is TryParse method line 123, just set Type straight to DataKind.Num since in other cases it can be never setup.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -334,66 +395,98 @@ internal bool TryUnparse(StringBuilder sb)
             }
         }
 
-        public class ArgumentsCore
+        /// <summary>
+        /// The settings for <see cref="TextLoader"/>
```
Ivanidzo4ka: >settings [](start = 16, length = 8)

options.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -334,66 +395,98 @@ internal bool TryUnparse(StringBuilder sb)
...
+        /// </summary>
+        public class Options
         {
+            /// <summary>
+            /// Whether the input may include quoted values, which can contain separator characters, colons,
+            /// and distinguish empty values from missing values. When true, consecutive separators denote a
+            /// missing value and an empty value is denoted by \"\". When false, consecutive separators denote an empty value.
```
Ivanidzo4ka: >\"\" [](start = 63, length = 4)

where should be specific documentation magic to properly represent such strings. `<value></value> maybe?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -334,66 +395,98 @@ internal bool TryUnparse(StringBuilder sb)
...
                     "Whether the input may include quoted values, which can contain separator characters, colons," +
                     " and distinguish empty values from missing values. When true, consecutive separators denote a" +
                     " missing value and an empty value is denoted by \"\". When false, consecutive separators" +
                     " denote an empty value.",
                 ShortName = "quote")]
-            public bool AllowQuoting = DefaultArguments.AllowQuoting;
+            public bool AllowQuoting = Defaults.AllowQuoting;
 
+            /// <summary>
+            /// Whether the input may include sparse representations.
```
Ivanidzo4ka: >Whether the input may include sparse representations. [](start = 16, length = 53)

[citation needed] ([https://xkcd.com/285/](https://xkcd.com/285/))
not sure how it should look like, but we should specify how sparse format looks like. it's a index:value separated pairs where each index should be bigger than previous one.

Ivanidzo4ka: I also not sure can we have multiple sparse columns or not.

---
In reply to: [255806745](https://github.com/dotnet/machinelearning/pull/2417#discussion_r255806745) [](ancestors = 255806745)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -334,66 +395,98 @@ internal bool TryUnparse(StringBuilder sb)
...
+            /// </summary>
             [Argument(ArgumentType.AtMostOnce, HelpText = "Whether the input may include sparse representations", ShortName = "sparse")]
-            public bool AllowSparse = DefaultArguments.AllowSparse;
+            public bool AllowSparse = Defaults.AllowSparse;
 
+            /// <summary>
+            /// Number of source columns in the text data. Default is that sparse rows contain their size information.
```
Ivanidzo4ka: >Number [](start = 16, length = 6)

would be  nice to have example
theoretically we can have sparse format like SparseSize separtor pairs of index:value devided by separator
743 1:2 5:-1 8:0.4 742:1 743:2
not sure what would happen if we have 744:7 in that line.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -334,66 +395,98 @@ internal bool TryUnparse(StringBuilder sb)
...
                 ShortName = "size")]
             public int? InputSize;
 
             [Argument(ArgumentType.AtMostOnce, Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly, HelpText = "Source column separator. Options: tab, space, comma, single character", ShortName = "sep")]
             // this is internal as it only serves the command line interface
-            internal string Separator = DefaultArguments.Separator.ToString();
+            internal string Separator = Defaults.Separator.ToString();
 
+            /// <summary>
+            /// The characters that should be used as separators column separator.
```
Ivanidzo4ka: > separators column separator [](start = 53, length = 28)

does it sound right for you?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1145,7 +1238,14 @@ private sealed class LoaderHolder
 
             // Verify that the current schema-defining arguments are default.
             // Get settings just for core arguments, not everything.
-            string tmp = CmdParser.GetSettings(host, args, new ArgumentsCore());
+            string tmp = CmdParser.GetSettings(host, options, new Options()
+            {
+                // It's fine if the user sets the following three arguments that are not considered core arguments.
+                // Setting the defaults to the user provided values will avoid these in the output of the call CmdParser.GetSettings.
```
Ivanidzo4ka: Why we override them here? Can you override them in upstream function?

artidoro: `GetSettings` builds a string with the fields in the second parameter `options` that are different from the fields in the third parameter in this case `new Options () {...}`. Previously, only the core arguments were considered when taking the difference between the user provided arguments and the defaults. Since now there is no distinction between core and non core arguments, I achieve exactly the same behavior as before by setting the non core arguments in the third parameter to the user provided values. 

---
In reply to: [255808097](https://github.com/dotnet/machinelearning/pull/2417#discussion_r255808097) [](ancestors = 255808097)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
artidoro: No, thank you for pointing out.

---
In reply to: [255807189](https://github.com/dotnet/machinelearning/pull/2417#discussion_r255807189) [](ancestors = 255807189)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
artidoro: I would like to do another PR focusing on improving the public surface of text loader. I think these are all important points, and I will come back to them to improve the documentation. 
Before I focus on improving the documentation for these public fields I would like to make sure that we want to expose them all and that the current form is the final one.

---
In reply to: [255806866](https://github.com/dotnet/machinelearning/pull/2417#discussion_r255806866) [](ancestors = 255806866,255806745)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
artidoro: See my comment, but let me know if you think otherwise.

---
In reply to: [255806299](https://github.com/dotnet/machinelearning/pull/2417#discussion_r255806299) [](ancestors = 255806299)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
artidoro: I agree it does not make much sense to have this as a nullable type. We should just have a default. I would like to address this in another PR as I mention in my other comment.

---
In reply to: [255805797](https://github.com/dotnet/machinelearning/pull/2417#discussion_r255805797) [](ancestors = 255805797)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
artidoro: See my other comment regarding another PR after discussing the public surface of TextLoader.

---
In reply to: [255807082](https://github.com/dotnet/machinelearning/pull/2417#discussion_r255807082) [](ancestors = 255807082)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1145,7 +1238,14 @@ private sealed class LoaderHolder
 
             // Verify that the current schema-defining arguments are default.
             // Get settings just for core arguments, not everything.
-            string tmp = CmdParser.GetSettings(host, args, new ArgumentsCore());
+            string tmp = CmdParser.GetSettings(host, options, new Options()
```
artidoro: Null check.

Ivanidzo4ka:
You still have conflicting file :)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1179,30 +1280,32 @@ private sealed class LoaderHolder
...
-                ch.Assert(parsed);
-                // Copy the core arguments to the new args.
-                if (!CmdParser.ParseArguments(host, loader.GetSettingsString(), argsNew, typeof(ArgumentsCore), msg => ch.Error(msg)))
+                var optionsNew = new Options();
+                // Copy the core arguments to the new options.
+                if (!CmdParser.ParseArguments(host, loader.GetSettingsString(), optionsNew, typeof(Options), msg => ch.Error(msg)))
                     goto LDone;
 
-                cols = argsNew.Columns;
+                // Set the non core arguments in the new options.
```
abgoswam: >non core arguments [](start = 27, length = 18)

perhaps we should clarify what are "non core" arguments here ?   would be especially useful for the future.

without #2046 i did not have enough context to know what this meant

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1131,20 +1224,28 @@ private sealed class LoaderHolder
...
         {
             host.AssertValue(host);
             host.AssertValue(files);
+            host.CheckValue(options, nameof(options));
 
             cols = null;
             error = false;
 
             // Verify that the current schema-defining arguments are default.
             // Get settings just for core arguments, not everything.
```
abgoswam: >core arguments [](start = 37, length = 14)

perhaps we should rephrase this , so it makes sense for future readers ?

Issue #2297, opened 2019-01-29
Naming overhaul for IDataView subsystem


eerhardt:
We should make sure the type names in the IDataView subsystem are the names we want to use forever.

See 

1. https://github.com/dotnet/machinelearning/pull/2220#discussion_r251141578

> If all of these types are column types, should they include column in the name? e.g. TextColumnType?

2. https://github.com/dotnet/machinelearning/pull/2254#discussion_r251259500

> The idea was broached I know, but resolution never reached: should we rename? Or else perhaps make Schema a nested class of something appropriate, so as to give it definition?

> We could rename the class to DataViewSchema, or ViewSchema.
We’d probably want to rename everything then. Row is a pretty generic name. So is ColumnType.

________

# Proposal

1. We will ensure the class names in the `Microsoft.Data.DataView` package are unique by adding a prefix to any name deemed too general.
    - Currently we have 2 options for the prefix:
        - `DataView`, ex. `DataViewSchema`, `DataViewRow`, `DataViewType`.
        -  `DV`, ex. `DVSchema`, `DVRow`, `DVType`. 
            - Note: this is following the `Db` prefix pattern in `System.Data`: `DbDataReader`, `DbConnection`, `DbColumn`, etc.
            - It could be argued to lower-case the `v`: `DvSchema` to follow the same pattern above.
2. Add the base type suffix to all types that derive from the current `ColumnType` class, using whatever name we decide for the base `ColumnType` class. For example, assuming we rename `ColumnType` => `DataViewType`, we would have:
    - `NumberDataViewType`
    - `TextDataViewType`
    - `BooleanDataViewType`
    - etc
3. Rename the properties on [`NumberType`](https://github.com/dotnet/machinelearning/blob/17b07d8ec34a843410b125cecd4205428d1b9c69/src/Microsoft.Data.DataView/ColumnType.cs#L103) to match the type name in `System.` namespace.
    - ex. use `Int16` instead of `I2`. use `Single` instead of `R4`.
    - For the `UG` property and corresponding `struct RowId`, use whatever we rename the `Row` type, followed by `Id`. Example, `DataViewRowId`.
4. Rename the `Schema.Metadata` class to `Schema.Annotations`, and rename the property on `Schema.Column` from `Metadata` to `Annotations`.
[...]

PR #2419, opened 2019-02-05
Mechanical rename of Schema to DataSchema.

### LINKED ISSUE #1500: Final interface of Schema

sfilipi:
Mechanical rename of Schema to DataSchema, through VisualStudio's rename. 

Closes #1500
This is the last item in the list of items for #1500. 

@TomFinley @eerhardt valuable to do this,  at this time?


eerhardt:
We have #2297 as well for naming of the types in the `Microsoft.Data.DataView` package.

I think we may want to take a holistic approach to naming of all the types.

sfilipi:
Closing, since Eric has it covered on #2297

Issue #2253, opened 2019-01-26
AvxIntrinsics.DotU is Slower than the native version


Anipik:
We can verify this by running the benchmark https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Benchmarks/StochasticDualCoordinateAscentClassifierBench.cs#L157

``` ini

BenchmarkDotNet=v0.11.3, OS=Windows 10.0.17763.253 (1809/October2018Update/Redstone5)
Intel Xeon CPU E5-1650 v4 3.60GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=3.0.100-preview-009812
  [Host]     : .NET Core 2.1.6 (CoreCLR 4.6.27019.06, CoreFX 4.6.27019.05), 64bit RyuJIT
  Job-RQXZIO : .NET Core 2.1.6 (CoreCLR 4.6.27019.06, CoreFX 4.6.27019.05), 64bit RyuJIT

Arguments=/p:Configuration=Release  Toolchain=netcoreapp2.1  MaxIterationCount=20  
WarmupCount=1  

```
|      Method |     (netcoreapp2.1)Mean |    Error |   StdDev |        netcoreapp3.0 |
|------------ |--------------:|---------:|---------:|--------------------:|
| PredictIris | 515.6 ns | 3.097 ns | 2.745 ns | 620.3ns |

if we use ```UseIntrinsics=false``` Then we get the same performance for netcoreapp3.0

I verified using the perfview that the cause of regression is DotU. The Inclusive time fot DotU on netcoreapp3.0  is 95ns and on native version of cpumath is 11 ns

cc @danmosemsft @adamsitnik @tannergooding @eerhardt

adamsitnik:
@Anipik great job finding this regression!

the reason is most probably the fact that ML.NET typically has a very small input for all the CpuMath operations, something like an array of 7 or 14 floats. Our managed implementation is not optimized for that, IMHO the best would be to change the architecture/data flow in ML.NET

Anipik:
Same is the case with ```MakeIrisPredictions``` Benchmark
old - netcoreapp2.1
new - netcoreapp3.0

Base (old) Time: | 707.5
-- | --
Test (new) Time: | 786.9
Delta: | 79.4
Delta %: | 11.2


I verified using the perfview that the cause of regression is DotU. The Inclusive time fot DotU on netcoreapp3.0 is 192ms and on native version of cpumath is 106 ms

[...]

PR #2428, opened 2019-02-06
Minimum Vector size check added

Anipik:
Fixes https://github.com/dotnet/machinelearning/issues/2253

The new c# implementation is slower for inputs of smaller size, so adding a size constraint for using these new c# apis

### DIFF
--- a/src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
+++ b/src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
@@ -163,14 +163,14 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(destination);
 
-            if (destination.Length < MinInputSize || (!Avx.IsSupported && !Sse.IsSupported))
+            if (destination.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
             {
                 for (int i = 0; i < destination.Length; i++)
                 {
                     destination[i] += value;
                 }
             }
-            if (Avx.IsSupported)
+            else if (Avx.IsSupported)
             {
                 AvxIntrinsics.AddScalarU(value, destination);
             }
@@ -190,20 +190,20 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(destination);
 
-            if (Avx.IsSupported)
-            {
-                AvxIntrinsics.Scale(value, destination);
-            }
-            else if (Sse.IsSupported)
-            {
-                SseIntrinsics.Scale(value, destination);
-            }
-            else
+            if (destination.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
             {
                 for (int i = 0; i < destination.Length; i++)
                 {
                     destination[i] *= value;
                 }
+            }
+            else if (Avx.IsSupported)
+            {
+                AvxIntrinsics.Scale(value, destination);
+            }
+            else
+            {
+                SseIntrinsics.Scale(value, destination);
             }
         }
 
@@ -491,15 +491,7 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(source);
 
-            if (Avx.IsSupported)
-            {
-                return AvxIntrinsics.Sum(source);
-            }
-            else if (Sse.IsSupported)
-            {
-                return SseIntrinsics.Sum(source);
-            }
-            else
+            if (source.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
             {
                 float sum = 0;
                 for (int i = 0; i < source.Length; i++)
@@ -507,6 +499,14 @@ namespace Microsoft.ML.Internal.CpuMath
                     sum += source[i];
                 }
                 return sum;
+            }
+            else if (Avx.IsSupported)
+            {
+                return AvxIntrinsics.Sum(source);
+            }
+            else
+            {
+                return SseIntrinsics.Sum(source);
             }
         }
 
--- a/src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
+++ b/src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
@@ -163,7 +163,7 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(destination);
 
-            if (destination.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (destination.Length < MinInputSize || !Sse.IsSupported)
             {
                 for (int i = 0; i < destination.Length; i++)
                 {
@@ -190,7 +190,7 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(destination);
 
-            if (destination.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (destination.Length < MinInputSize || !Sse.IsSupported)
             {
                 for (int i = 0; i < destination.Length; i++)
                 {
@@ -224,7 +224,7 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= source.Length);
             Contracts.Assert(count <= destination.Length);
 
-            if (destination.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (destination.Length < MinInputSize || !Sse.IsSupported)
             {
                 for (int i = 0; i < count; i++)
                 {
@@ -255,7 +255,7 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(destination);
 
-            if (destination.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (destination.Length < MinInputSize || !Sse.IsSupported)
             {
                 for (int i = 0; i < destination.Length; i++)
                 {
@@ -288,7 +288,7 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= source.Length);
             Contracts.Assert(count <= destination.Length);
 
-            if (destination.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (destination.Length < MinInputSize || !Sse.IsSupported)
             {
                 for (int i = 0; i < count; i++)
                 {
@@ -324,7 +324,7 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= indices.Length);
             Contracts.Assert(count < destination.Length);
 
-            if (count < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (count < MinInputSize || !Sse.IsSupported)
             {
                 for (int i = 0; i < count; i++)
                 {
@@ -361,7 +361,7 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= destination.Length);
             Contracts.Assert(count <= result.Length);
 
-            if (count < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (count < MinInputSize || !Sse.IsSupported)
             {
                 for (int i = 0; i < count; i++)
                 {
@@ -393,7 +393,7 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= source.Length);
             Contracts.Assert(count <= destination.Length);
 
-            if (count < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (count < MinInputSize || !Sse.IsSupported)
             {
                 for (int i = 0; i < count; i++)
                 {
@@ -428,7 +428,7 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= indices.Length);
             Contracts.Assert(count < destination.Length);
 
-            if (count < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (count < MinInputSize || !Sse.IsSupported)
             {
                 for (int i = 0; i < count; i++)
                 {
@@ -464,7 +464,7 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= right.Length);
             Contracts.Assert(count <= destination.Length);
 
-            if (count < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (count < MinInputSize || !Sse.IsSupported)
             {
                 for (int i = 0; i < count; i++)
                 {
@@ -491,7 +491,7 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(source);
 
-            if (source.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (source.Length < MinInputSize || !Sse.IsSupported)
             {
                 float sum = 0;
                 for (int i = 0; i < source.Length; i++)
@@ -520,7 +520,7 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(source);
 
-            if (source.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (source.Length < MinInputSize || !Sse.IsSupported)
             {
                 float result = 0;
                 for (int i = 0; i < source.Length; i++)
@@ -550,7 +550,7 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(source);
 
-            if (source.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (source.Length < MinInputSize || !Sse.IsSupported)
             {
                 float result = 0;
                 for (int i = 0; i < source.Length; i++)
@@ -579,7 +579,7 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(source);
 
-            if (source.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (source.Length < MinInputSize || !Sse.IsSupported)
             {
                 float sum = 0;
                 for (int i = 0; i < source.Length; i++)
@@ -609,7 +609,7 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(source);
 
-            if (source.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (source.Length < MinInputSize || !Sse.IsSupported)
             {
                 float sum = 0;
                 for (int i = 0; i < source.Length; i++)
@@ -638,7 +638,7 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(source);
 
-            if (source.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (source.Length < MinInputSize || !Sse.IsSupported)
             {
                 float max = 0;
                 for (int i = 0; i < source.Length; i++)
@@ -672,7 +672,7 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(source);
 
-            if (source.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (source.Length < MinInputSize || !Sse.IsSupported)
             {
                 float max = 0;
                 for (int i = 0; i < source.Length; i++)
@@ -711,7 +711,7 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(left.Length >= count);
             Contracts.Assert(right.Length >= count);
 
-            if (count < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (count < MinInputSize || !Sse.IsSupported)
             {
                 float result = 0;
                 for (int i = 0; i < count; i++)
@@ -749,7 +749,7 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= right.Length);
             Contracts.Assert(count <= indices.Length);
 
-            if (count < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (count < MinInputSize || !Sse.IsSupported)
             {
                 float result = 0;
                 for (int i = 0; i < count; i++)
@@ -785,7 +785,7 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= left.Length);
             Contracts.Assert(count <= right.Length);
 
-            if (count < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (count < MinInputSize || !Sse.IsSupported)
             {
                 float norm = 0;
                 for (int i = 0; i < count; i++)
@@ -897,7 +897,7 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= v.Length);
             Contracts.Assert(count <= w.Length);
 
-            if (count < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (count < MinInputSize || !Sse.IsSupported)
             {
                 for (int i = 0; i < count; i++)
                 {
@@ -939,7 +939,7 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= v.Length);
             Contracts.Assert(count <= w.Length);
 
-            if (count < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
+            if (count < MinInputSize || !Sse.IsSupported)
             {
                 for (int i = 0; i < count; i++)
                 {
--- a/src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
+++ b/src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
@@ -20,6 +20,8 @@ namespace Microsoft.ML.Internal.CpuMath
         // The count of bytes in a 32-bit float, corresponding to _cbAlign in AlignedArray
         private const int FloatAlignment = 4;
 
+        private const int MinVectorSize = 16;
+
         // If neither AVX nor SSE is supported, return basic alignment for a 4-byte float.
         [MethodImplAttribute(MethodImplOptions.AggressiveInlining)]
         public static int GetVectorAlignment()
@@ -160,11 +162,11 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(destination);
 
-            if (Avx.IsSupported)
+            if (Avx.IsSupported && destination.Length >= MinVectorSize)
             {
                 AvxIntrinsics.AddScalarU(value, destination);
             }
-            else if (Sse.IsSupported)
+            else if (Sse.IsSupported && destination.Length >= MinVectorSize)
             {
                 SseIntrinsics.AddScalarU(value, destination);
             }
@@ -249,11 +251,11 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(destination);
 
-            if (Avx.IsSupported)
+            if (Avx.IsSupported && destination.Length >= MinVectorSize)
             {
                 AvxIntrinsics.ScaleAddU(scale, addend, destination);
             }
-            else if (Sse.IsSupported)
+            else if (Sse.IsSupported && destination.Length >= MinVectorSize)
             {
                 SseIntrinsics.ScaleAddU(scale, addend, destination);
             }
@@ -281,11 +283,11 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= source.Length);
             Contracts.Assert(count <= destination.Length);
 
-            if (Avx.IsSupported)
+            if (Avx.IsSupported && count >= MinVectorSize)
             {
                 AvxIntrinsics.AddScaleU(scale, source, destination, count);
             }
-            else if (Sse.IsSupported)
+            else if (Sse.IsSupported && count >= MinVectorSize)
             {
                 SseIntrinsics.AddScaleU(scale, source, destination, count);
             }
@@ -316,11 +318,11 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= indices.Length);
             Contracts.Assert(count < destination.Length);
 
-            if (Avx.IsSupported)
+            if (Avx.IsSupported && count >= MinVectorSize)
             {
                 AvxIntrinsics.AddScaleSU(scale, source, indices, destination, count);
             }
-            else if (Sse.IsSupported)
+            else if (Sse.IsSupported && count >= MinVectorSize)
             {
                 SseIntrinsics.AddScaleSU(scale, source, indices, destination, count);
             }
@@ -352,11 +354,11 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= destination.Length);
             Contracts.Assert(count <= result.Length);
 
-            if (Avx.IsSupported)
+            if (Avx.IsSupported && count >= MinVectorSize)
             {
                 AvxIntrinsics.AddScaleCopyU(scale, source, destination, result, count);
             }
-            else if (Sse.IsSupported)
+            else if (Sse.IsSupported && count >= MinVectorSize)
             {
                 SseIntrinsics.AddScaleCopyU(scale, source, destination, result, count);
             }
@@ -383,11 +385,11 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= source.Length);
             Contracts.Assert(count <= destination.Length);
 
-            if (Avx.IsSupported)
+            if (Avx.IsSupported && count >= MinVectorSize)
             {
                 AvxIntrinsics.AddU(source, destination, count);
             }
-            else if (Sse.IsSupported)
+            else if (Sse.IsSupported && count >= MinVectorSize)
             {
                 SseIntrinsics.AddU(source, destination, count);
             }
@@ -417,11 +419,11 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= indices.Length);
             Contracts.Assert(count < destination.Length);
 
-            if (Avx.IsSupported)
+            if (Avx.IsSupported && count >= MinVectorSize)
             {
                 AvxIntrinsics.AddSU(source, indices, destination, count);
             }
-            else if (Sse.IsSupported)
+            else if (Sse.IsSupported && count >= MinVectorSize)
             {
                 SseIntrinsics.AddSU(source, indices, destination, count);
             }
@@ -452,11 +454,11 @@ namespace Microsoft.ML.Internal.CpuMath
             Contracts.Assert(count <= right.Length);
             Contracts.Assert(count <= destination.Length);
 
-            if (Avx.IsSupported)
+            if (Avx.IsSupported && count >= MinVectorSize)
             {
                 AvxIntrinsics.MulElementWiseU(left, right, destination, count);
             }
-            else if (Sse.IsSupported)
+            else if (Sse.IsSupported && count >= MinVectorSize)
             {
                 SseIntrinsics.MulElementWiseU(left, right, destination, count);
             }
@@ -506,11 +508,11 @@ namespace Microsoft.ML.Internal.CpuMath
         {
             Contracts.AssertNonEmpty(source);
 
-            if (Avx.IsSupported)
+            if (Avx.IsSupported && source.Length >= MinVectorSize)
             {
                 return AvxIntrinsics.SumSqU(source);
             }
-            else if (Sse.IsSupported)
+            else if (Sse.IsSupported && source.Length >= MinVectorSize)
             {
                 return SseIntrinsics.SumSqU(source);
             }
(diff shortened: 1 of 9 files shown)


adamsitnik:
I have one question: what is the assembly code that we generate?

The  `Avx.IsSupported` and `Sse.IsSupported` are JIT Intrinsics which means that they get some extra treatment from JIT. JIT checks at runtime if they are supported and if so, it performs dead code elimination for other things.

Let's consider the following example:

```cs
if (Avx.IsSupported)
{
    AvxIntrinsics.AddScalarU(value, destination);
}
else if (Sse.IsSupported)
{
    SseIntrinsics.AddScalarU(value, destination);
}
else
{
    for (int i = 0; i < destination.Length; i++)
    {
        destination[i] += value;
    }
}
```

If AVX is supported, JIT will generate the following code for this method:

```
AvxIntrinsics.AddScalarU(value, destination);
```

What I want to know is what assembly code is generated, when we add `&& destination.Length >= MinVectorSize` to every condition? I am simply afraid that JIT will not remove the other paths.

@Anipik could you please check what code is generated? https://mijailovic.net/2018/07/05/generated-code/ gives a great explanation how you can do that (small hint: if you want to use BDN for that please remember to disable tiered jit first due to https://github.com/dotnet/BenchmarkDotNet/issues/899)

adamsitnik:
@Anipik maybe following pattern would do the job:

```cs
if (destination.Length < MinVectorSize || (!Avx.IsSupported && !Sse.IsSupported))
{
    for (int i = 0; i < destination.Length; i++)
    {
        destination[i] += value;
    }
}
else if (Avx.IsSupported)
{
    AvxIntrinsics.AddScalarU(value, destination);
}
else
{
    SseIntrinsics.AddScalarU(value, destination);
}
```

which I hope should be translated to the following (if Avx is supported)

```cs
if (destination.Length < MinVectorSize)
{
    for (int i = 0; i < destination.Length; i++)
    {
        destination[i] += value;
    }
}
else
{
    AvxIntrinsics.AddScalarU(value, destination);
}
```

adamsitnik:
@Anipik also, we need to make sure if after adding the extra codintion the method still gets inlined (it's important for very small arrays)

In PerfView, there is an "Events" view which shows you all the events from the trace file. You should be able to search for "Inlining" and the  method name and see if it got inlined or not. 

obraz

My guess is that it did not get inlined (it contains a loop) and we will have to use `[MethodOptions(MethodImpl.AgressiveInlining)]`

Anipik:
Thanks @adamsitnik for all the info. i will conduct all the experiments that you suggestion and will update the PR with some answers about inlining and assembly code

Anipik:
@adamsitnik in this folloowing case
```cs
if (Avx.IsSupported)
{
    AvxIntrinsics.AddScalarU(value, destination);
}
else if (Sse.IsSupported)
{
    SseIntrinsics.AddScalarU(value, destination);
}
else
{
    for (int i = 0; i < destination.Length; i++)
    {
        destination[i] += value;
    }
}
```

the disassembly still contains the sse code
```
--- C:\git\machinelearning\src\Microsoft.ML.CpuMath\CpuMathUtils.netcoreapp.cs -
   141:             if (Avx.IsSupported)
 sub         rsp,68h  
 vzeroupper  
 lea         rbp,[rsp+70h]  
 lea         rdi,[rbp-40h]  
 mov         ecx,0Eh  
 xor         eax,eax  
 rep stos    dword ptr [rdi]  
 mov         qword ptr [rbp+18h],rdx  
 vmovss      dword ptr [rbp+10h],xmm0  
 xor         edx,edx  
 cmp         edx,1  
 je          00007FF9FB2BBF98  
   142:             {
   143:                 AvxIntrinsics.AddScalarU(value, destination);
 vmovss      xmm0,dword ptr [rbp+10h]  
 vmovss      dword ptr [rbp-2Ch],xmm0  
 mov         rdx,qword ptr [rbp+18h]  
 vmovdqu     xmm0,xmmword ptr [rdx]  
 vmovdqu     xmmword ptr [rbp-28h],xmm0  
 vmovss      xmm0,dword ptr [rbp-2Ch]  
 lea         rdx,[rbp-28h]  
 call        00007FF9FA97D528  
 nop  
 lea         rsp,[rbp-8]  
 pop         rdi  
 pop         rbp  
 ret  
   145:             else if (Sse.IsSupported)
 xor         edx,edx  
 cmp         edx,1  
 je          00007FF9FB2BBFCC  
   146:             {
   147:                 SseIntrinsics.AddScalarU(value, destination);
 vmovss      xmm0,dword ptr [rbp+10h]  
 vmovss      dword ptr [rbp-44h],xmm0  
 mov         rdx,qword ptr [rbp+18h]  
 vmovdqu     xmm0,xmmword ptr [rdx]  
 vmovdqu     xmmword ptr [rbp-40h],xmm0  
 vmovss      xmm0,dword ptr [rbp-44h]  
 lea         rdx,[rbp-40h]  
 call        00007FF9FA97DA58  
 nop  
 lea         rsp,[rbp-8]  
 pop         rdi  
 pop         rbp  
 ret  
   149:             else
   150:             {
   151:                 for (int i = 0; i < destination.Length; i++)
 xor         ecx,ecx  
 mov         dword ptr [rbp-0Ch],ecx  
 jmp         00007FF9FB2BC000  
   152:                 {
   153:                     destination[i] += value;
 mov         rcx,qword ptr [rbp+18h]  
 mov         edx,dword ptr [rbp-0Ch]  
 call        00007FF9FB283480  
 mov         qword ptr [rbp-18h],rax  
 mov         rax,qword ptr [rbp-18h]  
 vmovss      xmm0,dword ptr [rax]  
 vaddss      xmm0,xmm0,dword ptr [rbp+10h]  
 mov         rax,qword ptr [rbp-18h]  
 vmovss      dword ptr [rax],xmm0  
   151:                 for (int i = 0; i < destination.Length; i++)
 mov         eax,dword ptr [rbp-0Ch]  
 inc         eax  
 mov         dword ptr [rbp-0Ch],eax  
 mov         rcx,qword ptr [rbp+18h]  
 call        00007FF9FB283408  
 cmp         eax,dword ptr [rbp-0Ch]  
 jg          00007FF9FB2BBFD3  
 lea         rsp,[rbp-8]  
 pop         rdi  
 pop         rbp  
```

am I missing sumthing here ?

adamsitnik:
@Anipik  Which tool did you use? VS? If so, did you disable the Tiered Compilation first?

Anipik:
Yes i used the vs. I though  i have to disable tired jitting if i am using benchmarkdotnet ? and isnt it already disabled by default for netcoreapp3.0 ? a function gets jitted if it is being called more than 30 times ?

adamsitnik:
@Anipik there is a bug when it comes to getting the disassembly https://github.com/dotnet/BenchmarkDotNet/issues/899

adamsitnik:
> a function gets jitted if it is being called more than 30 times

and executed for longer than 100ms.

The best way is to simply disable the Tiered Jit and check the disassembly

Anipik:
@adamsitnik I checked the numbers and Disassembly for both the cases.

``` ini

BenchmarkDotNet=v0.11.3, OS=Windows 10.0.17763.253 (1809/October2018Update/Redstone5)
Intel Xeon CPU E5-1650 v4 3.60GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=3.0.100-preview-010184
  [Host]     : .NET Core 3.0.0-preview-27324-5 (CoreCLR 4.6.27322.0, CoreFX 4.7.19.7311), 64bit RyuJIT
  Job-UCTBXC : .NET Core 3.0.0-preview-27324-5 (CoreCLR 4.6.27322.0, CoreFX 4.7.19.7311), 64bit RyuJIT

Arguments=/p:Configuration=Release-Intrinsics  Toolchain=netcoreapp3.0  MaxIterationCount=20  
WarmupCount=1  

```
|                      Method | original |     Adam  |
|---------------------------- |---------:|----------:|
| MakeBreastCancerPredictions | 2.148 ms | 2.168  ms |
| MakeIrisPredictions         | 4.864 ms | 4.859  ms |
| PredictIris                 | 490.0 ns | 499.6  ns |


the numbers in the first 2 benchmarks are the same but for the last benchmark there is slight difference

Anipik:
The IL code for original approach

```IL
; Lcl frame size = 56

G_M5980_IG01:
       push     rdi
       push     rsi
       sub      rsp, 56
       vzeroupper 
       xor      rax, rax
       mov      qword ptr [rsp+28H], rax
       mov      qword ptr [rsp+30H], rax

G_M5980_IG02:
       mov      rsi, bword ptr [rdx]
       mov      edi, dword ptr [rdx+8]
       cmp      edi, 4
       jl       SHORT G_M5980_IG04
       lea      rdx, bword ptr [rsp+28H]
       mov      bword ptr [rdx], rsi
       mov      dword ptr [rdx+8], edi
       lea      rdx, bword ptr [rsp+28H]
       call     Program:AddScalarU(float,struct)
       nop      

G_M5980_IG03:
       add      rsp, 56
       pop      rsi
       pop      rdi
       ret      

G_M5980_IG04:
       xor      eax, eax
       test     edi, edi
       jle      SHORT G_M5980_IG06

G_M5980_IG05:
       movsxd   rdx, eax
       lea      rdx, bword ptr [rsi+4*rdx]
       vmovaps  xmm1, xmm0
       vaddss   xmm1, dword ptr [rdx]
       vmovss   dword ptr [rdx], xmm1
       inc      eax
       cmp      eax, edi
       jl       SHORT G_M5980_IG05

G_M5980_IG06:
       add      rsp, 56
       pop      rsi
       pop      rdi
       ret      

; Total bytes of code 99, prolog size 21 for method Program:Add(float,struct)
; ============================================================

````
The IL code for adam approach 
```IL
; Lcl frame size = 56

G_M5979_IG01:
       sub      rsp, 56
       vzeroupper 
       xor      rax, rax
       mov      qword ptr [rsp+28H], rax
       mov      qword ptr [rsp+30H], rax

G_M5979_IG02:
       mov      rax, bword ptr [rdx]
       mov      edx, dword ptr [rdx+8]
       cmp      edx, 4
       jge      SHORT G_M5979_IG05

G_M5979_IG03:
       xor      ecx, ecx
       test     edx, edx
       jle      SHORT G_M5979_IG05

G_M5979_IG04:
       movsxd   r8, ecx
       lea      r8, bword ptr [rax+4*r8]
       vmovaps  xmm1, xmm0
       vaddss   xmm1, dword ptr [r8]
       vmovss   dword ptr [r8], xmm1
       inc      ecx
       cmp      ecx, edx
       jl       SHORT G_M5979_IG04

G_M5979_IG05:
       lea      rcx, bword ptr [rsp+28H]
       mov      bword ptr [rcx], rax
       mov      dword ptr [rcx+8], edx
       lea      rdx, bword ptr [rsp+28H]
       call     Program:AddScalarU(float,struct)
       nop      

G_M5979_IG06:
       add      rsp, 56
       ret      

; Total bytes of code 90, prolog size 19 for method Program:Add(float,struct)
; ============================================================

```

In both the cases sse part of the conditions are not generated if the avx is supported on the machine.

Anipik:
@adamsitnik can you take another go at the comments as well as code ?

Anipik:
@adammodlin @eerhardt @tannergooding can you please take a look at this  ?

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
```diff
@@ -20,6 +20,8 @@ internal static partial class CpuMathUtils
         // The count of bytes in a 32-bit float, corresponding to _cbAlign in AlignedArray
         private const int FloatAlignment = 4;
 
+        private const int MinVectorSize = 16;
```
tannergooding: Maybe this should be called `MinVectorSizeX86`? Arm, for example, actually supports a `Vector64<>` type.

tannergooding: Ah, I see. This is saying only use this code path for 16 elements arrays. The name is slightly confusing.

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
```diff
@@ -156,15 +158,16 @@ public static void MatrixTimesSource(bool transpose, AlignedArray matrix, Aligne
         /// </summary>
         /// <param name="value">The value to add.</param>
         /// <param name="destination">The destination to add the value to.</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static void Add(float value, Span<float> destination)
         {
             Contracts.AssertNonEmpty(destination);
 
-            if (Avx.IsSupported)
+            if (Avx.IsSupported && destination.Length >= MinVectorSize)
```
tannergooding: I would actually expect the `Avx.IsSupported` and `Sse.IsSupported` cases to have different "MinVectorSize"

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
```diff
@@ -156,15 +158,16 @@ public static void MatrixTimesSource(bool transpose, AlignedArray matrix, Aligne
...
             {
                 AvxIntrinsics.AddScalarU(value, destination);
             }
-            else if (Sse.IsSupported)
+            else if (Sse.IsSupported && destination.Length >= MinVectorSize)
```
tannergooding: I would actually expect this to be beneficial for `4` or more elements, possibly with a check to special-case this scenario in order to avoid the "leading" vs "trailing" element handling...

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
```diff
@@ -915,11 +939,11 @@ public static void SdcaL1UpdateSparse(float primalUpdate, int count, ReadOnlySpa
             Contracts.Assert(count <= v.Length);
             Contracts.Assert(count <= w.Length);
 
-            if (Avx.IsSupported)
+            if (Avx.IsSupported && count >= MinVectorSize)
```
tannergooding: I think, if there is now a `count >= size` checks, the various implementations have some dead-code paths.

adamsitnik:
@Anipik could you also provide the C# code for the disassembly that you have provided?

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
Anipik: changed to ```MinInputSize```

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
Anipik: actually it tried a smaller count for sse (8) but it was still slower so i increased it to 16 and then is was faster

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
Anipik: Actually not. eg if the inputsize is 21 and avx.is supported

It will execute all three code blocks, avx codeblock (0-16), sseblock (16-20) and simple codeblock (21)

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
Anipik: > I would actually expect this to be beneficial for `4` or more elements,

The benchmarks were still slower with MinVectorSize = 4

> possibly with a check to special-case this scenario in order to avoid the "leading" vs "trailing" element handling.

i didnt understand this. can you explain it little bit more ? Did you mean to do go for the software implementation for trailing element ?

Anipik:
@adamsitnik I didn't have the C# code earlier so I just did everything again

the C# code is 
```C#
    class Program
    {
        private static int MinInputSize = 16;
        {
            Console.WriteLine("Hello World!");
            float[] testArray2 = new float[15] { 1.96f, -2.38f, -9.76f, 13.84f, -106.37f, -26.93f, 32.45f, 3.29f, 1.96f, -2.38f, -9.76f, 13.84f, -106.37f, -26.93f, 32.45f };
            Add(1.7f, testArray2);
        }

        public static void Add(float value, Span<float> destination)
        {
            if (destination.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
            {
                for (int i = 0; i < destination.Length; i++)
                {
                    destination[i] += value;
                }
            }
            else if (Avx.IsSupported)
            {
                AddScalarU(value, destination);
            }
            else
            {
                AddScalarU2(value, destination);
            }
        }

        public static unsafe void AddScalarU(float scalar, Span<float> dst)
        {
            fixed (float* pdst = &MemoryMarshal.GetReference(dst))
            {
                float* pDstEnd = pdst + dst.Length;
                float* pDstCurrent = pdst;
                float* pVectorizationEnd = pDstEnd - 4;

                Vector256<float> scalarVector256 = Vector256.Create(scalar);

                while (pDstCurrent + 8 <= pDstEnd)
                {
                    Vector256<float> dstVector = Avx.LoadVector256(pDstCurrent);
                    dstVector = Avx.Add(dstVector, scalarVector256);
                    Avx.Store(pDstCurrent, dstVector);

                    pDstCurrent += 8;
                }

                Vector128<float> scalarVector128 = Vector128.Create(scalar);

                if (pDstCurrent <= pVectorizationEnd)
                {
                    Vector128<float> dstVector = Sse.LoadVector128(pDstCurrent);
                    dstVector = Sse.Add(dstVector, scalarVector128);
                    Sse.Store(pDstCurrent, dstVector);

                    pDstCurrent += 4;
                }

                while (pDstCurrent < pDstEnd)
                {
                    Vector128<float> dstVector = Sse.LoadScalarVector128(pDstCurrent);
                    dstVector = Sse.AddScalar(dstVector, scalarVector128);
                    Sse.StoreScalar(pDstCurrent, dstVector);

                    pDstCurrent++;
                }
            }
        }

        public static unsafe void AddScalarU2(float scalar, Span<float> dst)
        {
            fixed (float* pdst = &MemoryMarshal.GetReference(dst))
            {
                float* pDstEnd = pdst + dst.Length;
                float* pDstCurrent = pdst;
                float* pVectorizationEnd = pDstEnd - 4;

                Vector128<float> scalarVector = Vector128.Create(scalar);

                while (pDstCurrent <= pVectorizationEnd)
                {
                    Vector128<float> dstVector = Sse.LoadVector128(pDstCurrent);
                    dstVector = Sse.Add(dstVector, scalarVector);
                    Sse.Store(pDstCurrent, dstVector);

                    pDstCurrent += 4;
                }

                while (pDstCurrent < pDstEnd)
                {
                    Vector128<float> dstVector = Sse.LoadScalarVector128(pDstCurrent);
                    dstVector = Sse.AddScalar(dstVector, scalarVector);
                    Sse.StoreScalar(pDstCurrent, dstVector);

                    pDstCurrent++;
                }
            }
        }

    }
```

When i produced disassembly the assembly code was as you suggested

Anipik:
Original Approach
```IL
Hello World!
; Assembly listing for method Program:Add(float,struct)
; Emitting BLENDED_CODE for X64 CPU with AVX - Windows
; optimized code
; rsp based frame
; fully interruptible
; Final local variable assignments
;
;  V00 arg0         [V00,T08] (  5,  7   )   float  ->  mm6        
;  V01 arg1         [V01,T02] (  4,  8   )   byref  ->  rdx         ld-addr-op
;  V02 loc0         [V02,T01] (  5, 16.50)     int  ->  rax        
;  V03 OutArgs      [V03    ] (  1,  1   )  lclBlk (32) [rsp+0x00]   "OutgoingArgSpace"
;  V04 tmp1         [V04,T00] (  3, 24   )   byref  ->  rdx         "dup spill"
;  V05 tmp2         [V05,T04] (  4,  6   )   byref  ->  rsi         V07._pointer(offs=0x00) P-INDEP "field V01._pointer (fldOffset=0x0)"
;  V06 tmp3         [V06,T03] (  7,  8   )     int  ->  rdi         V07._length(offs=0x08) P-INDEP "field V01._length (fldOffset=0x8)"
;* V07 tmp4         [V07    ] (  0,  0   )  struct (16) zero-ref    "Promoted implicit byref"
;  V08 tmp5         [V08    ] (  4,  4   )  struct (16) [rsp+0x20]   do-not-enreg[XSB] must-init addr-exposed "by-value struct argument"
;  V09 tmp6         [V09,T05] (  3,  3   )   byref  ->  rdx         stack-byref "BlockOp address local"
;  V10 tmp7         [V10,T09] (  2,  2   )   float  ->  mm0         "argument with side effect"
;  V11 tmp8         [V11,T06] (  3,  3   )   byref  ->  rdx         stack-byref "BlockOp address local"
;  V12 tmp9         [V12,T10] (  2,  2   )   float  ->  mm0         "argument with side effect"
;  V13 cse0         [V13,T07] (  3,  2.50)     int  ->  rbx         "ValNumCSE"
;
; Lcl frame size = 64

G_M20289_IG01:
       57                   push     rdi
       56                   push     rsi
       53                   push     rbx
       4883EC40             sub      rsp, 64
       C5F877               vzeroupper 
       C5F829742430         vmovaps  qword ptr [rsp+30H], xmm6
       33C0                 xor      rax, rax
       4889442420           mov      qword ptr [rsp+20H], rax
       4889442428           mov      qword ptr [rsp+28H], rax
       C5F828F0             vmovaps  xmm6, xmm0

G_M20289_IG02:
       488B32               mov      rsi, bword ptr [rdx]
       8B7A08               mov      edi, dword ptr [rdx+8]
       48B93025E231F97F0000 mov      rcx, 0x7FF931E22530
       BA01000000           mov      edx, 1
       E8663EB05F           call     CORINFO_HELP_GETSHARED_NONGCSTATIC_BASE
       8B1DF4D60800         mov      ebx, dword ptr [reloc classVar[0x31e23d38]]
       3BFB                 cmp      edi, ebx
       7C2A                 jl       SHORT G_M20289_IG04
       C5F828C6             vmovaps  xmm0, xmm6
       488D542420           lea      rdx, bword ptr [rsp+20H]
       488932               mov      bword ptr [rdx], rsi
       897A08               mov      dword ptr [rdx+8], edi
       488D542420           lea      rdx, bword ptr [rsp+20H]
       E8BB2FFFFF           call     Program:AddScalarU(float,struct)
       90                   nop      

G_M20289_IG03:
       C5F828742430         vmovaps  xmm6, qword ptr [rsp+30H]
       4883C440             add      rsp, 64
       5B                   pop      rbx
       5E                   pop      rsi
       5F                   pop      rdi
       C3                   ret      

G_M20289_IG04:
       3BFB                 cmp      edi, ebx
       7C2A                 jl       SHORT G_M20289_IG06
       C5F828C6             vmovaps  xmm0, xmm6
       488D542420           lea      rdx, bword ptr [rsp+20H]
       488932               mov      bword ptr [rdx], rsi
       897A08               mov      dword ptr [rdx+8], edi
       488D542420           lea      rdx, bword ptr [rsp+20H]
       E8972FFFFF           call     Program:AddScalarU2(float,struct)
       90                   nop      

G_M20289_IG05:
       C5F828742430         vmovaps  xmm6, qword ptr [rsp+30H]
       4883C440             add      rsp, 64
       5B                   pop      rbx
       5E                   pop      rsi
       5F                   pop      rdi
       C3                   ret      

G_M20289_IG06:
       33C0                 xor      eax, eax
       85FF                 test     edi, edi
       7E1C                 jle      SHORT G_M20289_IG08

G_M20289_IG07:
       4863D0               movsxd   rdx, eax
       488D1496             lea      rdx, bword ptr [rsi+4*rdx]
       C5F828C6             vmovaps  xmm0, xmm6
       C5FA5802             vaddss   xmm0, dword ptr [rdx]
       C5FA1102             vmovss   dword ptr [rdx], xmm0
       FFC0                 inc      eax
       3BC7                 cmp      eax, edi
       7CE7                 jl       SHORT G_M20289_IG07

G_M20289_IG08:
       C5F828742430         vmovaps  xmm6, qword ptr [rsp+30H]
       4883C440             add      rsp, 64
       5B                   pop      rbx
       5E                   pop      rsi
       5F                   pop      rdi
       C3                   ret      

; Total bytes of code 197, prolog size 32 for method Program:Add(float,struct)
; ============================================================
```
Adam approach 
```
Hello World!
; Assembly listing for method Program:Add(float,struct)
; Emitting BLENDED_CODE for X64 CPU with AVX - Windows
; optimized code
; rsp based frame
; fully interruptible
; Final local variable assignments
;
;  V00 arg0         [V00,T06] (  4,  6.50)   float  ->  mm6        
;  V01 arg1         [V01,T02] (  4,  8   )   byref  ->  rdx         ld-addr-op
;  V02 loc0         [V02,T01] (  5, 16.50)     int  ->  rdx        
;  V03 OutArgs      [V03    ] (  1,  1   )  lclBlk (32) [rsp+0x00]   "OutgoingArgSpace"
;  V04 tmp1         [V04,T00] (  3, 24   )   byref  ->  rax         "dup spill"
;  V05 tmp2         [V05,T04] (  3,  5.50)   byref  ->  rsi         V07._pointer(offs=0x00) P-INDEP "field V01._pointer (fldOffset=0x0)"
;  V06 tmp3         [V06,T03] (  5,  7   )     int  ->  rdi         V07._length(offs=0x08) P-INDEP "field V01._length (fldOffset=0x8)"
;* V07 tmp4         [V07    ] (  0,  0   )  struct (16) zero-ref    "Promoted implicit byref"
;  V08 tmp5         [V08    ] (  2,  2   )  struct (16) [rsp+0x20]   do-not-enreg[XSB] must-init addr-exposed "by-value struct argument"
;  V09 tmp6         [V09,T05] (  3,  3   )   byref  ->  rdx         stack-byref "BlockOp address local"
;  V10 tmp7         [V10,T07] (  2,  2   )   float  ->  mm0         "argument with side effect"
;
; Lcl frame size = 72

G_M20288_IG01:
       57                   push     rdi
       56                   push     rsi
       4883EC48             sub      rsp, 72
       C5F877               vzeroupper 
       C5F829742430         vmovaps  qword ptr [rsp+30H], xmm6
       33C0                 xor      rax, rax
       4889442420           mov      qword ptr [rsp+20H], rax
       4889442428           mov      qword ptr [rsp+28H], rax
       C5F828F0             vmovaps  xmm6, xmm0

G_M20288_IG02:
       488B32               mov      rsi, bword ptr [rdx]
       8B7A08               mov      edi, dword ptr [rdx+8]
       48B93025DD1CF97F0000 mov      rcx, 0x7FF91CDD2530
       BA01000000           mov      edx, 1
       E8673EB25F           call     CORINFO_HELP_GETSHARED_NONGCSTATIC_BASE
       3B3DF5D60800         cmp      edi, dword ptr [reloc classVar[0x1cdd3d38]]
       7D30                 jge      SHORT G_M20288_IG06

G_M20288_IG03:
       33D2                 xor      edx, edx
       85FF                 test     edi, edi
       7E1C                 jle      SHORT G_M20288_IG05

G_M20288_IG04:
       4863C2               movsxd   rax, edx
       488D0486             lea      rax, bword ptr [rsi+4*rax]
       C5F828C6             vmovaps  xmm0, xmm6
       C5FA5800             vaddss   xmm0, dword ptr [rax]
       C5FA1100             vmovss   dword ptr [rax], xmm0
       FFC2                 inc      edx
       3BD7                 cmp      edx, edi
       7CE7                 jl       SHORT G_M20288_IG04

G_M20288_IG05:
       C5F828742430         vmovaps  xmm6, qword ptr [rsp+30H]
       4883C448             add      rsp, 72
       5E                   pop      rsi
       5F                   pop      rdi
       C3                   ret      

G_M20288_IG06:
       C5F828C6             vmovaps  xmm0, xmm6
       488D542420           lea      rdx, bword ptr [rsp+20H]
       488932               mov      bword ptr [rdx], rsi
       897A08               mov      dword ptr [rdx+8], edi
       488D542420           lea      rdx, bword ptr [rsp+20H]
       E8922FFFFF           call     Program:AddScalarU(float,struct)
       90                   nop      

G_M20288_IG07:
       C5F828742430         vmovaps  xmm6, qword ptr [rsp+30H]
       4883C448             add      rsp, 72
       5E                   pop      rsi
       5F                   pop      rdi
       C3                   ret      

; Total bytes of code 148, prolog size 31 for method Program:Add(float,struct)
; ============================================================
````
```
```

**Comment by Anipik** [comment]
``` ini

BenchmarkDotNet=v0.11.3, OS=Windows 10.0.17763.253 (1809/October2018Update/Redstone5)
Intel Xeon CPU E5-1650 v4 3.60GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=3.0.100-preview-010184
  [Host]     : .NET Core 3.0.0-preview-27324-5 (CoreCLR 4.6.27322.0, CoreFX 4.7.19.7311), 64bit RyuJIT
  Job-UCTBXC : .NET Core 3.0.0-preview-27324-5 (CoreCLR 4.6.27322.0, CoreFX 4.7.19.7311), 64bit RyuJIT

Arguments=/p:Configuration=Release-Intrinsics  Toolchain=netcoreapp3.0  MaxIterationCount=20  
WarmupCount=1  

```
|                      Method | original |     Adam  |    percentage  |
|---------------------------- |---------:|----------:|----------:|
| MakeBreastCancerPredictions | 1.929 ms | 1.904  ms | 1.29% |
| MakeIrisPredictions         | 4.929 ms | 4.916  ms | 0.2%  |
| PredictIris                 | 489.1 ns | 497.6  ns |  -1.3% |

The numbers are same for the benchmarks. the only plus point with the original approach is that it is easier to read. @adamsitnik any thoughts ? which approach we should finalize

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
tannergooding: Some of the functions in `SseIntrinsics.cs` and `AvxIntrinsics.cs` have code paths like:
```cs
if (length < 4)
{
    // Handle cases where we have less than 128-bits total and can't ever use SIMD acceleration.
    switch (length)
    {
        case 3: dst[2] *= scale; goto case 2;
        case 2: dst[1] *= scale; goto case 1;
        case 1: dst[0] *= scale; break;
    }
    return;
}
```

These code paths should be considered dead now.

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
tannergooding: Take `Scale` as an example. It currently has 5-6 checks for the average code path, which allows it to  deal with with leading elements, the main block, and trailing elements.

For large inputs, this is fine and the branch checks are amortized. However, for small inputs, these branches add up and could be considered unnecessary (in cases like 4 elements, the branches are also likely more than the cost of the code actually executing).

I would expect that 4 or more elements is still the cutoff, but due to the current code structure, the various leading/trailing/alignment checks cause the code to look worse. Perhaps if we special-cased inputs that are between 4-16 elements (or some profiled upper bound) and only dealt with it via unaligned loads for these small cases, it would still be faster (on average) than the scalar code path.

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
tannergooding: Are you indicating that `Sse` for `destination.Length >= 8` was slower than `Avx` for `destination.Length >= 16`?

But for `Sse` when `destination.Length >= 16` it was faster than `Avx` for `destination.Length >= 16`?

Anipik: I am not doing a unit perf testing here. I am directly testing end to end benchmarks here. The cases I tried was
1- avx 8, sse 8
2- avx 16, sse 8
3- avx  16, sse 16

Among these, 3 gave the best results (although difference was not much difference between 2 and 3 and was within the error range)
So I am fine either way (using 2 or 3)

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
Anipik: Actually i didnt change those cpumathfunctions which are already using the software implementation for shorter paths. eg scale and sum. but sure i can make them similar to the other cpumathfunctions

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
tannergooding: I was trying to indicate that I believe that we should have two distinct constants: `SseMinVectorSize` (for the 128-bit code paths) and `AvxMinVectorSize` (for the 256-bit code paths).

I would expect that:
* Scalar is faster for 0 to X
* 128-bit is faster for X to Y
* 256-bit is faster for Y and above
Where X != Y and both X and Y are powers of 2.

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
Anipik: > Perhaps if we special-cased inputs that are between 4-16 elements (or some profiled upper bound) and only dealt with it via unaligned loads for these small cases, it would still be faster (on average) than the scalar code path.

yes we can try doing that. But we are using this updated algorithm of trailing and leading elements, only for sum and scale. Before that we should first need to move other functions to this new implementation. For this PR, i suggest just adding this check

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
tannergooding: Ah, I see. I don't think we need to do anything here then.

adamsitnik:
@Anipik personally I would go with "my approach" because it's producing smaller code size (197 vs 148 bytes) and if we want to enforce method inlining it's going to affect all the callers of these methods. I think that in this particular case performance is more important than readability.

@tannergooding what do you think? I might be biased here ;)

adamsitnik:
Also when it comes to PredictIris difference it's few nanoseconds, less than 2% so I would not worry about this, it can be just noise.

Anipik:
@adamsitnik i added your change and some other minor unrelated changes

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
Anipik: @tannergooding wont it depend upon on the operation being conducted also ? eg if we compare 2 functions one involving addition and one involving multiplication, wont the constants be different for both the cases ?

Anipik:
@tannergooding can you take another look ? The pr fixes some of the regressions , we can do other optimisations in a separate PR

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
```diff
@@ -156,24 +158,25 @@ public static void MatrixTimesSource(bool transpose, AlignedArray matrix, Aligne
         /// </summary>
         /// <param name="value">The value to add.</param>
         /// <param name="destination">The destination to add the value to.</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static void Add(float value, Span<float> destination)
         {
             Contracts.AssertNonEmpty(destination);
 
-            if (Avx.IsSupported)
+            if (destination.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
```
tannergooding: nit: You should only need to check `|| !Sse.IsSupported`.

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
```diff
@@ -281,20 +288,20 @@ public static void AddScale(float scale, ReadOnlySpan<float> source, Span<float>
             Contracts.Assert(count <= source.Length);
             Contracts.Assert(count <= destination.Length);
 
-            if (Avx.IsSupported)
+            if (destination.Length < MinInputSize || !(Avx.IsSupported || Sse.IsSupported))
             {
-                AvxIntrinsics.AddScaleU(scale, source, destination, count);
+                for (int i = 0; i < count; i++)
```
tannergooding: just a nit: Given that this is always less than 16, I imagine it would be worth checking the numbers for the unrolled version of the loop.

test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
```diff
@@ -33,11 +33,11 @@ public class CpuMathUtilsUnitTests
...
-            float[] testArray1 = new float[16] { 1.96f, -2.38f, -9.76f, 13.84f, -106.37f, -26.93f, 32.45f, 3.29f, 1.96f, -2.38f, -9.76f, 13.84f, -106.37f, -26.93f, 32.45f, 3.29f };
+            float[] testArray1 = new float[32] { 1.96f, -2.38f, -9.76f, 13.84f, -106.37f, -26.93f, 32.45f, 3.29f, 1.96f, -2.38f, -9.76f, 13.84f, -106.37f, -26.93f, 32.45f, 3.29f, 1f, 2f, 3f, 4f, 5f, 6f, 7f, 8f, 9f, 10f, 11f, 12f, 13f, 14f, 15f, 16f };
             // Unpadded array whose length is not a multiple of 4.
-            float[] testArray2 = new float[15] { 1.96f, -2.38f, -9.76f, 13.84f, -106.37f, -26.93f, 32.45f, 3.29f, 1.96f, -2.38f, -9.76f, 13.84f, -106.37f, -26.93f, 32.45f };
+            float[] testArray2 = new float[30] { 1.96f, -2.38f, -9.76f, 13.84f, -106.37f, -26.93f, 32.45f, 3.29f, 1.96f, -2.38f, -9.76f, 13.84f, -106.37f, -26.93f, 32.45f, 1f, 2f, 3f, 4f, 5f, 6f, 7f, 8f, 9f, 10f, 11f, 12f, 13f, 14f, 15f };
             _testArrays = new float[][] { testArray1, testArray2 };
-            _testIndexArray = new int[9] { 0, 2, 5, 6, 8, 11, 12, 13, 14 };
+            _testIndexArray = new int[18] { 0, 2, 5, 6, 8, 11, 12, 13, 14, 16, 18, 21, 22, 24, 26, 27, 28, 29};
```
tannergooding: nit: These are all cases that will hit the "fast" (intrinsic) path.

Anipik: Yes, This will go through the intrinsics path But when we will be running these tests in avx and sse disabled mode then we will be able to test the software implementation path

src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs
Anipik: sure i will run the numbers with unrolled version of the loop

test/Microsoft.ML.CpuMath.UnitTests.netcoreapp/UnitTests.cs
tannergooding: It would probably be good to have at least one code path that always hits the non-accelerated path, as it properly validates that it works in the machines default configuration (rather than only validating it works when the user explicitly disables the accelerated paths).

PR #2431, opened 2019-02-06
ITransformer derives from ICanSaveModel and explicit implementation for ICanSaveModel

### LINKED ISSUE #2336: ITransformer should derive from ICanSaveModel and we should explicitly implement ICanSaveModel
In our code base we make the implicit assumption that all classes implementing `ITransformer` also implement `ICanSaveModel`. `ITransformer` is the abstraction that transforms the data and both single and chained components in a pipeline are made into an `ITransformer` once they are trained. Since this is the abstraction that contains trained models and data transformations, it should also be savable to a file. The standard scenario is that after training, one saves trained models, or `ITransformer`, for later use for scoring/transforming new data.

The above highlights that this is an important assumption and that we should capture it in code by making `ITransformer` derive from `ICanSaveModel`. Especially going forward as we think about allowing custom components, we should make it explicit that new transformers also implement `ICanSaveModel`.

Furthermore, we should explicitly implement in the transformers the `Save` method of the `ICanSaveModel` interface. This would make it less visible to the users. The reasons for doing so are twofold:
1.  `ModelSaveContext` only has internal constructors, so a user cannot instantiate it. 
2. We only use the `Save` method internally, and we don't want the user to call it directly.

/cc @TomFinley @sfilipi

artidoro:
Fixes  #2336.

This PR makes `ITransformer` derive from `ICanSaveModel`. The reasons for changing the core `ITransformer/IEstimator/IDataView` API are illustrated in more detail in the issue #2336. But this essentially captures the idea that once trained the model should be savable. 
For this change I moved `ICanSaveModel, ModelSaveContext, ModelLoadContext` and a few related utility files to Microsoft.ML.Core as `ITransformer` lives in that assembly.

The second dimension of this change is the explicit implementation of the interface `ICanSaveModel` everywhere. This will make it a requirement to cast the classes to `ICanSaveModel` in order to access the method `Save(ModelSaveContext ctx)` so that it is less visible. I have covered the entire code base with my changes, including some classes that are internal and others that I think should be made internal shortly. This makes my change touch ~140 files. I decided to do so in order to make the way we implement the `ICanSaveModel` interface consistent everywhere.


src/Microsoft.ML.FastTree/TreeEnsemble/InternalRegressionTree.cs
```diff
@@ -409,7 +409,7 @@ protected void Save(ModelSaveContext ctx, TreeType code)
             writer.WriteDoubleArray(_previousLeafValue);
         }
 
-        public virtual void Save(ModelSaveContext ctx)
+        internal virtual void Save(ModelSaveContext ctx)
```
TomFinley: >internal [](start = 8, length = 8)

Since this is already an internal class there is no point in having this be internal I think.

src/Microsoft.ML.HalLearners/VectorWhitening.cs
```diff
@@ -464,7 +464,7 @@ private static void GetColTypesAndIndex(IHostEnvironment env, IDataView inputDat
             }
         }
 
-        public override void Save(ModelSaveContext ctx)
+        private protected override void SaveModel(ModelSaveContext ctx)
```
TomFinley: >SaveModel [](start = 40, length = 9)

It's good that we're finally locking this down, `protected` members like these are a little appreciated part of our public API we need to lock down. Thanks @artidoro!

src/Microsoft.ML.FastTree/TreeEnsembleFeaturizer.cs
```diff
@@ -387,7 +387,7 @@ public TreeEnsembleFeaturizerBindableMapper(IHostEnvironment env, ModelLoadConte
             _totalLeafCount = CountLeaves(_ensemble);
         }
 
-        public void Save(ModelSaveContext ctx)
+        void ICanSaveModel.Save(ModelSaveContext ctx)
```
TomFinley: >void ICanSaveModel.Save [](start = 8, length = 23)

Note that even though this is itself an internal class, I still appreciate making these implementations explicit, since I help it helps establish the pattern. (So I like this change, in case not clear. :) )(

src/Microsoft.ML.Ensemble/OutputCombiners/BaseStacking.cs
```diff
@@ -15,7 +15,7 @@
 
 namespace Microsoft.ML.Ensemble.OutputCombiners
 {
-    internal abstract class BaseStacking<TOutput> : IStackingTrainer<TOutput>
+    internal abstract class BaseStacking<TOutput> : IStackingTrainer<TOutput>, ICanSaveModel
```
TomFinley: >, ICanSaveModel [](start = 77, length = 15)

Geez, every time I look at this assembly it's something else... thanks for fixing this @artidoro!

src/Microsoft.ML.Core/Data/ModelLoading.cs
```diff
@@ -10,6 +10,11 @@
 
 namespace Microsoft.ML.Model
 {
+    /// <summary>
+    /// Signature for a repository based model loader. This is the dual of ICanSaveModel.
+    /// </summary>
+    public delegate void SignatureLoadModel(ModelLoadContext ctx);
```
TomFinley: >public delegate void SignatureLoadModel [](start = 4, length = 39)

Maybe we can `[BestFriend] internal` this as we are with other delegates.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1283,7 +1283,8 @@ internal static IDataLoader Create(IHostEnvironment env, Arguments args, IMultiS
         internal static IDataView ReadFile(IHostEnvironment env, Arguments args, IMultiStreamSource fileSource)
             => new TextLoader(env, args, fileSource).Read(fileSource);
 
-        public void Save(ModelSaveContext ctx)
+        void ICanSaveModel.Save(ModelSaveContext ctx) => SaveModel(ctx);
+        internal void SaveModel(ModelSaveContext ctx)
```
TomFinley: >internal [](start = 8, length = 8)

Can this be private?

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -126,7 +126,11 @@ public IRowToRowMapper GetRowToRowMapper(Schema inputSchema)
             return (IRowToRowMapper)Scorer.ApplyToData(Host, new EmptyDataView(Host, inputSchema));
         }
 
-        protected void SaveModel(ModelSaveContext ctx)
+        void ICanSaveModel.Save(ModelSaveContext ctx) => SaveModel(ctx);
+
+        private protected abstract void SaveModel(ModelSaveContext ctx);
+
+        protected void SaveModelCore(ModelSaveContext ctx)
```
TomFinley: Also, I think `SaveModelCore` should at least be `private protected`, probably best-friended as well. (Actually I might think that prediction transformers should be best-friended on `SaveModel` as well, but I don't insist on it.

src/Microsoft.ML.Data/Transforms/RowToRowTransformerBase.cs
```diff
@@ -23,7 +23,8 @@ protected RowToRowTransformerBase(IHost host)
             Host = host;
         }
 
-        public abstract void Save(ModelSaveContext ctx);
+        void ICanSaveModel.Save(ModelSaveContext ctx) => SaveModel(ctx);
+        private protected abstract void SaveModel(ModelSaveContext ctx);
```
TomFinley: >private protected abstract void SaveModel(ModelSaveContext ctx); [](start = 8, length = 64)

Even though we might not be making use of it right now, I think we still ought to make this `[BestFriend]` to signal at least to internal developers that this is *intended* as something that someone implementing this, that this is definitely an intended expansion point.

src/Microsoft.ML.Data/Transforms/GenerateNumberTransform.cs
```diff
@@ -163,7 +163,7 @@ public static Bindings Create(ModelLoadContext ctx, Schema input)
                 return new Bindings(useCounter, states, input, false, names);
             }
 
-            public void Save(ModelSaveContext ctx)
+            internal void Save(ModelSaveContext ctx)
```
TomFinley: >internal [](start = 12, length = 8)

This is on the private class `Bindings`, right? There's no need I think to make this internal.

src/Microsoft.ML.FastTree/TreeEnsemble/InternalQuantileRegressionTree.cs
```diff
@@ -50,7 +50,7 @@ public InternalQuantileRegressionTree(byte[] buffer, ref int position)
             _instanceWeights = buffer.ToDoubleArray(ref position);
         }
 
-        public override void Save(ModelSaveContext ctx)
+        internal override void Save(ModelSaveContext ctx)
```
TomFinley: >internal [](start = 8, length = 8)

Similar note that methods on internal classes don't need to themselves be marked internal.

src/Microsoft.ML.Data/Transforms/RowToRowTransformerBase.cs
TomFinley: Actually, wait, how is this actually working? I see `ImageGrayscalingTransformer`, which I *think* is overriding this method. That's definitely in an entirely separate assembly. So how is this working without this being declared a `[BestFriend]`? (I'm sort of working via scrolling through the text of your review, so I could easily have missed something, but I have to expect that this is the root of all the basic transformer functionality.) If it is somehow working, which I don't fully understand how it could, this means that there's probably some bug or unintended behavior in `[BestFriend]`? Or am I wrong?

---
In reply to: [254475884](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254475884) [](ancestors = 254475884)

src/Microsoft.ML.ImageAnalytics/VectorToImageTransform.cs
```diff
@@ -195,7 +195,7 @@ public ColInfoEx(ModelLoadContext ctx)
                 Interleave = ctx.Reader.ReadBoolByte();
             }
 
-            public void Save(ModelSaveContext ctx)
+            internal void Save(ModelSaveContext ctx)
```
TomFinley: >internal [](start = 12, length = 8)

Same notes about `ColInfoEx` being a private class, no need to internalize its methods.

src/Microsoft.ML.PCA/PcaTransformer.cs
```diff
@@ -140,7 +140,7 @@ public TransformInfo(ModelLoadContext ctx)
                 Contracts.CheckDecode(MeanProjected == null || (MeanProjected.Length == Rank && FloatUtils.IsFinite(MeanProjected)));
             }
 
-            public void Save(ModelSaveContext ctx)
+            internal void Save(ModelSaveContext ctx)
```
TomFinley: >internal [](start = 12, length = 8)

Same note about private classes etc. etc.

src/Microsoft.ML.TimeSeries/SequenceModelerBase.cs
```diff
@@ -75,6 +75,8 @@ private protected SequenceModelerBase()
         /// <summary>
         /// Implementation of <see cref="ICanSaveModel.Save(ModelSaveContext)"/>.
         /// </summary>
-        public abstract void Save(ModelSaveContext ctx);
+        void ICanSaveModel.Save(ModelSaveContext ctx) => SaveModel(ctx);
+
+        private protected abstract void SaveModel(ModelSaveContext ctx);
```
TomFinley: >SaveModel [](start = 40, length = 9)

Unlike the other things, I view this as not being `[BestFriend]` appropriate since it is *very* particular to time series. (So I'd keep this one the way it is.)

src/Microsoft.ML.Transforms/GroupTransform.cs
```diff
@@ -305,7 +305,7 @@ private Schema BuildOutputSchema(Schema sourceSchema)
                 return schemaBuilder.GetSchema();
             }
 
-            public void Save(ModelSaveContext ctx)
+            internal void Save(ModelSaveContext ctx)
```
TomFinley: >internal [](start = 12, length = 8)

Same note private grumble grumble nested mumble mumble...

src/Microsoft.ML.Transforms/OneHotEncoding.cs
```diff
@@ -165,7 +165,7 @@ internal OneHotEncodingTransformer(ValueToKeyMappingEstimator term, IEstimator<I
 
         public IDataView Transform(IDataView input) => _transformer.Transform(input);
 
-        public void Save(ModelSaveContext ctx) => _transformer.Save(ctx);
+        void ICanSaveModel.Save(ModelSaveContext ctx) => (_transformer as ICanSaveModel).Save(ctx);
```
TomFinley: >_transformer as ICanSaveModel [](start = 58, length = 29)

Do not use `as` style casts unless you are entertaining the possibility that it might not be this thing. We know for a fact it must be this thing, so `()` style casts are appropriate.

src/Microsoft.ML.Transforms/OneHotHashEncoding.cs
```diff
@@ -189,7 +189,7 @@ internal OneHotHashEncodingTransformer(HashingEstimator hash, IEstimator<ITransf
         /// </summary>
         public IDataView Transform(IDataView input) => _transformer.Transform(input);
 
-        public void Save(ModelSaveContext ctx) => _transformer.Save(ctx);
+        void ICanSaveModel.Save(ModelSaveContext ctx) => (_transformer as ICanSaveModel).Save(ctx);
```
TomFinley: >as ICanSaveModel [](start = 71, length = 16)

Similar about `()` style casts.

src/Microsoft.ML.TimeSeries/SequentialTransformerBase.cs
```diff
@@ -450,9 +452,9 @@ protected override RowCursor GetRowCursorCore(IEnumerable<Schema.Column> columns
             public override RowCursor[] GetRowCursorSet(IEnumerable<Schema.Column> columnsNeeded, int n, Random rand = null)
                 => new RowCursor[] { GetRowCursorCore(columnsNeeded, rand) };
 
-            public override void Save(ModelSaveContext ctx)
+            private protected override void SaveModel(ModelSaveContext ctx)
             {
-                _parent.Save(ctx);
+                (_parent as ICanSaveModel).Save(ctx);
```
TomFinley: >_parent as ICanSaveModel [](start = 17, length = 24)

Similar about `()` style casts when you know it must implement this interface.

src/Microsoft.ML.Transforms/ProduceIdTransform.cs
```diff
@@ -60,7 +60,7 @@ public static Bindings Create(ModelLoadContext ctx, Schema input)
                 return new Bindings(input, true, name);
             }
 
-            public void Save(ModelSaveContext ctx)
+            internal void Save(ModelSaveContext ctx)
```
TomFinley: >internal [](start = 12, length = 8)

Consternation, uproar!

src/Microsoft.ML.Transforms/Text/LdaTransform.cs
```diff
@@ -340,7 +340,7 @@ internal LdaSummary GetLdaSummary(VBuffer<ReadOnlyMemory<char>> mapping)
                 }
             }
 
-            public void Save(ModelSaveContext ctx)
+            internal void Save(ModelSaveContext ctx)
```
TomFinley: >internal [](start = 12, length = 8)

Grumble grumble mumble mumble private.

src/Microsoft.ML.Transforms/Text/NgramTransform.cs
```diff
@@ -162,7 +162,7 @@ public TransformInfo(ModelLoadContext ctx, bool readWeighting)
                 NonEmptyLevels = ctx.Reader.ReadBoolArray(NgramLength);
             }
 
-            public void Save(ModelSaveContext ctx)
+            internal void Save(ModelSaveContext ctx)
```
TomFinley: >internal [](start = 12, length = 8)

Grumble grumble mumble mumble private.

src/Microsoft.ML.Data/Transforms/RowToRowTransformerBase.cs
TomFinley: I guess *overriding* a method doesn't count as a *reference* to it, so the analyzer rule wouldn't be triggered. Huh. Not sure if I should care or not...

TomFinley:
Looks mostly good @artidoro except for one or two unnecessary internalizations. Do we have though a handle on why these `ReturnTypeIsSchemaShape` and `ReturnTypeIsSchemaShapeChained` tests are failing? Are you planning on de-WIPing this?

src/Microsoft.ML.FastTree/TreeEnsemble/InternalRegressionTree.cs
artidoro: Sounds good!

---
In reply to: [254471904](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254471904) [](ancestors = 254471904)

src/Microsoft.ML.FastTree/TreeEnsembleFeaturizer.cs
artidoro: Ok makes sense :)

---
In reply to: [254472337](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254472337) [](ancestors = 254472337)

src/Microsoft.ML.Core/Data/ModelLoading.cs
artidoro: Ok that makes sense

---
In reply to: [254473032](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254473032) [](ancestors = 254473032)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
artidoro: Yes it can! Good point thank you.

---
In reply to: [254473480](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254473480) [](ancestors = 254473480)

src/Microsoft.ML.Data/Transforms/RowToRowTransformerBase.cs
artidoro: It is the core of the transformers, so yes I guess that this is not being picked up by the analyzer.
But I agree in making it BestFriend. 

---
In reply to: [254554561](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254554561) [](ancestors = 254554561)

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
artidoro: I made the `SaveModel` BestFriend, I think that's what you meant. 

---
In reply to: [254475268](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254475268) [](ancestors = 254475268)

src/Microsoft.ML.Data/Transforms/GenerateNumberTransform.cs
artidoro: Sounds good, I will revert here.

---
In reply to: [254486483](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254486483) [](ancestors = 254486483)

src/Microsoft.ML.Transforms/OneHotEncoding.cs
artidoro: Thank you, I didn't know!

---
In reply to: [254489054](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254489054) [](ancestors = 254489054)

src/Microsoft.ML.Transforms/ProduceIdTransform.cs
artidoro: Resignation!

---
In reply to: [254489466](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254489466) [](ancestors = 254489466)

artidoro:
Thank you very much for taking a look Tom! I am resolving your comments, and de-WIPing. Will look into the test failures afterwards.

---
In reply to: [461297633](https://github.com/dotnet/machinelearning/pull/2431#issuecomment-461297633) [](ancestors = 461297633)

TomFinley:
OK. Except for that test failure I consider all my comments addressed so will approve.

---
In reply to: [461315617](https://github.com/dotnet/machinelearning/pull/2431#issuecomment-461315617) [](ancestors = 461315617,461297633)

src/Microsoft.ML.Core/Data/ModelHeader.cs
```diff
@@ -10,7 +10,8 @@
 
 namespace Microsoft.ML.Model
 {
-    [StructLayout(LayoutKind.Explicit, Size = ModelHeader.Size)]
+    [StructLayout(LayoutKind.Explicit, Size = Size)]
+    [BestFriend]
```
wschin: Do we want to sort attributes?

src/Microsoft.ML.Data/Model/Repository.cs
```diff
@@ -10,11 +10,6 @@
 
 namespace Microsoft.ML.Model
 {
-    /// <summary>
-    /// Signature for a repository based model loader. This is the dual of ICanSaveModel.
-    /// </summary>
-    public delegate void SignatureLoadModel(ModelLoadContext ctx);
-
```
wschin: I guess the reason of removing this is that we choose `ICanSaveModel` as the universal way to dump a trained model.

src/Microsoft.ML.Core/Data/Repository.cs
```diff
@@ -10,11 +10,6 @@
 
 namespace Microsoft.ML.Model
 {
-    /// <summary>
-    /// Signature for a repository based model loader. This is the dual of ICanSaveModel.
-    /// </summary>
-    public delegate void SignatureLoadModel(ModelLoadContext ctx);
-
     /// <summary>
     /// For saving a model into a repository.
```
wschin: I'd suggest to mention the two major design patterns introduced in this PR.
(1) A class inherited `ICanSaveModel` from a base class should overwrite the function invoked by `Save(ModelSaveContext ctx)` in that base class.
(2) A class declared as a `ICanSaveModel` should do explicit implementation of `Save(ModelSaveContext ctx)`.

src/Microsoft.ML.Data/DataView/LambdaColumnMapper.cs
```diff
@@ -140,7 +140,7 @@ private sealed class Impl<T1, T2, T3> : OneToOneTransformBase
                 Metadata.Seal();
             }
 
-            public override void Save(ModelSaveContext ctx)
+            private protected override void SaveModel(ModelSaveContext ctx)
```
wschin: >SaveModel [](start = 44, length = 9)

May we call it `SaveCore` following the convention of `TrainCore`?

src/Microsoft.ML.Data/Depricated/Instances/HeaderSchema.cs
```diff
@@ -151,7 +151,7 @@ private static VersionInfo GetVersionInfo()
                 loaderAssemblyName: typeof(FeatureNameCollection).Assembly.FullName);
         }
 
-        public static void Save(ModelSaveContext ctx, in VBuffer<ReadOnlyMemory<char>> names)
+        internal static void Save(ModelSaveContext ctx, in VBuffer<ReadOnlyMemory<char>> names)
```
wschin: >internal [](start = 8, length = 8)

private protected?

src/Microsoft.ML.Data/Scorers/GenericScorer.cs
```diff
@@ -114,7 +114,7 @@ public Bindings ApplyToSchema(IHostEnvironment env, Schema input)
                 return Create(env, bindable, input, roles, suffix, user: false);
             }
 
-            public override void Save(ModelSaveContext ctx)
+            internal override void SaveModel(ModelSaveContext ctx)
```
wschin: >internal [](start = 12, length = 8)

`private protected` if possible.

src/Microsoft.ML.Data/Transforms/PerGroupTransformBase.cs
```diff
@@ -133,7 +133,9 @@ protected PerGroupTransformBase(IHostEnvironment env, ModelLoadContext ctx, IDat
             GroupCol = ctx.LoadNonEmptyString();
         }
 
-        public virtual void Save(ModelSaveContext ctx)
+        void ICanSaveModel.Save(ModelSaveContext ctx) => SaveModel(ctx);
+
+        internal virtual void SaveModel(ModelSaveContext ctx)
```
wschin: >internal [](start = 8, length = 8)

private protected?

src/Microsoft.ML.Transforms/OneHotHashEncoding.cs
wschin: @tfinley@gmail.com, do you mean we need to use `(ICanSaveModel)_transformer` because `_transformer` must implement `ICanSaveModel`?

---
In reply to: [254489139](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254489139) [](ancestors = 254489139)

src/Microsoft.ML.Data/Evaluators/EvaluatorBase.cs
```diff
@@ -510,7 +510,9 @@ protected PerInstanceEvaluatorBase(IHostEnvironment env, ModelLoadContext ctx,
                 throw Host.ExceptSchemaMismatch(nameof(schema), "score", ScoreCol);
         }
 
-        public virtual void Save(ModelSaveContext ctx)
+        void ICanSaveModel.Save(ModelSaveContext ctx) => SaveModel(ctx);
+
+        private protected virtual void SaveModel(ModelSaveContext ctx)
```
wschin: >SaveModel [](start = 39, length = 9)

<summary>
Derived class, for example `A`, should overwrite 'SaveModel` so that `((ICanSaveModel)A).Save(ctx)` can correctly dump `A`.
</summary>

src/Microsoft.ML.Data/DataView/LambdaColumnMapper.cs
artidoro: There are quite a few instances where there is a `SaveCore` method that was referenced by `SaveModel` (the previous `Save` method). That's why I preferred not to name it `SaveCore`. If it's fine with you I would keep it like this.

src/Microsoft.ML.Core/Data/ModelHeader.cs
artidoro: Didin't know about this best practice, thank you for letting me know, it makes sense :)

---
In reply to: [254933949](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254933949) [](ancestors = 254933949)

src/Microsoft.ML.Data/Model/Repository.cs
artidoro: No I just moved it to the `ModelLoading` file.

---
In reply to: [254961987](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254961987) [](ancestors = 254961987)

src/Microsoft.ML.Data/Depricated/Instances/HeaderSchema.cs
artidoro: The class is internal, so I think I should actually revert this change.

---
In reply to: [254963619](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254963619) [](ancestors = 254963619)

src/Microsoft.ML.Data/Scorers/GenericScorer.cs
artidoro: In this case it is not possible, because `BindingsBase` is not `ICanSaveModel`, and other classes need to access this method. However, I went through all the instance of `internal.* Save(ModelSaveContext...` and fixed a couple places where we could make private protected, so thank you for pointing that out.

---
In reply to: [254963789](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254963789) [](ancestors = 254963789)

src/Microsoft.ML.Data/Evaluators/EvaluatorBase.cs
artidoro: Thank you !

---
In reply to: [254968381](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254968381) [](ancestors = 254968381)

src/Microsoft.ML.Core/Data/Repository.cs
artidoro: I agree, I will do this.

---
In reply to: [254963056](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254963056) [](ancestors = 254963056)

artidoro: Only can add the second part however, because the first part is not general. There can be a base class that knows how to save, and thus simply implements Save directly with no additional abstract function invocation. 

---
In reply to: [255227041](https://github.com/dotnet/machinelearning/pull/2431#discussion_r255227041) [](ancestors = 255227041,254963056)

src/Microsoft.ML.Data/Evaluators/RankerEvaluator.cs
```diff
@@ -715,7 +715,7 @@ public Transform(IHostEnvironment env, ModelLoadContext ctx, IDataView input)
                 _bindings = new Bindings(Host, input.Schema, false, LabelCol, ScoreCol, GroupCol, _truncationLevel);
             }
 
-            public override void Save(ModelSaveContext ctx)
+            internal override void SaveModel(ModelSaveContext ctx)
```
Ivanidzo4ka: >internal [](start = 12, length = 8)

why internal and not private protected as in other evalutors?

src/Microsoft.ML.Data/Scorers/RowToRowScorerBase.cs
```diff
@@ -49,7 +49,7 @@ private protected RowToRowScorerBase(IHost host, ModelLoadContext ctx, IDataView
             ctx.LoadModel<ISchemaBindableMapper, SignatureLoadModel>(host, out Bindable, "SchemaBindableMapper");
         }
 
-        public sealed override void Save(ModelSaveContext ctx)
+        private protected sealed override void SaveModel(ModelSaveContext ctx)
```
Ivanidzo4ka: >private protected sealed override [](start = 8, length = 33)

do you need all this modifiers?

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -134,7 +134,11 @@ public IRowToRowMapper GetRowToRowMapper(Schema inputSchema)
...
```
Ivanidzo4ka: >SaveModelCore [](start = 23, length = 13)

funny thing, this method get called only in one of three classes derived from PredictionTransformerBase

src/Microsoft.ML.Data/Transforms/GenerateNumberTransform.cs
Ivanidzo4ka: it's still internal

---
In reply to: [254568639](https://github.com/dotnet/machinelearning/pull/2431#discussion_r254568639) [](ancestors = 254568639,254486483)

src/Microsoft.ML.Data/Transforms/RowToRowTransformerBase.cs
Ivanidzo4ka: >private protected [](start = 8, length = 17)

I spent probably 10 mintues going through MSDN in attempt to understand how private protected which should be accessible only in this assembly works in OnnxTransform and other assemblies. And then I notice same question from Tom, but on top of file. Can you please stop force-pushing changes, it's screw ups codeflow in a bad way, and no way anyone will do codereview in web interface for 140 files.

src/Microsoft.ML.Core/Data/ModelLoading.cs
```diff
@@ -10,6 +10,11 @@
 
 namespace Microsoft.ML.Model
 {
+    /// <summary>
+    /// Signature for a repository based model loader. This is the dual of ICanSaveModel.
```
wschin: <see cref="ICanSaveModel">

src/Microsoft.ML.Data/Transforms/RowToRowTransformerBase.cs
Ivanidzo4ka: part of me want you to search for all 
```
void ICanSaveModel.Save(ModelSaveContext ctx) => SaveModel(ctx);
private protected abstract void SaveModel(ModelSaveContext ctx);
```
and replace them to
```
void ICanSaveModel.Save(ModelSaveContext ctx) => SaveModel(ctx);

private protected abstract void SaveModel(ModelSaveContext ctx);
```
but it's up to you

src/Microsoft.ML.Data/Transforms/TransformBase.cs
```diff
@@ -388,7 +390,7 @@ public VectorType GetSlotType(int col)
                 return new Bindings(parent, infos, inputSchema, false, names);
             }
 
-            public void Save(ModelSaveContext ctx)
+            internal void Save(ModelSaveContext ctx)
```
Ivanidzo4ka: >internal [](start = 12, length = 8)

it's in private class, any reason for this change?

src/Microsoft.ML.Core/Data/Repository.cs
wschin: I'd like to have every use documented. It can provide future devs a clear view of your clever idea.

---
In reply to: [255228894](https://github.com/dotnet/machinelearning/pull/2431#discussion_r255228894) [](ancestors = 255228894,255227041,254963056)

src/Microsoft.ML.Data/Transforms/TransformBase.cs
wschin: I guess everything should be as conservative as possible?

---
In reply to: [255247956](https://github.com/dotnet/machinelearning/pull/2431#discussion_r255247956) [](ancestors = 255247956)

Ivanidzo4ka:
>     public sealed class ProduceIdTransform : RowToRowTransformBase

if you have intent of pushing another iteration, you can make this transform internal :)

---
Refers to: src/Microsoft.ML.Transforms/ProduceIdTransform.cs:30 in 5b44cae. [](commit_id = 5b44cae0a50c3135d17c7bd4d8724bce9b552099, deletion_comment = False)

src/Microsoft.ML.Core/Data/IEstimator.cs
```diff
@@ -263,7 +264,7 @@ public interface IDataReaderEstimator<in TSource, out TReader>
     /// The transformer is a component that transforms data.
     /// It also supports 'schema propagation' to answer the question of 'how will the data with this schema look, after you transform it?'.
     /// </summary>
-    public interface ITransformer
+    public interface ITransformer : ICanSaveModel
```
Ivanidzo4ka: >ICanSaveModel [](start = 36, length = 13)

We need to reflect this change, and others in our [https://github.com/dotnet/machinelearning/blob/master/docs/code/MlNetHighLevelConcepts.md](https://github.com/dotnet/machinelearning/blob/master/docs/code/MlNetHighLevelConcepts.md)
But that probably required another issue.

test/Microsoft.ML.CodeAnalyzer.Tests/Helpers/DiagnosticVerifier.cs
```diff
@@ -267,7 +268,7 @@ private static string FormatDiagnostics(DiagnosticAnalyzer analyzer, params Diag
 
         private static readonly MetadataReference MSDataDataViewReference = RefFromType<IDataView>();
         private static readonly MetadataReference MLNetCoreReference = RefFromType<IHostEnvironment>();
-        private static readonly MetadataReference MLNetDataReference = RefFromType<Model.ModelLoadContext>();
+        private static readonly MetadataReference MLNetDataReference = RefFromType<KeyToValueMappingTransformer>();
```
Ivanidzo4ka: >KeyToValueMappingTransformer [](start = 83, length = 28)

maybe MlContext?

src/Microsoft.ML.Data/Transforms/RowToRowTransformerBase.cs
artidoro: I can replace that sure!

---
In reply to: [255247150](https://github.com/dotnet/machinelearning/pull/2431#discussion_r255247150) [](ancestors = 255247150)

artidoro: Only 4 occurrences :) 

---
In reply to: [255276034](https://github.com/dotnet/machinelearning/pull/2431#discussion_r255276034) [](ancestors = 255276034,255247150)

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
artidoro: Yes I had noticed that and found it quite strange!

---
In reply to: [255240879](https://github.com/dotnet/machinelearning/pull/2431#discussion_r255240879) [](ancestors = 255240879)

src/Microsoft.ML.Data/Transforms/RowToRowTransformerBase.cs
artidoro: OK sounds good sorry about that!

---
In reply to: [255246374](https://github.com/dotnet/machinelearning/pull/2431#discussion_r255246374) [](ancestors = 255246374)

test/Microsoft.ML.CodeAnalyzer.Tests/Helpers/DiagnosticVerifier.cs
artidoro: Done!

---
In reply to: [255255322](https://github.com/dotnet/machinelearning/pull/2431#discussion_r255255322) [](ancestors = 255255322)

src/Microsoft.ML.Core/Data/IEstimator.cs
artidoro: I am going to publish a separate issue about this, so that maybe we can also double check that it is up to date with everything

---
In reply to: [255254348](https://github.com/dotnet/machinelearning/pull/2431#discussion_r255254348) [](ancestors = 255254348)

src/Microsoft.ML.Data/Transforms/TransformBase.cs
artidoro: Yes you are right I can undo this

---
In reply to: [255253748](https://github.com/dotnet/machinelearning/pull/2431#discussion_r255253748) [](ancestors = 255253748,255247956)

src/Microsoft.ML.Core/Data/Repository.cs
artidoro: Sounds good, I added a conditional case at the end of the sentence so that it actually matches all cases.

---
In reply to: [255252654](https://github.com/dotnet/machinelearning/pull/2431#discussion_r255252654) [](ancestors = 255252654,255228894,255227041,254963056)

src/Microsoft.ML.Data/Scorers/RowToRowScorerBase.cs
artidoro: I think so,  sealed override means that derived classes cannot override this method. Since it is a base class, I think it could be fair that we want to tell developers deriving from this class that this should not be overwritten. 

the change from public to private protected comes from the fact that the base class is public, however, I agree that maybe in the future that could not be useful anymore. For the time being I would keep it like this.

---
In reply to: [255239221](https://github.com/dotnet/machinelearning/pull/2431#discussion_r255239221) [](ancestors = 255239221)

Issue #1519, opened 2018-11-03
Best friends to limit friend access


TomFinley:
While we stabilize the API, it is best if we keep hidden as much as possible, since once v1.0 the barrier to changing the public types on a publicly released library will be high. For that reason the thinking is to keep the scope of what we publicly release as limited as possible. If we suppose that *usage* of the library vs. *component authorship* inside the library, one obvious way to reduce the surface area is to ideally hide (or, in the worst case, render useless) all the items related to component authorship.

The problem is, over the years we have accumulated have a fairly huge library, so there are a tremendous number of types, and a fair amount of infrastructure. We cannot make that infrastructure internal simply and put everything in one big assembly, because some components or their dependencies are quite large, so we want to keep them in separate nugets, which means separate assemblies. But, that means public classes.

We could make the core infrastructure assemblies friends of all leaf assemblies via the `InternalsVisibleTo` attribute, but that would render *all* internal classes usable from the leaf assemblies, which again is not what we want from a code maintainability perspective.

We really want to have a mechanism that distinguishes internal to the assembly, and internal to the library. There is no concept in .NET, and based on conversations with the .NET folks there seems to be little belief such a thing might be useful in any other situation, so we'll roll our own.

The best idea we have settled on so far is that we set things to be friends as described above, make things internal, *but* for these internal items we want to share, make sure that there is an attribute (provisionally called `BestFriend`) on the item, and if not, the code analyzer we have internal to ML.NET will complain.

So, e.g.:

```csharp
internal sealed class Foo { }

[BestFriend]
internal sealed class Bar
{
    [BestFriend]
    internal int A { get; }
    internal int B { get; }
}

public abstract class Biz
{
    [BestFriend]
    private internal Biz() { }
}
```

In this scenario, if we consider some hypothetical other project code trying to access this, access to `Foo` and `Bar.B` would be restricted (despite being friends!), but access to `Bar`, `Bar.A`, and a class inheriting from `Biz` would be possible in.

Note here, `Biz` itself is still public since we very often have user-accessible classes descend from some class we would like to be "internal," but kind of can't, at least, not without introducing a significant absurd amount of wrapping. This level of "visible but unusable" though we should try to keep to a minimum.

You could imagine this attribute being an opt-out rather than opt-in, but opt-in seems a bit more reasonable since we want to treat these cross-assembly sharings as being somehow special, and it would be nice if in PRs it becomes obvious when there's something like this going on.

The work then is:

[...]

PR #2434, opened 2019-02-06
Add analyzer for detecting BestFriend usages on public declarations

### LINKED ISSUE #2415: BestFriend on public members ought to complain
We have the best friend attribute introduced in #1520, which is great, but sometimes I or someone makes a mistake and puts it on a public class. This is harmless, has no visible effects, but is just a bit odd and definitely not intended usage.

At least for me, the procedure is this: I make something `public` become `[BestFriend] internal`, but then I realize something is a bit more complicated than I thought, and I decide to delay that *specific* internalization work until some other conditions are met. (E.g., in #2404, I internalized `IPredictorProducing`, but then realized that the problems with calibrators represented a hard barrier, so I wrote #2378 to capture those difficulties, and changed it back to public, until that is resolved.) The trouble is sometimes I leave the `[BestFriend]` behind. This has happened multiple times.

It might be nice, though hardly essential, to have `[BestFriend]` itself be analyzed to see that any usage of it is not applied to any publicly accessible thing. This is not actually essential per se, it's just a code cleanliness type of thing.

This hypothetical work would go in `Microsoft.ML.InternalCodeAnalyzer`.

mareklinka:
As proposed in #2415, the `BestFriend` attribute should never be applied to public declarations. While this is not a problem per se it still introduces unnecessary noise. I implemented a code analyzer that will flag these occurrences as errors whenever a `[BestFriend]` attribute is applied to a:

1. public class/struct/enum/interface
2. public method/property/field/delegate
3. public constructor

I didn't implement an automated code fix for this diagnostic as fixing the problem is trivial - either remove the attribute or make the declaration internal. If we still want a code fix to be created I'll gladly take care of that as well.

Fixes #2415.

### DIFF
--- a/test/Microsoft.ML.CodeAnalyzer.Tests/Code/BestFriendOnPublicDeclarationTest.cs
+++ b/test/Microsoft.ML.CodeAnalyzer.Tests/Code/BestFriendOnPublicDeclarationTest.cs
@@ -1,0 +1,62 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Reflection;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.ML.CodeAnalyzer.Tests.Helpers;
+using Xunit;
+
+namespace Microsoft.ML.InternalCodeAnalyzer.Tests
+{
+    public sealed class BestFriendOnPublicDeclarationTest : DiagnosticVerifier<BestFriendOnPublicDeclarationsAnalyzer>
+    {
+        private readonly Lazy<string> SourceAttribute = TestUtils.LazySource("BestFriendAttribute.cs");
+        private readonly Lazy<string> SourceDeclaration = TestUtils.LazySource("BestFriendOnPublicDeclaration.cs");
+
+        [Fact]
+        public void BestFriendOnPublicDeclaration()
+        {
+            Solution solution = null;
+            var projA = CreateProject("ProjectA", ref solution, SourceDeclaration.Value, SourceAttribute.Value);
+
+            var analyzer = new BestFriendOnPublicDeclarationsAnalyzer();
+
+            var refs = new List<MetadataReference> {
+                RefFromType<object>(), RefFromType<Attribute>(),
+                MetadataReference.CreateFromFile(Assembly.Load("netstandard, Version=2.0.0.0").Location),
+                MetadataReference.CreateFromFile(Assembly.Load("System.Runtime, Version=0.0.0.0").Location)
+            };
+
+            var comp = projA.GetCompilationAsync().Result.WithReferences(refs.ToArray());
+            var compilationWithAnalyzers = comp.WithAnalyzers(ImmutableArray.Create((DiagnosticAnalyzer)analyzer));
+            var allDiags = compilationWithAnalyzers.GetAnalyzerDiagnosticsAsync().Result;
+
+            var projectTrees = new HashSet<SyntaxTree>(projA.Documents.Select(r => r.GetSyntaxTreeAsync().Result));
+            var diags = allDiags
+                .Where(d => d.Location == Location.None || d.Location.IsInMetadata || projectTrees.Contains(d.Location.SourceTree))
+                .OrderBy(d => d.Location.SourceSpan.Start).ToArray();
+
+            var diag = analyzer.SupportedDiagnostics[0];
+            var expected = new DiagnosticResult[] {
+                diag.CreateDiagnosticResult(8, 6, "PublicClass"),
+                diag.CreateDiagnosticResult(11, 10, "PublicField"),
+                diag.CreateDiagnosticResult(14, 10, "PublicProperty"),
+                diag.CreateDiagnosticResult(20, 10, "PublicMethod"),
+                diag.CreateDiagnosticResult(26, 10, "PublicDelegate"),
+                diag.CreateDiagnosticResult(29, 10, "PublicClass"),
+                diag.CreateDiagnosticResult(35, 6, "PublicStruct"),
+                diag.CreateDiagnosticResult(40, 6, "PublicEnum"),
+                diag.CreateDiagnosticResult(47, 6, "PublicInterface")
+            };
+
+            VerifyDiagnosticResults(diags, analyzer, expected);
+        }
+    }
+}
+
--- a/test/Microsoft.ML.CodeAnalyzer.Tests/Code/BestFriendOnPublicDeclarationTest.cs
+++ b/test/Microsoft.ML.CodeAnalyzer.Tests/Code/BestFriendOnPublicDeclarationTest.cs
@@ -52,7 +52,8 @@ namespace Microsoft.ML.InternalCodeAnalyzer.Tests
                 diag.CreateDiagnosticResult(29, 10, "PublicClass"),
                 diag.CreateDiagnosticResult(35, 6, "PublicStruct"),
                 diag.CreateDiagnosticResult(40, 6, "PublicEnum"),
-                diag.CreateDiagnosticResult(47, 6, "PublicInterface")
+                diag.CreateDiagnosticResult(47, 6, "PublicInterface"),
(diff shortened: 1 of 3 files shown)


tools-local/Microsoft.ML.InternalCodeAnalyzer/BestFriendOnPublicDeclarationsAnalyzer.cs
```diff
@@ -0,0 +1,102 @@
...
+    public sealed class BestFriendOnPublicDeclarationsAnalyzer : DiagnosticAnalyzer
+    {
+        private const string Category = "Access";
+        internal const string DiagnosticId = "MSML_BestFriendOnPublicDeclaration";
+
+        private const string Title = "Public declarations should not have " + AttributeName + " attribute.";
+        private const string Format = "The " + AttributeName + " should not be applied to publicly visible members.";
+
+        private const string Description =
+            "The " + AttributeName + " attribute is only valid on internal identifiers.";
```
Ivanidzo4ka: >valid on internal identifiers. [](start = 57, length = 30)

Well, technically this rule is only about public declaration, so no one will stop evil me to put `[BestFriend]`on private declarations, and this description wouldn't be right.
I'm not saying you should verify private declarations, it's just description and Title slightly out of sync :)

test/Microsoft.ML.CodeAnalyzer.Tests/Code/BestFriendOnPublicDeclarationTest.cs
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Microsoft.ML.InternalCodeAnalyzer.Tests
+{
+    public sealed class BestFriendOnPublicDeclarationTest : DiagnosticVerifier<BestFriendOnPublicDeclarationsAnalyzer>
+    {
+        private readonly Lazy<string> SourceAttribute = TestUtils.LazySource("BestFriendAttribute.cs");
+        private readonly Lazy<string> SourceDeclaration = TestUtils.LazySource("BestFriendOnPublicDeclaration.cs");
+
+        [Fact]
+        public void BestFriendOnPublicDeclaration()
```
sharwell: 💡 We should really get this switched over to Microsoft.CodeAnalysis.Testing

tools-local/Microsoft.ML.InternalCodeAnalyzer/BestFriendOnPublicDeclarationsAnalyzer.cs
```diff
@@ -0,0 +1,102 @@
...
+                DiagnosticSeverity.Warning, isEnabledByDefault: true, description: Description);
+
+        private const string AttributeName = "Microsoft.ML.BestFriendAttribute";
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics =>
+            ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterSemanticModelAction(Analyze);
```
sharwell: ❗️ *Needs* to call `EnableConcurrentExecution`

💡 *Should* call `ConfigureGeneratedCodeAnalysis`

sharwell: ❕ For efficiency purposes, this should be a *symbol* analyzer, not a *semantic model* analyzer

tools-local/Microsoft.ML.InternalCodeAnalyzer/BestFriendOnPublicDeclarationsAnalyzer.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        private void AnalyzeCore(SemanticModelAnalysisContext context, string attributeName)
+        {
+            var model = context.SemanticModel;
+            var comp = model.Compilation;
+
+            // Get the symbols of the key types we are analyzing. If we can't find it
+            // there is no point in going further.
+            var bestFriendAttributeType = comp.GetTypeByMetadataName(attributeName);
```
sharwell: 💡 Ideally this will be called once in a compilation start action and used directly in the symbol analyzer

sfilipi:
@mareklinka thanks for taking care of this. 
Have you accepted the license?  I see the license/cla in pending state.

mareklinka:
> Have you accepted the license? I see the license/cla in pending state.

I definitely should have it signed. I contributed to corefx as well some time ago so it should be sorted out. I also double checked yesterday (I saw the pending check as well) and the .NET foundation CLA page says I signed it. Not sure what's going on.

mareklinka:
> Have you accepted the license? I see the license/cla in pending state.

Just pushed new changes and now it reports correctly. Weird.

tools-local/Microsoft.ML.InternalCodeAnalyzer/BestFriendOnPublicDeclarationsAnalyzer.cs
mareklinka: Fixed.

mareklinka: Fixed.

test/Microsoft.ML.CodeAnalyzer.Tests/Code/BestFriendOnPublicDeclarationTest.cs
mareklinka: I feel like the whole analyzer test project could benefit from this. However, this extends beyond the scope of the original issue - could we track this as a separate issue?

sharwell: > could we track this as a separate issue?

Yes, of course 😄

tools-local/Microsoft.ML.InternalCodeAnalyzer/BestFriendOnPublicDeclarationsAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+                if (attribute == null)
+                    continue;
+                
+                context.RegisterSymbolAction(c => AnalyzeCore(c, attribute), SymbolKind.NamedType, SymbolKind.Method, SymbolKind.Field, SymbolKind.Property);
+            }
+        }
+
+        private void AnalyzeCore(SymbolAnalysisContext context, INamedTypeSymbol attributeType)
+        {
+            if (context.Symbol.DeclaredAccessibility != Accessibility.Public)
```
sharwell: 📝 This will report a diagnostic on something like the following. You'll want to review and determine if this is desirable.

```csharp
internal class NotPublicType
{
  [BestFriend]
  public void DeclaredPublicButNotExposed() { }
}
```

mareklinka: That is a good point. The effective visibility is `internal` here but a declaration like this is still not "correct". How do you imagine we handle this, @TomFinley?

We could either just keep reporting this as incorrect (simple 😄) or recursively compute the effective visibility. That would require us to walk all the way to the assembly level for each symbol. That might get expensive.

test/Microsoft.ML.CodeAnalyzer.Tests/Code/BestFriendOnPublicDeclarationTest.cs
mareklinka: Tracked as #2480.

tools-local/Microsoft.ML.InternalCodeAnalyzer/BestFriendOnPublicDeclarationsAnalyzer.cs
TomFinley: Hi just saw this. If it will report something like this, that is good, not bad. So I'd say @mareklinka's change here is desirable, if it in fact does warn on this.

Indeed, it's unclear to me what the above snippet would do? My best guess is, the thinking that references to `NotPublicType` should fail, but references to `NotPublicType.DeclaredPublicButNotExposed` should succeed? If my guess as to what the point is is correct, that is against the spirit of #1519 which is about making it obvious when one of these "cross-assembly-infrastructure" types and members were being declared. So the snippet's usage of `[BestFriend]` is functionally not useful (since it serves no purpose), which means that I'd think it should warn.

PR #2435, opened 2019-02-06
Typed Calibrated Predictors

### LINKED ISSUE #1307: Binary fast tree doesn't return tree predictor in pigsty
https://github.com/dotnet/machinelearning/blob/a285f8d5db15cd8bf77f3c03763bce16642f902b/src/Microsoft.ML.FastTree/FastTreeStatic.cs#L93
It returns IPredictorWithFeatureWeights<float> which I found weird since I expect tree predictor from fast tree.
Regression and Multiclass returns FastTreeRegressionPredictor and FastTreeRankingPredictor which is right.

### LINKED ISSUE #2378: State of CalibratorPredictorBase v1

wschin:
There are some [problems ](https://github.com/dotnet/machinelearning/issues/2378)around weekly-typed calibrated predictor, if we want to expose those trained calibrated predictor to users. To expose trained predictor in a type-safe manner, this PR adds two type parameters to `CalibratedPredictorBase`; that is,
```csharp
    /// <summary>
    /// <see cref="IWeeklyTypedCalibratedPredictor"/> provides a weekly-typed way to access strongly-typed
    /// <see cref="CalibratedPredictorBase{TSubPredictor, TCalibrator}.SubModelParameters"/> and
    /// <see cref="CalibratedPredictorBase{TSubPredictor, TCalibrator}.Calibrator"/>.
    /// <see cref="IWeeklyTypedCalibratedPredictor"/> is commonly used in weekly-typed expressions. The
    /// existence of this interface is just for supporting existing codebase, so we discourage its uses.
    /// </summary>
    [BestFriend]
    internal interface IWeeklyTypedCalibratedPredictor
    {
        IPredictorProducing<float> WeeklyTypedSubModelParameters { get; }
        ICalibrator WeeklyTypedCalibrator { get; }
    }

    public abstract class CalibratedPredictorBase<TSubModelParameters, TCalibrator> :
        IDistPredictorProducing<float, float>,
        ICanSaveInIniFormat,
        ICanSaveInTextFormat,
        ICanSaveInSourceCode,
        ICanSaveSummary,
        ICanGetSummaryInKeyValuePairs,
        IWeeklyTypedCalibratedPredictor
        where TSubModelParameters : class, IPredictorProducing<float>
        where TCalibrator : class, ICalibrator
    {
        ...
        // Strongly-typed members.
        public TSubModelParameters SubModelParameters { get; }
        public TCalibrator Calibrator { get; }

        // Type-unsafed accessors of strongly-typed members.
        IPredictorProducing<float> IWeeklyTypedCalibratedPredictor.WeeklyTypedSubModelParameters => SubModelParameters;
        ICalibrator IWeeklyTypedCalibratedPredictor.WeeklyTypedCalibrator => Calibrator
        ...
    }
```
and make changes making the code complied. This PR doesn't make every use of `CalibratedPredictorBase` type-safe, because there are some functionalities built upon C# interface.

Fix #2378, Fix #1307.

### DIFF
--- a/src/Microsoft.ML.Data/Prediction/Calibrator.cs
+++ b/src/Microsoft.ML.Data/Prediction/Calibrator.cs
@@ -116,20 +116,41 @@ namespace Microsoft.ML.Internal.Calibration
         IPredictor Calibrate(IChannel ch, IDataView data, ICalibratorTrainer caliTrainer, int maxRows);
     }
 
-    public abstract class CalibratedPredictorBase<TSubPredictor,TCalibrator> :
+    /// <summary>
+    /// <see cref="IHasWeeklyTypedCalibratedPredictor"/> provides a weekly-typed way to access strongly-typed
+    /// <see cref="CalibratedPredictorBase{TSubPredictor, TCalibrator}.SubPredictor"/> and
+    /// <see cref="CalibratedPredictorBase{TSubPredictor, TCalibrator}.Calibrator"/>.
+    /// <see cref="IHasWeeklyTypedCalibratedPredictor"/> is commonly used in weekly-typed expressions. The
+    /// existence of this interface is just for supporting existing codebase, so we discourage its uses.
+    /// </summary>
+    [BestFriend]
+    internal interface IHasWeeklyTypedCalibratedPredictor
+    {
+        IPredictorProducing<float> WeeklyTypedSubPredictor { get; }
+        ICalibrator WeeklyTypedCalibrator { get; }
+    }
+
+    public abstract class CalibratedPredictorBase<TSubPredictor, TCalibrator> :
         IDistPredictorProducing<float, float>,
         ICanSaveInIniFormat,
         ICanSaveInTextFormat,
         ICanSaveInSourceCode,
         ICanSaveSummary,
-        ICanGetSummaryInKeyValuePairs
-        where TSubPredictor :  class, IPredictorProducing<float>
+        ICanGetSummaryInKeyValuePairs,
+        IHasWeeklyTypedCalibratedPredictor
+        where TSubPredictor : class, IPredictorProducing<float>
         where TCalibrator : class, ICalibrator
     {
         protected readonly IHost Host;
 
+        // Strongly-typed members.
         public TSubPredictor SubPredictor { get; }
         public TCalibrator Calibrator { get; }
+
+        // Type-unsafed accessors of strongly-typed members.
+        IPredictorProducing<float> IHasWeeklyTypedCalibratedPredictor.WeeklyTypedSubPredictor => SubPredictor;
+        ICalibrator IHasWeeklyTypedCalibratedPredictor.WeeklyTypedCalibrator => Calibrator;
+
         public PredictionKind PredictionKind => SubPredictor.PredictionKind;
 
         private protected CalibratedPredictorBase(IHostEnvironment env, string name, TSubPredictor predictor, TCalibrator calibrator)
@@ -140,6 +161,13 @@ namespace Microsoft.ML.Internal.Calibration
             Host.CheckValue(predictor, nameof(predictor));
             Host.CheckValue(calibrator, nameof(calibrator));
 
+            SubPredictor = predictor;
+            Calibrator = calibrator;
+        }
+
+        private CalibratedPredictorBase(IHost host, TSubPredictor predictor, TCalibrator calibrator)
+        {
+            Host = host;
             SubPredictor = predictor;
             Calibrator = calibrator;
         }
--- a/src/Microsoft.ML.Data/Prediction/Calibrator.cs
+++ b/src/Microsoft.ML.Data/Prediction/Calibrator.cs
@@ -58,20 +58,20 @@ using Newtonsoft.Json.Linq;
     "Naive Calibration Executor",
     NaiveCalibrator.LoaderSignature)]
 
-[assembly: LoadableClass(typeof(CalibratedPredictor<IPredictorProducing<float>,ICalibrator>), null, typeof(SignatureLoadModel),
+[assembly: LoadableClass(typeof(ValueMapperCalibratedModelParameters<IPredictorProducing<float>, ICalibrator>), null, typeof(SignatureLoadModel),
     "Calibrated Predictor Executor",
-    CalibratedPredictor<IPredictorProducing<float>,ICalibrator>.LoaderSignature, "BulkCaliPredExec")]
+    ValueMapperCalibratedModelParameters<IPredictorProducing<float>, ICalibrator>.LoaderSignature, "BulkCaliPredExec")]
 
-[assembly: LoadableClass(typeof(FeatureWeightsCalibratedPredictor<IPredictorWithFeatureWeights<float>,ICalibrator>), null, typeof(SignatureLoadModel),
+[assembly: LoadableClass(typeof(FeatureWeightsCalibratedModelParameters<IPredictorWithFeatureWeights<float>, ICalibrator>), null, typeof(SignatureLoadModel),
     "Feature Weights Calibrated Predictor Executor",
-    FeatureWeightsCalibratedPredictor<IPredictorWithFeatureWeights<float>,ICalibrator>.LoaderSignature)]
+    FeatureWeightsCalibratedModelParameters<IPredictorWithFeatureWeights<float>, ICalibrator>.LoaderSignature)]
 
-[assembly: LoadableClass(typeof(ParameterMixingCalibratedPredictor<IPredictorWithFeatureWeights<float>,ICalibrator>), null, typeof(SignatureLoadModel),
+[assembly: LoadableClass(typeof(ParameterMixingCalibratedModelParameters<IPredictorWithFeatureWeights<float>, ICalibrator>), null, typeof(SignatureLoadModel),
     "Parameter Mixing Calibrated Predictor Executor",
-    ParameterMixingCalibratedPredictor<IPredictorWithFeatureWeights<float>,ICalibrator>.LoaderSignature)]
+    ParameterMixingCalibratedModelParameters<IPredictorWithFeatureWeights<float>, ICalibrator>.LoaderSignature)]
 
-[assembly: LoadableClass(typeof(SchemaBindableCalibratedPredictor<IPredictorProducing<float>,ICalibrator>), null, typeof(SignatureLoadModel),
-    "Schema Bindable Calibrated Predictor", SchemaBindableCalibratedPredictor<IPredictorProducing<float>,ICalibrator>.LoaderSignature)]
+[assembly: LoadableClass(typeof(SchemaBindableCalibratedModelParameters<IPredictorProducing<float>, ICalibrator>), null, typeof(SignatureLoadModel),
+    "Schema Bindable Calibrated Predictor", SchemaBindableCalibratedModelParameters<IPredictorProducing<float>, ICalibrator>.LoaderSignature)]
 
 [assembly: LoadableClass(typeof(void), typeof(Calibrate), null, typeof(SignatureEntryPointModule), "Calibrate")]
 
@@ -117,43 +117,57 @@ namespace Microsoft.ML.Internal.Calibration
     }
 
     /// <summary>
-    /// <see cref="IWeaklyTypedCalibratedPredictor"/> provides a weekly-typed way to access strongly-typed
-    /// <see cref="CalibratedPredictorBase{TSubPredictor, TCalibrator}.SubModelParameters"/> and
-    /// <see cref="CalibratedPredictorBase{TSubPredictor, TCalibrator}.Calibrator"/>.
-    /// <see cref="IWeaklyTypedCalibratedPredictor"/> is commonly used in weekly-typed expressions. The
+    /// <see cref="IWeaklyTypedCalibratedModelParameters"/> provides a weekly-typed way to access strongly-typed
+    /// <see cref="CalibratedModelParametersBase{TSubPredictor, TCalibrator}.SubModel"/> and
+    /// <see cref="CalibratedModelParametersBase{TSubPredictor, TCalibrator}.Calibrator"/>.
+    /// <see cref="IWeaklyTypedCalibratedModelParameters"/> is commonly used in weekly-typed expressions. The
     /// existence of this interface is just for supporting existing codebase, so we discourage its uses.
     /// </summary>
     [BestFriend]
-    internal interface IWeaklyTypedCalibratedPredictor
+    internal interface IWeaklyTypedCalibratedModelParameters
     {
-        IPredictorProducing<float> WeeklyTypedSubModelParameters { get; }
+        IPredictorProducing<float> WeeklyTypedSubModel { get; }
         ICalibrator WeeklyTypedCalibrator { get; }
     }
 
-    public abstract class CalibratedPredictorBase<TSubModelParameters, TCalibrator> :
+    /// <summary>
+    /// Class for allowing a post-processing step, defined by <see cref="Calibrator"/>, to <see cref="SubModel"/>'s
+    /// output. For example, in binary classification, <see cref="Calibrator"/> can convert support vector machine's
+    /// output value to the probability of belonging to the positive (or negative) class. Detailed math materials
+    /// can be found at <a href="https://www.csie.ntu.edu.tw/~cjlin/papers/plattprob.pdf">this paper</a>.
+    /// </summary>
+    /// <typeparam name="TSubModel">Type being calibrated.</typeparam>
+    /// <typeparam name="TCalibrator">Type used to calibrate.</typeparam>
+    public abstract class CalibratedModelParametersBase<TSubModel, TCalibrator> :
         IDistPredictorProducing<float, float>,
         ICanSaveInIniFormat,
         ICanSaveInTextFormat,
         ICanSaveInSourceCode,
         ICanSaveSummary,
         ICanGetSummaryInKeyValuePairs,
-        IWeaklyTypedCalibratedPredictor
-        where TSubModelParameters : class, IPredictorProducing<float>
+        IWeaklyTypedCalibratedModelParameters
+        where TSubModel : class, IPredictorProducing<float>
         where TCalibrator : class, ICalibrator
     {
         protected readonly IHost Host;
 
         // Strongly-typed members.
-        public TSubModelParameters SubModelParameters { get; }
+        /// <summary>
+        /// <see cref="SubModel"/>'s output would calibrated by <see cref="Calibrator"/>.
+        /// </summary>
+        public TSubModel SubModel { get; }
+        /// <summary>
+        /// <see cref="Calibrator"/> is used to post-process score produced by <see cref="SubModel"/>.
+        /// </summary>
         public TCalibrator Calibrator { get; }
 
         // Type-unsafed accessors of strongly-typed members.
-        IPredictorProducing<float> IWeaklyTypedCalibratedPredictor.WeeklyTypedSubModelParameters => SubModelParameters;
-        ICalibrator IWeaklyTypedCalibratedPredictor.WeeklyTypedCalibrator => Calibrator;
+        IPredictorProducing<float> IWeaklyTypedCalibratedModelParameters.WeeklyTypedSubModel => SubModel;
+        ICalibrator IWeaklyTypedCalibratedModelParameters.WeeklyTypedCalibrator => Calibrator;
 
-        public PredictionKind PredictionKind => SubModelParameters.PredictionKind;
+        public PredictionKind PredictionKind => SubModel.PredictionKind;
 
-        private protected CalibratedPredictorBase(IHostEnvironment env, string name, TSubModelParameters predictor, TCalibrator calibrator)
+        private protected CalibratedModelParametersBase(IHostEnvironment env, string name, TSubModel predictor, TCalibrator calibrator)
         {
             Contracts.CheckValue(env, nameof(env));
             env.CheckNonWhiteSpace(name, nameof(name));
@@ -161,21 +175,21 @@ namespace Microsoft.ML.Internal.Calibration
             Host.CheckValue(predictor, nameof(predictor));
             Host.CheckValue(calibrator, nameof(calibrator));
 
-            SubModelParameters = predictor;
+            SubModel = predictor;
             Calibrator = calibrator;
         }
 
         void ICanSaveInIniFormat.SaveAsIni(TextWriter writer, RoleMappedSchema schema, ICalibrator calibrator)
         {
             Host.Check(calibrator == null, "Too many calibrators.");
-            var saver = SubModelParameters as ICanSaveInIniFormat;
+            var saver = SubModel as ICanSaveInIniFormat;
             saver?.SaveAsIni(writer, schema, Calibrator);
         }
 
         void ICanSaveInTextFormat.SaveAsText(TextWriter writer, RoleMappedSchema schema)
         {
             // REVIEW: What about the calibrator?
-            var saver = SubModelParameters as ICanSaveInTextFormat;
+            var saver = SubModel as ICanSaveInTextFormat;
             if (saver != null)
                 saver.SaveAsText(writer, schema);
         }
@@ -183,7 +197,7 @@ namespace Microsoft.ML.Internal.Calibration
         void ICanSaveInSourceCode.SaveAsCode(TextWriter writer, RoleMappedSchema schema)
         {
             // REVIEW: What about the calibrator?
-            var saver = SubModelParameters as ICanSaveInSourceCode;
+            var saver = SubModel as ICanSaveInSourceCode;
             if (saver != null)
                 saver.SaveAsCode(writer, schema);
         }
@@ -191,7 +205,7 @@ namespace Microsoft.ML.Internal.Calibration
         void ICanSaveSummary.SaveSummary(TextWriter writer, RoleMappedSchema schema)
         {
             // REVIEW: What about the calibrator?
-            var saver = SubModelParameters as ICanSaveSummary;
+            var saver = SubModel as ICanSaveSummary;
             if (saver != null)
                 saver.SaveSummary(writer, schema);
         }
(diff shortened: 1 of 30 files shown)


src/Microsoft.ML.StandardLearners/Standard/Online/OnlineLinear.cs
```diff
@@ -253,7 +253,11 @@ private protected sealed override TModel TrainModelCore(TrainContext context)
         {
             Host.CheckValue(context, nameof(context));
             var initPredictor = context.InitialPredictor;
-            var initLinearPred = initPredictor as LinearModelParameters ?? (initPredictor as CalibratedPredictorBase)?.SubPredictor as LinearModelParameters;
+            LinearModelParameters initLinearPred = null;
+            if (initPredictor is LinearModelParameters)
+                initLinearPred = (LinearModelParameters)initPredictor;
+            else if (initPredictor is CalibratedPredictorBase<LinearBinaryModelParameters, PlattCalibrator>)
+                initLinearPred = ((CalibratedPredictorBase<LinearBinaryModelParameters, PlattCalibrator>)initPredictor).SubPredictor;
             Host.CheckParam(initPredictor == null || initLinearPred != null, nameof(context), "Not a linear predictor.");
```
TomFinley: >Host.CheckParam(initPredictor == null || initLinearPred != null, nameof(context), "Not a linear predictor."); [](start = 12, length = 109)

This is effectively repeating the test already in the prior if statements. So this can be improved by just using `else throw Host.ExceptParam(...)` and so on.

src/Microsoft.ML.StandardLearners/Standard/Online/OnlineLinear.cs
```diff
@@ -253,7 +253,11 @@ private protected sealed override TModel TrainModelCore(TrainContext context)
         {
             Host.CheckValue(context, nameof(context));
             var initPredictor = context.InitialPredictor;
-            var initLinearPred = initPredictor as LinearModelParameters ?? (initPredictor as CalibratedPredictorBase)?.SubPredictor as LinearModelParameters;
+            LinearModelParameters initLinearPred = null;
+            if (initPredictor is LinearModelParameters)
+                initLinearPred = (LinearModelParameters)initPredictor;
```
TomFinley: Can be simplfiied to

```csharp
if (initPredictor is LinearModelParameters linModel)
    initLinearPred = linModel;
```

Same with code below.

src/Microsoft.ML.FastTree/GamClassification.cs
```diff
@@ -31,7 +31,9 @@
 namespace Microsoft.ML.Trainers.FastTree
 {
     public sealed class BinaryClassificationGamTrainer :
-    GamTrainerBase<BinaryClassificationGamTrainer.Options, BinaryPredictionTransformer<CalibratedPredictorBase>, CalibratedPredictorBase>
+        GamTrainerBase<BinaryClassificationGamTrainer.Options,
+        BinaryPredictionTransformer<CalibratedPredictorBase<BinaryClassificationGamModelParameters,PlattCalibrator>>,
+        CalibratedPredictorBase<BinaryClassificationGamModelParameters,PlattCalibrator>>
```
TomFinley: >,P [](start = 70, length = 2)

Under default settings, visual studio's auto-formatting is I believe control-k-d.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -116,21 +116,44 @@ internal interface ISelfCalibratingPredictor
...
-    public abstract class CalibratedPredictorBase :
+    /// <summary>
+    /// <see cref="IHasWeeklyTypedCalibratedPredictor"/> provides a weekly-typed way to access strongly-typed
+    /// <see cref="CalibratedPredictorBase{TSubPredictor, TCalibrator}.SubPredictor"/> and
+    /// <see cref="CalibratedPredictorBase{TSubPredictor, TCalibrator}.Calibrator"/>.
+    /// <see cref="IHasWeeklyTypedCalibratedPredictor"/> is commonly used in weekly-typed expressions. The
+    /// existence of this interface is just for supporting existing codebase, so we discourage its uses.
+    /// </summary>
+    [BestFriend]
+    internal interface IHasWeeklyTypedCalibratedPredictor
```
TomFinley: >IHasWeeklyTypedCalibratedPredictor [](start = 23, length = 34)

Clever.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -190,23 +220,27 @@ private protected void SaveCore(ModelSaveContext ctx)
...
-            ICalibrator calibrator;
-            ctx.LoadModel<ICalibrator, SignatureLoadModel>(env, out calibrator, @"Calibrator");
+            TCalibrator calibrator;
+            ctx.LoadModel<TCalibrator, SignatureLoadModel>(env, out calibrator, @"Calibrator");
             return calibrator;
         }
     }
 
-    public abstract class ValueMapperCalibratedPredictorBase : CalibratedPredictorBase, IValueMapperDist, IFeatureContributionMapper, ICalculateFeatureContribution,
+    public abstract class ValueMapperCalibratedPredictorBase<TSubPredictor,TCalibrator> :
```
TomFinley: >ValueMapperCalibratedPredictorBase [](start = 26, length = 34)

Is there any reason why this should be exposed as a public type? More broadly, is there any reason why *user* code should ever need to know that the model is anything more specific than what we currently call a `CalibratedPredictorBase`?

It just strikes me that since we have made `ValueMapper

src/Microsoft.ML.Data/Prediction/Calibrator.cs
TomFinley: Though, I think probably you *could* have gotten away with just making a couple `[BestFriend] internal` fields on the calibrated predictor base.

---
In reply to: [254540575](https://github.com/dotnet/machinelearning/pull/2435#discussion_r254540575) [](ancestors = 254540575)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -116,21 +116,44 @@ internal interface ISelfCalibratingPredictor
...
+        IHasWeeklyTypedCalibratedPredictor
+        where TSubPredictor : class, IPredictorProducing<float>
+        where TCalibrator : class, ICalibrator
     {
         protected readonly IHost Host;
 
-        public IPredictorProducing<float> SubPredictor { get; }
-        public ICalibrator Calibrator { get; }
+        // Strongly-typed members.
+        public TSubPredictor SubPredictor { get; }
```
TomFinley: >SubPredictor [](start = 29, length = 12)

Note that we will not want this name in future, since we have deprecated the phrase "predictor" (since this style of object does not predict and has not for years). I do not suggest changing the name in this PR necessary.

TomFinley:
This looks pretty good @wschin ... I have only really two high level observations:

1. can we get away with just one public class (with all implementing classes being only different in ways that matter only to component authors),  and also

2. the naming of things being "predictors" relies on now deprecated terminology. I might consult @sfilipi and @rogancarr on preferred names for this sort of thing. (Model parameters seems a bit verbose, maybe they have better ideas.)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
wschin: I am not sure if we can do that. There are some places where we need to enumerate ALL possible combinations of `TSubPredictor` and `TCalibrator` in `CalibratedPredictorBase` to cast an object to the desired type; for example, in [SummarizePredictor](https://github.com/dotnet/machinelearning/blob/3dd3a1ebaff92dba34c2bb2ecd833c97beae1a23/src/Microsoft.ML.Data/EntryPoints/SummarizePredictor.cs#L51) we have
```csharp
        internal static IDataView GetSummaryAndStats(IHostEnvironment env, IPredictor predictor, RoleMappedSchema schema, out IDataView stats)
        {
            var calibrated = predictor as CalibratedPredictorBase;
            while (calibrated != null)
            {
                predictor = calibrated.SubPredictor;
                calibrated = predictor as CalibratedPredictorBase;
            }
            ...
       }
```
Removing `IHasWeeklyTypedCalibratedPredictor` may force us to extend that function significantly.

---
In reply to: [254541286](https://github.com/dotnet/machinelearning/pull/2435#discussion_r254541286) [](ancestors = 254541286,254540575)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
wschin: Sounds good I will do the following replacements.
`SubPredictor` ---> `SubModelParameters` 
`TSubPredictor` ---> `TSubModelParameters`

---
In reply to: [254541699](https://github.com/dotnet/machinelearning/pull/2435#discussion_r254541699) [](ancestors = 254541699)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
wschin: Let me make it internal.

---
In reply to: [254540984](https://github.com/dotnet/machinelearning/pull/2435#discussion_r254540984) [](ancestors = 254540984)

src/Microsoft.ML.FastTree/GamClassification.cs
wschin: Right. Thanks!

---
In reply to: [254540313](https://github.com/dotnet/machinelearning/pull/2435#discussion_r254540313) [](ancestors = 254540313)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
TomFinley: OK, that's fine, just throwing it out there. Thanks @wschin !

src/Microsoft.ML.StandardLearners/Standard/Online/OnlineLinear.cs
wschin: Sure. Many thanks for your education.

---
In reply to: [254540256](https://github.com/dotnet/machinelearning/pull/2435#discussion_r254540256) [](ancestors = 254540256)

src/Microsoft.ML.StandardLearners/Standard/Online/OnlineLinear.cs
```diff
@@ -253,10 +253,19 @@ private protected sealed override TModel TrainModelCore(TrainContext context)
...
-            var data = context.TrainingSet;
 
+            if (initPredictor is LinearModelParameters initLinearPred)
+                initLinearPred = (LinearModelParameters)initPredictor;
+            else if (initPredictor is CalibratedPredictorBase<LinearBinaryModelParameters, PlattCalibrator> calibrated)
+                initLinearPred = calibrated.SubModelParameters;
+            else
+            {
+                initLinearPred = null;
+                Host.CheckParam(initPredictor == null || initLinearPred != null, nameof(context),
```
TomFinley: >Host.CheckParam(initPredictor == null || initLinearPred != null, [](start = 16, length = 64)

What I think I meant though was, you can just nuke this test and replace it with a throw exception, since if the two tests above *fail*, then these two will definitely be null. So there is no point in having this be a check.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -116,59 +116,82 @@ internal interface ISelfCalibratingPredictor
...
-    public abstract class CalibratedPredictorBase :
+    /// <summary>
+    /// <see cref="IWeeklyTypedCalibratedPredictor"/> provides a weekly-typed way to access strongly-typed
+    /// <see cref="CalibratedPredictorBase{TSubPredictor, TCalibrator}.SubModelParameters"/> and
+    /// <see cref="CalibratedPredictorBase{TSubPredictor, TCalibrator}.Calibrator"/>.
+    /// <see cref="IWeeklyTypedCalibratedPredictor"/> is commonly used in weekly-typed expressions. The
+    /// existence of this interface is just for supporting existing codebase, so we discourage its uses.
+    /// </summary>
+    [BestFriend]
+    internal interface IWeeklyTypedCalibratedPredictor
```
artidoro: >IWeeklyTypedCalibratedPredictor [](start = 23, length = 31)

I think there is a typo in the naming, should be "weak" if I am not wrong.

TomFinley: Ah ha ha... yes somehow I missed this. Funny how the brain just elides over such things as this.

---
In reply to: [254937081](https://github.com/dotnet/machinelearning/pull/2435#discussion_r254937081) [](ancestors = 254937081)

wschin: Fixed....

src/Microsoft.ML.StandardLearners/Standard/Online/OnlineLinear.cs
wschin: We can't drop it because, when initial predictor is not provided, we will always throw.

---
In reply to: [254933689](https://github.com/dotnet/machinelearning/pull/2435#discussion_r254933689) [](ancestors = 254933689)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
TomFinley: Maybe. However, one thing that occurred to me is that in the other prominent place where we expose the model parameters, we don't call them `ModelParameters` but just `Model`. I might elect to be consistent with that usage there. (So, `SubModel` for the property name, and `TSubModel` for the predictor.)

https://github.com/dotnet/machinelearning/blob/834e4715afa070f5e5c3628693682943b3f0bd6d/src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs#L43

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -186,27 +209,31 @@ void ICanSaveSummary.SaveSummary(TextWriter writer, RoleMappedSchema schema)
...
-            ICalibrator calibrator;
-            ctx.LoadModel<ICalibrator, SignatureLoadModel>(env, out calibrator, @"Calibrator");
+            TCalibrator calibrator;
+            ctx.LoadModel<TCalibrator, SignatureLoadModel>(env, out calibrator, @"Calibrator");
             return calibrator;
         }
     }
 
-    public abstract class ValueMapperCalibratedPredictorBase : CalibratedPredictorBase, IValueMapperDist, IFeatureContributionMapper, ICalculateFeatureContribution,
+    internal abstract class ValueMapperCalibratedPredictorBase<TSubPredictor,TCalibrator> :
```
artidoro: >TSubPredictor [](start = 63, length = 13)

Could you rename to `TSubModelParameters`, to be more consistent with the above?

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -116,59 +116,82 @@ internal interface ISelfCalibratingPredictor
...
+    /// existence of this interface is just for supporting existing codebase, so we discourage its uses.
+    /// </summary>
+    [BestFriend]
+    internal interface IWeaklyTypedCalibratedPredictor
+    {
+        IPredictorProducing<float> WeeklyTypedSubModelParameters { get; }
+        ICalibrator WeeklyTypedCalibrator { get; }
+    }
+
+    public abstract class CalibratedPredictorBase<TSubModelParameters, TCalibrator> :
```
artidoro: >CalibratedPredictorBase [](start = 26, length = 23)

@tfinley@gmail.com since we transitioned to a place where Predictors are named ModelParameters, should we rename this class `CalibratedModelParametersBase`?

wschin: Sounds good. I am changing it.

---
In reply to: [255160688](https://github.com/dotnet/machinelearning/pull/2435#discussion_r255160688) [](ancestors = 255160688)

artidoro: Potentially the same should apply to the derived classes.

---
In reply to: [255160688](https://github.com/dotnet/machinelearning/pull/2435#discussion_r255160688) [](ancestors = 255160688)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
wschin: Sure.

---
In reply to: [255159935](https://github.com/dotnet/machinelearning/pull/2435#discussion_r255159935) [](ancestors = 255159935)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
wschin: Ah. Got it. Will scan through my changes to make it.

---
In reply to: [255158805](https://github.com/dotnet/machinelearning/pull/2435#discussion_r255158805) [](ancestors = 255158805)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -313,9 +340,12 @@ bool ISingleCanSaveOnnx.SaveAsOnnx(OnnxContext ctx, string[] outputNames, string
     }
 
     [BestFriend]
-    internal sealed class CalibratedPredictor : ValueMapperCalibratedPredictorBase, ICanSaveModel
+    internal sealed class CalibratedPredictor<TSubPredictor,TCalibrator> :
```
artidoro: >TSubPredictor [](start = 46, length = 13)

Same here please.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
artidoro: I was just thinking about the same thing looking at the test code! I think it's a very good idea.

---
In reply to: [255162189](https://github.com/dotnet/machinelearning/pull/2435#discussion_r255162189) [](ancestors = 255162189,255158805)

test/Microsoft.ML.Tests/TrainerEstimators/SymSgdClassificationTests.cs
```diff
@@ -35,7 +37,7 @@ public void TestEstimatorSymSgdInitPredictor()
             var initPredictor = ML.BinaryClassification.Trainers.StochasticDualCoordinateAscent().Fit(transformedData);
             var data = initPredictor.Transform(transformedData);
 
-            var withInitPredictor = new SymSgdClassificationTrainer(Env, new SymSgdClassificationTrainer.Options()).Train(transformedData, initialPredictor: initPredictor.Model);
+            var withInitPredictor = new SymSgdClassificationTrainer(Env, new SymSgdClassificationTrainer.Options()).Train(transformedData, initialPredictor: initPredictor.Model as CalibratedPredictorBase<LinearBinaryModelParameters, PlattCalibrator>);
             var outInitData = withInitPredictor.Transform(transformedData);
```
artidoro: Can you split into two lines? It's could be hard to read on a smaller screen (like my laptop right now :) )

wschin: Sure.

---
In reply to: [255170947](https://github.com/dotnet/machinelearning/pull/2435#discussion_r255170947) [](ancestors = 255170947)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
wschin: Sure. I also replace `Predictor` from this class name.

---
In reply to: [255162847](https://github.com/dotnet/machinelearning/pull/2435#discussion_r255162847) [](ancestors = 255162847)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -116,59 +116,96 @@ internal interface ISelfCalibratingPredictor
...
+
+    /// <summary>
+    /// Class for allowing a post-processing step, defined by <see cref="Calibrator"/>, to <see cref="SubModel"/>'s
+    /// output. For example, in binary classification, <see cref="Calibrator"/> can convert support vector machine's
+    /// output value to the probability of belonging to the positive (or negative) class. Detailed math materials
+    /// can be found at <a href="https://www.csie.ntu.edu.tw/~cjlin/papers/plattprob.pdf">this paper</a>.
+    /// </summary>
+    /// <typeparam name="TSubModel">Type being calibrated.</typeparam>
+    /// <typeparam name="TCalibrator">Type used to calibrate.</typeparam>
+    public abstract class CalibratedModelParametersBase<TSubModel, TCalibrator> :
```
Ivanidzo4ka: >TSubModel, TCalibrator [](start = 56, length = 22)

No intent of starting holywar, but don't you find it strange what we have CalibratedModel, and we specify model first and calibrator second?

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -116,59 +116,96 @@ internal interface ISelfCalibratingPredictor
...
+        IPredictorProducing<float> WeeklyTypedSubModel { get; }
+        ICalibrator WeeklyTypedCalibrator { get; }
+    }
+
+    /// <summary>
+    /// Class for allowing a post-processing step, defined by <see cref="Calibrator"/>, to <see cref="SubModel"/>'s
+    /// output. For example, in binary classification, <see cref="Calibrator"/> can convert support vector machine's
+    /// output value to the probability of belonging to the positive (or negative) class. Detailed math materials
+    /// can be found at <a href="https://www.csie.ntu.edu.tw/~cjlin/papers/plattprob.pdf">this paper</a>.
+    /// </summary>
```
Ivanidzo4ka: can you put it in remarks?
I don't see much value in links to pdf in intellisence tooltip.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
wschin: I like this model-first setting. Model is the major part of this class while calibrator just decorates it.

---
In reply to: [255176402](https://github.com/dotnet/machinelearning/pull/2435#discussion_r255176402) [](ancestors = 255176402)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
wschin: Ok.

---
In reply to: [255176861](https://github.com/dotnet/machinelearning/pull/2435#discussion_r255176861) [](ancestors = 255176861)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -116,59 +116,96 @@ internal interface ISelfCalibratingPredictor
...
+        // Strongly-typed members.
+        /// <summary>
+        /// <see cref="SubModel"/>'s output would calibrated by <see cref="Calibrator"/>.
+        /// </summary>
+        public TSubModel SubModel { get; }
+        /// <summary>
+        /// <see cref="Calibrator"/> is used to post-process score produced by <see cref="SubModel"/>.
+        /// </summary>
+        public TCalibrator Calibrator { get; }
+
```
artidoro: Do we ever expect the user to manipulate a Calibrator directly (as in create them through their constructors)?

I noticed that the calibrators (classes implementing `ICalibrator`) are public, and their constructors are public too. But I am yet to find a place where they are instantiated directly.
Should we assume that they should only be accessible through the `CalibratorEstimators`?

artidoro: Along these lines, do we want to expose this here in the `CalibratedModelParametersBase`?

---
In reply to: [255187091](https://github.com/dotnet/machinelearning/pull/2435#discussion_r255187091) [](ancestors = 255187091)

artidoro: Any thoughts @tfinley@gmail.com ?

---
In reply to: [255187736](https://github.com/dotnet/machinelearning/pull/2435#discussion_r255187736) [](ancestors = 255187736,255187091)

TomFinley: This is a worthy question, but maybe we can solve this past this issue, what you think @artidoro ?

artidoro: Sounds good!

---
In reply to: [255197677](https://github.com/dotnet/machinelearning/pull/2435#discussion_r255197677) [](ancestors = 255197677)

TomFinley: If you could file an issue about it @artidoro that might be helpful, I'm not entirely certain I understand the scenario.

Issue #2371, opened 2019-02-01
SelectColumns handles input string arrays differently than other schema transformations


rogancarr:
`SelectColumns` has a different input contract than related schema manipulation transforms. Look at the difference:

```cs
mlContext.Transforms.DropColumns("Column1", "Column2");
mlContext.Transforms.SelectColumns(new string[] { "Column1", "Column2" })
```

rogancarr:
I think this happens because there is an optional parameter that comes after `keepColumns`, which means we can't use it as a `params` parameter.

TomFinley:
Yes. You could potentially introduce an overload:

```csharp
public static ColumnSelectingEstimator SelectColumns(this TransformsCatalog catalog,
            params string[] keepColumns) => catalog.SelectColumns(keepColumns, false);
```

Or something along these lines.

As you've observed, it is due to (1) the presence of multiple parameters and (2) the preference that we like to put column names towards-ish the front, even if there are multiple of them.

PR #2439, opened 2019-02-06
Allowing SelectColumns to take a bare list of columns

rogancarr:
This PR overloads `SelectColumns` so that it can be used with a bare list of columns. The original API is changed to remove `keepHidden`, as this new API has the default behavior.

The PR also cleans up the sample, which had mismatches between the actual and expected outputs.

Fixes: #2371

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
+++ b/src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.Data.DataView;
 using Microsoft.ML.Data;
 using Microsoft.ML.Transforms;
 
@@ -67,9 +68,9 @@ namespace Microsoft.ML
         /// <remarks>
         /// <see cref="DropColumns"/> is commonly used to remove unwanted columns from the schema if the dataset is going to be serialized or
         /// written out to a file. It is not actually necessary to drop unused columns before training or
-        /// performing transforms, as IDataView's lazy evaluation won't actually materialize those columns.
+        /// performing transforms, as <see cref="IDataView"/>'s lazy evaluation won't actually materialize those columns.
         /// In the case of serialization, every column in the schema will be written out. If you have columns
-        /// that you don't want to save, you can use DropColumns to remove them from the schema.
+        /// that you don't want to save, you can use <see cref="DropColumns"/> to remove them from the schema.
         /// </remarks>
         /// <param name="catalog">The transform's catalog.</param>
         /// <param name="columnsToDrop">The array of column names to drop.</param>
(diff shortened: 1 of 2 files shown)


Ivanidzo4ka:
@artidoro Who I believe have same changes in his code.

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
```diff
@@ -107,8 +107,28 @@ public static ColumnSelectingEstimator DropColumns(this TransformsCatalog catalo
...
+        /// <format type="text/markdown">
+        /// <see cref="SelectColumns(TransformsCatalog, string[])"/> operates on the schema of an input IDataView, dropping unselected columns from the schema.
+        /// </format>
+        /// </remarks>
+        /// <param name="catalog">The transform's catalog.</param>
+        /// <param name="keepColumns">The array of column names to keep.</param>
+        /// <example>
+        /// <format type="text/markdown">
+        /// <![CDATA[
+        /// [!code-csharp[SelectColumns](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/SelectColumns.cs)]
```
TomFinley: >[!code-csharp[SelectColumns](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/SelectColumns.cs)] [](start = 12, length = 118)

Hi @rogancarr, just curious, do you have some sort of plugin or something that is helping you edit these without error? VS at least with its builtin facilities is giving me very little help here for my own work, just wondering how you are managing.

rogancarr: There are actually no debugging tools for these links at the moment. Just very careful coping and pasting and editing.

---
In reply to: [254457544](https://github.com/dotnet/machinelearning/pull/2439#discussion_r254457544) [](ancestors = 254457544)

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
```diff
@@ -107,8 +107,28 @@ public static ColumnSelectingEstimator DropColumns(this TransformsCatalog catalo
         /// </example>
         public static ColumnSelectingEstimator SelectColumns(this TransformsCatalog catalog,
             string[] keepColumns,
-            bool keepHidden = ColumnSelectingTransformer.Defaults.KeepHidden)
+            bool keepHidden)
             => new ColumnSelectingEstimator(CatalogUtils.GetEnvironment(catalog),
                 keepColumns, null, keepHidden, ColumnSelectingTransformer.Defaults.IgnoreMissing);
+
+        /// <summary>
+        /// ColumnSelectingEstimator is used to select a list of columns that user wants to keep from a given input.
```
artidoro: Could you use the same comment across the two extensions?
Also I would remove the reference to user, maybe something like: 

```suggestion
        /// Select a list of columns to keep in a given <see cref="IDataView"/>.
```

rogancarr: Yes, good catch!

---
In reply to: [254465617](https://github.com/dotnet/machinelearning/pull/2439#discussion_r254465617) [](ancestors = 254465617)

PR #2441, opened 2019-02-06
Make DataOperations catalog names more precise

### LINKED ISSUE #2440: Add "Row" to the title of ambiguous DataOperations catalog members
Certain operations in the `DataOperations` catalog are as to whether they operate on rows, columns, or slots. I suggest renaming the `FilterBy...`s and `Shuffle` to have `Rows` after the verb (e.g. `FilterRowsBy...`), similar to how we named `SkipRows` and `TakeRows`.

rogancarr:
This PR adds the word "Rows" to names in `DataOperations` catalog where the use may be ambiguous, e.g. between Rows, Columns, and Slots.

Fixes #2440

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/FilterRowsByColumn.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/FilterRowsByColumn.cs
@@ -1,13 +1,13 @@
 ﻿using System;
+using System.Collections.Generic;
 using Microsoft.ML.Data;
-using Microsoft.ML.SamplesUtils;
 
 namespace Microsoft.ML.Samples.Dynamic
 {
     /// <summary>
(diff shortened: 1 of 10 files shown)


docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/FilterRowsByColumn.cs
```diff
@@ -5,9 +5,9 @@
 namespace Microsoft.ML.Samples.Dynamic
 {
     /// <summary>
-    /// Sample class showing how to use <see cref="DataOperationsCatalog.FilterByColumn"/>.
+    /// Sample class showing how to use FilterRowsByColumn.
```
TomFinley: I wonder if `ColumnValue` might be better than just `Column` in this context, but maybe not. Probably fine...

Issue #2465, opened 2019-02-07
Cannot set the threshold on a binary predictor


rogancarr:
It is no longer possible to set a custom `Threshold` and `ThresholdColumn` on a binary classifier.

Previously, we had been using `BinaryPredictionTransformer`. Recently, `BinaryPredictionTransformer` was marked as `internal` and is no longer available for usage outside of the library.

Related to question #403

rogancarr:
One suggestion is to make the `Threshold` and `ThresholdColumn` properties of the prediction transform settable. They are currently only gettable.

Something like this?
```cs
var model = mlContext.BinaryClassification.Trainers.LogisticRegression().Fit(train);
var predictor = model.LastTransformer;
predictor.Threshold = 0.01;
// or
predictor.SetThreshold(0.01);
predictor.SetThresholdColumn("Bananas");
```

Ivanidzo4ka:
https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Tests/Scenarios/Api/Estimators/ReconfigurablePrediction.cs

rogancarr:
@Ivanidzo4ka  I am pulling these out of Tests/ and into a new Functional.Tests that is does not have `InternalsVisibleTo` available in accordance with #2306 .

This test relies on `internal` APIs, so it doesn't work from and end-user perspective.

eerhardt:
YAY! The first value of these tests is now realized.  😄

rogancarr:
For anyone looking for this functionality, there is a workaround discussed in #2645.

eerhardt:
Note that we are doing this exact scenario in our samples:

https://github.com/dotnet/machinelearning-samples/blob/73803ce861e16b48edefaad97771bcd721e268d6/samples/csharp/getting-started/BinaryClassification_SpamDetection/SpamDetectionConsoleApp/Program.cs#L61-L66

```C#
            // The dataset we have is skewed, as there are many more non-spam messages than spam messages.
            // While our model is relatively good at detecting the difference, this skewness leads it to always
            // say the message is not spam. We deal with this by lowering the threshold of the predictor. In reality,
            // it is useful to look at the precision-recall curve to identify the best possible threshold.
            var inPipe = new TransformerChain<ITransformer>(model.Take(model.Count() - 1).ToArray());
[...]

Issue #2420, opened 2019-02-05
SampleDatasetUtils is in disarray


rogancarr:
There is a project called `SamplesUtils` sitting *inside* the core ML.NET codebase. I believe this was created before the `[BestFriend]` concept came around. 

This project contains one `namespace`, `SampleDatasetUtils` that has one `static` `class`, `DatasetUtils`. Here are the following issues:
1. Everything is public and nothing is sealed.
2. This was built for Samples, but now Tests have taken a dependence on a couple of methods.
3. The file itself has no organization, everything is everywhere.

rogancarr:
I suggest the following fixes:

1. Mark everything as `internal` and all non-`static` classes as `sealed`.
2. Move this library class into `Samples` and out of the core codebase.
3. Mark everything appropriate as `BestFriend` and let Tests have access to them. I do think these will also be useful to Tests, and we *do* explicitly test some of our Sample code.

@sfilipi @TomFinley , what do you think about this?

sfilipi:
This class was created and is shipping inside the main Microsoft.ML package, so that the users can copy/paste the API references samples and they will just run. 

#2277 has some similar questions. 

We should definitely seal the classes, but leave public visibility to the methods that download the datasets, and other utilities, because they need to be referenced from outside our own assemblies. 

We don't want to use the [BestFriend] annotation from the samples project. 

We could organize the SamplesUtils file, as it grows.

rogancarr:
Thanks Senja!

I'll start with the move to sealing, but I do share @TomFinley 's concerns from #2277. If we ship this in the main NuGet with public APIs, then people, *including us* as evidenced by our Tests usage, will begin taking a dependence on these methods and classes, and then we'll have to support them forever.

TomFinley:
As already expressed in #2277 and, as @rogancarr has already noted, I feel fairly strongly that this class and the assembly containing it should be part of a separate nuget. It is *clearly* not core functionality of ML.NET, and it's perfectly fine for our samples to rely on multiple nugets (as, of course, they already do). Yet it is in `Microsoft.ML` nuget. So weird.

[Did you mean to link #1137, instead of #2393? (The latter appears to be irrelevant.) In #1137, it was established (correctly) that we should bundle the datasets in the nuget. It said nothing about the merits of making that samples class part of the main nuget.

sfilipi:
Seems like we're in agreement, to summarize this issue can be closed after making the SamplesUtils a separate package.

[...]

PR #2470, opened 2019-02-07
Add a project for functional tests without visibility into internals of ML.NET

### LINKED ISSUE #2306: Add a Functional.Tests project that doesn't have InternalsVisibleTo
Today, our ML.NET project has `InternalsVisibleTo` all of our tests projects. This is not ideal because we can't ensure that our public API meets scenarios that our customers can actually use. We could easily internalize something that is necessary to a scenario, but we wouldn't catch it by our tests because our tests all have internal-access.

We should add a new test project - or set of test projects - that do not have `InternalsVisibleTo` access. We can put all our public API tests in this project, and we can ensure that customers can use the same code to meet those scenarios. At first we should have at least one test in the project(s), and we can migrate and add new tests over time.

/cc @TomFinley @shauheen @glebuk

rogancarr:
This PR adds a new project, `Microsoft.ML.Functional.Tests` for adding end-to-end scenario tests for ML.NET. This project does not have visibility into the main library, and is not strongly named, so may not be added.

Two tests were also moved from `Microsoft.ML.Tests` into this project:
- `CrossValidation`: Migrated with additional tests
- `ReconfigurablePrediction`: Migrated but marked as `Skip` because Issue #2465 prevents us from completing the scenario.

Fixes #2306


test/Microsoft.ML.Functional.Tests/Microsoft.ML.Functional.Tests.csproj
```diff
@@ -0,0 +1,53 @@
...
+    <NativeAssemblyReference Include="MklImports" />
+  </ItemGroup>
+
+  <!-- TensorFlow is 64-bit only -->
+  <ItemGroup Condition="'$(NativeTargetArchitecture)' == 'x64'">
+    <NativeAssemblyReference Include="tensorflow" />
+    <NativeAssemblyReference Condition="'$(OS)' != 'Windows_NT'" Include="tensorflow_framework" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ML.TensorFlow.TestModels" Version="0.0.7-test" />
```
eerhardt: I think Zeeshan is in the process of publishing version `0.0.8-test`.  We should extract this value into a common place here:

[https://github.com/dotnet/machinelearning/blob/834e4715afa070f5e5c3628693682943b3f0bd6d/build/Dependencies.props#L42-L46](https://github.com/dotnet/machinelearning/blob/834e4715afa070f5e5c3628693682943b3f0bd6d/build/Dependencies.props#L42-L46)

Same for the Onnx TestModels below.

test/Microsoft.ML.Functional.Tests/Microsoft.ML.Functional.Tests.csproj
```diff
@@ -0,0 +1,53 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <AssemblyName>Microsoft.ML.Functional.Tests</AssemblyName>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
eerhardt: I don't think we need `unsafe` blocks (at least not right now). Let's leave this off for now.

test/Microsoft.ML.Functional.Tests/Microsoft.ML.Functional.Tests.csproj
```diff
@@ -0,0 +1,53 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <AssemblyName>Microsoft.ML.Functional.Tests</AssemblyName>
```
eerhardt: There is no need to set `AssemblyName` in new .csproj files. This value gets defaulted to the file's name. You can remove this line.

test/Microsoft.ML.Functional.Tests/Prediction.cs
```diff
@@ -0,0 +1,49 @@
...
+            var pipeline = mlContext.Transforms.Concatenate("Features", new string[] {
+                    "CrimesPerCapita", "PercentResidental", "PercentNonRetail", "CharlesRiver", "NitricOxides", "RoomsPerDwelling",
+                    "PercentPre40s", "EmploymentDistance", "HighwayDistance", "TaxRate", "TeacherRatio"})
+                .Append(mlContext.Transforms.CopyColumns("Label", "MedianHomeValue"))
+                .Append(mlContext.Regression.Trainers.OrdinaryLeastSquares());
+
+            var model = pipeline.Fit(train);
+
+            var scoredTest = model.Transform(test);
+            var metrics = mlContext.Regression.Evaluate(scoredTest);
```
eerhardt: Should we be asserting the metrics are in a certain range?

test/Microsoft.ML.Functional.Tests/Prediction.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    public partial class PredictionScenarios
+    {
+        /// <summary>
+        /// Reconfigurable predictions: The following should be possible: A user trains a binary classifier,
+        /// and through the test evaluator gets a PR curve, the based on the PR curve picks a new threshold
+        /// and configures the scorer (or more precisely instantiates a new scorer over the same predictor)
+        /// with some threshold derived from that.
+        /// </summary>
+        [Fact(Skip = "Blocked by issue #2465")]
```
eerhardt: The test can still be run, right?  Since you are commenting out the code that is being blocked.  Maybe remove the Skip here, and add a TODO to the commented out line below.

test/Microsoft.ML.Functional.Tests/Microsoft.ML.Functional.Tests.csproj
```diff
@@ -0,0 +1,53 @@
...
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <!-- We are turning off strong naming to ensure we never add `InternalsVisibleTo` for these tests -->
+    <SignAssembly>false</SignAssembly>
```
eerhardt: You probably need to set `PublicSign` to false as well, to unblock the Mac and Linux builds.

rogancarr: Thanks for the tip! I couldn't figure out why those builds broke!

---
In reply to: [255168732](https://github.com/dotnet/machinelearning/pull/2470#discussion_r255168732) [](ancestors = 255168732)

test/Microsoft.ML.Functional.Tests/Microsoft.ML.Functional.Tests.csproj
rogancarr: I've created global build variables for those and set them to the latest versions across the projects (they were out of sync). Will resolve once I see that the builds & tests pass.

---
In reply to: [255165705](https://github.com/dotnet/machinelearning/pull/2470#discussion_r255165705) [](ancestors = 255165705)

test/Microsoft.ML.Functional.Tests/Prediction.cs
rogancarr: Good call on checking valid ranges. I added a Common library to add those sorts of checks to.

---
In reply to: [255167478](https://github.com/dotnet/machinelearning/pull/2470#discussion_r255167478) [](ancestors = 255167478)

src/Microsoft.ML.SamplesUtils/SamplesDatasetUtils.cs
```diff
@@ -17,7 +17,12 @@ public static class DatasetUtils
         /// Downloads the housing dataset from the ML.NET repo.
         /// </summary>
         public static string DownloadHousingRegressionDataset()
-        => Download("https://raw.githubusercontent.com/dotnet/machinelearning/024bd4452e1d3660214c757237a19d6123f951ca/test/data/housing.txt", "housing.txt");
+        {
```
eerhardt: Is this change necessary in this PR?

I don't think our tests should be calling this method - BTW

test/Microsoft.ML.Functional.Tests/Microsoft.ML.Functional.Tests.csproj
```diff
@@ -0,0 +1,53 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <AllowUnsafeBlocks>false</AllowUnsafeBlocks>
```
eerhardt: We shouldn't need this line at all. It can be removed.

test/Microsoft.ML.Functional.Tests/Prediction.cs
eerhardt: I think you missed checking that new function in.

---
In reply to: [255228907](https://github.com/dotnet/machinelearning/pull/2470#discussion_r255228907) [](ancestors = 255228907,255167478)

src/Microsoft.ML.SamplesUtils/SamplesDatasetUtils.cs
rogancarr: This change is necessary if we want to use `LoadHousingRegressionDataset` in our tests because there is a race condition on the file lock, so tests will sometimes fail.

Can you explain a bit more why you don't want to use this in tests? Is it that we don't want to use the SamplesUtils project in Tests, or that we shouldn't be downloading data for tests?

If it's the former, check out issue #2420 . We're going to make this a standalone Datasets/ (or some such name) outside of the NuGet project to use in Samples and Tests.

If it's the latter, we are already downloading datasets for tests. But now that I mention it, we can actually add an optional input to LoadHousingRegressionDataset and friends that can load the file from the tests/data/ directory. I'll add this capability now.

---
In reply to: [255236393](https://github.com/dotnet/machinelearning/pull/2470#discussion_r255236393) [](ancestors = 255236393)

rogancarr: Added.

---
In reply to: [255256550](https://github.com/dotnet/machinelearning/pull/2470#discussion_r255256550) [](ancestors = 255256550,255236393)

eerhardt: There are 2 datasets our tests should use.

1. Datasets checked into `test\data`.
2. Datasets that are downloaded into `test\data\external` through the [`DownloadExternalTestFiles`](https://github.com/dotnet/machinelearning/blob/a254bf58c47c1a00348fcc628f48e26dca959ee6/build.proj#L88-L93) build step.

We shouldn't have the test code be downloading random things.

Microsoft.ML.sln
```diff
@@ -928,6 +930,18 @@ Global
 		{5E920CAC-5A28-42FB-936E-49C472130953}.Release-Intrinsics|Any CPU.Build.0 = Release-Intrinsics|Any CPU
 		{5E920CAC-5A28-42FB-936E-49C472130953}.Release-netfx|Any CPU.ActiveCfg = Release-netfx|Any CPU
 		{5E920CAC-5A28-42FB-936E-49C472130953}.Release-netfx|Any CPU.Build.0 = Release-netfx|Any CPU
+		{CFED9F0C-FF81-4C96-8D5E-0436264CA7B5}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{CFED9F0C-FF81-4C96-8D5E-0436264CA7B5}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{CFED9F0C-FF81-4C96-8D5E-0436264CA7B5}.Debug-Intrinsics|Any CPU.ActiveCfg = Debug|Any CPU
```
eerhardt: Did you manually add these?  They look wrong to me.

For example, this line should be:

```
  {CFED9F0C-FF81-4C96-8D5E-0436264CA7B5}.Debug-Intrinsics|Any CPU.ActiveCfg = Debug-Intrinsics|Any CPU
```

and the `-netfx` ones should have `-netfx` on the right side too.

This is probably why CI is failing.

rogancarr: Autogenerated. I'll fix these up.

---
In reply to: [255266128](https://github.com/dotnet/machinelearning/pull/2470#discussion_r255266128) [](ancestors = 255266128)

src/Microsoft.ML.SamplesUtils/SamplesDatasetUtils.cs
rogancarr: Got it. This has been updated to use the local dataset in the `test\data` folder. I'll chase down any other tests using these `Download` commands as I migrate API-Scenario tests to Functional.Tests/

---
In reply to: [255264819](https://github.com/dotnet/machinelearning/pull/2470#discussion_r255264819) [](ancestors = 255264819)

Microsoft.ML.sln
rogancarr: I copied the lines from Microsoft.ML.Tests/ and swapped in the correct GUID.

---
In reply to: [255266430](https://github.com/dotnet/machinelearning/pull/2470#discussion_r255266430) [](ancestors = 255266430,255266128)

test/Microsoft.ML.Functional.Tests/Microsoft.ML.Functional.Tests.csproj
```diff
@@ -0,0 +1,52 @@
...
+
+  <ItemGroup>
+    <NativeAssemblyReference Include="CpuMathNative" />
+    <NativeAssemblyReference Include="FastTreeNative" />
+    <NativeAssemblyReference Include="FactorizationMachineNative" />
+    <NativeAssemblyReference Include="MatrixFactorizationNative" />
+    <NativeAssemblyReference Include="LdaNative" />
+    <NativeAssemblyReference Include="SymSgdNative" />
+    <NativeAssemblyReference Include="MklProxyNative" />
+    <NativeAssemblyReference Include="MklImports" />
```
sfilipi: are those all needed a this point in time?

test/Microsoft.ML.Functional.Tests/Prediction.cs
```diff
@@ -0,0 +1,52 @@
...
+using Xunit;
+
+namespace Microsoft.ML.Functional.Tests
+{
+    public class PredictionScenarios
+    {
+        /// <summary>
+        /// Reconfigurable predictions: The following should be possible: A user trains a binary classifier,
+        /// and through the test evaluator gets a PR curve, the based on the PR curve picks a new threshold
+        /// and configures the scorer (or more precisely instantiates a new scorer over the same predictor)
```
sfilipi: >predictor [](start = 97, length = 9)

model Parameters

test/Microsoft.ML.Functional.Tests/Prediction.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// with some threshold derived from that.
+        /// </summary>
+        [Fact]
+        public void ReconfigurablePrediction()
+        {
+            var mlContext = new MLContext(seed: 789);
```
sfilipi: >seed: 78 [](start = 42, length = 8)

this intentional?

test/Microsoft.ML.Functional.Tests/Prediction.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            // Get the dataset, create a train and test
+            var dataset = DatasetUtils.LoadHousingRegressionDataset(mlContext, BaseTestClass.GetDataPath("housing.txt"));
```
sfilipi: >LoadHousingRegressionDataset [](start = 39, length = 28)

just reference the housing dataset like the other tests are doing. Let's leave DatasetUtils for the samples.

sfilipi: oh I see, you are defining the TextLoader etc in there. I believe there is a separate test file for those, in TestDataets. 

---
In reply to: [255278973](https://github.com/dotnet/machinelearning/pull/2470#discussion_r255278973) [](ancestors = 255278973)

sfilipi:
please take look at the comments about not using DatasetUtils in tests. Thanks. 

---
In reply to: [461637125](https://github.com/dotnet/machinelearning/pull/2470#issuecomment-461637125) [](ancestors = 461637125)

test/Microsoft.ML.Functional.Tests/Microsoft.ML.Functional.Tests.csproj
rogancarr: I think they will be when we're all done.

---
In reply to: [255278694](https://github.com/dotnet/machinelearning/pull/2470#discussion_r255278694) [](ancestors = 255278694)

test/Microsoft.ML.Functional.Tests/Prediction.cs
rogancarr: Good point. I went ahead and added this just like the other tests do. It looks like we'll need to do some refactoring so that we can push most of the loader and options into the dataset object as well, but that can wait for later.

---
In reply to: [255279066](https://github.com/dotnet/machinelearning/pull/2470#discussion_r255279066) [](ancestors = 255279066,255278973)

Issue #1209, opened 2018-10-10
API reference - Samples for Transforms


sfilipi:
We need to add samples on how to use the new transformer, and estimators than reference those samples from the XML documentation so that in docs.microsoft.com users can copy/paste the sample and have a head-starts. 

Mot of the tests that got added as part of the transformer work are a good start for creating a sample.  

# MLContext Catalogs

Catalog | Total APIs | Samples Owner | Samples Status /   ETA | 
-- | -- | -- | -- |
MLContext.Transforms   (root) | 19 | Senja | Remaining:   4 overrides for the normalizer multicolumn examples | 
MLContext.Transforms.Categorical | 2 | ZeeshanA | Done v1 | 
MLContext.Transforms.Conversion | 6 | Senja | DoneV1 | 
MLContext.Transforms.FeatureSelection | 4 | ZeeshanA | Done v1 | 
MLContext.Transforms.TimeSeries | 4 | Senja | Done V1 | 
MLContext.Transforms.Text | 29 | ZeeshanA | Done  V1 | 
MLContext.Data | 10 | Senja | DoneV1 | 
MLContext.Model   (root) | 4 | ZeeshanS  | DoneV1   | 

# P0+P1 Public API (extension methods) per Catalog 

| MLContext.Transforms (root) | Num Overloads | Documentation | Sample | API Owner |
| --------------------------------  | ------------- | ------------  | -----  | ----- |
CopyColumns| 2 | Yes | 2 Can remove dependency on DatasetUtils. | Zeeshan|
Concatenate| 1 | Yes, needs improvement.| 1 - Can remove dependency on DatasetUtils.| Zeeshan |
DropColumns| 1 | Yes| 1 Can remove dependency on DatasetUtils.|Zeeshan |
SelectColumns|2 | Yes, needs improvement. | 2 - Can remove dependency on DatasetUtils.|Zeeshan |
Normalize| 1 | Done. | 1  #3244 |Ivan|
CustomMapping | 1 | Yes, needs improvement. | Done-v1 #3275| Artidoro |
IndicateMissingValues |  2|  |  Done-v1 #3275 | Artidoro | 
ReplaceMissingValues  | 2 |  | Done-v1 #3275 | Artidoro | 
ConvertToGrayscale  | 1 |  Yes, needs fixes. Example not displaying.| 1 #3165 | Abhishek | 
LoadImages  | 1 | Yes, needs fixes. Example not displaying. | 1 #3165 | Abhishek | 
[...]

PR #2476, opened 2019-02-08
Lockdown of Microsoft.ML.LightGBM public surface.

zeahmed:
This PR fixes #2271. Fixes https://github.com/dotnet/machinelearning/issues/2534 fixes #2459

The changes included in this PR are as follows.
1. Changed unnecessary `public` items to `internal`.
2. Created samples in the samples project.
3. Referenced the samples as an example in public API exposed through MLContext.
4. Resolved a bug where one of the constructor of `LightGbmMulticlassTrainer` was using `MakeBoolScalarLabel` instead of `MakeU4ScalarColumn`.

PR #2493, opened 2019-02-10
Add Light GBM sample

jwood803:
Add Light GBM sample as part of #1209.


Ivanidzo4ka:
https://github.com/dotnet/machinelearning/pull/2476 @zeahmed  is already has PR which doing same thing and more.  So I would rather merge his PR.
On other hand we probably owe to update that issue with who is working on documentation. To prevent people from spending time on something which is already in work.

jwood803:
Sorry about that. I'll close this in favor of #2476

PR #2505, opened 2019-02-11
Move IModelCombiner to Microsoft.ML.Data and Lockdown Microsoft.ML.Ensemble

### LINKED ISSUE #2268: Lockdown Microsoft.ML.Ensemble public surface, or isolate it
We ought to internalize as much of that assembly to those things we want to be public. The trouble with this one is that the API for ensembling, it is unclear what a good public version would look like and we don't have time to develop it. So it may be that at least for now internalizing the *whole thing* is the answer.

Complicating matters, FastTree depends on this assembly; it may do so for good or bad reasons, I am not sure. That should be investigated. If it is not doing so for important reasons, we could break that dependency in some fashion, and then put Ensemble in a separate Nuget that we will keep in a sub-v1 version even as much of the API becomes v1.

yaeldekel:
We have an implementation of IModelCombiner in Microsoft.ML.FastTree. This PR is to move the interface to Microsoft.ML.Data so that Microsoft.ML.FastTree doesn't need to reference Microsoft.ML.Ensemble.

This PR also internalizes most of the classes in Microsoft.ML.Ensemble. Fixes #2268 .


Ivanidzo4ka:
>     public sealed class TreeEnsembleCombiner : IModelCombiner

shall we make it internal as well?

---
Refers to: src/Microsoft.ML.FastTree/TreeEnsemble/TreeEnsembleCombiner.cs:15 in a167c4d. [](commit_id = a167c4de63e6bbb2b651e5f5557fd2f3a11eb328, deletion_comment = False)

src/Microsoft.ML.Ensemble/OutputCombiners/Average.cs
```diff
@@ -12,7 +12,7 @@
 
 namespace Microsoft.ML.Trainers.Ensemble
 {
-    public sealed class Average : BaseAverager, IRegressionOutputCombiner
+    internal sealed class Average : BaseAverager, IRegressionOutputCombiner
```
Ivanidzo4ka: Do we have any tests/ examples of how use Ensembles? I'm just curious how user will set any of combiners if they all hidden

yaeldekel: The current API for ensembles is not the one we want to make public, so I am internalizing it until we get a chance to develop a good public API for it. 

---
In reply to: [256207741](https://github.com/dotnet/machinelearning/pull/2505#discussion_r256207741) [](ancestors = 256207741)

Ivanidzo4ka: So if everything in this Ensemble is hidden, shall we exclude it from Microsoft.ML package? See Microsoft.ML.Ensemble.csproj  <IncludeInPackage>Microsoft.ML</IncludeInPackage>

---
In reply to: [256514419](https://github.com/dotnet/machinelearning/pull/2505#discussion_r256514419) [](ancestors = 256514419,256207741)

yaeldekel: Hmm. Good question. If there are any consumers of the nuget that use maml or entry points then we should probably leave it. 

---
In reply to: [256542187](https://github.com/dotnet/machinelearning/pull/2505#discussion_r256542187) [](ancestors = 256542187,256514419,256207741)

PR #2509, opened 2019-02-12
Internalize IDataTransform

Ivanidzo4ka:
Hide the unhideable, bestfriend the unfrienable!
fixes https://github.com/dotnet/machinelearning/issues/1995

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/TransformBase.cs
+++ b/src/Microsoft.ML.Data/Transforms/TransformBase.cs
@@ -18,7 +18,8 @@ namespace Microsoft.ML.Data
     /// <summary>
     /// Base class for transforms.
     /// </summary>
-    public abstract class TransformBase : IDataTransform
+    [BestFriend]
+    internal abstract class TransformBase : IDataTransform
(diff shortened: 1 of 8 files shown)


src/Microsoft.ML.Data/Transforms/TransformBase.cs
```diff
@@ -244,7 +247,8 @@ public override bool IsColumnActive(int col)
     /// on multiple input columns.
     /// This class provides the implementation of ISchema and IRowCursor.
     /// </summary>
-    public abstract class OneToOneTransformBase : RowToRowMapperTransformBase, ITransposeDataView, ITransformCanSavePfa,
+    [BestFriend]
+    internal abstract class OneToOneTransformBase : RowToRowMapperTransformBase, ITransposeDataView, ITransformCanSavePfa,
         ITransformCanSaveOnnx
```
artidoro: There are a few `[BestFriend]` in this class, if you are doing another iteration, maybe we could make those public and remove the attribute. Not a requirement though.

Issue #2381, opened 2019-02-01
Add Ubuntu CI leg


danmosemsft:
Right now we build and test for Linux on only one distro -- Centos 7 (a container hosted on Ubuntu 1804). We use .NET Core 2.1.

We maybe should add a leg to test on a Debian/Ubuntu family distro (Centos is in the RedHat family). That should be able to use all the same binaries. If we did that it would probably be good to use .NET Core 3.0 instead of 2.1 to get Linux coverage with that.

.NET Core also supports Alpine, which some people use to minimise the size and security surface of their containers. It would be more "interesting" (takes= more work to support) because as well as having few default packages it is musl not glibc based so it needs a special build. We probably have to defer support for Alpine.

@eerhardt thoughts?

eerhardt:
We've also discussed adding an Ubuntu leg for the OnnxTransform tests, since the OnnxRuntime doesn't support CentOS/RedHat 7.

/cc @jignparm

jignparm:
PR #2516 adds an Ubuntu CI leg. 

It's still using .NET Core 2.1, but it looks like the upgrade to .Net Core 3.0 can be done after the buildtool changes mentioned on the PR.

PR #2516, opened 2019-02-12
Adding a new ubuntu CI leg

Anipik:
Related To https://github.com/dotnet/machinelearning/issues/2381

Currently we are just running the tests on centos linux. A couple of libraries dont run on centos distro but run on ubuntu so we are adding a new ci leg to increase coverafe and run these tests on ubuntu

### DIFF
--- a/.vsts-dotnet-ci.yml
+++ b/.vsts-dotnet-ci.yml
@@ -4,11 +4,11 @@
 
 resources:
   containers:
-  - container: LinuxCentosContainer
+  - container: CentosContainer
     image: microsoft/dotnet-buildtools-prereqs:centos-7-b46d863-20180719033416
   
-  - container: LinuxUbuntuContainer
-    image: microsoft/dotnet-buildtools-prereqs:ubuntu1604_cross_prereqs_v1
+  - container: UbuntuContainer
+    image: microsoft/dotnet-buildtools-prereqs:ubuntu-16.04-10fcdcf-20190208200917
 
 phases:
 - template: /build/ci/phase-template.yml
@@ -29,19 +29,37 @@ phases:
 
 - template: /build/ci/phase-template.yml
   parameters:
-    name: Linux_Centos
+    name: Centos
     buildScript: ./build.sh
+    customMatrixes:
+      Build_Debug_Intrinsics:
+        _configuration: Debug-Intrinsics
+        _config_short: DI
+        _includeBenchmarkData: false
+      Build_Release:
+        _configuration: Release
+        _config_short: R
+        _includeBenchmarkData: true
     queue:
       name: Hosted Ubuntu 1604
-      container: LinuxCentosContainer
+      container: CentosContainer
 
 - template: /build/ci/phase-template.yml
   parameters:
-    name: Linux_Ubuntu
+    name: Ubuntu
     buildScript: ./build.sh
+    customMatrixes:
+      Build_Debug_Intrinsics:
+        _configuration: Debug-Intrinsics
+        _config_short: DI
+        _includeBenchmarkData: false
+      Build_Release:
+        _configuration: Release
+        _config_short: R
+        _includeBenchmarkData: true
     queue:
       name: Hosted Ubuntu 1604
-      container: LinuxUbuntuContainer
+      container: UbuntuContainer
 
 - template: /build/ci/phase-template.yml
   parameters:
--- a/.vsts-dotnet-ci.yml
+++ b/.vsts-dotnet-ci.yml
@@ -4,8 +4,11 @@
 
 resources:
   containers:
-  - container: LinuxContainer
+  - container: LinuxCentosContainer
     image: microsoft/dotnet-buildtools-prereqs:centos-7-b46d863-20180719033416
+  
+  - container: LinuxUbuntuContainer
+    image: microsoft/dotnet-buildtools-prereqs:ubuntu1604_cross_prereqs_v1
 
 phases:
 - template: /build/ci/phase-template.yml
@@ -26,11 +29,19 @@ phases:
 
 - template: /build/ci/phase-template.yml
   parameters:
-    name: Linux
+    name: Linux_Centos
     buildScript: ./build.sh
     queue:
       name: Hosted Ubuntu 1604
-      container: LinuxContainer
+      container: LinuxCentosContainer
+
+- template: /build/ci/phase-template.yml
+  parameters:
+    name: Linux_Ubuntu
+    buildScript: ./build.sh
+    queue:
+      name: Hosted Ubuntu 1604
+      container: LinuxUbuntuContainer
 
 - template: /build/ci/phase-template.yml
   parameters:
--- a/.vsts-dotnet-ci.yml
+++ b/.vsts-dotnet-ci.yml
@@ -31,15 +31,6 @@ phases:
   parameters:
     name: Centos
     buildScript: ./build.sh
-    customMatrixes:
-      Build_Debug_Intrinsics:
-        _configuration: Debug-Intrinsics
-        _config_short: DI
-        _includeBenchmarkData: false
-      Build_Release:
-        _configuration: Release
-        _config_short: R
-        _includeBenchmarkData: true
     queue:
       name: Hosted Ubuntu 1604
       container: CentosContainer
@@ -48,15 +39,6 @@ phases:
   parameters:
     name: Ubuntu
     buildScript: ./build.sh
-    customMatrixes:
-      Build_Debug_Intrinsics:
-        _configuration: Debug-Intrinsics
-        _config_short: DI
-        _includeBenchmarkData: false
-      Build_Release:
-        _configuration: Release
-        _config_short: R
-        _includeBenchmarkData: true
     queue:
       name: Hosted Ubuntu 1604
       container: UbuntuContainer
--- a/init-tools.sh
+++ b/init-tools.sh
@@ -10,7 +10,7 @@ if [ -z "${__BUILDTOOLS_SOURCE:-}" ]; then __BUILDTOOLS_SOURCE=https://dotnet.my
 export __BUILDTOOLS_USE_CSPROJ=true
 __BUILD_TOOLS_PACKAGE_VERSION=$(cat "$__scriptpath/BuildToolsVersion.txt" | sed 's/\r$//') # remove CR if mounted repo on Windows drive
 
-if [[ $1 == *"Intrinsics"* ]]; then
+if [[ $2 == *"Intrinsics"* ]]; then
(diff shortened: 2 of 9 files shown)


.vsts-dotnet-ci.yml
```diff
@@ -26,11 +29,19 @@ phases:
...
+    name: Linux_Centos
     buildScript: ./build.sh
     queue:
       name: Hosted Ubuntu 1604
-      container: LinuxContainer
+      container: LinuxCentosContainer
+
+- template: /build/ci/phase-template.yml
+  parameters:
+    name: Linux_Ubuntu
```
eerhardt: Was one of the proposals that we would use `netcoreapp3.0` on Ubuntu?  That way would could get some non-Windows netcoreapp3.0 coverage?

.vsts-dotnet-ci.yml
```diff
@@ -4,8 +4,11 @@
 
 resources:
   containers:
-  - container: LinuxContainer
+  - container: LinuxCentosContainer
```
eerhardt: I don't think these need the prefix `Linux`.  `CentosContainer` and `UbuntuContainer` is enough IMO.

.vsts-dotnet-ci.yml
```diff
@@ -4,8 +4,11 @@
...
     image: microsoft/dotnet-buildtools-prereqs:centos-7-b46d863-20180719033416
+  
+  - container: LinuxUbuntuContainer
+    image: microsoft/dotnet-buildtools-prereqs:ubuntu1604_cross_prereqs_v1
```
eerhardt: From the test failures, it looks like this image doesn't have `libgdiplus` installed on it.

src/Native/CpuMathNative/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ set(SOURCES
 )
 
 if(NOT WIN32)
-    set_property(SOURCE Sse.cpp APPEND_STRING PROPERTY COMPILE_FLAGS " -msse3")
+    set_property(SOURCE Sse.cpp APPEND_STRING PROPERTY COMPILE_FLAGS " -msse3 -msse4.1 -msse4.2")
```
eerhardt: Are we sure these build changes are necessary? This worked before on Ubuntu - what changed?

.vsts-dotnet-ci.yml
MichaelSimons: This doesn't look like a tag I'd recommend using.  What are the requirements here?

test/Microsoft.ML.TestFramework/Attributes/OnnxTheoryAttribute.cs
```diff
@@ -19,7 +19,7 @@ public OnnxTheoryAttribute() : base("Onnx is 64-bit Windows only")
         /// <inheritdoc />
         protected override bool IsEnvironmentSupported()
         {
-            return Environment.Is64BitProcess && RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+            return Environment.Is64BitProcess && (RuntimeEnvironment.OperatingSystem.Contains("Windows") || RuntimeEnvironment.OperatingSystem.Contains("ubuntu"));
```
eerhardt: > (RuntimeEnvironment.OperatingSystem.Contains("Windows")

This should stay as `RuntimeInformation.IsOSPlatform(OSPlatform.Windows)`.

> RuntimeEnvironment.OperatingSystem.Contains("ubuntu")

I believe the more correct check here would be "get the glibc version. If it is greater than or equal to the supported `glibc` version Onnx, return `true`.

.vsts-dotnet-ci.yml
```diff
@@ -26,11 +29,37 @@ phases:
...
+      name: Hosted Ubuntu 1604
+      container: CentosContainer
+
+- template: /build/ci/phase-template.yml
+  parameters:
+    name: Ubuntu
+    buildScript: ./build.sh
+    customMatrixes:
+      Build_Debug_Intrinsics:
+        _configuration: Debug-Intrinsics
```
eerhardt: I thought we were going to switch to the `Release` to use `3.0`/`Intrinsics`. That way we get Debug and Release coverage across both versions:

```
CentOS|Debug|3.0
CentOS|Release|2.1
Ubuntu|Debug|2.1
Ubuntu|Release|3.0
```

init-tools.sh
```diff
@@ -9,7 +9,12 @@ __DOTNET_CMD="$__DOTNET_PATH/dotnet"
 if [ -z "${__BUILDTOOLS_SOURCE:-}" ]; then __BUILDTOOLS_SOURCE=https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json; fi
 export __BUILDTOOLS_USE_CSPROJ=true
 __BUILD_TOOLS_PACKAGE_VERSION=$(cat "$__scriptpath/BuildToolsVersion.txt" | sed 's/\r$//') # remove CR if mounted repo on Windows drive
-__DOTNET_TOOLS_VERSION=$(cat "$__scriptpath/DotnetCLIVersion.txt" | sed 's/\r$//') # remove CR if mounted repo on Windows drive
+
+if [[ $2 == *"Intrinsics"* ]]; then
+    __DOTNET_TOOLS_VERSION=$(cat "$__scriptpath/DotnetCLIVersion.netcoreapp.latest.txt" | sed 's/\r$//') # remove CR if mounted repo on Windows drive
```
eerhardt: (nit) there is some duplication here that can be factored out.  Create a new variable to hold the file name, setting it to different values based on the Configuration.  Then set ` __DOTNET_TOOLS_VERSION` in a single place, with the `| sed...` piece also in 1 place.

.vsts-dotnet-ci.yml
Anipik: sorry I didn't saw it correctly, I will correct it

test/Directory.Build.props
```diff
@@ -26,6 +26,7 @@
   </PropertyGroup>
   
   <ItemGroup>
+    <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="2.1.0" />
```
eerhardt: Let's not use this.  And for sure every project doesn't need this reference, even if we were going to use it in one or two projects.  `.props` files are only for things that are common across many/all projects.

test/Microsoft.ML.TestFramework/Attributes/OnnxTheoryAttribute.cs
jignparm: The "Windows" versus "Ubuntu" would probably work for ML.NET CI (which is where the tests would run). The glibc version check of 2.23+ is a better (more general check) for OnnxRuntime. However, there doesn't seem to be a very good way to check for this.  We could test the loader version (e.g. "ldd --version"), but that would spawn an external process. Also, I don't believe that will work if the OS has been patched.

eerhardt: This is how we get the `libc` version in the CLI for telemetry purposes:

https://github.com/dotnet/cli/blob/95c6eff6daa1a69f29c42b2d405400ad44bdec91/src/dotnet/Telemetry/ExternalTelemetryProperties.cs#L110-L132

.vsts-dotnet-ci.yml
Anipik: Moving back the CI to netcoreap2.1 untill the cli people resolves the https://github.com/dotnet/cli/issues/10802

Anipik:
@eerhardt @jignparm can you please review this one. I have addressed the feedback.

@eerhardt currently i moved back the ubuntu ci to the netcoreapp2.1. 
Most probably we will have to make some changes in the build tools script in order to make the netcoreapp3.0 non-windows ci work. I will change the ci leg to netcoreapp3.0 when the buildtools changes have been done

TomFinley:
Hi @Anipik thank you for working on this. Can't we narrow it down a little, to specific components of interest? As far as I understood there were only a few things that would be potentially problematic. It's just a little irritating to think of this 99% C# codebase getting yet another fully realized leg just to run what is really only one test of interest.

test/Microsoft.ML.TestFramework/Attributes/AttributeHelpers.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.ML.TestFramework.Attributes
+{
+    internal static class AttributeHelpers
+    {
+        internal static bool CheckLibcVersionGreaterThanMinimum(double minVersion)
+        {
+#if !NETFRAMEWORK
```
eerhardt: Why the `#if` here? The `if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))` will catch anytime we are running on Windows, and return `false`.

test/Microsoft.ML.TestFramework/Attributes/AttributeHelpers.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.ML.TestFramework.Attributes
+{
+    internal static class AttributeHelpers
+    {
+        internal static bool CheckLibcVersionGreaterThanMinimum(double minVersion)
```
eerhardt: Use `System.Version` here instead.

`1.12` > `1.2` when using versions.  But not in doubles.

test/Microsoft.ML.TestFramework/Attributes/AttributeHelpers.cs
```diff
@@ -0,0 +1,41 @@
...
+                return false;
+            }
+
+            try
+            {
+                version = Marshal.PtrToStringUTF8(gnu_get_libc_version());
+            }
+            catch (Exception e)
+            {
+                Debug.Assert(e is DllNotFoundException || e is EntryPointNotFoundException);
```
eerhardt: How about just catch those two exception types instead. And let any other exception types throw.

eerhardt:
> Hi @Anipik thank you for working on this. Can't we narrow it down a little, to specific components of interest? As far as I understood there were only a few things that would be potentially problematic. It's just a little irritating to think of this 99% C# codebase getting yet another fully realized leg just to run what is really only one test of interest.

@TomFinley - See #2381 for the reasoning.

test/Microsoft.ML.TestFramework/Attributes/AttributeHelpers.cs
Anipik: The reason behind this is that ```Marshal.PtrToStringUTF8``` is not available for netframework which leads to compile error

Issue #2488, opened 2019-02-10
ProductRecommander sample - System.AccessViolationException


Ikaer:
### System information

- **OS version/distro**: Win10
- **.NET Version (eg., dotnet --info)**: .Net Core 2.1

### Issue

- **What did you do?** I tried the sample Product recommander with my own sample of product IDs. I first try it with the provided data (Amazon0302) and it works fine. Then I try a different file (which is attached).
- **What happened?** An exception is throw at ITransformer model = est.Fit(traindata); System.AccessViolationException: 'Attempted to read or write protected memory. This is often an indication that other memory is corrupt.'
- **What did you expect?** To not have this kind of exception at least (this is scary) and to be able to use a different sample than the one provided.
I have the same error if, for example, I truncate the file Amazon0302 to only keep the first 10 lines.

### Source code / logs
[test_sample.txt](https://github.com/dotnet/machinelearning/files/2848725/test_sample.txt)
I just downloaded the ProductRecommander sample and replace the variable TrainingDataLocation with my own file. I reproduced the problem on two machines (one with Win10, another one with Windows Server 2016, both with .net core 2.1)

Ivanidzo4ka:
Thank you for reporting this. I'll investigate this issue.

Ivanidzo4ka:
If you look into sample all across it we use magic number 262111. This number is basically maximum ID +1 you have in your file. Everything above that number would be ignored.
You can put 375092 for your sample and it will work.

But that's not excuse for us to throw exception.
@wschin  I can see how matrix factorization throw exception on https://github.com/cjlin1/libmf/blob/dac39ab92d61b54806fbb214a8ccd1c3aa6eb9ee/mf.cpp#L3608 this line
If I set `u=262109` it works fine, if I set `u = 262110` it I get  `ptr` equal to null and `ptr->v` leads to exception (or b[v])

Ikaer:
Thanks to have check it out 👍 . Really nice of you.

Just a last question, I've provided a fix sample, so I can determine this maximum ID + 1 and set the KeyCount. But ultimately, I would like to provide a dynamic sample, so I can still determine this maximum ID + 1, and I can through code set it on this line:
![image](https://user-images.githubusercontent.com/7936075/52610105-ffde6500-2e7f-11e9-9ead-241b7efecd5d.png)
But I don't know how to dynamically set it on those lines:
[...]

PR #2525, opened 2019-02-13
Fix an initial-value problem caused by unseen row/column

wschin:
Update LIBMF to have this [fix](https://github.com/cjlin1/libmf/pull/25). The problem is that when the largest row index in the training set doesn't match the number of rows in the specified matrix shape, rows' starting addresses may look like `[ptr_to_first_element_in_row_1, ptr_to_first_element_in_row_8, ptr_to_the_end_of_element_array, nullptr, nullptr]` but the correct value should be `[ptr_to_first_element_in_row_1, ptr_to_first_element_in_row_8, ptr_to_the_end_of_element_array, ptr_to_the_end_of_element_array, ptr_to_the_end_of_element_array]`.

Fixes https://github.com/dotnet/machinelearning/issues/2488

### DIFF
--- a/test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
+++ b/test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
@@ -530,21 +530,20 @@ namespace Microsoft.ML.Tests.TrainerEstimators
         }
 
         [MatrixFactorizationFact]
-        public void OneClassMatrixFactorizationWithUnseenRow()
+        public void OneClassMatrixFactorizationWithUnseenColumnAndRow()
         {
             // Create an in-memory matrix as a list of tuples (column index, row index, value). For one-class matrix
             // factorization problem, unspecified matrix elements are all a constant provided by user. If that constant is 0.15,
             // the following list means a 3-by-2 training matrix with elements:
-            //   (0, 0, 1), (1, 0, 1), (1, 1, 1), (0, 1, 0.15), (0, 2, 0.15), (1, 2, 0.15).
-            // because matrix elements at (0, 1), (0, 2), and (1, 2) are not specified. Below is a visualization of the training matrix.
-            //   [1, 1]
-            //   |?, 1| where ? will be set to 0.15 by user when creating the trainer.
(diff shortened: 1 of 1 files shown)


Ivanidzo4ka:
Can we have a test?

wschin:
As you wish. :)

---
In reply to: [463012943](https://github.com/dotnet/machinelearning/pull/2525#issuecomment-463012943) [](ancestors = 463012943)

test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
```diff
@@ -528,5 +528,78 @@ public void MatrixFactorizationBackCompat()
             // Negative example (i.e., examples can not be found in dataMatrix) is close to 0.15 (specified by s.C = 0.15 in the trainer).
             CompareNumbersWithTolerance(0.141411, testResults[1].Score, digitsOfPrecision: 5);
         }
+
+        [MatrixFactorizationFact]
+        public void OneClassMatrixFactorizationWithUnseenRow()
```
Ivanidzo4ka: >OneClassMatrixFactorizationWithUnseenRow [](start = 20, length = 40)

Right now you limits for columns are 2 and rows is 3.

Your sample have 2 columns and 2 rows.
Problem in code was related to border cases and triggered when we provide data which doesn't get to the limits.
Can we test fully? Right now you data doesn't reach max rows, can we test what would happen if you also don't reach max columns?

wschin: Sure. Please take a look at the visualization of the latest training matrix below.

---
In reply to: [256537572](https://github.com/dotnet/machinelearning/pull/2525#discussion_r256537572) [](ancestors = 256537572)

artidoro: In the same spirit, it might be worth it to test an empty `IDataView` for training. 

---
In reply to: [256600662](https://github.com/dotnet/machinelearning/pull/2525#discussion_r256600662) [](ancestors = 256600662,256537572)

Issue #1871, opened 2018-12-13
Follow up on Calibrator estimators


sfilipi:
There are a few follow ups to the work to create calibrator estimators:

1- The public classes in Microsoft.ML.Core/Prediction/Calibrator.cs need to have descriptive XML documentation. 
2- The Calibrators need to be added as property to the BinaryClassificationContext
3- Some of the Calibrators have a public constructor, that allows initializing the parameters. Does  it make sense to expose and  wire those parameters to the CalibratorEstimators, and set them from there? 
So they will all look like [FixedPalttCalibratorEstimator](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs#L319)

cc @Zruty0

Zruty0:
There are 2 cases of calibrators: one where we set the parameters (slope, offset) manually, and one where we learn them. 

Both cases will need an estimator: first one will just be a regular trainer, and the second will be a trivial estimator. I world prefer these 2 cases to be invoked via 2 different methods, instead of a single overload.

TomFinley:
It seems to me that these when fit, given that they're parameterized with model parameters, should be producing a model with `CalibratedModelParametersBase`, with the sub-model being the input model. I do not see that the estimators currently are doing this. Instead what I see is that.

I am at least confident we should not have *two* mechanisms for doing this. Right now we have two, and I'm not sure why:

1. `CalibratorTransformer<TCalibrator>`, and

2. `BinaryPredictionTransformer<CalibratedModelParametersBase<TModel, TCalibrator>>`

Ivanidzo4ka:
I can be wrong, but for me it looks like done issue.
https://github.com/dotnet/machinelearning/pull/2601 does 1) and 3) and I believe @wschin  is done with  making sure we return calibrated or not calibrated model.

TomFinley:
> I can be wrong, but for me it looks like done issue.
> #2601 does 1) and 3)

Hi @Ivanidzo4ka , you are omitting 2 I think. The estimator for calibration is still not accessible through `MLContext`, the `IHostEnvironment` constructor is still public at the present time, as we see below.

https://github.com/dotnet/machinelearning/blob/eb60021712a729bbcd19abd28ef8e4669d7ee85b/src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs#L279-L282

[...]

PR #2526, opened 2019-02-13
Update calibrator estimators to be more suitable.

### LINKED ISSUE #2515: Calibrator estimators as successor to ICalibratorTrainer
One of the legacy interfaces we have is the calibrator trainer.

https://github.com/dotnet/machinelearning/blob/f269adcd30a33feeb969c17c2a9f9db235113cb1/src/Microsoft.ML.Data/Prediction/Calibrator.cs#L95

This interface is unsuitable to being a public class as part of the `IEstimator`/`ITransformer` idioms, being more of implementation interface for the infrastructure of the calibrators. So we don't want that exposed, but we do nonetheless want the ability to given some data, produce a calibrator out of it.

Happily we *almost* have it, in the form of the subclasses of the calibrator estimators.

https://github.com/dotnet/machinelearning/blob/f269adcd30a33feeb969c17c2a9f9db235113cb1/src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs#L52

The only real trouble is that this object assumes, at a fairly deep level, that internally there is an `IPredictor` in it, but as far as I know this is mostly due to some internal convenience code that I hope could be refactored to just take an `IDataView` directly.

This might be viewed as a sub-part of #1871.

/ccing @sfilipi for visibility.

TomFinley:
Fixes #2515, contributes towards #1871 and indirectly towards #2251.

* Internalize infrastructure only interface ICalibratorTrainer.
* Update calibrator estimators so they are a suitable replacement for calibrator trainers in the public surface, e.g., no longer take IPredictor.

Note that I did not add a calibrator catalog, since:

1. I believe @sfilipi already has assigned herself the issue #1871 to do so, and
2. Real reason, I am lazy! 😄

Calibrator estimators now are configured by the score/label/optionally weight column, *except* for the fixed Platt estimator, which only takes score (since it is not trained).

Note that ultimately the trainer estimator constructors *will* be internal, pending #1871.

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Calibrator.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Calibrator.cs
@@ -1,8 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
-
-using System;
+﻿using System;
 using System.Linq;
 using Microsoft.ML.Calibrator;
 using Microsoft.ML.Data;
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Calibrator.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Calibrator.cs
@@ -1,4 +1,8 @@
-﻿using System;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
 using System.Linq;
 using Microsoft.ML.Calibrator;
 using Microsoft.ML.Data;
@@ -75,7 +79,7 @@ namespace Microsoft.ML.Samples.Dynamic
 
             // Let's train a calibrator estimator on this scored dataset. The trained calibrator estimator produces a transformer
             // that can transform the scored data by adding a new column names "Probability". 
-            var calibratorEstimator = new PlattCalibratorEstimator(mlContext, model, "Sentiment", "Features");
+            var calibratorEstimator = new PlattCalibratorEstimator(mlContext, "Sentiment", "Score");
             var calibratorTransformer = calibratorEstimator.Fit(scoredData);
 
             // Transform the scored data with a calibrator transfomer by adding a new column names "Probability". 
--- a/src/Microsoft.ML.Data/Prediction/Calibrator.cs
+++ b/src/Microsoft.ML.Data/Prediction/Calibrator.cs
@@ -12,6 +12,7 @@ using Microsoft.Data.DataView;
 using Microsoft.ML;
 using Microsoft.ML.Calibrator;
 using Microsoft.ML.CommandLine;
+using Microsoft.ML.Core.Data;
 using Microsoft.ML.Data;
 using Microsoft.ML.EntryPoints;
 using Microsoft.ML.Internal.Calibration;
@@ -85,14 +86,24 @@ namespace Microsoft.ML.Internal.Calibration
     /// <summary>
     /// Signature for the loaders of calibrators.
     /// </summary>
-    public delegate void SignatureCalibrator();
+    [BestFriend]
+    internal delegate void SignatureCalibrator();
 
+    [BestFriend]
     [TlcModule.ComponentKind("CalibratorTrainer")]
-    public interface ICalibratorTrainerFactory : IComponentFactory<ICalibratorTrainer>
+    internal interface ICalibratorTrainerFactory : IComponentFactory<ICalibratorTrainer>
     {
     }
 
-    public interface ICalibratorTrainer
+    /// <summary>
+    /// This is a legacy interface still used for the command line and entry-points. All applications should transition away
+    /// from this interface and still work instead via <see cref="IEstimator{TTransformer}"/> of <see cref="CalibratorTransformer{TICalibrator}"/>,
+    /// for example, the subclasses of <see cref="CalibratorEstimatorBase{TICalibrator}"/>. However for now we retain this
+    /// until such time as those components making use of it can transition to the new way. No public surface should use
+    /// this, and even new internal code should avoid its use if possible.
+    /// </summary>
+    [BestFriend]
+    internal interface ICalibratorTrainer
     {
         /// <summary>
         /// True if the calibrator needs training, false otherwise.
@@ -105,6 +116,17 @@ namespace Microsoft.ML.Internal.Calibration
 
         /// <summary> Finish up training after seeing all examples </summary>
         ICalibrator FinishTraining(IChannel ch);
+    }
+
+    /// <summary>
+    /// This is a shim interface implemented only by <see cref="CalibratorEstimatorBase{TICalibrator}"/> to enable
+    /// access to the underlying legacy <see cref="ICalibratorTrainer"/> interface for those components that use
+    /// that old mechanism that we do not care to change right now.
+    /// </summary>
+    [BestFriend]
+    internal interface IHaveCalibratorTrainer
+    {
+        ICalibratorTrainer CalibratorTrainer { get; }
     }
 
     /// <summary>
@@ -842,6 +864,64 @@ namespace Microsoft.ML.Internal.Calibration
             return CreateCalibratedPredictor(env, (IPredictorProducing<float>)predictor, trainedCalibrator);
         }
 
+        public static ICalibrator TrainCalibrator(IHostEnvironment env, IChannel ch, ICalibratorTrainer caliTrainer, IDataView scored, string labelColumn, string scoreColumn, string weightColumn = null, int maxRows = _maxCalibrationExamples)
+        {
+            Contracts.CheckValue(env, nameof(env));
+            env.CheckValue(ch, nameof(ch));
+            ch.CheckValue(scored, nameof(scored));
+            ch.CheckValue(caliTrainer, nameof(caliTrainer));
+            ch.CheckParam(!caliTrainer.NeedsTraining || !string.IsNullOrWhiteSpace(labelColumn), nameof(labelColumn),
+                "If " + nameof(caliTrainer) + " requires training, then " + nameof(labelColumn) + " must have a value.");
+            ch.CheckNonWhiteSpace(scoreColumn, nameof(scoreColumn));
+
+            if (!caliTrainer.NeedsTraining)
+                return caliTrainer.FinishTraining(ch);
+
+            var labelCol = scored.Schema[labelColumn];
+            var scoreCol = scored.Schema[scoreColumn];
+
+            var weightCol = weightColumn == null ? null : scored.Schema.GetColumnOrNull(weightColumn);
+            if (weightColumn != null && !weightCol.HasValue)
+                throw ch.ExceptSchemaMismatch(nameof(weightColumn), "weight", weightColumn);
+
+            ch.Info("Training calibrator.");
+
+            var cols = weightCol.HasValue ?
+                new Schema.Column[] { labelCol, scoreCol, weightCol.Value } :
+                new Schema.Column[] { labelCol, scoreCol };
+
+            using (var cursor = scored.GetRowCursor(cols))
+            {
+                var labelGetter = RowCursorUtils.GetLabelGetter(cursor, labelCol.Index);
+                var scoreGetter = RowCursorUtils.GetGetterAs<Single>(NumberType.R4, cursor, scoreCol.Index);
+                ValueGetter<Single> weightGetter = !weightCol.HasValue ? (ref float dst) => dst = 1 :
+                    RowCursorUtils.GetGetterAs<Single>(NumberType.R4, cursor, weightCol.Value.Index);
+
+                int num = 0;
+                while (cursor.MoveNext())
+                {
+                    Single label = 0;
+                    labelGetter(ref label);
+                    if (!FloatUtils.IsFinite(label))
+                        continue;
+                    Single score = 0;
+                    scoreGetter(ref score);
+                    if (!FloatUtils.IsFinite(score))
+                        continue;
+                    Single weight = 0;
+                    weightGetter(ref weight);
+                    if (!FloatUtils.IsFinite(weight))
+                        continue;
+
+                    caliTrainer.ProcessTrainingExample(score, label > 0, weight);
+
+                    if (maxRows > 0 && ++num >= maxRows)
+                        break;
+                }
+            }
+            return caliTrainer.FinishTraining(ch);
+        }
+
         /// <summary>
         /// Trains a calibrator.
         /// </summary>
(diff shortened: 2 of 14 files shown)


src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -20,7 +19,8 @@ namespace Microsoft.ML
     /// </summary>
     public abstract class TrainCatalogBase
     {
-        protected internal readonly IHost Host;
+        [BestFriend]
+        private protected readonly IHost Host;
```
glebuk: >private [](start = 8, length = 7)

is this attribute valid on a private field? ("A private protected member is accessible by types derived from the containing class, but only within its containing assembly.")

TomFinley: This is not a private field. See the modifier `private protected`.

abgoswam: This field seems to be also used in RecommenderCatalog.cs  (which is in a separate assembly Microsoft.ML.Recommender).  

So the [BestFriend] attribute is valid  for both `internal` and `private protected` access modifiers. Am i right ?

glebuk: Yes, this is a very quirky behaviour of C# 7.2 Apparently they treat private protected as "internal" from the point of view of the InternalsVisibleTo attribute.  This is actual behaviour and under the hood the CLR marks the private protected as protectedAndInternal accesibility level.  Very confusing. But the bottom line is that "private protected" field is only visible to derived classes in the same assembly and any assembly that is a friend to it.

---
In reply to: [256224446](https://github.com/dotnet/machinelearning/pull/2526#discussion_r256224446) [](ancestors = 256224446)

glebuk: I understand, but it is confusing. Perhaps we can update the BestFriend attribute xml docs to say 
    /// <summary>
    /// Intended to be applied to types and members marked as internal OR PRIVATE PROTECTED to indicate that friend access of ...

---
In reply to: [256221035](https://github.com/dotnet/machinelearning/pull/2526#discussion_r256221035) [](ancestors = 256221035)

Ivanidzo4ka:
> namespace Microsoft.ML.Calibrator

Just curious, are you happy with this namespace? Don't you think it should be ML.Training?

---
Refers to: src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs:25 in f00b206. [](commit_id = f00b206921404ab8c5dfc192665dee6e987b6716, deletion_comment = False)

src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs
```diff
@@ -280,65 +263,67 @@ protected override Delegate MakeGetter(Row input, int iinfo, Func<int, bool> act
     /// The PlattCalibratorEstimator.
```
Ivanidzo4ka: >The PlattCalibratorEstimator. [](start = 8, length = 29)

/// <summary>
/// Creates Platt calibrator by fitting a logistic regression model to a classifier's scores
///</summary>

Ivanidzo4ka:
>     /// The naive binning-based calibratorEstimator.

two words?

---
Refers to: src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs:350 in f00b206. [](commit_id = f00b206921404ab8c5dfc192665dee6e987b6716, deletion_comment = False)

src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs
```diff
@@ -405,43 +389,42 @@ public sealed class NaiveCalibratorTransformer : CalibratorTransformer<NaiveCali
...
         // Factory method for SignatureLoadModel.
         internal NaiveCalibratorTransformer(IHostEnvironment env, ModelLoadContext ctx)
             : base(env, ctx, LoadName)
         {
-
         }
     }
 
     /// <summary>
     /// The PavCalibratorEstimator.
```
Ivanidzo4ka: >PavCalibratorEstimator [](start = 12, length = 22)

/// <summary>
/// Estimator  to fit <see cref="PavCalibrator"/>
///</summary>

docs/samples/Microsoft.ML.Samples/Dynamic/Calibrator.cs
```diff
@@ -1,4 +1,8 @@
-﻿using System;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
```
Ivanidzo4ka: I don't think we put this header to samples

src/Microsoft.ML.Data/TrainCatalog.cs
TomFinley: I can change "marked as internal" to "with internal scope," but I do not intend to push that change unless I have to update the PR anyway for some other reason. The root of your confusion, as far as I can tell, is not realizing that `private protected` implied the internal scope in its definition. I agree it's confusing, but the XML docs here are definitely *not* the place to resolve this confusion.

TomFinley:
> > ```
> > /// The naive binning-based calibratorEstimator.
> > ```
> 
> two words? or maybe
> 
> ```
> /// <summary>
> /// Estimator  to fit <see cref="NaiveCalibrator"/>
> ///</summary>
> ```
> Refers to: src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs:350 in [f00b206](https://github.com/dotnet/machinelearning/commit/f00b206921404ab8c5dfc192665dee6e987b6716). [](commit_id = [f00b206](https://github.com/dotnet/machinelearning/commit/f00b206921404ab8c5dfc192665dee6e987b6716), deletion_comment = False)

Let's go with two words (same with PAV also)... the information that the model being trained is this or that specific type is something they can just get from the type signature.

My favorite constructor documentation is `public Foo()` is `<summary>Constructor for <see cref="Foo"/>.</summary>`. 😆 Oh thank you doc writer, my cup runneth over.

The documentation in this file is really poor. Maybe we'll be able to fix it up later.

docs/samples/Microsoft.ML.Samples/Dynamic/Calibrator.cs
TomFinley: No idea why it put my reply here when I was replying to your other thing...

src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs
TomFinley: Hmmm. I'll just fix the words a little, but this needs a lot more work than just this. I'm pretty sure those remarks are going to be utterly incomprehensible when written out also. I'm not quite equal to that task.

sfilipi:
>         public static ICalibrator TrainCalibrator(IHostEnvironment env, IChannel ch, ICalibratorTrainer caliTrainer, IPredictor predictor, RoleMappedData data, int maxRows = _maxCalibrationExamples)

if this is public, does the other one need to be public as well?

---
Refers to: src/Microsoft.ML.Data/Prediction/Calibrator.cs:936 in ac3a9e0. [](commit_id = ac3a9e03c9f81f386a3230865b28f02d83ea7014, deletion_comment = False)

src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs
```diff
@@ -49,39 +48,35 @@ public interface ICalibrator
...
-            PredictedLabel = new SchemaShape.Column(DefaultColumnNames.PredictedLabel,
-                SchemaShape.Column.VectorKind.Scalar,
-                BoolType.Instance,
-                false,
-                new SchemaShape(MetadataUtils.GetTrainerOutputMetadata()));
+            if (!string.IsNullOrWhiteSpace(labelColumn))
+                LabelColumn = TrainerUtils.MakeBoolScalarLabel(labelColumn);
+            else
+                env.CheckParam(!calibratorTrainer.NeedsTraining, nameof(labelColumn), "For trained calibrators, " + nameof(labelColumn) + " must be specified.");
+            ScoreColumn = TrainerUtils.MakeR4ScalarColumn(scoreColumn); // Do we fanthom this being named anything else (renaming column)? Complete metadata?
```
sfilipi: >// Do we fanthom this being named anything else (renaming column)? Complete metadata? [](start = 72, length = 85)

want to keep this?

src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs
```diff
@@ -277,68 +260,72 @@ protected override Delegate MakeGetter(Row input, int iinfo, Func<int, bool> act
...
         /// <param name="env">The environment to use.</param>
-        /// <param name="predictor">The predictor used to train the data.</param>
         /// <param name="slope">The slope in the function of the exponent of the sigmoid.</param>
         /// <param name="offset">The offset in the function of the exponent of the sigmoid.</param>
-        /// <param name="labelColumn">The label column name.</param>
-        /// <param name="featureColumn">The feature column name.</param>
-        /// <param name="weightColumn">The weight column name.</param>
+        /// <param name="scoreColumn">The score column name. This is consumed both when this estimator
+        /// is fit and when the estimator is consumed.</param>
         public FixedPlattCalibratorEstimator(IHostEnvironment env,
```
sfilipi: >public [](start = 8, length = 6)

are we leaving the ctors of the calibratorEstimators public?

TomFinley: No. However, until #1871 is completed with the proper catalog for calibrators added, this should remain public. Incidentally this was what I meant by this sentence at the end of my PR description:

> Note that ultimately the trainer estimator constructors _will_ be internal, pending #1871.

src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs
TomFinley: You mean the comment? I didn't quite know what the original author had in mind, so I just fixed the obvious typo and moved on.

TomFinley:
> > ```
> >     public static ICalibrator TrainCalibrator(IHostEnvironment env, IChannel ch, ICalibratorTrainer caliTrainer, IPredictor predictor, RoleMappedData data, int maxRows = _maxCalibrationExamples)
> > ```
> 
> if this is public, does the other one need to be public as well? Do the trainers call that one?

Hi @sfilipi, that's a good point. I don't see anything other than the other utility method that calls it. (Maybe Nimus or the internal tools, who knows what they do.)

This one is kind of public, but it's a public member on an internal class, so I'm not *too* excited about it? Especially since I have that big, big green squash and merge button... tempting me...

rogancarr:
>     public class CalibratorExample

Calibrator

---
Refers to: docs/samples/Microsoft.ML.Samples/Dynamic/Calibrator.cs:11 in ac3a9e0. [](commit_id = ac3a9e03c9f81f386a3230865b28f02d83ea7014, deletion_comment = False)

rogancarr:
>         public static void Calibration()

Example

---
Refers to: docs/samples/Microsoft.ML.Samples/Dynamic/Calibrator.cs:13 in ac3a9e0. [](commit_id = ac3a9e03c9f81f386a3230865b28f02d83ea7014, deletion_comment = False)

docs/samples/Microsoft.ML.Samples/Dynamic/Calibrator.cs
```diff
@@ -75,7 +75,7 @@ public static void Calibration()
 
             // Let's train a calibrator estimator on this scored dataset. The trained calibrator estimator produces a transformer
             // that can transform the scored data by adding a new column names "Probability". 
-            var calibratorEstimator = new PlattCalibratorEstimator(mlContext, model, "Sentiment", "Features");
+            var calibratorEstimator = new PlattCalibratorEstimator(mlContext, "Sentiment", "Score");
```
rogancarr: >var calibratorEstimator = new PlattCalibratorEstimator(mlContext, "Sentiment", "Score"); [](start = 12, length = 88)

Can you give a bit more detail what's happening here? The SDCA trainer produces a transformer, which is calibrated, then you pull the original model out and recalibrate it? If so, this sample is a bit confusing because it's not necessarily something we would need to do, at least not as outlined here. I would at least spell out the steps precisely.

rogancarr:
>         private static void PrintRowViewValues(Data.DataDebuggerPreview data)

This is a bit weird that we can do useful things with the debugger. I'm a bit worried people will use this in production code, not for debugging.

---
Refers to: docs/samples/Microsoft.ML.Samples/Dynamic/Calibrator.cs:98 in ac3a9e0. [](commit_id = ac3a9e03c9f81f386a3230865b28f02d83ea7014, deletion_comment = False)

rogancarr:
>             //Preview of finalData.RowView

Expected output:

---
Refers to: docs/samples/Microsoft.ML.Samples/Dynamic/Calibrator.cs:88 in ac3a9e0. [](commit_id = ac3a9e03c9f81f386a3230865b28f02d83ea7014, deletion_comment = False)

rogancarr:
>             // Score - 0.458968    Probability 0.4670409

No break, two spaces indent

---
Refers to: docs/samples/Microsoft.ML.Samples/Dynamic/Calibrator.cs:90 in ac3a9e0. [](commit_id = ac3a9e03c9f81f386a3230865b28f02d83ea7014, deletion_comment = False)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -85,14 +86,24 @@ namespace Microsoft.ML.Internal.Calibration
...
 
+    [BestFriend]
     [TlcModule.ComponentKind("CalibratorTrainer")]
-    public interface ICalibratorTrainerFactory : IComponentFactory<ICalibratorTrainer>
+    internal interface ICalibratorTrainerFactory : IComponentFactory<ICalibratorTrainer>
     {
     }
 
-    public interface ICalibratorTrainer
+    /// <summary>
```
rogancarr: >/// <summary> [](start = 4, length = 13)

Technically, this clarification is a <remark> and <summary> is a short description of what it does.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -842,6 +864,64 @@ internal static class CalibratorUtils
             return CreateCalibratedPredictor(env, (IPredictorProducing<float>)predictor, trainedCalibrator);
         }
 
+        public static ICalibrator TrainCalibrator(IHostEnvironment env, IChannel ch, ICalibratorTrainer caliTrainer, IDataView scored, string labelColumn, string scoreColumn, string weightColumn = null, int maxRows = _maxCalibrationExamples)
```
rogancarr: >string labelColumn, string scoreColumn, string weightColumn = null [](start = 135, length = 66)

Defaults?

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -842,6 +864,64 @@ internal static class CalibratorUtils
...
+        {
+            Contracts.CheckValue(env, nameof(env));
+            env.CheckValue(ch, nameof(ch));
+            ch.CheckValue(scored, nameof(scored));
+            ch.CheckValue(caliTrainer, nameof(caliTrainer));
+            ch.CheckParam(!caliTrainer.NeedsTraining || !string.IsNullOrWhiteSpace(labelColumn), nameof(labelColumn),
```
rogancarr: >!caliTrainer.NeedsTraining || !string.IsNullOrWhiteSpace(labelColumn) [](start = 26, length = 69)

Nit: This is a bit confusing to read. Maybe ! the And version?

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -842,6 +864,64 @@ internal static class CalibratorUtils
...
+                "If " + nameof(caliTrainer) + " requires training, then " + nameof(labelColumn) + " must have a value.");
+            ch.CheckNonWhiteSpace(scoreColumn, nameof(scoreColumn));
+
+            if (!caliTrainer.NeedsTraining)
+                return caliTrainer.FinishTraining(ch);
+
+            var labelCol = scored.Schema[labelColumn];
+            var scoreCol = scored.Schema[scoreColumn];
+
+            var weightCol = weightColumn == null ? null : scored.Schema.GetColumnOrNull(weightColumn);
```
rogancarr: >== null ? null : [](start = 41, length = 16)

??

rogancarr: >weightColumn [](start = 28, length = 12)

I'm confused -- isn't this a string, and won't the GetColumnOrNull return a Column object?

rogancarr:
Oh, nevermind. Saw that you already merged this :)

src/Microsoft.ML.Data/TrainCatalog.cs
glebuk: In any case, merged https://github.com/dotnet/dotnet-api-docs/pull/1844 to clarify .NET docs.

PR #2532, opened 2019-02-13
Lockdown Microsoft.ML.EntryPoints public surface

### LINKED ISSUE #2265: Lockdown Microsoft.ML.EntryPoints public surface
We ought to internalize as much of the Lockdown Microsoft.ML.EntryPoints assembly to those things we want to be public.

This may be considered less urgent if we take the decision that the associated EntryPoints nuget is to be marked as experimental for v1.

yaeldekel:
Fixes #2265 .


Ivanidzo4ka:
https://github.com/dotnet/machinelearning/blob/656a42fd0c4f12112a05c40e9a15f7595157ee50/src/Microsoft.ML.EntryPoints/TrainTestMacro.cs#L298
can we remove that weird lonely pragma?

Ivanidzo4ka:
Thank you!

Issue #1204, opened 2018-10-09
Classification stratificationColumn not supported for boolean column


CESARDELATORRE:
For not balanced datasets, with stratified splitting, the data is divided in such a way that a percentage of each target column value is put in both training and test dataset.

However, the following line of code throws an error if the column 'Label' is Boolean, which is very common for binary classification.

`(trainData, testData) = classification.TrainTestSplit(data, testFraction: 0.2, stratificationColumn: "Label");`

It would work if the Label column would be float or other types.

I might be missing something, but why is Boolean not supported for the `stratificationColumn`?
Can we support it since it can be a common scenario for binary classifications?

Zruty0:
Stratification column is not what you think. Actually, if you read the documentation, it [states](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Training/TrainingStaticExtensions.cs#L31) that 

>If two examples share the same value of the `stratificationColumn` (if provided), they are guaranteed to appear in the same subset (train or test). Use this to make sure there is no label leakage from train to the test set

As you can see, it is ALWAYS a bad idea to use `Label` as a stratification column, and ALWAYS a bad idea to use a Boolean column as a stratification column.

That said, currently we only support float, key and string values for stratification (and I'm not sure about key). We should expand the coverage of Hash estimator (see #1031)

CESARDELATORRE:
Just highlighting that this issue still applies. 
The stratification column parameter was renamed to 'samplingKeyColumn' because stratification column means a very different thing:
**Stratification Column**: A Stratification column is important if the rows in the dataset should be divided evenly by some key column when sampling or producing split datasets, so it is a very different concept.

On the other hand, 'samplingKeyColumn' is used for grouping rows. If two examples share the same value of the samplingKeyColumn, they are guaranteed to appear in the same subset (train or test). This can be used to ensure no label leakage from the train to the test set. If null no row grouping will be performed.

So, the usage is very different. But that is not the issue/limitation.

The BUG/ISSUE is that looks like if I have a dataset where the Label is Boolean and there are few "True" rows, I'm finding that the test/train split is not stratifying as I'm getting a test dataset with all labels=false. That means it is not divided evenly by the label key column when sampling or producing split datasets.

Additionally, looks like we have no parameter to specify the "stratification column"?

codemzs:
I agree with Pete. Closing this.

PR #2537, opened 2019-02-13
Rename CV and TrainTest "stratification" parameter

### LINKED ISSUE #2536: StratificationColumn in CrossValidation and TrainTestSplit
`CrossValidation` and `TrainTestSplit` have a parameter called `StratificationColumn` that is used to preserve groupings of columns across splits (as discussed in #2487). This isn't actually stratification, so we should rename the column.

This is a forked sub-issue from #2487

Related to #1204

rogancarr:
This PR changes the `CrossValidation` and `TrainTest` parameter `StratificationColumn` to be `GroupPreservationColumn` and updates the docstrings to give a clearer explanation.

Fixes #2536 

See conversation in #2536 

Related to #1204, but that issue might be asking for further coverage.


rogancarr:
Something has gone weird with my master branch, and I'm seeing every merge from upstream as a commit. I'll look into this, but let's work with it for now ;)

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -48,34 +48,32 @@ internal TrainTestData(IDataView trainSet, IDataView testSet)
...
-        /// If this optional parameter is not provided, a stratification columns will be generated, and its values will be random numbers .</param>
-        /// <param name="seed">Optional parameter used in combination with the <paramref name="stratificationColumn"/>.
-        /// If the <paramref name="stratificationColumn"/> is not provided, the random numbers generated to create it, will use this seed as value.
-        /// And if it is not provided, the default value will be used.</param>
-        public TrainTestData TrainTestSplit(IDataView data, double testFraction = 0.1, string stratificationColumn = null, uint? seed = null)
+        /// <param name="idColumn">Name of a column to use as an ID for grouping rows. If two examples share the same value of the <paramref name="idColumn"/>,
+        /// they are guaranteed to appear in the same subset (train or test). This can be used to ensure no label leakage from train to the test set.
+        /// If <see langword="null"/> no row grouping will be performed.</param>
+        /// <param name="seed">Seed for the random number generator used to select rows for the train-test split.</param>
```
najeeb-kazmi: `seed` isn't actually doing this here. Looks like it is used only in `EnsureIdColumn`. Otherwise, we just return the first `testFraction` rows as the test set, and the complement as the train set. 

We should be doing train test splits with a seed where the split is random based on the seed provided, but given current behavior, the presence of this argument (or at least the description) would cause confusion to the user.

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -48,34 +48,32 @@ internal TrainTestData(IDataView trainSet, IDataView testSet)
...
-        /// <param name="stratificationColumn">Optional name of the column to use as a stratification column. If two examples share the same value of the <paramref name="stratificationColumn"/>
-        /// (if provided), they are guaranteed to appear in the same subset (train or test). Use this to make sure there is no label leakage from train to the test set.
-        /// If this optional parameter is not provided, a stratification columns will be generated, and its values will be random numbers .</param>
-        /// <param name="seed">Optional parameter used in combination with the <paramref name="stratificationColumn"/>.
-        /// If the <paramref name="stratificationColumn"/> is not provided, the random numbers generated to create it, will use this seed as value.
-        /// And if it is not provided, the default value will be used.</param>
-        public TrainTestData TrainTestSplit(IDataView data, double testFraction = 0.1, string stratificationColumn = null, uint? seed = null)
+        /// <param name="groupPreservationColumn">Name of a column to use as an ID for grouping rows. If two examples share the same value of the <paramref name="groupPreservationColumn"/>,
```
Ivanidzo4ka: I think you left that description from previous iteration with `idColumn`.
Do you want to rephrase it, since `ID` looks weird.

TomFinley:
> Something has gone weird with my master branch, and I'm seeing every merge from upstream as a commit. I'll look into this, but let's work with it for now ;)

You commited *directly* into `master` branch on your own fork about a week ago, somehow didn't catch it, pushed that cahnge, and you've been doing pulls ever since.

Once you're ready to restore your master branch, the easiest way to do this would be to `git reset --hard upstream/master` (assuming you set up the original remote name as `upstream`, as most people here seem to, if the original source repo remote you gave some other name, obviously replace that), which will update your local branches. Then `git push -f` to force push that update the master on your origin branch. Thenceforth remember that commiting directly to master is a really bad idea.

src/Microsoft.ML.Data/TrainCatalog.cs
TomFinley: Also ID is already a thing on IDs, naming something different that would be super confusing.

https://github.com/dotnet/machinelearning/blob/1f29d2ca32ccdb4a504cdc8de668daf91e24b448/src/Microsoft.Data.DataView/IDataView.cs#L132

rogancarr:
> You commited _directly_ into `master` branch on your own fork about a week ago, somehow didn't catch it, pushed that cahnge, and you've been doing pulls ever since.
> 
> Once you're ready to restore your master branch, the easiest way to do this would be to `git reset --hard upstream/master` (assuming you set up the original remote name as `upstream`, as most people here seem to, if the original source repo remote you gave some other name, obviously replace that), which will update your local branches. Then `git push -f` to force push that update the master on your origin branch. Thenceforth remember that commiting directly to master is a really bad idea.

Yes. I fixed that a couple days ago, but this PR still has the remnants :)

PR #2543, opened 2019-02-14
Exclude SamplesUtils from ML.NET nuget and make its own nuget

### LINKED ISSUE #2277: Lockdown Microsoft.ML.SamplesUtils public surface
We ought to internalize as much of that assembly to those things we want to be public, and finalize its public API.

This is an odd one. In spirit it seems similar to [similar utilites from sklearn](https://scikit-learn.org/stable/datasets/index.html), which are valuable. But, should it be part of the main nuget? Is this public API surface what we really want? It seems like maybe making it an extension on top of `DataOperationsCatalog` or somesuch might be a better choice. And should it be in a separate Nuget since its functionality is just to make some samples be self contained? (Certainly our samples could reference another nuget, since this one is so odd.)

These existential questions may be best answered by @sfilipi .

codemzs:
fixes #2277


eerhardt:
Do we need to update any docs to tell people to install this new `Microsoft.ML.SampleUtils` nuget package in order to call `DatasetUtils` functions?

PR #2544, opened 2019-02-14
Lockdown Microsoft.ML.OnnxTransform public surface, possibly rename 

### LINKED ISSUE #2272: Lockdown Microsoft.ML.OnnxTransform public surface, possibly rename
We ought to internalize as much of that assembly to those things we want to be public.

Also, I wonder if the assembly and associated nuget (both `Microsoft.ML.OnnxTransform`) is misnamed. We do not have "transforms," we have "transformers."

codemzs:
fixes #2272

There isn't anything obvious to lockdown and the only task I see here is the rename suggestion by @TomFinley


Ivanidzo4ka:
We have TensorFlow nuget but this one is OnnxTransform(OnnxTransformer). Functionality wise they do same thing, one score TensorFlow, second scores Onnx model. Are we happy with having different names for them?
I know we have Onnx nuget which is why we call this one OnnxTransform, but our current Onnx nuget should be called OnnxExporter or OnnxConverter, it doesn't deserve to be called Onnx. 

I know @TomFinley  propose to call this one OnnxTransformer, but I think it's because:
`We do not have "transforms," we have "transformers."`, not because it should have that exact name.

@glebuk @zeahmed @jignparm @TomFinley  ?

Ivanidzo4ka:
>         public IHostEnvironment Environment { get; }

why public?

---
Refers to: src/Microsoft.ML.OnnxTransformer/DnnImageFeaturizerTransform.cs:28 in 93c0ffa. [](commit_id = 93c0ffa0d55f5c587ddfd0555836998cec3c592e, deletion_comment = False)

Ivanidzo4ka:
>         public DnnImageFeaturizerInput(string outputColumnName, string inputColumnName, IHostEnvironment env, DnnImageModelSelector modelSelector)

why constructor is public?

---
Refers to: src/Microsoft.ML.OnnxTransformer/DnnImageFeaturizerTransform.cs:33 in 93c0ffa. [](commit_id = 93c0ffa0d55f5c587ddfd0555836998cec3c592e, deletion_comment = False)

TomFinley:
> We have TensorFlow nuget but this one is OnnxTransform(OnnxTransformer). Functionality wise they do same thing, one score TensorFlow, second scores Onnx model. Are we happy with having different names for them?
> I know we have Onnx nuget which is why we call this one OnnxTransform, but our current Onnx nuget should be called OnnxExporter or OnnxConverter, it doesn't deserve to be called Onnx.
> 
> I know @TomFinley propose to call this one OnnxTransformer, but I think it's because:
> `We do not have "transforms," we have "transformers."`, not because it should have that exact name.
> 
> @glebuk @zeahmed @jignparm @TomFinley ?

So, here is my thinking.

We used to have `Onnx` nuget package, that handled ONNX model conversion and export. Then we introduce the nuget that is the subject of this PR, `OnnxTransform`. If we accept the reasoning that these things are to be separate nugets (which seems to me a defensible choice, given the different requirements), then we need to distinguish these names. That we arrived at this seems to me natural: the facilities needed for *exporting* a model as ONNX are rather different than those required to actually run an ONNX model.

Tensorflow is different. What is tensorflow? It is not primarily a model format as such. It is rather some framework, into which we can interface. I do not relate tensorflow to ONNX; they just both have something to do with DNNs! I relate tensorflow to LightGBM and FAFMs and so forth, insofar as we're interacting with external trainers. If we had multiple interfaces to Tensorflow due to different requirements.

So I might say that if we wanted ONNX to be consistent, we could name this package just "ONNX," and renaming what is currently called ONNX to something else (as is in fact done in #2625). But I might also say since the only thing you can really do to onnx is *apply* it to data (unlike Tensorflow!), that naming it `OnnxTransformer` certainly seems fine to me.

So I am fine with this change. We can certainly agree I hope that changing the name from `transform` to `transformer` is more appropriate? So I view this as a good change. But like all things naming, my opinion is quite subjective.

PR #2547, opened 2019-02-14
DataView type rename

eerhardt:
This covers the first 3 changes proposed in #2297.

1. We will ensure the class names in the `Microsoft.Data.DataView` package are unique by adding a prefix to any name deemed too general. Add a `DataView` prefix.
2. Add the base type suffix to all types that derive from the current `ColumnType` class.
3. Rename the properties on NumberType to match the type name in `System` namespace.

The rest of the proposed changes will come in a separate PR (this one is big enough).


src/Microsoft.ML.Data/Commands/DataCommand.cs
```diff
@@ -170,9 +170,9 @@ protected void SendTelemetryMetric(Dictionary<string, IDataView>[] metricValues)
                             {
                                 var nameOfMetric = "TLC_" + cursor.Schema[currentIndex].Name;
```
shauheen: >TLC_ [](start = 52, length = 4)

Do we need to change this? not that I protest! :)

eerhardt: Change the "TLC" part? There are still a bunch of things with "TLC" in them... Maybe open an issue to do a mass "TLC" removal/rename?

Note this is part of the "telemetry" code, which is only invoked from `Command` classes, which are only invoked from MAML, as far as I can tell.

PR #2506, opened 2019-02-11
Typed SDCA binary trainers

wschin:
Make `SdcaBinaryTrainer` strongly-typed according to the type it produces. The existing `SdcaBinaryTrainer` can produce either calibrated linear model or linear model as mentioned in #2469. To fix #2469, we move common functionalities used in both cases to `SdcaBinaryTrainerBase<T>` where `T` is `CalibratedModelParametersBase<LinearBinaryModelParameters, PlattCalibrator` for calibrated case and `LinearBinaryModelParameters` otherwise. Top-level APIs are changed accordingly. For dynamic/static APIs, we have 4 SDCA binary trainers for

1. Calibrated linear model with simple arguments.
2. Calibrated linear model with advanced options.
3. Uncalibrated linear model with simple arguments.
4. Uncalibrated linear model with advanced options.

In addition, because we don't like auto-calibration, the output schema in uncalibrated cases should not contain a probability column. This PR also fixes this in the two derived trainer classes' `ComputeSdcaBinaryClassifierSchemaShape`; the only case with a probability column generated is training logistic regression with Sdca. We also have two separated `SdcaCalibratedBinaryTrainer.Options` and `SdcaBinaryTrainer.Options` derived from `BinaryArgumentBase` and therefore entry-points are changed too.

PR #2548, opened 2019-02-14
[Tiny & Simple Change] Example classes should be static

### LINKED ISSUE #2549: Example classes should be static
We have accumulated a bunch of examples while our coding guideline changes from time to time. We should as least make all example classes static to meet our latest decision.

wschin:
Polish DYNAMIC examples to fix #2549. This change is inspired by @TomFinley's good comment to #2506 and I sincerely want to address that.

- Example class should be static
- Example function should be called `Example()`
- Remove redundant `using` statements.

Please note that this PR only focuses on dynamic examples' `static` problem. I don't have much bandwidth to take care other cases.

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/IidSpikeDetectorTransform.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/IidSpikeDetectorTransform.cs
@@ -6,7 +6,7 @@ using Microsoft.ML.Transforms.TimeSeries;
 
 namespace Microsoft.ML.Samples.Dynamic
 {
-    public partial class TransformSamples
+    public static partial class TransformSamples
     {
         class IidSpikeData
         {
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/OnnxTransform.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/OnnxTransform.cs
@@ -6,12 +6,12 @@ using Microsoft.ML.Transforms;
 
 namespace Microsoft.ML.Samples.Dynamic
 {
-    class OnnxTransformExample
+    public static class OnnxTransformExample
     {
         /// <summary>
         /// Example use of OnnxEstimator in an ML.NET pipeline
         /// </summary>
-        public static void OnnxTransformSample()
+        public static void Example()
         {
             // Download the squeeznet image model from ONNX model zoo, version 1.2
             // https://github.com/onnx/models/tree/master/squeezenet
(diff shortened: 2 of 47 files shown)


docs/samples/Microsoft.ML.Samples/Dynamic/OnnxTransform.cs
```diff
@@ -6,12 +6,12 @@
 
 namespace Microsoft.ML.Samples.Dynamic
 {
-    class OnnxTransformExample
+    public static class OnnxTransformExample
```
eerhardt: (nit)  you've dropped the `Example` suffix in other places, but not here.

docs/samples/Microsoft.ML.Samples/Dynamic/ValueMapping.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.ML.Samples.Dynamic
 {
-    public class ValueMappingExample
+    public static partial class ValueMapping
```
Ivanidzo4ka: >partial [](start = 18, length = 7)

(nit) it doesn't have to be partial

docs/samples/Microsoft.ML.Samples/Dynamic/IidSpikeDetectorTransform.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.ML.Samples.Dynamic
 {
-    public partial class TransformSamples
+    public static partial class TransformSamples
```
Ivanidzo4ka: >TransformSamples [](start = 32, length = 16)

All timeseries example slightly different from our current samples.
They share same class and introduce data classes inside it, instead of using sample utils.
But I'm not sure they transformation to how we see samples should be done in your PR.

docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/PriorTrainerSample.cs
```diff
@@ -1,11 +1,9 @@
 ﻿using System;
-using System.Linq;
 using Microsoft.ML.Data;
-using Microsoft.ML.Trainers;
 
 namespace Microsoft.ML.Samples.Dynamic
 {
-    public class PriorTrainerSample
+    public class PriorTrainer
```
Ivanidzo4ka: >PriorTrainer [](start = 17, length = 12)

if you decide to update PR for nit comments, can you move this class and RandomTrainer to BinaryClassification folder? (If you decide to do it, don't forget to update catalog comments which looks on this file)

Ivanidzo4ka:
Thank you for doing this @wschin !

Issue #1287, opened 2018-10-17
LocalEnvironment to stop being IDisposable


Zruty0:
The only reason we have `IHostEnvironment` as an `IDisposable` is because we clean up the temp files created using `CreateTempFile`. This functionality is not really very useful. We should remove the functionality, remove the `IDisposable` and fix all the `using` calls in the codebase (preferably to just `new MLContext`, if possible)

TomFinley:
The implication is that anything that relies on it for temporary file creation and that requires must cease to do so. In order for us to do this, we must also remove temporary file creation from `IHostEnvironment`. (Also, regular file creation, though that is less urgent.) This I have found untangling this *relatively* easy to do, but there is one exception that is very troublesome because it explicitly relies on this behavior.

https://github.com/dotnet/machinelearning/blob/057c4b97bb4fab09b8f88db2300644660f46aa5d/src/Microsoft.ML.Data/EntryPoints/Cache.cs#L71-L73

That's bad, but I think there's a pretty easily solution. I think it is completely reasonable for someone asking for disk cache to provide a place *where* on the disk it should be stored. I think that will actually be an improvement -- certainly it is not an expectation that any "temporary" files will be many gigabytes or terabytes large (which is the only conceivable reason why anyone would use disk caching in the first place), so having some ability to configure *where* it is stored would I think be good. Or we provide access to the binary saver via entry-points and say that if someone wants that functionality they just have to do it themselves. This I think is fine, since entry-points were entry-points into ML.NET for *wrapping* code, not *user facing* code.

eerhardt:
Yes, this is still relevant.

We still have these methods on `IHostEnvironment` that we need to remove:

https://github.com/dotnet/machinelearning/blob/0c62e30b4d9eabb60322b2a3e75bc90e20007889/src/Microsoft.ML.Core/Data/IHostEnvironment.cs#L55-L77

Even though it doesn't implement `IDisposable`, we can't have these methods on this interface.

eerhardt:
Just an FYI - it looks like the only place in the product that is calling `Cache.CacheData` (which is the only place calling `CreateTempFile` pointed out by @TomFinley above) is coming from Entry-Points:

https://github.com/dotnet/machinelearning/blob/174de1a8dedb7bd72e1f43df6f07b9392c42c290/src/Microsoft.ML.Data/EntryPoints/InputBase.cs#L220-L224

[...]

PR #2539, opened 2019-02-13
Remove OpenInputFile and CreateOutputFile methods from IHostEnvironment

maryamariyan:
Didn't add new tests since the implementation is just moved from HostEnvironmentBase to HostEnvironmentUtils class.

On a separate PR will remove CreateTempFile which still part of IHostEnvironment interface.

Related to: #1287
cc: @eerhardt 

- [x] There's a descriptive title that will make sense to other developers some time from now. 
- [x] There's associated issues. All PR's should have issue(s) associated - unless a trivial self-evident change such as fixing a typo. You can use the format `Fixes #nnnn` in your description to cause GitHub to automatically close the issue(s) when your PR is merged.
- [x] Your change description explains what the change does, why you chose your approach, and anything else that reviewers should know.
- [ ] You have included any necessary tests in the same PR.

PR #2568, opened 2019-02-15
Removes IHostEnvironment.CreateTempFile

maryamariyan:
This PR is effectively removing CreateTempFile from IHostEnvironment and moving it to helper class and expects the caller to dispose of the handle safely instead.

Issue ask was to Remove Dispose from LocalEnvironment. The Dispose would have been useful if the class wanted to Dispose of the file handles created using CreateTempFile which is not expected to be the responsibility of LocalEnvironment (or ConsoleEnvironment) anymore. 

Fixes: #1287 
cc: @eerhardt 

---


src/Microsoft.ML.Data/EntryPoints/Cache.cs
```diff
@@ -34,7 +34,7 @@ public sealed class CacheOutput
         }
 
         [TlcModule.EntryPoint(Name = "Transforms.DataCache", Desc = "Caches using the specified cache option.", UserName = "Cache Data")]
-        public static CacheOutput CacheData(IHostEnvironment env, CacheInput input)
+        public static CacheOutput CacheData(IHostEnvironment env, CacheInput input, IFileHandle fileHandle)
```
maryamariyan: _From @eerhardt on Feb 14, 2019, 7:59 AM PST_

Since this is an `EntryPoint` itself, I don't know if we can add parameters to it. I'm not 100% exactly familiar with how EntryPoints work.

@ganik ?

_From @eerhardt on Feb 14, 2019, 8:10 AM PST_

Maybe it could work something like this:

If disk caching was specified the `CacheOutput` will have the path to the file. It is then up to the caller to clean the file up when they are done with it.

Optionally, if we feel it necessary, the `CacheInput` can take a path to where on the file system to put the cached file. But maybe that could be left up for a separate change in the future.

Thoughts?

---
In reply to: [256896499](https://github.com/dotnet/machinelearning/pull/2539#discussion_r256896499) [](ancestors = 256896499)

maryamariyan: moved from https://github.com/dotnet/machinelearning/pull/2539#discussion_r256896499

TomFinley:
I see this PR seems to be updating `src/Native/MatrixFactorizationNative/libmf`. Is that intended?

maryamariyan:
Good catch. no I need to remove diff to `src/Native/MatrixFactorizationNative/libmf`

TomFinley:
I know that the changes to the cache entry-point are not correct for various reasons, but I think the problem is that the cache entry-point and its fundamental needs that were addressed by it needs to be redesigned and reconsidered in a way that cannot be captured in this PR, but should be instead captured by whoever introduced it in the first place so they can design something proper -- you and I don't have the insight necessary to do that. This suggests to me deleting the entry-point *altogether* is the right choice, so that something proper can be put in its place.

You will observe when you do that that you'll have to update the entry-point manifest baseline file. (The test will fail.)

TomFinley:
When I say redesigned, let me just say I think the cache entry-point specifically should be removed as part of this PR. I don't think it is appropriate for a machine learning API to be responsible for the tracking and management of files on disk in a situation like this where it can clearly be avoided by just having client API code take care of it.

maryamariyan:
I pushed up another commit updating core_manifest.json to come with a working diff.

Next I'm going to go over the feedback by @TomFinley

maryamariyan:
As @TomFinley asked I am removing cache entry-point

src/Microsoft.ML.Data/Commands/TrainCommand.cs
```diff
@@ -70,9 +70,6 @@ public sealed class Arguments : DataCommand.ArgumentsBase
             [Argument(ArgumentType.AtMostOnce, IsInputFileName = true, HelpText = "The test data file", ShortName = "test")]
             public string TestFile;
 
-            [Argument(ArgumentType.LastOccurenceWins, HelpText = "Whether we should cache input training data", ShortName = "cache")]
-            public bool? CacheData;
```
eerhardt: I think removing the `Transforms.DataCache` entry point is fine, but I don't think we want to remove this option from our training commands, do we?

test/BaselineOutput/Common/EntryPoints/core_manifest.json
```diff
@@ -17869,6 +17869,11 @@
           "Name": "OutputData",
           "Type": "DataView",
           "Desc": "Dataset"
+        },
```
eerhardt: I think this is out of date?

src/Microsoft.ML.Data/Commands/TrainCommand.cs
TomFinley: The commands are self contained applications. It's just fine for them to cache.

To be clear, what I was talking about was removing the *cache entry point specifically*, and that is all. I was not suggesting anything broader or more ambitious than this. We *still want this command to have an option to cache*.

src/Microsoft.ML.Data/Commands/TrainCommand.cs
```diff
@@ -487,28 +482,6 @@ IDataView ApplyNormalizer(IHostEnvironment innerEnv, IDataView input)
             return false;
         }
 
-        private static bool AddCacheIfWanted(IHostEnvironment env, IChannel ch, ITrainer trainer, ref RoleMappedData data, bool? cacheData)
```
TomFinley: >AddCacheIfWanted [](start = 28, length = 16)

Please add this back, thanks./

maryamariyan:
Resolved issues. PR is ready for review

src/Microsoft.ML.Data/EntryPoints/InputBase.cs
```diff
@@ -188,25 +187,26 @@ public static string FindColumn(IExceptionContext ectx, DataViewSchema schema, O
                 var roleMappedData = new RoleMappedData(view, label, feature, group, weight, name, custom);
 
                 RoleMappedData cachedRoleMappedData = roleMappedData;
-                Cache.CachingType? cachingType = null;
+                IFileHandle fileHandle = null;
```
eerhardt: `fileHandle` appears to always be set to `null`, and never assigned a value.

maryamariyan: Good catch. I was using it in the CachingOptions.Disk switch case where I then removed.

maryamariyan: Opened https://github.com/dotnet/machinelearning/pull/2759 to address this

PR #2569, opened 2019-02-15
Stop using System.ComponentModel.Composition

### LINKED ISSUE #1595: Sentiment Analysis on Uwp - MissingMethodException
### System information

- **OS version/distro**: Windows 10 UWP app,   ML 0.7.0
- **.NET Version (eg., dotnet --info)**:  Net standard / Targeting min Fall Creator Update

### Issue

- **What did you do?**
Was trying to run  binary classification example on Uwp.

- **What happened?**

System.MissingMethodException: Method not found: 'System.ComponentModel.Composition.Hosting.CompositionContainer Microsoft.ML.Runtime.IHostEnvironment.GetCompositionContainer()'.

It works fine with  .Net Core Console .

### LINKED ISSUE #2422: Stop using MEF as part of the public API
We have received feedback that we should limit our usage of MEF as part of the public API, according to feedback by .NET team. We currently for the sake of writing "custom" transformers have the user interact with it. This can be seen here.

https://github.com/dotnet/machinelearning/blob/2d351eb73855cdc30e9c978c00820470cb38cf39/test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamplesDynamicApi.cs#L535

As we see, we are hsing on `MLContext` MEF containers and catalogs, using types exposing MEF `Import` and `Export` attributes, etc. etc. The feedback from .NET team is, stop, do something else, because MEF is apparently not "clean" enough.

(What is to be done instead? Possibly registering some "named" delegate to produce these things, or other. But anyway, we should not use MEF directly. Maybe not use attributes at all, and just require explicit registration somehow. I'm not sure.)

@eerhardt and others may have more specific recommendations or context on why we should not use MEF in this way.

eerhardt:
Replace our MEF usage, which is only used by custom mapping transforms, with the ComponentCatalog class.

Fix #1595
Fix #2422

### DIFF
--- a/src/Microsoft.ML.Transforms/CustomMappingTransformerFactory.cs
+++ b/src/Microsoft.ML.Transforms/CustomMappingTransformerFactory.cs
@@ -1,0 +1,44 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Microsoft.Data.DataView;
+
+namespace Microsoft.ML.Transforms
+{
+    /// <summary>
+    /// Place this attribute onto a type to cause it to be considered a custom mapping transformer factory.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
+    public sealed class CustomMappingTransformerFactoryAttribute : ExtensionBaseAttribute
+    {
+        public CustomMappingTransformerFactoryAttribute(string contractName)
+            : base(contractName)
+        {
+        }
+    }
+
+    internal interface ICustomMappingTransformerFactory
+    {
+        ITransformer CreateTransformerObject(IHostEnvironment env, string contractName);
+    }
+
+    /// <summary>
+    /// The base type for custom mapping transformer factories.
+    /// </summary>
+    /// <typeparam name="TSrc">The type that describes what 'source' columns are consumed from the input <see cref="IDataView"/>.</typeparam>
+    /// <typeparam name="TDst">The type that describes what new columns are added by this transform.</typeparam>
+    public abstract class CustomMappingTransformerFactory<TSrc, TDst> : ICustomMappingTransformerFactory
+        where TSrc : class, new()
+        where TDst : class, new()
+    {
+        public abstract Action<TSrc, TDst> GetTransformer();
+
+        ITransformer ICustomMappingTransformerFactory.CreateTransformerObject(IHostEnvironment env, string contractName)
+        {
+            Action<TSrc, TDst> mapAction = GetTransformer();
+            return new CustomMappingTransformer<TSrc, TDst>(env, mapAction, contractName);
+        }
+    }
+}
--- a/src/Microsoft.ML.Transforms/LambdaTransform.cs
+++ b/src/Microsoft.ML.Transforms/LambdaTransform.cs
@@ -68,13 +68,13 @@ namespace Microsoft.ML.Transforms
 
             var contractName = ctx.LoadString();
 
-            object factoryObject = env.ComponentCatalog.GetExtensionValue(env, typeof(CustomMappingTransformerFactoryAttribute), contractName);
-            if (!(factoryObject is ICustomMappingTransformerFactory transformerFactory))
+            object factoryObject = env.ComponentCatalog.GetExtensionValue(env, typeof(CustomMappingFactoryAttributeAttribute), contractName);
+            if (!(factoryObject is ICustomMappingFactory mappingFactory))
             {
-                throw env.Except($"The class with contract '{contractName}' must derive from '{typeof(CustomMappingTransformerFactory<,>).FullName}'.");
+                throw env.Except($"The class with contract '{contractName}' must derive from '{typeof(CustomMappingFactory<,>).FullName}'.");
             }
 
-            return transformerFactory.CreateTransformerObject(env, contractName);
+            return mappingFactory.CreateTransformer(env, contractName);
         }
 
         /// <summary>
--- a/src/Microsoft.ML.Transforms/LambdaTransform.cs
+++ b/src/Microsoft.ML.Transforms/LambdaTransform.cs
@@ -68,11 +68,13 @@ namespace Microsoft.ML.Transforms
 
             var contractName = ctx.LoadString();
 
-            var composition = env.GetCompositionContainer();
-            if (composition == null)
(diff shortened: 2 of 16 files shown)


src/Microsoft.ML.Transforms/CustomMappingTransformerFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        ITransformer CreateTransformerObject(IHostEnvironment env, string contractName);
+    }
+
+    /// <summary>
+    /// The base type for custom mapping transformer factories.
+    /// </summary>
+    /// <typeparam name="TSrc">The type that describes what 'source' columns are consumed from the input <see cref="IDataView"/>.</typeparam>
+    /// <typeparam name="TDst">The type that describes what new columns are added by this transform.</typeparam>
+    public abstract class CustomMappingTransformerFactory<TSrc, TDst> : ICustomMappingTransformerFactory
```
eerhardt: One of the things that I'm not 100% satisfied with yet is the names I used here. I wonder if we should drop the `er` and just be `CustomMappingTransformFactory` and `public abstract Action<TSrc, TDst> GetTransform();`.

One reason for this is that the things this creates aren't actually `Transformer`s, but instead just actions that the actual `Transformer` will call.

TomFinley: We've elsewhere called the things that aren't transformers but the simpler things that transformers *call* to do the transformation mappers... e.g., `IRowToRowMapper`, `ISchemaBindableMapper`, `IValueMapper`... now granted many of these are not public concepts and some of them don't quite map onto what you're doing, but if we think about this, we might consider that the thing you're creating in this factory is not the `CustomMappingTransformer`, but just the mapping. So maybe `ICustomMappingFactory` or `ICustomMapperFactory` might be the better choice, as it is more consistent with what we do elsewhere? Not insisting on that particular solution, just wondering if that helps clear your logjam to know how I would rename it.

However, just to be clear, while I don't insist on the above solution, I do feel strongly there needs to be a different name to what it is now. Calling this a "transformer factory" with the abstract method `GetTransformer` is inappropriate... `Transformer` is already a very well established central concept in this library, and this method definitely is *not* a factory for them, and must not advertise itself as such.

---
In reply to: [257331335](https://github.com/dotnet/machinelearning/pull/2569#discussion_r257331335) [](ancestors = 257331335)

src/Microsoft.ML.Transforms/LambdaTransform.cs
```diff
@@ -68,9 +68,13 @@ private static ITransformer Create(IHostEnvironment env, ModelLoadContext ctx)
 
             var contractName = ctx.LoadString();
 
-            var composition = env.GetCompositionContainer();
-            ITransformer transformer = composition.GetExportedValue<ITransformer>(contractName);
-            return transformer;
+            object factoryObject = env.ComponentCatalog.GetExtensionValue(env, typeof(CustomMappingTransformerFactoryAttribute), contractName);
+            if (!(factoryObject is ICustomMappingTransformerFactory transformerFactory))
+            {
+                throw env.Except($"The class with contract '{contractName}' must derive from '{typeof(CustomMappingTransformerFactory<,>).FullName}'.");
```
TomFinley: >CustomMappingTransformerFactory [](start = 102, length = 31)

Custom mappings, at least historically within Microsoft, have been *very* popular, so eventually we'll want to probably make this sort of rule part of our code analyzer. (Obviously not right now! Just something I'd like to think about for post v1.)

src/Microsoft.ML.Transforms/CustomMappingTransformerFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+        where TSrc : class, new()
+        where TDst : class, new()
+    {
+        public abstract Action<TSrc, TDst> GetTransformer();
```
TomFinley: >GetTransformer [](start = 43, length = 14)

You may be disinclined to do it now, and maybe someone else should do it, but this is definitely something we'll want to document very carefully. It is a bit odd, this pattern of mutable input as a return value. (I of course know exactly why we do it, but a user will not.) Not saying you should fix it here, just registering something we'll "eventually" need to fix.

src/Microsoft.ML.Transforms/CustomMappingTransformerFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Microsoft.Data.DataView;
+
+namespace Microsoft.ML.Transforms
+{
+    /// <summary>
+    /// Place this attribute onto a type to cause it to be considered a custom mapping transformer factory.
```
TomFinley: >custom mapping transformer factory [](start = 72, length = 34)

Actually, pursuant to my last comment about documentation, we'd probably want to do another thing here... `<see` tags to and from `CustomMappingEstimator`, the usage of that via `MLContext` (if appropriate), and so on, seems like a larger task that ought to be covered via documentation team. Maybe file an issue if you feel like approaching it right now is a bit much (which I could easily understand)?

src/Microsoft.ML.Transforms/LambdaTransform.cs
TomFinley: Ha ha, and considering the exclusion rules, the internal code analyzer would have to make sure that it is *never* used. :)

---
In reply to: [257425601](https://github.com/dotnet/machinelearning/pull/2569#discussion_r257425601) [](ancestors = 257425601)

src/Microsoft.ML.Transforms/CustomMappingTransformerFactory.cs
eerhardt: Agreed.  Logged #2657.

src/Microsoft.ML.Transforms/CustomMappingTransformerFactory.cs
eerhardt: #2657 will handle this as well.

src/Microsoft.ML.Transforms/LambdaTransform.cs
eerhardt: Logged #2670 to address this.

eerhardt:
OK, @TomFinley, I've renamed the classes. Can you take a look?

@Ivanidzo4ka @yaeldekel - can I get a second pair of eyes?

Issue #2567, opened 2019-02-15
Bad exception experience loading a model


eerhardt:
See this PR review comment: https://github.com/dotnet/machinelearning/pull/1951/files#r244808064.

https://github.com/dotnet/machinelearning/blob/13b33392dd3fbffc0922b2af72c8236b74c95fc0/src/Microsoft.ML.Data/DataLoadSave/TransformerChain.cs#L255-L263

This is causing a real problem with the Custom Mapping Transformer work that I'm doing.

When a user tries loading a model that contains a Custom Mapping Transformer, but we can't find that extension/contract we throw an exception during `ModelLoadContext.LoadModel` above saying `Can't find extension 'foo'`.

However, this code here eats that exception, tries doing something else and then throws a terrible exception `Repository doesn't contain entry DataLoaderModel\Model.key`, which makes absolutely no sense to the user.

We should change this code such that the original exception is thrown if we can't load the `Pipeline`. Maybe we could start writing a flag into a "pipeline" model file to tell if it is supposed to be a Pipeline or not...? If we don't see that flag, then for sure the exception from `ModelLoadContext.LoadModel` above should be thrown.

@TomFinley @Ivanidzo4ka - Thoughts?

eerhardt:
Side note: when fixing this, we should also fix this test:

https://github.com/dotnet/machinelearning/blob/174de1a8dedb7bd72e1f43df6f07b9392c42c290/test/Microsoft.ML.Tests/Transformers/CustomMappingTests.cs#L67-L75

Which would have caught the problem if it was well written. This test should be ensuring the correct exception is thrown, not just any exception.

TomFinley:
I actually ran into this myself. I spent about a half hour confirming , before realizing that *something* in our code stack was actually catching, then eating, the exception (in my case a null reference exception from some incorrect code I wrote), replacing it with this utterly unhelpful and incorrect message during model deserialization.

As you mentioned I had commented that doing this was a bad idea in principle, but I didn't realize at the time just how big of an impact that would have. I thought it was limited to a relatively limited scenario for backwards compatibility of some legacy models, so we could (regrettably) afford some ugliness, assuming it was actually explained why it was necessary. (Still not quite sure why it's needed?) If that same code is also being applied to *newly* serialized models, then that is a serious flaw for sure.

PR #2574, opened 2019-02-15
Respect Marked exception during model loading.

Ivanidzo4ka:
address https://github.com/dotnet/machinelearning/issues/2567

### DIFF
--- a/src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
+++ b/src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
@@ -185,16 +185,25 @@ namespace Microsoft.ML
             internal object CreateInstanceCore(object[] ctorArgs)
             {
                 Contracts.Assert(Utils.Size(ctorArgs) == CtorTypes.Length + ((RequireEnvironment) ? 1 : 0));
-
-                if (InstanceGetter != null)
+                try
                 {
-                    Contracts.Assert(Utils.Size(ctorArgs) == 0);
-                    return InstanceGetter.Invoke(null, null);
+                    if (InstanceGetter != null)
+                    {
+                        Contracts.Assert(Utils.Size(ctorArgs) == 0);
+                        return InstanceGetter.Invoke(null, null);
+                    }
+                    if (Constructor != null)
+                        return Constructor.Invoke(ctorArgs);
+                    if (CreateMethod != null)
+                        return CreateMethod.Invoke(null, ctorArgs);
                 }
-                if (Constructor != null)
-                    return Constructor.Invoke(ctorArgs);
-                if (CreateMethod != null)
-                    return CreateMethod.Invoke(null, ctorArgs);
+                catch (Exception ex)
+                {
+                    if (ex.InnerException != null && ex.InnerException.IsMarked())
+                        throw Contracts.Except(ex, "Error during class instantiation");
+                    else
+                        throw ex;
+                }
                 throw Contracts.Except("Can't instantiate class '{0}'", Type.Name);
             }
 
--- a/src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
+++ b/src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
@@ -202,7 +202,7 @@ namespace Microsoft.ML
                     if (ex.InnerException != null && ex.InnerException.IsMarked())
                         throw Contracts.Except(ex, "Error during class instantiation");
                     else
-                        throw ex;
+                        throw;
                 }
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
```diff
@@ -185,16 +185,25 @@ internal LoadableClassInfo(LoadableClassAttributeBase attr, MethodInfo getter, C
...
+                    }
+                    if (Constructor != null)
+                        return Constructor.Invoke(ctorArgs);
+                    if (CreateMethod != null)
+                        return CreateMethod.Invoke(null, ctorArgs);
+                }
+                catch (Exception ex)
+                {
+                    if (ex.InnerException != null && ex.InnerException.IsMarked())
+                        throw Contracts.Except(ex, "Error during class instantiation");
```
eerhardt: Does this actually add value? Why not just let the original exception be thrown? It will have the stacktrace of where it came from.

src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
```diff
@@ -185,16 +185,25 @@ internal LoadableClassInfo(LoadableClassAttributeBase attr, MethodInfo getter, C
...
+                    else
+                        throw ex;
```
eerhardt: Please never do this.

`throw;` is what you want.

src/Microsoft.ML.Data/DataLoadSave/TransformerChain.cs
```diff
@@ -257,8 +257,10 @@ public static TransformerChain<ITransformer> LoadFrom(IHostEnvironment env, Stre
                     ModelLoadContext.LoadModel<TransformerChain<ITransformer>, SignatureLoadModel>(env, out var transformerChain, rep, LoaderSignature);
                     return transformerChain;
                 }
-                catch
+                catch (FormatException ex)
```
eerhardt: I like this much better.

src/Microsoft.ML.Transforms/LambdaTransform.cs
```diff
@@ -69,6 +69,7 @@ private static ITransformer Create(IHostEnvironment env, ModelLoadContext ctx)
             var contractName = ctx.LoadString();
 
             var composition = env.GetCompositionContainer();
+            Contracts.CheckValue(composition,nameof(composition));
```
eerhardt: Note that this is being deleted with #2569

test/Microsoft.ML.Tests/Transformers/CustomMappingTests.cs
```diff
@@ -69,8 +69,10 @@ public void TestCustomTransformer()
                 TestEstimatorCore(customEst, data);
                 Assert.True(false, "Cannot work without MEF injection");
             }
-            catch (Exception)
+            catch (Exception ex)
             {
+                if (!ex.IsMarked())
+                    throw ex;
```
eerhardt: Again here - `throw;`.

src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
Ivanidzo4ka: Because of `Invoke` call. If anything happens inside it, we got TargetInvocationException and original exception would be inside InnerException.

src/Microsoft.ML.Transforms/LambdaTransform.cs
Ivanidzo4ka: I understand. 
I'm just trying to resolve this comment from TestCustomTrasformer :
`we should have a common mechanism that will make sure this is 'our' exception thrown.`

Only way I know how to do that is to check was exception marked or not.
Currently composition is null, and in next line it throws NullReference. Which is unmarked, and in order to make it marked, I need to run check on it.

src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
eerhardt: At least `catch (TargetInvocationException ex)` here then.

test/Microsoft.ML.Tests/Transformers/CustomMappingTests.cs
```diff
@@ -69,9 +69,10 @@ public void TestCustomTransformer()
                 TestEstimatorCore(customEst, data);
                 Assert.True(false, "Cannot work without MEF injection");
             }
-            catch (Exception)
+            catch (Exception ex)
```
eerhardt: Can you change this to be the specific type of exception we expect?

And possibly adding a check for the message - since I'm sure it will just be `InvalidOperationException`.

Issue #2553, opened 2019-02-14
Cleaning of overload Fit functions


Ivanidzo4ka:
- We shouldn't use `IPredictor` interface for `initialPredictor`, we should specify proper class type.
- `initialPredictor` should be renamed to `modelParameters`

Ivanidzo4ka:
related to https://github.com/dotnet/machinelearning/issues/2251

PR #2577, opened 2019-02-15
Replace IPredictor in Fit method to proper model parameters

Ivanidzo4ka:
Fixes https://github.com/dotnet/machinelearning/issues/2553

### DIFF
--- a/src/Microsoft.ML.HalLearners/SymSgdClassificationTrainer.cs
+++ b/src/Microsoft.ML.HalLearners/SymSgdClassificationTrainer.cs
@@ -181,10 +181,7 @@ namespace Microsoft.ML.Trainers.HalLearners
             {
                 var initPred = context.InitialPredictor;
                 var linearInitPred = initPred as LinearModelParameters;
-                if (linearInitPred == null)
-                    linearInitPred = ((initPred as IWeaklyTypedCalibratedModelParameters)?.WeeklyTypedSubModel) as LinearModelParameters;
-
-                // If initial predictor is set, it must be a linear model or calibrated linear model. Otherwise, we throw.
+                // If initial predictor is set, it must be a linear model.
                 // If initPred is null (i.e., not set), the following check will always be bypassed.
                 // If initPred is not null, then the following checks if a LinearModelParameters is loaded to linearInitPred.
                 Host.CheckParam(initPred == null || linearInitPred != null, nameof(context),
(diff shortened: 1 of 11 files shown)


src/Microsoft.ML.StandardLearners/Standard/LogisticRegression/LbfgsPredictorBase.cs
```diff
@@ -364,7 +364,7 @@ private protected virtual VBuffer<float> InitializeWeightsSgd(IChannel ch, Float
             return result;
         }
 
-        protected abstract VBuffer<float> InitializeWeightsFromPredictor(TModel srcPredictor);
+        protected abstract VBuffer<float> InitializeWeightsFromPredictor(IPredictor srcPredictor);
```
TomFinley: I guess we're trading type safety on the private surface for type safety on the public surface. 😄 That's a good exchange I suppose.

src/Microsoft.ML.HalLearners/SymSgdClassificationTrainer.cs
```diff
@@ -181,10 +181,7 @@ private protected override TPredictor TrainModelCore(TrainContext context)
             {
                 var initPred = context.InitialPredictor;
                 var linearInitPred = initPred as LinearModelParameters;
-                if (linearInitPred == null)
-                    linearInitPred = ((initPred as IWeaklyTypedCalibratedModelParameters)?.WeeklyTypedSubModel) as LinearModelParameters;
-
-                // If initial predictor is set, it must be a linear model or calibrated linear model. Otherwise, we throw.
+                // If initial predictor is set, it must be a linear model.
```
wschin: We might want to allow two model types.

- LinearModelParameters
- CalibratedModelParametersBase<BinaryLinearModelParameters, PlattCalibrator>
The change below seems removing the second case.

TomFinley: I'd rather not, honestly, since it's just a matter of feeding in `.SubModel` in the second case. So, if a user were to see *both* they might reasonably assume that there was a reason for us adding the second -- but there wasn't, actually! It was just a convenience to save the user 9 characters, since we couldn't really use the calibrator anyway, really, right?

TomFinley: My opinion might change in a situation where the calibrator was *somehow* being used in a way that was not completely trivial -- e.g., you might imagine a calibrated model in a logistic model getting model parameters that are somehow "adjusted" by the behavior of the calibrator. But that certainly is not the case now, so it should not be here.

Issue #2576, opened 2019-02-15
ReadFromTextLoader(..., supportsSparse, ...) defaults to true?


stephentoub:
ReadFromTextFile has a `supportsSparse = true` argument.  It seems like this should actually default to false?  Otherwise it tries to interpret all of the input as if it may contain sparse data and fails in the parsing on every line, unless there actually is sparse data.

eerhardt:
@TomFinley @Ivanidzo4ka @yaeldekel - do you know why the current default value for this is `true`? It seems like a very specific format, that the majority of data sets will not conform to.

TomFinley:
I agree with this. The same might be said of "quoting" which fouls people up more often than it helps them. Not only based on your observation from this issue, but if I think back on the years we've had this loader this has been a persistent and confusing problem.

The argument might be made that `TextLoader` and `TextSaver` arguments should be symmetrical, but as I argued in #2452 text saver by default writes out a settings pre-amble anyway which overrides defaults in the text loader, and there are other cases where we have decided that having these be disparate is appropriate. Currently this disparity is limited to header, but if we judge it appropriate there's no reason not to expand it.

If someone knows enough to conform to our format for quoting and sparse vectors, they certainly know enough to set this option, as well as the quoting option.

PR #2579, opened 2019-02-15
Remove exception due to supportsSparse:true in ReadFromTextLoader

stephentoub:
In the GitHubLabeler sample, for example, there are ~450 InvalidOperationExceptions getting thrown and caught internally, as the implementation tries to parse some text as an int and fails.  There's no need for an exception here; the implementation is already using a TryParse method, which it then turns into an exception and then immediately catches and ignores... we can just cut out the middle man and not throw.

Related to https://github.com/dotnet/machinelearning-samples/pull/257
Related to https://github.com/dotnet/machinelearning/issues/2576

### DIFF
--- a/src/Microsoft.ML.Data/Data/Conversion.cs
+++ b/src/Microsoft.ML.Data/Data/Conversion.cs
@@ -1585,6 +1585,17 @@ namespace Microsoft.ML.Data.Conversion
         {
             value = ParseI4(in span);
         }
+        public bool TryConvert(in TX span, ref I4 value)
+        {
+            TryParseSigned(I4.MaxValue, in span, out long? res);
+            if (res.HasValue)
+            {
+                value = (I4)res.GetValueOrDefault();
+                return true;
+            }
+
+            return false;
+        }
         public void Convert(in TX span, ref U4 value)
         {
             value = ParseU4(in span);
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderParser.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderParser.cs
@@ -1016,18 +1016,8 @@ namespace Microsoft.ML.Data
                                 }
                                 var spanT = Fields.Spans[Fields.Count - 1];
 
-                                // Note that Convert throws exception the text is unparsable.
-                                int csrc = default;
-                                try
-                                {
-                                    Conversions.Instance.Convert(in spanT, ref csrc);
-                                }
-                                catch
-                                {
-                                    Contracts.Assert(csrc == default);
-                                }
-
-                                if (csrc <= 0)
+                                int csrc = 0;
+                                if (!Conversions.Instance.TryConvert(in spanT, ref csrc) || csrc <= 0)
                                 {
                                     _stats.LogBadFmt(ref scan, "Bad dimensionality or ambiguous sparse item. Use sparse=- for non-sparse file, and/or quote the value.");
                                     break;


src/Microsoft.ML.Data/Data/Conversion.cs
```diff
@@ -1585,6 +1585,17 @@ public void Convert(in TX span, ref I4 value)
         {
             value = ParseI4(in span);
         }
+        public bool TryConvert(in TX span, ref I4 value)
+        {
+            TryParseSigned(I4.MaxValue, in span, out long? res);
+            if (res.HasValue)
+            {
+                value = (I4)res.GetValueOrDefault();
```
codemzs: >GetValueOrDefault [](start = 32, length = 17)

Why not just GetValue() since you have already checked it has value?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderParser.cs
```diff
@@ -1016,18 +1016,8 @@ public int GatherFields(ReadOnlyMemory<char> lineSpan, ReadOnlySpan<char> span,
...
-                                {
-                                    Conversions.Instance.Convert(in spanT, ref csrc);
-                                }
-                                catch
-                                {
-                                    Contracts.Assert(csrc == default);
-                                }
-
-                                if (csrc <= 0)
+                                int csrc = 0;
```
codemzs: >0 [](start = 43, length = 1)

default

src/Microsoft.ML.Data/Data/Conversion.cs
stephentoub: > Why not just GetValue() since you have already checked it has value?

Exactly because we already checked the value.  `GetValueOrDefault()` just does `return value;` whereas `.Value` does something like `if (!_hasValue) throw new Exception(); return _value;`.

codemzs: Thanks for the explanation. My understanding of these routines was the exact opposite of what they really do. Looks good!

TomFinley: Hmmm. We may have some code to rewrite a number of places...

src/Microsoft.ML.Data/Data/Conversion.cs
```diff
@@ -1585,6 +1585,17 @@ public void Convert(in TX span, ref I4 value)
         {
             value = ParseI4(in span);
         }
+        public bool TryConvert(in TX span, ref I4 value)
```
eerhardt: Why not use the existing `TryParse` method?

https://github.com/dotnet/machinelearning/blob/369a9b614b68793de8401c843df6f5d566088993/src/Microsoft.ML.Data/Data/Conversion.cs#L1184-L1201

stephentoub: I was trying to remain consistent with the existing code, which assigned a value, then passed it by ref, and then checked whether it remained what it was assigned.  In practice it wouldn't matter because it would end up with the default value in either case, but it would incur additional writes in the failure case, and I didn't know whether it would matter.  When I don't know a code base particularly well and what ramifications my changes might have, I try to maintain the existing code as much as possible.

eerhardt: I guess my concern is the consistency in the internal `Conversion` class. 

1. We now have `TryParse` and `TryConvert`.  Which one should new callers choose?
2. We have `TryConvert` for Int32, but no other types?

stephentoub: 1. Whichever signature they need. If they want an out, they choose TryParse, if they want ref, they choose TryConvert.
2. Yes.  You'd prefer we add a bunch of dead code?

If you feel strongly about this and want to change the call site to use TryParse, go for it.  As I noted, my goal was to muck minimally with the call site logic.

Issue #2589, opened 2019-02-16
Calibrators shouldn't have public constructors


Ivanidzo4ka:
We closed our ModelParameters constructors. We should do same thing with calibrators.
https://github.com/dotnet/machinelearning/blob/1f90f50f813ecd152e46e92deb4a7827c7f24e31/src/Microsoft.ML.Data/Prediction/Calibrator.cs#L1745

Ivanidzo4ka:
And `CalibrationDataStore` should be internal.

PR #2601, opened 2019-02-18
Cleanup calibrators

Ivanidzo4ka:
Fixes https://github.com/dotnet/machinelearning/issues/2589


src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -1180,7 +1180,7 @@ private void SaveCore(ModelSaveContext ctx)
             ctx.Writer.Write(sizeof(float));
             ctx.Writer.Write(BinSize);
             ctx.Writer.Write(Min);
-            ctx.Writer.WriteSingleArray(BinProbs);
+            ctx.Writer.WriteSingleArray(BinProbs as float[]);
```
TomFinley: > as float[] [](start = 48, length = 11)

Let's not please... let's do a backing private field of `float[]` that we actually operate over. This will have a couple of advantages, most notably not having this bad code smell, but also incidentally making access to this structure much faster since it won't have to work through the virtual method table on every single usage.

Ivanidzo4ka: Just to make completely clear.
We will have:
```
public IReadOnlyList<float> BinProbs => _binProbs;
private readonly float[] _binProbs;
```
And for loading and saving we will use `_binProbs`
Does it sound good?

TomFinley: And other operators too, yes.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -1205,11 +1207,6 @@ internal static int GetBinIdx(float output, float min, float binSize, int numBin
             return binIdx;
```
TomFinley: Ugh. This code makes my eyes bleed but if no one has complained after all these years we can probably let it stand for a bit. :)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -1218,8 +1215,91 @@ public string GetSummary()
     [BestFriend]
     internal abstract class CalibratorTrainerBase : ICalibratorTrainer
     {
+        public sealed class DataStore : IEnumerable<DataStore.DataItem>
+        {
+            public readonly struct DataItem
```
rogancarr: >public readonly struct DataItem [](start = 12, length = 31)

Really vague name for a class visible throughout the assembly.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -1218,8 +1215,91 @@ public string GetSummary()
     [BestFriend]
     internal abstract class CalibratorTrainerBase : ICalibratorTrainer
     {
+        public sealed class DataStore : IEnumerable<DataStore.DataItem>
```
rogancarr: >public sealed class DataStore [](start = 8, length = 29)

Really vague name. Can we think of something more precise?

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -1218,8 +1215,91 @@ public string GetSummary()
...
+            private readonly Random _random;
+
+            private static int _randSeed;
+
+            private readonly int _capacity;
+            private DataItem[] _data;
+            private bool _dataSorted;
+
+            public DataStore()
+                : this(1000000)
```
rogancarr: >1000000 [](start = 23, length = 7)

Any reason?

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -1218,8 +1215,91 @@ public string GetSummary()
...
+            {
+            }
+
+            public DataStore(int capacity)
```
rogancarr: >public DataStore(int capacity) [](start = 12, length = 30)

Maybe instead of two constructors, we just do `int capacity=defaultValue`?

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -1218,8 +1215,91 @@ public string GetSummary()
...
+            {
+                Contracts.CheckParam(capacity > 0, nameof(capacity), "must be positive");
+
+                _capacity = capacity;
+                _data = new DataItem[Math.Min(4, capacity)];
+                // REVIEW: Horrifying. At a point when we have the IHost stuff plumbed through
```
rogancarr: >Horrifying [](start = 27, length = 10)

The horror

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -1218,8 +1215,91 @@ public string GetSummary()
...
+            }
+
+            IEnumerator IEnumerable.GetEnumerator()
+            {
+                return GetEnumerator();
+            }
+
+            public void AddToStore(float score, bool isPositive, float weight)
+            {
+                // Can't calibrate NaN scores.
```
rogancarr: > // Can't calibrate NaN scores. [](start = 15, length = 31)

Isn't it BAD to get NaN scores?

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -1485,15 +1565,21 @@ private static VersionInfo GetVersionInfo()
 
         private readonly IHost _host;
 
+        /// <summary>
+        /// Slope value for this calibrator.
```
rogancarr: >Slope value for this calibrator. [](start = 12, length = 32)

Maybe use the phrases `bias` and `weight` instead of `Slope` and `Offset`? This is more consistent with our public API.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -1556,14 +1642,15 @@ private void SaveCore(ModelSaveContext ctx)
             }
         }
 
+        /// <summary> Given a classifier output, produce the probability.</summary>
```
rogancarr: >/ <summary [](start = 10, length = 10)

break pls

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -1703,12 +1785,17 @@ public override ICalibrator CreateCalibrator(IChannel ch)
...
+    /// The function that is implemented by this calibrator is:
+    /// P(x) =
+    /// <list type="bullet">
+    /// <item><description><see cref="Values"/>[i], if <see cref="Mins"/>[i] &lt;= x &lt;= <see cref="Maxes"/>[i]</description>></item>
+    /// <item> <description>Linear interpolation between <see cref="Values"/>[i] and <see cref="Values"/>[i+1], if <see cref="Maxes"/>[i] &lt; x &lt; <see cref="Mins"/>[i+1]</description></item>
+    /// <item><description><see cref="Values"/>[0], if x &lt; <see cref="Mins"/>[0]</description></item>
+    /// <item><description><see cref="Values"/>[n], if x &gt; <see cref="Maxes"/>[n]</description></item>
+    ///</list>
+    /// </remarks>
     public sealed class PavCalibrator : ICalibrator, ICanSaveInBinaryFormat
```
rogancarr: If it's not too hard, can you provide a link to any resources about this kind of calibration? You could also file an issue and assign it to the docs folks.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -1851,6 +1947,7 @@ private void SaveCore(ModelSaveContext ctx)
             _host.CheckDecode(valuePrev <= 1);
         }
 
+        /// <summary> Given a classifier output, produce the probability.</summary>
```
rogancarr: ><summary> [](start = 12, length = 9)

break

rogancarr:
Just a few nits, although I realize it's nits on copy/pasted code, so feel free to defer those to future maintenance fixes.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
Ivanidzo4ka: It fits the line, why it need break?

---
In reply to: [258187676](https://github.com/dotnet/machinelearning/pull/2601#discussion_r258187676) [](ancestors = 258187676)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
TomFinley: As far as I see it is a nested class inside the internal `CalibratorTrainerBase`?

Ivanidzo4ka: A) it's part of internal class.
B) I'm making it private protected.
C) I'm just moving this code around to hide it, I don't have much intent in improving it's right now

---
In reply to: [258183106](https://github.com/dotnet/machinelearning/pull/2601#discussion_r258183106) [](ancestors = 258183106)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
rogancarr: I thought the style was to break the xml tags onto their own lines, but I could be mistaken.

---
In reply to: [258188999](https://github.com/dotnet/machinelearning/pull/2601#discussion_r258188999) [](ancestors = 258188999,258187676)

PR #2605, opened 2019-02-19
Hide IPredictor and descendants,

### LINKED ISSUE #2251: IPredictor and related parts hiding

### LINKED ISSUE #2592: PredictionKind in TrainerEstimatorBase should be internalized
I don't see any reason to expose it publicly especially since all our trainers coupled with specific catalog.

### LINKED ISSUE #2603: Hide some parts of TrainerInfo
There is this structure `TrainerInfo` which was previously used on `ITrainer`. It is returned however not only by this old (internal) interface, but also by the newer `ITrainerEstimator`. It is still useful in this capacity, but a few things on it should be hidden.

https://github.com/dotnet/machinelearning/blob/b881fdc2eab55e184eb0856d4c10c732a58e0887/src/Microsoft.ML.Core/Prediction/TrainerInfo.cs#L11

This is probably still a good idea to return from `ITrainerEstimator`, but in a restricted form. We'd internalize some of these things though, as they are either inapplicable for one reason or another: `NeedCalibration`, whether a trainer estimator returns a calibrated model or not is obvious from the return type and intellisense. The `Supports*` properties are intended per #509 to indicate which fields of `TrainContext` will be supported in a call to `Train`, which is totally irrelevant w.r.t. trainer estimators.

TomFinley:
Fixes #2251. Fixes #2592. Fixes #2603.

Most significant change is internalization of IPredictor, IPredictorProducing, PredictionKind. Deletion of TrainerBase abstract base class, since it is used in few enough places that complicating the type heirarchy was no longer beneficial.

I changed feature contribution so that it does not descend from `IPredictor` since I think we still might need it (though I made its implementation internal on all implementors). Querying @rogancarr about this though, he may have some thoughts on its current state.

Also incidentally found some minor infrastructure that is specific to ITrainer lurking around.


Ivanidzo4ka:
>         public static IDataView ReadFromEnumerable<TRow>(this DataOperationsCatalog catalog, IEnumerable<TRow> data, SchemaDefinition schemaDefinition = null)

Funny how this is only remaining public method in class, and it doesn't looks like it belong here. (We don't create "component", right?).
Can be cleaned later anyway.

---
Refers to: src/Microsoft.ML.Data/Utilities/ComponentCreation.cs:78 in cc4b2ae. [](commit_id = cc4b2ae2f43238c7fbb0e2fa52b898e72db9bd84, deletion_comment = False)

codemzs:
Just glad PredictionKind is being internalized here, that used to be the sole reason a class would appear in public surface area ...

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
```diff
@@ -302,24 +304,29 @@ internal static class Defaults
         /// <summary>
         /// The <see cref="TrainerInfo"/> contains general parameters for this trainer.
         /// </summary>
-        public override TrainerInfo Info { get; }
+        TrainerInfo ITrainer.Info => _info;
```
Ivanidzo4ka: > TrainerInfo ITrainer.Info => _info; [](start = 7, length = 36)

Here and in FAFM we hide TrainerInfo.
In same time TrainerEstimatorBase class keep having it.
I would prefer to be consistent and make TrainerInfo public. Unless you have reason not to do that.

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
```diff
@@ -86,7 +86,7 @@ namespace Microsoft.ML.Trainers
     /// ]]>
     /// </format>
     /// </example>
-    public sealed class MatrixFactorizationTrainer : TrainerBase<MatrixFactorizationModelParameters>,
+    public sealed class MatrixFactorizationTrainer : ITrainer<MatrixFactorizationModelParameters>,
         IEstimator<MatrixFactorizationPredictionTransformer>
```
Ivanidzo4ka: @wschin do you know why it's not `ITrainerEstimator`  but TrainerBase+ IEstimator?

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
TomFinley: Hmm okay fair enough.

TomFinley: Let's keep it explicit for now. The fact that it's not producing an `ISingleFeaturePredictionTransformer` complicates the story a bit, and I'd rather not force a solution prior to v1.

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
TomFinley: Can't. If you look at `ITrainerEstimator` you'll see that it doesn't produce `ISingleFeaturePredictionTransformer`. That base interface is meant to capture some common cases, but it is not universal.

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
TomFinley: So, to be clear, despite initially looking like a mistake I think it's correct and intentional.

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
TomFinley: So just going to mark resolved, feel free to reactivate if this angers you! 😄

PR #2611, opened 2019-02-19
Address PR feedback from #2579

eerhardt:
We don't need a TryConvert method, we already have a TryParse, which can be used instead.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderParser.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderParser.cs
@@ -1016,8 +1016,8 @@ namespace Microsoft.ML.Data
                                 }
                                 var spanT = Fields.Spans[Fields.Count - 1];
 
-                                int csrc = 0;
-                                if (!Conversions.Instance.TryConvert(in spanT, ref csrc) || csrc <= 0)
+                                int csrc;
+                                if (!Conversions.Instance.TryParse(in spanT, out csrc) || csrc <= 0)
                                 {
                                     _stats.LogBadFmt(ref scan, "Bad dimensionality or ambiguous sparse item. Use sparse=- for non-sparse file, and/or quote the value.");
                                     break;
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderParser.cs
```diff
@@ -1016,8 +1016,8 @@ public int GatherFields(ReadOnlyMemory<char> lineSpan, ReadOnlySpan<char> span,
                                 }
                                 var spanT = Fields.Spans[Fields.Count - 1];
 
-                                int csrc = 0;
-                                if (!Conversions.Instance.TryConvert(in spanT, ref csrc) || csrc <= 0)
+                                int csrc;
+                                if (!Conversions.Instance.TryParse(in spanT, out csrc) || csrc <= 0)
```
stephentoub: Note that TryParse is going to be slightly more expensive, but it probably doesn't matter.

stephentoub:
This begs the question why the methods are named Convert and TryParse, rather than Convert and TryConvert, or Parse and TryParse.  As long as you're cleaning things up, seems it should be consistent.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderParser.cs
eerhardt: > slightly more expensive

Because of the extra write to the `out` parameter? Or because of the under/overflow check? Or something else?

stephentoub: The extra write in the case of success, two extra writes in the case of failure, the overflow check, the cast from the nullable rather than GetValueOrDefault, larger code that's even less likely to be inlined (though it's probably already not getting inlined).  In general this code is not particularly clean.

eerhardt:
@stephentoub - I've logged #2669 for your concerns on the existing `TryParse` code.

PR #2630, opened 2019-02-19
Change Default Settings in TextLoader

### LINKED ISSUE #2452: DataOperationsCatalog SaveAsText extension method is evil
I was attempting to migrate some of our tests, when I discovered we have a few problems in our new API on text saving. I know @artidoro had some thoughts on text saving/loading so mentioning him. Also I know @sfilipi and @rogancarr are handling many issues w.r.t. API completeness and consistency as they work on samples and docs, so maybe they have some thoughts on this. (Of course everyone is free to chime in.)

From what I can see, whoever authored this method confused the defaults with the text *loader* and the text *saver*. (Note how the defaults used in the text saving utility method are coming from `TextLoader`, which is incorrect.) It might seem intuitive if you don't think about it too hard that text saving and loading should have the same defaults, but practically it becomes clear they should not. The situations where one is "loading" into ML.NET and "saving" out of ML.NET are in fact very different situations. When someone is using a text loader ***with non-default settings*** they're usually asking us to ingest *their* format (so we chose the most helpful defaults for that more common scenario), whereas our text saver makes some attempt at schema. (Note also that under default settings, the text loader loads our own format without trouble, since it detects that a schema and settings is embedded in the file itself.)

This is the offending method:

https://github.com/dotnet/machinelearning/blob/e192a18d553a71ded3728495e9aaf4349169290d/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs#L147-L153

As point of reference, this is that constant:

https://github.com/dotnet/machinelearning/blob/e192a18d553a71ded3728495e9aaf4349169290d/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L401

Now, compare this with the *actual* defaults on our text saver:

https://github.com/dotnet/machinelearning/blob/e192a18d553a71ded3728495e9aaf4349169290d/src/Microsoft.ML.Data/DataLoadSave/Text/TextSaver.cs#L28-L43

* The default for whether header row is saved has gone from `true` to `false`. The primary practical effect of this is, we're now dropping feature names (or more precisely, slot names) by default. This seems silly. Feature names are *important*. I think we ought to keep them by default.

* We've lost the ability to force saving as dense format *at all* through this new API. This is often important for comprehensibility.

I ran into it while I was trying to clean up some of our tests to use more of the actual public surface. Consider this test.

https://github.com/dotnet/machinelearning/blob/e192a18d553a71ded3728495e9aaf4349169290d/test/Microsoft.ML.Tests/Transformers/ConcatTests.cs#L133-L138

That's kind of obnoxious, and not using our public API. I'd love to migrate it over to something like this:

```csharp
using (var fs = File.Create(outputPath)) 
    ML.Data.SaveAsText(data, fs, keepHidden: false, forceDense: true);
```

But I can't specify equivalent settings because there's no way to force dense.

***So I suggest this:*** we change the defaults of this text *saving* utility function to use the text *saver* defaults, instead of the text loader defaults, and also restore the ability to force a dense format.

## Why forcing dense is kind of useful sometimes...

It may not be obvious why forcing to dense is kind of useful. So imagine this input file `foo.txt`.

```
1,2,3,4,5,6
0,0,1,0,0,0
```

Then imagine I have this MML command line:

    dotnet MML.dll savedata loader=text{sep=comma} data=foo.txt dout=foo1.txt

The resulting `foo1.txt` file is this:

```
#@ TextLoader{
#@   header+
#@   sep=tab
#@   col=Label:R4:0
#@   col=Features:R4:1-5
#@ }
Label   5       0:""
1       2       3       4       5       6
6       2:1
```

What is that `6 2:1` line? It is encoding the information that this line has 6 fields, and the field at index 2 is the only one with a non-default value. That is, it has detected, "hey, this can be sparsely encoded," and it has done so. Similarly with this seemingly crazy `Label   5       0:""` line. But sometimes we find that confusing!! I've spent over the years probably, cumulatively, *days* trying to explain the ins and outs of the mixed sparse/dense format. So we have this setting to say, "you know what, I don't care about efficient, give me a dense format."

    dotnet MML.dll savedata loader=text{sep=comma} data=foo.txt dout=foo2.txt saver=text{dense+}

The result is this:

```
#@ TextLoader{
#@   header+
#@   sep=tab
#@   col=Label:R4:0
#@   col=Features:R4:1-5
#@ }
Label   ""      ""      ""      ""      ""
1       2       3       4       5       6
0       0       1       0       0       0
```

Less efficient? Sure. Eaier to understand? I'd say so. And lots of our tests use it, since our tests are usually writing small amounts of data and we found comprehensibility of test files to be valuable.

### LINKED ISSUE #2576: ReadFromTextLoader(..., supportsSparse, ...) defaults to true?

wschin:
To fix #2576, this PR makes the default to `sparse-` everywhere including both of command line and public APIs.

- [x] Make `sparse-` as default
- [x] Make `quote-` as default
- [x] Fix #2452, which is about TextSaver.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Text;
+using System.Text.RegularExpressions;
 using Microsoft.Data.DataView;
 using Microsoft.ML;
 using Microsoft.ML.CommandLine;
@@ -488,7 +489,7 @@ namespace Microsoft.ML.Data
         internal static class Defaults
         {
             internal const bool AllowQuoting = true;
-            internal const bool AllowSparse = true;
+            internal const bool AllowSparse = false;
             internal const char Separator = '\t';
             internal const bool HasHeader = false;
             internal const bool TrimWhitespace = false;
@@ -1068,15 +1069,16 @@ namespace Microsoft.ML.Data
         /// <param name="hasHeader">Whether the file has a header.</param>
         /// <param name="separatorChar"> The character used as separator between data points in a row. By default the tab character is used as separator.</param>
         /// <param name="dataSample">Allows to expose items that can be used for reading.</param>
-        internal TextLoader(IHostEnvironment env, Column[] columns, bool hasHeader = false, char separatorChar = '\t', IMultiStreamSource dataSample = null)
-            : this(env, MakeArgs(columns, hasHeader, new[] { separatorChar }), dataSample)
+        /// <param name="allowSparse">Whether the file can contain numerical vectors in sparse format.</param>
+        internal TextLoader(IHostEnvironment env, Column[] columns, bool hasHeader = false, char separatorChar = '\t', IMultiStreamSource dataSample = null, bool allowSparse = false)
+            : this(env, MakeArgs(columns, hasHeader, new[] { separatorChar }, allowSparse), dataSample)
         {
         }
 
-        private static Options MakeArgs(Column[] columns, bool hasHeader, char[] separatorChars)
+        private static Options MakeArgs(Column[] columns, bool hasHeader, char[] separatorChars, bool allowSparse)
         {
             Contracts.AssertValue(separatorChars);
-            var result = new Options { Columns = columns, HasHeader = hasHeader, Separators = separatorChars};
+            var result = new Options { Columns = columns, HasHeader = hasHeader, Separators = separatorChars, AllowSparse = allowSparse };
             return result;
         }
 
@@ -1255,6 +1257,7 @@ namespace Microsoft.ML.Data
                 HeaderFile = options.HeaderFile,
                 MaxRows = options.MaxRows
             });
+            tmp = Regex.Replace(tmp, @"[(sparse=\+)]", "");
 
             // Try to get the schema information from the file.
             string str = Cursor.GetEmbeddedArgs(files);
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -488,7 +488,7 @@ namespace Microsoft.ML.Data
 
         internal static class Defaults
         {
-            internal const bool AllowQuoting = true;
+            internal const bool AllowQuoting = false;
             internal const bool AllowSparse = false;
             internal const char Separator = '\t';
             internal const bool HasHeader = false;
@@ -1070,15 +1070,16 @@ namespace Microsoft.ML.Data
         /// <param name="separatorChar"> The character used as separator between data points in a row. By default the tab character is used as separator.</param>
         /// <param name="dataSample">Allows to expose items that can be used for reading.</param>
         /// <param name="allowSparse">Whether the file can contain numerical vectors in sparse format.</param>
-        internal TextLoader(IHostEnvironment env, Column[] columns, bool hasHeader = false, char separatorChar = '\t', IMultiStreamSource dataSample = null, bool allowSparse = false)
-            : this(env, MakeArgs(columns, hasHeader, new[] { separatorChar }, allowSparse), dataSample)
+        /// <param name="allowQuoting">Whether the file can contain numerical vectors in sparse format.</param>
+        internal TextLoader(IHostEnvironment env, Column[] columns, bool hasHeader = false, char separatorChar = '\t', IMultiStreamSource dataSample = null, bool allowSparse = false, bool allowQuoting = false)
+            : this(env, MakeArgs(columns, hasHeader, new[] { separatorChar }, allowSparse, allowQuoting), dataSample)
         {
         }
 
-        private static Options MakeArgs(Column[] columns, bool hasHeader, char[] separatorChars, bool allowSparse)
+        private static Options MakeArgs(Column[] columns, bool hasHeader, char[] separatorChars, bool allowSparse, bool allowQuoting)
         {
             Contracts.AssertValue(separatorChars);
-            var result = new Options { Columns = columns, HasHeader = hasHeader, Separators = separatorChars, AllowSparse = allowSparse };
+            var result = new Options { Columns = columns, HasHeader = hasHeader, Separators = separatorChars, AllowSparse = allowSparse, AllowQuoting = allowQuoting };
             return result;
         }
 
@@ -1257,7 +1258,7 @@ namespace Microsoft.ML.Data
                 HeaderFile = options.HeaderFile,
                 MaxRows = options.MaxRows
             });
-            tmp = Regex.Replace(tmp, @"[(sparse=\+)]", "");
+            tmp = Regex.Replace(tmp, @"[(sparse=\+)|(quote\+)]", "");
 
             // Try to get the schema information from the file.
             string str = Cursor.GetEmbeddedArgs(files);
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1071,7 +1071,9 @@ namespace Microsoft.ML.Data
         /// <param name="dataSample">Allows to expose items that can be used for reading.</param>
         /// <param name="allowSparse">Whether the file can contain numerical vectors in sparse format.</param>
         /// <param name="allowQuoting">Whether the file can contain numerical vectors in sparse format.</param>
-        internal TextLoader(IHostEnvironment env, Column[] columns, bool hasHeader = false, char separatorChar = '\t', IMultiStreamSource dataSample = null, bool allowSparse = false, bool allowQuoting = false)
+        internal TextLoader(IHostEnvironment env, Column[] columns, bool hasHeader = Defaults.HasHeader,
+            char separatorChar = Defaults.Separator, IMultiStreamSource dataSample = null,
+            bool allowSparse = Defaults.AllowSparse, bool allowQuoting = Defaults.AllowQuoting)
             : this(env, MakeArgs(columns, hasHeader, new[] { separatorChar }, allowSparse, allowQuoting), dataSample)
         {
         }
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1064,14 +1064,14 @@ namespace Microsoft.ML.Data
         /// </summary>
         /// <param name="env">The environment to use.</param>
         /// <param name="columns">Defines a mapping between input columns in the file and IDataView columns.</param>
-        /// <param name="hasHeader">Whether the file has a header.</param>
         /// <param name="separatorChar"> The character used as separator between data points in a row. By default the tab character is used as separator.</param>
-        /// <param name="dataSample">Allows to expose items that can be used for reading.</param>
+        /// <param name="hasHeader">Whether the file has a header.</param>
         /// <param name="allowSparse">Whether the file can contain numerical vectors in sparse format.</param>
         /// <param name="allowQuoting">Whether the content of a column can be parsed from a string starting and ending with quote.</param>
-        internal TextLoader(IHostEnvironment env, Column[] columns, bool hasHeader = Defaults.HasHeader,
-            char separatorChar = Defaults.Separator, IMultiStreamSource dataSample = null,
-            bool allowSparse = Defaults.AllowSparse, bool allowQuoting = Defaults.AllowQuoting)
+        /// <param name="dataSample">Allows to expose items that can be used for reading.</param>
+        internal TextLoader(IHostEnvironment env, Column[] columns, char separatorChar = Defaults.Separator,
+            bool hasHeader = Defaults.HasHeader, bool allowSparse = Defaults.AllowSparse,
+            bool allowQuoting = Defaults.AllowQuoting, IMultiStreamSource dataSample = null)
             : this(env, MakeArgs(columns, hasHeader, new[] { separatorChar }, allowSparse, allowQuoting), dataSample)
         {
         }
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1260,7 +1260,6 @@ namespace Microsoft.ML.Data
                 HeaderFile = options.HeaderFile,
                 MaxRows = options.MaxRows
             });
-            tmp = Regex.Replace(tmp, @"[(sparse=\+)|(quote\+)]", "");
 
             // Try to get the schema information from the file.
             string str = Cursor.GetEmbeddedArgs(files);
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -7,14 +7,12 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Text;
-using System.Text.RegularExpressions;
 using Microsoft.Data.DataView;
 using Microsoft.ML;
 using Microsoft.ML.CommandLine;
 using Microsoft.ML.Data;
 using Microsoft.ML.Internal.Utilities;
 using Microsoft.ML.Model;
-using Float = System.Single;
 
 [assembly: LoadableClass(TextLoader.Summary, typeof(IDataLoader), typeof(TextLoader), typeof(TextLoader.Options), typeof(SignatureDataLoader),
     "Text Loader", "TextLoader", "Text", DocName = "loader/TextLoader.md")]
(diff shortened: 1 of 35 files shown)


src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -395,7 +396,7 @@ public sealed class Options : ArgumentsCore
         internal static class DefaultArguments
         {
             internal const bool AllowQuoting = true;
```
TomFinley: >AllowQuoting [](start = 32, length = 12)

I feel like this needs to be changed too.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -19,12 +19,14 @@ public static class TextLoaderSaverCatalog
...
         /// <param name="dataSample">The optional location of a data sample. The sample can be used to infer column names and number of slots in each column.</param>
+        /// <param name="allowSparse">Whether the file can contain numerical vectors in sparse format.</param>
         public static TextLoader CreateTextLoader(this DataOperationsCatalog catalog,
             TextLoader.Column[] columns,
             bool hasHeader = TextLoader.DefaultArguments.HasHeader,
             char separatorChar = TextLoader.DefaultArguments.Separator,
-            IMultiStreamSource dataSample = null)
-            => new TextLoader(CatalogUtils.GetEnvironment(catalog), columns, hasHeader, separatorChar, dataSample);
+            IMultiStreamSource dataSample = null,
+            bool allowSparse = TextLoader.DefaultArguments.AllowSparse)
```
TomFinley: >TextLoader [](start = 31, length = 10)

Consider that since this is the code that needs to be changed for #2452, which you've self-assigned anyway, we may as well correct this while we're at it. (Otherwise I see no point in changing it.)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
wschin: Absolutely. Please take a look at Iteration 3.

---
In reply to: [258240404](https://github.com/dotnet/machinelearning/pull/2630#discussion_r258240404) [](ancestors = 258240404)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
wschin: Couldn't agree more. Proposed solution to #2452 can be found in Iteration 4!

---
In reply to: [258240786](https://github.com/dotnet/machinelearning/pull/2630#discussion_r258240786) [](ancestors = 258240786)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -975,15 +976,17 @@ private bool HasHeader
         /// <param name="hasHeader">Whether the file has a header.</param>
         /// <param name="separatorChar"> The character used as separator between data points in a row. By default the tab character is used as separator.</param>
         /// <param name="dataSample">Allows to expose items that can be used for reading.</param>
-        public TextLoader(IHostEnvironment env, Column[] columns, bool hasHeader = false, char separatorChar = '\t', IMultiStreamSource dataSample = null)
-            : this(env, MakeArgs(columns, hasHeader, new[] { separatorChar }), dataSample)
+        /// <param name="allowSparse">Whether the file can contain numerical vectors in sparse format.</param>
+        /// <param name="allowQuoting">Whether the file can contain numerical vectors in sparse format.</param>
+        public TextLoader(IHostEnvironment env, Column[] columns, bool hasHeader = false, char separatorChar = '\t', IMultiStreamSource dataSample = null, bool allowSparse = false, bool allowQuoting = false)
```
eerhardt: The new parameters should use the default constants.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1145,6 +1148,7 @@ private sealed class LoaderHolder
             // Verify that the current schema-defining arguments are default.
             // Get settings just for core arguments, not everything.
             string tmp = CmdParser.GetSettings(host, options, new ArgumentsCore());
+            tmp = Regex.Replace(tmp, @"[(sparse=\+)|(quote\+)]", "");
```
eerhardt: I don’t understand what this is for... can you explain?

wschin: The following code may throw if tmp is not empty (Line 1163) Thus, we need to remove those values here.

---
In reply to: [258307862](https://github.com/dotnet/machinelearning/pull/2630#discussion_r258307862) [](ancestors = 258307862)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
justinormont: I feel like having quoting on by default is better. Mainly this helps with CSV files w/ text columns (commas in the quoted strings). 

 | row | needs quotes |
|--|--|
| `true`, `"Schrödinger's cat walks into a bar, and doesn't."` |  yes | 
| `false`, `Wanted: Schrodinger's cat. Dead and alive.` | no |

I see more datasets with gains from default **on** quoting, then hurt by it.

wschin: I'd like to have non-standard settings to be non-default. There is no standard on parsing quoted strings.

---
In reply to: [258341863](https://github.com/dotnet/machinelearning/pull/2630#discussion_r258341863) [](ancestors = 258341863)

justinormont: To be clear, my recommendation is to keep  `AllowQuoting` defaulting to `true`. 

General reason is that it more datasets will work. The impact of the parameter falsely on, is less than falsely off. 

/cc @TomFinley: thoughts?

TomFinley: I'm actually rather surprised to hear you say this @justinormont, since if I think back, at least internally I recall you being one of the most frequent "victims" of the quoting. Indeed it was one of the most frequent questions we got over the years.

I think what you said would be true for a TSV and CSV parser, to adhere to that quasi-format. However this isn't a TSV or CSV parser, and differs from it in some important respects. I'd agree that an `IDataReader` implementation should adhere to the format of that, but that argument doesn't really apply here.

I'm not motivated exactly by any desire to read CSV format (this thing doesn't do that anyway, so, you can continue quoting as many APIs to read CSV format as you like), but more motivated by what I've perceived as more accidental problems caused by having this default *on* than would have been solved had we left it *off*.

justinormont: This is incorrect. Exactly zero of the issues I ran into would be solved by turning off quoting. 

The issue was our quoting support did not go far enough. I main issue I faced was that we haven't added support for newlines in the quoted string. I believe the argument at the time was that we treat lines independently for speed. 

Most datasets are in CSV/TSV format. We should be supporting the datasets of our users.

wschin: If quote is not fully-functioning, it'd be better to turn it off by default and have another TSC & CSV reader implementation, I guess.

---
In reply to: [258668384](https://github.com/dotnet/machinelearning/pull/2630#discussion_r258668384) [](ancestors = 258668384)

TomFinley: > This is incorrect. Exactly zero of the issues I ran into would be solved by turning off quoting.

I don't think that's true. I recall on multiple occasions seeing you run experiments where you were getting these warnings and you just sort of ignored it, even though it was probably corrupting your dataset. It definitely wasn't beneficial on the whole. It isn't a matter of it "not go far enough," because it wasn't intended to do that at any point. I believe this has also been explained to you multiple times by multiple people. At any rate, it again reinforces the point that this is not a CSV reader, which I believe I also mentioned in my reply.

There's also the issue that there are other people in the world, and on the whole if I look through my DRI email list, I see way more instances of turning quoting off than I see instances where someone somehow accidentally stumbled into using our quoting system.

So, off.

justinormont: I see this as lesser of two evils. 

Default ON:
* Has the potential to cause read errors (dropped rows) when there are newlines or quotes in the quoted string _(generally doesn't matter)_
* Non-text datasets are unaffected

Default OFF:
* Most text datasets fail _(bad)_
* Non-text datasets are unaffected

Let me know if I'm wrong, but I think the argument more directly is:
The user should be forced to mung their dataset in to the form we are expecting (eg: no commas/newlines/escaped-quotes in a string) vs. we should do our best on the data the user hands us (expectation of dirty data).

src/Microsoft.ML.Data/DataLoadSave/Text/TextSaver.cs
```diff
@@ -22,25 +22,33 @@ namespace Microsoft.ML.Data.IO
...
+            internal const bool OutputSchema = true;
+            internal const bool OutputHeader = true;
+        }
+
         // REVIEW: consider saving a command line in a separate file.
         public sealed class Arguments
         {
             [Argument(ArgumentType.AtMostOnce, HelpText = "Separator", ShortName = "sep")]
-            public string Separator = "tab";
+            public string Separator = DefaultArguments.Separator.ToString();
```
artidoro: I think that what we had before was correct, I think that `'\t'.ToString()` simply gives a tab as a string. Here it was meant to be an understandable description of separators. We since moved to separator chars in TextLoader and such, in which case we use the character directly to define the separator. Here I believe we have to keep it like it was "tab".

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
TomFinley: To me it's fairly simpler than that. If I think of the user questions and scenarios that I've handled with text loader over the years, and I tally up (1) those datasets that conformed to *our* scheme of quoting and (2) those that did not, there's a really clear tendency. I also think of all the problems it's caused over the years.

src/Microsoft.ML.Data/DataLoadSave/Text/TextSaver.cs
wschin: Ok.

---
In reply to: [258714604](https://github.com/dotnet/machinelearning/pull/2630#discussion_r258714604) [](ancestors = 258714604)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
wschin: Sure.

---
In reply to: [258307575](https://github.com/dotnet/machinelearning/pull/2630#discussion_r258307575) [](ancestors = 258307575)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -144,20 +148,22 @@ public static IDataView ReadFromTextFile(this DataOperationsCatalog catalog, str
         /// <param name="headerRow">Whether to write the header row.</param>
         /// <param name="schema">Whether to write the header comment with the schema.</param>
         /// <param name="keepHidden">Whether to keep hidden columns in the dataset.</param>
+        /// <param name="forceDense">Whether to save columns in dense format even if they are sparse vectors.</param>
         public static void SaveAsText(this DataOperationsCatalog catalog,
             IDataView data,
             Stream stream,
             char separatorChar = TextLoader.DefaultArguments.Separator,
```
artidoro: Should this be` TextSaver.DefaultArguments.Separator`?

artidoro: Actually we should match the signature of `TextLoader` which uses a separatorChar. If this logic works, then my comment about the need to use "tab" instead of `'\t'.ToString()` was wrong

---
In reply to: [259067133](https://github.com/dotnet/machinelearning/pull/2630#discussion_r259067133) [](ancestors = 259067133)

artidoro:
Could you also rebase/merge with master as I think I made some conflicting changes to `TextLoader`.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
wschin: If it didn't work, some tests may fail. I agree char is better.

---
In reply to: [259071246](https://github.com/dotnet/machinelearning/pull/2630#discussion_r259071246) [](ancestors = 259071246,259067133)

wschin:
Done. Hopefully that's my last time of doing so for this PR.....

---
In reply to: [466127204](https://github.com/dotnet/machinelearning/pull/2630#issuecomment-466127204) [](ancestors = 466127204)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
wschin: With changes made (probably) @artidoro, I have dropped the new regexp.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -7,6 +7,7 @@
 using System.Linq;
 using System.Reflection;
 using System.Text;
+using System.Text.RegularExpressions;
```
eerhardt: This isn't necessary anymore, right?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1068,15 +1069,19 @@ private bool HasHeader
         /// <param name="hasHeader">Whether the file has a header.</param>
         /// <param name="separatorChar"> The character used as separator between data points in a row. By default the tab character is used as separator.</param>
         /// <param name="dataSample">Allows to expose items that can be used for reading.</param>
-        internal TextLoader(IHostEnvironment env, Column[] columns, bool hasHeader = false, char separatorChar = '\t', IMultiStreamSource dataSample = null)
-            : this(env, MakeArgs(columns, hasHeader, new[] { separatorChar }), dataSample)
+        /// <param name="allowSparse">Whether the file can contain numerical vectors in sparse format.</param>
+        /// <param name="allowQuoting">Whether the file can contain numerical vectors in sparse format.</param>
```
eerhardt: copy-paste error here. I don't think this is what `allowQuoting` means :)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -19,12 +19,16 @@ public static class TextLoaderSaverCatalog
...
+        /// <param name="allowSparse">Whether the file can contain numerical vectors in sparse format.</param>
+        /// <param name="allowQuoting">Whether the file can contain column defined by a quoted string.</param>
         public static TextLoader CreateTextLoader(this DataOperationsCatalog catalog,
             TextLoader.Column[] columns,
             bool hasHeader = TextLoader.Defaults.HasHeader,
             char separatorChar = TextLoader.Defaults.Separator,
-            IMultiStreamSource dataSample = null)
-            => new TextLoader(CatalogUtils.GetEnvironment(catalog), columns, hasHeader, separatorChar, dataSample);
+            IMultiStreamSource dataSample = null,
+            bool allowSparse = TextLoader.Defaults.AllowSparse,
```
eerhardt: Should these go before `dataSample`?

test/data/adult.tiny.with-schema.txt
```diff
@@ -1,5 +1,6 @@
 #@ TextLoader{
 #@   header+
+#@   sparse+
```
eerhardt: What is sparse about this dataset?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -19,12 +19,16 @@ public static class TextLoaderSaverCatalog
         /// <param name="hasHeader">Whether the file has a header.</param>
         /// <param name="separatorChar">The character used as separator between data points in a row. By default the tab character is used as separator.</param>
         /// <param name="dataSample">The optional location of a data sample. The sample can be used to infer column names and number of slots in each column.</param>
+        /// <param name="allowSparse">Whether the file can contain numerical vectors in sparse format.</param>
+        /// <param name="allowQuoting">Whether the file can contain column defined by a quoted string.</param>
```
eerhardt: `ReadFromTextFile` calls these options:

```
            bool allowQuotedStrings = TextLoader.Defaults.AllowQuoting,
            bool supportSparse = TextLoader.Defaults.AllowSparse,
```

We should be consistent in the names everywhere.

wschin: Ok. I also checked other `quote` and `sparse` in this file.

---
In reply to: [259422639](https://github.com/dotnet/machinelearning/pull/2630#discussion_r259422639) [](ancestors = 259422639)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
wschin: Thanks. Got it changed to `/// <param name="allowQuoting">Whether the content of a column can be parsed from a string starting and ending with quote.</param>`.


---
In reply to: [259418380](https://github.com/dotnet/machinelearning/pull/2630#discussion_r259418380) [](ancestors = 259418380)

test/data/adult.tiny.with-schema.txt
wschin: Removed.

---
In reply to: [259421045](https://github.com/dotnet/machinelearning/pull/2630#discussion_r259421045) [](ancestors = 259421045)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
wschin: I reorder those arguments based on their usage frequencies (in my mind). In general, the more ML.NET-specific the later an argument appears.

---
In reply to: [259418984](https://github.com/dotnet/machinelearning/pull/2630#discussion_r259418984) [](ancestors = 259418984)

CESARDELATORRE:
We had a issue in one of the samples when migrating to 0.11 because the Label column in the dataset file had numeric values in quotes. Best workaround was to remove the quotes from the dataset file...

https://github.com/dotnet/machinelearning/issues/2821

PR #2632, opened 2019-02-19
Internalization of OneToOne and ManyToOne Column classes

### LINKED ISSUE #2631: Internalization of OneToOne and ManyToOne Column classes
https://github.com/dotnet/machinelearning/blob/25404b83df1b6bf9f548ee59396a87e43f0f9058/src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs#L17

https://github.com/dotnet/machinelearning/blob/25404b83df1b6bf9f548ee59396a87e43f0f9058/src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs#L106

https://github.com/dotnet/machinelearning/blob/25404b83df1b6bf9f548ee59396a87e43f0f9058/src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs#L112

Should be hidden

codemzs:
fixes #2631

### DIFF
--- a/src/Microsoft.ML.Transforms/GcnTransform.cs
+++ b/src/Microsoft.ML.Transforms/GcnTransform.cs
@@ -77,6 +77,7 @@ namespace Microsoft.ML.Transforms.Projections
             public float Scale = LpNormalizingEstimatorBase.Defaults.Scale;
         }
 
+        [BestFriend]
         internal abstract class ColumnBase : OneToOneColumn
         {
             [Argument(ArgumentType.AtMostOnce, HelpText = "Subtract mean from each value before normalizing")]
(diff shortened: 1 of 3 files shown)


src/Microsoft.ML.Transforms/GcnTransform.cs
```diff
@@ -77,6 +77,7 @@ internal sealed class GcnOptions : TransformInputBase
             public float Scale = LpNormalizingEstimatorBase.Defaults.Scale;
         }
 
+        [BestFriend]
```
yaeldekel: >BestFriend [](start = 9, length = 10)

Why is this needed?

codemzs: You are right, its not needed here.

Issue #2593, opened 2019-02-17
Docs: LoadColumn attribute mentions the column name, but no means to provide it


pkulikov:
The [LoadColumnAttribute](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.loadcolumnattribute?view=ml-dotnet) class description claims:
https://github.com/dotnet/machinelearning/blob/cf7e3d992b4b770a7345174673107c47e8bf4d13/src/Microsoft.ML.Data/DataLoadSave/Text/LoadColumnAttribute.cs#L10-L15

However, there are no means to provide the name of a column. Available three constructors only allow to specify the source indices.

That looks to be a copy-paste error from the [ColumnAttribute](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.columnattribute?view=ml-dotnet) documentation.

As I see from the usage of the `LoadColumn` attribute:

https://github.com/dotnet/machinelearning/blob/cf7e3d992b4b770a7345174673107c47e8bf4d13/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L1336-L1345

the `LoadColumn` attribute is used to define only the source indices of a column, while the name is defined either by the `ColumnName` attribute or the property name.

Ivanidzo4ka:
This need to be cleaned up.
First `ColumnAttribute` need to be deleted. It's pointless at current moment.
Second documentation for `LoadColumnAttribute` and `ColumnNameAttribute` need to be update.

Part of me want's to have option to provide Name in `LoadColumnAttribute`, but that would bring confusion what to do if you have two attributes on top of field/property `LoadColumn` and `ColumnName`. It's unclear which one should have more priority. 
In same time  we can't get rid of `ColumnName` because you can use it to decorate your custom class and it's useful.

Ivanidzo4ka:
If you want to chime in regarding wording for this attributes feel free to join https://github.com/dotnet/machinelearning/pull/2638

Issue #2437, opened 2019-02-06
VariableVectorType attribute


zeahmed:
`VectorType(0)` is currently used to designate a variable in a type as variable length vector e.g.

``` csharp
public class IMDBSentiment
{
    public string Sentiment_Text { get; set; }

    /// <summary>
    /// This is a variable length vector designated by VectorType(0) attribute.
    /// Variable length vectors are produced by applying operations such as 'TokenizeWords' on strings
    /// resulting in vectors of tokens of variable lengths.
    /// </summary>
    [VectorType(0)]
    public int[] VariableLenghtFeatures { get; set; }
}
```

This seems not very intuitive. Can we have attribute type that can be specifically used to designed variable length vectors? such as `VariableVectorType`

PR #2638, opened 2019-02-19
Cleaning our attributes classes

Ivanidzo4ka:
Fixes: https://github.com/dotnet/machinelearning/issues/2593
Fixes: https://github.com/dotnet/machinelearning/issues/2437


src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -389,9 +402,14 @@ public static SchemaDefinition Create(Type userType, Direction direction = Direc
...
                     throw Contracts.ExceptParam(nameof(userType), "Member {0} marked with VectorType attribute, but does not appear to be a vector type", memberInfo.Name);
+                var varVectorAttr = memberInfo.GetCustomAttribute<VariableVectorTypeAttribute>();
+                if (varVectorAttr != null && !isVector)
+                    throw Contracts.ExceptParam(nameof(userType), "Member {0} marked with VariableVectorType attribute, but does not appear to be a vector type", memberInfo.Name);
+                if (vectorAttr != null && varVectorAttr != null)
+                    throw Contracts.ExceptParam(nameof(userType), "Member {0} marked with both VariableVectorType and VectorType attribute", memberInfo.Name);
                 if (isVector)
                 {
-                    int[] dims = vectorAttr?.Dims;
+                    int[] dims = varVectorAttr != null ? new int[0] : vectorAttr?.Dims;
```
Ivanidzo4ka: >vectorAttr?.Dims; [](start = 70, length = 17)

This one makes me question do we even needed VariableVectorType attribute.
In past, and now we just treat any array as variable one, and presence of VariableVectorType is a bit wacky.
I can probably force system to fail if it's a vector and none of attributes is present.
Or I can just specify in documentation for VectorType what absence of VectorType means this array has unfixed size.

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -17,50 +17,65 @@ namespace Microsoft.ML.Data
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property, AllowMultiple = false, Inherited = true)]
     public sealed class KeyTypeAttribute : Attribute
     {
-        // REVIEW: Property based, but should I just have a constructor?
+        /// <summary>
+        /// Specifies expected size of key type.
+        /// </summary>
+        /// <param name="count">Size of key type.</param>
+        public KeyTypeAttribute(ulong count)
```
TomFinley: >KeyTypeAttribute [](start = 15, length = 16)

Can this one mirror our existing constructors for the key type itself, that is, parameterless constructor where the result winds up being a key type with the maximum applicable count for the type specified?

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -17,50 +17,65 @@ namespace Microsoft.ML.Data
...
+        /// Mark member with expected size of array.
+        /// </summary>
+        /// <param name="size">Expected size of array.</param>
+        public VectorTypeAttribute(int size)
+        {
+            Contracts.CheckParam(size > 0, nameof(size), "Should be positive number, if you want to specify variable size please use " + nameof(VariableVectorTypeAttribute));
+            Dims = new int[1] { size };
+        }
 
         public VectorTypeAttribute(params int[] dims)
```
TomFinley: >dims [](start = 48, length = 4)

So long as you're cleaning up the public surface, even though it's a parameter name we should probably still follow .NET naming guidelines and avoid abbreviations and so on.

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -389,9 +402,14 @@ public static SchemaDefinition Create(Type userType, Direction direction = Direc
                 var vectorAttr = memberInfo.GetCustomAttribute<VectorTypeAttribute>();
                 if (vectorAttr != null && !isVector)
                     throw Contracts.ExceptParam(nameof(userType), "Member {0} marked with VectorType attribute, but does not appear to be a vector type", memberInfo.Name);
+                var varVectorAttr = memberInfo.GetCustomAttribute<VariableVectorTypeAttribute>();
+                if (varVectorAttr != null && !isVector)
+                    throw Contracts.ExceptParam(nameof(userType), "Member {0} marked with VariableVectorType attribute, but does not appear to be a vector type", memberInfo.Name);
```
TomFinley: >VariableVectorType [](start = 90, length = 18)

Elsewhere I observe you correctly using `nameof`, maybe shift to that here?

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
Ivanidzo4ka: oops

---
In reply to: [258276176](https://github.com/dotnet/machinelearning/pull/2638#discussion_r258276176) [](ancestors = 258276176)

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
Ivanidzo4ka: So basically no size here, and in place where we consume this count, just look on member type and use maxvalue for that type.
I can do that.
But there is small issue with MF I think.
It actually rely on key size and have loops based on size of key.
Which can blew up any MF calculation.

@wschin is it possible to change MF code to not rely on key size?

---
In reply to: [258276031](https://github.com/dotnet/machinelearning/pull/2638#discussion_r258276031) [](ancestors = 258276031)

src/Microsoft.ML.Data/DataLoadSave/Text/LoadColumnAttribute.cs
```diff
@@ -8,14 +8,14 @@
 namespace Microsoft.ML.Data
 {
     /// <summary>
-    /// Describes column information such as name and the source columns indices that this
-    /// column encapsulates.
+    /// Allow member to specify mapping to column(s) in text file.
```
TomFinley: >column [](start = 43, length = 6)

Please do not use the word "column" here since that has an overloaded context in `IDataView` itself.

What I tend to favor, given that a different term is required, is what appears in [Wikipedia's description of CSV](https://en.wikipedia.org/wiki/Comma-separated_values), specifically, "field."

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
TomFinley: I think it's fine to keep specifying the key size. Just so long as we have another constructor where it is not specified.

---
In reply to: [258277629](https://github.com/dotnet/machinelearning/pull/2638#discussion_r258277629) [](ancestors = 258277629,258276031)

src/Microsoft.ML.Data/DataLoadSave/Text/LoadColumnAttribute.cs
TomFinley: Here and everywhere. I notice this is a problem with all the new documentation is this bad usage of the word "column." This would be troublesome anywhere, but it is especially troublesome since this is literally on a attribute where column is used to describe a rather entirely different concept. :P

---
In reply to: [258283784](https://github.com/dotnet/machinelearning/pull/2638#discussion_r258283784) [](ancestors = 258283784)

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -17,50 +17,76 @@ namespace Microsoft.ML.Data
...
+
+        /// <summary>
+        /// Marks member as <see cref="KeyType"/>.
+        /// </summary>
+        public KeyTypeAttribute()
+        {
+
+        }
+        /// <summary>
+        /// Marks member as <see cref="KeyType"/> and specifies <see cref="KeyType"/> cordinality.
```
pkulikov: Typo:
```suggestion
        /// Marks member as <see cref="KeyType"/> and specifies <see cref="KeyType"/> cardinality.
```

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -70,19 +96,18 @@ public ColumnAttribute(string ordinal, string name = null)
...
-        private readonly string _name;
         /// <summary>
         /// Column name.
         /// </summary>
-        public string Name { get { return _name; } }
+        internal string Name { get; }
 
         /// <summary>
         /// Allows one to specify a name to expose this column as, as opposed to simply
         /// the field name.
```
pkulikov: The attribute may be applied to a property as well:
```suggestion
        /// the field or property name.
```

artidoro:
>     /// Attach to a member of a class to indicate that the item type should be of class key.

maybe replace the end of the comment with:
"... should be <see cref="KeyType"/>.

---
Refers to: src/Microsoft.ML.Data/Data/SchemaDefinition.cs:15 in d4a1f58. [](commit_id = d4a1f58c6bdefe5aa1baf99bd10dd436b0c8dee8, deletion_comment = False)

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -17,50 +17,76 @@ namespace Microsoft.ML.Data
...
     public sealed class KeyTypeAttribute : Attribute
     {
-        // REVIEW: Property based, but should I just have a constructor?
+
+        /// <summary>
+        /// Marks member as <see cref="KeyType"/>.
+        /// </summary>
+        public KeyTypeAttribute()
+        {
+
```
artidoro: Nit: if you could remove the line in between the braces and add lines between constructors.

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -17,50 +17,76 @@ namespace Microsoft.ML.Data
...
+        /// <summary>
+        /// Mark member with expected size of array.
+        /// </summary>
+        /// <param name="size">Expected size of array.</param>
+        public VectorTypeAttribute(int size)
+        {
+            Contracts.CheckParam(size > 0, nameof(size), "Should be positive number, if you want to specify variable size please use " + nameof(VariableVectorTypeAttribute));
+            Dims = new int[1] { size };
+        }
+        /// <summary>
```
artidoro: >/// [](start = 8, length = 3)

nit: add a line between the constructors.

test/Microsoft.ML.Tests/TextLoaderTests.cs
```diff
@@ -700,7 +700,7 @@ public class IrisStartEnd
 
         public class IrisColumnIndices
         {
-            [LoadColumn(columnIndexes: new[] { 0, 2 })]
+            [LoadColumn( new []{0, 2})]
```
artidoro: > new []{0, 2}) [](start = 24, length = 14)

nit: (new[] { 0, 2 })

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -379,7 +403,10 @@ public static SchemaDefinition Create(Type userType, Direction direction = Direc
                 {
                     if (!KeyType.IsValidDataType(dataType))
                         throw Contracts.ExceptParam(nameof(userType), "Member {0} marked with KeyType attribute, but does not appear to be a valid kind of data for a key type", memberInfo.Name);
-                    itemType = new KeyType(dataType, keyAttr.Count);
+                    if (keyAttr.Count == null)
+                        itemType = new KeyType(dataType, dataType.ToMaxInt());
+                    else
+                        itemType = new KeyType(dataType, keyAttr.Count.GetValueOrDefault());
```
artidoro: Instead of rewriting the logic here, I would suggest using `KeyCount` and `ConstructKeyType` in TypeParsingUtils.cs as those are the objects and methods used to specify KeyTypes.

Ivanidzo4ka: ConstructKeyType  uses `DataKind` not type. But I like idea of KeyCount

---
In reply to: [258663366](https://github.com/dotnet/machinelearning/pull/2638#discussion_r258663366) [](ancestors = 258663366)

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -12,82 +12,100 @@
...
         /// <summary>
         /// Column name.
         /// </summary>
-        public string Name { get { return _name; } }
+        internal string Name { get; }
 
         /// <summary>
-        /// Allows one to specify a name to expose this column as, as opposed to simply
-        /// the field name.
+        /// Allows one to specify a name to expose this column as, as opposed as opposed to the default
```
pkulikov: Typo: duplicate "as opposed"

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
TomFinley: Maybe some people get nervous without explicit attributes identifying types, I don't know. I am unable to trace back its addition or any discussion regarding it. If we wanted an explicit attribute for *some* reason, maybe we can just get away with vector type attribute with no arguments? Which I think is also how our vector type actually does behave.

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
TomFinley: So, I think it's fine for MF to fail with an informative error message if it an input is somehow judged as "too big."

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -12,82 +12,101 @@
...
+
+        }
+
+        /// <summary>
+        /// Marks member as <see cref="KeyType"/> and specifies <see cref="KeyType"/> cardinality.
+        /// </summary>
+        /// <param name="count">Size of key type.</param>
+        public KeyTypeAttribute(ulong count)
+        {
+            Count = new KeyCount(count);
```
artidoro: >Count [](start = 12, length = 5)

If you make another iteration, maybe this could be renamed to `KeyCount` to reflect its type. Not a blocking comment though.

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
artidoro: Looks good! I guess you are right. If you feel like it you can extract that logic out of this method and `ConstructKeyType` but not sure it's worth it for four lines. 

---
In reply to: [258671672](https://github.com/dotnet/machinelearning/pull/2638#discussion_r258671672) [](ancestors = 258671672,258663366)

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -12,82 +12,108 @@
...
+        /// </remarks>
+        public KeyTypeAttribute()
+        {
+
+        }
+
+        /// <summary>
+        /// Marks member as <see cref="KeyType"/> and specifies <see cref="KeyType"/> cardinality.
+        /// </summary>
+        /// <param name="count">Cardinality of key type.</param>
```
artidoro: >key type [](start = 47, length = 8)

Could you cref this reference to KeyType?

Ivanidzo4ka:
@TomFinley @yaeldekel  @zeahmed  can you take a look on this PR?

Issue #2554, opened 2019-02-14
ColumnInfo as an API parameter


rogancarr:
The use of `ColumnInfo` in the public API can be rather jarring. I'd like to clarify when we should use it in the catalog API, and if the parameter name could be made to be more consistent with the API, such as `ColumnOptions` or `Options`.

Ivanidzo4ka:
Whatever we decide to call it, we also have `MeanVarColumn`, `LogMeanVarColumn`, `BinningColumn`, `SupervisedBinningColumn` and `MinMaxColumn` in `NormalizingEstimator` so they should be renamed to same thing. (Easy to miss if we sweep just ColumnInfo).

artidoro:
I think the suggestion `ColumnOptions` could work. It relates to the naming convention that we use for trainers, and it also indicates that this is column-specific compared to the trainers.

I would like however to hear @TomFinley 's opinion on this.

TomFinley:
I could definitely get behind `ColumnOptions`. It seems a far superior choice, so, let me say I agree wholeheartedly.

PR #2641, opened 2019-02-19
Remove value tuples from public TransformCatalog APIs

### LINKED ISSUE #2581: Remove value tuples from TransformsCatalog methods
Many of the methods in TransformsCatalog create estimators using a params array of (string outputColumnName, string inputColumnName). As mentioned in #2501, we should change all these to not use value tuples.

yaeldekel:
Fixes #2581 .

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
+++ b/src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
@@ -2,12 +2,36 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Linq;
 using Microsoft.Data.DataView;
 using Microsoft.ML.Data;
 using Microsoft.ML.Transforms;
 
 namespace Microsoft.ML
 {
+    public sealed class SimpleColumnInfo
+    {
+        private readonly string _outputColumnName;
+        private readonly string _inputColumnName;
+
+        public SimpleColumnInfo(string outputColumnName, string inputColumnName)
+        {
+            _outputColumnName = outputColumnName;
+            _inputColumnName = inputColumnName;
+        }
+
+        public static implicit operator SimpleColumnInfo((string outputColumnName, string inputColumnName) value)
+        {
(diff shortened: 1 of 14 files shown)


src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
```diff
@@ -2,12 +2,36 @@
...
 // See the LICENSE file in the project root for more information.
 
+using System.Linq;
 using Microsoft.Data.DataView;
 using Microsoft.ML.Data;
 using Microsoft.ML.Transforms;
 
 namespace Microsoft.ML
 {
+    public sealed class SimpleColumnInfo
```
Ivanidzo4ka: >SimpleColumnInfo [](start = 24, length = 16)

I think @rogancarr has issue where he don't like ColumnInfo name. [https://github.com/dotnet/machinelearning/issues/2554](https://github.com/dotnet/machinelearning/issues/2554)
I wonder how much he would hate this one.

`SimpleMapping`,`SimplePair`?

rogancarr: No strong opinion, I just think that we should have a conversation about naming these things :)

---
In reply to: [258285683](https://github.com/dotnet/machinelearning/pull/2641#discussion_r258285683) [](ancestors = 258285683)

yaeldekel: How about `ColumnNamePair` or something similar? I think something containing 'mapping' or 'pair' would be too general, wouldn't it?

Or I could leave it as `SimpleColumnInfo` and add a comment about it in the issue so we can change it when we find a good name for `ColumnInfo`?

---
In reply to: [258297894](https://github.com/dotnet/machinelearning/pull/2641#discussion_r258297894) [](ancestors = 258297894,258285683)

Ivanidzo4ka: I added mention of this column to #2554.
Let's merge this one, and figure out proper naming in that issue.

PR #2647, opened 2019-02-20
Increase build timeout for code coverage CI.

codemzs:
Recently it seems the code coverage build times have increased, the exact reason is unclear but it could be because of added tests. Since code coverage CI does not interfere with the main CI so its ok to increase the build timeouts for code coverage CI to prevent build failures due to [timeouts.](https://dev.azure.com/dnceng/public/_build/results?buildId=100301)

### DIFF
--- a/build/ci/phase-template.yml
+++ b/build/ci/phase-template.yml
@@ -15,9 +15,9 @@ phases:
       _codeCoverage: ${{ parameters.codeCoverage }}
     queue:
       ${{ if eq(variables._codeCoverage, 'false') }}:
-        timeoutInMinutes: 45
+        timeoutInMinutes: 30
       ${{ if eq(variables._codeCoverage, 'true') }}:
-        timeoutInMinutes: 90
+        timeoutInMinutes: 60
       parallel: 99
       matrix:
         ${{ if eq(parameters.customMatrixes, '') }}:
--- a/build/ci/phase-template.yml
+++ b/build/ci/phase-template.yml
@@ -14,7 +14,10 @@ phases:
       _arch: ${{ parameters.architecture }}
       _codeCoverage: ${{ parameters.codeCoverage }}
     queue:
-      timeoutInMinutes: 45
+      ${{ if eq(variables._codeCoverage, 'false') }}:
+        timeoutInMinutes: 45
+      ${{ if eq(variables._codeCoverage, 'true') }}:
+        timeoutInMinutes: 90
       parallel: 99
       matrix:
         ${{ if eq(parameters.customMatrixes, '') }}:


build/ci/phase-template.yml
```diff
@@ -14,7 +14,10 @@ phases:
       _arch: ${{ parameters.architecture }}
       _codeCoverage: ${{ parameters.codeCoverage }}
     queue:
-      timeoutInMinutes: 45
+      ${{ if eq(variables._codeCoverage, 'false') }}:
+        timeoutInMinutes: 45
```
Ivanidzo4ka: >45 [](start = 26, length = 2)

I've look on few recent builds and non coverage builds average 20, longest 25 minutes.
with code coverage it takes around 43.
Can we have reasonable 30/60 instead of 45/90?
If I have to wait hour and a half for build, we probably doing something wrong with our tests.

codemzs: Did you read the description of this PR? CodeCovergae CI  DOES NOT affect merge process and this has been made clear before as well, i.e you don’t have to wait for Code Coverage  build to finish to merge your PR. Regarding 45 mins for regular CI that has been the value we have always used, making it 30 minutes isn’t going to speed up the builds, may be I’m missing your point here so please feel free to clarify.

Ivanidzo4ka: >DOES NOT affect merge process 

It's doesn't give you big green checks as well. 

I see you trying to update build numbers.
Looks like you have assumption what code coverage build works twice amount of time than average build.
We used to have 45 minutes for both of them. Now we will have 45/90.

I don't know where 45 come from. Maybe from the times we had deadlocks in our code and fasttree test took 20 minutes to run.
I also see what on average build works 20 minutes. So this is why I propose 30/60 timeouts.
I also see builds like this: [https://dev.azure.com/dnceng/public/_build/results?buildId=100119](https://dev.azure.com/dnceng/public/_build/results?buildId=100119) see core30 Build_Release_Intrinsics which got stuck for some reason. 45 minute timeout just force us to wait till it cancel itself by timeout.

I understand it's currently on the edge of timeout. Let's increase it by twice is definitely solution. I just not sure it's a good one. 

---
In reply to: [258347010](https://github.com/dotnet/machinelearning/pull/2647#discussion_r258347010) [](ancestors = 258347010)

build/ci/phase-template.yml
```diff
@@ -14,7 +14,10 @@ phases:
...
+      ${{ if eq(variables._codeCoverage, 'true') }}:
+        timeoutInMinutes: 90
```
artidoro: I think 90 min might be too much if the builds take around 45 min. We unfortunately still have some builds that get stuck and the build completion time does not oscillate that much when they terminate. I would increase it to 60 min as Ivan suggested.

build/ci/phase-template.yml
artidoro: I agree with Ivan, it would be great to reduce the CI build timeout, but maybe not the scope of this PR.

build/ci/phase-template.yml
safern: Yes, I scanned a bunch of builds and the avg time is 42 mins. Let's start with 60 mins and see how it goes.

build/ci/phase-template.yml
codemzs: OK.

build/ci/phase-template.yml
codemzs: Done

PR #2651, opened 2019-02-20
Metadata utils internalization, migration of few useful methods

### LINKED ISSUE #2622: Metadata/annotations public surface of the API
Metadata (FYI to be later named annotations per #2297) is an essential mechanism for attaching optional information about columns. This ranges from publicly facing stuff that user's should be aware of (slot names, which we use for feature names on feature columns, and key values), versus a bunch of stuff that is arguably useful for users but primarily for our internal infrastructure (e.g., whether something has already been normalized), versus stuff intended purely for our internal infrastructure.

We ought to decide what we really want to be part of our initial public surface (as small as possible but no smaller), and internalize the rest of it.

So, we will keep as is `Metadata` (to be `Annotations`):

https://github.com/dotnet/machinelearning/blob/a56caeebaaa0076d6940bfdede90a4eb0a351a20/src/Microsoft.Data.DataView/DataViewSchema.cs#L172

This by itself is little more than an arbitrary string/object store, which is as intended. So that will not change. What will change however is the class we've made to make access a little more structured.

https://github.com/dotnet/machinelearning/blob/a56caeebaaa0076d6940bfdede90a4eb0a351a20/src/Microsoft.ML.Core/Data/MetadataUtils.cs#L17

This has stuff that is "good" in that we want to keep it as part of the public surface, but also stuff that is internal and should not be part of the public surface.

## The good

A small amount of this stuff we probably want to keep.

However we should probably move it somewhere else... perhaps, the static class `SchemaColumnAnnotationsExtensions` as a series of extension methods on top of `DataViewSchema.Column` to access the associated metadata.

This might include things like these methods.

https://github.com/dotnet/machinelearning/blob/a56caeebaaa0076d6940bfdede90a4eb0a351a20/src/Microsoft.ML.Core/Data/MetadataUtils.cs#L297

https://github.com/dotnet/machinelearning/blob/a56caeebaaa0076d6940bfdede90a4eb0a351a20/src/Microsoft.ML.Core/Data/MetadataUtils.cs#L321

## The bad

Much of this class though should be internal.

So for example, we have this static class of `Kinds` of metadata. Absolutely
quite nice a thing to have for our own infrastructure for consistency, but this is not what we want to show users. Similar with this sort of labels for types of scorings, which is a scenario irrelevant to the ML.NET API as defined (since people evaluate scores by saying, "here, evaluate these scores" explicitly by calling some code). Also stuff on the ranges of categorical variables which, while essential, are mostly for the benefit of trainers downstream consuming data. (User's that want the raw categoricals can, by just programming, consume the source data themselves, since they control the pipeline.)

There's also a lot of stuff built around *implementing* metadata, which is of questionable worth at this time given the changes that have happened to schema in the past year, and which is of no use whatever.

/cc @Ivanidzo4ka , @eerhardt , @rogancarr , @sfilipi

TomFinley:
Fixes #2622 by limiting the public surface area of metadata/annotations conveniences to a few places.

Note that that migrated methods are not exactly the same as the existing methods in the case of querying whether a key column type exists, since the existing methods predated the schema column type and therefore had some unnecessary parameters. Though I decided to expand it a bit so as to allow checks on whether other types of key values are present. Not sure if that's necessary or helpful.

Also internalized the rather innocuous but nonetheless probably unnecessary metadata builder extensions.

Commits structured incrementally as usual for the small steps.

### DIFF
--- a/src/Microsoft.ML.Core/Data/MetadataUtils.cs
+++ b/src/Microsoft.ML.Core/Data/MetadataUtils.cs
@@ -14,7 +14,8 @@ namespace Microsoft.ML.Data
     /// <summary>
     /// Utilities for implementing and using the metadata API of <see cref="DataViewSchema"/>.
     /// </summary>
-    public static class MetadataUtils
+    [BestFriend]
+    internal static class MetadataUtils
     {
         /// <summary>
         /// This class lists the canonical metadata kinds
@@ -114,20 +115,17 @@ namespace Microsoft.ML.Data
         /// <summary>
         /// Returns a standard exception for responding to an invalid call to GetMetadata.
         /// </summary>
-        [BestFriend]
-        internal static Exception ExceptGetMetadata() => Contracts.Except("Invalid call to GetMetadata");
+        public static Exception ExceptGetMetadata() => Contracts.Except("Invalid call to GetMetadata");
 
         /// <summary>
         /// Returns a standard exception for responding to an invalid call to GetMetadata.
         /// </summary>
-        [BestFriend]
-        internal static Exception ExceptGetMetadata(this IExceptionContext ctx) => ctx.Except("Invalid call to GetMetadata");
+        public static Exception ExceptGetMetadata(this IExceptionContext ctx) => ctx.Except("Invalid call to GetMetadata");
 
         /// <summary>
         /// Helper to marshal a call to GetMetadata{TValue} to a specific type.
         /// </summary>
-        [BestFriend]
-        internal static void Marshal<THave, TNeed>(this MetadataGetter<THave> getter, int col, ref TNeed dst)
+        public static void Marshal<THave, TNeed>(this MetadataGetter<THave> getter, int col, ref TNeed dst)
         {
             Contracts.CheckValue(getter, nameof(getter));
 
@@ -141,8 +139,7 @@ namespace Microsoft.ML.Data
         /// Returns a vector type with item type text and the given size. The size must be positive.
         /// This is a standard type for metadata consisting of multiple text values, eg SlotNames.
         /// </summary>
-        [BestFriend]
-        internal static VectorType GetNamesType(int size)
+        public static VectorType GetNamesType(int size)
         {
             Contracts.CheckParam(size > 0, nameof(size), "must be known size");
             return new VectorType(TextDataViewType.Instance, size);
@@ -154,8 +151,7 @@ namespace Microsoft.ML.Data
         /// This is a standard type for metadata consisting of multiple int values that represent
         /// categorical slot ranges with in a column.
         /// </summary>
-        [BestFriend]
-        internal static VectorType GetCategoricalType(int rangeCount)
+        public static VectorType GetCategoricalType(int rangeCount)
         {
             Contracts.CheckParam(rangeCount > 0, nameof(rangeCount), "must be known size");
             return new VectorType(NumberDataViewType.Int32, rangeCount, 2);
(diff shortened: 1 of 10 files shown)


src/Microsoft.ML.Data/Evaluators/MultiClassClassifierEvaluator.cs
```diff
@@ -998,16 +998,17 @@ private protected override IEnumerable<string> GetPerInstanceColumnsToSave(RoleM
         // Multi-class evaluator adds four per-instance columns: "Assigned", "Top scores", "Top classes" and "Log-loss".
         private protected override IDataView GetPerInstanceMetricsCore(IDataView perInst, RoleMappedSchema schema)
         {
-            // If the label column is a key without key values, convert it to I8, just for saving the per-instance
+            // If the label column is a key without text key values, convert it to I8, just for saving the per-instance
```
yaeldekel: I noticed that the code actually converts to R8, not I8.

TomFinley: Ahhh. Delightful... seems a pity to update such a simple PR for an internal class. How about this, I'll locally fix it, and if for some other reason I need to update the PR it'll go in, otherwise it'll just go into my next PR.

src/Microsoft.ML.Data/EntryPoints/PredictorModelImpl.cs
```diff
@@ -127,7 +127,7 @@ internal override string[] GetLabelInfo(IHostEnvironment env, out DataViewType l
             if (trainRms.Label != null)
             {
                 labelType = trainRms.Label.Value.Type;
```
codemzs: >labelType = trainRms.Label.Value.Type; [](start = 16, length = 38)

Dead code.

src/Microsoft.ML.Data/Data/SchemaAnnotationsExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+        public static void GetKeyValues<TValue>(this DataViewSchema.Column column, ref VBuffer<TValue> keyValues)
+            => column.Metadata.GetValue(MetadataUtils.Kinds.KeyValues, ref keyValues);
+
+        /// <summary>
+        /// Returns <see langword="true"/> if and only if <paramref name="column"/> has <c>IsNormalized</c> metadata
+        /// set to <see langword="true"/>.
+        /// </summary>
+        public static bool IsNormalized(this DataViewSchema.Column column)
+        {
+            var metaColumn = column.Metadata.Schema.GetColumnOrNull((MetadataUtils.Kinds.IsNormalized));
```
eerhardt: (super nit): double params here. Can remove one set.

src/Microsoft.ML.Core/Data/MetadataUtils.cs
```diff
@@ -337,55 +307,24 @@ internal static void GetSlotNames(RoleMappedSchema schema, RoleMappedSchema.Colu
...
-            return
-                metaColumn != null
-                && metaColumn.Value.Type is VectorType vectorType
-                && keyCount == (ulong)vectorType.Size
-                && vectorType.ItemType is TextDataViewType;
-        }
-
-        [BestFriend]
-        internal static bool HasKeyValues(this SchemaShape.Column col)
+        public static bool HasKeyValues(this SchemaShape.Column col)
```
eerhardt: Just for my information - what's the difference between this method and the `public static bool HasKeyValues` in `SchemaAnnotationsExtensions`?

Issue #2006, opened 2019-01-03
DataKind Enum is not well documented


sfilipi:
This issue was initially logged in the dotnet/docs repo [9278](https://github.com/dotnet/docs/issues/9278) and [9184](https://github.com/dotnet/docs/issues/9184). 

@TomFinley, @CESARDELATORRE, @eerhardt Is the DataKind Enum something that will remain user facing, or will it get behind the .net types?

Should we document the DataKind or cleanup our error messages, and usage to not contain references to the DataKind anymore like #1943 suggest?

cc @mairaw @JRAlexander

TomFinley:
We do not want `DataKind` visible anywhere. This is why `ColumnType` now has it as private. It ought to be deleted. Anywhere currently using `DataKind` should use some other mechanism. The only place we *might* want something like it is a place like `TextLoader` or whatever, to describe what type of item should be loaded from whatever fields of the text file we are reading, but even then it is questionable. It should not be a first class citizen in dataview land.

JRAlexander:
@TomFinley - What should be used in place of DataKind for TextLoader field definition? Is there a best practice that you would recommend as you've said that use case is questionable?

sfilipi:
Since I opened this, to make this issue actionable, and not just a discussion placeholder: 

1- I'll rename this issue to: Substitute internal usage of DataKind by the actual .net respective System.Type. IMO that can be done post Project 13 but please review its pri.  

2- There is already an issue to cleanup the error messages from mentioning `DataKind`.
This is already in Project 13. 

3- Open an issue about investigating the TextLoader to potentially using the respective System.Type to load. Address backwards compatibility.  

cc @shauheen @glebuk @CESARDELATORRE to arrange priorities.

TomFinley:
> 3- Open an issue about investigating the TextLoader to potentially using the respective System.Type to load. Address backwards compatibility.

See, *that* I might like even less than an enum, because it's less restrictive in a bad way. The set of types that the text loader can load is certainly finite. So, it should be somehow enumerated in a finite set, either in a finite set of an enum, or a finite set of calls with specific names (with the suggestion they are associated with specific types -- that type not being one provided by the user). Just clarifying, that while I don't like an enum, there are some things I might like even less than that. :)

PR #2661, opened 2019-02-20
Internalize DataKind

### LINKED ISSUE #2583: DataKind shouldn't be present in our public API.
We need to replace places like this 
https://github.com/dotnet/machinelearning/blob/13b33392dd3fbffc0922b2af72c8236b74c95fc0/src/Microsoft.ML.Data/Transforms/TypeConverting.cs#L542
https://github.com/dotnet/machinelearning/blob/832ecad57808bbe79f36dc6886882781fb5bd399/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L45
to proper .net type support.

wschin:
Fix #2583. In general, we don't want to see `DataKind` in any public area. Depending on the actual amount of works required, this PR might first address cases mentioned #2583 and then more PRs will be published later.

- [x] TextLoader
- [x] TypeConverting
- [x] Rest (DataKind is internal-best-friend now).

Note that to specify target types in type converters and text loader, we introduce a cleaner version of existing `DataKind`,
```csharp
    public enum DataKind : byte 
    { 
        SByte = 1, 
        Byte = 2, 
        Int16 = 3, 
        UInt16 = 4, 
        Int32 = 5, 
        UInt32 = 6, 
        Int64 = 7, 
        UInt64 = 8, 
        Single = 9, 
        Double = 10, 
        String = 11, 
        Boolean = 12, 
        TimeSpan = 13, 
        DateTime = 14, 
        DateTimeOffset = 15, 
    }
```
and the existing `DataKind` is renamed as `InternalDataKind`.
Other user-facing changes were made to work with `DataKind`. Command line and entry point are not affected at all.


docs/samples/Microsoft.ML.Samples/Dynamic/Calibrator.cs
```diff
@@ -34,8 +34,8 @@ public static void Example()
                 HasHeader = true,
                 Columns = new[]
                     {
-                        new TextLoader.Column("Sentiment", DataKind.BL, 0),
-                        new TextLoader.Column("SentimentText", DataKind.Text, 1)
+                        new TextLoader.Column("Sentiment", typeof(bool), 0),
```
TomFinley: In issue #2006, I had [a comment](https://github.com/dotnet/machinelearning/issues/2006#issuecomment-451511350). You will note that you are picking the option that I argued was the worst of all possible worlds. It offers absolutely no hints to the user of what might be an acceptable type. How does a user know that a `bool` is acceptable, but, say, a `List<bool>` is not>? This is much, *much* worse than what it is supposedly meant to replace.

wschin: Ok. I will make `ColumnInfo` strongly-typed based on the desired output type.

---
In reply to: [258657461](https://github.com/dotnet/machinelearning/pull/2661#discussion_r258657461) [](ancestors = 258657461)

TomFinley: That is one possibility. We should also remain open to the possibility that `DataKind` will persist, just in a changed form. So while doing this work, think about whether it is making the usage more or less clear.

It is a perfectly acceptable outcome to say, "you know what, this enumeration is still useful, just in a vastly abbreviated scope," if we can make that argument. It is unclear to me which is correct. I would not start with the assumption that datakind needs to be internalized and destroyed. (W.r.t. its prior role in `ColumnType`, absolutely!!)

wschin: How about this one for a new form of `DataKind`?
```csharp
    public static class ScalarTypes
    {
        public static Type Float => typeof(float);

        public static Type Double => typeof(double);
    }
```

---
In reply to: [258755422](https://github.com/dotnet/machinelearning/pull/2661#discussion_r258755422) [](ancestors = 258755422)

wschin: Discussed offline. We decided internalizing DataKind and creating an simplified alternative.

---
In reply to: [259025888](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259025888) [](ancestors = 259025888,258755422)

Ivanidzo4ka:
Before reviewing this code two questions:
- Are we happy with `ScalarType` name?
- Do we ever plan to use this enum outside of `TextLoader` and `TypeConverting`? If no, I would suggest to have multiple `ConvertToFloat`, `ConvertToDouble`, `ConvertToKey`, etc to replace `TypeConverting` and just move that enum to `TextLoader`

wschin:
@Ivanidzo4ka, your suggestion was one of the 3 solutions I have tried. Unfortunately, having specialized functions cannot tackle the case where multiple columns are casted to different types in one transformer. If you want to specialize `ColumnInfo` to different allowed types, you will end up with TWO groups of those `ColumnInfo` independently for `TextLoader` and `TypeConverting`. Therefore, @TomFinley and I feel it's ok to have a simplified version of `DataKind`.

eerhardt:
> * Are we happy with `ScalarType` name?

I had a similar concern initially. I tried to think of a better name, and the only name I thought was better was `DataKind`. What do you think about renaming the existing `DataKind` to `InternalDataKind` (or anything but `DataKind`), and then rename `public enum ScalarType` to `public enum DataKind`?

wschin:
> 
> 
> > * Are we happy with `ScalarType` name?
> 
> I had a similar concern initially. I tried to think of a better name, and the only name I thought was better was `DataKind`. What do you think about renaming the existing `DataKind` to `InternalDataKind` (or anything but `DataKind`), and then rename `public enum ScalarType` to `public enum DataKind`?

Having a new name sounds reasonable to me.

src/Microsoft.ML.Core/Data/DataKind.cs
```diff
@@ -8,40 +8,83 @@
 namespace Microsoft.ML.Data
 {
     /// <summary>
-    /// Data type specifier.
+    /// Data type specifier used in text loader and type converters.
     /// </summary>
+    /// <remarks>
+    /// <list type="bullet">
+    /// <item><description><see cref="SByte"/>: 1-byte integer, type of <see cref="System.SByte"/>.</description></item>
```
eerhardt: Don't these comments typically go on each individual value in the enum?

```C#
public enum ActionType
{
  /// <summary>The New Subscription is called</summary>
  Insert,
  /// <summary>The Cancel Subscription is called</summary>
  Update,
  /// <summary>The Cancel Subscription is called</summary>
  Cancel, 
  /// <summary>The Get Subscription is called</summary>
  Retrieve
}
```

src/Microsoft.ML.Core/Data/DataKind.cs
```diff
@@ -8,40 +8,83 @@
 namespace Microsoft.ML.Data
 {
     /// <summary>
-    /// Data type specifier.
+    /// Data type specifier used in text loader and type converters.
```
eerhardt: I don't really like giving where a type is used as part of its definition/summary of what it is. What happens if a new place starts using this enum?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -92,10 +98,18 @@ public Column(string name, DataKind? type, Range[] source, KeyCount keyCount = n
...
             [Argument(ArgumentType.AtMostOnce, HelpText = "Type of the items in the column")]
-            public DataKind? Type;
+            [BestFriend]
+            internal InternalDataKind? Type;
+
+            /// <summary>
+            /// <see cref="DataKind"/> of the items in the column.
+            /// </summary>
+            /// It's a public interface to access the information in an internal DataKind.
+            public DataKind ItemType => Type.HasValue ? Type.Value.ToScalarType() : DataKind.Single;
```
eerhardt: `ToScalarType()`?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -50,37 +49,44 @@ public sealed class Column
...
             /// <summary>
             /// Describes how an input column should be mapped to an <see cref="IDataView"/> column.
             /// </summary>
             /// <param name="name">Name of the column.</param>
-            /// <param name="type"><see cref="DataKind"/> of the items in the column. If <see langword="null"/> defaults to a float.</param>
+            /// <param name="kind"><see cref="InternalDataKind"/> of the items in the column. If <see langword="null"/> defaults to a float.</param>
             /// <param name="source">Source index range(s) of the column.</param>
             /// <param name="keyCount">For a key column, this defines the range of values.</param>
-            public Column(string name, DataKind? type, Range[] source, KeyCount keyCount = null)
+            private Column(string name, InternalDataKind? kind, Range[] source, KeyCount keyCount = null)
```
eerhardt: Who calls this constructor?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -50,37 +49,44 @@ public sealed class Column
...
             /// <param name="minIndex">The minimum inclusive index of the column.</param>
             /// <param name="maxIndex">The maximum-inclusive index of the column.</param>
-            public Column(string name, DataKind? type, int minIndex, int maxIndex)
-                : this(name, type, new[] { new Range(minIndex, maxIndex) })
+            public Column(string name, DataKind dataKind, int minIndex, int maxIndex)
+                : this(name, dataKind.ToDataKind(), new[] { new Range(minIndex, maxIndex) })
+            {
+            }
+
+            public Column(string name, DataKind dataKind, Range[] source, KeyCount keyCount = null)
```
eerhardt: public API should have XML doc.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
eerhardt: Why `ItemType`? Typically we use that name for the type of items inside of something - like a Vector. Why not just `Type` as the public name?

src/Microsoft.ML.Data/Transforms/TypeConverting.cs
```diff
@@ -539,7 +540,7 @@ public sealed class ColumnInfo
             /// <summary>
             /// The expected kind of the converted column.
             /// </summary>
-            public readonly DataKind OutputKind;
+            internal readonly DataKind OutputKind;
```
eerhardt: Why is this internal?

wschin: It shouldn't be. Thanks.

---
In reply to: [259521046](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259521046) [](ancestors = 259521046)

test/Microsoft.ML.Tests/Transformers/ConvertTests.cs
```diff
@@ -243,7 +243,7 @@ public void TypeConvertKeyBackCompatTest()
         {
             // Model generated using the following command before the change removing Min and Count from KeyType.
             // ML.Transforms.Conversion.ConvertType(new[] { new TypeConvertingEstimator.ColumnInfo("key", "convertedKey",
-            //      DataKind.U8, new KeyCount(4)) }).Fit(dataView);
+            //      ScalarType.Ulong, new KeyCount(4)) }).Fit(dataView);
```
eerhardt: `ScalarType` ?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
wschin: We can't. Internal name, nameof(Type), is used in naming command line arguments.

---
In reply to: [259520596](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259520596) [](ancestors = 259520596)

wschin: Will rename it to ToDataKind.

---
In reply to: [259519970](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259519970) [](ancestors = 259519970)

eerhardt: I'm not sure I understand this reasoning. Can you explain further?

In general, we shouldn't be sacrificing our public API for "command line" reasonings.

test/Microsoft.ML.Tests/Transformers/ConvertTests.cs
wschin: Should be `DataKind.UInt64`. Thanks.

---
In reply to: [259522599](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259522599) [](ancestors = 259522599)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
wschin: Only the three public constructors above.

---
In reply to: [259520279](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259520279) [](ancestors = 259520279)

src/Microsoft.ML.Core/Data/DataKind.cs
wschin: Then we need to update this doc string. We have this new enum only because of the two classes mentioned.

---
In reply to: [259519002](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259519002) [](ancestors = 259519002)

eerhardt: > We have this new enum only because of the two classes mentioned.

Right now, sure. But nothing is stopping some new API from using this enum. Then this doc would be out of date. Saying where something is used now doesn't explain what this thing is.

How about simply:

```
Specifies a simple data type.
```

other enum summaries are pretty simple:

https://docs.microsoft.com/en-us/dotnet/api/system?view=netframework-4.7.2#enums

>  TypeCode | Specifies the type of an object.
>  DayOfWeek | Specifies the day of the week.

If you want to add where it is used, that would be better in a `<remarks>` section.

src/Microsoft.ML.Core/Data/DataKind.cs
```diff
@@ -8,43 +8,39 @@
...
-    /// <item><description><see cref="String"/>: string, type of <see cref="System.String"/>.</description></item>
-    /// <item><description><see cref="Boolean"/>: boolean variable type, type of <see cref="System.Boolean"/>.</description></item>
-    /// <item><description><see cref="TimeSpan"/>: type of <see cref="System.TimeSpan"/>.</description></item>
-    /// <item><description><see cref="DateTime"/>: type of <see cref="System.DateTime"/>.</description></item>
-    /// <item><description><see cref="DateTimeOffset"/>: type of <see cref="System.DateTimeOffset"/>.</description></item>
-    /// </list>
-    /// </remarks>
     public enum DataKind : byte
     {
+        /// <summary><see cref="SByte"/>: 1-byte integer, type of <see cref="System.SByte"/>.</summary>
```
eerhardt: You don't need `<see cref` in the summary for the thing you are providing a summary for:

```
/// <summary>1-byte integer, type of <see cref="System.SByte"/>.</summary>
```

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
wschin: We use nameof(Type) as an argument name in entry point (classical APIs) and command line. If we rename `Type` to, for example, `InternalType`, old applications may be broken.

---
In reply to: [259523340](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259523340) [](ancestors = 259523340)

src/Microsoft.ML.Core/Data/DataKind.cs
wschin: Ok. I will move it to `//`.

---
In reply to: [259536248](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259536248) [](ancestors = 259536248)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
eerhardt: I'm still not convinced.... We are breaking all sorts of old applications with the API changes we've been making. We should define our public API as we want, and the other things will need to change.

@TomFinley - thoughts here?

wschin: TextLoader is a widely used class. Braking BC may cost a lot. Also, `Type` is not very accurate. When we load a float vector column, its type will still be `Float` instead of `Float[]`.

---
In reply to: [259541585](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259541585) [](ancestors = 259541585)

eerhardt: Also, we have this design guideline:

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/constructor

> ✓ DO use the same name for constructor parameters and a property if the constructor parameters are used to simply set the property.

> The only difference between such parameters and the properties should be casing.

So this name should be in sync with the constructor parameter name. Maybe that would solve all your problems?

`public DataKind DataKind` to match the constructor parameter name: `public Column(string name, DataKind dataKind, ...`

eerhardt: > TextLoader is a widely used class.

Agreed. That is why it is critical we get the API right. We can never change it after this.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -91,10 +105,18 @@ public Column(string name, DataKind? type, Range[] source, KeyCount keyCount = n
...
 
             /// <summary>
-            /// <see cref="DataKind"/> of the items in the column. If <see langword="null"/> defaults to a float.
+            /// <see cref="InternalDataKind"/> of the items in the column. If <see langword="null"/> defaults to a float.
+            /// Although <see cref="InternalDataKind"/> is internal, <see cref="Type"/>'s information can be publically accessed by <see cref="ItemType"/>.
             /// </summary>
             [Argument(ArgumentType.AtMostOnce, HelpText = "Type of the items in the column")]
-            public DataKind? Type;
+            [BestFriend]
+            internal InternalDataKind? Type;
```
Ivanidzo4ka: >InternalDataKind? Type; [](start = 21, length = 23)

Would be great if you make it non-nullable field.
It doesn't have to.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
wschin: Ok. Let's use `DataKind`.

---
In reply to: [259542584](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259542584) [](ancestors = 259542584)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
wschin: Can we break entry-point BC? I can remove nullable if your answer is yes.

---
In reply to: [259543165](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259543165) [](ancestors = 259543165)

Ivanidzo4ka: Please do. Just make it:
`internal InternalDataKind Type = InternalDataKind.R4`


---
In reply to: [259549751](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259549751) [](ancestors = 259549751,259543165)

Ivanidzo4ka: Why `InternalDataKind Type = default` default?

---
In reply to: [259551578](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259551578) [](ancestors = 259551578,259549751,259543165)

wschin: The reason is described (indirectly) in Iteration 9. TextLoader doesn't store default settings so that some baseline files need to be updated.

---
In reply to: [259555563](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259555563) [](ancestors = 259555563,259551578,259549751,259543165)

src/Microsoft.ML.Core/Data/DataKind.cs
```diff
@@ -55,50 +95,67 @@ public enum DataKind : byte
     [BestFriend]
     internal static class DataKindExtensions
```
eerhardt: (nit) - maybe this could be `InternalDataKindExtensions`?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -91,10 +105,18 @@ public Column(string name, DataKind? type, Range[] source, KeyCount keyCount = n
...
             [Argument(ArgumentType.AtMostOnce, HelpText = "Type of the items in the column")]
-            public DataKind? Type;
+            [BestFriend]
+            internal InternalDataKind Type = InternalDataKind.R4;
+
+            /// <summary>
+            /// <see cref="Data.DataKind"/> of the items in the column.
+            /// </summary>
+            /// It's a public interface to access the information in an internal DataKind.
+            public DataKind DataKind => Type.ToDataKind();
```
eerhardt: I think this should be settable. All the other fields are settable on this class:

`Name`, `KeyCount`, `Source`, but this one isn't. So if I use the default constructor - there is no way to set this property.

src/Microsoft.ML.Core/Data/DataKind.cs
wschin: Sure.

---
In reply to: [259953370](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259953370) [](ancestors = 259953370)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
wschin: Ok. Will make it
```csharp
            public DataKind DataKind
            {
                get { return Type.ToDataKind(); }
                set { Type = value.ToInternalDataKind(); }
            }
```

---
In reply to: [259954397](https://github.com/dotnet/machinelearning/pull/2661#discussion_r259954397) [](ancestors = 259954397)

Issue #2660, opened 2019-02-20
Rename GroupId column to RowGroup 


Ivanidzo4ka:
https://github.com/dotnet/machinelearning/blob/4a71e501f09846974e550f817f3f8f1243054c1b/src/Microsoft.ML.LightGBM/LightGbmCatalog.cs#L127

Conversation can be found here:
https://github.com/dotnet/machinelearning/issues/2536

abgoswam:
For the sake of consistency (as mentioned in issue #2177 )  should we call it 

- rowGroup**ColumnName**

abgoswam:
Fixed by PR #2665

Issue #2613, opened 2019-02-19
Scrubbing learners (Meta issue)


Ivanidzo4ka:
We need to clean our API for learners.
This includes following things:

1. No protected fields/members/method in public classes. Only private protected.

2. Learner class is sealed.

3. The trainer name types should follow the names used in the contexts (see https://github.com/dotnet/machinelearning/issues/2172)

4. ModelPameters for this Learner is also clean. (sealed, public documentation, no public constructor)

5. Options cleaning:
- They should be named Options, and all their base classes (except `LearnerInputBase*`)
- Option should have meaning and proper way it initialize it self. We shouldn't accept int if in reality we use enum (https://github.com/dotnet/machinelearning/issues/2521) or accept array of ints as a string separated by comma `public string CustomGains = "0,3,7,15,31";`
- No short names.
- Standard names like:
`/// <param   name="labelColumnName">The name of the label   column.</param>`
`/// <param   name="featureColumnName">The name of the feature   column.</param>`
`NumberOfIterations`, `NumberOfThreads`, `LearningRate`,`L2Regularization`, `L1Regularization`
no (`MaxIterations`,`NumIterations`, `NumThreads`, `L2Weight`, `L1Weight`)

6. If we communicate with user (exceptions, channel messages), don't use DataKind values. No R4, U4, and so on, it should be float, uint, etc.

ignore for now:

1. Everything public has proper documentation.

7. We have samples of how use base call to trainer (no options) and call with options.

8. We have baseline tests for learner.

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions#using-abbreviations-and-acronyms

`

rogancarr:
9. There should be consistency within `Tasks` (e.g. `BinaryClassification`, `Regression`) on how labels are specified (e.g. `R4`, `bool`, `KeyType`).

For instance, many learners accept R4 values and helpfully convert them to `bool` or `KeyType`, while some learners only accept `KeyType`. (see #2628 for a multiclass example). This inconsistency is confusing from a user perspective because you can't hot-swap Trainers in a pipeline.

wschin:
We also have `rowGroupColumnName` to replace `groupIdColumn`.

rogancarr:
The classnames themselves come in many different styles and they are surfaced through options:

For example, `SDCA` and `FastTree` have very different class names for binary classification:
```cs
[...]

Issue #2177, opened 2019-01-17
The parameter descriptions should distinguish between columns and column names. 


sfilipi:
As @wschin pointed out here: https://github.com/dotnet/machinelearning/pull/2170#issuecomment-455346083 the parmaters, and in general our XML documentation should distinguish between columns and column names. 

This issue will be considered closed when all the params of this language:

```
/// <param name="labelColumn">The label column.</param>
/// <param name="featureColumn">The featureColumn column.</param>
/// <param name="weights">The optional weights column.</param>
```
Get changed to: 
```
/// <param name="labelColumnName">The name of the label column.</param>
/// <param name="featureColumnName">The name of the feature column.</param>
/// <param name="weightsColumnName">The name of the optional weights column.</param>
```

glebuk:
Should the weights be renamed to weightsColumn, or perhaps exampleWeightsColumn?

sfilipi:
> Should the weights be renamed to weightsColumn, or perhaps exampleWeightsColum

HerraHak:
How can we contribute on this issue?

abgoswam:
@sfilipi . 

For some learners, the feature column name is a `string`

https://github.com/dotnet/machinelearning/blob/4a71e501f09846974e550f817f3f8f1243054c1b/src/Microsoft.ML.HalLearners/HalLearnersCatalog.cs#L31-L34

But for some others its a `string[]`

https://github.com/dotnet/machinelearning/blob/4a71e501f09846974e550f817f3f8f1243054c1b/src/Microsoft.ML.StandardLearners/FactorizationMachine/FactorizationMachineCatalog.cs#L29-L32 

For consistency of API  

(1)  we will call both of them `featureColumnName`.  is that fine ?

(2) regarding order of arguments in API  should we adopt the following order between  

- featureColumnName, labelColumnName, weightColumnName
- featureColumnName, labelColumnName, groupIdColumnName, weightColumnName  (for algos that need group id)

@Ivanidzo4ka @glebuk

abgoswam:
Assigning to myself  since this is superset of issues #2257, #2660 that i am fixing

abgoswam:
Fixed by PR #2665

wschin:
My feeling is that exampleWeightColumnName is a bit long, so #2873 is opened for further discussion.

PR #2665, opened 2019-02-20
Fixing parameters in ML.NET Public API

### LINKED ISSUE #2257: Rename `weights` parameter in ML.NET Public API
Couple of conventions to follow for `weights` parameter in ML.NET Public API

- Name of the `weight` argument in trainer API should be consistent with the other column names i.e it should end on *Column.
- Proposal to disambiguate this by calling it exampleWeightColumn.

@glebuk

abgoswam:
Fixes #2257, #2660, #2177 

Also related to #2680 #2613 

Using ApiReview tool, went through the public surface area making the changes (as described in the issue)


src/Microsoft.ML.KMeansClustering/KMeansCatalog.cs
```diff
@@ -29,7 +29,7 @@ public static class KMeansClusteringExtensions
         /// </example>
         public static KMeansPlusPlusTrainer KMeans(this ClusteringCatalog.ClusteringTrainers catalog,
            string featureColumn = DefaultColumnNames.Features,
```
TomFinley: >featureColumn [](start = 18, length = 13)

So we're calling this `exampleWeightColumnName`, but calling the above `featureColumn`. I'm not terribly insistent one way or the other, but we should at least be consistent I think.

In the original issue you proposed calling this `* exampleWeightColumn`, which seems to me to be consistent with the existing behavior *here*. Yet elsewhere I see that we are calling things `labelColumn` and `labelColumnName` here and there.

Is this on someone's radar? If we are not going to fix it (we should!) a second best alternative would be to be at least locally consistent. Not sure what @rogancarr and @eerhardt and @sfilipi think?

abgoswam: Issue #2177  proposes we use suffix *ColumnName consistently  

In this PR I am making that change for  weight, groupid  parameters.  In a follow up PR I will do it for label and features.  

So once  *both* PRs go in, we will mark #2177 as complete, and the following convention will be used
- labelColumnName
- featureColumnName
- exampleWeightColumnName
- rowGroupColumnName

This is certainly on my radar, as i drive towards consistency in the APIs . I have also updated the description of the issue so it has reference to issue #2177 .  Hope that sounds good ?

TomFinley: Very good, thank you @abgoswam ! Sorry missed that discussion.

src/Microsoft.ML.HalLearners/HalLearnersCatalog.cs
```diff
@@ -31,15 +31,15 @@ public static class HalLearnersCatalog
...
+            string exampleWeightColumnName = null)
         {
             Contracts.CheckValue(catalog, nameof(catalog));
             var env = CatalogUtils.GetEnvironment(catalog);
             var options = new OlsLinearRegressionTrainer.Options
             {
                 LabelColumn = labelColumnName,
                 FeatureColumn = featureColumnName,
-                WeightColumn = weightColumn != null ? Optional<string>.Explicit(weightColumn) : Optional<string>.Implicit(DefaultColumnNames.Weight)
+                WeightColumn = exampleWeightColumnName != null ? Optional<string>.Explicit(exampleWeightColumnName) : Optional<string>.Implicit(DefaultColumnNames.Weight)
```
TomFinley: Uh oh. Am I wrong or is there *literally no way* to have non-weighted training if you happen to have a column named `Weight`????

It looks like there's an existing issue #1065 that is capturing this.

TomFinley: Although I see it's incidentally assigned to you @abgoswam . 😄 Maybe after this PR we can think about how to address this.

abgoswam: Sounds good. We should fix this behavior as part of  #1065   (in a separate PR).

src/Microsoft.ML.PCA/PCACatalog.cs
```diff
@@ -15,21 +15,21 @@ public static class PcaCatalog
...
+        /// <param name="exampleWeightColumnName">The name of the example weight column (optional).</param>
         /// <param name="rank">The number of principal components.</param>
         /// <param name="overSampling">Oversampling parameter for randomized PrincipalComponentAnalysis training.</param>
         /// <param name="center">If enabled, data is centered to be zero mean.</param>
         /// <param name="seed">The seed for random number generation.</param>
         public static PrincipalComponentAnalysisEstimator ProjectToPrincipalComponents(this TransformsCatalog.ProjectionTransforms catalog,
             string outputColumnName,
             string inputColumnName = null,
-            string weightColumn = PrincipalComponentAnalysisEstimator.Defaults.WeightColumn,
+            string exampleWeightColumnName = PrincipalComponentAnalysisEstimator.Defaults.WeightColumn,
```
eerhardt: Defaulting this to `null` will be covered in #1065, right?

abgoswam: Yes, perhaps #1065 would be the right place to  set this to `null` as default value, since that issue related to changing the _behavior_ of the weights parameters.

The issues addressed by this PR are all related  to _consistency_ in API naming convention.

src/Microsoft.ML.KMeansClustering/KMeansCatalog.cs
abgoswam: @tomfinley. I have consolidated all the changes in this PR.  So  this PR  addresses   {features, labels, groupid, weight} column names in a consistent manner.

---
In reply to: [258779314](https://github.com/dotnet/machinelearning/pull/2665#discussion_r258779314) [](ancestors = 258779314)

abgoswam: @tomfinley. I have consolidated all the changes in this PR.  So  this PR  addresses   {features, labels, groupid, weight} column names in a consistent manner.

---
In reply to: [258776812](https://github.com/dotnet/machinelearning/pull/2665#discussion_r258776812) [](ancestors = 258776812)

src/Microsoft.ML.PCA/PCACatalog.cs
abgoswam: #1065 did not address this. hence taking care of this in this PR

---
In reply to: [259028009](https://github.com/dotnet/machinelearning/pull/2665#discussion_r259028009) [](ancestors = 259028009)

abgoswam: fixed in this PR itself.

---
In reply to: [259039000](https://github.com/dotnet/machinelearning/pull/2665#discussion_r259039000) [](ancestors = 259039000)

src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
```diff
@@ -97,27 +97,27 @@ public static class TreeExtensions
         /// Ranks a series of inputs based on their relevance, training a decision tree ranking model through the <see cref="FastTreeRankingTrainer"/>.
         /// </summary>
         /// <param name="catalog">The <see cref="RankingCatalog"/>.</param>
-        /// <param name="labelColumn">The labelColumn column.</param>
-        /// <param name="featureColumn">The featureColumn column.</param>
-        /// <param name="groupId">The groupId column.</param>
-        /// <param name="weights">The optional weights column.</param>
+        /// <param name="labelColumnName">The name of the label column.</param>
+        /// <param name="featureColumnName">The name of the feature column.</param>
+        /// <param name="rowGroupColumnName">The name of the group column.</param>
```
eerhardt: This isn't necessary for this PR, but I wonder if we need some documentation around the `row group` column. I'm obviously not an expert, but I have no idea what this is.

test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamplesDynamicApi.cs
```diff
@@ -513,7 +513,7 @@ private static void RunEndToEnd(MLContext mlContext, IDataView trainData, string
             // Construct the learning pipeline. Note that we are now providing a contract name for the custom mapping:
             // otherwise we will not be able to save the model.
             var estimator = mlContext.Transforms.CustomMapping<InputRow, OutputRow>(CustomMappings.IncomeMapping, nameof(CustomMappings.IncomeMapping))
-                .Append(mlContext.BinaryClassification.Trainers.FastTree(labelColumn: "Label"));
+                .Append(mlContext.BinaryClassification.Trainers.FastTree(labelColumnName: "Label"));
```
eerhardt: This usually means the cookbook .md file needs to be updated as well.

abgoswam: Thanks for making me take a second look at this.  Updated the cookbook .md file in a couple of places.

p.s. my initial search for "labelColumn" in the cookbook .md file did not return relevant results.  Apparently it used to be called just "label" at some point :)

---
In reply to: [259626158](https://github.com/dotnet/machinelearning/pull/2665#discussion_r259626158) [](ancestors = 259626158)

src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
abgoswam: fyi...#2536 gives some context on this.  we can add more documentation around this in  follow up documentation focused  PRs.

Issue #1098, opened 2018-09-28
MLContext to create them all


Zruty0:
During one of the in-person API reviews we agreed that it would be a good idea to have a single object `MLContext` that would serve as a 'factory of everything' (similar to the HTTP context / DB context in the .NET world).

- `MLContext` will explicitly implement `IHostEnvironment`, so you can create all the existing estimators by giving the context as the first argument.
- `MLContext` will have properties `BinaryClassification`, `Regression`, `Clustering` etc. for canonical ML tasks (the ones that are currently classes in themselves), complete with `Evaluate` and all corresponding trainers.
- It will have *extension methods* for non-canonical tasks like recommendation or anomaly detection etc.
- It will have properties `Transformation`, `Filtering`, `Loading` to instantiate all known transform estimators, filters and data readers (again via extension methods).
- It will have a pair of methods `SaveModel` and `LoadModel` that handle model serialization.

/cc @KrzysztofCwalina @TomFinley @eerhardt @markusweimer @asthana86

Ivanidzo4ka:
>MLContext will explicitly implement IHostEnvironment, so you can create all the existing estimators by giving the context as the first argument.
We don't like pass IHostEnvironment objects because they bloated, so let's create even more bloated object.

Can we have something like this instead:
```
     public sealed class MLContext
    {
        public readonly IHostEnvironment Env;
        public MLContext(IHostEnvironment env = null)
        {
            if (env == null)
                env = new ConsoleEnvironment();
            Env = env;
        }
    }
```
and just modify constructors in Estimators to accept MLContext?

TomFinley:
> We don't like pass IHostEnvironment objects because they bloated, so let's create even more bloated object.

It's not quite so bad as all that. One of the details here you will see is that @Zruty0 has said, originally suggested by @eerhardt, is that this hypothetical object will *explicitly* implement the interface. Further, the numerous extensions to `IHostEnvironment` that are useful only to component authors (most significantly, those in `Contracts.cs`) would be in a namespace generally used only by component authors, or otherwise rendered invisible to users. On the other hand, the user-facing properties themselves being on the *class* and not part of the *interface*, would not be visible to component authors.

So despite being a single object kinda, it has a "dual nature" that reflects the dual usage of ML.NET, on the one hand a tool for exploiting ML by a user, as well as a tool into which one can plug ones own components. All without having to have parallel "user context" vs. "component context" object, which seems like an elegant solution.

[...]

PR #2672, opened 2019-02-20
Internalization of TensorFlowUtils.cs and refactored TensorFlowCatalog.

### LINKED ISSUE #2552: Should TensorFlowModelInfo be public?
Related to #2280.
The TensorFlowModelInfo class is still public, and so are two methods for creating a TensorFlowEstimator from a TensorFlowModelInfo. I was wondering whether we should have these, since the only way to get a TensorFlowModelInfo is by calling TensorFlowUtils.LoadTensorFlowModel(), and that uses a file name - and users that have a file name, can instantiate a TensorFlowEstimator from that directly. 
There is one scenario that I can think of where a user would benefit from using this API: The TensorFlowModelInfo class has a method GetInputSchema(), that can be used to get the name of the input layer(s), and then the TensorFlowEstimator can be instantiated from the TensorFlowModelInfo instead of instantiating it from the file name and having to load it from disk again. Is this reason enough to keep these APIs?
If it is, then TensorFlowUtils.LoadTensorFlowModel should be moved to TensorFlowCatalog, and perhaps also the GetInputSchema() method of TensorFlowModelInfo.
@TomFinley , @zeahmed , @abgoswam , @ganik  - thoughts?

Another unrelated issue: Microsoft.ML.TensorFlow has two AssemblyInfo.cs files, one in src\Microsoft.ML.TensorFlow\ and one in  src\Microsoft.ML.TensorFlow\Properties. These should be unified.

### LINKED ISSUE #2572: TensorFlowUtils doesn't really fit with the 1.0 API
One of the scenarios we want to support in ML.NET 1.0 is the ability to examine the schema of a TensorFlow model (see example below [1]). To do this, we currently use helper methods in the `TensorFlowUtils` class [2].

Right now, this class and these methods are `public` and can be used as part of an ML.NET program, but that doesn't seem like the way we want to go in v1.0. What do you all think? Does it make sense to add these methods to mlContext and make them internal/BestFriend? If we add them to a catalog, where should they go?

[1] Example of why you would want to examine the schema before loading. Here, we want to get the layer names and their sizes before specifying them as outputs in our `IDataView`:
```cs
// Get the output size of a layer
var schema = TensorFlowUtils.GetModelSchema(mlContext, model_location);
var outputColumn = schema.GetColumnOrNull("Output");
if (outputColumn.HasValue)
    outputLength = ((VectorType)outputColumn.Value.Type).Dimensions[0];
```

[2] Helper Methods
```cs
public static Schema GetModelSchema(IHostEnvironment env, string modelPath);
public static IEnumerable<(string, string, ColumnType, string[])> GetModelNodes(IHostEnvironment env, string modelPath);
public static TensorFlowModelInfo LoadTensorFlowModel(IHostEnvironment env, string modelPath);
```
where `LoadTensorFlowModel` is already in the Transforms catalog.

Would `GetModelSchema` and `GetModelNodes` go in the `DataOperations` catalog?

And to pile on, `GetModelNodes` (which may not be necessary to expose) returns a list of returns, instead of an object, which we definitely can't have in V1.

Related to #2498

zeahmed:
This PR fixes #2552 and also fixes #2572.

The discussion in these issues are taken as initial feedback to create this PR. The PR groups all the TensorFlow related operations into TensorFlowCatalog. This catalog now appears as a separate catalog in TransformsCatalog e.g. `ScoreTensorFlowModel` that appeared under `Transforms` catalog as

``` csharp
mlContext.Transforms.ScoreTensorFlowModel
```
now appears under `TensorFlow` as 
``` csharp
mlContext.Transforms.TensorFlow.ScoreTensorFlowModel
```
How breaking is this change at this stage?

Also, the PR also include following methods in TensorFlow catalog. 
``` csharp
public static DataViewSchema GetModelSchema(this TransformsCatalog.TensorFlowTransforms catalog, string modelLocation)
    => TensorFlowUtils.GetModelSchema(CatalogUtils.GetEnvironment(catalog), modelLocation);

public static IEnumerable<(string, string, DataViewType, string[])> GetModelNodes(this TransformsCatalog.TensorFlowTransforms catalog, string modelLocation)
    => TensorFlowUtils.GetModelNodes(CatalogUtils.GetEnvironment(catalog), modelLocation);

public static TensorFlowModelInfo LoadTensorFlowModel(this TransformsCatalog.TensorFlowTransforms catalog, string modelLocation)
    => TensorFlowUtils.LoadTensorFlowModel(CatalogUtils.GetEnvironment(catalog), modelLocation);
```
There is also a suggestion to add these methods into `DataCatalog`. Feedback is requested in this regard.

CC: @TomFinley, @yaeldekel, @rogancarr, @Ivanidzo4ka


src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs
```diff
@@ -100,9 +102,42 @@ public static class TensorflowCatalog
...
+
+        /// <summary>
+        /// This is a convenience method for iterating over the nodes of a TensorFlow model graph. It
+        /// iterates over the columns of the <see cref="DataViewSchema"/> returned by <see cref="GetModelSchema(TransformsCatalog.TensorFlowTransforms, string)"/>,
+        /// and for each one it returns a tuple containing the name, operation type, column type and an array of input node names.
+        /// This method is convenient for filtering nodes based on certain criteria, for example, by the operation type.
+        /// </summary>
+        /// <param name="catalog">The transform's catalog.</param>
+        /// <param name="modelLocation">Location of the TensorFlow model.</param>
+        public static IEnumerable<(string, string, DataViewType, string[])> GetModelNodes(this TransformsCatalog.TensorFlowTransforms catalog, string modelLocation)
```
yaeldekel: >GetModelNodes [](start = 76, length = 13)

This doesn't need to be a part of the public API.

yaeldekel:
>         public DataViewSchema GetModelSchema()

I would make this internal (and `GetInputSchema` as well) and just expose them through the TensorflowCatalog.

---
Refers to: src/Microsoft.ML.TensorFlow/TensorFlowModelInfo.cs:46 in e5eef19. [](commit_id = e5eef190629e6e2dbe9e966a8fe719b76c73f2be, deletion_comment = False)

yaeldekel:
>         public DataViewSchema GetModelSchema()

If you do this, then this class doesn't need to have an `IHostEnvironment` member, because you can use `CatalogUtils.GetEnvironment(catalog)`.

---
In reply to: [465804802](https://github.com/dotnet/machinelearning/pull/2672#issuecomment-465804802) [](ancestors = 465804802)

---
Refers to: src/Microsoft.ML.TensorFlow/TensorFlowModelInfo.cs:46 in e5eef19. [](commit_id = e5eef190629e6e2dbe9e966a8fe719b76c73f2be, deletion_comment = False)

src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs
zeahmed: It is being use here.
[https://github.com/dotnet/machinelearning/blob/fb6ce5442f1af7b30dc48f6a438b2a7b63779390/src/Microsoft.ML.DnnAnalyzer/Microsoft.ML.DnnAnalyzer/DnnAnalyzer.cs#L20](https://github.com/dotnet/machinelearning/blob/fb6ce5442f1af7b30dc48f6a438b2a7b63779390/src/Microsoft.ML.DnnAnalyzer/Microsoft.ML.DnnAnalyzer/DnnAnalyzer.cs#L20)

If I don't expose it like this then I will have to make the internal one public.

---
In reply to: [258728605](https://github.com/dotnet/machinelearning/pull/2672#discussion_r258728605) [](ancestors = 258728605)

src/Microsoft.ML.Data/Transforms/TransformsCatalog.cs
```diff
@@ -37,6 +37,11 @@ public sealed class TransformsCatalog
         /// </summary>
         public FeatureSelectionTransforms FeatureSelection { get; }
 
+        /// <summary>
+        /// List of operations for using TensorFlow model.
+        /// </summary>
+        public TensorFlowTransforms TensorFlow { get; }
```
TomFinley: >TensorFlowTransforms [](start = 15, length = 20)

Please do not do this. Otherwise we will have an empty property unless someone imports the nuget, which is confusing and undesirable.

Please follow instead the pattern that we see in image processing. You'll note that we do not have an empty image processing nuget. Rather they are added to this catalog.

This is defensible since we can take someone directly importing a nuget as a strong signal that they want to actually use those transforms.

src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs
```diff
@@ -100,9 +102,41 @@ public static class TensorflowCatalog
...
+            => TensorFlowUtils.GetModelSchema(CatalogUtils.GetEnvironment(catalog), tensorFlowModel.Session.Graph);
+
+        /// <summary>
+        /// This method retrieves the information about the input graph nodes of a TensorFlow model as an <see cref="DataViewSchema"/>.
+        /// The nodes with OpType as "Placeholder" are classified as input nodes. This is the convenience method to inspect only input nodes.
+        /// For retrieving complete information please see <see cref="GetModelSchema(TransformsCatalog.TensorFlowTransforms, TensorFlowModelInfo)"/>
+        /// </summary>
+        /// <param name="catalog">The transform's catalog.</param>
+        /// <param name="tensorFlowModel">The pre-loaded TensorFlow model. Please see <see cref="LoadTensorFlowModel(TransformsCatalog.TensorFlowTransforms, string)"/> to know more about loading model into memory.</param>
+        public static DataViewSchema GetInputSchema(this TransformsCatalog.TensorFlowTransforms catalog, TensorFlowModelInfo tensorFlowModel)
```
TomFinley: >GetInputSchema [](start = 37, length = 14)

This API does not seem like any C# API I've seen. It seems more like a C API with a bunch of static functions that you call with an object, rather than anything like what I'd expect in an actual object oriented property. Why is not *everything* you've added as an extension method, except possibly the creation of what you call now this `TensorFlowModelInfo`, just a property *or* a method of that `TensorFlowModelInfo`?

yaeldekel: Tom, I was under the impression that we wanted to have as much of the functionality as possible accessible from MLContext, and expose as few classes as possible, that's why I suggested exposing these methods here. Do you think the `LoadTensorFlowModel` API should still be an extension method on `TransformsCatalog`, or should we expose the `TensorFlowUtils` class for that?

---
In reply to: [258769868](https://github.com/dotnet/machinelearning/pull/2672#discussion_r258769868) [](ancestors = 258769868)

src/Microsoft.ML.TensorFlow/TensorFlowModelInfo.cs
```diff
@@ -11,50 +11,21 @@ namespace Microsoft.ML.Transforms
...
-    ///    <item>
-    ///      <description>Get schema related to model input(s) by calling <see cref="GetInputSchema()"/>.</description>
-    ///    </item>
-    /// </list>
     /// </summary>
-    public class TensorFlowModelInfo
+    public sealed class TensorFlowModelInfo
     {
         internal TFSession Session { get; }
         public string ModelPath { get; }
```
yaeldekel: >public [](start = 8, length = 6)

I think this can also be internal.

src/Microsoft.ML.TensorFlow/TensorFlowModelInfo.cs
```diff
@@ -11,50 +11,21 @@ namespace Microsoft.ML.Transforms
...
-    ///    <item>
-    ///      <description>Get complete schema by calling <see cref="GetModelSchema()"/>.</description>
-    ///    </item>
-    ///    <item>
-    ///      <description>Get schema related to model input(s) by calling <see cref="GetInputSchema()"/>.</description>
-    ///    </item>
-    /// </list>
     /// </summary>
-    public class TensorFlowModelInfo
+    public sealed class TensorFlowModelInfo
```
yaeldekel: >TensorFlowModelInfo [](start = 24, length = 19)

Would it be possible to rename this to `TensorFlowModel`?

docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/TextClassification.cs
```diff
@@ -45,7 +45,7 @@ public static void Example()
             // Load the TensorFlow model once.
             //      - Use it for quering the schema for input and output in the model
             //      - Use it for prediction in the pipeline.
-            var modelInfo = TensorFlowUtils.LoadTensorFlowModel(mlContext, modelLocation);
+            var modelInfo = mlContext.Transforms.LoadTensorFlowModel(modelLocation);
```
TomFinley: >mlContext.Transforms.LoadTensorFlowModel [](start = 28, length = 40)

This doesn't seem like a transform. This seems like you're loading a model, so should it be under "model?"

src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs
TomFinley: I don't think that's how we've been doing things at all. We use MLContext for things like component creation, but once you, to give the most conspicuous example, create an estimator, the creation of a transformer, the getting of information *out* of the transformer, the usage of the transformer to transform a data view, *that* does not involve the MLContext.

You create a model. The model is the object, you use objects by calling methods and properties on those objects, just like everything else. Recall the title of #1098 that is I'd say the central MLContext... one MLContext to *create* them all, not to *use* them all.

I gave examples above about how we don't structure our APIs as they are structured here, and I think my reasoning is pretty solid.

docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/TextClassification.cs
zeahmed: Sounds reasonable. I see some of the ONNX model related methods are available via `model` catalog. I will move it. Thanks!

---
In reply to: [259530418](https://github.com/dotnet/machinelearning/pull/2672#discussion_r259530418) [](ancestors = 259530418)

src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs
zeahmed: Thanks @TomFinley, I hope I addressed your concern here and other places. I am changing it resolved for now.

---
In reply to: [259531678](https://github.com/dotnet/machinelearning/pull/2672#discussion_r259531678) [](ancestors = 259531678)

docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/TextClassification.cs
```diff
@@ -45,7 +45,7 @@ public static void Example()
...
             var schema = modelInfo.GetModelSchema();
             var featuresType = (VectorType)schema["Features"].Type;
```
yaeldekel: >Features [](start = 51, length = 8)

Can we add a sample that uses `modelInfo.GetInputSchema()` to find out what the name of the input node is?

zeahmed: I see its being used at couple of places in the tests e.g.
https://github.com/dotnet/machinelearning/blob/eb959c32a6215b1445df56a49761d065dfcdb7b3/test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs#L894

https://github.com/dotnet/machinelearning/blob/eb959c32a6215b1445df56a49761d065dfcdb7b3/test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs#L849

src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs
TomFinley: Marking unresolved... I do not see that anything has changed at all. `TensorFlowModel` has no meaningful methods on it, everything is still being exposed via these extension methods on top of properties.

---
In reply to: [259545815](https://github.com/dotnet/machinelearning/pull/2672#discussion_r259545815) [](ancestors = 259545815,259531678)

src/Microsoft.ML.TensorFlow/TensorFlowModel.cs
```diff
@@ -20,20 +19,20 @@ namespace Microsoft.ML.Transforms
     ///    </item>
     /// </list>
     /// </summary>
-    public class TensorFlowModelInfo
+    public sealed class TensorFlowModel
```
TomFinley: I think possibly maybe there's some misunderstanding. Just to be explicit, I expect to see methods used to *query* the model to be *on the model*, that is, they will be just methods and properties of the model. Creatingtransformers, querying the schema, or *whatever*, will be here. The implication is that nearly everything in `TensorflowCatalog` will be moved out of there, and into here. (Except for loading the model, which of course must be on the model operations catalog.)

src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs
```diff
@@ -59,7 +60,7 @@ public static class TensorflowCatalog
         /// <param name="inputColumnName"> The name of the model input.</param>
         /// <param name="outputColumnName">The name of the requested model output.</param>
         public static TensorFlowEstimator ScoreTensorFlowModel(this TransformsCatalog catalog,
```
TomFinley: >ScoreTensorFlowModel [](start = 42, length = 20)

This should be an operation on the model.

src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs
```diff
@@ -79,7 +80,7 @@ public static class TensorflowCatalog
         /// </format>
         /// </example>
         public static TensorFlowEstimator ScoreTensorFlowModel(this TransformsCatalog catalog,
```
TomFinley: >ScoreTensorFlowModel [](start = 42, length = 20)

Likewise on the model..

src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs
```diff
@@ -102,7 +103,16 @@ public static class TensorflowCatalog
         /// <param name="tensorFlowModel">The pre-loaded TensorFlow model.</param>
         public static TensorFlowEstimator TensorFlow(this TransformsCatalog catalog,
```
TomFinley: >TensorFlow [](start = 42, length = 10)

So this I find pretty confusing. Do we create estimators via *this* method, or do we work through the model object?

TomFinley:
>         public static TensorFlowEstimator TensorFlow(this TransformsCatalog catalog,

Just more broadly, should these be verbs?

---
Refers to: src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs:94 in fc188cd. [](commit_id = fc188cd65cea68216016b6b0aca859481d3259a6, deletion_comment = False)

yaeldekel:
>         public static TensorFlowEstimator TensorFlow(this TransformsCatalog catalog,

Also, do we generally expose APIs creating estimators from Options? From what I've seen the signature is either all the arguments separately, or a params array of ColumnInfo objects.

---
In reply to: [467198148](https://github.com/dotnet/machinelearning/pull/2672#issuecomment-467198148) [](ancestors = 467198148)

---
Refers to: src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs:94 in fc188cd. [](commit_id = fc188cd65cea68216016b6b0aca859481d3259a6, deletion_comment = False)

src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs
zeahmed: Thanks Tom for feedback. I have refactored the code according to the your comments.

---
In reply to: [260027159](https://github.com/dotnet/machinelearning/pull/2672#discussion_r260027159) [](ancestors = 260027159,259545815,259531678)

zeahmed:
>         public static TensorFlowEstimator TensorFlow(this TransformsCatalog catalog,

I think we are exposing APIs with Options. e.g. [https://github.com/dotnet/machinelearning/blob/2a1f2bb6a62393115f6cb8128fbbe9850068180d/src/Microsoft.ML.LightGBM/LightGbmCatalog.cs#L112](https://github.com/dotnet/machinelearning/blob/2a1f2bb6a62393115f6cb8128fbbe9850068180d/src/Microsoft.ML.LightGBM/LightGbmCatalog.cs#L112). Do you think it should not be exposed?

---
In reply to: [467552241](https://github.com/dotnet/machinelearning/pull/2672#issuecomment-467552241) [](ancestors = 467552241,467198148)

---
Refers to: src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs:94 in fc188cd. [](commit_id = fc188cd65cea68216016b6b0aca859481d3259a6, deletion_comment = False)

yaeldekel:
>         public static TensorFlowEstimator TensorFlow(this TransformsCatalog catalog,

In `TransformsCatalog` we usually don't expose the `Options`.
But I think that we do need an extension method on `TransformsCatalog` to create a `TensorFlowEstimator`.

---
In reply to: [467668930](https://github.com/dotnet/machinelearning/pull/2672#issuecomment-467668930) [](ancestors = 467668930,467552241,467198148)

---
Refers to: src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs:94 in fc188cd. [](commit_id = fc188cd65cea68216016b6b0aca859481d3259a6, deletion_comment = False)

zeahmed:
>         public static TensorFlowEstimator TensorFlow(this TransformsCatalog catalog,

Tom's idea is not to have those methods in `TransformsCatalog`. Let me know Tom what do you suggest?

---
In reply to: [467939019](https://github.com/dotnet/machinelearning/pull/2672#issuecomment-467939019) [](ancestors = 467939019,467668930,467552241,467198148)

---
Refers to: src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs:94 in fc188cd. [](commit_id = fc188cd65cea68216016b6b0aca859481d3259a6, deletion_comment = False)

TomFinley:
> @TomFinley's idea is not to have those methods in `TransformsCatalog`. Let me know @TomFinley what do you suggest?
> 
> In reply to: [467939019](https://github.com/dotnet/machinelearning/pull/2672#issuecomment-467939019) [](ancestors = 467939019,467668930,467552241,467198148)
> 
> Refers to: src/Microsoft.ML.TensorFlow/TensorflowCatalog.cs:94 in [fc188cd](https://github.com/dotnet/machinelearning/commit/fc188cd65cea68216016b6b0aca859481d3259a6). [](commit_id = [fc188cd](https://github.com/dotnet/machinelearning/commit/fc188cd65cea68216016b6b0aca859481d3259a6), deletion_comment = False)


There are I think two distinct things being discussed. One is the presence of advanced options. @yaeldekel is correct, we generally try to avoid having these advanced options where we can, but we definitely have it as part of the public surface area multiple places where it is unavoidable. It is more common in so-called *trainer* estimators rather than plain old featurizing estimators. If you can avoid it @zeahmed, I'd do so. But I could easily imagine that something as involved as an *interface to TensorFlow* might justify a bit more complexity.

The second question I think is about whether creation of the estimators should be on the transform catalog. I'd say in this case no, since we create the model out of `MLContext`, and once we have it, there's no longer any need to involve `MLContext`, so, let's not.

Nonetheless over the past week it's become clear to me that there is deep discomfort about this, so I'll try to address that, by telling you how I'm thinking about this. I tend to think of the introduction of `MLContext` as a mechanism by which we avoid throwing these `IHostEnvironment` objects around, when we create components that require an `IHostEnvironment`.

Now then, it just so happens that often these "initial components" are estimators, but I don't attach any special significance to that. We have plenty of cases where this is not the case: operations on `Data` often take and return `IDataView` directly, on `Model` we are often loading `ITransformer`, and so on).

So we see, naturally, this progression oftentimes, of `MLContext` operating over all of these things, as is appropriate in teh context.

    IEstimator -> ITransformer -> IDataView

and sometimes this parallel structure for ingestion via loaders:

    IDataLoaderEstimator -> IDataLoader -> IDataView

As far as I am aware we have hitherto always had the `MLContext` create components *somewhere* along this continuum, never (I think) outside of it. Yet ***in this case*** there is something happening before this continuum. We have this "chain of creation."

    TensorFlowModel -> IEstimator -> ITransformer -> IDataView

I am saying, I think, that I don't view this chain as being in any way special to the more typical chain. Given that we're creating `IEstimator` out of this tensorflow model, there's no longer any reason to involve `MLContext` beyond that initial call to create the model -- that is, the creation of the estimator seems not to rely in any way on the context, so, it should not. The model object itself is perfectly adequate and equal to the task. Better, really.

yaeldekel:
>         public sealed class Options : TransformInputBase

This can be internal now.

---
Refers to: src/Microsoft.ML.TensorFlow/TensorflowTransform.cs:1002 in 7cd88ed. [](commit_id = 7cd88edf013bc2058310f45f3612b2e673df2229, deletion_comment = False)

PR #973, opened 2018-09-21
PredictionEngine uses IRowToRowMapper, ITransformer can create IRowToRowMapper

TomFinley:
In which we reduce the overhead of calls to `Predict` by avoiding the creation of cursors.

`PredictionEngine` used to create a "fake" data view under the hood, by wrapping a `BatchPredictionEngine`, but only feeding in one item. This was a fairly heavy operation involving spooling up of cursors, including heavy reflection based processing on each row, etc. Now we are able to avoid this by using an existing interface `IRowToRowMapper`.

The other major part of the change is, of course, changing transformers so they *can* produce `IRowToRowMapper`, which they previously did not. In most cases this is relatively straightforward since nearly all transformers of interest are using these `IRowToRowMapper` under the hood to support their computation.

* `IRowToRowMapper` enhancements, unification of some interfaces.
* `ITransformer` has `IRowToRowMapper` accessor.
* `PredictionEngine` now uses `IRowToRowMapper`

The effects of this are most dramatic using prediction engines *outside* of the ML.NET v0.1 pipeline API (e.g., using `IEstimator` based pipelines, pre-ML.NET v0.1 API, and so forth). The effect of this is that when predicting, now actual computation dominates the cost, as opposed to reflection based stuff. 😄 Due to architectural limitations of pipeline API models, these do not see benefit.

Will perhaps update with timings later.

Fix #986.

PR #2673, opened 2019-02-21
Explicit implementation for IsRowToRowMapper and GetRowToRowMapper

### LINKED ISSUE #2540: Explicit implementation for IsRowToRowMapper and GetRowToRowMapper
The interface `ITransformer` defines two methods:

```csharp
/// <summary>
/// Whether a call to <see cref="GetRowToRowMapper(Schema)"/> should succeed, on an
/// appropriate schema.
/// </summary>
bool IsRowToRowMapper { get; }

/// <summary>
/// Constructs a row-to-row mapper based on an input schema. If <see cref="IsRowToRowMapper"/>
/// is <c>false</c>, then an exception should be thrown. If the input schema is in any way
/// unsuitable for constructing the mapper, an exception should likewise be thrown.
/// </summary>
/// <param name="inputSchema">The input schema for which we should get the mapper.</param>
/// <returns>The row to row mapper.</returns>
IRowToRowMapper GetRowToRowMapper(Schema inputSchema);
```

A transformer maps `IDataViews` to `IDataViews`, however this process can in some cases be optimized by mapping rows to rows with a `IRowToRowMapper`. Although for most scenarios a user should not be using them directly, they might be useful in custom extensions which we would like to enable in the future. This is why the above two methods are present in the interface of `ITransformer` and are necessary.  

Since, however, these are not usually employed directly by users, we should implement them explicitly. 

/cc: @TomFinley

artidoro:
Fixes #2540.

As explained in the issue I make the implementations of two methods of the`ITransformer` interface `IsRowToRowMapper` and `GetRowToRowMapper` explicit.

In a previous PR #2431 I made the implementation of `ICanSaveModel.Save` explicit. `Save` is part of `ITransformer` as it derives from `ICanSaveModel`. I found a few instances that have been changed since my last PR in the TimeSeries assembly and that still had a regular implementation for `Save. I made their implementation explicit in this PR as it is very similar in spirit.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/TransformWrapper.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/TransformWrapper.cs
@@ -33,7 +33,6 @@ namespace Microsoft.ML.Data
             _host.CheckValue(xf, nameof(xf));
             _xf = xf;
             _allowSave = allowSave;
-            IsRowToRowMapper = IsChainRowToRowMapper(_xf);
         }
 
         public DataViewSchema GetOutputSchema(DataViewSchema inputSchema)
@@ -108,7 +107,6 @@ namespace Microsoft.ML.Data
             }
 
             _xf = data;
-            IsRowToRowMapper = IsChainRowToRowMapper(_xf);
         }
 
         public IDataView Transform(IDataView input) => ApplyTransformUtils.ApplyAllTransformsToData(_host, _xf, input);
@@ -123,9 +121,9 @@ namespace Microsoft.ML.Data
             return true;
         }
 
-        public bool IsRowToRowMapper { get; }
+        bool ITransformer.IsRowToRowMapper => IsChainRowToRowMapper(_xf);
 
-        public IRowToRowMapper GetRowToRowMapper(DataViewSchema inputSchema)
+        IRowToRowMapper ITransformer.GetRowToRowMapper(DataViewSchema inputSchema)
         {
             _host.CheckValue(inputSchema, nameof(inputSchema));
             var input = new EmptyDataView(_host, inputSchema);
--- a/src/Microsoft.ML.Data/DataLoadSave/TransformWrapper.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/TransformWrapper.cs
@@ -25,6 +25,7 @@ namespace Microsoft.ML.Data
         private readonly IHost _host;
         private readonly IDataView _xf;
         private readonly bool _allowSave;
+        private readonly bool _isRowToRowMapper;
 
         public TransformWrapper(IHostEnvironment env, IDataView xf, bool allowSave = false)
         {
@@ -33,6 +34,7 @@ namespace Microsoft.ML.Data
             _host.CheckValue(xf, nameof(xf));
             _xf = xf;
             _allowSave = allowSave;
+            _isRowToRowMapper = IsChainRowToRowMapper(_xf);
         }
 
         public DataViewSchema GetOutputSchema(DataViewSchema inputSchema)
@@ -107,6 +109,7 @@ namespace Microsoft.ML.Data
             }
 
             _xf = data;
+            _isRowToRowMapper = IsChainRowToRowMapper(_xf);
         }
 
         public IDataView Transform(IDataView input) => ApplyTransformUtils.ApplyAllTransformsToData(_host, _xf, input);
@@ -121,7 +124,7 @@ namespace Microsoft.ML.Data
             return true;
         }
 
-        bool ITransformer.IsRowToRowMapper => IsChainRowToRowMapper(_xf);
+        bool ITransformer.IsRowToRowMapper => _isRowToRowMapper;
 
         IRowToRowMapper ITransformer.GetRowToRowMapper(DataViewSchema inputSchema)
         {
--- a/src/Microsoft.ML.Data/DataLoadSave/TransformerChain.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/TransformerChain.cs
@@ -59,7 +59,7 @@ namespace Microsoft.ML.Data
 
         private const string TransformDirTemplate = "Transform_{0:000}";
 
-        public bool IsRowToRowMapper => _transformers.All(t => t.IsRowToRowMapper);
+        bool ITransformer.IsRowToRowMapper => _transformers.All(t => t.IsRowToRowMapper);
 
         ITransformer[] ITransformerChainAccessor.Transformers => _transformers;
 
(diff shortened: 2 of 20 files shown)


src/Microsoft.ML.Data/DataLoadSave/TransformWrapper.cs
```diff
@@ -123,9 +121,9 @@ private static bool IsChainRowToRowMapper(IDataView view)
             return true;
         }
 
-        public bool IsRowToRowMapper { get; }
+        bool ITransformer.IsRowToRowMapper => IsChainRowToRowMapper(_xf);
```
TomFinley: >IsChainRowToRowMapper [](start = 46, length = 21)

This operation is somewhat involved, so I would prefer if you could have a backing field to store this result as we had in the initial implementation.

src/Microsoft.ML.TimeSeries/SsaSpikeDetector.cs
```diff
@@ -162,7 +162,7 @@ internal SsaSpikeDetector(IHostEnvironment env, ModelLoadContext ctx)
             InternalTransform.Host.CheckDecode(InternalTransform.IsAdaptive == false);
         }
 
-        public override void Save(ModelSaveContext ctx)
+        private protected override void SaveModel(ModelSaveContext ctx)
```
TomFinley: >private protected override void SaveModel [](start = 8, length = 41)

Ahh, thank you very much for finding more of these!!

TomFinley: Actually wait, these are already internal, so probably not necessary... still, eh... I don't know.

---
In reply to: [258775986](https://github.com/dotnet/machinelearning/pull/2673#discussion_r258775986) [](ancestors = 258775986)

TomFinley:
Thank you @artidoro I finished a pass, the only real high level things I saw were not storing the result of that somewhat expensive call in the wrapper. More minor was the somewhat needless internalization of members of an already internal class, but I consider that less urgent (especially if you feel strongly that being consistent in that fashion is desirable, which I could understand).

src/Microsoft.ML.TimeSeries/SequentialAnomalyDetectionTransformBase.cs
```diff
@@ -348,7 +348,7 @@ public DataViewSchema.DetachedColumn[] GetOutputColumns()
                         return col => false;
                 }
 
-                public void Save(ModelSaveContext ctx) => _parent.SaveModel(ctx);
+                void ICanSaveModel.Save(ModelSaveContext ctx) => _parent.SaveModel(ctx);
```
artidoro: This one was internal already, but in every other Mapper elsewhere we have made this one explicit despite it being internal for consistency.

src/Microsoft.ML.TimeSeries/SsaSpikeDetector.cs
artidoro: I think the containing classes are public. The classes in which I made the implementation of `ICanSaveModel,Save` explicit derive from  `SsaAnomalyDetectionBaseWrapper` or `IidAnomalyDetectionBaseWrapper` which are both public. 
There is just one that does not, and that's the Mapper class in `SequentialAnomalyDetectionTransformBase` which is internal. This was the only instance where a mapper class does not have explicit implementation of the `Save` method. In the previous PR, we decided the that despite this being an internal class it was beneficial to do an explicit implementation for consistency as it implements `ICanSaveModel`. So all mappers elsewhere have an explicit implementation of that method. 

---
In reply to: [258776058](https://github.com/dotnet/machinelearning/pull/2673#discussion_r258776058) [](ancestors = 258776058,258775986)

src/Microsoft.ML.Data/DataLoadSave/TransformWrapper.cs
artidoro: Sounds good let me do that!

---
In reply to: [258775778](https://github.com/dotnet/machinelearning/pull/2673#discussion_r258775778) [](ancestors = 258775778)

src/Microsoft.ML.Data/DataLoadSave/TransformerChain.cs
```diff
@@ -216,10 +216,11 @@ public void SaveTo(IHostEnvironment env, Stream outputStream)
 
         IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
 
-        public IRowToRowMapper GetRowToRowMapper(DataViewSchema inputSchema)
+        IRowToRowMapper ITransformer.GetRowToRowMapper(DataViewSchema inputSchema)
         {
             Contracts.CheckValue(inputSchema, nameof(inputSchema));
-            Contracts.Check(IsRowToRowMapper, nameof(GetRowToRowMapper) + " method called despite " + nameof(IsRowToRowMapper) + " being false.");
+            Contracts.Check(((ITransformer)this).IsRowToRowMapper, nameof(ITransformer.GetRowToRowMapper) + " method called despite " +
```
wschin: >IsRowToRowMapper [](start = 49, length = 16)

I feel we should hide `IsRowToRowMapper` as well?

artidoro: I think you are asking whether we can make `IsRowToRowMapper` internal. 

I discussed this with Tom at length and the answer is that we should not make it internal. The reason is that when a `ITransformer` is a `IRowToRowMapper`, the performance of the prediction engine can be significantly increased compared to when it is not. Tom talked of a difference of orders of magnitude. 

Since it is such an important characteristics of the `ITransformer` we think it is better to keep it as part of the public API. This is especially true when we think about possibly allowing expansion of the library with custom components in the future.

wschin: If its for potential use, I feel we should hide for this version because I don't think users will be able to control the performance of ML.NET in the near future.

---
In reply to: [259442579](https://github.com/dotnet/machinelearning/pull/2673#discussion_r259442579) [](ancestors = 259442579)

artidoro: I will let @TomFinley explain his reasons.

TomFinley: I think in general the policy is a good one, that we should try to limit the scope of the public API. However, I feel very strongly that something like `IRowToRowMapper` has to exist.

The reason why I think this is the case is that I think back to #973. Literally everything *can* be captured as mapping from `IDataView` to `IDataView`, as we have seen. But we see that if we have a more "specific" interface for this row-to-row mapping, we can gain some benefit -- in this case, orders of magnitude speedup and a far more clear interface from using this "bypass" mechanism, in the case where we *can* use it.

From this perspective, we might say, "very well, then `IRowToRowMapper` should just be internal and prediction engine will still be free to use it." Which is of course possible. But then I think to myself, if `PredictionEngine` is useful from a reflection-driven perspective (as it quite clearly is), what is the dynamic API's answer to this scenario? There is nothing except `IRowToRowMapper`.

So, I think the case for something like `IRowToRowMapper` is clear, so I think we should ship it. So does it have to be that specifically? I don't see any better proposal than that interface. (Though perhaps it could be refined, e.g., with @sfilipi making it operating over schema columns specifically rather than `int => bool` delegates, etc.) But I think the principle has proven itself useful and should be made public, just like `IDataView` has been around, proven itself useful, and should be made public.

TomFinley: (I know this already closed, but though I'd answer the direct question. 😄 )

Issue #1065, opened 2018-09-26
WeightColumn name defaults to "Weights" in one of the constructors of the estimators


artidoro:
Estimators that are instantiated through MAML without specifying a weightcolumn are instead created with default weightcolumn named "Weights".

**More details:**
The constructor for estimator objects that is used by MAML takes an Arguments object, which has a weight column name that defaults to "Weights". To be precise it defaults to an Optional<string> with implicit value of "Weights".

When we instantiate the estimator through that constructor, we need to pass a SchemaShape.Column object to the base class TrainerEstimatorBase. This column is in most cases constructed with a method that takes the column name as specified in the Arguments object. As the default value in the Arguments object is not null, it is "Weights", this method does not return null, but returns a new SchemaShape.Column named Weights.

**The rational:**
The new API's rational is as follows:
1.  the user needs to specify a weight column name, if the name is not null we check the presence of the column
2.  if the user does not define a weight column name (it is null), we assume there is no weight column

The old MAML's rational:
1.  same as above
2. if the user does not specify a weight column name:
    -  if there is a column named "Weights" it will be taken as the weights column automatically
    - if there is no column name "Weights", we assume that there is no weight column

**What we have to do:**
We have to fix the current instantiation of estimators through MAML, and decide if we want to continue using a different behavior for the command line and the C# API.

TomFinley:
I just realized that we have not done this and it has implications for our public surface. This must be done soon. We currently have a situation where if there *happens* to be a column named `Weight`, there is no way to have unweighted training, which is really bad!!

@eerhardt and @ganik have context on this. It also reflects something I was talking about with @vinodshanbhag earlier today where I (wrongly) claimed that it was possible to not set weighted training in this scenario of the "incidental" `Weight column. I realized when reviewing #2665 that we are still doing this wrong thing.

The behavior should be as @artidoro describes above, that if the weight column is left `null`, you simply do not have weighted training. If you want weighted training, you must provide the column name to a non-null value.

eerhardt:
@ganik - This was fixed with #2674, correct? Can it be closed?

ganik:
@eerhardt I believe so, but let me double check with Artidoro

ganik:
Fixed with #2674

PR #2674, opened 2019-02-21
Mark EntryPoints classes and APIs as internal

### LINKED ISSUE #2582: Mark EntryPoints classes and APIs as internal
We have a few classes in `Microsoft.ML.Core` that are public, but they shouldn't be.

https://github.com/dotnet/machinelearning/blob/13b33392dd3fbffc0922b2af72c8236b74c95fc0/src/Microsoft.ML.Core/EntryPoints/TransformModel.cs#L14

https://github.com/dotnet/machinelearning/blob/13b33392dd3fbffc0922b2af72c8236b74c95fc0/src/Microsoft.ML.Core/EntryPoints/PredictorModel.cs#L14

Basically, anything under the `Microsoft.ML.EntryPoints` namespace should be internal.

This is OK, because anything that uses EntryPoints already has `InternalsVisibleTo`.

@TomFinley @glebuk @ganik @shauheen

ganik:
fixes #2582 #1065


src/Microsoft.ML.StandardLearners/Standard/Online/LinearSvm.cs
```diff
@@ -69,7 +69,8 @@ public sealed class Options : OnlineLinearOptions
             /// Column to use for example weight.
             /// </summary>
             [Argument(ArgumentType.AtMostOnce, HelpText = "Column to use for example weight", ShortName = "weight", SortOrder = 4, Visibility = ArgumentAttribute.VisibilityType.EntryPointsOnly)]
-            public Optional<string> WeightColumn = Optional<string>.Implicit(DefaultColumnNames.Weight);
+            [BestFriend]
+            internal Optional<string> WeightColumn = Optional<string>.Implicit(DefaultColumnNames.Weight);
```
Ivanidzo4ka: >internal Optional<string> WeightColumn = Optional<string>.Implicit(DefaultColumnNames.Weight); [](start = 12, length = 94)

How user suppose to specify weight column during training?
This algorithm supports weights column in dataview see line 149

src/Microsoft.ML.Core/EntryPoints/ModuleArgs.cs
```diff
@@ -133,7 +133,8 @@ public sealed class OutputAttribute : Attribute
         /// A node can be run without optional input fields.
         /// </summary>
         [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
-        public sealed class OptionalInputAttribute : Attribute { }
+        [BestFriend]
+        internal sealed class OptionalInputAttribute : Attribute { }
```
Ivanidzo4ka: >OptionalInputAttribute [](start = 30, length = 22)

why only this get changed?
And it's a part of internal static class, isn't it internal already?

src/Microsoft.ML.StandardLearners/Standard/Online/LinearSvm.cs
```diff
@@ -69,7 +69,8 @@ public sealed class Options : OnlineLinearOptions
             /// Column to use for example weight.
             /// </summary>
             [Argument(ArgumentType.AtMostOnce, HelpText = "Column to use for example weight", ShortName = "weight", SortOrder = 4, Visibility = ArgumentAttribute.VisibilityType.EntryPointsOnly)]
-            public Optional<string> WeightColumn = Optional<string>.Implicit(DefaultColumnNames.Weight);
+            [BestFriend]
```
TomFinley: >[BestFriend] [](start = 12, length = 12)

I see no particular reason for this to be best-friended with anything. Your public surface will not involve this attribute. This attribute will only be used via entry-points and suchlike.

src/Microsoft.ML.StandardLearners/Standard/Online/LinearSvm.cs
TomFinley: @Ivanidzo4ka is correct... this must continue to be supported.

To remind and possibly rephrase earlier conversations since they were somewhat informal:

1. You will rename this field `WeightColumn` something appropriate, like, say, `WeightColumnOptional`. To maintain the entry-point name, the `Argument` attribute will have its `Name` property set to `WeightColumn`. (Thereby making the entry-points remain unchanged.)

2. You will introduce another field or property named `WeightColumn` intended for use from the API. This will be a public field, unadorned by an argument attribute, as it is not intended for API use. This will be a `string`, not an `Optional<string>`.

3. The code in LightGBM will appropriately adjust to the presence of either of these, thereby responding appropriately whether one or the other is set (or, as it may happen, not set). This may require some additional (hidden!) parameters somewhere to distinguish whether something was called from the command line/entry-points (in which case we'd expect the argument attribute field to be set), or the public surface of the API. There are many obvious ways of doing this, I'll trust we can figure out which is best.

The same pattern should be followed in other places where you are providing an "alternative" to the options. We want our `Options` object, where exposed, to expose all available options. (This in accordance of our usual attitude towards how these objects should behave, as captured in #1798.)

src/Microsoft.ML.Data/Properties/AssemblyInfo.cs
```diff
@@ -26,6 +26,7 @@
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.HalLearners" + PublicKey.Value)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.KMeansClustering" + PublicKey.Value)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.LightGBM" + PublicKey.Value)]
+[assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.LightGBM.StaticPipe" + PublicKey.Value)]
```
eerhardt: Will we still need this after @tomfinley's comments are addressed?  (Basically to keep a `public string WeightColumn` property)

src/Microsoft.ML.StandardLearners/Standard/Online/LinearSvm.cs
ganik: As agreed made Weight and GroupId columns explicit strings

---
In reply to: [258773450](https://github.com/dotnet/machinelearning/pull/2674#discussion_r258773450) [](ancestors = 258773450)

Ivanidzo4ka:
image

src/Microsoft.ML.StandardLearners/FactorizationMachine/FactorizationMachineTrainer.cs
```diff
@@ -147,7 +147,7 @@ internal FieldAwareFactorizationMachineTrainer(IHostEnvironment env, Options opt
                 FeatureColumns[i + 1] = new SchemaShape.Column(options.ExtraFeatureColumns[i], SchemaShape.Column.VectorKind.Vector, NumberDataViewType.Single, false);
 
             LabelColumn = new SchemaShape.Column(options.LabelColumn, SchemaShape.Column.VectorKind.Scalar, BooleanDataViewType.Instance, false);
-            WeightColumn = options.WeightColumn.IsExplicit ? new SchemaShape.Column(options.WeightColumn, SchemaShape.Column.VectorKind.Scalar, NumberDataViewType.Single, false) : default;
+            WeightColumn = options.WeightColumn != null ? new SchemaShape.Column(options.WeightColumn, SchemaShape.Column.VectorKind.Scalar, NumberDataViewType.Single, false) : default;
```
Ivanidzo4ka: >ptions.WeightColumn != null ? new SchemaShape.Column(options.WeightColumn, SchemaShape.Column.VectorKind.Scalar, NumberDataViewType.Single, false) : default [](start = 28, length = 156)

TrainerUtils.MakeR4ScalarWeightColumn?

Issue #2634, opened 2019-02-19
Ranking / Ranker API inconsistency


rogancarr:
In the API, we have references to "Ranking" and "Ranker". One (of many) examples is that `mlContext.Ranking.Evaluate` returns a `RankerMetrics` object. This leads to an inconsistency across the API surface.

Ivanidzo4ka:
Just to make it clear. Which one do we prefer?

TomFinley:
> Just to make it clear. Which one do we prefer?

The task is `BinaryClassification`, `MulticlassClassification`, `Regression`, `Clustering`, `Ranking`. The metrics are correspondingly named `BinaryClassificationMetrics`, etc. The rank*er* is the thing that does the rank*ing*, so that is fine, but the metrics should be related to the task. It should be `RankingMetrics`.

TomFinley:
High level comment is this, I think renaming the evaluators and metrics is correct, but I would not have renamed the signature.

PR #2675, opened 2019-02-21
Changed Ranker to Ranking in evaluation related files.

zeahmed:
This PR fixes https://github.com/dotnet/machinelearning/issues/2634

Some of the EntryPoint names got changed. Reviewers' are requested to please verify consequences of changing EntryPoint names.


src/Microsoft.ML.Core/Prediction/ITrainer.cs
```diff
@@ -25,7 +25,7 @@ namespace Microsoft.ML
     [BestFriend]
     internal delegate void SignatureMultiOutputRegressorTrainer();
     [BestFriend]
-    internal delegate void SignatureRankerTrainer();
+    internal delegate void SignatureRankingTrainer();
```
TomFinley: >SignatureRankingTrainer [](start = 27, length = 23)

Since this is internal it is not a big deal, but note that this is incorrect. So you see how it is `SignatureBinaryClassifierTrainer` above? This is because the thing that does the *task* binary classificat*ion* is a binary classif*ier*. So this should have been retained as signature *ranker* trainer. What is the task? Ranking. What is the thing doing the task? A ranker.

Sort of like how things are being transformer, and the thing that does this is called a transformer.

Not a huge deal, but if you have to post another iteration anyway, this is what I would call a hypercorrection.

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
```diff
@@ -44,7 +44,7 @@ private static class DefaultEvaluatorTable
                             { MetadataUtils.Const.ScoreColumnKind.Regression, env => new RegressionMamlEvaluator(env, new RegressionMamlEvaluator.Arguments()) },
                             { MetadataUtils.Const.ScoreColumnKind.MultiOutputRegression, env => new MultiOutputRegressionMamlEvaluator(env, new MultiOutputRegressionMamlEvaluator.Arguments()) },
                             { MetadataUtils.Const.ScoreColumnKind.QuantileRegression, env => new QuantileRegressionMamlEvaluator(env, new QuantileRegressionMamlEvaluator.Arguments()) },
-                            { MetadataUtils.Const.ScoreColumnKind.Ranking, env => new RankerMamlEvaluator(env, new RankerMamlEvaluator.Arguments()) },
+                            { MetadataUtils.Const.ScoreColumnKind.Ranking, env => new RankingMamlEvaluator(env, new RankingMamlEvaluator.Arguments()) },
```
TomFinley: >RankingMamlEvaluator [](start = 86, length = 20)

Note that this renaming is correct. You are not evaluating a ranker, you are evaluating a ranking, so this one is good.

src/Microsoft.ML.EntryPoints/MacroUtils.cs
```diff
@@ -22,7 +22,7 @@ public enum TrainerKinds
         {
             SignatureBinaryClassifierTrainer,
             SignatureMultiClassClassifierTrainer,
-            SignatureRankerTrainer,
+            SignatureRankingTrainer,
```
TomFinley: >SignatureRankingTrainer [](start = 12, length = 23)

Another thing where this is I think hypercorrected. See how it is now inconsistent with everything else here?

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
TomFinley: It is also you note now *more* consistent with its "brethren" here.

---
In reply to: [258754361](https://github.com/dotnet/machinelearning/pull/2675#discussion_r258754361) [](ancestors = 258754361)

TomFinley:
Another way we could go is change *all* these things into things like `SignatureBinaryClassificationTrainer`. That also seems like a defensible choice, but we'd want them all to be consistent. That strikes me as possibly more work than we want to do right now and its urgency does not seem clear.

zeahmed:
Sure! I am changing `SignatureRankingTrainer` back to `SignatureRankerTrainer` so that we don't have to change many things right now.

---
In reply to: [465837115](https://github.com/dotnet/machinelearning/pull/2675#issuecomment-465837115) [](ancestors = 465837115)

zeahmed:
@Ivanidzo4ka, I accidentally click the re-request-review icon next to the review name above which made me lost your review approval. Can you please approve it back? Thanks!

Ivanidzo4ka:
once it's gone, IT'S GONE!

---
In reply to: [466194690](https://github.com/dotnet/machinelearning/pull/2675#issuecomment-466194690) [](ancestors = 466194690)

zeahmed:
@Ivanidzo4ka and @TomFinley, thanks for the useful comments and suggestions.

PR #2684, opened 2019-02-21
Adding defaults for Label and GroupId to Ranking Evaluator

### LINKED ISSUE #2633: Ranking evaluator doesn't have defaults for Label and GroupId
The `Evaluate()` method for `Ranking` takes a `label` and `groupId` as input, but unlike other `Evaluators` and the ranking trainers, it does not have default values, so these must always be specified. This leads to an inconsistent feeling across the API surface.

rogancarr:
This PR adds default column names for Label and GroupId for the Ranking `Evaluator`, in line with all other Evaluators.

Fixes #2633

### DIFF
--- a/src/Microsoft.ML.Data/TrainCatalog.cs
+++ b/src/Microsoft.ML.Data/TrainCatalog.cs
@@ -623,7 +623,10 @@ namespace Microsoft.ML
         /// <param name="groupId">The name of the groupId column in <paramref name="data"/>.</param>
         /// <param name="score">The name of the score column in <paramref name="data"/>.</param>
         /// <returns>The evaluation results for these calibrated outputs.</returns>
-        public RankerMetrics Evaluate(IDataView data, string label, string groupId, string score = DefaultColumnNames.Score)
+        public RankerMetrics Evaluate(IDataView data,
+            string label = DefaultColumnNames.Label,
+            string groupId = DefaultColumnNames.GroupId,
+            string score = DefaultColumnNames.Score)
         {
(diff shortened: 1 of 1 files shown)


danmosemsft:
@safern do you know what the magic is to make all the legs show up above in github, like they do now in CoreFX, rather than just one meta leg "MachineLearning-CI"

safern:
> @safern do you know what the magic is to make all the legs show up above in github, like they do now in CoreFX, rather than just one meta leg "MachineLearning-CI"

We need to change the connection settings in the build definition itself. That will also enable comment triggers for this repo. @eerhardt @shauheen I can take care of that, I think using the new connection setting is worth it as it also gets a lot of fixes from the azure devops CI team. However, it adds an extra step to get to the builds. When you click on details, it sends you to the checks tab in GH.

eerhardt:
I'm not sure if we need anything different than what we have today... I feel I'm productive in the current environment.

But if others feel it is valuable to have something different, I won't object unless I think that something different is worse than what we have today.

safern:
> But if others feel it is valuable to have something different, I won't object unless I think that something different is worse than what we have today.

To just feel the difference you can go into a corefx PR and look at the builds, click on one of them and try to navigate to it, etc.

eerhardt:
Yea, that experience is worse IMO.

safern:
> Yea, that experience is worse IMO.

Yeah, but I think is getting cool feature, like retrying a single job, rather than the whole pipeline, and comment triggers, optional builds triggered by comments, etc. Stuff corefx did require.

PR #2687, opened 2019-02-21
Update build tools version and use netcoreapp3.0 on ubuntu CI

### LINKED ISSUE #2381: Add Ubuntu CI leg

Anipik:
Fixes #2381 

Updates the buildtools version to latest and also running tests and building machinelearning repo with netcorepp3.0 on non-windows machine


Anipik:
@safern one of ubuntu configuration(Debug) is netcoreapp2.1

Anipik:
@tannergooding 
i am trying to use these flags COMPlus_EnableAvx and COMPlus_EnableSse on ubuntu netcoreapp3.0, these flags are getting set but Avx.IsSupported and Sse.IsSupported is still coming out to be true
​
do we need to do something else here ?

tannergooding:
@Anipik pinged me offline, the issue was with the casing of the Environment Variables. They work on Windows, but not on Unix because Windows is case-insensitive (most of the time) and Unix is case-sensitive (also, most of the time).

Anipik:
@eerhardt can you review this one ? thanks :)

PR #2697, opened 2019-02-22
Move metrics from percentages to [0,1]

### LINKED ISSUE #2637: Add a note that NDCG returns values between 0 and 100
NDCG, the normalized discounted cumulative gain metric that we often use in ranking, ranges in value between 0 and 100 in ML.NET. In the wide world, NDCG is often considered to be between 0 and 1 (e.g. see [Wikipedia](https://en.wikipedia.org/wiki/Discounted_cumulative_gain#Normalized_DCG)). We should add a note to the ranking metrics object that the values range between 0 and 100 so that we don't confuse people who are new to the toolkit.

rogancarr:
This PR moves the metrics `LogLoss`, `LogLossReduction`, and `NDCG` to be between 0 and 1 instead of 0 and 100 (i.e. from percentage to fraction) to be in line with the community standards.

Fixes #2637


src/Microsoft.ML.Data/Evaluators/RankerEvaluator.cs
```diff
@@ -806,7 +806,7 @@ protected override void UpdateState(RowCursorState state)
                 RankerUtils.QueryDcg(_labelGains, _truncationLevel, state.QueryLabels, state.QueryOutputs, state.DcgCur);
                 for (int t = 0; t < _truncationLevel; t++)
                 {
-                    Double ndcg = state.MaxDcgCur[t] > 0 ? state.DcgCur[t] / state.MaxDcgCur[t] * 100 : 0;
```
Ivanidzo4ka: >100 [](start = 98, length = 3)

[https://github.com/dotnet/machinelearning/blob/master/docs/images/NDCG.png](https://github.com/dotnet/machinelearning/blob/master/docs/images/NDCG.png)
you need to update this one :D

rogancarr: I swear I did! :)

---
In reply to: [259520368](https://github.com/dotnet/machinelearning/pull/2697#discussion_r259520368) [](ancestors = 259520368)

rogancarr: Do you mean change back to having the multiplier?

---
In reply to: [259520368](https://github.com/dotnet/machinelearning/pull/2697#discussion_r259520368) [](ancestors = 259520368)

Ivanidzo4ka: No, I mean we use this image in our documents. Would be strange if we change values to be from 0 to 1, but in documentation it would still be multiplied by 100

rogancarr: Oh, good eyes. I'll update the docstrings too.

---
In reply to: [259548793](https://github.com/dotnet/machinelearning/pull/2697#discussion_r259548793) [](ancestors = 259548793)

test/BaselineOutput/Common/Command/CommandTrainingBinaryFactorizationMachineWithInitialization-1-out.txt
```diff
@@ -15,7 +15,7 @@ Precision ||   0.9682 |   0.9438 |
 OVERALL 0/1 ACCURACY: 0.951684
 LOG LOSS/instance:  0.260415
 Test-set entropy (prior Log-Loss/instance): 0.934003
-LOG-LOSS REDUCTION (RIG): 72.118377
+LOG-LOSS REDUCTION (RIG): 0.721184
```
Ivanidzo4ka: > 0.721184 [](start = 25, length = 9)

GAINZ!

PR #2698, opened 2019-02-22
Remove the IFourierDistributionSampler interface

### LINKED ISSUE #699: RffTransform has undesirable coupling to its MatrixGenerator
`RffTransform` has an undesirable coupling to which kind of MatrixGenerator it is using (gaussian or not). 

https://github.com/dotnet/machinelearning/blob/e77f24e68d92f033910a5d783576d1879763d979/src/Microsoft.ML.Transforms/RffTransform.cs#L421-L425

It is using the ComponentCatalog to determine which type of MatrixGenerator it was working with before actually creating it. And then depending on the Type, it is doing a different distance scaling algorithm.

This type of decision (which algorithm to use for distance scaling) should be done by the MatrixGenerator itself, and not necessarily decided by the RffTransform.

I am changing RffTransform to no longer use SubComponent. However, I can no longer tell which type of MatrixGenerator is going to be created without actually creating the generator, so I needed to make a "dummy" instance. I spoke with @yaeldekel, and we decided this was "OK" for now, since it typically only used with a small number of columns (i.e. 1).

We should consider refactoring RffTransform and the IFourierDistributionSampler types to allow RffTransform to not have to switch based on whether Gaussian sampling is used or not.

### LINKED ISSUE #2659: What to do with IFourierDistributionSampler and friends
This interface is related to random Fourier features projection. We may want this functionality as part of our public surface, but the specific API right now is still very command-line and entry-points centric. This should change.

Somewhat similar in spirit to the problem of loss functions as discussed in #1973, and has many of the same problems, and should probably have a similar resolution to whatever we choose to do with loss functions.

yaeldekel:
Fixes #2659,
fixes #699.


src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -5,205 +5,252 @@
...
     /// <summary>
-    /// Signature for an IFourierDistributionSampler constructor.
+    /// Signature for a <see cref="RngGeneratorBase"/> constructor.
     /// </summary>
     [BestFriend]
-    internal delegate void SignatureFourierDistributionSampler(float avgDist);
+    internal delegate void SignatureRngGenerator();
 
-    public interface IFourierDistributionSampler : ICanSaveModel
+    public abstract class RngGeneratorBase
```
eerhardt: In general, .NET APIs prefer to write out the names. `Rng`. Also I assume the `g` in `Rng` is "generator", so it is a little redundant.

[https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions#using-abbreviations-and-acronyms](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions#using-abbreviations-and-acronyms)

X DO NOT use abbreviations or contractions as part of identifier names.

For example, use GetWindow rather than GetWin.

___________

Can we add some public XML doc to this new abstract interface?

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -5,205 +5,252 @@
...
     {
-        float Next(Random rand);
+        internal abstract float Dist(in VBuffer<float> first, in VBuffer<float> second);
+
+        internal abstract RandomNumberGeneratorBase GetRandomNumberGenerator(float avgDist);
```
eerhardt: What's the difference between `RngGeneratorBase` and `RandomNumberGeneratorBase`?

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
eerhardt: Maybe it would be good to tie this with `Fourier` as well? Since this is in the `Microsoft.ML.Transforms` namespace, having a `RngGeneratorBase` class seems a bit broad, especially one that doesn't have any public members on it.

---
In reply to: [259486033](https://github.com/dotnet/machinelearning/pull/2698#discussion_r259486033) [](ancestors = 259486033)

yaeldekel:
The problem with `IFourierDistributionSampler`, as @eric.erhardt@microsoft.com mentioned in issue #699, is that creating one requires a float value (`avgDist`), but the way `avgDist` is computed, depends on the implementation of the interface. So in this PR, I am breaking up this interface into two base classes instead. The first one is used to help in the computation of `avgDist`, and then it is used to create an instance of the second one. The second one, has the `Next` API to produce random numbers.
I wasn't able to find good names for these classes, so I would appreciate any ideas :). I called the first class `RngGenerator` because it knows how to generate a random number generator...
If we leave the second class internal, perhaps it could stay `RandomNumberGeneratorBase`?

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
yaeldekel: Yes, I was going to mention that I'm not happy with these class names... I wrote an explanation about these classes in a separate comment, I'll also add some XML comments and post another iteration for this PR.


---
In reply to: [259486033](https://github.com/dotnet/machinelearning/pull/2698#discussion_r259486033) [](ancestors = 259486033)

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -5,205 +5,252 @@
...
+    "Gaussian Fourier Sampler Executor", "GaussianSamplerExecutor", GaussianRngGenerator.RandomNumberGenerator.LoaderSignature)]
 
 // This is for deserialization from a binary model file.
-[assembly: LoadableClass(typeof(LaplacianFourierSampler), null, typeof(SignatureLoadModel),
-    "Laplacian Fourier Sampler Executor", "LaplacianSamplerExecutor", LaplacianFourierSampler.LoaderSignature)]
+[assembly: LoadableClass(typeof(LaplacianRngGenerator.RandomNumberGenerator), null, typeof(SignatureLoadModel),
+    "Laplacian Fourier Sampler Executor", "LaplacianSamplerExecutor", LaplacianRngGenerator.RandomNumberGenerator.LoaderSignature)]
 
 // REVIEW: Roll all of this in with the RffTransform.
 namespace Microsoft.ML.Transforms
```
Ivanidzo4ka: I would remove this comment, and change namespace to `Microsoft.ML.Transforms.Projections` to be align with RFF

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -5,205 +5,252 @@
...
     /// </summary>
     [BestFriend]
-    internal delegate void SignatureFourierDistributionSampler(float avgDist);
+    internal delegate void SignatureRngGenerator();
 
-    public interface IFourierDistributionSampler : ICanSaveModel
+    public abstract class RngGeneratorBase
     {
-        float Next(Random rand);
+        internal abstract float Dist(in VBuffer<float> first, in VBuffer<float> second);
```
Ivanidzo4ka: >Dist [](start = 32, length = 4)

I know it's internal, but since we touching this, can we call it Distance?

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
Ivanidzo4ka: >RngGeneratorBase [](start = 26, length = 16)

I would call it `DistributionBase`, or KernelBase.

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -5,205 +5,252 @@
...
+
+        internal abstract RandomNumberGeneratorBase GetRandomNumberGenerator(float avgDist);
     }
 
-    [TlcModule.ComponentKind("FourierDistributionSampler")]
-    internal interface IFourierDistributionSamplerFactory : IComponentFactory<float, IFourierDistributionSampler>
+    internal abstract class RandomNumberGeneratorBase
```
Ivanidzo4ka: >internal abstract class RandomNumberGeneratorBase [](start = 4, length = 49)

any reason why it's not interface?

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -5,205 +5,252 @@
...
+                    loaderSignature: LoaderSignature,
+                    loaderAssemblyName: typeof(RandomNumberGenerator).Assembly.FullName);
+            }
+
+            private readonly float _gamma;
+
+            public RandomNumberGenerator(float gamma, float avgDist)
+                : base()
+            {
+                _gamma = gamma / avgDist;
```
Ivanidzo4ka: >avgDist [](start = 33, length = 7)

would be nice to have Assert for non zero value.

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -5,205 +5,252 @@
...
+                return (float)Stats.SampleFromGaussian(rand) * MathUtils.Sqrt(2 * _gamma);
+            }
         }
     }
 
-    public sealed class LaplacianFourierSampler : IFourierDistributionSampler
+    public sealed class LaplacianRngGenerator : RngGeneratorBase
     {
-        public sealed class Options : IFourierDistributionSamplerFactory
+        public sealed class Options : IComponentFactory<RngGeneratorBase>
```
Ivanidzo4ka: >Options [](start = 28, length = 7)

does it still need to be public?

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -5,205 +5,252 @@
...
+            }
+
+            internal const string LoaderSignature = "RandLaplacianFourierExec";
+            internal const string RegistrationName = "LaplacianRandom";
+
+            private readonly float _a;
+
+            public RandomNumberGenerator(float a, float avgDist)
+            {
+                _a = a / avgDist;
```
Ivanidzo4ka: >avgDist [](start = 25, length = 7)

assert for non zero value.

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -5,205 +5,270 @@
...
-namespace Microsoft.ML.Transforms
+namespace Microsoft.ML.Transforms.Projections
 {
     /// <summary>
-    /// Signature for an IFourierDistributionSampler constructor.
+    /// Signature for a <see cref="KernelBase"/> constructor.
     /// </summary>
     [BestFriend]
-    internal delegate void SignatureFourierDistributionSampler(float avgDist);
+    internal delegate void SignatureRngGenerator();
```
eerhardt: Should we rename this for the new name? - `SignatureKernelBase`?

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -5,205 +5,270 @@
...
 
-    public interface IFourierDistributionSampler : ICanSaveModel
+    /// <summary>
+    /// This class indicates which kernel should be approximated by the <see cref="RandomFourierFeaturizingTransformer"/>.
+    /// <seealso cref="RandomFourierFeaturizingEstimator"/>.
+    /// </summary>
+    public abstract class KernelBase
```
eerhardt: We should add a `private protected` constructor to this class, to explicitly tell external devs that they can't inherit from it.
The way it is now, it has an implicit `protected` constructor, which makes it seem like someone could inherit from it. But if they try, they will get a compile error saying it doesn't implement the abstract methods below. Having the `private protected` constructor shows that we intend that no one outside of ML.NET should inherit from this class.

src/Microsoft.ML.Transforms/RandomFourierFeaturizing.cs
```diff
@@ -367,23 +367,15 @@ private float[] GetAvgDistances(RandomFourierFeaturizingEstimator.ColumnInfo[] c
                             for (int i = 0; i < instanceCount; i++)
                             {
                                 for (int j = i + 1; j < instanceCount; j++)
-                                {
-                                    distances[count++] = gaussian ? VectorUtils.L2DistSquared(in res[i], in res[j])
-                                        : VectorUtils.L1Distance(in res[i], in res[j]);
-                                }
+                                    distances[count++] = columns[iinfo].Generator.Distance(in res[i], in res[j]);
```
eerhardt: You may want to factor out `columns[iinfo].Generator` into a new variable (here and below). That way we aren't indexing into the array over and over inside these nested for loops.

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -5,205 +5,270 @@
...
-                loaderAssemblyName: typeof(GaussianFourierSampler).Assembly.FullName);
+            public KernelBase CreateComponent(IHostEnvironment env) => new GaussianKernel(env, this);
         }
 
         internal const string LoadName = "GaussianRandom";
 
         private readonly float _gamma;
 
-        public GaussianFourierSampler(IHostEnvironment env, Options options, float avgDist)
+        public GaussianKernel(float gamma = 1)
```
Ivanidzo4ka: >GaussianKernel [](start = 15, length = 14)

would be nice to have comment.
Also `MathUtils.Sqrt(2 * _gamma)`
Since I doubt average distance can be negative and _gamma in random generator is this gamma / avgDistance, we need to specify what gamma is non-negative number. and assert it.

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -5,205 +5,270 @@
...
-
-        internal const string LoaderSignature = "RandLaplacianFourierExec";
-        internal const string RegistrationName = "LaplacianRandom";
+        internal const string LoadName = "LaplacianRandom";
 
-        private readonly IHost _host;
         private readonly float _a;
 
-        public LaplacianFourierSampler(IHostEnvironment env, Options options, float avgDist)
+        public LaplacianKernel(float a = 1)
```
Ivanidzo4ka: would be nice to have comment.

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -5,205 +5,270 @@
...
+    /// (<a href="https://en.wikipedia.org/wiki/Bochner%27s_theorem">Bochner's theorem</a>). This class
+    /// samples numbers from the non-negative measure corresponding to the given kernel.
+    /// </summary>
+    internal abstract class FourierRandomNumberGeneratorBase
     {
+        public abstract float Next(Random rand);
     }
 
-    public sealed class GaussianFourierSampler : IFourierDistributionSampler
+    public sealed class GaussianKernel : KernelBase
```
Ivanidzo4ka: >GaussianKernel [](start = 24, length = 14)

would be nice to have comment.

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -5,205 +5,270 @@
...
+
+            public override float Next(Random rand)
+            {
+                return (float)Stats.SampleFromGaussian(rand) * MathUtils.Sqrt(2 * _gamma);
+            }
         }
     }
 
-    public sealed class LaplacianFourierSampler : IFourierDistributionSampler
+    public sealed class LaplacianKernel : KernelBase
```
Ivanidzo4ka: >LaplacianKernel [](start = 24, length = 15)

would be nice to have comment

PR #2702, opened 2019-02-22
Make separator char[] everywhere (previous type is char sometime)

### LINKED ISSUE #2472: Any reason for inconsistency in params for TextLoader?
1) This Method :
https://github.com/dotnet/machinelearning/blob/834e4715afa070f5e5c3628693682943b3f0bd6d/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs#L55

2) and this one : 
https://github.com/dotnet/machinelearning/blob/834e4715afa070f5e5c3628693682943b3f0bd6d/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs#L22

Looks like the first one has params like : 
```
char separatorChar = '\t', bool allowQuotedStrings = true, bool supportSparse = true, bool trimWhitespace = false
```

Where as the second one doesn't have those. In order to use those params there is another method overload : 
```
CreateTextLoader(this DataOperationsCatalog catalog, TextLoader.Arguments args, IMultiStreamSource dataSample = null);
```

This takes Arguments class and particularly one param is interesting in this Arguments:
```
char [] Separators
```
 Any reason this is char[] and not char like in other overloads? and why inconsistency in params for the methods 1) and 2).

wschin:
Fix #2472. This PR makes all separators a `char[]` instead of `char` in the public area of `TextLoader`. Note that `TextLoader`'s advanced `options` is already using `char[]`.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -428,13 +428,13 @@ namespace Microsoft.ML.Data
 
             [Argument(ArgumentType.AtMostOnce, Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly, HelpText = "Source column separator. Options: tab, space, comma, single character", ShortName = "sep")]
             // this is internal as it only serves the command line interface
-            internal string Separator = "\t";
+            internal string Separator = Defaults.Separator;
 
             /// <summary>
             /// The characters that should be used as separators column separator.
             /// </summary>
             [Argument(ArgumentType.AtMostOnce, Name = nameof(Separator), Visibility = ArgumentAttribute.VisibilityType.EntryPointsOnly, HelpText = "Source column separator.", ShortName = "sep")]
-            public char[] Separators = Defaults.Separator;
+            public char[] Separators = Defaults.Separators;
 
             /// <summary>
             /// Specifies the input columns that should be mapped to <see cref="IDataView"/> columns.
@@ -488,7 +488,8 @@ namespace Microsoft.ML.Data
         {
             internal const bool AllowQuoting = false;
             internal const bool AllowSparse = false;
-            internal static char[] Separator => new[] { '\t' };
+            internal static char[] Separators => new[] { '\t' };
+            internal const string Separator = "\t";
             internal const bool HasHeader = false;
             internal const bool TrimWhitespace = false;
         }
@@ -1064,15 +1065,15 @@ namespace Microsoft.ML.Data
         /// </summary>
         /// <param name="env">The environment to use.</param>
         /// <param name="columns">Defines a mapping between input columns in the file and IDataView columns.</param>
-        /// <param name="separatorChar"> The character used as separator between data points in a row. By default the tab character is used as separator.</param>
+        /// <param name="separators"> The character used as separator between data points in a row. {'\t'} will be used if not specified.</param>
         /// <param name="hasHeader">Whether the file has a header.</param>
         /// <param name="allowSparse">Whether the file can contain numerical vectors in sparse format.</param>
         /// <param name="allowQuoting">Whether the content of a column can be parsed from a string starting and ending with quote.</param>
         /// <param name="dataSample">Allows to expose items that can be used for reading.</param>
-        internal TextLoader(IHostEnvironment env, Column[] columns, char[] separatorChar = null,
+        internal TextLoader(IHostEnvironment env, Column[] columns, char[] separators = null,
             bool hasHeader = Defaults.HasHeader, bool allowSparse = Defaults.AllowSparse,
             bool allowQuoting = Defaults.AllowQuoting, IMultiStreamSource dataSample = null)
-            : this(env, MakeArgs(columns, hasHeader,  separatorChar ?? Defaults.Separator, allowSparse, allowQuoting), dataSample)
+            : this(env, MakeArgs(columns, hasHeader,  separators ?? Defaults.Separators, allowSparse, allowQuoting), dataSample)
         {
         }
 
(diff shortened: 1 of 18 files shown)


src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -23,12 +23,12 @@ public static class TextLoaderSaverCatalog
         /// <param name="dataSample">The optional location of a data sample. The sample can be used to infer column names and number of slots in each column.</param>
         public static TextLoader CreateTextLoader(this DataOperationsCatalog catalog,
             TextLoader.Column[] columns,
-            char separatorChar = TextLoader.Defaults.Separator,
+            char[] separatorChar = null,
```
Ivanidzo4ka: >separatorChar [](start = 19, length = 13)

`separators` please

wschin: Fixed.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1064,15 +1065,15 @@ private bool HasHeader
         /// </summary>
         /// <param name="env">The environment to use.</param>
         /// <param name="columns">Defines a mapping between input columns in the file and IDataView columns.</param>
-        /// <param name="separatorChar"> The character used as separator between data points in a row. By default the tab character is used as separator.</param>
+        /// <param name="separators"> The character used as separator between data points in a row. {'\t'} will be used if not specified.</param>
```
Ivanidzo4ka: >The character used [](start = 38, length = 18)

Array of characters.

Ivanidzo4ka: >{'\t'} [](start = 100, length = 6)

what is wrong with `tab` ?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -45,7 +45,7 @@ public static class TextLoaderSaverCatalog
         /// Create a text loader <see cref="TextLoader"/> by inferencing the dataset schema from a data model type.
         /// </summary>
         /// <param name="catalog">The <see cref="DataOperationsCatalog"/> catalog.</param>
-        /// <param name="separatorChar">Column separator character. Default is '\t'</param>
+        /// <param name="separators">Column separator characters. {'\t'} will be used if not specified.</param>
```
Ivanidzo4ka: >Column separator characters. [](start = 37, length = 28)

Please be consistent with description for same property across different methods.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
wschin: `tab` is scalar, not an array.

---
In reply to: [259558700](https://github.com/dotnet/machinelearning/pull/2702#discussion_r259558700) [](ancestors = 259558700)

Ivanidzo4ka: ok, what is difference from user standpoint between array with one element and scalar?

---
In reply to: [259559041](https://github.com/dotnet/machinelearning/pull/2702#discussion_r259559041) [](ancestors = 259559041,259558700)

wschin: Their types are obviously different, right?

---
In reply to: [259559198](https://github.com/dotnet/machinelearning/pull/2702#discussion_r259559198) [](ancestors = 259559198,259559041,259558700)

artidoro:
Since the most common scenario is that we only have one separator, and not multiple separators I think we should add an overload with a `separatorChar` instead of a `separators` array of chars.

wschin:
That I am not sure. This sounds a `new feature` because it's already doable. In addition, I feel we should have only one way to do one thing.

---
In reply to: [466604514](https://github.com/dotnet/machinelearning/pull/2702#issuecomment-466604514) [](ancestors = 466604514)

artidoro:
There are quite a few instances in which we added an overload in the catalog extensions. 
See for example `DropColumns` and `KeepColumns` (https://github.com/dotnet/machinelearning/blob/44c3113c83ebc7ea2d1a43f404f602ec7202b52c/src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs#L108-L109).

The old version of `TextLoader` had a separatorchar for a reason: it is quite annoying to have to define an array of a single character every time you use `TextLoader`.

In fact it is quite common in our extensions to have a "simple" scenario and an advanced options scenario with two different extensions for the two scenarios.

artidoro:
Actually now that I take a deeper look at it, I would keep the single `separatorChar` everywhere for the simple scenario where we specify the arguments directly in the extension, while we have `char[] separators` in the `Options` class that is used as an argument in the advanced scenario.

wschin:
I don't quite agree. We have only one separator because all our text files are cleaned. Is this assumption applied to other places?

---
In reply to: [466606912](https://github.com/dotnet/machinelearning/pull/2702#issuecomment-466606912) [](ancestors = 466606912)

TomFinley:
Hi @wschin thanks for working on this. I will say that my initial thinking was that "multi-separator" files were fairly rare. (People know what TSV files are, people know what CSV files are, but I don't think too many people know or would be sympathetic to the idea of the "T or C SV" file.) For that reason, I think while the advanced options should continue to support multiple characters (I do know it does happen), the convenience API should keep the separator as `char`, not `char[]`. What you think?

Also including @sfilipi and @rogancarr.

TomFinley:
> I don't quite agree. We have only one separator because all our text files are cleaned. Is this assumption applied to other places?
> 
> In reply to: [466606912](https://github.com/dotnet/machinelearning/pull/2702#issuecomment-466606912) [](ancestors = 466606912)

I'm pretty sure. When I look back on DRI responses internally, I don't see the multi-separate capability used too often, even for user data.

wschin:
Ok, so I will close this PR.

---
In reply to: [467074220](https://github.com/dotnet/machinelearning/pull/2702#issuecomment-467074220) [](ancestors = 467074220)

PR #2706, opened 2019-02-23
Remove MD5Hasher.

### LINKED ISSUE #2696: Need to remove FastTree/Utils/MD5Hasher class
Currently FastTree uses a 16byte MD5 hash to ensure uniqueness of content. 

### There are several issues with this:
- The MD5 hash is a (weak) cryptographic hash, resulting in additional work for the release process with regards to export controls.
- MD5 hash is quite slow and memory intensive

### Suggested Fix:
- Rename MD5Hash to ContentHash or similar. 
- Same goes to MD5Hasher
- Update all Hash(,,,) methods to not use MD5CryptoServiceProvider but murmur hash or equivalent instead.
- Ensure no reference to System.Security.Cryptography namespace exists in the codebase afterwards.
- Alternatively, we can extend the murmurhash class to return

codemzs:
fixes #2696


src/Microsoft.ML.FastTree/BinFile/IniFileParserInterface.cs
```diff
@@ -254,8 +254,6 @@ public sealed class FeatureEvaluator
 
             public int Id { get; }
 
-            public MD5Hash ContentMD5Hash => MD5Hasher.Hash(Content);
```
glebuk: >ContentMD5Hash [](start = 27, length = 14)

Strange to have a fully unused field.  Looks like internal code does not use that either?

TomFinley:
I know @eerhardt had some interest in this, mentioning him as informational.

PR #2707, opened 2019-02-25
Make DataViewRowId not act like a number.

eerhardt:
- Remove it from the NumberDataViewType.
- Remove any method/operator that makes it feel like a number.

Working towards #2297

### DIFF
--- a/src/Microsoft.Data.DataView/DataViewType.cs
+++ b/src/Microsoft.Data.DataView/DataViewType.cs
@@ -199,17 +199,6 @@ namespace Microsoft.Data.DataView
             }
         }
 
-        private static volatile NumberDataViewType _instDataViewRowId;
-        public static NumberDataViewType DataViewRowId
-        {
-            get
-            {
-                return _instDataViewRowId ??
-                    Interlocked.CompareExchange(ref _instDataViewRowId, new NumberDataViewType(typeof(DataViewRowId), "UG"), null) ??
-                    _instDataViewRowId;
-            }
-        }
-
         private static volatile NumberDataViewType _instSingle;
         public static NumberDataViewType Single
         {
@@ -241,6 +230,41 @@ namespace Microsoft.Data.DataView
         }
 
         public override string ToString() => _name;
+    }
+
+    /// <summary>
+    /// The DataViewRowId type.
+    /// </summary>
+    public sealed class RowIdDataViewType : PrimitiveDataViewType
+    {
+        private static volatile RowIdDataViewType _instance;
+        public static RowIdDataViewType Instance
+        {
+            get
+            {
+                return _instance ??
+                    Interlocked.CompareExchange(ref _instance, new RowIdDataViewType(), null) ??
+                    _instance;
+            }
+        }
+
+        private RowIdDataViewType()
+            : base(typeof(DataViewRowId))
+        {
+        }
+
+        public override bool Equals(DataViewType other)
+        {
+            if (other == this)
+                return true;
+            Debug.Assert(!(other is RowIdDataViewType));
+            return false;
+        }
+
+        public override string ToString()
+        {
+            return "DataViewRowId";
+        }
     }
 
     /// <summary>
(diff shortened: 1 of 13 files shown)


src/Microsoft.Data.DataView/DataViewType.cs
```diff
@@ -243,6 +232,41 @@ public override bool Equals(DataViewType other)
...
+    /// </summary>
+    public sealed class RowIdDataViewType : PrimitiveDataViewType
+    {
+        private static volatile RowIdDataViewType _instance;
+        public static RowIdDataViewType Instance
+        {
+            get
+            {
+                return _instance ??
+                    Interlocked.CompareExchange(ref _instance, new RowIdDataViewType(), null) ??
```
rogancarr: >Interlocked.CompareExchange [](start = 20, length = 27)

Nice!

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
```diff
@@ -118,6 +120,8 @@ public static PrimitiveDataViewType PrimitiveTypeFromKind(DataKind kind)
                 return DateTimeDataViewType.Instance;
             if (kind == DataKind.DZ)
                 return DateTimeOffsetDataViewType.Instance;
+            if (kind == DataKind.UG)
```
rogancarr: >DataKind.UG [](start = 24, length = 11)

Since `DataKind.UG` is only for these `RowId` instances, perhaps we should rename `DataKind.UG` to `DataKind.RowId`. When I first ran across `UG`, I thought it was for holding large unsigned integers.... Could be a more user friendly nomenclature, I guess.

rogancarr:
>             private DataViewRowId _valueCount;

Perhaps this should just be a long?

It looks like we're just accumulating integers here, no?

---
Refers to: src/Microsoft.ML.Transforms/MissingValueReplacingUtils.cs:185 in f25882e. [](commit_id = f25882e08e897a2204c985ca44a20918c93ad19a, deletion_comment = False)

src/Microsoft.ML.Core/Data/ColumnTypeExtensions.cs
eerhardt: @wschin is currently working on #2661, which makes `DataKind.UG` internal. So it wouldn't be user facing.

rogancarr:
A comment and a quick question before I sign off:

**Comment**: It looks like this was designed so that we could take slices of an `IDataView`. We lose that functionality if we remove this. But in practice, we take slices by creating dummy columns anyways, so it's probably just not necessary.

**Question**: Can we just remove all the Add / Subtract / ToDouble / etc. methods? The whole business of counting the number of rows processed here could be done with a `ulong`, couldn't it? Or am I missing something?

TomFinley:
> Comment: It looks like this was designed so that we could take slices of an IDataView.

That definitely was not my intent, and I definitely did not design it with that in mind. Please [read here](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.Data.DataView/ICursor.md#getidgetter) for more details on what it is used for.

---
In reply to: [467108963](https://github.com/dotnet/machinelearning/pull/2707#issuecomment-467108963) [](ancestors = 467108963)

TomFinley:
>             private DataViewRowId _valueCount;

We're accumulating integers, but the critical point is how large those integers might conceivably get. Let's imagine we have a dataset with 1 billion features (sparse, of course), and 20 billion instances. The accumulation of lengths will exceed the capacity of a `ulong`. With sparse vectors, these values can easily grow to be fairly large.

Have I seen datasets where the feature array is 1 billion large? Yes, a few times. Have I seen datasets with many billions of examples? Also yes. So I don't view it as ridiculous that a `ulong` might be judged to be insufficient to calculate this quantity. Now, then, *could* this code be changed to do something else? Again, maybe yes, but not without changing the implementation significantly, and it's unclear to me what the benefit is. So I'd prefer to just let @eerhardt's changes as written here stand.

---
In reply to: [467106243](https://github.com/dotnet/machinelearning/pull/2707#issuecomment-467106243) [](ancestors = 467106243)

---
Refers to: src/Microsoft.ML.Transforms/MissingValueReplacingUtils.cs:185 in f25882e. [](commit_id = f25882e08e897a2204c985ca44a20918c93ad19a, deletion_comment = False)

PR #2709, opened 2019-02-25
Renaming ColumnInfo to ColumnOptions

### LINKED ISSUE #2554: ColumnInfo as an API parameter

artidoro:
Fixes #2554.

As discussed in the issue, we decided to rename `ColumnInfo` to `ColumnOptions`.
I used Visual Studio to find and replace instances of `ColumnInfo`. I did an operation of find and replace for each different casing. I also went over the changes to double check.


src/Microsoft.ML.Data/DataDebuggerPreview.cs
```diff
@@ -94,14 +94,14 @@ internal RowInfo(int n)
             }
         }
 
-        public sealed class ColumnInfo
+        public sealed class ColumnOptions
```
Ivanidzo4ka: >ColumnOptions [](start = 28, length = 13)

Why?

sfilipi: +1 :)

---
In reply to: [259962854](https://github.com/dotnet/machinelearning/pull/2709#discussion_r259962854) [](ancestors = 259962854)

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
```diff
@@ -9,24 +9,24 @@
 
 namespace Microsoft.ML
 {
-    public sealed class SimpleColumnInfo
+    public sealed class ColumnOptions
```
sfilipi: >ColumnOptions [](start = 24, length = 13)

SimpleColumnOptions

test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamplesDynamicApi.cs
```diff
@@ -243,9 +243,9 @@ private void NormalizationWorkout(string dataPath)
             // Apply all kinds of standard ML.NET normalization to the raw features.
             var pipeline =
                 mlContext.Transforms.Normalize(
-                    new NormalizingEstimator.MinMaxColumn("MinMaxNormalized", "Features", fixZero: true),
-                    new NormalizingEstimator.MeanVarColumn("MeanVarNormalized", "Features", fixZero: true),
-                    new NormalizingEstimator.BinningColumn("BinNormalized", "Features", numBins: 256));
+                    new NormalizingEstimator.MinMaxColumnOptions("MinMaxNormalized", "Features", fixZero: true),
```
artidoro: Update the cookbook

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
artidoro: It was intentional, why do you think that `SimpleColumnOptions` is better?
I thought that for consistency it was better like this.

src/Microsoft.ML.Data/DataDebuggerPreview.cs
artidoro: Will revert that.

Issue #2663, opened 2019-02-20
SlotNames for TextLoader are lost


Ivanidzo4ka:
Before refactoring if we had header in file and we read it we filled slot names metadata with values in that header for columns.
This way we can have mapping between field "A" in csv file and slot number 5 in feature vector.

This functionality is lost right now.
Mostly because we split functionality of schema construction which done without file and reading data from file with already defined schema.
If I have this header:
`Label A B C D E F G ....`
and this source code:
```
            var reader = mlContext.Data.CreateTextLoader(new Microsoft.ML.Data.TextLoader.Column[] { new Microsoft.ML.Data.TextLoader.Column("Label", Microsoft.ML.Data.DataKind.R4, 0), new Microsoft.ML.Data.TextLoader.Column("Features", Microsoft.ML.Data.DataKind.R4, 1, 100) },
                hasHeader: true, separatorChar: ' ');
            var data = reader.Read("data.txt");
```
I expect Features column to have SlotNames metadata with values `A B C D E F G`, etc.

Ivanidzo4ka:
Sparked by https://github.com/dotnet/machinelearning/issues/2240
>Additionally, the model is not explorable, because the class I have to represent a single prediction has each column mapped to a different field, but those fields cannot be identified in the model.

CESARDELATORRE:
@Ivanidzo4ka - Days ago, using 0.10, we changed the two multi-class classification samples in the ML.NET samples repo and both of them are using this info coming from the schema so we know the "best three labels" for the classification based on the scores.

This should work on 0.11 or it'll break the code that customers might also be using.

@prathyusha12345 - Be aware of this issue. Can yo test these apps with 0.11, as well?

prathyusha12345:
@CESARDELATORRE  Sure I will test this in 0.11 preview today at sometime and update whether its executing fine or not.

Ivanidzo4ka:
@CESARDELATORRE  Can you point me on changes? I think we speak about slightly different things.

What I'm talking is you have option of reading each column separately and then concat them into feature columns and I think we can look into SlotNames metadata for Features which would have information of previous column names.

Or you can read one big Feature column by specifying range of columns in text loader.
And in this case you have no way to trace back original names.

Ivanidzo4ka:
@rogancarr  I know you working on samples do you have any examples where slot names can be useful? Like feature importance?

CESARDELATORRE:
[...]

PR #2710, opened 2019-02-25
Make text loaders consistent

### LINKED ISSUE #2472: Any reason for inconsistency in params for TextLoader?

wschin:
Another attempt to fix #2472.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1089,32 +1089,6 @@ namespace Microsoft.ML.Data
         /// Loads a text file into an <see cref="IDataView"/>. Supports basic mapping from input columns to IDataView columns.
         /// </summary>
         /// <param name="env">The environment to use.</param>
-        /// <param name="columns">Defines a mapping between input columns in the file and IDataView columns.</param>
-        /// <param name="separatorChar"> The character used as separator between data points in a row. By default the tab character is used as separator.</param>
-        /// <param name="hasHeader">Whether the file has a header.</param>
-        /// <param name="allowSparse">Whether the file can contain numerical vectors in sparse format.</param>
-        /// <param name="allowQuoting">Whether the content of a column can be parsed from a string starting and ending with quote.</param>
-        /// <param name="dataSample">Allows to expose items that can be used for reading.</param>
-        /// <param name="trimWhitespace">Remove trailing whitespace from lines.</param>
-        internal TextLoader(IHostEnvironment env, Column[] columns, char separatorChar = Defaults.Separator,
-            bool hasHeader = Defaults.HasHeader, bool allowSparse = Defaults.AllowSparse,
-            bool allowQuoting = Defaults.AllowQuoting, IMultiStreamSource dataSample = null, bool trimWhitespace = Defaults.TrimWhitespace)
-            : this(env, MakeArgs(columns, hasHeader, new[] { separatorChar }, allowSparse, allowQuoting, trimWhitespace), dataSample)
-        {
-        }
-
-        private static Options MakeArgs(Column[] columns, bool hasHeader, char[] separatorChars, bool allowSparse, bool allowQuoting, bool trimWhitespace)
-        {
-            Contracts.AssertValue(separatorChars);
-            var result = new Options { Columns = columns, HasHeader = hasHeader, Separators = separatorChars,
-                AllowSparse = allowSparse, AllowQuoting = allowQuoting, TrimWhitespace = trimWhitespace };
-            return result;
-        }
-
-        /// <summary>
-        /// Loads a text file into an <see cref="IDataView"/>. Supports basic mapping from input columns to IDataView columns.
-        /// </summary>
-        /// <param name="env">The environment to use.</param>
         /// <param name="options">Defines the settings of the load operation.</param>
         /// <param name="dataSample">Allows to expose items that can be used for reading.</param>
         internal TextLoader(IHostEnvironment env, Options options = null, IMultiStreamSource dataSample = null)
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1095,17 +1095,19 @@ namespace Microsoft.ML.Data
         /// <param name="allowSparse">Whether the file can contain numerical vectors in sparse format.</param>
         /// <param name="allowQuoting">Whether the content of a column can be parsed from a string starting and ending with quote.</param>
         /// <param name="dataSample">Allows to expose items that can be used for reading.</param>
+        /// <param name="trimWhitespace">Remove trailing whitespace from lines.</param>
         internal TextLoader(IHostEnvironment env, Column[] columns, char separatorChar = Defaults.Separator,
             bool hasHeader = Defaults.HasHeader, bool allowSparse = Defaults.AllowSparse,
-            bool allowQuoting = Defaults.AllowQuoting, IMultiStreamSource dataSample = null)
-            : this(env, MakeArgs(columns, hasHeader, new[] { separatorChar }, allowSparse, allowQuoting), dataSample)
+            bool allowQuoting = Defaults.AllowQuoting, IMultiStreamSource dataSample = null, bool trimWhitespace = Defaults.TrimWhitespace)
+            : this(env, MakeArgs(columns, hasHeader, new[] { separatorChar }, allowSparse, allowQuoting, trimWhitespace), dataSample)
         {
         }
 
-        private static Options MakeArgs(Column[] columns, bool hasHeader, char[] separatorChars, bool allowSparse, bool allowQuoting)
+        private static Options MakeArgs(Column[] columns, bool hasHeader, char[] separatorChars, bool allowSparse, bool allowQuoting, bool trimWhitespace)
         {
             Contracts.AssertValue(separatorChars);
-            var result = new Options { Columns = columns, HasHeader = hasHeader, Separators = separatorChars, AllowSparse = allowSparse, AllowQuoting = allowQuoting };
+            var result = new Options { Columns = columns, HasHeader = hasHeader, Separators = separatorChars,
+                AllowSparse = allowSparse, AllowQuoting = allowQuoting, TrimWhitespace = trimWhitespace };
             return result;
         }
 
@@ -1462,7 +1464,7 @@ namespace Microsoft.ML.Data
         internal static TextLoader CreateTextReader<TInput>(IHostEnvironment host,
            bool hasHeader = Defaults.HasHeader,
            char separator = Defaults.Separator,
-           bool allowQuotedStrings = Defaults.AllowQuoting,
+           bool allowQuoting = Defaults.AllowQuoting,
            bool supportSparse = Defaults.AllowSparse,
            bool trimWhitespace = Defaults.TrimWhitespace)
         {
@@ -1519,7 +1521,7 @@ namespace Microsoft.ML.Data
             {
                 HasHeader = hasHeader,
                 Separators = new[] { separator },
-                AllowQuoting = allowQuotedStrings,
+                AllowQuoting = allowQuoting,
                 AllowSparse = supportSparse,
                 TrimWhitespace = trimWhitespace,
                 Columns = columns.ToArray()
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1466,7 +1466,8 @@ namespace Microsoft.ML.Data
            char separator = Defaults.Separator,
            bool allowQuoting = Defaults.AllowQuoting,
            bool supportSparse = Defaults.AllowSparse,
-           bool trimWhitespace = Defaults.TrimWhitespace)
+           bool trimWhitespace = Defaults.TrimWhitespace,
+           IMultiStreamSource dataSample = null)
         {
             var userType = typeof(TInput);
 
@@ -1527,7 +1528,7 @@ namespace Microsoft.ML.Data
                 Columns = columns.ToArray()
             };
 
-            return new TextLoader(host, options);
+            return new TextLoader(host, options, dataSample: dataSample);
         }
 
         private sealed class BoundLoader : IDataLoader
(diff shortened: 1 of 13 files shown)


src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -20,15 +20,17 @@ public static class TextLoaderSaverCatalog
...
+        /// <param name="trimWhitespace">Remove trailing whitespace from lines</param>
         /// <param name="dataSample">The optional location of a data sample. The sample can be used to infer column names and number of slots in each column.</param>
         public static TextLoader CreateTextLoader(this DataOperationsCatalog catalog,
             TextLoader.Column[] columns,
             char separatorChar = TextLoader.Defaults.Separator,
             bool hasHeader = TextLoader.Defaults.HasHeader,
             bool allowSparse = TextLoader.Defaults.AllowSparse,
             bool allowQuoting = TextLoader.Defaults.AllowQuoting,
+            bool trimWhitespace = TextLoader.Defaults.TrimWhitespace,
             IMultiStreamSource dataSample = null)
```
TomFinley: The data sample is ***really important***. (See #2663.) If there are feature names, if there is a header, that's where they're coming from. It's really important. Burying it... what... *eight* levels deep seems a bit much. To bury it beyond such things as allowing sparsity and quoting itself seems really tough.

I might say the following order is most sensible:

1. Columns,
2. Separator char,
3. Has header,
4. Data sample,
5. Quoting,
6. Trimming,
7. Sparsity.

I might even think that 5, 6, and 7 could be relegated to advanced options, but no need to insist.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -47,71 +49,78 @@ public static class TextLoaderSaverCatalog
...
         /// <param name="allowQuoting">Whether the input may include quoted values,
         /// which can contain separator characters, colons,
         /// and distinguish empty values from missing values. When true, consecutive separators
         /// denote a missing value and an empty value is denoted by \"\".
         /// When false, consecutive separators denote an empty value.</param>
-        /// <param name="allowSparse">Whether the input may include sparse representations for example,
-        /// if one of the row contains "5 2:6 4:3" that's mean there are 5 columns all zero
-        /// except for 3rd and 5th columns which have values 6 and 3</param>
         /// <param name="trimWhitespace">Remove trailing whitespace from lines</param>
         public static TextLoader CreateTextLoader<TInput>(this DataOperationsCatalog catalog,
```
TomFinley: >CreateTextLoader [](start = 33, length = 16)

This is not consistent. The file is missing altogether, which means it cannot be used to read slot names. Whoops.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1069,17 +1069,19 @@ private bool HasHeader
         /// <param name="allowSparse">Whether the file can contain numerical vectors in sparse format.</param>
         /// <param name="allowQuoting">Whether the content of a column can be parsed from a string starting and ending with quote.</param>
         /// <param name="dataSample">Allows to expose items that can be used for reading.</param>
+        /// <param name="trimWhitespace">Remove trailing whitespace from lines.</param>
         internal TextLoader(IHostEnvironment env, Column[] columns, char separatorChar = Defaults.Separator,
             bool hasHeader = Defaults.HasHeader, bool allowSparse = Defaults.AllowSparse,
-            bool allowQuoting = Defaults.AllowQuoting, IMultiStreamSource dataSample = null)
-            : this(env, MakeArgs(columns, hasHeader, new[] { separatorChar }, allowSparse, allowQuoting), dataSample)
+            bool allowQuoting = Defaults.AllowQuoting, IMultiStreamSource dataSample = null, bool trimWhitespace = Defaults.TrimWhitespace)
```
TomFinley: >IMultiStreamSource dataSample = null [](start = 55, length = 36)

If you wanted you could just get rid of this constructor opportunistically. It serves no purpose as far as I can tell.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -47,71 +49,78 @@ public static class TextLoaderSaverCatalog
...
         /// <param name="separatorChar">Column separator character. Default is '\t'</param>
         /// <param name="hasHeader">Does the file contains header?</param>
+        /// <param name="allowSparse">Whether the input may include sparse representations for example,
+        /// if one of the row contains "5 2:6 4:3" that's mean there are 5 columns all zero
+        /// except for 3rd and 5th columns which have values 6 and 3</param>
         /// <param name="allowQuoting">Whether the input may include quoted values,
         /// which can contain separator characters, colons,
         /// and distinguish empty values from missing values. When true, consecutive separators
         /// denote a missing value and an empty value is denoted by \"\".
         /// When false, consecutive separators denote an empty value.</param>
```
artidoro: If you could make the comments consistent as well it would be great.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
wschin: Ok. I wrongly thought TInput can be used to encode SlotNames.

---
In reply to: [259963380](https://github.com/dotnet/machinelearning/pull/2710#discussion_r259963380) [](ancestors = 259963380)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
wschin: Just got them reordered. Thanks for sharing the list!

---
In reply to: [259963185](https://github.com/dotnet/machinelearning/pull/2710#discussion_r259963185) [](ancestors = 259963185)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
wschin: Sure.

---
In reply to: [259975940](https://github.com/dotnet/machinelearning/pull/2710#discussion_r259975940) [](ancestors = 259975940)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
wschin: Dropped.

---
In reply to: [259963695](https://github.com/dotnet/machinelearning/pull/2710#discussion_r259963695) [](ancestors = 259963695)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
yaeldekel: I think that if hasHeader is true, then the dataSample should not be allowed to be null (because if it is, there can be no slot names). Does it make sense to check this? Another alternative is instead of having a bool argument, to have an `IMultiStreamSource header` instead. But on the other hand, if there are two `IMultiStreamSource`s that might be confusing. Not sure what the right signature is here. @ivanidzo4ka, @TomFinley, what do you think?

---
In reply to: [260017117](https://github.com/dotnet/machinelearning/pull/2710#discussion_r260017117) [](ancestors = 260017117,259963185)

TomFinley: > I think that if hasHeader is true, then the dataSample should not be allowed to be null (because if it is, there can be no slot names). Does it make sense to check this? Another alternative is instead of having a bool argument, to have an `IMultiStreamSource header` instead. But on the other hand, if there are two `IMultiStreamSource`s that might be confusing. Not sure what the right signature is here. @Ivanidzo4ka, @TomFinley, what do you think?
> 
> In reply to: [260017117](https://github.com/dotnet/machinelearning/pull/2710#discussion_r260017117) [](ancestors = 260017117,259963185)

So, let's imagine that I declare that there is a header, but I provide no sample. I think this is fine. I am merely describing *how* to read data, and I am saying that when you read the data with this thing, that you will have a header to contend with. If headers were always only used for defining slot names, I might agree, but sometimes files have headers just because they have headers, and there's no larger reason for it.

The presence of absence of a sample does not mean that we cannot fully define the schema, even if header is true. (This is in contrast to some other options, like, "this is fixed size, but I would like you to auto-determine what that size is, which obviously requires a sample.)

So no check would be my reaction.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -47,79 +61,98 @@ public static class TextLoaderSaverCatalog
         /// <param name="catalog">The <see cref="DataOperationsCatalog"/> catalog.</param>
         /// <param name="separatorChar">Column separator character. Default is '\t'</param>
         /// <param name="hasHeader">Does the file contains header?</param>
```
singlis: >hasHeader [](start = 25, length = 9)

Nit pick - I would rephrase this as "Set to true if the file contains a header row".

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -128,14 +161,19 @@ public static class TextLoaderSaverCatalog
         /// <param name="catalog">The <see cref="DataOperationsCatalog"/> catalog.</param>
         /// <param name="path">Specifies a file from which to read.</param>
         /// <param name="options">Defines the settings of the load operation.</param>
-        public static IDataView ReadFromTextFile(this DataOperationsCatalog catalog, string path, TextLoader.Options options = null)
+        /// <param name="dataSample">The optional location of a data sample. The sample can be used to infer column names and number of slots in each column.</param>
+        public static IDataView ReadFromTextFile(this DataOperationsCatalog catalog, string path,
+            TextLoader.Options options = null, IMultiStreamSource dataSample = null)
```
TomFinley: >, IMultiStreamSource dataSample = null [](start = 45, length = 38)

Hi @wschin, somehow I missed this -- definitely we don't need to have a data sample if we're talking about the path to a data file already.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -47,79 +61,98 @@ public static class TextLoaderSaverCatalog
...
         /// if one of the row contains "5 2:6 4:3" that's mean there are 5 columns all zero
         /// except for 3rd and 5th columns which have values 6 and 3</param>
-        /// <param name="trimWhitespace">Remove trailing whitespace from lines</param>
-        /// <param name="path">The path to the file.</param>
         /// <returns>The data view.</returns>
         public static IDataView ReadFromTextFile<TInput>(this DataOperationsCatalog catalog,
             string path,
             char separatorChar = TextLoader.Defaults.Separator,
             bool hasHeader = TextLoader.Defaults.HasHeader,
+            IMultiStreamSource dataSample = null,
```
TomFinley: >IMultiStreamSource dataSample [](start = 12, length = 29)

Similar here.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -47,79 +61,98 @@ public static class TextLoaderSaverCatalog
...
             bool allowQuoting = TextLoader.Defaults.AllowQuoting,
-            bool allowSparse = TextLoader.Defaults.AllowSparse,
-            bool trimWhitespace = TextLoader.Defaults.TrimWhitespace)
-            => TextLoader.CreateTextReader<TInput>(CatalogUtils.GetEnvironment(catalog), hasHeader, separatorChar, allowQuoting, allowSparse, trimWhitespace);
+            bool trimWhitespace = TextLoader.Defaults.TrimWhitespace,
+            bool allowSparse = TextLoader.Defaults.AllowSparse)
+            => TextLoader.CreateTextReader<TInput>(CatalogUtils.GetEnvironment(catalog), hasHeader, separatorChar, allowQuoting,
+                allowSparse, trimWhitespace, dataSample: dataSample);
```
singlis: >dataSample [](start = 45, length = 10)

why pass the arguments this way? Why not add dataSample after hasHeader?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -47,79 +61,98 @@ public static class TextLoaderSaverCatalog
...
+        /// <param name="path">The path to the file.</param>
         /// <param name="separatorChar">Column separator character. Default is '\t'</param>
+        /// <param name="hasHeader">Does the file contains header?</param>
+        /// <param name="dataSample">The optional location of a data sample. The sample can be used to infer column names and number of slots in each column.</param>
         /// <param name="allowQuoting">Whether the input may include quoted values,
         /// which can contain separator characters, colons,
         /// and distinguish empty values from missing values. When true, consecutive separators
         /// denote a missing value and an empty value is denoted by \"\".
         /// When false, consecutive separators denote an empty value.</param>
+        /// <param name="trimWhitespace">Remove trailing whitespace from lines</param>
```
singlis: minor nit - Add a period.

Issue #2279, opened 2019-01-28
Separate Nuget versioning for "final" and "experimental" nugets


TomFinley:
Right now all our nugets have one global version. Once we ship V1, however, there will be some assemblies and nugets that are "final," in that their API surface is locked and going forward we do not make breaking changes, and other assemblies doing more experimental work where we're not quite so sure what the final surface will look like.

An example of this is `Microsoft.ML.StaticPipe` and friends. The statically typed pipelines work is interesting and we want to continue working on it post v1. Similarly, `Microsoft.ML.Ensemble` seems like something we don't have time to get the API "right" for yet. (Though we might solve that specific one in a different way by just hiding it altogether for now.)

A different example of this would be things like `Microsoft.ML.EntryPoints`, which we have to make public for "reasons," but cannot reasonably be said to have a public API.

Another example would be some sort of "smuggler's cove" assembly/nuget (which does not yet exist) that exposes some infrastructure that some important scenario or other needs access to, but we are not comfortable making as part of our public API. (E.g., some conduit to "get at" the internal methods of the `ComponentCatalog`.)

 @eerhardt opined that if we do that, then of course there should be a separate versioning scheme for those. This would include changes to the package build infrastructure to accomodate this "two tier" structure we do not have yet.

eerhardt:
@maryamariyan will add the build infrastructure in order to have 2 groups of NuGet packages: "stable" and "experimental".

We still need to specify which packages will be "stable", but we can mark those packages after the build infrastructure is there.

Ivanidzo4ka:
So here is list of Stable and Unstable:
Stable:
Microsoft.ML
Microsoft.ML.CpuMath
Microsoft.ML.HalLearners
Microsoft.ML.ImageAnalytcs
Microsoft.ML.LightGBM
Microsoft.ML.Mkl.Redist

Unstable:

Microsoft.ML.DnnImageFeaturizer.AlexNet
Microsoft.ML.DnnImageFeaturizer.ResNet101
Microsoft.ML.DnnImageFeaturizer.ResNet18
Microsoft.ML.DnnImageFeaturizer.ResNet50
Microsoft.ML.EntryPoints
Microsoft.ML.OnnxConverter
Microsoft.ML.OnnxTransformer
Microsoft.ML.Parquet
Microsoft.ML.Recommender
Microsoft.ML.SampleUtils
Microsoft.ML.StaticPipe
Microsoft.ML.TensorFlow
[...]

PR #2718, opened 2019-02-25
Ensemble to it's own package

### LINKED ISSUE #2717: Ensemble project shouldn't be part of Microsoft.ML
From public perspective it's just dll with 0 public classes and none of other projects in Microsoft.ML rely on it.

Ivanidzo4ka:
fix #2717


src/Microsoft.ML.Ensemble/Microsoft.ML.Ensemble.csproj
```diff
@@ -2,7 +2,6 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <IncludeInPackage>Microsoft.ML</IncludeInPackage>
```
eerhardt: Should this assembly go in a different nuget package?

eerhardt: I guess there aren't any public types in this assembly....?

Ivanidzo4ka: Do we have way to mark stable nugets vs experimental?
It's needed by NimbusML, but in same time, it feels weird to create nuget with zero public surface.

eerhardt: We don’t have it yet. But #2279 will enable it. For now let’s make the new package and we can mark it experimental when we have that.

Btw We are creating a Parquet package now, but we don’t ship it.

Ivanidzo4ka:
We discuss and decided to leave it as it is for right now.

eerhardt:
I don't think the way it is right now is correct. Why are we shipping this assembly in the core nuget package?

TomFinley:
> We discuss and decided to leave it as it is for right now.

Sorry, who discussed? Where? I think we need a new nuget to capture this.

eerhardt:
@Ivanidzo4ka - can you just keep this PR to be for `Ensemble`?  We can do the FastTree work separate.  There is also the native library that needs to move out of `Microsoft.ML`. I'd feel more comfortable to do Ensemble first. FastTree separate.

Ivanidzo4ka:
Sure

---
In reply to: [467669272](https://github.com/dotnet/machinelearning/pull/2718#issuecomment-467669272) [](ancestors = 467669272)

Microsoft.ML.sln
```diff
@@ -260,6 +260,12 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Microsoft.Data.DataView", "
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "RemoteExecutorConsoleApp", "test\RemoteExecutorConsoleApp\RemoteExecutorConsoleApp.csproj", "{5E920CAC-5A28-42FB-936E-49C472130953}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Microsoft.ML.Ensemble", "Microsoft.ML.Ensemble", "{AD7058C9-5608-49A8-BE23-58C33A74EE91}"
+	ProjectSection(SolutionItems) = preProject
+		pkg\Microsoft.ML.Ensemble\Microsoft.ML.Ensemble.nupkgproj = pkg\Microsoft.ML.Ensemble\Microsoft.ML.Ensemble.nupkgproj
```
TomFinley: Not aware of us doing this elsewhere, but that's OK I guess.

PR #2722, opened 2019-02-25
Remove and combine Microsoft.ML.UniversalModelFormat.Onnx with Microsoft.ML.Model.OnnxConverter.

### LINKED ISSUE #2721: Remove and combine Microsoft.ML.UniversalModelFormat.Onnx with Microsoft.ML.Model.OnnxConverter
towards #2326

codemzs:
fixes #2721

### DIFF
--- a/src/Microsoft.ML.OnnxConverter/OnnxContextImpl.cs
+++ b/src/Microsoft.ML.OnnxConverter/OnnxContextImpl.cs
@@ -6,7 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Data.DataView;
-using Microsoft.ML.UniversalModelFormat.Onnx;
+using Microsoft.ML.Model.OnnxConverter;
 
 namespace Microsoft.ML.Model.OnnxConverter
 {
--- a/src/Microsoft.ML.OnnxConverter/OnnxExportExtensions.cs
+++ b/src/Microsoft.ML.OnnxConverter/OnnxExportExtensions.cs
@@ -7,7 +7,7 @@ using System.IO;
 using Google.Protobuf;
 using Microsoft.Data.DataView;
 using Microsoft.ML.Model.OnnxConverter;
-using static Microsoft.ML.UniversalModelFormat.Onnx.OnnxCSharpToProtoWrapper;
+using static Microsoft.ML.Model.OnnxConverter.OnnxCSharpToProtoWrapper;
 
 namespace Microsoft.ML
 {
--- a/src/Microsoft.ML.OnnxConverter/OnnxMl.cs
+++ b/src/Microsoft.ML.OnnxConverter/OnnxMl.cs
@@ -8,7 +8,7 @@
 using pb = global::Google.Protobuf;
 using pbc = global::Google.Protobuf.Collections;
 using pbr = global::Google.Protobuf.Reflection;
-namespace Microsoft.ML.UniversalModelFormat.Onnx
+namespace Microsoft.ML.Model.OnnxConverter
 {
     internal class OnnxCSharpToProtoWrapper
     {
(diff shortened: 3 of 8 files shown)


codemzs:
@sfilipi Some of us(@Ivanidzo4ka and @TomFinley ) had a discussion before to not have Microsoft.ML.Onnx namespace since its ambiguous because we have OnnxTransformer and OnnxCoverter now. That conversation let to the removal of Onnx namespace and creation of OnnxCoverter and OnnxTransformer namespace. Microsoft.ML.UniversalModelFormat.Onnx namespace is part of OnnxCoverter hence it needs to come under OnnxCoverter namespace.

TomFinley:
> @sfilipi Some of us(@Ivanidzo4ka and @TomFinley ) had a discussion before to not have Microsoft.ML.Onnx namespace since its ambiguous because we have OnnxTransformer and OnnxCoverter now. That conversation let to the removal of Onnx namespace and creation of OnnxCoverter and OnnxTransformer namespace. Microsoft.ML.UniversalModelFormat.Onnx namespace is part of OnnxCoverter hence it needs to come under OnnxCoverter namespace.

I think though @sfilipi's point is orthogonal. While it may be that I may have said this, I think my point was related to the assembly names and the names of nugets, not the names of the namespaces. Let's imagine a hypothetical world where the namespace was `Microsoft.ML.Onnx`. In a world where both namespaces were used for this, this would not interfere with my point about the need for separate assemblies. Indeed, I think we have defined classes in separate assemblies and nugets in the same *namespace* (the extension methods on `MLContext` being the most obvious example). So let us treat it as established that these are different matters.

I am not saying you are wrong and she is right -- I would need more thought to determine this -- I just think there was a misinterpretation about what I had said, and its scope. So therefore do not use my utterance in that PR or issue to determine what you think is correct or incorrect in this specific context, because I think it is based on a potential misinterpretation of the scope of my words.

TomFinley:
Actually, there's a more germane point that I'd somehow forgotten in my first reply. @sfilipi is as far as I can tell has shown greatest authority in the the namespace reconciliation effort, so we might take her word about what namespace should go where as having more authority in this matter. Not to say she is incontrovertible, any more than I am, but we might defer unless there is a strong reason to not. Which as far as I can tell in this case, does not exist.

codemzs:
@TomFinley Not only did I change the name of the assembly from Microsoft.ML.Onnx to Microsoft.ML.OnnxConverter but I also changed then namespace in that assembly from Microsoft.ML.Model.Onnx to Microsoft.ML.Model.OnnxConverter, here is the [PR ](https://github.com/dotnet/machinelearning/commit/685b0648cf0ce13548269874165d42fa5baf9e9e#diff-795a5a7d8658a0be6c58b031cc46e17f) that you signed-off on. It does not make sense to have a different namespace for a file in a project when its purpose is the same as other files. I strongly feel we should keep the namespace consistent and not introduce a new namespace for OnnxML.cs file whose sole purpose is to create protobuf file.

TomFinley:
Just because I sign off on a PR that touches a file doesn't mean that I am saying that this file cannot be changed forever. Generally, I consider the finalization of the API to be an iterative process. If a PR gets us 90% of the way to where we want to be, I will still generally sign off, because I know the 10% can be done later. (E.g., work around calibrators, and wahtnot.) Regarding this *specifically I did not offer specific comment on the choice of this namespace. I also do not have a strong opinion about namespace names, but I think @sfilipi has been tracking this more carefully. So my point is that referring to me as an authority to deny a renaming request, is what I am trying to say we should not do. :)

For sure we have not insisted elsewhere that namespaces and assemblies need to be isolated. In fact we have considered it a bad idea. (See the usage of `Microsoft.ML` namespace.) Also if we wanted these two to somehow "work together" at some future point (which is entirely reasonable), we'd probably want them to have the same namespace. So I am somehat sympathetic to @sfilipi's argument. What do you think?

---
In reply to: [467264618](https://github.com/dotnet/machinelearning/pull/2722#issuecomment-467264618) [](ancestors = 467264618)

codemzs:
@TomFinley Not sure what you are asking here but thanks for your input. I’ll discuss this offline tomorrow and resolve this issue.

sfilipi:
The reason i proposed `Microsoft.ML.Model.Onnx` a month ago, is because most of the types dealing with Onnx were in that namespace. 

I wrote the comment above  under the impression that @codemzs was creating another namespace, without realizing that `Microsoft.ML.Model.Onnx` had changed to `Microsoft.ML.Model.OnnxConverter`.  

Reading through the thread, and checking on the current state of the solution; the only namespace that groups Onnx converting types, is `Microsoft.ML.Model.OnnxConverter`. The Onnx Transform is in `Microsoft.ML.Transforms`; so IMO having those classes moved to Microsoft.ML.Model.OnnxConverter makes sense, and all those types are internal.

codemzs:
Thanks @sfilipi and @wschin

PR #2738, opened 2019-02-26
Polish GetColumn on IDataView

### LINKED ISSUE #2473: GetColumn method is extension for IDataView but not mlContext.
https://github.com/dotnet/machinelearning/issues/1708 moved CreateEnumerable from extension on top of DataView to become extension on mlContext object.

We still have GetColumn extension https://github.com/dotnet/machinelearning/blob/7fc7e50ce6f8fed24fc0b9528839a0ac8d0ed320/src/Microsoft.ML.Data/Utilities/ColumnCursor.cs#L25
which works on top of DataView. Shall we move it to mlContext?

`var enumerable = mlContext.CreateEnumerable<SamplesUtils.DatasetUtils.SampleTemperatureData>(filteredData, reuseRowObject: true);`
`var originalColumnBack = transformedData_default.GetColumn<VBuffer<ReadOnlyMemory<char>>>(mlContext, defaultColumnName);`

They look silly together.

wschin:
To fix #2473, we need to

- [x] Remove IHostEnviroment from GetColumn's argument list. Note that our strategy is using `Constracts`.
- [x] Replace column name with `DataViewSchema.Column` in GetColumn's argument list.

Review steps:
1. Start with the 2nd iteration because some changes in the 1st iteration got discarded.
2. Removal of IHostEnviroment happens in the 2nd iteration.
3. Uses of `DataViewSchema.Column` are done in the 3rd iteration.

### DIFF
--- a/src/Microsoft.ML.Data/Utilities/ColumnCursor.cs
+++ b/src/Microsoft.ML.Data/Utilities/ColumnCursor.cs
@@ -23,12 +23,11 @@ namespace Microsoft.ML.Data
         /// <param name="columnName">The name of the column to extract.</param>
         public static IEnumerable<T> GetColumn<T>(this IDataView data, string columnName)
         {
-            var env = RetrieveHost(data);
-            env.CheckValue(data, nameof(data));
-            env.CheckNonEmpty(columnName, nameof(columnName));
+            Contracts.CheckValue(data, nameof(data));
+            Contracts.CheckNonEmpty(columnName, nameof(columnName));
 
             if (!data.Schema.TryGetColumnIndex(columnName, out int col))
-                throw env.ExceptSchemaMismatch(nameof(columnName), "input", columnName);
+                throw Contracts.ExceptSchemaMismatch(nameof(columnName), "input", columnName);
 
             // There are two decisions that we make here:
             // - Is the T an array type?
@@ -56,7 +55,7 @@ namespace Microsoft.ML.Data
             {
                 // Output is an array type.
                 if (!(colType is VectorType colVectorType))
-                    throw env.ExceptSchemaMismatch(nameof(columnName), "input", columnName, "vector", "scalar");
+                    throw Contracts.ExceptSchemaMismatch(nameof(columnName), "input", columnName, "vector", "scalar");
                 var elementType = typeof(T).GetElementType();
                 if (elementType == colVectorType.ItemType.RawType)
                 {
@@ -75,24 +74,7 @@ namespace Microsoft.ML.Data
                 }
                 // Fall through to the failure.
             }
-            throw env.Except($"Could not map a data view column '{columnName}' of type {colType} to {typeof(T)}.");
-        }
-
-        /// <summary>
-        /// Return a <see cref="IHost"/> assigned in a given <see cref="IDataView"/> implementation.
-        /// </summary>
-        /// <param name="data">an <see cref="IDataView"/> implementation.</param>
-        private static IHost RetrieveHost(IDataView data)
-        {
-            // Search for the first (if there are multiples) field typed to IHost.
-            var fields = data.GetType().GetFields(System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
-            var fieldInfo = fields.Where((field, index) => field.FieldType == typeof(IHost)).FirstOrDefault();
-
-            // Check if a IHost really gets retrieved.
-            string errorMessage = nameof(data) + " should contains a field of " + typeof(IHost);
-            Contracts.CheckValue(fieldInfo, nameof(data), errorMessage);
-
-            return (IHost)fieldInfo.GetValue(data);
+            throw Contracts.Except($"Could not map a data view column '{columnName}' of type {colType} to {typeof(T)}.");
         }
 
         private static IEnumerable<T> GetColumnDirect<T>(IDataView data, int col)
--- a/src/Microsoft.ML.Data/Utilities/ColumnCursor.cs
+++ b/src/Microsoft.ML.Data/Utilities/ColumnCursor.cs
@@ -26,12 +26,15 @@ namespace Microsoft.ML.Data
             Contracts.CheckValue(data, nameof(data));
             Contracts.CheckNonEmpty(column.Name, nameof(column));
 
-            if (!data.Schema.TryGetColumnIndex(column.Name, out int colIndex))
-                throw Contracts.ExceptParam(nameof(column), string.Format("column name {0} cannot be found in {1}", column.Name, nameof(data)));
+            var colIndex = column.Index;
+            var colType = column.Type;
+            var colName = column.Name;
 
-            if (data.Schema[colIndex].Type != column.Type)
-                throw Contracts.ExceptParam(nameof(column), string.Format("column {0}'s type {1} doesn't match the expected type {2} in {3}",
-                    column.Name, column.Type, data.Schema[colIndex].Type, nameof(data)));
+            // Use column index as the principle address of the specified input column and check if that address in data contains
+            // the column indicated.
+            if (data.Schema[colIndex].Name != colName || data.Schema[colIndex].Type != colType)
+                throw Contracts.ExceptParam(nameof(column), string.Format("column with name {0}, type {1}, and index {2} cannot be found in {3}",
+                    colName, colType, colIndex, nameof(data)));
 
             // There are two decisions that we make here:
             // - Is the T an array type?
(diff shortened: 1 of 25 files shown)


src/Microsoft.ML.Data/Utilities/ColumnCursor.cs
```diff
@@ -79,6 +78,23 @@ public static IEnumerable<T> GetColumn<T>(this IDataView data, IHostEnvironment
             throw env.Except($"Could not map a data view column '{columnName}' of type {colType} to {typeof(T)}.");
         }
 
+        /// <summary>
+        /// Return a <see cref="IHost"/> assigned in a given <see cref="IDataView"/> implementation.
+        /// </summary>
+        /// <param name="data">an <see cref="IDataView"/> implementation.</param>
+        private static IHost RetrieveHost(IDataView data)
```
Ivanidzo4ka: >RetrieveHost [](start = 29, length = 12)

No.
Either IDataView will have IHost property/field as part of interface, or we don't do it at all.
This reflection magic will only work if you have IHost in IDataView, and there is no guarantee of that.

wschin: Fixed. We use `Constracts` now.

---
In reply to: [260409959](https://github.com/dotnet/machinelearning/pull/2738#discussion_r260409959) [](ancestors = 260409959)

src/Microsoft.ML.Data/Utilities/ColumnCursor.cs
```diff
@@ -20,16 +20,18 @@ public static class ColumnCursorExtensions
...
         {
-            Contracts.CheckValue(env, nameof(env));
-            env.CheckValue(data, nameof(data));
-            env.CheckNonEmpty(columnName, nameof(columnName));
+            Contracts.CheckValue(data, nameof(data));
+            Contracts.CheckNonEmpty(column.Name, nameof(column));
 
-            if (!data.Schema.TryGetColumnIndex(columnName, out int col))
-                throw env.ExceptSchemaMismatch(nameof(columnName), "input", columnName);
+            if (!data.Schema.TryGetColumnIndex(column.Name, out int colIndex))
```
Ivanidzo4ka: >if (!data.Schema.TryGetColumnIndex(column.Name, out int colIndex) [](start = 12, length = 65)

With this implementation I don't see any reason to have `DataViewSchema.Column column` instead of just `string columnName`.

So imagine I'm doing following:
I have column `A` of type string. I modify that column to `vector<float>` with same name `A`.
Then I apply ToKey to `A`, but preserve name.

In old code, I can say: give me `A` and i would get key A.
with your code, I can inspect `IDataview` and I can find 3 `A` columns, two are hidden and one is not.
I pick `A` which is `vector<float>` and pass to your function.

I would get exception, right?
What's the point of this new function, if it does same thing, but in more obfuscated way by forcing me to get `DataViewSchema.Column` first instead of just name?

wschin: How about this?
```csharp
            var colIndex = column.Index;
            var colType = column.Type;
            var colName = column.Name;

            // Use column index as the principle address of the specified input column and check if that address in data contains
            // the column indicated.
            if (data.Schema[colIndex].Name != colName || data.Schema[colIndex].Type != colType)
                throw Contracts.ExceptParam(nameof(column), string.Format("column with name {0}, type {1}, and index {2} cannot be found in {3}",
                    colName, colType, colIndex, nameof(data)));
```
We use input `column.Index` to retrieve information from `data`.

test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamples.cs
```diff
@@ -79,7 +79,7 @@ private void IntermediateData(string dataPath)
             // The same extension method also applies to the dynamic-typed data, except you have to
             // specify the column name and type:
             var dynamicData = transformedData.AsDynamic;
-            var sameFeatureColumns = dynamicData.GetColumn<string[]>(mlContext, "AllFeatures")
+            var sameFeatureColumns = dynamicData.GetColumn<string[]>(dynamicData.Schema["AllFeatures"])
```
Ivanidzo4ka: don't forget to update cookbook! both of them!

src/Microsoft.ML.Data/Utilities/ColumnCursor.cs
Ivanidzo4ka: Looks good.

---
In reply to: [260491233](https://github.com/dotnet/machinelearning/pull/2738#discussion_r260491233) [](ancestors = 260491233)

src/Microsoft.ML.Data/Utilities/ColumnCursor.cs
```diff
@@ -20,16 +20,21 @@ public static class ColumnCursorExtensions
         /// </summary>
         /// <typeparam name="T">The type of the values. This must match the actual column type.</typeparam>
         /// <param name="data">The data view to get the column from.</param>
-        /// <param name="env">The current host environment.</param>
-        /// <param name="columnName">The name of the column to extract.</param>
-        public static IEnumerable<T> GetColumn<T>(this IDataView data, IHostEnvironment env, string columnName)
+        /// <param name="column">The column to be extracted.</param>
+        public static IEnumerable<T> GetColumn<T>(this IDataView data, DataViewSchema.Column column)
```
Ivanidzo4ka: >column [](start = 93, length = 6)

I would prefer to have string version of this method as well, but I have impression Tom doesn't like that idea (not sure why).
No blocker, just fishing for explanation.

test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamples.cs
wschin: Sure.

src/Microsoft.ML.Data/Utilities/ColumnCursor.cs
wschin: Maybe for consistency with static APIs?

src/Microsoft.ML.Data/Utilities/ColumnCursor.cs
```diff
@@ -20,16 +20,21 @@ public static class ColumnCursorExtensions
...
+            Contracts.CheckValue(data, nameof(data));
+            Contracts.CheckNonEmpty(column.Name, nameof(column));
 
-            if (!data.Schema.TryGetColumnIndex(columnName, out int col))
-                throw env.ExceptSchemaMismatch(nameof(columnName), "input", columnName);
+            var colIndex = column.Index;
+            var colType = column.Type;
+            var colName = column.Name;
+
+            // Use column index as the principle address of the specified input column and check if that address in data contains
```
abgoswam: I believe this logic is to make sure the `column`  (type DataViewSchema.Column)  is indeed from the same `data`  (type IDataView). Can we add a test that exercises this ?  Perhaps by passing a `column` from a different dataview, and making sure it fails.

I see some tests which exercise the type (given the *same* dataview), but none which exercises the new logic added above

src/Microsoft.ML.Data/Utilities/ColumnCursor.cs
TomFinley: > I would prefer to have string version of this method as well, but I have impression Tom doesn't like that idea (not sure why).

I think this is how religious doctrine can get established in some cases. :)

But seriously, I think the *core* interfaces of `IDataView` itself and its accessors ought to work over `Schema.Column`. It's just safer. A convenience *like this* however, surely benefits from a `string`. I think it's quite fine to have both; in the one case, you get some degree of "safety." In the other, convenience. This is clearly a case where convenience *is the intent*, and insisting on a column is inconvenient.

---
In reply to: [260585414](https://github.com/dotnet/machinelearning/pull/2738#discussion_r260585414) [](ancestors = 260585414)

src/Microsoft.ML.Data/Utilities/ColumnCursor.cs
wschin: Added in Iteration 6.

---
In reply to: [260851354](https://github.com/dotnet/machinelearning/pull/2738#discussion_r260851354) [](ancestors = 260851354)

src/Microsoft.ML.Data/Utilities/ColumnCursor.cs
wschin: String overloading is added in Iteration 7.

---
In reply to: [260860073](https://github.com/dotnet/machinelearning/pull/2738#discussion_r260860073) [](ancestors = 260860073,260585414)

PR #2740, opened 2019-02-26
Microsoft.ML.Internal.Internallearn should be hidden/moved/renamed

### LINKED ISSUE #2714: Microsoft.ML.Internal.Internallearn should be hidden/moved/renamed.
towards #2326

codemzs:
fixes #2714


src/Microsoft.ML.Data/Training/EarlyStoppingCriteria.cs
```diff
@@ -21,7 +21,7 @@
 [assembly: EntryPointModule(typeof(PQEarlyStoppingCriterion))]
 [assembly: EntryPointModule(typeof(UPEarlyStoppingCriterion))]
 
-namespace Microsoft.ML.Internal.Internallearn
+namespace Microsoft.ML.Trainers.FastTree
```
sfilipi: >namespace Microsoft.ML.Trainers [](start = 0, length = 31)

would this be better under just `Microsoft.ML.Trainers`, since more than FastTree might use it. Does LightGBM uses its?

codemzs: I believe it’s just FastTree

TomFinley: So one thing that I am not seeing yet, is I do not see that this has moved to FastTree? Is that coming in a future commit? If we're saying something is useful only to one assembly even to the point where we are changing its namespace, why not just move that to that assembly outright?

---
In reply to: [260433610](https://github.com/dotnet/machinelearning/pull/2740#discussion_r260433610) [](ancestors = 260433610)

TomFinley:
>     public interface IEarlyStoppingCriterion

Probably should not be an interface.

---
Refers to: src/Microsoft.ML.Data/Training/EarlyStoppingCriteria.cs:29 in 07eb860. [](commit_id = 07eb860cb445d683e8420f432c99607bce1bff8a, deletion_comment = False)

src/Microsoft.ML.Data/Training/EarlyStoppingCriteria.cs
TomFinley: Also double check that this is only being used in FastTree also on internal code. That may change our thinking here, just something to double check.

---
In reply to: [260449769](https://github.com/dotnet/machinelearning/pull/2740#discussion_r260449769) [](ancestors = 260449769,260433610)

singlis:
Looks good pending the feedback from Tom. I agree if that is only used by FastTree (which looks to be only referenced in the BoostedTreeOptions), then I would move that code to the FastTree assembly.

src/Microsoft.ML.Data/Training/EarlyStoppingCriteria.cs
codemzs: Seems to also being used in NeuralNetworks but as we discussed offline we probably don't care for now. I have moved this file in fast tree project.

codemzs:
>     public interface IEarlyStoppingCriterion

Done.

---
In reply to: [467578760](https://github.com/dotnet/machinelearning/pull/2740#issuecomment-467578760) [](ancestors = 467578760)

---
Refers to: src/Microsoft.ML.Data/Training/EarlyStoppingCriteria.cs:29 in 07eb860. [](commit_id = 07eb860cb445d683e8420f432c99607bce1bff8a, deletion_comment = False)

codemzs:
thanks for reviewing. I have addressed Tom's comments.

---
In reply to: [467632167](https://github.com/dotnet/machinelearning/pull/2740#issuecomment-467632167) [](ancestors = 467632167)

Issue #2652, opened 2019-02-20
LightGbm ranking with Options does not set GroupId column even when specified in Options


najeeb-kazmi:
During the course of Fixing #2530 in PR #2650, I discovered that when creating a `LightGbmRankingTrainer` with `MLContext.Ranking.Trainers.LightGbm(Options options)`, the GroupId column is never set even though it is specified in the `Options`.

Digging a bit deeper, I found that in [LightGbmTrainerBase.cs](https://github.com/dotnet/machinelearning/blob/4a71e501f09846974e550f817f3f8f1243054c1b/src/Microsoft.ML.LightGBM/LightGbmTrainerBase.cs#L91-L92) we have
```csharp
        private protected LightGbmTrainerBase(IHostEnvironment env, string name, Options options, SchemaShape.Column label)
           : base(Contracts.CheckRef(env, nameof(env)).Register(name), TrainerUtils.MakeR4VecFeature(options.FeatureColumn), label, TrainerUtils.MakeR4ScalarWeightColumn(options.WeightColumn))
```

This does not use options.GroupIdColumn to create a `SchemaShape.Column` for GroupId to pass to the base constructor `TrainerEstimatorBaseWithGroupId`. Neither is there a method in `TrainerUtils` to create such a `SchemaShape.Column` object from options.GroupIdColumn for Key types such as GroupId.

Repro:
Run the sample in `Microsoft.ML.Samples.Dynamic.LightGbmRankingWithOptions.Example()`.

Unhandled Exception: System.ArgumentOutOfRangeException: Need a group column.
Parameter name: data
   at Microsoft.ML.Contracts.CheckParam(IExceptionContext ctx, Boolean f, String paramName, String msg) in C:\najeeb-kazmi\machinelearning\src\Microsoft.ML.Core\Utilities\Contracts.cs:line 543
   at Microsoft.ML.LightGBM.LightGbmRankingTrainer.CheckDataValid(IChannel ch, RoleMappedData data) in C:\najeeb-kazmi\machinelearning\src\Microsoft.ML.LightGBM\LightGbmRankingTrainer.cs:line 127
   at Microsoft.ML.LightGBM.LightGbmTrainerBase`3.LoadTrainingData(IChannel ch, RoleMappedData trainData, CategoricalMetaData& catMetaData) in C:\najeeb-kazmi\machinelearning\src\Microsoft.ML.LightGBM\LightGbmTrainerBase.cs:line 313
[...]

PR #2742, opened 2019-02-26
Respect group id in options for ranking task

Ivanidzo4ka:
Fix https://github.com/dotnet/machinelearning/issues/2652

### DIFF
--- a/src/Microsoft.ML.LightGBM/LightGbmTrainerBase.cs
+++ b/src/Microsoft.ML.LightGBM/LightGbmTrainerBase.cs
@@ -84,7 +84,8 @@ namespace Microsoft.ML.LightGBM
         }
 
         private protected LightGbmTrainerBase(IHostEnvironment env, string name, Options options, SchemaShape.Column label)
-           : base(Contracts.CheckRef(env, nameof(env)).Register(name), TrainerUtils.MakeR4VecFeature(options.FeatureColumn), label, TrainerUtils.MakeR4ScalarWeightColumn(options.WeightColumn))
+           : base(Contracts.CheckRef(env, nameof(env)).Register(name), TrainerUtils.MakeR4VecFeature(options.FeatureColumn), label,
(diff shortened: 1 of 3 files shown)


test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
```diff
@@ -115,7 +115,8 @@ public void FastTreeRankerEstimator()
                 new FastTreeRankingTrainer.Options
                 {
                     FeatureColumn = "NumericFeatures",
-                    NumTrees = 10
+                    NumTrees = 10,
+                    GroupIdColumn = "Group"
```
singlis: >GroupIdColumn [](start = 20, length = 13)

I feel like this should be "GroupColumn" to fall more inline with FeatureColumn and LabelColumn.

src/Microsoft.ML.LightGBM/LightGbmTrainerBase.cs
```diff
@@ -163,7 +164,7 @@ private protected virtual void CheckDataValid(IChannel ch, RoleMappedData data)
             ch.CheckParam(data.Schema.Label.HasValue, nameof(data), "Need a label column");
         }
 
-        protected virtual void GetDefaultParameters(IChannel ch, int numRow, bool hasCategarical, int totalCats, bool hiddenMsg=false)
+        protected virtual void GetDefaultParameters(IChannel ch, int numRow, bool hasCategarical, int totalCats, bool hiddenMsg = false)
```
singlis: >hasCategarical [](start = 82, length = 14)

misspelling -- should be hasCategorical.

singlis: >hiddenMsg [](start = 118, length = 9)

for the sake of not having short names -hiddenMessage?

Ivanidzo4ka: I'll leave that stuff to whoever would scrub lightgbm learners.

PR #2744, opened 2019-02-26
Make EarlyStoppingMetric an enum.

### LINKED ISSUE #2521: FastTree EarlyStoppingMetric is an `int` but only accepts specific values
In the current `Options` for `FastTree` (and `GAMs`), `EarlyStoppingMetric` is an `int`, but only specific values are accepted (e.g. 1 or 2 for regression, 1 or 3 for ranking).

It would be much cleaner if this was an enum.

codemzs:
fixes #2521

### DIFF
--- a/src/Microsoft.ML.FastTree/FastTreeArguments.cs
+++ b/src/Microsoft.ML.FastTree/FastTreeArguments.cs
@@ -44,7 +44,7 @@ namespace Microsoft.ML.Trainers.FastTree
         {
             public Options()
             {
-                EarlyStoppingMetrics = 1; // Use L1 by default.
+                EarlyStoppingMetrics = EarlyStoppingMetrics.RegressionL1; // Use L1 by default.
             }
 
             ITrainer IComponentFactory<ITrainer>.CreateComponent(IHostEnvironment env) => new FastTreeRegressionTrainer(env, this);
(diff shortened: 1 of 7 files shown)


src/Microsoft.ML.FastTree/FastTreeArguments.cs
```diff
@@ -442,6 +442,17 @@ internal virtual void Check(IExceptionContext ectx)
...
 
+    /// <summary>
+    /// Early stopping metrics for Regression and Ranking tasks in <see cref="FastTree"/>
+    /// </summary>
+    public enum EarlyStoppingMetrics
+    {
+        RegressionL1 = 1,
+        RegressionL2 = 2,
+        RankingNdcg1 = 1,
+        RankingNdcg3 = 3
```
TomFinley: Please have separate enums for each task. The argument subclasses for each task will have as a public property something that takes that as its input. The existing argument can be then made internal. I am also pretty curious what the state is for binary classification.

Note that given the current state of the code, assigning `RankingNdcg1` to a regression task is totally legal, and it will happen to just map to L1, which is surely nonsense. The goal of #2521 is about introducing an enum, it is sure, but the udnerlying point really is to have an API that makes some degree of sense.

rogancarr: Ditto what Tom said. I'd like the API to be like `TaskEarlyStoppingMetrics.Metric`.

Also about Binary Classification: When I was playing around with the API, the early stopping for Binary Classification wasn't clear. It is possible we have hard-coded the stopping metric we use.

rogancarr:
Thanks for taking a look at this!

TomFinley:
Is this still in progress @codemzs ?

codemzs:
@TomFinley @wschin has done this part of his FT scrub it seems based on what @Ivanidzo4ka told me yesterday.

PR #2745, opened 2019-02-26
Remove IMultiStreamSource when path (type: string) exists in text loader APIs

wschin:
Fix bugs introduced in #2710. If `string path` and `IMultiStreamSource dataExample` are present at the same time, we drop `IMultiStreamSource dataExample`.


artidoro:
>         public static TextLoader CreateTextLoader(this DataOperationsCatalog catalog,

Seems inconsistent to have `CreateTextLoader` accept `IMultiStreamSource dataSample` as an input while `ReadFromTextFile` accepts a `string path`. 
Why not `string path` in `CreateTextLoader`?

---
Refers to: src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs:25 in 8b127a3. [](commit_id = 8b127a353749cdb56e4561782b173d674d459e14, deletion_comment = False)

TomFinley:
Thank you @wschin !

Issue #1669, opened 2018-11-19
Float v.s float


wschin:
In ML.NET codebase, one can see two different names for 32-bit float type. Recent PR uses `float` while [old places](https://github.com/dotnet/machinelearning/blob/e0906608ca0d37ad0c10ba647e83b45990afde43/src/Microsoft.ML.Sweeper/ISweeper.cs#L5) have `Float`. It'd be nice to have only one of the two styles.

Ivanidzo4ka:
@wschin we are trying to get rid of `Float` in our code, and if you look into estimators PRs we almost always delete that `using Float = System.Single;` line and use `float` instead

TomFinley:
Just some history, in the long vanished past we used to have two builds, a "double" and a "single" build. (This was in the days before `IDataView` where pipelines could accommodate either.) But, when we removed the compiler switch, it was beneficial to keep the declaration around: by doing so you only had a couple thousand lines changed, instead of tens of thousands of lines changed.

We've been opportunistically removing `Float` though, though in a very very piecemeal fashion. Doing it all at once though -- well, it might be a good idea, but that would be one heck of a PR to manage conflicts on. 😛

Zruty0:
Yes, just recently I saw someone mass-renaming `Single` to `float`, and ending up with `Ldafloatbox` somewhere in comments :)

glebuk:
Seems to be fixed already

Ivanidzo4ka:
Not exactly finished.
https://github.com/dotnet/machinelearning/blob/a570da14a41f2870eb8f61d84496a58422398253/src/Microsoft.ML.StandardLearners/Optimizer/LineSearch.cs#L8
We have still plenty of places where we have `using Float = System.Single;` pattern.

In many places we clean it up via different unrelated PRs, but some code haven't been touch since we merge it from internal repo.

sfilipi:
The action item on this PR is to remove

`using Float = System.Single` from all the files, and through the codebase, replace `Float`, with `float`.

PR #2754, opened 2019-02-27
Replace Float with float

jwood803:
Fix for #1669.


src/Microsoft.ML.Sweeper/Parameters.cs
```diff
@@ -11,12 +11,11 @@
 using Microsoft.ML.CommandLine;
 using Microsoft.ML.Internal.Utilities;
 using Microsoft.ML.Sweeper;
-using Float = System.Single;
 
 [assembly: LoadableClass(typeof(LongValueGenerator), typeof(LongParamOptions), typeof(SignatureSweeperParameter),
     "Long parameter", "lp")]
 [assembly: LoadableClass(typeof(FloatValueGenerator), typeof(FloatParamOptions), typeof(SignatureSweeperParameter),
-    "Float parameter", "fp")]
+    "float parameter", "fp")]
```
TomFinley: >float [](start = 5, length = 5)

I would probably desist from changing this particular one.

src/Microsoft.ML.Sweeper/Parameters.cs
```diff
@@ -555,7 +554,7 @@ public bool TryParseParameter(string paramValue, Type paramType, string paramNam
 
             // Extract the minimum, and the maximum value of the list of suggested sweeps.
             // Positive lookahead splitting at the '-' character.
-            // It is used for the Float and Long param types.
+            // It is used for the float and Long param types.
```
TomFinley: >float [](start = 34, length = 5)

I think this is referring specifically to the `FloatParamOptions`, if you want to downcase it that's fine, but we should probably do the same with `long` here.

src/Microsoft.ML.Sweeper/Parameters.cs
jwood803: Good catch! I thought it was in reference to the `Float` type parameter that we were getting rid of.

PR #2766, opened 2019-02-27
Calibrators catalog introduction

Ivanidzo4ka:
should fix https://github.com/dotnet/machinelearning/issues/1871


src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -367,6 +368,20 @@ internal BinaryClassificationTrainers(BinaryClassificationCatalog catalog)
             return result.Select(x => new CrossValidationResult<CalibratedBinaryClassificationMetrics>(x.Model,
                 Evaluate(x.Scores, labelColumn), x.Scores, x.Fold)).ToArray();
         }
+
+        /// <summary>
+        /// The list of trainers for performing binary classification.
+        /// </summary>
+        public CalibratorsSubCatalog Calibrators { get; }
+
+        public sealed class CalibratorsSubCatalog : CatalogInstantiatorBase
```
TomFinley: >CalibratorsSubCatalog [](start = 28, length = 21)

If I look at other classes derived from `CatalogInstantiatorBase`, I don't see anything like `SubCatalog`. How about just `CalibratorsCatalog`? Why do we need the sub in there?

Don't forget to like and sub.

src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs
```diff
@@ -425,26 +419,129 @@ public sealed class PavCalibratorEstimator : CalibratorEstimatorBase<PavCalibrat
...
+        private PairAdjacentViolatorsCalibratorTransformer(IHostEnvironment env, ModelLoadContext ctx)
             : base(env, ctx, LoadName)
         {
         }
     }
 }
+
+namespace Microsoft.ML
+{
+    public static class CalibratorsCatalog
```
TomFinley: >CalibratorsCatalog [](start = 24, length = 18)

So, maybe we can just have `CalibratorsCatalog` be a class, with actual methods, that is, not extension methods, given that it's in the same assembly. We sometimes do extension methods, but we often also put methods directly on the thing when the "scope" of what is to be done is fairly limited, or there isn't some great question of broader consistency, or somesuch.

Ivanidzo4ka: by saying that you mean something like `mlContext.Calirbrators.Platt()` or you mean `CalibratorsCatalog.Platt(MlContext context, ...)`
Second feels outlier to way how we currently handle everything else.

TomFinley: No. I am saying that these should be actual methods on the catalog class itself.

TomFinley: Let me be more explicit. I could write the following code:

```chsarp
public class Foo { }

public static class FooExtensions
{
    public static void Bar(this Foo foo, ...) { ... }
    public static void Biz(this Foo foo, ...) { ... }
    public static void Blam(this Foo foo, ...) { ... }
}
```

As in fact we do when we want to have separation between multiple things. But, in this particular case, *everything* we'd want to do with this catalog lives in the same assembly where the assembly is. So I could have written the following.

```chsarp
public class Foo
{
    public void Bar(...) { ... }
    public void Biz(...) { ... }
    public void Blam(...) { ... }
}
```

I would argue the second is easier and more clear. I understand the reasons why we sometimes phrase these catalog operations as extension methods, what I'm saying is, in this case, it is not necessary or helpful to do so. If you disagree that's fine I guess, it just seems strange.

Ivanidzo4ka: Thank you. I'm sorry you had to write down this explanation for me. I've been slightly tired lately and with all other distractions it looks like I lost ability to focus and understand other people.

TomFinley: Ha ha, that's fine Ivan. I remember yesterday when I was so exhausted and sick I confused the heck out of people with what were (if I look back on them) some insane ramblings.

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -366,6 +368,99 @@ internal BinaryClassificationTrainers(BinaryClassificationCatalog catalog)
             return result.Select(x => new CrossValidationResult<CalibratedBinaryClassificationMetrics>(x.Model,
                 Evaluate(x.Scores, labelColumn), x.Scores, x.Fold)).ToArray();
         }
+
+        /// <summary>
+        /// The list of trainers for performing binary classification.
+        /// </summary>
+        public CalibratorsCatalog Calibrators { get; }
+
+        public sealed class CalibratorsCatalog : CatalogInstantiatorBase
```
sfilipi: >CalibratorsCatalog [](start = 28, length = 18)

i need xml doc too.

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -366,6 +368,99 @@ internal BinaryClassificationTrainers(BinaryClassificationCatalog catalog)
...
+        {
+            internal CalibratorsCatalog(BinaryClassificationCatalog catalog)
+                : base(catalog)
+            {
+            }
+            /// <summary>
+            /// Adds probability column by training naive binning-based calbirator.
```
sfilipi: >calbirator [](start = 72, length = 10)

spelling here and below.

sfilipi: >naive binning-based [](start = 52, length = 19)

can we link to any online resources? @yaeldekel, does this apply: [https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4410090/](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4410090/)

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -366,6 +368,99 @@ internal BinaryClassificationTrainers(BinaryClassificationCatalog catalog)
...
+            /// </format>
+            /// </example>
+            public NaiveCalibratorEstimator Naive(
+                  string labelColumnName = DefaultColumnNames.Label,
+                  string scoreColumnName = DefaultColumnNames.Score)
+            {
+                return new NaiveCalibratorEstimator(Owner.Environment, labelColumnName, scoreColumnName);
+            }
+            /// <summary>
+            /// Adds probability column by training platt calbirator.
```
sfilipi: >training platt [](start = 43, length = 14)

[https://en.wikipedia.org/wiki/Platt_scaling](https://en.wikipedia.org/wiki/Platt_scaling) link

docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/BinaryClassification/Calibrators/Platt.cs
```diff
@@ -0,0 +1,71 @@
...
+            // Score   4.18144
+            // Score  -14.10248
+            // Score   2.731951
+            // Score  -2.554229
+            // Score   5.36571
+
+            // Let's train a calibrator estimator on this scored dataset. The trained calibrator estimator produces a transformer
+            // that can transform the scored data by adding a new column names "Probability". 
+            var calibratorEstimator = mlContext.BinaryClassification.Calibrators.Platt();
+            var calibratorTransformer = calibratorEstimator.Fit(scoredData);
```
sfilipi: >scoredData [](start = 64, length = 10)

we should probably train the calibrator on another dataset, not the testSet, in those samples. So have training set, testing set, and a calibrator training set

docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/BinaryClassification/Calibrators/PairAdjacentViolators.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System;
+using System.Linq;
+
+namespace Microsoft.ML.Samples.Dynamic.Trainers.BinaryClassification.Calibrators
```
sfilipi: >Microsoft.ML.Samples.Dynamic.Trainers.BinaryClassification.Calibrators [](start = 10, length = 70)

please Microsoft.ML.Samples.Dynamic 

no need to be too fancy

src/Microsoft.ML.Data/TrainCatalog.cs
TomFinley: I doubt they're the same. That article seems to have been published in 2015, whereas this calibrator has AFAIK existed at least since 2013, and probably goes back a fair bit further than that. I imagine Misha didn't just conjure up the method out of absolute thin air, but whatever work he based the implementation on has probably be lost to the sands of time.

---
In reply to: [261052618](https://github.com/dotnet/machinelearning/pull/2766#discussion_r261052618) [](ancestors = 261052618)

docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/BinaryClassification/Calibrators/PairAdjacentViolators.cs
Ivanidzo4ka: I don't know. I told same thing to @shmoradims and he told me I'm silly. I don't want be part of any fights regarding samples and how we organize them. So I would just follow what he is doing.

---
In reply to: [261054779](https://github.com/dotnet/machinelearning/pull/2766#discussion_r261054779) [](ancestors = 261054779)

src/Microsoft.ML.Data/TrainCatalog.cs
Ivanidzo4ka: To be honest none of other `CatalogInstantiatorBase` inherited classes has one. And I don't plan to fix it right now. Documentation can wait till next month.

---
In reply to: [261052357](https://github.com/dotnet/machinelearning/pull/2766#discussion_r261052357) [](ancestors = 261052357)

Ivanidzo4ka:
@TomFinley  would be nice if you can take a look one more time, thx!

src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs
```diff
@@ -425,24 +419,24 @@ public sealed class PavCalibratorEstimator : CalibratorEstimatorBase<PavCalibrat
...
         [BestFriend]
         private protected override CalibratorTransformer<PavCalibrator> Create(IHostEnvironment env, PavCalibrator calibrator)
-            => new PavCalibratorTransformer(env, calibrator);
+            => new PairAdjacentViolatorsCalibratorTransformer(env, calibrator);
 
     }
 
     /// <summary>
-    /// The <see cref="ITransformer"/> implementation obtained by training a <see cref="PavCalibratorEstimator"/>
+    /// The <see cref="ITransformer"/> implementation obtained by training a <see cref="PoolAdjacentViolatorsCalibratorEstimator"/>
```
TomFinley: >PoolAdjacentViolatorsCalibratorEstimator [](start = 88, length = 40)

What is the difference between `PoolAdjacentViolators` and `PairAdjacentViolators`? They both seem to be an expansion of the single acronym that used to just be PAV, so I'm wondering if this distinction is on purpose.

TomFinley:
>     public sealed class PavCalibrator : ICalibrator, ICanSaveInBinaryFormat

So, I see we have this PAV calibrator that are I guess the model parameters from pair-adjacent calibrator transformers which are in turn trained from pool adjacent calibrator estimators? Right? I kind of feel like consistency here would be less confusing.

---
Refers to: src/Microsoft.ML.Data/Prediction/Calibrator.cs:1801 in 553aeb1. [](commit_id = 553aeb126128daab8ebd6cb9f46e97bef6491fdf, deletion_comment = False)

src/Microsoft.ML.Data/Prediction/CalibratorCatalog.cs
Ivanidzo4ka: Changed to Isotonic to avert confusion

---
In reply to: [262238146](https://github.com/dotnet/machinelearning/pull/2766#discussion_r262238146) [](ancestors = 262238146)

PR #2777, opened 2019-02-27
Correct documentation for MapKeyToVector.

### LINKED ISSUE #2772: KeyToVectorMappingEstimator catalog extension method summary is wrong
This documentation appears to be incorrect.

https://github.com/dotnet/machinelearning/blob/a0edc5c5724ed2895fd0c2fa5c2525d03de13455/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs#L93-L97

This is for taking a key value (or key values) and mapping it into a floating point vector suitable for learning (variously, one-hot-encoding, or bag-of-words vector, depending on the type of input data and other applications). But to say it "converts it back to its original vector" is in no way accurate.

Perhaps someone got a little excited copying and pasting from the `KeyToValue` methods above.

codemzs:
fixes #2772

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
+++ b/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
@@ -90,8 +90,7 @@ namespace Microsoft.ML
              => new KeyToValueMappingEstimator(CatalogUtils.GetEnvironment(catalog), ColumnOptions.ConvertToValueTuples(columns));
 
         /// <summary>
-        /// Maps key types (or key values) into a floating point vector suitable for learning (variously, one-hot-encoding,
-        /// or bag-of-words vector, depending on the type of input data and other applications)
+        /// Maps key types or key values into a floating point vector.
         /// </summary>
         /// <param name="catalog">The categorical transform's catalog.</param>
         /// <param name="columns">The input column to map back to vectors.</param>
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
```diff
@@ -90,7 +90,8 @@ public static KeyToValueMappingEstimator MapKeyToValue(this TransformsCatalog.Co
              => new KeyToValueMappingEstimator(CatalogUtils.GetEnvironment(catalog), ColumnOptions.ConvertToValueTuples(columns));
 
         /// <summary>
-        /// Convert the key types back to their original vectors.
+        /// Maps key types (or key values) into a floating point vector suitable for learning (variously, one-hot-encoding,
```
TomFinley: >variously [](start = 95, length = 9)

While I did use this language in the issue, it is not suitable language for documentation. The way in which I write is not generally suitable language for documentation.

codemzs: Oh come'on you are the [Shakespeare ](https://en.wikipedia.org/wiki/William_Shakespeare) of our team ... I'll change the language to be more ideal.

codemzs:
@TomFinley Can you please take another look?

Issue #2333, opened 2019-01-30
KeyType missing documentation for a few methods/parameters


rogancarr:
The following properties of KeyType are missing documentation
- Contiguous
- Equals(ColumnType)
- Equals(Object)
- GetHashCode()
- ToString()

TomFinley:
Contiguous should have been gone. Indeed I am surprised to see that it is still there. Maybe @artidoro has more info.

The others, I wonder if there's something like inheritdoc or something for implementations of common methods and overrides from stuff already in .NET standard library.

artidoro:
It's gone in the current version of the code. 
However, if you are looking at the MDocs it will probably still be there at least until we update it with the new nuget.

PR #2783, opened 2019-02-27
Add missing KeyType documentation

jwood803:
Fix for #2333.

Hopefully, this is a good start. I'm sure some wording could be updated. :)


src/Microsoft.ML.Core/Data/KeyType.cs
```diff
@@ -59,6 +59,13 @@ public static bool IsValidDataType(Type type)
         /// </summary>
         public ulong Count { get; }
 
+        /// <summary>
+        /// Determine if a DataViewType object is equal to another KeyType instance.
```
TomFinley: >DataViewType [](start = 27, length = 12)

We might prefer `<see` tags. Not that I think we'll rename them this late in the game, but the links would probably be nice.

TomFinley: >KeyType [](start = 67, length = 7)

I might reverse this, so, "if this `KeyType` instance is equal to another `DataViewType` instance." SInce *this* is assuredly a key type, and the other thing is assuredly a `DataView` Type. Note also, I might suggest being consistent in this sentence between the usage of instance and object. And the use of "this" instead of "a". Anyway, these latter points are not really big deals.

src/Microsoft.ML.Core/Data/KeyType.cs
```diff
@@ -59,6 +59,13 @@ public static bool IsValidDataType(Type type)
...
+        /// Checks if the other item is the type of KeyType, if the RawType
+        /// is the same, and if the Count is the same.
```
TomFinley: >Count [](start = 36, length = 5)

Similar notes on `<see` tags, here and everywhere.

src/Microsoft.ML.Core/Data/KeyType.cs
```diff
@@ -73,16 +80,31 @@ public override bool Equals(DataViewType other)
             return true;
         }
 
+        /// <summary>
+        /// Determine if a KeyType instance is equal to another KeyType instance.
+        /// Checks if any object is the type of KeyType, if the RawType
+        /// is the same, and if the Count is the same.
+        /// </summary>
+        /// <param name="other">The other object to compare against.</param>
+        /// <returns>A boolean if both objects are equal or not.</returns>
```
shmoradims: > /// <returns>A boolean if both objects are equal or not.</returns [](start = 7, length = 66)

for bool returns, please document them like this: [https://github.com/dotnet/machinelearning/blob/6e9023fa08ee2dba2d9b43e5485b7230c6dd29ed/src/Microsoft.ML.StandardLearners/Standard/Online/AveragedLinear.cs#L94](https://github.com/dotnet/machinelearning/blob/6e9023fa08ee2dba2d9b43e5485b7230c6dd29ed/src/Microsoft.ML.StandardLearners/Standard/Online/AveragedLinear.cs#L94)

src/Microsoft.ML.Core/Data/KeyType.cs
```diff
@@ -73,16 +80,31 @@ public override bool Equals(DataViewType other)
             return true;
         }
 
+        /// <summary>
+        /// Determine if a <see cref="KeyType"/> instance is equal to another <see cref="KeyType"/> instance.
+        /// Checks if any object is the type of <see cref="KeyType"/>, if the RawType
+        /// is the same, and if the Count is the same.
```
artidoro: Could you use the `<see`  tag for `Count` and for `RawType`?

src/Microsoft.ML.Core/Data/KeyType.cs
```diff
@@ -59,6 +59,13 @@ public static bool IsValidDataType(Type type)
         /// </summary>
         public ulong Count { get; }
 
+        /// <summary>
+        /// Determine if this <see cref="KeyType"/> object is equal to another <see cref="DataViewType"/> instance.
+        /// Checks if the other item is the type of <see cref="KeyType"/>, if the RawType
```
artidoro: Same here, could you use the `<see` tag for `RawType`?

PR #970, opened 2018-09-20
ComponentCatalog refactor

eerhardt:
This executes on the plan outlined in https://github.com/dotnet/machinelearning/issues/208#issuecomment-422136134 copied here for easy reading:

## New Proposal

1. We will move `ComponentCatalog` from being a static class to being an instance member on `Environment`. This has been a planned refactoring for ML.NET for a while, but hasn't been funded until now.
2. We will completely remove any implicit scanning for components in `ComponentCatalog` itself. It will have public APIs to register components, but will not do any registering itself - neither by loading assemblies from disk, nor by scanning loaded assemblies.
3. Other subsystems (like the GUI, command-line, Entry Points, and model loading) will be responsible for registering the components they require in the manner they require.
4. During model saving, we will write the `Assembly.FullName` into the model file. We will then register that assembly with the `env.ComponentCatalog` when loading the model.
    - Any model that was saved with a previous version of ML.NET, and loaded using the API, will need to explicitly register the components before loading the model. (Or they will need to save the model again with a current version of ML.NET that will save the assembly names.)

Under normal circumstances, API users won't have to explicitly register components with the `ComponentCatalog`. Using the API to train models won't require looking up components from a catalog - you just create .NET objects like normal. Loading a trained model from disk will register the components inside of it by loading the Assembly and scanning it for `LoadableClass` assembly attributes.

Fix #208

PR #2792, opened 2019-02-28
Rename Microsoft.ML.StandardLearners

### LINKED ISSUE #2786: Rename Microsoft.ML.StandardLearners to Microsoft.ML.StandardTrainers.
CC: @TomFinley , @eerhardt

Potapy4:
# Summary
I renamed `StandardLearners` to `StandardTrainers`. If I missed something, please let me know 👌 
Fixes #2786 

---


codemzs:
@Potapy4 Before picking up a task please make sure it isn’t assigned to anyone! In this case it was assigned to me and that means that person could already be working on it! Just be very careful moving forward.

TomFinley:
Hi @Potapy4 , thank you for working on this. Though, as @codemzs explains, we typically use the "assigned" field on issues to indicate that someone is actually actively working on it, which might lead to duplication of effort if someone else likewise takes on the work. But let's call that past praying for now. 😄

This test failure `LoadEntryPointModel` is an interesting one.

As we see of point 4 on this PR #970, we wrote the assembly name into the model file, so that we can find the loader signature later. This is on a whole a valuable change, and I think it will serve us well going forward. In this case it presents a difficulty, because in this case we are testing that our model loading is backwards compatible, as we do in several places. However those older models were saved with the old assembly name. It is here:

https://github.com/dotnet/machinelearning/blob/e1ab18ced02812a7185af293af3956cc9ebc3880/src/Microsoft.ML.Core/Data/ModelLoading.cs#L217-L227

I see several alternatives, of which I can identify two as possibly best. But between these I am not certain which is best. Note that this same difficulty @Potapy4 has faced here will crop up again when we rename the so-called `HalLearner`'s assembly, as well as fold `Microsoft.ML.Transforms` into `Microsoft.ML.Data`.

1. We take advantage of the fact that we are in preview mode (which is of course why we are undertaking all these breaking changes now, since we cannot do them later!), and say, "backwards compatibility with models built in preview are not a goal," and change the models in the test appropriately.

2. Internal to our own code, we have code that detects one of the "obsolete" `Microsoft.ML.*` assemblies of interest (in this case so far there is only one, `.StandardLearners`) we replace it with the desirable target assembly. Having magic strings in our codebase is a bit undesirable, but might be more desirable than just flat out failing. It would be relatively straightforward to add to the code, and when the time comes and we rename `HalLearners` in the coming days it may be desirable.


I myself believe 2 is correct, but I am not certain. It would be relatively easy to do I think

https://github.com/dotnet/machinelearning/blob/e1ab18ced02812a7185af293af3956cc9ebc3880/src/Microsoft.ML.Core/Data/ModelLoading.cs#L258-L265

We could imagine an auxiliary property `ForwardedLoaderAssemblyName` of roughly this form (pseudocode, don't take literally)...

```csharp
private string ForwardedLoaderAssemblyName
{ get {
switch (LoaderAssemblyName) {
case "Microsoft.ML.StandardLearners": return "Microsoft.ML.StandardTrainers";
default: return LoaderAssemblyName;
}
} }
```

Then we change the ensure loaded to work over this private `ForwardedLoaderAssemblyName` property, and when the time comes to rename the others, we have a mechanism in place to do so.

TomFinley:
In addition to people already involved on thread I'd welcome the thoughts of @eerhardt. He wrote this assembly storing code and might have anticipated this problem and already have a solution in mind. Of course anyone can participate.

eerhardt:
My initial reaction is to lean towards solution (1) above - we've been in preview mode for the past year, I don't think we've strictly guaranteed that we aren't going to break things from preview release to preview release. The API for sure has gone through vast breaking changes. Requiring people to re-build their models one more time might not be horrible.

That being said - this is the last time we get to say these things. Once `v1.0` is shipped, there is absolutely no questions here - we need to support model compatibility.

However, solution (2) is so simple that it may be worth doing, the cost of adding that code one time, and keeping it forever, is probably way less than the effort it will take to triage bugs and questions about why their old model no longer works. If it was more complex code, I would rethink that position.

So, in the end, it is probably "worth" doing solution (2).

TomFinley:
> So, in the end, it is probably "worth" doing solution (2).

I think so too. So maybe we say, "OK, let's just have this assembly forwarding." Are you comfortable doing that @Potapy4 ?

A natural followup question (which we do not have to answer now) is what .NET team considers a breaking change in the API, whether if we ship this capability in v1.0 to read pre-release versions, we are under obligation to keep it, *provided* the version v1.0 remains available for whoever might want to convert models (by simply reading and writing them). Sometimes I observe some libraries and tools say, "look, we dropped backward compatibility models for version X in version Z, but version Y reads version X format and writes in a format Z understands." (Where X < Y < Z.) This is quite common, but I don't know if it would be considered a breaking change in the API. I don't have in my mind a clear picture of all the things that might or might not be considered breaking changes in the API. (This is surely a sliding scale. By some very strict standards, even a change in `ToString` overload formatting could be considered a breaking change in an API. 😄)

But that followup question may not be one we want to answer right now. I would be interested though in your perspective.

eerhardt:
Here's what is documented in `corefx` for what is considered "breaking changes" to the BCL:

https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-changes.md

It links to

https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md

Potapy4:
Alright, first of all @codemzs I would like to apologize for taking this task from your plate - my bad 😔 next time I check the issues before I start working on them.

@TomFinley thanks for the good and detailed explanation about the tests. ❤️ I think we can use the second approach and make this "hack" right now. I also liked your pseudo-code, but I prefer `if` instead of `switch` - and I think that in this case `if` - this is more than enough. But if we gonna fix it later, do we have to somehow mark this place right now (I mean leave a link to issue or comment with `TODO` in code)?

TomFinley:
> @TomFinley thanks for the good and detailed explanation about the tests. ❤️ I think we can use the second approach and make this "hack" right now. I also liked your pseudo-code, but I prefer `if` instead of `switch` - and I think that in this case `if` - this is more than enough. But if we gonna fix it later, do we have to somehow mark this place right now (I mean leave a link to issue or comment with `TODO` in code)?

Sounds fine thanks @Potapy4 . My reasoning for preferring the switch is I know for a fact that there will be two others in the immediate future that can use the same mechanism, but of course this can be changed at any time. So that sounds fine.

I wouldn't worry too much about a TODO, since I would consider this work to be part of the other assembly renaming work, which I will be reviewing no doubt.

codemzs:
@Potapy4 I'm seeing you still haven't assigned the associated issue to yourself.

TomFinley:
> @Potapy4 I'm seeing you still haven't assigned the associated issue to yourself.

I don't think people can assign issues to themselves that do not have write access. I cannot assign him and I do have write access. @codemzs , can *you* do it?

eerhardt:
https://help.github.com/en/articles/assigning-issues-and-pull-requests-to-other-github-users

> If you have write access to a repository, you can assign issues and pull requests to yourself, collaborators on personal projects, or members of your organization with read permissions on the repository.

*members of your organization* is the key there. Non-members need to accept an invitation first before they can be assigned issues.

src/Microsoft.ML.Core/Data/ModelLoadContext.cs
```diff
@@ -40,6 +40,11 @@ internal sealed partial class ModelLoadContext : IDisposable
         /// </summary>
         public readonly string[] Strings;
 
+        /// <summary>
+        /// The name of the assembly that the loader lives in. This property is used for wrapping old ML.StandardLearners assembly.
+        /// </summary>
+        public readonly string ForwardedLoaderAssemblyName => LoaderAssemblyName == "Microsoft.ML.StandardLearners" ? "Microsoft.ML.StandardTrainers" : LoaderAssemblyName;
```
TomFinley: >readonly [](start = 15, length = 8)

Note that unlike the other things here which are fields, what you are writing here is a property, and the `readonly` modifier will not work for it.

codemzs: I would recommend removing this, syncing to master and updating the existing `[ForwardedLoaderAssemblyName](https://github.com/dotnet/machinelearning/blob/df2f3fb49d4299925325921ee31c2cdff78ba3b5/src/Microsoft.ML.Core/Data/ModelLoading.cs#L27)`, StandardLearners isn't the only assembly that requires forwarding.

codemzs:
@Potapy4 I just forced pushed the changes into your branch so that we can close this PR soon as its been open for a while. @TomFinley Lets review and close this stuff ....

docs/code/EntryPoints.md
```diff
@@ -220,11 +220,11 @@ parameter.
 
 ## How to create an entry point for an existing ML.NET component
 
-1. Add a `LoadableClass` assembly attribute with the `SignatureEntryPointModule` signature as shown [here](https://github.com/dotnet/machinelearning/blob/9db16c85888e7163c671543faee6ba1f47015d68/src/Microsoft.ML.StandardLearners/Standard/LogisticRegression/LogisticRegression.cs#L27).  
+1. Add a `LoadableClass` assembly attribute with the `SignatureEntryPointModule` signature as shown [here](https://github.com/dotnet/machinelearning/blob/9db16c85888e7163c671543faee6ba1f47015d68/src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LogisticRegression.cs#L27).
```
eerhardt: These links are now broken. The reason is that these links are using SHAs of a certain commit in the history of the repo. In that point in the history, this was named `ML.StandardLearners`.

So either you need to update the SHA in the link, or keep it `ML.StandardLearners`.

codemzs: Good catch. I have reverted this file so that we check-in stuff that is not broken. I will create another PR once this change is checked-in and I have SHA of the new commit that has these links in the new renamed directory.

PR #2796, opened 2019-02-28
Replace predicate with an IEnumerable<DataViewSchema.Column> for IRowToRowMapper.GetRow and  ISchemaBoundRowMapper.GetRow

### LINKED ISSUE #1529: IDataView Cleanup: Predicates from int to Column

sfilipi:
Closes #1529 

Note to reviewers: The first commit does all the replacement work. 
The second commit, has more files on it, because it deals with renaming the col to columnIndex in the other two methods of IDataView (@eerhardt comment on #1529)
The third commit is me reviewing the first commit, and making sure that GetRow is an explicit implementation everywhere. 

I have had most of the work on this PR done before breaking ISchemaBoundRowMapper deriving from IRowToRowMapper; so I kept the signature change of ISchemaBoundRowMapper.GetRow although that interface is internal. LMK if that is not desirable.

### DIFF
--- a/src/Microsoft.Data.DataView/IDataView.cs
+++ b/src/Microsoft.Data.DataView/IDataView.cs
@@ -132,9 +132,9 @@ namespace Microsoft.Data.DataView
         public abstract ValueGetter<DataViewRowId> GetIdGetter();
 
         /// <summary>
-        /// Returns whether the column with the given index, is active in this row.
+        /// Returns whether the give column is active in this row.
         /// </summary>
-        public abstract bool IsColumnActive(int columnIndex);
+        public abstract bool IsColumnActive(DataViewSchema.Column column);
 
         /// <summary>
         /// Returns a value getter delegate to fetch the value of the given <paramref name="column"/>, from the row.
--- a/src/Microsoft.Data.DataView/IDataView.cs
+++ b/src/Microsoft.Data.DataView/IDataView.cs
@@ -141,6 +141,8 @@ namespace Microsoft.Data.DataView
         /// This throws if the column is not active in this row, or if the type
         /// <typeparamref name="TValue"/> differs from this column's type.
         /// </summary>
+        /// <typeparam name="TValue"> is the output column's content type.</typeparam>
+        /// <param name="columnIndex"> is the index of a output column whose getter should be returned.</param>
         public abstract ValueGetter<TValue> GetGetter<TValue>(int columnIndex);
 
         /// <summary>
--- a/src/Microsoft.Data.DataView/IDataView.cs
+++ b/src/Microsoft.Data.DataView/IDataView.cs
@@ -132,7 +132,7 @@ namespace Microsoft.Data.DataView
         public abstract ValueGetter<DataViewRowId> GetIdGetter();
 
         /// <summary>
-        /// Returns whether the give column is active in this row.
+        /// Returns whether the given column is active in this row.
         /// </summary>
         public abstract bool IsColumnActive(DataViewSchema.Column column);
 
@@ -141,7 +141,7 @@ namespace Microsoft.Data.DataView
         /// This throws if the column is not active in this row, or if the type
         /// <typeparamref name="TValue"/> differs from this column's type.
         /// </summary>
-        /// <typeparam name="TValue"> is the output column's content type.</typeparam>
+        /// <typeparam name="TValue"> is the column's content type.</typeparam>
         /// <param name="column"> is the output column whose getter should be returned.</param>
         public abstract ValueGetter<TValue> GetGetter<TValue>(DataViewSchema.Column column);
 
--- a/src/Microsoft.Data.DataView/IDataView.cs
+++ b/src/Microsoft.Data.DataView/IDataView.cs
@@ -132,16 +132,16 @@ namespace Microsoft.Data.DataView
         public abstract ValueGetter<DataViewRowId> GetIdGetter();
 
         /// <summary>
-        /// Returns whether the given column is active in this row.
+        /// Returns whether the column with the given index, is active in this row.
         /// </summary>
-        public abstract bool IsColumnActive(int col);
+        public abstract bool IsColumnActive(int columnIndex);
 
         /// <summary>
-        /// Returns a value getter delegate to fetch the given column value from the row.
+        /// Returns a value getter delegate to fetch the valueof column with the given columnIndex, from the row.
         /// This throws if the column is not active in this row, or if the type
         /// <typeparamref name="TValue"/> differs from this column's type.
         /// </summary>
-        public abstract ValueGetter<TValue> GetGetter<TValue>(int col);
+        public abstract ValueGetter<TValue> GetGetter<TValue>(int columnIndex);
 
         /// <summary>
         /// Gets a <see cref="Schema"/>, which provides name and type information for variables
--- a/src/Microsoft.Data.DataView/IDataView.cs
+++ b/src/Microsoft.Data.DataView/IDataView.cs
@@ -137,13 +137,13 @@ namespace Microsoft.Data.DataView
         public abstract bool IsColumnActive(int columnIndex);
 
         /// <summary>
-        /// Returns a value getter delegate to fetch the valueof column with the given columnIndex, from the row.
+        /// Returns a value getter delegate to fetch the value of the given <paramref name="column"/>, from the row.
         /// This throws if the column is not active in this row, or if the type
         /// <typeparamref name="TValue"/> differs from this column's type.
         /// </summary>
         /// <typeparam name="TValue"> is the output column's content type.</typeparam>
-        /// <param name="columnIndex"> is the index of a output column whose getter should be returned.</param>
-        public abstract ValueGetter<TValue> GetGetter<TValue>(int columnIndex);
+        /// <param name="column"> is the output column whose getter should be returned.</param>
+        public abstract ValueGetter<TValue> GetGetter<TValue>(DataViewSchema.Column column);
 
         /// <summary>
         /// Gets a <see cref="Schema"/>, which provides name and type information for variables
--- a/src/Microsoft.Data.DataView/SchemaDebuggerProxy.cs
+++ b/src/Microsoft.Data.DataView/SchemaDebuggerProxy.cs
@@ -41,16 +41,16 @@ namespace Microsoft.Data.DataView
             foreach (var column in annotations.Schema)
             {
                 var name = column.Name;
-                var value = Utils.MarshalInvoke(GetValue<int>, column.Type.RawType, annotations, column.Index);
+                var value = Utils.MarshalInvoke(GetValue<DataViewSchema.Column>, column.Type.RawType, annotations, column);
                 result.Add(new KeyValuePair<string, object>(name, value));
             }
             return result;
         }
 
-        private static object GetValue<T>(DataViewSchema.Annotations annotations, int columnIndex)
+        private static object GetValue<T>(DataViewSchema.Annotations annotations, DataViewSchema.Column column)
         {
             T value = default;
-            annotations.GetGetter<T>(annotations.Schema[columnIndex])(ref value);
+            annotations.GetGetter<T>(column)(ref value);
             return value;
         }
     }
--- a/src/Microsoft.Data.DataView/SchemaDebuggerProxy.cs
+++ b/src/Microsoft.Data.DataView/SchemaDebuggerProxy.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Data.DataView
         private static object GetValue<T>(DataViewSchema.Annotations annotations, int columnIndex)
         {
             T value = default;
-            annotations.GetGetter<T>(columnIndex)(ref value);
+            annotations.GetGetter<T>(annotations.Schema[columnIndex])(ref value);
             return value;
         }
     }
--- a/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
+++ b/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
@@ -460,9 +460,22 @@ namespace Microsoft.ML.Data
             public override DataViewSchema Schema => _annotations.Schema;
             public override long Position => 0;
             public override long Batch => 0;
-            public override ValueGetter<TValue> GetGetter<TValue>(int col) => _annotations.GetGetter<TValue>(col);
+
+            /// <summary>
+            /// Returns a value getter delegate to fetch the valueof column with the given columnIndex, from the row.
+            /// This throws if the column is not active in this row, or if the type
+            /// <typeparamref name="TValue"/> differs from this column's type.
+            /// </summary>
+            /// <typeparam name="TValue"> is the output column's content type.</typeparam>
+            /// <param name="columnIndex"> is the index of a output column whose getter should be returned.</param>
+            public override ValueGetter<TValue> GetGetter<TValue>(int columnIndex) => _annotations.GetGetter<TValue>(columnIndex);
+
             public override ValueGetter<DataViewRowId> GetIdGetter() => (ref DataViewRowId dst) => dst = default;
-            public override bool IsColumnActive(int col) => true;
+
+            /// <summary>
+            /// Returns whether the given column is active in this row.
+            /// </summary>
+            public override bool IsColumnActive(int columnIndex) => true;
         }
 
         /// <summary>
--- a/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
+++ b/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
@@ -475,7 +475,7 @@ namespace Microsoft.ML.Data
             /// <summary>
             /// Returns whether the given column is active in this row.
             /// </summary>
-            public override bool IsColumnActive(int columnIndex) => true;
+            public override bool IsColumnActive(DataViewSchema.Column column) => true;
         }
 
         /// <summary>
--- a/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
+++ b/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
@@ -466,7 +466,7 @@ namespace Microsoft.ML.Data
             /// This throws if the column is not active in this row, or if the type
             /// <typeparamref name="TValue"/> differs from this column's type.
             /// </summary>
-            /// <typeparam name="TValue"> is the output column's content type.</typeparam>
+            /// <typeparam name="TValue"> is the column's content type.</typeparam>
             /// <param name="column"> is the output column whose getter should be returned.</param>
             public override ValueGetter<TValue> GetGetter<TValue>(DataViewSchema.Column column) => _annotations.GetGetter<TValue>(column);
 
--- a/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
+++ b/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
@@ -462,13 +462,13 @@ namespace Microsoft.ML.Data
             public override long Batch => 0;
 
             /// <summary>
-            /// Returns a value getter delegate to fetch the valueof column with the given columnIndex, from the row.
+            /// Returns a value getter delegate to fetch the value of column with the given columnIndex, from the row.
             /// This throws if the column is not active in this row, or if the type
             /// <typeparamref name="TValue"/> differs from this column's type.
             /// </summary>
             /// <typeparam name="TValue"> is the output column's content type.</typeparam>
-            /// <param name="columnIndex"> is the index of a output column whose getter should be returned.</param>
-            public override ValueGetter<TValue> GetGetter<TValue>(int columnIndex) => _annotations.GetGetter<TValue>(columnIndex);
+            /// <param name="column"> is the output column whose getter should be returned.</param>
+            public override ValueGetter<TValue> GetGetter<TValue>(DataViewSchema.Column column) => _annotations.GetGetter<TValue>(column);
 
             public override ValueGetter<DataViewRowId> GetIdGetter() => (ref DataViewRowId dst) => dst = default;
 
--- a/src/Microsoft.ML.Core/Utilities/Utils.cs
+++ b/src/Microsoft.ML.Core/Utilities/Utils.cs
@@ -782,6 +782,42 @@ namespace Microsoft.ML.Internal.Utilities
         /// map.
         /// </summary>
         /// <param name="lim">Indicates the exclusive upper bound on the tested values</param>
+        /// <param name="schema">The input schema where the predicate can checkif columns are active.</param>
+        /// <param name="pred">The predicate to test for various value</param>
+        /// <param name="map">An ascending array of values from 0 inclusive
+        /// to <paramref name="lim"/> exclusive, holding all values for which
+        /// <paramref name="pred"/> is true</param>
+        /// <param name="invMap">Forms an inverse mapping of <paramref name="map"/>,
+        /// so that <c><paramref name="invMap"/>[<paramref name="map"/>[i]] == i</c>,
+        /// and for other entries not appearing in <paramref name="map"/>,
+        /// <c><paramref name="invMap"/>[i] == -1</c></param>
+        public static void BuildSubsetMaps(int lim, DataViewSchema schema, Func<DataViewSchema.Column, bool> pred, out int[] map, out int[] invMap)
+        {
+            Contracts.CheckParam(lim >= 0, nameof(lim));
+            Contracts.CheckValue(pred, nameof(pred));
+            // REVIEW: Better names?
+            List<int> mapList = new List<int>();
+            invMap = new int[lim];
+            for (int c = 0; c < lim; ++c)
+            {
+                if (!pred(schema[c]))
+                {
+                    invMap[c] = -1;
+                    continue;
+                }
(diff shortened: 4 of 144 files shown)


src/Microsoft.Data.DataView/IDataView.cs
```diff
@@ -132,16 +132,18 @@ public abstract class DataViewRow : IDisposable
...
         /// <summary>
-        /// Returns whether the given column is active in this row.
+        /// Returns whether the column with the given index, is active in this row.
         /// </summary>
-        public abstract bool IsColumnActive(int col);
+        public abstract bool IsColumnActive(int columnIndex);
 
         /// <summary>
-        /// Returns a value getter delegate to fetch the given column value from the row.
+        /// Returns a value getter delegate to fetch the valueof column with the given columnIndex, from the row.
```
Ivanidzo4ka: >valueof [](start = 57, length = 7)

space

Ivanidzo4ka: >columnIndex [](start = 87, length = 11)

paramref?

src/Microsoft.ML.Core/Data/AnnotationUtils.cs
```diff
@@ -460,9 +460,22 @@ public AnnotationRow(DataViewSchema.Annotations annotations)
             public override DataViewSchema Schema => _annotations.Schema;
             public override long Position => 0;
             public override long Batch => 0;
-            public override ValueGetter<TValue> GetGetter<TValue>(int col) => _annotations.GetGetter<TValue>(col);
+
+            /// <summary>
+            /// Returns a value getter delegate to fetch the valueof column with the given columnIndex, from the row.
```
Ivanidzo4ka: >valueof [](start = 61, length = 7)

space, I guess across all files.

src/Microsoft.ML.Data/DataView/AppendRowsDataView.cs
```diff
@@ -187,18 +187,21 @@ protected Delegate CreateGetter(int col)
 
             protected abstract ValueGetter<TValue> CreateTypedGetter<TValue>(int col);
 
-            public sealed override ValueGetter<TValue> GetGetter<TValue>(int col)
+            public sealed override ValueGetter<TValue> GetGetter<TValue>(int columnIndex)
             {
-                Ch.Check(IsColumnActive(col), "The column must be active against the defined predicate.");
-                if (!(Getters[col] is ValueGetter<TValue>))
+                Ch.Check(IsColumnActive(columnIndex), "The column must be active against the defined predicate.");
+                if (!(Getters[columnIndex] is ValueGetter<TValue>))
```
Ivanidzo4ka: ` Ch.CheckParam(IsColumnActive(columnIndex), nameof(columnIndex), "requested column not active");`
from transposeLoader.cs
I like it more than some mythical `defined predicate`

src/Microsoft.ML.Data/DataView/AppendRowsDataView.cs
```diff
@@ -187,18 +187,21 @@ protected Delegate CreateGetter(int col)
...
-            public sealed override bool IsColumnActive(int col)
+            /// <summary>
+            /// Returns whether the given column is active in this row.
+            /// </summary>
+            public sealed override bool IsColumnActive(int columnIndex)
             {
-                Ch.Check(0 <= col && col < Schema.Count, "Column index is out of range");
-                return Getters[col] != null;
+                Ch.Check(0 <= columnIndex && columnIndex < Schema.Count, "Column index is out of range");
+                return Getters[columnIndex] != null;
```
Ivanidzo4ka: `Ch.CheckParam(0 <= columnIndex && columnIndex < _colToActive.Length, nameof(columnIndex));`
in files above this one we use this format. consistency would be nice. not a blocker tho.

src/Microsoft.ML.Data/DataView/ArrayDataViewBuilder.cs
```diff
@@ -304,17 +304,27 @@ public override ValueGetter<DataViewRowId> GetIdGetter()
...
+                /// </summary>
+                /// <typeparam name="TValue"> is the output column's content type.</typeparam>
+                /// <param name="columnIndex"> is the index of a output column whose getter should be returned.</param>
+                public override ValueGetter<TValue> GetGetter<TValue>(int columnIndex)
                 {
-                    Ch.Check(0 <= col & col < Schema.Count);
-                    Ch.Check(_active[col], "column is not active");
-                    var column = _view._columns[col] as Column<TValue>;
+                    Ch.Check(0 <= columnIndex & columnIndex < Schema.Count);
+                    Ch.Check(_active[columnIndex], "column is not active");
```
Ivanidzo4ka: >"column is not active" [](start = 51, length = 22)

for the sake of consistency: "requested column not active"

src/Microsoft.ML.Data/DataView/CacheDataView.cs
```diff
@@ -1177,11 +1191,11 @@ protected sealed override void Dispose(bool disposing)
...
             }
 
-            public sealed override ValueGetter<TValue> GetGetter<TValue>(int col)
+            public sealed override ValueGetter<TValue> GetGetter<TValue>(int columnIndex)
             {
-                if (!IsColumnActive(col))
-                    throw Ch.Except("Column #{0} is requested but not active in the cursor", col);
-                var getter = _getters[_colToActivesIndex[col]] as ValueGetter<TValue>;
+                if (!IsColumnActive(columnIndex))
+                    throw Ch.Except("Column #{0} is requested but not active in the cursor", columnIndex);
```
Ivanidzo4ka: in how many different ways we can do same thing :)
`Ch.CheckParam(IsColumnActive(columnIndex), nameof(columnIndex), "requested column not active");`?

src/Microsoft.ML.Data/DataView/CompositeRowToRowMapper.cs
```diff
@@ -49,18 +49,19 @@ IEnumerable<DataViewSchema.Column> IRowToRowMapper.GetDependencies(IEnumerable<D
             return columnsNeeded;
         }
 
-        public DataViewRow GetRow(DataViewRow input, Func<int, bool> active)
+        DataViewRow IRowToRowMapper.GetRow(DataViewRow input, IEnumerable<DataViewSchema.Column> activeColumns)
```
Ivanidzo4ka: >IEnumerable<DataViewSchema.Column> activeColumns [](start = 62, length = 48)

It's probably quite late considering how much work you did already for this issue.
So we replace functions to schema columns.
In past we had just indexes and we just check are they belong to schema range or not.
Now we accept schema columns. But we use only index position in them.
Which can lead to situation where i have two dataviews with following schemas: 
one is `A: int, B: float`,  and another is `C:string, D: bool`.
I want to get rows from first dataview. but I made mistake and I pass schema from second dataview as active columns.
I would assume it would work perfectly fine in current state of the code. I'm not sure it should.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -631,27 +631,20 @@ IEnumerable<DataViewSchema.Column> ISchemaBoundRowMapper.GetDependenciesForNewCo
...
-                    if (!predicate(i))
+                    if (column.Index == OutputSchema.Count - 1)
                         continue;
-                    getters[i] = Utils.MarshalInvoke(GetPredictorGetter<int>, type.RawType, predictorRow, i);
+                    var type = predictorRow.Schema[column.Index].Type;
+                    getters[column.Index] = Utils.MarshalInvoke(GetPredictorGetter<int>, type.RawType, predictorRow, column.Index);
                 }
-                if (predicate(OutputSchema.Count - 1))
+
+                if (activeColumns.Select(c => c.Name).Contains(DefaultColumnNames.Probability))
```
Ivanidzo4ka: > if (activeColumns.Select(c => c.Name).Contains(DefaultColumnNames.Probability)) [](start = 15, length = 80)

I would rather change if inside foreach loop to set bool flag to true if column.Index == OutputSchema.Count-1 instead of continue and check that flag here.
You have mix of two different conditions right now.
And who knows maybe we will allow user to define their own names for probability column in future, so index feels somewhat more secure.

src/Microsoft.ML.Data/Scorers/FeatureContributionCalculation.cs
```diff
@@ -385,9 +385,7 @@ public DataViewRow GetRow(DataViewRow input, Func<int, bool> active)
             }
 
             public Func<int, bool> GetGenericPredicate(Func<int, bool> predicate)
```
Ivanidzo4ka: >GetGenericPredicate [](start = 35, length = 19)

is it still needed?

src/Microsoft.ML.Data/Scorers/FeatureContributionCalculation.cs
```diff
@@ -360,21 +360,21 @@ IEnumerable<DataViewSchema.Column> ISchemaBoundRowMapper.GetDependenciesForNewCo
...
+            DataViewRow ISchemaBoundRowMapper.GetRow(DataViewRow input, IEnumerable<DataViewSchema.Column> activeColumns)
             {
                 Contracts.AssertValue(input);
-                Contracts.AssertValue(active);
+                Contracts.AssertValue(activeColumns);
                 var totalColumnsCount = 1 + _outputGenericSchema.Count;
                 var getters = new Delegate[totalColumnsCount];
 
-                if (active(totalColumnsCount - 1))
+                if (activeColumns.Select(c => c.Index).Contains(totalColumnsCount - 1))
```
Ivanidzo4ka: >totalColumnsCount - 1 [](start = 64, length = 21)

this is _outputGenericSchema.Count :)

src/Microsoft.ML.Data/Transforms/BindingsWrappedRowCursor.cs
```diff
@@ -34,21 +34,31 @@ public BindingsWrappedRowCursor(IChannelProvider provider, DataViewRowCursor inp
...
         }
 
-        public override bool IsColumnActive(int col)
+        /// <summary>
+        /// Returns whether the given column is active in this row.
+        /// </summary>
+        public override bool IsColumnActive(int columnIndex)
         {
-            Ch.Check(0 <= col & col < _bindings.ColumnCount, "col");
+            Ch.Check(0 <= columnIndex & columnIndex < _bindings.ColumnCount, "col");
```
Ivanidzo4ka: >"col" [](start = 77, length = 5)

nameof!

src/Microsoft.ML.Data/Transforms/TransformBase.cs
```diff
@@ -800,13 +809,13 @@ protected override int MapColumnIndex(out bool isSrc, int col)
...
 
-        protected override Delegate[] CreateGetters(DataViewRow input, Func<int, bool> active, out Action disposer)
+        protected override Delegate[] CreateGetters(DataViewRow input, IEnumerable<DataViewSchema.Column> activeColumns, out Action disposer)
         {
             Func<int, bool> activeInfos =
                 iinfo =>
                 {
                     int col = _bindings.MapIinfoToCol(iinfo);
-                    return active(col);
+                    return activeColumns.Select(c => c.Index).Contains(col);
```
Ivanidzo4ka: >activeColumns.Select(c => c.Index) [](start = 27, length = 34)

Make a hashset before you define activeInfos and use hashSet.Contains here.

src/Microsoft.ML.TimeSeries/PredictionFunction.cs
```diff
@@ -108,29 +108,30 @@ private static ITransformer CloneTransformers(ITransformer transformer)
...
         {
             Contracts.CheckValue(input, nameof(input));
-            Contracts.CheckValue(active, nameof(active));
+            Contracts.CheckValue(activeColumns, nameof(activeColumns));
 
             IRowToRowMapper[] innerMappers = new IRowToRowMapper[0];
             if (mapper is CompositeRowToRowMapper compositeMapper)
                 innerMappers = compositeMapper.InnerMappers;
 
+            var activeIndices = activeColumns.Select(c => c.Index);
```
Ivanidzo4ka: ToArray or HashSet.

src/Microsoft.ML.Transforms/OptionalColumnTransform.cs
```diff
@@ -363,13 +363,13 @@ protected override int MapColumnIndex(out bool isSrc, int col)
...
 
-        protected override Delegate[] CreateGetters(DataViewRow input, Func<int, bool> active, out Action disposer)
+        protected override Delegate[] CreateGetters(DataViewRow input, IEnumerable<DataViewSchema.Column> activeColumns, out Action disposer)
         {
             Func<int, bool> activeInfos =
                 iinfo =>
                 {
                     int col = _bindings.MapIinfoToCol(iinfo);
-                    return active(col);
+                    return activeColumns.Select(c => c.Index).Contains(col);
```
Ivanidzo4ka: >activeColumns.Select(c => c.Index) [](start = 27, length = 34)

hashset

src/Microsoft.ML.Data/Scorers/RowToRowScorerBase.cs
```diff
@@ -155,30 +155,25 @@ public override DataViewRowCursor[] GetRowCursorSet(IEnumerable<DataViewSchema.C
...
             var bindings = GetBindings();
             IEnumerable<DataViewSchema.Column> inputColumns;
-            Func<int, bool> predicateMapper;
-            IEnumerable<DataViewSchema.Column> activeColumns = OutputSchema.Where(col => active(col.Index));
-            GetActive(bindings, activeColumns, out inputColumns, out predicateMapper);
-            var output = bindings.RowMapper.GetRow(input, predicateMapper);
-            Func<int, bool> activeInfos = iinfo => active(bindings.MapIinfoToCol(iinfo));
+            GetActive(bindings, activeColumns, out inputColumns, out IEnumerable<DataViewSchema.Column> activeMapperColumns);
+            var output = bindings.RowMapper.GetRow(input, activeMapperColumns);
+            Func<int, bool> activeInfos = iinfo => activeColumns.Select(c => c.Index).Contains(bindings.MapIinfoToCol(iinfo));
```
Ivanidzo4ka: >activeColumns [](start = 51, length = 13)

this can be quite heavy

src/Microsoft.Data.DataView/IDataView.cs
```diff
@@ -132,16 +132,18 @@ public abstract class DataViewRow : IDisposable
         public abstract ValueGetter<DataViewRowId> GetIdGetter();
 
         /// <summary>
-        /// Returns whether the given column is active in this row.
+        /// Returns whether the column with the given index, is active in this row.
         /// </summary>
-        public abstract bool IsColumnActive(int col);
+        public abstract bool IsColumnActive(int columnIndex);
```
eerhardt: I thought we were changing these two methods to take in the column instead of the index?

`public abstract bool IsColumnActive(DataViewSchema.Column column);`

sfilipi: I think the predicates were deemed hard to debug. Ideally we'd pass the column and not deal with the indexes but is that a V1 issue? @shauheen @glebuk @TomFinley

eerhardt: The issue being closed here says:

> The use of integer indices here has sometimes led to confusion or even bugs.

Which seems to me that we shouldn't have integer indices in our public API. So if we ship V1 like this, we won't be able to remove/change this API. We can add a new `virtual` method that takes in the `Column`, but this current API will need to live forever.

src/Microsoft.ML.Data/DataView/AppendRowsDataView.cs
sfilipi: this does not have _colToActive. Leaving it. 

---
In reply to: [261485256](https://github.com/dotnet/machinelearning/pull/2796#discussion_r261485256) [](ancestors = 261485256)

src/Microsoft.ML.Data/DataView/CacheDataView.cs
sfilipi: in many. 

---
In reply to: [261485672](https://github.com/dotnet/machinelearning/pull/2796#discussion_r261485672) [](ancestors = 261485672)

src/Microsoft.ML.Data/DataView/CompositeRowToRowMapper.cs
sfilipi: Agree. should switch from tracking activeIndices to active columns at some point. I'll log an issue. 

---
In reply to: [261487085](https://github.com/dotnet/machinelearning/pull/2796#discussion_r261487085) [](ancestors = 261487085)

src/Microsoft.ML.Data/Transforms/TransformBase.cs
sfilipi: thanks. 

---
In reply to: [261497161](https://github.com/dotnet/machinelearning/pull/2796#discussion_r261497161) [](ancestors = 261497161)

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
```diff
@@ -1153,7 +1155,7 @@ private static IDataView AddVarLengthColumn<TSrc>(IHostEnvironment env, IDataVie
                             throw Contracts.Except("Multiple unweighted rows found in metrics data view.");
 
                         numResults++;
-                        UpdateSums(isWeightedCol, stratCol, stratVal, agg, numFolds > 1, metricNames, hasWeighted, hasStrats,
+                        UpdateSums(isWeightedColumn.Value.Index, stratCol, stratVal, agg, numFolds > 1, metricNames, hasWeighted, hasStrats,
```
sfilipi: switch to isWeightedCol

src/Microsoft.Data.DataView/IDataView.cs
```diff
@@ -132,16 +132,18 @@ public abstract class DataViewRow : IDisposable
         public abstract ValueGetter<DataViewRowId> GetIdGetter();
 
         /// <summary>
-        /// Returns whether the given column is active in this row.
+        /// Returns whether the give column is active in this row.
```
sfilipi: fix

sfilipi:
>         }

try removing this, as it has only one reference. 

---
Refers to: src/Microsoft.ML.Core/Utilities/Utils.cs:847 in 7aa80a1. [](commit_id = 7aa80a18914f4a6f906d64a44a1948537a61d196, deletion_comment = False)

src/Microsoft.ML.Data/Data/DataViewUtils.cs
```diff
@@ -523,7 +523,7 @@ private DataViewRowCursor[] SplitCore(IChannelProvider ch, DataViewRowCursor inp
                 // Create the mappings between active column index, and column index.
                 int[] activeToCol;
                 int[] colToActive;
-                Utils.BuildSubsetMaps(_schema.Count, input.IsColumnActive, out activeToCol, out colToActive);
+                Utils.BuildSubsetMaps(_schema.Count, _schema, input.IsColumnActive, out activeToCol, out colToActive);
```
sfilipi: >_schema.Count [](start = 38, length = 13)

try removing the count, if we're passing schema

sfilipi:
>         }

it is too involved atm. 

---
In reply to: [469811161](https://github.com/dotnet/machinelearning/pull/2796#issuecomment-469811161) [](ancestors = 469811161)

---
Refers to: src/Microsoft.ML.Core/Utilities/Utils.cs:847 in 7aa80a1. [](commit_id = 7aa80a18914f4a6f906d64a44a1948537a61d196, deletion_comment = False)

src/Microsoft.Data.DataView/IDataView.cs
```diff
@@ -132,16 +132,18 @@ public abstract class DataViewRow : IDisposable
...
+        public abstract bool IsColumnActive(DataViewSchema.Column column);
 
         /// <summary>
-        /// Returns a value getter delegate to fetch the given column value from the row.
+        /// Returns a value getter delegate to fetch the value of the given <paramref name="column"/>, from the row.
         /// This throws if the column is not active in this row, or if the type
         /// <typeparamref name="TValue"/> differs from this column's type.
         /// </summary>
-        public abstract ValueGetter<TValue> GetGetter<TValue>(int col);
+        /// <typeparam name="TValue"> is the output column's content type.</typeparam>
```
eerhardt: (nit) Is it confusing to say `output column` here and below?  To the consumer of an IDataView, they are just `column`s.

src/Microsoft.Data.DataView/SchemaDebuggerProxy.cs
```diff
@@ -50,7 +50,7 @@ public AnnotationsDebuggerProxy(DataViewSchema.Annotations annotations)
         private static object GetValue<T>(DataViewSchema.Annotations annotations, int columnIndex)
         {
             T value = default;
-            annotations.GetGetter<T>(columnIndex)(ref value);
+            annotations.GetGetter<T>(annotations.Schema[columnIndex])(ref value);
```
eerhardt: This method is only called in one place - which already has the Column. Can we just pass the column through here?

src/Microsoft.Data.DataView/SchemaDebuggerProxy.cs
```diff
@@ -41,16 +41,16 @@ public AnnotationsDebuggerProxy(DataViewSchema.Annotations annotations)
             foreach (var column in annotations.Schema)
             {
                 var name = column.Name;
-                var value = Utils.MarshalInvoke(GetValue<int>, column.Type.RawType, annotations, column.Index);
+                var value = Utils.MarshalInvoke(GetValue<DataViewSchema.Column>, column.Type.RawType, annotations, column);
```
TomFinley: >DataViewSchema.Column [](start = 57, length = 21)

Incidentally this was not necessary or helpful. The point is to merely have *a* method for the method to sink its teeth into, so the type parameter here was not informative. (If anything it is now more obscure.)

TomFinley: This is not essential -- the method will still work the same even with this change, and there should be no cost to it. Just pointing out that there was no need for it.

---
In reply to: [263101561](https://github.com/dotnet/machinelearning/pull/2796#discussion_r263101561) [](ancestors = 263101561)

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -776,6 +776,41 @@ public static T[] BuildArray<T>(int length, Func<int, T> func)
...
+        /// <param name="schema">The input schema where the predicate can check if columns are active.</param>
+        /// <param name="pred">The predicate to test for various value</param>
+        /// <param name="map">An ascending array of values from 0 inclusive
+        /// to <paramref name="schema.Count"/> exclusive, holding all values for which
+        /// <paramref name="pred"/> is true</param>
+        /// <param name="invMap">Forms an inverse mapping of <paramref name="map"/>,
+        /// so that <c><paramref name="invMap"/>[<paramref name="map"/>[i]] == i</c>,
+        /// and for other entries not appearing in <paramref name="map"/>,
+        /// <c><paramref name="invMap"/>[i] == -1</c></param>
+        public static void BuildSubsetMaps(DataViewSchema schema, Func<DataViewSchema.Column, bool> pred, out int[] map, out int[] invMap)
```
TomFinley: >Func<DataViewSchema.Column, bool> pred [](start = 66, length = 38)

Hmmm. Is the goal here to enable more elaborate checking I suppose? Totally fine, just interesting.

TomFinley:
Reviewing this reminds me, we'll have to deal with `ITransposeDataView` sooner or later. But, maybe later.

PR #2797, opened 2019-02-28
Refactor cancellation mechanism and make it internal, accessible via experimental nuget.

### LINKED ISSUE #2795: Refactor cancellation mechanism and make it internal, accessible via experimental nuget.
Some time ago I added `StopExecution` method and `IsCancelled` flag to IHost.
https://github.com/dotnet/machinelearning/blob/fbf282d982223a51cf6e4a9ad9f3f036e0f150ca/src/Microsoft.ML.Core/Data/IHostEnvironment.cs#L98

https://github.com/dotnet/machinelearning/blob/fbf282d982223a51cf6e4a9ad9f3f036e0f150ca/src/Microsoft.ML.Core/Data/IHostEnvironment.cs#L74

It was quick solution to make certain algorithms react to stop signal from user. I doubt in long run it's proper way to do it. And it probably required lengthy discussion how to expose stopping functionality (method, or `CancellationToken`, or something else).

But for now I think it's a good idea to remove method and flag from our public API. Especially considering what no code actually checking `IsCancelled` flag.

EDIT: We decided to keep this functionality but make it internal and still be accessible via experimental nuget.

codemzs:
fixes #2795

### DIFF
--- a/pkg/Microsoft.ML.Experimental/Microsoft.ML.Experimental.nupkgproj
+++ b/pkg/Microsoft.ML.Experimental/Microsoft.ML.Experimental.nupkgproj
@@ -1,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk" DefaultTargets="Pack">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageDescription>Microsoft.ML.Experimental contains extension method for MLContext to access internal function to cancel ML pipeline.</PackageDescription>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../Microsoft.ML.Data/Microsoft.ML.Data.nupkgproj" />
+  </ItemGroup>
+
+</Project>
--- a/pkg/Microsoft.ML.Experimental/Microsoft.ML.Experimental.nupkgproj
+++ b/pkg/Microsoft.ML.Experimental/Microsoft.ML.Experimental.nupkgproj
@@ -2,11 +2,11 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <PackageDescription>Microsoft.ML.Experimental contains extension method for MLContext to access internal function to cancel ML pipeline.</PackageDescription>
+    <PackageDescription>Microsoft.ML.Experimental contains experimental work such extension methods to access internal methods.</PackageDescription>
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="../Microsoft.ML.Data/Microsoft.ML.Data.nupkgproj" />
+    <ProjectReference Include="../Microsoft.ML/Microsoft.ML.nupkgproj" />
   </ItemGroup>
 
 </Project>
--- a/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
+++ b/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
@@ -51,12 +51,12 @@ namespace Microsoft.ML.Runtime
     }
 
     [BestFriend]
-    internal interface ICancellableEnvironment
+    internal interface ICancelableEnvironment
     {
         /// <summary>
         /// Flag which indicate should we stop any code execution in this host.
         /// </summary>
-        bool IsCancelled { get; }
+        bool IsCanceled { get; }
     }
 
     /// <summary>
@@ -78,7 +78,7 @@ namespace Microsoft.ML.Runtime
     }
 
     [BestFriend]
-    internal interface ICancellableHost
+    internal interface ICancelableHost
     {
         /// <summary>
         /// Signal to stop exection in this host and all its children.
--- a/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
+++ b/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
@@ -72,7 +72,7 @@ namespace Microsoft.ML.Runtime
     internal interface ICancelableHost : IHost
     {
         /// <summary>
-        /// Signal to stop exection in this host and all its children.
+        /// Signal to stop exection in this host.
         /// </summary>
         void CancelExecution();
 
--- a/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
+++ b/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
@@ -69,7 +69,7 @@ namespace Microsoft.ML.Runtime
     }
 
     [BestFriend]
-    internal interface ICancelableHost
+    internal interface ICancelableHost : IHost
     {
         /// <summary>
         /// Signal to stop exection in this host and all its children.
--- a/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
+++ b/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
@@ -63,11 +63,6 @@ namespace Microsoft.ML.Runtime
         IHost Register(string name, int? seed = null, bool? verbose = null);
 
         /// <summary>
-        /// Flag which indicate should we stop any code execution in this host.
-        /// </summary>
-        bool IsCancelled { get; }
-
-        /// <summary>
         /// The catalog of loadable components (<see cref="LoadableClassAttribute"/>) that are available in this host.
         /// </summary>
         ComponentCatalog ComponentCatalog { get; }
@@ -85,11 +80,6 @@ namespace Microsoft.ML.Runtime
         /// generators are NOT thread safe.
         /// </summary>
         Random Rand { get; }
-
-        /// <summary>
-        /// Signal to stop exection in this host and all its children.
-        /// </summary>
-        void StopExecution();
     }
 
     /// <summary>
--- a/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
+++ b/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
@@ -50,15 +50,6 @@ namespace Microsoft.ML.Runtime
         }
     }
 
-    [BestFriend]
-    internal interface ICancelableEnvironment
-    {
-        /// <summary>
-        /// Flag which indicate should we stop any code execution in this host.
-        /// </summary>
-        bool IsCanceled { get; }
-    }
-
     /// <summary>
     /// The host environment interface creates hosts for components. Note that the methods of
     /// this interface should be called from the main thread for the environment. To get an environment
@@ -83,7 +74,12 @@ namespace Microsoft.ML.Runtime
         /// <summary>
         /// Signal to stop exection in this host and all its children.
         /// </summary>
-        void StopExecution();
+        void CancelExecution();
+
+        /// <summary>
+        /// Flag which indicate should we stop any code execution in this host.
+        /// </summary>
+        bool IsCanceled { get; }
     }
 
     /// <summary>
--- a/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
+++ b/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
@@ -69,15 +69,15 @@ namespace Microsoft.ML.Runtime
     }
 
     [BestFriend]
-    internal interface ICancelableHost : IHost
+    internal interface ICancelable
     {
         /// <summary>
-        /// Signal to stop exection in this host.
+        /// Signal to stop exection in all the hosts.
         /// </summary>
         void CancelExecution();
 
         /// <summary>
-        /// Flag which indicate should we stop any code execution in this host.
+        /// Flag which indicates host execution has been stopped.
         /// </summary>
         bool IsCanceled { get; }
     }
--- a/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
+++ b/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
@@ -50,6 +50,15 @@ namespace Microsoft.ML.Runtime
         }
     }
 
+    [BestFriend]
+    internal interface ICancellableEnvironment
+    {
+        /// <summary>
+        /// Flag which indicate should we stop any code execution in this host.
+        /// </summary>
+        bool IsCancelled { get; }
+    }
+
     /// <summary>
     /// The host environment interface creates hosts for components. Note that the methods of
     /// this interface should be called from the main thread for the environment. To get an environment
@@ -66,6 +75,15 @@ namespace Microsoft.ML.Runtime
         /// The catalog of loadable components (<see cref="LoadableClassAttribute"/>) that are available in this host.
         /// </summary>
         ComponentCatalog ComponentCatalog { get; }
+    }
+
+    [BestFriend]
+    internal interface ICancellableHost
+    {
+        /// <summary>
+        /// Signal to stop exection in this host and all its children.
+        /// </summary>
+        void StopExecution();
     }
 
     /// <summary>
--- a/src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
+++ b/src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
@@ -101,7 +101,7 @@ namespace Microsoft.ML.Runtime
         /// to provide their own host class that derives from this class.
         /// This encapsulates the random number generator and name information.
         /// </summary>
-        public abstract class HostBase : HostEnvironmentBase<TEnv>, IHost
+        public abstract class HostBase : HostEnvironmentBase<TEnv>, IHost, ICancellableEnvironment, ICancellableHost
         {
             public override int Depth { get; }
 
@@ -110,10 +110,28 @@ namespace Microsoft.ML.Runtime
             // We don't have dispose mechanism for hosts, so to let GC collect children hosts we make them WeakReference.
             private readonly List<WeakReference<IHost>> _children;
 
+            public bool IsCancelled { get; set; }
+
             public HostBase(HostEnvironmentBase<TEnv> source, string shortName, string parentFullName, Random rand, bool verbose)
                 : base(source, rand, verbose, shortName, parentFullName)
             {
                 Depth = source.Depth + 1;
+                _children = new List<WeakReference<IHost>>();
+            }
+
+            public void StopExecution()
+            {
+                lock (_cancelLock)
+                {
+                    IsCancelled = true;
+                    foreach (var child in _children)
+                    {
+                        if (child.TryGetTarget(out IHost host))
+                            if (host is ICancellableHost)
+                                ((ICancellableHost)host).StopExecution();
+                    }
+                    _children.Clear();
+                }
             }
 
             public new IHost Register(string name, int? seed = null, bool? verbose = null)
@@ -124,7 +142,7 @@ namespace Microsoft.ML.Runtime
                 {
                     Random rand = (seed.HasValue) ? RandomUtils.Create(seed.Value) : RandomUtils.Create(_rand);
                     host = RegisterCore(this, name, Master?.FullName, rand, verbose ?? Verbose);
-                    if (!IsCancelled)
+                    if (host is ICancellableEnvironment && !((ICancellableEnvironment)host).IsCancelled)
                         _children.Add(new WeakReference<IHost>(host));
                 }
                 return host;
--- a/src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
+++ b/src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
@@ -342,8 +342,6 @@ namespace Microsoft.ML.Runtime
 
         protected readonly ProgressReporting.ProgressTracker ProgressTracker;
 
-
-
         public ComponentCatalog ComponentCatalog { get; }
 
         public override int Depth => 0;
--- a/src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
+++ b/src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
@@ -98,16 +98,21 @@ namespace Microsoft.ML.Runtime
     {
         public bool IsCanceled { get; protected set; }
 
-        private List<IHost> _hosts;
+        private readonly List<IHost> _hosts;
+
+        private readonly object _cancelEnvLock;
 
         [BestFriend]
         internal void CancelExecutionHosts()
         {
-            foreach (var host in _hosts)
-                if (host is ICancelableHost)
-                    ((ICancelableHost)host).CancelExecution();
+            lock (_cancelEnvLock)
+            {
+                foreach (var host in _hosts)
+                    if (host is ICancelableHost)
+                        ((ICancelableHost)host).CancelExecution();
 
-            _hosts.Clear();
+                _hosts.Clear();
+            }
         }
 
         /// <summary>
@@ -370,6 +375,7 @@ namespace Microsoft.ML.Runtime
             ListenerDict = new ConcurrentDictionary<Type, Dispatcher>();
             ProgressTracker = new ProgressReporting.ProgressTracker(this);
             _cancelLock = new object();
+            _cancelEnvLock = new object();
             Root = this as TEnv;
             ComponentCatalog = new ComponentCatalog();
             _hosts = new List<IHost>();
@@ -386,6 +392,7 @@ namespace Microsoft.ML.Runtime
             Contracts.CheckValueOrNull(rand);
             _rand = rand ?? RandomUtils.Create();
             _cancelLock = new object();
+            _cancelEnvLock = new object();
             _hosts = new List<IHost>();
 
             // This fork shares some stuff with the master.
--- a/src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
+++ b/src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
@@ -103,7 +103,7 @@ namespace Microsoft.ML.Runtime
         /// to provide their own host class that derives from this class.
         /// This encapsulates the random number generator and name information.
         /// </summary>
-        public abstract class HostBase : HostEnvironmentBase<TEnv>, IHost, ICancelableEnvironment, ICancelableHost
+        public abstract class HostBase : HostEnvironmentBase<TEnv>, IHost, ICancelableHost
         {
             public override int Depth { get; }
 
@@ -119,7 +119,7 @@ namespace Microsoft.ML.Runtime
                 _children = new List<WeakReference<IHost>>();
             }
 
-            public void StopExecution()
+            public void CancelExecution()
             {
                 lock (_cancelLock)
                 {
@@ -128,7 +128,7 @@ namespace Microsoft.ML.Runtime
                     {
                         if (child.TryGetTarget(out IHost host))
                             if (host is ICancelableHost cancelableHost)
-                                cancelableHost.StopExecution();
+                                cancelableHost.CancelExecution();
                     }
                     _children.Clear();
                 }
@@ -142,7 +142,7 @@ namespace Microsoft.ML.Runtime
                 {
                     Random rand = (seed.HasValue) ? RandomUtils.Create(seed.Value) : RandomUtils.Create(_rand);
                     host = RegisterCore(this, name, Master?.FullName, rand, verbose ?? Verbose);
-                    if (host is ICancelableEnvironment cancelableHost && !cancelableHost.IsCanceled)
+                    if (host is ICancelableHost cancelableHost && !cancelableHost.IsCanceled)
                         _children.Add(new WeakReference<IHost>(host));
                 }
                 return host;
--- a/src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
+++ b/src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
@@ -114,21 +114,6 @@ namespace Microsoft.ML.Runtime
                 : base(source, rand, verbose, shortName, parentFullName)
             {
                 Depth = source.Depth + 1;
-                _children = new List<WeakReference<IHost>>();
-            }
-
-            public void StopExecution()
-            {
-                lock (_cancelLock)
-                {
-                    IsCancelled = true;
-                    foreach (var child in _children)
-                    {
-                        if (child.TryGetTarget(out IHost host))
-                            host.StopExecution();
-                    }
-                    _children.Clear();
-                }
             }
 
             public new IHost Register(string name, int? seed = null, bool? verbose = null)
@@ -339,7 +324,7 @@ namespace Microsoft.ML.Runtime
 
         protected readonly ProgressReporting.ProgressTracker ProgressTracker;
 
-        public bool IsCancelled { get; protected set; }
+
 
         public ComponentCatalog ComponentCatalog { get; }
 
--- a/src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
+++ b/src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
@@ -96,6 +96,8 @@ namespace Microsoft.ML.Runtime
     internal abstract class HostEnvironmentBase<TEnv> : ChannelProviderBase, IHostEnvironment, IChannelProvider
         where TEnv : HostEnvironmentBase<TEnv>
     {
+        public bool IsCancelled { get; set; }
+
         /// <summary>
         /// Base class for hosts. Classes derived from  <see cref="HostEnvironmentBase{THostEnvironmentBase}"/> may choose
         /// to provide their own host class that derives from this class.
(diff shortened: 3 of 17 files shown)


src/Microsoft.ML.Core/Data/IHostEnvironment.cs
```diff
@@ -68,11 +68,6 @@ public interface IHostEnvironment : IChannelProvider, IProgressChannelProvider
         /// </summary>
         int ConcurrencyFactor { get; }
 
-        /// <summary>
-        /// Flag which indicate should we stop any code execution in this host.
-        /// </summary>
-        bool IsCancelled { get; }
```
glebuk: >IsCancelled [](start = 13, length = 11)

I'm fine with removing this infra for now as nothing uses it.  We should consider, in the future to allow iterative components to be cancellable (Boosting, linear, etc)

codemzs: Agreed, I have updated the issue with the same.

glebuk:
>     public interface IHostEnvironment : IChannelProvider, IProgressChannelProvider

Seems like there are potential scenarios where this feature might be critical, perhaps lets first evaluate various options on enable it short term and have a stable design for the long term

---
Refers to: src/Microsoft.ML.Core/Data/IHostEnvironment.cs:57 in b81c3f3. [](commit_id = b81c3f37ad625251a2b927d2d0227fb885e755cf, deletion_comment = False)

pkg/Microsoft.ML.Experimental/Microsoft.ML.Experimental.nupkgproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk" DefaultTargets="Pack">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageDescription>Microsoft.ML.Experimental contains extension method for MLContext to access internal function to cancel ML pipeline.</PackageDescription>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../Microsoft.ML.Data/Microsoft.ML.Data.nupkgproj" />
```
eerhardt: I don't think this .nupkgproj exists. I think you meant `Microsoft.ML`.

pkg/Microsoft.ML.Experimental/Microsoft.ML.Experimental.nupkgproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk" DefaultTargets="Pack">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageDescription>Microsoft.ML.Experimental contains extension method for MLContext to access internal function to cancel ML pipeline.</PackageDescription>
```
eerhardt: I assume there are more things we are going to want to add to this package - so this description seems a little to specific to me.

src/Microsoft.ML.Core/Data/IHostEnvironment.cs
```diff
@@ -50,6 +50,15 @@ public static IFileHandle CreateOutputFile(this IHostEnvironment env, string pat
         }
     }
 
+    [BestFriend]
+    internal interface ICancellableEnvironment
+    {
+        /// <summary>
+        /// Flag which indicate should we stop any code execution in this host.
+        /// </summary>
+        bool IsCancelled { get; }
```
eerhardt: `Task` uses the `Canceled` spelling (single `L`):

[https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.iscanceled?view=netframework-4.7.2](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.iscanceled?view=netframework-4.7.2)

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
```diff
@@ -125,7 +127,8 @@ public void StopExecution()
                     foreach (var child in _children)
                     {
                         if (child.TryGetTarget(out IHost host))
-                            host.StopExecution();
+                            if (host is ICancellableHost)
```
eerhardt: use pattern matching here:

```c#
if (host is ICancellableHost cancellableHost)
{
    cancellableHost.StopExecution();
}
```

eerhardt: and below.

---
In reply to: [264859870](https://github.com/dotnet/machinelearning/pull/2797#discussion_r264859870) [](ancestors = 264859870)

src/Microsoft.ML.Core/Utilities/Contracts.cs
```diff
@@ -737,22 +737,22 @@ public static void CheckIO(this IExceptionContext ctx, bool f, string msg)
...
-        /// Check state of the host and throw exception if host marked to stop all exection.
-        /// </summary>
-        public static void CheckAlive(this IHostEnvironment env)
+    /// <summary>
+    /// Check state of the host and throw exception if host marked to stop all exection.
+    /// </summary>
+    public static void CheckAlive(this IHostEnvironment env)
         {
-            if (env.IsCancelled)
+            if ((env is ICancellableEnvironment) && (env as ICancellableEnvironment).IsCancelled)
```
eerhardt: Use pattern matching.

src/Microsoft.ML.Experimental/Microsoft.ML.Experimental.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <IncludeInPackage>Microsoft.ML.Experimental</IncludeInPackage>
+    <RootNamespace>Microsoft.ML.Experimental</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Folder Include="Properties\" />
```
eerhardt: This can be deleted.

src/Microsoft.ML.Experimental/Microsoft.ML.Experimental.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <IncludeInPackage>Microsoft.ML.Experimental</IncludeInPackage>
+    <RootNamespace>Microsoft.ML.Experimental</RootNamespace>
```
eerhardt: This line in unnecessary, and can be deleted.

eerhardt:
We should add some unit tests for this capability.

src/Microsoft.ML.Data/MLContext.cs
```diff
@@ -104,12 +108,25 @@ private void ProcessMessage(IMessageSource source, ChannelMessage message)
...
+        }
+
         IChannel IChannelProvider.Start(string name) => _env.Start(name);
         IPipe<TMessage> IChannelProvider.StartPipe<TMessage>(string name) => _env.StartPipe<TMessage>(name);
         IProgressChannel IProgressChannelProvider.StartProgressChannel(string name) => _env.StartProgressChannel(name);
+
+        [BestFriend]
+        internal void StopExecution()
+        {
+            foreach(var host in _hosts)
```
eerhardt: Why is remembering the `_hosts` necessary?  Why can't we just call `_env.StopExecution()` here?

src/Microsoft.ML.Experimental/Experimental.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Microsoft.ML.Experimental
+{
+    public static class Experimental
+    {
+        /// <summary>
+        /// Stop the exeuction of pipeline in <see cref="MLContext"/>
+        /// </summary>
+        /// <param name="ctx"></param>
+        public static void StopExecution(this MLContext ctx) => ctx.StopExecution();
```
eerhardt: Do we also need to know if the `MLContext` has been canceled?

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
```diff
@@ -96,12 +96,14 @@ internal interface IMessageSource
     internal abstract class HostEnvironmentBase<TEnv> : ChannelProviderBase, IHostEnvironment, IChannelProvider
         where TEnv : HostEnvironmentBase<TEnv>
     {
+        public bool IsCancelled { get; set; }
```
eerhardt: I don't think this should be publicly settable.

codemzs:
There is a unit-test that checks for cancel signal but I think what you are referring to is testing cancellation on a pipeline. For that to happen we will need to implement checks in our components for this cancellation signal and cancel the pipeline but this is not within the scope of this PR.

---
In reply to: [472159903](https://github.com/dotnet/machinelearning/pull/2797#issuecomment-472159903) [](ancestors = 472159903)

src/Microsoft.ML.Data/MLContext.cs
codemzs: StopExecution is implemented by HostBase not EnvironmentBase.

---
In reply to: [264861094](https://github.com/dotnet/machinelearning/pull/2797#discussion_r264861094) [](ancestors = 264861094)

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
codemzs: Its within an internal class so its internal right?

---
In reply to: [264861769](https://github.com/dotnet/machinelearning/pull/2797#discussion_r264861769) [](ancestors = 264861769)

eerhardt:
You are adding a new public API - 

```
public static class Experimental
    {
        /// <summary>
        /// Stop the exeuction of pipeline in <see cref="MLContext"/>
        /// </summary>
        /// <param name="ctx"></param>
        public static void StopExecution(this MLContext ctx) => ctx.StopExecution();
```

Adding unit tests for that new public API is within the scope of this PR.

src/Microsoft.ML.Experimental/Experimental.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.ML.Experimental
+{
+    public static class Experimental
+    {
+        /// <summary>
+        /// Stop the exeuction of pipeline in <see cref="MLContext"/>
```
eerhardt: type-o `exeuction`

src/Microsoft.ML.Experimental/Experimental.cs
```diff
@@ -0,0 +1,15 @@
...
+        /// </summary>
+        /// <param name="ctx"></param>
```
eerhardt: This should be filled out. It is a public API.

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
eerhardt: Correct. And we don't want internal code to be able to set this property. Only this class (or subclasses) should be able to set it.

codemzs: Yep, made it private set.

---
In reply to: [264863680](https://github.com/dotnet/machinelearning/pull/2797#discussion_r264863680) [](ancestors = 264863680)

src/Microsoft.ML.Core/Data/IHostEnvironment.cs
codemzs: When in the USA ...

---
In reply to: [264859366](https://github.com/dotnet/machinelearning/pull/2797#discussion_r264859366) [](ancestors = 264859366)

src/Microsoft.ML.Experimental/Experimental.cs
codemzs: Right now this is a synchronous operation that just turns on a flag for cancellation. Later when we implement mechanisms that actually make use of this flag to cancel a pipeline then we can check if a pipeline has been canceled may be by checking for an exception or something,

---
In reply to: [264861512](https://github.com/dotnet/machinelearning/pull/2797#discussion_r264861512) [](ancestors = 264861512)

codemzs:
Done.

---
In reply to: [472161675](https://github.com/dotnet/machinelearning/pull/2797#issuecomment-472161675) [](ancestors = 472161675)

src/Microsoft.ML.Core/Data/IHostEnvironment.cs
```diff
@@ -50,6 +50,15 @@ public static IFileHandle CreateOutputFile(this IHostEnvironment env, string pat
         }
     }
 
+    [BestFriend]
+    internal interface ICancelableEnvironment
```
eerhardt: Why do we have 2 interfaces? `ICancelableEnvironment` and `ICancelableHost`? Can't there just be a single interface?

src/Microsoft.ML.Core/Data/IHostEnvironment.cs
```diff
@@ -63,14 +72,18 @@ public interface IHostEnvironment : IChannelProvider, IProgressChannelProvider
...
 
+    [BestFriend]
+    internal interface ICancelableHost
+    {
         /// <summary>
-        /// The catalog of loadable components (<see cref="LoadableClassAttribute"/>) that are available in this host.
+        /// Signal to stop exection in this host and all its children.
         /// </summary>
-        ComponentCatalog ComponentCatalog { get; }
+        void StopExecution();
```
eerhardt: (minor) it's a bit of inconsistent naming - this is a "cancelable" thing, but it doesn't have a "Cancel" method - it has `StopExecution()`.

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
```diff
@@ -96,12 +96,14 @@ internal interface IMessageSource
     internal abstract class HostEnvironmentBase<TEnv> : ChannelProviderBase, IHostEnvironment, IChannelProvider
         where TEnv : HostEnvironmentBase<TEnv>
     {
+        public bool IsCanceled { get; protected set; }
```
eerhardt: Why is this moved from all the other property definitions?

src/Microsoft.ML.Core/Utilities/Contracts.cs
```diff
@@ -737,22 +737,22 @@ public static void CheckIO(this IExceptionContext ctx, bool f, string msg)
...
         {
             if (!f)
                 throw ExceptIO(ctx, msg, args);
         }
 #if !CPUMATH_INFRASTRUCTURE
-        /// <summary>
-        /// Check state of the host and throw exception if host marked to stop all exection.
-        /// </summary>
-        public static void CheckAlive(this IHostEnvironment env)
+    /// <summary>
```
eerhardt: There are some whitespace mismatches here.

src/Microsoft.ML.Data/MLContext.cs
```diff
@@ -104,12 +108,27 @@ private void ProcessMessage(IMessageSource source, ChannelMessage message)
             log(this, new LoggingEventArgs(msg));
         }
 
-        bool IHostEnvironment.IsCancelled => _env.IsCancelled;
         string IExceptionContext.ContextDescription => _env.ContextDescription;
         TException IExceptionContext.Process<TException>(TException ex) => _env.Process(ex);
-        IHost IHostEnvironment.Register(string name, int? seed, bool? verbose) => _env.Register(name, seed, verbose);
+        IHost IHostEnvironment.Register(string name, int? seed, bool? verbose)
+        {
+            var host = _env.Register(name, seed, verbose);
```
eerhardt: Does this Register actually get called? Since we are passing the inner `_env` to all of the Catalogs above, that is the object that all the code is calling `Register` on right? There is a split here between `MLContext` and `HostEnivronmentBase`.

src/Microsoft.ML.Data/MLContext.cs
```diff
@@ -104,12 +108,27 @@ private void ProcessMessage(IMessageSource source, ChannelMessage message)
...
         IProgressChannel IProgressChannelProvider.StartProgressChannel(string name) => _env.StartProgressChannel(name);
+
+        [BestFriend]
+        internal void StopExecution()
+        {
+            foreach(var host in _hosts)
+                if (host is ICancelableHost)
+                    ((ICancelableHost)host).StopExecution();
+
+            _hosts.RemoveAll(h => true);
```
eerhardt: `_hosts.Clear()`

src/Microsoft.ML.Data/MLContext.cs
eerhardt: This doesn't make sense to me. The `HostEnvironmentBase` already remembers all of its `_children`. We should put the `StopExecution()` there, and MLContext doesn't need to do any of this. It can just call `_env.StopExecution()`.

---
In reply to: [264862395](https://github.com/dotnet/machinelearning/pull/2797#discussion_r264862395) [](ancestors = 264862395,264861094)

test/Microsoft.ML.Core.Tests/UnitTests/TestHosts.cs
```diff
@@ -55,23 +55,62 @@ public void TestCancellation()
...
+            IHostEnvironment env = new MLContext(seed: 42);
+            var mainHost = env.Register("Main");
+            var children = new ConcurrentDictionary<IHost, List<IHost>>();
+            var hosts = new BlockingCollection<Tuple<IHost, int>>();
+            hosts.Add(new Tuple<IHost, int>(mainHost.Register("1"), 1));
+            hosts.Add(new Tuple<IHost, int>(mainHost.Register("2"), 1));
+            hosts.Add(new Tuple<IHost, int>(mainHost.Register("3"), 1));
+            hosts.Add(new Tuple<IHost, int>(mainHost.Register("4"), 1));
+            hosts.Add(new Tuple<IHost, int>(mainHost.Register("5"), 1));
+            Random rand = new Random();
```
eerhardt: Why are we using randomness and multiple threads here?  Why can't we just register a few hosts to the MLContext, and then call `StopExecution()` and ensure all the hosts have `IsCanceled` set to true?

src/Microsoft.ML.Core/Data/IHostEnvironment.cs
```diff
@@ -50,6 +50,15 @@ public static IFileHandle CreateOutputFile(this IHostEnvironment env, string pat
         }
     }
 
+    [BestFriend]
+    internal interface ICancelableEnvironment
+    {
+        /// <summary>
+        /// Flag which indicate should we stop any code execution in this host.
```
glebuk: >code execution in this host [](start = 51, length = 27)

in this environment?

src/Microsoft.ML.Data/MLContext.cs
codemzs: Yes, it gets called. From an environment you can create multiple hosts and then from each host you can create multiple hosts. I have verified this with the test that I wrote.

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
codemzs: Didn't seem the need to include it there. The way I have it now is sufficient to cancel components.

src/Microsoft.ML.Data/MLContext.cs
eerhardt: Your test calls Register directly on the MLContext, which is not what users are going to do.

Users are going to write code like:

```C#

mlContext.Regression.Trainers.Sdca();
```

If you walk that code path above - `mlContext.Regression` is:

https://github.com/dotnet/machinelearning/blob/b6f94bcb86b4fa25aa0efed2ecc036ae981c98e0/src/Microsoft.ML.Data/MLContext.cs#L86

Which gets the underlying `_env` passed into it. NOT the `MLContext`. Therefore, when Sdca calls `Register` it will call it on the inner `_env`, not on the `MLContext`.

codemzs: I expect the user to cancel via MLContext. I will sync offline with on you on this.

src/Microsoft.ML.Core/Data/IHostEnvironment.cs
```diff
@@ -63,14 +63,23 @@ public interface IHostEnvironment : IChannelProvider, IProgressChannelProvider
...
         /// <summary>
-        /// Flag which indicate should we stop any code execution in this host.
+        /// The catalog of loadable components (<see cref="LoadableClassAttribute"/>) that are available in this host.
         /// </summary>
-        bool IsCancelled { get; }
+        ComponentCatalog ComponentCatalog { get; }
+    }
 
+    [BestFriend]
+    internal interface ICancelableHost
```
glebuk: >ICancelableHost [](start = 23, length = 15)

Shouldn;t it inherit from IHost? (for example, it should include Rand)

codemzs: No, it shouldn't.

src/Microsoft.ML.Core/Data/IHostEnvironment.cs
```diff
@@ -63,14 +63,23 @@ public interface IHostEnvironment : IChannelProvider, IProgressChannelProvider
...
+    {
         /// <summary>
-        /// The catalog of loadable components (<see cref="LoadableClassAttribute"/>) that are available in this host.
+        /// Signal to stop exection in this host and all its children.
```
glebuk: >host and all its children [](start = 44, length = 25)

but will it cancel parents, siblings and environment?

codemzs: Hmm...I added a test that verifies this. Did you get a chance to look at it?

src/Microsoft.ML.Data/MLContext.cs
codemzs: Done.

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
```diff
@@ -96,12 +96,26 @@ internal interface IMessageSource
...
+
+        private List<IHost> _hosts;
```
glebuk: >private [](start = 8, length = 7)

should this be readonly?

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
```diff
@@ -96,12 +96,26 @@ internal interface IMessageSource
...
+
+        [BestFriend]
+        internal void CancelExecutionHosts()
+        {
+            foreach (var host in _hosts)
+                if (host is ICancelableHost)
+                    ((ICancelableHost)host).CancelExecution();
+
+            _hosts.Clear();
```
glebuk: >_hosts [](start = 12, length = 6)

Will this class be used from multiple threads? If so then it is not currently threadsafe.

src/Microsoft.ML.Experimental/Experimental.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.ML.Experimental
+{
+    public static class Experimental
```
glebuk: >Experimental [](start = 24, length = 12)

Rname to MLContextExtensions. Let's not use up the "experimental" class name. We might have several experimental features.

src/Microsoft.ML.Experimental/Experimental.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        /// <summary>
+        /// Stop the execution of pipeline in <see cref="MLContext"/>
+        /// </summary>
+        /// <param name="ctx"><see cref="MLContext"/> reference.</param>
+        public static void StopExecution(this MLContext ctx) => ctx.CancelExecution();
```
glebuk: >StopExecution [](start = 27, length = 13)

consistent naming? CancelExecution()

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
codemzs: Why is it not thread safe? 

---
In reply to: [266555335](https://github.com/dotnet/machinelearning/pull/2797#discussion_r266555335) [](ancestors = 266555335)

codemzs:
>             string shortName = null, string parentFullName = null)

Not sure, doesn't seem related to my change.

---
In reply to: [474019719](https://github.com/dotnet/machinelearning/pull/2797#issuecomment-474019719) [](ancestors = 474019719)

---
Refers to: src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs:382 in c4a3415. [](commit_id = c4a3415587b941f433b602aba33686a481299e17, deletion_comment = False)

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
glebuk: You can have a race between two instances of CancelExecutionHost - if the _hosts is cleared while the other thread is enumerating you'll get an exception, 
Alternatively there can be a race between CancelExecutionHost() and Register() when item is removed but not cancelled.

---
In reply to: [266558718](https://github.com/dotnet/machinelearning/pull/2797#discussion_r266558718) [](ancestors = 266558718,266555335)

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
```diff
@@ -385,7 +407,9 @@ public IHost Register(string name, int? seed = null, bool? verbose = null)
         {
             Contracts.CheckNonEmpty(name, nameof(name));
             Random rand = (seed.HasValue) ? RandomUtils.Create(seed.Value) : RandomUtils.Create(_rand);
-            return RegisterCore(this, name, Master?.FullName, rand, verbose ?? Verbose);
+            var host = RegisterCore(this, name, Master?.FullName, rand, verbose ?? Verbose);
+            _hosts.Add(host);
```
glebuk: >_hosts [](start = 12, length = 6)

don't  you need to lock that too?

src/Microsoft.ML.Data/MLContext.cs
```diff
@@ -72,6 +73,8 @@ public sealed class MLContext : IHostEnvironment
         /// </summary>
         public ComponentCatalog ComponentCatalog => _env.ComponentCatalog;
 
+        private List<IHost> _hosts;
```
eerhardt: I don't see this field being used. Can it be removed?

test/Microsoft.ML.Core.Tests/UnitTests/TestHosts.cs
```diff
@@ -55,23 +55,57 @@ public void TestCancellation()
...
+            IHostEnvironment env = new MLContext(seed: 42);
+            var mainHost = env.Register("Main");
+            var children = new ConcurrentDictionary<IHost, List<IHost>>();
+            var hosts = new BlockingCollection<Tuple<IHost, int>>();
+            hosts.Add(new Tuple<IHost, int>(mainHost.Register("1"), 1));
+            hosts.Add(new Tuple<IHost, int>(mainHost.Register("2"), 1));
+            hosts.Add(new Tuple<IHost, int>(mainHost.Register("3"), 1));
+            hosts.Add(new Tuple<IHost, int>(mainHost.Register("4"), 1));
+            hosts.Add(new Tuple<IHost, int>(mainHost.Register("5"), 1));
+            var addThread = new Thread(
```
eerhardt: Why are we using a different thread here? I don't see any value in it. You are starting a thread, and then immediately joining it. You might as well just run the `for` loop below on the same thread.

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
```diff
@@ -96,12 +96,31 @@ internal interface IMessageSource
     internal abstract class HostEnvironmentBase<TEnv> : ChannelProviderBase, IHostEnvironment, IChannelProvider
         where TEnv : HostEnvironmentBase<TEnv>
     {
+        public bool IsCanceled { get; protected set; }
```
eerhardt: You are splitting the properties and fields of this class up. They were all declared after the nested types before your change - right above the constructor. Now you have some up here, and most down there. Please keep them grouped together.

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
```diff
@@ -96,12 +96,31 @@ internal interface IMessageSource
...
+        private readonly object _cancelEnvLock;
+
+        [BestFriend]
+        internal void CancelExecutionHosts()
+        {
+            lock (_cancelEnvLock)
+            {
+                foreach (var host in _hosts)
+                    if (host is ICancelableHost)
+                        ((ICancelableHost)host).CancelExecution();
```
eerhardt: You can use pattern matching here.

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
```diff
@@ -357,8 +375,10 @@ public void RemoveListener(Action<IMessageSource, TMessage> listenerFunc)
             ListenerDict = new ConcurrentDictionary<Type, Dispatcher>();
             ProgressTracker = new ProgressReporting.ProgressTracker(this);
             _cancelLock = new object();
+            _cancelEnvLock = new object();
```
eerhardt: This "smells" to me that we have both `_cancelLock` and `_cancelEnvLock`.

I think it would make more sense if we moved the `HostBase._children` field "up" to the HostEnvironmentBase class. Then we only need a single `_cancelLock` object. And we only need to a single "children" list.

codemzs: You have a good smell.

---
In reply to: [266936547](https://github.com/dotnet/machinelearning/pull/2797#discussion_r266936547) [](ancestors = 266936547)

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
```diff
@@ -96,36 +96,47 @@ internal interface IMessageSource
...
-
             public HostBase(HostEnvironmentBase<TEnv> source, string shortName, string parentFullName, Random rand, bool verbose)
                 : base(source, rand, verbose, shortName, parentFullName)
             {
                 Depth = source.Depth + 1;
-                _children = new List<WeakReference<IHost>>();
             }
 
-            public void StopExecution()
+            public void CancelExecution()
```
eerhardt: This appears to be duplicated with the base `CancelExecutionHosts`. Can we just use the base implementation?

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
```diff
@@ -93,9 +93,24 @@ internal interface IMessageSource
     /// query progress.
     /// </summary>
     [BestFriend]
-    internal abstract class HostEnvironmentBase<TEnv> : ChannelProviderBase, IHostEnvironment, IChannelProvider
+    internal abstract class HostEnvironmentBase<TEnv> : ChannelProviderBase, IHostEnvironment, IChannelProvider, ICancelable
         where TEnv : HostEnvironmentBase<TEnv>
     {
+        [BestFriend]
```
eerhardt: (nit) you probably don't need `[BestFriend]` here.

src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs
```diff
@@ -139,7 +136,7 @@ public new IHost Register(string name, int? seed = null, bool? verbose = null)
                 {
                     Random rand = (seed.HasValue) ? RandomUtils.Create(seed.Value) : RandomUtils.Create(_rand);
                     host = RegisterCore(this, name, Master?.FullName, rand, verbose ?? Verbose);
-                    if (!IsCancelled)
+                    if (host is ICancelable cancelableHost && !cancelableHost.IsCanceled)
```
eerhardt: Is this change correct? Shouldn't it still be `if (!IsCancelled)`?

src/Microsoft.ML.Data/MLContext.cs
```diff
@@ -104,12 +105,14 @@ private void ProcessMessage(IMessageSource source, ChannelMessage message)
...
-        bool IHostEnvironment.IsCancelled => _env.IsCancelled;
         string IExceptionContext.ContextDescription => _env.ContextDescription;
         TException IExceptionContext.Process<TException>(TException ex) => _env.Process(ex);
         IHost IHostEnvironment.Register(string name, int? seed, bool? verbose) => _env.Register(name, seed, verbose);
         IChannel IChannelProvider.Start(string name) => _env.Start(name);
         IPipe<TMessage> IChannelProvider.StartPipe<TMessage>(string name) => _env.StartPipe<TMessage>(name);
         IProgressChannel IProgressChannelProvider.StartProgressChannel(string name) => _env.StartProgressChannel(name);
+
+        [BestFriend]
+        internal void CancelExecution() => (_env as ICancelable).CancelExecution();
```
eerhardt: This should just be a straight cast.  `((ICancelable)_env).CancelExecution();`

PR #2800, opened 2019-02-28
Rename HalLearners assembly and nuget to Microsoft.ML.Mkl.Components.

### LINKED ISSUE #2756: Need to rename HalLearners assembly and nuget
The name `HalLearners` isn't a great public name. We should come up with a better name.

Current issues:

1. What does `Hal` mean?
2. We are using the term "trainers" instead of "learners" everywhere else.
3. This assembly contains more than just learners/trainers. It also has vector whitening, which isn't a trainer.

Possible suggestions to get the naming started:

1. `Microsoft.ML.HardwareAccelerated`
2. `Microsoft.ML.HardwareAccelerated.Components`
3. `Microsoft.ML.Mkl`
    - (we already have `Microsoft.ML.Mkl.Redist`. this would be an analog to `ML.TensorFlow` and `ML.TensorFlow.Redist`)
4. `Microsoft.ML.Mkl.Components`

@TomFinley @sfilipi @Ivanidzo4ka @shauheen @yaeldekel

codemzs:
fixes #2756


eerhardt:
>     <PackageDescription>ML.NET additional learners making use of hardware acceleration. They use Intel Mkl.</PackageDescription>

This description probably needs updating. As @TomFinley stated in the issue:

>"Hardware accelerated" is far too prone to misunderstanding in my opinion. Without knowing what I know, I think if I were to look at this library, and I saw, "hardware accelerated," my natural thought would be "OK, so, they are saying the rest the stuff isn't hardware accelerated." And, that would be incorrect while also making us look kind of silly.

---
Refers to: pkg/Microsoft.ML.Mkl.Components/Microsoft.ML.Mkl.Components.nupkgproj:5 in 93190e7. [](commit_id = 93190e711fd5ad0c64b8981b97484a4b75827bc3, deletion_comment = False)

eerhardt:
>     dependancy on MlNetMklDeps. From the users prospective, the HalLearners package depends on Microsoft.ML.Mkl.Redist. -->

`HalLearners` should be updated.

---
Refers to: src/Microsoft.ML.Mkl.Components/Microsoft.ML.Mkl.Components.csproj:18 in 93190e7. [](commit_id = 93190e711fd5ad0c64b8981b97484a4b75827bc3, deletion_comment = False)

eerhardt:
Won't you need to update:

https://github.com/dotnet/machinelearning/blob/129b47c489b91fe48ed622a6ec8c74aa64b3da7e/src/Native/build.proj#L92-L93

?

In general it would be good to do a "Find in files" search of the entire repo for the term "HalLearner" and make sure you aren't missing any.

TomFinley:
> [assembly: LoadableClass(typeof(OlsLinearRegressionModelParameters), null, typeof(SignatureLoadModel),

The same problem with assembly names will arise here as it arises in #2792, and will require basically the same solution.

---
Refers to: src/Microsoft.ML.Mkl.Components/OlsLinearRegression.cs:26 in 93190e7. [](commit_id = 93190e711fd5ad0c64b8981b97484a4b75827bc3, deletion_comment = False)

codemzs:
@eerhardt and @TomFinley can you please take another look?

Microsoft.ML.sln
```diff
@@ -100,7 +100,7 @@ Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Microsoft.ML.FSharp.Tests",
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ML.ImageAnalytics", "src\Microsoft.ML.ImageAnalytics\Microsoft.ML.ImageAnalytics.csproj", "{00E38F77-1E61-4CDF-8F97-1417D4E85053}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ML.HalLearners", "src\Microsoft.ML.HalLearners\Microsoft.ML.HalLearners.csproj", "{A7222F41-1CF0-47D9-B80C-B4D77B027A61}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ML.Mkl.Components", "src\Microsoft.ML.Mkl.Components\Microsoft.ML.Mkl.Components.csproj", "{A7222F41-1CF0-47D9-B80C-B4D77B027A61}"
```
sfilipi: >Microsoft.ML.Mkl.Components [](start = 53, length = 27)

i don't like the Mkl.Components name tbh, but nothing else comes to mind either..

docs/samples/Microsoft.ML.Samples/Dynamic/PermutationFeatureImportance/PFIHelper.cs
```diff
@@ -3,7 +3,7 @@
 using Microsoft.Data.DataView;
 using Microsoft.ML.Trainers;
 using Microsoft.ML.SamplesUtils;
-using Microsoft.ML.Trainers.HalLearners;
+using Microsoft.ML.Trainers.Mkl;
```
sfilipi: >Microsoft.ML.Trainers.Mkl [](start = 6, length = 25)

in the last API review we decided to not have sub namespaces in Microsoft.ML.Trainers. (except FastTree). 
So Microsoft.ML.Trainers.HalLearners; -> Microsoft.ML.Trainers

src/Microsoft.ML.Core/Data/ModelLoading.cs
```diff
@@ -21,6 +21,23 @@ internal sealed partial class ModelLoadContext : IDisposable
         public const string ModelStreamName = "Model.key";
         internal const string NameBinary = "Model.bin";
 
+        /// <summary>
+        /// Returns the new assembly name to maintain backward compatibility.
+        /// </summary>
+        private string ForwardedLoaderAssemblyName
```
sfilipi: >ForwardedLoaderAssemblyName [](start = 23, length = 27)

my $0.02 makes sense to do this post V1; i don't see it necessary to drag this legacy in the codebase.

src/Microsoft.ML.Mkl.Components/ComputeLRTrainingStdThroughHal.cs
```diff
@@ -5,13 +5,13 @@
 using System;
 using Microsoft.ML.Data;
 using Microsoft.ML.Internal.Utilities;
-using Microsoft.ML.Trainers.HalLearners;
+using Microsoft.ML.Trainers.Mkl;
```
sfilipi: >Mkl [](start = 28, length = 3)

commented in the samples; just Microsoft.ML.Trainers.

src/Microsoft.ML.Core/Data/ModelLoading.cs
```diff
@@ -259,7 +276,7 @@ private void EnsureLoaderAssemblyIsRegistered(ComponentCatalog catalog)
         {
             if (!string.IsNullOrEmpty(LoaderAssemblyName))
```
TomFinley: >LoaderAssemblyName [](start = 38, length = 18)

It may have been more sensible to do `ForwardedLoaderAssemblyName` here, but we can fix that later since the scenarios that would require that change can handle that change independently.

src/Microsoft.ML.Core/Data/ModelLoading.cs
TomFinley: That's reasonable, but there was a [discussion here](https://github.com/dotnet/machinelearning/pull/2792#issuecomment-468319048) where we decided that this sort of approach would be preferable from a practical point of view.

---
In reply to: [261829875](https://github.com/dotnet/machinelearning/pull/2800#discussion_r261829875) [](ancestors = 261829875)

docs/samples/Microsoft.ML.Samples/Dynamic/PermutationFeatureImportance/PFIHelper.cs
eerhardt: Agreed. I don’t think there are enough types to justify a new namespace.

codemzs: Ok, I’ll revert this change. Thanksz

Microsoft.ML.sln
codemzs: If you are not part of the solution then you are part of the ... just kidding! :)

Issue #2498, opened 2019-02-11
V1 Scenarios need to be covered by tests


rogancarr:
In issue #584, we laid out a set of scenarios that we'd like to cover for V1.0 of ML.NET. We need high-level functional tests to make sure that these work well in the 1.0 library.

Here is a list of tests that cover the scenarios. Let's use this issue as a top-level issue to track coverage of the APIs.

Category | Scenarios | Link to Test | Completed PR | Blocked by Issue
-- | -- | -- | -- | --
Data I/O| I can   use objects already in memory (as IEnumerable) as input to my ML   pipeline/experiment | [Link](https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Functional.Tests/DataIO.cs)  | #2518  |  
Data I/O | I can   use locally stored delimited files (.csv, .tsv, etc.) as input to my ML   pipeline/experiment | [Link](https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Functional.Tests/DataIO.cs)  | #2518  |    
Data I/O | I can   use locally stored binary files (.idv) as input to my ML   pipeline/experiment | [Link](https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Functional.Tests/DataIO.cs)  | #2518  |  
Data I/O | I can   go through any arbitrary data transformation / model training and save the   output to disk as a delimited file (.csv, .tsv, etc.). | [Link](https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Functional.Tests/DataIO.cs)  | #2518  |  
Data I/O | I can   go through any arbitrary data transformation / model training and save the   output to disk as a binary file (.idv). | [Link](https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Functional.Tests/DataIO.cs)  | #2518  |  
Data I/O | I can   go through any arbitrary data transformation / model training and convert the   output to an IEnumerable. | [Link](https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Functional.Tests/DataIO.cs)  | #2518  |  
Data I/O | I can   use data from a SQL database by reading it into memory or to disk using an   existing SQL reader and then use that as input to my ML pipeline/experiment | (May be a sample)  |   |  
Data   Transformation, Feature Engineering | I can   take an existing ONNX model and get predictions from it (as both final output   and as input to downstream pipelines) |   |   |  
[...]

PR #2803, opened 2019-02-28
Add V1 Scenario tests for data transformation

### LINKED ISSUE #2711: Create functional tests for all V1 Data Transformation scenarios
As laid out in #2498 , we need scenarios to cover the Data Transformation functionality we want fully supported in V1.

- Extensible transformation: It should be possible to write simple row-mapping transforms.
Examples: "I can add custom steps to my pipeline such as creating a new column that is the addition of two other columns, or easily add cosine similarity, without having to create my own build of ML.NET.
- I can modify settings in the TextFeaturizer to update the number of word-grams and char-grams used along with things like the normalization.	 	 	 
- I can apply normalization to the columns of my data

rogancarr:
Fixes #2711 
Towards #2498 

* Extensible transformation: It should be possible to write simple row-mapping transforms.
  Examples: "I can add custom steps to my pipeline such as creating a new column that is the addition of two other columns, or easily add cosine similarity, without having to create my own build of ML.NET.
* I can modify settings in the TextFeaturizer to update the number of word-grams and char-grams used along with things like the normalization.
* I can apply normalization to the columns of my data


test/Microsoft.ML.Functional.Tests/Datasets/FeatureColumn.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Microsoft.ML.Functional.Tests.Datasets
+{
+    /// <summary>
+    /// A class to hold a feature column.
+    /// </summary>
+    internal sealed class FeatureColumn
+    {
+        public float[] Features { get; set; }
+    }
```
sfilipi: maybe move it to the other file, the Iris.

test/Microsoft.ML.Functional.Tests/DataTransformation.cs
```diff
@@ -0,0 +1,243 @@
...
+                Assert.Equal(randomNumber1, row.Float1);
+                Assert.Equal(randomNumber2, row.Float2);
+            }
+        }
+
+        /// <summary>
+        /// Extensibility: Add a new column that is a function of other columns.
+        /// </summary>
+        [Fact]
+        void ExtensibilityAddAColumnAsAFunctionOfMultipleColumns()
```
sfilipi: >ExtensibilityAddAColumnAsAFunctionOfMultipleColumns [](start = 13, length = 51)

to me this test seems identical to the ExtensibilityAddingOneColumn. The function is different, but does that matter to the CustomMapping?

test/Microsoft.ML.Functional.Tests/DataTransformation.cs
```diff
@@ -0,0 +1,243 @@
...
+            var scoredData = model.Transform(data);
+            var metrics = mlContext.BinaryClassification.Evaluate(scoredData);
+
+            // Check that the metrics returned are valid.
+            Common.AssertMetrics(metrics);
+        }
+
+        /// <summary>
+        /// Extensibility: Featurize text using custom word-grams, char-grams, and normalization.
+        /// </summary>
```
artidoro: Could you change summary tag to match the test?

test/Microsoft.ML.Functional.Tests/DataTransformation.cs
artidoro: I agree, maybe I would only keep one of the two functions.

---
In reply to: [261830627](https://github.com/dotnet/machinelearning/pull/2803#discussion_r261830627) [](ancestors = 261830627)

test/Microsoft.ML.Functional.Tests/Datasets/FeatureColumn.cs
rogancarr: This is designed to be generic for tests against any features columns, not just the iris, so I'll keep it in its own file for now, similar to `FeatureContributionOutput'.

---
In reply to: [261830507](https://github.com/dotnet/machinelearning/pull/2803#discussion_r261830507) [](ancestors = 261830507)

PR #2807, opened 2019-03-01
Make accessor of linear coefficients unique to the public

### LINKED ISSUE #2763: LinearModelParameters have two ways to get Weights, should be one
we have this:
https://github.com/dotnet/machinelearning/blob/a0edc5c5724ed2895fd0c2fa5c2525d03de13455/src/Microsoft.ML.StandardLearners/Standard/LinearModelParameters.cs#L93
and we have this:
https://github.com/dotnet/machinelearning/blob/a0edc5c5724ed2895fd0c2fa5c2525d03de13455/src/Microsoft.ML.StandardLearners/Standard/LinearModelParameters.cs#L387

I would prefer to hide second one (and `IHaveFeatureWeights` interface) but it's up to discussion

wschin:
To fix #2763, we plan to hide one accessor in `LinearModelParameters` which is
```csharp
volid GetFeatureWeights(ref VBuffer<float> weights)
```
and make the its containing interface a best friend.
```csharp
internal interface IHaveFeatureWeights
```

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/FeatureContributionCalculationTransform.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/FeatureContributionCalculationTransform.cs
@@ -44,8 +44,7 @@ namespace Microsoft.ML.Samples.Dynamic
             var outData = featureContributionCalculator.Fit(scoredData).Transform(scoredData);
 
             // Let's extract the weights from the linear model to use as a comparison
-            var weights = new VBuffer<float>();
-            model.Model.GetFeatureWeights(ref weights);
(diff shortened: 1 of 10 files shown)


eerhardt:
>         public IReadOnlyList<float> Weights => new WeightsCollection(this);

While you are in here - can you make this property not create a new object every time it is accessed? Properties should be fast, and not allocate on every call.

We can lazily instantiate the object, if necessary.

---
Refers to: src/Microsoft.ML.StandardLearners/Standard/LinearModelParameters.cs:92 in e972912. [](commit_id = e972912de4589bcd05156816b993cb83896340e8, deletion_comment = False)

Issue #2785, opened 2019-02-28
GetExecutingAssemblyLocation in the DNNImageFeaturizers composes the wrong path


sfilipi:
Create a console project, add the MIcrosoft.ML, the Microsoft.Ml.OnnxTransformer and one of the Microsoft.ML.DnnImageFeaturizer packages. 

Try running the following snippet:

             // Downloading a few images, and an images.tsv file, which contains a list of the files from the dotnet/machinelearning/test/data/images/.
            // If you inspect the fileSystem, after running this line, an "images" folder will be created, containing 4 images, and a .tsv file
            // enumerating the images. 
            var imagesDataFile = Microsoft.ML.SamplesUtils.DatasetUtils.DownloadImages();

            // Preview of the content of the images.tsv file
            //
            // imagePath    imageType
            // tomato.bmp   tomato
            // banana.jpg   banana
            // hotdog.jpg   hotdog
            // tomato.jpg   tomato

            var data = mlContext.Data.CreateTextLoader(new TextLoader.Options()
            {
                Columns = new[]
                {
                        new TextLoader.Column("ImagePath", DataKind.String, 0),
                        new TextLoader.Column("Name", DataKind.String, 1),
                }
            }).Load(imagesDataFile);

            string codeBaseUri = typeof(Program).Assembly.CodeBase;
            string path = new Uri(codeBaseUri).AbsolutePath;
            var f =  Directory.GetParent(path).FullName;

            var imagesFolder = Path.GetDirectoryName(imagesDataFile);
            // Image loading pipeline. 
            var pipeline = mlContext.Transforms.LoadImages(imagesFolder, ("ImageObject", "ImagePath"))
                          .Append(mlContext.Transforms.ResizeImages("ImageObject", imageWidth: 224, imageHeight: 224))
                          .Append(mlContext.Transforms.ExtractPixels("Pixels", "ImageObject"))
                           //.Append(mlContext.Transforms.DnnFeaturizeImage("FeaturizedImage", m => m.ModelSelector.ResNet18(mlContext, m.OutputColumn, m.InputColumn, @"C:\Code\mlnet2\machinelearning-1\bin\AnyCPU.Debug\Microsoft.ML.Samples\netcoreapp2.1\DnnImageModels"), "Pixels"));
[...]

PR #2809, opened 2019-03-01
correcting path construction when Resnet projets are consumed as a nuget package

Anipik:
Fixes https://github.com/dotnet/machinelearning/issues/2785


src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AlexNetExtension.cs
```diff
@@ -21,7 +21,7 @@ public static class AlexNetExtension
         /// </summary>
         public static EstimatorChain<ColumnCopyingTransformer> AlexNet(this DnnImageModelSelector dnnModelContext, IHostEnvironment env, string outputColumnName, string inputColumnName)
         {
-            return AlexNet(dnnModelContext, env, outputColumnName, inputColumnName, Path.Combine(AssemblyPathHelpers.GetExecutingAssemblyLocation(), "DnnImageModels"));
+            return AlexNet(dnnModelContext, env, outputColumnName, inputColumnName, Path.Combine(Directory.GetCurrentDirectory(), "DnnImageModels"));
```
eerhardt: How does this work?

Anipik: So when we add a nuget reference to any of the resnets, these resnets are automatically copied to the output folder of that project, and they can be accessed through ```Directory.GetCurrentDirectory()```

The Assembly.codebase was not working when we were trying to use it as a nuget package because it was returing the system nuget package path i.e. C:\users\.nuget\packagename\version\

eerhardt: `Directory.GetCurrentDirectory()` is not going to be correct most of the time. There is nothing that says the process's current directory is the same as where the app is executing from.

Anipik: I modified it to use ```AppDomain.CurrentDomain.BaseDirectory```

eerhardt: have you tried it with a .NET Core 2.1 app when consuming the NuGet package? I don't think `
`AppDomain.CurrentDomain.BaseDirectory` is going to work in that scenario. The model will be in the NuGet cache, and the BaseDirectory is where the app is running from, which aren't the same.

Anipik: I tried it with 2.1 and its working.

eerhardt: Ah, I see why - I forgot that we added this part:

https://github.com/dotnet/machinelearning/blob/f78af891ea3339ec9ecabe2f35c965e8861b316d/pkg/common/DnnImageFeaturizer.props#L4-L7

That will copy the model to the output folder of the app.

PR #2815, opened 2019-03-01
[Tiny] Use string[] instead of IEnumerable<string> in column names

### LINKED ISSUE #2460: Modify API for FeaturizeText ?
In the MLContext for the text featurizer, the input column names are taken as a IEnumerable<string>

https://github.com/dotnet/machinelearning/blob/834e4715afa070f5e5c3628693682943b3f0bd6d/src/Microsoft.ML.Transforms/Text/TextCatalog.cs#L43-L46

https://github.com/dotnet/machinelearning/pull/2394#discussion_r254357200   recommends making them params instead.   

Should we modify this API ?

@sfilipi

### LINKED ISSUE #2801: TextFeaturizer API is non-standard
When specifying the input columns, `TextFeaturizer` specifies an `IEnumerable<string>`. The standard in the API is to specify a `string[]`.

This API should be updated to be consistent with the others. If we prefer `IEnumerable`, then we should switch all the other APIs.

wschin:
There are three possible ways to denote input column names, `string[]`, `IEnumerable<string>`, and `params`. I personally like `string[]` because

- `IEnumerable<string>` is not used elsewhere for the same purpose.
- `params` makes signature less typed and it forces input column names to be the last argument group.

Fix #2801, Fix #2460.

### DIFF
--- a/src/Microsoft.ML.Transforms/Text/TextCatalog.cs
+++ b/src/Microsoft.ML.Transforms/Text/TextCatalog.cs
@@ -45,7 +45,7 @@ namespace Microsoft.ML
         /// </example>
         public static TextFeaturizingEstimator FeaturizeText(this TransformsCatalog.TextTransforms catalog,
             string outputColumnName,
-            IEnumerable<string> inputColumnNames,
+            string[] inputColumnNames,
             TextFeaturizingEstimator.Options options)
             => new TextFeaturizingEstimator(Contracts.CheckRef(catalog, nameof(catalog)).GetEnvironment(),
                 outputColumnName, inputColumnNames, options);
--- a/src/Microsoft.ML.Transforms/Text/TextCatalog.cs
+++ b/src/Microsoft.ML.Transforms/Text/TextCatalog.cs
@@ -34,8 +34,8 @@ namespace Microsoft.ML
         /// </summary>
         /// <param name="catalog">The text-related transform's catalog.</param>
         /// <param name="outputColumnName">Name of the column resulting from the transformation of <paramref name="inputColumnNames"/>.</param>
-        /// <param name="inputColumnNames">Name of the columns to transform. If set to <see langword="null"/>, the value of the <paramref name="outputColumnName"/> will be used as source.</param>
         /// <param name="options">Advanced options to the algorithm.</param>
+        /// <param name="inputColumnNames">Name of the columns to transform. If set to <see langword="null"/>, the value of the <paramref name="outputColumnName"/> will be used as source.</param>
         /// <example>
         /// <format type="text/markdown">
         /// <![CDATA[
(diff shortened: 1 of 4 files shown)


TomFinley:
I understand that we wrote our samples this way, but writing it this way means we could not, for instance, take things that are the results of linq queries. It seems like we're just giving up capabilities for no particular benefit.

TomFinley:
I would prefer to have `IEnumerable` rather than arrays. It's a bit more flexible, allows linq queries, and other such things. Of course that's nothing that a user's `ToArray` can't fix, so maybe it's worthwhile.

src/Microsoft.ML.Transforms/Text/TextCatalog.cs
```diff
@@ -45,7 +45,7 @@ public static class TextCatalog
         /// </example>
         public static TextFeaturizingEstimator FeaturizeText(this TransformsCatalog.TextTransforms catalog,
             string outputColumnName,
-            IEnumerable<string> inputColumnNames,
```
sfilipi: >  IEnumerable<string> inputColumnNames, [](start = 10, length = 39)

my vote, and the original issues is to have this be the last parameter of the method, and be a params, to optimize for both cases where the users has one or more columns and wants to provied options. 

Having the IEumerable,  or with the string[] will require building an Enumerable/Array to hold just one element, which is unnecessary since params exists. 

Why would the users have to do either:

` var customized_pipeline = ml.Transforms.Text.FeaturizeText(customizedColumnName, options, new List<string> { "SentimentText" }, 
var customized_pipeline = ml.Transforms.Text.FeaturizeText(customizedColumnName, options, new string[] { "SentimentText" },`

when they can just go:

 ` var customized_pipeline = ml.Transforms.Text.FeaturizeText(customizedColumnName, options,  "SentimentText"  `

I do think the signature should be:

 public static TextFeaturizingEstimator FeaturizeText(this TransformsCatalog.TextTransforms catalog,
            string outputColumnName,
            TextFeaturizingEstimator.Options options
            params string[] inputColumnNames,
            )

and we are doing that everywhere else...

wschin: The assumption here is that output column is always single, but why can't it be multiple columns?

sfilipi: it can be both, hence params is the cleanest 

---
In reply to: [261709072](https://github.com/dotnet/machinelearning/pull/2815#discussion_r261709072) [](ancestors = 261709072)

wschin:
I switch to `params` now.

---
In reply to: [468742828](https://github.com/dotnet/machinelearning/pull/2815#issuecomment-468742828) [](ancestors = 468742828)

sharwell:
@wschin This commit appears to have broken the repo build

PR #2300, opened 2019-01-29
Hide much infrastructure in data

TomFinley:
Another of many steps towards #1602. Commits logically structured. No overall theme, just lots of hiding of individual components, most notably command line parsing, entry-point declarations, and other such things.

Issue #1602, opened 2018-11-12
Make non-essential members of Core and Data assembly internal


Zruty0:
#1519 made it possible to have members that are not public, but still accessible to other ML.NET assemblies. We now plan to use it to reduce the public surface of ML.NET assemblies substantially.

Namely, let's 'hide' the non-user-facing public members of `Microsoft.ML.Core` assembly

TomFinley:
In the next few days before we "finalize" the v1.0 RC API surface, we might conceivably hide one or two more things in these assemblies, but the usefulness of "catch-all" issues like this are less useful than they were back in the days when we were hiding hundreds or even thousands of items in our public surface with each PR. For that reason it is probably appropriate to close this issue at this time, since:

1. the core work it was meant to do has been almost entirely done,

2. even if it wasn't the opportunity for grand sweeping changes to the API in terms of public surface restriction are limited in the next few days,

3. and whatever little remains or can be done should have more specific issues.

Issue #2841, opened 2019-03-04
The curious case of TrainedWrapperEstimatorBase and friends


TomFinley:
So, while I was doing another round of internalization, one thing that I internalized was this. (Basically, something to handle the shimming from the now internal `IDataTransform` interface to the new `ITransformer` interface, during the regrettable situations -- thankfully few -- where such a thing is still necessary. So this:

https://github.com/dotnet/machinelearning/blob/63a7654edaddaf0eea6af98ea762b01594d083ee/src/Microsoft.ML.Data/DataLoadSave/TransformWrapper.cs#L20

and this

https://github.com/dotnet/machinelearning/blob/63a7654edaddaf0eea6af98ea762b01594d083ee/src/Microsoft.ML.Data/DataLoadSave/TransformWrapper.cs#L151

Now, that's all fine, but after doing the necessary work it seemed that I could delete the estimator wrapper entirely, but then I see this very intriguing note.

https://github.com/dotnet/machinelearning/blob/63a7654edaddaf0eea6af98ea762b01594d083ee/src/Microsoft.ML.Data/DataView/RowToRowMapperTransform.cs#L50-L56

I do not understand what is going on here. This is not essential -- everything here is internal -- but it seems at least odd. This seems to indicate that this method and the class I wanted to delete have something to do with each other, but as far as I can tell they have nothing to do with each other whatsoever. But then why the comment?

Anyway, there's clearly something odd going on. This is absolutely not critical, but I wanted to register an issue about the oddness, since the code underlying it passes my understanding.

rosebyte:
triage: we need to investigate if there is enough impact to justify the change.

PR #2842, opened 2019-03-05
Hide more of Microsoft.ML.Data

TomFinley:
Towards #1602. When performing what I hope is one of my final reviews of the public surface area of `Microsoft.ML.Data` I saw many "small" items, each too petty to warrant separate issues, but that we nonetheless do not want in the public surface. This is in the vein of #2300 and other similar PRs. Special emphasis was placed on making sure we don't have abstract protected members visible as part of the public surface, and other such things as this.

In my review of the assembly I did not address those types or members I knew were being taken care of through other channels. (E.g., the attribute bearing marker interfaces for `IComponentFactory`, or the calibrator.)

### DIFF
--- a/src/Microsoft.ML.Data/Commands/DefaultColumnNames.cs
+++ b/src/Microsoft.ML.Data/Commands/DefaultColumnNames.cs
@@ -4,7 +4,13 @@
 
 namespace Microsoft.ML.Data
 {
-    public static class DefaultColumnNames
+    /// <summary>
+    /// A set of string literals intended to be "canonical" names for column names intended for particular purpose.
+    /// While not part of the public API surface, its primary purpose is intended to be used in such a way as to encourage
+    /// uniformity on the public API surface, wherever it is judged where columns with default names should be consumed.
+    /// </summary>
+    [BestFriend]
+    internal static class DefaultColumnNames
     {
         public const string Features = "Features";
         public const string Label = "Label";
--- a/src/Microsoft.ML.Data/DataLoadSave/TransformWrapper.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/TransformWrapper.cs
@@ -15,9 +15,13 @@ using Microsoft.ML.Model;
 
 namespace Microsoft.ML.Data
 {
-    // REVIEW: this class is public, as long as the Wrappers.cs in tests still rely on it.
-    // It needs to become internal.
-    public sealed class TransformWrapper : ITransformer
+    /// <summary>
+    /// This is a shim class to present the legacy <see cref="IDataTransform"/> interface as an <see cref="ITransformer"/>.
+    /// Note that there are some important differences in usages that make this shimming somewhat non-seemless, so the goal
+    /// would be gradual removal of this as we do away with <see cref="IDataTransform"/> based code.
+    /// </summary>
+    [BestFriend]
+    internal sealed class TransformWrapper : ITransformer
     {
         internal const string LoaderSignature = "TransformWrapper";
         private const string TransformDirTemplate = "Step_{0:000}";
@@ -148,7 +152,7 @@ namespace Microsoft.ML.Data
     /// <summary>
     /// Estimator for trained wrapped transformers.
     /// </summary>
-    public abstract class TrainedWrapperEstimatorBase : IEstimator<TransformWrapper>
+    internal abstract class TrainedWrapperEstimatorBase : IEstimator<TransformWrapper>
     {
         [BestFriend]
         private protected readonly IHost Host;
--- a/src/Microsoft.ML.Data/DataLoadSave/TransformWrapper.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/TransformWrapper.cs
@@ -150,9 +150,11 @@ namespace Microsoft.ML.Data
     /// </summary>
     public abstract class TrainedWrapperEstimatorBase : IEstimator<TransformWrapper>
     {
-        protected readonly IHost Host;
+        [BestFriend]
+        private protected readonly IHost Host;
 
-        protected TrainedWrapperEstimatorBase(IHost host)
+        [BestFriend]
+        private protected TrainedWrapperEstimatorBase(IHost host)
         {
             Contracts.CheckValue(host, nameof(host));
             Host = host;
--- a/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
+++ b/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
@@ -10,20 +10,17 @@ namespace Microsoft.ML
     /// <summary>
     /// An object serving as a 'catalog' of available model operations.
     /// </summary>
-    public sealed class ModelOperationsCatalog
+    public sealed class ModelOperationsCatalog : IInternalCatalog
     {
-        /// <summary>
-        /// This is a best friend because an extension method defined in another assembly needs this field.
-        /// </summary>
-        [BestFriend]
-        internal IHostEnvironment Environment { get; }
+        IHostEnvironment IInternalCatalog.Environment => _env;
+        private readonly IHostEnvironment _env;
 
         public ExplainabilityTransforms Explainability { get; }
 
         internal ModelOperationsCatalog(IHostEnvironment env)
         {
             Contracts.AssertValue(env);
-            Environment = env;
+            _env = env;
 
             Explainability = new ExplainabilityTransforms(this);
         }
@@ -33,25 +30,26 @@ namespace Microsoft.ML
         /// </summary>
         /// <param name="model">The trained model to be saved.</param>
         /// <param name="stream">A writeable, seekable stream to save to.</param>
-        public void Save(ITransformer model, Stream stream) => model.SaveTo(Environment, stream);
+        public void Save(ITransformer model, Stream stream) => model.SaveTo(_env, stream);
 
         /// <summary>
         /// Load the model from the stream.
         /// </summary>
         /// <param name="stream">A readable, seekable stream to load from.</param>
         /// <returns>The loaded model.</returns>
-        public ITransformer Load(Stream stream) => TransformerChain.LoadFrom(Environment, stream);
+        public ITransformer Load(Stream stream) => TransformerChain.LoadFrom(_env, stream);
 
         /// <summary>
         /// The catalog of model explainability operations.
         /// </summary>
-        public sealed class ExplainabilityTransforms
+        public sealed class ExplainabilityTransforms : IInternalCatalog
         {
-            internal IHostEnvironment Environment { get; }
+            IHostEnvironment IInternalCatalog.Environment => _env;
+            private readonly IHostEnvironment _env;
 
             internal ExplainabilityTransforms(ModelOperationsCatalog owner)
             {
-                Environment = owner.Environment;
+                _env = owner._env;
             }
         }
 
@@ -68,7 +66,7 @@ namespace Microsoft.ML
             where TSrc : class
             where TDst : class, new()
         {
-            return new PredictionEngine<TSrc, TDst>(Environment, transformer, false, inputSchemaDefinition, outputSchemaDefinition);
+            return new PredictionEngine<TSrc, TDst>(_env, transformer, false, inputSchemaDefinition, outputSchemaDefinition);
         }
     }
 }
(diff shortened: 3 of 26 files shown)


src/Microsoft.ML.Data/Commands/DefaultColumnNames.cs
```diff
@@ -4,7 +4,13 @@
...
 namespace Microsoft.ML.Data
 {
-    public static class DefaultColumnNames
+    /// <summary>
+    /// A set of string literals intended to be "canonical" names for column names intended for particular purpose.
+    /// While not part of the public API surface, its primary purpose is intended to be used in such a way as to encourage
+    /// uniformity on the public API surface, wherever it is judged where columns with default names should be consumed.
+    /// </summary>
+    [BestFriend]
+    internal static class DefaultColumnNames
```
TomFinley: While helpful for *enabling* a uniform public surface, this is of course not the same as it needing to be in the public surface.

src/Microsoft.ML.Data/DataLoadSave/TransformWrapper.cs
```diff
@@ -148,11 +152,13 @@ IRowToRowMapper ITransformer.GetRowToRowMapper(DataViewSchema inputSchema)
     /// <summary>
     /// Estimator for trained wrapped transformers.
     /// </summary>
-    public abstract class TrainedWrapperEstimatorBase : IEstimator<TransformWrapper>
+    internal abstract class TrainedWrapperEstimatorBase : IEstimator<TransformWrapper>
```
TomFinley: This is a funny one. I was able to remove all subclasses of it, but some notes I read elsewhere led me to suspect there may be more than this than meets the eye. Started an issue about this for #2841. (Though that does not touch the public surface and is therefore inessential.)

src/Microsoft.ML.Data/DataLoadSave/TransformWrapper.cs
```diff
@@ -148,11 +152,13 @@ IRowToRowMapper ITransformer.GetRowToRowMapper(DataViewSchema inputSchema)
...
     {
-        protected readonly IHost Host;
+        [BestFriend]
+        private protected readonly IHost Host;
```
TomFinley: I have checked and in the public surface there is currently no `protected` member anywhere in `Microsoft.ML.Data`.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -28,16 +28,6 @@ internal ModelOperationsCatalog(IHostEnvironment env)
             Explainability = new ExplainabilityTransforms(this);
         }
 
-        public abstract class SubCatalogBase
```
TomFinley: I may do more work to get rid of this sort of thing. These mysterious subclasses mean nothing to the user, and as part of the "central" object `MLContext`, we should probably choose some alternate mechanism. (I got rid of it here, but there's similar code elsewhere.)

src/Microsoft.ML.Data/Commands/DefaultColumnNames.cs
eerhardt: (I agree we should internalize this - just raising as a general question in this area)

How do we imagine users to find out about these "magic" strings? Currently, one way is the default values in the MLContext APIs. But I wonder if that is enough. For example, when making a Prediction "output" class, how do users know to name/attribute their properties as `PredictedLabel`, `Probability`, `Score`?  Just through documentation?

---
In reply to: [262298366](https://github.com/dotnet/machinelearning/pull/2842#discussion_r262298366) [](ancestors = 262298366)

src/Microsoft.ML.Data/Prediction/PredictionEngine.cs
```diff
@@ -103,7 +103,7 @@ public abstract class PredictionEngineBase<TSrc, TDst> : IDisposable
         /// <summary>
         /// Provides output schema.
         /// </summary>
-        public DataViewSchema OutputSchema;
+        public DataViewSchema OutputSchema { get; }
```
TomFinley: A mutable schema on the base class of the prediction engine. *That* would have been embarrassing. :)

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
eerhardt: I like this change.  :+1:

---
In reply to: [262299101](https://github.com/dotnet/machinelearning/pull/2842#discussion_r262299101) [](ancestors = 262299101)

src/Microsoft.ML.Transforms/Text/WrappedTextTransformers.cs
```diff
@@ -14,8 +15,9 @@ namespace Microsoft.ML.Transforms.Text
     /// Produces a bag of counts of ngrams (sequences of consecutive words) in a given text.
     /// It does so by building a dictionary of ngrams and using the id in the dictionary as the index in the bag.
     /// </summary>
-    public sealed class WordBagEstimator : TrainedWrapperEstimatorBase
+    public sealed class WordBagEstimator : IEstimator<ITransformer>
```
TomFinley: The `TrainedWrapperEstimatorBase` was meant to facilitate the translation from the legacy `IDataTransform` interface to the new `ITransformer` interface. It is intended to be a temporary situation; however, by making this class descend from that abstract base class, that temporary shim would have had to become permanent -- you cannot "undescend" from a class without it being a breaking change in your API. Therefore we would have had to live with it forever.

So we had to duplicate the logic that existed in that base class. Fortunately there were only two subclasses remaining (both in this file, as it happened), so it was not such a big problem.

src/Microsoft.ML.Data/Commands/DefaultColumnNames.cs
TomFinley: Yes, unfortunately I think both helpful displays in the debugger (for live discovery) and documentation (for offline discovery) are the only real answer we have right now -- both seem insufficient or at least suboptimal. This is a problem I think in most places we've chosen to believe that producing "default" columns names is more helpful than requiring the user to specify them. I am unsure how to address that problem. Something *resembling* the static typed API will I think be a solution to that problem, since the columns will be typed and addressable and right there.

src/Microsoft.ML.Transforms/Text/WrappedTextTransformers.cs
```diff
@@ -119,16 +123,26 @@ public override TransformWrapper Fit(IDataView input)
                 Weighting = _weighting
             };
 
-            return new TransformWrapper(Host, WordBagBuildingTransformer.Create(Host, options, input), true);
+            return new TransformWrapper(_host, WordBagBuildingTransformer.Create(_host, options, input), true);
+        }
+
+        public SchemaShape GetOutputSchema(SchemaShape inputSchema)
+        {
```
artidoro: Can you add a summary for this new public method?

src/Microsoft.ML.Transforms/Text/WrappedTextTransformers.cs
```diff
@@ -259,7 +275,16 @@ public override TransformWrapper Fit(IDataView input)
                 InvertHash = _invertHash
             };
 
-            return new TransformWrapper(Host, WordHashBagProducingTransformer.Create(Host, options, input), true);
+            return new TransformWrapper(_host, WordHashBagProducingTransformer.Create(_host, options, input), true);
+        }
+
+        public SchemaShape GetOutputSchema(SchemaShape inputSchema)
```
artidoro: >SchemaShape [](start = 15, length = 11)

Same here: Can you add a summary for this new public method?

src/Microsoft.ML.Transforms/Text/WrappedTextTransformers.cs
artidoro: This is a really good idea! until we do the proper translation

---
In reply to: [262299912](https://github.com/dotnet/machinelearning/pull/2842#discussion_r262299912) [](ancestors = 262299912)

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
TomFinley: Very good, I have done as I suggested, and got rid of the rest. We typically hide infrastructure behind this using the tricks of explicit implementations of internal interfaces, so I used the same in introducing a new interface `IInternalCatalog` (not the best name, perhaps, but as it is internal we are free to improve it later). This also allowed me to simplify `CatalogUtils` considerably.

src/Microsoft.ML.Transforms/Text/WrappedTextTransformers.cs
TomFinley: I suppose I could. While acknowledging the worth of documentation, I would point out documentation can be added any time. We have three weeks and only three weeks to finalize the public shape of the API. Anything that is there must remain there forever.

This method is not really "new" as such. It is merely copied from the existing base class, where it has precisely the same documentation as you see now.

If other people are making similar requests on other PRs, I might advise them to desist.

src/Microsoft.ML.Data/Commands/DefaultColumnNames.cs
artidoro: I have seen quite a few instances where users use `DefaultColumnNames`. I don't understand why we want to hide it. We will always have default column names, why not allowing users to access them instead of hard coding them in their applications?

searching the samples repo shows there are quite a few instances where people use `DefaultColumnNames`.
[https://github.com/dotnet/machinelearning-samples/search?q=defaultcolumnnames&unscoped_q=defaultcolumnnames](https://github.com/dotnet/machinelearning-samples/search?q=defaultcolumnnames&unscoped_q=defaultcolumnnames)

or this other user provided code:
[https://github.com/dotnet/machinelearning/issues/2486#issuecomment-462248105](https://github.com/dotnet/machinelearning/issues/2486#issuecomment-462248105)


---
In reply to: [262302369](https://github.com/dotnet/machinelearning/pull/2842#discussion_r262302369) [](ancestors = 262302369)

artidoro:
>         /// </summary>

Should this be more specific to calibrators?

---
Refers to: src/Microsoft.ML.Data/TrainCatalog.cs:381 in 7b714fd. [](commit_id = 7b714fd48f5de10ec9aa86e0d90b507b01fc10d8, deletion_comment = False)

src/Microsoft.ML.Data/Commands/DefaultColumnNames.cs
TomFinley: In those places, what would have happened in those samples if instead of `DefaultColumnNames.Features` we instead wrote `"Features"`? I understand the appeal inside the tool of having a consistent "location" for them. For user code, I am less certain.

I am very uncomfortable with having as part of our public surface a class that is merely a collection of "magic strings" that we've used internally to denote certain concepts. That's not great API design.

The samples can and should change. To use this class of magic constants was, frankly, a bad move by the sample author, since it was totally needless.

artidoro:
>         /// </summary>

Why is this showing as a change? I see it was added as part of ivan's PR #2766

---
In reply to: [469501477](https://github.com/dotnet/machinelearning/pull/2842#issuecomment-469501477) [](ancestors = 469501477)

---
Refers to: src/Microsoft.ML.Data/TrainCatalog.cs:381 in 392cee6. [](commit_id = 392cee654b823f81570c49234afa8fc09c97c3ed, deletion_comment = False)

TomFinley:
>         /// </summary>

The documentation is wrong, but this line (381) is not at least showing as a change for me either on the Github web review tool, nor in the non-web tools we use to review, so, maybe something somehow got screwed up on your end?

The last line it shows that I've changed is 258, and the next line is 409. This line (381) definitely did not change.

I agree that the documentation should be changed to actually reflect reality, but I hope we can refine that later.

---
In reply to: [469504933](https://github.com/dotnet/machinelearning/pull/2842#issuecomment-469504933) [](ancestors = 469504933,469501477)

---
Refers to: src/Microsoft.ML.Data/TrainCatalog.cs:381 in 658a128. [](commit_id = 658a128d0423cb0b5c06d8860edc4c843d856fd8, deletion_comment = False)

src/Microsoft.ML.Transforms/Text/WrappedTextTransformers.cs
eerhardt: I think there’s an “inheritdoc” syntax you can use. 

Aspnet uses it:

https://github.com/aspnet/AspNetCore/blob/c565386a3ed135560bc2e9017aa54a950b4e35dd/src/Mvc/Mvc.Core/src/Infrastructure/NonDisposableStream.cs

It looks like some doc tools respect these. Do ours?

Here’s the C# Lang feature request to add it as an official support in the language.

https://github.com/dotnet/csharplang/issues/313

PR #2846, opened 2019-03-05
Remove ConcurrencyFactor from IHostEnvironment

### LINKED ISSUE #2051: Removal of IHostEnvironment.ConcurrencyFactor and replace with ... ?
Consider this property

https://github.com/dotnet/machinelearning/blob/e2f7075782764c9dab8f0a04013aaf8b9921d984/src/Microsoft.ML.Core/Data/IHostEnvironment.cs#L43

set for example via this `conc` parameter here:

https://github.com/dotnet/machinelearning/blob/e2f7075782764c9dab8f0a04013aaf8b9921d984/src/Microsoft.ML.Data/MLContext.cs#L76

That we should have some sort of indication hinting of degree of desired parallelism might have seemed useful at one point, but the practical reality has more closely resembled this: components where parallelism matters for performance or results invariably wind up having their own independent settings *anyway*, to such a degree even that the `IHostEnvironment.ConcurrencyFactory` property is for the most part totally ignored -- e.g., people set `conc = 1` thinking this will give them deterministic results, but of course things like SDCA and LR have separate multithreading behavior, which under default settings will not give deterministic results unless it is set.

As an integer property, and as originally intended, this property is, i'd claim, a failure, since everywhere it *could* matter we wind up inevitably ignoring and overriding its behavior anyway, and where we don't it's only because that is a situation where such control doesn't really matter.

We ought at least to remove this, I think. This must be done prior to v1.

Whether post v1 we want to replace it with *anything* is another matter. Clearly at least for *test* purposes we want to have something to ensure consistent deterministic results, but is this really a user concern? (Definitely in other ML frameworks it is not a guarantee at all, so the idea of providing it and making it some sort of global configuration option seems questionable.)

/cc @stephentoub

artidoro:
Fixes #2051.

In this PR I remove `ConcurrencyFactor` from `IHostEnvironment` and interfaces and classes deriving/implementing it. 

I had to skip a `RandomPredictor` as I am not sure how I can require a single threaded behavior without the `ConcurrencyFactor` for this specific trainer.


artidoro:
Note that the first and second commit contain the most important logical changes the rest are fixes to tests and samples to make sure they compile.

TomFinley:
> I had to skip a `RandomPredictor` as I am not sure how I can require a single threaded behavior without the `ConcurrencyFactor` for this specific trainer.

Actually that's interesting. We need to hide this for now, just as a we had to hide `GenerateNumberTransform`. (Not necessarily part of your PR, but if you wanted to do it it wouldn't be that hard, we'd just need to hide `RandomModelParameters` and `RandomTrainer`.) `RandomModelParameters` claims it implements the internal interface `IValueMapperDist`, which means that it is claiming (via our infrastructure) that it is possible to express as an `IRowToRowMapper`, but it can't since the outputs are not independent.

src/Microsoft.ML.KMeansClustering/KMeansPlusPlusTrainer.cs
```diff
@@ -219,15 +219,11 @@ private KMeansModelParameters TrainCore(IChannel ch, RoleMappedData data, int di
...
             // isn't going to get us a performance improvement.
-            int maxThreads;
-            if (host.ConcurrencyFactor < 1)
-                maxThreads = Environment.ProcessorCount / 2;
-            else
-                maxThreads = host.ConcurrencyFactor;
+            int maxThreads = Environment.ProcessorCount / 2;
 
             // If we specified a number of threads that's fine, but it must be below the
             // host-set concurrency factor.
```
TomFinley: >host-set concurrency factor. [](start = 15, length = 28)

Not quite right

Ivanidzo4ka:
>             //      The coordinates of centroid 0 are: (26, 6, 1)

I would double check output in samples.
Or just set numThreads in KMeans to 1.

---
Refers to: docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Clustering/KMeans.cs:45 in a8c7541. [](commit_id = a8c75418f4ee1a766b788fc2e670bc918a4542d0, deletion_comment = False)

Ivanidzo4ka:
>             //      The coordinates of centroid 0 are: (26, 6, 1)

Please validate what output in samples are same across all files in this PR

---
In reply to: [469970156](https://github.com/dotnet/machinelearning/pull/2846#issuecomment-469970156) [](ancestors = 469970156)

---
Refers to: docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Clustering/KMeans.cs:45 in a8c7541. [](commit_id = a8c75418f4ee1a766b788fc2e670bc918a4542d0, deletion_comment = False)

src/Microsoft.ML.Data/DebuggerExtensions.cs
```diff
@@ -45,7 +45,7 @@ public static DataDebuggerPreview Preview(this ITransformer transformer, IDataVi
             Contracts.CheckParam(maxRows >= 0, nameof(maxRows));
             Contracts.CheckParam(maxTrainingRows >= 0, nameof(maxTrainingRows));
 
-            var env = new LocalEnvironment(conc: 1);
+            var env = new LocalEnvironment();
```
Ivanidzo4ka: >LocalEnvironment [](start = 26, length = 16)

Do we have any tests for Preview method?
Can you double check it still works with you changes?

src/Microsoft.ML.Core/Data/IHostEnvironment.cs
```diff
@@ -59,14 +59,7 @@ public interface IHostEnvironment : IChannelProvider, IProgressChannelProvider
         /// <summary>
         /// Create a host with the given registration name.
         /// </summary>
```
wschin: Please explain what `seed` and `verbose` are.

src/Microsoft.ML.Data/DebuggerExtensions.cs
artidoro: We use it in some tests and I tried using it in the debugger and it seemed to be working.

---
In reply to: [262793799](https://github.com/dotnet/machinelearning/pull/2846#discussion_r262793799) [](ancestors = 262793799)

PR #2849, opened 2019-03-05
Remove random trainer and model parameters from the public surface.

### LINKED ISSUE #2848: Internalize random model parameters and trainer
In #2846 it came out that we still somehow had the random predictor lying around -- I was unaware of this. This structure, being inherently stateful, will introduce subtle bugs and problems when used as part of a transformer. (E.g., what happens when you construct multiple prediction engines over the resulting transform.) These bugs are subtle, but still there and points to a deep problem with this structure. It claims to be a value mapper, so it is even claiming to be a function, which it definitely is not. Perhaps its presence was viewed as merely harmless which is why we did not remove it in prior years, but now that we're trying to finalize the API we must make a stronger case here.

We have in other places introduced other mechanisms to deal with inherently stateful predictors (e.g., what we had to do with time series). So, I think that in time we can address the need for this thing, *if there is a need*. (I have been unaware of any scenario where this thing is useful beyond a mere comparison point for evaluation, which can be done far better in a variety of ways.)

The work is fairly simple: internalize this architecturally troublesome trainer and model. (This is similar to how we are still keeping the generate number transform internal, though, that one has actual real application.) We can decide what to do with it later.

TomFinley:
Fix #2848.


src/Microsoft.ML.StandardLearners/Properties/AssemblyInfo.cs
```diff
@@ -9,6 +9,7 @@
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.StaticPipe" + PublicKey.Value)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Core.Tests" + PublicKey.TestValue)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Predictor.Tests" + PublicKey.TestValue)]
+[assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Tests" + PublicKey.TestValue)]
```
TomFinley: I think @artidoro in his #2846 actually skipped this test, so we might be able to avoid this by simply removing the test. NOt sure what he thinks.

src/Microsoft.ML.StandardLearners/StandardLearnersCatalog.cs
```diff
@@ -742,7 +742,7 @@ private static ICalibratorTrainer GetCalibratorTrainerOrThrow(IExceptionContext
         ///  [!code-csharp[FastTree](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/BinaryClassification/RandomTrainerSample.cs)]
         /// ]]></format>
```
artidoro: If you want to remove the sample I think you also need to remove the reference here.

src/Microsoft.ML.StandardLearners/Properties/AssemblyInfo.cs
artidoro: Give what you highlighted I think we should remove the test, we already have a test with `TestEstimatorCore`. If we want to make this estimator public, I think we should write tests in the new API anyway. 

You can remove it here, or I can do it in my PR.

src/Microsoft.ML.StandardLearners/StandardLearnersCatalog.cs
```diff
@@ -736,13 +736,7 @@ private static ICalibratorTrainer GetCalibratorTrainerOrThrow(IExceptionContext
...
         /// </remarks>
         /// <param name="catalog">The <see cref="BinaryClassificationCatalog"/>.</param>
-        /// <example>
-        /// <format type="text/markdown">
-        /// <![CDATA[
-        ///  [!code-csharp[FastTree](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/BinaryClassification/RandomTrainerSample.cs)]
-        /// ]]></format>
-        /// </example>
-        public static RandomTrainer Random(this BinaryClassificationCatalog.BinaryClassificationTrainers catalog)
+        internal static RandomTrainer Random(this BinaryClassificationCatalog.BinaryClassificationTrainers catalog)
```
eerhardt: Can this just be deleted?

TomFinley: Yeah... actually from what I'm reading of the prior documentation it is simply not necessary at all. As far as I can tell, it existed in pre-`IDataView` days when evaluation didn't happen by evaluating predictions (which are of course just data), but evaluation happened over predictors. Nowadays evaluation of course just takes an `IDataView`, so involving predictors is not necessary at all (and in fact merely complicates the whole story). Even from those days I can't even find any instances of that actually becoming useful, so while I was somewhat reluctant to undertake that drastic step, I think it probably is appropriate now that I've read a bit more about this.

TomFinley: Keeping the predictor for now for potential back-compat (we can figure out what we really want to do about that later).

src/Microsoft.ML.StandardLearners/Standard/Simple/SimpleTrainers.cs
```diff
@@ -7,123 +7,34 @@
...
-    }
-
-    /// <summary>
-    /// The predictor implements the Predict() interface. The predictor returns a
-    ///  uniform random probability and classification assignment.
+    /// This is a faulty model parameters that claims to be a <see cref="IValueMapperDist"/>. We maintain the class for
+    /// now for backwards compatibility of models (assuming there are any), but might consider removing it in the future.
     /// </summary>
-    public sealed class RandomModelParameters :
+    [Obsolete("This class was fundamnetally misdesigned, is incapable of implementing the interfaces it claims to implement, " +
```
eerhardt: type-o `fundamnetally`

src/Microsoft.ML.StandardLearners/Properties/AssemblyInfo.cs
eerhardt: Do we still need this now?

TomFinley: Ah you're right.

PR #2850, opened 2019-03-05
Add API to save/load models with their input schema

### LINKED ISSUE #2735: Add API to save/load models with their input schema
Reasons for this are listed in issue #2663.

Currently, ModelOperationsCatalog offers the following API:

```
public void Save(ITransformer model, Stream stream) 
public ITransformer Load(Stream stream)
```
So when using a loaded model, users have to create the `IDataView` to be passed to the `ITransformer` themselves by creating a new `TextLoader`, (or another way?). 
I suggest adding these new APIs to ModelOperationsCatalog:

```
public void Save<TSource>(IDataReader<TSource> model, Stream stream);
public void Save<TSource>(IDataReader<TSource> reader, ITransformer model, Stream stream);
public IDataReader<TSource> Load<TSource>(Stream stream);
```
The last one would return a `CompositeDataReader` containing the loader and the `ITransformer` chain, so we could also add new APIs to `DataOperationsCatalog` to only load the reader:

```
public TextLoader CreateTextLoader(Stream stream);
```

Another option is to add an API that creates a `PredictionEngine` from a `Stream`, or an API that creates a `SchemaDefinition` from a `Stream` (that way users can use the existing API to create a `PredictionEngine`).

yaeldekel:
Fixes #2735. 

Currently, this PR adds a public API that saves and load an IDataLoader. Later iterations will add APIs to save/load the schema directly.

### DIFF
--- a/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
+++ b/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
@@ -45,12 +45,33 @@ namespace Microsoft.ML
         /// <param name="stream">A writeable, seekable stream to save to.</param>
         public void Save(ITransformer model, Stream stream) => model.SaveTo(Environment, stream);
 
+        public void Save<TSource>(IDataLoader<TSource> model, Stream stream)
+        {
+            using (var rep = RepositoryWriter.CreateNew(stream))
+            {
+                ModelSaveContext.SaveModel(rep, model, "Model");
+                rep.Commit();
+            }
+        }
+
+        public void Save<TSource>(IDataLoader<TSource> loader, ITransformer model, Stream stream) =>
+            Save(new CompositeDataLoader<TSource, ITransformer>(loader, new TransformerChain<ITransformer>(model)), stream);
+
         /// <summary>
         /// Load the model from the stream.
         /// </summary>
         /// <param name="stream">A readable, seekable stream to load from.</param>
         /// <returns>The loaded model.</returns>
         public ITransformer Load(Stream stream) => TransformerChain.LoadFrom(Environment, stream);
+
+        public IDataLoader<IMultiStreamSource> LoadAsCompositeDataLoader(Stream stream)
+        {
+            using (var rep = RepositoryReader.Open(stream))
+            {
+                ModelLoadContext.LoadModel<IDataLoader<IMultiStreamSource>, SignatureLoadModel>(Environment, out var model, rep, "Model");
+                return model;
+            }
+        }
 
         /// <summary>
         /// The catalog of model explainability operations.
--- a/src/Microsoft.ML.Data/Prediction/Calibrator.cs
+++ b/src/Microsoft.ML.Data/Prediction/Calibrator.cs
@@ -14,7 +14,6 @@ using Microsoft.ML.Calibrators;
 using Microsoft.ML.CommandLine;
 using Microsoft.ML.Data;
 using Microsoft.ML.EntryPoints;
-using Microsoft.ML.Internal.Internallearn;
 using Microsoft.ML.Internal.Utilities;
 using Microsoft.ML.Model;
 using Microsoft.ML.Model.OnnxConverter;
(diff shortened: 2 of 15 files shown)


eerhardt:
>         public void SaveTo(IHostEnvironment env, Stream outputStream)

Do we want this public API on `CompositeDataLoader`?

---
Refers to: src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs:104 in 2cf7ee4. [](commit_id = 2cf7ee4bbafd0faab4b1f44fe5fa1c1d15afe25f, deletion_comment = False)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -58,19 +57,19 @@
     "Naive Calibration Executor",
     NaiveCalibrator.LoaderSignature)]
 
-[assembly: LoadableClass(typeof(ValueMapperCalibratedModelParameters<IPredictorProducing<float>, ICalibrator>), null, typeof(SignatureLoadModel),
+[assembly: LoadableClass(typeof(ValueMapperCalibratedModelParameters<object, ICalibrator>), null, typeof(SignatureLoadModel),
```
TomFinley: >object [](start = 69, length = 6)

This is a bit stronger medicine than what I was suggesting! (I don't object per se, but it did entail a bit more work.) I suppose it's a bit more "honest" than the old way -- marker interfaces are more or less a promise you can't keep -- but they are still useful on internal code where you have complete control over the implementation (as we do as they are internal).

So I think this should be fine, I just wonder, are we fully comfortable with this?

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -45,13 +45,34 @@ protected SubCatalogBase(ModelOperationsCatalog owner)
         /// <param name="stream">A writeable, seekable stream to save to.</param>
         public void Save(ITransformer model, Stream stream) => model.SaveTo(Environment, stream);
 
+        public void Save<TSource>(IDataLoader<TSource> model, Stream stream)
```
eerhardt: need xml comments for new public API.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -45,13 +45,34 @@ protected SubCatalogBase(ModelOperationsCatalog owner)
...
+            Save(new CompositeDataLoader<TSource, ITransformer>(loader, new TransformerChain<ITransformer>(model)), stream);
+
         /// <summary>
         /// Load the model from the stream.
         /// </summary>
         /// <param name="stream">A readable, seekable stream to load from.</param>
         /// <returns>The loaded model.</returns>
         public ITransformer Load(Stream stream) => TransformerChain.LoadFrom(Environment, stream);
 
+        public IDataLoader<IMultiStreamSource> LoadAsCompositeDataLoader(Stream stream)
```
TomFinley: >IDataLoader [](start = 15, length = 11)

The subtle differences between usage here and usage of the above might be a bit hard of a pill to swallow.

eerhardt: Should this return `CompositeDataLoader` instead of `IDataLoader`?  The name is: `LoadAsCompositeDataLoader`, which makes me think it should return a `CompositeDataLoader`

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
TomFinley: This I think this is not correct... we need to save *both* the loader (or failing that, input schema) as well as the transform model to a stream. Not two separate things.

TomFinley: I recall the mistake that was made with predictor models and transform models in entry-points, where we viewed these as single things saved to a stream, as opposed to having to capture the entire pipeline. This mistake was inherited by ML.NET v0.1, which led to all sorts of grief when people realized, yes, saving the loader is *really freakin' important*. So I think we're just making the same mistake again. A model isn't just a transform model, or a loader, or an input schema, it I think needs to capture all of them, and I think we need a change a bit more fundamental than merely adding a few methods here and there.

---
In reply to: [262638162](https://github.com/dotnet/machinelearning/pull/2850#discussion_r262638162) [](ancestors = 262638162)

TomFinley:
Were we going to try to address the problem of saving schemas? Not all data sources we train a transform chain with will be coming from a loader.

yaeldekel:
Closing this PR and replacing it with #2858 that was created from the correct fork.

PR #2851, opened 2019-03-05
Polish early stop rules in fast tree

### LINKED ISSUE #2520: FastTree EarlyStoppingRule definition is inconsistent with API
In `FastTree`, the `EarlyStoppingRule` is defined as
```cs
[Argument(ArgumentType.Multiple, HelpText = "Early stopping rule. (Validation set (/valid) is required.)", ShortName = "esr", NullName = "<Disable>")]
[TGUI(Label = "Early Stopping Rule", Description = "Early stopping rule. (Validation set (/valid) is required.)")]
public IEarlyStoppingCriterionFactory EarlyStoppingRule;
```

This can be specified like so:
```cs
var fastTreeTrainer = mlContext.Regression.Trainers.FastTree(new 
    Trainers.FastTree.FastTreeRegressionTrainer.Options {
        EarlyStoppingMetrics = 2,
        EarlyStoppingRule = new GLEarlyStoppingCriterion.Arguments()
    });
```

This exposes the `IComponentFactory` way of doing things in the public API, which seems to be inconsistent with the public API. One suggestion is to use an `enum` over the existing options.

That said, this does support custom early stopping methods through implementing  `IEarlyStoppingCriterionFactory`.

wschin:
Fix #2520. The pattern implemented in this PR is
```csharp
        [BestFriend]
        [Argument(ArgumentType.Multiple, HelpText = "Early stopping rule. (Validation set (/valid) is required.)", ShortName = "esr", NullName = "<Disable>")]
        [TGUI(Label = "Early Stopping Rule", Description = "Early stopping rule. (Validation set (/valid) is required.)")]
        internal IEarlyStoppingCriterionFactory EarlyStoppingRuleFactory;

        /// <summary>
        /// The underlying state of <see cref="EarlyStoppingRuleFactory"/> and <see cref="EarlyStoppingRule"/>.
        /// </summary>
        private EarlyStoppingRuleBase _earlyStoppingRuleBase;

        /// <summary>
        /// Early stopping rule used to terminate training process once meeting a specified criterion. Possible choices are
        /// <see cref="EarlyStoppingRuleBase"/>'s implementations such as <see cref="TolerantEarlyStoppingRule"/> and <see cref="GeneralityLossRule"/>.
        /// </summary>
        public EarlyStoppingRuleBase EarlyStoppingRule
        {
            get { return _earlyStoppingRuleBase;  }
            set
            {
                _earlyStoppingRuleBase = value;
                EarlyStoppingRuleFactory = _earlyStoppingRuleBase.BuildFactory();
            }
        }
```
You can see that `EarlyStoppingRuleFactory` (used in old infra) is exposed to users by adding `EarlyStoppingRule`.

### DIFF
--- a/src/Microsoft.ML.FastTree/FastTreeArguments.cs
+++ b/src/Microsoft.ML.FastTree/FastTreeArguments.cs
@@ -621,9 +621,12 @@ namespace Microsoft.ML.Trainers.FastTree
         /// <summary>
         /// Early stopping rule. (Validation set (/valid) is required).
         /// </summary>
+        [BestFriend]
         [Argument(ArgumentType.Multiple, HelpText = "Early stopping rule. (Validation set (/valid) is required.)", ShortName = "esr", NullName = "<Disable>")]
         [TGUI(Label = "Early Stopping Rule", Description = "Early stopping rule. (Validation set (/valid) is required.)")]
-        public IEarlyStoppingCriterionFactory EarlyStoppingRule;
+        internal IEarlyStoppingCriterionFactory EarlyStoppingRuleFactory;
+
+        public EarlyStoppingRuleBase EarlyStoppingRule;
 
         /// <summary>
         /// Early stopping metrics. (For regression, 1: L1, 2:L2; for ranking, 1:NDCG@1, 3:NDCG@3).
--- a/src/Microsoft.ML.FastTree/FastTreeArguments.cs
+++ b/src/Microsoft.ML.FastTree/FastTreeArguments.cs
@@ -622,12 +622,19 @@ namespace Microsoft.ML.Trainers.FastTree
         /// Early stopping rule. (Validation set (/valid) is required).
         /// </summary>
         [BestFriend]
-        [Argument(ArgumentType.Multiple, HelpText = "Early stopping rule. (Validation set (/valid) is required.)", ShortName = "esr", NullName = "<Disable>")]
+        [Argument(ArgumentType.Multiple, HelpText = "Early stopping rule. (Validation set (/valid) is required.)", Name = "EarlyStoppingRule", ShortName = "esr", NullName = "<Disable>")]
         [TGUI(Label = "Early Stopping Rule", Description = "Early stopping rule. (Validation set (/valid) is required.)")]
         internal IEarlyStoppingCriterionFactory EarlyStoppingRuleFactory;
 
+        /// <summary>
+        /// The underlying state of <see cref="EarlyStoppingRuleFactory"/> and <see cref="EarlyStoppingRule"/>.
+        /// </summary>
         private EarlyStoppingRuleBase _earlyStoppingRuleBase;
 
+        /// <summary>
+        /// Early stopping rule used to terminate training process once meeting a specified criterion. Possible choices are
+        /// <see cref="EarlyStoppingRuleBase"/>'s implementations such as <see cref="TolerantEarlyStoppingRule"/> and <see cref="GeneralityLossRule"/>.
+        /// </summary>
         public EarlyStoppingRuleBase EarlyStoppingRule
         {
             get { return _earlyStoppingRuleBase;  }
--- a/src/Microsoft.ML.FastTree/FastTreeArguments.cs
+++ b/src/Microsoft.ML.FastTree/FastTreeArguments.cs
@@ -626,7 +626,17 @@ namespace Microsoft.ML.Trainers.FastTree
         [TGUI(Label = "Early Stopping Rule", Description = "Early stopping rule. (Validation set (/valid) is required.)")]
         internal IEarlyStoppingCriterionFactory EarlyStoppingRuleFactory;
 
-        public EarlyStoppingRuleBase EarlyStoppingRule;
+        private EarlyStoppingRuleBase _earlyStoppingRuleBase;
+
+        public EarlyStoppingRuleBase EarlyStoppingRule
+        {
+            get { return _earlyStoppingRuleBase;  }
+            set
+            {
+                _earlyStoppingRuleBase = value;
+                EarlyStoppingRuleFactory = _earlyStoppingRuleBase.BuildFactory();
+            }
+        }
 
         /// <summary>
         /// Early stopping metrics. (For regression, 1: L1, 2:L2; for ranking, 1:NDCG@1, 3:NDCG@3).
--- a/src/Microsoft.ML.FastTree/Training/EarlyStoppingCriteria.cs
+++ b/src/Microsoft.ML.FastTree/Training/EarlyStoppingCriteria.cs
@@ -8,24 +8,30 @@ using Microsoft.ML.CommandLine;
 using Microsoft.ML.EntryPoints;
 using Microsoft.ML.Trainers.FastTree;
 
-[assembly: LoadableClass(typeof(TolerantEarlyStoppingCriterion), typeof(TolerantEarlyStoppingCriterion.Options), typeof(SignatureEarlyStoppingCriterion), "Tolerant (TR)", "tr")]
-[assembly: LoadableClass(typeof(GLEarlyStoppingCriterion), typeof(GLEarlyStoppingCriterion.Options), typeof(SignatureEarlyStoppingCriterion), "Loss of Generality (GL)", "gl")]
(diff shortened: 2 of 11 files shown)


Ivanidzo4ka:
@zeahmed  @ganik  You both working on same problem in Loss function and Text featurizer. Would be nice if we stay sync regarding solution.

test/BaselineOutput/Common/EntryPoints/core_manifest.json
```diff
@@ -6622,7 +6622,7 @@
           "Default": "GradientDescent"
         },
         {
-          "Name": "EarlyStoppingRule",
+          "Name": "EarlyStoppingRuleFactory",
           "Type": {
             "Kind": "Component",
             "ComponentKind": "EarlyStoppingCriterion"
```
ganik: >"ComponentKind": "EarlyStoppingCriterion" [](start = 12, length = 41)

Manifest is broken. Here is a reference to ComponenKInd "EarlyStoppingCriterion" but the Components for this component kind is removed from manifest. This would  Entrypoint Compiler to fail to figure out fitting components during entrypoint graph parse

src/Microsoft.ML.FastTree/FastTreeArguments.cs
```diff
@@ -482,9 +482,12 @@ public enum OptimizationAlgorithmType { GradientDescent, AcceleratedGradientDesc
         /// <summary>
         /// Early stopping rule. (Validation set (/valid) is required).
         /// </summary>
+        [BestFriend]
         [Argument(ArgumentType.Multiple, HelpText = "Early stopping rule. (Validation set (/valid) is required.)", ShortName = "esr", NullName = "<Disable>")]
```
ganik: Add attribute Name = "EarlyStoppingRule"  this would keep core_manifest.json same and will avoid big changes in NimbusML

src/Microsoft.ML.FastTree/Training/EarlyStoppingCriteria.cs
```diff
@@ -35,24 +41,30 @@ public abstract class IEarlyStoppingCriterion
         /// <param name="isBestCandidate">True if the current result is the best ever.</param>
         /// <returns>If true, the learning should stop.</returns>
         public abstract bool CheckScore(float validationScore, float trainingScore, out bool isBestCandidate);
+
+        /// <summary>
+        /// Create <see cref="IEarlyStoppingCriterionFactory"/> for supporting legacy infra built upon <see cref="IComponentFactory"/>.
+        /// </summary>
+        internal abstract IEarlyStoppingCriterionFactory BuildFactory();
```
ganik: >internal abstract IEarlyStoppingCriterionFactory BuildFactory(); [](start = 8, length = 64)

How would 3rd party devs create their own EarlyStoppingRuleBase implementations to supply into FastTree algos? This abstract internal method will prevent that

TomFinley: Component authorship is not a goal for v1, even at the level of people being able to create their own `ITransformer` implementations. This is fine. Inf act I would go even further by making this thing have a `private protected` constructor.

---
In reply to: [262757300](https://github.com/dotnet/machinelearning/pull/2851#discussion_r262757300) [](ancestors = 262757300)

test/BaselineOutput/Common/EntryPoints/core_manifest.json
wschin: No (at least at iteration 3), those definitions of EarlyStoppingCriterion are still there. Please try searching for "EarlyStoppingCriterion".

---
In reply to: [262645293](https://github.com/dotnet/machinelearning/pull/2851#discussion_r262645293) [](ancestors = 262645293)

src/Microsoft.ML.FastTree/FastTreeArguments.cs
wschin: Fixed but I'd expect tons of breaking changes to entry-point had happened somewhere.

---
In reply to: [262650819](https://github.com/dotnet/machinelearning/pull/2851#discussion_r262650819) [](ancestors = 262650819)

src/Microsoft.ML.FastTree/Training/EarlyStoppingCriteria.cs
wschin: Sounds good Tom. I also think creating a proper stopping rule is a challenge to many C# developers.

---
In reply to: [263023672](https://github.com/dotnet/machinelearning/pull/2851#discussion_r263023672) [](ancestors = 263023672,262757300)

ganik: In this case if you have only restricted set of EarlyStoppingRule - why dont you go with enum? I dont understand creating such a complicated code where enum will suffice

---
In reply to: [263048215](https://github.com/dotnet/machinelearning/pull/2851#discussion_r263048215) [](ancestors = 263048215,263023672,262757300)

wschin: I have explained that this can't be enum in the original issue (#2520). Stopping rule itself can contain states, parameters, and so on; it's more than an enum.

---
In reply to: [263057699](https://github.com/dotnet/machinelearning/pull/2851#discussion_r263057699) [](ancestors = 263057699,263048215,263023672,262757300)

ganik: I would like to discuss this further. You are exposing EarlyStoppingRule and dont allow devs to use it. If you would decouple Factory from Class/Interface you would get a smaller code change and will allow 3rd party devs create their own stopping rule - same as I did for LossFunction

---
In reply to: [263057699](https://github.com/dotnet/machinelearning/pull/2851#discussion_r263057699) [](ancestors = 263057699,263048215,263023672,262757300)

wschin: Just make a note for our internal discussion --- public area should be as small as possible while all vital functions are still maintained. Thus, we don't allow the implementation of this base class.

---
In reply to: [263063854](https://github.com/dotnet/machinelearning/pull/2851#discussion_r263063854) [](ancestors = 263063854,263057699,263048215,263023672,262757300)

ganik: I am ok as far as this pattern is an exception and doesnt propagate to things like LossFunction

---
In reply to: [263067053](https://github.com/dotnet/machinelearning/pull/2851#discussion_r263067053) [](ancestors = 263067053,263063854,263057699,263048215,263023672,262757300)

wschin: This may not affect LossFunction because it's an interface. If LossFunction is a class, the same pattern may apply. As Tom and Ivan have mentioned somewhere, public area should be as small as possible. On the other hand, another pattern here may affect LossFunction where we have a property to hide a field.

---
In reply to: [263068114](https://github.com/dotnet/machinelearning/pull/2851#discussion_r263068114) [](ancestors = 263068114,263067053,263063854,263057699,263048215,263023672,262757300)

src/Microsoft.ML.FastTree/Training/EarlyStoppingCriteria.cs
```diff
@@ -262,35 +384,54 @@ public override bool CheckScore(float validationScore, float trainingScore, out
...
 
-        public PQEarlyStoppingCriterion(Options options, bool lowerIsBetter)
-            : base(options, lowerIsBetter) { }
+        /// <summary>
+        /// Create a rule which may terminate the training process when generality-to-progress ratio exceeds <paramref name="threshold"/>.
+        /// </summary>
+        /// <param name="threshold">The maximum ratio gap (in percentage such as 0.01 for 1% and 0.5 for 50%).</param>
+        /// <param name="windowSize">See <see cref="MovingWindowRule.WindowSize"/>.</param>
+        public GeneralityToProgressRatioRule(float threshold = 0.01f, int windowSize = 5)
+            : base(threshold, windowSize)
```
wschin:
```suggestion
            : base(threshold, windowSize: windowSize)
```

PR #2858, opened 2019-03-05
Add save/load APIs for IDataLoader

### LINKED ISSUE #2735: Add API to save/load models with their input schema

yaeldekel:
Fixes #2735.

### DIFF
--- a/src/Microsoft.ML.Core/Properties/AssemblyInfo.cs
+++ b/src/Microsoft.ML.Core/Properties/AssemblyInfo.cs
@@ -13,6 +13,7 @@ using Microsoft.ML;
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.InferenceTesting" + PublicKey.TestValue)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.StaticPipelineTesting" + PublicKey.TestValue)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.OnnxTransformerTest" + PublicKey.TestValue)]
+[assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Functional.Tests" + PublicKey.TestValue)]
 
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.EntryPoints" + PublicKey.Value)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Maml" + PublicKey.Value)]
--- a/src/Microsoft.ML.Core/Properties/AssemblyInfo.cs
+++ b/src/Microsoft.ML.Core/Properties/AssemblyInfo.cs
@@ -13,7 +13,6 @@ using Microsoft.ML;
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.InferenceTesting" + PublicKey.TestValue)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.StaticPipelineTesting" + PublicKey.TestValue)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.OnnxTransformerTest" + PublicKey.TestValue)]
-[assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Functional.Tests" + PublicKey.TestValue)]
 
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.EntryPoints" + PublicKey.Value)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Maml" + PublicKey.Value)]
--- a/src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.IO;
 using Microsoft.ML;
 using Microsoft.ML.Data;
 using Microsoft.ML.Runtime;
@@ -19,6 +18,10 @@ namespace Microsoft.ML.Data
     public sealed class CompositeDataLoader<TSource, TLastTransformer> : IDataLoader<TSource>
         where TLastTransformer : class, ITransformer
     {
+        private const string LoaderDirectory = "Loader";
+        private const string LegacyLoaderDirectory = "Reader";
+        private const string TransformerDirectory = TransformerChain.LoaderSignature;
+
         /// <summary>
         /// The underlying data loader.
         /// </summary>
@@ -39,9 +42,9 @@ namespace Microsoft.ML.Data
 
         private CompositeDataLoader(IHost host, ModelLoadContext ctx)
         {
-            if (!ctx.LoadModelOrNull<IDataLoader<TSource>, SignatureLoadModel>(host, out Loader, ModelOperationsCatalog.LegacyLoaderDirectory))
-                ctx.LoadModel<IDataLoader<TSource>, SignatureLoadModel>(host, out Loader, ModelOperationsCatalog.LoaderDirectory);
-            ctx.LoadModel<TransformerChain<TLastTransformer>, SignatureLoadModel>(host, out Transformer, ModelOperationsCatalog.TransformerDirectory);
+            if (!ctx.LoadModelOrNull<IDataLoader<TSource>, SignatureLoadModel>(host, out Loader, LegacyLoaderDirectory))
+                ctx.LoadModel<IDataLoader<TSource>, SignatureLoadModel>(host, out Loader, LoaderDirectory);
+            ctx.LoadModel<TransformerChain<TLastTransformer>, SignatureLoadModel>(host, out Transformer, TransformerDirectory);
         }
 
         private static CompositeDataLoader<TSource, TLastTransformer> Create(IHostEnvironment env, ModelLoadContext ctx)
@@ -90,11 +93,11 @@ namespace Microsoft.ML.Data
             ctx.CheckAtModel();
             ctx.SetVersionInfo(GetVersionInfo());
 
-            ctx.SaveModel(Loader, ModelOperationsCatalog.LoaderDirectory);
-            ctx.SaveModel(Transformer, ModelOperationsCatalog.TransformerDirectory);
+            ctx.SaveModel(Loader, LoaderDirectory);
+            ctx.SaveModel(Transformer, TransformerDirectory);
         }
 
-        internal const string Summary = "A loader that encapsulates a loader and a transformer chain.";
+        internal const string Summary = "A model loader that encapsulates a data loader and a transformer chain.";
 
         internal const string LoaderSignature = "CompositeLoader";
         private static VersionInfo GetVersionInfo()
--- a/src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
@@ -18,9 +18,9 @@ namespace Microsoft.ML.Data
     public sealed class CompositeDataLoader<TSource, TLastTransformer> : IDataLoader<TSource>
         where TLastTransformer : class, ITransformer
     {
+        internal const string TransformerDirectory = TransformerChain.LoaderSignature;
         private const string LoaderDirectory = "Loader";
         private const string LegacyLoaderDirectory = "Reader";
-        private const string TransformerDirectory = TransformerChain.LoaderSignature;
 
         /// <summary>
         /// The underlying data loader.
--- a/src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
@@ -98,29 +98,6 @@ namespace Microsoft.ML.Data
             ctx.SaveModel(Transformer, TransformerDirectory);
         }
 
-        /// <summary>
-        /// Save the contents to a stream, as a "model file".
-        /// </summary>
-        public void SaveTo(IHostEnvironment env, Stream outputStream)
-        {
-            Contracts.CheckValue(env, nameof(env));
-            env.CheckValue(outputStream, nameof(outputStream));
-
-            env.Check(outputStream.CanWrite && outputStream.CanSeek, "Need a writable and seekable stream to save");
-            using (var ch = env.Start("Saving pipeline"))
-            {
-                using (var rep = RepositoryWriter.CreateNew(outputStream, ch))
-                {
-                    ch.Trace("Saving data loader");
-                    ModelSaveContext.SaveModel(rep, Loader, LoaderDirectory);
-
-                    ch.Trace("Saving transformer chain");
-                    ModelSaveContext.SaveModel(rep, Transformer, TransformerDirectory);
-                    rep.Commit();
-                }
-            }
-        }
-
         internal const string Summary = "A loader that encapsulates a loader and a transformer chain.";
 
         internal const string LoaderSignature = "CompositeLoader";
--- a/src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
@@ -3,7 +3,12 @@
 // See the LICENSE file in the project root for more information.
 
 using System.IO;
+using Microsoft.ML;
+using Microsoft.ML.Data;
 using Microsoft.ML.Runtime;
+
+[assembly: LoadableClass(CompositeDataLoader<IMultiStreamSource, ITransformer>.Summary, typeof(CompositeDataLoader<IMultiStreamSource, ITransformer>), null, typeof(SignatureLoadModel),
+    "Composite Loader", CompositeDataLoader<IMultiStreamSource, ITransformer>.LoaderSignature)]
 
 namespace Microsoft.ML.Data
 {
@@ -14,6 +19,10 @@ namespace Microsoft.ML.Data
     public sealed class CompositeDataLoader<TSource, TLastTransformer> : IDataLoader<TSource>
         where TLastTransformer : class, ITransformer
     {
+        private const string LoaderDirectory = "Loader";
+        private const string LegacyLoaderDirectory = "Reader";
+        private const string TransformerDirectory = TransformerChain.LoaderSignature;
+
         /// <summary>
         /// The underlying data loader.
         /// </summary>
@@ -30,6 +39,24 @@ namespace Microsoft.ML.Data
 
             Loader = loader;
             Transformer = transformerChain ?? new TransformerChain<TLastTransformer>();
+        }
+
+        private CompositeDataLoader(IHost host, ModelLoadContext ctx)
+        {
+            if (!ctx.LoadModelOrNull<IDataLoader<TSource>, SignatureLoadModel>(host, out Loader, LegacyLoaderDirectory))
+                ctx.LoadModel<IDataLoader<TSource>, SignatureLoadModel>(host, out Loader, LoaderDirectory);
+            ctx.LoadModel<TransformerChain<TLastTransformer>, SignatureLoadModel>(host, out Transformer, TransformerDirectory);
+        }
+
+        private static CompositeDataLoader<TSource, TLastTransformer> Create(IHostEnvironment env, ModelLoadContext ctx)
+        {
+            Contracts.CheckValue(env, nameof(env));
+            IHost h = env.Register(LoaderSignature);
+
+            h.CheckValue(ctx, nameof(ctx));
+            ctx.CheckAtModel(GetVersionInfo());
+
+            return h.Apply("Loading Model", ch => new CompositeDataLoader<TSource, TLastTransformer>(h, ctx));
         }
 
         /// <summary>
@@ -61,6 +88,16 @@ namespace Microsoft.ML.Data
             return new CompositeDataLoader<TSource, TNewLast>(Loader, Transformer.Append(transformer));
         }
 
+        void ICanSaveModel.Save(ModelSaveContext ctx)
+        {
+            Contracts.CheckValue(ctx, nameof(ctx));
+            ctx.CheckAtModel();
+            ctx.SetVersionInfo(GetVersionInfo());
+
+            ctx.SaveModel(Loader, LoaderDirectory);
+            ctx.SaveModel(Transformer, TransformerDirectory);
+        }
+
         /// <summary>
         /// Save the contents to a stream, as a "model file".
         /// </summary>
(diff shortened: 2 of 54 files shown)


yaeldekel:
Copying all the comments from #2850.

yaeldekel:
>         public void SaveTo(IHostEnvironment env, Stream outputStream)

Do we want this public API on `CompositeDataLoader`?


---
Refers to: src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs:104 in 495740e. [](commit_id = 495740ed62857695a44e5e03dedfc88520daceb1, deletion_comment = False)

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -32,13 +32,34 @@ internal ModelOperationsCatalog(IHostEnvironment env)
         /// <param name="stream">A writeable, seekable stream to save to.</param>
         public void Save(ITransformer model, Stream stream) => model.SaveTo(_env, stream);
 
+        public void Save<TSource>(IDataLoader<TSource> model, Stream stream)
```
yaeldekel: >Save [](start = 20, length = 4)

Add xml comments for new public APIs.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -32,13 +32,34 @@ internal ModelOperationsCatalog(IHostEnvironment env)
...
+            Save(new CompositeDataLoader<TSource, ITransformer>(loader, new TransformerChain<ITransformer>(model)), stream);
+
         /// <summary>
         /// Load the model from the stream.
         /// </summary>
         /// <param name="stream">A readable, seekable stream to load from.</param>
         /// <returns>The loaded model.</returns>
         public ITransformer Load(Stream stream) => TransformerChain.LoadFrom(_env, stream);
 
+        public IDataLoader<IMultiStreamSource> LoadAsCompositeDataLoader(Stream stream)
```
yaeldekel: >IDataLoader [](start = 15, length = 11)

The subtle differences between usage here and usage of the above might be a bit hard of a pill to swallow.

yaeldekel:
@TomFinley commented in #2850: Were we going to try to address the problem of saving schemas? Not all data sources we train a transform chain with will be coming from a loader.

yaeldekel:
Yes, I will address that in the next iterations.

---
In reply to: [469851370](https://github.com/dotnet/machinelearning/pull/2858#issuecomment-469851370) [](ancestors = 469851370)

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
yaeldekel: >public IDataLoader<IMultiStreamSource> LoadAsCompositeDataLoader(Stream stream) [](start = 8, length = 79)

Should this return `CompositeDataLoader` instead of `IDataLoader`?  The name is: `LoadAsCompositeDataLoader`, which makes me think it should return a `CompositeDataLoader`

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -57,19 +57,19 @@
     "Naive Calibration Executor",
     NaiveCalibrator.LoaderSignature)]
 
-[assembly: LoadableClass(typeof(ValueMapperCalibratedModelParameters<IPredictorProducing<float>, ICalibrator>), null, typeof(SignatureLoadModel),
+[assembly: LoadableClass(typeof(ValueMapperCalibratedModelParameters<object, ICalibrator>), null, typeof(SignatureLoadModel),
```
yaeldekel: >object [](start = 69, length = 6)

@TomFinley's comment from #2850:
This is a bit stronger medicine than what I was suggesting! (I don't object per se, but it did entail a bit more work.) I suppose it's a bit more "honest" than the old way -- marker interfaces are more or less a promise you can't keep -- but they are still useful on internal code where you have complete control over the implementation (as we do as they are internal).

So I think this should be fine, I just wonder, are we fully comfortable with this?

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
yaeldekel: >public void Save<TSource>(IDataLoader<TSource> model, Stream stream) [](start = 8, length = 68)

@TomFinley's comment from #2850: 
This I think this is not correct... we need to save both the loader (or failing that, input schema) as well as the transform model to a stream. Not two separate things.

I recall the mistake that was made with predictor models and transform models in entry-points, where we viewed these as single things saved to a stream, as opposed to having to capture the entire pipeline. This mistake was inherited by ML.NET v0.1, which led to all sorts of grief when people realized, yes, saving the loader is really freakin' important. So I think we're just making the same mistake again. A model isn't just a transform model, or a loader, or an input schema, it I think needs to capture all of them, and I think we need a change a bit more fundamental than merely adding a few methods here and there. We could have done that after v1.0. I think the problem is that the API here is just misleading.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
yaeldekel: But this is in the public API, since when we load it from a serialized model this would be the type of the object. And then if you want to access `SubModel` (like in the test I added `LoadModelAndExtractPredictor`) you can't do it if it is generic on `IPredictorProducing`.

Another option, is to introduce an abstract class `CalibratedModelParametersBase` that is not generic, where `SubModel` is an object (I made this change to one of the classes in the last commit, so we can decide which option we prefer).

@TomFinley, @eerhardt


---
In reply to: [262684849](https://github.com/dotnet/machinelearning/pull/2858#discussion_r262684849) [](ancestors = 262684849)

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
yaeldekel: This is an API to save an `IDataLoader`, but it can be used to save the entire model like this:

```
var loader = ml.Model.CreateTextLoader(...);
var pipeline = loader.Append(/* define a transformer here */);
var model = pipeline.Fit(/* give it an IMultiStreamSource */);
ml.Model.Save(model, fs);
```
`model` here is a `CompositeDataLoader`. 
Perhaps I should define this method to take a `CompositeDataLoader` instead of an `IDataLoader`?




---
In reply to: [262685374](https://github.com/dotnet/machinelearning/pull/2858#discussion_r262685374) [](ancestors = 262685374)

yaeldekel: I will add the code above to the test.

---
In reply to: [262756744](https://github.com/dotnet/machinelearning/pull/2858#discussion_r262756744) [](ancestors = 262756744,262685374)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
yaeldekel: If we decide to go with the abstract base class, I can do something similar for the `PredictionTransformer`s, except there it would have to be an interface... between `ITransformer` and `IPredictionTransformer<TModel>`.

---
In reply to: [262739008](https://github.com/dotnet/machinelearning/pull/2858#discussion_r262739008) [](ancestors = 262739008,262684849)

TomFinley: The other way of course is to just have the type restriction. Then these things become `IPredictionTransformer<object>` via covariance/contravariance (can't remember which this is at the moment). In fact I'd kind of like that even if we go this route.

TomFinley: > The other way of course is to just have the type restriction. Then these things become `IPredictionTransformer<object>` via covariance/contravariance (can't remember which this is at the moment). In fact I'd kind of like that even if we go this route, and introduce a "super-interface."

TomFinley: I think we kind of needed it in the case you mentioned though, because calibrated models are a class, not an interface, and there's no such thing as these `in`/`out` modifiers on type parameters for classes as there are with interfaces. If we can use that to our advantage to avoid complicating the type hierarchy, I think I'd prefer to.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
TomFinley: The problem that I see is a bit different. I understand that it's *possible* to do the right thing, but I'd like our API to be changed in such a way that doing the right thing (saving the input schema with the model) becomes easy and convenient, and doing the wrong thing (not saving the input schema) becomes less convenient, or even impossible. (Since if an `ITransformer` was trained, surely there must have been an input schema *somewhere* in that process.)

So, what I am saying is the current scheme for saving and loading models that takes a stream and an `ITransformer` and *that's it*, should go away, since it is confusing and leading people down the right path. It's not that I think your change doesn't make the right thing possible, I'm saying that by having the old way alongside it we're going to forever have the question of, "I had this problem, I don't know my input features now," and we have to forever explain, "yeah, we have this booby trap in our code, you should have used this other method, your model is now incomplete forever." Which they will, because, on the face of it, it may not be obvious to people why merely saving an `ITransformer` is insufficient. I'd rather never get that question, and I'd do that by removing the current API and replacing it with something better, that *always* takes a schema or loader. (And on the flip side of loading, always returns a loader or schema.)

src/Microsoft.ML.Data/DataLoadSave/TransformerChain.cs
```diff
@@ -236,7 +236,7 @@ IRowToRowMapper ITransformer.GetRowToRowMapper(DataViewSchema inputSchema)
     /// <summary>
     /// Saving/loading routines for transformer chains.
     /// </summary>
-    internal static class TransformerChain
+    public static class TransformerChain
```
TomFinley: >TransformerChain [](start = 24, length = 16)

Why?

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -32,13 +32,34 @@ internal ModelOperationsCatalog(IHostEnvironment env)
         /// <param name="stream">A writeable, seekable stream to save to.</param>
         public void Save(ITransformer model, Stream stream) => model.SaveTo(_env, stream);
```
TomFinley: >Save [](start = 20, length = 4)

I'd also get rid of this.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
TomFinley: Actually I'm thinking about this now, and I don't like the idea of there being a super-interface, because we already have this hierarchy:

https://github.com/dotnet/machinelearning/blob/adb53c41fd8c612df30f8d0213b7f459ef113724/src/Microsoft.ML.Data/Prediction/IPredictionTransformer.cs#L16

And here:

https://github.com/dotnet/machinelearning/blob/adb53c41fd8c612df30f8d0213b7f459ef113724/src/Microsoft.ML.Data/Prediction/IPredictionTransformer.cs#L27

We could have another interface, another non-generic `IPredictionTransformer` with `object Model { get; }`, and another non-generic `ISingleFeaturePredictionTransformer` that descends from it, with the generic `IPredictionTransformer`. Or we could just make things so that we *know* we can always call them `IPredictionTransformer<object>`.

test/Microsoft.ML.Tests/Scenarios/Api/Estimators/DeserializationTests.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.IO;
+using System.Linq;
+using Microsoft.ML.Calibrators;
+using Microsoft.ML.Data;
+using Microsoft.ML.RunTests;
+using Microsoft.ML.Trainers.FastTree;
+using Xunit;
+
+namespace Microsoft.ML.Tests.Scenarios.Api
+{
```
artidoro: Should these go in Microsoft.ML.Functional.Tests?
They seem to be in the same spirit of Rogan's tests.

test/Microsoft.ML.Tests/Scenarios/Api/Estimators/DeserializationTests.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            var gam = (((loadedModel as TransformerChain<ITransformer>).LastTransformer
+                as BinaryPredictionTransformer<object>).Model
+                as CalibratedModelParametersBase).SubModel
+                as BinaryClassificationGamModelParameters;
+            Assert.NotNull(gam);
+        }
+
+        [Fact]
+        public void SaveAndLoadModelWithLoader()
```
artidoro: Could we also have a test for the scenario where the loader is part of the pipeline and we try to save the entire pipeline? Something that looks like the following:

```csharp
var pipeline = mlContext.Data.CreateTextLoader(....)
                        .Append(mlContext.Transforms.Conversions.Hash(...)
                        .Append(....)

var model = pipeline.Fit(stream);

mlContext.Model.Save(model, fs);
```

artidoro: Would be good to make sure we can then use `LoadAsCompositeDataLoader`.

---
In reply to: [263153288](https://github.com/dotnet/machinelearning/pull/2858#discussion_r263153288) [](ancestors = 263153288)

src/Microsoft.ML.Core/Properties/AssemblyInfo.cs
```diff
@@ -13,6 +13,7 @@
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.InferenceTesting" + PublicKey.TestValue)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.StaticPipelineTesting" + PublicKey.TestValue)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.OnnxTransformerTest" + PublicKey.TestValue)]
+[assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Functional.Tests" + PublicKey.TestValue)]
```
TomFinley: So, the express purpose of this "functional tests" is that it does not have access to internals. This to avoid the problem of us inadvertantly adding something we *need* to internals, so we have one project where we can be sure, "these tests are using the actual power of the public API." See ref #2470, and its motivations. Maybe @rogancarr and @eerhardt can express more about this, though I think the motivations are captured adequately by the PR and its linked issue.

src/Microsoft.ML.Data/DataLoadSave/TransformerChain.cs
```diff
@@ -250,13 +250,15 @@ private static TransformerChain<ITransformer> Create(IHostEnvironment env, Model
         public static void SaveTo(this ITransformer transformer, IHostEnvironment env, Stream outputStream)
             => new TransformerChain<ITransformer>(transformer).SaveTo(env, outputStream);
 
-        public static TransformerChain<ITransformer> LoadFrom(IHostEnvironment env, Stream stream)
+        public static ITransformer LoadFrom(IHostEnvironment env, Stream stream)
```
TomFinley: >ITransformer [](start = 22, length = 12)

Would it have been bad in the case where the source wasn't a chain (the `@"Model"` branch below, to just put it in a chain?

src/Microsoft.ML.Data/DataLoadSave/TransformerChain.cs
```diff
@@ -250,13 +250,15 @@ private static TransformerChain<ITransformer> Create(IHostEnvironment env, Model
...
         {
             using (var rep = RepositoryReader.Open(stream, env))
             {
                 try
                 {
-                    ModelLoadContext.LoadModel<TransformerChain<ITransformer>, SignatureLoadModel>(env, out var transformerChain, rep, LoaderSignature);
+                    ModelLoadContext.LoadModelOrNull<ITransformer, SignatureLoadModel>(env, out var transformerChain, rep, LoaderSignature);
+                    if (transformerChain == null)
+                        ModelLoadContext.LoadModel<ITransformer, SignatureLoadModel>(env, out transformerChain, rep, $@"Model\{LoaderSignature}");
```
TomFinley: >$ [](start = 117, length = 1)

I think this formatting is somewhat locale sensitive. I might have preferred the simpler `"Model\\" + LoaderSignature`.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -31,14 +40,100 @@ internal ModelOperationsCatalog(IHostEnvironment env)
...
 
         /// <summary>
-        /// Load the model from the stream.
+        /// Load the model and its input schema from the stream.
         /// </summary>
         /// <param name="stream">A readable, seekable stream to load from.</param>
+        /// <param name="inputSchema">Will contain the input schema for the model.</param>
         /// <returns>The loaded model.</returns>
-        public ITransformer Load(Stream stream) => TransformerChain.LoadFrom(_env, stream);
+        public ITransformer Load(Stream stream, out DataViewSchema inputSchema)
```
TomFinley: >DataViewSchema inputSchema [](start = 52, length = 26)

Sometimes this will not be present. Should we clarify that this can be null in some cases?

TomFinley: >Load(Stream stream, out DataViewSchema inputSchema [](start = 28, length = 50)

So I see above we have methods to save either a loader, or a schema, with a transform. I see this method to load the transform, with an input schema. What about to load the transformer, with the loader?

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -13,6 +17,11 @@ namespace Microsoft.ML
     /// </summary>
     public sealed class ModelOperationsCatalog : IInternalCatalog
     {
+        internal const string LoaderDirectory = "Loader";
```
TomFinley: >LoaderDirectory [](start = 30, length = 15)

So, as far as I can see, these are used entirely from `CompositeDataLoader`, though maybe I missed something. So why are they here?

src/Microsoft.ML.Core/Properties/AssemblyInfo.cs
eerhardt: YES! This should be noted to ALL devs - we must NEVER have InternalsVisibleTo the Functional.Tests. That defeats the purpose.

Note the comment in the Functional.Tests.csproj:

https://github.com/dotnet/machinelearning/blob/40abffc4eb98990c86804a05ca707a467edbe3d6/test/Microsoft.ML.Functional.Tests/Microsoft.ML.Functional.Tests.csproj#L3-L7

Maybe we need to make that comment stand out more. @yaeldekel - can you make it ALL CAPS or something that will catch people's eye, so they don't make the same change/mistake you just did? Maybe put it before and after the `SignAssembly` property.

```xml
  <PropertyGroup>
    <!-- DON'T CHANGE THIS!!! We are turning off strong naming to ensure we never add `InternalsVisibleTo` for these tests -->
    <!-- DON'T CHANGE THIS!!! We are turning off strong naming to ensure we never add `InternalsVisibleTo` for these tests -->
  </PropertyGroup>
```

src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
```diff
@@ -76,47 +113,27 @@ public void SaveTo(IHostEnvironment env, Stream outputStream)
...
-    /// </summary>
-    [BestFriend]
-    internal static class CompositeDataLoader
-    {
-        /// <summary>
-        /// Save the contents to a stream, as a "model file".
-        /// </summary>
-        public static void SaveTo<TSource>(this IDataLoader<TSource> loader, IHostEnvironment env, Stream outputStream)
-            => new CompositeDataLoader<TSource, ITransformer>(loader).SaveTo(env, outputStream);
+        internal const string Summary = "A loader that encapsulates a loader and a transformer chain.";
```
sfilipi: >A loader that encapsulates a loader [](start = 41, length = 35)

Maybe: 'A model loader that encapsulates a text loader and a transformer chain". so it sounds less tautological.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -147,8 +147,8 @@ internal interface ISelfCalibratingPredictor
     [BestFriend]
     internal interface IWeaklyTypedCalibratedModelParameters
     {
-        IPredictorProducing<float> WeeklyTypedSubModel { get; }
-        ICalibrator WeeklyTypedCalibrator { get; }
+        IPredictorProducing<float> WeaklyTypedSubModel { get; }
+        ICalibrator WeaklyTypedCalibrator { get; }
```
sfilipi: >WeaklyTypedCalibrator [](start = 20, length = 21)

nice :)

src/Microsoft.ML.Data/DataLoadSave/TransformerChain.cs
yaeldekel: I ended up doing the opposite of that - so the transform chain is never nested in the `Model` directory.

---
In reply to: [264740060](https://github.com/dotnet/machinelearning/pull/2858#discussion_r264740060) [](ancestors = 264740060)

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
yaeldekel: I am adding a Load overload that produces a transformer and a loader. If the model was saved by saving the transformer+schema, do we prefer throwing an exception in this case (since there is no loader), or defining this method as `TryLoad`?

---
In reply to: [264742808](https://github.com/dotnet/machinelearning/pull/2858#discussion_r264742808) [](ancestors = 264742808)

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -26,19 +32,126 @@ internal ModelOperationsCatalog(IHostEnvironment env)
             Explainability = new ExplainabilityTransforms(this);
         }
 
+        private void Save<TSource>(DataViewSchema schema, IDataLoader<TSource> model, Stream stream)
```
TomFinley: >ader<TSource> m [](start = 65, length = 15)

In this case where we have a data loader, we *have* the schema already, do we not?

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -26,19 +32,126 @@ internal ModelOperationsCatalog(IHostEnvironment env)
...
+        /// <param name="loader">The loader that was used to create data to train the model</param>
+        /// <param name="model">The trained model to be saved</param>
+        /// <param name="stream">A writeable, seekable stream to save to.</param>
+        public void Save<TSource>(IDataLoader<TSource> loader, ITransformer model, Stream stream) =>
+            Save(loader.GetOutputSchema(), new CompositeDataLoader<TSource, ITransformer>(loader, new TransformerChain<ITransformer>(model)), stream);
+
+        /// <summary>
+        /// Save a transformer model and the schema of the data that was used to train it to the stream.
+        /// </summary>
+        /// <param name="inputSchema">The schema of the input to the transformer.</param>
```
TomFinley: >The schema of the input to the transformer [](start = 38, length = 42)

While I think having the argument is important to *encourage* people to save the input model, we should also allow them to explicitly not save the input in cases where it is disallowed. The goal is to lead them in the right direction -- the old ways were merely misleading, not always wrong, so I think we should allow this to be `null`.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
TomFinley: The usage of `Try` suggests there's something wrong with the approach, and that the absence of a schema represents failure. I'm not sure I agree with that -- I think the situation where you have a transformer with no loader or input schema is a valid one we should *allow*, just not *encourage* (since it is a minority scenario). Now then, I think we should encourage people to *think* about saving it, and I think a good way to encourage people to make a deliberate decision is to force them to consciously feed us a `null`, as suggested above. But that also means that we have to allow for the possibility upon loading that the schema *will* be `null`, and I'd think we'd call this not an error condition. What you think?

---
In reply to: [265196459](https://github.com/dotnet/machinelearning/pull/2858#discussion_r265196459) [](ancestors = 265196459,264742808)

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -31,14 +38,161 @@ internal ModelOperationsCatalog(IHostEnvironment env)
...
+            }
+        }
+
+        /// <summary>
+        /// Load a transformer model and a data loader model from the stream.
+        /// </summary>
+        /// <param name="stream">A readable, seekable stream to load from.</param>
+        /// <param name="loader">The data loader from the model stream.</param>
+        /// <returns>The transformer model from the model stream.</returns>
+        public ITransformer Load(Stream stream, out IDataLoader<IMultiStreamSource> loader)
```
TomFinley: >Load [](start = 28, length = 4)

It would be nice if this was `LoadWithDataLoader`, so that we could have `mlContext.Model.Load(stream, out var thing)` without ambiguity.

src/Microsoft.ML.Data/DataView/DataViewConstructionUtils.cs
```diff
@@ -46,6 +46,57 @@ internal static class DataViewConstructionUtils
             return new StreamingDataView<TRow>(env, data, internalSchemaDefn);
         }
 
+        public static StreamingDataView<TRow> CreateFromEnumerable<TRow>(IHostEnvironment env, IEnumerable<TRow> data,
```
TomFinley: >CreateFromEnumerable [](start = 46, length = 20)

So, I don['t object exactly, it seems positive, but I'm not sure how this relates to the original issue?

yaeldekel: I added these so that I could add tests show what you can do with the schema once you load it with the new API.

---
In reply to: [266546399](https://github.com/dotnet/machinelearning/pull/2858#discussion_r266546399) [](ancestors = 266546399)

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -31,24 +37,160 @@ internal ModelOperationsCatalog(IHostEnvironment env)
...
-        /// Load the model from a file path.
+        /// Load the model and its input schema from the stream.
         /// </summary>
-        /// <param name="modelPath">Path to model.</param>
-        /// <returns>The loaded model.</returns>
-        public ITransformer Load(string modelPath)
+        /// <param name="stream">A readable, seekable stream to load from.</param>
+        /// <returns>A model of type <see cref="CompositeDataLoader{IMultiStreamSource, ITransformer}"/> containing the loader
+        /// and the transformer chain.</returns>
+        public IDataLoader<IMultiStreamSource> Load(Stream stream)
```
Ivanidzo4ka: >IDataLoader [](start = 15, length = 11)

it's kinda sad what now we have `IDataLoader` along side with `ITransformer` since it's increase complexity. Would be nice to have some documentation/samples which would explain in which case I need to save/load one thing, and in which another.

Also can I now have slot names in my model?

Issue #2828, opened 2019-03-04
Scrubbing schema related transformers


Ivanidzo4ka:
Sub task of #2827
Copy, SelectColumns, DropColumns, Concat

PR #2863, opened 2019-03-05
Scrubbing schema related transforms

Ivanidzo4ka:
Should fix https://github.com/dotnet/machinelearning/issues/2828

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/ColumnSelecting.cs
+++ b/src/Microsoft.ML.Data/Transforms/ColumnSelecting.cs
@@ -9,9 +9,7 @@ using Microsoft.Data.DataView;
 using Microsoft.ML;
 using Microsoft.ML.CommandLine;
 using Microsoft.ML.Data;
-using Microsoft.ML.EntryPoints;
 using Microsoft.ML.Internal.Utilities;
-using Microsoft.ML.Model;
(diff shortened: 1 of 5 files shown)


src/Microsoft.ML.Data/Transforms/ColumnSelecting.cs
```diff
@@ -140,12 +138,12 @@ public sealed class ColumnSelectingTransformer : ITransformer
...
-        public IEnumerable<string> SelectColumns => _selectedColumns.AsReadOnly();
+        internal IEnumerable<string> SelectColumns => _selectedColumns.AsReadOnly();
 
-        public bool KeepColumns { get; }
+        internal bool KeepColumns { get; }
 
-        public bool KeepHidden { get; }
-        public bool IgnoreMissing { get; }
+        internal bool KeepHidden { get; }
+        internal bool IgnoreMissing { get; }
```
wschin: Do we really need to hide them? They are some direct states of public options.

Ivanidzo4ka: Maybe it just me been paranoid and I want to hide everything I can.
On other hand we need to provide this information for each transformer, like term dictionary or list of columns we been concatenating for concat, and we don't have conclusion how to do it yet.

Issue #655, opened 2018-08-07
OpenMP support for MKL library and SymSGD.


codemzs:
Current MKL library was not built with parallel option to not have dependency on OpenMP because we are still waiting for guidance from legal department on redistributing open mp binary.

Issue #2211, opened 2019-01-23
Re-build the MKL binaries together with OpenMP and update SymSGD to make use of parallelism


sfilipi:
The current binaries for MlNetMKLDeps are build with sequential threading, and SymSgd has parallelism turned off. 

Now that the legal issues around packaging and re-distributing Open MP are clarified, re-build the ML bits with threading=parallel and update SymSGD.

singlis:
So when enabling OpenMP for mac this hit multiple issues as MKL uses Intel's OpenMP implementation and we are linking against default OpenMP (installed via brew install libomp).

While this does build, there is an error that comes up during tests that we are mixing two implementations of OpenMP:

> The active test run was aborted. Reason: OMP: Error #15: Initializing libiomp5.dylib, but found libomp.dylib already initialized.
> OMP: Hint This means that multiple copies of the OpenMP runtime have been linked into the program. That is dangerous, since it can degrade performance or cause incorrect results. The best thing to do is to ensure that only a single OpenMP runtime is linked into the process, e.g. by avoiding static linking of the OpenMP runtime in any library. As an unsafe, unsupported, undocumented workaround you can set the environment variable KMP_DUPLICATE_LIB_OK=TRUE to allow the program to continue to execute, but that may cause crashes or silently produce incorrect results. For more information, please see http://www.intel.com/software/products/support/.
> 

I am following up with Intel for guidance on how to resolve this.

PR #2867, opened 2019-03-06
Updating MKL

singlis:
- Updating ml.net to use mlnetDep nuget 0.0.0.9. This does a couple of things:
1) Updates ML.Net to use MKL version 2008.3.10
2) Enables OpenMP for MKL
3) Project changes for handling the OpenMP library, note that we only copy this library on windows builds. Both Mac and Linux require this library to be installed.
- Enables OpenMP for SymSGDNative
- NumberOfThreads parameter now set the number of threads to use for SymSGDNative.

This related to issue #2211 
fixes https://github.com/dotnet/machinelearning/issues/655


Ivanidzo4ka:
https://github.com/dotnet/machinelearning/tree/master/docs/building/MlNetMklDeps
does this one requires any update?

src/Native/SymSgdNative/SymSgdNative.cpp
```diff
@@ -240,7 +238,6 @@ void InitializeState(int totalNumInstances, int* instSizes, int** instIndices, f
         TuneNumLocIter(numLocIter, totalNumInstances, instSizes, numThreads);
 
     state->WeightScaling = 1.0f;
-#if defined(USE_OMP)
```
Ivanidzo4ka: >#if defined(USE_OMP)  [](start = 0, length = 21)

why it's going away?

src/Native/build.proj
```diff
@@ -72,7 +72,10 @@
     
     <Copy SourceFiles="$(PackagesDir)mlnetmkldeps\$(MlNetMklDepsPackageVersion)\runtimes\$(PackageRid)\native\$(NativeLibPrefix)MklImports$(NativeLibExtension)"
           DestinationFolder="$(NativeAssetsBuiltPath)" />   
-    
+
+    <Copy Condition="'$(OS)' == 'Windows_NT'" SourceFiles="$(PackagesDir)mlnetmkldeps\$(MlNetMklDepsPackageVersion)\runtimes\$(PackageRid)\native\libiomp5md$(NativeLibExtension)"
```
Ivanidzo4ka: >Windows_NT [](start = 33, length = 10)

No support for Linux and MacOS?

src/Native/SymSgdNative/SymSgdNative.cpp
eerhardt: Maybe we should just define `USE_OMP` in the CMake files instead?

build/ci/phase-template.yml
```diff
@@ -33,13 +33,13 @@ phases:
           ${{ insert }}: ${{ parameters.customMatrixes }}
       ${{ insert }}: ${{ parameters.queue }}
     steps:
+    - ${{ if eq(parameters.name, 'MacOS') }}:
```
eerhardt: Note - you are going to need to update the "official" build as well:

https://github.com/dotnet/machinelearning/blob/6f4e055b113b2bac024b7241ba469b9c20453fbe/build/vsts-ci.yml#L37-L60

singlis: thanks - ill update that as well.

singlis:
It does, I have another PR I will post here soon.

---
In reply to: [469972694](https://github.com/dotnet/machinelearning/pull/2867#issuecomment-469972694) [](ancestors = 469972694)

src/Native/SymSgdNative/SymSgdNative.cpp
singlis: I'm fine with adding back and adding the USE_OMP to  CMakefile file.

---
In reply to: [262995598](https://github.com/dotnet/machinelearning/pull/2867#discussion_r262995598) [](ancestors = 262995598)

src/Native/build.proj
singlis: So this works on all platforms. For Linux and MacOS, we rely on openmp to be installed on the system. 

---
In reply to: [262795888](https://github.com/dotnet/machinelearning/pull/2867#discussion_r262795888) [](ancestors = 262795888)

singlis:
This is pending changes to update the linux container.

singlis:
PR for adding openmp support: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/94

singlis:
Minor update on this - the build changes are committed, but the build that generates the docker containers failed (due to something unrelated to the new changes). I am currently waiting for the build to retried.

build/ci/phase-template.yml
```diff
@@ -33,13 +33,16 @@ phases:
...
+      - script: brew update && brew install libomp && brew link libomp --force
+        displayName: Install build dependencies
     - script: $(_buildScript) -$(_configuration) -buildArch=$(_arch)
       displayName: Build
-    - script: $(_buildScript) -- /t:DownloadExternalTestFiles /p:IncludeBenchmarkData=$(_includeBenchmarkData)
-      displayName: Download Benchmark Data
     - ${{ if eq(parameters.name, 'MacOS') }}:
-      - script: brew update && brew install libomp mono-libgdiplus gettext && brew link gettext --force
+      - script: brew update && brew install mono-libgdiplus gettext && brew link gettext --force
```
eerhardt: Why split this into 2 steps?  Why not just `brew install` all the stuff in one step?

build/vsts-ci.yml
```diff
@@ -49,6 +49,8 @@ phases:
     - agent.os -equals Darwin
   steps:
   # Only build native assets to avoid conflicts.
```
eerhardt: This comment should go below your new script - it applies to the `./build.sh` line.

build/ci/phase-template.yml
singlis: So I was differentiating between the build dependency and runtime dependencies. Its minor - so I will move to a single step.

---
In reply to: [265170687](https://github.com/dotnet/machinelearning/pull/2867#discussion_r265170687) [](ancestors = 265170687)

singlis:
@Ivanidzo4ka  and @eerhardt  the build is now passing, can you take a look at the latest update when you have a chance?

src/Native/SymSgdNative/CMakeLists.txt
```diff
@@ -1,10 +1,30 @@
 project (SymSgdNative)
 
+
 set(SOURCES
     SymSgdNative.cpp
 )
 
+if(APPLE)
```
eerhardt: Can we just put this in the `if (NOT WIN32)` section below?  Instead of having 2 different "If not windows" sections?

singlis: Hi Eric, sorry I am not quite following. This has special handling for APPLE - adding this under "NOT WIN32" would include both apple and linux which is not the same result. There is also the ELSE clause which is what I want to execute for windows and linux.  If I am misunderstanding let me know.  

---
In reply to: [266134616](https://github.com/dotnet/machinelearning/pull/2867#discussion_r266134616) [](ancestors = 266134616)

eerhardt: Sorry - I should have been clearer:

```
if(APPLE)
    # CMake has support for OpenMP, however, Apple has a version of Clang
    # that does not support openMP out of the box. Therefore
    # these commands are added to sepcifically handle the Apple Clang scenario
    # If the LLVM version of clang is used for Apple builds, this can be removed
    # and the else condition can be used instead.
    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Xpreprocessor -fopenmp")
    SET(OPENMP_LIBRARY "omp")
    include_directories("/usr/local/opt/libomp/include")
    link_directories("/usr/local/opt/libomp/lib")
else()
    find_package(OpenMP)
    if (OPENMP_FOUND)
        SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
        SET(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
    endif()
endif()

find_library(MKL_LIBRARY MklImports HINTS ${MKL_LIB_PATH})

if(NOT WIN32)
    list(APPEND SOURCES ${VERSION_FILE_PATH})
    if(NOT APPLE)
        SET(CMAKE_SKIP_BUILD_RPATH  FALSE)
        SET(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
        SET(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
        SET(CMAKE_INSTALL_RPATH "$ORIGIN/")
    endif()
endif()
````

You see how there are now 2 separate "if" checks for not Windows and Not Apple?

We can combine this together into a single set of "if" conditions:

```
if(NOT WIN32)
    // Do common non-Windows
    if(APPLE)
        // Apple stuff
    else()
        // Do non-Apple stuff
    endif()
endif()
```

Comment (singlis) [reply src/Native/SymSgdNative/CMakeLists.txt]: Here is what I came up with -- its merged with the least amount of duplication. The line that was duplicated is this: list(APPEND SOURCES ${VERSION_FILE_PATH})


---
In reply to: [266144477](https://github.com/dotnet/machinelearning/pull/2867#discussion_r266144477) [](ancestors = 266144477)

File: build/ci/phase-template.yml [hunk 7]
```diff
@@ -33,13 +33,13 @@ phases:
           ${{ insert }}: ${{ parameters.customMatrixes }}
       ${{ insert }}: ${{ parameters.queue }}
     steps:
+    - ${{ if eq(parameters.name, 'MacOS') }}:
+      - script: brew update && brew install libomp && brew install mono-libgdiplus gettext && brew link gettext --force && brew link libomp --force
```
Ivanidzo4ka: >libomp [](start = 44, length = 6)

do we still need to install libomp if we install it in vsts-ci.yml?

src/Microsoft.ML.Mkl.Components/SymSgdClassificationTrainer.cs
```diff
@@ -664,6 +664,10 @@ private TPredictor TrainCore(IChannel ch, RoleMappedData data, LinearModelParame
             int numFeatures = data.Schema.Feature.Value.Type.GetVectorSize();
             var cursorFactory = new FloatLabelCursor.Factory(data, CursOpt.Label | CursOpt.Features);
             int numThreads = 1;
+
+            if (_options.NumberOfThreads.HasValue)
+                numThreads = _options.NumberOfThreads.Value;
```
Ivanidzo4ka: >numThreads [](start = 16, length = 10)

So we now can have multithreading for SymSGD always?
If yes:
```
  /// <summary>
            /// Degree of lock-free parallelism. Determinism not guaranteed if this is set to higher than 1.
            /// Multi-threading is not supported currently.
            /// </summary>
            [Argument(ArgumentType.AtMostOnce, HelpText = "Degree of lock-free parallelism. Determinism not guaranteed. " +
                "Multi-threading is not supported currently.", ShortName = "nt")]
            public int? NumberOfThreads;
```
this need some polishing in description, but that minor.

So if it's null (which it is by default) it's would be shame to utilize only one thread.

`numThreads = NumberOfThreads ?? Environment.ProcessorCount` maybe?

build/ci/phase-template.yml
singlis: I believe it does as Eric's comment earlier was to make sure to also add this to vsts-ci.yml. My impression is that the vsts-ci.yml is for the gated build (post merge) and vsts-ci.yml is separate from phase-template.yml, but its a good question. @eerhardt can you clarify? Thanks!

src/Microsoft.ML.Mkl.Components/SymSgdClassificationTrainer.cs
singlis: Thanks for bringing this up - I had similar thoughts and actually meant to bring this up in the PR. I agree - - null should default to the number of threads that are available to openmp. Im happy to update this...

build/ci/phase-template.yml
eerhardt: they are 2 completely separate builds:

1. For PR validation and post merge CI.
2. For official build - which signs assemblies, publishes nuget packages and symbols, etc.

src/Native/SymSgdNative/CMakeLists.txt
```diff
@@ -1,25 +1,46 @@
...
     list(APPEND SOURCES ${VERSION_FILE_PATH})
-    if(NOT APPLE)
+else()
+    find_package(OpenMP)
+    if (OPENMP_FOUND)
+        SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
+        SET(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
+    endif()
+
+    if (NOT WIN32)
```
eerhardt: This is not correct. This check will never happen on an APPLE system.

Please fix this in a new PR.

eerhardt: Ah, I see it now. It is in both conditions.

We could have just made that a completely separate `if`... but I guess this is OK. It is just generally "odd" to have APPLE different and Windows/LINUX in the same condition.

Issue #2445, opened 2019-02-06
Inventory of the Microsoft.ML public types: what to rename, move, hide there. 


sfilipi:
This is the list of the types in the Microsoft.ML namespace, as seen from the docs site.  Taking a first pass at what doesn't need to be there, what needs to be hidden, and what needs to be moved to a different namespace. 

Microsoft.ML namespace type | Rename/Move/Hide |
--------------------------- | --------- |
BinaryClassificationCatalog |    | 
BinaryClassificationCatalog.BinaryClassificationTrainers|    | 
BinaryClassificationMetricsStatistics|    | 
BinaryLoaderSaverCatalog|    | 
CategoricalCatalog|    | 
ChannelMessage |   move to Microsoft.ML.Data | 
ChannelMessageKind|  move to Microsoft.ML.Data | 
ClusteringCatalog|    | 
ClusteringCatalog.ClusteringTrainers|    | 
ComponentCatalog|  move to Microsoft.ML.Core | 
ConversionsCatalog|    | 
ConversionsExtensionsCatalog|    | 
CustomMappingCatalog|    | 
DataOperationsCatalog|    | 
DataReaderExtensions|    | 
DebuggerExtensions|    | 
ExplainabilityCatalog |    | 
ExpLoss | move  to Microsoft.ML.Trainers.Loss | 
ExpLoss.Arguments |  move  to Microsoft.ML.Trainers.Loss | 
ExtensionsCatalog |    | 
FactorizationMachineExtensions |    | 
FeatureSelectionCatalog |    | 
HalLearnersCatalog |    | 
HingeLoss |  move to Microsoft.ML.Trainers.Loss | 
HingeLoss.Arguments |  move to Microsoft.ML.Trainers.Loss | 
IChannel |  move to Microsoft.ML.Data | 
IChannelProvider |  move to Microsoft.ML.Data | 
IClassificationLoss |  move to Microsoft.ML.Trainers.Loss | 
IComponentFactory | move to Microsoft.ML.Data | 
IComponentFactory<TComponent> | move to Microsoft.ML.Data | 
IComponentFactory<TArg1,TComponent> |  move to Microsoft.ML.Data | 
IComponentFactory<TArg1,TArg2,TComponent> |  move to Microsoft.ML.Data | 
IComponentFactory<TArg1,TArg2,TArg3,TComponent> |  move to Microsoft.ML.Data | 
IExceptionContext |  move to Microsoft.ML.Core | 
[...]

PR #2869, opened 2019-03-06
PFI statistics polish

TomFinley:
Fixes 2868. Also contributes minorly towards #2445, by making the statistics classes (not the context extension methods) in the `Microsoft.ML.Data` namespace.


docs/samples/Microsoft.ML.Samples/Dynamic/PermutationFeatureImportance/PfiBinaryClassificationExample.cs
```diff
@@ -66,7 +66,7 @@ public static void Example()
             // Third, some features show an *increase* in AUC. This means that the model actually improved
```
eerhardt: There's a few usages of `AUC` here and in the `Console.Writeline` below.

src/Microsoft.ML.Transforms/PermutationFeatureImportance.cs
```diff
@@ -9,19 +9,19 @@
...
 using Microsoft.Data.DataView;
 using Microsoft.ML.Data;
-using Microsoft.ML.Internal.Internallearn;
 using Microsoft.ML.Internal.Utilities;
 using Microsoft.ML.Model;
 
 namespace Microsoft.ML.Transforms
 {
-    internal static class PermutationFeatureImportance<TModel, TMetric, TResult> where TResult : MetricsStatisticsBase<TMetric>, new()
+    internal static class PermutationFeatureImportance<TModel, TMetric, TResult> where TResult : IMetricsStatistics<TMetric>
```
eerhardt: Is it not possible to keep the `new` constraint?  That seems cleaner than passing in a `Func<TResult>` IMO.

src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
```diff
@@ -411,69 +408,79 @@ protected MetricStatistics[] InitializeArray(int length)
...
+    }
+
+    /// <summary>
+    /// The <see cref="RegressionMetricsStatistics"/> class holds summary
+    /// statistics over multiple observations of <see cref="RegressionMetrics"/>.
+    /// </summary>
+    public sealed class RegressionMetricsStatistics : IMetricsStatistics<RegressionMetrics>
     {
         /// <summary>
         /// Summary Statistics for L1
```
eerhardt: do the `summary` comments here need to be updated?

src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
```diff
@@ -411,69 +408,79 @@ protected MetricStatistics[] InitializeArray(int length)
...
-    /// The BinaryClassificationMetricsStatistics class is computes summary
-    /// statistics over multiple observations of binary classification evaluation metrics.
+    /// The <see cref="BinaryClassificationMetricsStatistics"/> class holds summary
+    /// statistics over multiple observations of <see cref="BinaryClassificationMetrics"/>.
     /// </summary>
-    public sealed class BinaryClassificationMetricsStatistics : MetricsStatisticsBase<BinaryClassificationMetrics>
+    public sealed class BinaryClassificationMetricsStatistics : IMetricsStatistics<BinaryClassificationMetrics>
     {
         /// <summary>
         /// Summary Statistics for AUC
```
eerhardt: `AUC` => `AreaUnderRocCurve`?

src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
```diff
@@ -334,7 +338,10 @@ private static double[] ComputeArrayDeltas(double[] a, double[] b)
 
         #endregion
     }
+}
 
+namespace Microsoft.ML.Data
```
eerhardt: In general I like having one file per type, but am fine when we group a bunch of related types into one file.

However, having types from multiple namespaces in one file feels like it should really get split out into a separate file.

src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
```diff
@@ -508,42 +515,42 @@ public sealed class BinaryClassificationMetricsStatistics : MetricsStatisticsBas
         /// <summary>
         /// Summary statistics for AUPRC.
```
rogancarr: Area under the Precision-Recall Curve?

src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
TomFinley: I thought I'd opportunistically update them, since one was wrong and I think they all kind of benefit by describing (since it is no longer part of the type signature) what *sort* of thing they're accumulating. Do you have a concern here that I should not have updated them?

src/Microsoft.ML.Transforms/PermutationFeatureImportance.cs
TomFinley: I disagree in this case, since I do not want this thing to be publicly instantiable, and the `new()` constraint required that it have a *public* empty constructor, which I do not consider necessary for the purpose. (The public is the key problem here.)

So I had a choice: do I internally have this little factory method, or do I have a publicly accessible type just to enable this infrastructure. Seemed like a really straightforward choice to me.

src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
eerhardt: I'm happy you changed the property names to match the underlying metrics class.

My concern is that the summary comments here are now out of date. What is `L1`? `L2`? etc.

src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
TomFinley: I think what I should probably do here in the *documentation* is do a direct link against the relevant property in the corresponding metrics class, using `<see` tags. Let me update it to do that...

src/Microsoft.ML.Transforms/PermutationFeatureImportance.cs
eerhardt: Ok - the consideration I was missing here was that the type needed to have a public ctor. This approach seems right then.

src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
TomFinley: That's fine. I think what I'll do is I'll have two commits in my next push so it remains easy to review what i have done, one where I make changes, and another where I split out the file.

src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
TomFinley: Yes, will do `<see` against the relevant original metrics classes so that the relationship is a bit more explicit.

src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
TomFinley: Oh, you meant not this line, but the summary comments on the individual statistics. Yes, I am doing that.

docs/samples/Microsoft.ML.Samples/Dynamic/PermutationFeatureImportance/PfiBinaryClassificationExample.cs
TomFinley: During my original work, I did consider this but decided against this. So, the reason why we changed it at all is just to be consistent with .NET member naming guidelines, not because we had any illusions that anyone would be genuinely confused about what AUC actually meant. I think it's fine and desirable for samples to continue to call this metric by its canonical, accepted name in anything meant to be read and interpreted by a human. [Certainly elsewhere people don't have too many qualms about this.](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.auc.html#sklearn.metrics.auc)

src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
TomFinley: OK @eerhardt this is now done. Note that the commit "Move metrics statistics types to new file." is just a cut and paste job, no internal code was changed.

Issue #2778, opened 2019-02-27
Possible Bug Tensorflow in v0.11 - Invalid argument: input must be 4-dimensional[224,224,3]


CESARDELATORRE:
Starting from [a working ML.NET sample](https://github.com/dotnet/machinelearning-samples/tree/migration/v.11/samples/csharp/getting-started/DeepLearning_ImageClassification_TensorFlow/ImageClassification) scoring the standard TensorFlow Inception v3 model (it was working on 0.9 and 0.10) after migrating to 0.11 (I just needed to fix a couple of breaking changes in the API) the model is now throwing the following exception: 

Exception from TensorFlow:

```
2019-02-27 14:56:18.800272: W tensorflow/core/framework/op_kernel.cc:1273] OP_REQUIRES failed at conv_ops.cc:437 : Invalid argument: input must be 4-dimensional[224,224,3]

EXCEPTION
#########
input must be 4-dimensional[224,224,3]
         [[{{node conv2d0_pre_relu/conv}} = Conv2D[T=DT_FLOAT, data_format="NHWC", dilations=[1, 1, 1, 1], padding="SAME", strides=[1, 2, 2, 1], use_cudnn_on_gpu=true, _device="/job:localhost/replica:0/task:0/device:CPU:0"](_arg_input_0_0, conv2d0_w)]]
```

CESARDELATORRE:
This other [TensorFlow sample using the Estimator for training](https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/getting-started/DeepLearning_TensorFlowEstimator/ImageClassification.Train) is having the same issue/exception coming from TensorFlow.

prathyusha12345:
@zeahmed, @CESARDELATORRE  @yaeldekel  I am still getting the same error after migrating to Ml.Net v1.0.0-preview.

```
input must be 4-dimensional[3,224,224]
	 [[{{node conv2d0_pre_relu/conv}} = Conv2D[T=DT_FLOAT, data_format="NHWC", dilations=[1, 1, 1, 1], padding="SAME", strides=[1, 2, 2, 1], use_cudnn_on_gpu=true, _device="/job:localhost/replica:0/task:0/device:CPU:0"](_arg_input_0_0, conv2d0_w)]]
```

The pipeline is below

```
var pipeline = mlContext.Transforms.LoadImages(outputColumnName: "input", imageFolder: imagesFolder, inputColumnName: nameof(ImageNetData.ImagePath))
[...]

Issue #765, opened 2018-08-29
Add "Reshape Transform"


yaeldekel:
We need the following functionalities in order to score TensorFlow models:

1. Change the ColumnType of the column to be a different shape – this would not do anything to the data, it would just change the type of the column, to match the input dimensions the model expects (for example, if the data contains a column of length 784, and the model expects a 28x28 input).
2. Actually reshape the data – (C,H,W) to (H,W,C). This transform would also have to move data around in addition to changing the ColumnType.
3. Reshape data from RGB to other ordering of the channels. This transform would move the data around, but leave the ColumnType as it was.

2 is already implemented as an option in the PixelExtractorTransform, and 3 could also be implemented as an option in that transform, but it may be useful to have them as a separate transform, for cases where the input data doesn't necessarily come from the PixelExtractorTransform.

zeahmed:
Once we have this transform implemented, please revert the change in https://github.com/dotnet/machinelearning/pull/2935

JakeRadMSFT:
Can we increase the priority of this? It'll be useful in Azure Attach scenarios for Model Builder and ML NET CLI. Model's exported from Azure are ONNX and expect the format to have the shape [3,224,224].

JakeRadMSFT:
Just to follow up, we were able to work around it with a custom mapping. We also ended up getting a new model that didn't require it.

laurentpayot:
This issue is more than 6 years old, any update?

matouskozak:
Triage: This request is very old and might not fit the current need of ML.NET. We currently don't plan to invest in this feature. Please comment here if you have any blocked use cases which would benefit from this feature.

PR #2872, opened 2019-03-06
ImageModels in tensorflow are 4 dimensional.

Ivanidzo4ka:
Fixes https://github.com/dotnet/machinelearning/issues/2778
Well, not exactly fixes, it's more like a hack.
Proper solution would be to implement Reshape transform https://github.com/dotnet/machinelearning/issues/765

### DIFF
--- a/src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
+++ b/src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
@@ -730,7 +730,7 @@ namespace Microsoft.ML.Transforms
                     var originalShape = _parent.TFInputShapes[i];
                     var shape = originalShape.ToIntArray();
 
-                    var colTypeDims = vecType.Dimensions.Select(dim => (long)dim).ToArray();
+                    var colTypeDims = vecType.Dimensions.Prepend(1).Select(dim => (long)dim).ToArray();
                     if (shape == null)
                         _fullySpecifiedShapes[i] = new TFShape(colTypeDims);
                     else


src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -730,7 +730,7 @@ private sealed class Mapper : MapperBase
                     var originalShape = _parent.TFInputShapes[i];
                     var shape = originalShape.ToIntArray();
 
-                    var colTypeDims = vecType.Dimensions.Select(dim => (long)dim).ToArray();
+                    var colTypeDims = vecType.Dimensions.Prepend(1).Select(dim => (long)dim).ToArray();
                     if (shape == null)
```
wschin: This is not right. What if the expected shape is really 3-D? In addition, tensorflow internally can do reshape. Why can't we put the weight on tensorflow's shoulder?

Ivanidzo4ka: Well, it doesn't. At least for image models we use.
I'm actually not sure why, considering what usually models have variable tensors like [?,224,224,3], but in one which samples has problems they have [1,224,224,3]

wschin: Then we need to fix those models or edit them before their exestuation. It's not idea to have an assumption which is true only for some image models.

zeahmed: @Ivanidzo4ka, I agree with @wschin. It will create problems for models that don't have batch dimension. There were a couple of issues related to that on github. I already talked to @shauheen and @CESARDELATORRE on this and proposed following solution.

`The ultimate solution is to have a reshape transform that user can use to reshape their data. However, the temporary solution would be to add a parameter in options class called “AddBatchDimensionOnInput”. When user set it to true, batch dimension would be added to the inputs otherwise not.`

Can you instead do it for now in this PR? We can then discuss adding reshape transform later.

zeahmed: This is the inception model from google.
https://storage.googleapis.com/download.tensorflow.org/models/inception5h.zip

Which takes input data of any shape. So we cannot infer shape from the input. There is a convolution layer just after the input which requires 4-D input and that's the problem.

Issue #2827, opened 2019-03-04
Scrubbing transformers (Meta issue)


Ivanidzo4ka:
We need to make sure our transforms are in good shape.

1. No protected fields/members/method in public classes. Only private protected.

2. Transformer class is sealed.

3. Transformer/Estimator should and catalog method names should be sensible (#2853 fine if slightly different).

4. Column options cleaning:
- Option should have meaning and proper way it initialize it self. 
- No short names.
5. `public IReadOnlyCollection<(string outputColumnName, string inputColumnName)> Columns => ColumnPairs.AsReadOnly(); ` 
I don't think that public thing make sense to me, I would prefer to get rid of it.

6. No public value tuples (especially in catalog methods) (#2881)

List still in progress.

shauheen:
Thanks @Ivanidzo4ka , Please make sure that the transforms that are not going 1.0 are not in the scope of this work for now. We can file other issues outside Project 13 for those.

PR #2875, opened 2019-03-07
Scrubbing image transforms

### LINKED ISSUE #2833: Scrubbing Image related transformers
Sub task of #2827
ConvertToGrayScale
ResizeImage
ExtractPixel
ConvertImage
LoadImage

najeeb-kazmi:
Fixes #2833


docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ResizeImages.cs
```diff
@@ -37,7 +37,7 @@ public static void Example()
             var imagesFolder = Path.GetDirectoryName(imagesDataFile);
             // Image loading pipeline. 
             var pipeline = mlContext.Transforms.LoadImages(imagesFolder, ("ImageReal", "ImagePath"))
-                        .Append(mlContext.Transforms.ResizeImages("ImageReal", imageWidth: 100, imageHeight: 100));
```
sfilipi: >imageWidth [](start = 79, length = 10)

I'd vote to keep them at ImageWidth and ImageWeight. Doesn't feel like output adds information.

src/Microsoft.ML.ImageAnalytics/ExtensionsCatalog.cs
```diff
@@ -11,15 +11,15 @@ public static class ImageEstimatorsCatalog
     {
         /// <include file='doc.xml' path='doc/members/member[@name="ImageGrayscalingEstimator"]/*' />
         /// <param name="catalog">The transform's catalog.</param>
-        /// <param name="columnPairs">The name of the columns containing the name of the resulting output column (first item of the tuple), and the paths of the images to work on (second item of the tuple).</param>
+        /// <param name="outputInputColumnPair">Names of the resulting output column (first item of the tuple), and the column containing the paths of the images to work on (second item of the tuple).</param>
```
sfilipi: >outputInputColumnPair [](start = 25, length = 21)

they are not column pairs anymore. This might be leftover from when we wer eusing tuples.

sfilipi: >tuple [](start = 104, length = 5)

no more tuple.

src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
```diff
@@ -82,7 +82,7 @@ internal sealed class Options : TransformInputBase
         /// <summary>
         /// The columns passed to this <see cref="ITransformer"/>.
         /// </summary>
-        public IReadOnlyCollection<(string outputColumnName, string inputColumnName)> Columns => ColumnPairs.AsReadOnly();
+        internal IReadOnlyCollection<(string outputColumnName, string inputColumnName)> Columns => ColumnPairs.AsReadOnly();
```
sfilipi: >internal [](start = 8, length = 8)

why make them internal? the user might want to inspect them?

najeeb-kazmi: This one is specifically called out in the issue #2827 (item 5 in the list). @Ivanidzo4ka can you elaborate?

src/Microsoft.ML.ImageAnalytics/ExtensionsCatalog.cs
najeeb-kazmi: @sfilipi I see output-input tuples being used in the samples for several image transforms, despite the api signature accepting `params ColumnOptions[]`. 

For example, in the [`ConvertToGrayScale.cs`](https://github.com/dotnet/machinelearning/blob/178827f4e7a7ccd129a8ee60c96e9630ffc6c0b3/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs#L39-L40) sample, we have tuples being passed to both `LoadImages` and `ConvertToGrayscale`.
```csharp
            var pipeline = mlContext.Transforms.LoadImages(imagesFolder, ("ImageObject", "ImagePath"))
                           .Append(mlContext.Transforms.ConvertToGrayscale(("Grayscale", "ImageObject")));
```

Questions for you:
1. Are the tuples from the samples being cast to array or are the samples incorrect and need to be revised?
2. What name do you suggest for this?

najeeb-kazmi: @sfilipi see above.

test/Microsoft.ML.Tests/ImagesTests.cs
```diff
@@ -262,7 +262,7 @@ public void TestBackAndForthConversionWithoutAlphaInterleave()
             var pixels = new ImagePixelExtractingTransformer(env, "ImagePixels", "ImageCropped", interleave: true, scale: 2f / 19, offset: 30).Transform(cropped);
```
Ivanidzo4ka: >ImagePixelExtractingTransformer [](start = 29, length = 31)

Right now ImagePixel and VectorToImage are pair of opposite to each over transforms. it's weird to have in one of them `interleave` parameter and `interleavedColors` in another.

I know both this constructors are internal, but it's still looks weird.

src/Microsoft.ML.ImageAnalytics/ExtensionsCatalog.cs
Ivanidzo4ka: Not Senja, but maybe I can be useful as well.
1. ColumnOptions have implicit cast from tuples, so user can enter tuples if he wants to, but in general we work with array of ColumnsOptions. So examples are fine, at least for now.
2. ` /// <param name="columns">Specifies the names of the input columns for the transformation, and their respective output column names.</param>` I think this is our default choice for such cases.


---
In reply to: [263581917](https://github.com/dotnet/machinelearning/pull/2875#discussion_r263581917) [](ancestors = 263581917)

test/Microsoft.ML.Tests/ImagesTests.cs
najeeb-kazmi: In the public constructors and public api, I intentionally named the parameter differently because the two transforms are doing the opposite of each other. In ImagePixel, the parameter specifies whether to **interleave** the colors or not when producing the vector, so I named it `interleavePixelColors`. In VectorToImage, the parameter specifies whether the colors in the vector are **interleaved** or not, so I named it `interleavedColors`. The intention was to make the parameter names descriptive, like all the other parameters that I renamed.

Do you agree with this distinction in names, i.e. `interleavePixelColors` for ImagePixel and `interleavedColors` for VectorToImage? If so, I can rename the parameters in the internal constructors to be consistent with the public ones. If not, do you suggest I use the same name for both ImagePixel and VectorToImage and what should that name be? In either case, I agree the names in public and internal constructors should be consistent.

najeeb-kazmi: Also, is it fine to make the rest of the parameters descriptive as I have done or do you prefer the previous short names e.g. `offset` (old) vs `offsetPixelColor` (new) ?

@Ivanidzo4ka

src/Microsoft.ML.ImageAnalytics/ExtensionsCatalog.cs
najeeb-kazmi: renamed as suggested

najeeb-kazmi: renamed as suggested above

src/Microsoft.ML.ImageAnalytics/VectorToImageTransform.cs
```diff
@@ -336,7 +336,7 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
 
                 var type = _types[iinfo];
                 var ex = _parent._columns[iinfo];
-                bool needScale = ex.Offset != 0 || ex.Scale != 1;
+                bool needScale = ex.OffsetPixelColor != 0 || ex.ScalePixelColor != 1;
```
sfilipi: >OffsetPixelColor [](start = 36, length = 16)

is this correct? the offset are for the image, not for the pixel color? (one can argue that the pixel colors are the image, but i think it becomes too complicated. )

src/Microsoft.ML.ImageAnalytics/VectorToImageTransform.cs
```diff
@@ -192,32 +192,32 @@ internal VectorToImageConvertingTransformer(IHostEnvironment env, params VectorT
...
         /// for all the pixels of the image. </param>
-        /// <param name="scale">Scale color pixel value by this amount.</param>
-        /// <param name="offset">Offset color pixel value by this amount.</param>
-        /// <param name="defaultAlpha">Default value for alpha color, would be overriden if <paramref name="colors"/> contains <see cref="ImagePixelExtractingEstimator.ColorBits.Alpha"/>.</param>
-        /// <param name="defaultRed">Default value for red color, would be overriden if <paramref name="colors"/> contains <see cref="ImagePixelExtractingEstimator.ColorBits.Red"/>.</param>
-        /// <param name="defaultGreen">Default value for grenn color, would be overriden if <paramref name="colors"/> contains <see cref="ImagePixelExtractingEstimator.ColorBits.Green"/>.</param>
-        /// <param name="defaultBlue">Default value for blue color, would be overriden if <paramref name="colors"/> contains <see cref="ImagePixelExtractingEstimator.ColorBits.Blue"/>.</param>
+        /// <param name="scalePixelColor">Scale color pixel value by this amount.</param>
+        /// <param name="offsetPixelCOlor">Offset color pixel value by this amount.</param>
```
sfilipi: left a note below on whether those are correct. Offseting and scaling pixel color makes little sense to me, they make more sense in the image context.

src/Microsoft.ML.ImageAnalytics/VectorToImageTransform.cs
najeeb-kazmi: Agree. Changing to `ScaleImage` and `OffsetImage`.

PR #2878, opened 2019-03-07
Make array values intended to be immutable IReadOnlyList

TomFinley:
Just a bit of relatively minor polish of something revealed during review of the public surface. In a handful of places we were returning arrays as values intended to be immutable. As elsewhere where perf isn't critical I've shifted to having the return value be `IReadOnlyList`. We've been doing this elsewhere but evidently missed a handful spots.

In an *ideal* world we'd make them actually immutable, as via `System.Collections.Immutable`, as I indeed did one or two places, but for the time being I think it suffices that we just make them read-only (even if the user could technically still muck with them with a cast, I'd consider fixing that problem later to be a non-breaking change in the API).

I only undertook this review in the Core/Data/Transform assemblies.

### DIFF
--- a/src/Microsoft.ML.Data/Evaluators/Metrics/RankingMetrics.cs
+++ b/src/Microsoft.ML.Data/Evaluators/Metrics/RankingMetrics.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Collections.Generic;
+using System.Collections.Immutable;
 using Microsoft.Data.DataView;
 
 namespace Microsoft.ML.Data
@@ -15,7 +17,7 @@ namespace Microsoft.ML.Data
         /// Array of normalized discounted cumulative gains where i-th element represent NDCG@i.
         /// <image src="https://github.com/dotnet/machinelearning/tree/master/docs/images/NDCG.png"></image>
         /// </summary>
-        public double[] NormalizedDiscountedCumulativeGains { get; }
+        public IReadOnlyList<double> NormalizedDiscountedCumulativeGains { get; }
 
         /// <summary>
         /// Array of discounted cumulative gains where i-th element represent DCG@i.
@@ -25,7 +27,7 @@ namespace Microsoft.ML.Data
         /// <image src="https://github.com/dotnet/machinelearning/tree/master/docs/images/DCG.png"></image>
         /// </summary>
         /// <remarks><a href="https://en.wikipedia.org/wiki/Discounted_cumulative_gain">Discounted Cumulative gain.</a></remarks>
-        public double[] DiscountedCumulativeGains { get; }
+        public IReadOnlyList<double> DiscountedCumulativeGains { get; }
 
         private static T Fetch<T>(IExceptionContext ectx, DataViewRow row, string name)
         {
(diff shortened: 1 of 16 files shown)


src/Microsoft.ML.Data/Evaluators/Metrics/RankingMetrics.cs
```diff
@@ -40,16 +42,14 @@ internal RankingMetrics(IExceptionContext ectx, DataViewRow overallResult)
...
+            NormalizedDiscountedCumulativeGains = Fetch(RankingEvaluator.Ndcg).DenseValues().ToImmutableArray();
         }
 
         internal RankingMetrics(double[] dcg, double[] ndcg)
         {
-            DiscountedCumulativeGains = new double[dcg.Length];
-            dcg.CopyTo(DiscountedCumulativeGains, 0);
-            NormalizedDiscountedCumulativeGains = new double[ndcg.Length];
-            ndcg.CopyTo(NormalizedDiscountedCumulativeGains, 0);
+            DiscountedCumulativeGains = dcg.ToImmutableArray();
```
eerhardt: Do we _need_ to do a copy here? Would it be terrible if we just set the field to an array?

src/Microsoft.ML.Transforms/Text/NgramHashingTransformer.cs
```diff
@@ -878,7 +878,8 @@ public sealed class ColumnOptions
             /// <summary>Name of the column resulting from the transformation of <see cref="InputColumnNames"/>.</summary>
             public readonly string Name;
             /// <summary>Names of the columns to transform.</summary>
-            public readonly string[] InputColumnNames;
+            public IReadOnlyList<string> InputColumnNames => InputColumnNamesArray;
```
eerhardt: How can someone set these values?

src/Microsoft.ML.Transforms/Text/WordTokenizing.cs
```diff
@@ -444,7 +444,8 @@ public sealed class ColumnOptions
         {
             public readonly string Name;
             public readonly string InputColumnName;
-            public readonly char[] Separators;
+            public IReadOnlyList<char> Separators => SeparatorsArray;
```
eerhardt: Same question as above - how can users specify this?

src/Microsoft.ML.Transforms/Text/NgramHashingTransformer.cs
TomFinley: You set them in the constructor. This like the other parameters was not intended to be mutable.

src/Microsoft.ML.Transforms/Text/WordTokenizing.cs
TomFinley: Same answer.

src/Microsoft.ML.Data/Evaluators/Metrics/RankingMetrics.cs
TomFinley: I'm honestly not certain, but as I indicated in the description I think it's better to be more concerned at this moment with getting the right public surface. If it turns out that this is an area where we can do an optimization to avoid a copy later, I'm all for it, but I doubt given the nature of this code that a copy will really wind up making a big difference here (as opposed to, say, something in the actual data pipeline).

src/Microsoft.ML.Transforms/Text/NgramHashingTransformer.cs
eerhardt: Obviously this concern existed before your change, but

I think our "Options" classes are inconsistent here.

For example in TextLoader, I can set the properties after the constructor:

https://github.com/dotnet/machinelearning/blob/002d6b45e1fd9f699129a322180e135a20f65483/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L101-L129

I would have thought that these Options classes are mutable - just like how TextLoader's Column class is.

TomFinley: I agree that there is an inconsistency -- I think this is an effect of something that I observed that some things that had been renamed from `ColumnInfo` to `ColumnOptions` shouldn't have actually been renamed in #2709. But until that quesiton of renaming of the *class* is addressed, I can only make sure the object itself is behaving consistently w.r.t. itself, and ***this*** is very, very clearly intended to be a non-mutable structure.

src/Microsoft.ML.Transforms/Text/NgramHashingTransformer.cs
```diff
@@ -951,7 +952,7 @@ public sealed class ColumnOptions
                 }
                 FriendlyNames = null;
                 Name = name;
-                InputColumnNames = inputColumnNames;
+                InputColumnNamesArray = inputColumnNames;
```
wschin: Not sure if we really need to rename this. If yes, should we do the same to all transforms/trainers public APIs?

TomFinley: No. I wanted the *public* name to remain what it was, just over a different type. But I wasn't quite willing to change the *internal* implementation to work over that structure, since I've run into subtle issues with that approach before, and I was just trying to do a quick PR. (You'll note that this renaming is purely an internal change.) But I had to name what it had been *something* different, right?

But since I wanted the public surface to remain something named `InputColumnNames` (as it had been prior to my change), I needed the internal backing field to be named something else. `InputColumnNamesArray` seemed like an innocent enough name.

I changed no name anywhere in our public surface, just the apparent type from an external perspective. (As well as, inevitably, making this a property instead of a field.)

TomFinley:
> This change looks good. We should think about the Options classes being mutable/immutable separately.

I'll write an issue about that. I think there are a couple things to think about.

TomFinley:
FYI the regex I used to find array return values and properties and fields was:

    ^\s+(?: [\w<,>]+)+\[\]

PR #2883, opened 2019-03-07
Specify MaxNumberOfIterations for SDCA, K-Means

### LINKED ISSUE #2871: NumberOfIterations vs. MaxIterations
The recent naming changes to K-Means and SDCA (and perhaps others?) resulted in `MaxIterations` being renamed to `NumberOfIterations`. Since this parameter specifies the worst-case bound and not the actual number of iterations taken in most cases, I think we should keep it named `MaxIterations`. 

Whatever we call it, `NumberOfIterations` is not what this parameter specifies &mdash; this is a stopping criterion, not a guaranteed execution parameter.

rogancarr:
This PR updates SDCA and K-Means to specify `MaximumNumberOfIterations` rather than `NumberOfIterations` to make the name more precise. Essentially, for these learners, the maximum number of iterations is a worst-case scenario and a last-resort stopping criteria.

Fixes #2871


rogancarr:
Also saw a place where we forgot to put `NumberOfClusters` for K-Means, so fixing this there too.

Issue #2832, opened 2019-03-04
Scrubbing Text related transformers


Ivanidzo4ka:
Sub task of #2827

- [x] WordEmbedding #2891 
- [x] FeaturizeText #2944 
- [x] LDA #2890
- [x] NormalizeText #2918
- [x] ProduceHashedNGrams #2898
- [x] ProduceHashedWordBag #2898
- [x] ProduceNgrams #2898
- [x] ProduceWordBag #2898
- [x] RemoveStopWords #2916
- [x] RemoveCustomStopWords #2916
- [x] TokenizeChar #2916
- [x] TokenizeWords #2916

shauheen:
@Ivanidzo4ka , @wschin can this issue be closed now?

wschin:
Yes. Let me close it.

PR #2890, opened 2019-03-08
Scrub Latent Dirichlet Allocation Transform (Just Renaming)

wschin:
Toward #2832. This PR handles Latent Dirichlet Allocation.

### DIFF
--- a/src/Microsoft.ML.Transforms/Text/TextCatalog.cs
+++ b/src/Microsoft.ML.Transforms/Text/TextCatalog.cs
@@ -579,16 +579,16 @@ namespace Microsoft.ML
         /// <param name="catalog">The transform's catalog.</param>
         /// <param name="outputColumnName">Name of the column resulting from the transformation of <paramref name="inputColumnName"/>.</param>
         /// <param name="inputColumnName">Name of the column to transform. If set to <see langword="null"/>, the value of the <paramref name="outputColumnName"/> will be used as source.</param>
-        /// <param name="numTopic">The number of topics.</param>
+        /// <param name="numberOfTopics">The number of topics.</param>
         /// <param name="alphaSum">Dirichlet prior on document-topic vectors.</param>
         /// <param name="beta">Dirichlet prior on vocab-topic vectors.</param>
-        /// <param name="mhstep">Number of Metropolis Hasting step.</param>
-        /// <param name="numIterations">Number of iterations.</param>
+        /// <param name="samplingStepCount">Number of Metropolis Hasting step.</param>
+        /// <param name="maximumNumberOfIterations">Number of iterations.</param>
         /// <param name="likelihoodInterval">Compute log likelihood over local dataset on this iteration interval.</param>
-        /// <param name="numThreads">The number of training threads. Default value depends on number of logical processors.</param>
-        /// <param name="numMaxDocToken">The threshold of maximum count of tokens per doc.</param>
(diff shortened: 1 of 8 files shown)


test/Microsoft.ML.StaticPipelineTesting/StaticPipeTests.cs
```diff
@@ -675,7 +675,7 @@ public void LdaTopicModel()
             var est = data.MakeNewEstimator()
                 .Append(r => (
                     r.label,
-                    topics: r.text.ToBagofWords().ToLdaTopicVector(numTopic: 3, numSummaryTermPerTopic:5, alphaSum: 10, onFit: m => ldaSummary = m.LdaTopicSummary)));
+                    topics: r.text.ToBagofWords().ToLatentDirichletAllocationTopicVector(numberOfTopics: 3, numberOfSummaryTermsPerTopic:5, alphaSum: 10, onFit: m => ldaSummary = m.LdaTopicSummary)));
```
Ivanidzo4ka: >ToLatentDirichletAllocationTopicVector [](start = 50, length = 38)

It's not what I care much about Static pipeline right now, but I would call it `LatentDirichletAllocation` same as we use in Dynamic pipeline.

wschin: Why not. :)

---
In reply to: [263972927](https://github.com/dotnet/machinelearning/pull/2890#discussion_r263972927) [](ancestors = 263972927)

src/Microsoft.ML.Transforms/Text/TextCatalog.cs
```diff
@@ -600,20 +600,20 @@ public static class TextCatalog
         public static LatentDirichletAllocationEstimator LatentDirichletAllocation(this TransformsCatalog.TextTransforms catalog,
```
Ivanidzo4ka: >LatentDirichletAllocation [](start = 57, length = 25)

Whole text catalog is slightly different from any other catalogs.
In any other catalog, we expose constructor with few major options, and then just columnOptions array.
In this specific case we expose EVERYTHING.
Shall we cut it to bare minimum?

wschin: Ok.

---
In reply to: [263974378](https://github.com/dotnet/machinelearning/pull/2890#discussion_r263974378) [](ancestors = 263974378)

src/Microsoft.ML.Transforms/Text/TextCatalog.cs
```diff
@@ -600,20 +593,22 @@ public static class TextCatalog
...
```
sfilipi: >LatentDirichletAllocation [](start = 57, length = 25)

not directly related to your change, but the other names are usually verbs. If we wanted to keep to that, shall this be named ApplyLatentDirichletAllocation

wschin: Not sure. `ConvertLatentDirichletAllocation` looks a bit confusing. For example, we don't say `ConvertToMatrixFactorization`. One proper (mathematically) choice is `ProduceConvertLatentDirichletAllocationVector` but it's just too long..

---
In reply to: [264332826](https://github.com/dotnet/machinelearning/pull/2890#discussion_r264332826) [](ancestors = 264332826)

PR #2891, opened 2019-03-08
Scrub word embedding transform

wschin:
Toward #2832. This PR only contains renaming and internalizes a function,
```
-        public IReadOnlyCollection<(string outputColumnName, string inputColumnName)> Columns => ColumnPairs.AsReadOnly();
+        private IReadOnlyCollection<(string outputColumnName, string inputColumnName)> Columns => ColumnPairs.AsReadOnly();
```


src/Microsoft.ML.Transforms/Text/TextCatalog.cs
```diff
@@ -112,7 +112,7 @@ public static class TextCatalog
         public static WordEmbeddingsExtractingEstimator ExtractWordEmbeddings(this TransformsCatalog.TextTransforms catalog,
```
Ivanidzo4ka: >ExtractWordEmbeddings [](start = 56, length = 21)

Maybe `ApplyWordEmbeddings` ?

Issue #2884, opened 2019-03-07
Discussion: ColumnOptions actually a good name?


TomFinley:
In #2878, @eerhardt had a comment that we should consider, the gist of which was, since all of our `Options` classes have mutable properties, is it appropriate for `ColumnOptions` to be called this, since they are not and often cannot be mutable? We also have issue #2854 where @rogancarr thought he couldn't get normalization information out of the structure since it was named options, so this is not actually as academic an issue than I might have thought, say, a few days ago.

The approach taken in #2709 was that these structures created for configuration of the per-column options should be called options, and that it was (apparently) assumed to be irrelevant whether those items were mutable or not. Now, I'm not saying we should revert that PR necessarily, but it is something to consider, since it seems to be confusing people.

Now then, the structures themselves obviously must not be mutable, since they are often the same structures used in the associated estimators and transformers to project schema, e.g., here it is for the n-gram hashing estimator:

https://github.com/dotnet/machinelearning/blob/a5580108d6171ae8dfa5ba60210dc2d88c9dec75/src/Microsoft.ML.Transforms/Text/NgramHashingTransformer.cs#L1077

Here it is in the transformer:

https://github.com/dotnet/machinelearning/blob/a5580108d6171ae8dfa5ba60210dc2d88c9dec75/src/Microsoft.ML.Transforms/Text/NgramHashingTransformer.cs#L1077

So, just something to think about, whether it was in fact a good idea for this thing to be called "options" really, in all the cases we named it options. Maybe we could have a refinement on the policy of naming this thing? Or maybe we decide to just live with it, because the confusion of calling all these things "options" vs. "info" vs. "whatever" is greater than this inconsistency in roles?

I'm fine with leaving it as is, but I do see some confusion so I think we should think about it, and at least formulate a psoition.

/cc @eerhardt and @rogancarr and @sfilipi and @artidoro ...

TomFinley:
Another thing is in the case of `NgramHashing` this `ColumnOptions` never actually appears anywhere as an input or an output in the public API, but that's probably just an oversight (more usual would be, expose it as a `IReadOnlyCollection<ColumnOptions>` on the transformer, as we do elsewhere, we just forgot to do that, but we can do that post v1 since that's strictly additive.) More a question on the naming.

artidoro:
[...]

Issue #2064, opened 2019-01-07
Estimator arguments should take output column name as first parameter, any inputs as subsequent parameters


TomFinley:
So, a somewhat embarrassing blunder to report... not quite sure how to say this.

Throughout our codebase, for years we've ***always*** specified the name of the output column of a transform first, *then* the source(s).

That's good. But when estimators were introduced, somehow, nearly all of them were introduced in the reverse order: nearly all of them specify the inputs, *then* the outputs. This was probably an unconscious mistake, but it's one with fairly wide consequences, since that mistaken pattern was copied again and again as people did the estimator conversion work, to the point where most (not all!) estimators are now inconsistent with the whole rest of the codebase!

This should be corrected: it is at least inconsistent, and even if not inconsistent actually rather obnoxious practically because specifying the name of the output first has practical benefits, and makes a lot more sense, since if you're specifying a transformation the *most* important information someone will want to know is what you're calling your output!

# The Story Until a Few Months Ago

So, throughout our codebase, for years, it has been our practice that when specifying a transform, you specify the name of the output, *then* you specify the name of the input(s) (if any). The reason for this is practical: the outputs are usually the result of a well defined single calculation (the application of the transform), whereas what is taken as an "input" to a transform can have various meanings since it is sometimes a multivariate function in its *inputs* more often than in its *outputs*. (E.g., concatenate can have multiple inputs, text can have multiple inputs, MI feature selection takes multiple heterogeneous inputs, but all produce single output columns.)

This was our practice certainly when this code was a tool, as we see in the command line help string, specifying name *then* source.

https://github.com/dotnet/machinelearning/blob/bb46fdf405828134de9feebce74f21b4aacb15ed/src/Microsoft.ML.Data/Transforms/ColumnCopying.cs#L109-L110

This trend continued during our initial attempts at an API, as seen in this [early discussion in PR 405](https://github.com/dotnet/machinelearning/pull/405#discussion_r197980521), and as seen here:

https://github.com/zeahmed/machinelearning/blob/16f7883933b56f8fd86077bf0fd262b24374e9d0/src/Microsoft.ML.Data/Transforms/ConvertTransform.cs#L116

and here

https://github.com/zeahmed/machinelearning/blob/16f7883933b56f8fd86077bf0fd262b24374e9d0/src/Microsoft.ML.Data/Transforms/DropSlotsTransform.cs#L226

and numerous other places.

This is done for the practical reason that, when a transform produces an
[...]

PR #2893, opened 2019-03-08
WIP: Refactoring of ColumnOptions for ImagePixelExtractor

artidoro:
Related to #2884.

The change does the following things:
1. Internalizes the immutable class `ColumnInfos` and moves it to the transformer
2. Exposes the previously named `Column` class and renames it `ColumnOptions`
3. Internalizes entrypoint only fields in `ColumnOptions`, renames those that did not match the public API
4. Refactored the constructor so that it takes the new `ColumnOptions` class

The overall behavior of the Estimator/Transformer API is unchanged (besides the use of the new mutable class).
However, this change alters the entrypoints API behavior in the following way.
Past behavior:
- Can set column specific options.
- Can set overall options for all columns
- Can set overall options for some column and column specific options for others. 

After this change: 
- Can set column specific options.
- Can set overall options for all columns

I would also like to change the name of the input and output columns (in `SourceNameColumnBase`) to something more appropriate than `Name` and `Source`, but since that will touch a lot of files, I would like some feedback on the approach first.

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
+++ b/src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
@@ -15,7 +15,7 @@ using Microsoft.ML.Transforms;
 
 namespace Microsoft.ML.Data
 {
-    internal abstract class SourceNameColumnBase
+    public abstract class SourceNameColumnBase
     {
         [Argument(ArgumentType.AtMostOnce, HelpText = "Name of the new column", ShortName = "name")]
         public string Name;
@@ -104,8 +104,7 @@ namespace Microsoft.ML.Data
         }
     }
 
-    [BestFriend]
-    internal abstract class OneToOneColumn : SourceNameColumnBase
+    public abstract class OneToOneColumn : SourceNameColumnBase
     {
         [BestFriend]
         private protected OneToOneColumn() { }
--- a/src/Microsoft.ML.ImageAnalytics/ImagePixelExtractor.cs
+++ b/src/Microsoft.ML.ImageAnalytics/ImagePixelExtractor.cs
@@ -44,8 +44,11 @@ namespace Microsoft.ML.ImageAnalytics
     public sealed class ImagePixelExtractingTransformer : OneToOneTransformerBase
     {
         [BestFriend]
-        internal class Column : OneToOneColumn
+        internal class Options : TransformInputBase
         {
+            [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "New column definition(s) (optional form: name:src)", Name = "Column", ShortName = "col", SortOrder = 1)]
+            public ImagePixelExtractingEstimator.ColumnOptions[] Columns;
+
             [Argument(ArgumentType.AtMostOnce, HelpText = "Whether to use alpha channel", ShortName = "alpha")]
             public bool? UseAlpha;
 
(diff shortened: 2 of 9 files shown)


TomFinley:
>         public bool TrySanitize()

I feel we should make this internal.

---
Refers to: src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs:97 in be097f0. [](commit_id = be097f0e8e8d43d89e380256dcef1a23cf29076f, deletion_comment = False)

TomFinley:
>         protected virtual bool TryUnparseCore(StringBuilder sb)

If this is now to be in the public surface then we'll have to `private protected` these guys.

---
Refers to: src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs:197 in be097f0. [](commit_id = be097f0e8e8d43d89e380256dcef1a23cf29076f, deletion_comment = False)

TomFinley:
>         protected virtual bool TryUnparseCore(StringBuilder sb, string extra)

This too.

---
Refers to: src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs:222 in be097f0. [](commit_id = be097f0e8e8d43d89e380256dcef1a23cf29076f, deletion_comment = False)

TomFinley:
>         protected ColumnBindingsBase(DataViewSchema input, bool user, params string[] names)

This too.

---
Refers to: src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs:304 in be097f0. [](commit_id = be097f0e8e8d43d89e380256dcef1a23cf29076f, deletion_comment = False)

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
```diff
@@ -15,7 +15,7 @@
 
 namespace Microsoft.ML.Data
 {
-    internal abstract class SourceNameColumnBase
+    public abstract class SourceNameColumnBase
     {
         [Argument(ArgumentType.AtMostOnce, HelpText = "Name of the new column", ShortName = "name")]
         public string Name;
```
TomFinley: >Name [](start = 22, length = 4)

The names will have to change in accordance with the conclusion of #2064, but let's have agreement on the general approach first. (If we don't also want the entry-points to change yet, remember the `Name` property on the attribute.)

src/Microsoft.ML.ImageAnalytics/ImagePixelExtractor.cs
```diff
@@ -44,8 +44,11 @@ namespace Microsoft.ML.ImageAnalytics
...
     {
         [BestFriend]
-        internal class Column : OneToOneColumn
+        internal class Options : TransformInputBase
         {
+            [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "New column definition(s) (optional form: name:src)", Name = "Column", ShortName = "col", SortOrder = 1)]
+            public ImagePixelExtractingEstimator.ColumnOptions[] Columns;
+
             [Argument(ArgumentType.AtMostOnce, HelpText = "Whether to use alpha channel", ShortName = "alpha")]
             public bool? UseAlpha;
```
TomFinley: >UseAlpha [](start = 25, length = 8)

These are a little funky. In entry-point land, and command line land, there would be the arguments to the transform, and then the arguments to each column that could potentially override it.

TomFinley:
So I'm not sure about this. It seems like the existing column structure (being nullable to allow overrides of more global options which don't exist in our interface) don't quite work. Also this particular example brings up the case that not all transformers practically benefit from this "multi-column mapping" idiom that we felt like we had to obey in the command line (since a universal interface there was more essential). Also the size of the change makes me worried this is way too much to take on prior to v1, even if we were to take this on prior to v1.

What you think about just hiding the things for now, including the `ColumnOptions` extension methods and the classes, and just calling it a day, at least for now?

artidoro:
I agree that changing the behavior of the command line and the entrypoint API is such a fashion is too significant to be a valid option before V1. I will close this PR.

PR #2894, opened 2019-03-08
Renaming IterationsToRemember to HistorySize for L-BFGS learners.

### LINKED ISSUE #2882: Non-standard naming in L-BFGS Learners (LogisticRegression, PoissonRegression)
In `LogisticRegression` and `PoissonRegression` (which use the same L-BFGS base), we have a parameter `IterationsToRemember` that refers to the number of gradients to accumulate in the history. While this terminology makes sense, it's not what we usually encounter in the field.

In the literature, we see this referred to as the "history size" (see e.g. [wikipedia](https://en.wikipedia.org/wiki/Limited-memory_BFGS)).

In the various toolkits that expose an L-BFGS solver, they use:
Scikit Learn: Doesn't expose it.
Spark: [NumberOfCorrections](https://spark.apache.org/docs/2.3.0/api/java/org/apache/spark/mllib/optimization/LBFGS.html)
TensorFlow: [num_correction_pairs](https://www.tensorflow.org/probability/api_docs/python/tfp/optimizer/lbfgs_minimize)
PyTorch: [history_size](https://pytorch.org/docs/stable/_modules/torch/optim/lbfgs.html)

I would vote for `HistorySize`, with the docs explaining what it is.

What do you all think? Any big feelings around this?

rogancarr:
This PR renames the L-BFGS parameter `IterationsToRemember` to `HistorySize` to be more in line with the common nomenclature. (Previously it was `MemorySize`.)

Fixes #2882

### DIFF
--- a/src/Microsoft.ML.StandardLearners/Standard/LogisticRegression/LbfgsPredictorBase.cs
+++ b/src/Microsoft.ML.StandardLearners/Standard/LogisticRegression/LbfgsPredictorBase.cs
@@ -54,7 +54,7 @@ namespace Microsoft.ML.Trainers
                 ShortName = "m, MemorySize", SortOrder = 50)]
             [TGUI(Description = "Memory size for L-BFGS", SuggestedSweeps = "5,20,50")]
             [TlcModule.SweepableDiscreteParamAttribute("MemorySize", new object[] { 5, 20, 50 })]
-            public int IterationsToRemember = Defaults.IterationsToRemember;
+            public int HistorySize = Defaults.HistorySize;
 
             /// <summary>
             /// Number of iterations.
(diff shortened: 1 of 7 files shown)


Ivanidzo4ka:
It's always EntryPointCatalog

src/Microsoft.ML.StandardLearners/Standard/LogisticRegression/LbfgsPredictorBase.cs
```diff
@@ -54,7 +54,7 @@ public abstract class OptionsBase : TrainerInputBaseWithWeight
                 ShortName = "m, MemorySize", SortOrder = 50)]
             [TGUI(Description = "Memory size for L-BFGS", SuggestedSweeps = "5,20,50")]
             [TlcModule.SweepableDiscreteParamAttribute("MemorySize", new object[] { 5, 20, 50 })]
-            public int IterationsToRemember = Defaults.IterationsToRemember;
+            public int HistorySize = Defaults.HistorySize;
```
wschin: nit:
```suggestion
            public int HistoryLength = Defaults.HistorySize;
```
`Size` sounds like the number of parameters we can store.

rogancarr:
Oops, just saw your comment. HistorySize is the common nomenclature, and also used by PyTorch.

Issue #2438, opened 2019-02-06
We need documentation describing how to use the explainability features in ML.NET


rogancarr:
We have some great explainability features in ML.NET, but they are hidden here and there. It would be great to have an explainability doc, similar to the [ML.NET Cookbook](https://github.com/dotnet/machinelearning/blob/master/docs/code/MlNetCookBook.md), that showed how to use these features and what they do.

JRAlexander:
This should be in docs. Lets work together on it.

shmoradims:
Already resolved [here](https://docs.microsoft.com/en-us/dotnet/machine-learning/how-to-guides/explain-machine-learning-model-permutation-feature-importance-ml-net)

shmoradims:
We also need to cover FCC and GAM as explained by Rogan here:

"Permutation Feature Importance (PFI), which gives a global picture of feature importance for a model, Feature Contribution Calculator (FCC), which gives a local picture of feature importance for a sample, and GAM, which are a class of interpretable models that give both a global and local explanation. Along with functioning code, we will write documentation to explain how the methods work and point to relevant literature, and create samples showing how to use the features and how to interpret the results."

PR #2901, opened 2019-03-10
Explainability doc

jwood803:
Initial draft to add explainability documentation.

Fix for #2438.


docs/code/ExplainabilityCookBook.md
```diff
@@ -0,0 +1,90 @@
+# Explainability CookBook
+
+This document serves to provide different ways to use the explainability features of ML.NET.
+
+All of these samples will use the [housing data](https://github.com/dotnet/machinelearning/blob/master/test/data/housing.txt) and will reference the below data schema class and pipeline.
+
+```csharp
```
artidoro: We should try to make sure that the cookbooks code compiles and runs properly even with the ongoing changes. 
The way we do this is by adding all the code and utilities in a test in https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamplesDynamicApi.cs

We then copy the most important parts of the CookbookSamplesDynamicApi.cs to the .md file.

If you could do that it would be great!

docs/code/ExplainabilityCookBook.md
```diff
@@ -0,0 +1,90 @@
...
+    public float PercentPre40s { get; set; }
+    public float EmploymentDistance { get; set; }
+    public float HighwayDistance { get; set; }
+    public float TaxRate { get; set; }
+    public float TeacherRatio { get; set; }
+}
+
+MLContext context = new MLContext();
+
+IDataView data = context.Data.LoadFromTextFile("./housing.txt", new[]
```
artidoro: So this code for loading and specifying the data class would be necessary in the .cs file to run your tests. However, I think we should not add it to the .md file. We have other sections of the cookbook and samples in which we illustrate that.

docs/code/ExplainabilityCookBook.md
```diff
@@ -0,0 +1,90 @@
...
+hasHeader: true);
+
+var pipeline = context.Transforms.Concatenate("Features", "CrimesPerCapita", "PercentResidental", "PercentNonRetail", "CharlesRiver", "NitricOxides",
+    "RoomsPerDwelling", "PercentPre40s", "EmploymentDistance", "HighwayDistance", "TaxRate", "TeacherRatio")
+    .Append(context.Regression.Trainers.FastTree());
+
+var model = pipeline.Fit(data);
+```
+
+## How do I look at the global feature importance?
```
artidoro: I would suggest not to write these under a separate cookbook, but rather to add them to the main MlNetCookBook.md under a separate model explainability section after the model inspection section: "I want to look at my model's coefficients".

docs/code/MlNetCookBook.md
```diff
@@ -578,6 +578,48 @@ var biases = modelParameters.GetBiases();
 
 ```
 
+## How do I look at the global feature importance?
+The below snippet shows how to get a glimpse of the the feature importance, or how much each column of data impacts the performance of the model.
```
rogancarr: >column of data [](start = 93, length = 14)

"feature" rather than "column of data". The end features in the model might not be exactly the input columns.

docs/code/MlNetCookBook.md
```diff
@@ -578,6 +578,48 @@ var biases = modelParameters.GetBiases();
...
+
+```csharp
+var transformedData = model.Transform(data);
+
+var featureImportance = context.Regression.PermutationFeatureImportance(model.LastTransformer, transformedData);
+
+foreach (var metricsStatistics in featureImportance)
+{
+    Console.WriteLine($"Root Mean Squared - {metricsStatistics.Rms.Mean}");
```
rogancarr: >Console.WriteLine($"Root Mean Squared - {metricsStatistics.Rms.Mean}"); [](start = 4, length = 71)

Explain a bit above about what this is calculating. It's not giving the RMS, but the difference in RMS for each feature if the feature were to be replaced with a random value.

Also, I would print "Feature I: Difference in RMS" rather than just the RMS.

docs/code/MlNetCookBook.md
```diff
@@ -578,6 +578,48 @@ var biases = modelParameters.GetBiases();
...
+}
+```
+
+## How do I get a model's weights to look at the global feature importance?
```
rogancarr: Move this above PFI, as it's the most naïve way we have to ask this question.

docs/code/MlNetCookBook.md
```diff
@@ -578,6 +578,48 @@ var biases = modelParameters.GetBiases();
...
+The below snippet shows how to get a model's weights to help determine the feature importance of the model.
```
rogancarr: >The below [](start = 0, length = 9)

Note that for a linear model, the weights are only an approximation. It helps to standardize the variables before the fit, so that they are all on the same scale, and even then, the linear regression solution does not account for correlations between the variables, and therefore this isn't a great measure of explainability.

docs/code/MlNetCookBook.md
```diff
@@ -578,6 +578,48 @@ var biases = modelParameters.GetBiases();
...
+
+```csharp
+var linearModel = model.LastTransformer.Model;
+
+var weights = new VBuffer<float>();
+linearModel.GetFeatureWeights(ref weights);
```
rogancarr: Add for trees as well -- see the functional tests.

docs/code/MlNetCookBook.md
```diff
@@ -578,6 +578,48 @@ var biases = modelParameters.GetBiases();
...
+```
+
+## How do I look at the feature importance per row?
```
rogancarr: "Local feature importance"

"row" => "example" (the language we've shifted to using)

docs/code/MlNetCookBook.md
```diff
@@ -578,6 +578,48 @@ var biases = modelParameters.GetBiases();
...
+
+var linearModel = model.LastTransformer;
+
+var featureContributionCalculation = context.Transforms.CalculateFeatureContribution(linearModel, normalize: false);
+
+var featureContributionData = featureContributionCalculation.Fit(transfomedData).Transform(transfomedData);
+
+var shuffledSubset = context.Data.TakeRows(context.Data.ShuffleRows(featureContributionData), 10);
+
+var preview = shuffledSubset.Preview();
```
rogancarr: manually print the rows of data after casting to an enumerable. You can copy / paste from the FCC Sample.

rogancarr:
This looks great! Just a few comments.

docs/code/MlNetCookBook.md
```diff
@@ -578,6 +578,60 @@ var biases = modelParameters.GetBiases();
...
+
+```csharp
+var treeModel = model.LastTransformer.Model;
+
+var weights = new VBuffer<float>();
+treeModel.GetFeatureWeights(ref weights);
+```
+
+## How do I look at the global feature importance?
+The below snippet shows how to get a glimpse of the the feature importance, or how much each feature impacts the performance of the model. It also outputs the difference in root mean squared for each feature as though the feature were replaced with a random value.
```
rogancarr: >It also outputs the difference in root mean squared for each feature as though the feature were replaced with a random value. [](start = 139, length = 125)

"Permutation Feature Importance works by computing the change in the evaluation metrics when each feature is replaced by a random value. In this case, we are investigating the change in the root mean squared error".

docs/code/MlNetCookBook.md
```diff
@@ -578,6 +578,60 @@ var biases = modelParameters.GetBiases();
...
+
+```csharp
+var transformedData = model.Transform(data);
+
+var featureImportance = context.Regression.PermutationFeatureImportance(model.LastTransformer, transformedData);
+
+foreach (var metricsStatistics in featureImportance)
+{
+    Console.WriteLine($"Feature I: Difference in RMS - {metricsStatistics.Rms.Mean}");
```
rogancarr: >I [](start = 32, length = 1)

Sorry, I meant the number of the feature, like 0, 1, 2, ... "Feature 0: " "Feature 1: " etc.

docs/code/MlNetCookBook.md
```diff
@@ -578,6 +578,60 @@ var biases = modelParameters.GetBiases();
...
+}
+```
+
+## How do I look at the local feature importance per example?
+The below snippet shows how to get feature importance for each example of data.
```
rogancarr: >feature importance for each example of data [](start = 35, length = 43)

Can you link to the appropriate place in docs for more information for all of these? Maybe we don't actually need to go into major details here.

jwood803: The best doc I could find is [this one](https://docs.microsoft.com/en-us/dotnet/machine-learning/how-to-guides/determine-global-feature-importance-in-model). Is this ok to link to in each of these sections or would there be a doc for each of these?

rogancarr: Oh, I was thinking we could link to the code samples in the repo. But this is a moving target, so let's revisit later.

jwood803: Sounds good! Apologies for the misunderstanding. Was there anything else I missed for the PR? Just making sure no one is waiting for me to make more updates 😄

PR #2902, opened 2019-03-10
Add sample to get data from a SQL database

jwood803:
Sample to help with #2498.

This sample using the Entity Framework in-memory data provider to load and read data that can be used to create an ML.NET model.

Hopefully, this is close to what was in mind for the sample. 😄

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/EntityFrameworkEnumerable.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/EntityFrameworkEnumerable.cs
@@ -1,0 +1,128 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.EntityFrameworkCore;
+
+namespace Microsoft.ML.Samples.Dynamic
+{
+    public class HousingData
+    {
+        // Entity Framework requires the object to have a primary key
+        public int HousingDataId { get; set; }
+        public float MedianHomeValue { get; set; }
+        public float CrimesPerCapita { get; set; }
+        public float PercentResidental { get; set; }
+        public float PercentNonRetail { get; set; }
+        public float CharlesRiver { get; set; }
+        public float NitricOxides { get; set; }
+        public float RoomsPerDwelling { get; set; }
+        public float PercentPre40s { get; set; }
+        public float EmploymentDistance { get; set; }
+        public float HighwayDistance { get; set; }
+        public float TaxRate { get; set; }
+        public float TeacherRatio { get; set; }
+    }
+
+    public partial class MLNetExampleContext : DbContext
+    {
+        public DbSet<HousingData> Housing { get; set; }
+
+        public MLNetExampleContext(DbContextOptions<MLNetExampleContext> options)
+            : base(options)
+        {
+        }
+
+        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+        {
+            if (!optionsBuilder.IsConfigured)
+            {
+                optionsBuilder.UseSqlServer("Server=(localdb)\\mssqllocaldb;Database=EFProviders.InMemory;Trusted_Connection=True;ConnectRetryCount=0");
+            }
+        }
+    }
+
+    public static class EntityFrameworkEnumerable
+    {
+        private static IEnumerable<HousingData> housingData;
+
+        public static void Example()
+        {
+            // Download the housing data set and get the file name.
+            var housingFile = SamplesUtils.DatasetUtils.DownloadHousingRegressionDataset();
+
+            // Read data from file into an IEnumerable.
+            var data = File.ReadAllLines(housingFile)
+                .Skip(1) // Skip the header row
+                .Select(l => l.Split('\t'))
+                .Select(i => new HousingData
+                {
+                    MedianHomeValue = float.Parse(i[0]),
+                    CrimesPerCapita = float.Parse(i[1]),
+                    PercentResidental = float.Parse(i[2]),
+                    PercentNonRetail = float.Parse(i[3]),
+                    CharlesRiver = float.Parse(i[4]),
+                    NitricOxides = float.Parse(i[5]),
+                    RoomsPerDwelling = float.Parse(i[6]),
+                    PercentPre40s = float.Parse(i[7])
+                });
+
+            // Options to pass to the DbContext to use an in-memory database for
+            // Entity Framework and to give the database a name.
+            var options = new DbContextOptionsBuilder<MLNetExampleContext>()
+                .UseInMemoryDatabase(databaseName: "TestData")
+                .Options;
+
+            // Add the data to the in-memory database via Entity Framework.
+            using (var dbContext = new MLNetExampleContext(options))
+            {
+                foreach (var dataItem in data)
+                {
+                    dbContext.Add(dataItem);
+                }
+
+                var rowsAffected = dbContext.SaveChanges();
+
+                housingData = dbContext.Housing.ToList();
+            }
+
+            // Create a new context for ML.NET operations. It can be used for exception tracking and logging, 
+            // as a catalog of available operations and as the source of randomness.
+            var mlContext = new MLContext();
+
+            // Load the housing dataset into an IDataView from the IEnumerable that was
+            // retrieved from the database.
+            var dataView = mlContext.Data.LoadFromEnumerable(housingData);
+
+            //////////////////// Data Preview ////////////////////
+            // MedianHomeValue    CrimesPerCapita    PercentResidental    PercentNonRetail    CharlesRiver    NitricOxides    RoomsPerDwelling    PercentPre40s
+            // 24.00              0.00632            18.00                2.310               0               0.5380          6.5750              65.20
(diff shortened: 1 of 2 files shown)


CESARDELATORRE:
Adding @divega from the Entity Framework team. Diego, if possible, when you have the time could you review this short sample using EF and ML.NET?
Thanks a lot! 👍

CESARDELATORRE:
Adding @singlis nglis - Related: Scott Inglis is working on an Eitity Framework - SQL  sample for the ML.NET Samples repo, here:

https://github.com/dotnet/machinelearning-samples/pull/295

sfilipi:
Hi @jwood803, thanks for the ongoing contributions. 
The purpose of the samples project in the ml.net repo is to illustrate the usage of the API. Those samples are referenced from the ml.net api reference guide, in docs.microsoft.com, like the example in [Fast Tree](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.staticpipe.treeregressionextensions.fasttree?view=ml-dotnet#Microsoft_ML_StaticPipe_TreeRegressionExtensions_FastTree_Microsoft_ML_BinaryClassificationCatalog_BinaryClassificationTrainers_Microsoft_ML_StaticPipe_Scalar_System_Boolean__Microsoft_ML_StaticPipe_Vector_System_Single__Microsoft_ML_StaticPipe_Scalar_System_Single__Microsoft_ML_Trainers_FastTree_FastTreeBinaryClassificationTrainer_Options_System_Action_Microsoft_ML_Calibrators_CalibratedModelParametersBase_Microsoft_ML_Trainers_FastTree_FastTreeBinaryModelParameters_Microsoft_ML_Calibrators_PlattCalibrator___) here. 

Your sample is valuable, but it won't be discoverable from here, as there is currently no link to an ML.Net api. I think the best place for it is either the samples repository, or the docs repository. 

@JRAlexander, @luisquintanilla   what would you suggest? Is there a good place for this sample in the apis repo? under tutorials maybe?

singlis:
@jwood803 - for reference, here is the sample that I was working on in the samples repository:
https://github.com/dotnet/machinelearning-samples/pull/295

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/EntityFrameworkEnumerable.cs
```diff
@@ -0,0 +1,128 @@
...
+        public float NitricOxides { get; set; }
+        public float RoomsPerDwelling { get; set; }
+        public float PercentPre40s { get; set; }
+        public float EmploymentDistance { get; set; }
+        public float HighwayDistance { get; set; }
+        public float TaxRate { get; set; }
+        public float TeacherRatio { get; set; }
+    }
+
+    public partial class MLNetExampleContext : DbContext
```
rogancarr: >partial [](start = 11, length = 7)

Is this a partial class, or is it all in this file?

rogancarr:
Hi @JRAlexander Apologies for the delay &mdash; I just saw this PR. I think this is great, and seeing you implement this in this way, I think we should move this from the Samples repository to the FunctionalTests repository so we can validate this functionality at build time. Do you have bandwidth to do this?

singlis:
@jwood803 and @rogancarr - how would you feel about having this in machinelearning-samples? I have WIP PR that is currently there. @jwood803 I feel at this point that if I make any changes, I could be accused of pulling from your example as they are pretty similar :). I don't mind merging or just using your code. My example uses the dotnet migrations command line to build up a sql lite database which requires the user to run a couple of commands before running the code. Yours does not look to have that requirement since its all in-memory. 

With the machinelearning-samples there is additional documentation that needs to be there to explain what the sample addresses. Also we (we being me, @CESARDELATORRE  and @divega) wanted to get a better idea on what the behavior is with the Entity Framework as ML.Net is pull data through the IEnumerable.

Generating this sample can happen in multiple commits. If you wanted to commit this to machinlearning samples (or use my PR), I would be fine with that. I can follow up with documentation and investigating more about the EF behavior. 

@jwood803 what are your thoughts?

jwood803:
@singlis We can use your PR in the samples repository if everyone believes that's the best place to have this sample. Feel free to use any of this code that you need. I believe your PR was first, anyway 😄

If you need help with the documentation or anything I'll be glad to help.

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/EntityFrameworkEnumerable.cs
divega: From looking at the code, it seems this became partial because it was generated by reverse engineering the database. Partial definitively can be removed.

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/EntityFrameworkEnumerable.cs
```diff
@@ -0,0 +1,128 @@
...
+        public DbSet<HousingData> Housing { get; set; }
+
+        public MLNetExampleContext(DbContextOptions<MLNetExampleContext> options)
+            : base(options)
+        {
+        }
+
+        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+        {
+            if (!optionsBuilder.IsConfigured)
```
divega: This is probably ok to remove. We add it when the context is reverse engineered so that customers that want to pass the DbContext configuration externally don't need to change the generated code.

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/EntityFrameworkEnumerable.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                optionsBuilder.UseSqlServer("Server=(localdb)\\mssqllocaldb;Database=EFProviders.InMemory;Trusted_Connection=True;ConnectRetryCount=0");
```
divega: Worth adding some external configuration files to store the connection string or at least a TO-DO comment here acknowledging that in a real world application this shouldn't be hardcoded and checked in as part of the code.

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/EntityFrameworkEnumerable.cs
```diff
@@ -0,0 +1,128 @@
...
+            // Options to pass to the DbContext to use an in-memory database for
+            // Entity Framework and to give the database a name.
+            var options = new DbContextOptionsBuilder<MLNetExampleContext>()
+                .UseInMemoryDatabase(databaseName: "TestData")
+                .Options;
+
+            // Add the data to the in-memory database via Entity Framework.
+            using (var dbContext = new MLNetExampleContext(options))
+            {
+                foreach (var dataItem in data)
```
divega: Likely this whole loop can be replaced with dbContext.AddRange(data)

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/EntityFrameworkEnumerable.cs
```diff
@@ -0,0 +1,128 @@
...
+                    PercentNonRetail = float.Parse(i[3]),
+                    CharlesRiver = float.Parse(i[4]),
+                    NitricOxides = float.Parse(i[5]),
+                    RoomsPerDwelling = float.Parse(i[6]),
+                    PercentPre40s = float.Parse(i[7])
+                });
+
+            // Options to pass to the DbContext to use an in-memory database for
+            // Entity Framework and to give the database a name.
+            var options = new DbContextOptionsBuilder<MLNetExampleContext>()
```
divega: You may want to consider moving this inside the OnConfiguring method for simplicity, replacing the code currently there. On the other hand, I think an example that uses the in-memory provider (which BTW is recommended only for testing) misses an opportunity to show when you would actually want to pull the data from a real database using EF Core, which I believe @singlis's version of the sample addresses.

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/EntityFrameworkEnumerable.cs
```diff
@@ -0,0 +1,128 @@
...
+            using (var dbContext = new MLNetExampleContext(options))
+            {
+                foreach (var dataItem in data)
+                {
+                    dbContext.Add(dataItem);
+                }
+
+                var rowsAffected = dbContext.SaveChanges();
+
+                housingData = dbContext.Housing.ToList();
```
divega: This ToList() and the ReadAllLines() code that initially loads the data from a text file require that all the data fits in memory at the same time (twice actually). As @singlis mentioned, one of the goals with had in mind was to show how to avoid this.

singlis:
@jwood803 - I missed the context that this demo is being used as part of the v1 scenarios. After talking with @rogancarr he felt that this should go into the feature tests. So now worries on merging,  I will continue with my sample in machinelearning-samples.

jwood803:
Sounds good! Anything I can help with, just let me know. I'll leave the PR
here but it can be closed if needed.

rogancarr:
@singlis and I can take this branch and make it into a functional test.

jwood803:
I'll close this since [PR 405](https://github.com/dotnet/machinelearning-samples/pull/405) in the samples repository got merged in. Thanks all!

PR #2904, opened 2019-03-11
Hide some duplicate methods

### LINKED ISSUE #2897: Two Ways to Save a Model
The current API has two ways to save a model:

```cs
model.SaveTo(MlContext, stream);
mlContext.Model.Save(ITransformer, stream);
```

Do we just want one of these?

Maybe related to #2735

Ivanidzo4ka:
fixes #2897

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/TransformerChain.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/TransformerChain.cs
@@ -200,6 +200,7 @@ namespace Microsoft.ML.Data
                 LastTransformer = null;
         }
 
+        [BestFriend]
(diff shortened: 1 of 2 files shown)


TomFinley:
This seems like it intersects pretty closely with #2858 which as seen will be changing this API rather entirely to the point where what is noted here may be obviated. Should we instead focus on that one?

Ivanidzo4ka:
@TomFinley  I've check #2858 and SaveTo remain public and `CreatePredictEngine` remains in two places.
So I would prefer to check this one, and let @yaeldekel decide should we continue to have two public `CreatePredictionEngine` or we need only one.

PR #2905, opened 2019-03-11
TrainTestSplit should be inside MLContext.Data

### LINKED ISSUE #2337: TrainTestSplit should be inside MLContext.Data
Currently `TrainTestSplit` can be found in the various training tasks in `MLContext`:

https://github.com/dotnet/machinelearning/blob/533171a430063c5db234aff20e6d39e6673d3c4d/test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamples.cs#L604-L605 

Since that is a data operation that only involves splitting data and since it is independent on the training task, it would make more sense to have it in `mlContext.Data`. The previous line would look like:

```csharp
var (trainData, testData) = mlContext.Data.TrainTestSplit(data, testFraction: 0.1);
```

ganik:
fixes #2337


artidoro:
> // The .NET Foundation licenses this file to you under the MIT license.

Could you update the MlNetCookBook.md for both dynamic and static?

---
Refers to: test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamples.cs:2 in c1be3c8. [](commit_id = c1be3c8cd1568a270fa7e3cc87ca16990b1f992a, deletion_comment = False)

Issue #2559, opened 2019-02-14
Make `ISupportBoosterParameterFactory` internal in LightGBM project


zeahmed:
This interface was tried to be made internal in the following PR
https://github.com/dotnet/machinelearning/pull/2476

However, it seems like there is more work needed to make it internal. Thus, creating this issue to track the work.

PR #2909, opened 2019-03-11
Make IServerFactory internal

### LINKED ISSUE #1973: TlcModule.ComponentKind and the curse of the empty interfaces
Consider this attribute class:

https://github.com/dotnet/machinelearning/blob/7e21afa4612ddca53bbf9d67b84df2b58d4636c2/src/Microsoft.ML.Core/EntryPoints/ModuleArgs.cs#L82

It *looks* harmless enough, but kind of isn't. It exists, as far as I can tell, to give some sort of "name" to various types of entry-point nodes. These mostly serve to decorate specific `IComponentFactory` subinterfaces. However, because we need to attach this attribute to a definite type, we have the following implication: that we've littered our codebase with a bunch of interfaces as subinterfaces of `IComponentFactory`, and these exist only so we can attach attributes to them. To take just one example:

https://github.com/dotnet/machinelearning/blob/7e21afa4612ddca53bbf9d67b84df2b58d4636c2/src/Microsoft.ML.Data/Utils/LossFunctions.cs#L84-L87

That's fairly obnoxious. From an API user's perspective, this represents a completely pointless complication -- all an API user really wants to do is be able to set a loss function. But because we have this other "entry point" thing we also have to work through this generic way to create subcomponents (this `IComponentFactory` thing, which is subideal but possibly OK), but then because we also wanted to attach an attribute to this "type" of set of component factories which we did through attributes on the type, which required creating this empty interface that otherwise has no reason to exist -- etc., etc. Anyway, we see that things have piled up.

What I might like instead is the following:

1. In all cases, never exposing the empty `IComponentFactory` derived empty interfaces that exist just for the sake of this attribute never be publicly accessible.

2. Ideally have some alternative for entry-points that does not intersect with the public surface of the API in such a visible way. I am having difficulty with suggesting a specific course here, since I am having difficulty finding anyone that knows why this attribute was introduced in the first place, or how it is used in entry-points. (Though, it seems *to* be used.)

3. Where possibly and reasonable, replace the public surface of the API of `IComponentFactory<Foo>` with ` Foo` directly. As seen above for the case of loss functions, there is no reason not to do this. Indeed, in some places we have done this, which is good:

   https://github.com/dotnet/machinelearning/blob/7e21afa4612ddca53bbf9d67b84df2b58d4636c2/src/Microsoft.ML.StaticPipe/SdcaStaticExtensions.cs#L170

   Yet there are too many places in the code where we have not, e.g., here:
   https://github.com/dotnet/machinelearning/blob/7e21afa4612ddca53bbf9d67b84df2b58d4636c2/src/Microsoft.ML.StandardLearners/StandardLearnersCatalog.cs#L42
   or here:
 
 
 https://github.com/dotnet/machinelearning/blob/7e21afa4612ddca53bbf9d67b84df2b58d4636c2/src/Microsoft.ML.StandardLearners/Standard/SdcaBinary.cs#L1600

   The reasons for this are somewhat historical, in that we had the attitude for a while that the statically typed API would be the public face, but if the dynamic API is to be the public face then it should be a little less terrible than it is now.

ganik:
fixes #1973
This concludes the issue 1973. The only remaining issue of same kind is making ISupportBoosterParameterFactory internal and its addressed by #2559

### DIFF
--- a/src/Microsoft.ML.Core/Data/ServerChannel.cs
+++ b/src/Microsoft.ML.Core/Data/ServerChannel.cs
@@ -140,7 +140,8 @@ namespace Microsoft.ML.Runtime
         /// Entry point factory for creating <see cref="IServer"/> instances.
         /// </summary>
         [TlcModule.ComponentKind("Server")]
-        public interface IServerFactory : IComponentFactory<IChannel, IServer>
+        [BestFriend]
+        internal interface IServerFactory : IComponentFactory<IChannel, IServer>
         {
             new IServer CreateComponent(IHostEnvironment env, IChannel ch);
         }
--- a/src/Microsoft.ML.Data/Commands/DataCommand.cs
+++ b/src/Microsoft.ML.Data/Commands/DataCommand.cs
@@ -46,7 +46,8 @@ namespace Microsoft.ML.Data
             public bool? Verbose;
 
             [Argument(ArgumentType.AtMostOnce, Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly, HelpText = "The web server to publish the RESTful API", Hide = true)]
-            public ServerChannel.IServerFactory Server;
+            [BestFriend]
+            internal ServerChannel.IServerFactory Server;
 
             // This is actually an advisory value. The implementations themselves are responsible for
             // determining what they consider appropriate, and the actual heuristics is a bit more


src/Microsoft.ML.Data/Commands/DataCommand.cs
```diff
@@ -46,7 +46,8 @@ public abstract class ArgumentsBase
             public bool? Verbose;
 
             [Argument(ArgumentType.AtMostOnce, Visibility = ArgumentAttribute.VisibilityType.CmdLineOnly, HelpText = "The web server to publish the RESTful API", Hide = true)]
-            public ServerChannel.IServerFactory Server;
+            [BestFriend]
+            internal ServerChannel.IServerFactory Server;
```
Ivanidzo4ka: >internal [](start = 12, length = 8)

it's part of nested class in internal class. It's already invisible. You don't have to do that.

src/Microsoft.ML.Core/Data/ServerChannel.cs
```diff
@@ -140,7 +140,8 @@ public void Acknowledge(Action<Bundle> toDo)
         /// Entry point factory for creating <see cref="IServer"/> instances.
         /// </summary>
         [TlcModule.ComponentKind("Server")]
-        public interface IServerFactory : IComponentFactory<IChannel, IServer>
+        [BestFriend]
+        internal interface IServerFactory : IComponentFactory<IChannel, IServer>
```
Ivanidzo4ka: >IServerFactory [](start = 27, length = 14)

it's a nested interface inside internal class.
It's already hidden.

TomFinley:
Hi @ganik as mentioned by @Ivanidzo4ka  already hidden. Shall we close?

PR #2923, opened 2019-03-12
Logistic Regression NumberOfIterations to MaximumNumberOfIterations

### LINKED ISSUE #2922: Logistic Regression: NumberOfIterations should be MaximumNumberOfIterations
Similar to `SDCA` and `K-Means`, `Logistic Regression` has a `NumberOfIterations` option that stands for the maximum number of iterations.

I propose changing it to `MaximumNumberOfIterations`.

** I searched through all the learners, and this is the last learner that needs this change.

rogancarr:
This PR updates `Logistic Regression` (multiclass, regression aka poisson, and multiclass) to specify the `MaximumNumberOfIterations` instead of `NumberOfIterations` as it's a stopping criterion and not necessarily a tuning parameter.

I also took a look around the codebase, and this is the last change of this sort that will be needed.

Fixes #2922

### DIFF
--- a/src/Microsoft.ML.KMeansClustering/KMeansPlusPlusTrainer.cs
+++ b/src/Microsoft.ML.KMeansClustering/KMeansPlusPlusTrainer.cs
@@ -78,7 +78,7 @@ namespace Microsoft.ML.Trainers
             /// <summary>
             /// Maximum number of iterations.
             /// </summary>
-            [Argument(ArgumentType.AtMostOnce, HelpText = "Maximum number of iterations.", ShortName = "maxiter")]
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Maximum number of iterations.", ShortName = "maxiter, NumberOfIterations")]
             [TGUI(Label = "Max Number of Iterations")]
             public int MaximumNumberOfIterations = 1000;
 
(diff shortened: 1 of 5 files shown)


abgoswam:
Could you kindly cross-check with @Ivanidzo4ka on this change ?  I think we are calling it `NumberOfIterations`  uniformly across all the learners

rogancarr:
@abgoswam On the standards for naming, we made this decision in #2871 and addressed K-Means and SCDA in #2883. This PR makes the change for the L-BFGS family (`Logistic Regression`, `Poisson Regression`, and `MulticlassLogisticRegression`). I missed this family in the previous PR, and I've triple-checked that this is the last one with a max-iterations.

On calling it a "stopping criterion" vs. "hyperparameter", this family of algorithms use a maximum number of iterations as one of a two stopping criteria. Usually (almost always) the L-BFGS solver will stop based on a measure of goodness (in this version it's something like the change in the geometric mean of differences in loss functions between iterations? It's been a while since I have been in that part of the codebase, so don't take that as a given). Max iterations is a secondary stopping criterion that acts as a sanity check to stop the algorithm from spinning forever. You could technically use it as a hyperparameter for early stopping by setting it to a small number so that it's guaranteed to trigger before the goodness criterion. That said, I don't think you'll see this much in practice. One place you do run into it, though, is SparkML, where they have a pretty small default max iterations. I think that is for a good "first run" experience; to get a good model there you usually have to set it to something large so it won't trigger.

abgoswam:
Thanks @rogancarr for the explanation.

PR #2924, opened 2019-03-12
Added an extension method for saving statically typed model (#1286)

### LINKED ISSUE #1286: `MLContext.Model.Save` should accept statically typed models too
Add an extension method in `Microsoft.ML.StaticPipe` for that

llRandom:
Added an extension method for saving statically typed model
Fixes #1286


TomFinley:
Hi @llRandom thanks for the PR!

In #2735 we expressed a change in this API that they should accept not only the model, but also some description of the inputs to that model (so as to make certain information discoverable). In the base so-called "dynamic" API, this descriptor takes the form of either:

* `DataViewSchema` (whose static analogue is imperfect, but we might accept as being some subclass of `SchemaBearing<TShape>`,
* `IDataLoader<...>` whose static analogue is `DataLoader<TIn, TShape>` (note that this happens to descend from `SchemaBearing<TShape>`, but due to the way overload resolution works that should be fine).

Note also the new presence of overloads for taking file paths and not just streams, for convenience sake.

Note the APIs were removed altogether and put into their final form in #3044, so I might think this code may not build at this time.

src/Microsoft.ML.StaticPipe/ModelOperationsCatalog.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.IO;
+
+namespace Microsoft.ML.StaticPipe
+{
+    public static class ModelOperationsCatalog
```
TomFinley: >ModelOperationsCatalog [](start = 24, length = 22)

It may be preferable if the static class containing the extension methods for this did not have the same name as the type they're operating on, that is `ModelOperationsCatalogExtensions` or suchlike may be preferable.

src/Microsoft.ML.StaticPipe/ModelOperationsCatalog.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        /// <summary>
+        /// Save statically typed model to the stream.
+        /// </summary>
+        /// <param name="catalog">The model explainability operations catalog.</param>
+        /// <param name="model">The trained model to be saved.</param>
+        /// <param name="stream">A writeable, seekable stream to save to.</param>
+        public static void Save<TInShape, TOutShape, TTransformer>(this ML.ModelOperationsCatalog catalog, Transformer<TInShape, TOutShape, TTransformer> model, Stream stream)
```
TomFinley: >ModelOperationsCatalog catalog [](start = 75, length = 30)

This is an interesting question I'm not sure I know the answer to. The experimental "static" API predates the concept of the `MLContext`, and solved the problem `MLContext` seeks to solve -- that of discoverability, and (at a lower level) propagation of the `IHostEnvironment`, in a rather entirely different way -- that is, you apply estimators and transformers in static-pipe land through discoverability on the input data to those estimators and transformers -- that is, intellisense on the columns.

Dynamic pipe land can't do that -- there is no (and kind of cannot be) concept of a column as "separate object" in dynamic-pipe land. So we achieve discoverability of relevant instantiable components via the taxonomy as originally proposed in #1098.

Here we're, AFAIK for the first time, mixing the two metaphors. This may be appropriate. Indeed I think we might shift the static API to use more of these idioms (which, again, predates `MLContext` altogether during the period we were engineering its core principles).

Not asking you to answer that question, just registering my thoughts since this change brought it to my mind, if we continue to iterate on the static pipe...

test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamples.cs
```diff
@@ -149,7 +149,7 @@ private void TrainRegression(string trainDataPath, string testDataPath, string m
             using (var stream = File.Create(modelPath))
             {
                 // Saving and loading happens to 'dynamic' models, so the static typing is lost in the process.
-                mlContext.Model.Save(model.AsDynamic, stream);
+                mlContext.Model.Save(model, stream);
```
TomFinley: >stream [](start = 44, length = 6)

Note that when we update these samples in the `CookbookSamples.cs`, there is usually a section in the actual cookbook that needs to be updated. (In this case I think [it is this one](https://github.com/dotnet/machinelearning/blob/9caafb04d2b4f298f2a9479a9550597afee32c94/docs/code/experimental/MlNetCookBookStaticApi.md#how-do-i-save-and-load-the-model).)

src/Microsoft.ML.StaticPipe/ModelOperationsCatalog.cs
llRandom: fixed

test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamples.cs
llRandom: fixed

llRandom:
Are there any other changes needed for this PR?

eerhardt:
> Are there any other changes needed for this PR?

@TomFinley - were all of your comments addressed?

eerhardt:
The static API has been removed. Closing as this PR is no longer necessary.

Issue #191, opened 2018-05-19
Project will not build because of space in build path


atotalnoob:
### System information

- **OS version/distro**: Windows 10 Home
- **.NET Version (eg., dotnet --info)**: 2.1.200, concerning ML .net V0.1.0

### Issue

- **What did you do?** Try to build from source
- **What happened?** When trying to build, I have a space in my user profile, for example let's say "John Smith".  So when I cloned to C:/users/John Smith/source/repo/machinelearning/ and tried to build using a command prompt (in the before directory) using command `start build.cmd -Release -TargetArchitecture:x64` 

Another window opened (expected) and it said `'C:\Users\John' is not recognized as an internal or external command, operable program or batch file`

Moving the folder up to C:/ it builds just fine. 
- **What did you expect?** To build

### Source code / logs

Its probably caused by %~dp0 in the build.cmd file.

danmosemsft:
@atotalnoob I'm surprised this wasn't reported in older repos like CoreFX that seem to have the same issue. But this requires both that you have a space in your profile and you clone your repo into your user profile. 

Could you try putting quotes around the `%~dp0...` in build.cmd, run.cmd, and native\build.cmd and see whether that is enough?

PR #2925, opened 2019-03-12
handle space in the directory path.

andrewkittredge:

fixes https://github.com/dotnet/machinelearning/issues/191

### DIFF
--- a/init-tools.cmd
+++ b/init-tools.cmd
@@ -75,7 +75,7 @@ if NOT exist "%DOTNET_LOCAL_PATH%" (
 if exist "%BUILD_TOOLS_PATH%" goto :afterbuildtoolsrestore
 echo Restoring BuildTools version %BUILDTOOLS_VERSION%...
 echo Running: "%DOTNET_CMD%" restore "%INIT_TOOLS_RESTORE_PROJECT%" --no-cache --packages "%PACKAGES_DIR%" --source "%BUILDTOOLS_SOURCE%" /p:BuildToolsPackageVersion=%BUILDTOOLS_VERSION% /p:ToolsDir=%TOOLRUNTIME_DIR% >> "%INIT_TOOLS_LOG%"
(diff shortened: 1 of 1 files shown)


init-tools.cmd
```diff
@@ -75,7 +75,7 @@ if NOT exist "%DOTNET_LOCAL_PATH%" (
 if exist "%BUILD_TOOLS_PATH%" goto :afterbuildtoolsrestore
 echo Restoring BuildTools version %BUILDTOOLS_VERSION%...
 echo Running: "%DOTNET_CMD%" restore "%INIT_TOOLS_RESTORE_PROJECT%" --no-cache --packages "%PACKAGES_DIR%" --source "%BUILDTOOLS_SOURCE%" /p:BuildToolsPackageVersion=%BUILDTOOLS_VERSION% /p:ToolsDir=%TOOLRUNTIME_DIR% >> "%INIT_TOOLS_LOG%"
```
eerhardt: Can you fix up this line to match the line below?

andrewkittredge: This still does not build because I have a space in my path.


When I run -
C:\Users\Andrew Kittredge\Source\Repos\machinelearning\src\Native> .\build.cmd.

I get 

CMake Error: The source directory "C:/Users/Andrew Kittredge/Source/Repos/machinelearning/bin/obj/x64.Debug/Native/Kittredge/Source/Repos/machinelearning/src/Native"".

eerhardt: @andrewkittredge - I don't understand your comment. Are you saying your change didn't fix the issue you said it is fixing?

My comment here is regarding that this line `echo command` is supposed to be a duplicate of `command`. So if you change the `command` line - we should keep the `echo command` line up-to-date.

shauheen: Was this fixed before merge? @TomFinley , @eerhardt

Issue #2751, opened 2019-02-26
Namespace reorg for the public surface 


sfilipi:
Related to #2326, and the feedback of the ML.Net public surface API recommended the following changes:
 
1 - Microsoft.ML.Trainers.* everything should go to Microsoft.ML.Trainers, except FastTree. 
2- Microsoft.ML.Transforms.* should go to Microsoft.ML.Transforms with the exception of Text and Images

cc @Ivanidzo4ka , @eerhardt @TomFinley

sfilipi:
There is more to do on this:
remove:
Microsoft.ML.Transforms.TimeSeries 
 Microsoft.ML.Transforms.FeatureSelection
Microsoft.ML.Transforms.Tensorflow
Microsoft.ML.Trainers.Recommender

Ivanidzo4ka:
Well issue states Microsoft.ML package, and none of this namespaces is in that package.

sfilipi:
Proposal for the final namespace landscape of the public surface:

*  Microsoft.ML.ImageAnalytics - contains both the ImageLoader and the other image transforms. 
The ImageLoader gets moved into ML.Data, where TextLoader is. The other image transforms get placed into Microsoft.ML.Transforms.Image (consistent with Microsoft.ML.Transforms.Text).

* Microsoft.ML.LightGBM -> changes to Microsoft.ML.Trainers.LightGBM*

* Microsoft.ML.Trainers.Recommender  -> stays as is, if we follow the pattern of: external trainer code wrapped from ML.NET  gets its own namespace (e.g. Light GBM).

*  Microsoft.ML.Transforms.FeatureSelection -> moves to Microsoft.ML.Transforms

The proposal above is me synthesizing the discussion we have had in the past about trainers. 

**My personal opinion** is that  we should not have Microsoft.ML.Trainers.FastTree and Microsoft.ML.Trainers.LightGBM. We can fold them both into a Microsoft.ML.Trainers.Trees, call the linear ones Microsoft.ML.Trainers.Linear (ensembles can come on Microsoft.ML.Trainers.Ensembles, when time. ). 

The:
* [Onnx Transformer/Estimator](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs),  
* The[ Onnx catalog](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.OnnxTransformer/OnnxCatalog.cs) 
[...]

PR #2929, opened 2019-03-12
Move ONNX Transformer into Microsoft.ML.Transforms.Onnx namespace.

codemzs:
towards #2751

### DIFF
--- a/src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AlexNetExtension.cs
+++ b/src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AlexNetExtension.cs
@@ -7,7 +7,7 @@ using System.IO;
 using Microsoft.ML.Data;
 using Microsoft.ML.Runtime;
 
-namespace Microsoft.ML.Transforms
+namespace Microsoft.ML.Transforms.Onnx
 {
     /// <summary>
     /// This is an extension method to be used with the <see cref="DnnImageFeaturizerEstimator"/> in order to use a pretrained AlexNet model.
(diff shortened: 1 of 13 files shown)


src/Microsoft.ML.DnnImageFeaturizer.AlexNet/AlexNetExtension.cs
```diff
@@ -7,7 +7,7 @@
 using Microsoft.ML.Data;
 using Microsoft.ML.Runtime;
 
-namespace Microsoft.ML.Transforms
+namespace Microsoft.ML.Transforms.Onnx
```
sfilipi: all other extensions seem to be in the Microsoft.ML namespace. Let's leave this and the Resnet's there too.

sfilipi: Signing off, but please move the extensions.

codemzs: thanks, done.

src/Microsoft.ML.DnnImageFeaturizer.ResNet18/ResNet18Extension.cs
```diff
@@ -6,8 +6,10 @@
 using System.IO;
 using Microsoft.ML.Data;
 using Microsoft.ML.Runtime;
+using Microsoft.ML.Transforms;
+using Microsoft.ML.Transforms.Onnx;
 
-namespace Microsoft.ML.Transforms
+namespace Microsoft.ML
```
TomFinley: Ah, nice, I had forgotten about these extensions.

PR #2935, opened 2019-03-12
Added support for inserting batch dimension in inputs in TensorFlow.

### LINKED ISSUE #2778: Possible Bug Tensorflow in v0.11 - Invalid argument: input must be 4-dimensional[224,224,3]

zeahmed:
This PR fixes #2778.

It is difficult to induce shape of the inputs from the data or model when the model accepts input of any shape but internal operators requires the input in particular shape. This is the problem with the inception model available at the following location.

https://storage.googleapis.com/download.tensorflow.org/models/inception5h.zip

The model takes input data of any shape. There is a convolution layer just after the input which requires 4-D input. The first dimension for Conv2D operation in TensorFlow is the batch dimension. That's causing failure of samples in #2778. The ML.NET input is [224, 244, 3] while convolution layer in the above model requires [-1, 224, 224, 3]. The ultimate solution to this problem is to have reshape transform #765.

However, it will take time implement. To unblock #2778, the temporary solution implemented here is to add a parameter in options class or other public interfaces called “AddBatchDimensionInput”. When user set it to true, batch dimension would be added to the inputs otherwise not. 

NOTE: Once we have the `ReshapeTransform`. This change needs to be reverted.

### DIFF
--- a/src/Microsoft.ML.TensorFlow.StaticPipe/TensorFlowStaticExtensions.cs
+++ b/src/Microsoft.ML.TensorFlow.StaticPipe/TensorFlowStaticExtensions.cs
@@ -14,14 +14,14 @@ namespace Microsoft.ML.Transforms.StaticPipe
         {
             public PipelineColumn Input { get; }
 
-            public OutColumn(Vector<float> input, string modelFile)
-                : base(new Reconciler(modelFile), input)
+            public OutColumn(Vector<float> input, string modelFile, bool addBatchDimensionInput)
+                : base(new Reconciler(modelFile, addBatchDimensionInput), input)
             {
                 Input = input;
             }
 
-            public OutColumn(Vector<float> input, TensorFlowModel tensorFlowModel)
-                : base(new Reconciler(tensorFlowModel), input)
+            public OutColumn(Vector<float> input, TensorFlowModel tensorFlowModel, bool addBatchDimensionInput)
+                : base(new Reconciler(tensorFlowModel, addBatchDimensionInput), input)
             {
                 Input = input;
             }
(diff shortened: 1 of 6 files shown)


Ivanidzo4ka:
>while convolution layer in the above model requires [-1, 224, 224, 3]

Are you sure about that? I though we properly handle variable length tensors, and that [-1,244,244,3] would be casted to [1,244,244,3] and we internally would stretch our image [244,244,3] into it. 

Then I look into it - input layer had [1,244,244,3] and that was reason why we can't handle it, since it's not variable length, it's fixed 4 dimensional tensor instead of tensor of variable length.

zeahmed:
Yes, the input is variable length with shape [-1]. However, model actually wants input in [B, H, W, C] format and this shape cannot be inferred automatically.

---
In reply to: [472221487](https://github.com/dotnet/machinelearning/pull/2935#issuecomment-472221487) [](ancestors = 472221487)

PR #2936, opened 2019-03-12
Hide more things in Data assembly

### LINKED ISSUE #1704: Make members of Data and Transforms assemblies internal
Similar to #1602, let's make as many members of these assemblies internal as possible

### LINKED ISSUE #2926: More hiding of unnecessary public classes
https://github.com/dotnet/machinelearning/blob/40abffc4eb98990c86804a05ca707a467edbe3d6/src/Microsoft.ML.Data/Commands/EvaluateCommand.cs#L30

https://github.com/dotnet/machinelearning/blob/40abffc4eb98990c86804a05ca707a467edbe3d6/src/Microsoft.ML.Data/Commands/TrainCommand.cs#L28

https://github.com/dotnet/machinelearning/blob/40abffc4eb98990c86804a05ca707a467edbe3d6/src/Microsoft.ML.Data/DataLoadSave/TrivialLoaderEstimator.cs#L10

https://github.com/dotnet/machinelearning/blob/40abffc4eb98990c86804a05ca707a467edbe3d6/src/Microsoft.ML.StandardTrainers/Standard/MultiClass/OneVersusAllTrainer.cs#L244 Would ask @sfilipi  since she is mentioned she works on OVA output columns, maybe this enum would proof useful.

https://github.com/dotnet/machinelearning/blob/40abffc4eb98990c86804a05ca707a467edbe3d6/src/Microsoft.ML.Data/Transforms/SlotsDroppingTransformer.cs#L38 Would double check with @TomFinley  right now, all we have is public class, and everything inside it is hidden, so it's useless for user. We don't have `IEstimator` for it, so I'm not sure how valuable it's to have it.

Ivanidzo4ka:
Fixes #2926. Fixes #1704.

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/SlotsDroppingTransformer.cs
+++ b/src/Microsoft.ML.Data/Transforms/SlotsDroppingTransformer.cs
@@ -35,7 +35,8 @@ namespace Microsoft.ML.Transforms
     /// If all the slots are to be dropped, a vector valued column will be changed to a vector of length 1 (a scalar column will retain its type) and
     /// the value will be the default value.
     /// </summary>
-    public sealed class SlotsDroppingTransformer : OneToOneTransformerBase
+    [BestFriend]
+    internal sealed class SlotsDroppingTransformer : OneToOneTransformerBase
     {
         internal sealed class Options
         {
--- a/src/Microsoft.ML.Data/Commands/EvaluateCommand.cs
+++ b/src/Microsoft.ML.Data/Commands/EvaluateCommand.cs
@@ -27,7 +27,8 @@ namespace Microsoft.ML.Data
     /// This class contains information about an overall metric, namely its name and whether it is a vector
     /// metric or not.
     /// </summary>
-    public sealed class MetricColumn
+    [BestFriend]
+    internal sealed class MetricColumn
     {
         /// <summary>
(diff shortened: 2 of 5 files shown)


src/Microsoft.ML.Data/Transforms/SlotsDroppingTransformer.cs
```diff
@@ -35,7 +35,8 @@ namespace Microsoft.ML.Transforms
     /// If all the slots are to be dropped, a vector valued column will be changed to a vector of length 1 (a scalar column will retain its type) and
     /// the value will be the default value.
     /// </summary>
-    public sealed class SlotsDroppingTransformer : OneToOneTransformerBase
+    [BestFriend]
+    internal sealed class SlotsDroppingTransformer : OneToOneTransformerBase
```
TomFinley: Thanks @Ivanidzo4ka ! Your changes seem good for the most part, though I might question this one in particular: if we had a feature selection estimator that used this (which I believe is one of the applications), it may be worthwhile to expose, since the most natural question one would want to ask upon fitting that is, "ok, which slots did you drop/keep", which would require exposing this anyway. But as you like. The other hidings seem good!

sfilipi:
Closes #1704

Issue #980, opened 2018-09-21
Avoids pointer overflow in bound checking of SSE/AVX intrinsics


briancylui:
#821 references and aims to solve this issue.
---
Suggested by @ahsonkhan to avoid integer overflow in bound checking inside SSE/AVX intrinsics implementation, i.e. change all `while (pCurrent + 8 OR 4 <= pEnd)` into `while (pEnd - pCurrent >= 8 OR 4)`.

Perf tests results before and after the change are shown below:
## Before the change:
``` ini
BenchmarkDotNet=v0.11.1, OS=Windows 10.0.17134.228 (1803/April2018Update/Redstone4)
Intel Core i7-7700 CPU 3.60GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.0.100-alpha1-20180720-2
  [Host] : .NET Core 3.0.0-preview1-26710-03 (CoreCLR 4.6.26710.05, CoreFX 4.6.26708.04), 64bit RyuJIT

Toolchain=InProcessToolchain
```
|                   Type | Method |     Mean |    Error |    StdDev |
|----------------------- |------- |---------:|---------:|----------:|
|    AvxPerformanceTests |   SumU | 159.4 us | 1.104 us | 0.9784 us |
| NativePerformanceTests |   SumU | 283.5 us | 5.492 us | 4.8687 us |
|    SsePerformanceTests |   SumU | 281.2 us | 1.472 us | 1.3045 us |
|    AvxPerformanceTests |   AddU | 276.1 us | 3.018 us | 2.520 us |
| NativePerformanceTests |   AddU | 330.1 us | 3.585 us | 3.178 us |
|    SsePerformanceTests |   AddU | 325.6 us | 6.883 us | 7.926 us |

## After the change:
``` ini
BenchmarkDotNet=v0.11.1, OS=Windows 10.0.17134.228 (1803/April2018Update/Redstone4)
Intel Core i7-7700 CPU 3.60GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.0.100-alpha1-20180720-2
  [Host] : .NET Core 3.0.0-preview1-26710-03 (CoreCLR 4.6.26710.05, CoreFX 4.6.26708.04), 64bit RyuJIT

Toolchain=InProcessToolchain
```
|                   Type | Method |     Mean |    Error |   StdDev |
|----------------------- |------- |---------:|---------:|---------:|
[...]

PR #821, opened 2018-09-05
Change bound checking in SSE/AVX intrinsics to avoid pointer overflow

briancylui:
Aims to solve #980
---
Suggested by @ahsonkhan to avoid integer overflow in bound checking inside SSE/AVX intrinsics implementation, i.e. change all `while (pCurrent + 8 OR 4 <= pEnd)` into `while (pEnd - pCurrent >= 8 OR 4)`.

Perf tests results before and after the change are shown below:
## Before the change:
``` ini
BenchmarkDotNet=v0.11.1, OS=Windows 10.0.17134.228 (1803/April2018Update/Redstone4)
Intel Core i7-7700 CPU 3.60GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.0.100-alpha1-20180720-2
  [Host] : .NET Core 3.0.0-preview1-26710-03 (CoreCLR 4.6.26710.05, CoreFX 4.6.26708.04), 64bit RyuJIT

Toolchain=InProcessToolchain
```
|                   Type | Method |     Mean |    Error |    StdDev |
|----------------------- |------- |---------:|---------:|----------:|
|    AvxPerformanceTests |   SumU | 159.4 us | 1.104 us | 0.9784 us |
| NativePerformanceTests |   SumU | 283.5 us | 5.492 us | 4.8687 us |
|    SsePerformanceTests |   SumU | 281.2 us | 1.472 us | 1.3045 us |
|    AvxPerformanceTests |   AddU | 276.1 us | 3.018 us | 2.520 us |
| NativePerformanceTests |   AddU | 330.1 us | 3.585 us | 3.178 us |
|    SsePerformanceTests |   AddU | 325.6 us | 6.883 us | 7.926 us |

## After the change:
``` ini
BenchmarkDotNet=v0.11.1, OS=Windows 10.0.17134.228 (1803/April2018Update/Redstone4)
Intel Core i7-7700 CPU 3.60GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.0.100-alpha1-20180720-2
  [Host] : .NET Core 3.0.0-preview1-26710-03 (CoreCLR 4.6.26710.05, CoreFX 4.6.26708.04), 64bit RyuJIT

Toolchain=InProcessToolchain
```
|                   Type | Method |     Mean |    Error |   StdDev |
|----------------------- |------- |---------:|---------:|---------:|
[...]

PR #2939, opened 2019-03-13
CpuMath Enhancement: Make bound checking of loops in hardware intrinsics more efficient

### LINKED ISSUE #835: CpuMath Enhancement: Make bound checking of loops in hardware intrinsics more efficient
Style changes needed to solve part of https://github.com/dotnet/machinelearning/issues/823

## Details
-  In `src\Microsoft.ML.CpuMath\SseIntrinsics.cs` and `src\Microsoft.ML.CpuMath\AvxIntrinsics.cs`, changing `while (pDstCurrent + 4 <= pDstEnd)` for the loop bound checking into `while (pDstCurrent <= pDstEnd - 4)` to save an instruction (ref: https://github.com/dotnet/machinelearning/pull/668#issuecomment-412121893)
- It may probably be a CoreCLR issue

llRandom:
Fixes #835

### DIFF
--- a/src/Microsoft.ML.CpuMath/AvxIntrinsics.cs
+++ b/src/Microsoft.ML.CpuMath/AvxIntrinsics.cs
@@ -425,7 +425,7 @@ namespace Microsoft.ML.Internal.CpuMath
 
                 Vector256<float> scalarVector256 = Vector256.Create(scalar);
 
-                while (pDstCurrent + 8 <= pDstEnd)
+                while (pDstCurrent <= pDstEnd - 8)
                 {
                     Vector256<float> dstVector = Avx.LoadVector256(pDstCurrent);
                     dstVector = Avx.Add(dstVector, scalarVector256);
@@ -577,7 +577,7 @@ namespace Microsoft.ML.Internal.CpuMath
 
                 Vector256<float> scaleVector256 = Vector256.Create(scale);
 
-                while (pDstCurrent + 8 <= pDstEnd)
+                while (pDstCurrent <= pDstEnd - 8)
                 {
                     Vector256<float> srcVector = Avx.LoadVector256(pSrcCurrent);
                     srcVector = Avx.Multiply(srcVector, scaleVector256);
@@ -623,7 +623,7 @@ namespace Microsoft.ML.Internal.CpuMath
                 Vector256<float> a256 = Vector256.Create(a);
                 Vector256<float> b256 = Vector256.Create(b);
 
-                while (pDstCurrent + 8 <= pDstEnd)
+                while (pDstCurrent <= pDstEnd - 8)
                 {
                     Vector256<float> dstVector = Avx.LoadVector256(pDstCurrent);
                     dstVector = Avx.Add(dstVector, b256);
@@ -671,7 +671,7 @@ namespace Microsoft.ML.Internal.CpuMath
 
                 Vector256<float> scaleVector256 = Vector256.Create(scale);
 
-                while (pDstCurrent + 8 <= pEnd)
+                while (pDstCurrent <= pEnd - 8)
                 {
                     Vector256<float> dstVector = Avx.LoadVector256(pDstCurrent);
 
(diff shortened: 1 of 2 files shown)


eerhardt:
We also have other issues in this area that should be addressed if we are working in this area - basically we need to ensure we won't pointer overflow. See #980 and #821.

cc @Anipik @tannergooding @GrabYourPitchforks

src/Microsoft.ML.CpuMath/AvxIntrinsics.cs
```diff
@@ -425,7 +425,7 @@ public static unsafe void AddScalarU(float scalar, Span<float> dst)
 
                 Vector256<float> scalarVector256 = Vector256.Create(scalar);
 
-                while (pDstCurrent + 8 <= pDstEnd)
+                while (pDstCurrent <= pDstEnd - 8)
```
tannergooding: nit: Explicitly cache the loop-end in a local to ensure it also isn't recomputed?

GrabYourPitchforks: If the destination span is empty, wouldn't `pDstEnd - 8` integer underflow back to a very large number, which could lead to an AV?

src/Microsoft.ML.CpuMath/AvxIntrinsics.cs
```diff
@@ -425,7 +425,7 @@ public static unsafe void AddScalarU(float scalar, Span<float> dst)
...
                 {
                     Vector256<float> dstVector = Avx.LoadVector256(pDstCurrent);
```
GrabYourPitchforks: @tannergooding, do you know if there's any instruction with corresponds roughly to `addfloats xmm ptr [rax], xmm0` and if this trio of instructions collapses to that? It seems like it'd be more efficient and would avoid the xmm register spill.

src/Microsoft.ML.CpuMath/AvxIntrinsics.cs
eerhardt: The spans passed into these functions are guaranteed not to be empty in the calling code.

https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.CpuMath/CpuMathUtils.netcoreapp.cs

src/Microsoft.ML.CpuMath/AvxIntrinsics.cs
tannergooding: No, there is only load-forms, so this will compile down to two instructions (ideally):
```
addps tmp, scalarVector128, [pDstCurrent]
movps [pDstCurrent], tmp
```

src/Microsoft.ML.CpuMath/AvxIntrinsics.cs
GrabYourPitchforks: This is also guaranteed in release builds? The contracts routines look like they're surrounded by `[Conditional("DEBUG")]`.

eerhardt: Yes, it is an invariant that these methods are not called with empty spans. Even higher in the callstack (in many places) are the real checks. For example:

https://github.com/dotnet/machinelearning/blob/d38a35edda4e3b3bb9e74fd78f57e7a49a906186/src/Microsoft.ML.Transforms/GcnTransform.cs#L479-L487

Here, when `count == 0`, the code doesn't even get into `CpuMathUtils.Add` below - it exits early.

On the "other side" of these CpuMathUtils (when are are compiling for `netstandard` where C# intrinsics don't exist), we make similar assumptions:

https://github.com/dotnet/machinelearning/blob/a570da14a41f2870eb8f61d84496a58422398253/src/Microsoft.ML.CpuMath/CpuMathUtils.netstandard.cs#L95-L101

`MemoryMarshal.GetReference` is going to fail if you pass in an empty span.

eerhardt: Also note - the CpuMathUtils methods are not public.

llRandom: fixed

llRandom:
Are there any other changes needed for this PR?

codemzs:
@llRandom Do we have any benchmarks after this change? 
@tannergooding @eerhardt  Are you still reviewing this?

eerhardt:
> Are you still reviewing this?

No checks are passing. This needs to pass CI before I can review it.

codemzs:
@llRandom I'm closing this PR since CI legs are failing. Please feel free to re-open this PR and then make sure CI passes and also please post benchmark results and then we can review it,.

PR #2942, opened 2019-03-13
Sort out seed types

### LINKED ISSUE #2917: Unify way of setting random seed
I can understand why we use nullable seed in one place and defaults in another, but why it's uint in one places and int in others is beyond my understanding.

**UInt:**
https://github.com/dotnet/machinelearning/blob/b861b5d64841cbe0f2c866ee7586872aac450a51/src/Microsoft.ML.Data/Transforms/Hashing.cs#L1164

https://github.com/dotnet/machinelearning/blob/b861b5d64841cbe0f2c866ee7586872aac450a51/src/Microsoft.ML.Transforms/Text/NgramHashingTransformer.cs#L892

https://github.com/dotnet/machinelearning/blob/24f782d306f58d31778eddde181c16939be17292/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs#L377

https://github.com/dotnet/machinelearning/blob/b861b5d64841cbe0f2c866ee7586872aac450a51/src/Microsoft.ML.Transforms/Text/WrappedTextTransformers.cs#L184

https://github.com/dotnet/machinelearning/blob/b861b5d64841cbe0f2c866ee7586872aac450a51/src/Microsoft.ML.Transforms/OneHotHashEncoding.cs#L241

**Int:**

https://github.com/dotnet/machinelearning/blob/24f782d306f58d31778eddde181c16939be17292/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs#L135

https://github.com/dotnet/machinelearning/blob/24f782d306f58d31778eddde181c16939be17292/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs#L293

https://github.com/dotnet/machinelearning/blob/b861b5d64841cbe0f2c866ee7586872aac450a51/src/Microsoft.ML.Transforms/RandomFourierFeaturizing.cs#L659

https://github.com/dotnet/machinelearning/blob/b861b5d64841cbe0f2c866ee7586872aac450a51/src/Microsoft.ML.PCA/PCACatalog.cs#L31

https://github.com/dotnet/machinelearning/blob/b861b5d64841cbe0f2c866ee7586872aac450a51/src/Microsoft.ML.Data/MLContext.cs#L79

Ivanidzo4ka:
fixes #2917

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
@@ -374,7 +374,7 @@ namespace Microsoft.ML
         /// they are guaranteed to appear in the same subset (train or test). This can be used to ensure no label leakage from the train to the test set.
         /// If <see langword="null"/> no row grouping will be performed.</param>
         /// <param name="seed">Seed for the random number generator used to select rows for the train-test split.</param>
-        public TrainTestData TrainTestSplit(IDataView data, double testFraction = 0.1, string samplingKeyColumn = null, uint? seed = null)
+        public TrainTestData TrainTestSplit(IDataView data, double testFraction = 0.1, string samplingKeyColumn = null, int? seed = null)
         {
             _env.CheckValue(data, nameof(data));
             _env.CheckParam(0 < testFraction && testFraction < 1, nameof(testFraction), "Must be between 0 and 1 exclusive");
@@ -403,7 +403,7 @@ namespace Microsoft.ML
         /// <summary>
         /// Ensures the provided <paramref name="samplingKeyColumn"/> is valid for <see cref="RangeFilter"/>, hashing it if necessary, or creates a new column <paramref name="samplingKeyColumn"/> is null.
         /// </summary>
-        internal static void EnsureGroupPreservationColumn(IHostEnvironment env, ref IDataView data, ref string samplingKeyColumn, uint? seed = null)
+        internal static void EnsureGroupPreservationColumn(IHostEnvironment env, ref IDataView data, ref string samplingKeyColumn, int? seed = null)
         {
             // We need to handle two cases: if samplingKeyColumn is provided, we use hashJoin to
             // build a single hash of it. If it is not, we generate a random number.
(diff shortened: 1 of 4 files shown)


artidoro:
I noticed that there are some occurrences of `uint seed`, for example in the constructor of `OneHotHashEncodingEstimator.ColumnOptions`. Maybe it would be a good idea to take care of those as well?

Ivanidzo4ka:
We decided if that's hash seed it should be uint, if it's random seed it should be int.
I've check through all code, and I think I unify all seed params to that behavior, but if you see something what doesn't comply with this policy, let me know!

---
In reply to: [472603765](https://github.com/dotnet/machinelearning/pull/2942#issuecomment-472603765) [](ancestors = 472603765)

artidoro:
Ah sorry I missed that in the issue. Ok makes sense!

---
In reply to: [472604860](https://github.com/dotnet/machinelearning/pull/2942#issuecomment-472604860) [](ancestors = 472604860,472603765)

src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
```diff
@@ -403,15 +403,15 @@ public TrainTestData TrainTestSplit(IDataView data, double testFraction = 0.1, s
...
+        internal static void EnsureGroupPreservationColumn(IHostEnvironment env, ref IDataView data, ref string samplingKeyColumn, int? seed = null)
         {
             // We need to handle two cases: if samplingKeyColumn is provided, we use hashJoin to
             // build a single hash of it. If it is not, we generate a random number.
 
             if (samplingKeyColumn == null)
             {
                 samplingKeyColumn = data.Schema.GetTempColumnName("SamplingKeyColumn");
-                data = new GenerateNumberTransform(env, data, samplingKeyColumn, seed);
+                data = new GenerateNumberTransform(env, data, samplingKeyColumn, (uint?)seed);
```
artidoro: >GenerateNumberTransform [](start = 27, length = 23)

Seems that the see for `GenerateRandomNumerTransform` is more a random see right? Why not making that a `int`?

TomFinley: Probably because the `GenerateNumberTransform` is not part of the public surface area is why he (correctly, I think) judged it as being less important.

---
In reply to: [265333130](https://github.com/dotnet/machinelearning/pull/2942#discussion_r265333130) [](ancestors = 265333130)

PR #2947, opened 2019-03-13
FeaturizeText: Add instructions to turn off char- or word-gram generation to the tooltip.

### LINKED ISSUE #2946: It is awkward to turn off char-grams with FeaturizeText
`FeaturizeText` was upgraded to allow specification of n-grams for words and characters. However, now it awkward to use `FeaturizeText` *without* specifying n-grams. It is now necessary to explicitly set `CharFeatureExtractor` as `null`.

This is how to compose a bag-of-words with the current API:
```cs
var pipeline = mlContext.Transforms.Text.FeaturizeText(
    "Features",
    new TextFeaturizingEstimator.Options
    {
        KeepPunctuations = false,
        OutputTokens = true,
        CharFeatureExtractor = null,
        WordFeatureExtractor = new WordBagEstimator.Options { NgramLength = 1},
        VectorNormalizer = TextFeaturizingEstimator.NormFunction.None
    },
    "SentimentText");
```

I would expect to be able to do something like
```cs
CharFeatureExtractor = new WordBagEstimator.Options { NgramLength = 0},
```
But this throws an error that `Skipgrams` is not less-than `NgramLength`, and `Skipgrams` must be positive.

Overall, it is a bit awkward and not obvious that you have to manually null a option. Is this the API we want to ship in v1.0?

rogancarr:
Adding a note on how to turn off char-grams and word-grams in `FeaturizeText`.

Fixes #2946

### DIFF
--- a/src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
+++ b/src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
@@ -128,7 +128,7 @@ namespace Microsoft.ML.Transforms.Text
 
             /// <summary>
             /// Ngram feature extractor to use for words (WordBag/WordHashBag).
-            /// Set to null to turn off n-gram generation for words.
+            /// Set to <see langword="null" /> to turn off n-gram generation for words.
             /// </summary>
             public WordBagEstimator.Options WordFeatureExtractor
             {
@@ -161,7 +161,7 @@ namespace Microsoft.ML.Transforms.Text
 
             /// <summary>
             /// Ngram feature extractor to use for characters (WordBag/WordHashBag).
-            /// Set to null to turn off n-gram generation for characters.
+            /// Set to <see langword="null" /> to turn off n-gram generation for characters.
             /// </summary>
             public WordBagEstimator.Options CharFeatureExtractor
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
```diff
@@ -128,6 +128,7 @@ public sealed class Options : TransformInputBase
 
             /// <summary>
             /// Ngram feature extractor to use for words (WordBag/WordHashBag).
+            /// Set to null to turn off n-gram generation for words.
```
wschin:
```suggestion
            /// Set to <see langword="null"> to turn off n-gram generation for words.
```

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
```diff
@@ -160,6 +161,7 @@ public WordBagEstimator.Options WordFeatureExtractor
 
             /// <summary>
             /// Ngram feature extractor to use for characters (WordBag/WordHashBag).
+            /// Set to null to turn off n-gram generation for characters.
```
wschin: Please make it `langword`.

PR #2949, opened 2019-03-13
Allow only binary classifiers in OVA

### LINKED ISSUE #2920: OVA Multiclass Classification can be instantiated for variety of sub-trainer training tasks
In writing a test for #2859, I have found that it is possible to create an `OVA` (aka "One-Versus-All") multiclass classification trainer learners other than a binary classifier as an input.

One of our V1 Goals is:
- Metacomponents smartly restrict their use to compatible components.   Example: "When specifying what trainer OVA should use, a user will be able to specify any binary classifier. If they specify a regression or multi-class classifier ideally that should be a compile error."

For all types
- **Anomaly Detection** throws on `Fit()` due to data mismatch. Suggests that this could work under ideal conditions.
- **Binary classification** works, as expected.
- **Clustering** is a runtime error upon pipeline construction (OVA checks the model type produced).
- **Multiclass classification** is a runtime error upon pipeline construction (OVA checks the model type produced).
- **Ranking** pipeline can be instantiated, but `Fit()` fails on "mismatch for label column".
- **Regression** pipeline can be instantiated, but `Fit()` fails on "mismatch for label column".

(Note that I didn't run a recommender check because this needs to be coded x64-only at this point.)

**Updated**: With the recent change to one-label-type-per task, Ranking and Regression now fail on the `Fit()`. I've updated the title to reflect that we can no longer train working models, but we can instantiate the pipeline.

ganik:
fixes #2920

### DIFF
--- a/src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
+++ b/src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
@@ -628,7 +628,7 @@ namespace Microsoft.ML
         /// <param name="useProbabilities">Use probabilities (vs. raw outputs) to identify top-score category.</param>
         /// <typeparam name="TModel">The type of the model. This type parameter will usually be inferred automatically from <paramref name="binaryEstimator"/>.</typeparam>
         public static OneVersusAllTrainer OneVersusAll<TModel>(this MulticlassClassificationCatalog.MulticlassClassificationTrainers catalog,
-            ITrainerEstimator<ISingleFeaturePredictionTransformer<TModel>, TModel> binaryEstimator,
+            ITrainerEstimator<BinaryPredictionTransformer<TModel>, TModel> binaryEstimator,
             string labelColumnName = DefaultColumnNames.Label,
             bool imputeMissingLabelsAsNegative = false,
             IEstimator<ISingleFeaturePredictionTransformer<ICalibrator>> calibrator = null,


src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
```diff
@@ -628,7 +628,7 @@ private static ICalibratorTrainer GetCalibratorTrainerOrThrow(IExceptionContext
         /// <param name="useProbabilities">Use probabilities (vs. raw outputs) to identify top-score category.</param>
         /// <typeparam name="TModel">The type of the model. This type parameter will usually be inferred automatically from <paramref name="binaryEstimator"/>.</typeparam>
         public static OneVersusAllTrainer OneVersusAll<TModel>(this MulticlassClassificationCatalog.MulticlassClassificationTrainers catalog,
-            ITrainerEstimator<ISingleFeaturePredictionTransformer<TModel>, TModel> binaryEstimator,
+            ITrainerEstimator<BinaryPredictionTransformer<TModel>, TModel> binaryEstimator,
```
rogancarr: >BinaryPredictionTransformer [](start = 30, length = 27)

Might not be so easy because any given `BinaryPredictionTransformer` might not be an `ISingleFeaturePredictionTransformer`; from what I understand, `OVA` is currently limited to learners with single feature vectors (e.g. not `FieldAwareFactorizationMachines`).

ganik: Heer is the definition for BinaryPredictionTransformer:

    /// <summary>
    /// Base class for the <see cref="ISingleFeaturePredictionTransformer{TModel}"/> working on binary classification tasks.
    /// </summary>
    /// <typeparam name="TModel">An implementation of the <see cref="IPredictorProducing{TResult}"/></typeparam>
    public sealed class BinaryPredictionTransformer<TModel> : SingleFeaturePredictionTransformerBase<TModel>


---
In reply to: [265408746](https://github.com/dotnet/machinelearning/pull/2949#discussion_r265408746) [](ancestors = 265408746)

rogancarr:
Broken build ! I'm fixing.

PR #2950, opened 2019-03-13
Get rid of public tuples

### LINKED ISSUE #2881: Value-tuple stragglers in the public API
There was a prior PR #2581 and issue #2501 related to value-tuples and why they should not be part of our public surface. I have noticed that there are some "stragglers" still remaining in the public API. So, the work is perhaps not yet complete.

The following list is I *believe* complete for `Core`/`Data`/`Transforms`/`FastTree`/`ImageAnalytics`/`KMeansClustering`/`LightGBM`/`PCA`/`Tensorflow`/`StandardLearners`/`Data.DataView` assemblies.

There are three distinct categories where this flaw has remained. (Though the last "category" other has only one item.)

# Properties on transformers

Some transformers are exposing information about themselves via this mechanism.

* KeyToBinaryVectorMappingTransformer.Columns
* MissingValueDroppingTransformer.Columns
* MissingValueIndicatorTransformer.Columns
* CustomStopWordsRemovingTransformer.Columns
* TextNormalizingTransformer.Columns
* TokenizingByCharactersTransformer.Columns
* WordEmbeddingsExtractingTransformer.Columns
* ImageGrayscalingTransformer.Columns
* ImageLoadingTransformer.Columns
* LatentDirichletAllocationTransformer.ItemScoresPerTopic and WordScoresPerTopic

# MLContext estimator creation extension methods

There are some overloads of MLContext extension methods on various catalogs that are stil using it. I view this as a *lesser* sin since this is at least something that could conceivably be fixed using an overload if we decide it is necessary, but I'd still prefer to be consistent.

* ProduceHashedNgrams extension method
* ProduceHashedWordBags extension method
* ProduceNgrams extension method
* ProduceWordBags extension method
* RemoveDefaultStopWords extension method
* TokenizeWords extension method

# Others

Lastly, I see a Microsoft.ML.ColumnOptions global class with an implicit operator from value-tuples. This one is *probably* harmless, since that specific class is for representing a simple case.

/cc @yaeldekel @Ivanidzo4ka

Ivanidzo4ka:
fixes #2881

### DIFF
--- a/src/Microsoft.ML.StaticPipe/LdaStaticExtensions.cs
+++ b/src/Microsoft.ML.StaticPipe/LdaStaticExtensions.cs
@@ -20,8 +20,8 @@ namespace Microsoft.ML.StaticPipe
         /// <param name="result"></param>
         public delegate void OnFit(LatentDirichletAllocationFitResult result);
 
-        public LatentDirichletAllocationTransformer.LdaModelParameters LdaTopicSummary;
-        public LatentDirichletAllocationFitResult(LatentDirichletAllocationTransformer.LdaModelParameters ldaTopicSummary)
+        public LatentDirichletAllocationTransformer.ModelParameters LdaTopicSummary;
+        public LatentDirichletAllocationFitResult(LatentDirichletAllocationTransformer.ModelParameters ldaTopicSummary)
         {
             LdaTopicSummary = ldaTopicSummary;
         }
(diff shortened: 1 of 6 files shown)


src/Microsoft.ML.Transforms/MissingValueDroppingTransformer.cs
```diff
@@ -339,55 +339,4 @@ private void DropNAs<TDst>(ref VBuffer<TDst> src, ref VBuffer<TDst> dst, InPredi
             }
         }
     }
-    /// <summary>
-    /// Drops missing values from columns.
-    /// </summary>
-    public sealed class MissingValueDroppingEstimator : TrivialEstimator<MissingValueDroppingTransformer>
```
wschin: Don't we need this estimator anymore? Is there any alternative?

Ivanidzo4ka: We don't have extensions for this estimator. 
We use transformer as part of other transformers, but I don't see any work with estimator.
I'm tempted to make transformer internal as well, since I don't see how it can be useful for user...

---
In reply to: [265372049](https://github.com/dotnet/machinelearning/pull/2950#discussion_r265372049) [](ancestors = 265372049)

src/Microsoft.ML.StaticPipe/LdaStaticExtensions.cs
```diff
@@ -20,8 +20,8 @@ public sealed class LatentDirichletAllocationFitResult
         /// <param name="result"></param>
         public delegate void OnFit(LatentDirichletAllocationFitResult result);
 
-        public LatentDirichletAllocationTransformer.LdaSummary LdaTopicSummary;
-        public LatentDirichletAllocationFitResult(LatentDirichletAllocationTransformer.LdaSummary ldaTopicSummary)
+        public LatentDirichletAllocationTransformer.LdaModelParameters LdaTopicSummary;
```
singlis: >LdaModelParameters [](start = 52, length = 18)

Since the model paramaters is an internal class, why not call it ModelParameters instead of LdaModelParameters?

src/Microsoft.ML.Transforms/MissingValueDroppingTransformer.cs
TomFinley: Unless we have a scenario I would certainly welcome internalizing it, since we can always de-internalize it once we do have a scenario. But if we view this as too far then I'd understand that as well.

PR #2951, opened 2019-03-13
Separate nuget packaging into Stable and Experimental

maryamariyan:
- [x] I added `StableProjects` property in a new `StablePackagesInfo.props` file, which has name of stable projects separated by semicolon.
- [x] I set the strategy for having two separate versions by adding another file next to default `BranchInfo.props`, called `BranchInfo.Stable.props`. The files contain version information. Either of the two would be used depending on whether or not a project is listed in `StableProjects` property.
- [x] I tested the assembly versions and nuget package versions using the diff in this PR.
- [x] The conditions could perhaps be simplified more - needs review

Currently `BranchInfo.props` generates `0.12.0-preview` for all projects. The Major/Minor/Patch versions to set for stable projects are specified in `BranchInfo.Stable.props` and have identical values to `BranchInfo.props` until we decide to change them in the near future.

Fixes: #2279 
cc: @eerhardt @TomFinley @Ivanidzo4ka @sfilipi @Anipik


build/BranchInfo.Stable.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <PropertyGroup>
+    <MajorVersion>1</MajorVersion>
+    <MinorVersion>13</MinorVersion>
```
Ivanidzo4ka: should it be in sync with BranchInfo.props?

maryamariyan: good point. to keep as same value for now.

build/StablePackagesInfo.props
```diff
@@ -0,0 +1,5 @@
+<Project>
+    <PropertyGroup>
+        <StableProjects>Microsoft.ML;Microsoft.ML.CpuMath;Microsoft.ML.ImageAnalytics;Microsoft.ML.LightGBM;Microsoft.ML.Mkl.Components;Microsoft.ML.Mkl.Redist</StableProjects>
```
Ivanidzo4ka: it's `LightGbm` after recent changes (casing of GBM got change)

Ivanidzo4ka: Since creation of Issue we also created Microsoft.ML.FastTree package which I believe should be part of stable packages.

build/StablePackagesInfo.props
```diff
@@ -0,0 +1,14 @@
...
+        <StableProject Include="Microsoft.ML"/>
+        <StableProject Include="Microsoft.ML.CpuMath"/>
+        <StableProject Include="Microsoft.ML.FastTree"/>
+        <StableProject Include="Microsoft.ML.ImageAnalytics"/>
+        <StableProject Include="Microsoft.ML.LightGbm"/>
+        <StableProject Include="Microsoft.ML.Mkl.Components"/>
+        <StableProject Include="Microsoft.ML.Mkl.Redist"/>
+    </ItemGroup>
+    <PropertyGroup>
+        <StableProjects>@(StableProject, ';')</StableProjects>
```
TomFinley: Ha ha, that's one way I guess!

eerhardt: Does this actually work? Properties in MSBuild are evaluated before Items. Which means when the `StableProjects` property gets defined, `@(StableProject)` will be empty.

---
In reply to: [265677312](https://github.com/dotnet/machinelearning/pull/2951#discussion_r265677312) [](ancestors = 265677312)

Directory.Build.props
```diff
@@ -1,7 +1,10 @@
 <Project>
   <!-- Directory.Build.props contains the common build settings for all projects in the repo. -->
 
-  <Import Project="build/BranchInfo.props" />
```
eerhardt: I think all 3 of these files `BranchInfo`, `BranchInfo.Stable` and `StablePackagesInfo` could all be combined into a single file - `build/BranchInfo.props`.

build/StablePackagesInfo.props
eerhardt: To fix this, you can just define

```xml
<PropertyGroup>
    <StableProjects>
        Microsoft.ML;
        Microsoft.ML.CpuMath;
        Microsoft.ML.FastTree;
    ....
    </StableProjects>
</PropertyGroup>
```

---
In reply to: [265710272](https://github.com/dotnet/machinelearning/pull/2951#discussion_r265710272) [](ancestors = 265710272,265677312)

build/StablePackagesInfo.props
```diff
@@ -0,0 +1,14 @@
+<Project>
+    <ItemGroup>
+        <StableProject Include="Microsoft.ML"/>
```
eerhardt: Note there are a few assemblies inside of this nuget package - so we will want to put all those assemblies in this list as well. Like `ML.Data`, `ML.Transforms`, `ML.StandardTrainers`.

Directory.Build.props
TomFinley: Oh... I interpreted this as being at the *nuget* level, not the assembly level. Is that wrong? SHouldn't it be at nuget level?

eerhardt: It needs to be for both. We use these versions for both NuGet package versions and for AssemblyVersions.

So if you only did it for nuget - you would have a `1.0.0` NuGet package with an assembly inside of it with AssemblyVersion `0.13.0.0`.

build/BranchInfo.props
```diff
@@ -1,8 +1,25 @@
...
+    <IsStableProject Condition="'$(MSBuildProjectName.Contains(.symbols))' == 'false'">$(StableProjects.Contains($(MSBuildProjectName)))</IsStableProject>
+    <IsStableProject Condition="'$(MSBuildProjectName.Contains(.symbols))' == 'true'">$(StableProjects.Contains($(MSBuildProjectName.Substring(0, $(MSBuildProjectName.IndexOf(.symbols))))))</IsStableProject>
     <MajorVersion>0</MajorVersion>
     <MinorVersion>12</MinorVersion>
     <PatchVersion>0</PatchVersion>
     <PreReleaseLabel>preview</PreReleaseLabel>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(IsStableProject)' == 'true'">
+    <MajorVersion>0</MajorVersion>
+    <MinorVersion>15</MinorVersion>
```
eerhardt: I assume this is just for testing?

maryamariyan: yes. 
P.S. I am still working on updating versions for ML.Data and others similar to it.

build/BranchInfo.props
```diff
@@ -1,8 +1,25 @@
...
+        Microsoft.ML.Mkl.Redist;
+    </StableProjects>
+    <IsStableProject Condition="'$(MSBuildProjectName.Contains(.symbols))' == 'false'">$(StableProjects.Contains($(MSBuildProjectName)))</IsStableProject>
+    <IsStableProject Condition="'$(MSBuildProjectName.Contains(.symbols))' == 'true'">$(StableProjects.Contains($(MSBuildProjectName.Substring(0, $(MSBuildProjectName.IndexOf(.symbols))))))</IsStableProject>
     <MajorVersion>0</MajorVersion>
     <MinorVersion>12</MinorVersion>
     <PatchVersion>0</PatchVersion>
     <PreReleaseLabel>preview</PreReleaseLabel>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(IsStableProject)' == 'true'">
```
eerhardt: One other place that uses these versions is the native build. In that world we generate a `version.h` or `version.c` file during the build using these properties.

We may also need to check for the `Native` build here as well.

build/BranchInfo.props
```diff
@@ -1,8 +1,25 @@
 <Project>
   <PropertyGroup>
+    <StableProjects>
+        Microsoft.ML;
+        Microsoft.ML.CpuMath;
```
eerhardt: Are we going to add the other assemblies?

build/BranchInfo.props
eerhardt: You can see it here:

https://github.com/dotnet/buildtools/blob/4642e67685b9c21c4d5160ea7665843a48ed8155/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets#L102-L108

build/BranchInfo.props
maryamariyan: Yes. Seems like only adding them here won't fix their versioning when they're used inside another package

build/BranchInfo.props
maryamariyan: @eerhardt could you please elaborate more on the actionable item here?

eerhardt: We probably won't be able to separate the native assets from non-"1.0" and "1.0" without a decent amount of native build work. Honestly, it wouldn't be worth it IMO, as the native version is more for debugging/informational purposes. It isn't as important as the NuGet package version or the AssemblyVersion.

So I'd say let's call the native assets "stable" in this work. When we ship "1.0", all the native assets will be versioned as "1.0". The only native assembly we build that may not be in a stable package is the `MatrixFactorizationNative` assembly. But again, there isn't much value in having that native assembly versioned as non-"1.0".

src/Native/build.proj
```diff
@@ -2,6 +2,8 @@
   <PropertyGroup>
     <!-- versioning.targets will import this file, so don't import it again -->
     <DisableImportVersioningProps>true</DisableImportVersioningProps>
+    <!-- Use this flag to ensure native assets take stable version -->
+    <UseStableVersionForNativeAssets>true</UseStableVersionForNativeAssets>
```
eerhardt: Does this work? I think the `version.h` file gets generated during this target:

[https://github.com/dotnet/machinelearning/blob/0835c52f64ecdc2225966a6bdadc0a4494b5bec2/build.proj#L33](https://github.com/dotnet/machinelearning/blob/0835c52f64ecdc2225966a6bdadc0a4494b5bec2/build.proj#L33)

Which runs during the build of the root `build.proj` file.

maryamariyan: The Native dlls under `bin\x64.Debug\Native` get the stable version successfully

build/BranchInfo.props
```diff
@@ -1,5 +1,28 @@
...
+      Microsoft.ML.LightGbm;
+      Microsoft.ML.PCA;
+      Microsoft.ML.StandardTrainers;
+      Microsoft.ML.Transforms;
+      Microsoft.ML.Mkl.Components;
+      Microsoft.ML.Mkl.Redist;
+    </StableProjects>
+    <IsStableProject Condition="'$(MSBuildProjectName.Contains(.symbols))' == 'false'">$(StableProjects.Contains($(MSBuildProjectName)))</IsStableProject>
+    <IsStableProject Condition="'$(MSBuildProjectName.Contains(.symbols))' == 'true'">$(StableProjects.Contains($(MSBuildProjectName.Substring(0, $(MSBuildProjectName.IndexOf(.symbols))))))</IsStableProject>
+    <MajorVersion>0</MajorVersion>
```
eerhardt: It may make more sense to split these properties into their own PropertyGroup. Then it will be easier to read which are "stable" and which are "preview". Along with that, I think `UseStableVersionForNativeAssets` should just be rolled into the `IsStableProject` condition. So something like this:

```xml
<Project>
  <PropertyGroup>
    <StableProjects>
      Microsoft.ML;
      Microsoft.ML.Core;
      Microsoft.ML.CpuMath;
      Microsoft.ML.Data;
      Microsoft.ML.FastTree;
      Microsoft.ML.ImageAnalytics;
      Microsoft.ML.KMeansClustering;
      Microsoft.ML.LightGbm;
      Microsoft.ML.PCA;
      Microsoft.ML.StandardTrainers;
      Microsoft.ML.Transforms;
      Microsoft.ML.Mkl.Components;
      Microsoft.ML.Mkl.Redist;
    </StableProjects>
    <IsStableProject Condition="'$(UseStableVersionForNativeAssets)' == 'true'">true</IsStableProject>
    <IsStableProject Condition="'$(IsStableProject)' == '' AND '$(MSBuildProjectName.Contains(.symbols))' == 'false'">$(StableProjects.Contains($(MSBuildProjectName)))</IsStableProject>
    <IsStableProject Condition="'$(IsStableProject)' == '' AND '$(MSBuildProjectName.Contains(.symbols))' == 'true'">$(StableProjects.Contains($(MSBuildProjectName.Substring(0, $(MSBuildProjectName.IndexOf(.symbols))))))</IsStableProject>
  </PropertyGroup>

  <PropertyGroup Condition="'$(IsStableProject)' == 'true'
    <MajorVersion>0</MajorVersion>
    <MinorVersion>12</MinorVersion>
    <PreReleaseLabel>preview</PreReleaseLabel>
  </PropertyGroup>
  <PropertyGroup Condition="'$(IsStableProject)' != 'true'">
    <MajorVersion>0</MajorVersion>
    <MinorVersion>12</MinorVersion>
    <PreReleaseLabel>preview</PreReleaseLabel>
  </PropertyGroup>
</Project>
```

src/Native/build.proj
eerhardt: Ah, OK good. Looks like I was mis-remembering how this was working. The native version files get generate during this proj's build:

https://github.com/dotnet/machinelearning/blob/df2f3fb49d4299925325921ee31c2cdff78ba3b5/src/Native/build.proj#L40

and

https://github.com/dotnet/machinelearning/blob/df2f3fb49d4299925325921ee31c2cdff78ba3b5/src/Native/build.proj#L57

This works out great then.

maryamariyan:
@dotnet-bot test MachineLearning-CI

maryamariyan:
@dotnet-bot test this

maryamariyan:
@dotnet-bot test MachineLearning-CodeCoverage

Issue #2955, opened 2019-03-14
Update tests for OVA


rogancarr:
Fixes for OVA went in independently of updates to the tests. The tests need to be updated so that master builds.

PR #2956, opened 2019-03-14
Updating OVA tests

rogancarr:
Fix for OVA #2949  went in without updates to tests. This PR updates the tests to work with the new OVA API.

Fixes: #2955


rogancarr:
This PR updates the OVA test to only test the correct (and working) path.

Issue #838, opened 2018-09-05
Make TextFeaturizer's extractors etc. configurable again


Zruty0:
#801 is making the parameters default and hardcoded for the following options of Text featurizer:
- Stop word remover (defaults to none) - Fixed in #2962 
- Custom term dictionary (defaults to none) - This parameter is hidden
- Word feature extractor (defaults to unigrams) - Fixed in #2911 
- Char feature extractor (defaults to 3-char) - Fixed in #2911 

Once individual building blocks become estimators, we should bring these parameters back (in a form of estimator for word/char extractor etc.). 

Or maybe we shouldn't, and instead just demonstrate how to compose your version of text transform from the individual building blocks?

eerhardt:
Is this strictly adding new API? Can this be done without a public API breaking change? If so, I think we can remove it from Project 13, and it can be added after v1.0.

But if this requires a public API breaking change, then it can be left in Project 13.

PR #2962, opened 2019-03-14
Made 'StopWordsRemover' in TextFeaturizer configurable again.

zeahmed:
This PR addresses the StopWordsRemover issue in https://github.com/dotnet/machinelearning/issues/838

This PR make available two options to remove stop words.
1. PredefinedStopWordsRemover: removes stops words using built in technique in ML.NET.
2. UseCustomStopWordsRemover: User provides the list of stop words to remove. Currently `string[]` is exposed to collect that list.

No support is added to load stop words from file. I think if we make No.2 take list as `IEnumerable<string>` instead of `string[]` then user can load list on their own. Let me know if reviewers have any thought.


src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
```diff
@@ -96,8 +129,42 @@ public sealed class Options : TransformInputBase
             [Argument(ArgumentType.AtMostOnce, HelpText = "Dataset language or 'AutoDetect' to detect language per row.", ShortName = "lang", SortOrder = 3)]
             public Language Language = DefaultLanguage;
```
Ivanidzo4ka: >Language [](start = 28, length = 8)

it needs to be part of StopWordsRemovingTransformer

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
```diff
@@ -85,6 +86,38 @@ internal bool TryUnparse(StringBuilder sb)
             }
         }
 
+        /// <summary>
+        /// Defines the different type of stop words remover supported.
+        /// </summary>
+        public enum StopWordsRemoverType
```
Ivanidzo4ka: >StopWordsRemoverType [](start = 20, length = 20)

I think it would be much cleaner if we just introduce IStopWordRemover interface and inherit both stop word removers from it and make it part of public interface.

zeahmed: Interface make it difficult to find out different type of `StopWordsRemover` in the system at API level. When user will use this enum they immediately know that what type of `StopWordsRemover` are available and how to use them.

---
In reply to: [265808886](https://github.com/dotnet/machinelearning/pull/2962#discussion_r265808886) [](ancestors = 265808886)

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
zeahmed: I think `Language` parameter is used by a couple components in TextTransform e.g. it used by tokenizer also. I know currently we are using very naïve whitespace tokenizer which does not use it. However, in future if we develop a sophisticated tokenizer for different language then this parameter will be needed. What do you say?

---
In reply to: [265808271](https://github.com/dotnet/machinelearning/pull/2962#discussion_r265808271) [](ancestors = 265808271)

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
```diff
@@ -85,6 +86,38 @@ internal bool TryUnparse(StringBuilder sb)
...
+        {
+            /// <summary>
+            /// Type of stop words remover to use.
+            /// </summary>
+            public StopWordsRemoverType StopWordsRemover;
+
+            /// <summary>
+            /// List of stop words. It is used when <see cref="StopWordsRemover"/> is equal to <see cref="StopWordsRemoverType.UseCustomStopWordsRemover"/>.
+            /// </summary>
+            public string[] StopWords;
```
artidoro: >string [](start = 19, length = 6)

Why not `IEnumerable<string>` ?

artidoro:
Could you add a test with a custom list of stop words?

zeahmed:
I had opened an issue already for adding more tests for TextFeaturizer. The rationale behind this is test all the options in the text featurizer. So, your concern can be addressed in that issue.
[https://github.com/dotnet/machinelearning/issues/2967](https://github.com/dotnet/machinelearning/issues/2967)

---
In reply to: [473395783](https://github.com/dotnet/machinelearning/pull/2962#issuecomment-473395783) [](ancestors = 473395783)

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
zeahmed: I doubt that EntryPoint will break because of this. The actual interface is still using `string[]`
[https://github.com/dotnet/machinelearning/blob/91a8703c6e43078dc8516e6e7383a179f8ab5ea1/src/Microsoft.ML.Transforms/Text/StopWordsRemovingTransformer.cs#L643](https://github.com/dotnet/machinelearning/blob/91a8703c6e43078dc8516e6e7383a179f8ab5ea1/src/Microsoft.ML.Transforms/Text/StopWordsRemovingTransformer.cs#L643)

I have mentioned that in the comments in the my PR. I am waiting for the people to comment on it. So that I can decide what to do.

---
In reply to: [266099423](https://github.com/dotnet/machinelearning/pull/2962#discussion_r266099423) [](ancestors = 266099423)

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
zeahmed: I have done what you proposed for the interface thing.

---
In reply to: [265811781](https://github.com/dotnet/machinelearning/pull/2962#discussion_r265811781) [](ancestors = 265811781,265808886)

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
Ivanidzo4ka: I don't know what kind of future lays ahead of us, but for now, this option just look silly.

---
In reply to: [265812513](https://github.com/dotnet/machinelearning/pull/2962#discussion_r265812513) [](ancestors = 265812513,265808271)

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
```diff
@@ -85,6 +86,27 @@ internal bool TryUnparse(StringBuilder sb)
...
+
+        /// <summary>
+        /// Use stop words remover that can removes language-specific list of stop words (most common words) already defined in the system.
+        /// </summary>
+        public sealed class PredefinedStopWordsRemoverOptions : IStopWordsRemoverOptions { }
+
+        /// <summary>
+        /// Use stop words remover that can removes language-specific list of stop words (most common words) already defined in the system.
+        /// </summary>
+        public sealed class CustomStopWordsRemoverOptions : IStopWordsRemoverOptions
```
Ivanidzo4ka: >CustomStopWordsRemoverOptions [](start = 28, length = 29)

Why this interface and class is part of text transform and not part of StopWordRemover itself?

Ivanidzo4ka: My question is why it's not implemented same way as word ngram and char ngram implemented?

---
In reply to: [266565265](https://github.com/dotnet/machinelearning/pull/2962#discussion_r266565265) [](ancestors = 266565265)

zeahmed: It can be part of StopWordRemover. I will move it.
Word ngram and char ngram are do different options that can be exclusively applied in Text transform (two parameter for two different options). On the contrary,  only one type of StopWordRemover can be applied (one parameter to take two different options).  Any idea if it can be done better than this?

---
In reply to: [266565861](https://github.com/dotnet/machinelearning/pull/2962#discussion_r266565861) [](ancestors = 266565861,266565265)

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
zeahmed: hmm...so you think that it should not be here? it should instead be in `PredefinedStopWordsRemoverOptions`?

---
In reply to: [266564835](https://github.com/dotnet/machinelearning/pull/2962#discussion_r266564835) [](ancestors = 266564835,265812513,265808271)

zeahmed:
Thanks all for the useful feedback!

PR #2969, opened 2019-03-14
Configurable Threshold for binary models

### LINKED ISSUE #2465: Cannot set the threshold on a binary predictor

Ivanidzo4ka:
fixes #2465


src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -274,6 +275,40 @@ internal BinaryClassificationTrainers(BinaryClassificationCatalog catalog)
...
+                    transformers.Add(transform);
+            }
+
+            transformers.Add(new BinaryPredictionTransformer<TModel>(Environment, predictionTransformer.Model,
+                predictionTransformer.TrainSchema, predictionTransformer.FeatureColumn,
+                threshold, predictionTransformer.ThresholdColumn));
+            return new TransformerChain<BinaryPredictionTransformer<TModel>>(transformers.ToArray());
+        }
+
+        public BinaryPredictionTransformer<TModel> ChangeModelThreshold<TModel>(BinaryPredictionTransformer<TModel> model, float threshold)
```
Ivanidzo4ka: needs documentation

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -274,6 +275,40 @@ internal BinaryClassificationTrainers(BinaryClassificationCatalog catalog)
...
+             where TModel : class
+        {
+            if (model.Threshold == threshold)
```
sfilipi: >if (model.Threshold == threshold) [](start = 12, length = 33)

do you want to warn here?

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -274,6 +275,40 @@ internal BinaryClassificationTrainers(BinaryClassificationCatalog catalog)
...
+        /// <param name="chain">Chain of transformers.</param>
+        /// <param name="threshold">New threshold.</param>
+        /// <returns></returns>
+        public TransformerChain<BinaryPredictionTransformer<TModel>> ChangeModelThreshold<TModel>(TransformerChain<BinaryPredictionTransformer<TModel>> chain, float threshold)
+            where TModel : class
+        {
+            if (chain.LastTransformer.Threshold == threshold)
+                return chain;
+            List<ITransformer> transformers = new List<ITransformer>();
+            var predictionTransformer = chain.LastTransformer;
```
sfilipi: >chain.LastTransformer [](start = 40, length = 21)

I don't like the assumption that the predictor is the last one, it might not be. Minimally, we should check, but that is not safe either. 

IMO the only API existing for this should be the second one.

test/Microsoft.ML.Functional.Tests/Prediction.cs
```diff
@@ -2,14 +2,26 @@
...
 namespace Microsoft.ML.Functional.Tests
 {
-    public class PredictionScenarios
+    public class PredictionScenarios : BaseTestClass
     {
+        public PredictionScenarios(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        class Answer
```
sfilipi: >Answer [](start = 14, length = 6)

Prediction

test/Microsoft.ML.Functional.Tests/Prediction.cs
```diff
@@ -19,36 +31,57 @@ public class PredictionScenarios
...
-            // This is no longer possible in the API
-            //var newModel = new BinaryPredictionTransformer<IPredictorProducing<float>>(ml, model.Model, trainData.Schema, model.FeatureColumnName, threshold: 0.01f, thresholdColumn: DefaultColumnNames.Probability);
-            //var newScoredTest = newModel.Transform(pipeline.Transform(testData));
-            //var newMetrics = mlContext.BinaryClassification.Evaluate(scoredTest);
-            // And the Threshold and ThresholdColumn properties are not settable.
-            //var predictor = model.LastTransformer;
-            //predictor.Threshold = 0.01; // Not possible
+            var mlContext = new MLContext(seed: 1);
+
+            var data = mlContext.Data.LoadFromTextFile<TweetSentiment>(GetDataPath(TestDatasets.Sentiment.trainFilename),
```
wschin: Can we try not to load file everywhere? It will be faster to just use in-memory data.

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -274,6 +275,40 @@ internal BinaryClassificationTrainers(BinaryClassificationCatalog catalog)
...
         }
 
+        /// <summary>
+        /// Change threshold for binary model.
+        /// </summary>
+        /// <typeparam name="TModel">An implementation of the <see cref="IPredictorProducing{TResult}"/></typeparam>
+        /// <param name="chain">Chain of transformers.</param>
+        /// <param name="threshold">New threshold.</param>
+        /// <returns></returns>
+        public TransformerChain<BinaryPredictionTransformer<TModel>> ChangeModelThreshold<TModel>(TransformerChain<BinaryPredictionTransformer<TModel>> chain, float threshold)
```
wschin:
```suggestion
        public TransformerChain<BinaryPredictionTransformer<TModel>> ChangeDecisionThreshold<TModel>(TransformerChain<BinaryPredictionTransformer<TModel>> chain, float threshold)
```
Maybe?

wschin: I am not sure if this should be a new function. Could we add a parameter, `threshold`, to all binary trainers?

test/Microsoft.ML.Functional.Tests/Prediction.cs
rogancarr: We have standard test datasets saved to files that we use in tests.

src/Microsoft.ML.Data/TrainCatalog.cs
Ivanidzo4ka: Ok, we can add that as parameter to binary trainer. Question is if you train your model, how you gonna change threshold? Retrain model?
I think this method has right to live.

---
In reply to: [266062138](https://github.com/dotnet/machinelearning/pull/2969#discussion_r266062138) [](ancestors = 266062138)

wschin: Retrain looks fine to me. I really don't feel adding a helper function is a good idea. This is not a Transformer, so I expect it will become a orphan in the future. Like FFM, PFI and so on don't care about it because it's not a standard binary classifier.

---
In reply to: [266088129](https://github.com/dotnet/machinelearning/pull/2969#discussion_r266088129) [](ancestors = 266088129,266062138)

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -274,6 +275,40 @@ internal BinaryClassificationTrainers(BinaryClassificationCatalog catalog)
...
+            where TModel : class
+        {
+            if (chain.LastTransformer.Threshold == threshold)
+                return chain;
+            List<ITransformer> transformers = new List<ITransformer>();
+            var predictionTransformer = chain.LastTransformer;
+            foreach (var transform in chain)
+            {
+                if (transform != predictionTransformer)
```
TomFinley: >predictionTransformer [](start = 33, length = 21)

Can we change this just a little please? I would prefer that we just add all transforms except the last unconditionally, which would be a fairly easy thing to do.

src/Microsoft.ML.Data/TrainCatalog.cs
TomFinley: That's a good point @sfilipi. I think you're probably right about this.

---
In reply to: [266034490](https://github.com/dotnet/machinelearning/pull/2969#discussion_r266034490) [](ancestors = 266034490)

src/Microsoft.ML.Data/TrainCatalog.cs
TomFinley: I think we should provide the same warning that C# does when you have a variable like `int a = 5` and then assign `5` to it later.

---
In reply to: [265862991](https://github.com/dotnet/machinelearning/pull/2969#discussion_r265862991) [](ancestors = 265862991)

src/Microsoft.ML.Data/TrainCatalog.cs
TomFinley: > I am not sure if this should be a new function. Could we add a parameter, `threshold`, to all binary trainers? #Pending

Historically we have found that adding options to "all" trainers just invites inconsistency and is a nightmare from a maintainability perspective. For those reasons we no longer do that. So I strongly object to that.

TomFinley:
Helper function for `IEstimator<...>`?

src/Microsoft.ML.Data/TrainCatalog.cs
TomFinley: As I think about it more, there's something about this idea of getting `ITransformer` implementors from existing `ITransformer` implementors that I find very appealing. Not just for this (which is a worthy use of this idea), but many other scenarios as well.

So for example, certain regressor algorithms are parametric w.r.t. their labels (in fact, most are). But there's a problem with merely normalizing the label, because then the predicted label is according to that same scale. In `sklearn` you could accomplish this fairly easily via the [`inverse_transform` method on their equivalent of what we call a normalizer, the `StandardScalar`](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html#sklearn.preprocessing.StandardScaler). So imagine you could get from a `NormalizerTransformer` another `NormalizerTransformer` that provides the inverse offset and scaling for any affine normalization, and whatnot. That would be pretty nice, would it not be?

So far from discouraging this pattern, I think we should do more of it.

src/Microsoft.ML.Data/TrainCatalog.cs
Ivanidzo4ka: My idea was to provide helper function for user since work with transform chain is kinda painful, at least from my point.
But I can have one method.

---
In reply to: [268206172](https://github.com/dotnet/machinelearning/pull/2969#discussion_r268206172) [](ancestors = 268206172,266034490)

Ivanidzo4ka:
I'm a bit struggle to understand what should it do.
Can you help me?



---
In reply to: [475655779](https://github.com/dotnet/machinelearning/pull/2969#issuecomment-475655779) [](ancestors = 475655779)

TomFinley:
Well, we can add it later if you like. But the idea is, an extension method on top of `ITrainerEstimator` or `IEstimator`, that is producing one of these `BinaryPredictionTransformer<TModel>` things, and is just a thin wrapper (the implementation of this wrapped thing would be minimal, private, and really only a few lines long.) It would store the `ITrainerEstimator` is is wrapping, and pass through everything on it, *except* for the `Fit` method, which would get the fit method then call the method you've added to apply.

Just so you could add this thing to an estimator pipeline. But, if it's unclear to you, we can always add it later. It is not essential, it was just something we had talked about doing.

---
In reply to: [476308360](https://github.com/dotnet/machinelearning/pull/2969#issuecomment-476308360) [](ancestors = 476308360,475655779)

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -256,6 +256,14 @@ internal BinaryClassificationTrainers(BinaryClassificationCatalog catalog)
                 Evaluate(x.Scores, labelColumnName), x.Scores, x.Fold)).ToArray();
         }
 
+        public BinaryPredictionTransformer<TModel> ChangeModelThreshold<TModel>(BinaryPredictionTransformer<TModel> model, float threshold)
+             where TModel : class
+        {
+            if (model.Threshold == threshold)
+                return model;
+            return new BinaryPredictionTransformer<TModel>(Environment, model.Model, model.TrainSchema, model.FeatureColumnName, threshold, model.ThresholdColumn);
```
rogancarr: >model.ThresholdColumn [](start = 140, length = 21)

Technically, Issue #2465 was that we should be able to set the `Threshold` and `ThresholdColumn` properties of the `BinaryPredictionTransformer`. That said, I think that the actual use cases that we care about are changing the `Threshold`; the `ThresholdColumn` seems more important when we are creating a _new_ `BinaryPredictionTransformer`. I actually don't think we need to modify that property. I'll update the issue that just setting a `Threshold` would be nice.

(Note that in practice, we have a `Score` column and a `Probability` column; modulo floating point error, these are a 1:1 mapping, so we will get the same results no matter which one we threshold on. Setting a custom threshold column seems more like a backdoor for crazy things: e.g. creating a custom score based on a few models and / or heuristics and then modifying a `BinaryPredictionTransformer` to score that column instead.)

test/Microsoft.ML.Functional.Tests/Prediction.cs
```diff
@@ -2,14 +2,30 @@
...
 namespace Microsoft.ML.Functional.Tests
 {
-    public class PredictionScenarios
+    public class PredictionScenarios : BaseTestClass
     {
+        public PredictionScenarios(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        class Prediction
```
rogancarr: >class Prediction [](start = 8, length = 16)

Please add to `Datasets/CommonColumns.cs`, and call it `PredictionColumns`.

test/Microsoft.ML.Functional.Tests/Prediction.cs
```diff
@@ -19,36 +35,64 @@ public class PredictionScenarios
...
+                .Append(mlContext.BinaryClassification.Trainers.LogisticRegression(
+                    new LogisticRegressionBinaryTrainer.Options { NumberOfThreads = 1 }));
+
+            // Train the model.
+            var model = pipeline.Fit(data);
+            var engine = mlContext.Model.CreatePredictionEngine<TweetSentiment, Prediction>(model);
+            var pr = engine.Predict(new TweetSentiment() { SentimentText = "Good Bad job" });
+            // Score is 0.64 so predicted label is true.
+            Assert.True(pr.PredictedLabel);
+            Assert.True(pr.Score > 0);
```
rogancarr: > 0) [](start = 34, length = 3)

0.5?

test/Microsoft.ML.Functional.Tests/Prediction.cs
```diff
@@ -19,36 +35,64 @@ public class PredictionScenarios
...
+            var transformers = new List<ITransformer>();
+            foreach (var transform in model)
+            {
+                if (transform != model.LastTransformer)
+                    transformers.Add(transform);
+            }
+            transformers.Add(mlContext.BinaryClassification.ChangeModelThreshold(model.LastTransformer, 0.7f));
+            var newModel = new TransformerChain<BinaryPredictionTransformer<CalibratedModelParametersBase<LinearBinaryModelParameters, PlattCalibrator>>>(transformers.ToArray());
```
rogancarr: >new TransformerChain<BinaryPredictionTransformer<CalibratedModelParametersBase<LinearBinaryModelParameters, PlattCalibrator>>> [](start = 27, length = 126)

Would it be better to do an in-place change rather than making a whole new chain? The new `TransformerChain` that comes back after still has references to the previous objects anyways. The only new thing here is `BinaryPredictionTransformer`.

test/Microsoft.ML.Functional.Tests/Prediction.cs
```diff
@@ -19,36 +35,64 @@ public class PredictionScenarios
...
-                .Append(mlContext.Regression.Trainers.Ols());
-
-            var model = pipeline.Fit(split.TrainSet);
-
-            var scoredTest = model.Transform(split.TestSet);
-            var metrics = mlContext.Regression.Evaluate(scoredTest);
-
-            Common.AssertMetrics(metrics);
-
-            // Todo #2465: Allow the setting of threshold and thresholdColumn for scoring.
```
rogancarr: >Todo #2465: [](start = 15, length = 11)

Thank you!

test/Microsoft.ML.Functional.Tests/Prediction.cs
TomFinley: > Would it be better to do an in-place change rather than making a whole new chain?

No. We rely upon `ITransformer`s not being mutable objects in many, many places.

The reason why they must be immutable is one of the "practical corollaries" to the [IDataView design principles](https://github.com/dotnet/machinelearning/blob/master/docs/code/IDataViewDesignPrinciples.md). We rely on several places on the `ITransformer` implementors being consistent. (Not least inside the `ITransformer` implementations themselves!) We often form chains of `ITransformer`s (pipelines, in other words), where each transform is the result of fitting to the result of the last. From a practical, user facing perspective, if these were suddenly to become mutable objects in their behavior w.r.t. transformation, the basic assumption that underlies why that is reliable at all would be compromised. So if you have a chain of transformers `A`, `B`, `C`, if you can suddenly mutate `B`'s behavior, the assumptions under which `C` was fit no longer hold, in ways that are impossible to detect. (Transformers when fit often depend on the distributional behavior of their inputs.)

Now then, nothing prevents someone from forming another transform `B'` derived from `B`, taking the original `A` and `C`, and forming the chain `A`, `B'`, and `C`. But I think this is understood to be a generally more risky operation. We have, and I think users have, I think a strong assumption that if `ITransformer` (including chains) works once, it shall continue to work of they don't do anything to it, something that would break if we were to allow them to be mutable.

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -256,6 +256,14 @@ internal BinaryClassificationTrainers(BinaryClassificationCatalog catalog)
                 Evaluate(x.Scores, labelColumnName), x.Scores, x.Fold)).ToArray();
         }
 
+        public BinaryPredictionTransformer<TModel> ChangeModelThreshold<TModel>(BinaryPredictionTransformer<TModel> model, float threshold)
```
eerhardt: We should put XML comments on all public members.

Ivanidzo4ka: https://media.giphy.com/media/xT5LMzIK1AdZJ4cYW4/giphy.gif

Issue #2726, opened 2019-02-26
Text loader v.s in-memory data structure in API reference samples


wschin:
We often starts our trainer examples with text loader but recently I feel loading text into `IDataView` is not directly related to the actual training procedure. If we use
```csharp
/// <summary>
/// Example with one binary label and 10 feature values.
/// </summary>
public class BinaryLabelFloatFeatureVectorSample
{
    public bool Label;

    [VectorType(_simpleBinaryClassSampleFeatureLength)]
    public float[] Features;
}
```
as our in-memory example, we can create more [flexible examples](https://github.com/dotnet/machinelearning/blob/7cc208c36edec554b6353a3a268cfb5e49274d17/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/BinaryClassification/SDCASupportVectorMachine.cs#L12) like [scikit-learn ones (where data matrix is float matrix)](https://scikit-learn.org/stable/modules/svm.html#multi-class-classification) and make ML.NET's learning curve smoother (because users don't need to learn text loader, the loaded data, and trainer at the same time).

cc @shmoradims, @rogancarr, @sfilipi, @shauheen 

#2780 shows a scikit-learn-style example for ML.NET. It is

- Self-contained --- To understand it, user doesn't need to look another document or use Visual Studio to search for those used functions. Notice that we can't rely on Visual Studio because not everyone is using it (1st partry and 3rd party experiences should the same!).
- End-to-end to C# developers --- because it trains a model over a C# List and get the prediction back as a C# List (The two ends are not IDataView so user doesn't need to learn IDataView to play with that API).
- Independent to external packages --- We shouldn’t expect a user who needs doc knows SamplesUtils.
- Production-friendly --- Doing prediction with C# data structure is included. That's how a trained model will be used in production.

shmoradims:
I chatted with Wei-Sheng offline about this. I can summarize the cons and pros as follows:

* Pros of in-memory (Cons of Textloader): Simpler data pipeline and more focused on the actual trainer that's being documented. It doesn't involve downloading, TextLoader, featurization, etc., so there's less friction for users to understand the sample and adapt it to their scenario.
* Cons of in-memory (Pros of Textloader): More meaningful data (Adult UCI vs randomly generated data)

Wei-Sheng has strong preference for adopting in-memory data for our samples. I'm ambivalent. @rogancarr @sfilipi what do you think?

bartczernicki:
[...]

PR #2979, opened 2019-03-15
In-memory & self-contained sample template.

shmoradims:
Related to #2726 I created this in-memory and self-contained sample for FastTree. I'll use the final version from this PR as template for the following samples.

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/FastTree.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/FastTree.cs
@@ -20,18 +20,48 @@ namespace Microsoft.ML.Samples.Dynamic.Trainers.Regression
             var examples = GenerateRandomDataPoints(1000);
 
             // Convert the examples list to an IDataView object, which is consumable by ML.NET API.
-            var data = mlContext.Data.LoadFromEnumerable(examples);
+            var trainingData = mlContext.Data.LoadFromEnumerable(examples);
 
             // Define the trainer.
-            var pipeline = mlContext.BinaryClassification.Trainers.FastTree();
+            var pipeline = mlContext.Regression.Trainers.FastTree();
 
             // Train the model.
-            var model = pipeline.Fit(data);
+            var model = pipeline.Fit(trainingData);
+
+            // Create testing examples. Use different random seed to make it different from training data.
+            var testData = mlContext.Data.LoadFromEnumerable(GenerateRandomDataPoints(500, seed:123));
+
+            // Run the model on test data set.
+            var transformedTestData = model.Transform(testData);
+
+            // Convert IDataView object to a list.
+            var predictions = mlContext.Data.CreateEnumerable<Prediction>(transformedTestData, reuseRowObject: false).ToList();
+
+            // Look at 5 predictions
+            foreach (var p in predictions.Take(5))
+                Console.WriteLine($"Label: {p.Label:F3}, Prediction: {p.Score:F3}");
+
+            // Expected output:
+            //   Label: 0.985, Prediction: 0.938
+            //   Label: 0.155, Prediction: 0.131
+            //   Label: 0.515, Prediction: 0.517
+            //   Label: 0.566, Prediction: 0.519
+            //   Label: 0.096, Prediction: 0.089
+
+            // Evaluate the overall metrics
+            var metrics = mlContext.Regression.Evaluate(transformedTestData);
+            SamplesUtils.ConsoleUtils.PrintMetrics(metrics);
+
+            // Expected output:
+            //   Mean Absolute Error: 0.05
+            //   Mean Squared Error: 0.00
+            //   Root Mean Squared Error: 0.06
+            //   RSquared: 0.95
         }
 
-        private static IEnumerable<DataPoint> GenerateRandomDataPoints(int count)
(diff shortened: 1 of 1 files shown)


docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/FastTree.cs
```diff
@@ -45,11 +75,21 @@ private static IEnumerable<DataPoint> GenerateRandomDataPoints(int count)
...
         }
+
+        // Class used to capture predictions.
+        private class Prediction
+        {
+            // Original label.
+            public float Label { get; set; }
+            // Predicted score from the trainer.
+            public float Score { get; set; }
+        }
```
sfilipi: I like extending the DataPoint class, because that is effectively what happens, columns get added. not a bigge though.

docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/FastTree.cs
```diff
@@ -20,18 +20,48 @@ public static void Example()
...
+
+            // Expected output:
+            //   Mean Absolute Error: 0.05
+            //   Mean Squared Error: 0.00
+            //   Root Mean Squared Error: 0.06
+            //   RSquared: 0.95
         }
 
-        private static IEnumerable<DataPoint> GenerateRandomDataPoints(int count)
+        private static IEnumerable<DataPoint> GenerateRandomDataPoints(int count, int seed=0)
```
sfilipi: > private static IEnumerable<DataPoint> GenerateRandomDataPoints(int count, int seed=0) [](start = 7, length = 86)

i like this, but do you think it will be quick  to create them artificially for each task? Ranking and time series come to mind.

shmoradims: sometimes it's not easy, ranking being an example. For those, I'll keep the text-loader style.

so this template is mostly suitable for regression and binary classification.

---
In reply to: [266151216](https://github.com/dotnet/machinelearning/pull/2979#discussion_r266151216) [](ancestors = 266151216)

docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/FastTree.cs
shmoradims: There's cons and pros here. I think keeping input and output separate is easier to understand for users. 

---
In reply to: [266150723](https://github.com/dotnet/machinelearning/pull/2979#discussion_r266150723) [](ancestors = 266150723)

docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/FastTree.cs
zeahmed: Text data is also not easy to randomly generated.

---
In reply to: [266157156](https://github.com/dotnet/machinelearning/pull/2979#discussion_r266157156) [](ancestors = 266157156,266151216)

docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/FastTree.cs
```diff
@@ -45,11 +75,21 @@ private static IEnumerable<DataPoint> GenerateRandomDataPoints(int count)
             }
         }
 
+        // Example with label and 50 feature values. A data set is a collection of such examples.
         private class DataPoint
         {
             public float Label { get; set; }
             [VectorType(50)]
             public float[] Features { get; set; }
         }
```
zeahmed: I assume all the samples are going to use same DataPoints (to be consistent), right?

wschin: I guess so. We now have samples, examples, instances, which are kind of less precise than data point.

docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/FastTree.cs
wschin: I believe Zeeshan A meant meaningful text data but I don't think we need meaningful data to demonstrate the functionality of a module. The amount of data might be a problem to trainers, but to my knowledge, there is no trainer directly consuming strings.

docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/FastTree.cs
shmoradims: the label type and number of features might change depending on the scenario, but the outline is the same.

---
In reply to: [266160360](https://github.com/dotnet/machinelearning/pull/2979#discussion_r266160360) [](ancestors = 266160360)

docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/FastTree.cs
shmoradims: This sample is just a template. If in some scenarios it doesn't make sense, we can try text-loader instead.

---
In reply to: [266160094](https://github.com/dotnet/machinelearning/pull/2979#discussion_r266160094) [](ancestors = 266160094,266157156,266151216)

Issue #2980, opened 2019-03-15
ONNX outputs don't match model outputs after serialization


rogancarr:
When I save a model to ONNX, load, and apply it with `ApplyOnnxModel`, it adds a zero as a suffix to all columns, including the expected output. This includes input columns, so the resulting `IDataView` now has double the columns, plus the output. To top it off, it's not clear how to find the output because it's been renamed.

Example:
ML.NET model
  Input:  `Features`
  Output: `Score`
  Resulting Schema: `Features`, `Score`

After Onnx Serialization / Deserialization:
  Input: `Features`
  Output: `Features0`, `Score0`
  Resulting Schema: `Features`, `Features0`, `Score0`

I am not sure if this is by design, but it feels like a bug.

jignparm:
What does the model above produce as outputs  (try this viewer https://github.com/lutzroeder/netron/releases/tag/v2.9.4) ? The zeros at the end of the names seems odd -- unless those are the names of the model outputs.

The OnnxTransform will generate output **all** output columns by default, with  names, type and shape specified by the model, unless the user specifies explicitly the list of columns they are interested in.

rogancarr:
@jignparm Check out this PR: #2984 . I think the issue is the way we save or load the model.

jignparm:
@rogancarr , looks like the PR is merged.

For issues #2980, #2981 and #2982, the likely reason for the behaviour is the way the model is generated, as you mentioned above, using the function **mlContext.Model.ConvertToOnnx(model, data, file);**. 

The OnnxTransform should honor the type, shape and names as specified by the model. In case there's a discrepancy with the OnnxTransform, that can be tracked in a separate issue.

rogancarr:
@jignparm The PR for tests is merged, but I'm still investigating this particular behavior. I'll post here once I've updated. I do think that it's in the function you call out.

singlis:
The "0" is because we are preserving the input column to imitate the behavior of what we would have in ml.net where the original columns are preserved as part of the output. In order to preserve the input, we are passing the input through an identity transform. ONNX can't have multiple inputs/outputs of the same name, so ML.NET generates the new output name by appending a 0 to the name:
![image](https://user-images.githubusercontent.com/17055319/55597042-7f5e1700-5700-11e9-8cdf-5a7b9e9652d3.png)

[...]

PR #2984, opened 2019-03-15
Add functional tests for ONNX scenarios

### LINKED ISSUE #2963: Create functional tests for all ONNX scenarios
As laid out in #2498 , we need scenarios to cover the ONNX functionality we want fully supported in V1.

Scenarios:
- I can take an existing ONNX model and get predictions from it (as both final output and as input to downstream pipelines)
- P1: I can export ML.NET models to ONNX (limited to the existing internal functionality) (In Model Files section, but can be fleshed out a bit better with other ONNX tests)

rogancarr:
As laid out in #2498 , we need scenarios to cover the ONNX functionality we want fully supported in V1.

Scenarios:
- I can take an existing ONNX model and get predictions from it (as both final output and as input to downstream pipelines)
- P1: I can export ML.NET models to ONNX (limited to the existing internal functionality) (In Model Files section, but can be fleshed out a bit better with other ONNX tests)

Fixes #2963


rogancarr:
These tests roundtrip the main ONNX exportable classes in ML.NET, and show that it's possible to serialize, deserialize, and use them as scorers / transforms.

test/Microsoft.ML.Functional.Tests/ONNX.cs
```diff
@@ -0,0 +1,195 @@
...
+                Assert.Equal(originalPrediction.Score, onnxPrediction.Score[0], precision: 4); // Note the low-precision equality!
+            }
+        }
+
+        private class ScoreColumn
+        {
+            public float Score { get; set; }
+        }
+
+        private class OnnxScoreColumn
```
wschin: What's the difference between `OnnxScoreColumn` and `ClusteringScoreColumn`? Without specifying the dimension of `Score` field, I am not sure the code is safe. Maybe we can do
```csharp
[VectorType(dimension)]
public float[] Score { get; set; }
```

rogancarr: The `ScoreColumn` class will be replaced with a general one being added by the `ModelFiles` PR.

The `OnnxScoreColumn` is there to make it explicit that we are working around a ?bug? in our ONNX implementation.

The `ClusteringScoreColumn` I added specifically for clustering.

Having wrote that out, we can delete the last two and make a `ScoreArrayColumn` class in the general helper class files (in `ModelFiles`, but I'll pull it into this PR instead and rebase that one).

On the topic of `VectorType`, we don't need to specify the dimension. Specifying a dimension just guarantees that the vector will be the same length for each row. That has the downside of making the classes non-reusable, so for helper classes in tests, we usually don't specify this attribute.

---
In reply to: [266636611](https://github.com/dotnet/machinelearning/pull/2984#discussion_r266636611) [](ancestors = 266636611)

test/Microsoft.ML.Functional.Tests/ONNX.cs
```diff
@@ -0,0 +1,195 @@
...
+
+            // Get the dataset.
+            var data = mlContext.Data.LoadFromTextFile<HousingRegression>(GetDataPath(TestDatasets.housing.trainFilename), hasHeader: true);
+
+            // Create a pipeline to train on the housing data.
+            var pipeline = mlContext.Transforms.Concatenate("Features", HousingRegression.Features)
+                .Append(mlContext.Transforms.Normalize("Features"))
+                .AppendCacheCheckpoint(mlContext)
+                .Append(mlContext.Regression.Trainers.FastTree(
+                    new FastTreeRegressionTrainer.Options { NumberOfThreads = 1, NumberOfTrees = 10 }));
```
shmoradims: >= 1 [](start = 76, length = 3)

why threads equal to 1? is there a know issue with multi-threading?

test/Microsoft.ML.Functional.Tests/ONNX.cs
```diff
@@ -0,0 +1,195 @@
...
+                mlContext.Model.ConvertToOnnx(model, data, file);
+
+            // Load the model as a transform.
+            var onnxEstimator = mlContext.Transforms.ApplyOnnxModel(modelPath);
+            var onnxModel = onnxEstimator.Fit(data);
+
+            // TODO #2980: ONNX outputs don't match the outputs of the model, so we must hand-correct this for now.
+            // TODO #2981: ONNX models cannot be fit as part of a pipeline, so we must use a workaround like this.
+            var onnxWorkaroundPipeline = onnxModel.Append(
+                mlContext.Transforms.CopyColumns("Score", "Score0").Fit(onnxModel.Transform(data)));
```
zeahmed: >Score0 [](start = 59, length = 6)

Just a question: where does `Score0` come from? is it produced by onnx transform?

rogancarr: I am not sure where the '0' suffices are entering the pipeline yet. I am digging into it as part of this bug: #2980.

---
In reply to: [267067407](https://github.com/dotnet/machinelearning/pull/2984#discussion_r267067407) [](ancestors = 267067407)

test/Microsoft.ML.Functional.Tests/ONNX.cs
rogancarr: There are no issues with multithreading. This is for convenience. Namely, we run these sorts of tests single-threaded because
- Many are running at the same time;
- Some algorithms are non-deterministic when multitheaded (e.g. SDCA) and make testing harder (boxing vs exact).

---
In reply to: [267028362](https://github.com/dotnet/machinelearning/pull/2984#discussion_r267028362) [](ancestors = 267028362)

PR #2966, opened 2019-03-14
Clean FeatureContributionCalculation and PermutationFeatureImportance

artidoro:
Fixes #2965.

In this PR:
1.  I move and rename the extension `MlContext.Model.Explainability.FeatureContributionCalculation` to `MlContext.Transforms.CalculateFeatureContribution`.

2. I also remove the `MlContext.Model.Explainability` catalog. 

3. Clean up the `FeatureContributionCalculation` extension

4. Clean up PFI extensions

PR #2990, opened 2019-03-18
Fix FeatureColumnName in the public API

### LINKED ISSUE #2975: FeatureColumn vs FeatureColumnName in Model Transformers.
In the public API, we've been switching `FeatureColumn` to `FeatureColumnName`. However, trained models like `SDCA` have `FeatureColumnName` in the input, but produce a `Transformer` that has a `FeatureColumn` property. I think this is widespread.

abgoswam:
Fixes #2975

- also fixes an incorrect parameter name `featureColumn` parameter in ExplanabilityCatalog.cs

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/ExplainabilityCatalog.cs
+++ b/src/Microsoft.ML.Data/Transforms/ExplainabilityCatalog.cs
@@ -18,7 +18,7 @@ namespace Microsoft.ML
         /// </summary>
         /// <param name="catalog">The model explainability operations catalog.</param>
         /// <param name="modelParameters">Trained model parameters that support Feature Contribution Calculation and which will be used for scoring.</param>
-        /// <param name="featureColumn">The name of the feature column that will be used as input.</param>
+        /// <param name="featureColumnName">The name of the feature column that will be used as input.</param>
         /// <param name="numPositiveContributions">The number of positive contributions to report, sorted from highest magnitude to lowest magnitude.
         /// Note that if there are fewer features with positive contributions than <paramref name="numPositiveContributions"/>, the rest will be returned as zeros.</param>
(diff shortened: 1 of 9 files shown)


src/Microsoft.ML.Data/Transforms/ExplainabilityCatalog.cs
```diff
@@ -18,18 +18,18 @@ public static class ExplainabilityCatalog
...
+        /// <param name="featureColumnName">The name of the feature column that will be used as input.</param>
         /// <param name="numPositiveContributions">The number of positive contributions to report, sorted from highest magnitude to lowest magnitude.
         /// Note that if there are fewer features with positive contributions than <paramref name="numPositiveContributions"/>, the rest will be returned as zeros.</param>
         /// <param name="numNegativeContributions">The number of negative contributions to report, sorted from highest magnitude to lowest magnitude.
         /// Note that if there are fewer features with negative contributions than <paramref name="numNegativeContributions"/>, the rest will be returned as zeros.</param>
         /// <param name="normalize">Whether the feature contributions should be normalized to the [-1, 1] interval.</param>
         public static FeatureContributionCalculatingEstimator FeatureContributionCalculation(this ModelOperationsCatalog.ExplainabilityTransforms catalog,
             ICalculateFeatureContribution modelParameters,
-            string featureColumn = DefaultColumnNames.Features,
```
rogancarr: >featureColumn =  [](start = 19, length = 16)

#2966 rewrites these functions, so you can drop changes to `ExplainabilityCatalog`.

artidoro: That would be much appreciated indeed :)

PR #2993, opened 2019-03-18
Include the save file action (.ZIP file) as part of model.SaveFile().

### LINKED ISSUE #1689: Include the "save file" action (.ZIP file) as part of model.SaveFile()
Currently, whenever we are saving a .ZIP model file, you always need to handle the code for the File stream class.

I think that is repetitive code that could be included in the API.
Instead of:

```
using (var fs = new FileStream(ModelPath, FileMode.Create, FileAccess.Write, FileShare.Write))
                mlContext.Model.Save(trainedModel, fs);
```

We could have an overridden method so I could simply do the following:

`mlContext.Model.Save(trainedModel, modelPath);`

Or even the following, if those methods were part of the model class itself instead of a utility class in the MLContext:

`model.Save(modelPath);`

Currently, this kind of line using the file stream class is something you need to repeat over and over in every/most training app (even when the constructor can be simplified):

`using (var fs = new FileStream(ModelPath, FileMode.Create, FileAccess.Write, FileShare.Write))`

Handling a FileStream object might should not be mandatory for the user **but optional**:

Of course, I would also maintain the current methods because in some cases you might want to just provide an existing stream object, but as opt-in.

codemzs:
fixes #1689

### DIFF
--- a/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
+++ b/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
@@ -50,7 +50,18 @@ namespace Microsoft.ML
         }
 
         /// <summary>
-        /// Save a transformer model and the loader used to create its input data to the stream.
+        /// Save the model to the file.
+        /// </summary>
+        /// <param name="model">The trained model to be saved.</param>
+        /// <param name="modelPath">Path where model should be saved.</param>
+        public void Save(ITransformer model, string modelPath)
+        {
+            using (var stream = File.Create(modelPath))
+                Save(model, stream);
+        }
+
+        /// <summary>
+        /// Load the model from the stream.
         /// </summary>
         /// <param name="loader">The loader that was used to create data to train the model</param>
         /// <param name="model">The trained model to be saved</param>
--- a/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
+++ b/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
@@ -53,11 +53,23 @@ namespace Microsoft.ML
         /// Save the model to the file.
         /// </summary>
         /// <param name="model">The trained model to be saved.</param>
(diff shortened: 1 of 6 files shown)


src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -33,6 +33,17 @@ internal ModelOperationsCatalog(IHostEnvironment env)
         /// <param name="stream">A writeable, seekable stream to save to.</param>
         public void Save(ITransformer model, Stream stream) => model.SaveTo(_env, stream);
 
+        /// <summary>
+        /// Save the model to the file.
+        /// </summary>
+        /// <param name="model">The trained model to be saved.</param>
+        /// <param name="modelPath">Path where model should be saved.</param>
```
eerhardt: (nit) - I wonder if `modelPath` should be changed to `filePath` instead. `modelPath` sounds like it is the path where the model is already.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -50,14 +50,49 @@ public void Save<TSource>(IDataLoader<TSource> model, Stream stream)
...
+        /// Load the model from the stream.
         /// </summary>
         /// <param name="loader">The loader that was used to create data to train the model</param>
         /// <param name="model">The trained model to be saved</param>
         /// <param name="stream">A writeable, seekable stream to save to.</param>
         public void Save<TSource>(IDataLoader<TSource> loader, ITransformer model, Stream stream) =>
             Save(new CompositeDataLoader<TSource, ITransformer>(loader, new TransformerChain<ITransformer>(model)), stream);
 
+        /// <summary>
+        /// Save the model to the file.
```
codemzs: >Save the model to the file. [](start = 12, length = 27)

Save a transformer model and the loader used to create its input data to the stream.

PR #2995, opened 2019-03-18
Clean up the SchemaDefinition class

### LINKED ISSUE #2978: Public API for SchemaDefinition class
There are no samples/tests in the code that make use of this class, even though it is part of the public API.

yaeldekel:
Fixes #2978 .

### DIFF
--- a/src/Microsoft.ML.Data/Data/SchemaDefinition.cs
+++ b/src/Microsoft.ML.Data/Data/SchemaDefinition.cs
@@ -203,15 +203,14 @@ namespace Microsoft.ML.Data
         /// </summary>
         public sealed class Column
         {
-            private readonly Dictionary<string, AnnotationInfo> _annotations;
-            internal Dictionary<string, AnnotationInfo> Annotations { get { return _annotations; } }
+            internal Dictionary<string, AnnotationInfo> AnnotationInfos { get; }
 
             /// <summary>
             /// The name of the member the column is taken from. The API
             /// requires this to not be null, and a valid name of a member of
             /// the type for which we are creating a schema.
             /// </summary>
-            public string MemberName { get; set; }
+            public string MemberName { get; }
             /// <summary>
             /// The name of the column that's created in the data view. If this
             /// is null, the API uses the <see cref="MemberName"/>.
@@ -224,33 +223,20 @@ namespace Microsoft.ML.Data
             public DataViewType ColumnType { get; set; }
 
             /// <summary>
-            /// Whether the column is a computed type.
-            /// </summary>
-            public bool IsComputed { get { return Generator != null; } }
-
-            /// <summary>
             /// The generator function. if the column is computed.
             /// </summary>
-            public Delegate Generator { get; set; }
+            internal Delegate Generator { get; set; }
 
-            public Type ReturnType => Generator?.GetMethodInfo().GetParameters().LastOrDefault().ParameterType.GetElementType();
+            internal Type ReturnType => Generator?.GetMethodInfo().GetParameters().LastOrDefault().ParameterType.GetElementType();
 
-            public Column(IExceptionContext ectx, string memberName, DataViewType columnType,
-                string columnName = null, IEnumerable<AnnotationInfo> annotationInfos = null, Delegate generator = null)
+            public Column(string memberName, DataViewType columnType,
+                string columnName = null)
             {
-                ectx.CheckNonEmpty(memberName, nameof(memberName));
+                Contracts.CheckNonEmpty(memberName, nameof(memberName));
                 MemberName = memberName;
                 ColumnName = columnName ?? memberName;
                 ColumnType = columnType;
-                Generator = generator;
-                _annotations = annotationInfos != null ?
-                    annotationInfos.ToDictionary(m => m.Kind, m => m)
-                    : new Dictionary<string, AnnotationInfo>();
-            }
-
-            public Column()
-            {
-                _annotations = _annotations ?? new Dictionary<string, AnnotationInfo>();
+                AnnotationInfos = new Dictionary<string, AnnotationInfo>();
             }
 
             /// <summary>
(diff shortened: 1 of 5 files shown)


src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -224,34 +223,21 @@ public sealed class Column
...
             /// </summary>
-            public Delegate Generator { get; set; }
+            internal Delegate Generator { get; set; }
 
-            public Type ReturnType => Generator?.GetMethodInfo().GetParameters().LastOrDefault().ParameterType.GetElementType();
+            internal Type ReturnType => Generator?.GetMethodInfo().GetParameters().LastOrDefault().ParameterType.GetElementType();
 
-            public Column(IExceptionContext ectx, string memberName, DataViewType columnType,
-                string columnName = null, IEnumerable<AnnotationInfo> annotationInfos = null, Delegate generator = null)
+            public Column(string memberName, DataViewType columnType,
```
yaeldekel: >public [](start = 12, length = 6)

I would like to make this internal as well. Is there any scenario where users would want to add new columns to the `SchemaDefinition` generated from the type?

test/Microsoft.ML.Functional.Tests/SchemaDefinitionTests.cs
```diff
@@ -0,0 +1,120 @@
...
+using Microsoft.Data.DataView;
+using Microsoft.ML.Data;
+using Microsoft.ML.RunTests;
+using Microsoft.ML.TestFramework;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.ML.Functional.Tests
+{
+    public class PredictionEngineScenarios : BaseTestClass
```
eerhardt: (nit) I would name the file and the class name the same.

test/Microsoft.ML.Functional.Tests/SchemaDefinitionTests.cs
```diff
@@ -0,0 +1,120 @@
...
+        public PredictionEngineScenarios(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        protected override void Initialize()
+        {
+            base.Initialize();
+
+            _ml = new MLContext(42);
+            _ml.AddStandardComponents();
```
eerhardt: Do you need to call `AddStandardComponents`? That should only be necessary when you are doing things like using the MAML syntax. When you are strictly using the API, it shouldn't be necessary.

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
TomFinley: I think so. In fact I think that's one of the major scenarios, isn't it? You start with an empty schema definition, then you add columns to it to describe how you want fields mapped to columns, etc.? This object exists in large part for those situations where you don't or somehow can't rely purely on the reflection based mechanism.

---
In reply to: [266437352](https://github.com/dotnet/machinelearning/pull/2995#discussion_r266437352) [](ancestors = 266437352)

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -262,38 +248,37 @@ public Column()
...
-                    return Annotations.Select(x => new KeyValuePair<string, DataViewType>(x.Key, x.Value.AnnotationType));
+                    var builder = new DataViewSchema.Annotations.Builder();
+                    foreach (var kvp in AnnotationInfos)
+                        builder.Add(kvp.Key, kvp.Value.AnnotationType, kvp.Value.GetGetterDelegate());
+                    return builder.ToAnnotations();
                 }
             }
         }
 
+        private SchemaDefinition()
```
TomFinley: >SchemaDefinition [](start = 16, length = 16)

So, why is this private? I'm thinking about how I'd like to use it. I have my class, I create a new schema definition (but empty), then I populate the mapping. Do I have any other way to create an empty one of these guys?

TomFinley:
>         public static SchemaDefinition Create(Type userType, Direction direction = Direction.Both)

Is this actually the only way to create a schema definition? As far as I see this auto-populates everything, which is sort of the opposite of what someone actually trying to use this structure would want to do, most of the time. (Since the entire reason someone wants to create this thing is to be explicit about the dataview-type mapping.)

---
Refers to: src/Microsoft.ML.Data/Data/SchemaDefinition.cs:326 in de633a3. [](commit_id = de633a3d121aabe8f5ecc87e45f6838a3b0f6c17, deletion_comment = False)

TomFinley:
>         public static SchemaDefinition Create(Type userType, Direction direction = Direction.Both)

Maybe if you hadn't made the constructor private that would be sufficient.

---
In reply to: [474069107](https://github.com/dotnet/machinelearning/pull/2995#issuecomment-474069107) [](ancestors = 474069107)

---
Refers to: src/Microsoft.ML.Data/Data/SchemaDefinition.cs:326 in de633a3. [](commit_id = de633a3d121aabe8f5ecc87e45f6838a3b0f6c17, deletion_comment = False)

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
yaeldekel: I was thinking you start with a schema that's auto-generated from the type, then modify the columns that are already in it. The schema generated from the type contains all the possible columns, why would we need to add new columns (as opposed to modifying existing columns)?

---
In reply to: [266605420](https://github.com/dotnet/machinelearning/pull/2995#discussion_r266605420) [](ancestors = 266605420,266437352)

test/Microsoft.ML.Functional.Tests/SchemaDefinitionTests.cs
yaeldekel: I added this line back, because of issue #2996.

---
In reply to: [266474155](https://github.com/dotnet/machinelearning/pull/2995#discussion_r266474155) [](ancestors = 266474155)

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
eerhardt: It appears that not only is there no way to create an empty one, but there is no longer a way to add a new column to it, since the column constructor is now internal.

Are we sure there are no scenarios that need to do this?

PR #3008, opened 2019-03-19
MLContext.Model.Load overloads that take file path instead of stream.

codemzs:
fixes #2991

### DIFF
--- a/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
+++ b/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
@@ -200,7 +200,7 @@ namespace Microsoft.ML
         /// <param name="stream">A readable, seekable stream to load from.</param>
         /// <returns>A model of type <see cref="CompositeDataLoader{IMultiStreamSource, ITransformer}"/> containing the loader
         /// and the transformer chain.</returns>
-        public IDataLoader<IMultiStreamSource> Load(Stream stream)
+        public IDataLoader<IMultiStreamSource> LoadDataLoader(Stream stream)
         {
             _env.CheckValue(stream, nameof(stream));
 
@@ -224,10 +224,10 @@ namespace Microsoft.ML
         /// <param name="modelPath">Path to model.</param>
         /// <returns>A model of type <see cref="CompositeDataLoader{IMultiStreamSource, ITransformer}"/> containing the loader
         /// and the transformer chain.</returns>
-        public IDataLoader<IMultiStreamSource> Load(string modelPath)
+        public IDataLoader<IMultiStreamSource> LoadDataLoader(string modelPath)
         {
             using (var stream = File.OpenRead(modelPath))
-                return Load(stream);
+                return LoadDataLoader(stream);
         }
 
         /// <summary>
@@ -240,7 +240,7 @@ namespace Microsoft.ML
         {
             _env.CheckValue(stream, nameof(stream));
 
-            loader = Load(stream);
+            loader = LoadDataLoader(stream);
             if (loader is CompositeDataLoader<IMultiStreamSource, ITransformer> composite)
             {
                 loader = composite.Loader;
--- a/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
+++ b/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
@@ -182,7 +182,20 @@ namespace Microsoft.ML
         }
 
         /// <summary>
-        /// Load the model and its input schema from the stream.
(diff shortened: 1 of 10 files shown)


src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -209,6 +222,18 @@ public IDataLoader<IMultiStreamSource> Load(Stream stream)
             }
         }
 
+        /// <summary>
+        /// Load the model from the file.
+        /// </summary>
+        /// <param name="modelPath">Path to model.</param>
+        /// <returns>A model of type <see cref="CompositeDataLoader{IMultiStreamSource, ITransformer}"/> containing the loader
+        /// and the transformer chain.</returns>
+        public IDataLoader<IMultiStreamSource> Load(string modelPath)
```
eerhardt: This API seems wrong to me.

‘mlContext.Model.Load(“path”)’ returning an ‘IDataLoader’ doesn’t seem right.

Can/should it be renamed to ‘LoadDataLoader’ or similar?

TomFinley: Yes, probably -- overloads that return fundamentally different objects strike me as confusing.

This would affect not just this added method but the one above, just to be clear.

---
In reply to: [266724059](https://github.com/dotnet/machinelearning/pull/3008#discussion_r266724059) [](ancestors = 266724059)

codemzs: Done.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -182,12 +182,25 @@ public ITransformer Load(Stream stream, out DataViewSchema inputSchema)
...
+        }
+
+        /// <summary>
+        /// Load the model from the stream.
         /// </summary>
         /// <param name="stream">A readable, seekable stream to load from.</param>
         /// <returns>A model of type <see cref="CompositeDataLoader{IMultiStreamSource, ITransformer}"/> containing the loader
         /// and the transformer chain.</returns>
-        public IDataLoader<IMultiStreamSource> Load(Stream stream)
+        public IDataLoader<IMultiStreamSource> LoadDataLoader(Stream stream)
```
eerhardt: Should we make a similar change for the `Save` method?

```C#
        /// <summary>
        /// Save the model to the file.
        /// </summary>
        /// <param name="model">The trained model to be saved.</param>
        /// <param name="filePath">Path where model should be saved.</param>
        public void Save<TSource>(IDataLoader<TSource> model, string filePath)
        {
            using (var stream = File.Create(filePath))
                Save(model, stream);
        }
```

These are no longer symmetric.

codemzs: Yes but this is Load API PR. I have created another PR to rename this for Save API https://github.com/dotnet/machinelearning/pull/3018

codemzs: @eerhardt @TomFinley feels we shouldn't change for Save because in .Net we have ReadChar, ReadInt, ReadDouble, etc but Write() is just one loaded set of methods. Thoughts?

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -182,12 +182,25 @@ public ITransformer Load(Stream stream, out DataViewSchema inputSchema)
         }
 
         /// <summary>
-        /// Load the model and its input schema from the stream.
+        /// Load the model and its input schema from the file.
+        /// </summary>
+        /// <param name="modelPath">Path to model.</param>
+        /// <param name="inputSchema">Will contain the input schema for the model. If the model was saved using older APIs
+        /// it may not contain an input schema, in this case <paramref name="inputSchema"/> will be null.</param>
```
singlis: Is it worth calling out the behavior for older models having a null input schema? That should never be the case now right?

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
singlis: I dont quite follow is there more discussion on github?

---
In reply to: [267100385](https://github.com/dotnet/machinelearning/pull/3008#discussion_r267100385) [](ancestors = 267100385)

TomFinley:
Similar question to that posed in #3019. Seems to be obviated by a chance in direction with these APIs now checked in, shall we close @codemzs?

PR #3011, opened 2019-03-19
Fix bug in TextLoader

### LINKED ISSUE #2996: Cannot create TextLoader without loading assemblies into ComponentCatalog
In the `TextLoader` constructor, it is allows to pass a null `Options`, or an `Options` object with a null `Columns` (see https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L1095 and https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L1111).
When this is the case, the constructor goes to `TryParseSchema` which uses the `ComponentCatalog` to load the schema, so we get a null reference exception here: https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L1293, if no assemblies are loaded.

yaeldekel:
Fixes #2996.


src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1291,11 +1291,6 @@ private sealed class LoaderHolder
                 if (loader == null || string.IsNullOrWhiteSpace(loader.Name))
                     goto LDone;
 
-                // Make sure the loader binds to us.
-                var info = host.ComponentCatalog.GetLoadableClassInfo<SignatureDataLoader>(loader.Name);
-                if (info.Type != typeof(ILegacyDataLoader) || info.ArgType != typeof(Options))
```
TomFinley: >info [](start = 20, length = 4)

Could we explain this change? As far as I can tell, we're considering not finding this loader in the component factory to be no longer an error condition. This may be right, but I'd feel a bit more comfortable with this change if we can explain why?

I did read the attached issue, but it was a little vague on precisely why removing this check is a right and desirable thing to do.

yaeldekel: It seems that this check was there in order to make sure that the schema defined in the file has the correct syntax for creating a legacy TextLoader. The schema defined in the file should always be in the format `TextLoader{col=... }`, since it is generated by the `TextSaver`. Instead of the check that I deleted, should we check that `loader.Name==TextLoader.LoaderSignature`?

---
In reply to: [267105243](https://github.com/dotnet/machinelearning/pull/3011#discussion_r267105243) [](ancestors = 267105243)

sfilipi: Since it is a text file, might be manually altered after saving it with the TextSaver?

---
In reply to: [267424076](https://github.com/dotnet/machinelearning/pull/3011#discussion_r267424076) [](ancestors = 267424076,267105243)

yaeldekel:
Closing this since other commits got included in it somehow. Will open a new one soon.

PR #3019, opened 2019-03-19
Rename Save API for DataLoader to SaveDataLoader.

codemzs:
towards #2991

### DIFF
--- a/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
+++ b/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
@@ -33,7 +33,7 @@ namespace Microsoft.ML
         /// </summary>
         /// <param name="model">The trained model to be saved.</param>
         /// <param name="stream">A writeable, seekable stream to save to.</param>
-        public void Save<TSource>(IDataLoader<TSource> model, Stream stream)
+        public void SaveDataLoader<TSource>(IDataLoader<TSource> model, Stream stream)
         {
             _env.CheckValue(model, nameof(model));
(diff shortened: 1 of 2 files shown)


TomFinley:
I don't necessarily agree with this change. As far as I can tell, this PR was the result of an off hand comment by @eerhardt [here](https://github.com/dotnet/machinelearning/pull/3008#pullrequestreview-216372871).

I think it is confusing to have methods that have different results and return values and whatnot to be overloads, that is, have the same name, which is why I [agreed with the other point here]().

Consider for an analogous situation, [BinaryReader](https://docs.microsoft.com/en-us/dotnet/api/system.io.binaryreader?view=netframework-4.7.2). Note all the `Read*` methods -- all differently named, all different return values. Now consider [BinaryWriter](https://docs.microsoft.com/en-us/dotnet/api/system.io.binarywriter.write?view=netframework-4.7.2). Basically one `Write` with different arguments. So the idea that there has to be a "master symmetry" . This loading and saving strikes me as rather similar.

It wouldn't be an absolute disaster if this PR was checked in of course, but I consider the original premise that led to this PR to be questionable.

TomFinley:
Incidentally I am removing this from public surface in #3044. Maybe we close this @codemzs, and if you would be so kind as to review that one?

TomFinley:
> I think this change is now obviated by #3044, since after discussions with @eerhardt and @yaeldekel as captured in #3025, we decided perhaps the more radical step of removing this entirely (as its presence is confusing, no matter what we name it)

So, what I mean is, should we close this @codemzs ?

PR #3022, opened 2019-03-19
Move KeyType, VectorType and VBuffer to ML.DataView

### LINKED ISSUE #2986: Co-locate VBuffer and VectorType with IDataView
Once we do #2974, to move `IDataView` into the `ML` namespace, we should move `VBuffer`, `VectorType` and `KeyType` into the same assembly/package as `IDataView`.

`VBuffer` and `KeyType` were left in `Microsoft.ML` because they were too specific to machine learning usages to be in `Microsoft.Data`. If we are moving `IDataView` back into the `Microsoft.ML` namespace, then we should be able to locate these types in the same package.

cc @TomFinley @shauheen @glebuk

eerhardt:
Fix #2986 

Easiest to review commit-by-commit. Major changes:

1. Moved the types into the ML.DataView assembly.
2. Renamed KeyType and VectorType to have `DataViewType` suffixes.
3. VBufferEditor had an "internal-only" option that is now public: `maxValuesCapacity`. This was needed so VBufferUtils could operate effectively. If VBufferUtils needs it, chances are someone else will need it.
4. Removed VectorType's constructor that takes a `VectorType template`, and replaced it with a new constructor that takes an `ImmutableArray<int> dims`.


src/Microsoft.ML.DataView/KeyDataViewType.cs
```diff
@@ -3,55 +3,73 @@
...
-            Contracts.AssertValue(type);
-            if (count == 0 || type.ToMaxInt() < count)
-                throw Contracts.ExceptParam(nameof(count), "The cardinality of a {0} must not exceed {1}.MaxValue" +
-                    " and must be strictly positive but got {2}.", nameof(KeyType), type.Name, count);
+            if (!IsValidDataType(type))
+                throw new ArgumentException(
+                    $"Type is not valid, it must be {typeof(byte).FullName}, {typeof(ushort).FullName}, {typeof(uint).FullName}, or {typeof(ulong).FullName}.",
+                    nameof(type));
+            if (count == 0 || GetMaxInt(type) < count)
+                throw new ArgumentOutOfRangeException(
```
artidoro: >throw new ArgumentOutOfRangeException( [](start = 16, length = 38)

nit: you can use `Contracts.ExceptParam` since you moved it.

src/Microsoft.ML.DataView/KeyDataViewType.cs
```diff
@@ -3,55 +3,73 @@
...
-    /// For example, if a KeyType has range 0-5000, then it has a <see cref="Count"/> of 5001, but
+    /// For example, if a KeyDataViewType has range 0-5000, then it has a <see cref="Count"/> of 5001, but
     /// the representational values are 1-5001. The representation value zero is reserved
     /// to mean a missing value (similar to NaN).
     /// </summary>
-    public sealed class KeyType : PrimitiveDataViewType
+    public sealed class KeyDataViewType : PrimitiveDataViewType
     {
-        public KeyType(Type type, ulong count)
+        public KeyDataViewType(Type type, ulong count)
```
singlis: >KeyDataViewType [](start = 15, length = 15)

Hi @eric.erhardt  - There are a couple of things that I am not crazy about the name KeyDataViewType. First that it has the word Type in the name (I know, I know KeyType does too and I am not saying that name is better), but there are a lot of classes that do not have the name Type in it  (File, Socket, etc) where the assumption that the object is a Type is implied.

Also having DataView in the name associates that it is used with DataViews -- but this is where namespaces semi-manages that grouping and doesn't need to be added to the class name. I think it would be great to name it more along the lines of what it does -- maybe Identifier or KeyIdentifier? 

I could be missing a whole lot of context here -- like maybe there is thread where this name was agreed upon or maybe this came out of an api review. So if that's the case disregard all of this because that means I could have added feedback prior to the PR :)

src/Microsoft.ML.DataView/VBuffer.cs
```diff
@@ -331,5 +333,76 @@ internal VBufferEditor<T> GetEditor()
...
+        /// <see cref="VBuffer{T}"/> itself, due to a bug in the C++/CLI compiler. (DevDiv 1097919:
+        /// [C++/CLI] Nested generic types are not correctly imported from metadata). So, if we want to use
+        /// <see cref="VBuffer{T}"/> in C++/CLI projects, we cannot have a generic struct with a nested class
+        /// that has the outer struct type as a field.
+        /// </summary>
+        private static IEnumerable<KeyValuePair<int, T>> Items(T[] values, int[] indices, int length, int count, bool all)
+        {
+            Debug.Assert(0 <= count && count <= ArrayUtils.Size(values));
+            Debug.Assert(count <= length);
+            Debug.Assert(count == length || count <= ArrayUtils.Size(indices));
```
singlis: >Assert [](start = 18, length = 6)

should this also check the size of indices? Or can indices be an empty array (like when count == length)?

src/Microsoft.ML.DataView/VBufferEditor.cs
```diff
@@ -47,26 +51,16 @@ public static class VBufferEditor
             ref VBuffer<T> destination,
             int newLogicalLength,
             int? valuesCount = null,
+            int? maxValuesCapacity = null,
```
singlis: >maxValuesCapacity [](start = 17, length = 17)

maximumValuesCapacity?

src/Microsoft.ML.DataView/KeyDataViewType.cs
eerhardt: Good call. Fixing.

src/Microsoft.ML.DataView/VBufferEditor.cs
eerhardt: I don't think `max` has to be spelled out to `maximum`. There are plenty of APIs in the BCL that just use "max". Here are some examples from the Http APIs:

https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/ref/System.Net.Http.cs#L93-L96

```C#
        public int MaxAutomaticRedirections { get { throw null; } set { } }
        public int MaxConnectionsPerServer { get { throw null; } set { } }
        public long MaxRequestContentBufferSize { get { throw null; } set { } }
        public int MaxResponseHeadersLength { get { throw null; } set { } }
```

src/Microsoft.ML.DataView/VBuffer.cs
eerhardt: Indices can be an empty array - actually `null`. This is useful when it is a dense VBuffer (when count == length), because then you don't need to allocate a useless array.

src/Microsoft.ML.DataView/KeyDataViewType.cs
eerhardt: > First that it has the word Type in the name 

These things _are_ types. They represent what type a column in an IDataView is. They are metadata that is part of the schema class.

> having DataView in the name associates that it is used with DataViews

The thing we are differentiating from is explicitly `System.Type` - which is a .NET Type concept. DataViewTypes and .NET System.Types have a relationship, but they are not 1:1. Since `KeyType` and `VectorType` inherit from the abstract base `DataViewType` class, we decided in #2297 to put the `DataViewType` suffix on all these types. I just happened to miss these two.  (And come to think of it - I missed 

https://github.com/dotnet/machinelearning/blob/70ef7ecd43b031b481a4047ea361da5e2f360336/src/Microsoft.ML.ImageAnalytics/ImageType.cs#L12

as well.  I logged #3050 for that.

eerhardt:
Can I get a review on this?

TomFinley:
> Can I get a review on this?

Really sorry @eerhardt -- I've been looking forward to this PR for the past few days and I somehow totally missed that you'd already posted it. 😛

src/Microsoft.ML.DataView/Contracts.cs
```diff
@@ -0,0 +1,50 @@
...
+// See the LICENSE file in the project root for more information.
+
+// We want the conditional code in this file to always be available to
+// client assemblies that might be DEBUG versions. That is, if someone uses
+// the release build of this assembly to build a DEBUG version of their code,
+// we want Contracts.Assert to be fully functional for that client.
+
+using System;
+
+namespace Microsoft.ML.Internal.DataView
```
TomFinley: >DataView [](start = 32, length = 8)

Ah interesting. We can do this later, but if we're doing another "local contracts" we probably ought to add it to the analysis rules here.

[https://github.com/dotnet/machinelearning/blob/db4ecc0135baffc8930201a85fb3e9a101cdfa46/tools-local/Microsoft.ML.InternalCodeAnalyzer/ContractsCheckAnalyzer.cs#L201-L202](https://github.com/dotnet/machinelearning/blob/db4ecc0135baffc8930201a85fb3e9a101cdfa46/tools-local/Microsoft.ML.InternalCodeAnalyzer/ContractsCheckAnalyzer.cs#L201-L202)

But this can be done later... it is more important that we get the shape of the API right at this point.

src/Microsoft.ML.DataView/KeyDataViewType.cs
```diff
@@ -3,55 +3,73 @@
...
     /// class numbers, etc. For example, in multi-class classification, the label is typically
-    /// a class number which is naturally a KeyType.
+    /// a class number which is naturally a KeyDataViewType.
     ///
-    /// KeyTypes have a cardinality (i.e., <see cref="Count"/>) that is strictly positive.
+    /// KeyDataViewTypes have a cardinality (i.e., <see cref="Count"/>) that is strictly positive.
     ///
     /// Note that the underlying representation value does not necessarily match the logical value.
-    /// For example, if a KeyType has range 0-5000, then it has a <see cref="Count"/> of 5001, but
+    /// For example, if a KeyDataViewType has range 0-5000, then it has a <see cref="Count"/> of 5001, but
```
TomFinley: >KeyDataViewType [](start = 26, length = 15)

Eh... once we freeze for 1.0 we really ought to work on the documentation. These things really ought to be `<see`, but not the priority right now. Just registering the opinion somewhere.

src/Microsoft.ML.DataView/VectorType.cs
```diff
@@ -57,33 +57,20 @@ public VectorType(PrimitiveDataViewType itemType, params int[] dimensions)
...
+        /// <param name="dimensions">The dimensions. Note that, like <see cref="Dimensions"/>, must be non-empty, with all
+        /// non-negative values. Also, because <see cref="Size"/> is the product of <see cref="Dimensions"/>, the result of
+        /// multiplying all these values together must not overflow <see cref="int"/>.</param>
+        public VectorType(PrimitiveDataViewType itemType, ImmutableArray<int> dimensions)
             : base(GetRawType(itemType))
         {
-            Contracts.CheckValue(template, nameof(template));
-            Contracts.CheckParam(Utils.Size(dims) > 0, nameof(dims));
-            Contracts.CheckParam(dims.All(d => d >= 0), nameof(dims));
+            Contracts.CheckParam(dimensions.Length > 0, nameof(dimensions));
```
TomFinley: >dimensions.Length > 0, nameof(dimensions)); [](start = 33, length = 43)

In an ideal world this check would produce a useful error message, but, I see we don't do it above anyway, so...

src/Microsoft.ML.DataView/KeyDataViewType.cs
```diff
@@ -3,55 +3,83 @@
...
+
+        private string GetRawTypeName()
+        {
+            if (RawType == typeof(byte))
+                return NumberDataViewType.Byte.ToString();
+            else if (RawType == typeof(ushort))
+                return NumberDataViewType.UInt16.ToString();
+            else if (RawType == typeof(uint))
+                return NumberDataViewType.UInt32.ToString();
+            else if (RawType == typeof(ulong))
```
TomFinley: This is one of these situations where since this specific one `uint` *is* going to be the type 99.9% of the time, we should probably structure the chained if-else statement so that this is the first condition.

eerhardt: oh, right. I knew this, but forgot when I did this. Thanks for the catch.

src/Microsoft.ML.DataView/Contracts.cs
eerhardt: Let me see if I can add it quick.

src/Microsoft.ML.DataView/VectorType.cs
TomFinley: I don't know. I'd rather not, since I'm not sure what would go in the issue. To identify this specific line would be too petty... but then what should we write in the issue, "many of our usage of `Check` and `CheckParam` have no useful guiding error messages describing what went wrong?" I mean, that's *true*, but I feel like that's not quite capturing the scale of the work, or something.

PR #3026, opened 2019-03-19
Add cancellation checkpoints in SDCA.

### LINKED ISSUE #3024: Cancellation checkpoint in StochasticDualCoordinateAscent
**Goal:**
Implement a way to a cancel SDCA training during runtime without impacting the performance too much.

**Solution proposal:**

Below is the performance graph of SDCA run from [this ](https://github.com/dotnet/machinelearning/blob/c38f81b3957fed6aa88ea0e6b295522d5bf3f9ec/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/StochasticDualCoordinateAscent.cs#L9)example.

image

We see that **MoveNext() in TrainWithoutLock** seems to be a good place to check for cancellation signal as it seems to consume bulk of CPU cycles as evident from the red stripes in the graph.

The second checkout point would be **CheckConvergence** as evident from the below graph (look for red stripes):

image

CC: @TomFinley , @eerhardt , @shauheen , @glebuk

codemzs:
fixes #3024

### DIFF
--- a/src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs
+++ b/src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs
@@ -803,6 +803,7 @@ namespace Microsoft.ML.Trainers
                 Func<DataViewRowId, long> getIndexFromId = GetIndexFromIdGetter(idToIdx, biasReg.Length);
                 while (cursor.MoveNext())
                 {
+                    Host.CheckAlive();
                     long idx = getIndexFromId(cursor.Id);
                     VBuffer<float> features = cursor.Features;
                     var label = cursor.Label;
@@ -966,6 +967,7 @@ namespace Microsoft.ML.Trainers
                 // Iterates through data to compute loss function.
                 while (cursor.MoveNext())
                 {
+                    Host.CheckAlive();
                     var instanceWeight = GetInstanceWeight(cursor);
                     var features = cursor.Features;
                     var output = WDot(in features, in weights[0], biasTotal);
--- a/src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs
+++ b/src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs
@@ -378,6 +378,7 @@ namespace Microsoft.ML.Trainers
                 pch.SetHeader(new ProgressHeader("examples"), e => e.SetProgress(0, count));
                 while (cursor.MoveNext())
                 {
+                    Host.CheckAlive();
                     DataViewRowId id = cursor.Id;
                     if (id.High > 0 || id.Low >= (ulong)maxTrainingExamples)
                     {
@@ -568,6 +569,7 @@ namespace Microsoft.ML.Trainers
                     pch.SetHeader(new ProgressHeader("examples"), e => e.SetProgress(0, row, count));
                     while (cursor.MoveNext())
                     {
+                        Host.CheckAlive();
                         long longIdx = getIndexFromIdAndRow(cursor.Id, row);
                         Contracts.Assert(0 <= longIdx & longIdx < invariants.Length, $"longIdx={longIdx}, invariants.Length={invariants.Length}");
                         int idx = (int)longIdx;
--- a/src/Microsoft.ML.StandardTrainers/Standard/SdcaMulticlass.cs
+++ b/src/Microsoft.ML.StandardTrainers/Standard/SdcaMulticlass.cs
@@ -151,6 +151,7 @@ namespace Microsoft.ML.Trainers
                 Func<DataViewRowId, long> getIndexFromId = GetIndexFromIdGetter(idToIdx, biasReg.Length);
                 while (cursor.MoveNext())
                 {
+                    Host.CheckAlive();
                     long idx = getIndexFromId(cursor.Id);
                     long dualIndexInitPos = idx * numClasses;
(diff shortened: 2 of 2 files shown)


src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs
```diff
@@ -802,6 +802,7 @@ private void InitializeConvergenceMetrics(out string[] names, out Double[] initi
                 Func<DataViewRowId, long> getIndexFromId = GetIndexFromIdGetter(idToIdx, biasReg.Length);
                 while (cursor.MoveNext())
                 {
+                    Host.CheckAlive();
```
singlis: This looks good - would it be worth adding a couple of tests that cancel the training?

codemzs: Yea we have unit-tests that test cancellation signal has been set correctly. For real-time testing it will require much more work. I still need to figure if that is something we do with a test or manually test it. All in all the current mechanism for cancellation is temporary, long term goal is to use cancellation token. I will highly encourage you read the issue associated with this PR and comment on my analysis.

src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs
```diff
@@ -965,6 +966,7 @@ private void InitializeConvergenceMetrics(out string[] names, out Double[] initi
                 // Iterates through data to compute loss function.
                 while (cursor.MoveNext())
                 {
+                    Host.CheckAlive();
```
wschin: Is it a cheap operation? We call this function on every data point.

codemzs: It’s  a two line operation but if you haven’t read the issue you should.

src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs
singlis: Thanks Zeeshan.

As for analysis - I think where you have the cancellation check is reasonable. Even looking through the SDCA code, this looks like the two places where we loop and you would want to check the cancellation condition. I did not see any additional places where this needs to be checked. 

As for testsing, I don't think the test would be much more work  - you can start training an SDCA pipeline, call cancel on the environment (CancelExecution I believe) and then check if the OperationCancelledExecption is thrown. Since Train is a blocking call, you would have to call it in a separate thread -- but I believe that is already expected since we don't have any TrainAsync APIs.

I do have a couple additional questions:
1) CheckAlive is ifdef'ed by a !CPUMATH_INFRASTRUCTURE -- however CPUMATH_INFRASTRUCTURE is defined in Microsoft.ML.CpuMath - how is CheckAlive enabled?
2) CheckAlive throws an OperationCancelledException - is ml.net supposed to catch that exception or is this up to the user to catch?

codemzs: Its not as simple, there are all sorts of timing issues. I'm manually testing this with a customer and once I validate this works in practice then I will check-in as it is. The user catches the exception.

sfilipi:
>                             count++;

is there one more necessary here?

---
Refers to: src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs:2016 in 95c978b. [](commit_id = 95c978b0158ae6204fd0d426ed074dc5a79d2548, deletion_comment = False)

sfilipi:
>                     long idx = getIndexFromId(cursor.Id);

is a check needed here too?

---
Refers to: src/Microsoft.ML.StandardTrainers/Standard/SdcaMulticlass.cs:153 in 95c978b. [](commit_id = 95c978b0158ae6204fd0d426ed074dc5a79d2548, deletion_comment = False)

src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs
wschin: There is no description about this function in your issue....

codemzs:
>                             count++;

This is SGDBinary not SDCABinary that we care about, so no.

---
In reply to: [475791046](https://github.com/dotnet/machinelearning/pull/3026#issuecomment-475791046) [](ancestors = 475791046)

---
Refers to: src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs:2016 in 95c978b. [](commit_id = 95c978b0158ae6204fd0d426ed074dc5a79d2548, deletion_comment = False)

codemzs:
>                     long idx = getIndexFromId(cursor.Id);

SdcaMultiClass is something we don't care about but since I have made the change below in this class so might as well keep it consistent here.

---
In reply to: [475791293](https://github.com/dotnet/machinelearning/pull/3026#issuecomment-475791293) [](ancestors = 475791293)

---
Refers to: src/Microsoft.ML.StandardTrainers/Standard/SdcaMulticlass.cs:153 in 95c978b. [](commit_id = 95c978b0158ae6204fd0d426ed074dc5a79d2548, deletion_comment = False)

wschin:
>                     {

CheckAlive?

---
Refers to: src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs:569 in 95c978b. [](commit_id = 95c978b0158ae6204fd0d426ed074dc5a79d2548, deletion_comment = False)

PR #3028, opened 2019-03-20
Add cancellation signal checkpoints in FastTree.

### LINKED ISSUE #3027: Cancellation checkpoints in FastTree
**Goal:**
Implement a way to a cancel FastTree training during runtime without impacting the performance too much.

**Solution proposal:**

Below are the performance graphs of FastTree run from [this ](https://github.com/dotnet/machinelearning/blob/c38f81b3957fed6aa88ea0e6b295522d5bf3f9ec/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/FastTree.cs#L12) example. The graph will have a function selected and red stripes indicate the position in the graph it is called.

The first checkpoint would be **`InitializeBins`** as evident from the below graph. We see before training there is the data prep step that is CPU intensive. In this step transpose of the dataset is created in-memory and binning of features is done. 

image

The second place would be somewhere in the Feature Flock creation. Here we think `CreateFlocksCore` would be ideal. Based on the below graph.

image

The third place would be at `FeatureHistogram` ctr called by `CreateSufficientStats`. This function is called by constructor of `LeastSquaresRegressionTreeLearner` within a loop that creates `CreateSufficientStats`. We will just place a checkpoint in this loop.

image

The last place would be somewhere we do the splitting of the nodes and partitioning of the data to construct the tree. We see that **`FindBestThresholdForFlockThreadWorker()`** seems to be a good place to check for cancellation signal as it seems to consume bulk of CPU cycles as evident from the red stripes in the graph. We can also place the checkpoints into the functions that it calls such as `FindBestThresholdFromHistogram`, `Sumup`, `SubtractCore`, `FillSplitCandidates`, `FindBestSplitForFeature` but I think place checks in the inner function could degrade the performance. 

image

CC: @TomFinley , @eerhardt , @shauheen , @glebuk

codemzs:
fixes #3027

Please read the issue before reviewing this PR.

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/FastTree.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/FastTree.cs
@@ -17,7 +17,7 @@ namespace Microsoft.ML.Samples.Dynamic.Trainers.Regression
             var mlContext = new MLContext(seed: 0);
 
             // Create a list of training examples.
-            var examples = GenerateRandomDataPoints(100000);
+            var examples = GenerateRandomDataPoints(1000);
 
             // Convert the examples list to an IDataView object, which is consumable by ML.NET API.
             var trainingData = mlContext.Data.LoadFromEnumerable(examples);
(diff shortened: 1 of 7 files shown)


TomFinley:
Hi @codemzs this looks pretty good I think. While doing this check during the histogram calcaulation is good, one area that can take a significant amount of time that is not covered at all by this change as far as I see is the dataset preparation. What do you think?

codemzs:
@TomFinley You are right. My analysis was done without disk transpose so I missed the transpose function but now I have added checkpoints there.

PR #3029, opened 2019-03-20
Polish standard trainers' catalog (Just rename some variables)

wschin:
To fix #2680, we rename several parameters and improve some doc strings.

PR #3030, opened 2019-03-20
Polish train catalog (renaming only)

wschin:
Related to #3029 (for StandardTrainersCatalog.cs) but for TrainCatalog.cs.

- Rename `topK` to `topPredictionCount`
- Rename `k` to `falsePositiveCount`

### DIFF
--- a/src/Microsoft.ML.Data/TrainCatalog.cs
+++ b/src/Microsoft.ML.Data/TrainCatalog.cs
@@ -484,22 +484,22 @@ namespace Microsoft.ML
         /// <param name="labelColumnName">The name of the label column in <paramref name="data"/>.</param>
         /// <param name="scoreColumnName">The name of the score column in <paramref name="data"/>.</param>
         /// <param name="predictedLabelColumnName">The name of the predicted label column in <paramref name="data"/>.</param>
-        /// <param name="topPredictionCount">If given a positive value, the <see cref="MulticlassClassificationMetrics.TopKAccuracy"/> will be filled with
+        /// <param name="topKPredictionCount">If given a positive value, the <see cref="MulticlassClassificationMetrics.TopKAccuracy"/> will be filled with
         /// the top-K accuracy, that is, the accuracy assuming we consider an example with the correct class within
         /// the top-K values as being stored "correctly."</param>
         /// <returns>The evaluation results for these calibrated outputs.</returns>
         public MulticlassClassificationMetrics Evaluate(IDataView data, string labelColumnName = DefaultColumnNames.Label, string scoreColumnName = DefaultColumnNames.Score,
-            string predictedLabelColumnName = DefaultColumnNames.PredictedLabel, int topPredictionCount = 0)
+            string predictedLabelColumnName = DefaultColumnNames.PredictedLabel, int topKPredictionCount = 0)
         {
             Environment.CheckValue(data, nameof(data));
             Environment.CheckNonEmpty(labelColumnName, nameof(labelColumnName));
             Environment.CheckNonEmpty(scoreColumnName, nameof(scoreColumnName));
             Environment.CheckNonEmpty(predictedLabelColumnName, nameof(predictedLabelColumnName));
-            Environment.CheckUserArg(topPredictionCount >= 0, nameof(topPredictionCount), "Must be non-negative");
+            Environment.CheckUserArg(topKPredictionCount >= 0, nameof(topKPredictionCount), "Must be non-negative");
 
             var args = new MulticlassClassificationEvaluator.Arguments() { };
-            if (topPredictionCount > 0)
-                args.OutputTopKAcc = topPredictionCount;
+            if (topKPredictionCount > 0)
+                args.OutputTopKAcc = topKPredictionCount;
             var eval = new MulticlassClassificationEvaluator(Environment, args);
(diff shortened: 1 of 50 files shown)


artidoro:
Did you check the RecommenderCatalog? I think it's on a separate file.

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -484,21 +484,21 @@ internal MulticlassClassificationTrainers(MulticlassClassificationCatalog catalo
         /// <param name="labelColumnName">The name of the label column in <paramref name="data"/>.</param>
         /// <param name="scoreColumnName">The name of the score column in <paramref name="data"/>.</param>
         /// <param name="predictedLabelColumnName">The name of the predicted label column in <paramref name="data"/>.</param>
-        /// <param name="topK">If given a positive value, the <see cref="MulticlassClassificationMetrics.TopKAccuracy"/> will be filled with
+        /// <param name="topPredictionCount">If given a positive value, the <see cref="MulticlassClassificationMetrics.TopKAccuracy"/> will be filled with
```
abgoswam: >If given a positive value [](start = 45, length = 25)

it seems the existing behavior is to just ignore negative values. is that correct ? should we raise an exception if user gives a negative value

rogancarr: Did you also rename `TopKAccuracy`? If we make this change, we won't have parallelism with the evaluation metrics.

wschin: We don't need to. `TopKAccuracy` is as good as `Accuracy`. This original `topK` is a parameter when computing `TopKAccuracy`, not `TopKAccuracy` itself.

wschin: No problem. We throw now.

wschin: Discussed offline. We will do `topKPredictionCount` to associate it with `TopKAccuracy`.

test/Microsoft.ML.Tests/AnomalyDetectionTests.cs
```diff
@@ -30,7 +30,7 @@ public void RandomizedPcaTrainerBaselineTest()
             var transformedData = DetectAnomalyInMnistOneClass(trainPath, testPath);
 
             // Evaluate
-            var metrics = ML.AnomalyDetection.Evaluate(transformedData, k: 5);
+            var metrics = ML.AnomalyDetection.Evaluate(transformedData, falsePositiveCount: 5);
```
Ivanidzo4ka: >falsePositiveCount [](start = 72, length = 18)

`AnomalyDetectionMetrics` has `DetectionRateAtKFalsePositives`
If you remove `K` from here, I think you need to remove `K` from metric class as well.

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -484,21 +484,22 @@ internal MulticlassClassificationTrainers(MulticlassClassificationCatalog catalo
...
         /// <param name="scoreColumnName">The name of the score column in <paramref name="data"/>.</param>
         /// <param name="predictedLabelColumnName">The name of the predicted label column in <paramref name="data"/>.</param>
-        /// <param name="topK">If given a positive value, the <see cref="MulticlassClassificationMetrics.TopKAccuracy"/> will be filled with
+        /// <param name="topKPredictionCount">If given a positive value, the <see cref="MulticlassClassificationMetrics.TopKAccuracy"/> will be filled with
         /// the top-K accuracy, that is, the accuracy assuming we consider an example with the correct class within
         /// the top-K values as being stored "correctly."</param>
         /// <returns>The evaluation results for these calibrated outputs.</returns>
         public MulticlassClassificationMetrics Evaluate(IDataView data, string labelColumnName = DefaultColumnNames.Label, string scoreColumnName = DefaultColumnNames.Score,
-            string predictedLabelColumnName = DefaultColumnNames.PredictedLabel, int topK = 0)
+            string predictedLabelColumnName = DefaultColumnNames.PredictedLabel, int topKPredictionCount = 0)
```
Ivanidzo4ka: >topKPredictionCount [](start = 85, length = 19)

Can we align it with `MulticlassClassificationMetrics` class?
```
       /// <summary>
        /// If positive, this is the top-K for which the <see cref="TopKAccuracy"/> is calculated.
        /// </summary>
        public int TopK { get; }
```

wschin: No problem.

---
In reply to: [267591599](https://github.com/dotnet/machinelearning/pull/3030#discussion_r267591599) [](ancestors = 267591599)

test/Microsoft.ML.Tests/AnomalyDetectionTests.cs
wschin: Let's do `DetectionRateAtFalsePositiveCount`.

---
In reply to: [267591549](https://github.com/dotnet/machinelearning/pull/3030#discussion_r267591549) [](ancestors = 267591549)

wschin:
Yes, it looks good.

---
In reply to: [474958115](https://github.com/dotnet/machinelearning/pull/3030#issuecomment-474958115) [](ancestors = 474958115)

PR #3032, opened 2019-03-20
Add cancellation checkpoint in logistic regression.

### LINKED ISSUE #3031: Cancellation checkpoints in LogisticRegression
**Goal:**
Implement a way to a cancel Logistic Regression training during runtime without impacting the performance too much.

**Solution proposal:**

Below is the performance graphs of Logistic Regression run from [this ](https://github.com/dotnet/machinelearning/blob/3af9a5d96ade88e888894af23baef8fe4598f826/docs/samples/Microsoft.ML.Samples/Dynamic/LogisticRegression.cs#L8) example. The graph will have a function selected and red stripes indicate the position in the graph it is called.

image

The checkpoint should be at `cursor.MoveNext()` in `TrainCore` method. As we can see in the graph this function is called periodically in the training process. Everything before this is CPU cycles consumed by the transforms and it is not related to Logistic Regression training cycles.

codemzs:
fixes #3031

Please read the issue before reviewing this PR.

### DIFF
--- a/src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LbfgsPredictorBase.cs
+++ b/src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LbfgsPredictorBase.cs
@@ -475,6 +475,7 @@ namespace Microsoft.ML.Trainers
                     e => e.SetProgress(0, exCount, totalCount));
                 while (cursor.MoveNext())
                 {
+                    Host.CheckAlive();
                     WeightSum += cursor.Weight;
                     if (ShowTrainingStats)
                         ProcessPriorDistribution(cursor.Label, cursor.Weight);
--- a/src/Microsoft.ML.StandardTrainers/Optimizer/Optimizer.cs
+++ b/src/Microsoft.ML.StandardTrainers/Optimizer/Optimizer.cs
@@ -624,6 +624,7 @@ namespace Microsoft.ML.Numeric
                 state.UpdateDir();
                 while (!finished)
                 {
+                    Env.CheckAlive();
                     bool success = state.LineSearch(ch, false);
                     if (!success)
                     {


rogancarr:
What's the performance implications here?

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LbfgsPredictorBase.cs
```diff
@@ -475,6 +475,7 @@ private protected virtual void TrainCore(IChannel ch, RoleMappedData data)
                     e => e.SetProgress(0, exCount, totalCount));
                 while (cursor.MoveNext())
                 {
+                    Host.CheckAlive();
                     WeightSum += cursor.Weight;
```
wschin: I feel it's not the only place we need a check point.

wschin:
>             }

This looks very suspicious. Could you add some check points to this function? I also feel we need to perf algs before adding checking points. 

---
Refers to: src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LbfgsPredictorBase.cs:567 in 5540101. [](commit_id = 554010138d9d9b3288e47863b9dc2ba5df73e099, deletion_comment = False)

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LbfgsPredictorBase.cs
codemzs: Yep, added one more in Line Search Minimize function.

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LbfgsPredictorBase.cs
```diff
@@ -475,6 +475,7 @@ private protected virtual void TrainCore(IChannel ch, RoleMappedData data)
                     e => e.SetProgress(0, exCount, totalCount));
                 while (cursor.MoveNext())
                 {
+                    Host.CheckAlive();
```
sfilipi: >  Host.CheckAlive(); [](start = 18, length = 20)

is it too much to do it in every row fetch? would it be enough to do it every 10 cursor moves, or some other number > 1. 
(idk if there are any best practiced on how to determine the frequency of checks , from maybe the CancellationToken implementations)

codemzs: And how is that any more efficient than what we have now? You will end up executing an if condition on every row fetch ... based on my analysis of the current solution this doesn’t add any significant overhead.

Cancellation token works differently. You register a callback with it and when a signal is sent it invokes the callback and you do work to gracefully shutdown a process. Our plan is to implement cancellation token post 1.0.

rogancarr: We spoke offline. I think this is the best we can do until we get cancellation tokens into the mix. `CheckAlive` only checks a `bool` property, so it's probably faster than checking to see if it's the 10th iteration or not.

justinormont: To add some late flavor to this, the branch predictor should be slightly better at the (almost perfectly) constant `bool` property, than the return value from `iteration % 10` (including the hidden division operation..could check every 8 as the compiler should optimize to a bitwise `AND`).

That said, there's the overhead of the `CheckAlive()` function call which maybe greater if not inlined.

PR #3033, opened 2019-03-20
Refactoring of Options for ImagePixelExtractingEstimator

artidoro:
This PR is an example solution for #2884. Once I receive feedback on this, I will continue with the rest of the transforms.

The purpose of this PR is twofold:

- does the ground work to more easily enable the refactoring of `Options`  in other transforms (commit 1)
    - most of the work is in the file ColumnBindingsBase.cs.
    - made `OneToOneColumn` public, removed empty class
    - renamed `Name` and `Source` to `OutputColumnName` and `InputColumnName`
    - added implicit operators to `OneToOneColumn` to simplify the multicolumn mapping scenario
- refactors the `Options` class for `ImagePixelExtractingEstimator` (commit 2)
    - moved immutable `ColumnOptions` to transformer and renamed `ColumnInfo`
    - moved `Options` and `Column` to estimator
    - refactored extension and constructors

The third commit fixes tests and entrypoint catalog.

Note that with the combination of the implicit operators on `OneToOneColumn` and the constructor taking `OneToOneColumn` in the `Options` class, it is easier to define the multicolumn mapping scenario where the columns don't require column specific settings. For an example of the behavior see the test TestImagePixelExtractOptions in ImagesTests.cs:

```csharp
// options1 and 2 should be exactly the same.
var options1 = new ImagePixelExtractingEstimator.Options
{
    ColumnOptions = new[]
    {
        new ImagePixelExtractingEstimator.ColumnOptions { OutputColumnName = "outputColumn1", InputColumnName = "inputColumn1" },
        new ImagePixelExtractingEstimator.ColumnOptions { OutputColumnName = "outputColumn2", InputColumnName = "inputColumn2" },
        new ImagePixelExtractingEstimator.ColumnOptions { OutputColumnName = "outputColumn3", InputColumnName = "inputColumn3" }
    }
};
var options2 = new ImagePixelExtractingEstimator.Options(("outputColumn1", "inputColumn1"), ("outputColumn2", "inputColumn2"), ("outputColumn3", "inputColumn3"));
// options3 has the same OutputColumnName as the previous two.
var options3 = new ImagePixelExtractingEstimator.Options("outputColumn1", "outputColumn2", "outputColumn3");

```


src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
```diff
@@ -14,16 +14,40 @@
 
 namespace Microsoft.ML.Data
 {
-    internal abstract class SourceNameColumnBase
+    public class OneToOneColumn
```
artidoro: >OneToOneColumn [](start = 17, length = 14)

Needs summary.

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
```diff
@@ -14,16 +14,40 @@
...
+
+        /// <summary>
+        /// Instantiates a <see cref="ColumnOptions"/> from the column name.
+        /// </summary>
+        public static implicit operator OneToOneColumn(string outputColumnName)
+        {
+            return new OneToOneColumn(outputColumnName);
+        }
+
+        public OneToOneColumn(string outputColumnName, string inputColumnName = null)
```
artidoro: >public [](start = 8, length = 6)

Needs summary.

src/Microsoft.ML.ImageAnalytics/ExtensionsCatalog.cs
```diff
@@ -109,14 +109,25 @@ internal static ImageLoadingEstimator LoadImages(this TransformsCatalog catalog,
             float offsetImage = ImagePixelExtractingEstimator.Defaults.Offset,
             float scaleImage = ImagePixelExtractingEstimator.Defaults.Scale,
             bool outputAsFloatArray = ImagePixelExtractingEstimator.Defaults.Convert)
-            => new ImagePixelExtractingEstimator(CatalogUtils.GetEnvironment(catalog), outputColumnName, inputColumnName, colorsToExtract, orderOfExtraction, interleavePixelColors, offsetImage, scaleImage, outputAsFloatArray);
+        {
+            var options = new ImagePixelExtractingEstimator.Options
```
artidoro: >ImagePixelExtractingEstimator [](start = 30, length = 29)

Can use the `Options` constructor.

src/Microsoft.ML.Data/Transforms/TransformBase.cs
```diff
@@ -333,19 +333,19 @@ public VectorType GetSlotType(int col)
...
-                    host.CheckUserArg(item.TrySanitize(), nameof(OneToOneColumn.Name), "Invalid new column name");
-                    names[i] = item.Name;
+                    host.CheckUserArg(item.TrySanitize(), nameof(OneToOneColumn.OutputColumnName), "Invalid new column name");
+                    names[i] = item.OutputColumnName;
 
                     int colSrc;
-                    if (!inputSchema.TryGetColumnIndex(item.Source, out colSrc))
-                        throw host.ExceptUserArg(nameof(OneToOneColumn.Source), "Source column '{0}' not found", item.Source);
+                    if (!inputSchema.TryGetColumnIndex(item.InputColumnName, out colSrc))
+                        throw host.ExceptUserArg(nameof(OneToOneColumn.InputColumnName), "Source column '{0}' not found", item.InputColumnName);
```
sfilipi: >Source [](start = 90, length = 6)

Input column

src/Microsoft.ML.Data/Transforms/NormalizeColumn.cs
```diff
@@ -290,8 +290,8 @@ internal static IDataTransform Create(IHostEnvironment env, MinMaxArguments args
 
             var columns = args.Columns
                 .Select(col => new NormalizingEstimator.MinMaxColumnOptions(
-                    col.Name,
-                    col.Source ?? col.Name,
+                    col.OutputColumnName,
```
TomFinley: >OutputColumnName [](start = 24, length = 16)

I don't think that this is necessarily important, but in prior discussions we've discussed that in the case of a column options *specifically*, given that it's clear this is talking about columns we have discussed that it may be wise to name it `InputName` as opposed to `InputColumnName`, since it's obvious we're talking about columns here.

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
```diff
@@ -14,16 +14,44 @@
 
 namespace Microsoft.ML.Data
 {
-    internal abstract class SourceNameColumnBase
+    /// <summary>
+    /// Specifies input and output column names for a transformation
+    /// </summary>
+    public class OneToOneColumn
```
wschin: >OneToOneColumn [](start = 17, length = 14)

Is OneToOneColumn a good name? Usually `one-to-one` describes a mapping or a function, which doesn't sound a `Column`. Maybe we could try `OneToOneBind`?

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
```diff
@@ -14,16 +14,44 @@
...
     {
+        /// <summary>Name of the column resulting from the transformation of <see cref="InputColumnName"/>.</summary>
         [Argument(ArgumentType.AtMostOnce, HelpText = "Name of the new column", ShortName = "name")]
```
wschin: Maybe add `Name="Name"` to avoid EP changes.

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
```diff
@@ -14,16 +14,44 @@
...
-        [Argument(ArgumentType.AtMostOnce, HelpText = "Name of the source column", ShortName = "src")]
-        public string Source;
+        /// <summary>Name of the column to transform. If set to <see langword="null"/>, the value of the <see cref= "OutputColumnName"/> will be used as source.</summary>
+        [Argument(ArgumentType.AtMostOnce, HelpText = "Name of the source column", ShortName = "source, src")]
+        public string InputColumnName = null;
+
+        /// <summary>
+        /// Instantiates a <see cref="ColumnOptions"/> from a tuple of input and output column names.
+        /// </summary>
+        public static implicit operator OneToOneColumn((string outputColumnName, string inputColumnName) value)
```
wschin: >(string outputColumnName, string inputColumnName) value [](start = 55, length = 55)

I have an impression that tuples are not very welcomed.

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
artidoro: I don't think they are case sensitive. So this should be fine!

---
In reply to: [267456175](https://github.com/dotnet/machinelearning/pull/3033#discussion_r267456175) [](ancestors = 267456175)

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
```diff
@@ -14,16 +14,40 @@
...
-    internal abstract class SourceNameColumnBase
+    public class OneToOneColumn
     {
+        /// <summary>Name of the column resulting from the transformation of <see cref="InputColumnName"/>.</summary>
         [Argument(ArgumentType.AtMostOnce, HelpText = "Name of the new column", ShortName = "name")]
-        public string Name;
+        public string OutputColumnName;
+
+        /// <summary>Name of the column to transform. If set to <see langword="null"/>, the value of the <see cref= "OutputColumnName"/> will be used as source.</summary>
+        [Argument(ArgumentType.AtMostOnce, HelpText = "Name of the source column", ShortName = "source, src")]
```
TomFinley: >ShortName [](start = 83, length = 9)

Note that you could have used the `Name` field here. That might have been preferable if we care about not changing the behavior of command line and entry-points, at least not yet. (We can always adjust later if we really want to.)

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
```diff
@@ -14,16 +14,40 @@
...
+        public string InputColumnName = null;
```
TomFinley: >null [](start = 40, length = 4)

This `null` declaration serves no purpose. The default value of the field is null if unassigned, as we see above with `OutputColumnName`.

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
TomFinley: >OneToOneColumn [](start = 17, length = 14)

Do you mean this class to be used and directly instantiable? I would really not like that. I would prefer that our API be over sealed classes -- otherwise we'll have weird things like it being possible to assign `OneToOneColumn` taking APIs.

As a general policy, we prefer all classes to be either abstract or sealed. We have [generally followed this advice](https://blogs.msdn.microsoft.com/ericlippert/2004/01/22/why-are-so-many-of-the-framework-classes-sealed/).

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
```diff
@@ -14,16 +14,40 @@
...
+        /// </summary>
+        public static implicit operator OneToOneColumn((string outputColumnName, string inputColumnName) value)
+        {
+            return new OneToOneColumn(value.outputColumnName, value.inputColumnName);
+        }
+
+        /// <summary>
+        /// Instantiates a <see cref="ColumnOptions"/> from the column name.
+        /// </summary>
+        public static implicit operator OneToOneColumn(string outputColumnName)
```
TomFinley: >implicit [](start = 22, length = 8)

I get a little nervous when I see implicit operators from common types, especially since in the same APIs that we're going to be using these sorts of types we're going to have `string` parameters being used. We could sort of get away with it with the value-tuples above -- we don't use value-tuples anywhere else -- but this makes me very nervous.

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
TomFinley: >OneToOneColumn [](start = 15, length = 14)

Note that we have a non-sealed directly instantiable class...

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
TomFinley: Reactivating. What @wschin is pointing out, correctly, is that by doing this the name of the entry-point node, and the command line, will change, which seems like an unnecessary change.

---
In reply to: [267456928](https://github.com/dotnet/machinelearning/pull/3033#discussion_r267456928) [](ancestors = 267456928,267456175)

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
artidoro: I am making it abstract, thanks for pointing it out.

---
In reply to: [267543430](https://github.com/dotnet/machinelearning/pull/3033#discussion_r267543430) [](ancestors = 267543430)

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
artidoro: Ok will use Name = .... to prevent further changes

---
In reply to: [267542069](https://github.com/dotnet/machinelearning/pull/3033#discussion_r267542069) [](ancestors = 267542069)

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
artidoro: Ok will remove the constructor

---
In reply to: [267546107](https://github.com/dotnet/machinelearning/pull/3033#discussion_r267546107) [](ancestors = 267546107)

src/Microsoft.ML.Data/Transforms/ColumnBindingsBase.cs
artidoro: Ok will remove the implicit operators, and simply have a constructor in the options class that will take tuples of strings. 

---
In reply to: [267545690](https://github.com/dotnet/machinelearning/pull/3033#discussion_r267545690) [](ancestors = 267545690)

glebuk:
The concern of not taking this PR is that users will have to specify a transform per column. This can result in N*M slowdown where N is number of columns and M is the number of trainable transforms per column. The most common example is categorical, We would strongly prefer to learn all the categories for all columns in a single pass. If 1.0 has no such mechanism to specify multiple columns per transform then we would encourage suboptimal usage within our user base.

artidoro:
Closing in favor of a simpler approach prior to v1.

sfilipi:
Been thinking that the multicolumn case might be more common than we think, for a few transforms, after seeing this PR: [https://github.com/dotnet/machinelearning-samples/pull/321/files#diff-5f3b1d72514e8cdce1dcac64aaf8a097](https://github.com/dotnet/machinelearning-samples/pull/321/files#diff-5f3b1d72514e8cdce1dcac64aaf8a097) 


---
In reply to: [475059668](https://github.com/dotnet/machinelearning/pull/3033#issuecomment-475059668) [](ancestors = 475059668)

PR #2976, opened 2019-03-15
Make Multiclass Linear Trainers Typed Based on Output Model Types.

wschin:
Multiclass SDCA can train multi-class SVM but it always outputs multi-class logistic regression model. This is not correct because we should not apply softmax to SVM model.

To fix #1100, we got the following working items.
- [x] Clean code.
- [x] Create two multi-class linear models.
- [x] Make multiclass SDCA trainers typed.
- [x] Give the two new model classes better.

Framework changes:
- LogisticRegressionMulticlassClassificationTrainer (renamed to) ---> LbfgsMaximumEntropyTrainer 
- MulticlassLogisticRegressionModelParameters (refactorized to) ---> MaximumEntropyModelParameters (for multi-class LR) and LinearMulticlassModelParametersBase (for uncalibrated cases). The two new classes are also derived from LinearMulticlassModelParametersBase. MulticlassLogisticRegressionModelParameters.
- SdcaMulticlassClassificationTrainer (refactorized to) ---> SdcaMulticlassClassificationTrainer and SdcaNonCalibratedMulticlassClassificationTrainer. These two new classes' are derived from SdcaMulticlassClassificationTrainerBase.

API changes:
- (rename, static API) MulticlassLogisticRegression ---> LbfgsMaximumEntropy
- (rename, dynamic API) LogisticRegression ---> LbfgsMaximumEntropy
- (add, static API) SdcaNonCalibrated for multi-class linear models without calibration.
- (add, dynamic API) SdcaNonCalibrated for multi-class linear models without calibration.

Issue #2938, opened 2019-03-13
More trainer related naming alignment


sfilipi:
Take another pass over the trainers and the model parameter types, and align them, because now we have:

`LogisticRegressionMulticlassClassificationTrainer` but `MulticlassLogisticRegressionModelParameters`. 

I think following the same principles on the ModelParams might make them more relatable; so change `MulticlassLogisticRegressionModelParameters` to `LogisticRegressionMulticlassModelParameters`

sfilipi:
@shauheen would this be a candidate for Project 13?

abgoswam:
Following #2919 .. should we call it : 

- LogisticRegressionMulticlass**Classification**ModelParameters

That way it would be consistent across the convention we are following across the codebase now. This looks a bit long, but we can treat this as a special case.

abgoswam:
In general, it seems there isn't a 1:1  mapping between trainer and model parameter types. e.g `SdcaNonCalibratedBinaryClassificationTrainer`  uses `LinearBinaryModelParameters`  . 

However, whenever possible we should string to align the names. 

Listing some of the trainers where we can fix this :
 
- MulticlassLogisticRegressionModelParameters
- BinaryClassificationGamModelParameters
- MultiClassNaiveBayesModelParameters

abgoswam:
Update:
- To keep consistency between **Class** names of Trainers and ModelParameters,  we will **not** use the word "Classification"  in either the Trainer class or the  ModelParameter class

- We feel its OK to drop the word "Classification" for 2 main reasons : 
  - we have sufficient context  to just use **suffix**  :  BinaryTrainer ,  BinaryModelParameters etc. without using the word "Classification"
  - adding the word "Classification" leads to long names 

Here is a summary of the **Trainer** and **ModelParameter**  class names

| Updated Trainer Class Name |
| ------------------------ | 
| FastTreeBinaryTrainer   |
| FastTreeRegressionTrainer |
| FastTreeRankingTrainer |
| FastTreeTweedieTrainer |
| FastForestRegressionTrainer   |
| FastForestBinaryTrainer  |
| MatrixFactorizationTrainer | 
| GamBinaryTrainer |
| GamRegressionTrainer |
| LogisticRegressionBinaryTrainer |
| LogisticRegressionMulticlassTrainer |  
| AveragedPerceptronTrainer  |
| OnlineGradientDescentTrainer |
| PoissonRegressionTrainer |   
| KMeansTrainer |
[...]

Issue #2762, opened 2019-02-27
TrainersName pattern (Discussion)


Ivanidzo4ka:
https://github.com/dotnet/machinelearning/issues/2172 follow up on this one.
So right now we have mix of trainers names.
I want to standardize them.

We have following zoo of naming patterns:
`LogisticRegression`
`SdcaMultiClassTrainer`
`SgdNonCalibratedBinaryTrainer`
`LinearSvmTrainer`
`LightGbmMulticlassTrainer`
`MulticlassLogisticRegression`

My proposal is following:

`{AlgoName}(optional){Calibrated/NonCalibrated}{TypeOfTask}Trainer`

Where AlgoName is full name without abbreviations (SDCA->StochasticDualCoordinateAscent, Linearsvm ->LinearSupportVectorMachines) with exception of LightGBM.

I would also prefer to explicitly specify TypeOfTask even if algorithm exist only for one type. (Which would create weird abominations like OneVersusAllMulticlassTrainer, but i'm fine with that)

Does that sound good for you?
@sfilipi  @TomFinley  @eerhardt  @yaeldekel

wschin:
Maybe `{AlgoName}{**ModelName**}(optional){Calibrated/NonCalibrated}{TypeOfTask}Trainer`?

Ivanidzo4ka:
What ModelName means? Can you give example?

wschin:
StochasticGradient-FieldAwareFactorizationMachine-Calibrated-BinaryClassification
AlternatingNewton-FieldAwareFactorizationMachine-Calibrated-BinaryClassification

abgoswam:
This is also related to #2172 .  The trainer estimator  _Class names_ we finally decide upon  should also match the corresponding _MLContext names_

So if a user has to specify an SDCA binary trainer with advanced options, they would have to say something like : 


// Pipeline.
var pipeline = 
    ml.BinaryClassification.Trainers.**StochasticDualCoordinateAscentNonCalibratedBinaryTrainer**(
       new **StochasticDualCoordinateAscentNonCalibratedBinaryTrainer**.Options { NumThreads = 1 }
    );

It may be overly verbose. Thoughts ?

TomFinley:
> ml.BinaryClassification.Trainers.**StochasticDualCoordinateAscentNonCalibratedBinaryTrainer**(
> new **StochasticDualCoordinateAscentNonCalibratedBinaryTrainer**.Options { NumThreads = 1 }
> );
> 
[...]

PR #3034, opened 2019-03-20
Better names to calibreated linear classification models

### LINKED ISSUE #3016: The "Binary" in "LogisticRegressionBInaryTrainer" Looks Redundant
As title. Logistic regression is clearly a binary classifier, so we can consider dropping "Binary" among "LogisticRegressionBinaryTrainer". In addition, as L-BFGS is outperformed by SDCA and SymSGD, I feel we should not call this class `LogisticRegressionTrainer`. A suggested solution is to prepend `Lbfgs` to this class' name.

wschin:
Fix #3016 by renaming `LogisticRegressionBinaryTrainer` to `LbfgsLogisticRegressionTrainer`. Note that for multiclass case, we have `LbfgsMaximumEntropyTrainer`. In addition, as [SDCA (aka dual coordinate descent methods) outperforms L-BFGS](https://www.csie.ntu.edu.tw/~cjlin/papers/maxent_dual.pdf) when training logistic regression models, we should not make L-BFGS looks like the default trainer of logistic regression model.

**[This link](https://github.com/dotnet/machinelearning/issues/3016#issuecomment-476369026) contains our the conclusion of those new names.**


src/Microsoft.ML.StaticPipe/LbfgsStatic.cs
```diff
@@ -9,15 +9,15 @@
...
 namespace Microsoft.ML.StaticPipe
 {
-    using Options = LogisticRegressionBinaryTrainer.Options;
+    using Options = LbfgsLogisticRegressionTrainer.Options;
 
     /// <summary>
     /// Binary Classification trainer estimators.
     /// </summary>
-    public static class LbfgsBinaryClassificationStaticExtensions
+    public static class LbfgsBinaryExtensions
```
artidoro: >LbfgsBinaryExtensions [](start = 24, length = 21)

I think these are just static extensions, if possible could you keep static in the name of the class?

yaeldekel:
>         [TlcModule.EntryPoint(Name = "Trainers.LogisticRegressionBinaryClassifier",

Should we update the name of the entry point as well?

---
Refers to: src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LogisticRegression.cs:407 in ba22e12. [](commit_id = ba22e1265508967b70815118ca485a8ddc542de2, deletion_comment = False)

wschin:
>         [TlcModule.EntryPoint(Name = "Trainers.LogisticRegressionBinaryClassifier",

No for BC.

---
In reply to: [474926238](https://github.com/dotnet/machinelearning/pull/3034#issuecomment-474926238) [](ancestors = 474926238)

---
Refers to: src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LogisticRegression.cs:407 in ba22e12. [](commit_id = ba22e1265508967b70815118ca485a8ddc542de2, deletion_comment = False)

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LogisticRegression.cs
```diff
@@ -15,21 +15,21 @@
...
-[assembly: LoadableClass(typeof(void), typeof(LogisticRegressionBinaryTrainer), null, typeof(SignatureEntryPointModule), LogisticRegressionBinaryTrainer.LoadNameValue)]
+[assembly: LoadableClass(typeof(void), typeof(LbfgsLogisticRegressionTrainer), null, typeof(SignatureEntryPointModule), LbfgsLogisticRegressionTrainer.LoadNameValue)]
 
 namespace Microsoft.ML.Trainers
 {
 
     /// <include file='doc.xml' path='doc/members/member[@name="LBFGS"]/*' />
     /// <include file='doc.xml' path='docs/members/example[@name="LogisticRegressionBinaryClassifier"]/*' />
-    public sealed partial class LogisticRegressionBinaryTrainer : LbfgsTrainerBase<LogisticRegressionBinaryTrainer.Options,
+    public sealed partial class LbfgsLogisticRegressionTrainer : LbfgsTrainerBase<LbfgsLogisticRegressionTrainer.Options,
```
Ivanidzo4ka: >LbfgsLogisticRegressionTrainer [](start = 32, length = 30)

Can you call it `LbfgsBinaryTrainer`?

wschin: It looks fine if we consider this trainer and its SDCA counterpart (they all solve LR). However, it will break the consistency between LBFGS trainers. Please take a look at Iteration 3.

---
In reply to: [267460936](https://github.com/dotnet/machinelearning/pull/3034#discussion_r267460936) [](ancestors = 267460936)

Ivanidzo4ka: @sfilipi has comment on your issue, and I share it as well.
Putting word `Regression` without `Binary`  in the trainer is confusing.
If I would saw name of this trainer I would assume it's part of regression task.
Which is not.


---
In reply to: [267475700](https://github.com/dotnet/machinelearning/pull/3034#discussion_r267475700) [](ancestors = 267475700,267460936)

src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
```diff
@@ -516,7 +516,7 @@ public IClassificationLoss CreateComponent(IHostEnvironment env)
         /// ]]>
         /// </format>
         /// </example>
-        public static LogisticRegressionBinaryTrainer LogisticRegression(this BinaryClassificationCatalog.BinaryClassificationTrainers catalog,
+        public static LbfgsLogisticRegressionTrainer LbfgsLogisticRegressio(this BinaryClassificationCatalog.BinaryClassificationTrainers catalog,
```
Ivanidzo4ka: >LbfgsLogisticRegressio [](start = 53, length = 22)

you miss `n` in the end.
And I don't think this new verb any better than previous one.
More importantly I don't understand why you even changing this? We had @agoswami issue and PR regarding names, where we spend quite a lot of effort in attempt to figure out names, why you change it out of blue now?

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LogisticRegression.cs
sfilipi: yeah, given the target user base, IMO should do a one off for this one, and keep binary in the name. 

(also this feels like a classic academic vs engineering solution of  a problem :D )

---
In reply to: [267477762](https://github.com/dotnet/machinelearning/pull/3034#discussion_r267477762) [](ancestors = 267477762,267475700,267460936)

src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
wschin: I don't think my changes violate his rule.

---
In reply to: [267478532](https://github.com/dotnet/machinelearning/pull/3034#discussion_r267478532) [](ancestors = 267478532)

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LogisticRegression.cs
wschin: It will make the API looks like `mlContext.BinaryClassification.LbfgsBinary(...)` so I switch to SDCA-style name `LbfgsCalibrated`.

---
In reply to: [267481155](https://github.com/dotnet/machinelearning/pull/3034#discussion_r267481155) [](ancestors = 267481155,267477762,267475700,267460936)

docs/samples/Microsoft.ML.Samples/Dynamic/LogisticRegression.cs
```diff
@@ -63,7 +63,7 @@ public static void Example()
                 .Append(ml.Transforms.Text.FeaturizeText("TextFeatures", "Text"))
                 .Append(ml.Transforms.Concatenate("Features", "TextFeatures", "age", "fnlwgt",
                     "education-num", "capital-gain", "capital-loss", "hours-per-week"))
-                .Append(ml.BinaryClassification.Trainers.LogisticRegression());
+                .Append(ml.BinaryClassification.Trainers.LbfgsCalibrated());
```
rogancarr: >LbfgsCalibrated [](start = 57, length = 15)

I don't think we should call it this. `LogisticRegression` is fine but `LbfgsCalibrated` is too esoteric.

src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
rogancarr: This PR makes me nervous so close to ship date. Let's get consensus in #3016 before we make the change.

---
In reply to: [267481247](https://github.com/dotnet/machinelearning/pull/3034#discussion_r267481247) [](ancestors = 267481247,267478532)

docs/samples/Microsoft.ML.Samples/Dynamic/LogisticRegression.cs
rogancarr: Plus it's not a calibrated linear model. It's fitting the to the logistic loss function.

wschin: As long as it returns `CalibratedModelParametersBase<LinearBinaryModelParameters, PlattCalibrator`, I feel `LbfgsCalibrated` is fine. SDCA is doing the same.

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LogisticRegression.cs
wschin: > 
> 
> yeah, given the target user base, IMO should do a one off for this one, and keep binary in the name.
> 
> (also this feels like a classic academic vs engineering solution of a problem :D )
> 
> Also, IMO not all details need to be in the name. IMO we don't need to distingush that this is Lbfgs logistic regression through renaming. It can be in the documentations. shorter names are better...

There are several trainers returning the same model types. Only this one is called `LogisticRegression` and this one has been outperformed by other algorithms. Thus, I want to append `Lbfgs` to emphasize that 'this is not the only trainer for logistic regression`.

> 
> In reply to: [267477762](https://github.com/dotnet/machinelearning/pull/3034#discussion_r267477762) [](ancestors = 267477762,267475700,267460936)

docs/samples/Microsoft.ML.Samples/Dynamic/LogisticRegression.cs
rogancarr: Technically true, but wouldn't it be clearer from a user perspective to call it `LogisticRegression`? Everybody knows what that is. Behind the scenes, many packages solve LR with variants of L-BFGS (e.g. Spark), so I don't think it would be misleading to say that it's called `LogisticRegression`. We can put technical details like "Implemented with L-BFGS etc." in the docs.

Look at `FastTree`. We don't call that `CalibratedLambdaMART` ;)

---
In reply to: [267857743](https://github.com/dotnet/machinelearning/pull/3034#discussion_r267857743) [](ancestors = 267857743)

wschin: A problem here is that L-BFGS is not good for large-scale sparse data sets. SDCA in that case is usually much faster. If we call LBFGS-LogisticRegression just LogisticRegression, external users may view LBFGS as the default solver to all logistic regression problems.

docs/samples/Microsoft.ML.Samples/Dynamic/LogisticRegression.cs
```diff
@@ -63,7 +63,7 @@ public static void Example()
                 .Append(ml.Transforms.Text.FeaturizeText("TextFeatures", "Text"))
                 .Append(ml.Transforms.Concatenate("Features", "TextFeatures", "age", "fnlwgt",
                     "education-num", "capital-gain", "capital-loss", "hours-per-week"))
-                .Append(ml.BinaryClassification.Trainers.LogisticRegression());
+                .Append(ml.BinaryClassification.Trainers.LbfgsLogisticRegression());
```
rogancarr: How do you feel about making the other calibrated linear trainers, like SDCA into `XyzLogisticRegression()`.

rogancarr: We can always add that at a later time.

abgoswam:
Here is my understanding of how we arrived at naming convention of Trainers/ModelParameters

- we had couple of issues #2762 and #2938  discussing this, summarized in the comments :
  - https://github.com/dotnet/machinelearning/issues/2762#issuecomment-471845715
  - https://github.com/dotnet/machinelearning/issues/2938#issuecomment-474131883
- MulticlassLR was discussed, and finalized while fixing issue #2976


The PR proposes to **modify the API** as follows : 
- Renames several MLContext names  
  - `LogisticRegression` to `LbfgsLogisticRegression` 
  - `PoissonRegression` to `LbfgsPoissonRegression`
- Renames Class names
  - `LogisticRegressionBinaryTrainer` to `LbfgsLogisticRegressionTrainer`
 
Why do we think adding prefix "Lbfgs" in the name is a better alternative than what we have currently ?

@eerhardt

eerhardt:
> Why do we think adding prefix "Lbfgs" in the name is a better alternative than what we have currently ? In fact to me it seems having 'Lbfgs' as prefix might throw users off

I believe @wschin answered this in the first comment:

> In addition, as SDCA (aka dual coordinate descent methods) outperforms L-BFGS when training logistic regression models, we should not make L-BFGS looks like the default trainer of logistic regression model.

It seems to me that is his exact intention here. Don't use this one when you want to do LR. Or at least, don't think this is the "good"/"default" one. We have a better one.

rogancarr:
I think that it would also make sense to convert the binary `SDCA` trainer to be called `SdcaLogisticRegression` if we make this change to `Lbfgs`. That way, they both get the "LR Brand Name".

abgoswam:
> In addition, as SDCA (aka dual coordinate descent methods) outperforms L-BFGS when training logistic regression models, we should not make L-BFGS looks like the default trainer of logistic regression model.

- The PR  proposes adding "LbfgsLogisticRegression" in API name itself to make it  concrete  that this  particular  API  uses  "Lbfgs" as the optimization algorithm.  Is that the correct approach ? 

- Are we sure we are using "LBFGS" as the optimizing algorithm ?  The internal repo  refers to this [paper](https://www.microsoft.com/en-us/research/publication/scalable-training-of-l1-regularized-log-linear-models/?from=http%3A%2F%2Fresearch.microsoft.com%2Fapps%2Fpubs%2Fdefault.aspx%3Fid%3D78900),  which leads me to believe  that what we use the the  optimization algo is  not LBFGS itself, but rather  one of its variants "OWN-QN"

- Scikit-learn [solves this](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html) by adding parameter  called "solver" 

> solver : str, {‘newton-cg’, ‘lbfgs’, ‘liblinear’, ‘sag’, ‘saga’}, default: ‘liblinear’.
Algorithm to use in the optimization problem.

wschin:
> 
> 
> > In addition, as SDCA (aka dual coordinate descent methods) outperforms L-BFGS when training logistic regression models, we should not make L-BFGS looks like the default trainer of logistic regression model.
> 
>     * The PR  proposes adding "LbfgsLogisticRegression" in API name itself to make it  concrete  that this  particular  API  uses  "Lbfgs" as the optimization algorithm.  Is that the correct approach ?
> 

LBFGS can mean LBFGS family algorithms.

>     * Are we sure we are using "LBFGS" as the optimizing algorithm ?  The internal repo  refers to this [paper](https://www.microsoft.com/en-us/research/publication/scalable-training-of-l1-regularized-log-linear-models/?from=http%3A%2F%2Fresearch.microsoft.com%2Fapps%2Fpubs%2Fdefault.aspx%3Fid%3D78900),  which leads me to believe that the optimization algo that ML.NET uses is  not LBFGS itself, but rather  one of its variants "OWN-QN". All the more reason to _not_ include Lbfgs in the name of the API itself
> 

It's a variant of LBFGS, as stated in your link.

>     * Scikit-learn [solves this](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html) by adding parameter  called "solver"
> 
> 
> > solver : str, {‘newton-cg’, ‘lbfgs’, ‘liblinear’, ‘sag’, ‘saga’}, default: ‘liblinear’.
> > Algorithm to use in the optimization problem.

No. It's not necessary. L1-norm --> a variant of LBFGS, No L1-norm --> LBFGS.

abgoswam:
Couple of notes after discussing with @wschin 

- Since ML.NET is type-safe, we cannot really  follow the same paradigm as Scikit-Learn's way of handling this (using the `solver` parameter)

- We can take inspiration from Spark.ML  to finilize the API  

  - Spark.ML package exposes just 1 API
    -  [`LogisticRegression`](https://spark.apache.org/docs/2.2.0/api/scala/index.html#org.apache.spark.ml.classification.LogisticRegression)  and clarifies in documentation it uses LBFGS / OWLQN 

  - Spark.Mllib package exposes 2 APIs
    -  [`LogisticRegressionWithLBFGS`](https://spark.apache.org/docs/2.2.0/api/scala/index.html#org.apache.spark.mllib.classification.LogisticRegressionWithLBFGS)  
    -  [`LogisticRegressionWithSGD`](https://spark.apache.org/docs/2.2.0/api/scala/index.html#org.apache.spark.mllib.classification.LogisticRegressionWithSGD)  

- SDCA has other subtelities to it e.g.  supports SVM loss,  calibration/non-calibration etc.  We can keep the SDCA API as is.

src/Microsoft.ML.Mkl.Components/SymSgdClassificationTrainer.cs
```diff
@@ -19,20 +19,20 @@
 using Microsoft.ML.Trainers;
 using Microsoft.ML.Transforms;
 
-[assembly: LoadableClass(typeof(SymbolicSgdTrainer), typeof(SymbolicSgdTrainer.Options),
+[assembly: LoadableClass(typeof(SymbolicSgdLogisticRegressionBinaryTrainer), typeof(SymbolicSgdLogisticRegressionBinaryTrainer.Options),
```
rogancarr: >SymbolicSgdLogisticRegressionBinaryTrainer [](start = 32, length = 42)

`LogisticRegressionBinaryTrainer` => `LbfgsLogisticRegressionTrainer`. Shall `SymbolicSgdBinaryTrainer` => `SymbolicSgdLogisticRegressionTrainer` instead of `SymbolicSgdLogisticRegressionBinaryTrainer`? Right now, the namings don't line up.

src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
```diff
@@ -181,7 +181,7 @@ public static class StandardTrainersCatalog
         }
 
         /// <summary>
-        /// Predict a target using a linear classification model trained with <see cref="SdcaCalibratedBinaryTrainer"/>.
+        /// Predict a target using a linear classification model trained with <see cref="SdcaLogisticRegressionBinaryTrainer"/>.
```
rogancarr: >SdcaLogisticRegressionBinaryTrainer [](start = 89, length = 35)

Similar to `SymSgd`, now `Sdca` has `LogisticRegression` and `Binary`.

src/Microsoft.ML.Mkl.Components/SymSgdClassificationTrainer.cs
abgoswam: @rogancarr..could you cross-check ? To me the *Class* names do match up `LbfgsLogisticRegressionBinaryTrainer` and `SymbolicSgdLogisticRegressionBinaryTrainer`

---
In reply to: [268859710](https://github.com/dotnet/machinelearning/pull/3034#discussion_r268859710) [](ancestors = 268859710)

src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
wschin: Yes, adding both of `LogisticRegression` and `Binary` to trainer classes was our decision. For API, we only append `LogisticRegression` because `Binary` naturally comes from their context.

---
In reply to: [268859905](https://github.com/dotnet/machinelearning/pull/3034#discussion_r268859905) [](ancestors = 268859905)

src/Microsoft.ML.Mkl.Components/SymSgdClassificationTrainer.cs
wschin: Yeah. Our conclusion was having `LR` and `Binary` for trainer classes and only `LR` for APIs.

---
In reply to: [268865148](https://github.com/dotnet/machinelearning/pull/3034#discussion_r268865148) [](ancestors = 268865148,268859710)

rogancarr: Got it. I was mixing up catalogs & classes.

PR #3039, opened 2019-03-20
Add API to get Precision-Recall Curve data

### LINKED ISSUE #2645: Cannot produce a Precision-Recall Curve from our Binary Classification Evaluator
In the Supported Scenarios for V1 (see #2498) we want to be able to easily plot a PR curve from output of the evaluator.

This is not currently possible in the current set of APIs*. As detailed in #2465, we cannot set a threshold on a binary classification prediction, so we cannot loop over the various thresholds. Furthermore, the Evaluator, while it has enough information to return lists of precisions and recalls at the various cutoffs, does not have an avenue to return this information.

\* Well, actually this is possible using custom mappers like so:
1. Get a list of all unique probability scores
  (e.g. by reading the IDataView as an enumerable and creating a dictionary of observed values)
2. For each unique value of the probability:
  a. Write a custom mapper to produce PredictedLabel at that probability threshold.
  b. Calculate Precision and Recall with these labels.

However, it doesn't seem like this workaround qualifies as satisfying the V1 Scenario &mdash; it's a more advanced use of the APIs, and it requires an individual to understand the ins-and-outs of calculating PR curves

ganik:
fixes #2645

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Static/SDCABinaryClassification.cs
+++ b/docs/samples/Microsoft.ML.Samples/Static/SDCABinaryClassification.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Collections.Generic;
+using Microsoft.ML.Data;
 using Microsoft.ML.StaticPipe;
 
 namespace Microsoft.ML.Samples.Static
@@ -89,7 +91,7 @@ namespace Microsoft.ML.Samples.Static
             // Evaluate how the model is doing on the test data
             var dataWithPredictions = model.Transform(testData);
 
-            var metrics = mlContext.BinaryClassification.Evaluate(dataWithPredictions, row => row.Label, row => row.Score);
+            var metrics = mlContext.BinaryClassification.EvaluateWithPRCurve(dataWithPredictions, row => row.Label, row => row.Score, out IEnumerable<PrecisionRecallMetrics> prCurve);
 
             Console.WriteLine($"Accuracy: {metrics.Accuracy}"); // 0.83
             Console.WriteLine($"AUC: {metrics.AreaUnderRocCurve}"); // 0.88
@@ -98,7 +100,16 @@ namespace Microsoft.ML.Samples.Static
             Console.WriteLine($"Negative Precision: {metrics.NegativePrecision}"); // 0.87
             Console.WriteLine($"Negative Recall: {metrics.NegativeRecall}"); // 0.91
             Console.WriteLine($"Positive Precision: {metrics.PositivePrecision}"); // 0.65
-            Console.WriteLine($"Positive Recall: {metrics.PositiveRecall}"); // 0.55          
+            Console.WriteLine($"Positive Recall: {metrics.PositiveRecall}"); // 0.55      
+
+            var prData = prCurve.GetEnumerator();
+            while (prData.MoveNext())
+            {
+                Console.Write($"Threshold: {prData.Current.Threshold} ");
+                Console.Write($"Precision: {prData.Current.Precision} ");
+                Console.Write($"Recall: {prData.Current.Recall} ");
+                Console.WriteLine($"FPR: {prData.Current.FalsePositiveRate}");
+            }
         }
     }
 }
--- a/docs/samples/Microsoft.ML.Samples/Static/SDCABinaryClassification.cs
+++ b/docs/samples/Microsoft.ML.Samples/Static/SDCABinaryClassification.cs
@@ -91,7 +91,7 @@ namespace Microsoft.ML.Samples.Static
             // Evaluate how the model is doing on the test data
             var dataWithPredictions = model.Transform(testData);
 
-            var metrics = mlContext.BinaryClassification.EvaluateWithPRCurve(dataWithPredictions, row => row.Label, row => row.Score, out IEnumerable<PrecisionRecallMetrics> prCurve);
+            var metrics = mlContext.BinaryClassification.EvaluateWithPRCurve(dataWithPredictions, row => row.Label, row => row.Score, out List<BinaryPrecisionRecallDataPoint> prCurve);
 
             Console.WriteLine($"Accuracy: {metrics.Accuracy}"); // 0.83
             Console.WriteLine($"AUC: {metrics.AreaUnderRocCurve}"); // 0.88
@@ -100,15 +100,14 @@ namespace Microsoft.ML.Samples.Static
             Console.WriteLine($"Negative Precision: {metrics.NegativePrecision}"); // 0.87
             Console.WriteLine($"Negative Recall: {metrics.NegativeRecall}"); // 0.91
             Console.WriteLine($"Positive Precision: {metrics.PositivePrecision}"); // 0.65
-            Console.WriteLine($"Positive Recall: {metrics.PositiveRecall}"); // 0.55      
+            Console.WriteLine($"Positive Recall: {metrics.PositiveRecall}"); // 0.55
 
-            var prData = prCurve.GetEnumerator();
-            while (prData.MoveNext())
+            foreach(var prData in prCurve)
             {
-                Console.Write($"Threshold: {prData.Current.Threshold} ");
-                Console.Write($"Precision: {prData.Current.Precision} ");
-                Console.Write($"Recall: {prData.Current.Recall} ");
-                Console.WriteLine($"FPR: {prData.Current.FalsePositiveRate}");
+                Console.Write($"Threshold: {prData.Threshold} ");
+                Console.Write($"Precision: {prData.Precision} ");
+                Console.Write($"Recall: {prData.Recall} ");
+                Console.WriteLine($"FPR: {prData.FalsePositiveRate}");
             }
         }
     }
--- a/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
+++ b/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
@@ -829,7 +829,7 @@ namespace Microsoft.ML.Data
         }
 
         /// <summary>
-        /// Evaluates scored binary classification data.
+        /// Evaluates scored binary classification data and generates precision recall curve data.
         /// </summary>
         /// <param name="data">The scored data.</param>
         /// <param name="label">The name of the label column in <paramref name="data"/>.</param>
@@ -915,6 +915,58 @@ namespace Microsoft.ML.Data
                 moved = cursor.MoveNext();
                 Host.Assert(!moved);
             }
+            return result;
+        }
+
+        /// <summary>
+        /// Evaluates scored binary classification data, without probability-based metrics
+        /// and generates precision recall curve data.
+        /// </summary>
+        /// <param name="data">The scored data.</param>
+        /// <param name="label">The name of the label column in <paramref name="data"/>.</param>
+        /// <param name="score">The name of the score column in <paramref name="data"/>.</param>
+        /// <param name="predictedLabel">The name of the predicted label column in <paramref name="data"/>.</param>
+        /// <param name="prCurve">The generated precision recall curve data. Up to 100000 of samples are used for p/r curve generation.</param>
+        /// <returns>The evaluation results for these uncalibrated outputs.</returns>
+        /// <seealso cref="Evaluate(IDataView, string, string, string)"/>
+        public BinaryClassificationMetrics EvaluateWithPRCurve(IDataView data, string label, string score, string predictedLabel, out IEnumerable<PrecisionRecallMetrics> prCurve)
+        {
+            Host.CheckValue(data, nameof(data));
+            Host.CheckNonEmpty(label, nameof(label));
+            Host.CheckNonEmpty(score, nameof(score));
+            Host.CheckNonEmpty(predictedLabel, nameof(predictedLabel));
+
+            var roles = new RoleMappedData(data, opt: false,
+                RoleMappedSchema.ColumnRole.Label.Bind(label),
+                RoleMappedSchema.CreatePair(AnnotationUtils.Const.ScoreValueKind.Score, score),
+                RoleMappedSchema.CreatePair(AnnotationUtils.Const.ScoreValueKind.PredictedLabel, predictedLabel));
+
+            var resultDict = ((IEvaluator)this).Evaluate(roles);
+            Host.Assert(resultDict.ContainsKey(MetricKinds.PrCurve));
+            var prCurveView = resultDict[MetricKinds.PrCurve];
+            Host.Assert(resultDict.ContainsKey(MetricKinds.OverallMetrics));
+            var overall = resultDict[MetricKinds.OverallMetrics];
+
+            List<PrecisionRecallMetrics> prCurveResult = new List<PrecisionRecallMetrics>();
+            using (var cursor = prCurveView.GetRowCursorForAllColumns())
+            {
+                while (cursor.MoveNext())
+                {
+                    prCurveResult.Add(new PrecisionRecallMetrics(Host, cursor));
+                }
+            }
+            prCurve = prCurveResult;
+
+            BinaryClassificationMetrics result;
+            using (var cursor = overall.GetRowCursorForAllColumns())
+            {
+                var moved = cursor.MoveNext();
+                Host.Assert(moved);
+                result = new BinaryClassificationMetrics(Host, cursor);
+                moved = cursor.MoveNext();
+                Host.Assert(!moved);
+            }
+
             return result;
         }
     }
--- a/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
+++ b/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
@@ -838,7 +838,13 @@ namespace Microsoft.ML.Data
         /// <param name="predictedLabel">The name of the predicted label column in <paramref name="data"/>.</param>
         /// <param name="prCurve">The generated precision recall curve data. Up to 100000 of samples are used for p/r curve generation.</param>
         /// <returns>The evaluation results for these calibrated outputs.</returns>
-        public CalibratedBinaryClassificationMetrics EvaluateWithPRCurve(IDataView data, string label, string score, string probability, string predictedLabel, out IEnumerable<PrecisionRecallMetrics> prCurve)
+        public CalibratedBinaryClassificationMetrics EvaluateWithPRCurve(
+            IDataView data,
+            string label,
+            string score,
+            string probability,
+            string predictedLabel,
+            out List<BinaryPrecisionRecallDataPoint> prCurve)
         {
             Host.CheckValue(data, nameof(data));
             Host.CheckNonEmpty(label, nameof(label));
@@ -858,12 +864,18 @@ namespace Microsoft.ML.Data
             Host.Assert(resultDict.ContainsKey(MetricKinds.OverallMetrics));
             var overall = resultDict[MetricKinds.OverallMetrics];
 
-            List<PrecisionRecallMetrics> prCurveResult = new List<PrecisionRecallMetrics>();
+            var prCurveResult = new List<BinaryPrecisionRecallDataPoint>();
             using (var cursor = prCurveView.GetRowCursorForAllColumns())
             {
+                PrecisionRecallDataPointGetters(prCurveView, cursor,
+                    out ValueGetter<float> thresholdGetter,
+                    out ValueGetter<double> precisionGetter,
+                    out ValueGetter<double> recallGetter,
+                    out ValueGetter<double> fprGetter);
+
                 while (cursor.MoveNext())
                 {
-                    prCurveResult.Add(new PrecisionRecallMetrics(Host, cursor));
+                    prCurveResult.Add(new BinaryPrecisionRecallDataPoint(thresholdGetter, precisionGetter, recallGetter, fprGetter));
                 }
             }
             prCurve = prCurveResult;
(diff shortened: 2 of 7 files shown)


src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
```diff
@@ -789,6 +789,34 @@ private void ComputePrCurves()
...
+        /// <summary>
+        /// Generates precision recall curve data points.
+        /// </summary>
+        /// <param name="data">The scored data.</param>
+        /// <param name="label">The name of the label column in <paramref name="data"/>.</param>
+        /// <param name="score">The name of the score column in <paramref name="data"/>.</param>
+        /// <param name="probability">The name of the probability column in <paramref name="data"/>, the calibrated version of <paramref name="score"/>.</param>
+        /// <param name="predictedLabel">The name of the predicted label column in <paramref name="data"/>.</param>
+        /// <returns>The precision recall curve data points.</returns>
+        public IDataView GetPrecisionRecallCurve(IDataView data, string label, string score, string probability, string predictedLabel)
```
ganik: Unit tests & sample is tbd

Ivanidzo4ka: Do you plan add `mlContext.BinaryClassification.GetPresicionRecallCurve`?

TomFinley: Hi @ganik, thanks for working on this! I view the API in its current form as having a few potential improvements...

The most obvious thing is this return value... an `IDataView`. Probably not a very good transmission format for metrics results.

It may be informative to note how we surface other metrics results in our public API. You'll note that we *do* have `IEvaluator` (which is an internal interface) return metrics via `IDataView` (because, it needs to communicate them in *some* common format, and also because that's one of the types that entry-points uses), but note what we do with them for the sake of our public API.

https://github.com/dotnet/machinelearning/blob/5b22420d28c0cacc9b265d043555b6d11a017b91/src/Microsoft.ML.Data/Evaluators/Metrics/BinaryClassificationMetrics.cs#L87-L98

We actually take the data out, and store them in an actual strongly typed structure, with intellisense, and whatnot.

We *could* have done what you did here -- just sort of take the `IDataView` even though it was one whose structure we knew *exactly* what was in it, called it a day, and said, "here, good luck figuring out what to do with this," but we didn't do that because that would be a needlessly terrible experience. Here we caneasy to use, discoverable API for something as important as metrics, and there was an obvious way to do it, just as there are with this PR. So, I was expecting, and I kind of think people might appreciate something more closely resembling that, with something with actual types, since unlike in some scenarios (general dataflow) the types and column names are well defined.

Note, here is where this data view is composed, for your own reference, so you can see what goes in it.

https://github.com/dotnet/machinelearning/blob/5b22420d28c0cacc9b265d043555b6d11a017b91/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs#L373-L392

So, a few columns that are always there. We'll need to figure out what's up with these "strat" columns, that is, "stratification" columns (we don't want to expose this as such re: @rogancarr), so perhaps we skip that one, or figure out some other structure that's appropriate. (We'd probably group them, somehow, I don't quite know offhand what that's for. CV?) Anyway, you'll want to figure out what's up with that.

You could imagine some *alternate* structure, whereby this "stuff" is put into a strongly typed structure. In the short term you could copy the results of the `IDataView` out into this structure -- that's a little inefficient but will suffice. Could be as simple as an array of a struct containing the relevant PR points. 

The second problem with the API is a bit less fundamnetal, and probably easier to fix. You will also note that that you are discarding all the other results of evaluation, for no particular reason. What I had suggested in the original issue, and I still think is a good idea, is `EvaluateWithPRCurve` that still produces the metric results (since they're already ***right there*** and being calculated alongside the PR curves!). So, maybe return the metrics, but also have the PR curve structure, whatever it is (not just an `IDataView`!!), as an `out` parameter. Or reverse the two, I don't know.

ganik: Awesome, thank you for great suggestions as usually @TomFinley

yaeldekel: >predictedLabel [](start = 120, length = 14)

Does the `BinaryClassifierEvaluator` need that column to evaluate? (same question for the API below).

rogancarr: I like this idea: `EvaluateWithPRCurve`

One thing that Gani and I talked about was also being able to return the data for an ROC curve too. The original issue (and related v1 scenario) was specifically about a PR curve, but I think it should be on the table to return an ROC curve too. These are the two major curves that people may want to plot.

Also, we may want to think about being able to extend this design to any sort of curve someone may want to plot, as different fields of applications have slightly different target metrics.

---
In reply to: [267839855](https://github.com/dotnet/machinelearning/pull/3039#discussion_r267839855) [](ancestors = 267839855)

rogancarr: >string label, string score, string probability, string predictedLabel [](start = 65, length = 69)

Do these column names have default values? If so, let's specify them here as defaults.

yaeldekel: These are the internal methods that are called by the public `Evaluate` API. The public API does have default values for the column names.

---
In reply to: [267922053](https://github.com/dotnet/machinelearning/pull/3039#discussion_r267922053) [](ancestors = 267922053)

ganik: it should be added

---
In reply to: [267595498](https://github.com/dotnet/machinelearning/pull/3039#discussion_r267595498) [](ancestors = 267595498)

ganik: this should be done. ROC Curve I suggest to do post V1.

---
In reply to: [267921726](https://github.com/dotnet/machinelearning/pull/3039#discussion_r267921726) [](ancestors = 267921726,267839855)

ganik: as Yael says

---
In reply to: [267926650](https://github.com/dotnet/machinelearning/pull/3039#discussion_r267926650) [](ancestors = 267926650,267922053)

ganik: sample is done, tests tbd

---
In reply to: [267591483](https://github.com/dotnet/machinelearning/pull/3039#discussion_r267591483) [](ancestors = 267591483)

ganik: Yes it is needed, for ex. for accuracy metrics.

---
In reply to: [267902785](https://github.com/dotnet/machinelearning/pull/3039#discussion_r267902785) [](ancestors = 267902785)

docs/samples/Microsoft.ML.Samples/Static/SDCABinaryClassification.cs
```diff
@@ -98,7 +100,16 @@ public static void SdcaBinaryClassification()
             Console.WriteLine($"Negative Precision: {metrics.NegativePrecision}"); // 0.87
             Console.WriteLine($"Negative Recall: {metrics.NegativeRecall}"); // 0.91
             Console.WriteLine($"Positive Precision: {metrics.PositivePrecision}"); // 0.65
-            Console.WriteLine($"Positive Recall: {metrics.PositiveRecall}"); // 0.55          
+            Console.WriteLine($"Positive Recall: {metrics.PositiveRecall}"); // 0.55
```
glebuk: >       [](start = 84, length = 6)

extra white space?

docs/samples/Microsoft.ML.Samples/Static/SDCABinaryClassification.cs
```diff
@@ -98,7 +100,16 @@ public static void SdcaBinaryClassification()
             Console.WriteLine($"Negative Precision: {metrics.NegativePrecision}"); // 0.87
             Console.WriteLine($"Negative Recall: {metrics.NegativeRecall}"); // 0.91
             Console.WriteLine($"Positive Precision: {metrics.PositivePrecision}"); // 0.65
-            Console.WriteLine($"Positive Recall: {metrics.PositiveRecall}"); // 0.55          
+            Console.WriteLine($"Positive Recall: {metrics.PositiveRecall}"); // 0.55      
+
+            var prData = prCurve.GetEnumerator();
```
glebuk: >prCurve [](start = 25, length = 7)

why not just foreach(var data in prCurve) ?

src/Microsoft.ML.Data/Evaluators/Metrics/PrecisionRecallMetrics.cs
```diff
@@ -0,0 +1,42 @@
...
+
+using System.Collections.Generic;
+using Microsoft.ML.Runtime;
+
+namespace Microsoft.ML.Data
+{
+    /// <summary>
+    /// This class represents one data point on Precision-Recall curve.
+    /// </summary>
+    public sealed class PrecisionRecallMetrics
```
glebuk: >PrecisionRecallMetrics [](start = 24, length = 22)

This appears to be specific to binary PR, yet the name is generic (you use binaryevaluator.)  
If it is for binary, don't you need to also supply positive/negative {recall/precistion|} metrics?
If it  multiclass, you would need a pr matrix afaik

src/Microsoft.ML.StaticPipe/EvaluatorStaticExtensions.cs
```diff
@@ -50,6 +51,44 @@ public static class EvaluatorStaticExtensions
...
+        /// <param name="label">The index delegate for the label column.</param>
+        /// <param name="pred">The index delegate for columns from calibrated prediction of a binary classifier.
+        /// Under typical scenarios, this will just be the same tuple of results returned from the trainer.</param>
+        /// <param name="prCurve">The generated precision recall curve data. Up to 100000 of samples are used for p/r curve generation.</param>
+        /// <returns>The evaluation results for these calibrated outputs.</returns>
+        public static CalibratedBinaryClassificationMetrics EvaluateWithPRCurve<T>(
+            this BinaryClassificationCatalog catalog,
+            DataView<T> data,
+            Func<T, Scalar<bool>> label,
+            Func<T, (Scalar<float> score, Scalar<float> probability, Scalar<bool> predictedLabel)> pred,
```
glebuk: >predictedLabel [](start = 82, length = 14)

is this always a bool? don't we sometimes predict ints or keys?

src/Microsoft.ML.StaticPipe/EvaluatorStaticExtensions.cs
```diff
@@ -50,6 +51,44 @@ public static class EvaluatorStaticExtensions
...
+            out IEnumerable<PrecisionRecallMetrics> prCurve)
```
glebuk: >prCurve [](start = 52, length = 7)

find the out param awkward. Can we extend the CalibratedBinaryClassificationMetrics instead?

src/Microsoft.ML.StaticPipe/EvaluatorStaticExtensions.cs
```diff
@@ -50,6 +51,44 @@ public static class EvaluatorStaticExtensions
...
+            (var scoreCol, var probCol, var predCol) = pred(indexer.Indices);
+            env.CheckParam(scoreCol != null, nameof(pred), "Indexing delegate resulted in null score column.");
+            env.CheckParam(probCol != null, nameof(pred), "Indexing delegate resulted in null probability column.");
+            env.CheckParam(predCol != null, nameof(pred), "Indexing delegate resulted in null predicted label column.");
+            string scoreName = indexer.Get(scoreCol);
+            string probName = indexer.Get(probCol);
+            string predName = indexer.Get(predCol);
+
+            var eval = new BinaryClassifierEvaluator(env, new BinaryClassifierEvaluator.Arguments() { NumRocExamples = 100000 });
+            return eval.EvaluateWithPRCurve(data.AsDynamic, labelName, scoreName, probName, predName, out prCurve);
```
glebuk: super simmilar to another new method. Refactor?

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
```diff
@@ -828,6 +828,59 @@ public CalibratedBinaryClassificationMetrics Evaluate(IDataView data, string lab
...
+            var resultDict = ((IEvaluator)this).Evaluate(roles);
+            Host.Assert(resultDict.ContainsKey(MetricKinds.PrCurve));
+            var prCurveView = resultDict[MetricKinds.PrCurve];
+            Host.Assert(resultDict.ContainsKey(MetricKinds.OverallMetrics));
+            var overall = resultDict[MetricKinds.OverallMetrics];
+
+            List<PrecisionRecallMetrics> prCurveResult = new List<PrecisionRecallMetrics>();
+            using (var cursor = prCurveView.GetRowCursorForAllColumns())
+            {
+                while (cursor.MoveNext())
```
TomFinley: Just to put in the PR what we had talked about, it is important that we get the getters *before* the tight loop, and we use the getters *inside* the tight loop to retrieve the values, since constructing the delegates is extremely expensive but using them is very cheap. (See e.g., [here](https://github.com/dotnet/machinelearning/blob/master/docs/code/IDataViewImplementation.md#why-getgetter).)

src/Microsoft.ML.StaticPipe/EvaluatorStaticExtensions.cs
ganik: this is a twin method for public static CalibratedBinaryClassificationMetrics Evaluate(..) on line 28. It does everything the same way plus spits out PR Curve. The scope of this PR is to add spitting of the PRCurve, To change parameter types or names will mean refactoring the original Evaluate(..) method which is out oof scope for this PR

---
In reply to: [268282834](https://github.com/dotnet/machinelearning/pull/3039#discussion_r268282834) [](ancestors = 268282834)

src/Microsoft.ML.StaticPipe/EvaluatorStaticExtensions.cs
ganik: this would mean refactoring both Evaluate(..) on lines 28 & 102. As these new methods are based off them. I dont want to make bigger changes at this point and just want to follow the established pattern here to minimize risks

---
In reply to: [268285448](https://github.com/dotnet/machinelearning/pull/3039#discussion_r268285448) [](ancestors = 268285448)

src/Microsoft.ML.StaticPipe/EvaluatorStaticExtensions.cs
ganik: this would mean unsealing CalibratedBinaryClassificationMetrics class, its sealed currently

---
In reply to: [268283796](https://github.com/dotnet/machinelearning/pull/3039#discussion_r268283796) [](ancestors = 268283796)

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
TomFinley: Incidentally @yaeldekel is correct, the original command line didn't *need* it -- it would use it if required, but it would not insist on it. That we have it for the API is somewhat unfortunate, but not that big of a deal. In future we could in fact improve this story: we could have *this* overload, but also another overload that *doesn't* take it, and takes a threshold for it. Thereby capturing the existing functionality of this C# API, while retaining the "flexibility" we used to achieve in our command line.

That however is beyond the scope of @ganik's work here, and not something we should add now.

src/Microsoft.ML.Data/Evaluators/Metrics/PrecisionRecallMetrics.cs
ganik: changed name to reflect its not a metrics but data point for binary classifications

---
In reply to: [268281758](https://github.com/dotnet/machinelearning/pull/3039#discussion_r268281758) [](ancestors = 268281758)

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
ganik: thx

---
In reply to: [268291075](https://github.com/dotnet/machinelearning/pull/3039#discussion_r268291075) [](ancestors = 268291075)

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
ganik: thank you for explanation.

---
In reply to: [268327749](https://github.com/dotnet/machinelearning/pull/3039#discussion_r268327749) [](ancestors = 268327749)

ganik: test is added

---
In reply to: [268037217](https://github.com/dotnet/machinelearning/pull/3039#discussion_r268037217) [](ancestors = 268037217,267591483)

rogancarr: How much extra work is an ROC Curve on top of getting the PR Curve? It's weird to have one but not the other from the API.

---
In reply to: [268027498](https://github.com/dotnet/machinelearning/pull/3039#discussion_r268027498) [](ancestors = 268027498,267921726,267839855)

ganik: there is no code internally that calculates ROC. PR data has been already calculated internally and in this PR i jsut expose it. For ROC i ll need to implement code that calculates it, I expect it to be few days change (including reviews) and should be done in separate PR

---
In reply to: [268340685](https://github.com/dotnet/machinelearning/pull/3039#discussion_r268340685) [](ancestors = 268340685,268027498,267921726,267839855)

src/Microsoft.ML.Data/Evaluators/Metrics/BinaryPrecisionRecallDataPoint.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// Gets the true positive rate for the current threshold.
+        /// </summary>
+        public double TruePositiveRate => Recall;
+
+        /// <summary>
+        /// Gets the false positive rate for the given threshold.
+        /// </summary>
+        public double FalsePositiveRate { get; }
+
+        internal BinaryPrecisionRecallDataPoint(ValueGetter<float> thresholdGetter, ValueGetter<double> precisionGetter, ValueGetter<double> recallGetter, ValueGetter<double> fprGetter)
```
TomFinley: >ValueGetter<double> precisionGetter [](start = 84, length = 35)

I would recommend that all logic persuant to the getters remain in the same block of code as the cursor (anything else is just a pointless increase of complexity), but this at least avoids the main problem so, I guess we can fix it later.

src/Microsoft.ML.Data/Evaluators/Metrics/BinaryPrecisionRecallDataPoint.cs
```diff
@@ -0,0 +1,54 @@
...
+        public double Precision { get; }
+        /// <summary>
+        /// Gets the recall for the current threshold.
+        /// </summary>
+        public double Recall { get; }
+
+        /// <summary>
+        /// Gets the true positive rate for the current threshold.
+        /// </summary>
+        public double TruePositiveRate => Recall;
```
TomFinley: >public double TruePositiveRate => Recall; [](start = 8, length = 41)

So, what is the point of this?

ganik: This is to make life easier for users to plot ROC curve. ROC curve is plotted as TPR vs FPR. We do have FPR, but spent some time with Rogan looking for TPR in order to get ROC data. It appears Recall is synonymous with TPR

---
In reply to: [268370748](https://github.com/dotnet/machinelearning/pull/3039#discussion_r268370748) [](ancestors = 268370748)

src/Microsoft.ML.Data/Evaluators/Metrics/BinaryPrecisionRecallDataPoint.cs
ganik: sure, we can do it in this or in a separate PR.

---
In reply to: [268370725](https://github.com/dotnet/machinelearning/pull/3039#discussion_r268370725) [](ancestors = 268370725)

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
```diff
@@ -828,6 +828,93 @@ public CalibratedBinaryClassificationMetrics Evaluate(IDataView data, string lab
...
+            out ValueGetter<float> thresholdGetter,
+            out ValueGetter<double> precisionGetter,
+            out ValueGetter<double> recallGetter,
+            out ValueGetter<double> fprGetter)
+        {
+            var thresholdColumn = prCurveView.Schema.GetColumnOrNull(BinaryClassifierEvaluator.Threshold);
+            var precisionColumn = prCurveView.Schema.GetColumnOrNull(BinaryClassifierEvaluator.Precision);
+            var recallColumn = prCurveView.Schema.GetColumnOrNull(BinaryClassifierEvaluator.Recall);
+            var fprColumn = prCurveView.Schema.GetColumnOrNull(BinaryClassifierEvaluator.FalsePositiveRate);
+            Host.Assert(thresholdColumn != null);
```
TomFinley: >Host.Assert [](start = 12, length = 11)

In future, consider `AssertValue` to make this a bit more elegant.

test/Microsoft.ML.StaticPipelineTesting/Training.cs
```diff
@@ -18,6 +18,7 @@
 using Microsoft.ML.Trainers.Recommender;
 using Xunit;
 using Xunit.Abstractions;
+using System.Collections.Generic;
```
TomFinley: We generally like to keep our namespaces sorted.

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
ganik: tried it, didnt work here bcs DataViewSchema.Column is not a reference type, its a struct

---
In reply to: [268380273](https://github.com/dotnet/machinelearning/pull/3039#discussion_r268380273) [](ancestors = 268380273)

PR #3041, opened 2019-03-21
Data catalog done

sfilipi:
Towards #1209 

Adding a sample for the `Hash` extension. 
Disabling the compiler check for whether the members of a class are in use.


sfilipi:
Closing because the branch used has stale changes.

PR #3044, opened 2019-03-21
Model

### LINKED ISSUE #3025: About models, ITransform, and IDataLoader, saving/loading
So, I had a conversation with @eerhardt and @yaeldekel, about the nature of models, in particular relating to the saving and loading routines. This is very important for us to get right, since the artifacts of what we learn and how we transform data, and their persistability, is probably the most important thing we have to do correctly.

We take the view initially that the model is the `ITransformer` (note that a chain of `ITransformer`s is itself an `ITransformer`). But, by itself this is an insufficient description, was we saw in in #2663 and its subsequent "child" #2735, from the point of view of model being practically "the stuff you need to keep," there's a lot more to a machine learning model than merely just the `ITransformer` chain -- you also need to preserve some notion of what the input is to that. So we added these things to take either a loader, or the input schema, to be saved as part of the model.

Yet, is the loader a model itself? Sometimes that's precisely what we call it:

https://github.com/dotnet/machinelearning/blob/c38f81b3957fed6aa88ea0e6b295522d5bf3f9ec/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs#L36

And in the same file we call it something else:

https://github.com/dotnet/machinelearning/blob/c38f81b3957fed6aa88ea0e6b295522d5bf3f9ec/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs#L65

It is a model in one sense because it is yet another things that takes input data and produces output data -- the fact that `ITransformer` does it specifically over `IDataView` as input *specifically* does not give it some magical, special status to allow it to be called a model, and yet not. If I take a loader, and append a transform to it, then the whole aggregate thing is *still* a loader. If it isn't a model, it only isn't one by the mere skin of its teeth. Hence the presence of the original thing, and why we have in the model operations catalog operations to save and load `IDataLoader` itself specifically.

But at the same time this duality of the term "model" is, as I understand @eerhardt, confusing. We have two things we're calling model. In an ideal world, I feel like if we *can* get away with just one story of what the model is, we should take it, and if there must be only one it must be `ITransformer`. We even have the situation where if you have `mlContext.Model.Save(`, the first thing that pops up is the `IDataLoader` thing, which is kind of strange.

I am not quite sure what I think, since in this case I agree with whoever talks to me last with an even a vaguely convincing argument. But I think in this case I will see about getting rid of the `IDataLoader`-only APIs -- people can, if it is important, continue to save and load such things by using empty `ITransformer` chains (again, any chain of `ITransformer` is itself an `ITransformer`, including the empty chain).

Since we are approaching v1, I view it as a bit more important to be conservative w.r.t. potentially confusing additions to the API, especially around something as central as the saving and loading of models. We might be able to add it back later if there's some really compelling scenario for them, that we somehow did not anticipate.

We will of course retain the saving and loading of transformers *with* loaders, since that is really important to be able to capture, but I think being consistent around the story that "models are transformers" as we are most places is kind of important.

TomFinley:
Fixes #3025.


src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -28,64 +28,69 @@ internal ModelOperationsCatalog(IHostEnvironment env)
...
+        /// Save a transformer model and the loader used to create its input data to the stream.
         /// </summary>
-        /// <param name="model">The trained model to be saved.</param>
+        /// <param name="model">The trained model to be saved. Note that this can be <see langword="null"/>, as a shorthand
+        /// for an empty data set. Upon loading with <see cref="LoadWithDataLoader(Stream, out IDataLoader{IMultiStreamSource})"/>
+        /// the returned valued will be an empty <see cref="TransformerChain{TLastTransformer}"/>.</param>
+        /// <param name="loader">The loader that was used to create data to train the model.</param>
         /// <param name="stream">A writeable, seekable stream to save to.</param>
-        public void Save<TSource>(IDataLoader<TSource> model, Stream stream)
+        public void Save<TSource>(ITransformer model, IDataLoader<TSource> loader, Stream stream)
```
TomFinley: Major things on the top of my mind, the save/load methods are still inconsistent w.r.t. where they put the file or stream argument... here we see, it is the *last* argument, whereas during loading it is the *first* argument.

One possibility is we leave it as is, since in this case it is where the model is going (so maybe last is most clear), whereas during loading it is where the model is coming from.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -28,64 +28,69 @@ internal ModelOperationsCatalog(IHostEnvironment env)
...
         /// </summary>
-        /// <param name="loader">The loader that was used to create data to train the model</param>
-        /// <param name="model">The trained model to be saved</param>
+        /// <param name="model">The trained model to be saved. Note that this can be <see langword="null"/>, as a shorthand
+        /// for an empty data set. Upon loading with <see cref="LoadWithDataLoader(Stream, out IDataLoader{IMultiStreamSource})"/>
+        /// the returned valued will be an empty <see cref="TransformerChain{TLastTransformer}"/>.</param>
+        /// <param name="loader">The loader that was used to create data to train the model.</param>
         /// <param name="filePath">Path where model should be saved.</param>
-        public void Save<TSource>(IDataLoader<TSource> loader, ITransformer model, string filePath)
+        public void Save<TSource>(ITransformer model, IDataLoader<TSource> loader, string filePath)
```
TomFinley: I'm also curious about whether we agree that what is effectively the descriptor of input to the model should come after the model itself. The API had been inconsistent, in that in one save method we had loader/model (so input to model came first), whereas in another we had model/schema (so input to model came second).

In all cases I feel like the load methods should always just return `ITransform`, so maybe just always putting `ITransform` 
front-and-center for saving was probably the best approach, so I resolved the inconsistency that way, but I don't insist on it.

I kept it as a separate commit in case we wanted to go the other way, maybe... so in that case, the schema.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -28,64 +28,69 @@ internal ModelOperationsCatalog(IHostEnvironment env)
...
+        public void Save<TSource>(ITransformer model, IDataLoader<TSource> loader, Stream stream)
         {
-            _env.CheckValue(model, nameof(model));
+            _env.CheckValue(loader, nameof(loader));
+            _env.CheckValueOrNull(model);
             _env.CheckValue(stream, nameof(stream));
 
+            // For the sake of consistency of this API specifically, when called upon we save any transformer
+            // in a single element transformer chain.
+            var chainedModel = model == null ? null : new TransformerChain<ITransformer>(model);
```
yaeldekel: >chainedModel [](start = 16, length = 12)

Should you use the method introduced above (`CreateChain`)? That method creates an empty chain when `model==null`, and here you set it to null, is that intentional?

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -28,64 +28,69 @@ internal ModelOperationsCatalog(IHostEnvironment env)
...
+        /// <param name="model">The trained model to be saved. Note that this can be <see langword="null"/>, as a shorthand
+        /// for an empty data set. Upon loading with <see cref="LoadWithDataLoader(Stream, out IDataLoader{IMultiStreamSource})"/>
+        /// the returned valued will be an empty <see cref="TransformerChain{TLastTransformer}"/>.</param>
+        /// <param name="loader">The loader that was used to create data to train the model.</param>
         /// <param name="filePath">Path where model should be saved.</param>
-        public void Save<TSource>(IDataLoader<TSource> loader, ITransformer model, string filePath)
+        public void Save<TSource>(ITransformer model, IDataLoader<TSource> loader, string filePath)
         {
+            _env.CheckValueOrNull(model);
+            _env.CheckValue(loader, nameof(loader));
```
artidoro: nit: It looks like these checks will be performed in the method above anyways. Should we remove them from here?

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -28,64 +28,69 @@ internal ModelOperationsCatalog(IHostEnvironment env)
             _env = env;
         }
 
+        private static TransformerChain<ITransformer> CreateChain(ITransformer transformer)
```
artidoro: >CreateChain [](start = 54, length = 11)

I can't find it, where are you are using this method?

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
yaeldekel: I kind of like this possibility.
If we're going with the "intuitive" ordering, then maybe `loader` should come before `model`?

---
In reply to: [267790407](https://github.com/dotnet/machinelearning/pull/3044#discussion_r267790407) [](ancestors = 267790407)

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -28,64 +28,69 @@ internal ModelOperationsCatalog(IHostEnvironment env)
...
+            return new TransformerChain<ITransformer>(transformer);
+        }
+
         /// <summary>
-        /// Save the model to the stream.
+        /// Save a transformer model and the loader used to create its input data to the stream.
         /// </summary>
-        /// <param name="model">The trained model to be saved.</param>
+        /// <param name="model">The trained model to be saved. Note that this can be <see langword="null"/>, as a shorthand
+        /// for an empty data set. Upon loading with <see cref="LoadWithDataLoader(Stream, out IDataLoader{IMultiStreamSource})"/>
```
yaeldekel: >data set [](start = 25, length = 8)

transformer chain?

test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamplesDynamicApi.cs
```diff
@@ -120,15 +120,13 @@ private void TrainRegression(string trainDataPath, string testDataPath, string m
...
             mlContext.Model.Save(model, trainData.Schema, modelPath);
 
             // Potentially, the lines below can be in a different process altogether.
-
-            // When you load the model, it's a 'dynamic' transformer. 
-            ITransformer loadedModel;
-            using (var stream = File.OpenRead(modelPath))
-                loadedModel = mlContext.Model.Load(stream, out var schema);
+            // When you load the model, it's a non-specific ITransformer. We also recover
+            // the original schema.
```
glebuk: >schema [](start = 28, length = 6)

the original input schema (to disambiguate with the output schema)

glebuk:
Should we also update ConvertToOnnx methods in the same way

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -171,57 +182,100 @@ public ITransformer Load(Stream stream, out DataViewSchema inputSchema)
...
+                // that sort, and so return it.
+                var accessor = (ITransformerChainAccessor)chain;
+                if (accessor.Transformers.Length == 1)
+                    return accessor.Transformers[0];
+                // If it is some other length than 1 due to, say, some legacy model saving, just return that
+                // chain. Using the above API this is not possible, since the chain saved will always be of length
+                // one, but older APIs behaved differently so we should retain flexibility with those schemes.
+                // (Those schemes are BTW by no means incorrect, they just aren't what the API in this particular
+                // class will specifically do.)
+                return composite.Transformer;
```
yaeldekel: >composite.Transformer [](start = 23, length = 21)

nit: you can return `chain` here.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
TomFinley: Yes I considered it. Actually intuitive cuts two ways here, since this `Save` method interacts with intellisense in a way that I feel makes this arrangement most natural. But I don't insist on it.

test/Microsoft.ML.Functional.Tests/ModelLoading.cs
```diff
@@ -16,7 +16,7 @@
 
 namespace Microsoft.ML.Functional.Tests
 {
-    public partial class ModelLoadingTests : BaseTestClass
+    public partial class ModelLoadingTests : TestDataPipeBase
     {
         private MLContext _ml;
```
yaeldekel: >_ml [](start = 26, length = 3)

`TestDataPipeBase` may have an `MLContext` field.

test/Microsoft.ML.Functional.Tests/ModelLoading.cs
```diff
@@ -74,35 +81,49 @@ public void LoadModelAndExtractPredictor()
             {
                 // This model can be loaded either as a composite data loader,
                 // a transformer model + an input schema, or a transformer model + a data loader.
-                var t = _ml.Model.LoadWithDataLoader(fs, out IDataLoader<IMultiStreamSource> l);
+                var t = _ml.Model.LoadWithDataLoader(fs, out loadedCompositeLoader);
+                // This is a bit strange, as it seems to test that it can reload from the same
+                // stream twice opened only once, which as far as I know is not really a requirement
```
yaeldekel: It also tests that you can use the same file to load using both `LoadWithDataLoader` and `Load` (which loads with the schema). Maybe we should open a separate stream for that then, to make sure that if it fails we know that it's a real problem and not just a problem with reading twice from the same stream?

test/Microsoft.ML.Functional.Tests/ModelLoading.cs
```diff
@@ -74,35 +81,49 @@ public void LoadModelAndExtractPredictor()
...
-                as CalibratedModelParametersBase).SubModel
-                as GamBinaryModelParameters;
-            Assert.NotNull(gam);
-
-            gam = (((loadedTransformerModel1 as TransformerChain<ITransformer>).LastTransformer
-                as ISingleFeaturePredictionTransformer<object>).Model
-                as CalibratedModelParametersBase).SubModel
-                as GamBinaryModelParameters;
-            Assert.NotNull(gam);
+            void AssertIsGam(ITransformer trans)
```
yaeldekel: >void [](start = 12, length = 4)

Interesting. I didn't know this can be done :-).

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
TomFinley: My addition is deliberate and we should try to do this more often if we aren't already. It's useful for errors about method misuse to happen "close" to the user. So I corrected it here.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
TomFinley: Ah thanks I'll nuke it

test/Microsoft.ML.Functional.Tests/ModelLoading.cs
yaeldekel: By the way, we already have overloads for loading that take file names, don't we?

---
In reply to: [267891929](https://github.com/dotnet/machinelearning/pull/3044#discussion_r267891929) [](ancestors = 267891929)

test/Microsoft.ML.Functional.Tests/ModelLoading.cs
TomFinley: Ah good point thanks.

test/Microsoft.ML.Functional.Tests/ModelLoading.cs
TomFinley: Sure! I changed some tests to use those so both paths are tested, but this seems like a fine place to use them.

test/Microsoft.ML.Functional.Tests/ModelLoading.cs
TomFinley: New in one of the c# 7 versions... Good for tests especially I've found when you have lots of this only locally useful but often repeated here code.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
TomFinley: It was more a situation that I changed my mind but forgot to get rid of the method I'd written

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -28,64 +28,69 @@ internal ModelOperationsCatalog(IHostEnvironment env)
...
+        }
+
         /// <summary>
-        /// Save the model to the stream.
+        /// Save a transformer model and the loader used to create its input data to the stream.
         /// </summary>
-        /// <param name="model">The trained model to be saved.</param>
+        /// <param name="model">The trained model to be saved. Note that this can be <see langword="null"/>, as a shorthand
+        /// for an empty data set. Upon loading with <see cref="LoadWithDataLoader(Stream, out IDataLoader{IMultiStreamSource})"/>
+        /// the returned valued will be an empty <see cref="TransformerChain{TLastTransformer}"/>.</param>
```
abgoswam: >valued [](start = 25, length = 6)

typo

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
eerhardt: One thing that I find a little bit odd about this behavior is when a user calls:

```
ITransformer model = ...;
mlContext.Model.Save(model, null, filePath);
```

The `null` argument in the 2nd argument is going to be ambiguous between the two overloads, and they will need to explicitly pick one.

If they choose the `DataViewSchema` overload, the above code will succeed.
If they choose the `IDataLoader` overload, it will throw an ArgumentNullException.

This probably isn't a huge issue. Just something I thought I'd bring up for discussion.

I almost wonder if we should have a `mlContext.ModelSave(model, filePath)` overload that doesn't require the loader or schema.

TomFinley:
> Should we also update ConvertToOnnx methods in the same way?

Hi @glebuk. Definitely not in scope, and possibly more involved than you might think. Anyway, ONNX is one of the nugets to be released in preview mode.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
TomFinley: > This probably isn't a huge issue. Just something I thought I'd bring up for discussion.

It is interesting @eerhardt, but let's remember, as per the discussion of #2663, that this scenario is extremely rare. Most of the time people want to save a complete model, not just bits and pieces of it. (I *did* notice that though, and we do have test code that exercises it as you will see, and due to the way the overload resolution works, it does in fact do the right thing.)

> I almost wonder if we should have a `mlContext.ModelSave(model, filePath)` overload that doesn't require the loader or schema.

That would be a repetition of the exact mistake that led to #2663, and the necessary changes in this API. I *want* the suggestion that the input schema (whether by the schema or the loader that generates the schema) to be right in the user's face. We know what happens when that suggestion is not there, because the situation where you don't want to do that is actually quite rare.

[I had a comment](https://github.com/dotnet/machinelearning/pull/2858#discussion_r262992588) in #2858 where I brought up why this was a mistake... to summarize and rephrase, the danger is, if you have a simpler looking API, people will tend to *use* that simpler looking API. (Because, heck, it looks simpler.) That's *very* dangerous if that simpler looking API does something that is wrong in 99.9% of situations. Even more dangerous if it's not *obvious* it is doing the wrong thing -- and this was *so* non-obvious that it was doing the wrong thing that @Ivanidzo4ka filed an issue and even went so far as to label it as an actual bug, when he's probably one of the people that knows our architecture the best.

No, this suggestion, "here's where you save the starting point of your model," is a really, really important thing, and I would consider acting on this specific suggestion a mistake.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
TomFinley: Thanks @abgoswam ... fixed this `valued` here and in three other places.

test/Microsoft.ML.Functional.Tests/ModelLoading.cs
TomFinley: I will have two commits in my next push -- one to resolve other comments, and one to do this rename, since it winds up being a little annoying and distracting in this file.

test/Microsoft.ML.Functional.Tests/ModelLoading.cs
TomFinley: Um, sorry, nevermind -- this is a little annoying dealing with all the local variables and changing of scope. Let's just leave it for now...

test/Microsoft.ML.Functional.Tests/ModelLoading.cs
TomFinley: Actually, no sorry, I don't have time to do this right now and need to run to another meeting, so I'll just leave as is and fix later (since I think other more "public facing" changes might be blocking the PR more than this).

test/Microsoft.ML.Functional.Tests/ModelLoading.cs
```diff
@@ -220,25 +248,33 @@ public void SaveCompositeLoaderAndLoad()
             var model = composite.Fit(file);
 
             string modelPath = GetOutputPath(FullTestName + "-model.zip");
-            _ml.Model.Save(model, modelPath);
+            _ml.Model.Save(null, model, modelPath);
```
eerhardt: should this be `_ml.Model.Save(model, null, modelPath);` ?

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
eerhardt: > because the situation where you don't want to do that is actually quite rare.

> That's very dangerous if that simpler looking API does something that is wrong in 99.9% of situations.

The part of this whole situation that has me confused is:

How was anything working previously? People are saving models today without the loader/schema, and they are successfully using the models to make predictions in their applications.

I get that there are scenarios where you will need the loader/schema information. But is it _really_ 99.9% of the time where you _need_ these things? How are people successfully using the Save/Load APIs today without it? Are they really into the 0.1% use case?

eerhardt: I get that we want to lead people down a "pit of success" - which is ensuring that they at least know that they should be saving the loader/schema. So I'll resolve this conversation. I get the concern. Thanks for the discussion.

TomFinley: So, I am on my phone so I can't find them now. The fact that we were failing to save the data loader has been a problem since the v0.1 API.

The fact that it appears to work sometimes is what makes it so pernicious. You don't need the input schema until you do, then if you don't have it you can't do simple things like figure out how many features a model expects, what the names were, etc. Then that information is not there unless you reconstruct the input... Which all of our samples happen to do.

Being able to get the output schema of a model without knowing what the input was is really important...

test/Microsoft.ML.Functional.Tests/ModelLoading.cs
TomFinley: No in this case the test is calling the loader model. I agree it's confusing but fixing up naming in a test was not my priority.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -29,63 +29,60 @@ internal ModelOperationsCatalog(IHostEnvironment env)
...
-        public void Save<TSource>(IDataLoader<TSource> loader, ITransformer model, Stream stream) =>
-            Save(new CompositeDataLoader<TSource, ITransformer>(loader, new TransformerChain<ITransformer>(model)), stream);
-
         /// <summary>
         /// Save a transformer model and the loader used to create its input data to the file.
         /// </summary>
-        /// <param name="loader">The loader that was used to create data to train the model</param>
-        /// <param name="model">The trained model to be saved</param>
+        /// <param name="model">The trained model to be saved. Note that this can be <see langword="null"/>, as a shorthand
+        /// for an empty data set. Upon loading with <see cref="LoadWithDataLoader(Stream, out IDataLoader{IMultiStreamSource})"/>
```
yaeldekel: >data set [](start = 25, length = 8)

Oops. "transformer chain" here too? And in the other two overloads of `Save` too.

test/Microsoft.ML.Functional.Tests/ModelLoading.cs
TomFinley: OK! I have finally done this, sorry for not doing this earlier @yaeldekel but it was one of these situations where i had literally 30 seconds before I had to go somewhere and the change would take me 1 minute. 😄 But I just got that 1 minute.

test/Microsoft.ML.Functional.Tests/ModelLoading.cs
TomFinley: OK, got some time, changed the name here... now it's `loaderWithEmbeddedModel`.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
TomFinley: D'oh!

Issue #3045, opened 2019-03-21
DataViewType ToString should not use old DataKind names


eerhardt:
All the DataViewType ToString methods are returning DataKind names instead of "normal .NET" type names.

```
I4
U8
R4
etc.
```

https://github.com/dotnet/machinelearning/blob/5b22420d28c0cacc9b265d043555b6d11a017b91/src/Microsoft.ML.DataView/DataViewType.cs#L203-L222

We should change these to use typical .NET type names, like we have in the public API.

Note: there are a bunch of tests that will need to be fixed with this since it baselines these names in baseline files.

Issue #3037, opened 2019-03-20
Getting started with ML .NET with in-memory data is *painful*.


isaacabraham:
Working with latest F#4.5 and net standard I'm having huge problems trying to do even the most basic explorations with the latest ML .NET. Is there any example showing an absolutely basic example for an in-memory dataset using a simple ML algorithm?

I'm talking something as simple as an example from e.g. [scikit-learn](https://mcalglobal.com/2018/02/22/machine-learning-hello-world-using-python/) e.g. the following hello world is 7 lines of code, and if you leave at the data loading side of things and just focus on the ML side of things - which is exactly what I want to do - it's the following *three lines of code*.

```
model = linear_model.LinearRegression()
model.fit(sqfeet, price)
model.predict( pd.DataFrame([1750]))
```

Lets try and port this into F#. Here's the source data as a simple F# list.

```fsharp
type Observation = { Area:int; Price:int }
let data =
    [ { Area = 1100; Price = 119000 }
      { Area = 1200; Price = 126000 }
      { Area = 1300; Price = 133000 }
      { Area = 1400; Price = 150000 }
      { Area = 1500; Price = 161000 }
      { Area = 1600; Price = 163000 }
      { Area = 1700; Price = 169000 }
      { Area = 1800; Price = 182000 }
      { Area = 1900; Price = 201000 }
      { Area = 2000; Price = 209000 } ]
```

I've spent a good few hours fighting with the API to try and get some - any - results. **I can't figure it out**.

Issues I've encountered:

1. Discoverability. The API is pretty large and not (in my personal opinion) easy to navigate your way around. The namespaces need to be reworked so that the most obvious types are easy and obvious to get to.
1. F# scripts are a pain because of the "occasional" reliance on native DLLs. However, you can work around this (or fall back to console applications if needed).
1. Error messages are painful - `I4`, `R4` etc. etc. - most people will not know what these are.
1. Vector types - it seems that in order to "use" data with a trainer you need to "convert" data from e.g. `float32` into a "vector" of `float32`. There's no explanation of what a "vector" in the context of ML .NET is, nor how to create one. Is it a .NET type? How do I create it? More than that, why as a developer should I have to care about it? I just want to give some of my data to the library as quickly and easily as possible.
1. Why do I need to convert from ints or floats into float32s to do some machine learning? Again, this raises the barrier to entry. This is an internal implementation detail of ML .NET, it's nothing that should be forced on the developer.
[...]

PR #3046, opened 2019-03-21
Attempt to fix error messages types

Ivanidzo4ka:
fixes https://github.com/dotnet/machinelearning/issues/3045

https://github.com/dotnet/machinelearning/issues/3037
>Error messages are painful - I4, R4 etc. etc. - most people will not know what these are.

Let's see is it so easy as I think, or I need to change all baseline


src/Microsoft.ML.Data/Data/RowCursorUtils.cs
```diff
@@ -341,7 +341,7 @@ public static string TestGetLabelGetter(DataViewType type, bool allowKeys)
             if (allowKeys && type is KeyType)
                 return null;
 
-            return allowKeys ? "Expected R4, R8, Bool or Key type" : "Expected R4, R8 or Bool type";
+            return allowKeys ? "Expected Single, Double, Boolean or Key type" : "Expected Single, RDouble8 or Boolean type";
```
Ivanidzo4ka: Double

src/Microsoft.ML.Data/Evaluators/RankingEvaluator.cs
```diff
@@ -89,13 +89,13 @@ private protected override void CheckScoreAndLabelTypes(RoleMappedSchema schema)
             if (t != NumberDataViewType.Single && !(t is KeyType))
             {
                 throw Host.ExceptSchemaMismatch(nameof(RankingMamlEvaluator.Arguments.LabelColumn),
-                    "label", schema.Label.Value.Name, "R4 or a key", t.ToString());
+                    "label", schema.Label.Value.Name, "Single or a key", t.ToString());
```
Ivanidzo4ka: Capitalize

test/BaselineOutput/SingleDebug/SavePipe/PartitionedNamedDirectories-Schema.txt
```diff
@@ -1,5 +1,5 @@
 ---- PartitionedFileLoader ----
 3 columns:
-  L0: Text
-  Year: Text
-  Month: Text
+  L0: String
```
Ivanidzo4ka: I'll move them to Common folder after successful build run.
Otherwise it's gonna be messy PR.

test/BaselineOutput/SingleDebug/SavePipe/SavePipeSlidingWindow-Schema.txt
```diff
@@ -1,7 +1,7 @@
 ---- BoundLoader ----
 1 columns:
-  Input: R4
+  Input: Single
 ---- SlidingWindowTransform ----
 2 columns:
-  Input: R4
-  Output: Vec<R4>
+  Input: Single
+  Output: Vec<Single>
```
Ivanidzo4ka: >Vec [](start = 10, length = 3)

Should I name you Vector?

src/Microsoft.ML.Data/Transforms/NormalizeColumn.cs
```diff
@@ -400,7 +400,7 @@ public static AffineColumnFunction Create(ModelLoadContext ctx, IHost host, Data
                     if (vectorType.ItemType == NumberDataViewType.Double)
                         return Dbl.ImplVec.Create(ctx, host, vectorType);
                 }
-                throw host.ExceptUserArg(nameof(AffineArgumentsBase.Columns), "Wrong column type. Expected: R4, R8, Vec<R4, n> or Vec<R8, n>. Got: {0}.", typeSrc.ToString());
+                throw host.ExceptUserArg(nameof(AffineArgumentsBase.Columns), "Wrong column type. Expected: Single, Double, Vector<Single, n> or Vector<Double, n>. Got: {0}.", typeSrc.ToString());
```
yaeldekel: >Vector [](start = 124, length = 6)

In other places we write something like "known-size vector of ...". Might be a bit less confusing than having this mysterious "n" here.

src/Microsoft.ML.DataView/DataViewType.cs
```diff
@@ -128,7 +128,7 @@ public static NumberDataViewType Byte
             get
             {
                 return _instByte ??
-                    Interlocked.CompareExchange(ref _instByte, new NumberDataViewType(typeof(byte), "U1"), null) ??
+                    Interlocked.CompareExchange(ref _instByte, new NumberDataViewType(typeof(byte), "ushort"), null) ??
```
Ivanidzo4ka: >ushort [](start = 101, length = 6)

byte and sbyte

src/Microsoft.ML.Data/Transforms/NormalizeColumn.cs
Ivanidzo4ka: It doesn't have to be known size, does this one look ok?

---
In reply to: [267931607](https://github.com/dotnet/machinelearning/pull/3046#discussion_r267931607) [](ancestors = 267931607)

docs/samples/Microsoft.ML.Samples/Dynamic/LdaTransform.cs
```diff
@@ -11,7 +11,7 @@ public static void Example()
             // Create a new ML context, for ML.NET operations. It can be used for exception tracking and logging, 
             // as well as the source of randomness.
             var ml = new MLContext();
-
+
```
Ivanidzo4ka: random change is random

forki:
Wow that's a big PR to improve a error message

src/Microsoft.ML.Data/Transforms/NormalizeColumn.cs
yaeldekel: Oh, you're right, the code above doesn't require the vector to have a known size :-).

But, isn't this a bug? How can a normalizer be defined on a column of unknown size?


---
In reply to: [267945509](https://github.com/dotnet/machinelearning/pull/3046#discussion_r267945509) [](ancestors = 267945509,267931607)

yaeldekel: Also, I just noticed that this is the method that loads from `ModelLoadContext`, shouldn't it be `ExceptDecode`?


---
In reply to: [267972968](https://github.com/dotnet/machinelearning/pull/3046#discussion_r267972968) [](ancestors = 267972968,267945509,267931607)

src/Microsoft.ML.DataView/DataViewType.cs
```diff
@@ -94,7 +94,7 @@ public override bool Equals(DataViewType other)
             return false;
         }
 
-        public override string ToString() => "Text";
+        public override string ToString() => "String";
```
TomFinley: >String [](start = 46, length = 6)

This one is a little tricky, since of course the type as we see a few lines above is *not* `String` at all, but rather `ReadOnlyMemory<char>`.

src/Microsoft.ML.Data/Transforms/FeatureContributionCalculationTransformer.cs
```diff
@@ -218,7 +218,7 @@ public Mapper(FeatureContributionCalculatingTransformer parent, DataViewSchema s
                     throw Host.ExceptSchemaMismatch(nameof(schema), "input", _parent.ColumnPairs[0].inputColumnName);
                 _featureColumnType = schema[_featureColumnIndex].Type as VectorType;
                 if (_featureColumnType == null || _featureColumnType.ItemType != NumberDataViewType.Single)
-                    throw Host.ExceptSchemaMismatch(nameof(schema), "feature", _parent.ColumnPairs[0].inputColumnName, "vector of float.", _featureColumnType.ItemType.ToString());
+                    throw Host.ExceptSchemaMismatch(nameof(schema), "feature", _parent.ColumnPairs[0].inputColumnName, "Vector of Single", _featureColumnType.ItemType.ToString());
```
TomFinley: >Vector [](start = 120, length = 6)

While `Single` is appropriate, I think we generally call these `vector` in the context of a larger sentence.

So I would say "I want a vector of Single," I would not say "I want a Vector of Single."

src/Microsoft.ML.Data/Transforms/NormalizeColumnDbl.cs
```diff
@@ -932,7 +932,7 @@ public ImplVec(IHost host, TFloat[] mean, TFloat[] stddev, bool useLog)
 
                     public static ImplVec Create(ModelLoadContext ctx, IHost host, VectorType typeSrc)
                     {
-                        host.Check(typeSrc.ItemType.RawType == typeof(TFloat), "The column type must be vector of R8.");
+                        host.Check(typeSrc.ItemType.RawType == typeof(TFloat), "The column type must be vector of Double.");
```
TomFinley: >vector [](start = 104, length = 6)

Here's an example of a place where we've left it lowercase `vector`. Not sure what approach you want to take here.

src/Microsoft.ML.Data/Transforms/RangeFilter.cs
```diff
@@ -153,7 +153,7 @@ private RangeFilter(IHost host, ModelLoadContext ctx, IDataView input)
 
             _type = schema[_index].Type;
             if (_type != NumberDataViewType.Single && _type != NumberDataViewType.Double && _type.GetKeyCount() == 0)
-                throw Host.ExceptSchemaMismatch(nameof(schema), "source", column, "float, double or KeyType", _type.ToString());
+                throw Host.ExceptSchemaMismatch(nameof(schema), "source", column, "Single, Double or KeyType", _type.ToString());
```
TomFinley: >KeyType [](start = 101, length = 7)

So, this is an interesting one... the *type* is of key, but we are calling it `KeyType`... yet, we say `Single` not `SingleType`. Probably not worth resolving now, but something to think about in future, no doubt.

src/Microsoft.ML.DataView/DataViewType.cs
```diff
@@ -117,7 +117,7 @@ public static NumberDataViewType SByte
             get
             {
                 return _instSByte ??
-                    Interlocked.CompareExchange(ref _instSByte, new NumberDataViewType(typeof(sbyte), "I1"), null) ??
+                    Interlocked.CompareExchange(ref _instSByte, new NumberDataViewType(typeof(sbyte), "sbyte"), null) ??
```
TomFinley: >sbyte [](start = 103, length = 5)

We are elsewhere using the .NET type names for these things, are we not? So should this be `SByte`, not `sbyte`

src/Microsoft.ML.DataView/DataViewType.cs
```diff
@@ -128,7 +128,7 @@ public static NumberDataViewType Byte
             get
             {
                 return _instByte ??
-                    Interlocked.CompareExchange(ref _instByte, new NumberDataViewType(typeof(byte), "U1"), null) ??
+                    Interlocked.CompareExchange(ref _instByte, new NumberDataViewType(typeof(byte), "byte"), null) ??
```
TomFinley: >byte [](start = 101, length = 4)

So, by a similar token, ought this to be `Byte`?

src/Microsoft.ML.DataView/DataViewType.cs
```diff
@@ -150,7 +150,7 @@ public static NumberDataViewType UInt16
             get
             {
                 return _instUInt16 ??
-                    Interlocked.CompareExchange(ref _instUInt16, new NumberDataViewType(typeof(ushort), "U2"), null) ??
+                    Interlocked.CompareExchange(ref _instUInt16, new NumberDataViewType(typeof(ushort), "UIUnt16"), null) ??
```
TomFinley: >UIUnt16 [](start = 105, length = 7)

Type.

TomFinley: Note that if you had, say, done `nameof(UInt16)` you wouldn't have had this problem. I think in fact this would probably be a good idea to adopt...

---
In reply to: [268029020](https://github.com/dotnet/machinelearning/pull/3046#discussion_r268029020) [](ancestors = 268029020)

src/Microsoft.ML.Data/Transforms/NormalizeColumn.cs
TomFinley: > Also, I just noticed that this is the method that loads from `ModelLoadContext`, shouldn't it be `ExceptDecode`?
> 
> In reply to: [267972968](https://github.com/dotnet/machinelearning/pull/3046#discussion_r267972968) [](ancestors = 267972968,267945509,267931607)

Not really. Except decode is for faults of format, which this is not. What this is describing is a properly deserialzed and formatted model being applied to user data, and that user data is not compatible with the transform. So, it would be inappropriate to throw an error explicitly for this.

src/Microsoft.ML.Data/Data/RowCursorUtils.cs
```diff
@@ -341,7 +341,7 @@ public static string TestGetLabelGetter(DataViewType type, bool allowKeys)
             if (allowKeys && type is KeyType)
                 return null;
 
-            return allowKeys ? "Expected R4, R8, Bool or Key type" : "Expected R4, R8 or Bool type";
+            return allowKeys ? "Expected Single, Double, Boolean or Key type" : "Expected Single, Double or Boolean type";
```
sfilipi: >Single [](start = 90, length = 6)

 i feel like "float" "double" or "bool" are even more user friendly.  

we don't usually go :
Boolean k = true;

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
```diff
@@ -1098,18 +1098,18 @@ private void CheckInputColumnTypes(DataViewSchema schema)
 
             var t = schema[(int)LabelIndex].Type;
             if (t != NumberDataViewType.Single && t != NumberDataViewType.Double && t != BooleanDataViewType.Instance && t.GetKeyCount() != 2)
-                throw Host.ExceptSchemaMismatch(nameof(schema), "label", LabelCol, "float, double, bool or a KeyType with cardinality 2", t.ToString());
```
sfilipi: >"float, double, bool  [](start = 83, length = 21)

i'd live them

src/Microsoft.ML.Data/Evaluators/QuantileRegressionEvaluator.cs
```diff
@@ -446,12 +446,12 @@ private void CheckInputColumnTypes(DataViewSchema schema)
...
             var t = schema[(int)LabelIndex].Type;
             if (t != NumberDataViewType.Single)
-                throw Host.ExceptSchemaMismatch(nameof(schema), "label", LabelCol, "float", t.ToString());
+                throw Host.ExceptSchemaMismatch(nameof(schema), "label", LabelCol, "Single", t.ToString());
 
             VectorType scoreType = schema[ScoreIndex].Type as VectorType;
             if (scoreType == null || scoreType.Size == 0 || (scoreType.ItemType != NumberDataViewType.Single && scoreType.ItemType != NumberDataViewType.Double))
             {
-                throw Host.ExceptSchemaMismatch(nameof(schema), "score", ScoreCol, "known-size vector of float or double", t.ToString());
+                throw Host.ExceptSchemaMismatch(nameof(schema), "score", ScoreCol, "known-size vector of Single or Double", t.ToString());
```
sfilipi: >vector [](start = 95, length = 6)

would array be more correct? vector still needs some mapping to a C# type.

src/Microsoft.ML.DataView/DataViewType.cs
eerhardt: I think we should just get rid of this string and use `RawType.Name` instead.

src/Microsoft.ML.Data/Transforms/RangeFilter.cs
Ivanidzo4ka: while `Single` and `Double` is `.net` types, Key is something we introduce. So I fill need to distinguish them somehow.
In same time Eric has PR to rename `KeyType` to `KeyDataViewType`

---
In reply to: [268028603](https://github.com/dotnet/machinelearning/pull/3046#discussion_r268028603) [](ancestors = 268028603)

src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
Ivanidzo4ka: [https://docs.microsoft.com/en-us/dotnet/standard/language-independence-and-language-independent-components?view=netframework-4.7.2#types-and-type-member-signatures](https://docs.microsoft.com/en-us/dotnet/standard/language-independence-and-language-independent-components?view=netframework-4.7.2#types-and-type-member-signatures)
If you scroll down it has list of CLS compliant types.
float and double are C# specific, while Single and Double I believe should work on vbasic, F# and other languages.


---
In reply to: [268223604](https://github.com/dotnet/machinelearning/pull/3046#discussion_r268223604) [](ancestors = 268223604)

src/Microsoft.ML.Data/Transforms/FeatureContributionCalculationTransformer.cs
Ivanidzo4ka: I try to figure out which one looks better, and come to conclusion what lower case vector is more appealing for me, but I guess i miss capital case in some places.

---
In reply to: [268028098](https://github.com/dotnet/machinelearning/pull/3046#discussion_r268028098) [](ancestors = 268028098)

src/Microsoft.ML.DataView/DataViewType.cs
Ivanidzo4ka: Well, it would be internally `ReadOnlyMemory`, but for user who construct class would be fine with `String` for person who works with `TextLoader` we have `DataKind.String`

---
In reply to: [268027832](https://github.com/dotnet/machinelearning/pull/3046#discussion_r268027832) [](ancestors = 268027832)

src/Microsoft.ML.Data/Transforms/NormalizeColumnDbl.cs
Ivanidzo4ka: lowercase

---
In reply to: [268028513](https://github.com/dotnet/machinelearning/pull/3046#discussion_r268028513) [](ancestors = 268028513)

src/Microsoft.ML.Data/Data/RowCursorUtils.cs
Ivanidzo4ka: from C# developer point, yes.
But as soon as you work with TextLoader or you inside Visual Basic you no longer have `bool` our Datakind is `Boolean` CLS wise i think it's also `Boolean`

---
In reply to: [268223385](https://github.com/dotnet/machinelearning/pull/3046#discussion_r268223385) [](ancestors = 268223385)

src/Microsoft.ML.Data/Evaluators/QuantileRegressionEvaluator.cs
Ivanidzo4ka: My problem with array, is what we also can work with `VBuffer` in our code.
Would like to hear other people opinion as well, tho.

---
In reply to: [268224361](https://github.com/dotnet/machinelearning/pull/3046#discussion_r268224361) [](ancestors = 268224361)

src/Microsoft.ML.Data/Data/RowCursorUtils.cs
eerhardt: For our C# code, we prefer to use the language specific keywords, like `float`, `bool`, `short`, etc.

However, for public APIs, strings, etc. where these words/names are visible to other languages we shouldn't be using the C# language specific keywords. Instead we should be using the .NET Type names like `Boolean`, `Single`, `Int16`, etc. We made those changes in our public API, like the `public DataKind enum`. We should use the same names in these strings.

src/Microsoft.ML.Data/Transforms/RangeFilter.cs
Ivanidzo4ka: Ok, let's name it Key.

---
In reply to: [268253632](https://github.com/dotnet/machinelearning/pull/3046#discussion_r268253632) [](ancestors = 268253632,268028603)

src/Microsoft.ML.DataView/DataViewType.cs
TomFinley: That's fine, just so long as you've thought about it. Though, incidentally, this is not an "internal" detail. We provide convenience conversions to and from string for the sake of our reflection based API, but `IDataView` itself works exclusively over `ReadOnlyMemory<char>`, never `string`.

---
In reply to: [268261843](https://github.com/dotnet/machinelearning/pull/3046#discussion_r268261843) [](ancestors = 268261843,268027832)

src/Microsoft.ML.Data/Transforms/FeatureContributionCalculationTransformer.cs
TomFinley: I agree, thank you @Ivanidzo4ka

---
In reply to: [268257594](https://github.com/dotnet/machinelearning/pull/3046#discussion_r268257594) [](ancestors = 268257594,268028098)

src/Microsoft.ML.Data/Transforms/NormalizeColumnDbl.cs
TomFinley: Good, I agree, thank you.

---
In reply to: [268267564](https://github.com/dotnet/machinelearning/pull/3046#discussion_r268267564) [](ancestors = 268267564,268028513)

PR #3052, opened 2019-03-21
Expose advanced options for the NormalizingEstimator 

### LINKED ISSUE #3047: Enable setting NormalizingEstimator options
In PR #2959 we hid the NormalizingEstimator `ColumnOptions`. Currently users cannot set advanced parameters for the estimator.

The reason for not doing that was that `NormalizingEstimator` has 5 `ColumnOptions` classes and would have required at least that many overloads. So we opted for **waiting for the full `Options`** object to be made accessible and make sure that **no breaking API change** was going to be required. 

However, as per further discussion in #2884 and offline we decided that we would not make `Options` available before v1. We therefore need another solution.

One possible approach would be to add five new extensions to MLContext each presenting all the settings in one of the NormalizingEstiamtor `ColumnOptions`. The extensions could indicate in their name the type of normalization to perform.

Note that this is not a breaking API change.

abgoswam:
Fixes  #3047 

- Introduces 5 APIs required to expose advanced options for each of the normalizing estimators in ML.NET {MinMax, MeanVariance, LogMeanVariance, Binning, SupervisedBinning}
- The APIs are added into the Experimental nuget, till  we finalize a  proper design for #2884 
- Added tests

### DIFF
--- a/src/Microsoft.ML.Experimental/MLContextExtensions.cs
+++ b/src/Microsoft.ML.Experimental/MLContextExtensions.cs
@@ -16,37 +16,37 @@ namespace Microsoft.ML.Experimental
         public static void CancelExecution(this MLContext ctx) => ctx.CancelExecution();
 
         /// <summary>
-        /// Normalize (rescale) the column according the <see cref="NormalizingEstimator.NormalizationMode.MinMax"/> mode.
+        /// Normalize (rescale) the column according to the <see cref="NormalizingEstimator.NormalizationMode.MinMax"/> mode.
         /// It normalizes the data based on the observed minimum and maximum values of the data.
         /// </summary>
         /// <param name="catalog">The transform catalog</param>
         /// <param name="outputColumnName">Name of the column resulting from the transformation of <paramref name="inputColumnName"/>.</param>
         /// <param name="inputColumnName">Name of the column to transform. If set to <see langword="null"/>, the value of the <paramref name="outputColumnName"/> will be used as source.</param>
-        /// <param name="fixZero">Whether to map zero to zero, preserving sparsity.</param>
         /// <param name="maximumExampleCount">Maximum number of examples used to train the normalizer.</param>
+        /// <param name="fixZero">Whether to map zero to zero, preserving sparsity.</param>
         public static NormalizingEstimator NormalizeMinMax(this TransformsCatalog catalog,
            string outputColumnName, string inputColumnName = null,
-           bool fixZero = NormalizingEstimator.Defaults.EnsureZeroUntouched,
-           long maximumExampleCount = NormalizingEstimator.Defaults.MaximumExampleCount)
+           long maximumExampleCount = NormalizingEstimator.Defaults.MaximumExampleCount,
+           bool fixZero = NormalizingEstimator.Defaults.EnsureZeroUntouched)
         {
             var columnOptions = new NormalizingEstimator.MinMaxColumnOptions(outputColumnName, inputColumnName, maximumExampleCount, fixZero);
             return new NormalizingEstimator(CatalogUtils.GetEnvironment(catalog), columnOptions);
         }
 
         /// <summary>
-        /// Normalize (rescale) the column according the <see cref="NormalizingEstimator.NormalizationMode.MeanVariance"/> mode.
+        /// Normalize (rescale) the column according to the <see cref="NormalizingEstimator.NormalizationMode.MeanVariance"/> mode.
         /// It normalizes the data based on the computed mean and variance of the data.
         /// </summary>
         /// <param name="catalog">The transform catalog</param>
         /// <param name="outputColumnName">Name of the column resulting from the transformation of <paramref name="inputColumnName"/>.</param>
         /// <param name="inputColumnName">Name of the column to transform. If set to <see langword="null"/>, the value of the <paramref name="outputColumnName"/> will be used as source.</param>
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.Experimental/MLContextExtensions.cs
```diff
@@ -11,5 +14,105 @@ public static class MLContextExtensions
...
         public static void CancelExecution(this MLContext ctx) => ctx.CancelExecution();
+
+        /// <summary>
+        /// Normalize (rescale) the column according the <see cref="NormalizingEstimator.NormalizationMode.MinMax"/> mode.
+        /// It normalizes the data based on the observed minimum and maximum values of the data.
+        /// </summary>
+        /// <param name="catalog">The transform catalog</param>
+        /// <param name="outputColumnName">Name of the column resulting from the transformation of <paramref name="inputColumnName"/>.</param>
+        /// <param name="inputColumnName">Name of the column to transform. If set to <see langword="null"/>, the value of the <paramref name="outputColumnName"/> will be used as source.</param>
+        /// <param name="ensureZeroUntouched">Whether to map zero to zero, preserving sparsity.</param>
```
TomFinley: >ensureZeroUntouched [](start = 25, length = 19)

Do we like this "untouched?" I mean, I guess it's fine, but...

test/Microsoft.ML.Tests/Transformers/NormalizerTests.cs
```diff
@@ -255,6 +256,71 @@ public void SimpleConstructorsAndExtensions()
             CheckSameValues(data6, data7);
             CheckSameValues(data6, data8);
 
+            // Tests for LogMeanVariance
+            var est9 = new NormalizingEstimator(Env, NormalizingEstimator.NormalizationMode.LogMeanVariance, ("float4", "float4"));
```
TomFinley: > "float4" [](start = 119, length = 9)

I'd prefer in *new* tests to avoid using non-public API. Is that impossible in this case for some reason?

abgoswam: I carved out a separate test `NormalizerExperimentalExtensions`  to exercise the experimental APIs

---
In reply to: [268185782](https://github.com/dotnet/machinelearning/pull/3052#discussion_r268185782) [](ancestors = 268185782)

src/Microsoft.ML.Experimental/MLContextExtensions.cs
```diff
@@ -11,5 +14,105 @@ public static class MLContextExtensions
         /// </summary>
         /// <param name="ctx"><see cref="MLContext"/> reference.</param>
         public static void CancelExecution(this MLContext ctx) => ctx.CancelExecution();
+
+        /// <summary>
+        /// Normalize (rescale) the column according the <see cref="NormalizingEstimator.NormalizationMode.MinMax"/> mode.
```
artidoro: >according the [](start = 43, length = 13)

I think it would sound better if you say "according to the".

src/Microsoft.ML.Experimental/MLContextExtensions.cs
abgoswam: updated to use "fixZero" in the API

internally, our codebase uses a mix of 'fixZero' and 'ensureZeroUntouched' , will create a separate bug to clean this up  post  v1

---
In reply to: [268184652](https://github.com/dotnet/machinelearning/pull/3052#discussion_r268184652) [](ancestors = 268184652)

src/Microsoft.ML.Experimental/MLContextExtensions.cs
```diff
@@ -11,5 +14,105 @@ public static class MLContextExtensions
...
+        /// <param name="catalog">The transform catalog</param>
+        /// <param name="outputColumnName">Name of the column resulting from the transformation of <paramref name="inputColumnName"/>.</param>
+        /// <param name="inputColumnName">Name of the column to transform. If set to <see langword="null"/>, the value of the <paramref name="outputColumnName"/> will be used as source.</param>
+        /// <param name="ensureZeroUntouched">Whether to map zero to zero, preserving sparsity.</param>
+        /// <param name="maximumExampleCount">Maximum number of examples used to train the normalizer.</param>
+        /// <param name="useCdf">Whether to use CDF as the output.</param>
+        public static NormalizingEstimator NormalizeMeanVariance(this TransformsCatalog catalog,
+            string outputColumnName, string inputColumnName = null,
+            bool ensureZeroUntouched = NormalizingEstimator.Defaults.EnsureZeroUntouched,
+            long maximumExampleCount = NormalizingEstimator.Defaults.MaximumExampleCount,
```
artidoro: I noticed that you reversed the order of `maximumExampleCount` and `ensureZeroUntouched` here and in the other extensions compared to the order in which they appeared in the constructors of the `ColumnOptions`,
It seems that `maximumExampleCount` is more important than the other argument, or at least  I would expect people to use it more.

abgoswam: good point. 

Subject to  whether the internal constructors use them, lets keep the following order : 

- maximumExampleCount  
- fixZero 
- useCdf
- maximumBinCount
- mininimumExamplesPerBin

---
In reply to: [268249207](https://github.com/dotnet/machinelearning/pull/3052#discussion_r268249207) [](ancestors = 268249207)

PR #3055, opened 2019-03-21
Rename ImageType to ImageDataViewType

### LINKED ISSUE #3050: Rename ImageType to ImageDataViewType
With #2297 we decided to suffix all the IDataView type classes with `DataViewType` instead of just `Type`. However, we missed `ImageType`:

https://github.com/dotnet/machinelearning/blob/70ef7ecd43b031b481a4047ea361da5e2f360336/src/Microsoft.ML.ImageAnalytics/ImageType.cs#L12

We should rename this class as well to match the rest in the hierarchy. This will need to be done before v1.0 since it would be a breaking change to rename it later.

abgoswam:
Fixes #3050 

- Renamed ImageType to ImageDataViewType

### DIFF
--- a/src/Microsoft.ML.ImageAnalytics/ImageType.cs
+++ b/src/Microsoft.ML.ImageAnalytics/ImageType.cs
@@ -9,11 +9,11 @@ using Microsoft.ML.Runtime;
 
 namespace Microsoft.ML.Transforms.Image
 {
-    public sealed class ImageType : StructuredDataViewType
+    public sealed class ImageDataViewType : StructuredDataViewType
     {
         public readonly int Height;
         public readonly int Width;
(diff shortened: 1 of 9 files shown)


src/Microsoft.ML.ImageAnalytics/ImageType.cs
```diff
@@ -9,11 +9,11 @@
 
 namespace Microsoft.ML.Transforms.Image
 {
-    public sealed class ImageType : StructuredDataViewType
+    public sealed class ImageDataViewType : StructuredDataViewType
```
TomFinley: If for some other reason you have to update this PR, please rename this file.

abgoswam:
Thanks folks for the reviews!

PR #3056, opened 2019-03-21
Fix bug in TextLoader

### LINKED ISSUE #2996: Cannot create TextLoader without loading assemblies into ComponentCatalog

yaeldekel:
Fixes #2996. (This PR replaces #3011).

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1288,12 +1288,9 @@ namespace Microsoft.ML.Data
                 ch.Assert(h.Loader == null || h.Loader is ICommandLineComponentFactory);
                 var loader = h.Loader as ICommandLineComponentFactory;
 
-                if (loader == null || string.IsNullOrWhiteSpace(loader.Name))
-                    goto LDone;
-
-                // Make sure the loader binds to us.
-                var info = host.ComponentCatalog.GetLoadableClassInfo<SignatureDataLoader>(loader.Name);
-                if (info.Type != typeof(ILegacyDataLoader) || info.ArgType != typeof(Options))
+                // Make sure that the schema is described using either the syntax TextLoader{<settings>} or the syntax Text{<settings>},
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1288,12 +1288,9 @@ private sealed class LoaderHolder
...
 
-                if (loader == null || string.IsNullOrWhiteSpace(loader.Name))
-                    goto LDone;
-
-                // Make sure the loader binds to us.
-                var info = host.ComponentCatalog.GetLoadableClassInfo<SignatureDataLoader>(loader.Name);
-                if (info.Type != typeof(ILegacyDataLoader) || info.ArgType != typeof(Options))
+                // Make sure that the schema is described using either the syntax TextLoader{<settings>} or the syntax Text{<settings>},
+                // where "settings" is a string that can be parsed by CmdParser into an object of type TextLoader.Options.
+                if (loader == null || string.IsNullOrWhiteSpace(loader.Name) || (loader.Name != LoaderSignature && loader.Name != "Text"))
```
yaeldekel: >Name [](start = 88, length = 4)

The previous check was to verify that `loader.Name` is one of the names defined in `TextLoader`'s `LoadableClassAttribute`. Since now we cannot rely on this attribute being in the `ComponentCatalog`, we explicitly require `loader.Name` to be one of the load names of `TextLoader`.

PR #3066, opened 2019-03-22
Multicolumn mapping for some estimators

### LINKED ISSUE #2884: Discussion: ColumnOptions actually a good name?

### LINKED ISSUE #3068: Add multicolumn mapping for the estimators that need it the most
This is not a Project 13 issue any longer, as it does not require any breaking API change.

We should enable multi-column mapping for all the estimators that support it. Since we don't have time to work on a general solution, we will go for option 1 presented in https://github.com/dotnet/machinelearning/issues/2884#issuecomment-475389060.

This will only be done for the estimators indicated in the list in the issue #2884. Going forward we should do it for all others.

artidoro:
Adding multicolumn mapping for some estimators (as per list by @TomFinley and @glebuk):

- OneHotEncodingEstimator
- TypeConvertingEstimator
- KeyToVectorMappingEstimator
- ValueToKeyMappingEstimator
- OneHotHashEncodingEstimator
- MissingValueEstimator
- FeatureSelectionCatalog.*
- KeyToValueMappingEstiamtor

Leaving out:
- TextFeaturizingEstimator (probably requires column specific settings most of the time)
- NoramlizingEstiamtor (in experimental nuget)

Let me know if I should add more estimators.

Fixes #3068
Related to #2884

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
+++ b/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.ML.Data;
+using Microsoft.ML.Runtime;
 using Microsoft.ML.Transforms;
 
 namespace Microsoft.ML
@@ -75,8 +76,10 @@ namespace Microsoft.ML
             InputOutputColumnPair[] columns,
             DataKind outputKind = ConvertDefaults.DefaultOutputKind)
         {
+            var env = CatalogUtils.GetEnvironment(catalog);
+            env.CheckValue(columns, nameof(columns));
             var columnOptions = columns.Select(x => new TypeConvertingEstimator.ColumnOptions(x.OutputColumnName, outputKind, x.InputColumnName)).ToArray();
-            return new TypeConvertingEstimator(CatalogUtils.GetEnvironment(catalog), columnOptions);
+            return new TypeConvertingEstimator(env, columnOptions);
         }
 
         /// <summary>
@@ -109,7 +112,11 @@ namespace Microsoft.ML
         /// <param name="catalog">The conversion transform's catalog.</param>
         /// <param name="columns">Specifies the names of the columns on which to apply the transformation.</param>
         public static KeyToValueMappingEstimator MapKeyToValue(this TransformsCatalog.ConversionTransforms catalog, InputOutputColumnPair[] columns)
-            => new KeyToValueMappingEstimator(CatalogUtils.GetEnvironment(catalog), columns.Select(x => (x.OutputColumnName, x.InputColumnName)).ToArray());
+        {
+            var env = CatalogUtils.GetEnvironment(catalog);
+            env.CheckValue(columns, nameof(columns));
+            return new KeyToValueMappingEstimator(env, columns.Select(x => (x.OutputColumnName, x.InputColumnName)).ToArray());
+        }
 
         /// <summary>
         /// Maps key types or key values into a floating point vector.
@@ -143,8 +150,10 @@ namespace Microsoft.ML
         public static KeyToVectorMappingEstimator MapKeyToVector(this TransformsCatalog.ConversionTransforms catalog,
             InputOutputColumnPair[] columns, bool outputCountVector = KeyToVectorMappingEstimator.Defaults.OutputCountVector)
         {
+            var env = CatalogUtils.GetEnvironment(catalog);
+            env.CheckValue(columns, nameof(columns));
             var columnOptions = columns.Select(x => new KeyToVectorMappingEstimator.ColumnOptions(x.OutputColumnName, x.InputColumnName, outputCountVector)).ToArray();
-            return new KeyToVectorMappingEstimator(CatalogUtils.GetEnvironment(catalog), columnOptions);
+            return new KeyToVectorMappingEstimator(env, columnOptions);
 
         }
 
@@ -197,8 +206,10 @@ namespace Microsoft.ML
             bool addKeyValueAnnotationsAsText = ValueToKeyMappingEstimator.Defaults.AddKeyValueAnnotationsAsText,
             IDataView keyData = null)
         {
+            var env = CatalogUtils.GetEnvironment(catalog);
+            env.CheckValue(columns, nameof(columns));
             var columnOptions = columns.Select(x => new ValueToKeyMappingEstimator.ColumnOptions(x.OutputColumnName, x.InputColumnName, maximumNumberOfKeys, keyOrdinality, addKeyValueAnnotationsAsText)).ToArray();
-            return new ValueToKeyMappingEstimator(CatalogUtils.GetEnvironment(catalog), columnOptions, keyData);
+            return new ValueToKeyMappingEstimator(env, columnOptions, keyData);
         }
 
         /// <summary>
@@ -278,9 +289,11 @@ namespace Microsoft.ML
             IEnumerable<KeyValuePair<TInputType, TOutputType>> keyValuePairs,
             params InputOutputColumnPair[] columns)
         {
+            var env = CatalogUtils.GetEnvironment(catalog);
+            env.CheckValue(columns, nameof(columns));
             var keys = keyValuePairs.Select(pair => pair.Key);
             var values = keyValuePairs.Select(pair => pair.Value);
-            return new ValueMappingEstimator<TInputType, TOutputType>(CatalogUtils.GetEnvironment(catalog), keys, values, InputOutputColumnPair.ConvertToValueTuples(columns));
+            return new ValueMappingEstimator<TInputType, TOutputType>(env, keys, values, InputOutputColumnPair.ConvertToValueTuples(columns));
         }
 
         /// <summary>
@@ -306,9 +319,11 @@ namespace Microsoft.ML
             bool treatValuesAsKeyType,
             params InputOutputColumnPair[] columns)
         {
+            var env = CatalogUtils.GetEnvironment(catalog);
+            env.CheckValue(columns, nameof(columns));
             var keys = keyValuePairs.Select(pair => pair.Key);
             var values = keyValuePairs.Select(pair => pair.Value);
-            return new ValueMappingEstimator<TInputType, TOutputType>(CatalogUtils.GetEnvironment(catalog), keys, values, treatValuesAsKeyType,
+            return new ValueMappingEstimator<TInputType, TOutputType>(env, keys, values, treatValuesAsKeyType,
                   InputOutputColumnPair.ConvertToValueTuples(columns));
         }
 
@@ -367,9 +382,11 @@ namespace Microsoft.ML
             IEnumerable<KeyValuePair<TInputType, TOutputType[]>> keyValuePairs,
             params InputOutputColumnPair[] columns)
         {
+            var env = CatalogUtils.GetEnvironment(catalog);
+            env.CheckValue(columns, nameof(columns));
             var keys = keyValuePairs.Select(pair => pair.Key);
             var values = keyValuePairs.Select(pair => pair.Value);
-            return new ValueMappingEstimator<TInputType, TOutputType>(CatalogUtils.GetEnvironment(catalog), keys, values,
+            return new ValueMappingEstimator<TInputType, TOutputType>(env, keys, values,
                     InputOutputColumnPair.ConvertToValueTuples(columns));
         }
 
@@ -422,7 +439,11 @@ namespace Microsoft.ML
         internal static ValueMappingEstimator MapValue(
             this TransformsCatalog.ConversionTransforms catalog,
             IDataView lookupMap, DataViewSchema.Column keyColumn, DataViewSchema.Column valueColumn, params InputOutputColumnPair[] columns)
-            => new ValueMappingEstimator(CatalogUtils.GetEnvironment(catalog), lookupMap, keyColumn.Name, valueColumn.Name,
+        {
+            var env = CatalogUtils.GetEnvironment(catalog);
+            env.CheckValue(columns, nameof(columns));
+            return new ValueMappingEstimator(env, lookupMap, keyColumn.Name, valueColumn.Name,
                 InputOutputColumnPair.ConvertToValueTuples(columns));
+        }
     }
 }
--- a/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
+++ b/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
@@ -66,6 +66,20 @@ namespace Microsoft.ML
             => new TypeConvertingEstimator(CatalogUtils.GetEnvironment(catalog), new[] { new TypeConvertingEstimator.ColumnOptions(outputColumnName, outputKind, inputColumnName) });
 
         /// <summary>
+        /// Changes column type of the input columns.
+        /// </summary>
+        /// <param name="catalog">The conversion transform's catalog.</param>
+        /// <param name="columns">Specifies the names of the columns on which to apply the transformation.</param>
+        /// <param name="outputKind">The expected kind of the output column.</param>
+        public static TypeConvertingEstimator ConvertType(this TransformsCatalog.ConversionTransforms catalog,
+            InputOutputColumnPair[] columns,
+            DataKind outputKind = ConvertDefaults.DefaultOutputKind)
+        {
+            var columnOptions = columns.Select(x => new TypeConvertingEstimator.ColumnOptions(x.OutputColumnName, outputKind, x.InputColumnName)).ToArray();
+            return new TypeConvertingEstimator(CatalogUtils.GetEnvironment(catalog), columnOptions);
+        }
+
+        /// <summary>
         /// Changes column type of the input column.
         /// </summary>
         /// <param name="catalog">The conversion transform's catalog.</param>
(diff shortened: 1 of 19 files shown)


src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
```diff
@@ -8,6 +8,32 @@
...
+        public readonly string OutputColumnName;
+
+        /// <summary>
+        /// Specifies input and output column names for a transformation.
+        /// </summary>
+        /// <param name="outputColumnName">Name of the column resulting from the transformation of <paramref name="inputColumnName"/>.</param>
+        /// <param name="inputColumnName">Name of the column to transform. If set to <see langword="null"/>, the value of the <paramref name="outputColumnName"/> will be used as source.</param>
+        public InputOutputColumnPair(string outputColumnName, string inputColumnName = null)
+        {
+            InputColumnName = inputColumnName;
```
Ivanidzo4ka: > InputColumnName = inputColumnName [](start = 11, length = 34)

`InputColumnName = inputColumnName?? outputColumnName`

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
```diff
@@ -8,6 +8,32 @@
...
 {
+    /// <summary>
+    /// Specifies input and output column names for a transformation.
+    /// </summary>
+    public sealed class InputOutputColumnPair
+    {
+        /// <summary>
+        /// Name of the column to transform. If set to <see langword="null"/>, the value of the <see cref="OutputColumnName"/> will be used as source.
+        /// </summary>
+        public readonly string InputColumnName;
```
Ivanidzo4ka: > public readonly string InputColumnName [](start = 7, length = 39)

I'm slightly confuse.
We get rid of ColumnOptions because they were immutable, and now we add another immutable class...

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
artidoro: Thank you Ivan! I have created and erased too many of these classes :( 

---
In reply to: [268332864](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268332864) [](ancestors = 268332864)

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
```diff
@@ -8,6 +8,32 @@
 
 namespace Microsoft.ML
 {
+    /// <summary>
+    /// Specifies input and output column names for a transformation.
+    /// </summary>
+    public sealed class InputOutputColumnPair
```
Ivanidzo4ka: >InputOutputColumnPair [](start = 24, length = 21)

what is difference between this one and `ColumnOptions`?

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
artidoro: Will invite you to check with Tom about this, I am just executing what he asked

---
In reply to: [268333577](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268333577) [](ancestors = 268333577)

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
artidoro: Not much, I would have used the bellow if it were me, but Tom asked it to be different, for some reason!

---
In reply to: [268334077](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268334077) [](ancestors = 268334077)

TomFinley: Yes, a very good reason. `ColumnOptions` is a struct meant to serve a *specific* transformer base, and that is involved in the type heirarchy, and in particular something that captured all of the individual settings and state for each mapping. Our goals here were comparatively more modest: we just needed to . Well designed code does what it is designed to do, and in the simplest possible way. There is no need for this to be part of an elaborate type hierarchy -- this was in fact the mistake that led to the issue #2884 being filed, that we'd conflated two *distinct* techniques for the extremely bad reasoning that they both had to do with "column." (And, of course, if something deals with the same *sort* of object, obviously they belong in the same type heirarchy, right?)

---
In reply to: [268334389](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268334389) [](ancestors = 268334389,268334077)

Ivanidzo4ka: Just in case, I'm not talking about `ColumnOptions` inside Estimator, (for example `HashingEstimator.ColumnOptions`.
I'm talking about `ColumnOptions` in this exact file few lines below (line 41)
Only difference I see is Input/output column names is public in this one instead of private in other, and name of class.
Ok, two difference, one below has implicit converter from tuple.

Can we delete `ColumnOptions` in this file?

---
In reply to: [268337035](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268337035) [](ancestors = 268337035,268334389,268334077)

artidoro:
Since @ivanbasov  asked for it the third commit shows the how this would look like if we removed the `ColumnOptions` class, and used `InputOutputColumnPair`.

Let me know what looks better! I will either eliminate this commit or keep it

src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
```diff
@@ -65,6 +65,20 @@ internal static HashingEstimator Hash(this TransformsCatalog.ConversionTransform
...
             => new TypeConvertingEstimator(CatalogUtils.GetEnvironment(catalog), new[] { new TypeConvertingEstimator.ColumnOptions(outputColumnName, outputKind, inputColumnName) });
 
+        /// <summary>
+        /// Changes column type of the input columns.
+        /// </summary>
+        /// <param name="catalog">The conversion transform's catalog.</param>
+        /// <param name="columns">Specifies the names of the columns on which to apply the transformation.</param>
+        /// <param name="outputKind">The expected kind of the output column.</param>
+        public static TypeConvertingEstimator ConvertType(this TransformsCatalog.ConversionTransforms catalog,
+            InputOutputColumnPair[] columns,
```
glebuk: >columns [](start = 36, length = 7)

Don't you need to check for null for the columns arg to avoid null reference exception?

src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
```diff
@@ -90,20 +104,12 @@ public static KeyToValueMappingEstimator MapKeyToValue(this TransformsCatalog.Co
...
-        /// <![CDATA[
-        ///  [!code-csharp[KeyToValueMappingEstimator](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/ValueMappingStringToKeyType.cs)]
-        /// ]]></format>
-        /// </example>
-        [BestFriend]
-        internal static KeyToValueMappingEstimator MapKeyToValue(this TransformsCatalog.ConversionTransforms catalog, params ColumnOptions[] columns)
-             => new KeyToValueMappingEstimator(CatalogUtils.GetEnvironment(catalog), ColumnOptions.ConvertToValueTuples(columns));
+        /// <param name="catalog">The conversion transform's catalog.</param>
+        /// <param name="columns">Specifies the names of the columns on which to apply the transformation.</param>
+        public static KeyToValueMappingEstimator MapKeyToValue(this TransformsCatalog.ConversionTransforms catalog, InputOutputColumnPair[] columns)
```
glebuk: >columns [](start = 140, length = 7)

also check for null

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
```diff
@@ -11,35 +11,32 @@ namespace Microsoft.ML
...
-        {
-            _outputColumnName = outputColumnName;
-            _inputColumnName = inputColumnName ?? outputColumnName;
-        }
-
-        /// <summary>
-        /// Instantiates a <see cref="ColumnOptions"/> from a tuple of input and output column names.
-        /// </summary>
-        public static implicit operator ColumnOptions((string outputColumnName, string inputColumnName) value)
+        public InputOutputColumnPair(string outputColumnName, string inputColumnName = null)
```
glebuk: >public InputOutputColumnPair(string outputColumnName, string inputColumnName = null) [](start = 8, length = 84)

Add another overload for the case when input name = output name. That would make it a lot clearer vs setting one to null.

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
```diff
@@ -11,35 +11,32 @@ namespace Microsoft.ML
...
-    [BestFriend]
-    internal sealed class ColumnOptions
+    public sealed class InputOutputColumnPair
     {
-        private readonly string _outputColumnName;
-        private readonly string _inputColumnName;
+        /// <summary>
+        /// Name of the column to transform. If set to <see langword="null"/>, the value of the <see cref="OutputColumnName"/> will be used as source.
+        /// </summary>
+        public readonly string InputColumnName;
```
glebuk: >InputColumnName [](start = 31, length = 15)

Shouldn't it be in reverse - the input be set, but output be optional and equal to input?

src/Microsoft.ML.Transforms/CategoricalCatalog.cs
```diff
@@ -40,6 +41,28 @@ public static class CategoricalCatalog
...
+        /// <param name="keyData">Specifies an ordering for the encoding. If specified, this should be a single column data view,
+        /// and the key-values will be taken from that column. If unspecified, the ordering will be determined from the input data upon fitting.</param>
+        public static OneHotEncodingEstimator OneHotEncoding(this TransformsCatalog.CategoricalTransforms catalog,
+                InputOutputColumnPair[] columns,
+                OneHotEncodingEstimator.OutputKind outputKind = OneHotEncodingEstimator.Defaults.OutKind,
+                int maximumNumberOfKeys = ValueToKeyMappingEstimator.Defaults.MaximumNumberOfKeys,
+                ValueToKeyMappingEstimator.KeyOrdinality keyOrdinality = ValueToKeyMappingEstimator.Defaults.Ordinality,
+                IDataView keyData = null)
+        {
+            var columnOptions = columns.Select(x => new OneHotEncodingEstimator.ColumnOptions(x.OutputColumnName, x.InputColumnName, outputKind, maximumNumberOfKeys, keyOrdinality)).ToArray();
```
glebuk: >columns [](start = 32, length = 7)

check for null here and elsewhere

test/Microsoft.ML.Tests/Transformers/FeatureSelectionTests.cs
```diff
@@ -182,9 +182,9 @@ public void MutualInformationSelectionWorkout()
 
             var est = ML.Transforms.FeatureSelection.SelectFeaturesBasedOnMutualInformation("FeatureSelect", "VectorFloat", slotsInOutput: 1, labelColumnName: "Label")
                 .Append(ML.Transforms.FeatureSelection.SelectFeaturesBasedOnMutualInformation(labelColumnName: "Label", slotsInOutput: 2, numberOfBins: 100,
-                    columns: new ColumnOptions[] {
-                        ("out1", "VectorFloat"),
-                        ("out2", "VectorDouble")
+                    columns: new[] {
+                        new InputOutputColumnPair("out1", "VectorFloat"),
```
glebuk: >InputOutputColumnPair [](start = 28, length = 21)

why do we have to use the more verbose initializer here?  Ideally we should use the old syntax is possible as it is a lot more compact.

glebuk:
>         public void ValueMappingValueTypeIsVectorWorkout()

Add a test for when InputOutputColumnPair is null.

---
Refers to: test/Microsoft.ML.Tests/Transformers/ValueMappingTests.cs:523 in 802e4de. [](commit_id = 802e4de79822d44a04eefc13c0ac9a4780185e3e, deletion_comment = False)

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
artidoro: As per [https://github.com/dotnet/machinelearning/issues/2064](https://github.com/dotnet/machinelearning/issues/2064) we use the `outputColumnName` as `inputColumnName` when `inputColumnName` is null. We are doing this across the code base.

---
In reply to: [268727730](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268727730) [](ancestors = 268727730)

test/Microsoft.ML.Tests/Transformers/FeatureSelectionTests.cs
artidoro: Because the other one uses tuples. We have decided not to have tuples in the public surface any longer.

---
In reply to: [268729946](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268729946) [](ancestors = 268729946)

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
TomFinley: *That's* fine. What Artidoro and I had discussed was actually somewhat different. (Or we were talking about two separate things without realizing it.)

---
In reply to: [268339442](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268339442) [](ancestors = 268339442,268337035,268334389,268334077)

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
TomFinley: >outputColumnName [](start = 44, length = 16)

Check non-empty on `outputColumName` probably.

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
TomFinley: >InputColumnName [](start = 31, length = 15)

Should these be properties?

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
artidoro: I guess we were talking about something different, glad we are on the same page.

---
In reply to: [268760195](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268760195) [](ancestors = 268760195,268339442,268337035,268334389,268334077)

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
artidoro: Unfortunately I believe that everywhere in the codebase we have the same pattern `string outputColumnName, string inputColumnName = null`. This is found in all the mlContext extensions for transforms. I can add it here, but I think it would make more sense to stick to the general pattern.

---
In reply to: [268726871](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268726871) [](ancestors = 268726871)

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
artidoro: Ok making these properties!

---
In reply to: [268761458](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268761458) [](ancestors = 268761458)

src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
artidoro: I added the checks thanks for pointing out. I only fixed the extensions that are public. Not those that are internal.

---
In reply to: [268724517](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268724517) [](ancestors = 268724517)

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
TomFinley: We must be consistent. Sorry @glebuk!

---
In reply to: [268779384](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268779384) [](ancestors = 268779384,268726871)

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
TomFinley: This seems good to me. The `ColumnOptions` I had been talking about (which is to say, the vast majority of things with that name) are as I said meant to serve a different purpose.

We should probably rename them back to `ColumnInfo` at some point but this can be delayed as they are internal... ummm except one. Whoops. Opened #3078. :) Aside from that, yeah.

---
In reply to: [268771510](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268771510) [](ancestors = 268771510,268760195,268339442,268337035,268334389,268334077)

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
glebuk: Ok  then.

---
In reply to: [268819893](https://github.com/dotnet/machinelearning/pull/3066#discussion_r268819893) [](ancestors = 268819893,268779384,268726871)

artidoro:
Thank you for reviewing, I am updating now with the latest changes.

PR #3071, opened 2019-03-22
TextLoader throws when type is missing LoadColumnAttribute

### LINKED ISSUE #2037: Need better exception for TextLoader when column attribute is not present
In PR #2033 we converted all uses of the legacy TextLoader to the new TextLoader.

It emerged that when reading from a file providing the schema, if there is no column attribute, then there are two different exceptions thrown depending on whether it is a debug or release build.

We need a single exception for both debug and release, and we need the name to be more informative. We also need a single message for both exceptions.

This is the test that exerts these exceptions: ThrowsExceptionWithPropertyName in TextLoaderTests.cs

### LINKED ISSUE #3051: LoadFromTextFile invalid model throw NullReferenceException
### System information

.NET Core SDK (reflecting any global.json):
 Version:   2.2.104
 Commit:    73f036d4ac

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17763
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.2.104\

### Issue

- **What did you do?**
Tried to load a model from [this Kaggle](https://www.kaggle.com/harlfoxem/housesalesprediction/kernels). My model definition skipped a column index.

- **What happened?**

NullReferenceException

- **What did you expect?**

A user-friendly error message. 

### Source code / logs

```
public class KingCountyHouseData
{
        [LoadColumn(0)]
        public string Id { get; set; }

        [LoadColumn(1)]
        public string SaleDate { get; set; }
        public float Price { get; set; } // <== this was missed

        [LoadColumn(2)]  // <=== this should be +1
        public float Bedrooms { get; set; }

        [LoadColumn(3)]  // <=== this should be +1
        public float Bathrooms { get; set; }

        [LoadColumn(4)]  // <=== this should be +1
        public float LivingAreaSquareFoot { get; set; }

        [LoadColumn(5)]  // <=== this should be +1
        public float Floors { get; set; }

        [LoadColumn(6)]  // <=== this should be +1
        public bool Waterfront { get; set; }

        //.... more fields below.
}
```

yaeldekel:
Fixes #3051.

Edit: This also fixes #2037.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1460,7 +1460,8 @@ namespace Microsoft.ML.Data
                 var memberInfo = memberInfos[index];
                 var mappingAttr = memberInfo.GetCustomAttribute<LoadColumnAttribute>();
 
-                host.Assert(mappingAttr != null, $"Field or property {memberInfo.Name} is missing the {nameof(LoadColumnAttribute)} attribute");
+                if (mappingAttr == null)
+                    throw host.Except($"{(memberInfo is FieldInfo ? "Field" : "Property")} '{memberInfo.Name}' is missing the {nameof(LoadColumnAttribute)} attribute");
 
                 var mappingAttrName = memberInfo.GetCustomAttribute<ColumnNameAttribute>();
 
--- a/test/Microsoft.ML.Tests/TextLoaderTests.cs
+++ b/test/Microsoft.ML.Tests/TextLoaderTests.cs
@@ -617,8 +617,6 @@ namespace Microsoft.ML.EntryPoints.Tests
             {
                 Assert.Contains($"Property 'Feature' is missing the {nameof(LoadColumnAttribute)} attribute", e.Message);
             }
-
-
         }
 
         [Fact]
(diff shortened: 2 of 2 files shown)


src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1460,7 +1460,8 @@ void ICanSaveModel.Save(ModelSaveContext ctx)
                 var memberInfo = memberInfos[index];
                 var mappingAttr = memberInfo.GetCustomAttribute<LoadColumnAttribute>();
 
-                host.Assert(mappingAttr != null, $"Field or property {memberInfo.Name} is missing the {nameof(LoadColumnAttribute)} attribute");
+                if (mappingAttr == null)
+                    throw host.Except($"{(memberInfo is FieldInfo ? "Field" : "Property")} '{memberInfo.Name}' is missing the {nameof(LoadColumnAttribute)} attribute");
```
TomFinley: >memberInfo.Name [](start = 93, length = 15)

Just a note, while we often use `'` to indicate quoting of things that can contain potentially confusing items like spaces, since C# identifiers cannot contain spaces I view their presence as less essential than I would in, say, a column name or some other such entity. (Not a big deal at all, just registering a random thought since I am half allergic to letting PRs pass with no comments whatsoever. :D )

PR #3080, opened 2019-03-25
Move transform catalog extensions into its own file and class in experimental nuget.

### LINKED ISSUE #3081: Move transform catalog extensions into its own class in experimental nuget.
Move transform catalog extensions into its own class in experimental nuget.

codemzs:
Transform catalog extension methods are misplaced in MlContextExtensions.cs file. They need to go in a file of its own.

fixes #3081


artidoro:
Hey @codemzs, I think we need to add them back to another file. Did you mean to add that to this commit?

codemzs:
@artidoro That is correct, thanks.

PR #3084, opened 2019-03-25
Fixing API call in LoadImages sample

### LINKED ISSUE #3079: DnnFeaturizeImage sample doesn't match the signature in the current API
The API for `LoadImage` is

```
LoadImages(string outputColumnName, string imageFolder, string inputColumnName = null)
```

but in the sample it is used like this:

```
mlContext.Transforms.LoadImages(imagesFolder, "ImageObject", "ImagePath")
```

One of them should be fixed, not sure which one? @TomFinley , @artidoro , @wschin 
I kind of like the second option (the folder first, output and input column names together).

rogancarr:
The `LoadImages` sample had the wrong API call. This PR fixes it so that the sample runs correctly.

Fixes #3079

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/LoadImages.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/LoadImages.cs
@@ -36,7 +36,7 @@ namespace Microsoft.ML.Samples.Dynamic
 
             var imagesFolder = Path.GetDirectoryName(imagesDataFile);
             // Image loading pipeline. 
-            var pipeline = mlContext.Transforms.LoadImages(imagesFolder, "ImageReal", "ImagePath");
+            var pipeline = mlContext.Transforms.LoadImages("ImageReal", imagesFolder, "ImagePath");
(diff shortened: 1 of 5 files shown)


docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/LoadImages.cs
```diff
@@ -36,7 +36,7 @@ public static void Example()
 
             var imagesFolder = Path.GetDirectoryName(imagesDataFile);
             // Image loading pipeline. 
-            var pipeline = mlContext.Transforms.LoadImages(imagesFolder, "ImageReal", "ImagePath");
+            var pipeline = mlContext.Transforms.LoadImages("ImageReal", imagesFolder, "ImagePath");
```
yaeldekel: >LoadImages [](start = 48, length = 10)

Hi @rogancarr, thanks for addressing this issue.
Sorry, I should have been more specific in the issue itself - there are a few more references to this API in the samples project (DnnFeaturizeImage, ResizeImage, ConvertToGrayScale, ExtractPixels), could you update these as well?

PR #3086, opened 2019-03-25
Updating the ApplyOnnxModel transform to meet the API parameter ordering standards

### LINKED ISSUE #3082: ApplyOnnxModel API parameters are in the wrong order
The `ApplyOnnxModel` `Transformer` has a parameter order based on the `ONNX` model (see below) but as a `Transformer`, it should have parameters ordered to reflect the modification of the input `IDataView`.

Here is the current API:
```cs
public static OnnxScoringEstimator ApplyOnnxModel(this TransformsCatalog catalog,
            string modelFile,
            string outputColumnName,
            string inputColumnName,
            int? gpuDeviceId = null,
            bool fallbackToCpu = false);
```

This should be re-ordered to be:

```cs
public static OnnxScoringEstimator ApplyOnnxModel(this TransformsCatalog catalog,
            string outputColumnName,
            string inputColumnName,
            string modelFile,
            int? gpuDeviceId = null,
            bool fallbackToCpu = false);
```

Related to #3079

rogancarr:
This PR updates the `ApplyOnnxModel` transform to meet the API standards: Moving to `outputColumnName`, `inputColumnName`, `modelFile`.

Fixes #3082

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/OnnxTransform.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/OnnxTransform.cs
@@ -34,7 +34,7 @@ namespace Microsoft.ML.Samples.Dynamic
             var mlContext = new MLContext();
             var data = GetTensorData();
             var idv = mlContext.Data.LoadFromEnumerable(data);
-            var pipeline = mlContext.Transforms.ApplyOnnxModel(modelPath, new[] { outputInfo.Key }, new[] { inputInfo.Key });
+            var pipeline = mlContext.Transforms.ApplyOnnxModel(new[] { outputInfo.Key }, new[] { inputInfo.Key }, modelPath);
 
             // Run the pipeline and get the transformed values
             var transformedValues = pipeline.Fit(idv).Transform(idv);
--- a/src/Microsoft.ML.OnnxTransformer/OnnxCatalog.cs
+++ b/src/Microsoft.ML.OnnxTransformer/OnnxCatalog.cs
@@ -33,15 +33,15 @@ namespace Microsoft.ML
         /// Applies a pre-trained Onnx model.
         /// </summary>
         /// <param name="catalog">The transform's catalog.</param>
-        /// <param name="modelFile">The path of the file containing the ONNX model.</param>
         /// <param name="outputColumnName">The output column resulting from the transformation.</param>
         /// <param name="inputColumnName">The input column.</param>
+        /// <param name="modelFile">The path of the file containing the ONNX model.</param>
         /// <param name="gpuDeviceId">Optional GPU device ID to run execution on, <see langword="null" /> to run on CPU.</param>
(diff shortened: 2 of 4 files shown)


src/Microsoft.ML.OnnxTransformer/OnnxCatalog.cs
```diff
@@ -33,15 +33,15 @@ public static class OnnxCatalog
...
-        /// <param name="modelFile">The path of the file containing the ONNX model.</param>
         /// <param name="outputColumnName">The output column resulting from the transformation.</param>
         /// <param name="inputColumnName">The input column.</param>
+        /// <param name="modelFile">The path of the file containing the ONNX model.</param>
         /// <param name="gpuDeviceId">Optional GPU device ID to run execution on, <see langword="null" /> to run on CPU.</param>
         /// <param name="fallbackToCpu">If GPU error, raise exception or fallback to CPU.</param>
         public static OnnxScoringEstimator ApplyOnnxModel(this TransformsCatalog catalog,
-            string modelFile,
             string outputColumnName,
             string inputColumnName,
```
sfilipi: >inputColumnName [](start = 19, length = 15)

should this default to the outputColumnName like the other inputColumns?

shmoradims: my comment is the same. 

---
In reply to: [268883194](https://github.com/dotnet/machinelearning/pull/3086#discussion_r268883194) [](ancestors = 268883194)

rogancarr: I went back and forth on this. I don't think we need to have this as an "in-place" transform because we expect models to create new columns, similar to the TensorFlow transform.

docs/samples/Microsoft.ML.Samples/Dynamic/OnnxTransform.cs
```diff
@@ -34,7 +34,7 @@ public static void Example()
             var mlContext = new MLContext();
             var data = GetTensorData();
             var idv = mlContext.Data.LoadFromEnumerable(data);
-            var pipeline = mlContext.Transforms.ApplyOnnxModel(modelPath, new[] { outputInfo.Key }, new[] { inputInfo.Key });
+            var pipeline = mlContext.Transforms.ApplyOnnxModel(new[] { outputInfo.Key }, new[] { inputInfo.Key }, modelPath);
```
zeahmed: >modelPath [](start = 114, length = 9)

I see at line 20 where session is loaded. Does the `ApplyOnnxModel` reload the model? If it does then there needs to be a way to reuse the` session` object created above.

rogancarr: That is a good point. Can you file an issue on this? We can do this as we start to fix outstanding issues in the ONNX implementation.

---
In reply to: [268890248](https://github.com/dotnet/machinelearning/pull/3086#discussion_r268890248) [](ancestors = 268890248)

PR #3088, opened 2019-03-25
API to return ColumnPairs for a OneToOneTransformer.

### LINKED ISSUE #3087: API to return ColumnPairs for a transformer.
We need an API to return ColumnPairs for transformers descending from OneToOneTransformerBase. Create such an API in the experimental nuget.

codemzs:
fixes #3087

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
+++ b/src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
@@ -36,10 +36,12 @@ namespace Microsoft.ML
         }
 
         [BestFriend]
-        internal static (string outputColumnName, string inputColumnName)[] ConvertToValueTuples(InputOutputColumnPair[] infos)
-        {
-            return infos.Select(info => (info.OutputColumnName, info.InputColumnName)).ToArray();
-        }
+        internal static (string outputColumnName, string inputColumnName)[] ConvertToValueTuples(InputOutputColumnPair[] infos) =>
+            infos.Select(info => (info.OutputColumnName, info.InputColumnName)).ToArray();
+
+        [BestFriend]
+        internal static  InputOutputColumnPair[] ConvertFromValueTuples((string outputColumnName, string inputColumnName)[] infos) =>
+            infos.Select(v => new InputOutputColumnPair(v.outputColumnName, v.inputColumnName)).ToArray();
     }
 
     /// <summary>
--- a/src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
+++ b/src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Collections.Generic;
 using System.Linq;
 using Microsoft.ML.Data;
 using Microsoft.ML.Runtime;
@@ -40,8 +41,8 @@ namespace Microsoft.ML
             infos.Select(info => (info.OutputColumnName, info.InputColumnName)).ToArray();
 
         [BestFriend]
-        internal static InputOutputColumnPair[] ConvertFromValueTuples((string outputColumnName, string inputColumnName)[] infos) =>
-            infos.Select(info => new InputOutputColumnPair(info.outputColumnName, info.inputColumnName)).ToArray();
+        internal static IReadOnlyList<InputOutputColumnPair> ConvertFromValueTuples((string outputColumnName, string inputColumnName)[] infos) =>
+            infos.Select(info => new InputOutputColumnPair(info.outputColumnName, info.inputColumnName)).ToList().AsReadOnly();
     }
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
```diff
@@ -36,10 +36,12 @@ public InputOutputColumnPair(string outputColumnName, string inputColumnName = n
...
         [BestFriend]
-        internal static (string outputColumnName, string inputColumnName)[] ConvertToValueTuples(InputOutputColumnPair[] infos)
-        {
-            return infos.Select(info => (info.OutputColumnName, info.InputColumnName)).ToArray();
-        }
+        internal static (string outputColumnName, string inputColumnName)[] ConvertToValueTuples(InputOutputColumnPair[] infos) =>
+            infos.Select(info => (info.OutputColumnName, info.InputColumnName)).ToArray();
+
+        [BestFriend]
+        internal static  InputOutputColumnPair[] ConvertFromValueTuples((string outputColumnName, string inputColumnName)[] infos) =>
```
artidoro: >   [](start = 23, length = 2)

nit: there are two spaces here I think.

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
```diff
@@ -36,10 +36,12 @@ public InputOutputColumnPair(string outputColumnName, string inputColumnName = n
...
+            infos.Select(v => new InputOutputColumnPair(v.outputColumnName, v.inputColumnName)).ToArray();
```
artidoro: >v [](start = 25, length = 1)

nit: could be a good thing if we reconcile with the above `info => ...` instead of `v => ...`

src/Microsoft.ML.Experimental/OneToOneTransformerBaseExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.ML.Data;
+
+namespace Microsoft.ML.Experimental
+{
+    public static class OneToOneTransformerBaseExtensions
+    {
+        /// <summary>
+        /// Returns Input/Output column pair(s) for a <see cref="OneToOneTransformerBase"/>
```
artidoro:
```suggestion
        /// Returns the names of the input-output column pairs on which the transformation is applied.
```

src/Microsoft.ML.Experimental/OneToOneTransformerBaseExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+using Microsoft.ML.Data;
+
+namespace Microsoft.ML.Experimental
+{
+    public static class OneToOneTransformerBaseExtensions
+    {
+        /// <summary>
+        /// Returns the names of the input-output column pairs on which the transformation is applied.
+        /// </summary>
+        public static InputOutputColumnPair[] GetColumnPairs(this OneToOneTransformerBase transformer) => InputOutputColumnPair.ConvertFromValueTuples(transformer.ColumnPairs);
```
wschin: Do we want to return `IReadOnlyList`, which is not mutable?

codemzs:
@TomFinley Please leave any comments you may have and I will address this in another PR. I have sign-offs from @artidoro, @wschin and @Ivanidzo4ka for now. 

CC: @artidoro

TomFinley:
> @TomFinley Please leave any comments you may have and I will address them in another PR. I have sign-offs from @artidoro, @wschin and @Ivanidzo4ka for now.

Hi @codemzs and @artidoro, I am replying here with my opinions. Short reaction is, I think this is a good change, and entirely in the spirit of what I intended to be done with the experimental nuget, pursuant to #2279, and I am happy where it ended up. I see no reason for a new PR at this time.

One complaint: It would have been a bit cleaner to just make an `internal` method on `OneToOneTransformerBase` directly that returns these pairs, and just call that method simply. The way in which it was actually written (to make the *field* public, then operate on that field across the assembly) represents a needless breaking of the principle of encapsulation. I mean, there was no point to it, and now the class is significantly harder to understand. (Every time you make a central structure in a class more broadly accessible, you make that class harder to comprehend.)

Aside from breaking encapsulation, looking at it from a more general perspective, if we consider that the experimental API represents candidates for things that *might* become public, it is fair cleaner if the `internal` method in the stable code mirrors precisely the `public` extension method on the experimental code. That way, if we decide, "you know what, this should be public," it's just a matter of deleting the extension method then changing the `internal` to `public`. So that represents a simpler transition.

But, these are things that can be fixed later, as they deal with internal structures. No urgency.

Beyond that, in side communications I sensed concern that this might not be enough information, that the return type of this input/output pair is not right. My answer is: yes, you're probably right, but I've heard so many requests for information out of `ITransformer`s (ranging the gamut from the reasonable, as this is, to the ridiculous), that I start to suspect an API that could make everyone happy is impossible. I suspect that where we wind up will be something a bit more sophisticated than this, but what the shape of that will look like, ***I have no idea.*** So, given that I have no idea, I'm happy to start with something simple and easy to do (as @codemzs has done here) made to address a very specific, limited request for information from some people, so we at least have a starting point. And this sort of working in a 'new area' is precisely why we have this experimental project, and other experimental projects!

So, I think it's good.

PR #3101, opened 2019-03-26
Restore OVA ability to preserve key names on predicted label

### LINKED ISSUE #3090: Exception when converting PredictedLabel from Key To Value
When trying to convert the output PredictedLabel from key back to value, the exception
```
System.InvalidOperationException
  HResult=0x80131509
  Message=Metadata KeyValues does not exist
  Source=Microsoft.ML.Core
  StackTrace:
   at Microsoft.ML.Runtime.Contracts.Check(IExceptionContext ctx, Boolean f, String msg) in C:\MLDotNet2\src\Microsoft.ML.Core\Utilities\Contracts.cs:line 497
```
is thrown.

Code to repro:
```C#

var textLoaderOptions = new TextLoader.Options()
{
	Columns = new[]
	{
		new TextLoader.Column("Label", DataKind.Single, 0),
		new TextLoader.Column("Row", DataKind.Single, 1),
		new TextLoader.Column("Column", DataKind.Single, 2),
	},
	HasHeader = true,
	Separators = new[] { '\t' }
};
var textLoader = mlContext.Data.CreateTextLoader(textLoaderOptions);
var data = textLoader.Load(@"C:\MLDotNet2\test\data\trivial-train.tsv");

var ap = mlContext.BinaryClassification.Trainers.AveragedPerceptron();
var ova = mlContext.MulticlassClassification.Trainers.OneVersusAll(ap);

var pipeline = mlContext.Transforms.Conversion.MapValueToKey("Label")
	.Append(mlContext.Transforms.Concatenate("Features", "Row", "Column"))
	.Append(ova)
	.Append(mlContext.Transforms.Conversion.MapKeyToValue("PredictedLabel"));

var model = pipeline.Fit(data);
```

Replace `C:\MLDotNet2\` with path to ML.NET repo on your local machine

Ivanidzo4ka:
Fixes #3090.
I found usage of slot names _slightly_ confusing. In same time we have `TrainingLabelValues` which should do the trick.

### DIFF
--- a/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
+++ b/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
@@ -441,8 +441,10 @@ namespace Microsoft.ML.Data
         public static IEnumerable<SchemaShape.Column> AnnotationsForMulticlassScoreColumn(SchemaShape.Column? labelColumn = null)
         {
             var cols = new List<SchemaShape.Column>();
-            if (labelColumn != null && labelColumn.Value.IsKey && NeedsSlotNames(labelColumn.Value))
-                cols.Add(new SchemaShape.Column(Kinds.SlotNames, SchemaShape.Column.VectorKind.Vector, TextDataViewType.Instance, false));
+            if (labelColumn != null && labelColumn.Value.IsKey &&
+                labelColumn.Value.Annotations.TryFindColumn(Kinds.KeyValues, out var metaCol) &&
+                metaCol.Kind == SchemaShape.Column.VectorKind.Vector)
+                cols.Add(new SchemaShape.Column(Kinds.TrainingLabelValues, SchemaShape.Column.VectorKind.Vector, metaCol.ItemType, false));
             cols.AddRange(GetTrainerOutputAnnotation());
             return cols;
         }
--- a/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
+++ b/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
@@ -442,7 +442,7 @@ namespace Microsoft.ML.Data
         {
             var cols = new List<SchemaShape.Column>();
             if (labelColumn != null && labelColumn.Value.IsKey)
-
+            {
                 if (labelColumn.Value.Annotations.TryFindColumn(Kinds.KeyValues, out var metaCol) &&
                     metaCol.Kind == SchemaShape.Column.VectorKind.Vector)
                 {
@@ -450,6 +450,7 @@ namespace Microsoft.ML.Data
                         cols.Add(new SchemaShape.Column(Kinds.SlotNames, SchemaShape.Column.VectorKind.Vector, TextDataViewType.Instance, false));
                     cols.Add(new SchemaShape.Column(Kinds.TrainingLabelValues, SchemaShape.Column.VectorKind.Vector, metaCol.ItemType, false));
                 }
+            }
             cols.AddRange(GetTrainerOutputAnnotation());
             return cols;
         }
--- a/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
+++ b/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
@@ -441,10 +441,15 @@ namespace Microsoft.ML.Data
         public static IEnumerable<SchemaShape.Column> AnnotationsForMulticlassScoreColumn(SchemaShape.Column? labelColumn = null)
         {
             var cols = new List<SchemaShape.Column>();
-            if (labelColumn != null && labelColumn.Value.IsKey &&
-                labelColumn.Value.Annotations.TryFindColumn(Kinds.KeyValues, out var metaCol) &&
-                metaCol.Kind == SchemaShape.Column.VectorKind.Vector)
-                cols.Add(new SchemaShape.Column(Kinds.TrainingLabelValues, SchemaShape.Column.VectorKind.Vector, metaCol.ItemType, false));
+            if (labelColumn != null && labelColumn.Value.IsKey)
+
+                if (labelColumn.Value.Annotations.TryFindColumn(Kinds.KeyValues, out var metaCol) &&
+                    metaCol.Kind == SchemaShape.Column.VectorKind.Vector)
+                {
+                    if (metaCol.ItemType is TextDataViewType)
+                        cols.Add(new SchemaShape.Column(Kinds.SlotNames, SchemaShape.Column.VectorKind.Vector, TextDataViewType.Instance, false));
+                    cols.Add(new SchemaShape.Column(Kinds.TrainingLabelValues, SchemaShape.Column.VectorKind.Vector, metaCol.ItemType, false));
+                }
             cols.AddRange(GetTrainerOutputAnnotation());
             return cols;
         }
--- a/src/Microsoft.ML.Data/Scorers/MulticlassClassificationScorer.cs
+++ b/src/Microsoft.ML.Data/Scorers/MulticlassClassificationScorer.cs
@@ -390,7 +390,7 @@ namespace Microsoft.ML.Data
             if (trainSchema?.Label == null)
                 return mapper; // We don't even have a label identified in a training schema.
             var keyType = trainSchema.Label.Value.Annotations.Schema.GetColumnOrNull(AnnotationUtils.Kinds.KeyValues)?.Type as VectorDataViewType;
-            if (keyType == null || !CanWrap(mapper, keyType))
+            if (keyType == null)
                 return mapper;
 
             // Great!! All checks pass.
@@ -409,11 +409,19 @@ namespace Microsoft.ML.Data
         /// from the model of a bindable mapper)</param>
         /// <returns>Whether we can call <see cref="LabelNameBindableMapper.CreateBound{T}"/> with
         /// this mapper and expect it to succeed</returns>
-        internal static bool CanWrap(ISchemaBoundMapper mapper, DataViewType labelNameType)
+        internal static bool CanWrapTrainingLabels(ISchemaBoundMapper mapper, DataViewType labelNameType)
+        {
+            if (GetTypesForWrapping(mapper, labelNameType, AnnotationUtils.Kinds.TrainingLabelValues, out var scoreType))
+                // Check that the type is vector, and is of compatible size with the score output.
+                return labelNameType is VectorDataViewType vectorType && vectorType.Size == scoreType.GetVectorSize();
+            return false;
+        }
+
+        internal static bool GetTypesForWrapping(ISchemaBoundMapper mapper, DataViewType labelNameType, string metaKind, out DataViewType scoreType)
         {
             Contracts.AssertValue(mapper);
             Contracts.AssertValue(labelNameType);
-
+            scoreType = null;
             ISchemaBoundRowMapper rowMapper = mapper as ISchemaBoundRowMapper;
             if (rowMapper == null)
                 return false; // We could cover this case, but it is of no practical worth as far as I see, so I decline to do so.
@@ -423,12 +431,18 @@ namespace Microsoft.ML.Data
             var scoreCol = outSchema.GetColumnOrNull(AnnotationUtils.Const.ScoreValueKind.Score);
             if (!outSchema.TryGetColumnIndex(AnnotationUtils.Const.ScoreValueKind.Score, out scoreIdx))
                 return false; // The mapper doesn't even publish a score column to attach the metadata to.
-            if (outSchema[scoreIdx].Annotations.Schema.GetColumnOrNull(AnnotationUtils.Kinds.SlotNames)?.Type != null)
+            if (outSchema[scoreIdx].Annotations.Schema.GetColumnOrNull(AnnotationUtils.Kinds.TrainingLabelValues)?.Type != null)
                 return false; // The mapper publishes a score column, and already produces its own slot names.
-            var scoreType = outSchema[scoreIdx].Type;
+            scoreType = outSchema[scoreIdx].Type;
+            return true;
+        }
 
-            // Check that the type is vector, and is of compatible size with the score output.
-            return labelNameType is VectorDataViewType vectorType && vectorType.Size == scoreType.GetVectorSize();
+        internal static bool CanWrapSlotNames(ISchemaBoundMapper mapper, DataViewType labelNameType)
+        {
+            if (GetTypesForWrapping(mapper, labelNameType, AnnotationUtils.Kinds.SlotNames, out var scoreType))
+                // Check that the type is vector, and is of compatible size with the score output.
+                return labelNameType is VectorDataViewType vectorType && vectorType.Size == scoreType.GetVectorSize() && vectorType.ItemType == TextDataViewType.Instance;
+            return false;
         }
 
         internal static ISchemaBoundMapper WrapCore<T>(IHostEnvironment env, ISchemaBoundMapper mapper, RoleMappedSchema trainSchema)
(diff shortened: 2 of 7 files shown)


src/Microsoft.ML.Data/Scorers/MulticlassClassificationScorer.cs
```diff
@@ -450,7 +450,7 @@ internal static ISchemaBoundMapper WrapCore<T>(IHostEnvironment env, ISchemaBoun
                     trainSchema.Label.Value.GetKeyValues(ref value);
                 };
 
-            return LabelNameBindableMapper.CreateBound<T>(env, (ISchemaBoundRowMapper)mapper, type as VectorDataViewType, getter, AnnotationUtils.Kinds.SlotNames, CanWrap);
+            return LabelNameBindableMapper.CreateBound<T>(env, (ISchemaBoundRowMapper)mapper, type as VectorDataViewType, getter, AnnotationUtils.Kinds.TrainingLabelValues, CanWrap);
```
TomFinley: >AnnotationUtils.Kinds.TrainingLabelValues [](start = 130, length = 41)

Hi @Ivanidzo4ka, could you explain this to me a bit more? It looks like we are no longer publishing these as slot names. Is that correct, or do I misread?

More broadly: I understand from the issue that there has been some regression from 0.11, but I am not certain I understand the nature of the regression completely (how did it regress, by which change?). That lack of knowledge makes this code difficult to review for me, even though the change itself is not too large.

test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs
```diff
@@ -37,22 +37,22 @@ void PredictAndMetadata()
...
             // During prediction we will get Score column with 3 float values.
             // We need to find way to map each score to original label.
-            // In order to do what we need to get SlotNames from Score column.
-            // Slot names on top of Score column represent original labels for i-th value in Score array.
-            VBuffer<ReadOnlyMemory<char>> slotNames = default;
-            engine.OutputSchema[nameof(IrisPrediction.Score)].GetSlotNames(ref slotNames);
+            // In order to do what we need to get TrainingLabelValues from Score column.
+            // TrainingLabelValues on top of Score column represent original labels for i-th value in Score array.
+            VBuffer<ReadOnlyMemory<char>> originalLabels = default;
+            engine.OutputSchema[nameof(IrisPrediction.Score)].Annotations.GetValue(AnnotationUtils.Kinds.TrainingLabelValues, ref originalLabels);
```
TomFinley: >TrainingLabelValues [](start = 105, length = 19)

Is the distinction with slot names that slots names *must* be text, while these might be any type?

test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs
```diff
@@ -37,22 +37,22 @@ void PredictAndMetadata()
...
             // Since we apply MapValueToKey estimator with default parameters, key values
             // depends on order of occurence in data file. Which is "Iris-setosa", "Iris-versicolor", "Iris-virginica"
             // So if we have Score column equal to [0.2, 0.3, 0.5] that's mean what score for
             // Iris-setosa is 0.2
             // Iris-versicolor is 0.3
             // Iris-virginica is 0.5.
-            Assert.True(slotNames.GetItemOrDefault(0).ToString() == "Iris-setosa");
-            Assert.True(slotNames.GetItemOrDefault(1).ToString() == "Iris-versicolor");
-            Assert.True(slotNames.GetItemOrDefault(2).ToString() == "Iris-virginica");
+            Assert.True(originalLabels.GetItemOrDefault(0).ToString() == "Iris-setosa");
```
TomFinley: >Assert.True [](start = 12, length = 11)

XUnit has something called `Assert.Equal`. The benefit to using something like that is, *if* the test fails, you get a more descriptive message than merely knowing that a boolean test failed somewhere.

Something to think about when writing tests.

src/Microsoft.ML.Data/Scorers/MulticlassClassificationScorer.cs
Ivanidzo4ka: I ask you some time ago regarding nature of SlotNames, and what they are always should be Text instances. So I made change in `CanWrap` method which validates what KeyValues are actually TextInstances.
Which bring this bug.
I don't want to continue to use `SlotNames`, since they are limited by type. In same time `TrainingLabelValues` don't look like they have any restriction on data type of it content, so I just want to use it.

test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs
Ivanidzo4ka: No we don't, but I can change that. Any reason while we want continue to propagate slotnames?

TomFinley: Well, let's imagine I write out a text file, and I have this scores column. With slot names, I get a descriptive header. Without it I don't. Does that make sense?

Ivanidzo4ka: Only if original labels were string, but ok.
Just before I make crucial mistake.
Do you prefer to have two wrappers on top of multiclass scorer one for `TrainingLabelValues` one for `SlotNames` or you would prefer to extend `LabelNameBindableMapper` to support multiple getters/ metakinds?

test/Microsoft.ML.Tests/TrainerEstimators/MetalinearEstimators.cs
```diff
@@ -51,6 +51,36 @@ public void OVAUncalibrated()
             Done();
         }
 
+        /// <summary>
+        /// Test what OVA preserves key values for label.
+        /// </summary>
+        [Fact]
+        public void OvaKeyNames()
```
TomFinley: >OvaKeyNames [](start = 20, length = 11)

Just curious, I heard something fairly troubling from @eerhardt, could we test this sort of scenario still works for things other than OVA?

eerhardt:
@Ivanidzo4ka - this issue doesn't appear to be OVA specific. I'm also hitting it in upgrading [XamlBrewer.Uwp.MachineLearningSample](https://github.com/XamlBrewer/UWP-MachineLearning-Sample/blob/150b11ed4941451e0aa48c7f71500069efa2c2fb/XamlBrewer.Uwp.MachineLearningSample/Models/MulticlassClassification/MulticlassClassificationModel.cs#L19-L30) to the latest build.

test/Microsoft.ML.Tests/TrainerEstimators/MetalinearEstimators.cs
eerhardt: Can we add this to `FunctionalTests` instead? Let's not keep adding tests that the product allows InternalsVisibleTo. That way we can test like a customer uses the product.

test/Microsoft.ML.Tests/TrainerEstimators/MetalinearEstimators.cs
```diff
@@ -51,6 +51,36 @@ public void OVAUncalibrated()
...
+
+            var ap = ML.BinaryClassification.Trainers.AveragedPerceptron();
+            var ova = ML.MulticlassClassification.Trainers.OneVersusAll(ap);
+
+            var pipeline = ML.Transforms.Conversion.MapValueToKey("Label")
+                .Append(ML.Transforms.Concatenate("Features", "Row", "Column"))
+                .Append(ova)
+                .Append(ML.Transforms.Conversion.MapKeyToValue("PredictedLabel"));
+
+            var model = pipeline.Fit(data);
```
eerhardt: We should do something with the model to ensure it was created correctly.

Ivanidzo4ka:
I've check LR, SDCA and multiclass, all of them working fine with my changes.

---
In reply to: [477327070](https://github.com/dotnet/machinelearning/pull/3101#issuecomment-477327070) [](ancestors = 477327070)

test/Microsoft.ML.Tests/TrainerEstimators/MetalinearEstimators.cs
Ivanidzo4ka: Any tests which use `TestEstimatorCore` will compare metadata from expected and resulted output schema.
I've change `AnnotationsForMulticlassScoreColumn` to always have TrainingLabelValue, and slotNames if key was text type, so technically we check that everywhere where we use `TestEstimatorCore`

---
In reply to: [269746627](https://github.com/dotnet/machinelearning/pull/3101#discussion_r269746627) [](ancestors = 269746627)

eerhardt:
Can you address this comment? https://github.com/dotnet/machinelearning/pull/3101#discussion_r269747281

```
eerhardt 5 days ago
Can we add this to FunctionalTests instead? Let's not keep adding tests that the product allows InternalsVisibleTo. That way we can test like a customer uses the product.
````

File: test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs [hunk 6]
```diff
@@ -37,22 +37,22 @@ void PredictAndMetadata()
...
+
             // During prediction we will get Score column with 3 float values.
             // We need to find way to map each score to original label.
-            // In order to do what we need to get SlotNames from Score column.
-            // Slot names on top of Score column represent original labels for i-th value in Score array.
-            VBuffer<ReadOnlyMemory<char>> slotNames = default;
-            engine.OutputSchema[nameof(IrisPrediction.Score)].GetSlotNames(ref slotNames);
+            // In order to do what we need to get TrainingLabelValues from Score column.
+            // TrainingLabelValues on top of Score column represent original labels for i-th value in Score array.
+            VBuffer<ReadOnlyMemory<char>> originalLabels = default;
```
TomFinley: >ReadOnlyMemory [](start = 20, length = 14)

In this particular case we should be propagating *both* slot names and label names, right? Since they're string in both cases? While I see the point in augmenting the test to cover this new metadata type, is there any particular reason to remove the test that the vector has teh appropriate slot names?

test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs
TomFinley: Multiple metadata kinds is good, thanks Ivan.

---
In reply to: [269742478](https://github.com/dotnet/machinelearning/pull/3101#discussion_r269742478) [](ancestors = 269742478)

test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs
Ivanidzo4ka: This is old scenario test. Purpose of it to show user how to do work with metadata.
In this particular test label is string, and it has slotnames, but it wont in case of non string label.
If I add slotnames here it would be confusing? What for do I get slotnames?

All tests with TestEstimatorCore routing would test on presence of slotnames and TrainingLabelValues. And we have plenty of them,. why should we do anything here with slotname?

---
In reply to: [271046772](https://github.com/dotnet/machinelearning/pull/3101#discussion_r271046772) [](ancestors = 271046772)

TomFinley: OK, that's fine. I had the idea that these "showing the user" things were more the sake of functional tests, but as you like.

---
In reply to: [271050961](https://github.com/dotnet/machinelearning/pull/3101#discussion_r271050961) [](ancestors = 271050961,271046772)

test/Microsoft.ML.Functional.Tests/Training.cs
```diff
@@ -467,5 +467,37 @@ public void MetacomponentsFunctionAsExpectedOva()
             // Evaluate the model.
             var binaryClassificationMetrics = mlContext.MulticlassClassification.Evaluate(binaryClassificationPredictions);
         }
+
+        /// <summary>
+        /// Training: Meta-compononts function as expected. For OVA (one-versus-all), a user will be able to specify only
```
eerhardt: (nit) `compononts`

test/Microsoft.ML.Functional.Tests/Training.cs
```diff
@@ -467,5 +467,37 @@ public void MetacomponentsFunctionAsExpectedOva()
...
+        /// binary classifier trainers. If they specify a different model class there should be a compile error.
+        /// </summary>
+        [Fact]
+        public void MetacomponentsFunctionWithKeyHandeling()
```
eerhardt: (nit) `Handeling`

test/Microsoft.ML.Functional.Tests/Training.cs
```diff
@@ -467,5 +467,37 @@ public void MetacomponentsFunctionAsExpectedOva()
...
+                .Append(mlContext.Transforms.Conversion.MapKeyToValue("PredictedLabel"));
+
+            // Fit the binary classification pipeline.
+            var binaryClassificationModel = binaryClassificationPipeline.Fit(data);
+
+            // Transform the data
+            var binaryClassificationPredictions = binaryClassificationModel.Transform(data);
+
+            // Evaluate the model.
+            var binaryClassificationMetrics = mlContext.MulticlassClassification.Evaluate(binaryClassificationPredictions);
```
eerhardt: Can we spot check some of these values to make sure the code isn't returning garbage?

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
```diff
@@ -131,7 +131,8 @@ private protected override NaiveBayesMulticlassModelParameters TrainModelCore(Tr
                     int size = cursor.Label + 1;
                     Utils.EnsureSize(ref labelHistogram, size);
                     Utils.EnsureSize(ref featureHistogram, size);
-                    Utils.EnsureSize(ref featureHistogram[cursor.Label], featureCount);
+                    if (featureHistogram[cursor.Label] == null)
+                        featureHistogram[cursor.Label] = new int[featureCount];
```
eerhardt: Was this change discovered by a test?

src/Microsoft.ML.Core/Data/AnnotationUtils.cs
```diff
@@ -441,8 +441,15 @@ public static IEnumerable<SchemaShape.Column> GetTrainerOutputAnnotation(bool is
         public static IEnumerable<SchemaShape.Column> AnnotationsForMulticlassScoreColumn(SchemaShape.Column? labelColumn = null)
         {
             var cols = new List<SchemaShape.Column>();
-            if (labelColumn != null && labelColumn.Value.IsKey && NeedsSlotNames(labelColumn.Value))
-                cols.Add(new SchemaShape.Column(Kinds.SlotNames, SchemaShape.Column.VectorKind.Vector, TextDataViewType.Instance, false));
+            if (labelColumn != null && labelColumn.Value.IsKey)
+
```
eerhardt: It would probably be good to wrap this in curly braces, now that it is more than one line (and that you have this blank line in between the if and the body.

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
Ivanidzo4ka: I wanted to make sure all multiclass learners works fine with my changes (so I run bunch of different one, on my test)
My test has only 2 features, and I got exception.
Mainly because Utils.EnsureSize use 4 as length for array even if you specify 1 or 2 or 3.
It make sense for VBuffer (since we have Count or Length, i'm always confused about which one is actual size of whole collection and which is size of elements in it), not sure why we do it for arrays as well.

---
In reply to: [271070400](https://github.com/dotnet/machinelearning/pull/3101#discussion_r271070400) [](ancestors = 271070400)

test/Microsoft.ML.Functional.Tests/Training.cs
Ivanidzo4ka: how it's different from what I have now?


---
In reply to: [271069790](https://github.com/dotnet/machinelearning/pull/3101#discussion_r271069790) [](ancestors = 271069790)

Ivanidzo4ka: did you mean `Handling`?

---
In reply to: [271072076](https://github.com/dotnet/machinelearning/pull/3101#discussion_r271072076) [](ancestors = 271072076,271069790)

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
eerhardt: Would it make sense to put all those tests into regression? That way we can catch bugs like this in the future?

test/Microsoft.ML.Functional.Tests/Training.cs
eerhardt: `Handeling` isn't a word (at least in English).

PR #3115, opened 2019-03-27
Support for CNTK

wwExpert:
Hello, what is the reason that there is no support for CNTK? I think it´s easier to support CNTK instead of tensorflow.


Ivanidzo4ka:
Here is activities in tensorflow repository: https://github.com/tensorflow/tensorflow/pulse
and here is activities in CNTK repo: https://github.com/Microsoft/CNTK/pulse
Also I think you meant to create issue, not PR.

PR #3116, opened 2019-03-27
Remove generic normalizer estimator catalog methods.

### LINKED ISSUE #3161: Remove generic normalizer estimator catalog methods.
We now have a method for each of the normalizer type so we should get rid of methods that accept enum for normalizer type.

codemzs:
fixes #3161

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Normalizer.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Normalizer.cs
@@ -1,8 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
 using Microsoft.ML.Data;
-using Microsoft.ML.Transforms;
+using Microsoft.ML.Experimental;
 
 namespace Microsoft.ML.Samples.Dynamic
 {
@@ -28,7 +27,7 @@ namespace Microsoft.ML.Samples.Dynamic
             // 35   1       6-11yrs     1         3         32            5  ...
 
             // A pipeline for normalizing the Induced column.
-            var pipeline = ml.Transforms.Normalize("Induced");
+            var pipeline = ml.Transforms.NormalizeMinMax("Induced");
             // The transformed (normalized according to Normalizer.NormalizerMode.MinMax) data.
             var transformer = pipeline.Fit(trainData);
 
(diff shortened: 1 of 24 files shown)


src/Microsoft.ML.Transforms/NormalizerCatalog.cs
```diff
@@ -28,29 +28,6 @@ public static class NormalizationCatalog
...
-        /// <param name="columns">The pairs of input and output columns.</param>
-        /// <example>
-        /// <format type="text/markdown">
-        /// <![CDATA[
-        /// [!code-csharp[Normalize](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Normalizer.cs)]
-        /// ]]>
-        /// </format>
-        /// </example>
-        [BestFriend]
-        internal static NormalizingEstimator Normalize(this TransformsCatalog catalog,
```
Ivanidzo4ka: @artidoro made all multiple columns extensions methods internal, and I think he did that for a reason, rather than just deleting them.
I would ask him regarding deleting this internal methods.

codemzs: @Ivanidzo4ka @eerhardt and @TomFinley feel we should remove such methods that are taking enum for normalizer kind because a dedicated method for each normalize kind exists.

docs/samples/Microsoft.ML.Samples/Dynamic/Normalizer.cs
```diff
@@ -1,8 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
 using Microsoft.ML.Data;
-using Microsoft.ML.Transforms;
+using Microsoft.ML.Experimental;
```
wschin: Is it just experimental? We are removing the generic estimator, so these new methods must be supported officially and therefore deserve a namespace like `ML.Transform`. Is my understanding correct?

codemzs: https://github.com/dotnet/machinelearning/pull/3118

TomFinley:
As elsewhere noted, this should not be checked in independent of #3118 since, if it were to go in, it would leave the code in a bad state where there is no non-experimental means of normalizing columns. I believe you've correspondingly merged these changes in with #3116, so, close @codemzs ?

PR #3118, opened 2019-03-27
Move Normalizer extension method from experimental to stable nuget and remove Normalizer generic APIs

### LINKED ISSUE #3109: Microsoft.ML nuget package no longer has a way to specify number of bins for binning normalization
In `v0.11.0`, it was possible to write code like the following:

```C#
var normalizer = mLContext.Transforms.Normalize(
    new NormalizingEstimator.BinningColumnOptions(outputColumnName: "Label", numBins: 2));
```

This allowed you to create a `binning` normalizer with the number of bins set to `2`.

However, this API is no longer public in `Microsoft.ML`. There is a public API where you can specify `mode: NormalizingEstimator.NormalizationMode.Binning`, but there is no way to set the number of bins. So when you use this mode, you always `MaximumBinCount` set to the default `1024`.

### LINKED ISSUE #3161: Remove generic normalizer estimator catalog methods.

codemzs:
fixes #3109
fixes #3161


eerhardt:
Do we also want to get rid of this method?

```C#
public static NormalizingEstimator Normalize(this TransformsCatalog catalog,
           string outputColumnName, string inputColumnName = null,
            NormalizingEstimator.NormalizationMode mode = NormalizingEstimator.NormalizationMode.MinMax)
            => new NormalizingEstimator(CatalogUtils.GetEnvironment(catalog), outputColumnName, inputColumnName ?? outputColumnName, mode);
```

Do we really need a public method that takes an `enum` (but doesn't allow you to specify the specific values) as well as other public methods that basically just extract that `enum` into individual overloads.

TomFinley:
Wonder what @artidoro thinks, should we add the multi-column mapping capability here? It seems more or less harmless to do so.

src/Microsoft.ML.Transforms/NormalizerCatalog.cs
```diff
@@ -51,6 +51,106 @@ public static class NormalizationCatalog
             return new NormalizingEstimator(env, mode, InputOutputColumnPair.ConvertToValueTuples(columns));
```
TomFinley: This new method looks good... the new method looks a lot better than the old method. :) I wonder if the old method that just took an "enum" to specify the type of normalization is a good idea. Or if it's already too late to remove. :P Maybe @artidoro has opinion?

artidoro: Since we are adding these new extensions to the stable api, I think it would be a good idea to remove the old method if still possible.

artidoro:
Regarding multi-column mapping, I am not opposed to it here. My only concern is that we will have a lot of overloads for the normalizers, especially if we add another one with an options class in the future. But since they are widely used it might be worth it to add them. They were on the list of transforms that @TomFinley  and @glebuk  identified as candidates for multi-column mapping with `InputOutputColumnPair`.

TomFinley:
> Do we also want to get rid of this method?
> 
> ```cs
> public static NormalizingEstimator Normalize(this TransformsCatalog catalog,
>            string outputColumnName, string inputColumnName = null,
>             NormalizingEstimator.NormalizationMode mode = NormalizingEstimator.NormalizationMode.MinMax)
>             => new NormalizingEstimator(CatalogUtils.GetEnvironment(catalog), outputColumnName, inputColumnName ?? outputColumnName, mode);
> ```
> 
> Do we really need a public method that takes an `enum` (but doesn't allow you to specify the specific values) as well as other public methods that basically just extract that `enum` into individual overloads.

@eerhardt I'd totally get behind that... only trouble is that's definitely a breaking change, so wondering what @shauheen thinks.

eerhardt:
If we really think that it's the right thing to do, we have time to do it now.

The alternative is to live with this API forever.

src/Microsoft.ML.Transforms/NormalizerCatalog.cs
TomFinley: Good. Perhaps we could prioritize this @codemzs since this will involve a breaking change in the API, and we must do that immediately?

---
In reply to: [270174092](https://github.com/dotnet/machinelearning/pull/3118#discussion_r270174092) [](ancestors = 270174092)

codemzs: @TomFinley Yes, I should be able to do this in the next hour or so.

codemzs:
@eric.erhardt@microsoft.com @tfinley@gmail.com This PR removes such redundant methods, please review and leave your autograph here: [https://github.com/dotnet/machinelearning/pull/3116](https://github.com/dotnet/machinelearning/pull/3116)

---
In reply to: [478105437](https://github.com/dotnet/machinelearning/pull/3118#issuecomment-478105437) [](ancestors = 478105437)

TomFinley:
As discussed elsewhere it is usually a bad idea to have PRs that leave the codebase in a suboptimal state until they're all checked in, this is usually a sign that they belonged together in the first place. Anyway, thanks for combining them.

---
In reply to: [478783520](https://github.com/dotnet/machinelearning/pull/3118#issuecomment-478783520) [](ancestors = 478783520,478105437)

test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamplesDynamicApi.cs
```diff
@@ -93,7 +92,7 @@ private void TrainRegression(string trainDataPath, string testDataPath, string m
             var pipeline =
                 // First 'normalize' the data (rescale to be
                 // between -1 and 1 for all examples), and then train the model.
-                mlContext.Transforms.Normalize("FeatureVector")
+                mlContext.Transforms.NormalizeMinMax("FeatureVector")
```
Ivanidzo4ka: [https://github.com/dotnet/machinelearning/blob/master/docs/code/MlNetCookBook.md](https://github.com/dotnet/machinelearning/blob/master/docs/code/MlNetCookBook.md)
you need to update that file with your changes

Ivanidzo4ka:
> using Microsoft.ML.Experimental;

is this one still needed?

---
Refers to: test/Microsoft.ML.Tests/Transformers/NormalizerTests.cs:10 in d705dc2. [](commit_id = d705dc2019e011b67126fdb12f6b9022b57c6020, deletion_comment = False)

src/Microsoft.ML.Transforms/NormalizerCatalog.cs
```diff
@@ -51,6 +32,106 @@ public static class NormalizationCatalog
...
         }
 
+        /// <summary>
+        /// Normalize (rescale) the column according to the <see cref="NormalizingEstimator.NormalizationMode.MinMax"/> mode.
```
Ivanidzo4ka: >NormalizingEstimator.NormalizationMode.MinMax [](start = 71, length = 45)

If we don't use it anymore, shall it become internal?

src/Microsoft.ML.Transforms/NormalizerCatalog.cs
```diff
@@ -51,6 +32,106 @@ public static class NormalizationCatalog
...
+        /// </summary>
+        /// <param name="catalog">The transform catalog</param>
+        /// <param name="outputColumnName">Name of the column resulting from the transformation of <paramref name="inputColumnName"/>.</param>
+        /// <param name="inputColumnName">Name of the column to transform. If set to <see langword="null"/>, the value of the <paramref name="outputColumnName"/> will be used as source.</param>
+        /// <param name="maximumExampleCount">Maximum number of examples used to train the normalizer.</param>
+        /// <param name="fixZero">Whether to map zero to zero, preserving sparsity.</param>
+        public static NormalizingEstimator NormalizeMinMax(this TransformsCatalog catalog,
+           string outputColumnName, string inputColumnName = null,
+           long maximumExampleCount = NormalizingEstimator.Defaults.MaximumExampleCount,
+           bool fixZero = NormalizingEstimator.Defaults.EnsureZeroUntouched)
```
wschin: >fixZero [](start = 16, length = 7)

ensureZeroUntouched.

test/Microsoft.ML.Benchmarks/KMeansAndLogisticRegressionBench.cs
```diff
@@ -35,7 +35,7 @@ public class KMeansAndLogisticRegressionBench
             }, hasHeader: true);
 
             var estimatorPipeline = ml.Transforms.Categorical.OneHotEncoding("CatFeatures")
-                .Append(ml.Transforms.Normalize("NumFeatures"))
+                .Append(ml.Transforms.NormalizeMinMax("NumFeatures"))
```
TomFinley: >NormalizeMinMax [](start = 38, length = 15)

Great, thanks @codemzs. Only real problem I see with this API is that it may not be apparent which normalization method is the best first choice as opposed to having it be named min-max, but we can always add alias and documentation later.

src/Microsoft.ML.Transforms/NormalizerCatalog.cs
TomFinley: Good point...

---
In reply to: [271101091](https://github.com/dotnet/machinelearning/pull/3118#discussion_r271101091) [](ancestors = 271101091)

src/Microsoft.ML.Transforms/NormalizerCatalog.cs
```diff
@@ -51,6 +32,106 @@ public static class NormalizationCatalog
...
+        /// <param name="catalog">The transform catalog</param>
+        /// <param name="outputColumnName">Name of the column resulting from the transformation of <paramref name="inputColumnName"/>.</param>
+        /// <param name="inputColumnName">Name of the column to transform. If set to <see langword="null"/>, the value of the <paramref name="outputColumnName"/> will be used as source.</param>
+        /// <param name="maximumExampleCount">Maximum number of examples used to train the normalizer.</param>
+        /// <param name="fixZero">Whether to map zero to zero, preserving sparsity.</param>
+        /// <param name="useCdf">Whether to use CDF as the output.</param>
+        public static NormalizingEstimator NormalizeMeanVariance(this TransformsCatalog catalog,
+            string outputColumnName, string inputColumnName = null,
+            long maximumExampleCount = NormalizingEstimator.Defaults.MaximumExampleCount,
+            bool fixZero = NormalizingEstimator.Defaults.EnsureZeroUntouched,
```
wschin: ensureZeroUntouched. Please also double-check functions below.

codemzs: [https://github.com/dotnet/machinelearning/pull/3052#pullrequestreview-217776054](https://github.com/dotnet/machinelearning/pull/3052#pullrequestreview-217776054)

---
In reply to: [271101741](https://github.com/dotnet/machinelearning/pull/3118#discussion_r271101741) [](ancestors = 271101741)

src/Microsoft.ML.Transforms/NormalizerCatalog.cs
TomFinley: Can we clarify this point? I had [this comment here](https://github.com/dotnet/machinelearning/pull/3052#pullrequestreview-217776054). I haven't checked, but it *looks* like that's not part of our public API. What is the source of this @wschin?

It's just such an odd name.

src/Microsoft.ML.Transforms/NormalizerCatalog.cs
```diff
@@ -51,6 +32,106 @@ public static class NormalizationCatalog
...
+            bool useCdf = NormalizingEstimator.Defaults.MeanVarCdf)
+        {
+            var columnOptions = new NormalizingEstimator.MeanVarianceColumnOptions(outputColumnName, inputColumnName, maximumExampleCount, fixZero, useCdf);
+            return new NormalizingEstimator(CatalogUtils.GetEnvironment(catalog), columnOptions);
+        }
+
+        /// <summary>
+        /// Normalize (rescale) the column according to the <see cref="NormalizingEstimator.NormalizationMode.LogMeanVariance"/> mode.
+        /// It normalizes the data based on the computed mean and variance of the logarithm of the data.
+        /// </summary>
```
wschin: This is not quite clear. Does this normalizer first applied log to each element in the input column? Or it's equivalent to a function, a * (x - b), where a is the log(variance of x) and b is average of x?

src/Microsoft.ML.Transforms/NormalizerCatalog.cs
```diff
@@ -51,6 +32,106 @@ public static class NormalizationCatalog
...
+            long maximumExampleCount = NormalizingEstimator.Defaults.MaximumExampleCount,
+            bool useCdf = NormalizingEstimator.Defaults.LogMeanVarCdf)
+        {
+            var columnOptions = new NormalizingEstimator.LogMeanVarianceColumnOptions(outputColumnName, inputColumnName, maximumExampleCount, useCdf);
+            return new NormalizingEstimator(CatalogUtils.GetEnvironment(catalog), columnOptions);
+        }
+
+        /// <summary>
+        /// Normalize (rescale) the column according to the <see cref="NormalizingEstimator.NormalizationMode.Binning"/> mode.
+        /// The values are assigned into bins with equal density.
```
wschin: May we add more details if documentation is a part of this PR?

codemzs:
> using Microsoft.ML.Experimental;

Yep for GetColumnPairs API...

---
In reply to: [478799620](https://github.com/dotnet/machinelearning/pull/3118#issuecomment-478799620) [](ancestors = 478799620)

---
Refers to: test/Microsoft.ML.Tests/Transformers/NormalizerTests.cs:10 in d705dc2. [](commit_id = d705dc2019e011b67126fdb12f6b9022b57c6020, deletion_comment = False)

src/Microsoft.ML.Transforms/NormalizerCatalog.cs
codemzs: documentation is not part of this PR. We have this month to improve documentation.

src/Microsoft.ML.Transforms/NormalizerCatalog.cs
```diff
@@ -51,6 +32,106 @@ public static class NormalizationCatalog
...
+        /// Normalize (rescale) the column according to the <see cref="NormalizingEstimator.NormalizationMode.MinMax"/> mode.
+        /// It normalizes the data based on the observed minimum and maximum values of the data.
+        /// </summary>
+        /// <param name="catalog">The transform catalog</param>
+        /// <param name="outputColumnName">Name of the column resulting from the transformation of <paramref name="inputColumnName"/>.</param>
+        /// <param name="inputColumnName">Name of the column to transform. If set to <see langword="null"/>, the value of the <paramref name="outputColumnName"/> will be used as source.</param>
+        /// <param name="maximumExampleCount">Maximum number of examples used to train the normalizer.</param>
+        /// <param name="fixZero">Whether to map zero to zero, preserving sparsity.</param>
+        public static NormalizingEstimator NormalizeMinMax(this TransformsCatalog catalog,
+           string outputColumnName, string inputColumnName = null,
```
wschin: We used to have example link in the removed API. Can we add it back here?

codemzs: I did add it back.

---
In reply to: [271102840](https://github.com/dotnet/machinelearning/pull/3118#discussion_r271102840) [](ancestors = 271102840)

codemzs:
Spoke offline. We will add it in a separate PR with its own issue and it needs to go through shiproom.

---
In reply to: [477744551](https://github.com/dotnet/machinelearning/pull/3118#issuecomment-477744551) [](ancestors = 477744551)

src/Microsoft.ML.Transforms/NormalizerCatalog.cs
codemzs: That example link is good for min max normalizer only.

src/Microsoft.ML.Transforms/NormalizerCatalog.cs
codemzs: We could probably address this as part for documentation work this month, no?

Issue #3095, opened 2019-03-26
Code documentation: Updating existing code documents


TomFinley:
Many of the [documents describing the code and the principles that make it work](https://github.com/dotnet/machinelearning/tree/366332047340bdb12cf5fe2a11f9a6e257af9b84/docs/code) need to be updated. Sometimes this updating will take the form of just updating the type names (e.g., what we used to call `ColumnType` is not `DataViewType`, `ICursor` interface has been subsumed by the `RowCursor` abstract class), and sometimes will take the form of changing the content rather entirely (e.g., we used to consider throwing in cursors bad practice, now we need to explain why we changed our mind, also, we've restricted key-types so that they are explicitly enumerations over sets rather than trying to do a bunch of additional things we never once found useful).

Note that this is distinct from #2054, which is more about user-facing documentation and samples. This is more about the documentation of the principles upon which our code relies.

shmoradims:
Seems like most of the issues are addressed by Tom's 4 PRs above. @TomFinley if there's any topic that's not covered by your recent PRs above, please let us know and I'll create a separate issue for it. Closing for now.

PR #1580, opened 2018-11-08
Introduce VBufferMutationContext and hide VBuffer.Count

eerhardt:
This PR executes on proposed changes (1) and (5) in https://github.com/dotnet/machinelearning/issues/608#issuecomment-433185895.

> 1. Keep Length, but hide/private Count

> 5. The most drastic proposed change is how to actually mutate a VBuffer.

This introduces `VBufferMutationContext`, and uses it in almost all the places where VBuffers are mutated. There are a few stragglers that I will fix in a subsequent PR, but I thought this change was large enough, so I cut it off once I could make `.Count` private.

Working towards #608.

PR #3136, opened 2019-03-29
Update VBuffer documentation

TomFinley:
Towards #3095, specifically about `VBuffer`s. Changes documentation to reflect the changes made by @eerhardt in his refactoring of #1580.

### DIFF
--- a/docs/code/VBufferCareFeeding.md
+++ b/docs/code/VBufferCareFeeding.md
@@ -155,9 +155,8 @@ point, probably the first call, `value` is replaced with a `VBuffer<float>`
 that has actual values, stored in freshly allocated buffers. In subsequent
 calls, *perhaps* these are judged as insufficiently large, and new arrays are
 internally allocated, but we would expect at some point the arrays would
-become "large enough" to accommodate many values, so reallocations would
-become increasingly rare. In this way, we avoid nearly all allocations and
-garbage collection.
+become "large enough," and we would no longer have to allocate new buffers and
+garbage collect old ones after that point.
 
 A common mistake made by first time users is to do something like move the
 `var value` declaration inside the `while` loop, thus dooming `getter` to have
@@ -233,11 +232,11 @@ stores the scaled result in `dst`.
 ScaleBy(in VBuffer<float> src, ref VBuffer<float> dst, float c)
 ```
 
-What this does is, copy the values from `src` to `dst`, while scaling each
-value from `src` seen by the factor `c`.
+What this does is, copy the values from `src` to `dst` while scaling the
+values we are copying by the factor `c`.
 
 One possible alternate (wrong) implementation of this would be to just say
-`dst=src` then edit `dst` and scale its contents by `c`. But, then `dst` and
+`dst=src` then edit `dst` and scale each value by `c`. But, then `dst` and
 `src` would share references to their internal arrays, completely compromising
 the caller's ability to do anything useful with `src`: if the caller were to
 pass `dst` into some other method that modified it, this could easily
--- a/docs/code/VBufferCareFeeding.md
+++ b/docs/code/VBufferCareFeeding.md
@@ -10,43 +10,32 @@ A `VBuffer<T>` is a generic type that supports both dense and sparse vectors
 over items of type `T`. This is the representation type for all
 [`VectorType`](IDataViewTypeSystem.md#vector-representations) instances in the
 `IDataView` ecosystem. When an instance of this is passed to a row cursor
-getter, the callee is free to take ownership of and re-use the arrays
-(`Values` and `Indices`).
+getter, the callee is free to take ownership of and re-use the buffers, which
+internally are arrays, and accessed externally as `ReadOnlySpan`s through the
+`GetValues` and `GetIndices` methods.
 
-A `VBuffer<T>` is a struct, and has the following `readonly` fields:
+A `VBuffer<T>` is a struct, and has the following most important members:
 
 * `int Length`: The logical length of the buffer.
 
-* `int Count`: The number of items explicitly represented. This equals `Length`
-when the representation is dense and is less than `Length` when sparse.
+* `bool IsDense`: Whether the vector is dense (if `true`) or sparse (if
+  `false`).
 
-* `T[] Values`: The values. Only the first `Count` of these are valid.
+* `ReadOnlySpan<T> GetValues()`: The values. If `IsDense` is `true`, then this
+  is of length exactly equal to `Length`. Otherwise, it will have length less
+  than `Length`.
 
-* `int[] Indices`: The indices. For a dense representation, this array is not
-  used, and may be `null`. For a sparse representation it is parallel to
-  values and specifies the logical indices for the corresponding values. Only
-  the first `Count` of these are valid.
+* `ReadOnlySpan<int> GetIndices()`: For a dense representation, this span will
+  have length of `0`, but for a sparse representation, this will be of the
+  same length as the span returned from `GetValues()`, and be parallel to it.
 
-`Values` must have length equal to at least `Count`. If the representation is
-sparse, that is, `Count < Length`, then `Indices` must have length also
-greater than or equal to `Count`. If `Count == 0`, then it is entirely legal
-for `Values` or `Indices` to be `null`, and if dense then `Indices` can always
-be `null`.
-
-On the subject of `Count == 0`, note that having no valid values in `Indices`
-and `Values` merely means that no values are explicitly defined, and the
-vector should be treated, logically, as being filled with `default(T)`.
-
-For sparse vectors, `Indices` must have length equal to at least `Count`, and
-the first `Count` indices must be increasing, with all indices between `0`
-inclusive and `Length` exclusive.
-
-Regarding the generic type parameter `T`, the only real assumption made about
-this type is that assignment (that is, using `=`) is sufficient to create an
-*independent* copy of that item. All representation types of the [primitive
-types](IDataViewTypeSystem.md#standard-column-types) have this property (for example,
-`DvText`, `DvInt4`, `Single`, `Double`, etc.), but for example, `VBuffer<>`
-itself does not have this property. So, no `VBuffer` of `VBuffer`s for you.
+Regarding the generic type parameter `T`, the basic assumption made about this
+type is that assignment (that is, using `=`) is sufficient to create an
+*independent* copy of that  item. All representation types of the [primitive
+types](IDataViewTypeSystem.md#standard-column-types) have this property (for
+example, `ReadOnlyMemory<char>`, `int`, `float`, `double`, etc.), but for
+example, `VBuffer<>` itself does not have this property. So, no `VBuffer` of
+`VBuffer`s is possible.
 
 ## Sparse Values as `default(T)`
 
@@ -73,7 +62,7 @@ As a corollary to the above note about equivalence of sparse and dense
 representations, since they are equivalent it follows that any code consuming
 `VBuffer`s must work equally well with *both*. That is, there must never be a
 condition where data is read and assumed to be either sparse, or dense, since
-implementers of `IDataView` and related interfaces are perfectly free to
+implementors of `IDataView` and related interfaces are perfectly free to
 produce either.
 
 The only "exception" to this rule is a necessary acknowledgment of the reality
@@ -83,57 +72,62 @@ not commutative, operations over sparse `VBuffer<float>` or `VBuffer<double>`
 vectors can sometimes result in modestly different results than the "same"
 operation over dense values.
 
-## Why Buffer Reuse
+There is also another point about sparsity, that is, about the increasing
+indices. While it is a *requirement*, it is regrettably not one we can
+practically enforce without a heavy performance cost. So, if you are creating
+your own `VBuffer`s, be careful that you are creating indices that are in
+strictly increasing order, and in the right range (non-negative and less than
+`Length`). We cannot check!
+
+# Buffer Reuse
 
 The question is often asked by people new to this codebase: why bother with
 buffer reuse at all? Without going into too many details, we used to not and
-suffered for it. We had a far simpler system where examples were yielded
-through an
-[`IEnumerable<>`](https://msdn.microsoft.com/en-us/library/9eekhta0.aspx), and
-our vector type at the time had `Indices` and `Values` arrays as well, but
-their sizes were there actual sizes, and being returned through an
-`IEnumerable<>` there was no plausible way to "recycle" the buffers.
+suffered for it. Long ago we had a far simpler system where examples were
+yielded through an [`IEnumerable<>`][1], and our vector type at the time had
+`Indices` and `Values`, but exposed as arrays, and their sizes were the actual
+"logical" sizes, and being returned through an `IEnumerable<>` there was no
+plausible way to reuse buffers.
 
-Also: who "owned" a fetched example (the caller, or callee) was not clear.
-Because it was not clear, code was inevitably written and checked in that made
+Also: who "owned" a fetched example (the caller, or callee) was not clear;
+code was at one time written under both assumptions, which was a mess. Because
+it was not clear, code was inevitably written and checked in that made
 *either* assumption, which meant, ultimately, that everything that touched
-these would try to duplicate everything by default, because doing anything
-else would fail in some case.
+these would try to duplicate everything, because doing anything else would
+fail in *some* case.
 
-The reason why this becomes important is because [garbage
-collection](https://msdn.microsoft.com/en-us/library/0xy59wtx.aspx) in the
-.NET framework is not free. Creating and destroying these arrays *can* be
-cheap, provided that they are sufficiently small, short lived, and only ever
-exist in a single thread. But, violate any of these, there is a possibility
-these arrays could be allocated on the large object heap, or promoted to gen-2
+The reason why this becomes important is because [garbage collection][2] in
+.NET is not free. Creating and destroying these arrays *can* be cheap,
+provided that they are sufficiently small, short lived, and only ever exist in
+a single thread. But, if that ever does not hold, there is a possibility these
+arrays could be allocated on the large object heap, or promoted to gen-2
 collection. The results could be disastrous: in one particularly memorable
 incident regarding neural net training, the move to `IDataView` and its
 `VBuffer`s resulted in a more than tenfold decrease in runtime performance,
 because under the old regime the garbage collection of the feature vectors was
 just taking so much time.
 
-This is somewhat unfortunate: a joke-that's-not-really-a-joke on the team was
-that we were writing C# as though it were C code. Be that as it may, buffer
-reuse is essential to our performance, especially on larger problems.
-
 This design requirement of buffer reuse has deeper implications for the
 ecosystem merely than the type here. For example, it is one crucial reason why
 so many value accessors in the `IDataView` ecosystem fill in values passed in
 through a `ref` parameter, rather than, say, being a return value.
 
+[1]: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerable-1?view=netstandard-2.0
+[2]: https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/index
+
 ## Buffer Re-use as a User
 
-Let's imagine we have an `IDataView` in a variable `dataview`, and we just so
+Let's imagine we have an `IDataView` in a variable `data`, and we just so
 happen to know that the column with index 5 has representation type
 `VBuffer<float>`. (In real code, this would presumably we achieved through
-more complicated involving an inspection of `dataview.Schema`, but we omit
-such details here.)
+more complicated involving an inspection of `data.Schema`, but we omit such
+details here.)
 
 ```csharp
-using (IRowCursor cursor = dataview.GetRowCursor(col => col == 5))
+using (DataViewRowCursor cursor = data.GetRowCursor(data.Schema[5]))
 {
     ValueGetter<VBuffer<float>> getter = cursor.GetGetter<VBuffer<float>>(5);
-    var value = default(VBuffer<float>);
+    VBuffer<float> value = default;
     while (cursor.MoveNext())
     {
         getter(ref value);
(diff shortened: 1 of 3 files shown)


src/Microsoft.ML.DataView/VBuffer.cs
```diff
@@ -197,13 +246,23 @@ public void CopyTo(ref VBuffer<T> dst, int srcMin, int length)
...
+        /// <summary>
+        /// Copy from this buffer to the given destination span, starting at the specified index. This "densifies."
+        /// </summary>
+        /// <param name="dst">The destination buffer. This <see cref="Span{T}.Length"/> must be at least <see cref="Length"/>
+        /// plus <paramref name="ivDst"/>.</param>
+        /// <param name="ivDst">The starting index of <paramref name="dst"/> at which to start copying</param>
+        /// <param name="defaultValue">The value to fill in for the implicit sparse entries. This is a potential exception to
+        /// general expectation of sparse <see cref="VBuffer{T}"/> that the implicit sparse entries have the default value
+        /// of <typeparamref name="T"/>.</param>
         public void CopyTo(Span<T> dst, int ivDst, T defaultValue = default(T))
```
TomFinley: >ivDst [](start = 44, length = 5)

Looks like we still have some acronyms and whatever this is lurking around. Not sure if care enough to change the API for v1...

glebuk:
>         /// The logical length of the buffer.

perhaps make it a bit clearer:
The logical length of the buffer.  The value would include counts of both sparse and dense values.

---
Refers to: src/Microsoft.ML.DataView/VBuffer.cs:49 in ce4618d. [](commit_id = ce4618de2b2b5370297b9a27191377134f81f340, deletion_comment = False)

src/Microsoft.ML.DataView/VBuffer.cs
```diff
@@ -251,16 +310,34 @@ public static void Copy(T[] src, int srcIndex, ref VBuffer<T> dst, int length)
...
         }
 
+        /// <summary>
+        /// Gets the item stored in this structure. In the case of a dense vector this is a simple lookup.
+        /// In the case of a sparse vector, it will try to find the entry with that index, and set <paramref name="dst"/>
+        /// to that stored value, or if no such value was found, assign it the default value.
+        /// </summary>
+        /// <param name="slot">The slot index, which must be a non-negative number less than <see cref="Length"/>.
+        /// If it is not in that </param>
+        /// <param name="dst">The value stored </param>
```
glebuk: >d  [](start = 45, length = 2)

oh no! missing Dot!

TomFinley: Missing a sentence more like.

src/Microsoft.ML.DataView/VBufferEditor.cs
```diff
@@ -139,15 +148,16 @@ public VBuffer<T> Commit()
...
+        /// This method allows to modify the length of the explicitly defined values.
         /// This is useful in sparse situations where the <see cref="VBufferEditor{T}"/>
         /// was created with a larger physical value count than was needed
         /// because the final value count was not known at creation time.
         /// </remarks>
         public VBuffer<T> CommitTruncated(int physicalValuesCount)
         {
-            Contracts.CheckParam(physicalValuesCount <= Values.Length, nameof(physicalValuesCount), "Updating physicalValuesCount during CommitTruncated cannot be greater than the original physicalValuesCount value used in Create.");
+            Contracts.CheckParam(physicalValuesCount <= Values.Length, nameof(physicalValuesCount),
+                "Updating " + nameof(physicalValuesCount) + " during " + nameof(CommitTruncated) +
```
glebuk: >"Updating "  [](start = 16, length = 12)

nit: perhaps convert it to a more modern version of
$"Updating {nameof(physicalValuesCount) } during { nameof(CommitTruncated)}"+ ...

glebuk:
> A `VBuffer<T>` is a generic type that supports both dense and sparse vectors

Perhaps, it should mention how to represent multi-dimensional data, such as multi-color pixels? (either howto or mention the limitation.)

---
Refers to: docs/code/VBufferCareFeeding.md:9 in ce4618d. [](commit_id = ce4618de2b2b5370297b9a27191377134f81f340, deletion_comment = False)

src/Microsoft.ML.DataView/VBufferEditor.cs
TomFinley: Hi @glebuk, just a quick note, this advice is incorrect. Try the following for yourself (relies on there being a name called values.

```csharp
const string a = "Hello " + nameof(values);
const string b = $"Hello {nameof(values)}";
```

You will note that the first will succeed, but the second will fail because this is not actually a constant. The reason this is relevant is, you'll remember I hope our advice that the strings in checks or asserts and whatnot should always be *constants*, so as to avoid needless composition of error messages that are never used.

Happily, even if i had committed this error, I actually wrote a code analyzer to stop people from making such a mistake, but it might be good to remember why this is a bad idea.

src/Microsoft.ML.DataView/VBuffer.cs
TomFinley: I think I was going to write something about explicit and implicit values, went off and wrote that on the class documentation so that it was a defined term, but then forgot to come back here to make use of that clarifying note. 😄

src/Microsoft.ML.DataView/VBufferEditor.cs
glebuk: Thanks!

---
In reply to: [270520617](https://github.com/dotnet/machinelearning/pull/3136#discussion_r270520617) [](ancestors = 270520617)

TomFinley:
> Perhaps, it should mention how to represent multi-dimensional data, such as multi-color pixels? (either howto or mention the limitation.)

Hmmm. I agree that's worth mentioning, but that really has nothing to do with `VBuffer<T>`, but `VectorDataViewType`. (I mean, the concept of multiple dimensions does not appear anywhere in any method related to `VBuffer` itself as far as I can see, even once.)

For that reason, would you mind if I saved this clarification to the point when I am working on updating the `IDataViewTypeSystem.md` document, where I'd also presumably change the documentation on some of the `DataViewType` XML comments? I'm not sure this change belongs on `VBuffer`, since again by itself `VBuffer` knows literally not one thing about multiple dimensions.

TomFinley:
> > ```
> >     /// The logical length of the buffer.
> > ```
> 
> perhaps make it a bit clearer:
> The logical length of the buffer. The value would include counts of both sparse and dense values.
> 
> Refers to: src/Microsoft.ML.DataView/VBuffer.cs:49 in [ce4618d](https://github.com/dotnet/machinelearning/commit/ce4618de2b2b5370297b9a27191377134f81f340). [](commit_id = [ce4618d](https://github.com/dotnet/machinelearning/commit/ce4618de2b2b5370297b9a27191377134f81f340), deletion_comment = False)

Sounds good. I also took the opportunity to reinforce the idea of implicit and explicit values.

TomFinley:
Thanks @glebuk, I updated it, let me know if you're happy with it yet.

src/Microsoft.ML.DataView/VBuffer.cs
```diff
@@ -301,32 +395,35 @@ internal VBufferEditor<T> GetEditor()
...
-            Contracts.CheckParam(newLogicalLength >= 0, nameof(newLogicalLength));
-            Contracts.CheckParam(valuesCount == null || valuesCount.Value <= newLogicalLength, nameof(valuesCount));
+            Contracts.CheckParam(newLogicalLength >= 0, nameof(newLogicalLength), "Must be non-negative.");
+            Contracts.CheckParam(valuesCount == null || valuesCount.Value >= 0, nameof(valuesCount),
+                "If specified, must be non-negative.");
+            Contracts.CheckParam(valuesCount == null || valuesCount.Value <= newLogicalLength, nameof(valuesCount),
+                "If specified, must be no greater than " + nameof(newLogicalLength));
 
-            valuesCount = valuesCount ?? newLogicalLength;
+            int logicallValuesCount = valuesCount ?? newLogicalLength;
```
eerhardt: type-o `logicall`

src/Microsoft.ML.DataView/VBuffer.cs
eerhardt: Note that changing a parameter name after you ship an API is a breaking change. So it's now or never.

src/Microsoft.ML.DataView/VBufferEditor.cs
eerhardt: Just a side note to this conversation (it won't directly affect this scenario),

The .NET team is working on making these string formatting more efficient. Check out https://github.com/dotnet/csharplang/issues/2302 if you are interested in learning more.

docs/code/VBufferCareFeeding.md
```diff
@@ -10,43 +10,32 @@ A `VBuffer<T>` is a generic type that supports both dense and sparse vectors
...
-greater than or equal to `Count`. If `Count == 0`, then it is entirely legal
-for `Values` or `Indices` to be `null`, and if dense then `Indices` can always
-be `null`.
-
-On the subject of `Count == 0`, note that having no valid values in `Indices`
-and `Values` merely means that no values are explicitly defined, and the
-vector should be treated, logically, as being filled with `default(T)`.
-
-For sparse vectors, `Indices` must have length equal to at least `Count`, and
-the first `Count` indices must be increasing, with all indices between `0`
```
eerhardt: I think some of this `GetIndices` information is still valuable:

> must be increasing, with all indices between `0` inclusive and `Length` exclusive.

TomFinley: Well, I wrote at some length on this subject in the XML docs where I gave more information on the absolute requirements, but a quick mention about the ranges and the fact that it (should be!) sorted is probably something worth repeating, I agree...

docs/code/VBufferCareFeeding.md
```diff
@@ -83,57 +72,62 @@ not commutative, operations over sparse `VBuffer<float>` or `VBuffer<double>`
...
-cheap, provided that they are sufficiently small, short lived, and only ever
-exist in a single thread. But, violate any of these, there is a possibility
-these arrays could be allocated on the large object heap, or promoted to gen-2
+these would try to duplicate everything, because doing anything else would
+fail in *some* case.
+
+The reason why this becomes important is because [garbage collection][2] in
+.NET is not free. Creating and destroying these arrays *can* be cheap,
+provided that they are sufficiently small, short lived, and only ever exist in
+a single thread. But, if that ever does not hold, there is a possibility these
```
eerhardt: `But, if that ever does not hold` 

maybe could be rewritten as

`But, when that does not hold`

src/Microsoft.ML.DataView/VBuffer.cs
TomFinley: Sorry, that's intentional. Mr. Spock and I are starting a mobile wireless service provider, called "Logicall." We're going to bring the big players down! This is just a bit of viral guerrilla marketing he and I are doing.

It's going great, by the way; we save a lot of time by avoiding lengthy strategy meetings. A quick mind meld, and we're totally on the same page. The only real trouble is he insisted we make our branded Bluetooth ear-headsets *also* work with pointed ears. You wouldn't believe how hard that was to get right. We eventually went with a modular silicone "ear grip" that can be swapped out as needed. What a pain!

...I'm sorry, I'm very confused. I just had a very vivid dream last night and only just now realized it was not real. Correcting.

docs/code/VBufferCareFeeding.md
TomFinley: That is better -- "if that ever" suggests the occurrence is rare, whereas "when" doesn't have the same implicit suggestion, thanks @eerhardt.

docs/code/VBufferCareFeeding.md
```diff
@@ -142,129 +136,130 @@ using (IRowCursor cursor = dataview.GetRowCursor(col => col == 5))
...
+again. Presumably the first time, or several, memory is allocated (though,
+internally, and invisibly to us). Initially `VBuffer<float> value = default`,
+that is, it has zero `Length` and similarly empty spans. Presumably at some
+point, probably the first call, `value` is replaced with a `VBuffer<float>`
+that has actual values, stored in freshly allocated buffers. In subsequent
+calls, *perhaps* these are judged as insufficiently large, and new arrays are
+internally allocated, but we would expect at some point the arrays would
 become "large enough" to accommodate many values, so reallocations would
-become increasingly rare.
+become increasingly rare. In this way, we eventually avoid nearly all
```
eerhardt: `we eventually avoid nearly all allocations and garbage collection.`

=>

`we avoid future buffer allocations and require less garbage collection.`

?

src/Microsoft.ML.DataView/VBuffer.cs
glebuk: ^ this should be included in this  code comment., for future generations.

---
In reply to: [270549701](https://github.com/dotnet/machinelearning/pull/3136#discussion_r270549701) [](ancestors = 270549701)

docs/code/VBufferCareFeeding.md
```diff
@@ -142,129 +136,130 @@ using (IRowCursor cursor = dataview.GetRowCursor(col => col == 5))
...
+
+First, one creates the `VBufferEditor<T>` structure using one of the
+`VBufferEditor.Create` methods. You pass in a `VBuffer<T>` as a `ref`
+parameter. (For this purpose, it is useful to remember that a
+`default(VBuffer<T>)` is itself a completely valid, though completely empty,
+`VBuffer<T>`.) The editor is past that line considered to "own" the internal
+structure of the `VBuffer<T>` you passed in. (So, correct code should not
+continue to use that `VBuffer<T>` structure from that point onwards.)
+
+Then, one parties on the `Span<T>` for values and, if in a sparse application,
```
eerhardt: ``one parties on the `Span<T>` for values``

=>

``one mutates the `Span<T>` Values``

?

docs/code/VBufferCareFeeding.md
```diff
@@ -142,129 +136,130 @@ using (IRowCursor cursor = dataview.GetRowCursor(col => col == 5))
...
+The `ValueGetter` indicated above is the most important example of this
+principle. By passing in an existing `VBuffer<T>` into that delegate by
+reference, they are giving the implementation control over it to use (or,
+reallocate) as necessary to store the resulting value. But, once the delegate
+returns with the value returned, the caller is now considered to own that
+buffer.
+
 There is a corollary on this point: because the caller owns any `VBuffer`,
-then you shouldn't do anything that irrevocably destroys their usefulness to
+then could should not do anything that irrevocably destroys its usefulness to
```
eerhardt: Needs re-wording - `then could should not do anything`

docs/code/VBufferCareFeeding.md
TomFinley: Well... I sort of consider eventually a more accurate statement, because if you think about sparse vectors, or variable size vectors, every time you have a value that exceeds the current allocation you must reallocate, so it's not like you avoid *future* allocations right away, you merely reach a state where the buffer has reached what is likely its maximum size and you become increasingly unlikely to need a reallocation, but that is something that happens eventually. So my word choice here was deliberate. But... I see your point it's a suitably subtle point that your simpler language will probably be best, so I will change.

docs/code/VBufferCareFeeding.md
TomFinley: Yes, I independently noticed this a few minutes before you commented on it. 😛 I wound up rephrasing the entire sentence rather than just change a single phrase, maybe see if you like the result when I post the next commit...

docs/code/VBufferCareFeeding.md
TomFinley: I meant to say "code" not "could." I think I'll post my commits now and not wait for you to finish, since this is the second time in a row we've independently noticed the same mistake in the docs, and I might save you some time potentially...

docs/code/VBufferCareFeeding.md
```diff
@@ -142,129 +136,130 @@ using (IRowCursor cursor = dataview.GetRowCursor(col => col == 5))
...
 stores the scaled result in `dst`.
 
 ```csharp
-VectorUtils.ScaleBy(ref VBuffer<float> src, ref VBuffer<float> dst, float c)
+ScaleBy(in VBuffer<float> src, ref VBuffer<float> dst, float c)
 ```
 
 What this does is, copy the values from `src` to `dst`, while scaling each
-value seen by `c`.
+value from `src` seen by the factor `c`.
```
eerhardt: ``value from `src` seen by the factor `c`.``

=>

``value in `src` by the factor `c`.``

?

docs/code/VBufferCareFeeding.md
eerhardt: Maybe we can also rework the previous sentence and tie this whole thing together. The point I was thinking is that: `at some point the arrays would become "large enough", and we would no longer have to allocate new buffers after that point`. That is a true statement - at some point we've reached the max it will need, and no longer have to allocate.

src/Microsoft.ML.DataView/VBuffer.cs
TomFinley: Well... I was hoping to not make this one of the cherry picked PRs. Let me do this, I'll do the review comments in one commit, then do the changes to parameters in an isolated commit, and then we can decide what we want to do.

docs/code/VBufferCareFeeding.md
TomFinley: OK. It's also a bit more compact. I also added an additional phrase to also mention garbage collection, since that's *really* the thing we used to observe killing us in terms of perf.

docs/code/VBufferCareFeeding.md
TomFinley: Well... that's not right now, is it though? We *don't* scale each value in `src` by `c`. In fact the entire point of the paragraph following that is an explanation of why that is precisely the wrong thing to do. I can try to rephrase to try to make it a bit more direct and awkwardly phrased, but I would not welcome the specific edit suggested.

src/Microsoft.ML.DataView/VBuffer.cs
TomFinley: OK, so, last commit titled `Update parameter names to avoid abbreviations, e.g., src => source and suchlike.` is up, I updated the parameter names in `VBuffer`. Names in `VBufferEditor` were fine. I also updated the documenation in the `.md` file, but I did not update the actual code that inspired the example since that is in the internal API. (But, the documentation uses the idea merely as a simple, easy to understand example, so I think that's fine.) I took some queues from things like [Array.Copy](https://docs.microsoft.com/en-us/dotnet/api/system.array.copy?view=netstandard-2.0#System_Array_Copy_System_Array_System_Int32_System_Array_System_Int32_System_Int32_), but without going to far as calling the parameters `VBuffer<T> sourceBuffer` since I felt that might be going a bit too far, I just named them things like `source` rather than go quasi-hungarian w.r.t. the names, so, just `source`, `destination`, etc., as I believe you also did in `VBufferEditor` @eerhardt (e.g., your `Create` methods take `destination`, not `destinationBuffer`).

src/Microsoft.ML.DataView/VBufferEditor.cs
```diff
@@ -26,12 +29,10 @@ public static class VBufferEditor
         /// Creates a <see cref="VBufferEditor{T}"/> using
         /// <paramref name="destination"/>'s values and indices buffers.
         /// </summary>
-        /// <param name="destination">
-        /// The destination buffer.
-        /// </param>
-        /// <param name="newLogicalLength">
-        /// The logical length of the new buffer being edited.
-        /// </param>
+        /// <param name="destination">The destination buffer. Note that the resulting <see cref="VBufferEditor{T}"/> is assumed to take ownership
```
eerhardt: (nit) I find it easier to read XML comments when the wording is separate from the XML, like it is in the `summary` comment above.

```C#
        /// <param name="destination">
        /// The destination buffer. Note that the resulting <see cref="VBufferEditor{T}"/> is assumed to take ownership
        /// of this passed in object, and so whatever <see cref="VBuffer{T}"/> was passed in as this parameter should not be used again, since its
        /// underlying buffers are being potentially reused.
        /// </param>
```
```

vs

```C#
        /// <param name="destination">The destination buffer. Note that the resulting <see cref="VBufferEditor{T}"/> is assumed to take ownership
        /// of this passed in object, and so whatever <see cref="VBuffer{T}"/> was passed in as this parameter should not be used again, since its
        /// underlying buffers are being potentially reused.</param>
```

Formatting it this way forces the reader to parse the XML in order to find where the words start.

src/Microsoft.ML.DataView/VBufferEditor.cs
```diff
@@ -68,6 +69,14 @@ public static class VBufferEditor
     /// An object capable of editing a <see cref="VBuffer{T}"/> by filling out
     /// <see cref="Values"/> (and <see cref="Indices"/> if the buffer is not dense).
     /// </summary>
+    /// <remarks>
+    /// The <see cref="VBuffer{T}"/> structure by itself is immutable. However, the purpose of <see cref="VBuffer{T}"/>
+    /// in general, and its typical usage through <see cref="ValueGetter{TValue}"/>, is to enable buffer re-use. That is
+    /// the purpose of this structure: for situations where one wants to exploit buffer re-use to avoid costly memory
```
eerhardt: (minor) I think you can get rid of `That is the purpose of this structure` at the beginning, since we later say: `there is this structure, as created....` below. Having both sounds redundant.

src/Microsoft.ML.DataView/VBufferEditor.cs
```diff
@@ -85,12 +94,13 @@ public static class VBufferEditor
         public readonly Span<int> Indices;
 
         /// <summary>
-        /// Gets a value indicating whether a new Values array was allocated.
+        /// Gets a value indicating whether a new <see cref="Values"/> array was allocated. If this is <see langword="false"/>,
+        /// then the resulting <see cref="VBuffer{T}"/> from <see cref="Commit"/>
```
eerhardt: Was this the end of the sentence?

TomFinley: Right, thanks Eric.

src/Microsoft.ML.DataView/VBufferEditor.cs
TomFinley: Yeah, I don't know, I think I was struggling here with my wording and wrote a whole lot of nothing somehow without realizing it...

src/Microsoft.ML.DataView/VBufferEditor.cs
TomFinley: Well, all right. Maybe just for the longer more discursive ones.

src/Microsoft.ML.DataView/VBuffer.cs
```diff
@@ -11,47 +11,75 @@
...
+        /// indices should be treated as being implicitly defined with the default value of <typeparamref name="T"/>.
         /// </summary>
         /// <remarks>
-        /// For example, if GetIndices() returns [3, 5] and GetValues() produces [98, 76], this VBuffer
-        /// stands for a vector with:
-        ///  - non-zeros values 98 and 76 respectively at the 4th and 6th coordinates
-        ///  - zeros at all other coordinates
+        /// To give one example, if <see cref="GetIndices"/> returns [3, 5] and <see cref="GetValues"/>() produces [98, 76],
+        /// this <see cref="VBuffer{T}"/> stands for a vector with non-zeros values 98 and 76 respectively at the 4th and 6th
+        /// coordinates, and zeros at all other coordinates. (Zero, because that is the default value for all .NET numeric
```
eerhardt: I think the two `coordinate` usages should be changed to `indices`. We use the term "Index" and "Indices" everywhere else. Calling it "coordinate" here is confusing IMO.

src/Microsoft.ML.DataView/VBuffer.cs
```diff
@@ -77,16 +117,27 @@ public VBuffer(int length, T[] values, int[] indices = null)
         }
 
         /// <summary>
-        /// Construct a possibly sparse representation.
+        /// Construct a possibly sparse vector representation.
         /// </summary>
+        /// <param name="length">The length of the constructed buffer.</param>
+        /// <param name="count">The count of explicit entries. This must be between 0 and <paramref name="count"/>, both
```
eerhardt: `This must be between 0 and <paramref name="count"/>`

=>

`This must be between 0 and <paramref name="length"/>`

src/Microsoft.ML.DataView/VBuffer.cs
```diff
@@ -77,16 +117,27 @@ public VBuffer(int length, T[] values, int[] indices = null)
...
+        /// inclusive. If it equals <paramref name="length"/> the result is a dense vector, and if less this will be a
+        /// sparse vector.</param>
+        /// <param name="values">The values to be used. This must be at least as long as <paramref name="count"/>. If
+        /// <paramref name="count"/> is 0, it is legal for this to be <see langword="null"/>.</param>
+        /// <param name="indices">The values to be used. If we are constructing a dense representation, or
```
eerhardt: `The values to be used.`  This must be a copy-paste error.  This is the `indices`.

src/Microsoft.ML.DataView/VBuffer.cs
```diff
@@ -228,50 +294,79 @@ public void CopyTo(Span<T> dst, int ivDst, T defaultValue = default(T))
...
 
-        public void GetItemOrDefault(int slot, ref T dst)
+        /// <summary>
+        /// Gets the item stored in this structure. In the case of a dense vector this is a simple lookup.
+        /// In the case of a sparse vector, it will try to find the entry with that index, and set <paramref name="destination"/>
+        /// to that stored value, or if no such value was found, assign it the default value.
+        /// </summary>
+        /// <remarks>
+        /// In the case where <see cref="IsDense"/> is <see langword="true"/>, this will take constant time since it an
+        /// directly lookup. For sparse vectors, however, because it must perform a bisection seach on the indices to
```
eerhardt: `seach` => `search`

src/Microsoft.ML.DataView/VBuffer.cs
```diff
@@ -228,50 +294,79 @@ public void CopyTo(Span<T> dst, int ivDst, T defaultValue = default(T))
...
+        ///
+        /// For that reason, a single completely isolated lookup, since constructing <see cref="ReadOnlySpan{T}"/> as
+        /// <see cref="GetValues"/> does is not a free operation, it may be more efficient to use this method. However
+        /// if one is doing a more involved computation involving many operations, it may be faster to utiltize
+        /// <see cref="GetValues"/> and, if appropriate, <see cref="GetIndices"/> directly.
+        /// </remarks>
+        /// <param name="slot">The slot index, which must be a non-negative number less than <see cref="Length"/>.</param>
+        /// <param name="destination">The value stored at that index, or if this is a sparse vector where this is an implicit
+        /// entry, the default value for <typeparamref name="T"/>.</param>
+        public void GetItemOrDefault(int slot, ref T destination)
```
eerhardt: Should we rename `slot`? Everywhere else we call this `index`.... We even had to specify in the param doc: `The slot index`.

src/Microsoft.ML.DataView/VBuffer.cs
```diff
@@ -111,32 +162,32 @@ public VBuffer(int length, int count, T[] values, int[] indices)
         /// <summary>
         /// Copy from this buffer to the given destination, forcing a dense representation.
         /// </summary>
-        public void CopyToDense(ref VBuffer<T> dst)
+        public void CopyToDense(ref VBuffer<T> destination)
```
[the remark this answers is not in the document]
TomFinley: Well, the person that originally wrote this file had the idea that if the summary documentation sufficed there was no need to go nuts on the parameter documentation, but, I'll just fill it in here, and in the other nearby method.

src/Microsoft.ML.DataView/VBuffer.cs
TomFinley: Yes, I agree, thanks Eric

src/Microsoft.ML.DataView/VBuffer.cs
TomFinley: Geez, all right, I guess if we're renaming `src` and whatnot we may as well change this too while we're at it.

src/Microsoft.ML.DataView/VBuffer.cs
TomFinley: BTW on the same subject of renamings, there was one place @eerhardt pointed out where an index was being named `slot` for some reason (since I guess we conceptually separate the concepts of "columns" and "slots" in data view, but, for VBuffer specifically, this should have just been `index`).

PR #3173, opened 2019-04-02
Put product version with git commit into model.zip/version.txt

### LINKED ISSUE #3132: Version.txt in model .zip should use the FileVersion, not AssemblyVersion
When you save a model, it generates a `Version.txt` file in the .zip, so we can tell which version of ML.NET was used to generate the model.

https://github.com/dotnet/machinelearning/blob/b861b5d64841cbe0f2c866ee7586872aac450a51/src/Microsoft.ML.Core/Data/Repository.cs#L307-L310

However, the issue is we are using the AssemblyVersion there, which isn't great information as all the models produced are showing:

```
1.0.0.0
```

We should write the FileVersion or AssemblyInformationalVersion instead, which will give us the exact build and SHA:

```
[assembly: AssemblyFileVersion("0.12.27526.3")]
[assembly: AssemblyInformationalVersion("0.12.27526.3 @BuiltBy: dlab14-DDVSOWINAGE101 @Branch: master @SrcCode: https://github.com/dotnet/machinelearning/tree/9caafb04d2b4f298f2a9479a9550597afee32c94+9caafb04d2b4f298f2a9479a9550597afee32c94")]
````

Ivanidzo4ka:
Fixes #3132


Ivanidzo4ka:
>Do we want to take this for 1.0?
Well, in 1.0 version file would have 1.0.0.0 version in it, so I don't think it's super critical. For 1.1 or whatever future branch would be ,it would have this change, so, i'm not sure it's worth 1.0.
By I don't mind create PR against 1.0 release.

eerhardt:
If/when we service the `release/1.0` branch with `1.0.1` or `1.0.2`, could we tell the difference? Or even if someone made a .zip file using a local build, instead of our released build.

Ivanidzo4ka:
Sure. I'll create PR towards 1.0 release. But first I think I need approvals for this one. @artidoro  @abgoswam @wschin any one of you can take a look?

PR #3181, opened 2019-04-02
Fix IncludeBuildNumberInPackageVersion for official builds

eerhardt:
When doing an official build and setting a AzDO build variable, the variable turns into an environment variable. When MSBuild props/targets files declare a property, if they don't check if the property is already set, the MSBuild props file will override the environment variable. This causes the AzDO build variable to be ignored.

Adding a check if the IncludeBuildNumberInPackageVersion property is already set before setting it in Directory.Build.props.


eerhardt:
Merging to unblock the build. The CI legs got passed the build part, so this change is good.

PR #3185, opened 2019-04-03
Improve saving and loading of TextFeaturizingEstimator

### LINKED ISSUE #3128: Creating a prediction engine from a model with a TextFeaturizer takes too long
### System information

- **OS version/distro**: Windows
- **.NET Version (eg., dotnet --info)**:  .NET Core 2.1

### Issue

- **What did you do?**
Load the model and call `CreatePredictionEngine` on a pipeline that was created like this:

```C#
            var ml = new MLContext(seed: 1);
            var data = ml.Data.LoadFromTextFile<SentimentData>(GetDataPath(TestDatasets.Sentiment.trainFilename), hasHeader: true);

            // Pipeline.
            var pipeline = ml.Transforms.Text.FeaturizeText("Features", "SentimentText")
                .AppendCacheCheckpoint(ml)
                .Append(ml.BinaryClassification.Trainers.SdcaNonCalibrated(
                    new SdcaNonCalibratedBinaryTrainer.Options { NumberOfThreads = 1 }));

            // Train.
            var model = pipeline.Fit(data);

            var modelPath = GetOutputPath("temp.zip");
            // Save model. 
            ml.Model.Save(model, data.Schema, modelPath);

            // Load model.
            var loadedModel = ml.Model.Load(modelPath, out var inputSchema);

            // Create prediction engine and test predictions.
            var engine = ml.Model.CreatePredictionEngine<SentimentData, SentimentPrediction>(loadedModel, inputSchema);
```

- **What happened?**
It takes an incredible amount of time to `CreatePredictionEngine`. Using the attached model.zip
[SentimentModel.zip](https://github.com/dotnet/machinelearning/files/3015931/SentimentModel.zip) it takes roughly 1.5 seconds for `CreatePredictionEngine` on my machine in a benchmark:

|                 Method |    Mean |    Error |   StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|----------------------- |--------:|---------:|---------:|------------:|------------:|------------:|--------------------:|
| CreatePredictionEngine | 1.516 s | 0.0302 s | 0.0461 s |  14000.0000 |   6000.0000 |           - |           146.35 MB |

- **What did you expect?**
I expected the model to load in a much shorter time with a lot less memory being allocated.

### Source code / logs

Doing a profile of the [Benchmark](https://github.com/glennc/Extensions.ML/blob/840211d64e1e165a124fd17020339479aac79231/benchmarks/micro/CreationBenchmarks.cs#L25-L29), I was able to get this flame graph:

image

As you can see, the two boxes show where we are writing a model and compressing it twice, just during the call to `CreatePredictionEngine`. This is because of the following code:

https://github.com/dotnet/machinelearning/blob/beb67a128544ee98b8c4309859f7e5a9f679ac4e/src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs#L663-L679

Here, `TextFeaturizingEstimator.Transformer` is doing things the "old way" using `ApplyTransformUtils.ApplyAllTransformsToData`, which operates on `IDataTransform`s. This eventually gets into:

https://github.com/dotnet/machinelearning/blob/70ef7ecd43b031b481a4047ea361da5e2f360336/src/Microsoft.ML.Data/Utilities/ApplyTransformUtils.cs#L35-L52

which is serializing and deserializing the transform, which is super expensive.

This happens twice because once inside `GetRowToRowMapper` and once inside `GetOutputSchema`.

We should optimize the text featurizing transform so creating a prediction engine isn't so slow.

@TomFinley @glennc @CESARDELATORRE

yaeldekel:
Fixes #3128.


test/Microsoft.ML.Benchmarks/TextPredictionEngineCreation.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            // Save model. 
+            _context.Model.Save(model, data.Schema, modelPath);
+
+            // Load model.
+            _trainedModel = _context.Model.Load(modelPath, out var inputSchema);
+        }
+
+        [Benchmark]
+        public PredictionEngine<SentimentData, SentimentPrediction> CreatePredictionEngine()
```
yaeldekel: >CreatePredictionEngine [](start = 68, length = 22)

This improved from ~300 milliseconds to ~850 microseconds.

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
```diff
@@ -711,27 +692,33 @@ public Transformer(IHostEnvironment env, ModelLoadContext ctx)
...
             private static VersionInfo GetVersionInfo()
             {
                 return new VersionInfo(
                     modelSignature: "TEXT XFR",
-                    verWrittenCur: 0x00010001, // Initial
-                    verReadableCur: 0x00010001,
+                    //verWrittenCur: 0x00010001, // Initial
+                    verWrittenCur: 0x00010002, // Save as TransformerChain instead of an array of IDataTransform
+                    verReadableCur: 0x00010002,
                     verWeCanReadBack: 0x00010001,
```
yaeldekel: >0x00010001 [](start = 38, length = 10)

I will add a backwards compatibility test tomorrow.

test/Microsoft.ML.Benchmarks/TextPredictionEngineCreation.cs
eerhardt: I tried the original benchmark from the issue - 

https://github.com/glennc/Extensions.ML/blob/840211d64e1e165a124fd17020339479aac79231/benchmarks/micro/CreationBenchmarks.cs#L25-L29

Using the "SentimentModel.zip" that is uploaded on the issue. And the performance got 10x worse than originally reported:

|                 Method |    Mean |    Error |   StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|----------------------- |--------:|---------:|---------:|------------:|------------:|------------:|--------------------:|
| CreatePredictionEngine | 10.92 s | 0.2412 s | 0.5962 s | 101000.0000 |  37000.0000 |   1000.0000 |          1001.46 MB |

Running the same benchmark using the latest `master` branch packages, I am still seeing:

|                 Method |    Mean |    Error |   StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|----------------------- |--------:|---------:|---------:|------------:|------------:|------------:|--------------------:|
| CreatePredictionEngine | 1.526 s | 0.0218 s | 0.0193 s |  14000.0000 |   6000.0000 |           - |           146.34 MB |

Which is exactly what I saw originally.

I did an ETW perf trace, and I am still seeing `ApplyAllTransformsToData` being called, which is serializing and deserializing the model (using compression), and that is taking up a lot of time:

image

eerhardt:
> test/data/backcompat/SentimentModel.zip

we shouldn't check this file into this repo. Instead put it in https://github.com/dotnet/machinelearning-testdata/tree/master/Microsoft.ML.TestModels/TestModels

test/Microsoft.ML.Tests/Transformers/TextFeaturizerTests.cs
```diff
@@ -644,5 +645,18 @@ public void TestLdaCommandLine()
         {
             Assert.Equal(Maml.Main(new[] { @"showschema loader=Text{col=A:R4:0-10} xf=lda{col=B:A} in=f:\2.txt" }), (int)0);
         }
+
+        [Fact]
+        public void TestTextFeaturizerBackCompat()
+        {
+            var modelPath = GetDataPath("backcompat", "SentimentModel.zip");
+            var model = ML.Model.Load(modelPath, out var inputSchema);
```
eerhardt: Do we want to use the model as well? Just to make sure everything is hooked up correctly?

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
```diff
@@ -711,27 +692,54 @@ public Transformer(IHostEnvironment env, ModelLoadContext ctx)
...
+                    for (int i = 0; i < n; i++)
+                    {
+                        var dirName = string.Format(TransformDirTemplate, i);
+                        ITransformer transformer;
+                        // Try to load as an ITransformer.
+                        try
+                        {
+                            ctx.LoadModelOrNull<ITransformer, SignatureLoadModel>(env, out transformer, dirName);
+                        }
+                        catch
```
eerhardt: Are we OK with eating all exceptions here?

eerhardt:
I've taken my original scenario and tried it on the new code. The improvement is phenomenal. 


Method | Mean | Error | StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op
-- | -- | -- | -- | -- | -- | -- | --
CreatePredictionEngine | 935.2 us | 13.82 us | 11.54 us | 83.9844 | 13.6719 | - | 344.89 KB

vs. without these changes:

|                 Method |    Mean |    Error |   StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|----------------------- |--------:|---------:|---------:|------------:|------------:|------------:|--------------------:|
| CreatePredictionEngine | 1.516 s | 0.0302 s | 0.0461 s |  14000.0000 |   6000.0000 |           - |           146.35 MB |

Nice work @yaeldekel!

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
```diff
@@ -416,13 +416,14 @@ public ITransformer Fit(IDataView input)
             string charFeatureCol = null;
             List<string> tempCols = new List<string>();
             IDataView view = input;
+            TransformerChain<ITransformer> chain = new TransformerChain<ITransformer>();
```
artidoro: Why not building an `EstimatorChain` along the way according to the specified options instead of a `TransformerChain`?
You are essentially creating a pipeline of other estimators so it could make sense to act at that level. If you create a pipeline of estimators and call `.Fit(input)` at the end, the `EstimatorChain.Fit()` method runs code that is essentially the same as what you do in `AddToChainAndTransform`. 

Did you choose this approach because some of the transforms don't have an estimator version?

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
```diff
@@ -711,27 +692,54 @@ public Transformer(IHostEnvironment env, ModelLoadContext ctx)
...
             }
 
             private static VersionInfo GetVersionInfo()
             {
                 return new VersionInfo(
                     modelSignature: "TEXT XFR",
-                    verWrittenCur: 0x00010001, // Initial
-                    verReadableCur: 0x00010001,
+                    //verWrittenCur: 0x00010001, // Initial
+                    verWrittenCur: 0x00010002, // Save as TransformerChain instead of an array of IDataTransform
```
TomFinley: >TransformerChain [](start = 58, length = 16)

Wow, nice.

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
```diff
@@ -636,46 +643,37 @@ internal static IDataTransform Create(IHostEnvironment env, Options args, IDataV
...
             }
 
             public IDataView Transform(IDataView input)
             {
                 _host.CheckValue(input, nameof(input));
-                return ApplyTransformUtils.ApplyAllTransformsToData(_host, _xf, input);
+                return _chain.Transform(input);
             }
 
             bool ITransformer.IsRowToRowMapper => true;
```
TomFinley: >true [](start = 50, length = 4)

If we're going to do that, that's fine, but it might be best to have an assert that the chain is a row to row mapper in the constructor then. (Only if you have to make another change for some other reason anyway.)

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
yaeldekel: I tried to follow your suggestion and create an `EstimatorChain` instead of a `TransformerChain`, but I got stuck in the drop-missing-values part (WordBagTransform.cs line 326). It turns out that the `MissingValueDroppingTransformer` doesn't have an estimator.

---
In reply to: [272266922](https://github.com/dotnet/machinelearning/pull/3185#discussion_r272266922) [](ancestors = 272266922)

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
```diff
@@ -711,27 +692,54 @@ public Transformer(IHostEnvironment env, ModelLoadContext ctx)
                 _host.CheckValue(ctx, nameof(ctx));
 
                 ctx.CheckAtModel(GetVersionInfo());
-                int n = ctx.Reader.ReadInt32();
-
-                ctx.LoadModel<ILegacyDataLoader, SignatureLoadDataLoader>(env, out var loader, "Loader", new MultiFileSource(null));
 
-                IDataView data = loader;
-                for (int i = 0; i < n; i++)
+                if (ctx.Header.ModelVerReadable == VerIDataTransform)
```
TomFinley: >VerIDataTransform [](start = 51, length = 17)

Gee, that's a lot of backcompat code.

src/Microsoft.ML.Transforms/Text/TextFeaturizingEstimator.cs
```diff
@@ -636,46 +643,37 @@ internal static IDataTransform Create(IHostEnvironment env, Options args, IDataV
         private sealed class Transformer : ITransformer
```
TomFinley: >Transformer [](start = 29, length = 11)

While we need this class for the backcompat case, would anything bad actually happened if we retained it *only* for back-compat, and just returned chains?

Maybe not -- we might want to expose parts of it later in future API, just an idle thought.

TomFinley:
It is good, but, still, that's almost a millisecond... but, yeah, maybe, I guess, there's a lot of reflection and other stuff going into the prediction engine.

---
In reply to: [479647033](https://github.com/dotnet/machinelearning/pull/3185#issuecomment-479647033) [](ancestors = 479647033)

PR #3210, opened 2019-04-05
Fix marshalling of bool flags in MF

### LINKED ISSUE #3003: MatrixFactorizationTrainer prints log statements to console even if Quiet = true
Even when MatrixFactorizationTrainer option Quiet=true, the iterations print to console.

https://github.com/dotnet/machinelearning/blob/481c37774afdbc6f8b1eab9c9f894c83f3ac984d/src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs#L160

image

wschin:
Fix #3003. The marshalled Boolean size from C# to C should be 1 byte, not 4. Also, the change from LIBMF is for supporting quiet option in new implicit-feedback MF.

### DIFF
--- a/src/Microsoft.ML.Recommender/SafeTrainingAndModelBuffer.cs
+++ b/src/Microsoft.ML.Recommender/SafeTrainingAndModelBuffer.cs
@@ -133,19 +133,19 @@ namespace Microsoft.ML.Recommender.Internal
             /// Specify if the factor matrices should be non-negative.
             /// </summary>
             [FieldOffset(48)]
-            public int DoNmf;
+            public byte DoNmf;
 
             /// <summary>
             /// Set to true so that LIBMF may produce less information to STDOUT.
             /// </summary>
-            [FieldOffset(52)]
-            public int Quiet;
+            [FieldOffset(49)]
+            public byte Quiet;
 
             /// <summary>
             /// Set to false so that LIBMF may reuse and modifiy the data passed in.
             /// </summary>
-            [FieldOffset(56)]
-            public int CopyData;
+            [FieldOffset(50)]
+            public byte CopyData;
         }
 
         [StructLayout(LayoutKind.Explicit)]
@@ -223,9 +223,9 @@ namespace Microsoft.ML.Recommender.Internal
             _mfParam.Eta = (float)eta;
             _mfParam.Alpha = (float)alpha;
             _mfParam.C = (float)c;
-            _mfParam.DoNmf = doNmf ? 1 : 0;
-            _mfParam.Quiet = quiet ? 1 : 0;
-            _mfParam.CopyData = copyData ? 1 : 0;
+            _mfParam.DoNmf = doNmf ? (byte)1 : (byte)0;
+            _mfParam.Quiet = quiet ? (byte)1 : (byte)0;
+            _mfParam.CopyData = copyData ? (byte)1 : (byte)0;
         }
 
         ~SafeTrainingAndModelBuffer()
--- a/src/Microsoft.ML.Recommender/SafeTrainingAndModelBuffer.cs
+++ b/src/Microsoft.ML.Recommender/SafeTrainingAndModelBuffer.cs
@@ -133,19 +133,19 @@ namespace Microsoft.ML.Recommender.Internal
             /// Specify if the factor matrices should be non-negative.
             /// </summary>
             [FieldOffset(48)]
-            public byte DoNmf;
+            public sbyte DoNmf;
 
             /// <summary>
             /// Set to true so that LIBMF may produce less information to STDOUT.
             /// </summary>
             [FieldOffset(49)]
-            public byte Quiet;
+            public sbyte Quiet;
 
             /// <summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.ML.Recommender/SafeTrainingAndModelBuffer.cs
```diff
@@ -133,19 +133,19 @@ private struct MFParameter
             /// Specify if the factor matrices should be non-negative.
             /// </summary>
             [FieldOffset(48)]
-            public int DoNmf;
+            public byte DoNmf;
```
eerhardt: While this is better than what we had before, this is still not correct.

[https://docs.microsoft.com/en-us/cpp/cpp/fundamental-types-cpp?view=vs-2019](https://docs.microsoft.com/en-us/cpp/cpp/fundamental-types-cpp?view=vs-2019)

>  bool Type bool is an integral type that can have one of the two values true or false. Its size is unspecified.

[https://en.cppreference.com/w/cpp/language/types](https://en.cppreference.com/w/cpp/language/types)

> bool - type, capable of holding one of the two values: true or false. The value of sizeof(bool) is implementation defined and might differ from 1.

Instead, to make interop/marshalling work correctly on all computers - the technique we use in .NET is to ensure the ABI (application binary interface) is stable no matter the compiler/processor/etc. This means only stable-sized types can be used in the ABI that C# code P/Invokes into.

So to fix this, the libmf C code should change to have a stable ABI. I see it uses `float`, `int`, and `bool`. For sure `int` and `bool` can change sizes depending on the compiler/processor. Instead, we should be using `int32_t` and `int8_t` on the C side.  (Note: you can just have a wrapper export function in the C library that exposes the stable ABI and calls into your current code. So the main C/C++ code can remain unchanged.)

wschin: Thank you for pointing this out. Another iteration is pushed. Hopefully, it can address this comment.

src/Microsoft.ML.Recommender/SafeTrainingAndModelBuffer.cs
```diff
@@ -133,19 +133,19 @@ private struct MFParameter
             /// Specify if the factor matrices should be non-negative.
             /// </summary>
             [FieldOffset(48)]
-            public int DoNmf;
+            public sbyte DoNmf;
```
eerhardt: Why `sbyte`?  Why not just `byte`?

wschin: It's int8 in C side.

src/Native/MatrixFactorizationNative/UnmanagedMemory.cpp
```diff
@@ -9,25 +9,48 @@
 
 using namespace mf;
 
+EXPORT_API(mf_parameter) make_param(const mf_parameter_bridge *param_bridge)
```
eerhardt: I don't think this needs `EXPORT_API` does it?

src/Native/MatrixFactorizationNative/UnmanagedMemory.cpp
```diff
@@ -9,25 +9,48 @@
...
+    param.nr_bins = param_bridge->nr_bins;
+    param.nr_iters = param_bridge->nr_iters;
+    param.lambda_p1 = param_bridge->lambda_p1;
+    param.lambda_p2 = param_bridge->lambda_p2;
+    param.lambda_q1 = param_bridge->lambda_q1;
+    param.lambda_q2 = param_bridge->lambda_q2;
+    param.eta = param_bridge->eta;
+    param.alpha = param_bridge->alpha;
+    param.c = param_bridge->c;
+    param.do_nmf = static_cast<bool>(param_bridge->do_nmf);
```
eerhardt: Instead of static cast, would it be better to use `param_bridge->do_nmf != 0 ? true : false`?

src/Microsoft.ML.Recommender/SafeTrainingAndModelBuffer.cs
eerhardt: Make it `uint8_t` on the C side then?

Signed-ness doesn't really matter here. So using `byte` is more appropriate.

src/Native/MatrixFactorizationNative/UnmanagedMemory.cpp
wschin: Will do `> 0 ? true : false`. Thanks.

src/Native/MatrixFactorizationNative/UnmanagedMemory.cpp
wschin: Yes!

src/Microsoft.ML.Recommender/SafeTrainingAndModelBuffer.cs
wschin: Ok. I will change other place to match this decision.

wschin:
I will make change to `mf_problem` and `mf_model` in other PRs because they incur more complicated changes.

PR #3227, opened 2019-04-05
Polish marshalling of MF model and MF problem and enable 32-bit tests

### LINKED ISSUE #1441: Matrix factorization should work in x86
Matrix factorization fails in x86 (MatrixFactorizationSimpleTrainAndPredict test was disabled as part of #1432).

The reason is that the MFModel struct is architecture dependent: https://github.com/dotnet/machinelearning/blob/eae76959e6714af44caa212e102a5f06f0110e72/src/Microsoft.ML.Recommender/SafeTrainingAndModelBuffer.cs#L67-L80

The struct matches the output data structure produced by c++ code in the libmf library. 

See the following comment for how this issue should be solved using P/Invoke instead: https://github.com/dotnet/machinelearning/pull/1432#discussion_r229095196

wschin:
Fix #1441. Also related to some suggestion made in #3210.

### DIFF
--- a/src/Microsoft.ML.Recommender/SafeTrainingAndModelBuffer.cs
+++ b/src/Microsoft.ML.Recommender/SafeTrainingAndModelBuffer.cs
@@ -17,31 +17,50 @@ namespace Microsoft.ML.Recommender.Internal
     /// </summary>
     internal sealed class SafeTrainingAndModelBuffer : IDisposable
     {
-        [StructLayout(LayoutKind.Explicit)]
+        [StructLayout(LayoutKind.Sequential)]
         private struct MFNode
         {
-            [FieldOffset(0)]
+            /// <summary>
+            /// Row index.
+            /// </summary>
             public int U;
-            [FieldOffset(4)]
+
+            /// <summary>
+            /// olumn index;
+            /// </summary>
             public int V;
-            [FieldOffset(8)]
+
+            /// <summary>
+            /// Matrix element's value at <see cref="U"/>-th row and <see cref="V"/>-th column.
+            /// </summary>
             public float R;
         }
 
-        [StructLayout(LayoutKind.Explicit)]
+        [StructLayout(LayoutKind.Sequential)]
         private unsafe struct MFProblem
         {
-            [FieldOffset(0)]
+            /// <summary>
+            /// Number of rows.
+            /// </summary>
             public int M;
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.Recommender/SafeTrainingAndModelBuffer.cs
```diff
@@ -17,31 +17,50 @@ namespace Microsoft.ML.Recommender.Internal
...
         {
-            [FieldOffset(0)]
+            /// <summary>
+            /// Row index.
+            /// </summary>
             public int U;
-            [FieldOffset(4)]
+
+            /// <summary>
+            /// olumn index;
```
eerhardt: `olumn`

test/Microsoft.ML.TestFramework/Attributes/MatrixFactorizationFactAttribute.cs
```diff
@@ -17,7 +17,7 @@ public MatrixFactorizationFactAttribute() : base("Disabled - this test is being
         /// <inheritdoc />
         protected override bool IsEnvironmentSupported()
         {
-            return Environment.Is64BitProcess;
+            return true;
```
eerhardt: Can probably just remove this whole thing, I think.

wschin: I'd keep it. We will be enabling LIBMF's SSE/AVX and I guess this might be useful if LIBMF's numerical stability changes.

src/Microsoft.ML.Recommender/SafeTrainingAndModelBuffer.cs
wschin: Fixed. Thanks.

test/Microsoft.ML.TestFramework/Attributes/MatrixFactorizationFactAttribute.cs
eerhardt: Can you at least remove `base("Disabled - this test is being fixed as part of https://github.com/dotnet/machinelearning/issues/1441")`, as that appears to no longer be valid?

wschin: No problem.

PR #3230, opened 2019-04-08
Fix bug in ONNX scorer sample.

codemzs:
ONNX scorer transformer sample throws an exception today because it cannot find the model file. This PR fixes that.

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/OnnxTransform.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/OnnxTransform.cs
@@ -13,7 +13,8 @@ namespace Microsoft.ML.Samples.Dynamic
         public static void Example()
         {
             // Download the squeeznet image model from ONNX model zoo, version 1.2
-            // https://github.com/onnx/models/tree/master/squeezenet
+            // https://github.com/onnx/models/tree/master/squeezenet or use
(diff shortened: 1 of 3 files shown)


docs/samples/Microsoft.ML.Samples/Program.cs
```diff
@@ -1,12 +1,13 @@
 ﻿using Microsoft.ML.Samples.Dynamic;
+using Samples.Dynamic;
 
 namespace Microsoft.ML.Samples
 {
     internal static class Program
     {
         static void Main(string[] args)
         {
-            ReplaceMissingValues.Example();
```
sfilipi: typically we don't check changes to this in. Unecessary.

docs/samples/Microsoft.ML.Samples/Dynamic/OnnxTransform.cs
```diff
@@ -14,7 +14,7 @@ public static void Example()
         {
             // Download the squeeznet image model from ONNX model zoo, version 1.2
```
sfilipi: Should it have a note about referencing the Microsoft.ML.Onnx.TestModels nuget?

codemzs: Done.

PR #3249, opened 2019-04-08
Fix runtime exception in ImageClassification.

### LINKED ISSUE #3248: Fix runtime exception in Image classification sample.
This sample throws an exception for two reasons 1) model not found, should download the mode and unzip 2) even if the model exist it needs to set addBatchDimensionInput to true.

codemzs:
fixes #3248

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/ImageClassification.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/ImageClassification.cs
@@ -1,10 +1,10 @@
 using System;
+using System.IO;
 using System.Linq;
 using System.Net;
-using Microsoft.ML.Data;
-using System.IO;
-using ICSharpCode.SharpZipLib.Tar;
 using ICSharpCode.SharpZipLib.GZip;
+using ICSharpCode.SharpZipLib.Tar;
+using Microsoft.ML.Data;
 
 namespace Microsoft.ML.Samples.Dynamic
 {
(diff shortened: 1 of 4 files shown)


docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/ImageClassification.cs
```diff
@@ -86,5 +97,31 @@ class OutputScores
...
+            using (WebClient client = new WebClient())
+            {
+                client.DownloadFile(new Uri($"{baseGitPath}"), dataFile);
+            }
+
+            return dataFile;
+        }
+
+        /// <summary>
+        /// Taken from https://github.com/icsharpcode/SharpZipLib/wiki/GZip-and-Tar-Samples.
```
zeahmed: >https://github.com/icsharpcode/SharpZipLib/wiki/GZip-and-Tar-Samp [](start = 23, length = 65)

Don't we have anything in .NET for zip/unzip. Are we fine to use external libraries?

The link is for wiki section. I think it should be for main repo, right?

codemzs: Nope. I tried but this is the most talked library.

docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/ImageClassification.cs
```diff
@@ -1,5 +1,9 @@
 using System;
+using System.IO;
 using System.Linq;
+using System.Net;
+using ICSharpCode.SharpZipLib.GZip;
```
wschin: Just make sure its license is good. :)

codemzs: Done.

PR #3250, opened 2019-04-08
Exposing the confusion matrix

### LINKED ISSUE #2009: Bring back support for `ConfusionMatrix`  in the new API
While updating tests from the old AP to the new API, we found that  the new API does not support  `ConfusionMatrix`  in the evaluation output.

https://github.com/dotnet/machinelearning/blob/21d64aa5781750bfaddad262100c54989e8901f8/src/Microsoft.ML.Data/Evaluators/MultiClassClassifierEvaluator.cs#L506-L520

In the Legacy API,  there was support for  `ConfusionMatrix`

https://github.com/dotnet/machinelearning/blob/21d64aa5781750bfaddad262100c54989e8901f8/src/Microsoft.ML.Data/Evaluators/MultiClassClassifierEvaluator.cs#L1031-L1049

We should add back the  checks for `ConfusionMatrix`  in the following tests : 
- TrainAndPredictIrisModelTest
- TrainAndPredictIrisModelWithStringLabelTest

@codemzs  @Ivanidzo4ka

sfilipi:
Fixes #2009 by exposing the confusion matrix.

### DIFF
--- a/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
+++ b/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
@@ -440,8 +440,8 @@ namespace Microsoft.ML.Data
         /// <param name="labelColumn">Label column.</param>
         public static IEnumerable<SchemaShape.Column> AnnotationsForMulticlassScoreColumn(SchemaShape.Column? labelColumn = null)
         {
-            var cols = new List<SchemaShape.Column>();
-            if (labelColumn.HasValue && labelColumn.Value.IsKey)
+           var cols = new List<SchemaShape.Column>();
+            if (labelColumn != null && labelColumn.Value.IsKey)
             {
                 if (labelColumn.Value.Annotations.TryFindColumn(Kinds.KeyValues, out var metaCol) &&
                     metaCol.Kind == SchemaShape.Column.VectorKind.Vector)
--- a/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
+++ b/src/Microsoft.ML.Core/Data/AnnotationUtils.cs
@@ -440,7 +440,7 @@ namespace Microsoft.ML.Data
         /// <param name="labelColumn">Label column.</param>
         public static IEnumerable<SchemaShape.Column> AnnotationsForMulticlassScoreColumn(SchemaShape.Column? labelColumn = null)
         {
-           var cols = new List<SchemaShape.Column>();
+            var cols = new List<SchemaShape.Column>();
             if (labelColumn != null && labelColumn.Value.IsKey)
             {
                 if (labelColumn.Value.Annotations.TryFindColumn(Kinds.KeyValues, out var metaCol) &&
--- a/src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
+++ b/src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
@@ -1359,7 +1359,7 @@ namespace Microsoft.ML.Data
             var confusionMatrix = GetConfusionMatrix(host, confusionDataView, binary, sample, false);
             var confusionTableString = GetConfusionTableAsString(confusionMatrix, false);
 
-            // if there is a Weight column, return the weighted confusionMatrix as well, from this function.
+            // If there is a Weight column, return the weighted confusionMatrix as well, from this function.
             if (isWeighted)
             {
                 confusionMatrix = GetConfusionMatrix(host, confusionDataView, binary, sample, true);
--- a/src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
+++ b/src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
@@ -1382,15 +1382,14 @@ namespace Microsoft.ML.Data
                 host.CheckParam(weightColumn.HasValue, nameof(getWeighted), "There is no Weight column in the confusionMatrix data view.");
 
             // Get the counts names.
-            var countColumn = confusionDataView.Schema.GetColumnOrNull(MetricKinds.ColumnNames.Count);
-            host.Check(countColumn.HasValue, "Did not find the count column");
-            var type = countColumn.Value.Annotations.Schema.GetColumnOrNull(AnnotationUtils.Kinds.SlotNames)?.Type as VectorDataViewType;
-            host.Check(type != null && type.IsKnownSize && type.ItemType is TextDataViewType, "The Count column does not have a text vector metadata of kind SlotNames.");
+            var countColumn = confusionDataView.Schema[MetricKinds.ColumnNames.Count];
+            var type = countColumn.Annotations.Schema.GetColumnOrNull(AnnotationUtils.Kinds.SlotNames)?.Type as VectorDataViewType;
+            host.Assert(type != null && type.IsKnownSize && type.ItemType is TextDataViewType, "The Count column does not have a text vector metadata of kind SlotNames.");
 
             // Get the class names
             var labelNames = default(VBuffer<ReadOnlyMemory<char>>);
-            countColumn.Value.Annotations.GetValue(AnnotationUtils.Kinds.SlotNames, ref labelNames);
-            host.Check(labelNames.IsDense, "Slot names vector must be dense");
+            countColumn.Annotations.GetValue(AnnotationUtils.Kinds.SlotNames, ref labelNames);
+            host.Assert(labelNames.IsDense, "Slot names vector must be dense");
 
             int numConfusionTableLabels = sample < 0 ? labelNames.Length : Math.Min(labelNames.Length, sample);
 
@@ -1418,23 +1417,30 @@ namespace Microsoft.ML.Data
             double[] recallSums;
             double[][] confusionTable;
 
-            if(getWeighted)
+            if (getWeighted)
                 confusionTable = GetConfusionTableAsArray(confusionDataView, weightColumn.Value.Index, labelNames.Length,
                   labelIndexToConfIndexMap, numConfusionTableLabels, out precisionSums, out recallSums);
             else
-                confusionTable = GetConfusionTableAsArray(confusionDataView, countColumn.Value.Index, labelNames.Length,
+                confusionTable = GetConfusionTableAsArray(confusionDataView, countColumn.Index, labelNames.Length,
                 labelIndexToConfIndexMap, numConfusionTableLabels, out precisionSums, out recallSums);
+
+            double[] precision = new double[numConfusionTableLabels];
+            double[] recall = new double[numConfusionTableLabels];
+            for (int i = 0; i < numConfusionTableLabels; i++)
+            {
+                recall[i] = recallSums[i] > 0 ? confusionTable[i][i] / recallSums[i] : 0;
+                precision[i] = precisionSums[i] > 0 ? confusionTable[i][i] / precisionSums[i] : 0;
+            }
 
             var predictedLabelNames = GetPredictedLabelNames(in labelNames, labelIndexToConfIndexMap);
             bool sampled = numConfusionTableLabels < labelNames.Length;
 
-            return new ConfusionMatrix(host, precisionSums, recallSums, confusionTable, predictedLabelNames, sampled, binary);
-
+            return new ConfusionMatrix(host, precision, recall, confusionTable, predictedLabelNames, sampled, binary, countColumn.Annotations);
         }
 
         private static List<ReadOnlyMemory<char>> GetPredictedLabelNames(in VBuffer<ReadOnlyMemory<char>> labelNames, int[] labelIndexToConfIndexMap)
         {
-            List<ReadOnlyMemory<char>> result = new List<ReadOnlyMemory<char>>();
+            List <ReadOnlyMemory<char>> result = new List<ReadOnlyMemory<char>>();
             var values = labelNames.GetValues();
             for (int i = 0; i < values.Length; i++)
             {
@@ -1578,10 +1584,10 @@ namespace Microsoft.ML.Data
         internal static string GetConfusionTableAsString(ConfusionMatrix confusionMatrix, bool isWeighted)
         {
             string prefix = isWeighted ? "Weighted " : "";
-            int numLabels = Utils.Size(confusionMatrix.ConfusionTableCounts);
+            int numLabels = confusionMatrix?.ConfusionTableCounts == null? 0: confusionMatrix.ConfusionTableCounts.Length;
 
             int colWidth = numLabels == 2 ? 8 : 5;
-            int maxNameLen = confusionMatrix.PredictedLabelNames.Max(name => name.Length);
+            int maxNameLen = confusionMatrix.PredictedClassesIndicators.Max(name => name.Length);
             // If the names are too long to fit in the column header, we back off to using class indices
             // in the header. This will also require putting the indices in the row, but it's better than
             // the alternative of having ambiguous abbreviated column headers, or having a table potentially
@@ -1594,7 +1600,7 @@ namespace Microsoft.ML.Data
             {
                 // The row label will also include the index, so a user can easily match against the header.
                 // In such a case, a label like "Foo" would be presented as something like "5. Foo".
-                rowDigitLen = Math.Max(confusionMatrix.PredictedLabelNames.Count - 1, 0).ToString().Length;
+                rowDigitLen = Math.Max(confusionMatrix.PredictedClassesIndicators.Count - 1, 0).ToString().Length;
                 Contracts.Assert(rowDigitLen >= 1);
                 rowLabelLen += rowDigitLen + 2;
             }
@@ -1617,7 +1623,7 @@ namespace Microsoft.ML.Data
             var sb = new StringBuilder();
             if (numLabels == 2 && confusionMatrix.Binary)
             {
-                var positiveCaps = confusionMatrix.PredictedLabelNames[0].ToString().ToUpper();
+                var positiveCaps = confusionMatrix.PredictedClassesIndicators[0].ToString().ToUpper();
 
                 var numTruePos = confusionTable[0][0];
                 var numFalseNeg = confusionTable[0][1];
@@ -1642,7 +1648,7 @@ namespace Microsoft.ML.Data
             sb.AppendFormat("PREDICTED {0}||", pad);
             string format = string.Format(" {{{0},{1}}} |", useNumbersInHeader ? 0 : 1, colWidth);
             for (int i = 0; i < numLabels; i++)
-                sb.AppendFormat(format, i, confusionMatrix.PredictedLabelNames[i]);
+                sb.AppendFormat(format, i, confusionMatrix.PredictedClassesIndicators[i]);
             sb.AppendLine(" Recall");
             sb.AppendFormat("TRUTH     {0}||", pad);
             for (int i = 0; i < numLabels; i++)
@@ -1654,11 +1660,10 @@ namespace Microsoft.ML.Data
                 string.IsNullOrWhiteSpace(prefix) ? "N0" : "F1");
             for (int i = 0; i < numLabels; i++)
             {
-                sb.AppendFormat(rowLabelFormat, i, confusionMatrix.PredictedLabelNames[i]);
+                sb.AppendFormat(rowLabelFormat, i, confusionMatrix.PredictedClassesIndicators[i]);
                 for (int j = 0; j < numLabels; j++)
                     sb.AppendFormat(format2, confusionTable[i][j]);
-                Double recall = confusionMatrix.RecallSums[i] > 0 ? confusionTable[i][i] / confusionMatrix.RecallSums[i] : 0;
-                sb.AppendFormat(" {0,5:F4}", recall);
+                sb.AppendFormat(" {0,5:F4}", confusionMatrix.PerClassRecall[i]);
                 sb.AppendLine();
             }
             sb.AppendFormat("          {0}||", pad);
@@ -1668,10 +1673,8 @@ namespace Microsoft.ML.Data
             sb.AppendFormat("Precision {0}||", pad);
             format = string.Format("{{0,{0}:N4}} |", colWidth + 1);
             for (int i = 0; i < numLabels; i++)
-            {
-                Double precision = confusionMatrix.PrecisionSums[i] > 0 ? confusionTable[i][i] / confusionMatrix.PrecisionSums[i] : 0;
-                sb.AppendFormat(format, precision);
-            }
+                sb.AppendFormat(format, confusionMatrix.PerClassPrecision[i]);
+
             sb.AppendLine();
             return sb.ToString();
         }
@@ -1724,7 +1727,7 @@ namespace Microsoft.ML.Data
             if (metrics.TryGetValue(MetricKinds.Warnings, out warnings))
             {
                 var warningTextColumn = warnings.Schema.GetColumnOrNull(MetricKinds.ColumnNames.WarningText);
-                if (warningTextColumn !=null && warningTextColumn.HasValue && warningTextColumn.Value.Type is TextDataViewType)
+                if (warningTextColumn != null && warningTextColumn.HasValue && warningTextColumn.Value.Type is TextDataViewType)
                 {
                     using (var cursor = warnings.GetRowCursor(warnings.Schema[MetricKinds.ColumnNames.WarningText]))
                     {
--- a/src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
+++ b/src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
@@ -1353,14 +1353,43 @@ namespace Microsoft.ML.Data
             host.CheckValue(confusionDataView, nameof(confusionDataView));
             host.CheckParam(sample == -1 || sample >= 2, nameof(sample), "Should be -1 to indicate no sampling, or at least 2");
 
-            // Get the class names.
-            int countCol;
-            host.Check(confusionDataView.Schema.TryGetColumnIndex(MetricKinds.ColumnNames.Count, out countCol), "Did not find the count column");
-            var type = confusionDataView.Schema[countCol].Annotations.Schema.GetColumnOrNull(AnnotationUtils.Kinds.SlotNames)?.Type as VectorDataViewType;
+            var weightColumn = confusionDataView.Schema.GetColumnOrNull(MetricKinds.ColumnNames.Weight);
+            bool isWeighted = weightColumn.HasValue;
+
+            var confusionMatrix = GetConfusionTableAsType(host, confusionDataView, binary, sample, false);
+            var confusionTableString = GetConfusionTableAsString(confusionMatrix, false);
+
+            // if there is a Weight column, return the weighted confusionMatrix as well, from this function.
+            if (isWeighted)
+            {
+                confusionMatrix = GetConfusionTableAsType(host, confusionDataView, binary, sample, false);
+                weightedConfusionTable = GetConfusionTableAsString(confusionMatrix, true);
+            }
+            else
+                weightedConfusionTable = null;
+
+            return confusionTableString;
+        }
+
+        public static ConfusionMatrix GetConfusionTableAsType(IHost host, IDataView confusionDataView, bool binary = true, int sample = -1, bool getWeighted = false)
+        {
+            host.CheckValue(confusionDataView, nameof(confusionDataView));
+            host.CheckParam(sample == -1 || sample >= 2, nameof(sample), "Should be -1 to indicate no sampling, or at least 2");
+
+            // check that there is a Weight column, if isWeighted parameter is set to true.
+            var weightColumn = confusionDataView.Schema.GetColumnOrNull(MetricKinds.ColumnNames.Weight);
+            if (getWeighted)
+                host.CheckParam(weightColumn.HasValue, nameof(getWeighted), "There is no Weight column in the confusionMatrix data view.");
+
+            // Get the counts names.
+            var countColumn = confusionDataView.Schema.GetColumnOrNull(MetricKinds.ColumnNames.Count);
+            host.Check(countColumn.HasValue, "Did not find the count column");
+            var type = countColumn.Value.Annotations.Schema.GetColumnOrNull(AnnotationUtils.Kinds.SlotNames)?.Type as VectorDataViewType;
             host.Check(type != null && type.IsKnownSize && type.ItemType is TextDataViewType, "The Count column does not have a text vector metadata of kind SlotNames.");
 
+            // Get the class names
             var labelNames = default(VBuffer<ReadOnlyMemory<char>>);
-            confusionDataView.Schema[countCol].Annotations.GetValue(AnnotationUtils.Kinds.SlotNames, ref labelNames);
+            countColumn.Value.Annotations.GetValue(AnnotationUtils.Kinds.SlotNames, ref labelNames);
             host.Check(labelNames.IsDense, "Slot names vector must be dense");
 
             int numConfusionTableLabels = sample < 0 ? labelNames.Length : Math.Min(labelNames.Length, sample);
@@ -1387,27 +1416,20 @@ namespace Microsoft.ML.Data
 
             double[] precisionSums;
             double[] recallSums;
-            var confusionTable = GetConfusionTableAsArray(confusionDataView, countCol, labelNames.Length,
+            double[][] confusionTable;
+
+            if(getWeighted)
+                confusionTable = GetConfusionTableAsArray(confusionDataView, weightColumn.Value.Index, labelNames.Length,
+                  labelIndexToConfIndexMap, numConfusionTableLabels, out precisionSums, out recallSums);
+            else
+                confusionTable = GetConfusionTableAsArray(confusionDataView, countColumn.Value.Index, labelNames.Length,
                 labelIndexToConfIndexMap, numConfusionTableLabels, out precisionSums, out recallSums);
 
             var predictedLabelNames = GetPredictedLabelNames(in labelNames, labelIndexToConfIndexMap);
-            var confusionTableString = GetConfusionTableAsString(confusionTable, recallSums, precisionSums,
-               predictedLabelNames,
-               sampled: numConfusionTableLabels < labelNames.Length, binary: binary);
+            bool sampled = numConfusionTableLabels < labelNames.Length;
 
-            int weightIndex;
-            if (confusionDataView.Schema.TryGetColumnIndex(MetricKinds.ColumnNames.Weight, out weightIndex))
-            {
-                confusionTable = GetConfusionTableAsArray(confusionDataView, weightIndex, labelNames.Length,
-                   labelIndexToConfIndexMap, numConfusionTableLabels, out precisionSums, out recallSums);
-                weightedConfusionTable = GetConfusionTableAsString(confusionTable, recallSums, precisionSums,
-                    predictedLabelNames,
-                    sampled: numConfusionTableLabels < labelNames.Length, prefix: "Weighted ", binary: binary);
-            }
-            else
-                weightedConfusionTable = null;
+            return new ConfusionMatrix(host, precisionSums, recallSums, confusionTable, predictedLabelNames, sampled, binary);
 
-            return confusionTableString;
         }
 
         private static List<ReadOnlyMemory<char>> GetPredictedLabelNames(in VBuffer<ReadOnlyMemory<char>> labelNames, int[] labelIndexToConfIndexMap)
@@ -1553,13 +1575,13 @@ namespace Microsoft.ML.Data
         }
 
         // Get a string representation of a confusion table.
-        private static string GetConfusionTableAsString(double[][] confusionTable, double[] rowSums, double[] columnSums,
-            List<ReadOnlyMemory<char>> predictedLabelNames, string prefix = "", bool sampled = false, bool binary = true)
+        internal static string GetConfusionTableAsString(ConfusionMatrix confusionMatrix, bool isWeighted)
         {
-            int numLabels = Utils.Size(confusionTable);
+            string prefix = isWeighted ? "Weighted " : "";
+            int numLabels = Utils.Size(confusionMatrix.ConfusionTableCounts);
 
             int colWidth = numLabels == 2 ? 8 : 5;
-            int maxNameLen = predictedLabelNames.Max(name => name.Length);
+            int maxNameLen = confusionMatrix.PredictedLabelNames.Max(name => name.Length);
             // If the names are too long to fit in the column header, we back off to using class indices
             // in the header. This will also require putting the indices in the row, but it's better than
             // the alternative of having ambiguous abbreviated column headers, or having a table potentially
@@ -1572,7 +1594,7 @@ namespace Microsoft.ML.Data
             {
                 // The row label will also include the index, so a user can easily match against the header.
                 // In such a case, a label like "Foo" would be presented as something like "5. Foo".
-                rowDigitLen = Math.Max(predictedLabelNames.Count - 1, 0).ToString().Length;
+                rowDigitLen = Math.Max(confusionMatrix.PredictedLabelNames.Count - 1, 0).ToString().Length;
                 Contracts.Assert(rowDigitLen >= 1);
                 rowLabelLen += rowDigitLen + 2;
             }
@@ -1591,10 +1613,11 @@ namespace Microsoft.ML.Data
             else
                 rowLabelFormat = string.Format("{{1,{0}}} ||", paddingLen);
 
+            var confusionTable = confusionMatrix.ConfusionTableCounts;
             var sb = new StringBuilder();
-            if (numLabels == 2 && binary)
+            if (numLabels == 2 && confusionMatrix.Binary)
             {
-                var positiveCaps = predictedLabelNames[0].ToString().ToUpper();
+                var positiveCaps = confusionMatrix.PredictedLabelNames[0].ToString().ToUpper();
 
                 var numTruePos = confusionTable[0][0];
                 var numFalseNeg = confusionTable[0][1];
@@ -1607,7 +1630,7 @@ namespace Microsoft.ML.Data
 
             sb.AppendLine();
             sb.AppendFormat("{0}Confusion table", prefix);
-            if (sampled)
+            if (confusionMatrix.Sampled)
                 sb.AppendLine(" (sampled)");
             else
                 sb.AppendLine();
@@ -1619,7 +1642,7 @@ namespace Microsoft.ML.Data
             sb.AppendFormat("PREDICTED {0}||", pad);
             string format = string.Format(" {{{0},{1}}} |", useNumbersInHeader ? 0 : 1, colWidth);
             for (int i = 0; i < numLabels; i++)
-                sb.AppendFormat(format, i, predictedLabelNames[i]);
+                sb.AppendFormat(format, i, confusionMatrix.PredictedLabelNames[i]);
             sb.AppendLine(" Recall");
             sb.AppendFormat("TRUTH     {0}||", pad);
             for (int i = 0; i < numLabels; i++)
@@ -1631,10 +1654,10 @@ namespace Microsoft.ML.Data
                 string.IsNullOrWhiteSpace(prefix) ? "N0" : "F1");
             for (int i = 0; i < numLabels; i++)
             {
-                sb.AppendFormat(rowLabelFormat, i, predictedLabelNames[i]);
+                sb.AppendFormat(rowLabelFormat, i, confusionMatrix.PredictedLabelNames[i]);
                 for (int j = 0; j < numLabels; j++)
                     sb.AppendFormat(format2, confusionTable[i][j]);
-                Double recall = rowSums[i] > 0 ? confusionTable[i][i] / rowSums[i] : 0;
+                Double recall = confusionMatrix.RecallSums[i] > 0 ? confusionTable[i][i] / confusionMatrix.RecallSums[i] : 0;
                 sb.AppendFormat(" {0,5:F4}", recall);
                 sb.AppendLine();
             }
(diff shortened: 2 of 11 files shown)


sfilipi:
I am sending this as WIP, because i wanted to get early feedback on how much to expose, and where to place it. 

I am not adding support to build the weighted confusion matrix, or adding capabilities to sample a subset of classes for it thinking that it might be enough for V1. The infrastructure is in there, let me know if it is desired.

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Collections.Generic;
+using Microsoft.ML.Runtime;
+
+namespace Microsoft.ML.Data
+{
+    public sealed class ConfusionMatrix
+    {
+        public double[] PrecisionSums { get; }
+        public double[] RecallSums { get; }
+        public ReadOnlyMemory<char>[] Labels => PredictedLabelNames.ToArray(); //sefilipi: return as string[]
```
TomFinley: >ReadOnlyMemory [](start = 15, length = 14)

Remember, as before, that we provide access to the columns, but do not copy out the annotations data. Remember the discussion around having names.

Remember also, as we've found in #3101 that the labels may not even be text, or even exist.

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,49 @@
...
+            RecallSums = recallSums;
+            ConfusionTableCounts = confusionTableCounts;
+            PredictedLabelNames = labelNames;
+            Sampled = sampled;
+            Binary = binary;
+        }
+
+        public override string ToString() => MetricWriter.GetConfusionTableAsString(this, false);
+
+        public double GetCountForClassPair(string predictedLabel, string actualLabel)
```
TomFinley: >string predictedLabel, string actualLabel [](start = 43, length = 41)

As with all the remainder, we index slots by numeric index. They may *optionally* have a string label (or not), but that is not a requirement, and what the string label is (again, if it even exists) cannot be used as any sort of primary key.

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            PrecisionSums = precisionSums;
+            RecallSums = recallSums;
+            ConfusionTableCounts = confusionTableCounts;
+            PredictedLabelNames = labelNames;
+            Sampled = sampled;
+            Binary = binary;
+        }
+
+        public override string ToString() => MetricWriter.GetConfusionTableAsString(this, false);
```
TomFinley: >MetricWriter.GetConfusionTableAsString [](start = 45, length = 38)

This is dangerous. Do we consider `ToString` overrides breaking changes?

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.ML.Runtime;
+
+namespace Microsoft.ML.Data
+{
+    public sealed class ConfusionMatrix
+    {
+        public double[] PrecisionSums { get; }
```
TomFinley: >double[] [](start = 15, length = 8)

Arrays are mutable -- never ever make them part of the public surface of an API that's intended to be a mutable structure.

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
TomFinley: >ToArray(); [](start = 68, length = 10)

THis should not exist anyway, but note that we generally there is an expectation in C# APIs that property accesses are "lightweight" in terms of their memory footprint.

src/Microsoft.ML.Data/Evaluators/Metrics/BinaryClassificationMetrics.cs
```diff
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Linq;
 using Microsoft.ML.Runtime;
```
TomFinley: If I look through this file, it's unclear to me which of the changes require this.

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
```diff
@@ -1353,14 +1353,43 @@ public static string GetConfusionTable(IHost host, IDataView confusionDataView,
...
+        public static ConfusionMatrix GetConfusionTableAsType(IHost host, IDataView confusionDataView, bool binary = true, int sample = -1, bool getWeighted = false)
+        {
+            host.CheckValue(confusionDataView, nameof(confusionDataView));
+            host.CheckParam(sample == -1 || sample >= 2, nameof(sample), "Should be -1 to indicate no sampling, or at least 2");
+
+            // check that there is a Weight column, if isWeighted parameter is set to true.
+            var weightColumn = confusionDataView.Schema.GetColumnOrNull(MetricKinds.ColumnNames.Weight);
+            if (getWeighted)
+                host.CheckParam(weightColumn.HasValue, nameof(getWeighted), "There is no Weight column in the confusionMatrix data view.");
+
```
TomFinley: Sorry, is this something we anticipate might actually happen? We control the creation of this matrix, as far as I am aware, and this is a purely internal utility.

There are two possible answer to your questions, "no" in which case this should have been an assert (with the message as a *comment*, not a string!), or "yes" in which case this is something the user will potentially have to deal with, in which case having something that user might have a prayer of understanding would be required. (I do not view this message as actionable or even diagnostic in any way from a user's perspective.)

Same for other check below...

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,49 @@
...
+            PredictedLabelNames = labelNames;
+            Sampled = sampled;
+            Binary = binary;
+        }
+
+        public override string ToString() => MetricWriter.GetConfusionTableAsString(this, false);
+
+        public double GetCountForClassPair(string predictedLabel, string actualLabel)
+        {
+            int predictedLabelIndex = PredictedLabelNames.IndexOf(predictedLabel.AsMemory());
```
TomFinley: >predictedLabel.AsMemory() [](start = 66, length = 25)

This API is not what we want anyway (since it relies on strings to index when there's no requirement that the strings *even exist* much less be unique), but in point of interest, please note that this code absolutely will not work.

```csharp
var m = "hello my friend".AsMemory();
var tokens = new List<ReadOnlyMemory<char>>();
tokens.Add(m.Slice(0, 5));
tokens.Add(m.Slice(6, 2));
tokens.Add(m.Slice(9, 6));

foreach (var token in tokens)
    Console.WriteLine($"'{token}'");

Console.WriteLine(tokens.IndexOf("friend".AsMemory()));
Console.WriteLine(tokens.IndexOf(tokens[2]));
```

If compiled and run on .NET core 2.1, you get this output.

```
'hello'
'my'
'friend'
-1
2
```

So, relying on `ReadOnlyMemory<char>`'s equality w.r.t. other similar items just is not a safe thing to do at all. You'll also recall all the bugs and barriers we had to fix back when we did the initial switch from our own `DvText` to `ReadOnlyMemory<char>`.

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
sfilipi: You're right, it is a breaking change. 
I'll create a separate API for it. 


"Changing the returned values for a property, field, return or 'out' value, such as the value returned from ToString"
on 
[https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md)



---
In reply to: [273615670](https://github.com/dotnet/machinelearning/pull/3250#discussion_r273615670) [](ancestors = 273615670)

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
sfilipi: I looked a bit on how is the ConfusionMatrix idv build, and it looks like an indicator for the predictedLabels will always be there, and it will be of the ReadOnlyMemory<char> type. 

[https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Evaluators/MulticlassClassificationEvaluator.cs#L222](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Evaluators/MulticlassClassificationEvaluator.cs#L222)
[https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs#L361](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs#L361)

The metrics aggregators have the class names in the ctor arguments, and if the class names are missing, those get generated from the size of the Score column. 

[https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Evaluators/MulticlassClassificationEvaluator.cs#L114](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Evaluators/MulticlassClassificationEvaluator.cs#L114)
[https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs#L184](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs#L184)

So given that, i left the labels as  ReadOnlyMemory<char> instead of providing the annotations of the Count column, and just renamed it to PredictedClassIndicators, since they are not going to be the class names all the time. 

Let me know if I am missing something else. 

---
In reply to: [273615151](https://github.com/dotnet/machinelearning/pull/3250#discussion_r273615151) [](ancestors = 273615151)

TomFinley: > I looked a bit on how is the ConfusionMatrix idv build, and it looks like an indicator for the predictedLabels will always be there, and it will be of the `ReadOnlyMemory<char>` type.

Yes, but AFAIK that is mostly because it existed in order to provide a string to write to console, for the console application, back when this was purely a command line tool and not an API. So it has to have *some* string description, that doesn't mean that it's a good string description. I think in fact it'll just render the number as a string in case there is none. So... *not* something we want to expose as part of the API!

src/Microsoft.ML.Core/Data/AnnotationUtils.cs
```diff
@@ -440,7 +440,7 @@ public static IEnumerable<SchemaShape.Column> GetTrainerOutputAnnotation(bool is
         /// <param name="labelColumn">Label column.</param>
         public static IEnumerable<SchemaShape.Column> AnnotationsForMulticlassScoreColumn(SchemaShape.Column? labelColumn = null)
         {
-            var cols = new List<SchemaShape.Column>();
+           var cols = new List<SchemaShape.Column>();
```
TomFinley: As far as I see the only change in this file is this incorrect indentation. Perhaps we want to fix this?

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,79 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using Microsoft.ML.Runtime;
```
shauheen: Please add the header lines

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,79 @@
...
+{
+    /// <summary>
+    /// Represents the <a href="https://en.wikipedia.org/wiki/Confusion_matrix">confusion matrix</a> of the classification results.
+    /// </summary>
+    public sealed class ConfusionMatrix
+    {
+        /// <summary>
+        /// The calculated value of <a href="https://en.wikipedia.org/wiki/Precision_and_recall#Precision">precision</a> for each class.
+        /// </summary>
+        public ImmutableArray<double> PerClassPrecision { get; }
```
TomFinley: >ImmutableArray [](start = 15, length = 14)

That's fine, note that elsewhere in similar situations we've been using `IReadOnlyList`, but I guess this works too?

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <returns></returns>
+        public string GetFormattedConfusionTable() => MetricWriter.GetConfusionTableAsString(this, false);
+
+        /// <summary>
+        /// Gets the confusion table count for the pair <paramref name="predictedClassIndicatorIndex"/>/<paramref name="actualClassIndicatorIndex"/>.
+        /// </summary>
+        /// <param name="predictedClassIndicatorIndex">The index of the predicted label indicator, in the <see cref="PredictedClassesIndicators"/>.</param>
+        /// <param name="actualClassIndicatorIndex">The index of the actual label indicator, in the <see cref="PredictedClassesIndicators"/>.</param>
+        /// <returns></returns>
+        public double GetCountForClassPair(uint predictedClassIndicatorIndex, uint actualClassIndicatorIndex)
```
TomFinley: >actualClassIndicatorIndex [](start = 83, length = 25)

Indices are usually `int`, as far as I understand. I understand that as far as they are key types they are represented as `int`, but when it comes down to, say, doing slot indexing or whatever, they're `int`s. What you think?

TomFinley: Also, might be nice if there was a special property to indicate the number of classes, maybe. I guess they can get it from the above lists, but, might be nice otherwise.

---
In reply to: [275005689](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275005689) [](ancestors = 275005689)

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// It might be the classes names, or just indices of the predicted classes, if the name mapping is missing.
+        /// </summary>
+        public IReadOnlyList<ReadOnlyMemory<char>> PredictedClassesIndicators;
+
+        internal readonly bool Sampled;
+        internal readonly bool Binary;
+
+        private readonly IHost _host;
+
+        internal ConfusionMatrix(IHost host, double[] precision, double[] recall, double[][] confusionTableCounts,
```
TomFinley: >precision [](start = 54, length = 9)

Might be nice to have some asserts here to make poor readers of the code (like me!) aware of the intent here -- I assume `precision` and `recall` have the same length, that `confusionTableCounts` is a jagged array with that same length along both dimensions, but these are just guesses based on my understanding of the setting, which many people coming to the code fresh would not have. And even I'm not sure.

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// It might be the classes names, or just indices of the predicted classes, if the name mapping is missing.
+        /// </summary>
+        public int NumberOfPredictedClasses { get; }
+
+        /// <summary>
+        /// The <see cref="DataViewSchema.Annotations"/> associated with the Confusion Matrix counts.
+        /// It contains information about the predicted classes, if that information is available.
+        /// It can be the classes names, or their indices.
+        /// </summary>
+        public DataViewSchema.Annotations ClassIndicators { get; }
```
sfilipi: > public DataViewSchema.Annotations ClassIndicators { get; } [](start = 7, length = 59)

@tfinley@gmail.com is exposing this desirable?

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        /// <summary>
+        /// The indicators of the predicted classes.
+        /// It might be the classes names, or just indices of the predicted classes, if the name mapping is missing.
+        /// </summary>
+        internal IReadOnlyList<ReadOnlyMemory<char>> PredictedClassesIndicators;
```
sfilipi: >PredictedClassesIndicators [](start = 53, length = 26)

if the above is ok, i can remove this, and just pass in the annotations everywhere.

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
```diff
@@ -1353,15 +1353,43 @@ public static string GetConfusionTable(IHost host, IDataView confusionDataView,
...
+            var weightColumn = confusionDataView.Schema.GetColumnOrNull(MetricKinds.ColumnNames.Weight);
+            bool isWeighted = weightColumn.HasValue;
 
+            var confusionMatrix = GetConfusionTableAsType(host, confusionDataView, binary, sample, false);
+            var confusionTableString = GetConfusionTableAsString(confusionMatrix, false);
+
+            // if there is a Weight column, return the weighted confusionMatrix as well, from this function.
+            if (isWeighted)
+            {
+                confusionMatrix = GetConfusionTableAsType(host, confusionDataView, binary, sample, false);
```
Ivanidzo4ka: >confusionMatrix = GetConfusionTableAsType(host, confusionDataView, binary, sample, false); [](start = 16, length = 90)

why we doing this again?
We already did that in line 1359

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
```diff
@@ -1353,15 +1353,43 @@ public static string GetConfusionTable(IHost host, IDataView confusionDataView,
...
+                weightedConfusionTable = GetConfusionTableAsString(confusionMatrix, true);
+            }
+            else
+                weightedConfusionTable = null;
+
+            return confusionTableString;
+        }
+
+        public static ConfusionMatrix GetConfusionTableAsType(IHost host, IDataView confusionDataView, bool binary = true, int sample = -1, bool getWeighted = false)
```
Ivanidzo4ka: >GetConfusionTableAsType [](start = 38, length = 23)

Maybe it make sense to rename this method to `GetConfusionMatrix` and `GetConfusionTable` to `GetConfusionMatrixAsString` or `GetConfusionString`?

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
Ivanidzo4ka: it should be true instead of false.

---
In reply to: [275079866](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275079866) [](ancestors = 275079866)

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
```diff
@@ -1553,13 +1581,13 @@ private static string GetFoldMetricsAsString(IHostEnvironment env, IDataView dat
         }
 
         // Get a string representation of a confusion table.
-        private static string GetConfusionTableAsString(double[][] confusionTable, double[] rowSums, double[] columnSums,
-            List<ReadOnlyMemory<char>> predictedLabelNames, string prefix = "", bool sampled = false, bool binary = true)
+        internal static string GetConfusionTableAsString(ConfusionMatrix confusionMatrix, bool isWeighted)
```
Ivanidzo4ka: >GetConfusionTableAsString [](start = 31, length = 25)

Mix of `ConfusionMatrix` and `ConfusionTable` makes me 
...
[http://i.imgur.com/lgSDkvC.gif](http://i.imgur.com/lgSDkvC.gif)
...
Confuse!

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
```diff
@@ -1591,10 +1619,11 @@ private static string GetFoldMetricsAsString(IHostEnvironment env, IDataView dat
             else
                 rowLabelFormat = string.Format("{{1,{0}}} ||", paddingLen);
 
+            var confusionTable = confusionMatrix.ConfusionTableCounts;
```
Ivanidzo4ka: >var confusionTable = confusionMatrix.ConfusionTableCounts; [](start = 12, length = 58)

SO MUCH CONFUSION!
no, seriously, how confusionTable can be  ConfusionTableCounts in  ConfusionMatrix?

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
```diff
@@ -1591,10 +1619,11 @@ private static string GetFoldMetricsAsString(IHostEnvironment env, IDataView dat
...
             var sb = new StringBuilder();
-            if (numLabels == 2 && binary)
+            if (numLabels == 2 && confusionMatrix.Binary)
```
Ivanidzo4ka: `IsBinary`?

src/Microsoft.ML.Data/Evaluators/Metrics/BinaryClassificationMetrics.cs
```diff
@@ -74,6 +74,8 @@ public class BinaryClassificationMetrics
         /// </remarks>
         public double AreaUnderPrecisionRecallCurve { get; }
 
+        public ConfusionMatrix ConfusionMatrix { get; }
```
Ivanidzo4ka: >ConfusionMatrix [](start = 31, length = 15)

need documentation

src/Microsoft.ML.Data/Evaluators/Metrics/MulticlassClassificationMetrics.cs
```diff
@@ -82,9 +82,12 @@ public sealed class MulticlassClassificationMetrics
         /// </remarks>
         public IReadOnlyList<double> PerClassLogLoss { get; }
 
-        internal MulticlassClassificationMetrics(IExceptionContext ectx, DataViewRow overallResult, int topKPredictionCount)
+        // The confusion matrix.
+        public ConfusionMatrix ConfusionMatrix { get; }
```
Ivanidzo4ka: >ConfusionMatrix [](start = 31, length = 15)

need documentation

src/Microsoft.ML.Data/Evaluators/Metrics/MulticlassClassificationMetrics.cs
```diff
@@ -93,8 +96,9 @@ internal MulticlassClassificationMetrics(IExceptionContext ectx, DataViewRow ove
             if (topKPredictionCount > 0)
                 TopKAccuracy = FetchDouble(MulticlassClassificationEvaluator.TopKAccuracy);
 
-            var perClassLogLoss = RowCursorUtils.Fetch<VBuffer<double>>(ectx, overallResult, MulticlassClassificationEvaluator.PerClassLogLoss);
+            var perClassLogLoss = RowCursorUtils.Fetch<VBuffer<double>>(host, overallResult, MulticlassClassificationEvaluator.PerClassLogLoss);
             PerClassLogLoss = perClassLogLoss.DenseValues().ToImmutableArray();
+            ConfusionMatrix = MetricWriter.GetConfusionTableAsType(host, confusionMatrix, binary: false, perClassLogLoss.Length, getWeighted:false);
```
Ivanidzo4ka: >false [](start = 141, length = 5)

it's false by default, you can probably omit it.

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
Ivanidzo4ka: I would also prefer to expose IReadOnlyList just for the sake of consistency. (I don't think we use ImmutableArray in public interface in metrics at all).

---
In reply to: [275005362](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275005362) [](ancestors = 275005362)

test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs
```diff
@@ -67,5 +67,68 @@ void PredictAndMetadata()
...
+                    new SdcaMaximumEntropyMulticlassTrainer.Options { MaximumNumberOfIterations = 100, Shuffle = true, NumberOfThreads = 1, }));
+
+            var model = pipeline.Fit(data);
+
+            // Evaluate the model.
+            var scoredData = model.Transform(data);
+            var metrics = mlContext.MulticlassClassification.Evaluate(scoredData);
+
+            // Check that the SlotNames column is there. 
+            Assert.NotNull(scoredData.Schema["Score"].Annotations.Schema.GetColumnOrNull(AnnotationUtils.Kinds.SlotNames));
```
Ivanidzo4ka: >SlotNames [](start = 111, length = 9)

SlotNames would be present only if your label class originally was string. We have TrainingLabelValues which would be available always!

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
sfilipi: thanks, i move the Weight inside the method, to avoid doing this twice, than forgot to update here. 

---
In reply to: [275080596](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275080596) [](ancestors = 275080596,275079866)

test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs
sfilipi: it will always be there for the Count column of the confusion matrix. This is what the test is testing, that even when the SlotNames is missing in the Score column, the Count column of the confusion matrix still has the names:


see my comment here: [https://github.com/dotnet/machinelearning/pull/3250#discussion_r274744013](https://github.com/dotnet/machinelearning/pull/3250#discussion_r274744013)

---
In reply to: [275083136](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275083136) [](ancestors = 275083136)

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
sfilipi: GetConfusionMatrixAsString exists. So   I used GetConfusionTableAsFormattedString. 

and I hate it already. lmk if anything better comes to mind. 

---
In reply to: [275080299](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275080299) [](ancestors = 275080299)

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
sfilipi: there is also confusionDataView

---
In reply to: [275081264](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275081264) [](ancestors = 275081264)

src/Microsoft.ML.Data/Evaluators/Metrics/MulticlassClassificationMetrics.cs
```diff
@@ -82,9 +82,15 @@ public sealed class MulticlassClassificationMetrics
         /// </remarks>
         public IReadOnlyList<double> PerClassLogLoss { get; }
 
-        internal MulticlassClassificationMetrics(IExceptionContext ectx, DataViewRow overallResult, int topKPredictionCount)
+        /// <summary>
+        /// The <a href="https://en.wikipedia.org/wiki/Confusion_matrix">confusion matrix</a> giving the counts of the
```
Ivanidzo4ka: >confusion matrix [](start = 73, length = 16)

in BinaryClassificationMetrics you use:`2 dimentional confusion matrix`
is this one k-dimensional?

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
Ivanidzo4ka: Can you at least rename this one to `GetConfusionMatrix` ?

---
In reply to: [275088372](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275088372) [](ancestors = 275088372,275080299)

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
```diff
@@ -1348,20 +1348,48 @@ internal static class MetricWriter
...
+
+        public static ConfusionMatrix GetConfusionTable(IHost host, IDataView confusionDataView, bool binary = true, int sample = -1, bool getWeighted = false)
+        {
+            host.CheckValue(confusionDataView, nameof(confusionDataView));
+            host.CheckParam(sample == -1 || sample >= 2, nameof(sample), "Should be -1 to indicate no sampling, or at least 2");
+
+            // check that there is a Weight column, if isWeighted parameter is set to true.
+            var weightColumn = confusionDataView.Schema.GetColumnOrNull(MetricKinds.ColumnNames.Weight);
+            if (getWeighted)
+                host.CheckParam(weightColumn.HasValue, nameof(getWeighted), "There is no Weight column in the confusionMatrix data view.");
```
Ivanidzo4ka: >confusionMatrix data view [](start = 110, length = 25)

`nameof(confusionDataView)` ?

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
```diff
@@ -1348,20 +1348,48 @@ internal static class MetricWriter
...
+
+            // Get the counts names.
+            var countColumn = confusionDataView.Schema[MetricKinds.ColumnNames.Count];
+            var type = countColumn.Annotations.Schema.GetColumnOrNull(AnnotationUtils.Kinds.SlotNames)?.Type as VectorDataViewType;
+            host.Assert(type != null && type.IsKnownSize && type.ItemType is TextDataViewType, "The Count column does not have a text vector metadata of kind SlotNames.");
```
Ivanidzo4ka: >Count [](start = 100, length = 5)

{MetricKinds.ColumnNames.Count}

Ivanidzo4ka: >SlotNames [](start = 158, length = 9)

AnnotationUtils.Kinds.SlotNames

test/Microsoft.ML.Functional.Tests/Common.cs
```diff
@@ -219,6 +227,38 @@ public static void AssertMetrics(MulticlassClassificationMetrics metrics)
...
+            Assert.NotNull(confusionMatrix.ClassIndicators);
+
+            foreach (var precision in confusionMatrix.PerClassPrecision)
+                Assert.InRange(precision, 0, 1);
+
+            foreach (var recall in confusionMatrix.PerClassRecall)
+                Assert.InRange(recall, 0, 1);
+
+
+            // Get the values in the annotations
```
Ivanidzo4ka: DOUBLE LINES! DOUBLE SHOCK!

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        internal readonly bool IsSampled;
+        internal readonly bool IsBinary;
+
+        private readonly IHost _host;
+
+        /// <summary>
+        /// The confusion matrix as a structured type, built from the counts of the confusion table idv.
```
Ivanidzo4ka: >idv [](start = 100, length = 3)

We don't have idv here. Or what is `idv`?

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// </summary>
+        /// <param name="host">The IHost instance. </param>
+        /// <param name="precision">The values of precision per class.</param>
+        /// <param name="recall">The vales of recall per class.</param>
+        /// <param name="confusionTableCounts">The counts of the confusion table. The actual classes values are in the rows of the 2D array,
+        /// and the counts of the predicted classes are in the columns.</param>
+        /// <param name="labelNames">The predicted classes names, or the indexes of the classes, if the names are missing.</param>
+        /// <param name="isSampled">Whether the classes are sampled.</param>
+        /// <param name="isBinary">Whether the confusion table is the result of a binary classification. </param>
+        /// <param name="classIndicators">The Annotations of the Count column, in the confusionTable idv.</param>
```
Ivanidzo4ka: >confusionTable idv [](start = 86, length = 18)

`<paramref name="confusionTableCounts"/>`?

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            Counts = counts.ToImmutableArray();
+            ClassIndicators = classIndicators;
+        }
+
+        /// <summary>
+        /// Returns a human readable representation of the confusion table.
+        /// </summary>
+        /// <returns></returns>
+        public string GetFormattedConfusionTable() => MetricWriter.GetConfusionTableAsString(this, false);
```
Ivanidzo4ka: >MetricWriter.GetConfusionTableAsString [](start = 54, length = 38)

Can we cache it?
Or just calculate it constructor?
I'm not sure in which case this would produce you different results in different times of call.

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
TomFinley: So @sfilipi I notice this was marked as resolved, if you disagree with me and @Ivanidzo4ka that's fine, but would you mind sharing justification? It is somewhat inconsistent?

---
In reply to: [275082837](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275082837) [](ancestors = 275082837,275005362)

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
TomFinley: I'd prefer to not? AFAIK you get the class indicators out of the column they pass to the evaluate method. So why store it here? Why not just say, "get the indicators?" What does this API buy you? So my initial reaction is, "no, why is this object so complicated?"

---
In reply to: [275067993](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275067993) [](ancestors = 275067993)

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
TomFinley: >, "The Count column does not have a text vector metadata of kind SlotNames." [](start = 93, length = 76)

Assert mesages are pointless so I'd prefer to not have them, and instead have a comment, but that's OK.

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
TomFinley: I mean, if I think about a confusion matrix, all I really want is the matrix, maybe a few summary statistics. Making this API for this limited object responsible for querying the metadata on the predicted label column *when we already have APIs to do that that are more general* seems like it is just making the story harder than it needs to be.

---
In reply to: [275473328](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275473328) [](ancestors = 275473328,275067993)

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
sfilipi: Ivan has commented after i marked it as resolved, and i marked it as resolved because i thought your comment was just an observation, not a suggestion for change. I'll swap it. 

---
In reply to: [275472667](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275472667) [](ancestors = 275472667,275082837,275005362)

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
sfilipi: I meant the Annotations in the original IDV that gets created by the evaluator. Expanding the comment with the above note. 

---
In reply to: [275468277](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275468277) [](ancestors = 275468277)

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
sfilipi: You can't do extrapolation in ChecParam, they need to be string literal. 

---
In reply to: [275462838](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275462838) [](ancestors = 275462838)

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
sfilipi: moved it to a comment, per Tom's suggestion. 

---
In reply to: [275463006](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275463006) [](ancestors = 275463006)

sfilipi: moved it to a comment, per Tom's suggestion. 

---
In reply to: [275463153](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275463153) [](ancestors = 275463153)

test/Microsoft.ML.Functional.Tests/Common.cs
sfilipi: ofc it'll happen, there is not compiler complaining about it :)

---
In reply to: [275464491](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275464491) [](ancestors = 275464491)

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
sfilipi: caching it and calculating it when first asked for.

---
In reply to: [275469566](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275469566) [](ancestors = 275469566)

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
sfilipi: Actually removed the param, so description went away too. 

---
In reply to: [275527727](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275527727) [](ancestors = 275527727,275468277)

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
sfilipi: Did the change. Just wanted to mention that the reason i used arrays in the first place, is because i believe arrays are preferred over lists when it comes to small collections that don't get updated frequently. 

---
In reply to: [275504885](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275504885) [](ancestors = 275504885,275472667,275082837,275005362)

TomFinley: Thanks @sfilipi. You may be right, but for things like this I feel like using the more "consistent" solution with what we've done elsewhere, may be more desirable. (Even if one or two places we do use immutable arrays. See also `VectorDataViewType.Dimensions`, but that one was deliberate.)

It may be that in future we may decide using immutable arrays is more desirable. Incidentally this is allowed by [the rules](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md) is "Returning a value of a more derived type for a property, field, return or out value", so in principle we *could* change what we think of as read only lists to immutable arrays.

---
In reply to: [275554556](https://github.com/dotnet/machinelearning/pull/3250#discussion_r275554556) [](ancestors = 275554556,275504885,275472667,275082837,275005362)

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <param name="confusionTableCounts">The counts of the confusion table. The actual classes values are in the rows of the 2D array,
+        /// and the counts of the predicted classes are in the columns.</param>
+        /// <param name="labelNames">The predicted classes names, or the indexes of the classes, if the names are missing.</param>
+        /// <param name="isSampled">Whether the classes are sampled.</param>
+        /// <param name="isBinary">Whether the confusion table is the result of a binary classification. </param>
+        /// <param name="classIndicators">The Annotations of the Count column, in the confusionTable idv.</param>
+        internal ConfusionMatrix(IHost host, double[] precision, double[] recall, double[][] confusionTableCounts,
+            List<ReadOnlyMemory<char>> labelNames, bool isSampled, bool isBinary, DataViewSchema.Annotations classIndicators)
+        {
+            _host = host;
```
TomFinley: >_host = host; [](start = 12, length = 13)

Not a big deal but while you're at it, might help to `Contracts.AssertValue(host)` before assigning it.

sfilipi:
@Ivanidzo4ka @TomFinley as you find time. Thanks.

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,123 @@
...
+                counts.Add(confusionTableCounts[i].ToList().AsReadOnly());
+
+            Counts = counts.AsReadOnly();
+        }
+
+        /// <summary>
+        /// Returns a human readable representation of the confusion table.
+        /// </summary>
+        /// <returns></returns>
+        public string GetFormattedConfusionTable() {
```
TomFinley: >  [](start = 50, length = 1)

Nit: formatting, in C# code the first brace tends to go on next line.

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        internal readonly bool IsSampled;
+        internal readonly bool IsBinary;
+
+        private readonly IHost _host;
+        private string _formattedConfusionMatrix;
+
+        /// <summary>
+        /// The confusion matrix as a structured type, built from the counts of the confusion table <see cref="IDataView"/> that the <see cref="BinaryClassifierEvaluator"/> or
+        /// the <see cref="MulticlassClassificationEvaluator"/> construct.
```
TomFinley: >construct [](start = 64, length = 9)

I think we meant constructor, but since it's internal anyway and will remain so I wouldn't say it's a big deal.

sfilipi: someone leaves sentences in half, and someone leaves words :)
Thanks for the catch.

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,123 @@
...
+        public IReadOnlyList<double> PerClassPrecision { get; }
+
+        /// <summary>
+        /// The calculated value of <a href="https://en.wikipedia.org/wiki/Precision_and_recall#Recall">recall</a> for each class.
+        /// </summary>
+        public IReadOnlyList<double> PerClassRecall { get; }
+
+        /// <summary>
+        /// The confsion matrix counts for the combinations actual class/predicted class.
+        /// The actual classes are in the rows of the table, and the predicted classes in the columns.
```
artidoro: >rows of the table, and the predicted classes in the columns [](start = 42, length = 59)

I guess that what is a row and what is a column is subjective. Maybe we could talk about the outer/inner index or first/second index?

src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// </summary>
+        public IReadOnlyList<IReadOnlyList<double>> Counts { get; }
+
+        /// <summary>
+        /// The indicators of the predicted classes.
+        /// It might be the classes names, or just indices of the predicted classes, if the name mapping is missing.
+        /// </summary>
+        public int NumberOfPredictedClasses { get; }
```
artidoro: Is this just meant to give the  number of classes?
Maybe we could simplify both the name and the comment.

Suggestion: `NumberOfClasses`

src/Microsoft.ML.Data/Evaluators/EvaluatorUtils.cs
```diff
@@ -1348,20 +1348,49 @@ internal static class MetricWriter
...
-            host.Check(confusionDataView.Schema.TryGetColumnIndex(MetricKinds.ColumnNames.Count, out countCol), "Did not find the count column");
-            var type = confusionDataView.Schema[countCol].Annotations.Schema.GetColumnOrNull(AnnotationUtils.Kinds.SlotNames)?.Type as VectorDataViewType;
-            host.Check(type != null && type.IsKnownSize && type.ItemType is TextDataViewType, "The Count column does not have a text vector metadata of kind SlotNames.");
+            var weightColumn = confusionDataView.Schema.GetColumnOrNull(MetricKinds.ColumnNames.Weight);
+            bool isWeighted = weightColumn.HasValue;
 
+            var confusionMatrix = GetConfusionMatrix(host, confusionDataView, binary, sample, false);
+            var confusionTableString = GetConfusionTableAsString(confusionMatrix, false);
+
+            // if there is a Weight column, return the weighted confusionMatrix as well, from this function.
```
artidoro: nit: capital letter.

Issue #1795, opened 2018-11-30
Evaluator of one-class matrix factorizaition is missing


wschin:
One class matrix factorization is solved as a regression problem but common metrics in practices are top-k accuracy, nDCG, etc.

wschin:
My plan is first convert `GroupTransform` to `Transformers` because we need it to convert
| User        | Item           | Score  |
| ------------- |:-------------:| -----:|
| 0      | 1 | 0 |
| 3    | 1      |   1 |
| 0 | 3      |    1 |
to

codemzs:
@wschin do we still need this? I'm closing but feel free to reopen.

PR #3251, opened 2019-04-08
[WIP] Matrix factorization transformer

wschin:
In addition to a scorer, matrix factorization also commonly acts like a featurizer. It can map matrix row/column index to its representation. If we factorize matrix `R` into `P^TQ`, then the row index u's latent representation is defined as the u-th column of P. Similarly, the column index v's latent representation is the v-th column of Q. In addition, this transformer can find columns/rows similar to the given row/column by computing the inner products between row/column and all other columns/rows. The following data structure summarizes this transformer's output schema.
```csharp
        private class LatentVectorAndNeighbors
        {
            [VectorType(2)]
            public float[] ColumnLatentVector { get; set; }

            [VectorType(2)]
            public float[] RowLatentVector { get; set; }

            [VectorType(1)]
            [KeyType(_oneClassMatrixRowCount)]
            public uint[] SimilarRows { get; set; }

            [VectorType(1)]
            [KeyType(_oneClassMatrixColumnCount)]
            public uint[] SImilarColumns { get; set; }
        }
```
This PR fix partially #1795 because we still need another evaluator.

### DIFF
--- a/src/Microsoft.ML.Recommender/MatrixFactorizationPredictor.cs
+++ b/src/Microsoft.ML.Recommender/MatrixFactorizationPredictor.cs
@@ -110,7 +110,7 @@ namespace Microsoft.ML.Trainers.Recommender
             MatrixRowIndexType = matrixRowIndexType;
         }
 
-        private MatrixFactorizationModelParameters(IHostEnvironment env, ModelLoadContext ctx)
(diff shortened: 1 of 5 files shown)


eerhardt:
This PR is out of date. Please resolve conflicts and re-open if you are interested in continuing this work.

PR #3252, opened 2019-04-08
Renamed WhatTheFeature and their abbreviations  to FeatureContributionScorer.

glebuk:
### Change: 
We need to rename some outdated component names to fit the term review policy.  Specifically we need to rename all instances of WhatTheFeature to the FutureContributionScorer for it and all abbreviations.  The change cleans up the last instances of old name of this feature.

### Impact:
Note that this change will break model compatibility for all previous versions of models containing FCC scorer.  

### Workaround:
The workaround for legacy models exists.  It requires users to open the model ZIP file and manually rename the binary model file identifier string that contains the FCC model from "WTF SCBI" to "FCC SCBI"


glebuk:
Fixes #3255

TomFinley:
I'm sorry I can see why we would change the name but is it really necessary to break b a ckcompat? Could we perhaps obfuscate the legacy signature in some fashion?

glebuk:
@TomFinley  - the issue here is that the offending string is in the binary file and visible there.  When someone opens it it would be noticeable. Moreover it is outdated as we have changed the name.
The question is - how many binary models in the wild would contain that component?  I hope we have some license with the 4.0 release to make a few breaking changes.

rogancarr:
@glebuk The feature is very popular. I would recommend we create a conversion tool we ship with the 4.0 to migrate models.

TomFinley:
> I hope we have some license with the 4.0 release to make a few breaking changes.

*Some* license. And, TBH, I think it will be fine. Yet, the string you refer to would only be visible to people that had saved this under a prior version, either of ML.NET or the internal tool. The damage there is already done. My hope would be we would have some mechanism of not inconveniencing these people.

So, let me point you to @rogancarr's statement to say why I consider this ill-advised...

> Approve, but we need documentation to preserve backwards-compatibility with the internal repository.

So, we're going to have to have, *somewhere* in our documentation, ***a description of that old signature so that they know what they must look for, what they have to replace to make the old models work, and they will have to do it by hand -- no code of ours can, by the thoughtless application of this policy helpful.*** So that, you know, they have enough information to actually pursue that. And what do you think? Is that going to make this issue *more* conspicuous? Or *less* conspicuous?

This strikes me as a perfect example of someone following the letter of the law in such a mechanical fashion that they do not notice when they have completely violated its entire spirit and purpose.

rogancarr:
@TomFinley 

>And, TBH, I think it will be fine. Yet, the string you refer to would only be visible to people that had saved this under a prior version, either of ML.NET or the internal tool.

The string is currently in our public source code, which I believe is the actual problem.

We *could* consider supporting the old signature in the internal codebase, couldn't we?

PR #3253, opened 2019-04-09
Fixing ONNX test

### LINKED ISSUE #2981: Cannot fit an Onnx Transform as part of a pipeline
It is possible to fit an ONNX model by itself:

```cs
var onnxEstimator = mlContext.Transforms.ApplyOnnxModel(modelPath)
var onnxModel = onnxEstimator.Fit(data);
```

But it throws when it is part of a pipeline:
```cs
var onnxEstimator = mlContext.Transforms.ApplyOnnxModel(modelPath)
    // TODO #2980: ONNX outputs don't match the outputs of the model, so we must hand-correct this for now.
    .Append(mlContext.Transforms.CopyColumns("Score", "Score0"));
    .Append(mlContext.Transforms.CopyColumns("Label", "Fable"));
    .Append(mlContext.Transforms.NormalizeLpNorm("Features2", "Features", LpNormNormalizingEstimatorBase.NormFunction.L2));
var onnxModel = onnxEstimator.Fit(data);
```
Any of these `Append` statements cause a throw. They do not affect the ONNX model at all, and use either the result of the calculation (`Score0`), rows unused by the model (`Label`) or rows also used by the transform (`Features`).

In this case, the error message is:
`System.ArgumentOutOfRangeException : Schema mismatch for input column 'Label': expected vector, got R4
Parameter name: inputSchema`

singlis:
The test pipeline for consuming an ONNX model would fail due to the Score column being named "Score0". The ONNX model will rename the output columns by design, therefore a different class with the ColumnName of "Score0" is needed. This fixes the test pipeline to address this issue.

Fixes #2981


test/Microsoft.ML.Functional.Tests/ONNX.cs
```diff
@@ -51,15 +58,9 @@ public void SaveOnnxModelLoadAndScoreFastTree()
             var onnxEstimator = mlContext.Transforms.ApplyOnnxModel(modelPath);
             var onnxModel = onnxEstimator.Fit(data);
 
-            // TODO #2980: ONNX outputs don't match the outputs of the model, so we must hand-correct this for now.
-            // TODO #2981: ONNX models cannot be fit as part of a pipeline, so we must use a workaround like this.
```
rogancarr: >// TODO #2981: [](start = 12, length = 14)

Maybe keep this TODO in until it's fixed?

test/Microsoft.ML.Functional.Tests/ONNX.cs
```diff
@@ -51,15 +58,9 @@ public void SaveOnnxModelLoadAndScoreFastTree()
             var onnxEstimator = mlContext.Transforms.ApplyOnnxModel(modelPath);
             var onnxModel = onnxEstimator.Fit(data);
 
-            // TODO #2980: ONNX outputs don't match the outputs of the model, so we must hand-correct this for now.
```
rogancarr: >puts don't match the outputs o [](start = 35, length = 30)

Can we keep a comment about this somewhere in the test? This is By Design, but nonetheless confusing.

test/Microsoft.ML.Functional.Tests/ONNX.cs
```diff
@@ -14,6 +15,12 @@
 
 namespace Microsoft.ML.Functional.Tests
 {
+    internal sealed class OnnxScoreColumn
```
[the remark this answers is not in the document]
rogancarr: Ditto

---
In reply to: [273552588](https://github.com/dotnet/machinelearning/pull/3253#discussion_r273552588) [](ancestors = 273552588)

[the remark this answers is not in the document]
singlis: I noticed there is a VectorScoreColumn class, using that instead.

test/Microsoft.ML.Functional.Tests/ONNX.cs
singlis: @rogancarr  - Thanks! I ended up fixing this, see the changes in GetOutputSchema in OnnxTransform.

src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
```diff
@@ -564,8 +564,6 @@ public override SchemaShape GetOutputSchema(SchemaShape inputSchema)
                 var input = Transformer.Inputs[i];
                 if (!inputSchema.TryFindColumn(input, out var col))
                     throw Host.ExceptSchemaMismatch(nameof(inputSchema), "input", input);
-                if (!(col.Kind == SchemaShape.Column.VectorKind.VariableVector || col.Kind == SchemaShape.Column.VectorKind.Vector))
-                    throw Host.ExceptSchemaMismatch(nameof(inputSchema), "input", input, "vector", col.GetTypeString());
```
jignparm: if input col.Kind == VariableVector, how is that case handled? If not handled, should throw exception.

rogancarr:
LGTM @singlis

src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
singlis: so prior to this change, we allowed VariableVector in that it would not throw -- this change now allows for vector and simple types. I would prefer to have a test that uses  VariableVector, but I have yet to find something that provides a VariableVector output AND can be saved to ONNX.

singlis: @jignparm from our conversation, you mentioned that we should not have allowed VariableVector to begin with and should fix as part of this change. I have updated the code to reflect that, but we should have a test that confirms if this works correctly. I have created the following issue for tracking #3375

PR #3263, opened 2019-04-09
Add DataView Type Register (supporting loading in-memory image and other in-memory custom types)

### LINKED ISSUE #445: Direct API: RoleMappedSchema/Data Cleanup, Improvement
Another followup to #371, in which we discuss changes to `RoleMappedSchema` and `RoleMappedData` to make them less idiosyncratic.

`RoleMappedSchema` and `RoleMappedData` are structures that solve the following problem: Once you create a pipeline, before you feed it to an `ITrainer` or similar structure, you must have some mechanism to communicate to consumers of that pipeline, what all the columns were actually for... e.g., which column(s) were feature columns, which the label, and so on. (Before this structure existed, our "solution" to this was that every component consumed an `IDataView` directly and had configurable options for someone to declare which was which. This is good in that each trainer had the chance to be explicit about what it wanted, but still was somewhat troublesome since having to tell absolutely every component we wanted to use, "OK, these are still the feature columns" became somewhat troublesome, and a source of user error. So a structure to make this assignment more "sticky" was invented.

So that is all fine, more or less. And, I'd say on the whole it is a pretty good class, insofar that it seems to have worked well for its purpose. However there are wrinkles we probably ought to clean up.

Nearly all architecture effort went into making it easy to consume, as opposed to being easy or sensible to create. Previously, this made sense, since it was only instantiated in a handful of places, and used in hundreds of places. With API usage, the situation is reversed: we expect everyone to create it, and there will be "only" hundreds of consumers.

* On that subject, creation is somewhat odd: there are `Create` and `CreateOpt` methods, as opposed to how most people would imagine an object is created, through an actual constructor (maybe with a `bool opt = false` parameter.)

* "Reapplication" of an existing role-mapping to new data is a common operation performed in the code-base, yet there is no convenience for it, and it's something we'd want people to be able to do relatively easily. (E.g., when applying caching, for example.)

* The common convenience helpers for the most common cases of creating `RoleMappedData` exist (e.g., "these are my features, these are my labels) exist in a `TrainUtils` class. This makes them impossible to discover unless you know where to look. Probably the easiest to discover place to have these conveniences would be on the classes themselves. (This would also be a start at cleaning up `TrainUtils`, which is basically a haphazard bag of vaguely useful things.)

* General cleanup of the code. A relic from a bygone time, `Id`, was never removed, despite being irrelevant and never used any more. It's been years since it was replaced with ids on `IRowCursor` directly. Also a fair amount of code exists in the class to detect conditions that cannot possibly happen.

### LINKED ISSUE #2121: ImageLoadingEstimator for TensorFlow scoring should allow in-memory image streams as input in addition to images from files on drive
Right now the only way for ML.NET to load images is via ImageLoadingEstimator, which can load them *only* from disk files (as confirmed by @yaeldekel and Pete a few weeks ago).

However, it is a very common scenario in applications, such as a web app, where users submit images through Http, then the DataView/pipeline would load in-memory image streams (either BitMap, byte[], Image) instead of loading images from files in a folder on a disk/drive. 

That's the right way to do it for many scenarios in web apps and services (Web APIs).
And for instance, you can do that when using TensorFlowSharp in C#. But we cannot in ML.NET, as of today.

When implementing this feature improvement in ML.NET, there could be the following two approaches:

- Modify schema comprehension to be able to map Bitmap fields/properties to Image columns of a data view.

- Add another version of ImageLoading transformer that loads/decodes the image from a byte vector, rather than from a disk file identified by path.

In any case, this is an important scenario to implement because not being able to load images from in-memory streams and only from files can be a big handicap in performance for on-line scenarios like the ones mentioned.

With the current implementation in ML.NET, the only workaround is to save the upcoming image from http and in-memory into a temporary file on the disk and load it from there. But that is a very "coarse/poor" workaround, not performant at all for a real application in production. 

The following is a sample app I created for this online scenario where the user uploads an image from the browser into a service (Web API) and ultimately you get it as an in-memory image stream.

SEE CODE HERE:

https://github.com/CESARDELATORRE/TensorFlowImageClassificationWebAPI

image

- That web form uploads the image through Http into a service (Web API) in the server-side. At that point, the image is an in-memory image stream.

- In this implementation the sample app works because I implemented a workaround so the submitted image is temporarily stored as a file, then loaded from the file into the DataView through the pipeline...)

Basically, when the C# method in the Web API gets the image as an in-memory stream it should be able to load it directly in the DataView. The following code is an example:

```cs
        // Controller's method from Web API 
        [HttpPost]
        [ProducesResponseType(200)]
        [ProducesResponseType(400)]
        [Route("classifyimage")]
        public async Task<IActionResult> ClassifyImage(IFormFile imageFile)
        {
                if (imageFile.Length == 0)
                    return BadRequest();

                // WORKAROUND: Save image into a temporal file
                //Save the temp image image into the temp-folder 
                string fileName = await _imageWriter.UploadImageAsync(imageFile, _imagesTmpFolder);
                string imageFilePath = Path.Combine(_imagesTmpFolder, fileName);

                // Use image filename as the workaround...
                // Rest of the implementation with ML.NET API for scoring TensorFlow model...
                // ...
           
        }
```

To sum up:

I believe it is "a must" for ML.NET to be able to load in-memory image streams into the DataView to use those images when scoring TensorFlow models (in addition "from files") because of the mentioned on-line and in-memory scenarios that are pretty common.

### LINKED ISSUE #2495: Create sample/test on ImageLoadingEstimator for TensorFlow scoring for loading an in-memory image stream (instead of files from the drive)
Related to [this issue](https://github.com/dotnet/machinelearning/issues/2121), we need a clear/simple sample code/test for ImageLoadingEstimator for TensorFlow scoring so it loads an in-memory image stream as input (instead image files from a drive).

I couldn't see a clear way to do it here (Mentioned by Gleb): https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Tests/ImagesTests.cs

### LINKED ISSUE #3162: Schema comprehension fails for type System.Drawing.Bitmap
Several of the image transforms in ML.NET produce intermediate columns of type System.Drawing.Bitmap

https://github.com/dotnet/machinelearning/blob/e5cbca78683e3c1036966e07b7fd7d1b180c1562/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs#L39-L42

If we try to use  schema comprehension to  map some intermediate data (e.g. `transformedData`) to an IEnumerable,  we  hit an exception  when  dealing with  a column of type ` System.Drawing.Bitmap`

**Error**: Unhandled Exception: System.ArgumentOutOfRangeException: Could not determine an IDataView type for member ImageObject

@Ivanidzo4ka

### LINKED ISSUE #3274: IDataView Type system needs to be extensible to properly support image, date, and other types
To support loading `Image` from `IDataView` (as shown in #3263), user must modify ML.NET code, which means our type system is not extensive. Do we want to invest on the extensibility of ML.NET type system?

### LINKED ISSUE #3369: Example code on how to load in-memory images into ONNX models and TensorFlow models
I'm reviving these issues. Basically, we need example code on how  to load in-memory images into pipeline (for ONNX and TensorFlow models):

https://github.com/dotnet/machinelearning/issues/2495

https://github.com/dotnet/machinelearning/issues/2121

There are many scenarios that require that way instead of loading from image files on the drive.

Scenarios:
- Live-streaming images coming from a laptop camera for object detection ONNX models
- Web/Service Http apps where images come through HTTP and need to be processed directly

Can the team create example code of ONNX model and a second for a TensorFlow model where in-memory images are provided instead of loading from files on the drive?

### LINKED ISSUE #3460: Is there any way to pass pre-loaded image into MLContext pipeline?
### System information

- **OS version/distro**: Windows 10.0.17134
- **.NET Version (eg., dotnet --info)**:  .NET Core 2.2.202

### Issue

- **What did you do?**
My intention was to implement object detection in video. Please see attached code sample and explanation below.
- **What happened?**
I wasn't able to find proper API methods for configuring MLContext pipeline.
- **What did you expect?**
I expected it should be possible to pass individual in-memory representation of image(as a Bitmap object for example) into pipeline. 

### Source code / logs

Let's pretend we have a task to run object detection model on frames coming from video stream. 
It seems natural to pass 1 frame one by one as they arriving from camera. 

I can't find such possibility with current MLContext pipeline and API available. 

Here's example pipeline from [machinelearning-samples](https://github.com/dotnet/machinelearning-samples/blob/master/samples/csharp/getting-started/DeepLearning_ObjectDetection_Onnx/ObjectDetectionConsoleApp/OnnxModelScorer.cs#L61) :

```
            var pipeline = mlContext.Transforms.LoadImages(outputColumnName: "image", imageFolder: imagesFolder, inputColumnName: nameof(ImageNetData.ImagePath))
                            .Append(mlContext.Transforms.ResizeImages(outputColumnName: "image", imageWidth: ImageNetSettings.imageWidth, imageHeight: ImageNetSettings.imageHeight, inputColumnName: "image"))
                            .Append(mlContext.Transforms.ExtractPixels(outputColumnName: "image"))
                            .Append(mlContext.Transforms.ApplyOnnxModel(modelFile: modelLocation, outputColumnNames: new[] { TinyYoloModelSettings.ModelOutput }, inputColumnNames: new[] { TinyYoloModelSettings.ModelInput }));
```

Having LoadImages extension method overload accepting in-memory image representation (Bitmap or whatever) would definetly solve a problem:

`public static ImageLoadingEstimator LoadImages(this TransformsCatalog catalog, params Bitmap[] images);`

Am I missing some way implement this without storing frames onto the disk? Or does it mean ML.NET wasn't designed for realtime video processing?

Thanks in advance.

### LINKED ISSUE #3723: IDataView Type System needs to be extensible to support Image scenarios properly
### The Problem
Currently it is impossible make prediction from images if they do not come from file.  PredictionEngine<IT,OT> and IEnumerable<T> to IDV is impossible if T contains images/

### The Root Cause
The fundamental issue is that in several places in our system,  our types are hardcoded and limited to a small set. Actual IDataView type system IS open, so it should be possible to support any type. However, some of our machinery is not so extensible.   For example, if we need to add full support of images, we would need to update:
- converters between IEnumerable<T> <-> IDataView, 
- PredictionEngine<> 
- API Utils, 
- Schema utils 
- etc.

What we need is a dependency-injection mechanism to allow component developers to inject definitions of types supported by our IDV machinery.  Such mechanism should be flexible and extensible.  For an example of hard-coded implementation, check #3263.  It is illustrative but incorrect as it is not flexible.  Note @TomFinley's comment on different options on how to do this properly.  We need to define a mechanism that is extensible, so that a dev can add new transforms to work on new data types, such as sound and enjoy all benefits of ML.NET

### Definition of Done:
- Mechanism to register additional IDV types so that they are supported by PredictionEngine and IEnumerable converters.
- Add such registration for images and date types
- Ensure that the registration (at least for images)  occurs automatically whenever appropriate component is used (for example, if image resizing transform is used, the image type is registered automatically )
- Ensure that prediction engine, conversion to and from IEnumerable work correctly with tests.

### The following issues will also be solved by this issue:
#3369, #3460, #3582, #2121, #2495, #3582, #3274
In addition, this work is needed to properly support ONNX and TF scenarios for structured data such as image, speech, video, or Audio.

### LINKED ISSUE #3784: Bitmap data
What is (I assume a variable) Enumerable and how do you get it from a Bitmap or Image, neither of which is Enumerable? Two loops to examine the pixels and put it into a linear array?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 8b9c0132-7901-8114-b8bc-faa09ad74797
* Version Independent ID: 8274abb2-6952-674b-e8bc-1e314aeee4cc
* Content: [TensorFlowModel.ScoreTensorFlowModel Method (Microsoft.ML.Transforms)](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.transforms.tensorflowmodel.scoretensorflowmodel?view=ml-dotnet-preview#feedback)
* Content Source: [dotnet/xml/Microsoft.ML.Transforms/TensorFlowModel.xml](https://github.com/dotnet/ml-api-docs/blob/live/dotnet/xml/Microsoft.ML.Transforms/TensorFlowModel.xml)
* Product: **dotnet-ml-api**
* GitHub Login: @sfilipi
* Microsoft Alias: **johalex**

wschin:
Fix #3162 by considering image type in our bridge code. Also fix #3723, fix #3369, fix #3274, fix #445, fix #3460, fix #2121, fix #2495, fix #3784.


test/Microsoft.ML.Tests/ImagesTests.cs
```diff
@@ -184,6 +184,53 @@ public void TestGreyscaleTransformImages()
...
+            }).Load(imagesDataFile);
+
+            var imagesFolder = Path.GetDirectoryName(imagesDataFile);
+            // Image loading and conversion pipeline.
+            var pipeline = ML.Transforms.LoadImages("ImageObject", imagesFolder, "ImagePath")
+                           .Append(ML.Transforms.ConvertToGrayscale("Grayscale", "ImageObject"));
+
+            var transformedData = pipeline.Fit(data).Transform(data);
+
+            var transformedDataPoints = ML.Data.CreateEnumerable<TransformedImageDataPoint>(transformedData, true).ToList();
```
wschin: This place we successfully assign images created in ML.NET to a C# enumerable.

src/Microsoft.ML.Core/Microsoft.ML.Core.csproj
```diff
@@ -10,9 +10,9 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.ML.DataView\Microsoft.ML.DataView.csproj" />
-    
     <PackageReference Include="System.Collections.Immutable" Version="$(SystemCollectionsImmutableVersion)" />
     <PackageReference Include="System.Memory" Version="$(SystemMemoryVersion)" />
+    <PackageReference Include="System.Drawing.Common" Version="$(SystemDrawingCommonPackageVersion)" />
```
Ivanidzo4ka: >SystemDrawingCommonPackageVersion [](start = 65, length = 33)

I think this is main reason why ImageDataType is in separate assembly.
I personally would prefer to make dataview system extensible, some kind of register mechanism in dataview type so we can have types necessary for certain workflow be separated from main nuget.

wschin: The parsing framework now depends on image type. How can it be in another assembly? It sounds a circular dependency, right? In addition, refactoring ML.NET to support customized type doesn't sound a small task, so I think we need shiproom's approval.

---
In reply to: [273717323](https://github.com/dotnet/machinelearning/pull/3263#discussion_r273717323) [](ancestors = 273717323)

Ivanidzo4ka: >The parsing framework now depends on image type. 
What? [https://media1.tenor.com/images/d1fbf17571ea3260dee054ceebf65fd7/tenor.gif?itemid=8605657](https://media1.tenor.com/images/d1fbf17571ea3260dee054ceebf65fd7/tenor.gif?itemid=8605657)

All I'm trying to say is I'm not sure it's worth to move image type inside basic nugets for the price of adding another dependency.

I agree support for customized type is quite big task, and shouldn't be part of 1.0.
But it can be part of 1.1


---
In reply to: [273720375](https://github.com/dotnet/machinelearning/pull/3263#discussion_r273720375) [](ancestors = 273720375,273717323)

wschin: This is a standard library downloaded by million times and also support .NET Core, so maybe adding it is fine?

---
In reply to: [273721913](https://github.com/dotnet/machinelearning/pull/3263#discussion_r273721913) [](ancestors = 273721913,273720375,273717323)

src/Microsoft.ML.DataView/ImageDataViewType.cs
```diff
@@ -3,11 +3,9 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Drawing;
-using Microsoft.ML.Data;
-using Microsoft.ML.Internal.Utilities;
-using Microsoft.ML.Runtime;
+using Microsoft.ML.Internal.DataView;
 
-namespace Microsoft.ML.Transforms.Image
+namespace Microsoft.ML.Data
```
eerhardt: These kinds of changes are API breaking. We can't be making these changes after 1.0.

src/Microsoft.ML.Core/Microsoft.ML.Core.csproj
```diff
@@ -10,7 +10,6 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.ML.DataView\Microsoft.ML.DataView.csproj" />
-
```
eerhardt: Can you revert this file?

src/Microsoft.ML.DataView/ImageDataViewType.cs
```diff
@@ -3,11 +3,9 @@
...
 {
     public sealed class ImageDataViewType : StructuredDataViewType
```
eerhardt: I thought the whole point of having the `ImageDataViewType` in a separate assembly is to show that the DataView type system is open and extensible. If we must have all the DataViewType class in `Microsoft.ML.DataView`, that completely breaks this extensibility story.

I'm not sure we should move ImageDataViewType into this assembly.

cc @tomfinley

src/Microsoft.ML.DataView/Microsoft.ML.DataView.csproj
```diff
@@ -8,6 +8,7 @@
   <ItemGroup>
     <PackageReference Include="System.Collections.Immutable" Version="$(SystemCollectionsImmutableVersion)" />
     <PackageReference Include="System.Memory" Version="$(SystemMemoryVersion)" />
+    <PackageReference Include="System.Drawing.Common" Version="$(SystemDrawingCommonPackageVersion)" />
```
eerhardt: If we really do go through with this (I hope we can do something so we don't have to), you will need to update the PackageReference in the corresponding `nupkgproj` under the `pkg` folder as well.

src/Microsoft.ML.Core/Microsoft.ML.Core.csproj
eerhardt: Note: the kind of change you are making here is not possible to do after `1.0` ships. You cannot change namespaces on types - it is an API breaking change.

Also, if you move one type from one assembly to another, it is a breaking change (unless you add a `TypeForwardedTo` attribute).

TomFinley: So, thanks so much for bringing this problem up. I definitely agree it's a real problem. That said, let me just give a hard "no" on this particular solution, not because it doesn't work for *this* custom type, but because there's no conceivable way it could work for the hypothetical next custom type (for the reasons @eerhardt has already mentioned).

Let me just propose the plausible workable alternative, which has the benefit of actually doing something we've successfully relied on for years and years. Our model loading logic is built on a system where the model loading/saving core relies in a fairly central nuget, but it does not have direct references. It does this through the component catalog. We also use this, for example, for custom mappings, where user code (which we obviously can't depend on!!) can register their own plugins. If we delve into the internal Microsoft usage of this tool, we also have the experience of our command line -- plenty of `ICommand` implementations rely upon the existence of things like trainers and transforms that they cannot possibly directly reference. I could imagine this same strategy -- the schema deinfition and prediction engine code having access to components that handle this "mapping" -- working quite similarly.

More specifically, I could easily imagine that when someone defines a new derived type from `StructuredDataViewType`, they have the *option* of registering a plugin to this schema definition object that our reflection-based mechanism relies on, to provide whatever the appropriate mapping logic is, to facilitate their reading-in and writing-out from ML pipelines. If they do, great. If not, so be it, we don't offer it.

If you don't like that `ComponentCatalog` based solution, you could even imagine in some hypothetical alternate mechanism for creating these prediction engine that takes the mapping logic *directly* as another (again optional) parameter... or works through an attribute on the property or field on the user class (with an abstract base class in the core assemblies and derived classes in the more leaf assemblies) to describe the logic. Or a static registration mechanism by which user code can tell this, "hey, when you see this StructuredDataViewType and a field/property with this .NET representation type, this is what you do."

The good news is that all this can be done post 1.0. The only reasonable solution I could imagine that could perhaps not be done post 1.0, is if we chose to make the logic of how to handle a type embedded in the `StructuredDataViewType` class itself -- that is, how to "safely" copy from one type to another, and such things as this. But even that could be done post-1.0, really. Imagine a new interface, `IAmATypeThatKnowsHowToReadAndWrite<T>`, that's something we could always add later I think, without it being a breaking change. Who knows.

Anyway, plenty of options I think, all of which can be done post 1.0 and are I *think* superior to this solution, if you'll forgive my saying so?

The happy part about that is, I don't think any of that needs to be done prior to 1.0. I could easily imagine, say, a new abstract class, or a new plugin class, to be -- just how, hypothetically, the custom mapper transform *could* have been implemented after 1.0 (it obviously wasn't, it was one of the first things we did, but, if we hypothetically hadn't it would not represent a breaking cahnge)

src/Microsoft.ML.Data/Utils/ApiUtils.cs
```diff
@@ -23,7 +23,8 @@ private static OpCode GetAssignmentOpCode(Type t)
             if (t == typeof(ReadOnlyMemory<char>) || t == typeof(string) || t.IsArray ||
                 (t.IsGenericType && t.GetGenericTypeDefinition() == typeof(VBuffer<>)) ||
                 (t.IsGenericType && t.GetGenericTypeDefinition() == typeof(Nullable<>)) ||
-                t == typeof(DateTime) || t == typeof(DateTimeOffset) || t == typeof(TimeSpan) || t == typeof(DataViewRowId))
+                t == typeof(DateTime) || t == typeof(DateTimeOffset) || t == typeof(TimeSpan) ||
```
glebuk: >DateTimeOffset [](start = 53, length = 14)

Should date*  types go into TypeManager?  They are structs, I guess.

src/Microsoft.ML.DataView/TypeManager.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        public static Type GetRawTypeOrNull(DataViewType type)
+        {
+            if (_dataViewTypeToRawTypeMap.ContainsKey(type))
+                return _dataViewTypeToRawTypeMap[type];
+            else
+                return null;
+        }
+
+        public static void Register(Type rawType, DataViewType dataViewType)
```
glebuk: >Register [](start = 27, length = 8)

This should be idempotent.  Because various dlls might register the same type again and again (imagine same type is used in 2 nugets, then they would both have that same code.)  We should only throw if one registration differes from already complete

src/Microsoft.ML.DataView/TypeManager.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        /// <summary>
+        /// Constructor to initialize type mappings.
+        /// </summary>
+        static TypeManager()
+        {
+            _notAllowedRawTypes = new HashSet<Type>() {
+                typeof(Boolean), typeof(SByte), typeof(Byte),
+                typeof(Int16), typeof(UInt16), typeof(Int32), typeof(UInt32),
+                typeof(Int64), typeof(UInt64), typeof(string), typeof(ReadOnlySpan<char>)
```
glebuk: >ReadOnlySpan [](start = 70, length = 12)

what about datetimes here?  they either are registered or not allowed.

src/Microsoft.ML.DataView/TypeManager.cs
```diff
@@ -0,0 +1,59 @@
...
+        private static ConcurrentDictionary<Type, DataViewType> _rawTypeToDataViewTypeMap;
+        private static ConcurrentDictionary<DataViewType, Type> _dataViewTypeToRawTypeMap;
+
+        /// <summary>
+        /// Constructor to initialize type mappings.
+        /// </summary>
+        static TypeManager()
+        {
+            _notAllowedRawTypes = new HashSet<Type>() {
+                typeof(Boolean), typeof(SByte), typeof(Byte),
```
glebuk: >SByte [](start = 40, length = 5)

float and double?

src/Microsoft.ML.DataView/TypeManager.cs
```diff
@@ -0,0 +1,59 @@
...
+            else
+                return null;
+        }
+
+        public static void Register(Type rawType, DataViewType dataViewType)
+        {
+            if (_notAllowedRawTypes.Contains(rawType))
+                throw Contracts.ExceptParam(nameof(rawType), $"Type {rawType} has been registered as ML.NET's default type. " +
+                    $"so it can't not be registered again.");
+            if (_rawTypeToDataViewTypeMap.ContainsKey(rawType))
```
glebuk: >_rawTypeToDataViewTypeMap [](start = 16, length = 25)

this is not atomic.  Prefer regular dictionaries and locks.

test/Microsoft.ML.Tests/ImagesTests.cs
```diff
@@ -184,6 +184,78 @@ public void TestGreyscaleTransformImages()
             Done();
         }
 
+        [Fact]
+        public void TestGrayScaleInMemory()
```
glebuk: >TestGrayScaleInMemory [](start = 20, length = 21)

Need to add a testfor some other random type, such as Point and a custom transform that uses it from IEnumerable to IDV.

test/Microsoft.ML.Tests/ImagesTests.cs
```diff
@@ -184,6 +184,78 @@ public void TestGreyscaleTransformImages()
...
+            var imagesDataFile = SamplesUtils.DatasetUtils.DownloadImages();
+
+            var data = ML.Data.CreateTextLoader(new TextLoader.Options()
+            {
+                Columns = new[]
+                {
+                    new TextLoader.Column("ImagePath", DataKind.String, 0),
+                    new TextLoader.Column("Name", DataKind.String, 1),
+             }
+            }).Load(imagesDataFile);
```
glebuk: Simplify the test by creating bitmaps on the fly - for example 100% red or blue bitmap.  This would make it more realistic and simple code.

test/Microsoft.ML.Tests/ImagesTests.cs
glebuk: >TestGrayScaleInMemory [](start = 20, length = 21)

Need to add test for prediction engine!

glebuk:
> using System;

Add copyright header.

---
Refers to: src/Microsoft.ML.DataView/TypeManager.cs:1 in 8aface6. [](commit_id = 8aface60208fc1b92486380e4839443c16d95477, deletion_comment = False)

src/Microsoft.ML.DataView/TypeManager.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using Microsoft.ML.Internal.DataView;
+
+namespace Microsoft.ML.Data
+{
+    public static class TypeManager
+    {
+        // Types have been used in ML.NET type systems. They can have multiple-to-one type mapping.
+        // For example, UInt32 and Key can be mapped to uint. We enforce one-to-one mapping for all
```
glebuk: >We enforce [](start = 62, length = 10)

Who is "we" - make it less personal

src/Microsoft.ML.DataView/TypeManager.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using Microsoft.ML.Internal.DataView;
+
+namespace Microsoft.ML.Data
+{
+    public static class TypeManager
```
glebuk: add XML comment on what it is and why it's useful.  When to use it!

glebuk: >TypeManager [](start = 24, length = 11)

TypeManager is too generic of a name.  Perhaps rename to DataViewTypeManager?

src/Microsoft.ML.DataView/TypeManager.cs
```diff
@@ -0,0 +1,59 @@
...
+                typeof(Int16), typeof(UInt16), typeof(Int32), typeof(UInt32),
+                typeof(Int64), typeof(UInt64), typeof(string), typeof(ReadOnlySpan<char>)
+            };
+            _rawTypeToDataViewTypeMap = new ConcurrentDictionary<Type, DataViewType>();
+            _dataViewTypeToRawTypeMap = new ConcurrentDictionary<DataViewType, Type>();
+        }
+
+        public static DataViewType GetDataViewTypeOrNull(Type type)
+        {
+            if (_rawTypeToDataViewTypeMap.ContainsKey(type))
```
glebuk: >_rawTypeToDataViewTypeMap [](start = 16, length = 25)

can be simplified with TryGetValue to get it once, not twice.

src/Microsoft.ML.DataView/TypeManager.cs
glebuk: >public [](start = 8, length = 6)

Add xml comment to explain what it is for, what types you can use and so forth.

src/Microsoft.ML.DataView/TypeManager.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        public static void Register(Type rawType, DataViewType dataViewType)
+        {
+            if (_notAllowedRawTypes.Contains(rawType))
+                throw Contracts.ExceptParam(nameof(rawType), $"Type {rawType} has been registered as ML.NET's default type. " +
+                    $"so it can't not be registered again.");
+            if (_rawTypeToDataViewTypeMap.ContainsKey(rawType))
+                throw Contracts.ExceptParam(nameof(rawType), $"Repeated type registration. The raw type {rawType} " +
+                    $"has been associated with {_rawTypeToDataViewTypeMap[rawType]}.");
```
glebuk: >has [](start = 22, length = 3)

blank line after if statement

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -392,37 +392,42 @@ public static SchemaDefinition Create(Type userType, Direction direction = Direc
...
-                else
-                    itemType = ColumnTypeExtensions.PrimitiveTypeFromType(dataType);
-
                 // Get the column type.
                 DataViewType columnType;
-                var vectorAttr = memberInfo.GetCustomAttribute<VectorTypeAttribute>();
-                if (vectorAttr != null && !isVector)
-                    throw Contracts.ExceptParam(nameof(userType), $"Member {memberInfo.Name} marked with {nameof(VectorTypeAttribute)}, but does not appear to be a vector type", memberInfo.Name);
-                if (isVector)
+                if (TypeManager.GetDataViewTypeOrNull(dataType) == null)
```
glebuk: >GetDataViewTypeOrNull [](start = 32, length = 21)

this usage pattern implies an overload like:
bool TryGetDataViewType(out datatype) overload.

src/Microsoft.ML.Data/DataView/InternalSchemaDefinition.cs
```diff
@@ -187,7 +187,7 @@ public static void GetVectorAndItemType(Type rawType, string name, out bool isVe
 
             if (itemType == typeof(string))
                 itemType = typeof(ReadOnlyMemory<char>);
-            else if (!itemType.TryGetDataKind(out _))
+            else if (!itemType.TryGetDataKind(out _) && TypeManager.GetDataViewTypeOrNull(itemType) == null)
```
glebuk: >GetDataViewTypeOrNull [](start = 68, length = 21)

add a method for  TypeManager.Contains(itemtype)

src/Microsoft.ML.Data/DataView/DataViewConstructionUtils.cs
```diff
@@ -269,6 +269,10 @@ private Delegate CreateGetter(DataViewType colType, InternalSchemaDefinition.Col
                         return Utils.MarshalInvoke(delForKey, keyRawType, peek, colType);
                     }
                 }
+                else if (TypeManager.GetRawTypeOrNull(colType) != null)
+                {
+                    del = CreateDirectGetterDelegate<int>;
```
glebuk: >int [](start = 53, length = 3)

why int?

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
shmoradims: +1

---
In reply to: [284945327](https://github.com/dotnet/machinelearning/pull/3263#discussion_r284945327) [](ancestors = 284945327)

shmoradims: alternatively, you can assign on line 369 and remove the else condition on line 430.
columnType = TypeManager.GetDataViewTypeOrNull(dataType);

---
In reply to: [285245746](https://github.com/dotnet/machinelearning/pull/3263#discussion_r285245746) [](ancestors = 285245746,284945327)

src/Microsoft.ML.DataView/TypeManager.cs
```diff
@@ -0,0 +1,59 @@
...
+namespace Microsoft.ML.Data
+{
+    public static class TypeManager
+    {
+        // Types have been used in ML.NET type systems. They can have multiple-to-one type mapping.
+        // For example, UInt32 and Key can be mapped to uint. We enforce one-to-one mapping for all
+        // user-registered types.
+        private static HashSet<Type> _notAllowedRawTypes;
+        private static ConcurrentDictionary<Type, DataViewType> _rawTypeToDataViewTypeMap;
+        private static ConcurrentDictionary<DataViewType, Type> _dataViewTypeToRawTypeMap;
```
shmoradims: these can all be made 'static readonly' I think.

src/Microsoft.ML.DataView/TypeManager.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        /// <summary>
+        /// Constructor to initialize type mappings.
+        /// </summary>
+        static TypeManager()
```
shmoradims: >TypeManager [](start = 15, length = 11)

IMO 'Manager' is too broad. This type is only managing mappings, and it's just a dictionary with special logic around adding keys. So the following changes might make things clearer:

1) Rename it to DataViewTypeMappings
2) Rename Register() to AddMapping() or Add()
3) Change GetDataViewTypeOrNull() and the other one to indexers, so the code changes to DataViewTypeMappings[someType]

src/Microsoft.ML.DataView/TypeManager.cs
shmoradims: +1

---
In reply to: [284939889](https://github.com/dotnet/machinelearning/pull/3263#discussion_r284939889) [](ancestors = 284939889)

src/Microsoft.ML.DataView/TypeManager.cs
shmoradims: >Register [](start = 27, length = 8)

please add basic unit tests for this to test all the paths in the functions

src/Microsoft.ML.Data/Utils/ApiUtils.cs
```diff
@@ -23,7 +23,8 @@ private static OpCode GetAssignmentOpCode(Type t)
...
+                t == typeof(DataViewRowId) || TypeManager.GetDataViewTypeOrNull(t) != null)
```
shmoradims: >GetDataViewTypeOrNull [](start = 58, length = 21)

let's just add a Contains() method instead of checking for null.

src/Microsoft.ML.Data/DataView/InternalSchemaDefinition.cs
shmoradims: >else if (!itemType.TryGetDataKind(out _) && TypeManager.GetDataViewTypeOrNull(itemType) == null) [](start = 12, length = 96)

what would happen if
!itemType.TryGetDataKind(out _) && TypeManager.GetDataViewTypeOrNull(itemType) != null

test/Microsoft.ML.Tests/ImagesTests.cs
```diff
@@ -184,6 +184,78 @@ public void TestGreyscaleTransformImages()
...
+
+            var imagesFolder = Path.GetDirectoryName(imagesDataFile);
+            // Image loading and conversion pipeline.
+            var pipeline = ML.Transforms.LoadImages("ImageObject", imagesFolder, "ImagePath")
+                           .Append(ML.Transforms.ConvertToGrayscale("Grayscale", "ImageObject"));
+
+            // Test path: image files -> IDataView -> Enumerable of Bitmaps.
+            var transformedData = pipeline.Fit(data).Transform(data);
+            var transformedDataPoints = ML.Data.CreateEnumerable<TransformedImageDataPoint>(transformedData, true).ToList();
```
shmoradims: >true [](start = 109, length = 4)

you're realizing the enumerable as a list, so reuse option should be false, otherwise your list will end up containing the same image (the last one).

src/Microsoft.ML.Data/Utils/ApiUtils.cs
wschin: Ok but will call it `Knows(...)`.

---
In reply to: [285269922](https://github.com/dotnet/machinelearning/pull/3263#discussion_r285269922) [](ancestors = 285269922)

src/Microsoft.ML.DataView/TypeManager.cs
wschin: This will be removed and the initializers of static fields will be used.

---
In reply to: [285262825](https://github.com/dotnet/machinelearning/pull/3263#discussion_r285262825) [](ancestors = 285262825)

src/Microsoft.ML.DataView/TypeManager.cs
wschin: Sure.

---
In reply to: [284944816](https://github.com/dotnet/machinelearning/pull/3263#discussion_r284944816) [](ancestors = 284944816)

src/Microsoft.ML.DataView/TypeManager.cs
wschin: `DataViewTypeMappings` looks even more strange to me... This is type register and this is how ONNXMLTools being used.

---
In reply to: [285266926](https://github.com/dotnet/machinelearning/pull/3263#discussion_r285266926) [](ancestors = 285266926)

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
wschin: Just add `Knows(...)`.

---
In reply to: [285260595](https://github.com/dotnet/machinelearning/pull/3263#discussion_r285260595) [](ancestors = 285260595,285245746,284945327)

src/Microsoft.ML.ImageAnalytics/ImageType.cs
```diff
@@ -13,6 +13,12 @@ public sealed class ImageDataViewType : StructuredDataViewType
     {
         public readonly int Height;
         public readonly int Width;
+
+        static ImageDataViewType()
+        {
+            DataViewTypeManager.Register(typeof(Bitmap), new ImageDataViewType());
```
eerhardt: Is it OK to not have height and width associated with this `ImageDataViewType`? Should we support setting those through an attribute?

src/Microsoft.ML.DataView/DataViewType.cs
```diff
@@ -21,7 +21,7 @@ public abstract class DataViewType : IEquatable<DataViewType>
         /// <summary>
         /// Constructor for extension types, which must be either <see cref="PrimitiveDataViewType"/> or <see cref="StructuredDataViewType"/>.
         /// </summary>
-        private protected DataViewType(Type rawType)
+        protected DataViewType(Type rawType)
```
eerhardt: This is intentional. It shouldn't be changed. See the summary comment.

src/Microsoft.ML.DataView/TypeManager.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            typeof(Boolean), typeof(SByte), typeof(Byte),
+            typeof(Int16), typeof(UInt16), typeof(Int32), typeof(UInt32),
+            typeof(Int64), typeof(UInt64), typeof(Single), typeof(Double),
+            typeof(string), typeof(ReadOnlySpan<char>)
+        };
+
+        private static Dictionary<Type, DataViewType> _rawTypeToDataViewTypeMap = new Dictionary<Type, DataViewType>();
+        private static Dictionary<DataViewType, Type> _dataViewTypeToRawTypeMap = new Dictionary<DataViewType, Type>();
+        private static SpinLock _lock = new SpinLock();
```
eerhardt: Why use a SpinLock, and not just a regular `lock`?

wschin: From .NET doc,
```
A mutual exclusion lock primitive that causes the thread that is trying to acquire the lock to wait in a loop, or spin, for a period of time before yielding its quantum. In scenarios where the wait for the lock is expected to be short, SpinLock offers better performance than other forms of locking. For more information, see SpinLock.
```
Because we only lock for a few instructions, I feel `SpinLock` can provide better performance.

---
In reply to: [285626445](https://github.com/dotnet/machinelearning/pull/3263#discussion_r285626445) [](ancestors = 285626445)

src/Microsoft.ML.Data/DataView/InternalSchemaDefinition.cs
wschin: It will just not throw any error. Let me rewrite the `else if` branch here, which is only for checking if the passed-in type is known.

---
In reply to: [285271321](https://github.com/dotnet/machinelearning/pull/3263#discussion_r285271321) [](ancestors = 285271321)

src/Microsoft.ML.DataView/TypeManager.cs
eerhardt: My suggestion would be to gather data to support the `I feel` statement above. See the answer from Stephen Toub here:

https://social.msdn.microsoft.com/Forums/en-US/b5b8373a-6d56-427f-b90b-1f67f78010c4/spinlock-usageperformance

src/Microsoft.ML.Data/DataView/DataViewConstructionUtils.cs
wschin: It can be anything you want. The generic type will be over-write at the end of this function.

---
In reply to: [284945609](https://github.com/dotnet/machinelearning/pull/3263#discussion_r284945609) [](ancestors = 284945609)

src/Microsoft.ML.DataView/DataViewType.cs
wschin: Ok. Will derive custom types from `StructuredDataViewType`.

---
In reply to: [285623815](https://github.com/dotnet/machinelearning/pull/3263#discussion_r285623815) [](ancestors = 285623815)

src/Microsoft.ML.DataView/TypeManager.cs
wschin: Let me just use `lock`. This is definitely not a computation  bottleneck and `lock` can simplify my code a lot.

---
In reply to: [285669394](https://github.com/dotnet/machinelearning/pull/3263#discussion_r285669394) [](ancestors = 285669394)

src/Microsoft.ML.ImageAnalytics/ImageType.cs
wschin: The entire image system in ML.NET is not strongly typed. I think to fix this, we need at least one week and it won't be BC.

---
In reply to: [285623486](https://github.com/dotnet/machinelearning/pull/3263#discussion_r285623486) [](ancestors = 285623486)

src/Microsoft.ML.Data/Utils/ApiUtils.cs
wschin: I don't want to mix users types with ML.NET's native types.

---
In reply to: [284939178](https://github.com/dotnet/machinelearning/pull/3263#discussion_r284939178) [](ancestors = 284939178)

src/Microsoft.ML.Data/DataView/InternalSchemaDefinition.cs
```diff
@@ -147,11 +147,11 @@ public static void GetVectorAndItemType(MemberInfo memberInfo, out bool isVector
             switch (memberInfo)
             {
                 case FieldInfo fieldInfo:
-                    GetVectorAndItemType(fieldInfo.FieldType, fieldInfo.Name, out isVector, out itemType);
+                    GetVectorAndItemType(fieldInfo.FieldType, fieldInfo.Name, out isVector, out itemType, fieldInfo.GetCustomAttributes().ToArray());
```
eerhardt: GetCustomAttributes already returns an array - why the need to call `ToArray()` again?

src/Microsoft.ML.ImageAnalytics/ImageType.cs
```diff
@@ -2,23 +2,72 @@
...
+        /// </summary>
+        public ImageTypeAttribute(int height, int width)
+        {
+            Contracts.CheckParam(width > 0, nameof(width), "Should be positive number");
+            Contracts.CheckParam(height > 0, nameof(height), "Should be positive number");
+            Height = height;
+            Width = width;
+        }
+
+        public override int GetHashCode()
```
eerhardt: If you override GetHashCode, you will also want to override Equals.

src/Microsoft.ML.ImageAnalytics/ImageType.cs
```diff
@@ -2,23 +2,72 @@
...
 using System.Drawing;
 using Microsoft.ML.Data;
 using Microsoft.ML.Internal.Utilities;
 using Microsoft.ML.Runtime;
 
 namespace Microsoft.ML.Transforms.Image
 {
+    /// <summary>
+    /// Allows a member to be marked as a <see cref="VectorDataViewType"/>, primarily allowing one to set
+    /// the dimensionality of the resulting array.
```
eerhardt: ` resulting array` what resulting array?

test/Microsoft.ML.Tests/ImagesTests.cs
```diff
@@ -184,6 +184,94 @@ public void TestGreyscaleTransformImages()
...
+        {
+            [ImageType(10, 10)]
+            public Bitmap Image { get; set; }
+
+            [ImageType(10, 10)]
+            public Bitmap GrayImage { get; set; }
+
+            static ImageDataPoint()
+            {
+                DataViewTypeManager.Register(new ImageDataViewType(10, 10), typeof(Bitmap), new ImageTypeAttribute(10, 10));
```
eerhardt: Is it really up to the user to register each possible value? It seems like we should be able to create an `ImageDataViewType` instance from any `ImageTypeAttribute`, without the user having to register each size.

wschin: Do you mean calling register function in a constructor of `ImageTypeAttribute`?

src/Microsoft.ML.Data/DataView/InternalSchemaDefinition.cs
wschin: It's not an array. It's an IEnumerable. Anyway, I change `params` to array.

---
In reply to: [286691422](https://github.com/dotnet/machinelearning/pull/3263#discussion_r286691422) [](ancestors = 286691422)

src/Microsoft.ML.ImageAnalytics/ImageType.cs
wschin: Sorry for making copy-and-paste..

---
In reply to: [286693530](https://github.com/dotnet/machinelearning/pull/3263#discussion_r286693530) [](ancestors = 286693530)

test/Microsoft.ML.Tests/ImagesTests.cs
wschin: If user don't want to specify shapes, they can use `unknown` size image.

---
In reply to: [286694386](https://github.com/dotnet/machinelearning/pull/3263#discussion_r286694386) [](ancestors = 286694386)

src/Microsoft.ML.ImageAnalytics/ImageType.cs
```diff
@@ -2,23 +2,90 @@
...
+            DataViewTypeManager.Register(new ImageDataViewType(Height, Width), typeof(Bitmap), new[] { this });
+        }
+    }
+
     public sealed class ImageDataViewType : StructuredDataViewType
     {
         public readonly int Height;
         public readonly int Width;
+
+        static ImageDataViewType()
```
eerhardt: This is no longer necessary.

src/Microsoft.ML.DataView/TypeManager.cs
```diff
@@ -0,0 +1,189 @@
...
+using Microsoft.ML.Internal.DataView;
+
+namespace Microsoft.ML.Data
+{
+    /// <summary>
+    /// A singleton class for managing the map between ML.NET <see cref="DataViewType"/> and C# <see cref="Type"/>.
+    /// To support custom column type in <see cref="IDataView"/>, the column's underlying type (e.g., a C# class's type)
+    /// should be registered with a class derived from <see cref="DataViewType"/>.
+    /// </summary>
+    public static class DataViewTypeManager
```
eerhardt: Can you keep the file name and the class name sync'd? The file should be named `DataViewTypeManager.cs`.

src/Microsoft.ML.DataView/DataViewType.cs
```diff
@@ -461,4 +461,16 @@ public override bool Equals(DataViewType other)
 
         public override string ToString() => "TimeSpan";
     }
+
+    public abstract class DataViewTypeAttribute : Attribute
```
eerhardt: all public classes should have XML summary docs on them.

src/Microsoft.ML.DataView/DataViewType.cs
```diff
@@ -461,4 +461,16 @@ public override bool Equals(DataViewType other)
...
+    {
+        protected DataViewTypeAttribute() : base()
+        {
+        }
+
+        /// <summary>
+        /// A function implicitly invoked by ML.NET when processing a custom type. It binds a DataViewType to a custome type plus its attributes.
```
eerhardt: spelling `custome`

src/Microsoft.ML.Data/Utils/ApiUtils.cs
```diff
@@ -16,14 +17,15 @@ namespace Microsoft.ML
 
     internal static class ApiUtils
     {
-        private static OpCode GetAssignmentOpCode(Type t)
+        private static OpCode GetAssignmentOpCode(Type t, params Attribute[] attributes)
```
eerhardt: This doesn't need to be an array anymore - it can be an IEnumerable - so the upstream callers don't need to call `ToArray`

src/Microsoft.ML.Data/DataView/InternalSchemaDefinition.cs
```diff
@@ -185,9 +186,13 @@ public static void GetVectorAndItemType(Type rawType, string name, out bool isVe
...
             }
 
+            // The internal type of string is ReadOnlyMemory<char>. That is, string will be stored as ReadOnlyMemory<char> in IDataView.
             if (itemType == typeof(string))
                 itemType = typeof(ReadOnlyMemory<char>);
-            else if (!itemType.TryGetDataKind(out _))
+
+            // Check if the itemType extracted from rawType is supported by ML.NET's type system.
+            // It must be one of either ML.NET's pre-defined types or custom types registered by the user.
+            if (!itemType.TryGetDataKind(out _) && !DataViewTypeManager.Knows(itemType, attributes))
```
eerhardt: Why did you remove the `else` here? We don't need to call this for `string`.

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -392,37 +402,42 @@ public static SchemaDefinition Create(Type userType, Direction direction = Direc
...
-                else
-                    itemType = ColumnTypeExtensions.PrimitiveTypeFromType(dataType);
-
                 // Get the column type.
                 DataViewType columnType;
-                var vectorAttr = memberInfo.GetCustomAttribute<VectorTypeAttribute>();
-                if (vectorAttr != null && !isVector)
-                    throw Contracts.ExceptParam(nameof(userType), $"Member {memberInfo.Name} marked with {nameof(VectorTypeAttribute)}, but does not appear to be a vector type", memberInfo.Name);
-                if (isVector)
+                if (!DataViewTypeManager.Knows(dataType, memberInfo.GetCustomAttributes()))
```
eerhardt: We are calling `memberInfo.GetCustomAttributes()` multiple times in this method. We should probably cache it in a local variable.

src/Microsoft.ML.DataView/TypeManager.cs
```diff
@@ -0,0 +1,189 @@
...
+        /// </summary>
+        public static DataViewType GetDataViewType(Type rawType, IEnumerable<Attribute> rawTypeAttributes = null)
+        {
+            // Overall flow:
+            //   type (Type) + attrs ----> type ID ----------------> associated DataViewType's ID ----------------> DataViewType
+            //                     (hashing)      (dictionary look-up)                           (dictionary look-up)
+            lock (_lock)
+            {
+                // Compute the ID of type with extra attributes.
+                var typeId = ComputeHashCode(rawType, rawTypeAttributes);
```
eerhardt: HashCodes aren't necessarily unique, we shouldn't be using them as keys in a dictionary.

Instead, we should have a "Key" type that implements GetHashCode and Equals, and we should be putting the "Key" types into the dictionary, not an integer.

wschin: Just pushed iteration 15 for addressing this comment.

---
In reply to: [287034735](https://github.com/dotnet/machinelearning/pull/3263#discussion_r287034735) [](ancestors = 287034735)

src/Microsoft.ML.Data/DataView/InternalSchemaDefinition.cs
wschin: The semantic here was not quite consistent. The string branch is to replace a type. However the `else if` was for checking if the type is allowed.

---
In reply to: [287032519](https://github.com/dotnet/machinelearning/pull/3263#discussion_r287032519) [](ancestors = 287032519)

src/Microsoft.ML.DataView/DataViewType.cs
```diff
@@ -37,11 +37,25 @@ private protected DataViewType(Type rawType)
...
+        /// <summary>
+        /// Return <see langword="true"/> if <see langword="this"/> is equivalent to <paramref name="other"/> and <see langword="false"/> otherwise.
+        /// </summary>
+        /// <param name="other">Another <see cref="DataViewType"/> to be compared with <see langword="this"/>.</param>
         public abstract bool Equals(DataViewType other);
+
+        /// <summary>
+        /// Produce the hashing code of <see langword="this"/>. It's the implementation of <see cref="GetHashCode()"/>.
+        /// </summary>
+        public abstract int GetDataViewTypeHashCode();
```
eerhardt: This is actually a breaking change and shouldn't be checked in. You can't add new `abstract` methods to classes that have already shipped.

src/Microsoft.ML.DataView/DataViewType.cs
```diff
@@ -461,4 +480,48 @@ public override bool Equals(DataViewType other)
...
+        /// <summary>
+        /// A function implicitly invoked by ML.NET when processing a custom type. It binds a DataViewType to a custom type plus its attributes.
+        /// </summary>
+        public abstract void Register();
+
+        /// <summary>
+        /// Return <see langword="true"/> if <see langword="this"/> is equivalent to <paramref name="other"/> and <see langword="false"/> otherwise.
+        /// </summary>
+        /// <param name="other">Another <see cref="DataViewType"/> to be compared with <see langword="this"/>.</param>
+        public abstract bool Equals(DataViewTypeAttribute other);
```
eerhardt: This pattern isn't a typical .NET pattern. I think the derived classes should just override `Equals` and `GetHashCode`, just like any typical .NET class.

src/Microsoft.ML.DataView/DataViewType.cs
```diff
@@ -461,4 +480,48 @@ public override bool Equals(DataViewType other)
...
         public override string ToString() => "TimeSpan";
     }
+
+    /// <summary>
+    /// <see cref="DataViewTypeAttribute"/> should be used to decorated class properties and fields, if that class' instances will be loaded as ML.NET <see cref="IDataView"/>.
+    /// The function <see cref="Register"/> will be called to register a <see cref="DataViewType"/> for a <see cref="Type"/> with its <see cref="Attribute"/>s.
+    /// Whenever a value typed to the registered <see cref="Type"/> and its <see cref="Attribute"/>s, its <see cref="DataViewSchema.Column.Type"/> would be the
+    /// associated <see cref="DataViewType"/>.
+    /// </summary>
+    public abstract class DataViewTypeAttribute : Attribute
```
eerhardt: This class could implement `IEquatable<DataViewTypeAttribute>`, just like `DataViewType` does.

src/Microsoft.ML.DataView/DataViewTypeManager.cs
```diff
@@ -0,0 +1,274 @@
...
+            public DataViewType TargetType { get; }
+
+            public DataViewTypeId(DataViewType type)
+            {
+                TargetType = type;
+            }
+
+            public override bool Equals(object obj)
+            {
+                if (obj is DataViewTypeId)
```
eerhardt: You can use pattern matching here:

```C#
if (obj is DataViewTypeId other)
{
  return TargetType.Equals(other.TargetType);
}
```

src/Microsoft.ML.DataView/DataViewTypeManager.cs
```diff
@@ -0,0 +1,274 @@
...
+
+                    if (AssociatedAttributes == null && other.AssociatedAttributes == null)
+                        sameAttributeConfig = true;
+                    else if (AssociatedAttributes == null && other.AssociatedAttributes != null)
+                        sameAttributeConfig = false;
+                    else if (AssociatedAttributes != null && other.AssociatedAttributes == null)
+                        sameAttributeConfig = false;
+                    else
+                    {
+                        var zipped = AssociatedAttributes.Zip(other.AssociatedAttributes, (attr, otherAttr) => (attr, otherAttr));
```
eerhardt: We should check that `AssociatedAttributes` are the same length as well, right?

src/Microsoft.ML.DataView/DataViewType.cs
eerhardt: @artidoro @ericstj - this should be caught by ApiCompat, correct? Is that working in this repo?

src/Microsoft.ML.Data/Utils/ApiUtils.cs
```diff
@@ -56,7 +59,7 @@ private static OpCode GetAssignmentOpCode(Type t)
                 case FieldInfo fieldInfo:
                     Type fieldType = fieldInfo.FieldType;
 
-                    var assignmentOpCode = GetAssignmentOpCode(fieldType);
+                    var assignmentOpCode = GetAssignmentOpCode(fieldType, fieldInfo.GetCustomAttributes().ToArray());
```
eerhardt: The `ToArray()` calls are no longer needed.

src/Microsoft.ML.Data/DataView/InternalSchemaDefinition.cs
eerhardt: I don't understand your response. Why would we need to check if `string` is allowed?  Of course it is. If `itemType == typeof(string)`, replace it to `ROM<char>` and no need to check if it if valid, because we know it is valid.

---
In reply to: [287136944](https://github.com/dotnet/machinelearning/pull/3263#discussion_r287136944) [](ancestors = 287136944,287032519)

wschin: That's fine. I will change it back.

---
In reply to: [287165481](https://github.com/dotnet/machinelearning/pull/3263#discussion_r287165481) [](ancestors = 287165481,287136944,287032519)

src/Microsoft.ML.DataView/DataViewTypeManager.cs
wschin: You're right.

---
In reply to: [287164576](https://github.com/dotnet/machinelearning/pull/3263#discussion_r287164576) [](ancestors = 287164576)

src/Microsoft.ML.DataView/DataViewType.cs
ericstj: Should be https://github.com/dotnet/arcade/blob/767154ac5bb8715ab1357c63f346244de6d4aa6b/src/Microsoft.DotNet.ApiCompat/Rules/Compat/CannotAddAbstractMembers.cs

I can’t see the previous build result, did APICompat run?

eerhardt: Yes, ApiCompat does run, I tested it locally.

This appears to be a bug in the rule.  Here is the scenario:

# V1
```C#
public abstract class InternalConstructorBase
{
    private protected InternalConstructorBase()
    { }
}

public abstract class IntermediateBase
{
    protected InternalConstructorBase()
    { }
}
```

# V2 - add an abstract member to `InternalConstructorBase`
```C#
public abstract class InternalConstructorBase
{
    private protected InternalConstructorBase()
    { }

    public abstract void NewMember();
}

public abstract class IntermediateBase
{
    protected InternalConstructorBase()
    { }
}
```

Here, `InternalConstructorBase` is "effectively sealed" so it sneaks by the rule in 

https://github.com/dotnet/arcade/blob/767154ac5bb8715ab1357c63f346244de6d4aa6b/src/Microsoft.DotNet.ApiCompat/Rules/Compat/CannotAddAbstractMembers.cs#L26-L27

But, `IntermediateBase` now has a new `abstract` member, from its base class.

I've opened https://github.com/dotnet/arcade/issues/2885 for it.

src/Microsoft.ML.DataView/DataViewTypeManager.cs
```diff
@@ -0,0 +1,247 @@
...
+                    code = Hashing.CombineHash(code, attr.GetHashCode());
+                return code;
+            }
+
+        }
+
+        /// <summary>
+        /// An instance of <see cref="DataViewTypeId"/> represents an unique key of its <see cref="TargetType"/>.
+        /// </summary>
+        private class DataViewTypeId
```
eerhardt: Is this class necessary? It just wraps `DataViewType` and relies on its underlying Equals and GetHashCode. I think instead, we could just use `DataViewType` as the key to the dictionary.

wschin: Nice catch. I will remove it.

src/Microsoft.ML.DataView/DataViewTypeManager.cs
```diff
@@ -0,0 +1,226 @@
...
+            typeof(Boolean), typeof(SByte), typeof(Byte),
+            typeof(Int16), typeof(UInt16), typeof(Int32), typeof(UInt32),
+            typeof(Int64), typeof(UInt64), typeof(Single), typeof(Double),
+            typeof(string), typeof(ReadOnlySpan<char>), typeof(ReadOnlyMemory<char>),
+            typeof(VBuffer<>), typeof(Nullable<>), typeof(DateTime), typeof(DateTimeOffset),
+            typeof(TimeSpan), typeof(DataViewRowId)
+        };
+
+        /// <summary>
+        /// Mapping from hashing ID of a <see cref="Type"/> and its <see cref="Attribute"/>s to hashing ID of a <see cref="DataViewType"/>.
```
wschin:
```suggestion
        /// Mapping from a <see cref="Type"/> plus its <see cref="Attribute"/>s to a <see cref="DataViewType"/>.
```

src/Microsoft.ML.DataView/DataViewTypeManager.cs
```diff
@@ -0,0 +1,226 @@
...
+            typeof(TimeSpan), typeof(DataViewRowId)
+        };
+
+        /// <summary>
+        /// Mapping from a <see cref="Type"/> plus its <see cref="Attribute"/>s to a <see cref="DataViewType"/>.
+        /// </summary>
+        private static Dictionary<TypeWithAttributes, DataViewType> _rawTypeToDataViewTypeMap = new Dictionary<TypeWithAttributes, DataViewType>();
+
+        /// <summary>
+        /// Mapping from hashing ID of a <see cref="DataViewType"/> to hashing ID of a <see cref="Type"/> and its <see cref="Attribute"/>s.
```
wschin:
```suggestion
        /// Mapping from a <see cref="DataViewType"/> to a <see cref="Type"/> plus its <see cref="Attribute"/>s.
```

artidoro:
The PR looks great, but we will need a sample for this work. It's fine if it comes in another PR. Could you adapt a test and make it a sample?

artidoro:
Also would it be useful to surface the `DataViewTypeManager` in `MLContext`? Or some of it's methods?

wschin:
For this advanced functionality, I'd keep it as is for now. If we add an alias for it, we will have to support it for years.

---
In reply to: [496640481](https://github.com/dotnet/machinelearning/pull/3263#issuecomment-496640481) [](ancestors = 496640481)

wschin:
Added two samples.

---
In reply to: [496639086](https://github.com/dotnet/machinelearning/pull/3263#issuecomment-496639086) [](ancestors = 496639086)

ADD-eNavarro:
Hi! 
Is there a date for this to make it into ML.NET? I really could use this feature, but it seems a little stuck.
Thanks!

johncalvowtg:
Hello!
We have actually the same problem here at Wisetech. I am trying to consume a TensorFlow model with multiple multi-dimensional inputs . Same message and no workaround on this.

I've tried some of your examples, but none of them worked for me.

Please let us know if you manage to fully support TensorFlow integration with ML.NET.

Regards,
John Calvo-Martinez Ph.D.

jamsoft:
Phew, it's not just me going mad. I've been trying to use in memory images all day and have so far failed to get any of this to work. I have  [SO question](https://stackoverflow.com/questions/66515905/loading-existing-ml-net-image-classification-model-and-adapting-predict-using-in) open as well in case anyone wants to follow.

michaelgsharp:
@Webreaper I thought we had a sample or documentation about it, but I just went and looked and couldn't find anything myself. I'll make sure we have something. The basic idea is that you have 2 pipelines. 1 pipeline loads the images and stops when it has the images loaded as bytes, and the second pipeline starts from where you would use these image bytes already.

I can' upload a zip file here since its not a github issue, but I do have a sample interactive notebook that shows it. If you can create a new issue I can get attach it for you there.

PR #3267, opened 2019-04-09
Remove Microsoft.ML prefix from namespaces in samples binary.

### LINKED ISSUE #3266: Remove Microsoft.ML prefix from namespaces in Samples binary
This forces the samples to include Microsoft.ML namespace.

codemzs:
fixes #3266


rogancarr:
This PR didn't update SampleUtils, so the samples that use it won't run :)

PR #3287, opened 2019-04-11
Fix runtime exception in MapKeyToVector sample.

### LINKED ISSUE #3286: MapKeyToValue sample throws exception
Seems like the issue is cast exception between vector of float to vector of uint.

codemzs:
fixes #3286

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/Conversion/MapKeyToVector.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/Conversion/MapKeyToVector.cs
@@ -67,8 +67,8 @@ namespace Samples.Dynamic
 
         private class TransformedData : DataPoint
         {
-            public uint[] TimeframeVector { get; set; }
-            public uint[] CategoryVector { get; set; }
+            public float[] TimeframeVector { get; set; }
+            public float[] CategoryVector { get; set; }
         }
     }
 }


docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/Conversion/MapKeyToVector.cs
```diff
@@ -67,8 +67,8 @@ private class DataPoint
 
         private class TransformedData : DataPoint
         {
-            public uint[] TimeframeVector { get; set; }
-            public uint[] CategoryVector { get; set; }
+            public float[] TimeframeVector { get; set; }
```
codemzs: the other option is leave it as uint[] but add another transform in the pipeline that converts these column types from float to uint but this is not currently possibly because we don't have float to uint converter in Conversions.cs but of course that can be added and we can use C#'s convention for conversion to uint by casting.

CC: @TomFinley

TomFinley: I think changing the type is appropriate -- this is a key-to-vector sample, the output type of key-to-vector is a float vector, so, this change seems appropriate at least to me.

codemzs: Thanks @TomFinley

PR #3295, opened 2019-04-11
Fix runtime exception in prior trainer sample and update the comments.

### LINKED ISSUE #3294: Prior Trainer sample throws an exception and the output comments do not match the actual output
Prior trainer used to take label type of float but this was changed to take label type of boolean to make it consistent with other binary trainers. Update the sample to change the schema of the data loader to accept label type of boolean and also update the comments to be consistent with the actual output from ML.NET

codemzs:
fixes #3294

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/BinaryClassification/PriorTrainerSample.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/BinaryClassification/PriorTrainerSample.cs
@@ -26,7 +26,7 @@ namespace Samples.Dynamic
             var loader = mlContext.Data.CreateTextLoader(
                 columns: new[]
                     {
-                        new TextLoader.Column("Sentiment", DataKind.Single, 0),
+                        new TextLoader.Column("Sentiment", DataKind.Boolean, 0),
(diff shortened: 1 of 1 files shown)


docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/BinaryClassification/PriorTrainerSample.cs
```diff
@@ -63,7 +63,7 @@ public static void Example()
             //  Positive Precision: 0.50
             //  Positive Recall: 1.00
             //  LogLoss: 1.05
-            //  LogLossReduction: -4.89
```
wschin: Why does it change the baseline?

codemzs: It didn't the documentation was stable. I verified this without the prior trainer change from the master branch (removed the change that changes label type to boolean) and below is the output:

Accuracy: 0.50
AUC: 0.50
F1 Score: 0.67
Negative Precision: 0.00
Negative Recall: 0.00
Positive Precision: 0.50
Positive Recall: 1.00
Log Loss: 1.05
Log Loss Reduction: -0.05
Entropy: 1.00

PR #3268, opened 2019-04-09
 Updated CopyColumns, DropColumns and SelectColumns samples.

zeahmed:
Related to #1209.

PR #3262, opened 2019-04-09
Updated sample for Concatenate API.

zeahmed:
Related to #1209.

PR #3300, opened 2019-04-11
Cherry pick sample update (Concate, Select, Drop, Copy)

zeahmed:
This PR cherry picks the following commits from the master related to Concatenate, Select, Drop and Copy transforms.

#3262
#3268


shauheen:
Done as part of #3333 . Thanks @zeahmed .

Issue #3205, opened 2019-04-04
API reference - Remove Microsoft.ML prefix from samples' namespace


shmoradims:
Currently the samples are written like this:

```csharp
namespace **Microsoft.ML**.Samples.Dynamic.*
{ 
}
```
This is causing the samples to not require `using Microsoft.ML`. We want to drop the Microsoft.ML prefix from samples and write them like this:

```csharp
namespace Samples.Dynamic.*
{ 
}
```

codemzs:
I fixed for all the samples and checked-in my change.

PR #3302, opened 2019-04-11
Simple IDataView implementation sample.

### LINKED ISSUE #3301: Simple IDataView Sample
Though we have many methods or samples to create data views from existing structures, like files, databases, enumerables, lists, and so on, we can't possibly handle everything, so a sample as to how to create a data view may be desirable.

In this first simple sample, my goal would be to show a fairly minimal `IDataView` implementation but from absolute scratch, one that is very simple without being, I hope, simplistic. I might however stick to scalar inputs for that first simple example, and definitely avoid anything having to do with enabling shuffling or parallel cursor evaluation.

Related to #1209, but given that this does not describe a sample for anything *directly* related to `MLContext`, it was thought best to have a separate issue for it.

TomFinley:
Fixes #3301.

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
@@ -1,0 +1,258 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.ML;
+using Microsoft.ML.Data;
+
+namespace Samples.Dynamic
+{
+    /// <summary>
+    /// The <see cref="IDataView"/> interface is the central concept of "data" in ML.NET. While many conveniences exist
+    /// to create pre-baked implementations, it is also useful to know how to create one completely from scratch. We also
+    /// take this opportunity to illustrate and motivate the basic principles of how the IDataView system is architected,
+    /// since people interested in implementing <see cref="IDataView"/> need at least some knowledge of those principles.
+    /// </summary>
+    public static class SimpleDataViewImplementation
+    {
+        public static void Example()
+        {
+            // First we create an array of these objects, which we "present" as this IDataView implementation so that it
+            // can be used in a simple ML.NET pipeline.
+            var inputArray = new[]
+            {
+                new InputObject(false, "Hello my friend."),
+                new InputObject(true, "Stay awhile and listen."),
+                new InputObject(true, "Masterfully done hero!")
+            };
+            var dataView = new InputObjectDataView(inputArray);
+
+            // So, this is a very simple pipeline: a transformer that tokenizes Text, does nothing with the Label column
+            // at all.
+            var mlContext = new MLContext();
+            var transformedDataView = mlContext.Transforms.Text.TokenizeIntoWords(
+                "TokenizedText", "Text").Fit(dataView).Transform(dataView);
+
+            var textColumn = transformedDataView.Schema["Text"];
+            var tokensColumn = transformedDataView.Schema["TokenizedText"];
+
+            using (var cursor = transformedDataView.GetRowCursor(new[] { textColumn, tokensColumn }))
+            {
+                // Note that it is best to get the getters and values *before* iteration, so as to faciliate buffer
+                // sharing (if applicable), and column-type validation once, rather than many times.
+                ReadOnlyMemory<char> textValue = default;
+                VBuffer<ReadOnlyMemory<char>> tokensValue = default;
+
+                var textGetter = cursor.GetGetter<ReadOnlyMemory<char>>(textColumn);
+                var tokensGetter = cursor.GetGetter<VBuffer<ReadOnlyMemory<char>>>(tokensColumn);
+
+                while (cursor.MoveNext())
+                {
+                    textGetter(ref textValue);
+                    tokensGetter(ref tokensValue);
+
+                    Console.WriteLine($"{textValue} => {string.Join(", ", tokensValue.DenseValues())}");
+                }
+
+                // The output to console is this:
+
+                // Hello my friend. => Hello, my, friend.
+                // Stay awhile and listen. => Stay, awhile, and, listen.
+                // Masterfully done hero! => Masterfully, done, hero!
+
+                // Note that it may be interesting to set a breakpoint on the Console.WriteLine, and explore
+                // what is going on with the cursor, and the buffers. In particular, on the third iteration,
+                // while `tokensValue` is logically presented as a three element array, internally you will
+                // see that the arrays internal to that structure have (at least) four items, specifically:
+                // `Masterfully`, `done`, `hero!`, `listen.`. In this way we see a simple example of the details
+                // of how buffer sharing from one iteration to the next actually works.
+            }
+        }
+
+        private sealed class InputObject
+        {
+            public bool Label { get; }
+            public string Text { get; }
+
+            public InputObject(bool label, string text)
+            {
+                Label = label;
+                Text = text;
+            }
+        }
+
+        /// <summary>
+        /// This is an implementation of <see cref="IDataView"/> that wraps an <see cref="IEnumerable{T}"/>
+        /// of the above <see cref="InputObject"/>. Note that normally under these circumstances, the first
+        /// recommendation would be to use a convenience like
+        /// <see cref="DataOperationsCatalog.LoadFromEnumerable{TRow}(IEnumerable{TRow}, SchemaDefinition)"/>
+        /// or something like that, rather than implementing <see cref="IDataView"/> outright. However, sometimes when
+        /// code generation is impossible on some situations, like Unity or other similar platforms, implementing
+        /// something even closely resembling this may become necessary.
+        ///
+        /// This implementation of <see cref="IDataView"/> is, in fact, much simpler than practically anything you would
+        /// find in the ML.NET codebase. In this case we have a completely fixed schema (the two fields of
+        /// <see cref="InputObject"/>), with fixed types. However, for the purpose of the sample, we keep things as
+        /// simple as possible, but without being simplistic, I hope.
+        ///
+        /// For <see cref="Schema"/>, note that we keep a very simple schema based off the members of the object. You
+        /// may in fact note that it is possible in this specific case, this implementation of
+        /// <see cref="IDatView"/> could share the same <see cref="DataViewSchema"/> object across all instances of this
+        /// object, but since this is almost never the case, I do not take advantage of that.
+        ///
+        /// We have chosen to wrap an <see cref="IEnumerable{T}"/>, so in fact only a very simple implementation is
+        /// possible. Specifically: we cannot meaningfully shuffle (so <see cref="CanShuffle"/> is
+        /// <see langword="false"/>, and even if a <see cref="Random"/> parameter were passed to
+        /// <see cref="GetRowCursor(IEnumerable{DataViewSchema.Column}, Random)"/>, we could not make use of it), we do
+        /// not know the count of the item right away without counting (so, it is most correct for
+        /// <see cref="GetRowCount"/> to return <see langword="null"/>, even after we might hypothetically know after
+        /// the first pass, given the immutability principle of <see cref="IDatView"/>), and the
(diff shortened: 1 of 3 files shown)


docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
```diff
@@ -0,0 +1,258 @@
...
+
+            public InputObject(bool label, string text)
+            {
+                Label = label;
+                Text = text;
+            }
+        }
+
+        /// <summary>
+        /// This is an implementation of <see cref="IDataView"/> that wraps an <see cref="IEnumerable{T}"/>
```
TomFinley: In my first attempt at this, I had written much of this material as separate XML docs for each member, but I found that this made it hard to tell a connected "story" about what is going on here, so I stuck to one big block on the `IDataView` implementation and one block on the `DataViewRowCursor` derived class. This made it easier to tell a cohesive story, while also having it I think look a lot less busy.

They are XML comments so I can use `<see` tags, mostly so that someone inspecting this sample can sometimes F12 to see what I'm talking about...

docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
```diff
@@ -5,6 +5,7 @@
     <OutputType>Exe</OutputType>
     <SignAssembly>false</SignAssembly>
     <PublicSign>false</PublicSign>
+    <RootNamespace>Samples</RootNamespace>
```
TomFinley: Thought I might opportunistically do this, since our new policy per #3205 is that samples not have `Microsoft.ML` in their namespace, but *by default* we were still doing the "wrong" thing when we create a new file. I think it might be nice if we avoided the problem by default.

TomFinley:
Not sure offhand who the typical sample reviewers are... @shmoradims and @sfilipi are obvious, but beyond that not entirely certain.

docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// recommendation would be to use a convenience like
+        /// <see cref="DataOperationsCatalog.LoadFromEnumerable{TRow}(IEnumerable{TRow}, SchemaDefinition)"/>
+        /// or something like that, rather than implementing <see cref="IDataView"/> outright. However, sometimes when
+        /// code generation is impossible on some situations, like Unity or other similar platforms, implementing
+        /// something even closely resembling this may become necessary.
+        ///
+        /// This implementation of <see cref="IDataView"/> is, in fact, much simpler than practically anything you would
+        /// find in the ML.NET codebase. In this case we have a completely fixed schema (the two fields of
+        /// <see cref="InputObject"/>), with fixed types. However, for the purpose of the sample, we keep things as
+        /// simple as possible, but without being simplistic, I hope.
```
artidoro: Maybe we can remove this sentence? I am not sure what it's objective is. You already say above that this is simplification of what normally happens in ML.NET.

docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
```diff
@@ -0,0 +1,258 @@
...
+                builder.AddColumn("Text", TextDataViewType.Instance);
+                Schema = builder.ToSchema();
+            }
+
+            public long? GetRowCount() => null;
+
+            public DataViewRowCursor GetRowCursor(IEnumerable<DataViewSchema.Column> columnsNeeded, Random rand = null)
+            {
+                return new Cursor(this, columnsNeeded.Any(c => c.Index == 0), columnsNeeded.Any(c => c.Index == 1));
+            }
```
artidoro: nit: use => instead of return here.

docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
```diff
@@ -0,0 +1,258 @@
...
+
+            public DataViewRowCursor[] GetRowCursorSet(IEnumerable<DataViewSchema.Column> columnsNeeded, int n, Random rand = null)
+                => new[] { GetRowCursor(columnsNeeded, rand) };
+
+            /// <summary>
+            /// Having this be a private sealed nested class follows the typical pattern: in most <see
+            /// cref="IDataView"/>
+            /// implementations, the cursor instance is almost always that. The only "common" exceptions to this
```
artidoro: nit: fix new line.

docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
```diff
@@ -0,0 +1,258 @@
...
+            /// tendency are those implementations that are such thin wrappings of existing <see cref="IDataView"/>
+            /// without even bothering to change the schema.
+            ///
+            /// On the subject of schema, note that there is an expectation that the <see cref="Schema"/> object is
+            /// reference equal to the <see cref="IDataView.Schema"/> object that created this cursor, as we see here.
+            ///
+            /// Note that <see cref="Batch"/> returns <c>0</c>. As described in the documentation of that property, that
+            /// is meant to facilitate the reconciliation of the partitioning of the data in the case where multiple
+            /// cursors are returned from
+            /// <see cref="GetRowCursorSet(IEnumerable{DataViewSchema.Column}, int, Random)"/>, but since only one is
```
artidoro: nit: fix new line.

artidoro:
Is this something that we want to link in our API docs? Or should it be somewhere else in the ML.NET documentation website?
I am thinking it could be nice to have this as an example under the `IDataView` interface, but maybe there are better places or ways to include it in the documentation.

docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
TomFinley: Sigh... @artidoro do you happen to know of any tool that does line breaking well without doing *this* stuff?

docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
TomFinley: Well... I am saying it's simpler than anything you'll find in the codebase, but I think I'm trying to say that's for the sake of the example. I guess I could compress it by mentioning in the first sentence that it is didactic, then removing this sentence... that's shorter at least.

TomFinley:
> Looks great! Is this something that we want to link in our API docs? Or should it be somewhere else in the ML.NET documentation website?
> I am thinking it could be nice to have this as an example under the `IDataView` interface, but maybe there are better places or ways to include it in the documentation.

Sure @artidoro, I can try to do that... I don't actually know how, so I just did some copy-pasta from one of @rogancarr's PRs and adapted it, let me know if you get a chance whether my `IDataView.cs` changes in next commit look good.

docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
```diff
@@ -0,0 +1,254 @@
...
+using Microsoft.ML.Data;
+
+namespace Samples.Dynamic
+{
+    /// <summary>
+    /// The <see cref="IDataView"/> interface is the central concept of "data" in ML.NET. While many conveniences exist
+    /// to create pre-baked implementations, it is also useful to know how to create one completely from scratch. We also
+    /// take this opportunity to illustrate and motivate the basic principles of how the IDataView system is architected,
+    /// since people interested in implementing <see cref="IDataView"/> need at least some knowledge of those principles.
+    /// </summary>
```
sfilipi: we're using simple, non-xml comments since the sample displays as is.

docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
```diff
@@ -0,0 +1,254 @@
...
+                // Hello my friend. => Hello, my, friend.
+                // Stay awhile and listen. => Stay, awhile, and, listen.
+                // Masterfully done hero! => Masterfully, done, hero!
+
+                // Note that it may be interesting to set a breakpoint on the Console.WriteLine, and explore
+                // what is going on with the cursor, and the buffers. In particular, on the third iteration,
+                // while `tokensValue` is logically presented as a three element array, internally you will
+                // see that the arrays internal to that structure have (at least) four items, specifically:
+                // `Masterfully`, `done`, `hero!`, `listen.`. In this way we see a simple example of the details
+                // of how buffer sharing from one iteration to the next actually works.
```
sfilipi: > next actually works [](start = 66, length = 20)

do you want to add a line about the Length.

docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
shmoradims: +1

---
In reply to: [274765737](https://github.com/dotnet/machinelearning/pull/3302#discussion_r274765737) [](ancestors = 274765737)

docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
shmoradims: A few points:

1) The docs engine will paste this examples verbatim, so none of the xml docstrings will be parsed and the F12 for cref won't work as you hoped them to.

2) If there's a way to put these in the <remarks> section for IDataView we can get all the xml features like cref. But you're writing a story that's attached to specific code blocks, so lumping it all in <remarks> would probably break the story.

3) This is ideal to be written as mixed xml + code pages like below but those pages are editorials on a different repo and are manually updated.
[https://docs.microsoft.com/en-us/dotnet/machine-learning/how-to-guides/inspect-intermediate-data-ml-net](https://docs.microsoft.com/en-us/dotnet/machine-learning/how-to-guides/inspect-intermediate-data-ml-net)

Having said all that, I think this works for V1 as is, with the only caveat the xml features like cref won't work.


---
In reply to: [274705835](https://github.com/dotnet/machinelearning/pull/3302#discussion_r274705835) [](ancestors = 274705835)

TomFinley: Very good points @shmoradims. Thanks for keeping track of this. Fair enough. We can think about that later. TBH I find the documentation system perhaps more confusing than other parts of our infrastructure, so I'm grateful to have your guidance in this area.

I'm still not sure why the `<see` tags wouldn't work if someone were to try to run the sample in VS -- certainly we reference types outside of our own library all the time, e.g., lots of things in the BCL -- but we can try as you say to work that out later.

docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
TomFinley: But why? Maybe we've had wildly difference experiences, but I've generally observed that people don't *read* samples as an intellectual exercise from beginning to end, they *might* read a little, but immediately fall to *running* them and see what the heck is going on. At least, such has been my observation. It's definitely true for me.

But, you're the expert, but I don't know that I like this idea of making the documentation non-browsable if someone actually tries to run the sample. Perhaps that's what you're doing, but I might prefer to have some understanding of why it's actually a good idea.

docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs
TomFinley: You mean describe that `Masterfully`, `done`, `hero!` has three items, but `Stay`, `awhile`, `and`, `listen` has four items? I know I usually err on the side of overdescription so this may seem a bit out of character for me to say but, I kind of feel like this particular point is so completely obvious that including it *explicitly* in the narrative I've structured serves more to de-focus that narrative than clarify it. But maybe you mean something else?

PR #3307, opened 2019-04-12
Rewrite image transform samples and dispose images after use.

### LINKED ISSUE #3306: Image samples don't dispose bitmap object after use causes exceptions.
When running samples all at once there is an exception between image samples because image transforms open and return image as a Bitmap object, however our samples don't dispose the bitmap object due to this when the next sample opens the same image it gets an exception saying the file is in use by the same or another process. Even though it is a new pipeline, we see this exception because garbage collector has not yet disposed the Bitmap object. The correct way to fix this problem is to dispose all bitmap objects after a sample has executed, this will not only solve the issue when running all samples at once in the same process but will also educate the user on disposing the bitmaps after use and make our samples better.

codemzs:
fixes #3306
towards #3350

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs
@@ -57,27 +57,30 @@ namespace Samples.Dynamic
             Console.WriteLine("{0, -25} {1, -25} {2, -25} {3, -25}", "ImagePath", "Name", "ImageObject", "Grayscale");
             using (var cursor = transformedData.GetRowCursor(transformedData.Schema))
             {
+                // Note that it is best to get the getters and values *before* iteration, so as to faciliate buffer
+                // sharing (if applicable), and column-type validation once, rather than many times.
+                ReadOnlyMemory<char> imagePath = default;
+                ReadOnlyMemory<char> name = default;
+                Bitmap imageObject = null;
+                Bitmap grayscaleImageObject = null;
+
                 var imagePathGetter = cursor.GetGetter<ReadOnlyMemory<char>>(cursor.Schema["ImagePath"]);
                 var nameGetter = cursor.GetGetter<ReadOnlyMemory<char>>(cursor.Schema["Name"]);
                 var imageObjectGetter = cursor.GetGetter<Bitmap>(cursor.Schema["ImageObject"]);
                 var grayscaleGetter = cursor.GetGetter<Bitmap>(cursor.Schema["Grayscale"]);
                 while (cursor.MoveNext())
                 {
-                    ReadOnlyMemory<char> imagePath = null;
                     imagePathGetter(ref imagePath);
-                    ReadOnlyMemory<char> name = null;
                     nameGetter(ref name);
-                    Bitmap imageObject = null;
                     imageObjectGetter(ref imageObject);
-                    Bitmap grayscaleImageObject = null;
                     grayscaleGetter(ref grayscaleImageObject);
 
                     Console.WriteLine("{0, -25} {1, -25} {2, -25} {3, -25}", imagePath, name, imageObject.PhysicalDimension, grayscaleImageObject.PhysicalDimension);
-
-                    // Dispose the image.
-                    imageObject.Dispose();
-                    grayscaleImageObject.Dispose();
                 }
+
+                // Dispose the image.
+                imageObject.Dispose();
+                grayscaleImageObject.Dispose();
             }
         }
     }
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs
@@ -44,6 +44,16 @@ namespace Samples.Dynamic
 
             var transformedData = pipeline.Fit(data).Transform(data);
 
+            PrintColumns(transformedData);
+            // ImagePath    Name         ImageObject              Grayscale
+            // tomato.bmp   tomato       {Width=800, Height=534}  {Width=800, Height=534}
+            // banana.jpg   banana       {Width=800, Height=288}  {Width=800, Height=288}
+            // hotdog.jpg   hotdog       {Width=800, Height=391}  {Width=800, Height=391}
+            // tomato.jpg   tomato       {Width=800, Height=534}  {Width=800, Height=534}
+        }
+
+        private static void PrintColumns(IDataView transformedData)
+        {
             Console.WriteLine("{0, -25} {1, -25} {2, -25} {3, -25}", "ImagePath", "Name", "ImageObject", "Grayscale");
             using (var cursor = transformedData.GetRowCursor(transformedData.Schema))
             {
@@ -62,19 +72,13 @@ namespace Samples.Dynamic
                     Bitmap grayscaleImageObject = null;
                     grayscaleGetter(ref grayscaleImageObject);
 
-                    Console.WriteLine("{0, -25} {1, -25} {2, -25} {3, -25}", imagePath, name, imageObject, grayscaleImageObject);
+                    Console.WriteLine("{0, -25} {1, -25} {2, -25} {3, -25}", imagePath, name, imageObject.PhysicalDimension, grayscaleImageObject.PhysicalDimension);
 
                     //Dispose the image.
                     imageObject.Dispose();
                     grayscaleImageObject.Dispose();
                 }
             }
-
-            // ImagePath    Name         ImageObject            Grayscale
-            // tomato.bmp   tomato       System.Drawing.Bitmap  System.Drawing.Bitmap
-            // banana.jpg   banana       System.Drawing.Bitmap  System.Drawing.Bitmap
-            // hotdog.jpg   hotdog       System.Drawing.Bitmap  System.Drawing.Bitmap
-            // tomato.jpg   tomato       System.Drawing.Bitmap  System.Drawing.Bitmap
         }
     }
 }
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs
@@ -74,7 +74,7 @@ namespace Samples.Dynamic
 
                     Console.WriteLine("{0, -25} {1, -25} {2, -25} {3, -25}", imagePath, name, imageObject.PhysicalDimension, grayscaleImageObject.PhysicalDimension);
 
-                    //Dispose the image.
+                    // Dispose the image.
                     imageObject.Dispose();
                     grayscaleImageObject.Dispose();
                 }
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs
@@ -43,54 +43,31 @@ namespace Samples.Dynamic
                            .Append(mlContext.Transforms.ConvertToGrayscale("Grayscale", "ImageObject"));
 
             var transformedData = pipeline.Fit(data).Transform(data);
-            // The transformedData IDataView contains the loaded images column, and the grayscaled column.
 
-            // Preview the transformedData. 
-            var transformedDataPreview = transformedData.Preview();
-            PrintPreview(transformedDataPreview);
+            // The transformedData IDataView contains the loaded images column, and the grayscaled column.
+            PrintColumns(transformedData);
+
             // ImagePath    Name         ImageObject            Grayscale
             // tomato.bmp   tomato       System.Drawing.Bitmap  System.Drawing.Bitmap
             // banana.jpg   banana       System.Drawing.Bitmap  System.Drawing.Bitmap
             // hotdog.jpg   hotdog       System.Drawing.Bitmap  System.Drawing.Bitmap
             // tomato.jpg   tomato       System.Drawing.Bitmap  System.Drawing.Bitmap
-
-            // Images are stored as Bitmap. Dispose them after use.
-            DisposeImages(transformedDataPreview);
         }
 
-        private static void PrintPreview(DataDebuggerPreview data)
+        private static void PrintColumns(IDataView transformedData)
         {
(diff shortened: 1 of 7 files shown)


eerhardt:
> In the meantime, these samples should be rewritten to not use this "preview" structure (I have no idea why we're writing sample code using a structure expressly and clearly intended purely for debugging in a preview window anyway) but, whatever the reason, let's do something else.

👍

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs
```diff
@@ -42,29 +43,31 @@ public static void Example()
...
+            var grayScaleColumn = transformedData.GetColumn<Bitmap>("Grayscale").GetEnumerator();
+            Console.WriteLine("{0, -25} {1, -25} {2, -25} {3, -25}", "ImagePath", "Name", "ImageObject", "Grayscale");
+            while (imagePathColumn.MoveNext() && namePathColumn.MoveNext() && imageObjectColumn.MoveNext() && grayScaleColumn.MoveNext())
             {
-                foreach (var kvPair in row.Values)
-                    Console.Write("{0,-25}", kvPair.Value);
-                Console.WriteLine();
+                Console.WriteLine("{0, -25} {1, -25} {2, -25} {3, -25}", imagePathColumn.Current, namePathColumn.Current, imageObjectColumn.Current, grayScaleColumn.Current);
+
+                //Dispose bitmap image.
```
wschin:
```suggestion
                // Dispose bitmap image.
```

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToImage.cs
```diff
@@ -32,35 +33,31 @@ public static void Example()
...
-                    else
-                        Console.Write("{0,-25}", kvPair.Value);
-                }
-                Console.WriteLine();
+                Console.WriteLine("{0, -25} {1, -25} {2, -25}",
+                    string.Join(",", featuresColumn.Current.DenseValues().Take(5)) + "...",
+                    imageColumn.Current,
+                    string.Join(",", pixelsColumn.Current.DenseValues().Take(5)) + "...");
+
+                //Dispose bitmap image.
```
wschin:
```suggestion
                // Dispose bitmap image.
```

codemzs:
@TomFinley Can you please review again and remove the block?

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs
```diff
@@ -42,29 +43,31 @@ public static void Example()
...
+            var namePathColumn = transformedData.GetColumn<string>("Name").GetEnumerator();
+            var imageObjectColumn = transformedData.GetColumn<Bitmap>("ImageObject").GetEnumerator();
+            var grayScaleColumn = transformedData.GetColumn<Bitmap>("Grayscale").GetEnumerator();
+            Console.WriteLine("{0, -25} {1, -25} {2, -25} {3, -25}", "ImagePath", "Name", "ImageObject", "Grayscale");
+            while (imagePathColumn.MoveNext() && namePathColumn.MoveNext() && imageObjectColumn.MoveNext() && grayScaleColumn.MoveNext())
             {
-                foreach (var kvPair in row.Values)
-                    Console.Write("{0,-25}", kvPair.Value);
-                Console.WriteLine();
+                Console.WriteLine("{0, -25} {1, -25} {2, -25} {3, -25}", imagePathColumn.Current, namePathColumn.Current, imageObjectColumn.Current, grayScaleColumn.Current);
```
TomFinley: Please use `IDataView`.

TomFinley: >Console.WriteLine("{0, -25} {1, -25} {2, -25} {3, -25}", imagePathColumn.Current, namePathColumn.Current, imageObjectColumn.Current, grayScaleColumn.Current); [](start = 16, length = 158)

I feel like we're not really doing anything interesting with this data. What does writing the bitmap to string actually do that is informative to anyone?

It seems like this is just replicating the effect of the preview dataview structure without actually inspecting why it was helpful or necessary to do that in the first place, with the end result that we now have a somewhat odd sample.

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs
```diff
@@ -42,29 +43,31 @@ public static void Example()
...
         {
-            foreach (var colInfo in data.ColumnView)
-                Console.Write("{0,-25}", colInfo.Column.Name);
-
-            Console.WriteLine();
-            foreach (var row in data.RowView)
+            var imagePathColumn = transformedData.GetColumn<string>("ImagePath").GetEnumerator();
+            var namePathColumn = transformedData.GetColumn<string>("Name").GetEnumerator();
+            var imageObjectColumn = transformedData.GetColumn<Bitmap>("ImageObject").GetEnumerator();
+            var grayScaleColumn = transformedData.GetColumn<Bitmap>("Grayscale").GetEnumerator();
```
TomFinley: >grayScaleColumn [](start = 16, length = 15)

We shouldn't have written this code this way, but note that enumerators implement `IDisposable`...

codemzs:
@TomFinley I have resolved your concerns. Please take another look.

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs
```diff
@@ -42,29 +43,41 @@ public static void Example()
...
-                    Console.Write("{0,-25}", kvPair.Value);
-                Console.WriteLine();
+                var imagePathGetter = cursor.GetGetter<ReadOnlyMemory<char>>(cursor.Schema["ImagePath"]);
+                var nameGetter = cursor.GetGetter<ReadOnlyMemory<char>>(cursor.Schema["Name"]);
+                var imageObjectGetter = cursor.GetGetter<Bitmap>(cursor.Schema["ImageObject"]);
+                var grayscaleGetter = cursor.GetGetter<Bitmap>(cursor.Schema["Grayscale"]);
+                while (cursor.MoveNext())
+                {
+                    ReadOnlyMemory<char> imagePath = null;
+                    imagePathGetter(ref imagePath);
```
TomFinley: While you have the getter outside the while loop, which is good, I see we have the variable inside the while loop. This we generally try to avoid as it is a bad practice. We try to do this throughout our codebase and implementations, and we should continue to encourage this "habit" in our example code, since it is something for people to learn from.

As it happens a few days ago I wrote a sample where I made this advice explicit.

[https://github.com/dotnet/machinelearning/blob/66ff419a95ae92dc80478b5e4af21002aae9ded0/docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs#L40-L49](https://github.com/dotnet/machinelearning/blob/66ff419a95ae92dc80478b5e4af21002aae9ded0/docs/samples/Microsoft.ML.Samples/Dynamic/SimpleDataViewImplementation.cs#L40-L49)

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToImage.cs
```diff
@@ -32,35 +33,39 @@ public static void Example()
...
                 {
-                    if (kvPair.Key == "Pixels" || kvPair.Key == "Features")
-                    {
-                        var rawValues = ((VBuffer<float>)kvPair.Value).DenseValues().Take(5);
-                        Console.Write("{0,-25}", string.Join(",", rawValues));
-                    }
-                    else
-                        Console.Write("{0,-25}", kvPair.Value);
+                    VBuffer<float> features = default;
+                    featuresGetter(ref features);
```
TomFinley: So, here's where my note becomes quite practically important. You've broken buffer sharing on `VBuffer`, by putting the declaration inside the while loop. So you're forcing reallocation, a big no-no.

Try [this documentation](https://github.com/dotnet/machinelearning/blob/master/docs/code/VBufferCareFeeding.md#buffer-re-use-as-a-user), I also talk about this principle here. Good idea to know how to use `IDataView` correctly.

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs
codemzs: Ok, great. thanks will update.

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToImage.cs
codemzs: sounds good. May be will put a comment as well to educate the users as they wouldn't know this away.

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/LoadImages.cs
```diff
@@ -41,29 +42,39 @@ public static void Example()
...
             {
-                foreach (var kvPair in row.Values)
-                    Console.Write("{0,-25}", kvPair.Value);
-                Console.WriteLine();
+                var imagePathGetter = cursor.GetGetter<ReadOnlyMemory<char>>(cursor.Schema["ImagePath"]);
+                var nameGetter = cursor.GetGetter<ReadOnlyMemory<char>>(cursor.Schema["Name"]);
+                var imageObjectGetter = cursor.GetGetter<Bitmap>(cursor.Schema["ImageObject"]);
+                while (cursor.MoveNext())
+                {
+                    ReadOnlyMemory<char> imagePath = null;
```
TomFinley: >null [](start = 53, length = 4)

Do not do `ReadOnlyMemory<char> imagePath = null`, prefer `ReadOnlyMemory<char> imagePath = default`. There is no need to invoke the implicit conversion since it can be so easily avoided.

TomFinley: In addition of course to the earlier advice to put this declaration *outside* the loop, but of course that advice applies everywhere.

---
In reply to: [275877068](https://github.com/dotnet/machinelearning/pull/3307#discussion_r275877068) [](ancestors = 275877068)

codemzs: Good idea.

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/LoadImages.cs
```diff
@@ -41,29 +42,39 @@ public static void Example()
...
+                    imagePathGetter(ref imagePath);
+                    ReadOnlyMemory<char> name = null;
+                    nameGetter(ref name);
+                    Bitmap imageObject = null;
+                    imageObjectGetter(ref imageObject);
+
+                    Console.WriteLine("{0, -25} {1, -25} {2, -25}", imagePath, name, imageObject.PhysicalDimension);
+
+                    // Dispose the image.
+                    imageObject.Dispose();
```
TomFinley: >imageObject.Dispose(); [](start = 20, length = 22)

Once you move the declaration outside the loop, move this dispose call outside as well.

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToImage.cs
TomFinley: Certainly can't hurt to reinforce the lesson, since we directly use `IDataView` in relatively few samples.

---
In reply to: [275874929](https://github.com/dotnet/machinelearning/pull/3307#discussion_r275874929) [](ancestors = 275874929)

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/LoadImages.cs
TomFinley: Similar with others of course.

---
In reply to: [275877915](https://github.com/dotnet/machinelearning/pull/3307#discussion_r275877915) [](ancestors = 275877915)

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToGrayScale.cs
```diff
@@ -42,29 +43,41 @@ public static void Example()
...
+                    ReadOnlyMemory<char> name = null;
+                    nameGetter(ref name);
+                    Bitmap imageObject = null;
+                    imageObjectGetter(ref imageObject);
+                    Bitmap grayscaleImageObject = null;
+                    grayscaleGetter(ref grayscaleImageObject);
+
+                    Console.WriteLine("{0, -25} {1, -25} {2, -25} {3, -25}", imagePath, name, imageObject.PhysicalDimension, grayscaleImageObject.PhysicalDimension);
```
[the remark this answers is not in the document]
codemzs: Cannot agree more! Thanks, @TomFinley!

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/DnnFeaturizeImage.cs
```diff
@@ -61,6 +62,9 @@ public static void Example()
             // 0.159010679
             // 0.394427955
 
+            // Dispose bitmap image objects.
+            foreach (var img in transformedData.GetColumn<Bitmap>("ImageObject"))
+                img.Dispose();
         }
```
TomFinley: What is the purpose of this here?

TomFinley: The reason why I ask is, it *might* be easy for someone to misinterpret and think that it is a required step, for some reason.

---
In reply to: [275888856](https://github.com/dotnet/machinelearning/pull/3307#discussion_r275888856) [](ancestors = 275888856)

codemzs: yep, this is not necessary because this is a direct getter that does not go through any other image transform that might lead to the creation of intermediate images.

PR #3309, opened 2019-04-12
Standalone app to run all samples to catch run time exceptions.

### LINKED ISSUE #3308: Standalone app to run all samples to catch run time exceptions
We need an app that run as a build step to verify samples not only compile but also execute with no exceptions. We have recently seen as evident by #3187 that samples are being checked-in that when run by a user will not produce any meaningful output but instead cause runtime exception and leave a very bad user experience.

codemzs:
fixes #3308

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
+++ b/docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
@@ -4,9 +4,10 @@
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <OutputType>Exe</OutputType>
     <SignAssembly>false</SignAssembly>
+    <!--This ensures that we can never make the mistake of adding this as a friend assembly. Please don't remove.-->
     <PublicSign>false</PublicSign>
     <RootNamespace>Samples</RootNamespace>
-	<StartupObject>Microsoft.ML.Samples.Program</StartupObject>
+    <StartupObject>Microsoft.ML.Samples.Program</StartupObject>
   </PropertyGroup>
 
   <ItemGroup>
--- a/docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
+++ b/docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
@@ -7,7 +7,6 @@
     <!--This ensures that we can never make the mistake of adding this as a friend assembly. Please don't remove.-->
     <PublicSign>false</PublicSign>
     <RootNamespace>Samples</RootNamespace>
-    <StartupObject>Microsoft.ML.Samples.Program</StartupObject>
   </PropertyGroup>
 
   <ItemGroup>
--- a/docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
+++ b/docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
@@ -6,6 +6,7 @@
     <SignAssembly>false</SignAssembly>
     <PublicSign>false</PublicSign>
     <RootNamespace>Samples</RootNamespace>
+	<StartupObject>Microsoft.ML.Samples.Program</StartupObject>
   </PropertyGroup>
 
   <ItemGroup>
--- a/docs/samples/Microsoft.ML.Samples/Program.cs
+++ b/docs/samples/Microsoft.ML.Samples/Program.cs
@@ -11,7 +11,7 @@ namespace Microsoft.ML.Samples
         {
             //DataViewEnumerable.Example();
 
-            //if (args[1] == "-runall")
+            if (args[1] == "-runall")
                 RunAll();
         }
 
--- a/docs/samples/Microsoft.ML.Samples/Program.cs
+++ b/docs/samples/Microsoft.ML.Samples/Program.cs
@@ -1,4 +1,5 @@
-﻿using System.Reflection;
+﻿using System;
+using System.Reflection;
 
 namespace Microsoft.ML.Samples
 {
@@ -8,8 +9,19 @@ namespace Microsoft.ML.Samples
 
         internal static void RunAll()
         {
+            int samples = 0;
             foreach (var type in Assembly.GetExecutingAssembly().GetTypes())
-                type.GetMethod("Example", BindingFlags.Public | BindingFlags.Static | BindingFlags.FlattenHierarchy)?.Invoke(null, null);
+            {
+                var sample = type.GetMethod("Example", BindingFlags.Public | BindingFlags.Static | BindingFlags.FlattenHierarchy);
+
+                if(sample != null)
+                {
+                    sample.Invoke(null, null);
+                    samples++;
+                }
+            }
+
+            Console.WriteLine("Number of samples that ran without any exception: " + samples);
         }
     }
 }
--- a/docs/samples/Microsoft.ML.Samples/Program.cs
+++ b/docs/samples/Microsoft.ML.Samples/Program.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Reflection;
 using Samples.Dynamic;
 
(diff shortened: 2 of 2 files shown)


docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
```diff
@@ -3,8 +3,8 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <OutputType>Exe</OutputType>
-    <SignAssembly>false</SignAssembly>
-    <PublicSign>false</PublicSign>
```
Ivanidzo4ka: I think this properties were added by @artidoro for a reason.

docs/samples/Microsoft.ML.Samples/Program.cs
```diff
@@ -1,12 +1,43 @@
...
+                        method.Invoke(null, null);
+                    }
+                    catch (Exception ex)
+                    {
+                        // Print the exception for debugging.
+                        Console.Write(type);
+                        Console.Write(ex);
+
+                        // Throw to fail.
+                        throw ex;
```
TomFinley: So, generally, if you have an exception thrown it will print out the stack trace and message anyway, just by virtue of you running it. So, I'm not sure why we have this `try`/`catch` block? The only thing that it seems to accomplish is that it makes it 

Also, even if you wanted to do this, which, I really don't think you want to do, which I'm not sure why you do, please just re-throw the exception, so `throw;` instead of `throw ex;`.

So let's just point out, even if we do things this way:

```csharp
private static void EvilMethod()
{
    string a = null;
    Console.WriteLine(a.Length);
}

private static void Main(string[] args)
{
    try
    {
        EvilMethod();
    }
    catch
    {
        throw;
    }
}
```

If I run the debugger, the debugger stops on that last `throw;`. I have no information on anything...

If on the other hand, I just run this simpler code, under the debugger I get it stopped right where the exception was thrown (so I have some ability to inspect the actual state of the objects). Also from the command line perspective even if run without debugging I still get the console.

```csharp
private static void EvilMethod()
{
    string a = null;
    Console.WriteLine(a.Length);
}

private static void Main(string[] args)
{
    EvilMethod();
}
```

So, it seems like the only thing all this extra material accomplishes is, it just makes it harder to debug the actual issue?

docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
TomFinley: Right. This ensures that we can never make the mistake of adding this as a friend assembly... please don't remove...

docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
```diff
@@ -3,8 +3,8 @@
...
+    <AssemblyName>Samples</AssemblyName>
```
TomFinley: Is there any reason to make the assembly name distinct from the project name? I got the point with the console MAML since that was intended to be used, but I don't think it's a good idea here.

docs/samples/Microsoft.ML.Samples/Program.cs
```diff
@@ -1,12 +1,43 @@
...
 
 namespace Microsoft.ML.Samples
 {
-    internal static class Program
+    public static class Program
     {
-        static void Main(string[] args)
+        public static void Main(string[] args)
         {
             CalculateFeatureContribution.Example();
```
shmoradims: >CalculateFeatureContribution.Example(); [](start = 12, length = 39)

let's remove this

docs/samples/Microsoft.ML.Samples/Program.cs
```diff
@@ -1,12 +1,43 @@
...
+
+            if (args.Length > 0 && args[0] == "-runall")
```
shmoradims: >if (args.Length > 0 && args[0] == "-runall") [](start = 12, length = 44)

we want this run all samples without any arguments, so that devs can just hit F5 while creating samples w/o having to mock around with debug arguments.

docs/samples/Microsoft.ML.Samples/Program.cs
```diff
@@ -1,12 +1,43 @@
...
+        }
+
+        internal static void RunAll()
+        {
+            foreach (var type in Assembly.GetExecutingAssembly().GetTypes())
+            {
+                var method = type.GetMethod("Example", BindingFlags.NonPublic | BindingFlags.Public
+                    | BindingFlags.Static | BindingFlags.FlattenHierarchy);
+
+                if (method != null)
```
shmoradims: >if (method != null) [](start = 16, length = 19)

let's also handle the case where (method == null), which are all the cases where we don't conform with pattern below:
public static Example()

docs/samples/Microsoft.ML.Samples/Program.cs
```diff
@@ -1,12 +1,43 @@
...
+
+            if (args.Length > 0 && args[0] == "-runall")
+                RunAll();
+        }
+
+        internal static void RunAll()
+        {
+            foreach (var type in Assembly.GetExecutingAssembly().GetTypes())
+            {
+                var method = type.GetMethod("Example", BindingFlags.NonPublic | BindingFlags.Public
```
shmoradims: >NonPublic [](start = 68, length = 9)

I don't think we care about NonPublic

docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
shauheen: Can you also please add a comment explaining this for posterior knowledge.

docs/samples/Microsoft.ML.Samples/Program.cs
eerhardt: I've never seen a valid case of `throw ex;` in a catch block in my time as a developer. It always should have been just `throw;`

docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
eerhardt: If you want to rename the AssemblyName, just rename the `.csproj` file instead.

docs/samples/Microsoft.ML.Samples/Program.cs
codemzs: @shmoradims there are types not related to sample classes and they don't have example methods and there is nothing we can do about it, i.e throw exceptions. I see your point you want to catch samples that are not following the pattern but then there is that.

docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
```diff
@@ -4,8 +4,10 @@
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <OutputType>Exe</OutputType>
     <SignAssembly>false</SignAssembly>
+    <!--This ensures that we can never make the mistake of adding this as a friend assembly. Please don't remove.-->
     <PublicSign>false</PublicSign>
     <RootNamespace>Samples</RootNamespace>
+    <StartupObject>Microsoft.ML.Samples.Program</StartupObject>
```
eerhardt: Why is this necessary?

docs/samples/Microsoft.ML.Samples/Program.cs
```diff
@@ -1,12 +1,15 @@
...
     {
-        static void Main(string[] args)
+        public static void Main(string[] args) => RunAll();
+
+        internal static void RunAll()
         {
-            CalculateFeatureContribution.Example();
+            foreach (var type in Assembly.GetExecutingAssembly().GetTypes())
+                type.GetMethod("Example", BindingFlags.Public | BindingFlags.Static | BindingFlags.FlattenHierarchy)?.Invoke(null, null);
         }
```
shmoradims: >} [](start = 8, length = 1)

please add a 1-liner report like "20 sample ran successfully" so that sample-writers get a concrete feedback the the code actually ran

PR #3313, opened 2019-04-12
put product version instead of version 

### LINKED ISSUE #3132: Version.txt in model .zip should use the FileVersion, not AssemblyVersion

Ivanidzo4ka:
Fixes #3132
master PR https://github.com/dotnet/machinelearning/pull/3173

### DIFF
--- a/src/Microsoft.ML.Core/Data/Repository.cs
+++ b/src/Microsoft.ML.Core/Data/Repository.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
 using Microsoft.ML.Internal.Utilities;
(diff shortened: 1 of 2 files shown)


shauheen:
Done as part of #3333 . Thanks @Ivanidzo4ka .

PR #3314, opened 2019-04-12
Renaming CI legs to make them more readable and understandable

Anipik:


build/ci/phase-template.yml
```diff
@@ -21,11 +21,11 @@ phases:
       parallel: 99
       matrix:
         ${{ if eq(parameters.customMatrixes, '') }}:
-          Build_Debug:
+          Build_Debug_2.1:
```
eerhardt: Do we really need the word `Build` in our all legs?

.vsts-dotnet-ci.yml
```diff
@@ -63,14 +48,30 @@ phases:
...
-    name: MacOS
+    name: MacOS_x64
     buildScript: ./build.sh
     queue:
       name: Hosted macOS
 
 - template: /build/ci/phase-template.yml
   parameters:
-    name: Windows_NetFx
+    name: Windows_x64_1
```
eerhardt: I'm not sure `_1` and `_2` help any on the readability, does it?

Anipik: I didnt want to add these suffixes  but names for the jobs need to be unique.

eerhardt: They look unique to me without the suffixes:

image

Anipik: ```Debug_3.0``` comes from custom Matrix attribute (It will be always added automatically) the name is the first part before with space.

eerhardt: It seems like something better should be able to be done here.  Maybe @safern or @chcosta would have some ideas.

chcosta: If you were to merge the matrices such that you were using a single job with a larger matrix, you wouldn't hit collision issues with job names.

Also, phases are now a deprecated syntax.  Azure DevOps has moved to "jobs".  There are some actual changes associated with the naming change and I'd suggest you move to the supported syntax.  Here's our (Arcade's) documentation which is related but at least will point you in the right direction - https://github.com/dotnet/arcade/blob/master/Documentation/AzureDevOps/PhaseToJobSchemaChange.md

Anipik: @eerhardt I used a bigger custom matrix. Although it increased the number of CI legs but its just more combinations of configurations.

eerhardt:
It looks like you've added 8 more legs to the CI. I don't think we want this. Please keep the number of CI legs and the coverage the same.

Anipik:
> It looks like you've added 8 more legs to the CI. I don't think we want this. Please keep the number of CI legs and the coverage the same.

i did that and i was able to remove the suffix from the windows.  @eerhardt  i dont think we want to move to using the more recent   "Job Version" of Azure dev ops at this point ?

eerhardt:
I think this is looking better. I just have a few inconsistencies that would be nice to be fixed:

image

1. `CentOS` and `Ubuntu` have "Debug/Release" first, but everyone else has it last.
2. Sometimes it is `Debug` and sometimes `debug`.
3. Why does `Windows_x64_Netcore30 Debug_3.0` have `3.0` in the name twice?
4. I'm not sure I like `Netcore21` and `Netcore30`. Nowhere else uses that terminology/wording. How about either the product name: `.NET Core 2.1` or the TFM: `netcoreapp2.1`?

Anipik:
> CentOS and Ubuntu have "Debug/Release" first, but everyone else has it last.

I am not sure how to correct this. In all other configurations we are using the same configuration for debug and release but not here

> Sometimes it is Debug and sometimes debug.
Why does Windows_x64_Netcore30 Debug_3.0 have 3.0 in the name twice?

I can correct both of these

> I'm not sure I like Netcore21 and Netcore30. Nowhere else uses that terminology/wording. How about either the product name: .NET Core 2.1 or the TFM: netcoreapp2.1?

The Name has all the variable restrictions that are all alphanumeric characterer and ```_```. i can change it to netcoreapp21 and netcoreapp30 ?

.vsts-dotnet-ci.yml
```diff
@@ -63,21 +48,37 @@ phases:
...
+    name: Windows_x64_Netfx
     buildScript: build.cmd
     customMatrixes:
-      Build_Debug_netfx:
+      Debug:
         _configuration: Debug-netfx
         _config_short: DFX
         _includeBenchmarkData: false
-      Build_Release_netfx:
+      Release:
```
chcosta: I'd suggest not naming any of these matrix legs anything as simple as "Release" or "Debug".  In the past, we've seen issues where if the Job name (agent variable) is "Release", then some tasks (some MicroBuild task) thinks it is running in a release pipeline and exhibits odd behavior.

Anipik: Wont the job name here will Windows_x64_Netfx Debug ?

artidoro:
The position of debug/release in the name of the CI leg does not seem to be consistent across all the legs.
By the way @Anipik are you still working on this? It would be great to have it in!

Anipik:
i am not able to find a perfect way to position and name these ci legs. https://github.com/dotnet/machinelearning/pull/3314#issuecomment-483286941 

looked good but the windows ci has suffix attached so that was an issue. so still working on finding some perfect combination. any suggestions are welcomed

artidoro:
Ok I'll write my suggestion. Why do we only have Debug for 3.0 and Release for 2.1 on Centos and Ubuntu?

artidoro:
My suggestion is as follows:

- [OS Name]_[Platform architecture x86 or x64]_[Net type and version]_[Release or Debug]

I know in the current CI there is a differentiation using the intrinsics key word. If that's necessary for the new CI legs, we could append that at the end as well:

- [OS Name]_[Platform architecture x86 or x64]_[Net type and version]_[Release or Debug]_[Intrinsics if needed]

Let me give a few examples (not exhaustive):

- Centos_x64_NetCore30_Debug
- Centos_x64_NetCore30_Release
- Ubuntu_x64_NetCore21_Debug
- Ubuntu_x64_NetCore21_Release
- Windows_x86_NetCore30_Release
- Windows_x64_NetFx_Debug

PR #3315, opened 2019-04-12
Tom cherry pick

TomFinley:
Cherry pick of some of my commits into release/1.0.  Which PRs they correspond to is obvious from the commit descriptions. Only one, the last one, required conflict resolution. I have verified the sample introduced still runs.

Note also that when I run the test locally, the test `EntryPointLinearPredictorSummary` fails. This appears to have nothing to do with my changes, but it is probably something that shiproom should be aware of. @shauheen , @glebuk .


shauheen:
Done as part of #3333 . Thanks @TomFinley .

PR #3328, opened 2019-04-13
Dispose source image object in image transfomers.

### LINKED ISSUE #3327: Image transfomers don't dispose source image.
Image transformers such as grayscale and resize have images as input. After they process input images to produce output image they should dispose the input image because if they don't then we are at mercy of garbage collector cleaning the bitmap objects and if the GC has not done that and some code in the same process tries to open the same image referenced by the bitmap object that is not disposed then we get file in use exception.

codemzs:
fixes #3327

### DIFF
--- a/src/Microsoft.ML.ImageAnalytics/ImageGrayscale.cs
+++ b/src/Microsoft.ML.ImageAnalytics/ImageGrayscale.cs
@@ -218,6 +218,9 @@ namespace Microsoft.ML.Transforms.Image
                             g.DrawImage(src, srcRectangle, 0, 0, src.Width, src.Height, GraphicsUnit.Pixel, attributes);
                         }
                         Contracts.Assert(dst.Width == src.Width && dst.Height == src.Height);
+
+                        src.Dispose();
+                        src = null;
                     };
 
                 return del;
--- a/src/Microsoft.ML.ImageAnalytics/ImageResizer.cs
+++ b/src/Microsoft.ML.ImageAnalytics/ImageResizer.cs
@@ -396,6 +396,9 @@ namespace Microsoft.ML.Transforms.Image
                             g.DrawImage(src, destRectangle, srcRectangle, GraphicsUnit.Pixel);
                         }
                         Contracts.Assert(dst.Width == info.ImageWidth && dst.Height == info.ImageHeight);
+
+                        src.Dispose();
+                        src = null;
                     };
 
                 return del;


src/Microsoft.ML.ImageAnalytics/ImageGrayscale.cs
```diff
@@ -218,6 +218,9 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
                             g.DrawImage(src, srcRectangle, 0, 0, src.Width, src.Height, GraphicsUnit.Pixel, attributes);
                         }
                         Contracts.Assert(dst.Width == src.Width && dst.Height == src.Height);
+
+                        src.Dispose();
```
eerhardt: Is Disposing it the correct thing to do here? Are you given ownership of the `src` object? If so, then it is fine. But if `getSrc` caches that object and tries using it, people will get an ObjectDisposedException if they try to use it.

Also - there is a case on line 210 where this method could return early without disposing `src`.

codemzs: I’m aware about line 210 but that is an error condition and not sure bitmap object has the ownership of the file but I’ll check. I did also consider the caching issue and didn’t see it being cached anywhere. None of the image transforms seem to cache the bitmap object.

wschin: ML.NET pipeline runs in a read-only mode. `Disposing something` means read-and-write.

src/Microsoft.ML.ImageAnalytics/ImageGrayscale.cs
```diff
@@ -218,6 +218,9 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
...
+                        src = null;
```
TomFinley: So, we see the correct pattern above at line 205, where `dst` is disposed if non-`null`. So the getter should in principle be disposing. Is it not? If it is not it should be.

So this seems to be disposing potentially in *both* places, but this makes me think we just have a bad getter delegate somewhere.

src/Microsoft.ML.ImageAnalytics/ImageGrayscale.cs
codemzs: You need to dispose otherwise it is a bug. Not entirely sure what your argument here is but we need a solution.

TomFinley:
>                         getSrc(ref src);

So to be clear, I'd expect this getter (whatever it is!) to dispose of the thing if necessary. I'd also expect the `disposer` above to handle any final cleanup on the `DataViewRow` object that is created using this infrastructure.

Is there some reason this does not work?

---
Refers to: src/Microsoft.ML.ImageAnalytics/ImageGrayscale.cs:208 in 4bd3d1b. [](commit_id = 4bd3d1b4de920f3e950e1d7201c09f6497f26a0d, deletion_comment = False)

src/Microsoft.ML.ImageAnalytics/ImageGrayscale.cs
codemzs: That is dst, this is src.

src/Microsoft.ML.ImageAnalytics/ImageGrayscale.cs
TomFinley: Sure. We need a solution. My question is more, why the existing solution and infrastructure is insufficient, since *in principle* the `src` should be disposed by the getter, then *finally* disposed when the row cursor is itself disposed. I am wondering what has broken down. #3327 is not sufficient.

---
In reply to: [275461003](https://github.com/dotnet/machinelearning/pull/3328#discussion_r275461003) [](ancestors = 275461003)

src/Microsoft.ML.ImageAnalytics/ImageGrayscale.cs
TomFinley: The point is that 205 is showing the correct pattern, and we also have the dispose method. If this pattern is being adhered to everywhere, the dispose method is being composed correct, and the cursor is being disposed, `src` should be being disposed already. Investigate why this is not so.

---
In reply to: [275461358](https://github.com/dotnet/machinelearning/pull/3328#discussion_r275461358) [](ancestors = 275461358)

PR #3330, opened 2019-04-13
[AutoML] try to fix mlnet tests on netfx

Dmitry-A:


src/mlnet/mlnet.csproj
```diff
@@ -115,4 +104,7 @@
     </None>
   </ItemGroup>
 
+  <!-- CLI tool is only compatible with Core so don't try to build it on FX --> 
+  <Import Condition="'$(TargetFramework)' != 'netcoreapp2.1'" Project="$(RepoRoot)build\Empty.targets" />
```
eerhardt: Try moving this to the `mlnet.tests.csproj` instead. That should fix your netfx leg.

PR #3331, opened 2019-04-13
[AutoML] disable netfx build leg for now

Dmitry-A:
will re-enable once I figure out the fix for not building/running mlnet test there


Dmitry-A:
Hmm looks like random ml.net test failure, will re-run

PR #3339, opened 2019-04-15
Clarify documentation for GetFeatureWeights in TreeEnsembleModelParameters

### LINKED ISSUE #1850: OLS FeatureWeights are not the model weights
For the `LinearPredictor`, the `GetFeatureWeights` returns the model weights, as expected:

```
public virtual void GetFeatureWeights(ref VBuffer<Float> weights)
{
    Weight.CopyTo(ref weights);
}
```

The `OlsLinearRegressionPredictor` has the same method, but it does not return the model weights. It returns the [`-log(p-value)` for each feature](https://github.com/dotnet/machinelearning/blob/14c7a472579afa3ce98bad2e3495e4c0524471b9/src/Microsoft.ML.HalLearners/OlsLinearRegression.cs#L782). This is weird because it overrides the `LinearRegressionPredictor`'s GetFeatureWeights method but returns a different kind of value, essentially the magnitudes of the p-values.

Now, this goes back to the meaning of `feature weight`. Many predictors that are not linear models implement `GetFeatureWeight`, and it looks like it's a measure of importance. What's the right thing here? Do we want to return a measure of relative importance or do we want to return the model weights?

Currently, I find this API to be super confusing because there is no docstring to explain what you are getting back, and I would expect the "Feature Weights" of a linear model to the model weight parameters.

Also, this is a nit, but if we really want to return the magnitude of the p-values, doesn't `-log10(p-value)` make more sense? The base e log puts these onto a weird scale.

rogancarr:
This PR clarifies the documentation for `GetFeatureWeights()` in `TreeEnsembleModelParameters`. Since `IHaveFeatureWeights` has become `internal`, the `TreeEnsembleModelParameters` are now the only place where this method is visible, and it is unclear from the API what this returns. This PR adds documentation to specify that the "weights" returned are the cumulative split gains for all the trees in the ensemble.

Fixes #1850

### DIFF
--- a/src/Microsoft.ML.FastTree/FastTree.cs
+++ b/src/Microsoft.ML.FastTree/FastTree.cs
@@ -3229,10 +3229,10 @@ namespace Microsoft.ML.Trainers.FastTree
         }
 
         /// <summary>
-        /// Copy the weights of all training features to <paramref name="weights"/>.
+        /// Get the cumulative split gains for each feature across all trees.
         /// </summary>
-        /// <param name="weights">a <see cref="VBuffer{T}"/> where feature weights would be assigned to.
-        /// The i-th element in <paramref name="weights"/> stores the weight of the i-th feature.</param>
+        /// <param name="weights">A <see cref="VBuffer{T}"/> to hold the cumulative split gain value for each feature.
+        /// The i-th element in <paramref name="weights"/> stores the cumulative split gain of the i-th feature.</param>
         public void GetFeatureWeights(ref VBuffer<float> weights)
         {
             var numFeatures = Math.Max(NumFeatures, MaxSplitFeatIdx + 1);


src/Microsoft.ML.FastTree/FastTree.cs
```diff
@@ -3229,10 +3229,10 @@ private void ToCSharp(InternalRegressionTree tree, TextWriter writer, int node,
         }
 
         /// <summary>
-        /// Copy the weights of all training features to <paramref name="weights"/>.
+        /// Get the cumulative split gains for each feature across all trees.
         /// </summary>
-        /// <param name="weights">a <see cref="VBuffer{T}"/> where feature weights would be assigned to.
-        /// The i-th element in <paramref name="weights"/> stores the weight of the i-th feature.</param>
+        /// <param name="weights">A <see cref="VBuffer{T}"/> to hold the cumulative split gain value for each feature.
+        /// The i-th element in <paramref name="weights"/> stores the cumulative split gain of the i-th feature.</param>
```
TomFinley: Thank you @rogancarr! As discussed "weights" is a bit of an unfortunate word for "cumulative split gains," but it's probably too late to change the API at this point.

Issue #3255, opened 2019-04-09
Complete term review of all sources


glebuk:
Need to ensure that source content complies to term policies

glebuk:
Fixed in PR #3252.

PR #3341, opened 2019-04-15
Cherrypick to release: "Renamed WhatTheFeature and their abbreviations to FeatureContributionScorer."

glebuk:
….  Note that this breaks model backwards compatibility. (#3252)

Cherrypicks PR #3252 and issue #3255  into release branch


shauheen:
Thanks @glebuk , I will update the release from master again.

PR #3351, opened 2019-04-16
[AutoML] Add AutoML XML documentation to all public members; migrate AutoML projects & tests into ML.NET solution; AutoML test fixes

daholste:


src/Microsoft.ML.Auto/API/AutoCatalog.cs
```diff
@@ -0,0 +1,242 @@
...
+        {
+            return new MulticlassClassificationExperiment(_context, experimentSettings);
+        }
+
+        /// <summary>
+        /// Infers information about the columns of a dataset in a file located at <paramref name="path"/>.
+        /// </summary>
+        /// <param name="path">Path to a dataset file.</param>
+        /// <param name="labelColumnName">The name of the label column.</param>
+        /// <param name="separatorChar">The character used as separator between data points in a row. If <see langword="null"/>, AutoML will try to infer this value.</param>
```
justinormont: I think of an entire row as (potentially) a singular multi-dimensional "data point", perhaps "data element" is better?
```suggestion
        /// <param name="separatorChar">The character used as separator between data elements in a row. If <see langword="null"/>, AutoML will try to infer this value.</param>
```

daholste: this is much better -- thank you!

src/Microsoft.ML.Auto/API/AutoCatalog.cs
```diff
@@ -0,0 +1,242 @@
...
+        /// <param name="allowQuoting">Whether the file can contain column defined by a quoted string. If <see langword="null"/>, AutoML will try to infer this value.</param>
```
justinormont:
```suggestion
        /// <param name="allowQuoting">Whether the file can contain columns defined by a quoted string. If <see langword="null"/>, AutoML will try to infer this value.</param>
```

daholste: thank you!

daholste: this was a copy-paste error from https://github.com/dotnet/machinelearning/blob/70ef7ecd43b031b481a4047ea361da5e2f360336/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs#L22

daholste: will make a Github issue

src/Microsoft.ML.Auto/API/AutoCatalog.cs
justinormont: We likely should move to `char[] Separators` to match ML.NET.

Ref: https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.textloader.options.separators?view=ml-dotnet

src/Microsoft.ML.Auto/API/AutoCatalog.cs
daholste: https://github.com/dotnet/machinelearning/pull/3355/files

src/Microsoft.ML.Auto/API/AutoCatalog.cs
daholste: Agreed that this warrants further discussion --
I believe we went this direction b/c ML.NET's analog method takes a single separator character as well --https://github.com/dotnet/machinelearning/blob/70ef7ecd43b031b481a4047ea361da5e2f360336/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs#L27

src/Microsoft.ML.Auto/API/ExperimentSettings.cs
```diff
@@ -23,10 +51,10 @@ public class ExperimentSettings
         /// <summary>
         /// This setting controls whether or not an AutoML experiment will make use of ML.NET-provided caching.
         /// If set to true, caching will be forced on for all pipelines. If set to false, caching will be forced off.
-        /// If set to null (default value), AutoML will decide whether to enable caching for each model.
+        /// If set to <see langword="null"/> (default value), AutoML will decide whether to enable caching for each model.
         /// </summary>
         public bool? CacheBeforeTrainer = null;
```
justinormont: We had a quorum of folks decide this should be an enum: on/off/auto. Welcome to add into this PR; or we'll handle later.

src/Microsoft.ML.Auto/API/AutoCatalog.cs
justinormont: Ok. Sounds good.

src/Microsoft.ML.Auto/API/ExperimentSettings.cs
daholste: +1 -- will handle in a PR shortly w/ a few other API revs

PR #3352, opened 2019-04-16
Add samples for MLContext.Model (root) and add links to documentation.

codemzs:
Towards #1209

### DIFF
--- a/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
+++ b/src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
@@ -105,7 +105,7 @@ namespace Microsoft.ML
         /// <example>
         /// <format type="text/markdown">
         /// <![CDATA[
-        /// [!code-csharp[Save](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/CustomMappingSaveAndLoad.cs)]
+        /// [!code-csharp[Save](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/ModelOperations/SaveLoadModel.cs)]
         /// ]]>
         /// </format>
         /// </example>
@@ -145,7 +145,7 @@ namespace Microsoft.ML
         /// <example>
         /// <format type="text/markdown">
         /// <![CDATA[
-        /// [!code-csharp[Load](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TimeSeries/DetectChangePointBySsa.cs]
+        /// [!code-csharp[Save](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/ModelOperations/SaveLoadModel.cs)]
         /// ]]>
         /// </format>
         /// </example>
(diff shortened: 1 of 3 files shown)


src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -281,6 +302,13 @@ public ITransformer LoadWithDataLoader(string filePath, out IDataLoader<IMultiSt
         /// <typeparamref name="TDst"/>.</param>
         /// <param name="inputSchemaDefinition">Additional settings of the input schema.</param>
         /// <param name="outputSchemaDefinition">Additional settings of the output schema.</param>
+        /// <example>
+        /// <format type="text/markdown">
+        /// <![CDATA[
+        /// [!code-csharp[Load](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/TextClassification.cs)]
```
shmoradims: >TextClassification [](start = 103, length = 18)

this looks too complicated. let's use a simpler pipeline with bare minimum (no metrics, no testing, no textloader, just Fit and CreatePredictEngine). You can take one of our simpler examples and cut the unnecessary parts.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -187,6 +201,13 @@ public ITransformer Load(Stream stream, out DataViewSchema inputSchema)
         /// <param name="inputSchema">Will contain the input schema for the model. If the model was saved without
         /// any description of the input, there will be no input schema. In this case this can be <see langword="null"/>.</param>
         /// <returns>The loaded model.</returns>
+        /// <example>
+        /// <format type="text/markdown">
+        /// <![CDATA[
+        /// [!code-csharp[Load](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/CustomMappingSaveAndLoad.cs)]
```
shmoradims: >CustomMappingSaveAndLoad [](start = 103, length = 24)

I think majority of the Load()/Save() operations don't require mlContext.ComponentCatalog.RegisterAssembly that CustomMapping requires, which makes this sample the exception not the rule. If that's assumption is correct, let's stick to a sample that doesn't scare users with using  RegisterAssembly, CustomMappingFactoryAttribute, and Action<>. You can use the same approach I suggested for CreatePredictionEngine sample below (simple data, simple pipeline) and just simple Save() then Load().

CustomMappingSaveAndLoad is best suited for CustomMapping API which we already use it there.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
```diff
@@ -135,6 +142,13 @@ private void SaveInputSchema(DataViewSchema inputSchema, RepositoryWriter rep)
         /// <param name="inputSchema">Will contain the input schema for the model. If the model was saved without
         /// any description of the input, there will be no input schema. In this case this can be <see langword="null"/>.</param>
         /// <returns>The loaded model.</returns>
+        /// <example>
+        /// <format type="text/markdown">
+        /// <![CDATA[
+        /// [!code-csharp[Load](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TimeSeries/DetectChangePointBySsa.cs]
```
shmoradims: >DetectChangePointBySsa [](start = 114, length = 22)

Same here. please use/create a sample with the bare minimum required to show the API usage.

src/Microsoft.ML.Data/Model/ModelOperationsCatalog.cs
sfilipi: Agree for this one, CustomMapping might distract them. Maybe a simpler pipeline. 

---
In reply to: [275778414](https://github.com/dotnet/machinelearning/pull/3352#discussion_r275778414) [](ancestors = 275778414)

PR #3359, opened 2019-04-16
API reference - Fix placeholder links

### LINKED ISSUE #2356: Curate resource links for API reference
There are many topics that repeat throughout the API reference. For each topic, we want to have a separate link that contains more details, instead of duplicating content, or worse, stating the obvious like the following:
* Bias: The predictor's bias term.
* L2Weight: L2 regularization weight for trainer

Table below captures all the topics that we want to cover. These are the topics that apply to many APIs. For example, we have 4 or 5 cross validation extension methods, one per ML task. We can have one page explaining CV, instead of duplicating content. 

Some topics already exist in [glossary](https://docs.microsoft.com/en-us/dotnet/machine-learning/resources/glossary). We need to have a discussion about our linking strategy later. For now let's keep the list here. I suggest using the LinkPlaceholder string (tmpurl_*) in the XMLs, so that we can easily find-replace with the final link.

| Topic | Description | In Glossary? | Link Placeholder | Link |
| ------------- |-------------| -----| -----| -----|
| Regularization | L1, L2, and other topics related to regularization | no | tmpurl_regularization | https://en.wikipedia.org/wiki/Regularization_(mathematics) |
| Loss  | Loss functions  | no  | tmpurl_loss  | https://en.wikipedia.org/wiki/Loss_function |
| Calibration / calibrators  | calibrators used for producing probabilities  | no  | tmpurl_calib  | https://en.wikipedia.org/wiki/Calibration_(statistics)  |
| Learning rate  | what is learning rate in training  | no | tmpurl_lr  | There's no wikipage talking about learning rate. It's always mixed with SGD or the training algorithm using it. Deleting this since it's not an independent concept. |

shmoradims:
Fixes #2356 with actual links.

### DIFF
--- a/src/Microsoft.ML.StandardTrainers/Standard/Online/AveragedLinear.cs
+++ b/src/Microsoft.ML.StandardTrainers/Standard/Online/AveragedLinear.cs
@@ -61,7 +61,7 @@ namespace Microsoft.ML.Trainers
         public bool LazyUpdate = true;
 
         /// <summary>
-        /// The L2 weight for <a href='tmpurl_regularization'>regularization</a>.
+        /// The L2 weight for <a href='https://en.wikipedia.org/wiki/Regularization_(mathematics)'>regularization</a>.
         /// </summary>
         [Argument(ArgumentType.AtMostOnce, HelpText = "L2 Regularization Weight", ShortName = "reg,L2RegularizerWeight", SortOrder = 50)]
         [TGUI(Label = "L2 Regularization Weight")]
--- a/src/Microsoft.ML.StandardTrainers/Standard/Online/AveragedLinear.cs
+++ b/src/Microsoft.ML.StandardTrainers/Standard/Online/AveragedLinear.cs
@@ -21,7 +21,7 @@ namespace Microsoft.ML.Trainers
     public abstract class AveragedLinearOptions : OnlineLinearOptions
     {
         /// <summary>
-        /// <a href="tmpurl_lr">Learning rate</a>.
+        /// Learning rate.
         /// </summary>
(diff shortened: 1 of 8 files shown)


src/Microsoft.ML.StandardTrainers/Standard/Online/AveragedLinear.cs
```diff
@@ -21,7 +21,7 @@ namespace Microsoft.ML.Trainers
     public abstract class AveragedLinearOptions : OnlineLinearOptions
     {
         /// <summary>
-        /// <a href="tmpurl_lr">Learning rate</a>.
+        /// Learning rate.
```
singlis: Should there be the uri here for what tempurl_lr referenced?

singlis:
@shmoradims I have one comment to review, but it does look good. I am a little confused on why we are replacing the tempurl with the actual link when it looks like in 2356 the point is to use the tempurl.

src/Microsoft.ML.StandardTrainers/Standard/Online/AveragedLinear.cs
```diff
@@ -61,7 +61,7 @@ public abstract class AveragedLinearOptions : OnlineLinearOptions
         public bool LazyUpdate = true;
 
         /// <summary>
-        /// The L2 weight for <a href='tmpurl_regularization'>regularization</a>.
+        /// The L2 weight for <a href='https://en.wikipedia.org/wiki/Regularization_(mathematics)'>regularization</a>.
```
wschin:
```suggestion
        /// The coefficient of L2-norm for <a href='https://en.wikipedia.org/wiki/Regularization_(mathematics)'>regularization on the weight vector</a>.
```

PR #3367, opened 2019-04-16
Setting metadata back to defaults for Microsoft.ML.DataView

### LINKED ISSUE #3366: Microsoft.ML.DataView metadat needs to be fixed
In #2974 we moved IDataView back to Microsoft.ML, but did not update the logo for the nuget back to ML.NET

shauheen:
Fixes #3366


shauheen:
> This is going into 1.0 right?

That is correct.

PR #3370, opened 2019-04-17
Increase Native build warning level  to W3 and fix build warnings

glebuk:
Before, we actually only reported W1 warning level during native build.
Now we warn on L3 warnings
Fixed three warnings from downcasting variables. In all cases the downcasting was deemed safe and replaced with ```static_cast<>```
Note that all those warnings were deemed required to fix for compliance criteria.
This tracks #3370

### DIFF
--- a/src/Native/CMakeLists.txt
+++ b/src/Native/CMakeLists.txt
@@ -32,7 +32,6 @@ if(WIN32)
     add_compile_options(/FC) # use full pathnames in diagnostics
     add_compile_options(/DEBUG)
     add_compile_options(/GS)
-    add_compile_options(/W1)
     add_compile_options(/Zc:inline)
     add_compile_options(/fp:precise)
     add_compile_options(/EHsc)
--- a/src/Native/CpuMathNative/Sse.cpp
+++ b/src/Native/CpuMathNative/Sse.cpp
@@ -158,7 +158,7 @@ EXPORT_API(void) MatMulTran(_In_ const float * pmat, _In_ const float * psrc, _I
     }
 
     pm += 3 * crow;
-    
+
     for (; ps < psLim; ps += 4)
     {
         __m128 x01 = _mm_load_ps(ps);
@@ -219,9 +219,9 @@ EXPORT_API(void) Scale(float a, _Inout_ float * pd, int c)
     {
         switch (c)
         {
-            case 3: pd[2] *= a;
-            case 2: pd[1] *= a;
-            case 1: pd[0] *= a;
+        case 3: pd[2] *= a;
+        case 2: pd[1] *= a;
+        case 1: pd[0] *= a;
         }
         return;
     }
@@ -266,7 +266,7 @@ EXPORT_API(void) Scale(float a, _Inout_ float * pd, int c)
             _mm_storeu_ps(pd, result);
 
             pd += misalignment;
-			// safe to downcast as misalignment < 16.
+            // safe to downcast as misalignment <= 128.
             c -= static_cast<int>(misalignment);
         }
 
@@ -490,9 +490,9 @@ EXPORT_API(float) Sum(const float* pValues, int length)
 
         switch (length)
         {
-            case 3: result += pValues[2];
-            case 2: result += pValues[1];
-            case 1: result += pValues[0];
+        case 3: result += pValues[2];
+        case 2: result += pValues[1];
+        case 1: result += pValues[0];
         }
 
         return result;
@@ -533,8 +533,8 @@ EXPORT_API(float) Sum(const float* pValues, int length)
             result = _mm_add_ps(result, temp);
 
             pValues += misalignment;
-			// safe to downcastcast as misalignment < 16.
-			length -= static_cast<int>(misalignment);
+            // safe to downcast as misalignment < 16.
+            length -= static_cast<int>(misalignment);
         }
 
         if (length > 3)
--- a/src/Native/CpuMathNative/Sse.cpp
+++ b/src/Native/CpuMathNative/Sse.cpp
@@ -266,7 +266,8 @@ EXPORT_API(void) Scale(float a, _Inout_ float * pd, int c)
             _mm_storeu_ps(pd, result);
 
             pd += misalignment;
-            c -= misalignment;
+			// safe to downcast as misalignment < 16.
+            c -= static_cast<int>(misalignment);
         }
 
         if (c > 3)
@@ -532,7 +533,8 @@ EXPORT_API(float) Sum(const float* pValues, int length)
             result = _mm_add_ps(result, temp);
 
             pValues += misalignment;
-            length -= misalignment;
+			// safe to downcastcast as misalignment < 16.
+			length -= static_cast<int>(misalignment);
         }
 
         if (length > 3)
--- a/src/Native/LdaNative/alias_multinomial_rng_int.hpp
+++ b/src/Native/LdaNative/alias_multinomial_rng_int.hpp
@@ -21,6 +21,8 @@ Fast generation of discrete random variables
 */
 namespace wood
 {
+	using namespace std;
+
     struct alias_k_v
     {
         int32_t k_;
@@ -108,7 +110,7 @@ namespace wood
             int32_t H_head = 0;
             int32_t H_tail = 0;
 
-			if (proportion_int_.size() > numeric_limits<int32_t>::max())
+			if (proportion_int_.size() > static_cast<uint32_t>(numeric_limits<int32_t>::max()))
 				bad_array_new_length();
 
 			int32_t size = static_cast<int32_t>(proportion_int_.size());
--- a/src/Native/LdaNative/alias_multinomial_rng_int.hpp
+++ b/src/Native/LdaNative/alias_multinomial_rng_int.hpp
@@ -108,7 +108,11 @@ namespace wood
             int32_t H_head = 0;
             int32_t H_tail = 0;
 
-            for (size_t i = 0; i < proportion_int_.size(); ++i)
+			if (proportion_int_.size() > numeric_limits<int32_t>::max())
+				bad_array_new_length();
+
+			int32_t size = static_cast<int32_t>(proportion_int_.size());
+            for (int32_t i = 0; i < size; ++i)
             {
                 auto val = proportion_int_[i];
                 if (val < a_int_)
--- a/src/Native/LdaNative/alias_multinomial_rng_int.hpp
+++ b/src/Native/LdaNative/alias_multinomial_rng_int.hpp
@@ -110,7 +110,7 @@ namespace wood
 
 
             // note that i must fit into int32_t of L_[L_tail].first
-            if (std::numeric_limits<int32_t>::max() < proportion_int_.size() )
+            if (static_cast<uint32_t>(std::numeric_limits<int32_t>::max()) < proportion_int_.size() )
                 throw std::bad_alloc();
             int32_t size = static_cast<int32_t>(proportion_int_.size());
             for (int32_t i = 0; i < size; ++i)
--- a/src/Native/LdaNative/alias_multinomial_rng_int.hpp
+++ b/src/Native/LdaNative/alias_multinomial_rng_int.hpp
@@ -14,15 +14,13 @@
 #include <iostream>
 #include <assert.h>
 /*
-Algorithm described in 
+Algorithm described in
 https://www.jstatsoft.org/v11/i03/paper
 George Marsaglia
 Fast generation of discrete random variables
 */
 namespace wood
 {
-	using namespace std;
-
     struct alias_k_v
     {
         int32_t k_;
@@ -110,7 +108,11 @@ namespace wood
             int32_t H_head = 0;
             int32_t H_tail = 0;
 
-			int32_t size = static_cast<int32_t>(proportion_int_.size());
+
+            // note that i must fit into int32_t of L_[L_tail].first
+            if (std::numeric_limits<int32_t>::max() < proportion_int_.size() )
+                throw std::bad_alloc();
+            int32_t size = static_cast<int32_t>(proportion_int_.size());
             for (int32_t i = 0; i < size; ++i)
             {
                 auto val = proportion_int_[i];
@@ -157,7 +159,7 @@ namespace wood
                 auto first = L_[L_head].first;
                 auto second = L_[L_head].second;
                 alias_kv[first].k_ = first;
-                alias_kv[first].v_ = first  * a_int_ + second;
+                alias_kv[first].v_ = first * a_int_ + second;
                 ++L_head;
             }
             while (H_head != H_tail)
@@ -230,7 +232,7 @@ namespace wood
                 *p = i;  p++;
                 *p = (i + 1) * a_int_;
             }
-            
+
             int32_t L_head = 0;
             int32_t L_tail = 0;
 
@@ -298,8 +300,8 @@ namespace wood
                 *p = first; p++;
                 *p = first * a_int_ + second;
                 ++H_head;
-            }    
-            memcpy(memory, internal_memory_, sizeof(int32_t)* 2 * n_);
+            }
+            memcpy(memory, internal_memory_, sizeof(int32_t) * 2 * n_);
         }
 
         inline void SetProportionMass(std::vector<float> &proportion,
--- a/src/Native/LdaNative/alias_multinomial_rng_int.hpp
+++ b/src/Native/LdaNative/alias_multinomial_rng_int.hpp
@@ -108,7 +108,7 @@ namespace wood
             int32_t H_head = 0;
             int32_t H_tail = 0;
 
-            for (auto i = 0; i < proportion_int_.size(); ++i)
+            for (size_t i = 0; i < proportion_int_.size(); ++i)
             {
                 auto val = proportion_int_[i];
                 if (val < a_int_)
--- a/src/Native/LdaNative/alias_multinomial_rng_int.hpp
+++ b/src/Native/LdaNative/alias_multinomial_rng_int.hpp
@@ -110,9 +110,6 @@ namespace wood
             int32_t H_head = 0;
             int32_t H_tail = 0;
 
-			if (proportion_int_.size() > static_cast<uint32_t>(numeric_limits<int32_t>::max()))
-				bad_array_new_length();
-
 			int32_t size = static_cast<int32_t>(proportion_int_.size());
             for (int32_t i = 0; i < size; ++i)
             {
--- a/src/Native/LdaNative/data_block.cpp
+++ b/src/Native/LdaNative/data_block.cpp
@@ -51,7 +51,7 @@ namespace lda
     {
 		num_documents_ = num_document;
 
-		if (corpus_size > numeric_limits<size_t>::max() / sizeof(int64_t) + 1)
+		if (corpus_size < 0 || static_cast<uint64_t>(corpus_size) > numeric_limits<size_t>::max())
 			bad_alloc();
         corpus_size_ = static_cast<size_t>(corpus_size);
 
--- a/src/Native/LdaNative/data_block.cpp
+++ b/src/Native/LdaNative/data_block.cpp
@@ -3,11 +3,15 @@
 // See the LICENSE file in the project root for more information.
 
 #include <iostream>
+#include <stdexcept>
+#include <limits>
 #include "data_block.h"
 #include "lda_document.h"
 
 namespace lda
 {
+	using namespace std;
+
     LDADataBlock::LDADataBlock(int32_t num_threads) : 
         num_threads_(num_threads), has_read_(false), index_document_(0), documents_buffer_(nullptr), offset_buffer_(nullptr)
     {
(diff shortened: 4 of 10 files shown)


src/Native/CpuMathNative/Sse.cpp
```diff
@@ -266,7 +266,8 @@ EXPORT_API(void) Scale(float a, _Inout_ float * pd, int c)
             _mm_storeu_ps(pd, result);
 
             pd += misalignment;
-            c -= misalignment;
+			// safe to downcast as misalignment < 16.
```
wschin: I am not sure if it's still 16 for AVX or longer vector instruction sets, but I believe it's safe.

src/Native/LdaNative/alias_multinomial_rng_int.hpp
```diff
@@ -21,6 +21,8 @@ Fast generation of discrete random variables
 */
 namespace wood
 {
+	using namespace std;
+
```
wschin: Would this be too much for two new lines? Maybe we can do "std::"?

src/Native/LdaNative/data_block.cpp
```diff
@@ -45,8 +49,12 @@ namespace lda
 
     void LDADataBlock::Allocate(const int32_t num_document, const int64_t corpus_size)
```
wschin: If we change `cropus_size` to `size_t`, do we still need the following check?

src/Native/LdaNative/model_block.cpp
```diff
@@ -92,14 +96,21 @@ namespace lda
...
 
-        mem_block_size_ = 2 * upper_bound(load_factor_ * nonzero_num);
-        alias_mem_block_size_ = nonzero_num * 3; 
+		uint64_t size = 2 * upper_bound(load_factor_ * nonzero_num);
+		if (nonzero_num < 0 || size > numeric_limits<size_t>::max())
+			throw bad_alloc();
+
+        mem_block_size_ = static_cast<size_t>(size);
+
+		size = nonzero_num * 3;
```
wschin: >nonzero_num * 3; [](start = 9, length = 16)

Overflow can happen here. Let's if nonzero_num < max(sizes_t) / 3.

src/Native/LdaNative/alias_multinomial_rng_int.hpp
```diff
@@ -108,7 +110,8 @@ namespace wood
             int32_t H_head = 0;
             int32_t H_tail = 0;
 
-            for (auto i = 0; i < proportion_int_.size(); ++i)
+			int32_t size = static_cast<int32_t>(proportion_int_.size());
```
eerhardt: Isn't it more correct to use `std::vector<int>::size_type` here?

```
for(std::vector<int>::size_type i = 0; i != v.size(); i++) {
    /* std::cout << v[i]; ... */
}
```

src/Native/CpuMathNative/Sse.cpp
```diff
@@ -532,7 +533,8 @@ EXPORT_API(float) Sum(const float* pValues, int length)
             result = _mm_add_ps(result, temp);
 
             pValues += misalignment;
-            length -= misalignment;
+			// safe to downcastcast as misalignment < 16.
+			length -= static_cast<int>(misalignment);
```
danmosemsft: Looks like you have some tabs creeping in ..

src/Native/LdaNative/data_block.cpp
```diff
@@ -45,8 +49,12 @@ namespace lda
...
-        num_documents_ = num_document;
-        corpus_size_ = corpus_size;
+		num_documents_ = num_document;
+
+		if (corpus_size < 0 || static_cast<uint64_t>(corpus_size) > numeric_limits<size_t>::max())
+			bad_alloc();
+        corpus_size_ = static_cast<size_t>(corpus_size);
+
 
         offset_buffer_ = new int64_t[num_documents_ + 1]; // +1: one for the end of last document,
```
wschin: Please check if `num_documents_ + 1` overflows.

src/Native/LdaNative/model_block.cpp
```diff
@@ -324,14 +337,20 @@ namespace lda
             alias_offset += alias_row_size;
         }
 
-        mem_block_size_ = dict_[num_vocabs_ - 1].end_offset_;
+		uint64_t size = dict_[num_vocabs_ - 1].end_offset_;
+		if (size > numeric_limits<size_t>::max())
+			throw bad_alloc();
```
eerhardt: I see sometimes you `throw bad_alloc()`, but I've seen 2 other places where you just call `bad_alloc()` without throwing. Why the difference?

src/Native/MatrixFactorizationNative/UnmanagedMemory.cpp
```diff
@@ -115,7 +115,7 @@ EXPORT_API(float) MFCrossValidation(const mf_problem_bridge *prob_bridge, int32_
 {
     auto param = TranslateToParam(param_bridge);
     auto prob = TranslateToProblem(prob_bridge);
-    return mf_cross_validation(&prob, nr_folds, param);
+    return static_cast<float>(mf_cross_validation(&prob, nr_folds, param));
```
eerhardt: It appears `mf_cross_validation` returns `mf_double`

[https://github.com/cjlin1/libmf/blob/dadfd21ef3f6af633bb1b4c35f79a0da8df5be82/mf.h#L98](https://github.com/cjlin1/libmf/blob/dadfd21ef3f6af633bb1b4c35f79a0da8df5be82/mf.h#L98)

which is a `double`

[https://github.com/cjlin1/libmf/blob/dadfd21ef3f6af633bb1b4c35f79a0da8df5be82/mf.h#L16](https://github.com/cjlin1/libmf/blob/dadfd21ef3f6af633bb1b4c35f79a0da8df5be82/mf.h#L16)

Is `static_cast` really the appropriate thing to do here?

eerhardt: Would it be more appriopriate to return a `double` instead?

---
In reply to: [276750656](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276750656) [](ancestors = 276750656)

src/Native/CMakeLists.txt
```diff
@@ -32,7 +32,6 @@ if(WIN32)
     add_compile_options(/FC) # use full pathnames in diagnostics
     add_compile_options(/DEBUG)
     add_compile_options(/GS)
-    add_compile_options(/W1)
```
eerhardt: Is the default level 3?

singlis: Minor nit - but I would prefer this to be explicitly set.

---
In reply to: [276751232](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276751232) [](ancestors = 276751232)

src/Native/CpuMathNative/Sse.cpp
```diff
@@ -266,7 +266,8 @@ EXPORT_API(void) Scale(float a, _Inout_ float * pd, int c)
             _mm_storeu_ps(pd, result);
 
             pd += misalignment;
-            c -= misalignment;
+			// safe to downcast as misalignment < 16.
+            c -= static_cast<int>(misalignment);
```
singlis: >static_cast [](start = 17, length = 11)

you can cast with int, i.e. (int)misalignment;. I dont think the static_cast is buying anything here.

src/Native/CpuMathNative/Sse.cpp
```diff
@@ -532,7 +533,8 @@ EXPORT_API(float) Sum(const float* pValues, int length)
             result = _mm_add_ps(result, temp);
 
             pValues += misalignment;
-            length -= misalignment;
+			// safe to downcastcast as misalignment < 16.
```
singlis: >downcastcast [](start = 14, length = 12)

downcastcast

src/Native/CpuMathNative/Sse.cpp
singlis: >static_cast [](start = 13, length = 11)

same here, you can use (int)misalignment;

src/Native/CMakeLists.txt
eerhardt: Just an FYI that this code was copied from [https://github.com/dotnet/core-setup/blob/97d93381785a25e433fce74d3dd848f1bb07e0d9/src/settings.cmake#L160](https://github.com/dotnet/core-setup/blob/97d93381785a25e433fce74d3dd848f1bb07e0d9/src/settings.cmake#L160), which is what builds the `dotnet.exe` host. I'm a bit surprised that we've gone through almost 3 major versions of .NET Core using `/W1` and no one has complained.

---
In reply to: [276751232](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276751232) [](ancestors = 276751232)

src/Native/CpuMathNative/Sse.cpp
glebuk: true, there is  >>=2 there.  It is within 128 then

---
In reply to: [276744696](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276744696) [](ancestors = 276744696)

src/Native/LdaNative/model_block.cpp
glebuk: bug, need to throw

---
In reply to: [276749841](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276749841) [](ancestors = 276749841)

src/Native/LdaNative/alias_multinomial_rng_int.hpp
glebuk: yes, will fix

---
In reply to: [276748210](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276748210) [](ancestors = 276748210)

src/Native/CpuMathNative/Sse.cpp
glebuk: I thought in C++ we are to use *_cast<>() in most cases as per guidelines... ([http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-casts](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-casts))

---
In reply to: [276761019](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276761019) [](ancestors = 276761019)

src/Native/MatrixFactorizationNative/UnmanagedMemory.cpp
glebuk: Perhaps, but that would require changes all the way through to the C#  side.  Do we want to do this?  In most cases ML.NET uses single precision anyway...

---
In reply to: [276751151](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276751151) [](ancestors = 276751151,276750656)

src/Native/CMakeLists.txt
glebuk: No kidding, I was quite surprised when I found this. Perhaps this was not found because if you look for W3 you will find it and assume that it is set.  Fixing thishas exposed some potentially bad situations.  Perhaps the .NET core script needs to be update as well?

---
In reply to: [276763498](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276763498) [](ancestors = 276763498,276751232)

src/Native/LdaNative/alias_multinomial_rng_int.hpp
singlis: I agree - it is more correct, but you would have to cast when setting L_[L_tail].first and H_[H_tail].first to int32_t.

Gleb - should this fail if the size is bigger than int32_t max? If so, it may warrant a check first then cast the size as this cast will not throw during runtime. Also the cast can change to size=(int32_t)proportion_int.size();


---
In reply to: [276748210](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276748210) [](ancestors = 276748210)

src/Native/LdaNative/data_block.h
```diff
@@ -44,7 +44,7 @@ namespace lda
         int64_t used_size_;
```
singlis: >int64_t [](start = 8, length = 7)

should used_size_ also be unsigned?

src/Native/LdaNative/model_block.cpp
```diff
@@ -113,20 +124,22 @@ namespace lda
...
         for (int i = 0; i < num_vocabs_; ++i)
         {
-            // This warning is a false positive. Supressing it similar to the existing one on Line 140 below.
-#pragma warning(suppress: 6386)
             dict_[i].is_dense_ = 0;
             dict_[i].is_alias_dense_ = 0;
         }
 
-        mem_block_size_ = mem_block_size;
+		if (mem_block_size < 0 || static_cast<uint64_t>(mem_block_size) > numeric_limits<size_t>::max())
```
singlis: >uint64_t [](start = 40, length = 8)

Is it worth checking if mem_block_size is negative and throwing?

src/Native/LdaNative/model_block.cpp
```diff
@@ -113,20 +124,22 @@ namespace lda
...
+			throw bad_alloc();
+        mem_block_size_ = static_cast<size_t>(mem_block_size);
         mem_block_ = new int32_t[mem_block_size_]();   // NOTE : force to initialize the values to be zero
 
-        alias_mem_block_size_ = alias_mem_block_size;
+		if (mem_block_size < 0 || static_cast<uint64_t>(alias_mem_block_size) > numeric_limits<size_t>::max())
```
singlis: >alias_mem_block_size [](start = 50, length = 20)

I noticed when outputting the size, you are multiplying by 4. Should the *4 apply here in the check? And what is the *4? Is this supposed to be sizeof(float)?

src/Native/LdaNative/model_block.cpp
```diff
@@ -113,20 +124,22 @@ namespace lda
...
+			throw bad_alloc();
+        alias_mem_block_size_ = static_cast<size_t>(alias_mem_block_size);
         alias_mem_block_ = new int32_t[alias_mem_block_size_]();    //NOTE: force to initialize the values to be zero
 
-        std::cout << "mem_block_size = " << mem_block_size_ * 4 << std::endl;
-        std::cout << "alias_mem_block_size = " << alias_mem_block_size_ * 4 << std::endl;
+        cout << "mem_block_size = " << mem_block_size_ * 4 << endl;
```
singlis: >4 [](start = 57, length = 1)

sizeof(float)?

src/Native/LdaNative/model_block.cpp
```diff
@@ -157,7 +170,7 @@ namespace lda
         if (fullSparse)
         {
             // use a very large threshold to ensure every row of word-topic-table using a sparse representation
-            hot_thresh = std::numeric_limits<int>::max();
+            hot_thresh = numeric_limits<int>::max();
```
singlis: >int [](start = 40, length = 3)

why not max of int32_t?

src/Native/LdaNative/model_block.cpp
```diff
@@ -167,7 +180,7 @@ namespace lda
         if (fullSparse)
         {
             // use a very large threshold to ensure every row of alias table using a sparse representation
-            alias_hot_thresh = std::numeric_limits<int>::max();
+            alias_hot_thresh = numeric_limits<int>::max();
```
singlis: >alias_hot_thresh [](start = 12, length = 16)

same here

src/Native/CpuMathNative/Sse.cpp
singlis: ok thanks Gleb, I wont argue with Herb Sutter

---
In reply to: [276795215](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276795215) [](ancestors = 276795215,276761019)

src/Native/LdaNative/model_block.cpp
glebuk: you mean  sizeof(int32_t)? (those are arrays of int32_t )

---
In reply to: [276830689](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276830689) [](ancestors = 276830689)

src/Native/LdaNative/data_block.h
glebuk: Perhaps, but that would require a lot of changes.  I have only touched those variables that go directly into the size of the array that can under-allocate. In this case, the used_size_ is just a counter variable that measures some document size.  

---
In reply to: [276827553](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276827553) [](ancestors = 276827553)

src/Native/LdaNative/model_block.cpp
glebuk: I think that is the size of the array item (sizeof(int32_t) I'll update

---
In reply to: [276830624](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276830624) [](ancestors = 276830624)

src/Native/CpuMathNative/Sse.cpp
glebuk: just following patterns... they are not equivalent afaik. ([https://stackoverflow.com/questions/1609163/what-is-the-difference-between-static-cast-and-c-style-casting](https://stackoverflow.com/questions/1609163/what-is-the-difference-between-static-cast-and-c-style-casting))

---
In reply to: [276760574](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276760574) [](ancestors = 276760574)

src/Native/LdaNative/data_block.cpp
glebuk: It should not:  num_documents_ is signed int 32, while size_t is either uint32 or uint64

---
In reply to: [276748897](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276748897) [](ancestors = 276748897)

src/Native/LdaNative/data_block.cpp
glebuk: Yes, I can not change the corpus size to size_t because that value changes per platform bitness (32/64)  I do not have a mechanism to support this in C# / C++ interface.

---
In reply to: [276746121](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276746121) [](ancestors = 276746121)

src/Native/LdaNative/model_block.cpp
glebuk: indeed, why not?  - fixed.

---
In reply to: [276830864](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276830864) [](ancestors = 276830864)

src/Native/LdaNative/alias_multinomial_rng_int.hpp
glebuk: I think I fixed it. Please check.

---
In reply to: [276826843](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276826843) [](ancestors = 276826843,276748210)

src/Native/LdaNative/data_block.cpp
glebuk: You are right. fixed - added check

---
In reply to: [276748897](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276748897) [](ancestors = 276748897)

src/Native/CMakeLists.txt
glebuk: I want explicit setting.

---
In reply to: [276796288](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276796288) [](ancestors = 276796288,276763498,276751232)

src/Native/LdaNative/model_block.cpp
glebuk: updated to use sizeof(int32)

---
In reply to: [276865464](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276865464) [](ancestors = 276865464,276830624)

src/Native/LdaNative/model_block.cpp
```diff
@@ -324,14 +337,20 @@ namespace lda
...
-        alias_mem_block_size_ = dict_[num_vocabs_ - 1].alias_end_offset_;
+        size = dict_[num_vocabs_ - 1].alias_end_offset_;
+        if (size > numeric_limits<size_t>::max())
+            throw bad_alloc();
+        alias_mem_block_size_ = static_cast<size_t>(size);
         alias_mem_block_ = new int32_t[alias_mem_block_size_]();    //NOTE: force to initialize the values to be zero
 
-        std::cout << "mem_block_size = " << mem_block_size_ * 4 << std::endl;
-        std::cout << "alias_mem_block_size = " << alias_mem_block_size_ * 4 << std::endl;
+        cout << "mem_block_size = " << mem_block_size_ * 4 << endl;
```
singlis: >4 [](start = 57, length = 1)

same here with sizeof(int32_t)

singlis: actually instead of using mem_block_size_ * sizeof(uint32_t) -- if you just do sizeof(alias_mem_block_) that will give you the full of the allocated memory.

---
In reply to: [276881438](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276881438) [](ancestors = 276881438)

src/Native/LdaNative/alias_multinomial_rng_int.hpp
```diff
@@ -108,7 +108,12 @@ namespace wood
             int32_t H_head = 0;
             int32_t H_tail = 0;
 
-            for (auto i = 0; i < proportion_int_.size(); ++i)
+
+            // note that i must fit into int32_t of L_[L_tail].first
+            if (static_cast<uint32_t>(std::numeric_limits<int32_t>::max()) < proportion_int_.size() )
```
singlis: >int32_t [](start = 58, length = 7)

shouldnt this be uint32_t? Then you could remove the static_cast. Also why not have if (proportion_int_.size() > std::numeric_limits<uint32_t>::max())?

singlis: oh I see why you are using the int32_t, ok so you are getting the max of int32_t, then cast that to uint32_t for the size comparison?

---
In reply to: [276882246](https://github.com/dotnet/machinelearning/pull/3370#discussion_r276882246) [](ancestors = 276882246)

PR #3379, opened 2019-04-17
DataOperations: Replace PREVIEW API usage with proper IDataView based Enumerable API.

### LINKED ISSUE #3350: Stop using Preview structure and API in samples.
From PR #3307 and @TomFinley 's [comment ](https://github.com/dotnet/machinelearning/pull/3307#pullrequestreview-225891973) it was learnt that samples should be rewritten to not use "preview" structure because we're writing sample code using a structure expressly and clearly intended purely for debugging in a preview window anyway. There are a number of samples that use preview structure, I will start converting these samples in batches and replace the usage of preview API/structure with something more appropriate like IDataView + getter or Enumerable(with class). We definitely don't want to be educating our users the wrong thing.

CC: @TomFinley @shmoradims @natke @shauheen

codemzs:
fixes #3350

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/CrossValidationSplit.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/CrossValidationSplit.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using Microsoft.ML;
-using static Microsoft.ML.DataOperationsCatalog;
 
 namespace Samples.Dynamic
 {
@@ -26,7 +25,9 @@ namespace Samples.Dynamic
             // So below, we specify Group column as the column containing the sampling keys.
             // If we pass that column to cross validation it would be used to break data into certain chunks.
             var folds = mlContext.Data.CrossValidationSplit(dataview, numberOfFolds: 3, samplingKeyColumnName: "Group");
-            PrintPreviewRows(folds[0]);
+            var trainSet = mlContext.Data.CreateEnumerable<DataPoint>(folds[0].TrainSet, reuseRowObject: false);
+            var testSet = mlContext.Data.CreateEnumerable<DataPoint>(folds[0].TestSet, reuseRowObject: false);
+            PrintPreviewRows(trainSet, testSet);
 
             // The data in the Train split.
             // [Group, 1], [Features, 0.8173254]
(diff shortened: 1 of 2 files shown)


docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/CrossValidationSplit.cs
```diff
@@ -26,7 +25,9 @@ public static void Example()
             // So below, we specify Group column as the column containing the sampling keys.
             // If we pass that column to cross validation it would be used to break data into certain chunks.
             var folds = mlContext.Data.CrossValidationSplit(dataview, numberOfFolds: 3, samplingKeyColumnName: "Group");
-            PrintPreviewRows(folds[0]);
+            var trainSet = mlContext.Data.CreateEnumerable<DataPoint>(folds[0].TrainSet, reuseRowObject: false);
+            var testSet = mlContext.Data.CreateEnumerable<DataPoint>(folds[0].TestSet, reuseRowObject: false);
```
shmoradims: for CrossValidationSplit we only need one dataset. we just need to cv-split it and show the fold. no need for a separate train and test. in fact cv is used when you don't have a test set, so it's better to remove it.

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/CrossValidationSplit.cs
```diff
@@ -26,7 +25,9 @@ public static void Example()
             // So below, we specify Group column as the column containing the sampling keys.
             // If we pass that column to cross validation it would be used to break data into certain chunks.
             var folds = mlContext.Data.CrossValidationSplit(dataview, numberOfFolds: 3, samplingKeyColumnName: "Group");
```
shmoradims: >samplingKeyColumnName: "Group" [](start = 88, length = 30)

good idea to show usage both with and without groups.

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/CrossValidationSplit.cs
codemzs: Spoke offline and resolved. This is good as it is.

ganik:
> using static Microsoft.ML.DataOperationsCatalog;

need this?

---
Refers to: docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/TrainTestSplit.cs:4 in 2cb353e. [](commit_id = 2cb353ea681d1ac878bd64607653cfb2d30aaca9, deletion_comment = False)

codemzs:
> using static Microsoft.ML.DataOperationsCatalog;

Yep, for line 28. TrainTestData object. but I can use var and remove it. Thanks!

---
In reply to: [484237998](https://github.com/dotnet/machinelearning/pull/3379#issuecomment-484237998) [](ancestors = 484237998)

---
Refers to: docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/TrainTestSplit.cs:4 in 2cb353e. [](commit_id = 2cb353ea681d1ac878bd64607653cfb2d30aaca9, deletion_comment = False)

Issue #2522, opened 2019-02-13
Docs and samples for the API reference site (P0 & P1 Trainers)


shmoradims:
# List of trainers per catalog

## BinaryClassification.Trainers

BinaryClassification Trainers | Category | Priority | Docs (1st pass) | Docs (2nd pass) | Samples (1st   pass) | Samples (2nd   pass)
-- | -- | -- | -- | -- | -- | --
AveragedPerceptron | Linear | 0 | Shahab #2517 | Done Shahab #3310  | Shahab #2517 |  Artidoro #3311 |
LogisticRegression | Linear | 0 | Wei-Sheng #3385  | Done Wei-Sheng #3385   | ZeeshanA #3099 |  Artidoro #3311|
Stochastic GradientDescent | Linear | 0 | Shahab #2688 |  @codemzs #3392 | Shahab #2688 |  Artidoro #3311|
StochasticDual CoordinateAscent | Linear | 0 | Shahab #2771 |  @codemzs #3395  | Shahab #2771 |  Artidoro #3311|
SymbolicStochastic GradientDescent | Linear | 0 | Ivan  | @codemzs #3396   |   |  Artidoro #3311|
LightGbm | Tree | 0 | Shahab #2886 | Done @codemzs #3397 | Shahab #2729 |  Artidoro #3311|
FastTree | Tree | 0 | Shahab #2970 | Done @codemzs #3398  | Shahab #3035 |  Artidoro #3311|
FastForest | Tree | 0 | Shahab #2970 | Done @codemzs #3399  | Shahab #3035 |  Artidoro #3311|
Field Aware FactorizationMachine | FFM | 0 | Wei-Sheng #3374  | Done Wei-Sheng #3374  | #3312 Wei-Sheng  | Wei-Sheng #3374  |
Generalized AdditiveModels | GAM | 1 | Shahab #2970 | Done @codemzs #3425 | @codemzs   |  Artidoro #3311|
Linear Support VectorMachines | Linear | 2 | Yael | Done #3401 Yael  |   |  Artidoro #3311|
Prior | Baseline | 2 | Aridoro #2510 |  Done  @codemzs #3426  | Artidoro #2510 |  Artidoro #3311|

## MulticlassClassification.Trainers

MulticlassClassification   Trainers | Category | Priority | Docs (1st pass) | Docs (2nd pass) | Samples | Samples (2nd   pass)
-- | -- | -- | -- | -- | -- | --
[...]

PR #3396, opened 2019-04-18
XML documentation for Symbolic Stochastic Gradient Descent Trainer.

codemzs:
towards #2522

### DIFF
--- a/src/Microsoft.ML.Mkl.Components/SymSgdClassificationTrainer.cs
+++ b/src/Microsoft.ML.Mkl.Components/SymSgdClassificationTrainer.cs
@@ -36,8 +36,8 @@ namespace Microsoft.ML.Trainers
     /// </summary>
     /// <remarks>
     /// <format type="text/markdown"><![CDATA[
-    /// To create this trainer, use [SymbolicStochasticGradientDescent](xref:Microsoft.ML.MklComponentsCatalog.SymbolicSgdLogisticRegression(Microsoft.ML.BinaryClassificationCatalog.BinaryClassificationTrainers,System.String,System.String,System.Int32)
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Mkl.Components/SymSgdClassificationTrainer.cs
```diff
@@ -48,6 +48,7 @@ namespace Microsoft.ML.Trainers
     /// | Is normalization required? | Yes |
     /// | Is caching required? | No |
     /// | Required NuGet in addition to Microsoft.ML |Microsoft.ML.Mkl.Components |
+    ///
     /// ### Training Algorithm Details
     /// The symbolic SGD is a classification algorithm that makes its predictions by finding a separating hyperplane.
```
wschin: SGD is not a classification algorithm.

src/Microsoft.ML.Mkl.Components/SymSgdClassificationTrainer.cs
```diff
@@ -48,6 +48,7 @@ namespace Microsoft.ML.Trainers
...
     /// For instance, with feature values $f0, f1,..., f_{D-1}$, the prediction is given by determining what side of the hyperplane the point falls into.
```
wschin: This description doesn't directly reflect the true prediction rule: sign of score is the predicted label. If score > 0, predicted label is true and false otherwise.

wschin: The actual scoring function is $\boldsymbol{w}^T \boldsymbol{x} + b$ but this description only mentions $\boldsymbol{x}$.

codemzs: That is the same as the sign of the features' weighted sum, i.e. $\sum_{i = 0}^{D-1} (w_i * f_i)$, where $w_0, w_1,..., w_{D-1}$ are the weights computed by the algorithm.

---
In reply to: [277147921](https://github.com/dotnet/machinelearning/pull/3396#discussion_r277147921) [](ancestors = 277147921)

Issue #2252, opened 2019-01-26
Lack of support for UWP apps in ML.NET


CESARDELATORRE:
Current version (0.9 and 0.10) don't support UWP apps properly.

See additional details in this Blog Post:
https://xamlbrewer.wordpress.com/2019/01/25/machine-learning-with-ml-net-in-uwp-clustering/ 

Related issues:
https://github.com/dotnet/machinelearning/issues/1736
https://github.com/dotnet/machinelearning/issues/1595

SUGGESTED APPROACH:

Usually, the common scenario for UWP apps (visual desktop applications) is just about scoring a model.
The same common scenario for ARM based platforms like Xamarin on iOS and Android.
(ARM has nothing to do with this #2252 issue, it just happens that we still don't support ARM, neither)

The suggested approach would be to split ML.NET components/NuGet packages so the scoring components are segregated from the rest of ML.NET. 

Achieving support in UWP and ARM just for the "scoring part" of ML.NET might be easier and require less cost/work in testing and development than achieving support for the whole ML.NET (training/test model area).

CESARDELATORRE:
@TomFinley - Since it'll be breaking the compilation, I think it can be considered a breaking change..
Splitting the "scoring" assemblies from the training bits should be done as soon as possible, in my opinion. It will give us a lot more flexibility when incrementally updating/improving in areas like UWP, Unity and ARM support.

sharpwood:
When will ML.NET officially support UWP? UWP is Microsoft's latest development framework. This project does not even support UWP? @TomFinley

XamlBrewer:
The ML.NET v0.11.0 NuGet package largely supports UWP (and that is just awesome). 

There's still one big issue: compiling a Release build fails. (ilc.exe code 1200).

Reproduction steps: 

- create UWP app in Visual Studio 2017, 
- add NuGet package, 
- set to Release mode, 
- compile.

ianier:
Hi @TomFinley, I agree with @sharpwood that UWP should be treated as a first class citizen. In my opinion this issue should be given much higher priority and addressed before RTM.

codemzs:
CC: @abetaha

MattWhilden:
@XamlBrewer I'm able to build with the latest set of tools for UWP. Here's my set up:

- VS 2019 (don't think that matters)
- New UWP project
- Min OS version of at least 16299 (latest .NET Native needs this)
[...]

PR #3402, opened 2019-04-18
Towards #2522 - RandomizedPcaTrainer documentation

yaeldekel:
Adding documentation for Randomized PCA, as specified in #2252.


docs/api-reference/io-columns-anomaly-detection.md
```diff
@@ -0,0 +1,6 @@
+### Input and Output Columns
+The input features column data must be a known-sized vector of <xref:System.Single>. This trainer outputs the following columns:
+
+| Output Column Name | Column Type | Description|
+| -- | -- | -- |
+| `Score` | <xref:System.Single> | The score that was calculated by the anomaly detection model.|
```
shmoradims: >  [](start = 38, length = 1)

if the range is (-inf,+inf) we call them 'unbounded score' to be clear that it's not normalized or anything like that.

src/Microsoft.ML.PCA/PcaTrainer.cs
```diff
@@ -33,12 +33,35 @@ namespace Microsoft.ML.Trainers
...
+    /// | -- | -- |
+    /// | Machine learning task | Anomaly Detection |
+    /// | Is normalization required? | Yes |
+    /// | Is caching required? | No |
+    /// | Required NuGet in addition to Microsoft.ML | None |
+    ///
+    /// ### Training Algorithm Details
+    /// This trainer trains an approximate PCA using a randomized method for computing the singular value decomposition (SVD) of
+    /// the matrix whose rows are the input vectors.
+    /// The method is described [here](https://web.stanford.edu/group/mmds/slides2010/Martinsson.pdf).
```
wschin: My wish is that user can see the scoring rule in this document.

yaeldekel: I added it, but are you sure this is needed? isn't it too much math?

shmoradims:
Duplicate of #3429. Closing for now.

PR #3411, opened 2019-04-18
XML documentation for FastForest regression trainer.

codemzs:
towards #2522

### DIFF
--- a/src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
+++ b/src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
@@ -362,7 +362,7 @@ namespace Microsoft.ML
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
```diff
@@ -362,7 +362,7 @@ public static class TreeExtensions
         }
 
         /// <summary>
-        /// Predict a target using a decision tree regression model trained with the <see cref="FastForestRegressionTrainer"/> and advanced options.
+        /// Create <see cref="FastForestRegressionTrainer"/> using advanced options, which predicts a target using a decision tree regression model.
```
shmoradims: >using [](start = 61, length = 5)

with

Issue #3204, opened 2019-04-04
API reference - XML documentation template for transforms


sfilipi:
The XML documentation for the transforms should contain information about the schema: requirements about the type of the columns to work on, and information about the type of the columns produced. 

@shmoradims

wschin:
I'd suggest to document `GetOutputSchema` function and transformer's XML can reference `GetOutputSchema`'s document.

singlis:
Referencing #3127 as this feels same or related, maybe we could track this work under one issue? I think there are going to be a number of subsections that will need to be tracked. And should this go on the transforms? Or the extensions?

sfilipi:
> I'd suggest to document `GetOutputSchema` function and transformer's XML can reference `GetOutputSchema`'s document.

It won't work, because GetOutputSchema is on the base class, and is calling GetOutputSchemaCore of every class, which is internal. 
Also the types of the columns for the input columns need to be documented.

sfilipi:
Proposal for the transforms XML template, mirroring the trainers template from issue: #3218, 

1- XML  on Transform extension method:
- **Summary** - One line description of what the transform does. 
- **Remarks** - See cref the estimator for more details. 

- **Parameters**: 
inputColumnName - Name of column to transform. If set to <see langword="null"/>, the value of the <paramref name="outputColumnName"/> will be used as source. 
The data type on this column should be `<ref required data type.>` | The data type on this column can be any type.
- **Example**

2 - XML of the Estimator, Transformer 
- **Summary** - One line description of what the transform does.
- **Remarks** - More details about the transform and its implementation.
Does it do a pass through the data? Yes/No
Is convertible to Onnx format? Yes/No
Additional NuGet: "Link to NuGet" OR None of all that are included already in Microsoft.ML
- SeeAlso cref the extension method for an example usage.

natke:
Looks good @sfilipi. One small comment: for the summary of the extension method, would we want to say: "Create an xxx transform ..."?

artidoro:
Shouldn't it say "Create an xxx **estimator**" instead of "transform"?
@sfilipi

sfilipi:
> Shouldn't it say "Create an xxx **estimator**" instead of "transform"?
> @sfilipi

Updated, thanks @Artidoro.

shmoradims:
* Do we need the remarks? Can the summary just say: "Create an <cref the estimator>"?
* Let's add a short description of the transform in the summary, similar to trainers.

[...]

PR #3424, opened 2019-04-19
Towards #3204 -FeatureSelection

sfilipi:
Adhering to the #3204 template for the FeatureSelection estimator and transformer.


src/Microsoft.ML.Transforms/CountFeatureSelection.cs
```diff
@@ -18,7 +18,47 @@
 
 namespace Microsoft.ML.Transforms
 {
-    /// <include file='doc.xml' path='doc/members/member[@name="CountFeatureSelection"]' />
+    /// <summary>
+    ///  Selects the slots for which the count of non-default values is greater than or equal to a threshold.
```
shmoradims: >default values [](start = 54, length = 14)

create new issue for later: we need to specify for each input type what's the default value. for e.g. it's not clear if for text, default is null, empty string, or whitespaces. same goes with key type.

src/Microsoft.ML.Transforms/MutualInformationFeatureSelection.cs
```diff
@@ -19,7 +19,54 @@
...
+    /// |False|0,7,0 |
+    ///
+    /// This is how the dataset above would look, after fitting the estimator, and transforming the data with the resulting transformer:
+    ///
+    /// | Label |  Features |
+    /// | -- | -- |
+    /// |True |4,0 |
+    /// |False|0,5 |
+    /// |True |4,0 |
+    /// |False|0,5 |
```
shmoradims: do we need these examples? we already have similar samples.

src/Microsoft.ML.Transforms/CountFeatureSelection.cs
```diff
@@ -18,7 +18,47 @@
...
+    /// | Output column data type | Same as the input column.|
+    ///
+    /// This transform uses a set of aggregators to count the number of non-default values for each slot(vector element) and
+    /// instantiates a <xref:Microsoft.ML.Transforms.SlotsDroppingTransformer> to actually drop the slots that have the default values for the types
+    /// keeping at least 'count' non-default values.
+    /// This transform is useful when applied together with a <xref:Microsoft.ML.Transforms.OneHotHashEncodingTransformer>.
+    /// The count feature selection can remove the features generated by the hash transform that have no data in the examples.
+    ///
+    /// For example, if we set the count to 3 and fit the estimator, apply the transformer to the following Features column,
+    /// we would see the second solot, containing: NaN, 5, 5, NaN values being dropped because that slot has only two non-default values, the two 5 values.
```
shmoradims: >solot [](start = 32, length = 5)

slot

please use spell checker. we won't catch all the typos manually.
[https://marketplace.visualstudio.com/items?itemName=EWoodruff.VisualStudioSpellChecker](https://marketplace.visualstudio.com/items?itemName=EWoodruff.VisualStudioSpellChecker)

src/Microsoft.ML.Transforms/MutualInformationFeatureSelection.cs
sfilipi: it is really hard to understand without them, i think. 
The samples should be for how to use it, not to understand what is it doing. @natke what do you think?

natke: I like the samples. It helped me understand when I read it.

src/Microsoft.ML.Transforms/CountFeatureSelection.cs
sfilipi: Created: [https://github.com/dotnet/machinelearning/issues/3443](https://github.com/dotnet/machinelearning/issues/3443)

---
In reply to: [277027260](https://github.com/dotnet/machinelearning/pull/3424#discussion_r277027260) [](ancestors = 277027260)

PR #3425, opened 2019-04-19
XML documentation for GAM binary classifier trainer.

codemzs:
towards #2522


docs/api-reference/algo-details-gam.md
```diff
@@ -0,0 +1,18 @@
+### Training Algorithm Details
+Generalized Additive Models, or GAMs, model the data as a set of linearly
+independent features similar to a linear model. For each feature, the GAM
+trainer learns a non-linear function, called a "shape function", that computes
+the response as a function of the feature's value. (In contrast, a linear model
+fits a linear response (e.g. a line) to each feature.) To score an example, the
```
natke: Would "input" be a better choice of word here than "example"?

PR #3430, opened 2019-04-19
XML documentation for FastTree Ranking trainer.

codemzs:
towards #2522

### DIFF
--- a/src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
+++ b/src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
@@ -122,11 +122,11 @@ namespace Microsoft.ML
         }
 
         /// <summary>
-        /// Ranks a series of inputs based on their relevance, training a decision tree ranking model with the <see cref="FastTreeRankingTrainer"/>.
+        /// Creates a <see cref="FastTreeRankingTrainer"/>, which ranks a series of inputs based on their relevance and training a decision tree ranking model.
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
```diff
@@ -122,11 +122,11 @@ public static class TreeExtensions
         }
 
         /// <summary>
-        /// Ranks a series of inputs based on their relevance, training a decision tree ranking model with the <see cref="FastTreeRankingTrainer"/>.
+        /// Create a <see cref="FastTreeRankingTrainer"/>, which ranks a series of inputs based on their relevance and training a decision tree ranking model.
```
natke: Suggest: which ranks a series of inputs based on their relevance, using a decision tree ranking model

src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
```diff
@@ -156,7 +156,7 @@ public static class TreeExtensions
         }
 
         /// <summary>
-        /// Ranks a series of inputs based on their relevance, training a decision tree ranking model with the <see cref="FastTreeRankingTrainer"/> and advanced options.
+        /// Create a <see cref="FastTreeRankingTrainer"/> with advanced options, which ranks a series of inputs based on their relevance and training a decision tree ranking model.
```
natke: Same comment as above. Suggest: which ranks a series of inputs based on their relevance, using a decision tree ranking model

PR #3434, opened 2019-04-19
Cherry-picking - fixing the link (#3406)

sfilipi:
Cherry-picking #3406


artidoro:
@sfilipi  I think we are not doing any cherry picking, and the release branch will be updated to master, so if you want to change the link I would do it directly in master.

shauheen:
Thanks @sfilipi , I will update release one more time.

PR #3435, opened 2019-04-19
Normalizer multicolumn example

sfilipi:
Adding a sample for the bin normalizer multicolumn extenion.


sfilipi:
@shmoradims, @Ivanidzo4ka @natke how do you feel about leaving this samples simple?
The other overload has a ton more information, but maybe there is no need to duplicate all of that?

shmoradims:
> @shmoradims, @Ivanidzo4ka @natke how do you feel about leaving this samples simple?
> The other overload has a ton more information, but maybe there is no need to duplicate all of that?

Let's not spend time creating **new** samples. Our sample coverage is close to 100%. We just have to make sure the **existing** sample from people on vacation are checked in properly.

PR #3446, opened 2019-04-20
Refine descriptions of Microsoft.ML classes

natke:
Update summary description of all Microsoft.ML classes.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
@@ -10,7 +10,8 @@ using Microsoft.ML.Runtime;
 namespace Microsoft.ML
 {
     /// <summary/>
-    /// Collection of extension methods for the <see cref="DataOperationsCatalog"/>.
+    /// Collection of extension methods for the <see cref="DataOperationsCatalog"/> to read from delimited text
+    /// files such as csv and tsv.
     /// <summary>
     public static class TextLoaderSaverCatalog
     {
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
@@ -9,6 +9,9 @@ using Microsoft.ML.Runtime;
 
 namespace Microsoft.ML
 {
+    /// <summary/>
+    /// Collection of extension methods for the <see cref="DataOperationsCatalog"/>.
+    /// <summary>
     public static class TextLoaderSaverCatalog
     {
         /// <summary>
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
@@ -12,7 +12,7 @@ namespace Microsoft.ML
     /// <summary>
     /// Collection of extension methods for the <see cref="DataOperationsCatalog"/> to read from delimited text
     /// files such as csv and tsv.
-    /// <summary/>
+    /// </summary>
     public static class TextLoaderSaverCatalog
     {
         /// <summary>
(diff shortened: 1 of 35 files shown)


shmoradims:
>         /// </summary>

we also need to standardize this

---
Refers to: src/Microsoft.ML.Data/TrainCatalog.cs:645 in fda7183. [](commit_id = fda7183971d477d2b25cadc0ceaf3f4465faf5eb, deletion_comment = False)

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -635,7 +635,8 @@ internal RankingTrainers(RankingCatalog catalog)
     }
 
     /// <summary>
-    /// The central catalog for anomaly detection tasks and trainers.
+    /// Class used to create instances of anomaly detection components,
```
shmoradims: >Class [](start = 8, length = 5)

Suggestion: we somehow need to convey to the user that this class is not directly used by the user, and they don't create it directly. Just a suggestion.

natke: That's a good idea. I could reference MLContext

natke: Class used **by an MLContext object** to create ...?

src/Microsoft.ML.Data/DebuggerExtensions.cs
```diff
@@ -9,7 +9,7 @@
 namespace Microsoft.ML
 {
     /// <summary>
-    /// Static extensions for data preview.
+    /// Class used to create instances of preview objects for debugging.
```
shmoradims: >debugging. [](start = 62, length = 10)

please explicitly explain that this class and all the methods here are only used for debugging and should **not** be used in production code.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -9,6 +9,9 @@
 
 namespace Microsoft.ML
 {
+    /// <summary/>
+    /// Collection of extension methods for the <see cref="DataOperationsCatalog"/>.
```
glebuk: Well, this is one "Collection of extension methods for the <…. catalog> to read from delimited text files such as CSV TSV"

shmoradims:
Awesome. Thanks for doing this. :)

src/Microsoft.ML.Data/MLContext.cs
```diff
@@ -10,8 +10,9 @@
 namespace Microsoft.ML
 {
     /// <summary>
-    /// The <see cref="MLContext"/> is a starting point for all ML.NET operations. It is instantiated by user,
-    /// provides mechanisms for logging and entry points for training, prediction, model operations etc.
+    /// The starting point for all ML.NET applications. Once instantiated by the user, it
```
glebuk: ML.NET Operations is more correct.  (Applicatoin is a hosting app that contains ML.NET components.)
"The common context for all ML.NET operations.  Once instantiated by the user it provides a way to create variety of components for data preparation.... etc.  It also allows logging, execution control and ability set repeatable random numbers

src/Microsoft.ML.Data/Transforms/ExtensionsCatalog.cs
```diff
@@ -11,7 +11,7 @@
 namespace Microsoft.ML
 {
     /// <summary>
-    /// Specifies input and output column names for a transformation.
+    /// Specifies input and output column names for transformer components that operate on multiple input/output columns.
```
glebuk: columns.  It is often advantageous to transform several columns at once as all of the changes can be done in a single data pass.

src/Microsoft.ML.Transforms/CustomMappingCatalog.cs
```diff
@@ -10,7 +10,8 @@
 namespace Microsoft.ML
 {
     /// <summary>
-    /// Extension methods for custom mapping transformers.
+    /// Class containing an extension method for <see cref="TransformsCatalog"/> to create instances of
+    /// custom mapping transformer components.
```
glebuk: /// user-defined one-to-one row mapping transformer components.

src/Microsoft.ML.Transforms/NormalizerCatalog.cs
```diff
@@ -10,7 +10,8 @@
 namespace Microsoft.ML
 {
     /// <summary>
-    /// Extensions for normalizer operations.
+    /// Collection of extension methods for <see cref="TransformsCatalog"/> to create instances of normalization
```
glebuk: to create instances of **numerical** normalization.

Issue #206, opened 2018-05-22
Taxi fare dataset is almost 50MB


Ivanidzo4ka:
These two files is almost 50mb altogether
https://github.com/dotnet/machinelearning/blob/master/test/data/taxi-fare-test.csv
https://github.com/dotnet/machinelearning/blob/master/test/data/taxi-fare-train.csv

https://github.com/dotnet/machinelearning/pull/170 allows to download files from external sources. Can we move these files to separate repository and clean history?

TomFinley:
Hmmm. This means an interactive rebase of `master`. What a nightmare. If we have to we have to I guess.

Would have been nice had this been caught in PR. Maybe whatever bot we're using to validate/build releases can flag a PR if it contains huge files like this, going forward.

justinormont:
I'd checkout: https://help.github.com/articles/removing-sensitive-data-from-a-repository/

TomFinley:
Hi Justin! Yup, I know how to do it more or less, but here's what I'm imagining. One way or another this will involve repointing master to a completely different commit id. Once that happens, people that are attempting to write PRs against master will have an ***interesting*** experience when they attempt to merge master. (I think rebases might be fine.)

Which is why I'd love to get to the point where there's some check. The people that approved that PR... who knows what they were thinking. But I'd love to give the people that have the power to approve the PR some "help" so they get a hint that not all is as it should be.

codemzs:
@OliaG @aditidugar  These datasets were [checked-in](https://github.com/dotnet/machinelearning/pull/45/files) by you. Can you please clarify why you need nearly 2 million rows for training a sample?

shauheen:
@TomFinley that PR was right before the release, they should not have been merged with that size, however now they are there and we will find a way to clean them up.

aditidugar:
Yep, @shauheen covered it - we can certainly trim this down if necessary, it wasn't something that we consciously considered before the initial check in.

Ivanidzo4ka:
Out pure curiosity @eerhardt if we rebase this repo, what would happen to devdiv VSO?

Ivanidzo4ka:
Should we create another repository like "dotnet/machinelearning/datasets" and store this files there?
I have code to download files and put them into repo, but I need place to keep this files, since many of them either behind authorization page, or get slightly modified in order to make them more readable.

terrajobst:
[...]

PR #3458, opened 2019-04-22
[AutoML] Add AutoML example code

daholste:


docs/samples/Microsoft.ML.AutoML.Samples/Data/wikipedia-detox-250-line-test.tsv
```diff
@@ -0,0 +1,19 @@
+Sentiment	SentimentText
+1	  ==You're cool==  You seem like a really cool guy... *bursts out laughing at sarcasm*.
```
justinormont: This same line is in the training set.

docs/samples/Microsoft.ML.AutoML.Samples/Data/wikipedia-detox-250-line-data.tsv
```diff
@@ -0,0 +1,250 @@
+Sentiment	SentimentText
```
justinormont: This file already exists in the repo:
https://github.com/dotnet/machinelearning/blob/7b59ce5031769a75f07cc4daeeeb9fc60ee53177/test/data/wikipedia-detox-250-line-data.tsv#L2

daholste:
> We should reference the existing datasets in the repo.
> 
> For example, the Taxi dataset is 24MB, and adding another copy into the history of the repo is not good. This makes `git clone ...` very slow.

Thanks for reviewing, @justinormont . I was using a smaller taxi dataset not already in the repo. Also, since opt-digits dataset wasn't in the repo, the only existing dataset was wikidetox, which is some kilobytes and was duplicated locally for convenience.
(I see that in the samples repo, for instance, larger datasets are duplicated between csharp and fsharp in the same way for convenience.)
However, I think you're totally right here. Let's aim to use test/data as a central repo for datasets. Updated

docs/samples/Microsoft.ML.AutoML.Samples/BinaryClassificationExperiment.cs
```diff
@@ -0,0 +1,61 @@
...
+            // STEP 2: Run AutoML experiment
+            Console.WriteLine($"Running AutoML binary classification experiment for {ExperimentTime} seconds...");
+            ExperimentResult<BinaryClassificationMetrics> experimentResult = mlContext.Auto()
+                .CreateBinaryClassificationExperiment(ExperimentTime)
+                .Execute(trainDataView);
+
+            // STEP 3: Print metric from the best model
+            RunDetail<BinaryClassificationMetrics> bestRun = experimentResult.BestRun;
+            Console.WriteLine($"Total models produced: {experimentResult.RunDetails.Count()}");
+            Console.WriteLine($"Best model's trainer: {bestRun.TrainerName}");
```
justinormont: I'm not sure how to best do it, but printing the featurization would be nice.

docs/samples/Microsoft.ML.AutoML.Samples/Microsoft.ML.AutoML.Samples.csproj
```diff
@@ -0,0 +1,46 @@
...
+
+  <ItemGroup>
+    <Folder Include="Data\" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Include="..\..\..\test\data\optdigits-test.csv" Link="Data\optdigits-test.csv">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Include="..\..\..\test\data\optdigits-train.csv" Link="Data\optdigits-train.csv">
```
justinormont: Can we reference the dataset thru the Dataset Utils? Given its part of examples, I expect it's accessible thru nugets. 

We could add the optdigits to SamplesDatasetUtils.cs:
https://github.com/dotnet/machinelearning/blob/fde1ab718bcf0da10293e5ca32245373e4669e2a/src/Microsoft.ML.SamplesUtils/SamplesDatasetUtils.cs#L18-L24

docs/samples/Microsoft.ML.AutoML.Samples/MulticlassClassificationExperiment.cs
```diff
@@ -0,0 +1,62 @@
...
+            Console.WriteLine($"Running AutoML multiclass classification experiment for {ExperimentTime} seconds...");
+            ExperimentResult<MulticlassClassificationMetrics> experimentResult = mlContext.Auto()
+                .CreateMulticlassClassificationExperiment(ExperimentTime)
+                .Execute(trainDataView, LabelColumnName);
+
+            // STEP 3: Print metric from the best model
+            RunDetail<MulticlassClassificationMetrics> best = experimentResult.BestRun;
+            Console.WriteLine($"Total models produced: {experimentResult.RunDetails.Count()}");
+            Console.WriteLine($"Best model's trainer: {best.TrainerName}");
+            Console.WriteLine($"AccuracyMacro of best model from validation data: {best.ValidationMetrics.MacroAccuracy}");
```
justinormont: Why MacroAccuracy? Generally, I'd just print all metrics.

If we're going to display MacroAccracy, I'd recommend training towards it (or a proxy metric). Macro & Micro accuracy are sometimes at odds, for instance adding a weight column to balance the effect of each class tends to increase macro accuracy at the cost of micro.

docs/samples/Microsoft.ML.AutoML.Samples/BinaryClassificationExperiment.cs
daholste: Great point. +1 that not sure how to do it. Let's think through this sometime together, and then circle back & add it

docs/samples/Microsoft.ML.AutoML.Samples/MulticlassClassificationExperiment.cs
daholste: You're right -- I copied this part over from the old repo. Thanks. Let me print MicroAccuracy, since we're sweeping to optimize it.
Re: printing all metrics -- in our draft into the samples repo, where we assume the user will execute the sample code, we have a helper function that prints all the metrics like you suggest. The example code here will appear in the API docs for the user to view, but may not be targeted for stand-alone runs. Given this, do you think it would be helpful to print all metrics? If so, I can go ahead and do this for the samples here

docs/samples/Microsoft.ML.AutoML.Samples/Microsoft.ML.AutoML.Samples.csproj
daholste: Yeah, the opt-digits dataset is part of ML.NET samples. However, DatasetUtils doesn't download the dataset from NuGet, but instead pulls it from the ML.NET samples Git repository. You're right -- since this dataset is already in one Git repository (the ML.NET samples Git repo), no need to add it to another Git repository (the ML.NET one), too. Of course, questions about inconsistency come to mind (like, why have 2 out of 3 datasets for these samples (including two 25 MB ones), checked in directly, and download the other dataset (< 1MB) from another Git repo owned by our team), but I'm okay with that. Long-term, it'd be great to nail down a solid story around how we store datasets ofc. (Are datasets housed in their own repo? In Git LFS?  Etc)

daholste:
> LGTM, though I'd remove the optdigits dataset, and reference it thru the dataset utils. Currently, this PR needlessly adds 825KB to the repo.
"Needlessly" may be a bit strong. DatasetUtils actually downloads the dataset from another ML.NET-owned Git repo (the samples Git repository). Although, you're right -- since this dataset is already in one Git repository (the ML.NET samples Git repo), no need to add it to another Git repository (the ML.NET one), too. Of course, questions about inconsistency come to mind (like, why have the datasets for 2 out of 3 samples here (including two 25 MB files), checked in directly, and download the other dataset (< 865 KB) from another Git repo owned by our team), but I'm okay with that. Long-term, it'd be great to nail down a solid story around how we store datasets ofc. (Are datasets housed in their own repo? In Git LFS? Etc)

docs/samples/Microsoft.ML.AutoML.Samples/Microsoft.ML.AutoML.Samples.csproj
justinormont: Correct. I was wondering if the DatasetUtils class is public in the nuget. The dataset files are downloaded via https. 

Longer term, I'm a fan of Git LFS, but we haven't set it up for the repo. There's an ok argument to keep all datasets in the samples repo too. Besides for unit test datasets which tend to be trivially small, I like having the main repo be rather slim by keeping datasets elsewhere. 

The DatasetUtils method also lets us reference larger datasets for our examples.

docs/samples/Microsoft.ML.AutoML.Samples/MulticlassClassificationExperiment.cs
justinormont: How obvious is helper function? For clarity, we could have the 4 print statements directly.

I like seeing training time for my runs; do we pass that back to the user?

justinormont:
You're correct, "needlessly" is a bit too strong. My apologies.

It's easy to add the file later if we want to, but it's nearly impossible to remove it. Hence why the HelperUtils method is better for the moment. 

The taxi dataset shouldn't have been checked into the repo ([background](https://github.com/dotnet/machinelearning/issues/206)); it's currently ~40% of the repo's size. Even if you delete the file, it's still in the git history. It is possible to [rewrite the git history](https://help.github.com/en/articles/removing-sensitive-data-from-a-repository), but there's good reasons to not do that.

docs/samples/Microsoft.ML.AutoML.Samples/MulticlassClassificationExperiment.cs
daholste: rev-ed multiclass sample https://github.com/dotnet/machinelearning/blob/4b5bfd5eb69e9026b07bb98f18e3ed2268094e45/docs/samples/Microsoft.ML.AutoML.Samples/MulticlassClassificationExperiment.cs#L43
I can update metrics printing for remaining samples if that looks alright
I agree that printing iteration time for each model would be nice to have. Let's brainstorm together with Vinod all the other things the samples here should include, and what they should look like

daholste: for the sake of time, going to iterate on metrics printing for Regression & Binary in the same way, and then check-in. if there's a different way that would be preferable, will submit a PR to update

daholste:
Thank you, as always, for the careful review, solid comments, and deep care about the ML.NET product, @justinormont 👍 . Thanks also for catching / preventing repo overload

PR #3484, opened 2019-04-22
[AutoML] API docs for automl experiment

vinodshanbhag:
Fixing API docs for automl experiment


src/Microsoft.ML.Auto/API/AutoCatalog.cs
```diff
@@ -102,7 +102,7 @@ public MulticlassClassificationExperiment CreateMulticlassClassificationExperime
         }
 
         /// <summary>
-        /// Creates a new AutoML experiment to run on a binary classification dataset.
+        /// Creates a new AutoML experiment to run on a multiclass classification dataset.
```
daholste: thank you!

Issue #3474, opened 2019-04-22
fix paths in documentation


codemzs:
# Incorrect sample path

File | Message | Fix
-- | -- | --
dotnet/api/Microsoft.ML.BinaryClassificationCatalog.CalibratorsCatalog.yml | Cannot resolve '~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/BinaryClassification/Calibrators/Isotonic.cs' relative to 'api/Microsoft.ML.BinaryClassificationCatalog.CalibratorsCatalog.yml'. |  
dotnet/api/Microsoft.ML.BinaryClassificationCatalog.CalibratorsCatalog.yml | Cannot resolve '~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/BinaryClassification/Calibrators/Naive.cs' relative to 'api/Microsoft.ML.BinaryClassificationCatalog.CalibratorsCatalog.yml'. |  
dotnet/api/Microsoft.ML.BinaryClassificationCatalog.CalibratorsCatalog.yml | Cannot resolve '~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/BinaryClassification/Calibrators/FixedPlatt.cs' relative to 'api/Microsoft.ML.BinaryClassificationCatalog.CalibratorsCatalog.yml'. |  
dotnet/api/Microsoft.ML.BinaryClassificationCatalog.CalibratorsCatalog.yml | Cannot resolve '~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/BinaryClassification/Calibrators/Platt.cs' relative to 'api/Microsoft.ML.BinaryClassificationCatalog.CalibratorsCatalog.yml'. |  
dotnet/api/Microsoft.ML.Transforms.NormalizingTransformer.AffineNormalizerModelParameters-1.yml | Cannot resolve '~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Normalizer.cs' relative to 'api/Microsoft.ML.Transforms.NormalizingTransformer.AffineNormalizerModelParameters-1.yml'. |  
[...]

PR #3522, opened 2019-04-23
Fix samples link in XML documentation.

codemzs:
towards #3474

### DIFF
--- a/src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
+++ b/src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
@@ -643,7 +643,7 @@ namespace Microsoft.ML
         /// <example>
         /// <format type="text/markdown">
         /// <![CDATA[
-        ///  [!code-csharp[PoissonRegression](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/MulticlassClassification/LbfgsPoissonMaximumEntropy.cs)]
+        ///  [!code-csharp[PoissonRegression](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/MulticlassClassification/LbfgsMaximumEntropy.cs)]
         /// ]]></format>
         /// </example>
         public static LbfgsMaximumEntropyMulticlassTrainer LbfgsMaximumEntropy(this MulticlassClassificationCatalog.MulticlassClassificationTrainers catalog,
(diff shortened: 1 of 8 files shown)


src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
```diff
@@ -643,7 +643,7 @@ public static LbfgsPoissonRegressionTrainer LbfgsPoissonRegression(this Regressi
         /// <example>
         /// <format type="text/markdown">
         /// <![CDATA[
-        ///  [!code-csharp[PoissonRegression](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/MulticlassClassification/LbfgsPoissonMaximumEntropy.cs)]
+        ///  [!code-csharp[PoissonRegression](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/MulticlassClassification/LbfgsMaximumEntropy.cs)]
```
shmoradims: >PoissonRegression [](start = 27, length = 17)

nit: Lbfgs

src/Microsoft.ML.TensorFlow/TensorFlowModel.cs
```diff
@@ -56,13 +56,6 @@ public DataViewSchema GetInputSchema()
         /// <param name="outputColumnName">The name of the requested model output.</param>
         /// <param name="addBatchDimensionInput">Add a batch dimension to the input e.g. input = [224, 224, 3] => [-1, 224, 224, 3].
         /// This parameter is used to deal with models that have unknown shape but the internal operators in the model require data to have batch dimension as well.</param>
-        /// <example>
-        /// <format type="text/markdown">
-        /// <![CDATA[
-        /// [!code-csharp[ScoreTensorFlowModel](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlowTransform.cs)]
-        /// ]]>
-        /// </format>
-        /// </example>
```
shmoradims: please check with ZeeshanA. TextClassification.cs seems to be the correct sample for this API.

src/Microsoft.ML.TensorFlow/TensorFlowModel.cs
```diff
@@ -56,13 +56,6 @@ public DataViewSchema GetInputSchema()
         /// <param name="outputColumnName">The name of the requested model output.</param>
         /// <param name="addBatchDimensionInput">Add a batch dimension to the input e.g. input = [224, 224, 3] => [-1, 224, 224, 3].
         /// This parameter is used to deal with models that have unknown shape but the internal operators in the model require data to have batch dimension as well.</param>
-        /// <example>
-        /// <format type="text/markdown">
-        /// <![CDATA[
-        /// [!code-csharp[ScoreTensorFlowModel](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlowTransform.cs)]
```
natke: We don't have this sample?

src/Microsoft.ML.TensorFlow/TensorFlowModel.cs
shmoradims: /cc @zeahmed

codemzs: Yes but it is more array of input/output column and this API is just for single input/output column pair.

src/Microsoft.ML.TensorFlow/TensorFlowModel.cs
codemzs: We didn't but I just modified an existing sample so that we can have a sample for this API :)

PR #3534, opened 2019-04-23
Polish some options' XML document strings

### LINKED ISSUE #3507: Inheritance of LbfgsLogisticRegressionBinaryTrainer.Options class doesn't have links
Inheritance
Microsoft.ML.Trainers.LbfgsTrainerBase`3+OptionsBaseLbfgsLogisticRegressionBinaryTrainer.Options

wschin:
Fix #3507.

Note that some options don't have API references because they are not used to create trainers. For example, we cannot create SdcaMulticlassTrainerBase because it's a base class.

### DIFF
--- a/src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LbfgsPredictorBase.cs
+++ b/src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LbfgsPredictorBase.cs
@@ -25,7 +25,6 @@ namespace Microsoft.ML.Trainers
     {
         /// <summary>
         /// Options for the <see cref="LbfgsTrainerBase{TOptions, TTransformer, TModel}"/>.
-        /// It is used in <a href='https://en.wikipedia.org/wiki/Limited-memory_BFGS'>L-BFGS</a>-based trainers.
         /// </summary>
         public abstract class OptionsBase : TrainerInputBaseWithWeight
         {
--- a/src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LbfgsPredictorBase.cs
+++ b/src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LbfgsPredictorBase.cs
@@ -24,7 +24,8 @@ namespace Microsoft.ML.Trainers
       where TOptions : LbfgsTrainerBase<TOptions, TTransformer, TModel>.OptionsBase, new()
     {
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LbfgsPredictorBase.cs
```diff
@@ -24,7 +24,8 @@ public abstract class LbfgsTrainerBase<TOptions, TTransformer, TModel> : Trainer
       where TOptions : LbfgsTrainerBase<TOptions, TTransformer, TModel>.OptionsBase, new()
     {
         /// <summary>
-        /// Base options for <a href='https://en.wikipedia.org/wiki/Limited-memory_BFGS'>L-BFGS</a>-based trainers.
+        /// Options for the <see cref="LbfgsTrainerBase{TOptions, TTransformer, TModel}"/>.
+        /// It is used in <a href='https://en.wikipedia.org/wiki/Limited-memory_BFGS'>L-BFGS</a>-based trainers.
```
natke: Would it be better to have this wikipedia link on the trainer itself?

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LbfgsPredictorBase.cs
```diff
@@ -24,7 +24,8 @@ public abstract class LbfgsTrainerBase<TOptions, TTransformer, TModel> : Trainer
       where TOptions : LbfgsTrainerBase<TOptions, TTransformer, TModel>.OptionsBase, new()
     {
         /// <summary>
-        /// Base options for <a href='https://en.wikipedia.org/wiki/Limited-memory_BFGS'>L-BFGS</a>-based trainers.
+        /// Options for the <see cref="LbfgsTrainerBase{TOptions, TTransformer, TModel}"/>.
```
yaeldekel: >Options [](start = 12, length = 7)

Suggestion:
`Base Options class for trainer estimators deriving from ...`.
(For the SDCA base class too).

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LbfgsPredictorBase.cs
wschin: I kind of feel adding link to this abstract class could make users thinking they can create it.

PR #3545, opened 2019-04-23
Fix the ApproximatedKernelMappingEstimator XML doc

### LINKED ISSUE #3544: Fix the ApproximatedKernelMappingEstimator XML doc
The mathematical expression for the kernel functions is written in XML and the input and output types should be mentioned in the extension.

artidoro:
Fixes #3544.

### DIFF
--- a/src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
+++ b/src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
@@ -73,10 +73,10 @@ namespace Microsoft.ML.Transforms
     /// <remarks>
     /// <format type="text/markdown"><![CDATA[
     /// - The Gaussian kernel is defined as $k(x,y) = \exp(-\gamma||x-y||^2)$.
-    /// - The Fourier transform of this kernel is the [Normal](https://en.wikipedia.org/wiki/Normal_distribution) distribution with variance $2 \gamma$.
     /// ]]></format>
     /// </remarks>
-
+    // The Fourier transform of this kernel is the <a href="https://en.wikipedia.org/wiki/Normal_distribution">Normal</a> distribution with variance $2 \gamma$
+    // (<a href="https://en.wikipedia.org/wiki/Bochner%27s_theorem">Bochner's theorem</a>)).
     public sealed class GaussianKernel : KernelBase
     {
         internal sealed class Options : IComponentFactory<KernelBase>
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -68,9 +68,15 @@ internal abstract class FourierRandomNumberGeneratorBase
...
 
     /// <summary>
-    /// The Gaussian kernel is defined as k(x,y)=exp(-gamma*|x-y|_2^2). The distribution that is the Fourier transform of
-    /// this kernel is the Normal distribution with variance 2*gamma.
+    /// The Gaussian kernel.
     /// </summary>
+    /// <remarks>
+    /// <format type="text/markdown"><![CDATA[
+    /// The Gaussian kernel is defined as $k(x,y) = \exp(-\gamma||x-y||^2)$. The distribution that is the Fourier transform of
+    /// this kernel is the Normal distribution with variance $2 \gamma$.
```
wschin: What does this mean?
```The distribution that is the Fourier transform of this kernel is the Normal distribution with variance $2 \gamma$.```

File: src/Microsoft.ML.Transforms/FourierDistributionSampler.cs [hunk 2]
```diff
@@ -183,9 +189,14 @@ public override float Next(Random rand)
     }
 
     /// <summary>
-    /// The Laplacian kernel is defined as k(x,y)=exp(-a*|x-y|_1). The distribution that is the Fourier transform of this
-    /// kernel is the Cauchy distribution with parameters (0, a).
+    /// The Laplacian kernel.
     /// </summary>
+    /// <remarks>
+    /// <format type="text/markdown"><![CDATA[
+    /// The Laplacian kernel is defined as $k(x,y)=\exp(-a||x-y||)$. The distribution that is the Fourier transform of
```
wschin: The second sentence is not quite clear to me.

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
```diff
@@ -183,9 +189,14 @@ public override float Next(Random rand)
...
+    /// this kernel is the Cauchy distribution with parameters $(0, a)$.
```
wschin: Please add a link to Cauchy distribution.

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
yaeldekel: This refers to the comment on the base class: that the Fourier transform of a continuous positive definite kernel is a non-negative measure. 

Perhaps we should mention this in the documentation of each implementation? Alternatively, we can remove the sentences mentioning the distributions, since the methods on these classes that sample from the distributions are internal anyway.

What do you guys think?

---
In reply to: [277855942](https://github.com/dotnet/machinelearning/pull/3545#discussion_r277855942) [](ancestors = 277855942)

shmoradims: it's not clear to me either. please reword.

---
In reply to: [277889125](https://github.com/dotnet/machinelearning/pull/3545#discussion_r277889125) [](ancestors = 277889125,277855942)

src/Microsoft.ML.Transforms/FourierDistributionSampler.cs
shmoradims: +1. please reword the second sentence about fourier transform.

---
In reply to: [277856329](https://github.com/dotnet/machinelearning/pull/3545#discussion_r277856329) [](ancestors = 277856329)

PR #3546, opened 2019-04-23
Fix Time Series XML Doc

### LINKED ISSUE #3517: Time series documentation should follow transform template
As title.

wschin:
Fix #3517 by

1.  Change Trainer Template to Transformer Template
2.  Decompose scoring paragraph into two files so that Spike Detection and Change Point Detection can reference files independently.
3.  Some minor polishment.

### DIFF
--- a/docs/api-reference/io-time-series-change-point.md
+++ b/docs/api-reference/io-time-series-change-point.md
@@ -1,7 +1,10 @@
 ### Input and Output Columns
-There is only one input column and its type is <xref:System.Single>.
-This estimator adds the following output columns:
+There is only one input column and produces a column typed to a 4-element vector.
+The output vector sequentially contains alert level (non-zero value means a change point), score, p-value, and martingale value.
 
-| Output Column Name | Column Type | Description|
-| -- | -- | -- |
-| `Prediction` | 4-element vector of <xref:System.Double> | It sequentially contains alert level (non-zero value means a change point), score, p-value, and martingale value. |
+###  Estimator Characteristics 
+|  |  | 
+| -- | -- | 
+| Does this estimator need to look at the data to train its parameters? | No | 
+| Input column data type | Vector of <xref:System.Single> | 
+| Output column data type | 4-element vector of <xref:System.Double> |
--- a/docs/api-reference/io-time-series-spike.md
+++ b/docs/api-reference/io-time-series-spike.md
@@ -1,7 +1,10 @@
 ### Input and Output Columns
-There is only one input column and its type is <xref:System.Single>.
-This estimator adds the following output columns:
+There is only one input column and produces a column typed to a 3-element vector.
+The output vector sequentially contains alert level (non-zero value means a change point), score, and p-value.
 
-| Output Column Name | Column Type | Description|
-| -- | -- | -- |
-| `Prediction` | 3-element vector of <xref:System.Double> | It sequentially contains alert level (non-zero value means a change point), score, and p-value. |
+###  Estimator Characteristics 
+|  |  | 
+| -- | -- | 
+| Does this estimator need to look at the data to train its parameters? | No | 
+| Input column data type | Vector of <xref:System.Single> | 
+| Output column data type | 3-element vector of <xref:System.Double> |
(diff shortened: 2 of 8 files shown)


docs/api-reference/io-time-series-change-point.md
```diff
@@ -1,7 +1,10 @@
 ### Input and Output Columns
-There is only one input column and its type is <xref:System.Single>.
-This estimator adds the following output columns:
+There is only one input column and produces a column typed to a 4-element vector.
```
natke: What are the constraints on the input column?

docs/api-reference/io-time-series-change-point.md
```diff
@@ -1,7 +1,10 @@
...
+The output vector sequentially contains alert level (non-zero value means a change point), score, p-value, and martingale value.
 
-| Output Column Name | Column Type | Description|
-| -- | -- | -- |
-| `Prediction` | 4-element vector of <xref:System.Double> | It sequentially contains alert level (non-zero value means a change point), score, p-value, and martingale value. |
+###  Estimator Characteristics 
+|  |  | 
+| -- | -- | 
+| Does this estimator need to look at the data to train its parameters? | No |
```
natke: Is that correct?

docs/api-reference/io-time-series-change-point.md
natke: Separate input and output into two sentences

docs/api-reference/io-time-series-spike.md
```diff
@@ -1,7 +1,10 @@
 ### Input and Output Columns
-There is only one input column and its type is <xref:System.Single>.
-This estimator adds the following output columns:
+There is only one input column and produces a column typed to a 3-element vector.
```
natke: Separate input and output into two sentences

docs/api-reference/time-series-pvalue.md
```diff
@@ -0,0 +1,12 @@
+### Anomaly Scorer
+Once the raw score at a timestamp is computed, it is fed to the anomaly scorer component to calculate the final anomaly score at that timestamp.
+
+#### Spike detection based on p-value
+The p-value score indicates the p-value of the current computed raw score according to a distribution of raw scores.
```
natke: Suggest leading with the interpretation of the p-value. Something like:
```
The p-value score indicates whether the current point is an outlier (also known as a spike). The lower its value, the more likely it is a spike. The p-value score is always in $[0, 1]$.

This score is the p-value of the current computed raw score according to a distribution of raw scores. Here, the distribution is estimated based on the most recent raw score values up to certain depth back in the history. More specifically, this distribution is estimated using [kernel density estimation (https://en.wikipedia.org/wiki/Kernel_density_estimation) with the Gaussian [kernels](https://en.wikipedia.org/wiki/Kernel_(statistics)#In_non-parametric_statistics) of adaptive bandwidth.

If the p-value score exceeds $1 - \frac{\text{confidence}}{100}$, the associated timestamp may get a non-zero alert value in spike detection, which means a spike point is detected.
Note that $\text{confidence}$ is defined in the signatures of [DetectIidSpike](xref:Microsoft.ML.TimeSeriesCatalog.DetectIidSpike(Microsoft.ML.TransformsCatalog,System.String,System.String,System.Int32,System.Int32,Microsoft.ML.Transforms.TimeSeries.AnomalySide))
and [DetectSpikeBySsa](xref:Microsoft.ML.TimeSeriesCatalog.DetectSpikeBySsa(Microsoft.ML.TransformsCatalog,System.String,System.String,System.Int32,System.Int32,System.Int32,System.Int32,Microsoft.ML.Transforms.TimeSeries.AnomalySide,Microsoft.ML.Transforms.TimeSeries.ErrorFunction)).
```

docs/api-reference/io-time-series-change-point.md
wschin: Will do
```
There is only one input column.
The input column must be <xref:System.Single> where a <xref:System.Single> value indicates a value at a timestamp in the time series.
```

---
In reply to: [278258149](https://github.com/dotnet/machinelearning/pull/3546#discussion_r278258149) [](ancestors = 278258149)

docs/api-reference/io-time-series-change-point.md
wschin: Wrong for SSA.

---
In reply to: [278258440](https://github.com/dotnet/machinelearning/pull/3546#discussion_r278258440) [](ancestors = 278258440)

PR #3553, opened 2019-04-23
Renaming CI legs

Anipik:
The previous pr was alwayus using the mac machines with missing dependencies so just creating a new pr to avoid those failures.

@eerhardt is it fine to run centos just with netcoreapp3.0 and ubuntu with netcoreapp2.1 ?

### DIFF
--- a/.vsts-dotnet-ci.yml
+++ b/.vsts-dotnet-ci.yml
@@ -22,7 +22,7 @@ phases:
         _includeBenchmarkData: false
       Release_Build:
         _configuration: Release-Intrinsics
-        _config_short: R
+        _config_short: RI
         _includeBenchmarkData: true
     queue:
       name: Hosted Ubuntu 1604
--- a/.vsts-dotnet-ci.yml
+++ b/.vsts-dotnet-ci.yml
@@ -13,31 +13,15 @@ resources:
 phases:
 - template: /build/ci/phase-template.yml
   parameters:
-    name: core30
-    buildScript: build.cmd
-    customMatrixes:
-      Build_Debug_Intrinsics:
-        _configuration: Debug-Intrinsics
-        _config_short: DI
-        _includeBenchmarkData: false
-      Build_Release_Intrinsics:
-        _configuration: Release-Intrinsics
-        _config_short: RI
-        _includeBenchmarkData: true
-    queue:
-      name: Hosted VS2017
-
-- template: /build/ci/phase-template.yml
-  parameters:
-    name: Centos
+    name: Centos_x64_Netcoreapp30
     buildScript: ./build.sh
     customMatrixes:
-      Build_Debug_Intrinsics:
+      Debug_Build:
         _configuration: Debug-Intrinsics
         _config_short: DI
         _includeBenchmarkData: false
-      Build_Release:
-        _configuration: Release
+      Release_Build:
+        _configuration: Release-Intrinsics
         _config_short: R
         _includeBenchmarkData: true
     queue:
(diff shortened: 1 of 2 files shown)


Anipik:
@artidoro the mac ci legs are still failing

artidoro:
Looks a lot better already!

I have the following comments:
- Could you Pascal case the names? Netcoreapp -> NetCoreApp
- Could you remove the app from the name? NetCoreApp -> NetCore
- Could you do the same for netfx? netfx -> NetFx
- Could you specify which version of NetFx we are using? (if you think it's useful)

artidoro:
On a separate note, is it possible to add a period in the name? 
For example NetCore_3.0 or NetCore3.0 instead of NetCore30.

Anipik:
>  Could you remove the app from the name? NetCoreApp -> NetCore

Actually eric told me to keep that as we dont use netcore anywhere

> Could you do the same for netfx? netfx -> NetFx

Its actually azure thats changing it automatiically from uppercase to lowercase 

> Could you specify which version of NetFx we are using? (if you think it's useful)

Its not that useful but I can add that, as it may fix our lowercase problem

artidoro:
It is strange that it is not keeping the Pascal case on NetCoreApp!

Anipik:
> It is strange that it is not keeping the Pascal case on NetCoreApp!

It does weird things  :P

.vsts-dotnet-ci.yml
```diff
@@ -13,71 +13,69 @@ resources:
...
-      Build_Debug_Intrinsics:
+      Debug_Build:
         _configuration: Debug-Intrinsics
         _config_short: DI
         _includeBenchmarkData: false
-      Build_Release_Intrinsics:
+      Release_Build:
         _configuration: Release-Intrinsics
-        _config_short: RI
+        _config_short: R
```
danmosemsft: RI?

.vsts-dotnet-ci.yml
```diff
@@ -13,71 +13,69 @@ resources:
...
+    buildScript: build.cmd
     queue:
-      name: Hosted macOS
+      name: Hosted VS2017
 
 - template: /build/ci/phase-template.yml
   parameters:
-    name: Windows_NetFx
+    name: Windows_x64_NetFx461
     buildScript: build.cmd
```
danmosemsft: explicit `architecture: x64` ?

Anipik:
> Next time if you can avoid reordering, the diff is easier to read.

Actually ML team wanted these reorderings. but yeah i will split the reordering into a separate commit from next time

.vsts-dotnet-ci.yml
Anipik: Actually the default value of this architecture variable is x64 which we set here https://github.com/dotnet/machinelearning/pull/3553/files/251954988b8ebd8e2dd8c1eb420a5e7c09095dac#diff-51ce39ab4212d73da272ba04e66b0b3bR3

eerhardt:
Why do we need “Build” in all the names?

Anipik:
> Why do we need “Build” in all the names?

chocsta mentioned in the previous PR and naming the job matrices as debug and release can cause some problems with azure dev ops https://github.com/dotnet/machinelearning/pull/3314#discussion_r275455062

eerhardt:
> @eerhardt is it fine to run centos just with netcoreapp3.0 and ubuntu with netcoreapp2.1 ?

It was nice getting a mix of the two that we were doing previously, but I think we should be OK.

shauheen:
@Anipik this seems to have broken the build badges on the main readme.

Anipik:
@shauheen We need to  update the job names there as well. i Will put up the PR for it.

Anipik:
We need to update the links

Anipik:
i found the new links to apis

PR #3564, opened 2019-04-24
Publish nugets to temporary myget feed (#3400)

shmoradims:
Re-enable myget feed.


eerhardt:
If we want to stop publishing the DNN packages (because they take up so much room), we can exclude or remove them here:

https://github.com/dotnet/machinelearning/blob/3c168ce23cbea9b30f4bbcd5729f9a1a016793b0/build/publish.proj#L19-L20

Issue #3527, opened 2019-04-23
Trainer documentation fixes


sfilipi:
**MatrixFactorizationTrainer** 
* [x] collaborative filtering needs to link to an explanation.
* [x] key-typed should be "key type". 
* [ ] low-rank factor matrix needs a link. 

**OlsTrainer** 
* [x] ordinary least squares needs a link. 

**OnlineGradientDescentTrainer** 
* [x] Online Gradient Descent (OGD) needs a link. 
* [x] "// This trainer is not numerically stable. Please see issue #2425. "Should no appear twice in the example. 

**LinearSvm** 
* [x] Has no samples. 
* [x] Has no proper description of what it does. 

**LbfgsMaximumEntropyMulticlassTrainer** 
* [ ] errors in equations
* [x] "no reource" typo

**SdcaMaximumEntropyMulticlassTrainer** 
* [x] Scoring function has ill rendered equations.
* [x] The SdcaMulticlassTrainerBase in the "Training Algorithm Details" has a broken link, that tries to search for an app in the store upon clicking. 

**SdcaMulticlassTrainerBase**
* [x] equations are not properly rendered.

sfilipi:
Closed with #3575 and #3565

PR #3565, opened 2019-04-24
Towards #3527

sfilipi:
Addresses the typos documented on #3527

### DIFF
--- a/src/Microsoft.ML.Mkl.Components/OlsLinearRegression.cs
+++ b/src/Microsoft.ML.Mkl.Components/OlsLinearRegression.cs
@@ -33,7 +33,7 @@ namespace Microsoft.ML.Trainers
 {
     /// <summary>
     /// The <see cref="IEstimator{TTransformer}"/> for training a linear regression model using
-    /// ordinary least squares (OLS) for estimating the parameters of the linear regression model.
+    /// <a href="https://en.wikipedia.org/wiki/Ordinary_least_squares">ordinary least squares (OLS)</a> for estimating the parameters of the linear regression model.
     /// </summary>
     /// <remarks>
     /// <format type="text/markdown"><![CDATA[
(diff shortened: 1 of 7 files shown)


src/Microsoft.ML.Mkl.Components/OlsLinearRegression.cs
```diff
@@ -33,7 +33,7 @@ namespace Microsoft.ML.Trainers
 {
     /// <summary>
     /// The <see cref="IEstimator{TTransformer}"/> for training a linear regression model using
-    /// ordinary least squares (OLS) for estimating the parameters of the linear regression model.
+    /// <a href="https://en.wikipedia.org/wiki/Ordinary_least_squares">ordinary least squares (OLS)</a> for estimating the parameters of the linear regression model.
```
artidoro: I remember some links being broken when there were parentheses. 
If it is confirmed to be an issue maybe we could omit (OLS) from the link.

src/Microsoft.ML.StandardTrainers/Standard/Online/OnlineGradientDescent.cs
```diff
@@ -25,7 +25,7 @@ namespace Microsoft.ML.Trainers
 {
     /// <summary>
     /// The <see cref="IEstimator{TTransformer}"/> for training a linear regression model using
-    /// Online Gradient Descent (OGD) for estimating the parameters of the linear regression model.
+    /// <a href="https://en.wikipedia.org/wiki/Gradient_descent">Online Gradient Descent (OGD)</a> for estimating the parameters of the linear regression model.
```
artidoro: Same comment here. But again, it might not be an issue.

sfilipi: don't think we have issues with `<a href>` inside xml. 
Probably markdown

src/Microsoft.ML.Mkl.Components/OlsLinearRegression.cs
sfilipi: Should not be an issue for `<a>` inside xml.

PR #3569, opened 2019-04-24
Improve initialization algorithm documentation for kmeans

### LINKED ISSUE #3495: KMeansTrainer InitializationAlgorithm documentation is confusing
From the "Training Algorithm Details" section in `KMeansTrainer` documentation, it seems that the difference between regular KMeans and KMeans++ is that the former uses random initialization of clusters, and the latter uses the Yinyang method. However, in the `Options` class there is an 'InitializationAlgorithm' enum for the initialization algorithm that contains three members: 'KMeansPlusPlus', 'KMeansYinYang' and 'Random'. So which is which?

Also, the fields in the in the `KMeansTrainer.InitializationAlgorithm` enum page are not in order.

yaeldekel:
Fixes #3495.

### DIFF
--- a/src/Microsoft.ML.KMeansClustering/KMeansPlusPlusTrainer.cs
+++ b/src/Microsoft.ML.KMeansClustering/KMeansPlusPlusTrainer.cs
@@ -47,13 +47,20 @@ namespace Microsoft.ML.Trainers
     /// ### Training Algorithm Details
     /// [K-means](https://en.wikipedia.org/wiki/K-means_clustering) is a popular clustering algorithm.
     /// With K-means, the data is clustered into a specified number of clusters in order to minimize the within-cluster sum of squared distances.
-    /// K-means++ improves upon K-means by using the [Yinyang K-Means](https://research.microsoft.com/apps/pubs/default.aspx?id=252149)
-    /// method for choosing the initial cluster centers.
-    /// K-Means++ accelerates K-Means up to an order of magnitude while producing exactly the same clustering results (modulo floating point precision issues).
-    /// K-Means++ observes that there is a lot of redundancy across iterations in the KMeans algorithms and most points do not change their clusters during an iteration.
-    /// It uses various bounding techniques to identify this redundancy and eliminate many distance computations and optimize centroid computations.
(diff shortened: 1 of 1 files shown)


yaeldekel:
>             KMeansYinyang = 2

This name is wrong. 
From what I understood, Yinyang refers to the assignment/update steps of the algorithm, and the three methods for initialization are Random, KMeans++ and KMeans||.

However, changing this would be a breaking change.

What do you guys think?

---
Refers to: src/Microsoft.ML.KMeansClustering/KMeansPlusPlusTrainer.cs:89 in 11fc3c3. [](commit_id = 11fc3c3fb1cf8e0f8f91a57d62ef4cc35f8b540e, deletion_comment = False)

src/Microsoft.ML.KMeansClustering/KMeansPlusPlusTrainer.cs
```diff
@@ -47,13 +47,20 @@ namespace Microsoft.ML.Trainers
...
+    /// For choosing the initial cluster centeroids, one of three options can be used:
+    /// - Random initialization. This might lead to potentially bad approximations of the optimal clustering.
+    /// - The K-means++ method. This is an [improved initialization algorithm](https://en.wikipedia.org/wiki/K-means%2b%2b#Improved_initialization_algorithm)
+    /// introduced [here](http://ilpubs.stanford.edu:8090/778/1/2006-13.pdf) by Ding et al., that guarantees to find
+    /// a solution that is $O(log K)$ competitive to the optimal K-means solution.
+    /// - The K-means|| method. This method was introduced [here](https://theory.stanford.edu/~sergei/papers/vldb12-kmpar.pdf) by Bahmani et al., and uses
+    /// a parallel method that drastically reduces the number of passes needed to obtain a good initialization.
     ///
-    /// ### Scorning Function
+    /// The latter is the default initialization method. The other methods can be specified in the [Options](xref:Microsoft.ML.Trainers.KMeansTrainer.Options)
```
najeeb-kazmi: K-means|| instead of latter - former/latter are used to distinguish between two things only.

PR #3572, opened 2019-04-24
Fix incorrect link in ComputeLogisticRegressionStandardDeviation class documentation

### LINKED ISSUE #3479: Incorrect link in ComputeLogisticRegressionStandardDeviation class documentation
It recommends using the implementation in Microsoft.ML.Mkl.Components, but it links to the page of the abstract class `ComputeLogisticRegressionStandardDeviation`.

codemzs:
fixes #3479

### DIFF
--- a/src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LogisticRegression.cs
+++ b/src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LogisticRegression.cs
@@ -477,7 +477,7 @@ namespace Microsoft.ML.Trainers
     /// <summary>
     /// Computes the standard deviation matrix of each of the non-zero training weights, needed to calculate further the standard deviation,
     /// p-value and z-Score.
-    /// If you need fast calculations, use this class' implementation in the Microsoft.ML.Mkl.Components package which uses Intel Math Kernel Library.
+    /// Use this class' implementation in the Microsoft.ML.Mkl.Components package which uses Intel Math Kernel Library.
     /// Due to the existence of regularization, an approximation is used to compute the variances of the trained linear coefficients.
     /// </summary>
     public abstract class ComputeLogisticRegressionStandardDeviation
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LogisticRegression.cs
```diff
@@ -477,8 +477,7 @@ internal static CommonOutputs.BinaryClassificationOutput TrainBinary(IHostEnviro
     /// <summary>
     /// Computes the standard deviation matrix of each of the non-zero training weights, needed to calculate further the standard deviation,
     /// p-value and z-Score.
-    /// If you need fast calculations, use the <see cref="ComputeLogisticRegressionStandardDeviation"/> implementation in the Microsoft.ML.Mkl.Components package,
-    /// which makes use of hardware acceleration.
+    /// If you need fast calculations, use this class' implementation in the Microsoft.ML.Mkl.Components package which uses Intel Math Kernel Library.
```
yaeldekel: >If you need fast calculations [](start = 8, length = 29)

Looked at the code, and the MKL implementation is actually the only implementation, so we probably don't need to say "if you need fast calculations" (who would prefer slow calculations anyway? :-)).

codemzs: lol

yaeldekel: So change to something like "To compute the standard deviation, use this..."

---
In reply to: [278327642](https://github.com/dotnet/machinelearning/pull/3572#discussion_r278327642) [](ancestors = 278327642)

codemzs: Changed to Use this class' implementation in the Microsoft.ML.Mkl.Components package which uses Intel Math Kernel Library....also the hardware acceleration part was a little dramatic

PR #3577, opened 2019-04-25
Reference to See Also section for example of usage in all estimators

### LINKED ISSUE #3485: All transform estimator documentation should point the user that the See Also section contains the example
Make sure:

"Check the See Also section for links to examples of the usage." is in all transform estimator documentation.

artidoro:
FIxes #3485.


src/Microsoft.ML.Transforms/Text/NgramHashingTransformer.cs
```diff
@@ -876,6 +876,8 @@ public VBuffer<ReadOnlyMemory<char>>[] SlotNamesMetadata(out VectorDataViewType[
     /// <xref:Microsoft.ML.Transforms.Text.NgramHashingEstimator> is different from <xref:Microsoft.ML.Transforms.Text.WordHashBagEstimator>
     /// in a way that The former takes tokenized text as input while the latter tokenizes text internally.
     /// Check the See Also section for links to examples of the usage.
```
artidoro: >Check [](start = 8, length = 5)

Duplicate.

docs/api-reference/algo-details-fastforest.md
```diff
@@ -28,4 +28,6 @@ For more see:
 * [Quantile regression
   forest](http://jmlr.org/papers/volume7/meinshausen06a/meinshausen06a.pdf)
 * [From Stumps to Trees to
-  Forests](https://blogs.technet.microsoft.com/machinelearning/2014/09/10/from-stumps-to-trees-to-forests/)
\ No newline at end of file
+  Forests](https://blogs.technet.microsoft.com/machinelearning/2014/09/10/from-stumps-to-trees-to-forests/)
+
+Check the See Also section for links to examples of the usage.
```
sfilipi: "to examples of the usage" doesn't sound well.. 

I think they should all be "to usage examples."

Sorry..

natke: I agree with @sfilipi but we could live with what is here

PR #3583, opened 2019-04-25
Add the correct name to the InitializationAlgorithm enum

yaeldekel:
The name Yinyang actually refers to the implementation of the algorithm itself, not the initialization of the clusters. The algorithm for cluster initialization either random, kmeans++ or kmeans||
(see the links in the trainer documentation [here](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.KMeansClustering/KMeansPlusPlusTrainer.cs#L48-L57)).

### DIFF
--- a/src/Microsoft.ML.KMeansClustering/KMeansPlusPlusTrainer.cs
+++ b/src/Microsoft.ML.KMeansClustering/KMeansPlusPlusTrainer.cs
@@ -88,7 +88,8 @@ namespace Microsoft.ML.Trainers
         {
             KMeansPlusPlus = 0,
             Random = 1,
-            KMeansYinyang = 2
+            KMeansYinyang = 2,
+            KMeansBarBar = 2
         }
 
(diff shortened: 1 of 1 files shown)


najeeb-kazmi:
>     /// The latter is the default initialization method. The other methods can be specified in the [Options](xref:Microsoft.ML.Trainers.KMeansTrainer.Options)

Can you say "K-means||" here instead of "latter"? Former/latter are only used to distinguish two items.

---
Refers to: src/Microsoft.ML.KMeansClustering/KMeansPlusPlusTrainer.cs:59 in c78631a. [](commit_id = c78631a16886e93e5830e76035d5e491a97a5560, deletion_comment = False)

eerhardt:
@yaeldekel - are you interested in merging this?

codemzs:
@yaeldekel closing this because of lack of response from you but if you feel this needs to merged please reopen and merge.

PR #3591, opened 2019-04-25
[AutoML] If first three iterations all fail, short-circuit AutoML experiment

daholste:
Closes https://github.com/dotnet/machinelearning-automl/issues/341


src/Microsoft.ML.Auto/Experiment/Experiment.cs
```diff
@@ -89,6 +89,12 @@ public IList<TRunDetail> Execute()
                     break;
                 }
 
+                // If after third run, all runs have failed so far, break
+                if (_history.Count() == 3 && _history.All(r => !r.RunSucceded))
+                {
+                    break;
```
vinodshanbhag: throw

daholste: you're right -- thanks!

PR #3610, opened 2019-04-27
[AutoML] take dependency on a specific ml.net version

Dmitry-A:


nuget.config
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+    <packageSources>
+        <add key="DotNetMyGet" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" protocolVersion="3" />
```
eerhardt: Did you actually need this file?  You shouldn't have needed this because it is already specified here:

https://github.com/dotnet/machinelearning/blob/a819b46a371fab556842d4c6fee728710f22fa14/Directory.Build.props#L17-L22

I don't believe this nuget.config does anything because we define the feeds in that msbuild property.

Dmitry-A: woops, no. thanks for catching this!

PR #3617, opened 2019-04-29
[AutoML] turn off line pragmas in .tt files to play nice with signing

Dmitry-A:


src/mlnet/Templates/Console/ModelProject.tt
```diff
@@ -1,4 +1,5 @@
-﻿<#@ template language="C#" #>
+﻿<#@ template language="C#" linePragmas="false" #>
+<#@ template language="C#" #>
```
srsaggam: This is duplicate?

Dmitry-A: yeah, silly, fixed.  thanks!

PR #3618, opened 2019-04-29
[AutoML] Fix error handling in CLI. 

srsaggam:
Looks like there are lost exceptions in CLI due to spawning threads and the reason for doing that is Progress Bar in CLI.

Things to try in future:
*  Use tasks. (Which has already been tried and had side effects in progress bar animation)
* Get rid of Progress bar code from CLI

I personally like to remove Progress bar code from CLI and use some existing nuget package in open source community.

We are bringing in some new set of problems because of the GUI work in CLI which seems to be a whole new project.

### DIFF
--- a/src/mlnet/CodeGenerator/CodeGenerationHelper.cs
+++ b/src/mlnet/CodeGenerator/CodeGenerationHelper.cs
@@ -77,7 +77,7 @@ namespace Microsoft.ML.CLI.CodeGenerator
             // i.e there is no common class/interface to handle all three tasks together.
 
             ExperimentResult<BinaryClassificationMetrics> binaryExperimentResult = default;
-            ExperimentResult<MulticlassClassificationMetrics> multiClassExperimentResult = default;
+            ExperimentResult<MulticlassClassificationMetrics> multiclassExperimentResult = default;
             ExperimentResult<RegressionMetrics> regressionExperimentResult = default;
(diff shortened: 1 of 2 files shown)


src/mlnet/CodeGenerator/CodeGenerationHelper.cs
```diff
@@ -278,5 +280,52 @@ private void ConsumeAutoMLSDKLogs(MLContext context)
...
+            {
+                multiClassExperimentResult = p.Invoke();
+                ex = null;
+            }
+            catch (Exception e)
+            {
+                ex = e;
+                multiClassExperimentResult = null;
+                pbar.Dispose();
+                //((ManualResetEvent)pbar.CompletedHandle).Set();
```
daholste: maybe remove

justinormont:
> I personally like to remove Progress bar code from CLI and use some existing nuget package in open source community.
> 
> We are bringing in some new set of problems because of the GUI work in CLI which seems to be a whole new project.

I fully agree w/ the above.

CESARDELATORRE:
I would agree on that, too, but the one we are using is the only one I found which was compelling enough. And that was just a few weeks ago.

Have you found any other concrete possibility?

What I’ll try to get when we have additional development support from the .NET team is to segregate the progress bar into a different project and even NuGet package and make it production ready (unless we find a better and production ready progress bar for .NET Core console apps in the community..

PR #3623, opened 2019-04-30
API Compat tool in ML.NET

### LINKED ISSUE #3602: Need to add API breaking change definition and enforce it
With version `1.0.0` release we need to formally adopt the API breaking change definitions similar to [CoreFx](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-definitions.md) and [here](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-changes.md). We also need to add a test to run [APICompat](https://github.com/dotnet/arcade/tree/master/src/Microsoft.DotNet.ApiCompat) tool to enforce these rules. @eerhardt @TomFinley @glebuk

artidoro:
Fixes #3602.

We need to ensure that future changes to ML.NET will not break the stable API released in 1.0.0.

This PR introduces the [API Compat](https://github.com/dotnet/arcade/tree/e5c20e707876cb9596d1cbcc996f80c25698fdfe/src/Microsoft.DotNet.ApiCompat) tool from dotnet/Arcade. The API Compat tool runs as part of the build process and compares the assemblies with those found in the stable nugets referenced in the `Microsoft.ML.StableAPI` project.

The tool is only run for the assemblies which will be part of the stable nugets. Here is a list of those assemblies and the relative nugets:

| Stable Nuget | Stable Assemblies|
|--|--|
|Microsoft.ML| Microsoft.ML.Core, Microsoft.ML.Data, Microsoft.ML.KMeansClustering, Microsoft.ML.PCA, Microsoft.ML.StandardTrainers, MIcrosoft.ML.Transforms, Microsoft.ML.Analyzer|
|Microsoft.ML.DataView|  Microsoft.ML.DataView|
|Microsoft.ML.CpuMath |  Microsoft.ML.CpuMath|
|Microsoft.ML.FastTree |  Microsoft.ML.FastTree|
|Microsoft.ML.LightGbm |  Microsoft.ML.LightGbm|
|Microsoft.ML.ImageAnalytics |  Microsoft.ML.ImageAnalytics|
|Microsoft.ML.MklComponents |  Microsoft.ML.Mkl.Components|

Note: the tool does not run on Microsoft.ML.Analyzer as it does not have a public API, so no need to check backwards compatibility.

Still to do:
- [x] Fix possible bug in API Compat on handling of Attributes.
- [x] ~~Add a build setting to disable the API Compat tool from command line. This will allow people to work on breaking changes.~~ Not needed see comments below.
- [x] Update the version of the ML.NET nugets to 1.0.0 when available

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -124,4 +124,14 @@
   <PropertyGroup>
     <CustomAfterMicrosoftCommonTargets>$(RepoRoot)build\AfterCommonTargets.targets</CustomAfterMicrosoftCommonTargets>
   </PropertyGroup>
+  
+  <!-- API Compat -->
+  <PropertyGroup>
+    <!-- only validate that the current implementation is compatible
+         with the old one, not vice-versa (since latest may have new
+         API missing from old) -->
+    <RunApiCompatForSrc>true</RunApiCompatForSrc>
+    <RunMatchingRefApiCompat>false</RunMatchingRefApiCompat>
+    <MicrosoftDotNetApiCompatPackageVersion>1.0.0-beta.19225.5</MicrosoftDotNetApiCompatPackageVersion>
+  </PropertyGroup>
 </Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -132,6 +132,5 @@
          API missing from old) -->
     <RunApiCompatForSrc>true</RunApiCompatForSrc>
     <RunMatchingRefApiCompat>false</RunMatchingRefApiCompat>
-    <MicrosoftDotNetApiCompatPackageVersion>1.0.0-beta.19225.5</MicrosoftDotNetApiCompatPackageVersion>
   </PropertyGroup>
 </Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -124,13 +124,4 @@
   <PropertyGroup>
     <CustomAfterMicrosoftCommonTargets>$(RepoRoot)build\AfterCommonTargets.targets</CustomAfterMicrosoftCommonTargets>
   </PropertyGroup>
-  
-  <!-- API Compat -->
-  <PropertyGroup>
-    <!-- only validate that the current implementation is compatible
-         with the old one, not vice-versa (since latest may have new
-         API missing from old) -->
-    <RunApiCompatForSrc>true</RunApiCompatForSrc>
-    <RunMatchingRefApiCompat>false</RunMatchingRefApiCompat>
-  </PropertyGroup>
 </Project>
--- a/Microsoft.ML.sln
+++ b/Microsoft.ML.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2026
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.28803.352
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ML.Core", "src\Microsoft.ML.Core\Microsoft.ML.Core.csproj", "{A6CA6CC6-5D7C-4D7F-A0F5-35E14B383B0A}"
 EndProject
@@ -274,7 +274,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Microsoft.ML.FastTree", "Mi
 		pkg\Microsoft.ML.FastTree\Microsoft.ML.FastTree.symbols.nupkgproj = pkg\Microsoft.ML.FastTree\Microsoft.ML.FastTree.symbols.nupkgproj
 	EndProjectSection
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ML.StableApi", "tools-local\Microsoft.ML.StableApi\Microsoft.ML.StableApi.csproj", "{C1884169-7330-42DF-B401-9427C6EFB092}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ML.StableApi", "tools-local\Microsoft.ML.StableApi\Microsoft.ML.StableApi.csproj", "{F308DC6B-7E59-40D7-A581-834E8CD99CFE}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
@@ -970,17 +970,18 @@ Global
 		{E02DA82D-3FEE-4C60-BD80-9EC3C3448DFC}.Release-Intrinsics|Any CPU.Build.0 = Release-Intrinsics|Any CPU
 		{E02DA82D-3FEE-4C60-BD80-9EC3C3448DFC}.Release-netfx|Any CPU.ActiveCfg = Release-netfx|Any CPU
 		{E02DA82D-3FEE-4C60-BD80-9EC3C3448DFC}.Release-netfx|Any CPU.Build.0 = Release-netfx|Any CPU
-		{C1884169-7330-42DF-B401-9427C6EFB092}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{C1884169-7330-42DF-B401-9427C6EFB092}.Debug-Intrinsics|Any CPU.ActiveCfg = Debug-Intrinsics|Any CPU
-		{C1884169-7330-42DF-B401-9427C6EFB092}.Debug-Intrinsics|Any CPU.Build.0 = Debug-Intrinsics|Any CPU
-		{C1884169-7330-42DF-B401-9427C6EFB092}.Debug-netfx|Any CPU.ActiveCfg = Debug-netfx|Any CPU
-		{C1884169-7330-42DF-B401-9427C6EFB092}.Debug-netfx|Any CPU.Build.0 = Debug-netfx|Any CPU
-		{C1884169-7330-42DF-B401-9427C6EFB092}.Release|Any CPU.ActiveCfg = Release|Any CPU
-		{C1884169-7330-42DF-B401-9427C6EFB092}.Release|Any CPU.Build.0 = Release|Any CPU
-		{C1884169-7330-42DF-B401-9427C6EFB092}.Release-Intrinsics|Any CPU.ActiveCfg = Release-Intrinsics|Any CPU
-		{C1884169-7330-42DF-B401-9427C6EFB092}.Release-Intrinsics|Any CPU.Build.0 = Release-Intrinsics|Any CPU
-		{C1884169-7330-42DF-B401-9427C6EFB092}.Release-netfx|Any CPU.ActiveCfg = Release-netfx|Any CPU
-		{C1884169-7330-42DF-B401-9427C6EFB092}.Release-netfx|Any CPU.Build.0 = Release-netfx|Any CPU
+		{F308DC6B-7E59-40D7-A581-834E8CD99CFE}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{F308DC6B-7E59-40D7-A581-834E8CD99CFE}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{F308DC6B-7E59-40D7-A581-834E8CD99CFE}.Debug-Intrinsics|Any CPU.ActiveCfg = Debug-Intrinsics|Any CPU
+		{F308DC6B-7E59-40D7-A581-834E8CD99CFE}.Debug-Intrinsics|Any CPU.Build.0 = Debug-Intrinsics|Any CPU
+		{F308DC6B-7E59-40D7-A581-834E8CD99CFE}.Debug-netfx|Any CPU.ActiveCfg = Debug-netfx|Any CPU
+		{F308DC6B-7E59-40D7-A581-834E8CD99CFE}.Debug-netfx|Any CPU.Build.0 = Debug-netfx|Any CPU
+		{F308DC6B-7E59-40D7-A581-834E8CD99CFE}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{F308DC6B-7E59-40D7-A581-834E8CD99CFE}.Release|Any CPU.Build.0 = Release|Any CPU
+		{F308DC6B-7E59-40D7-A581-834E8CD99CFE}.Release-Intrinsics|Any CPU.ActiveCfg = Release-Intrinsics|Any CPU
+		{F308DC6B-7E59-40D7-A581-834E8CD99CFE}.Release-Intrinsics|Any CPU.Build.0 = Release-Intrinsics|Any CPU
+		{F308DC6B-7E59-40D7-A581-834E8CD99CFE}.Release-netfx|Any CPU.ActiveCfg = Release-netfx|Any CPU
+		{F308DC6B-7E59-40D7-A581-834E8CD99CFE}.Release-netfx|Any CPU.Build.0 = Release-netfx|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -1068,7 +1069,7 @@ Global
 		{AD7058C9-5608-49A8-BE23-58C33A74EE91} = {D3D38B03-B557-484D-8348-8BADEE4DF592}
 		{E02DA82D-3FEE-4C60-BD80-9EC3C3448DFC} = {09EADF06-BE25-4228-AB53-95AE3E15B530}
 		{B1B3F284-FA3D-4D76-A712-FF04495D244B} = {D3D38B03-B557-484D-8348-8BADEE4DF592}
-		{C1884169-7330-42DF-B401-9427C6EFB092} = {7F13E156-3EBA-4021-84A5-CD56BA72F99E}
+		{F308DC6B-7E59-40D7-A581-834E8CD99CFE} = {7F13E156-3EBA-4021-84A5-CD56BA72F99E}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {41165AF1-35BB-4832-A189-73060F82B01D}
--- a/Microsoft.ML.sln
+++ b/Microsoft.ML.sln
@@ -971,7 +971,6 @@ Global
 		{E02DA82D-3FEE-4C60-BD80-9EC3C3448DFC}.Release-netfx|Any CPU.ActiveCfg = Release-netfx|Any CPU
 		{E02DA82D-3FEE-4C60-BD80-9EC3C3448DFC}.Release-netfx|Any CPU.Build.0 = Release-netfx|Any CPU
 		{C1884169-7330-42DF-B401-9427C6EFB092}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{C1884169-7330-42DF-B401-9427C6EFB092}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{C1884169-7330-42DF-B401-9427C6EFB092}.Debug-Intrinsics|Any CPU.ActiveCfg = Debug-Intrinsics|Any CPU
 		{C1884169-7330-42DF-B401-9427C6EFB092}.Debug-Intrinsics|Any CPU.Build.0 = Debug-Intrinsics|Any CPU
 		{C1884169-7330-42DF-B401-9427C6EFB092}.Debug-netfx|Any CPU.ActiveCfg = Debug-netfx|Any CPU
--- a/Microsoft.ML.sln
+++ b/Microsoft.ML.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28803.352
+# Visual Studio 15
+VisualStudioVersion = 15.0.27130.2026
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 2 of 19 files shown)


Directory.Build.props
```diff
@@ -124,4 +124,14 @@
...
   </PropertyGroup>
+  
+  <!-- API Compat -->
+  <PropertyGroup>
+    <!-- only validate that the current implementation is compatible
+         with the old one, not vice-versa (since latest may have new
+         API missing from old) -->
+    <RunApiCompatForSrc>true</RunApiCompatForSrc>
+    <RunMatchingRefApiCompat>false</RunMatchingRefApiCompat>
+    <MicrosoftDotNetApiCompatPackageVersion>1.0.0-beta.19225.5</MicrosoftDotNetApiCompatPackageVersion>
```
ericstj: This property should be moved to https://github.com/dotnet/machinelearning/blob/master/build/Dependencies.props

src/Directory.Build.targets
```diff
@@ -26,4 +26,29 @@
...
 
+  <!-- API Compat -->
+  <PropertyGroup Condition="'$(RunApiCompat)' == 'true'">
+    <!-- resolve contract for APICompat as part of resolve references -->
+    <ResolveReferencesDependsOn>
+      $(ResolveReferencesDependsOn);
+      ResolveMatchingContract
+    </ResolveReferencesDependsOn>
+    
+    <StabelApiProject>$(RepoRoot)tools-local\Microsoft.ML.StableApi\Microsoft.ML.StableApi.csproj</StabelApiProject>
```
ericstj: nit: spelling

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
```diff
@@ -0,0 +1,31 @@
...
+  <!-- The purpose of this target is to return a path from a referenced
+       package / project -->
+  <Target Name="GetContract"
+          DependsOnTargets="ResolveReferences"
+          Returns="@(_contractReferencePath)">
+    <Error Condition="'$(ContractName)' == ''" Text="ContractName must be specified" />
+    <ItemGroup>
+      <_contractReferencePath Include="@(ReferencePath)" Condition="'%(FileName)' == '$(ContractName)'" />
+    </ItemGroup>
+    <Error Condition="'$(ContractName)' == ''" Text="Could not locate $(ContractName)" />
```
ericstj: This should be `'@(_contractReferencePath)' == ''`.

src/Microsoft.ML.Analyzer/Microsoft.ML.Analyzer.csproj
```diff
@@ -1,8 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard1.3</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
     <IncludeAnalyzerInPackage>Microsoft.ML</IncludeAnalyzerInPackage>
+    <RunApiCompat>true</RunApiCompat>
```
ericstj: Do you actually ship this analyzer?  Also changing targetFramework is a breaking change since it drops compatibility with what you previously targeted.

src/Directory.Build.targets
```diff
@@ -26,4 +26,29 @@
 
   </Target>
 
+  <!-- API Compat -->
```
ericstj: Consider moving these to a seperate targets file if that is a convention you'd like to follow in this repo.

ericstj:
>  Add a build setting to disable the API Compat tool from command line. This will allow people to work on breaking changes.

The right way to do this is to build once specifying /p:BaselineAllAPICompatError=true.  This will generate text files in the repo that list all the errors and let the build succeed despite the compatibility issues.  Typically you then file issues to address all the baselines by the time you ship the next version.

src/Microsoft.ML.Analyzer/Microsoft.ML.Analyzer.csproj
eerhardt: Yes we ship this analyzer.

But this project has no public APIs, so I don't believe we need to Run API Compat on it.

---
In reply to: [279938491](https://github.com/dotnet/machinelearning/pull/3623#discussion_r279938491) [](ancestors = 279938491)

artidoro: Should the targetFramework remain 1.3 @eric.erhardt@microsoft.com?
I'll remove the `RunApiCompat`.

---
In reply to: [279940309](https://github.com/dotnet/machinelearning/pull/3623#discussion_r279940309) [](ancestors = 279940309,279938491)

src/Microsoft.ML.Core/Microsoft.ML.Core.csproj
```diff
@@ -6,11 +6,11 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <DefineConstants>CORECLR</DefineConstants>
     <RootNamespace>Microsoft.ML</RootNamespace>
+    <RunApiCompat>true</RunApiCompat>
```
eerhardt: Can we instead set `RunApiCompat` based on if `IsStableProject` is set to true?

[https://github.com/dotnet/machinelearning/blob/master/build/BranchInfo.props#L2-L23](https://github.com/dotnet/machinelearning/blob/master/build/BranchInfo.props#L2-L23)

In the root Directory.Build.props, you should be able to say:

```xml
  <PropertyGroup>
    <RunApiCompat Condition="'$(RunApiCompat)` == '' AND '$(IsStableProject)' == 'true'">true</RunApiCompat>
  </PropertyGroup>
```

or simply:

```xml
  <PropertyGroup>
    <RunApiCompat Condition="'$(RunApiCompat)` == ''">$(IsStableProject)</RunApiCompat>
  </PropertyGroup>
```

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
```diff
@@ -0,0 +1,31 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <!-- needs to contain all frameworks which src projects wish to restore -->
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ML" Version="1.0.0-preview-27625-16"/>
```
eerhardt: (nit) you can use the latest 1.0.0-preview build: `1.0.0-preview-27630-5`. We just spun it and will have the official `1.0.0` up soon.

src/Directory.Build.targets
artidoro: I think we only have one for the src repo. But let me know if there is a better way.

---
In reply to: [279938766](https://github.com/dotnet/machinelearning/pull/3623#discussion_r279938766) [](ancestors = 279938766)

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
artidoro: Will update to the new build and change to 1.0.0 as soon as it is available.

---
In reply to: [279941849](https://github.com/dotnet/machinelearning/pull/3623#discussion_r279941849) [](ancestors = 279941849)

src/Microsoft.ML.Core/Microsoft.ML.Core.csproj
artidoro: Sure that seems to represent more clearly the logic, I'll do it.

---
In reply to: [279941488](https://github.com/dotnet/machinelearning/pull/3623#discussion_r279941488) [](ancestors = 279941488)

ericstj:
```
D:\a\1\s\packages\microsoft.dotnet.apicompat\1.0.0-beta.19225.5\build\Microsoft.DotNet.ApiCompat.targets(72,5): error : CannotChangeAttribute : Attribute 'System.AttributeUsageAttribute' on 'Microsoft.ML.ExtensionBaseAttribute' changed from '[AttributeUsageAttribute(4)]' in the contract to '[AttributeUsageAttribute(AttributeTargets.Class)]' in the implementation. [D:\a\1\s\src\Microsoft.ML.Core\Microsoft.ML.Core.csproj]
```

This is happening because APICompat is not resolving the System.AttributeTargets enum from the contract.   It's not resolving because you aren't providing the dependencies.  Today these can be passed in via the `$(ContractOutputPath)` property but that isn't very sensible.  I'll submit a change to APICompat to make that easier.
https://github.com/dotnet/arcade/blob/ac8d88df02d246d3147338fcfb03b1b93dc84b53/src/Microsoft.DotNet.ApiCompat/build/Microsoft.DotNet.ApiCompat.targets#L46-L52

artidoro:
That would be great! Thank you

---
In reply to: [488135991](https://github.com/dotnet/machinelearning/pull/3623#issuecomment-488135991) [](ancestors = 488135991)

artidoro:
> > Add a build setting to disable the API Compat tool from command line. This will allow people to work on breaking changes.
> 
> The right way to do this is to build once specifying /p:BaselineAllAPICompatError=true. This will generate text files in the repo that list all the errors and let the build succeed despite the compatibility issues. Typically you then file issues to address all the baselines by the time you ship the next version.

@ericstj is it possible to specify the path for the generated file?
Also, how would we specify known diffs going forward? I would like to write some documentation on how to do that.

ericstj:
by default the generated file gets written to the same directory as the project with the name, but you can customize it: https://github.com/dotnet/arcade/blob/ac8d88df02d246d3147338fcfb03b1b93dc84b53/src/Microsoft.DotNet.ApiCompat/build/Microsoft.DotNet.ApiCompat.targets#L22-L23

> how would we specify known diffs going forward?

You can use the same option I mentioned when building individual projects, you can also copy the errors and paste them into the baseline file.  I want to stress that these aren't "known diffs" these are compatibility bugs that break your customers and you really shouldn't be baselining them or introducing the suppressions.

ericstj:
To provide paths to contract dependencies, add the following in your `GetContract` target:
```
      <_allReferenceDirectories Include="%(ReferencePath.RootDir)%(ReferencePath.Directory)" />
      <_contractReferencePath Include="@(ReferencePath)" Condition="'%(FileName)' == '$(ContractName)'" />
      <_contractReferencePath DependencyPaths="@(_allReferenceDirectories)" />
```

Then in `ResolveMatchingContract` add the following after you get the ResolvedMatchingContract item:
```
<PropertyGroup>
  <ContractOutputPath>%(ResolvedMatchingContract.DependencyPaths)</ContractOutputPath>
</PropertyGroup>
```

I'm putting a feature into the APICompat targets that will make the latter property unnecessary.

src/Microsoft.ML.Analyzer/Microsoft.ML.Analyzer.csproj
eerhardt: I wouldn't change the TFM unless absolutely necessary. I think we should be able to revert this file.

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
eerhardt: It's available on myget now. You should be able to use `1.0.0`.

src/Microsoft.ML.Core/Microsoft.ML.Core.csproj
```diff
@@ -10,7 +10,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.ML.DataView\Microsoft.ML.DataView.csproj" />
-    
+
```
eerhardt: I think you can revert all these .csproj files. No need cluttering the history making unnecessary changes.

src/Microsoft.ML.Analyzer/Microsoft.ML.Analyzer.csproj
```diff
@@ -3,6 +3,9 @@
   <PropertyGroup>
     <TargetFramework>netstandard1.3</TargetFramework>
     <IncludeAnalyzerInPackage>Microsoft.ML</IncludeAnalyzerInPackage>
+
+    <!-- No public surface in this assembly -->
+    <RunApiCompat>false</RunApiCompat>
```
eerhardt: This shouldn't be necessary since ML.Analyzer isn't listed as a stable project. Can be reverted.

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
```diff
@@ -0,0 +1,31 @@
...
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ML" Version="1.0.0-preview-27630-5"/>
+    <PackageReference Include="Microsoft.ML.DataView" Version="1.0.0-preview-27630-5"/>
+    <PackageReference Include="Microsoft.ML.CpuMath" Version="1.0.0-preview-27630-5"/>
+    <PackageReference Include="Microsoft.ML.FastTree" Version="1.0.0-preview-27630-5"/>
+    <PackageReference Include="Microsoft.ML.LightGbm" Version="1.0.0-preview-27630-5"/>
+    <PackageReference Include="Microsoft.ML.ImageAnalytics" Version="1.0.0-preview-27630-5"/>
+    <PackageReference Include="Microsoft.ML.Mkl.Components" Version="1.0.0-preview-27630-5"/>
+    <PackageReference Include="Microsoft.ML.Mkl.Redist" Version="1.0.0-preview-27630-5"/>
```
eerhardt: The Redist package doesn't have public APIs. Do we need to include it here?

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
```diff
@@ -0,0 +1,31 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
eerhardt: This project will actually build a `.dll` when we build the `.sln` file. That seems unnecessary. Maybe we should override the `Build` target or maybe not name it `.csproj` and instead just `.proj`? It's mission in life isn't to build `.cs` files into a `.dll`, but instead pull down external packages and provide the `GetContract` target.

Directory.Build.props
```diff
@@ -124,4 +124,13 @@
   <PropertyGroup>
     <CustomAfterMicrosoftCommonTargets>$(RepoRoot)build\AfterCommonTargets.targets</CustomAfterMicrosoftCommonTargets>
   </PropertyGroup>
+  
+  <!-- API Compat -->
```
eerhardt: Can these properties go into `src\Directory.Build.props` instead? Grouping them with the `RunApiCompat` property would make sense IMO.

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
artidoro: Yes you are right I'll remove that.

---
In reply to: [280104238](https://github.com/dotnet/machinelearning/pull/3623#discussion_r280104238) [](ancestors = 280104238)

Directory.Build.props
artidoro: Yes I'll do that!

---
In reply to: [280106010](https://github.com/dotnet/machinelearning/pull/3623#discussion_r280106010) [](ancestors = 280106010)

artidoro:
Thank you that seems to make it work!

---
In reply to: [488287870](https://github.com/dotnet/machinelearning/pull/3623#issuecomment-488287870) [](ancestors = 488287870)

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
artidoro: I tried renaming, but that did not work (it gave an error saying that project.assets.json was not generated.
I have also tried to overwrite the Build target, but I must have made some mistake since I still found the generated .dll.

What I did was adding:
 <Target Name="Build">
    
  </Target>

to the .csproj file. 

What is the correct way to overwrite it?



---
In reply to: [280105436](https://github.com/dotnet/machinelearning/pull/3623#discussion_r280105436) [](ancestors = 280105436)

ericstj:
failing on linux because dotnet is not on the path and the $(ToolHostCmd) is not set to point to it.

artidoro:
Thank you @ericstj  and @eerhardt for looking at the PR, I have addressed your comments.
Let me know if there are other things I should look into!

ericstj:
Cool, this is looking pretty good.  You may want to wait for my changes https://github.com/dotnet/arcade/pull/2672 which help simplify some of this.

artidoro:
I would prefer checking in this change as soon as possible and updating it with the new buyer as soon as it is available.

Since we have released yesterday, it would be great to have the API Compat tool in the repo.

Microsoft.ML.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2026
+# Visual Studio Version 16
```
eerhardt: You may want to revert these two lines. I don't know what happens if someone tries opening the solution with VS 2017 (which is version 15) and this .sln says it is for VS 2019.

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
eerhardt: Yes, I think that is fine.

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
```diff
@@ -0,0 +1,45 @@
...
+
+  <PropertyGroup>
+    <!-- needs to contain all frameworks which src projects wish to restore -->
+    <TargetFramework Condition="'$(UseIntrinsics)' != 'true'">netstandard2.0</TargetFramework>
+    <TargetFrameworks Condition="'$(UseIntrinsics)' == 'true'">netstandard2.0;netcoreapp3.0</TargetFrameworks>
```
ericstj: Why should we have the condition here?  Is there any harm in always resolving for both TFMs?

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
```diff
@@ -0,0 +1,45 @@
...
+    <ItemGroup>
+      <_contractReferencePath Include="@(ReferencePath)" Condition="'%(FileName)' == '$(ContractName)'" />
+      <_allReferenceDirectories Include="%(ReferencePath.RootDir)%(ReferencePath.Directory)" />
+      <_contractReferencePath Include="@(ReferencePath)" Condition="'%(FileName)' == '$(ContractName)'" />
+      <_contractReferencePath DependencyPaths="@(_allReferenceDirectories)" />
+    </ItemGroup>
+    <Error Condition="'@(_contractReferencePath)' == ''" Text="Could not locate $(ContractName)" />
+  </Target>
+
+  <Target Name="Build">
```
ericstj: This could be improved by changing the project extension and defining a couple targets.  We do that elsewhere: 
https://github.com/dotnet/project-system/issues/4647

IOW: call this a .proj, or a .restoreproj or something.  SLN entry looks the same (you have to do it manually in text editor, IDE won't let you).  Add the workaround I linked, define stub targets for anything that fails, and it should avoid 
the confusion around a CSProj that doesn't build anything.
You don't need to do that now, but I think it would make this more sensible.

ericstj:
Have you made sure you see a failure when making a breaking API change?

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
eerhardt: Usually `netcoreapp3.0` causes problems when you are building in VS 2017 where the SDK throws an error saying `I don't support netcoreapp3.0`.

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
artidoro: I have tried to add the lines found in the issue to the new .restoreproj file but that did not work. 
I will keep it as is for now then.

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
ericstj: I see, makes sense.

artidoro:
Yes I have done a few tests where I rename methods and such and the build was failing.

tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
```diff
@@ -0,0 +1,45 @@
...
+  </ItemGroup>
+
+  <!-- The purpose of this target is to return a path from a referenced
+       package / project -->
+  <Target Name="GetContract" DependsOnTargets="ResolveReferences" Returns="@(_contractReferencePath)">
+    <Error Condition="'$(ContractName)' == ''" Text="ContractName must be specified" />
+    <ItemGroup>
+      <_contractReferencePath Include="@(ReferencePath)" Condition="'%(FileName)' == '$(ContractName)'" />
+      <_allReferenceDirectories Include="%(ReferencePath.RootDir)%(ReferencePath.Directory)" />
+      <_contractReferencePath Include="@(ReferencePath)" Condition="'%(FileName)' == '$(ContractName)'" />
```
RussKie: Lines 24 and 26 are identical... 😕

eerhardt: @RussKie - can you log an issue (or even submit a PR for the fix)?

PR #3625, opened 2019-04-30
Download images only when not present on disk and print warning messages when converting unsupported pixel format.

### LINKED ISSUE #3631: Image transformers such as Resize and Pixel extractor should print warning message when converting unsupported pixel type to supported pixel type.
Often images contain pixel formats that is not supported by the transformers. Instead of throwing exceptions we currently convert them to supported types. It would be nice if we also printed a warning line in logs that indicated the name of the image, its current pixel format and the new pixel format the code is converting to.

### LINKED ISSUE #3638: Download files only if they are not present on disk.
In addition to wastage of network and other computing resources this also causes issues when an a Bitmap object is holding a write lock on an image and download tries to overwrite that image and results in file in use exception.

codemzs:
fixes #3631
fixes #3638

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToImage.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToImage.cs
@@ -68,6 +68,9 @@ namespace Samples.Dynamic
                     Console.WriteLine("{0, -25} {1, -25} {2, -25}", string.Join(",", features.DenseValues().Take(5)) + "...",
                         imageObject.PhysicalDimension, string.Join(",", pixels.DenseValues().Take(5)) + "...");
                 }
+
+                // Dispose the image.
+                imageObject.Dispose();
             }
         }
 
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToImage.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToImage.cs
@@ -68,9 +68,6 @@ namespace Samples.Dynamic
                     Console.WriteLine("{0, -25} {1, -25} {2, -25}", string.Join(",", features.DenseValues().Take(5)) + "...",
                         imageObject.PhysicalDimension, string.Join(",", pixels.DenseValues().Take(5)) + "...");
                 }
-
-                // Dispose the image.
-                imageObject.Dispose();
             }
         }
 
--- a/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
+++ b/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
@@ -197,13 +197,7 @@ namespace Microsoft.ML.Data
                                 if (!string.IsNullOrWhiteSpace(_parent.ImageFolder))
                                     path = Path.Combine(_parent.ImageFolder, path);
 
-                                MemoryStream inMemoryCopy = new MemoryStream();
-                                using (FileStream fs = File.OpenRead(path))
-                                {
-                                    fs.CopyTo(inMemoryCopy);
-                                }
-
-                                dst = new Bitmap(inMemoryCopy) { Tag = path };
+                                dst = new Bitmap(path) { Tag = path };
                             }
                             catch (Exception)
                             {
--- a/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
+++ b/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
@@ -197,10 +197,13 @@ namespace Microsoft.ML.Data
                                 if (!string.IsNullOrWhiteSpace(_parent.ImageFolder))
                                     path = Path.Combine(_parent.ImageFolder, path);
 
-                                using (var stream = new FileStream(path, FileMode.Open, FileAccess.Read))
+                                MemoryStream inMemoryCopy = new MemoryStream();
+                                using (FileStream fs = File.OpenRead(path))
                                 {
-                                    dst = new Bitmap(stream) { Tag = path };
+                                    fs.CopyTo(inMemoryCopy);
                                 }
+
+                                dst = new Bitmap(inMemoryCopy) { Tag = path };
                             }
                             catch (Exception)
                             {
--- a/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
+++ b/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
@@ -190,23 +190,11 @@ namespace Microsoft.ML.Data
 
                         if (src.Length > 0)
                         {
-                            // Catch exceptions and pass null through. Should also log failures...
-                            try
-                            {
-                                string path = src.ToString();
-                                if (!string.IsNullOrWhiteSpace(_parent.ImageFolder))
-                                    path = Path.Combine(_parent.ImageFolder, path);
+                            string path = src.ToString();
+                            if (!string.IsNullOrWhiteSpace(_parent.ImageFolder))
+                                path = Path.Combine(_parent.ImageFolder, path);
 
-                                dst = new Bitmap(path) { Tag = path };
-                            }
-                            catch (Exception)
-                            {
-                                // REVIEW: We catch everything since the documentation for new Bitmap(string)
-                                // appears to be incorrect. When the file isn't found, it throws an ArgumentException,
-                                // while the documentation says FileNotFoundException. Not sure what it will throw
-                                // in other cases, like corrupted file, etc.
-                                throw Host.Except($"Image {src.ToString()} was not found.");
-                            }
+                            dst = new Bitmap(path) { Tag = path };
 
                             // Check for an incorrect pixel format which indicates the loading failed
                             if (dst.PixelFormat == System.Drawing.Imaging.PixelFormat.DontCare)
--- a/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
+++ b/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
@@ -196,7 +196,11 @@ namespace Microsoft.ML.Data
                                 string path = src.ToString();
                                 if (!string.IsNullOrWhiteSpace(_parent.ImageFolder))
                                     path = Path.Combine(_parent.ImageFolder, path);
-                                dst = new Bitmap(path);
+
+                                using (var stream = new FileStream(path, FileMode.Open, FileAccess.Read))
+                                {
+                                    dst = new Bitmap(stream) { Tag = path };
+                                }
                             }
                             catch (Exception)
                             {
--- a/src/Microsoft.ML.ImageAnalytics/ImagePixelExtractor.cs
+++ b/src/Microsoft.ML.ImageAnalytics/ImagePixelExtractor.cs
@@ -352,6 +352,7 @@ namespace Microsoft.ML.Transforms.Image
                         if (src.PixelFormat != PixelFormat.Format32bppArgb && src.PixelFormat != PixelFormat.Format24bppRgb)
                         {
                             var clone = src.Clone(new Rectangle(0, 0, src.Width, src.Height), PixelFormat.Format32bppArgb);
+                            clone.Tag = src.Tag;
                             src.Dispose();
                             src = clone;
                             using (var ch = Host.Start(nameof(ImagePixelExtractingTransformer)))
--- a/src/Microsoft.ML.ImageAnalytics/ImagePixelExtractor.cs
+++ b/src/Microsoft.ML.ImageAnalytics/ImagePixelExtractor.cs
@@ -357,8 +357,7 @@ namespace Microsoft.ML.Transforms.Image
                             src = clone;
                             using (var ch = Host.Start(nameof(ImagePixelExtractingTransformer)))
                             {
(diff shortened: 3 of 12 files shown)


src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
```diff
@@ -196,7 +196,11 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
                                 string path = src.ToString();
                                 if (!string.IsNullOrWhiteSpace(_parent.ImageFolder))
                                     path = Path.Combine(_parent.ImageFolder, path);
-                                dst = new Bitmap(path);
+
+                                using (var stream = new FileStream(path, FileMode.Open, FileAccess.Read))
```
eerhardt: https://docs.microsoft.com/en-us/dotnet/api/system.drawing.bitmap.-ctor?view=netframework-4.8#System_Drawing_Bitmap__ctor_System_IO_Stream_

The Remarks section of Bitmap says:

> You must keep the stream open for the lifetime of the Bitmap.

Are we sure we can dispose this?
cc @safern

src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
```diff
@@ -196,7 +196,11 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
...
+                                {
```
glebuk: "You must keep the stream open for the lifetime of the Bitmap. "
https://docs.microsoft.com/en-us/dotnet/api/system.drawing.bitmap.-ctor?view=netframework-4.8#System_Drawing_Bitmap__ctor_System_IO_Stream_

codemzs: I don't it matters if you are not writing to the image. In our case we are not doing that.

src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
codemzs: I know for a fact it gets disposed the way our transfomers are written. When we move to the new row the object created in the previous row is disposed by the transform's that uses ImageLoader transformer.

src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
TomFinley: > It doesn't matters if you are not writing to the image. In our case we are not doing that. The reason they want this is because of `Save` routine we have in Bitmap class.

Interesting. We ought to note this in documentation then, what you think?

codemzs: I have reverted the change now.

---
In reply to: [279979459](https://github.com/dotnet/machinelearning/pull/3625#discussion_r279979459) [](ancestors = 279979459)

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToImage.cs
```diff
@@ -68,9 +68,6 @@ private static void PrintColumns(IDataView transformedData)
                     Console.WriteLine("{0, -25} {1, -25} {2, -25}", string.Join(",", features.DenseValues().Take(5)) + "...",
                         imageObject.PhysicalDimension, string.Join(",", pixels.DenseValues().Take(5)) + "...");
                 }
-
-                // Dispose the image.
-                imageObject.Dispose();
```
glebuk: So when are those objects disposed off then? Someone somewhere has to do this... Aren't they holding handle and other windows objects?

codemzs: Can you please clarify what handle and other window objects you are referring to?

I had added that dispose call because in the past bitmap object was created such that it was holding lock on the file but that is no longer the case after this change. GC will collect the bitmap object and dispose it since there is no reference to it after sample execution. I have tested this by calling GC.Finalize(). This is sample code that should be simple and also educational. If you feel from educational point of view it makes sense to add back then I’ll do it.

src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
```diff
@@ -196,7 +196,14 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
                                 string path = src.ToString();
                                 if (!string.IsNullOrWhiteSpace(_parent.ImageFolder))
                                     path = Path.Combine(_parent.ImageFolder, path);
-                                dst = new Bitmap(path);
+
+                                MemoryStream inMemoryCopy = new MemoryStream();
+                                using (FileStream fs = File.OpenRead(path))
+                                {
+                                    fs.CopyTo(inMemoryCopy);
+                                }
```
TomFinley: Probably simpler as `var inMemoryCopy = new MemoryStream(File.ReadAllBytes());`.

TomFinley: Would also be nice if it had some protection again, say, being pointed to a huge non-image file.

---
In reply to: [280130176](https://github.com/dotnet/machinelearning/pull/3625#discussion_r280130176) [](ancestors = 280130176)

TomFinley:
>                                 throw Host.Except($"Image {src.ToString()} was not found.");

Not your code, but there's something a bit odd about it that it assumes that any exception must be due to teh file not being found.

---
Refers to: src/Microsoft.ML.ImageAnalytics/ImageLoader.cs:208 in 133fbe1. [](commit_id = 133fbe1205c74865e9827d03c23381db8ad583c7, deletion_comment = False)

src/Microsoft.ML.ImageAnalytics/ImagePixelExtractor.cs
```diff
@@ -352,8 +352,14 @@ private ValueGetter<VBuffer<TValue>> GetGetterCore<TValue>(DataViewRow input, in
                         if (src.PixelFormat != PixelFormat.Format32bppArgb && src.PixelFormat != PixelFormat.Format24bppRgb)
                         {
                             var clone = src.Clone(new Rectangle(0, 0, src.Width, src.Height), PixelFormat.Format32bppArgb);
+                            clone.Tag = src.Tag;
                             src.Dispose();
                             src = clone;
+                            using (var ch = Host.Start(nameof(ImagePixelExtractingTransformer)))
+                            {
+                                ch.Warning($"Encountered image {{0}} of unsupported pixel format {{1}} but converting it to {{2}}.",
```
TomFinley: >{0} [](start = 64, length = 3)

You're formatting this string using `$`... so that you can convert `{{0}}` to `{0}`, which is an old style format string, which you then use with `src.Tag`. It might be more straightforward to just use the `$` operator only.

src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
```diff
@@ -196,7 +196,8 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
                                 string path = src.ToString();
                                 if (!string.IsNullOrWhiteSpace(_parent.ImageFolder))
                                     path = Path.Combine(_parent.ImageFolder, path);
-                                dst = new Bitmap(path);
+
+                                dst = new Bitmap(path) { Tag = path };
                             }
                             catch (Exception)
```
glebuk: >catch (Exception) [](start = 28, length = 17)

Why not catch both argument and filenotfound then?  Catching any exception seems dangerous, moreover, you are not capturing inner exception.

codemzs: I'm sorry but what you are talking about does not seem related to my change. I haven't made any change regarding catching exception...

src/Microsoft.ML.ImageAnalytics/ImagePixelExtractor.cs
glebuk: Tom, didn't you say that we should be using the old style string formatting in errors to reduce number of strings?
Also if you are using the $ style, you should be making it like $"Encountered image {src.Tag } ...

---
In reply to: [280178155](https://github.com/dotnet/machinelearning/pull/3625#discussion_r280178155) [](ancestors = 280178155)

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -836,45 +837,90 @@ public class MNISTPrediction
...
 
             trans.Schema.TryGetColumnIndex("Output", out int output);
             using (var cursor = trans.GetRowCursor(trans.Schema["Output"]))
             {
-                var buffer = default(VBuffer<float>);
-                var getter = cursor.GetGetter<VBuffer<float>>(trans.Schema["Output"]);
-                var numRows = 0;
-                while (cursor.MoveNext())
+                using (var cursor2 = trans.GetRowCursor(trans.Schema["Output"]))
                 {
```
glebuk: nit: Common formatting pattern in such cases should  reduce the indentation and make it more readable
            using (var cursor = trans.GetRowCursor(trans.Schema["Output"]))
            using (var cursor2 = trans.GetRowCursor(trans.Schema["Output"]))
            {
                var buffer = defa

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -836,45 +837,90 @@ public class MNISTPrediction
...
+            // taco_invalidpixelformat.jpg has '8207' pixel format on Windows but this format translates to Format32bppRgb
+            // on macOS and Linux, hence on Windows this image's pixel format is converted in resize transformer to Format32bppArgb
+            // and on linux and macOS it is not converted in resize transform since pixel format 'Format32bppRgb' can be resized but
+            // in ImagePixelExtractingTransformer it is converted to Format32bppArgb since there we just support two 
+            // pixel formats, i.e Format32bppArgb and Format16bppArgb.
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) || RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                Assert.Contains(
+                    @"[Source=Mapper; ImagePixelExtractingTransformer, Kind=Warning] Encountered image " +
+                    GetDataPath("images/taco_invalidpixelformat.jpg") +
```
glebuk: >/ [](start = 39, length = 1)

will the slash be different on different platforms?

codemzs:
>                                 throw Host.Except($"Image {src.ToString()} was not found.");

Seems like the comment above this line explains why the original author wrote like.

---
In reply to: [488380679](https://github.com/dotnet/machinelearning/pull/3625#issuecomment-488380679) [](ancestors = 488380679)

---
Refers to: src/Microsoft.ML.ImageAnalytics/ImageLoader.cs:208 in 133fbe1. [](commit_id = 133fbe1205c74865e9827d03c23381db8ad583c7, deletion_comment = False)

src/Microsoft.ML.ImageAnalytics/ImagePixelExtractor.cs
TomFinley: > Tom, didn't you say that we should be using the old style string formatting in errors to reduce number of created strings?

I don't think I said that. I have generally been an enthusastic user of `$` style formatting since its introduction, so it would be odd of me to say that. Maybe someone else said it? (Maybe if no one is listening on the pipe, it doesn't bother to format the string? That might be a valid point. But I don't think I'd make that point.)

But even if I bought that, we'd still be having what amounts to the worst of both worlds with this unnecessary `$` which is not a constant. So you start with the `$` style string, which is creating a format string just for the sake of using the old style. It's like I did something like: `string.Format($"{{0}}", a)` instead of just `$"{a}".

> Also if you are using the $ style, you should be making it inline, like so: $"Encountered image {src.Tag } ...

Right, that is correct.

codemzs:
>                                 throw Host.Except($"Image {src.ToString()} was not found.");

Spoke offline and we decided the to remove try catch altogether. Right now docs do say on file not found it will throw FileNotFoundException instead of ArgumentException as before..

---
In reply to: [488396420](https://github.com/dotnet/machinelearning/pull/3625#issuecomment-488396420) [](ancestors = 488396420,488380679)

---
Refers to: src/Microsoft.ML.ImageAnalytics/ImageLoader.cs:208 in 133fbe1. [](commit_id = 133fbe1205c74865e9827d03c23381db8ad583c7, deletion_comment = False)

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
codemzs: ah nice thanks 

---
In reply to: [280192257](https://github.com/dotnet/machinelearning/pull/3625#discussion_r280192257) [](ancestors = 280192257)

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ImageAnalytics/ConvertToImage.cs
glebuk: According to doc it holds the GDI+ bitmap, which is the unmanaged resource.  Since it's disposable, we should call dispose of it once it's used, because the GC might significantly underestimate the cost of the object by not tracking unmanaged resources and result in a low resource condition..  What's wrong with disposing it here?

---
In reply to: [279999356](https://github.com/dotnet/machinelearning/pull/3625#discussion_r279999356) [](ancestors = 279999356)

PR #3632, opened 2019-05-01
[AutoML] Trim the values of column names provided in --ignore-columns in CLI

srsaggam:
Trim spaces for the values provided to --ignore-columns through CLI


justinormont:
Recommend not trimming the column names. If a user says their column name is " MyLabel" (with the space), I'd recommend throwing.

srsaggam:
There is no way to distinguish "a, b" i.e If the user meant space after "," or " b"

srsaggam:
Closing this since we expect columns with spaces.

PR #3633, opened 2019-05-01
[AutoML] use Name instead of FullName for telemetry filename hash

Dmitry-A:
minor nit fix, better get it in now

### DIFF
--- a/src/mlnet/Commands/New/NewCommandHandler.cs
+++ b/src/mlnet/Commands/New/NewCommandHandler.cs
@@ -21,7 +21,7 @@ namespace Microsoft.ML.CLI.Commands.New
 
         public void Execute()
         {
-            telemetry.LogAutoTrainMlCommand(settings.Dataset.FullName, settings.MlTask.ToString(), settings.Dataset.Length);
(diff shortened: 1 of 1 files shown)


src/mlnet/Commands/New/NewCommandHandler.cs
```diff
@@ -21,7 +21,7 @@ internal NewCommand(NewCommandSettings settings, MlTelemetry telemetry)
 
         public void Execute()
         {
-            telemetry.LogAutoTrainMlCommand(settings.Dataset.FullName, settings.MlTask.ToString(), settings.Dataset.Length);
+            telemetry.LogAutoTrainMlCommand(settings.Dataset.Name, settings.MlTask.ToString(), settings.Dataset.Length);
```
justinormont: What is FullName & Name? Is this the filepath & filename?

Dmitry-A: FullName would have both name and path. We log this mostly to try to tell how much use is from samples, folks will have different folders but same files so Name is what we really want.

PR #3645, opened 2019-05-02
Adds missing license headers

shauheen:
Adding the license header to files that were missing it


codemzs:
Are these the only files where header is missing? You could do a regex search on all cs files ..

shauheen:
@codemzs I have ran a grep script for these obviously. If you found others please fix them.

PR #3651, opened 2019-05-02
Adds missing license headers.

codemzs:
In addition to #3645, I wrote a small C# program to open every ".cs" file in src directory and read the first non-null/empty line and compare it against "// Licensed to the .NET Foundation under one or more agreements." and found few more places where license was missing. 

```csharp
public static class Program
    {
        public static void Main(string[] args) => RunAll(@"E:\machinelearning\src");

        internal static void RunAll(string sDir)
        {
           foreach (string d in Directory.GetDirectories(sDir))
            {
                foreach (string f in Directory.GetFiles(d))
                {
                    if (Path.GetExtension(f) == ".cs" || Path.GetExtension(f) == ".tt")
                    {
                        using (var fileStream = File.OpenRead(f))
                        using (var streamReader = new StreamReader(fileStream, Encoding.UTF8, true))
                        {
                            string line;
                            while ((line = streamReader.ReadLine()) != null)
                            {
                                if(String.IsNullOrEmpty(line))
                                    continue;

                                if (line !=
                                    @"// Licensed to the .NET Foundation under one or more agreements.")
                                {
                                    Console.WriteLine(f + " : " + line);
                                }

                                break;

                            }
                        }
                    }
                }
                RunAll(d);
            }
        }
    }
```

### DIFF
--- a/src/Microsoft.ML.TensorFlow/TensorFlow/TensorGeneric.cs
+++ b/src/Microsoft.ML.TensorFlow/TensorFlow/TensorGeneric.cs
@@ -1,5 +1,6 @@
-// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
 
 using System;
 using System.Text;
--- a/src/Microsoft.ML.FastTree/Utils/BufferPoolManager.cs
+++ b/src/Microsoft.ML.FastTree/Utils/BufferPoolManager.cs
@@ -1,8 +1,6 @@
-﻿// -----------------------------------------------------------------------
-// <copyright file="BufferPoolManager.cs" company="Microsoft Corporation">
-//     Copyright (C) All Rights Reserved
-// </copyright>
-// -----------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


src/Microsoft.ML.TensorFlow/TensorFlow/TensorGeneric.cs
```diff
@@ -1,5 +1,6 @@
-// Copyright (c) Microsoft. All rights reserved.
```
eerhardt: This file appears to be generated from:

https://github.com/dotnet/machinelearning/blob/fad2b23f1c14082ba03c2ad1f41ccc0f90ae5110/src/Microsoft.ML.TensorFlow/TensorFlow/TensorGeneric.tt#L8-L9

you'll want to update the .tt file so the next time someone wants to edit this code, they don't regress this change.

codemzs: very good catch! May be I should run the tool on tt files too

zeahmed: Yes, I see that the license comment is after some headers in .tt. May be you can put the license on top and then your tool will do the job on .tt too.

---
In reply to: [280540275](https://github.com/dotnet/machinelearning/pull/3651#discussion_r280540275) [](ancestors = 280540275)

PR #3653, opened 2019-05-02
[AutoML] Cleaner way to exit the main()

srsaggam:
We could just return the exit code from main()
Instead of Environment.Exit(exitcode)

The reason it wasn't working before was the fact that we were spawning foreground threads.

It seems to be the standard.
https://github.com/dotnet/cli/blob/21cda4adb6d88d240978e24fcfeb3ebe73f5fdd7/src/dotnet/Program.cs#L30-L83

### DIFF
--- a/src/mlnet/CodeGenerator/CodeGenerationHelper.cs
+++ b/src/mlnet/CodeGenerator/CodeGenerationHelper.cs
@@ -138,6 +138,8 @@ namespace Microsoft.ML.CLI.CodeGenerator
                                 logger.Log(LogLevel.Error, Strings.UnsupportedMlTask);
                                 break;
                         }
+
+                        t.IsBackground = true;
                         t.Start();
 
                         pbar.CompletedHandle.WaitOne(wait);
@@ -187,6 +189,8 @@ namespace Microsoft.ML.CLI.CodeGenerator
                             logger.Log(LogLevel.Error, Strings.UnsupportedMlTask);
                             break;
                     }
+
+                    t.IsBackground = true;
                     t.Start();
                     Thread.Sleep(wait);
                     context.Log -= ConsumeAutoMLSDKLog;
--- a/src/mlnet/Program.cs
+++ b/src/mlnet/Program.cs
@@ -20,7 +20,7 @@ namespace Microsoft.ML.CLI
     class Program
     {
         private static Logger logger = LogManager.GetCurrentClassLogger();
-        public static void Main(string[] args)
+        public static int Main(string[] args)
         {
             var telemetry = new MlTelemetry();
             int exitCode = 1;
@@ -101,7 +101,7 @@ namespace Microsoft.ML.CLI
             }
 
             parser.InvokeAsync(parseResult).Wait();
-            Environment.Exit(exitCode);
+            return exitCode;
         }
     }
 }


srsaggam:
@eerhardt @vinodshanbhag @justinormont

src/mlnet/CodeGenerator/CodeGenerationHelper.cs
```diff
@@ -138,6 +138,8 @@ public void GenerateCode()
                                 logger.Log(LogLevel.Error, Strings.UnsupportedMlTask);
                                 break;
                         }
+
+                        t.IsBackground = true;
```
justinormont: Fyi for others... as I had to look it up:

> A thread is either a background thread or a foreground thread. Background threads are identical to foreground threads, except that **background threads do not prevent a process from terminating**. Once all foreground threads belonging to a process have terminated, the common language runtime ends the process. Any remaining background threads are stopped and do not complete.

src/mlnet/CodeGenerator/CodeGenerationHelper.cs
```diff
@@ -187,6 +189,8 @@ public void GenerateCode()
                             logger.Log(LogLevel.Error, Strings.UnsupportedMlTask);
                             break;
                     }
+
+                    t.IsBackground = true;
```
justinormont: While we have control that our threads are background threads, we don't know that (of have control of) the learner's threads are background threads.

This will lead to not exiting properly.

srsaggam: Yeah that could be a problem.

eerhardt: I would think that any threads spawned by ML.NET should be background threads. And if they aren't, I would venture a guess that it is a bug that should be fixed in the learners.

srsaggam: Yep looks like all of them are background threads except for: 
https://github.com/dotnet/machinelearning/blob/6ebfeb33cd891db7328d032c903c3571d91a9265/src/Microsoft.ML.FastTree/SumupPerformanceCommand.cs#L291

Looks like this is the only place where a foreground thread is started. This one looks like a command used by some tool to measuring performance. So that should be okay I guess.

eerhardt: @TomFinley @codemzs - any thoughts on why this thread needs to run on the foreground?

eerhardt: Oh, this is an `ICommand`, which will only run from the MAML program. I don't think it is possible to run this code from a public API in ML.NET.

eerhardt: Looking even further, this appears to be more of a baseline test than anything.  Maybe `SumupPerformanceCommand` can just be deleted and/or turned into a BenchmarkDotNet test instead. I don't see why this code needs to be part of the shipping product.

justinormont: Any idea what occurs with LightGBM? Is there other outside could we are unsure of?

Seems simple enough to stay with `Environment.Exit(exitcode)` and be more future proof.

eerhardt: Environment.Exit isn't the "typical" way of terminating the process.

What do you expect would happen with this code:

```C#
IDisposable resource = ....;
try
{
   // use the resource
   // ...

   Environment.Exit(0);
}
finally
{
   resource.Dispose();
}
```

or even:

```C#
using (IDisposable resource = ....)
{
   // use the resource
   // ...

   Environment.Exit(0);
}
```

Would you expect the resource to be disposed?

vinodshanbhag: Isn't disposing only meaningful if the process is going to stay alive?  The resources belonging to exiting process will automatically released by OS ?

eerhardt: It depends on what you are disposing. `Dispose()` could be closing a connection somewhere else, or flushing log files so they get written to disk.

PR #3658, opened 2019-05-03
[AutoML] bump version to 0.4.0

Dmitry-A:


CESARDELATORRE:
Are we bumping the CLI to version to 0.4.0 as well?

eerhardt:
Yes. They both get their versions from the same file.

PR #3667, opened 2019-05-06
Fix to the official build (API Compat tool)

### LINKED ISSUE #3666: Official build failure due to the API Compat tool PR
PR #3623 breaks the official tool.

image

artidoro:
PR #3623 breaks the official build because the .Net Core 2.1 runtime is not found.
We have it locally in the tools directory, so a solution is to make it point to that local installation of the runtime. 

Fixes #3666.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -24,8 +24,7 @@
 <!-- API Compat -->
   <PropertyGroup>
     <RunApiCompat Condition="'$(RunApiCompat)' == ''">$(IsStableProject)</RunApiCompat>
-    <ToolHostCmd Condition="'$(OS)' != 'Windows_NT'">$(ToolsDir)dotnetcli/dotnet</ToolHostCmd>
-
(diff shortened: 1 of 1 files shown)


src/Directory.Build.props
```diff
@@ -24,8 +24,7 @@
 <!-- API Compat -->
   <PropertyGroup>
     <RunApiCompat Condition="'$(RunApiCompat)' == ''">$(IsStableProject)</RunApiCompat>
-    <ToolHostCmd Condition="'$(OS)' != 'Windows_NT'">$(ToolsDir)dotnetcli/dotnet</ToolHostCmd>
```
TomFinley: Oh, that's interesting. So we were assuming that if it was windows, *then* we wouldn't need to use the tools directory. Fun stuff. 😄

artidoro: Yes, on the CI windows machines most likely it was installed by default, but not on the official build machines.

PR #3696, opened 2019-05-09
Sample of using LoadFromEnumerable with a SchemaDefinition

sfilipi:
I have seen a few asks for how to bypass having to define the size of the vector in the data models. 
Adding an example for it.

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/LoadFromEnumerable.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/LoadFromEnumerable.cs
@@ -1,8 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
-
-using System;
+﻿using System;
 using System.Collections.Generic;
 using Microsoft.ML;
 using Microsoft.ML.Data;
@@ -11,7 +7,11 @@ namespace Samples.Dynamic
 {
     public static class LoadFromEnumerable
     {
-        // A simple case of creating IDataView from IEnumerable.
+        // Creating IDataView from IEnumerable, and setting the size of the vector at runtime. 
+        // When the data model is defined through types, setting the size of the vector is done through the VectorType 
+        // annotation. When the size of the data is not known at compile time, the Schema can be directly modified at runtime
+        // and the size of the vector set there. 
+        // This is important, because most of the ML.NET trainers require the Features vector to be of known size. 
         public static void Example()
         {
             // Create a new context for ML.NET operations. It can be used for exception tracking and logging, 
@@ -45,7 +45,8 @@ namespace Samples.Dynamic
                new DataPoint{ Features = new float[]{ 1.6f, 3.5f, 4.5f } },
             };
 
-            // The feature dimension retrievable at runtime.
+            // The feature dimension (typically this will be the Count of the array of the features vector
+            // known at runtime).
             int featureDimension = 3;
             var definedSchema = SchemaDefinition.Create(typeof(DataPoint));
             featureColumn = definedSchema["Features"].ColumnType as VectorDataViewType;
@@ -54,7 +55,7 @@ namespace Samples.Dynamic
             // Preview
             //
             // Is the size of the Features column known? False.
-            //Size: 0.
+            // Size: 0.
 
             // Set the column type to be a known-size vector.
             var vectorItemType = ((VectorDataViewType)definedSchema[0].ColumnType).ItemType;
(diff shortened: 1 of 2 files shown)


docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/LoadFromEnumerable.cs
```diff
@@ -0,0 +1,83 @@
+﻿using System;
```
shauheen: Please add headers

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/LoadFromEnumerable.cs
```diff
@@ -0,0 +1,87 @@
...
+            // If the size of the vector is unknown at compile time, it can be set at runtime.
+            IEnumerable<DataPoint> enumerableUnknownSize = new DataPoint[]
+            {
+               new DataPoint{ Features = new float[]{ 1.2f, 3.4f, 4.5f } },
+               new DataPoint{ Features = new float[]{ 4.2f, 3.4f, 1.6f } },
+               new DataPoint{ Features = new float[]{ 1.6f, 3.5f, 4.5f } },
+            };
+
+            // The feature dimension retrievable at runtime.
+            int featureDimension = 3;
```
abgoswam: is this a misplaced comment ?  .. featureDimension is not used until line 61   for setting , not retrieving ..

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/LoadFromEnumerable.cs
```diff
@@ -0,0 +1,87 @@
...
+            // Size: 3.
+        }
+    }
+
+    public class DataPoint
+    {
+        public float[] Features { get; set; }
+    }
+
+    public class DataPointVector
```
glebuk: >DataPointVector [](start = 17, length = 15)

Optionally define a constructor with float[] params features so that it's more compact to use.

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/LoadFromEnumerable.cs
```diff
@@ -0,0 +1,87 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.ML;
+using Microsoft.ML.Data;
+
+namespace Samples.Dynamic
+{
+    public static class LoadFromEnumerable
```
glebuk: >LoadFromEnumerable [](start = 24, length = 18)

Add comment explaining that you are not only loading from enumerable but also setting vector metadata properties. Why whould this be useful IRL?

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/LoadFromEnumerable.cs
shmoradims: Do we need headers for samples? whatever we put here, will show up in the docs verbatim, so we didn't include the license header in any of the samples. .NET samples don't have headers either (example below). So I suggest not including headers for samples files.

[https://docs.microsoft.com/en-us/dotnet/api/system.io.file?view=netframework-4.8](https://docs.microsoft.com/en-us/dotnet/api/system.io.file?view=netframework-4.8)


---
In reply to: [282706294](https://github.com/dotnet/machinelearning/pull/3696#discussion_r282706294) [](ancestors = 282706294)

sfilipi: i forgot, i'll remove them.

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/LoadFromEnumerable.cs
sfilipi: The sample is for cases when we don't know the size at compile time, so this is is set to a constant, but in real life would be the size of some array. 

But good point, i don't have all that language there, and i don't think it is necessary; so I'll remove the comment :)

---
In reply to: [283514291](https://github.com/dotnet/machinelearning/pull/3696#discussion_r283514291) [](ancestors = 283514291)

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/LoadFromEnumerable.cs
```diff
@@ -0,0 +1,88 @@
...
+            // Preview
+            //
+            // Is the size of the Features column known? False.
+            // Size: 0.
+
+            // Set the column type to be a known-size vector.
+            var vectorItemType = ((VectorDataViewType)definedSchema[0].ColumnType).ItemType;
+            definedSchema[0].ColumnType = new VectorDataViewType(vectorItemType, featureDimension);
+
+            // Read the data into an IDataView with the schema supplied in
```
wschin:
```suggestion
            // Read the data into an IDataView with the modified schema supplied in
```

PR #3721, opened 2019-05-14
Brew install libomp 7.0.0 only.

codemzs:
Stop gap fix until #3722 is fixed.


codemzs:
@eerhardt I’m no home brew expert either but based on my research this afternoon it seems this is apparently the right way to install an earlier version of libomp. Thanks for approving.

Issue #3729, opened 2019-05-15
CLI does not support other cultures besides en-US


tbombach:
### System information

- **OS version/distro**: Windows 10
- **.NET Version (eg., dotnet --info)**: 3.0.100-preview-5-011568

### Issue

- **What did you do?**
Ran CLI auto-train on a TSV file with culture set to Finnish.
1. Set Windows Regional format to Finnish.
    - Download the Finnish language pack from the Microsoft store (it's called suomi): https://www.microsoft.com/store/productId/9MW3PQ7SD3QK
    - Change the format to "Finnish (Finland)" in `Windows settings -> Time & Language -> Region -> Regional format`
2. Run CLI:
    - `mlnet auto-train --task regression --dataset taxi-fare-train.csv --label-column-name fare_amount -- max-exploration-time 10`
    - File: https://github.com/dotnet/machinelearning-samples/blob/master/samples/CLI/Regression_CLI/taxi-fare-train.csv

- **What happened?**
Error:
```
Exception occured while exploring pipelines:
Provided label column 'fare_amount' was of type String, but only type Single is allowed.
Please see the log file for more info.
Exiting ...
```
- **What did you expect?**
Successful model training. Switching the regional format back to "English (United States)" allows training to complete successfully.

### Source code / logs

Please paste or attach the code or logs or traces that would be helpful to diagnose the issue you are reporting.

rustd:
This was reported by customers using Model Builder. Adding automl and gui labels to highlight that.

srsaggam:
Thank you for filing this issue. Looks like we missed out on culture info on some of the places in AutoML API. This shall be fixed in coming versions.

Please note that this bug is different from #3677 
It is the same issue but happened in the code generation engine of CLI.

justinormont:
Closed by https://github.com/dotnet/machinelearning/pull/3731

PR #3731, opened 2019-05-16
[AutoML] Fix for Column inference for R4 type column in different cultures 

srsaggam:
Looks like the column inference is getting wrong in some of the cultures like Deutsch and Finnish . 
So making the parsing InvariantCulture

### DIFF
--- a/src/Microsoft.ML.AutoML/Utils/MLNetUtils/Conversions.cs
+++ b/src/Microsoft.ML.AutoML/Utils/MLNetUtils/Conversions.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Globalization;
 
 namespace Microsoft.ML.AutoML
 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.AutoML/Utils/MLNetUtils/Conversions.cs
```diff
@@ -25,7 +26,7 @@ public static bool TryParse(in TX src, out R4 dst)
                 dst = R4.NaN;
                 return true;
             }
-            if (float.TryParse(str, out dst))
+            if (float.TryParse(str, NumberStyles.Float, CultureInfo.InvariantCulture, out dst))
```
justinormont: Any idea what the ML.NET TextLoader does? Does the TextLoader handle reading as `InvariantCulture` correctly?

I'm worried that infer the value is a number in the AutoML code, then the TextLoader won't be able to read the value since it will read as the local culture. If so, we'd likely want to also correct the TextLoader.

justinormont:
Is this to fix https://github.com/dotnet/machinelearning/issues/3729?

srsaggam:
@justinormont yes thats the issue. Thanks for adding reference

PR #3776, opened 2019-05-25
[AutoML] Reservoir sample dataset statistics

### LINKED ISSUE #3778: [AutoML] Reservoir sample dataset statistics
Currently dataset statistics within AutoML are calculated from the first 1,000 rows of a dataset. Instead, we should be calculating statistics from a random sample of 1,000 rows. (First 1,000 rows could be biased if they are sorted by label, any other column, time of collection, etc.) We can use reservoir sampling to obtain a random sample of a fixed size in a single pass over the dataset

daholste:
Closes #3778

### DIFF
--- a/src/Microsoft.ML.AutoML/DatasetDimensions/ReservoirSampledDataView.cs
+++ b/src/Microsoft.ML.AutoML/DatasetDimensions/ReservoirSampledDataView.cs
@@ -226,7 +226,7 @@ namespace Microsoft.ML.AutoML
         /// </summary>
         private class SingleSlotSampledColumn<T> : SampledColumn<T>
         {
-            public SingleSlotSampledColumn(int size) : base(size)
+            public SingleSlotSampledColumn(long size) : base(size)
             {
             }
 
@@ -241,7 +241,7 @@ namespace Microsoft.ML.AutoML
         /// </summary>
         private class VectorSampledColumn<T> : SampledColumn<VBuffer<T>>
         {
-            public VectorSampledColumn(int size) : base(size)
+            public VectorSampledColumn(long size) : base(size)
             {
             }
 
--- a/src/Microsoft.ML.AutoML/DatasetDimensions/ReservoirSampledDataView.cs
+++ b/src/Microsoft.ML.AutoML/DatasetDimensions/ReservoirSampledDataView.cs
@@ -248,7 +248,7 @@ namespace Microsoft.ML.AutoML
         }
 
         /// <summary>
-        /// This class pulls from an original / parent data view column
+        /// Pulls from an original / parent data view column
         /// to populate the corresponding sampled column.
         /// </summary>
         private class ColumnSampleBuilder<T> : ColumnSampleBuilder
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.AutoML/DatasetDimensions/ReservoirSampledDataView.cs
```diff
@@ -0,0 +1,321 @@
...
+        private readonly object[] _sampledColumns;
+        private readonly long _sampleSize;
+
+        private bool _initialized = false;
+        private long _originalDataRowCount;
+
+        public ReservoirSampledDataView(IDataView data, long sampleSize)
+        {
+            _originalData = data;
+            _random = new Random();
```
justinormont: May want to expose a seed value as an optional parameter. Having a stable sample should make unit tests easier, both in the reservoir sample tests, and the statistics unit tests.

src/Microsoft.ML.AutoML/DatasetDimensions/ReservoirSampledDataView.cs
```diff
@@ -0,0 +1,321 @@
...
+            _sampledColumns = new object[Schema.Count];
+            _sampleSize = sampleSize;
+        }
+
+        public bool CanShuffle => false;
```
justinormont: Shouldn't be too hard to add shuffle support. Granted it's unneeded AFIK and adding `ShuffleRows()` afterwards will work at the cost of memory usage.

src/Microsoft.ML.AutoML/DatasetDimensions/ReservoirSampledDataView.cs
```diff
@@ -0,0 +1,315 @@
...
+            {
+                // Cache first _sampleSize rows
+                if (rowIdx < _sampleSize)
+                {
+                    sampleBuilder.StoreNextRow(rowIdx);
+                }
+                else
+                {
+                    // Reservoir sample row
+                    var idx = GetRandomLong(rowIdx + 1);
```
justinormont: I suspect a multiply is a faster than the current function w/ a mod...
```suggestion
                    var idx = rand.NextDouble() * (rowIdx + 1);
```

src/Microsoft.ML.AutoML/DatasetDimensions/ReservoirSampledDataView.cs
daholste: Added

src/Microsoft.ML.AutoML/DatasetDimensions/ReservoirSampledDataView.cs
daholste: You're right lol -- I timed this out
Generating 10 million random #s w/ the rand.NextDouble() took 228 milliseconds
Generating 10 million random #s w/ the existing method took 1081 milliseconds
Both are quite fast, which is good.
This is probably a theoretical concern instead of an actual issue -- but b/c of the fixed double precision, I find myself wondering if rand.NextDouble() * a long that is sufficiently large could produce a less than random distribution. For instance, in the extreme, I observe that 1.0 / (long.MaxValue/100 - 5) == 1.0 / (long.MaxValue/100 - 4) evaluates to true. Clearly, we would never have anywhere NEAR long.MaxValue rows, but this is just to illustrate the point at the extreme. I have a strong hunch this is more of a theoretical concern than a practical one, but just throwing it out there

PR #3788, opened 2019-05-29
Fix TextLoader constructor and add exception message

### LINKED ISSUE #3705: Context.Data.CreateTextLoader<T> throws error Can't determine the number of source columns without valid data
### System information

- **OS version/distro**: Windows 10
AutoML 0.3.0
Microsoft.ML 1.0.0.0

### Issue
I thought one can map a class and load it with the annotation but get an error when calling it on a parameter that's not available for me.

- **What did you do?**
I wrote a little test to play with Microsoft.ML.AutoML and test it against a label and a vector
I add my litle [Program](https://github.com/dotnet/machinelearning/files/3168931/Program.zip)
where I call in the ticket

basically I cal:
```
public static IDataView GetDataView<T>(MLContext mlContext, FileInfo trainingFile)
{
    var loader = mlContext.Data.CreateTextLoader<T>(separatorChar: '|', hasHeader: false);
    return loader.Load(trainingFile.FullName);
            
}
public class Data
{
     [LoadColumn(0)]
     public string Label { get; }

     [LoadColumn(1, 40_731)]
     [VectorType(40_730)]
     public float[] Features { get; }
}
```

- **What happened?**
I get an error stating
> System.ArgumentOutOfRangeException: 'Can't determine the number of source columns without valid data
> Parameter name: Source'

Stack
```
   at Microsoft.ML.Data.TextLoader.Bindings..ctor(TextLoader parent, Column[] cols, IMultiStreamSource headerFile, IMultiStreamSource dataSample)
   at Microsoft.ML.Data.TextLoader..ctor(IHostEnvironment env, Options options, IMultiStreamSource dataSample)
   at Microsoft.ML.Data.TextLoader.CreateTextLoader[TInput](IHostEnvironment host, Boolean hasHeader, Char separator, Boolean allowQuoting, Boolean supportSparse, Boolean trimWhitespace, IMultiStreamSource dataSample)
   at ConsoleMLWizard.Program.GetDataView[T](MLContext mlContext, FileInfo trainingFile) in C:\Users\W2307\source\repos\ConsoleMLWizard\Program.cs:line 99
   at ConsoleMLWizard.Program.Main(String[] args) in C:\Users\W2307\source\repos\ConsoleMLWizard\Program.cs:line 37
```

- **What did you expect?**

I expected the anotations to work, I can load the data fine using:
```
var loader = context.Data.CreateTextLoader(options: new TextLoader.Options()
{
    Columns = new[] {
        new TextLoader.Column(name:"Label", dataKind: DataKind.String, index: 0),
        new TextLoader.Column(name:"Features",dataKind:DataKind.Single,minIndex:0,maxIndex:40731)
    },
    HasHeader = false,
    Separators = new[] { ',' },
                
});
var data = loader.Load(dataFile.FullName);
```

artidoro:
Fixes #3705.

In issue #3705 it appeared that the class defining the data was required to have both get and **set** auto-properties. In `TextLoader` however, we don't use the set property, so I removed that from the conditions.

Second, I added an exception message in case no public readable property or field is found in the class defining the data and no dataSample is passed to the constructor.

I added a related test which tests the various scenarios.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1460,8 +1460,8 @@ namespace Microsoft.ML.Data
 
             var memberInfos = (fieldInfos as IEnumerable<MemberInfo>).Concat(propertyInfos).ToArray();
 
-            if (memberInfos.Length == 0 && dataSample == null)
-                throw host.ExceptParam(nameof(TInput), $"Should define at least one public, readable field or property in {nameof(TInput)} or provide a {nameof(dataSample)}.");
+            if (memberInfos.Length == 0)
+                throw host.ExceptParam(nameof(TInput), $"Should define at least one public, readable field or property in {nameof(TInput)}.");
 
             var columns = new List<Column>();
 
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1456,9 +1456,12 @@ namespace Microsoft.ML.Data
             var propertyInfos =
                 userType
                 .GetProperties(BindingFlags.Public | BindingFlags.Instance)
-                .Where(x => x.CanRead && x.CanWrite && x.GetGetMethod() != null && x.GetSetMethod() != null && x.GetIndexParameters().Length == 0);
+                .Where(x => x.CanRead && x.GetGetMethod() != null && x.GetIndexParameters().Length == 0);
 
             var memberInfos = (fieldInfos as IEnumerable<MemberInfo>).Concat(propertyInfos).ToArray();
+
+            if (memberInfos.Length == 0 && dataSample == null)
+                throw host.ExceptParam(nameof(TInput), $"Should define at least one public, readable field or property in {nameof(TInput)} or provide a {nameof(dataSample)}.");
 
             var columns = new List<Column>();
 
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
@@ -62,6 +62,9 @@ namespace Microsoft.ML
         /// <summary>
         /// Create a text loader <see cref="TextLoader"/> by inferencing the dataset schema from a data model type.
         /// </summary>
+        /// <typeparam name="TInput">Defines the schema of the data to be loaded. Use public fields or properties
+        /// decorated with <see cref="LoadColumnAttribute"/> (and possibly other attributes) to specify the column
+        /// names and their data types in the schema of the loaded data.</typeparam>
         /// <param name="catalog">The <see cref="DataOperationsCatalog"/> catalog.</param>
         /// <param name="separatorChar">Column separator character. Default is '\t'</param>
         /// <param name="hasHeader">Does the file contains header?</param>
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
@@ -68,11 +68,8 @@ namespace Microsoft.ML
         /// <param name="catalog">The <see cref="DataOperationsCatalog"/> catalog.</param>
         /// <param name="separatorChar">Column separator character. Default is '\t'</param>
         /// <param name="hasHeader">Does the file contains header?</param>
-        /// <param name="dataSample">The optional location of a data sample. The sample can be used to infer column
-        /// names and number of slots in each column. If <typeparamref name="TInput"/> does not contain any public
-        /// fields, we assume that the first column in the <paramref name="dataSample"/> is the Label column and
-        /// the second is the Features column. If there are more than two columns, the Features column is taken
-        /// to be a vector column.</param>
+        /// <param name="dataSample">The optional location of a data sample. The sample can be used to infer information
+        /// about the columns, such as slot names.</param>
         /// <param name="allowQuoting">Whether the input may include quoted values,
         /// which can contain separator characters, colons,
         /// and distinguish empty values from missing values. When true, consecutive separators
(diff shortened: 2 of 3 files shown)


test/Microsoft.ML.Tests/TextLoaderTests.cs
```diff
@@ -795,5 +796,77 @@ public void TestTextLoaderKeyTypeBackCompat()
                 Assert.True(result.Schema[featureIdx].Type is KeyDataViewType keyType && keyType.Count == typeof(uint).ToMaxInt());
             }
         }
+
+        public class IrisNoFields
```
wschin: Is it possible to do `private class`?

test/Microsoft.ML.Tests/TextLoaderTests.cs
```diff
@@ -795,5 +796,77 @@ public void TestTextLoaderKeyTypeBackCompat()
...
+        }
+
+        [Fact]
+        public void TestTextLoaderNoFields()
+        {
+            var dataPath = GetDataPath(TestDatasets.irisData.trainFilename);
+            var mlContext = new MLContext();
+
+            // Class with get property only.
+            var dataIris = mlContext.Data.CreateTextLoader<IrisPublicGetProperties>(separatorChar: ',').Load(dataPath);
```
wschin: Could you update the doc of `CreateTextLoader<T>` to explain what is allowed?

artidoro: Will do!

artidoro:
I realize I need to use starts with, since the new line characters change for ubuntu/mac, which is what is causing the test failures.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1449,10 +1449,13 @@ void ICanSaveModel.Save(ModelSaveContext ctx)
             var propertyInfos =
                 userType
                 .GetProperties(BindingFlags.Public | BindingFlags.Instance)
-                .Where(x => x.CanRead && x.CanWrite && x.GetGetMethod() != null && x.GetSetMethod() != null && x.GetIndexParameters().Length == 0);
+                .Where(x => x.CanRead && x.GetGetMethod() != null && x.GetIndexParameters().Length == 0);
 
             var memberInfos = (fieldInfos as IEnumerable<MemberInfo>).Concat(propertyInfos).ToArray();
 
+            if (memberInfos.Length == 0 && dataSample == null)
```
yaeldekel: >dataSample [](start = 43, length = 10)

Maybe this should throw even if `dataSample` is not null?
If it is not null, then `TextLoader` will create a schema with two columns `Label` and `Features`, but it will not be able to read `Label` and `Features` from objects of type `TInput` - wouldn't it throw somewhere downstream in that case?

artidoro: I did not check I will see what happens in that case.

test/Microsoft.ML.Tests/TextLoaderTests.cs
artidoro: If your question is whether the class defining the data can be private, the answer is yes.
If you are asking to make these classes private in the test, I would prefer to keep them public since right above they were defined this way. I would rather be consistent than changing here only.

wschin: I feel public surface should be as small as possible. It helps both devs (they know which thing not to look for example) and users (user only accepts necessary information) a lot.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -62,10 +62,17 @@ public static class TextLoaderSaverCatalog
...
         /// </summary>
+        /// <typeparam name="TInput">Defines the schema of the data to be loaded. Use public fields or properties
+        /// decorated with <see cref="LoadColumnAttribute"/> (and possibly other attributes) to specify the column
+        /// names and their data types in the schema of the loaded data.</typeparam>
         /// <param name="catalog">The <see cref="DataOperationsCatalog"/> catalog.</param>
         /// <param name="separatorChar">Column separator character. Default is '\t'</param>
         /// <param name="hasHeader">Does the file contains header?</param>
-        /// <param name="dataSample">The optional location of a data sample. The sample can be used to infer column names and number of slots in each column.</param>
+        /// <param name="dataSample">The optional location of a data sample. The sample can be used to infer column
+        /// names and number of slots in each column. If <typeparamref name="TInput"/> does not contain any public
```
wschin: What are the types of feature and label columns? What do you mean by `first column` in `dataSample`? Column only exists in `IDataView`.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
yaeldekel: I feel that if the user wants `TextLoader` to automatically load a `Label` and a `Features` column, they should use one of the non-generic overloads of the `CreateTextLoader` API. If they choose to use the generic overload, we should assume that their intent is to load the fields/properties in the class, and if `memberInfos` ends up being empty, then it means they didn't define the class properly, thus we should throw whether or not they provide a `dataSample`. What do you think?

---
In reply to: [290417237](https://github.com/dotnet/machinelearning/pull/3788#discussion_r290417237) [](ancestors = 290417237)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -62,10 +62,17 @@ public static class TextLoaderSaverCatalog
...
+        /// fields, we assume that the first column in the <paramref name="dataSample"/> is the Label column and
+        /// the second is the Features column. If there are more than two columns, the Features column is taken
```
yaeldekel: This is great! Could you add this point to the other overloads of this API as well? (the one taking the `Options` object, and the one taking the `Column` array directly)

test/Microsoft.ML.Tests/TextLoaderTests.cs
artidoro: Ok I understand your point! I will update this to be private.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
artidoro: I agree I did that after we discussed.

PR #3789, opened 2019-05-30
[AutoML] CLI telemetry rev

daholste:
Add telemetry around:
 - Infer Columns
 - System Info
 - Experiment iteration results
 - Overall experiment results
 - Application Exit

Also, add error logging and duration tracking for events
Closes https://github.com/dotnet/machinelearning-automl/issues/417 by componentizing the instrumentation of the CLI tool


justinormont:
PR could use a description. Perhaps:
```Closes https://github.com/dotnet/machinelearning-automl/issues/417 by componentizing the instrumentation of the AutoML CLI tool and adding xyz"

File: src/mlnet/Telemetry/Telemetry.cs [hunk 1]
```diff
@@ -1,144 +1,239 @@
...
+            {
+                sb.Append(transformNode.Name);
+                sb.Append(",");
+            }
+            if (trainerNode != null)
+            {
+                sb.Append(trainerNode.Name);
+                sb.Append("{");
+                var serializedHyperparams = trainerNode.Properties
+                    .Where(p => SweepableParams.AllHyperparameterNames.Contains(p.Key))
```
Dmitry-A: trying to understand the intent here..  why not log all of them?  defaults can change from dataset to dataset and also over time for hyperparams we don't explicitly set no?  which would be an interesting bit of info as well.

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
```diff
@@ -172,5 +177,29 @@ public static IEnumerable<SweepableParam> BuildSymSgdLogisticRegressionParams()
                 new SweepableDiscreteParam("UpdateFrequency", new object[] { "<Auto>", 5, 20 })
             };
         }
+
+        /// <summary>
+        /// Gets the name of every hyperparameter swept across all trainers.
+        /// </summary>
+        public static ISet<string> GetAllSweepableParameterNames()
+        {
+            var sweepableParams = new List<SweepableParam>();
```
Dmitry-A: is there a mechanism to enforce not to forget to update this list when new components are added?  looks to me this somehow needs to work off of a catalog and not have an explicit list at all

src/mlnet/Program.cs
```diff
@@ -20,24 +21,33 @@ namespace Microsoft.ML.CLI
...
+            var mlNetCommandEvent = new MLNetCommandEvent();
+
             // Create handler outside so that commandline and the handler is decoupled and testable.
             var handler = CommandHandler.Create<NewCommandSettings>(
                 (options) =>
              {
                  try
                  {
+                     // Send telemetry event for command issued
+                     mlNetCommandEvent.AutoTrainCommandSettings = options;
```
Dmitry-A: Not grok'ing the design here.  If it's a generic ml.net command event, then it shouldn't have anything automl specific.  If it's automl.net specific thing then why not call it that.  I'm trying to prevent an uber (garbage can) class where every command tosses whatever it wants.    The initial MlTelemetry class was flawed in that way too since I didn't anticipate having more than one event per command, but at least it didn't have any command- or event-specific members, just methods plus global stuff like logging command line params.

src/mlnet/Telemetry/Events/MLNetCommandEvent.cs
```diff
@@ -0,0 +1,59 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.DotNet.Cli.Telemetry;
+using Microsoft.ML.CLI.Data;
+using Microsoft.ML.CLI.Utilities;
+
+namespace Microsoft.ML.CLI.Telemetry.Events
+{
+    internal class MLNetCommandEvent
```
Dmitry-A: comment or a more descriptive name would be nice here, something like "AutoTrainCommandInvocationEvent".

src/mlnet/Telemetry/Events/SystemInfoEvent.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.ML.CLI.Telemetry.Events
+{
+    /// <summary>
+    /// System info telemetry event.
+    /// </summary>
+    internal class SystemInfoEvent
```
Dmitry-A: why is this not a part of the initial invocation event?

src/mlnet/Telemetry/Telemetry.cs
```diff
@@ -1,144 +1,239 @@
...
                 return;
             }
-
-            //continue task in existing parallel thread
-            _trackEventTask = _trackEventTask.ContinueWith(
-                x => TrackEventTask(eventName, properties, measurements)
-            );
+            // Increment count of outstanding telemetry events.
+            Interlocked.Increment(ref _outstandingTelemetryEventCount);
+            _initializationTask.ContinueWith(x => TrackEventTask(eventName, properties, duration, ex));
```
Dmitry-A: since the task is used for not just initialization, the _intializationTask name is a bit decieving

src/mlnet/Telemetry/Telemetry.cs
```diff
@@ -1,144 +1,239 @@
...
 
-        public void ThreadBlockingTrackEvent(string eventName, IDictionary<string, string> properties, IDictionary<string, double> measurements)
+        /// <summary>
+        /// Wait for all outstanding telemetry tasks to complete.
+        /// </summary>
+        public static void Flush(TimeSpan timeout)
         {
-            if (!Enabled)
+            var sw = Stopwatch.StartNew();
+            while (sw.Elapsed < timeout)
```
Dmitry-A: can you just check _intializationTask for IsCompleted()?

src/mlnet/Telemetry/Telemetry.cs
```diff
@@ -1,144 +1,239 @@
...
         {
-            if (_client == null)
+            if (_client != null)
             {
-                return;
+                try
+                {
+                    var eventProperties = GetEventProperties(properties, duration, ex);
+                    _client.TrackEvent(eventName, eventProperties);
+                    _client.Flush();
```
Dmitry-A: Thinking out loud here.  I am not convinced doing a Flush() here makes sense anymore.  Some events like iteration events with a high probability of the process continuing to run would do well with Flush() right there.  But in worst case scenario where we posted some events and now want to exit, doing a Flush() on each of them would likely make it not fit in 200ms time box that you have.  I think a more generic way would be to have an event queue that would accept events and only do a flush if explicitly told so (on exit of the process) or if there's been say more than 100ms idle time with the queue empty.

src/mlnet/Telemetry/Telemetry.cs
```diff
@@ -1,144 +1,239 @@
...
         }
+
+        /// <summary>
+        /// Get exception string to log. Exclude the exception message, as it
+        /// may contain PII.
+        /// </summary>
+        private static string GetSanitizedExceptionStr(Exception ex)
+        {
+            return $@"{ex.GetType()}
+{ex.StackTrace}";
```
Dmitry-A: Thinking out loud..  if the exception somehow happened in user code, we will get their names as part of the stack trace.  Maybe also sanitize the stack trace itself to only include stuff from Microsoft.* namespace?

src/mlnet/Telemetry/Telemetry.cs
Dmitry-A: Thinking out loud here too: if the user has any custom stuff in the pipeline, can we end up logging PII here?  Or I guess at this point there can't be any custom components so we should be Ok but I don't like the fact that this is here and when we implement a feature to prepend custom things to pipelines someone might not think about telemetry/PII

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
```diff
@@ -61,6 +61,11 @@ private static IEnumerable<SweepableParam> BuildLbfgsArgsParams()
             };
         }
 
+        /// <summary>
+        /// The names of every hyperparameter swept across all trainers.
+        /// </summary>
+        public static ISet<string> AllHyperparameterNames = GetAllSweepableParameterNames();
```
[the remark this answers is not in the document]
daholste: thanks for catching

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
daholste: no, there is not. I had the same concern. hopefully, a dev would click 'explore usages' of the existing methods in VS, and see all the usages, but this is not guaranteed, +1.
'LabelColumnName', 'ExampleWeightColumnName' etc are trainer properties here. We could use a blacklist instead, but I chose a whitelist to err hard on the side of not logging PII
would you be able to elaborate on your catalog idea in a concrete way? I would love a better solution to this, +1

src/mlnet/CodeGenerator/CodeGenerationHelper.cs
```diff
@@ -51,7 +52,9 @@ public void GenerateCode()
                 {
                     inputColumnInformation.IgnoredColumnNames.Add(value);
                 }
+                var inferColumnsStopwatch = Stopwatch.StartNew();
                 columnInference = automlEngine.InferColumns(context, inputColumnInformation);
+                InferColumnsEvent.TrackEvent(columnInference.ColumnInformation, inferColumnsStopwatch.Elapsed);
```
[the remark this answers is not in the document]
daholste: Let's say we have a per-iteration event, and on-start event, and on-exit event. If the user does 100 iterations, that's 102 log statements. If we increase from 102 to the current 104, from a network round trip perspective, I think we're in the same boat. However, from a telemetry / logging / style perspective, it could be cleaner to have a separate events for InferColumns & ApplicationStart / etc.

src/mlnet/Program.cs
daholste: +1. If I were designing this from scratch, I would not do this.
The original telemetry event was called "mlnet-command". It has AutoML-specific properties like hash of the train file name and name of the learning task (ie multiclass classification).
I would like to change the name of this event from "mlnet-command" to "auto-train-command", but if we do that, then we'd bifurcate our prod metrics & events. For backwards compatibility, I went this route -- but perhaps we should just make this breaking change. I'm not sure

src/mlnet/Telemetry/Events/MLNetCommandEvent.cs
daholste: +1. If I were designing this from scratch, I would not have chosen this name. However, this was the original name chosen for the AutoML invocation event. Please see further description in comment above / possible concern about backwards compatibility

src/mlnet/Telemetry/Events/SystemInfoEvent.cs
daholste: @justinormont pointed out that having a separate log statement here could be preferable, as not to clog up the "start command" event telemetry as the amount of system information grows

src/mlnet/Telemetry/Telemetry.cs
Dmitry-A: worth a look at how .NET CLI handles this now or how it did handle it before they implemented persisting unsent telemetry between invocations

src/mlnet/Telemetry/Telemetry.cs
daholste: Is anything else aside from initialization being done in the initialization task?

src/mlnet/Telemetry/Telemetry.cs
daholste: Even if _initializationTask is completed, there could still be outstanding telemetry events

src/mlnet/Telemetry/Telemetry.cs
daholste: "But in worst case scenario where we posted some events and now want to exit, doing a Flush() on each of them would likely make it not fit in 200ms time box that you have."
I could definitely be wrong, but I was thinking the opposite is true. These tasks are all done in parallel. The earlier we start each network request, the more likely it is that telemetry event has been sent in the case of an abrupt exit

src/mlnet/Telemetry/Telemetry.cs
daholste: Interesting. Right now, I don't think this is possible, of course. Are you thinking of this as a defensive move against future development / capabilities in this repository?

src/mlnet/Telemetry/Telemetry.cs
daholste: 'LabelColumnName', 'ExampleWeightColumnName' etc are trainer properties here. I erred hard by choosing only to log properties that were whitelisted in a file that only contained a bunch of hyperparameter names to err as hard away from logging PII as possible. let me know if there's a better approach

src/Microsoft.ML.AutoML/ColumnInference/ColumnInformationUtil.cs
```diff
@@ -106,6 +106,24 @@ public static IEnumerable<string> GetColumnNames(ColumnInformation columnInforma
             AddStringsToListIfNotNull(columnNames, columnInformation.TextColumnNames);
             return columnNames;
         }
+        
+        public static IDictionary<ColumnPurpose, int> CountColumnsByPurpose(ColumnInformation columnInformation)
+        {
+            var result = new Dictionary<ColumnPurpose, int>();
+            var columnNames = GetColumnNames(columnInformation);
+            foreach (var columnName in columnNames)
```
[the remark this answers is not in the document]
daholste: I love LINQ! Here, I think the existing way is more readable, but I see where you're coming from. Style is so idiosyncratic

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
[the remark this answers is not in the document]
daholste: actually, because this is a static property, apparently StyleCop is okay with it

src/mlnet/Telemetry/Events/SystemInfoEvent.cs
daholste: I realize that this is also laid out more here https://github.com/dotnet/machinelearning-automl/issues/417

src/mlnet/CodeGenerator/CodeGenerationHelper.cs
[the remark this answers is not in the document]
daholste: I realize that this is also laid out more here https://github.com/dotnet/machinelearning-automl/issues/417

daholste:
Thank you @Dmitry-A for taking the time to review this (in comments here and offline thru chats), and (among other improvements) working out improved parallelism and telemetry flush scheme

src/mlnet/Telemetry/Telemetry.cs
```diff
@@ -1,97 +1,146 @@
...
+        /// Flush outstanding telemetry, and wait for the specified timeout for this to complete.
+        /// </summary>
+        public static void Flush(TimeSpan timeout)
         {
-            if (!Enabled)
+            if (!_enabled || _client == null)
             {
                 return;
             }
+            Task.Run(() => _client.Flush()).Wait(timeout);
```
Dmitry-A: A bit if a nit but this will block the thread.  Probably not relevant in the way it is used currently since it's only when everything else is done.  But can become an issue in the future.   It's possible to do it async:  await Task.WhenAny(task, Task.Delay(timeout))

PR #3794, opened 2019-05-30
[AutoML] Cross validation fixes; validate empty training / validation input data

daholste:
Updates --
- In the AutoML API -- when calling the `Execute` method signature that contains both training & validation data, if validation data is null, route to the `Execute` method that doesn't contain validation data / that splits the data in a train/test or cross-val manner depending on the # of rows.
- Don't crash if some cross val splits have empty train/test data. As long as not all splits have empty train or test data, proceed only using splits w/ non-empty train & test sets. Explicitly throw an exception if all splits either have empty train or test sets.
- Validate input training & validation data (if explicitly provided) are not empty

### DIFF
--- a/src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
+++ b/src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
@@ -10,28 +10,37 @@ namespace Microsoft.ML.AutoML
 {
     internal static class SplitUtil
     {
-        private const string CrossValEmptyFoldErrorMsg = @"Cross validation split has 0 rows. Perhaps " +
-            "try increasing number of rows provided in training data, or lowering specified number of " +
-            "cross validation folds.";
-
         public static (IDataView[] trainDatasets, IDataView[] validationDatasets) CrossValSplit(MLContext context, 
             IDataView trainData, uint numFolds, string samplingKeyColumn)
         {
             var originalColumnNames = trainData.Schema.Select(c => c.Name);
             var splits = context.Data.CrossValidationSplit(trainData, (int)numFolds, samplingKeyColumnName: samplingKeyColumn);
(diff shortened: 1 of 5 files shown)


justinormont:
Does this also handle the error caused by the binary-classification scoring code crashing when both 1/0 are not in a particular fold? 

Error:
`System.ArgumentOutOfRangeException: AUC is not definied 
when there is no positive class in the data`

User reported dataset: https://github.com/dotnet/machinelearning/issues/3800#issue-450909708

daholste:
> Does this also handle the error caused by the binary-classification scoring code crashing when both 1/0 are not in a particular fold?
> 
> Error:
> `System.ArgumentOutOfRangeException: AUC is not definied when there is no positive class in the data`
> 
> User reported dataset: [#3800 (comment)](https://github.com/dotnet/machinelearning/issues/3800#issue-450909708)

No it does not fix this customer-encountered issue. Thank you for responding to the customer & pointing this out. After this PR goes in, I'll submit a fix on top of it for the binary AUC error

src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
```diff
@@ -10,28 +10,37 @@ namespace Microsoft.ML.AutoML
...
+                var validationDataset = DropAllColumnsExcept(context, split.TestSet, originalColumnNames);
+
+                trainDatasets.Add(trainDataset);
+                validationDatasets.Add(validationDataset);
             }
-            return (trainDatasets, validationDatasets);
+
+            if (!trainDatasets.Any())
+            {
+                throw new InvalidOperationException("All cross validation folds have empty train or test data. Perhaps " +
```
Dmitry-A: Nit: perhaps sounds unnecessary to me, usually it's just "Try bla".

PR #3796, opened 2019-05-30
Change ensembles trainer to work with ITrainerEstimators instead of ITrainers

### LINKED ISSUE #3709: Ensembles don't work
The trainer argument of Ensembles is of type `IComponentFactory<ITrainer<IPredictorProducing<>>>`.

However, none of the trainers in ML.NET implement `ITrainer<IPredictorProducing<>>`, they just implement `ITrainer<IPredictor>`, so there is no way to specify a different trainer in the arguments
(when running the maml command line).

yaeldekel:
This is needed for using ensembles with non-default values for the sub-trainers.
Fixes #3709 .


test/Microsoft.ML.Predictor.Tests/TestPredictors.cs
```diff
@@ -2181,6 +2181,135 @@ public void MulticlassNaiveBayes()
...
             Done();
         }
+
+        [Fact]
+        public void EnsemblesMultiClassBootstrapSelectorTest()
+        {
+            var pa = new PredictorAndArgs(new SubComponent("WeightedEnsembleMulticlass", "bp=mlr{t-} nm=20 st=BootstrapSelector{} tp=-"), "WE-Bootstrap");
+            Run_TrainTest(pa, TestDatasets.iris);
+            Done();
+        }
```
artidoro: I guess that these tests were taken from TLC. Did you have to regenerate the results or were they the same?

yaeldekel: Good question. I regenerated all the results, but I went back and compared them to the old baselines, and all but one were identical. The different one was for the CV case of the `EnsemblesDefaultTest`. The differences in the scores were something like this:

7.057208
-0.569736
-3.265584
-2.8438797
-3.42607021
-3.27301645
4.12974262
3.61101723
3.70798635
0.8509629

(the new scores) vs.:

7.346347
-0.308689833
-3.056211
-2.46506
-3.10131264
-2.99300337
4.336238
3.53593183
4.37107658
0.862161636

Do you think we can attribute these changes to a random seed change? Or should we investigate this further?


---
In reply to: [290953192](https://github.com/dotnet/machinelearning/pull/3796#discussion_r290953192) [](ancestors = 290953192)

artidoro: I think so! They look quite similar 

---
In reply to: [292166898](https://github.com/dotnet/machinelearning/pull/3796#discussion_r292166898) [](ancestors = 292166898,290953192)

PR #3802, opened 2019-06-02
Bump master to 1.2 and preview to 0.14.

codemzs:

### DIFF
--- a/build/BranchInfo.props
+++ b/build/BranchInfo.props
@@ -15,6 +15,7 @@
       Microsoft.ML.LightGbm;
       Microsoft.ML.Mkl.Components;
       Microsoft.ML.Mkl.Redist;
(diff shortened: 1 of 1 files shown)


build/BranchInfo.props
```diff
@@ -15,20 +15,21 @@
       Microsoft.ML.LightGbm;
       Microsoft.ML.Mkl.Components;
       Microsoft.ML.Mkl.Redist;
+      Microsoft.ML.TensorFlow;
```
eerhardt: Have we done the appropriate work to ensure the APIs we have in this package are the correct long term APIs we want to go with?

eerhardt: We are confident with the APIs even with respect to supporting TensorFlow 2.0?

PR #3809, opened 2019-06-03
Check for number of input columns in concat transform

### LINKED ISSUE #3061: Null Reference Exception when Concatenating with a single value
### Issue
Discovered from #3037, a user can call `Concatenate `and specify a single string. When this happens, a NullReference exception is thrown. Here is the code sample:
```
        EstimatorChain()
            .Append(mlContext.Transforms.Conversion.ConvertType("Features", "Price", DataKind.Double))
            .Append(mlContext.Transforms.Conversion.ConvertType("Label", "Area", DataKind.Double))
            .Append(mlContext.Transforms.Concatenate("Features"))  // This causes the error, should be ("Features", "Features")
            .AppendCacheCheckpoint(mlContext)
            .Append(mlContext.Regression.Trainers.Sdca("Label", "Features"))
            , mlContext

```

Here is the callstack:

```
>	Microsoft.ML.Core.dll!Microsoft.ML.SchemaShape.Column.GetTypeString() Line 111	C#
 	Microsoft.ML.Data.dll!Microsoft.ML.Trainers.TrainerEstimatorBase<Microsoft.ML.Data.RegressionPredictionTransformer<Microsoft.ML.Trainers.LinearRegressionModelParameters>, Microsoft.ML.Trainers.LinearRegressionModelParameters>.CheckInputSchema(Microsoft.ML.SchemaShape inputSchema) Line 111	C#
 	Microsoft.ML.Data.dll!Microsoft.ML.Trainers.TrainerEstimatorBase<Microsoft.ML.Data.RegressionPredictionTransformer<Microsoft.ML.Trainers.LinearRegressionModelParameters>, Microsoft.ML.Trainers.LinearRegressionModelParameters>.GetOutputSchema(Microsoft.ML.SchemaShape inputSchema) Line 83	C#
 	Microsoft.ML.Data.dll!Microsoft.ML.Data.EstimatorChain<Microsoft.ML.Data.RegressionPredictionTransformer<Microsoft.ML.Trainers.LinearRegressionModelParameters>>.GetOutputSchema(Microsoft.ML.SchemaShape inputSchema) Line 83	C#
 	Microsoft.ML.Data.dll!Microsoft.ML.Data.EstimatorChain<Microsoft.ML.Data.RegressionPredictionTransformer<Microsoft.ML.Trainers.LinearRegressionModelParameters>>.Fit(Microsoft.ML.IDataView input) Line 60	C#
 	ConsoleApp32.dll!Program.main(string[] argv) Line 33	F#
```
The problem is that a NullReference exception looks like a bug and its not obvious to the user on what is the cause of the problem.

## Expected
We should instead notify the user that: 
1) A bad argument was passed in
2) That its the Concatenate transform that has the bad argument

### Solution A
We simply check the length of the name array that is passed to Concatenate and throw the correct exception.

### Solution B
Another possible solution is to change the behavior so that when one column is specified for `Concatenate`, the name is treated as the source and destination -- so this:
```
            .Append(mlContext.Transforms.Concatenate("Features"))
```
would be the same as this:
```
            .Append(mlContext.Transforms.Concatenate("Features", "Features"))
```

cc @glebuk for additional feedback

abgoswam:
Fixes #3061 

The PR makes the following changes:
- Follows Solution A mentioned in the issue
- Added a unit test

### DIFF
--- a/test/Microsoft.ML.Tests/Transformers/ConcatTests.cs
+++ b/test/Microsoft.ML.Tests/Transformers/ConcatTests.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.IO;
 using Microsoft.ML.Data;
 using Microsoft.ML.Data.IO;
@@ -15,6 +16,25 @@ namespace Microsoft.ML.Tests.Transformers
     {
         public ConcatTests(ITestOutputHelper output) : base(output)
         {
+        }
+
+        [Fact]
+        void TestConcatNoInputColumns()
+        {
+            var thrown = false;
+
+            try
+            {
+                var pipe = ML.Transforms.Concatenate("Features");
+            }
+            catch(Exception ex)
+            {
+                Assert.Contains("Input columns not specified", ex.Message);
+                thrown = true;
+
+            }
+            Assert.True(thrown);
+            Done();
         }
 
         [Fact]
--- a/src/Microsoft.ML.Data/Transforms/ColumnConcatenatingEstimator.cs
+++ b/src/Microsoft.ML.Data/Transforms/ColumnConcatenatingEstimator.cs
@@ -54,6 +54,7 @@ namespace Microsoft.ML.Transforms
 
             _host.CheckNonEmpty(outputColumnName, nameof(outputColumnName));
             _host.CheckValue(inputColumnNames, nameof(inputColumnNames));
+            _host.CheckParam(inputColumnNames.Length > 0, nameof(inputColumnNames), "Input columns not specified");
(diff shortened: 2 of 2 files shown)


test/Microsoft.ML.Tests/Transformers/ConcatTests.cs
```diff
@@ -17,6 +18,25 @@ public ConcatTests(ITestOutputHelper output) : base(output)
...
         }
 
+        [Fact]
+        void TestConcatNoInputColumns()
+        {
+            var thrown = false;
+
+            try
+            {
+                var pipe = ML.Transforms.Concatenate("Features");
```
Ivanidzo4ka: >Concatenate [](start = 41, length = 11)

maybe it make sense to remove `params` from inputColumnNames in Concatenate method?

Issue #2957, opened 2019-03-14
FeaturizeText outputTokens uses a magical string to name a new column


rogancarr:
When using `OutputTokens=true`, `FeaturizeText` creates a new column called `${OutputColumnName}_TransformedText`. This isn't really well documented anywhere, and it's odd behavior. I suggest that we make the tokenized text column name explicit in the API.

My suggestion would be the following:
- Change `OutputTokens = [bool]` to `OutputTokensColumn = [string]`, and a `string.NullOrWhitespace(OutputTokensColumn)` signifies that this column will not be created.

What do you all think?

rogancarr:
@justinormont @TomFinley @sfilipi @Ivanidzo4ka  Your thoughts?

Ivanidzo4ka:
I would vote to hide that option in general and switch our examples to chain of estimators like we do here:
https://github.com/dotnet/machinelearning/blob/9f87099801ea5dcda16cd8638cd7b34bff26037f/docs/samples/Microsoft.ML.Samples/Dynamic/WordEmbeddingTransform.cs#L28

rogancarr:
@Ivanidzo4ka  I agree with what you're saying, but `FeaturizeText` is a "grab-bag" transform that does a bunch of stuff in one go, so unless we drop it from the APIs, we should fix this directly.

justinormont:
Exploding the text transform in to it's subparts will make the WordEmbedding quite error prone to use. For instance, users will often forget to lowercase their text, causing a silent degradation as it will still semi-work. I strongly prefer the explorability and simplicity of using a single transform instead of the seven or so sub-components.

On the original topic of creating an explicit `tokens` column: 
Does the API allow for multiple columns to be handled independently? Eg: `Title` => `NgramsTitle` and `Description` => `NgramsTitle`? If so we also produce `NgramsTitle_TransformedText` & `NgramsTitle_TransformedText`; reducing to just `token` means we can only handle one input column.

Ivanidzo4ka:
So we have this chunky transform which do lot of stuff, and have plenty of knobs. One of this knobs is `OutputTokens`. It's involves half of other knobs, and independent from extractors and normalization part.

I see main purpose of `FeaturizeText` transform to, you know, take text and produce something we can consume in trainers, which is vector of floats. And I like then transform or function in general does one thing. 

`OutputTokens` for me feels like some dangling peace of functionality which we added because we could add it, but unnecessary because we should add it.

[...]

PR #3812, opened 2019-06-03
Tree-based featurization

### LINKED ISSUE #2482: TreeEnsembleFeaturizer is not a Transformer yet
[TreeEnsembleFeaturizer](
https://github.com/dotnet/machinelearning/blob/6b1a0d337f74274509cfd439974dca2305a6bfbb/src/Microsoft.ML.FastTree/TreeEnsembleFeaturizer.cs#L44) is a very useful tool in many applications. Do we have a plan to make it publicly available?

cc @shauheen, @Ivanidzo4ka, @TomFinley, @yaeldekel.

wschin:
Fix #2482. Generating features using tree structure has been a popular technique in data mining. This PR exposes this internal-only feature to the public.

Since I don't have enough time to handle multiple different assignments at the same time, please don't put nit comments and create new issues instead. Thanks a lot.

### DIFF
--- a/docs/api-reference/io-columns-tree-featurization-binary-classification.md
+++ b/docs/api-reference/io-columns-tree-featurization-binary-classification.md
@@ -1,0 +1,13 @@
+### Input and Output Columns
+The input column must be a known-sized vector of <xref:System.Single>.
+
+This estimator outputs the following columns:
+
+| Output Column Name | Column Type | Description|
+| -- | -- | -- |
+| `Trees` | Vector of <xref:System.Single> | The output values of all trees. |
+| `Leaves` | Vector of <xref:System.Single> | The IDs of all leaves where the input feature vector falls into. |
+| `Paths` | Vector of <xref:System.Single> | The paths the input feature vector passed through to reach the leaves. |
+
+Those output columns are all optional and user can change their names.
+Please set the names of skipped columns to null so that they would not be produced.
--- a/docs/api-reference/io-columns-tree-featurization-binary-classification.md
+++ b/docs/api-reference/io-columns-tree-featurization-binary-classification.md
@@ -1,5 +1,6 @@
 ### Input and Output Columns
-The input column must be a known-sized vector of <xref:System.Single>.
+The input label column data must be <xref:System.Boolean>.
+The input features column data must be a known-sized vector of <xref:System.Single>.
 
 This estimator outputs the following columns:
 
--- a/docs/api-reference/io-columns-tree-featurization-binary-classification.md
+++ b/docs/api-reference/io-columns-tree-featurization-binary-classification.md
@@ -6,9 +6,9 @@ This estimator outputs the following columns:
 
 | Output Column Name | Column Type | Description|
 | -- | -- | -- |
-| `Trees` | Vector of <xref:System.Single> | The output values of all trees. |
-| `Leaves` | Vector of <xref:System.Single> | The IDs of all leaves where the input feature vector falls into. |
-| `Paths` | Vector of <xref:System.Single> | The paths the input feature vector passed through to reach the leaves. |
+| `Trees` | Known-sized vector of <xref:System.Single> | The output values of all trees. Its size is identical to the total number of trees in the tree ensemble model. |
+| `Leaves` | Known-sized vector of <xref:System.Single> | 0-1 vector representation to the IDs of all leaves where the input feature vector falls into. Its size is the number of total leaves in the tree ensemble model. |
+| `Paths` | Known-sized vector of <xref:System.Single> | 0-1 vector representation to the paths the input feature vector passed through to reach the leaves. Its size is the number of non-leaf nodes in the tree ensemble model. |
 
 Those output columns are all optional and user can change their names.
 Please set the names of skipped columns to null so that they would not be produced.
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/BinaryClassificationFeaturization.ttinclude
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/BinaryClassificationFeaturization.ttinclude
@@ -19,7 +19,7 @@ namespace Samples.Dynamic.Transforms.TreeFeaturization
             var mlContext = new MLContext(seed: 0);
 
             // Create a list of data points to be transformed.
-            var dataPoints = GenerateRandomDataPoints(1000).ToList();
+            var dataPoints = GenerateRandomDataPoints(100).ToList();
 
             // Convert the list of data points to an IDataView object, which is consumable by ML.NET API.
             var dataView = mlContext.Data.LoadFromEnumerable(dataPoints);
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/BinaryClassificationFeaturization.ttinclude
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/BinaryClassificationFeaturization.ttinclude
@@ -1,0 +1,110 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.ML;
+using Microsoft.ML.Data;
+<# if (TrainerOptions != null) { #>
+<#=OptionsInclude#>
+<# } #>
+
+namespace Samples.Dynamic.Transforms.TreeFeaturization
+{
+    public static class <#=ClassName#>
+    {<#=Comments#>
+        public static void Example()
+        {
+            // Create a new context for ML.NET operations. It can be used for exception tracking and logging, 
+            // as a catalog of available operations and as the source of randomness.
+            // Setting the seed to a fixed number in this example to make outputs deterministic.
+            var mlContext = new MLContext(seed: 0);
+
+            // Create a list of data points to be transformed.
+            var dataPoints = GenerateRandomDataPoints(1000).ToList();
+
+            // Convert the list of data points to an IDataView object, which is consumable by ML.NET API.
+            var dataView = mlContext.Data.LoadFromEnumerable(dataPoints);
+<# if (CacheData) { #>
+
+            // ML.NET doesn't cache data set by default. Therefore, if one reads a data set from a file and accesses it many times,
+            // it can be slow due to expensive featurization and disk operations. When the considered data can fit into memory,
+            // a solution is to cache the data in memory. Caching is especially helpful when working with iterative algorithms 
+            // which needs many data passes.
+            dataView = mlContext.Data.Cache(dataView);
+<# } #>
+
+            // Define input and output columns of tree-based featurizer.
+            string labelColumnName = nameof(DataPoint.Label);
+            string featureColumnName = nameof(DataPoint.Features);
+            string treesColumnName = nameof(TransformedDataPoint.Trees);
+            string leavesColumnName = nameof(TransformedDataPoint.Leaves);
+            string pathsColumnName = nameof(TransformedDataPoint.Paths);
+
+			// Define the configuration of the trainer used to train a tree-based model.
+            var trainerOptions = new <#=TrainerOptions#>;
+
+			// Define the tree-based featurizer's configuration.
+            var options = new <#=Options#>;
+
+            // Define the featurizer.
+            var pipeline = mlContext.Transforms.<#=Trainer#>(options);
+
+            // Train the model.
+            var model = pipeline.Fit(dataView);
+
+			// Apply the trained transformer to the considered data set.
+			var transformed = model.Transform(dataView);
+
+			// Convert IDataView object to a list. Each element in the resulted list corresponds to a row in the IDataView.
+            var transformedDataPoints = mlContext.Data.CreateEnumerable<TransformedDataPoint>(transformed, false).ToList();
+
+            // Print out the transformation of the first 3 data points.
+            for (int i = 0; i < 3; ++i)
+            {
+                var dataPoint = dataPoints[i];
+                var transformedDataPoint = transformedDataPoints[i];
+                Console.WriteLine($"The original feature vector [{String.Join(",", dataPoint.Features)}] is transformed to three different tree-based feature vectors:");
+                Console.WriteLine($"  Trees' output values: [{String.Join(",", transformedDataPoint.Trees)}].");
+                Console.WriteLine($"  Leave IDs' 0-1 representation: [{String.Join(",", transformedDataPoint.Leaves)}].");
+                Console.WriteLine($"  Paths IDs' 0-1 representation: [{String.Join(",", transformedDataPoint.Paths)}].");
+            }
+
+            <#=ExpectedOutput#>
+        }
+
+        private static IEnumerable<DataPoint> GenerateRandomDataPoints(int count, int seed=0)
+        {
+            var random = new Random(seed);
+            float randomFloat() => (float)random.NextDouble();
+            for (int i = 0; i < count; i++)
+            {
+                var label = randomFloat() > <#=LabelThreshold#>;
+                yield return new DataPoint
+                {
+                    Label = label,
+                    // Create random features that are correlated with the label.
+                    // For data points with false label, the feature values are slightly increased by adding a constant.
+                    Features = Enumerable.Repeat(label, 3).Select(x => x ? randomFloat() : randomFloat() + <#=DataSepValue#>).ToArray()
+                };
+            }
+        }
+
+        // Example with label and 3 feature values. A data set is a collection of such examples.
+        private class DataPoint
+        {
+            public bool Label { get; set; }
+            [VectorType(3)]
+            public float[] Features { get; set; }
+        }
+
+        // Class used to capture the output of tree-base featurization.
+        private class TransformedDataPoint : DataPoint
+        {
+		    // The i-th value is the output value of the i-th decision tree.
+            public float[] Trees { get; set; }
+		    // The 0-1 encoding of leaves the input feature vector falls into.
+            public float[] Leaves { get; set; }
+		    // The 0-1 encoding of paths the input feature vector reaches the leaves.
+            public float[] Paths { get; set; }
+        }
+    }
+}
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/BinaryClassificationFeaturization.ttinclude
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/BinaryClassificationFeaturization.ttinclude
@@ -39,10 +39,10 @@ namespace Samples.Dynamic.Transforms.TreeFeaturization
             string leavesColumnName = nameof(TransformedDataPoint.Leaves);
             string pathsColumnName = nameof(TransformedDataPoint.Paths);
 
-			// Define the configuration of the trainer used to train a tree-based model.
+            // Define the configuration of the trainer used to train a tree-based model.
             var trainerOptions = new <#=TrainerOptions#>;
 
-			// Define the tree-based featurizer's configuration.
+            // Define the tree-based featurizer's configuration.
             var options = new <#=Options#>;
 
             // Define the featurizer.
@@ -51,10 +51,10 @@ namespace Samples.Dynamic.Transforms.TreeFeaturization
             // Train the model.
             var model = pipeline.Fit(dataView);
 
-			// Apply the trained transformer to the considered data set.
-			var transformed = model.Transform(dataView);
+            // Apply the trained transformer to the considered data set.
+            var transformed = model.Transform(dataView);
 
-			// Convert IDataView object to a list. Each element in the resulted list corresponds to a row in the IDataView.
+            // Convert IDataView object to a list. Each element in the resulted list corresponds to a row in the IDataView.
             var transformedDataPoints = mlContext.Data.CreateEnumerable<TransformedDataPoint>(transformed, false).ToList();
 
             // Print out the transformation of the first 3 data points.
@@ -99,11 +99,11 @@ namespace Samples.Dynamic.Transforms.TreeFeaturization
         // Class used to capture the output of tree-base featurization.
         private class TransformedDataPoint : DataPoint
         {
-		    // The i-th value is the output value of the i-th decision tree.
+            // The i-th value is the output value of the i-th decision tree.
             public float[] Trees { get; set; }
-		    // The 0-1 encoding of leaves the input feature vector falls into.
+            // The 0-1 encoding of leaves the input feature vector falls into.
             public float[] Leaves { get; set; }
-		    // The 0-1 encoding of paths the input feature vector reaches the leaves.
+            // The 0-1 encoding of paths the input feature vector reaches the leaves.
             public float[] Paths { get; set; }
         }
     }
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/FastForestBinaryFeaturizationWithOptions.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/FastForestBinaryFeaturizationWithOptions.cs
@@ -87,17 +87,17 @@ namespace Samples.Dynamic.Transforms.TreeFeaturization
 
             // Expected output:
             //  The original feature vector [0.8173254,0.7680227,0.5581612] is transformed to three different tree-based feature vectors:
-            //    Trees' output values: [-0.3333333,0.005309734].
-            //    Leave IDs' 0-1 representation: [0,0,0,0,0,1,0,0,0,1,0,0].
-            //    Paths IDs' 0-1 representation: [1,1,1,1,1,1,1,1,1,1].
+            //    Trees' output values: [0.1111111,0.8823529].
+            //    Leave IDs' 0-1 representation: [0,0,0,0,1,0,0,0,0,1,0].
+            //    Paths IDs' 0-1 representation: [1,1,1,1,1,1,0,1,0].
             //  The original feature vector [0.5888848,0.9360271,0.4721779] is transformed to three different tree-based feature vectors:
-            //    Trees' output values: [-0.02077151,0.005309734].
-            //    Leave IDs' 0-1 representation: [0,1,0,0,0,0,0,0,0,1,0,0].
-            //    Paths IDs' 0-1 representation: [1,1,1,1,1,1,1,1,1,1].
+            //    Trees' output values: [0.4545455,0.8].
+            //    Leave IDs' 0-1 representation: [0,0,0,1,0,0,0,0,0,0,1].
+            //    Paths IDs' 0-1 representation: [1,1,1,1,0,1,0,1,1].
             //  The original feature vector [0.2737045,0.2919063,0.4673147] is transformed to three different tree-based feature vectors:
-            //    Trees' output values: [-0.02077151,0.005309734].
-            //    Leave IDs' 0-1 representation: [0,1,0,0,0,0,0,0,0,1,0,0].
-            //    Paths IDs' 0-1 representation: [1,1,1,1,1,1,1,1,1,1].
+            //    Trees' output values: [0.4545455,0.1111111].
+            //    Leave IDs' 0-1 representation: [0,0,0,1,0,0,1,0,0,0,0].
+            //    Paths IDs' 0-1 representation: [1,1,1,1,0,1,0,1,1].
         }
 
         private static IEnumerable<DataPoint> GenerateRandomDataPoints(int count, int seed=0)
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/FastForestBinaryFeaturizationWithOptions.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/FastForestBinaryFeaturizationWithOptions.cs
@@ -19,7 +19,7 @@ namespace Samples.Dynamic.Transforms.TreeFeaturization
             var mlContext = new MLContext(seed: 0);
 
             // Create a list of data points to be transformed.
-            var dataPoints = GenerateRandomDataPoints(1000).ToList();
+            var dataPoints = GenerateRandomDataPoints(100).ToList();
 
             // Convert the list of data points to an IDataView object, which is consumable by ML.NET API.
             var dataView = mlContext.Data.LoadFromEnumerable(dataPoints);
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/FastForestBinaryFeaturizationWithOptions.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/FastForestBinaryFeaturizationWithOptions.cs
@@ -1,0 +1,139 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.ML;
+using Microsoft.ML.Data;
+using Microsoft.ML.Trainers.FastTree;
+
+namespace Samples.Dynamic.Transforms.TreeFeaturization
+{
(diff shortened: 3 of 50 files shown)


src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
```diff
@@ -437,5 +436,45 @@ public static class TreeExtensions
             var env = CatalogUtils.GetEnvironment(catalog);
             return new FastForestBinaryTrainer(env, options);
         }
+
+        public static PretrainedTreeFeaturizationEstimator PretrainTreeEnsembleFeaturizing(this TransformsCatalog catalog,
```
eerhardt: XML Doc on all public classes and APIs.

src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
```diff
@@ -437,5 +436,45 @@ public static class TreeExtensions
...
+            PretrainedTreeFeaturizationEstimator.Options options)
+        {
+            Contracts.CheckValue(catalog, nameof(catalog));
+            var env = CatalogUtils.GetEnvironment(catalog);
+            return new PretrainedTreeFeaturizationEstimator(env, options);
+        }
+
+        public static FastForestRegressionFeaturizationEstimator FastForestRegressionFeaturizing(this TransformsCatalog catalog,
```
eerhardt: This naming pattern reads a little funny. How about turning it into `FeaturizeXXX`? Like we have with `FeaturizeText`.

wschin: I can do `FeaturizeBy...`.

src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
wschin: No problem. I am working on them.

src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
eerhardt: That sounds better (to me at least).

src/Microsoft.ML.FastTree/TreeEnsembleFeaturizationEstimator.cs
```diff
@@ -0,0 +1,279 @@
...
+            /// The column type must be a vector of <see cref="System.Single"/>.
+            /// </summary>
+            public string InputColumnName;
+
+            /// <summary>
+            /// The estimator has three output columns. Their names would be "Trees" + <see cref="OutputColumnsSuffix"/>,
+            /// "Leaves" + <see cref="OutputColumnsSuffix"/>, and "Paths" + <see cref="OutputColumnsSuffix"/>. If <see cref="OutputColumnsSuffix"/>
+            /// is <see langword="null"/>, the output names would be "Trees", "Leaves", and "Paths".
+            /// </summary>
+            public string OutputColumnsSuffix;
```
justinormont: We went away from magic strings in the TextTransform. Previously with `tokens=+`, we produced a new column named `{OutputColName}_TokenizedText`.

For the estimators API we have users directly enter the column name for the tokens.  We may want to do the same for the Trees/Leaves/Paths of the TreeFeat.

Perhaps:
`OutputColumnTreeName`, `OutputColumnLeavesName`, `OutputColumnPathsName`.

Ivanidzo4ka: And don't add that column if they empty or equal to null.
That way you can actually configure which parts of tree structure do you want.

justinormont: Conversation about TextTransform: 
via @rogancarr in https://github.com/dotnet/machinelearning/issues/2957 PR
>  When using `OutputTokens=true`, `FeaturizeText` creates a new column called `${OutputColumnName}_TransformedText`. This isn't really well documented anywhere, and it's odd behavior. I suggest that we make the tokenized text column name explicit in the API.
> 
> My suggestion would be the following:
> - Change `OutputTokens = [bool]` to `OutputTokensColumn = [string]`, and a `string.NullOrWhitespace(OutputTokensColumn)` signifies that this column will not be created.

wschin: Now we can do optional output columns and custom output column names. Please see tests for examples (or wait for formal API samples).

src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
```diff
@@ -437,5 +436,103 @@ public static class TreeExtensions
...
+            return new FastTreeRankingFeaturizationEstimator(env, options);
+        }
+
+        /// <summary>
+        /// Create <see cref="FastTreeTweedieFeaturizationEstimator"/>, which uses <see cref="FastTreeTweedieTrainer"/> to train <see cref="TreeEnsembleModelParameters"/> to create tree-based features.
+        /// </summary>
+        /// <param name="catalog">The context <see cref="TransformsCatalog"/> to create <see cref="FastTreeTweedieFeaturizationEstimator"/>.</param>
+        /// <param name="options">The options to configure <see cref="FastTreeTweedieFeaturizationEstimator"/>. See <see cref="FastTreeTweedieFeaturizationEstimator.Options"/> and
+        /// <see cref="TreeEnsembleFeaturizationEstimatorBase.CommonOptions"/> for available settings.</param>
+        public static FastTreeTweedieFeaturizationEstimator FeaturizeByFastTreeTweedie(this TransformsCatalog catalog,
```
justinormont: May want to note in its name that FastTreeTweedie is regression: _(the naming of the others list their task types)_
```suggestion
        public static FastTreeTweedieRegressionFeaturizationEstimator FeaturizeByFastTreeTweedieRegression(this TransformsCatalog catalog,
```

test/Microsoft.ML.Tests/TrainerEstimators/TreeEnsembleFeaturizerTest.cs
```diff
@@ -108,5 +115,633 @@ public void TreeEnsembleFeaturizerOutputSchemaTest()
...
+            var options = new FastTreeBinaryFeaturizationEstimator.Options()
+            {
+                InputColumnName = "Features",
+                TreesColumnName = "Trees",
+                LeavesColumnName = "Leaves",
+                PathsColumnName = "Paths",
+                TrainerOptions = trainerOptions
+            };
+
+            var pipeline = ML.Transforms.FeaturizeByFastTreeBinary(options).
```
justinormont: Creating (8) seperate `ML.Transforms.FeaturizeBy{ FastTreeBinary, FastForestRegression, FastTreeRegression, FastTreeTweedie, ... }` featurizers under `ML.Transforms.*` seem a bit large. Seems to clutter up the namespace. 

In the future, this should grow as I think we should have LightGBM variants too (and Catboost if we take it in). The number of independent featurizers will be: 
{ FastTree, FastTreeTweedie, FastForest, LightGBM, CatBoost } x { BinaryClassification, Multiclass, Regression, Ranking }`. _(with some combinations missing)_

Would it be more clean to have one `ML.Transforms.TreeFeaturizer()`, and put the specific instance type as a parameter? Would it be doable to have one return type?

test/Microsoft.ML.Tests/TrainerEstimators/TreeEnsembleFeaturizerTest.cs
```diff
@@ -108,5 +115,633 @@ public void TreeEnsembleFeaturizerOutputSchemaTest()
...
+            var options = new FastForestRegressionFeaturizationEstimator.Options()
+            {
+                InputColumnName = "Features",
+                TreesColumnName = "Trees",
+                LeavesColumnName = "Leaves",
+                PathsColumnName = "Paths",
+                TrainerOptions = trainerOptions
+            };
+
+            var pipeline = ML.Transforms.FeaturizeByFastForestRegression(options).
```
justinormont: Can you add a test of `ML.Transforms.FeaturizeByFastForestRegression()` using FastForest's [`ShuffleLabels`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.trainers.fasttree.fastforestregressiontrainer.options?view=ml-dotnet) option? I believe this is the only way to use TreeFeat w/ multi-class classification.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEnsembleFeaturizerTest.cs
wschin: There are two reasons that I don't like a single `TreeFeaturizer`.

1.  It goes an opposite direction of the C# API's design. Most of them are strongly typed to the underlying data structures. You can see we have a lot of `FastTree...` and `LightGbm...`.
2. It may requires user to specify `TreeFeaturizer<TTrainer>(options)` and the user manually need to make sure the type of `options` is `TTrainer.Options`, which is easy to make mistakes.

src/Microsoft.ML.FastTree/TreeTrainersCatalog.cs
wschin: Yes if the model name doesn't tell the task. Given that `Tweedie` somehow implies a regression case, we don't have `Regression` appended to any of public `Tweedie` modules. This pattern can be seen in `FastTreeTweedieTrainer` and `FastTreeTweedieModelParameters`.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEnsembleFeaturizerTest.cs
justinormont: You're right, the current pattern will likely have less mistakes by the user. 

For the plethora of `FastTree...` and `LightGBM...`, those are namespaced under the task `mlContext.Regression.Trainers.FastTree()`, hence the duplication isn't visible.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEnsembleFeaturizerTest.cs
wschin: Let's make it in another PR. This PR has been too large..

justinormont: Seems the UnitTests should be in the main PR.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEnsembleFeaturizerTest.cs
```diff
@@ -108,5 +115,672 @@ public void TreeEnsembleFeaturizerOutputSchemaTest()
...
+            Assert.True(metrics.Accuracy > 0.98);
+            Assert.True(metrics.LogLoss < 0.05);
+            Assert.True(metrics.AreaUnderPrecisionRecallCurve > 0.98);
+        }
+
+        [Fact]
+        public void TreeEnsembleFeaturizingPipelineMulticlass()
+        {
+            int dataPointCount = 200;
+            var data = SamplesUtils.DatasetUtils.GenerateFloatLabelFloatFeatureVectorSamples(dataPointCount).ToList();
```
justinormont: For a multi-class UnitTest, I think we'd want to create a multi-class dataset:
```suggestion
            var data = SamplesUtils.DatasetUtils.GenerateRandomMulticlassClassificationExamples(dataPointCount).ToList();
```

Otherwise, it's creating a regression dataset.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEnsembleFeaturizerTest.cs
```diff
@@ -108,5 +115,672 @@ public void TreeEnsembleFeaturizerOutputSchemaTest()
...
+                TreesColumnName = "Trees",
+                LeavesColumnName = "Leaves",
+                PathsColumnName = "Paths",
+                TrainerOptions = trainerOptions
+            };
+
+            var pipeline = ML.Transforms.Conversion.Hash("HashedLabel", "Label", numberOfBits: 3).
+                Append(ML.Transforms.FeaturizeByFastForestRegression(options)).
+                Append(ML.Transforms.Concatenate("CombinedFeatures", "Features", "Trees", "Leaves", "Paths")).
+                Append(ML.MulticlassClassification.Trainers.SdcaMaximumEntropy("HashedLabel", "CombinedFeatures"));
```
justinormont: The SDCA will be able to take in the original Labels from the multi-class dataset.
```suggestion
                Append(ML.MulticlassClassification.Trainers.SdcaMaximumEntropy("Label", "CombinedFeatures"));
```

docs/api-reference/io-columns-tree-featurization-binary-classification.md
```diff
@@ -0,0 +1,14 @@
+### Input and Output Columns
+The input label column data must be <xref:System.Boolean>.
+The input features column data must be a known-sized vector of <xref:System.Single>.
+
+This estimator outputs the following columns:
+
+| Output Column Name | Column Type | Description|
+| -- | -- | -- |
+| `Trees` | Vector of <xref:System.Single> | The output values of all trees. |
```
Ivanidzo4ka: >The output values of all trees [](start = 45, length = 30)

One value per tree I would assume?

Ivanidzo4ka: >Vector [](start = 12, length = 6)

Known size?

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/BinaryClassificationFeaturization.ttinclude
```diff
@@ -0,0 +1,110 @@
...
+<# } #>
+
+            // Define input and output columns of tree-based featurizer.
+            string labelColumnName = nameof(DataPoint.Label);
+            string featureColumnName = nameof(DataPoint.Features);
+            string treesColumnName = nameof(TransformedDataPoint.Trees);
+            string leavesColumnName = nameof(TransformedDataPoint.Leaves);
+            string pathsColumnName = nameof(TransformedDataPoint.Paths);
+
+			// Define the configuration of the trainer used to train a tree-based model.
```
Ivanidzo4ka: >			 [](start = 0, length = 3)

nit: you had some tab heresy in your `tt` files.

docs/api-reference/io-columns-tree-featurization-binary-classification.md
```diff
@@ -0,0 +1,14 @@
...
+| `Leaves` | Vector of <xref:System.Single> | The IDs of all leaves where the input feature vector falls into. |
+| `Paths` | Vector of <xref:System.Single> | The paths the input feature vector passed through to reach the leaves. |
```
Ivanidzo4ka: From what I see in your samples it's more like a bitmap than IDS and paths.

src/Microsoft.ML.FastTree/TreeEnsembleFeaturizationEstimator.cs
```diff
@@ -0,0 +1,464 @@
...
+        private protected readonly IHostEnvironment Env;
+
+        private protected TreeEnsembleFeaturizationEstimatorBase(IHostEnvironment env, CommonOptions options)
+        {
+            Env = env;
+            if (options.InputColumnName == null)
+                throw Env.Except(nameof(options), "The " + nameof(options.InputColumnName) + " cannot be null.");
+            if (options.TreesColumnName == null && options.LeavesColumnName == null && options.PathsColumnName == null)
+                throw Env.Except($"{nameof(CommonOptions.TreesColumnName)}, {nameof(CommonOptions.LeavesColumnName)}, and {nameof(CommonOptions.PathsColumnName)} cannot be all null at the same time. " +
+                    $"At least one output column name should be provided so that at least one output column may be generated.");
```
Ivanidzo4ka: >that [](start = 76, length = 4)

This `that` looks out of place. Could be wrong

docs/api-reference/io-columns-tree-featurization-binary-classification.md
wschin: Yes, details are included in `tree-featurization-prediction.md`
```md
### Prediction Details
This estimator produces several output columns from a tree ensemble model. Assume that the model contains only one decision tree:

                   Node 0
                   /    \
                 /        \
               /            \
             /                \
           Node 1            Node 2
           /    \            /    \
         /        \        /        \
       /            \     Leaf -3  Node 3
      Leaf -1      Leaf -2         /    \
                                 /        \
                                Leaf -4  Leaf -5

Assume that the input feature vector falls into `Leaf -1`. The output `Trees` may be a 1-element vector where
the only value is the decision value carried by `Leaf -1`. The output `Leaves` is a 0-1 vector. If the reached
leaf is the $i$-th (indexed by $-(i+1)$ so the first leaf is `Leaf -1`) leaf in the tree, the $i$-th value in `Leaves`
would be 1 and all other values would be 0. The output `Paths` is a 0-1 representation of the nodes passed
through before reaching the leaf. The $i$-th element in `Paths` indicates if the $i$-th node (indexed by $i$) is touched.
For example, reaching `Leaf -1` lead to $[1, 1, 0, 0]$ as the `Paths`. If there are multiple trees, this estimator
just concatenates `Trees`'s, `Leaves`'s, `Paths`'s from all trees (first tree's information comes first in the concatenated vectors).

Check the See Also section for links to usage examples.
```

src/Microsoft.ML.FastTree/TreeEnsembleFeaturizationEstimator.cs
```diff
@@ -0,0 +1,464 @@
...
+    /// produces three columns: (1) the prediction values of all trees, (2) the IDs of leaves the input feature vector falling into, and (3)
+    /// the binary vector which encodes the paths to those destination leaves.
+    /// </summary>
+    public abstract class TreeEnsembleFeaturizationEstimatorBase : IEstimator<TreeEnsembleFeaturizationTransformer>
+    {
+        /// <summary>
+        /// The common options of tree-based featurizations such as <see cref="FastTreeBinaryFeaturizationEstimator"/>, <see cref="FastForestBinaryFeaturizationEstimator"/>,
+        /// <see cref="FastTreeRegressionFeaturizationEstimator"/>, <see cref="FastForestRegressionFeaturizationEstimator"/>, and <see cref="PretrainedTreeFeaturizationEstimator"/>.
+        /// </summary>
+        public class CommonOptions
```
eerhardt: Why is this named `CommonOptions`? I think we normally name a base class `OptionsBase`.

For example: [https://github.com/dotnet/machinelearning/blob/a1b5eaabedda87554328682c4dfd3aa311463002/src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs#L157](https://github.com/dotnet/machinelearning/blob/a1b5eaabedda87554328682c4dfd3aa311463002/src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs#L157)

eerhardt: Should this class be `abstract`? Or do we expect external users to new one of these up directly?

src/Microsoft.ML.FastTree/TreeEnsembleFeaturizer.cs
```diff
@@ -390,9 +427,15 @@ public TreeEnsembleFeaturizerBindableMapper(IHostEnvironment env, ModelLoadConte
 
             // *** Binary format ***
             // ensemble
+            // string: treesColumnName
+            // string: leavesColumnName
+            // string: pathsColumnName
 
             ctx.LoadModel<TreeEnsembleModelParameters, SignatureLoadModel>(env, out _ensemble, "Ensemble");
             _totalLeafCount = CountLeaves(_ensemble);
+            _treesColumnName = ctx.LoadStringOrNull();
```
eerhardt: Do we need to bump the version number?

wschin: It's BC, so maybe no.

eerhardt: What does "BC" mean?

docs/api-reference/io-columns-tree-featurization-binary-classification.md
wschin: Fixed. Thanks.

src/Microsoft.ML.FastTree/TreeEnsembleFeaturizer.cs
wschin: This module is backward compatible. All previously saved models work as what they were doing.

wschin: Ah.. No, you're right. Let me bump the version.

src/Microsoft.ML.FastTree/TreeEnsembleFeaturizationEstimator.cs
wschin: Yes. Let's make it `abstract`. Thank you.

---
In reply to: [291369721](https://github.com/dotnet/machinelearning/pull/3812#discussion_r291369721) [](ancestors = 291369721)

wschin: No problem. I am renaming it.

---
In reply to: [291369495](https://github.com/dotnet/machinelearning/pull/3812#discussion_r291369495) [](ancestors = 291369495)

src/Microsoft.ML.FastTree/TreeEnsembleFeaturizationEstimator.cs
wschin: Removed.

---
In reply to: [291367319](https://github.com/dotnet/machinelearning/pull/3812#discussion_r291367319) [](ancestors = 291367319)

docs/api-reference/io-columns-tree-featurization-binary-classification.md
wschin: Yes. I will extend this string a bit.

---
In reply to: [291364613](https://github.com/dotnet/machinelearning/pull/3812#discussion_r291364613) [](ancestors = 291364613)

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/BinaryClassificationFeaturization.ttinclude
wschin: Removed. Thanks.

---
In reply to: [291365453](https://github.com/dotnet/machinelearning/pull/3812#discussion_r291365453) [](ancestors = 291365453)

src/Microsoft.ML.FastTree/TreeEnsembleFeaturizer.cs
```diff
@@ -390,9 +428,15 @@ public TreeEnsembleFeaturizerBindableMapper(IHostEnvironment env, ModelLoadConte
...
```
eerhardt: Do you need to check if the model has the correct version before calling `LoadStringOrNull`?

wschin: Sure. Added. Thanks.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEnsembleFeaturizerTest.cs
```diff
@@ -108,5 +115,700 @@ public void TreeEnsembleFeaturizerOutputSchemaTest()
...
+            try
+            {
+                var wrongPipeline = ML.Transforms.FeaturizeByFastTreeBinary(options).
+                    Append(ML.Transforms.Concatenate("CombinedFeatures", "Features", "Trees", "Leaves", "Paths")).
+                    Append(ML.BinaryClassification.Trainers.SdcaLogisticRegression("Label", "CombinedFeatures"));
+                var wrongModel = wrongPipeline.Fit(dataView);
+            }
+            catch
+            {
+                isWrong = true; // Only "Leaves" is produced by tree featurizer, so accessing "Trees" and "Paths" may lead to an error.
```
justinormont:
```suggestion
                isWrong = true; // Only "Leaves" is produced by the tree featurizer, so accessing "Trees" and "Paths" will lead to an error.
```

test/Microsoft.ML.Tests/TrainerEstimators/TreeEnsembleFeaturizerTest.cs
```diff
@@ -108,5 +115,700 @@ public void TreeEnsembleFeaturizerOutputSchemaTest()
...
+            var options = new PretrainedTreeFeaturizationEstimator.Options()
+            {
+                InputColumnName = "Features",
+                TreesColumnName = "Trees",
+                LeavesColumnName = "Leaves",
+                PathsColumnName = "Paths",
+                ModelParameters = treeModel.Model.SubModel
+            };
+            var pipeline = ML.Transforms.FeaturizeByPretrainTreeEnsemble(options).
+                Append(ML.Transforms.Concatenate("CombinedFeatures", "Features", "Trees", "Leaves", "Paths")).
```
justinormont: Isn't the normal style to have a period at the start?
```suggestion
                .Append(ML.Transforms.Concatenate("CombinedFeatures", "Features", "Trees", "Leaves", "Paths"))
```

wschin: You're right. Let me fix them. Thanks.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEnsembleFeaturizerTest.cs
```diff
@@ -108,5 +115,700 @@ public void TreeEnsembleFeaturizerOutputSchemaTest()
...
+            var prediction = model.Transform(dataView);
+            var metrics = ML.BinaryClassification.Evaluate(prediction);
+
+            Assert.True(metrics.Accuracy > 0.98);
+            Assert.True(metrics.LogLoss < 0.05);
+            Assert.True(metrics.AreaUnderPrecisionRecallCurve > 0.98);
+        }
+
+        [Fact]
+        public void TreeEnsembleFeaturizingPipelineMulticlass()
```
justinormont: @daholste and I were able to map the `Key` to a `float` using a `CustomMapping()`.

I'd recommend the multiclass unit test be:
```suggestion
        [Fact]
        public void TreeEnsembleFeaturizingPipelineMulticlass()
        {
            int dataPointCount = 1000;
            var data = SamplesUtils.DatasetUtils.GenerateRandomMulticlassClassificationExamples(dataPointCount).ToList();
            var dataView = ML.Data.LoadFromEnumerable(data);
            dataView = ML.Data.Cache(dataView);
 
            var trainerOptions = new FastForestRegressionTrainer.Options
            {
                NumberOfThreads = 1,
                NumberOfTrees = 10,
                NumberOfLeaves = 4,
                MinimumExampleCountPerLeaf = 10,
                FeatureColumnName = "Features",
                LabelColumnName = "FloatLabel",
                ShuffleLabels = true
            };
           
            var options = new FastForestRegressionFeaturizationEstimator.Options()
            {
                InputColumnName = "Features",
                TreesColumnName = "Trees",
                LeavesColumnName = "Leaves",
                PathsColumnName = "Paths",
                TrainerOptions = trainerOptions
            };
 
            Action<RowWithKey, RowWithFloat> actionConvertKeyToFloat = (RowWithKey rowWithKey, RowWithFloat rowWithFloat) =>
            {
                rowWithFloat.FloatLabel = rowWithKey.KeyLabel == 0 ? float.NaN : rowWithKey.KeyLabel - 1;
            };
 
            var split = ML.Data.TrainTestSplit(dataView, 0.5);
            var trainData = split.TrainSet;
            var testData = split.TestSet;
 
            var pipeline = ML.Transforms.Conversion.MapValueToKey("KeyLabel", "Label")
                .Append(ML.Transforms.CustomMapping(actionConvertKeyToFloat, "KeyLabel"))
                .Append(ML.Transforms.FeaturizeByFastForestRegression(options))
                .Append(ML.Transforms.Concatenate("CombinedFeatures", "Trees", "Leaves", "Paths"))
                .Append(ML.MulticlassClassification.Trainers.SdcaMaximumEntropy("KeyLabel", "CombinedFeatures"));
 
            var model = pipeline.Fit(trainData);
            var prediction = model.Transform(testData);
            var metrics = ML.MulticlassClassification.Evaluate(prediction, labelColumnName: "KeyLabel");
 
            Assert.True(metrics.MacroAccuracy > 0.6);
            Assert.True(metrics.MicroAccuracy > 0.6);
       }
       
        class RowWithKey
        {
            [KeyType()]
            public uint KeyLabel { get; set; }
        }
       
        class RowWithFloat
        {
            public float FloatLabel { get; set; }
        }
```

Specifically, this is using a `CustomMapping()` to convert the `Key` to a `float` for use in the TreeFeat's FastForest regression. The current method in the unit test requires a user to know/list all of the values in their Label (and their type). The `CustomMapping()` style is easier for a user to replicate for their dataset.

We also added a `TrainTestSplit()` so we're not testing on the training set, and we removed the original features from the `Concatenate()` to ensure the TreeFeat's output features are useful.

src/Microsoft.ML.FastTree/TreeEnsembleFeaturizationEstimator.cs
```diff
@@ -0,0 +1,464 @@
...
+        /// </summary>
+        public sealed class Options : OptionsBase
+        {
+            /// <summary>
+            /// The pretrained tree model used to do tree-based featurization. Note that <see cref="TreeEnsembleModelParameters"/> contains a collection of decision trees.
+            /// </summary>
+            public TreeEnsembleModelParameters ModelParameters;
+        };
+
+        private TreeEnsembleModelParameters _modelParameters;
```
artidoro: >TreeEnsembleModelParameters [](start = 16, length = 27)

Should this be readonly or something similar to make sure it is not altered?

artidoro:
Since you have already built all this infrastructure why are we not providing the featurizers for `LightGbm` trainers?

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TreeFeaturization/FastForestBinaryFeaturizationWithOptions.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        // Class used to capture the output of tree-base featurization.
+        private class TransformedDataPoint : DataPoint
+        {
+            // The i-th value is the output value of the i-th decision tree.
+            public float[] Trees { get; set; }
+            // The 0-1 encoding of leaves the input feature vector falls into.
+            public float[] Leaves { get; set; }
+            // The 0-1 encoding of paths the input feature vector reaches the leaves.
+            public float[] Paths { get; set; }
```
artidoro: Maybe:
// The 0-1 encoding of paths the input feature vector follows to reach the leaves.

src/Microsoft.ML.FastTree/TreeEnsembleFeaturizationEstimator.cs
```diff
@@ -0,0 +1,464 @@
...
+            /// necessary to be the same as the feature column used to train the underlying tree model.
+            /// </summary>
+            public string InputColumnName;
+
+            /// <summary>
+            /// The name of the column that stores the prediction values of all trees. Its type is a vector of <see cref="System.Single"/>
+            /// and the i-th vector element is the prediction value predicted by the i-th tree.
+            /// If <see cref="TreesColumnName"/> is <see langword="null"/>, this output column may not be generated.
+            /// </summary>
+            public string TreesColumnName;
```
artidoro: >TreesColumnName [](start = 26, length = 15)

Suggested renaming:

TreesColumnName -> TreeOutputsColumnName

I think it would be easier to understand, but this is not necessary.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEnsembleFeaturizerTest.cs
justinormont: Thanks for adding.
#Resolved

PR #3819, opened 2019-06-04
NativeAssemblyReference Include "MklProxyNative" in Samples project.

### LINKED ISSUE #3818: NativeAssemblyReference Include "MklProxyNative" in Samples project.
Need to the above dependency to have some timer series samples work that use MKL

codemzs:
fixes #3818

### DIFF
--- a/src/Microsoft.ML.TimeSeries/Microsoft.ML.TimeSeries.csproj
+++ b/src/Microsoft.ML.TimeSeries/Microsoft.ML.TimeSeries.csproj
@@ -6,6 +6,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.ML.Mkl.Redist" Version="1.0.0" />
+  </ItemGroup>
+
+  <ItemGroup>
     <ProjectReference Include="..\Microsoft.ML.Core\Microsoft.ML.Core.csproj" />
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.TimeSeries/Microsoft.ML.TimeSeries.csproj
```diff
@@ -5,6 +5,10 @@
     <IncludeInPackage>Microsoft.ML.TimeSeries</IncludeInPackage>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ML.Mkl.Redist" Version="1.0.0" />
```
eerhardt: Isn't this needed in `pkg\Microsoft.ML.TimeSeries\Microsoft.ML.TimeSeries.nupkgproj` instead?

eerhardt: Instead of `PackageReference`, shouldn't it be a `ProjectReference`?

shmoradims: how was timeseries working so far without adding this dependency?

codemzs: It works with the nuget, not in samples repo for some time series samples that require mkl. I have fixed it now.,

Issue #3800, opened 2019-05-31
Issue training


woanware:
I have tried creating a simple data and performing the training like so

```
dotnet .\mlnet.dll auto-train --task binary-classification
 --dataset "logons.csv" --label-column-index 0 
--has-header true --max-exploration-time 10
```

Here is an example of the data set which is reduced from my original, but shows the format:

```
Valid	 Data
0	 09:00
0	 09:01
0	 09:02
0	 09:03
0	 09:04
0	 09:05
0	 09:06
0	 09:07
1	 12:08
0	 09:09
0	 09:10
0	 09:00
0	 09:01
0	 09:02
0	 09:03
0	 09:04
0	 09:05
0	 09:06
0	 09:07
1	 13:08
0	 09:09
0	 09:10
0	 09:00
0	 09:01
0	 09:02
0	 09:03
0	 09:04
0	 09:05
0	 09:06
0	 09:07
1	 14:08
0	 09:09
0	 09:10
```

Every time I try and run the command I get the following error:

```
Exception occured while exploring pipelines:
Training failed with the exception: 
System.ArgumentOutOfRangeException: AUC is not definied 
when there is no positive class in the data
Parameter name: PosSample
```

I originally tried it via VS2019 and the latest version of ML.Net, but that failed, so I tried it using the binary directly

woanware:
I have now altered my test data to have a 30+% split of positive results, and the training works. Thanks!

justinormont:
@woanware: You may want to set a weight column too, which will preserve the original true/false ratio. 

Upsampling your positive class (or downsampling your negative class) changes the ratio of true/false that your trainer sees. This will cause the model to predict `true` more often than your original dataset. If that is unwanted, you can use a weight column to down-weight your positive class.

Also, if you're upsampling, ensure you split your dataset first, then upsample. Otherwise duplicate rows will be seen again in the test set causing your metrics to be no longer representative, which is a form of data leakage.

PR #3820, opened 2019-06-04
For binary classification, discard cross validation splits where there is not at least one true & one false label in the validation set

daholste:
For binary classification, discard cross validation splits where there is not at least one true & one false label in the validation set. Otherwise, scoring the dataset crashes because AUC cannot be computed like in https://github.com/dotnet/machinelearning/issues/3800

### DIFF
--- a/src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
+++ b/src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
@@ -29,7 +29,7 @@ namespace Microsoft.ML.AutoML
             for (var i = 0; i < trainDatasets.Count(); i++)
             {
                 var validationDataset = validationDatasets.ElementAt(i);
-                var labelColumn = validationDataset.Schema.First(c => c.Name == labelColumnName);
+                var labelColumn = validationDataset.Schema.First(c => !c.IsHidden && c.Name == labelColumnName);
                 if (DatasetDimensionsUtil.ComputeCardinality<bool>(validationDataset, labelColumn, 2) < 2)
                 {
                     continue;
--- a/src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
+++ b/src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
@@ -10,7 +10,48 @@ namespace Microsoft.ML.AutoML
 {
     internal static class SplitUtil
     {
-        public static (IDataView[] trainDatasets, IDataView[] validationDatasets) CrossValSplit(MLContext context, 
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
```diff
@@ -10,7 +10,48 @@ namespace Microsoft.ML.AutoML
...
+            // If we're running binary classification, discard splits where there is not at least one 
+            // true & one false label in the test set. Otherwise, scoring the dataset crashes because
+            // AUC cannot be computed.
+            var filteredTrainDatasets = new List<IDataView>();
+            var filteredValidationDatasets = new List<IDataView>();
+            for (var i = 0; i < trainDatasets.Count(); i++)
+            {
+                var validationDataset = validationDatasets.ElementAt(i);
+                var labelColumn = validationDataset.Schema.First(c => !c.IsHidden && c.Name == labelColumnName);
+                if (DatasetDimensionsUtil.ComputeCardinality<bool>(validationDataset, labelColumn, 2) < 2)
```
justinormont: This is an expensive solution for the uncommon issue; and all binary classification pipeline will entail the cost (normally) unnecessarily.  The specific issue being handled is binary classification learner need both classes in all folds. 

Recommend instead letting the cross-val fold fail, and catching the error; then removing the CV result. We need to check that some CV folds were successful.

Post checking is better as only the bad pipelines/splits entail the cost. The datasets most likely to hit the issue are small and fast. Otherwise, the large datasets (which are least likely to hit the issue) entail the most overhead.

daholste: We default to cross-val mode only for small datasets. Additionally, the cardinality counting stops as soon as one value of each label is encountered. Combining these two facts, most of the time, I would imagine this operation would be fast. Although +1 -- worst-case run time (user manually specifies a high # of splits for a dataset that is sorted by label) would be 1 pass over each of the folds. Although, since each iteration requires several featurization & training passes over each fold, even in this edge case that a user likely wouldn't encounter, run time may not be too abysmal.

+1 that we should make cross val robust to failures on individual folds. However, we may in addition want to preserve these kinds of up-front checks. If we can inexpensively determine that a fold is invalid up front, why, in general, featurize & train that fold, and and have it fail every iteration? Also, as we're passing iteration exceptions back to the user, we may not want to pass back the semi-cryptic AUC exception (https://github.com/dotnet/machinelearning/issues/3800) back as a failure that occurred each iteration. In general, we may also not want to clog up our iteration failed exception telemetry unnecessarily

PR #3827, opened 2019-06-05
 Add Microsoft.Extensions.ML integration package.

### LINKED ISSUE #3239: Make it easier to use ML.NET in an ASP.NET app/service
## Problem
With `1.0.0-preview` bits, it is currently harder than it needs to be to use ML.NET inside an ASP.NET service or application. The first problem users hit is whether they can cache a `PredictionEngine` statically and reuse it for multiple requests. As described in #1789, you cannot use a PredictionEngine on multiple threads at the same time. Doing so will cause problems in your application.

Thus the recommendation is to use a pooling technique, but writing one from scratch is rather hard and potentially error prone.

Also, by default the MLContext's Log operation is not aware of any logging infrastructure currently used by ASP.NET apps/services. Thus the log goes nowhere, and is lost.

## Proposal
We propose to add a new library (`Microsoft.ML.Extensions?`, `Microsoft.Extensions.ML?`) that is aware of both `Microsoft.ML` and `Microsoft.Extensions.DependencyInjection`/`Microsoft.Extensions.Logging` and glues the two together. This should make it much easier to consume ML.NET models inside ASP.NET apps/services, as well as any other app model that integrates with the `Microsoft.Extensions.*` libraries.

### Usage
Adding a new ML.NET model into an ASP.NET application could be as simple as two steps:

1. Add a PredictionEnginePool in your Startup.cs:

```C#
    // This method gets called by the runtime. Use this method to add services to the container.
    public void ConfigureServices(IServiceCollection services)
    {
        services
            .AddPredictionEnginePool<SentimentIssue, SentimentPrediction>("SentimentModel.zip");

        // other service configuration
    }
```

2. In any controller that needs to make a prediction, inject the PredictionEngine pool in the constructor, and use it where necessary:

```C#
[ApiController]
public class PredictionController : ControllerBase
{
    private PredictionEnginePool<SentimentIssue, SentimentPrediction> _predictionEnginePool;

    public PredictionController(PredictionEnginePool<SentimentIssue, SentimentPrediction> predictionEnginePool)
    {
        _predictionEnginePool = predictionEnginePool;
    }

    [HttpGet()]
    public ActionResult<SentimentPrediction> GetSentiment([FromQuery]SentimentIssue input)
    {
        return _predictionEnginePool.Predict(input);
    }
}
```

### Other potential scenarios
1. Being able to add a model `.zip` file from sources other than a file path
    1. Azure Blob Storage
    2. A SQL Database
    3. Any URL
2. Being able to automatically reload an updated model, if the file/URL changes (using FileWatcher or ETag or some other mechanism).
3. Being able to have different "named" models for scenarios like A/B testing where you want 90% of users to get Model A and 10% to get Model B.

@glennc @CESARDELATORRE @glebuk @TomFinley

eerhardt:
This package makes it easier to use ML.NET with app models that support Microsoft.Extensions - i.e. ASP.NET and Azure Functions.

Specifically it contains functionality for:

- Dependency Injection
- Pooling PredictionEngines
- Reloading models when the file or URI has changed
- Hooking ML.NET logging to Microsoft.Extensions.Logging

Fix #3239

cc @glennc - Note that the only major change I made to your code was to remove the `IPredictionEnginePoolBuilder` interface in favor of using a simple class instead.

After this is merged to master, we can remove the `features/IntegrationPackage` branch.


src/Microsoft.Extensions.ML/Builder/BuilderExtensions.cs
```diff
@@ -0,0 +1,213 @@
...
+    public static class BuilderExtensions
+    {
+        /// <summary>
+        /// Adds the model at the specified location to the builder.
+        /// </summary>
+        /// <param name="builder">The builder to which to add the model.</param>
+        /// <param name="uri">The location of the model.</param>
+        /// <returns>
+        /// The updated <see cref="PredictionEnginePoolBuilder{TData, TPrediction}"/>.
+        /// </returns>
```
wschin: Is this a public C# API? Would it need a sample file? The same question to the subsequent functions.

src/Microsoft.Extensions.ML/Builder/PredictionEnginePoolBuilder.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <param name="services">The <see cref="IServiceCollection"/> to add services to.</param>
+        public PredictionEnginePoolBuilder(IServiceCollection services)
+        {
+            Services = services ?? throw new ArgumentException(nameof(services));
+        }
+
+        /// <summary>
+        /// The <see cref="IServiceCollection"/> to add services to.
+        /// </summary>
+        public IServiceCollection Services { get; private set; }
```
wschin: I am not sure what `IServiceCollection` is. Would each `Service` be used to create a `PredictionEnginePool`? Am the a targeted audience of this function (I don't have background of Azure function and ASP.NET)? Feel free to ignore this comment if I am not. :)

src/Microsoft.Extensions.ML/MLContextOptions.cs
```diff
@@ -0,0 +1,60 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.ML;
+
+namespace Microsoft.Extensions.ML
+{
+    /// <summary>
+    /// Provides an options class for MLContext objects.
+    /// </summary>
+    public class MLContextOptions
```
wschin: This looks like a wrapper of `MLContext`. Couldn't we directly use `MLContext`?

src/Microsoft.Extensions.ML/ModelReloadToken.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// </summary>
+        public bool HasChanged => _cts.IsCancellationRequested;
+
+        /// <summary>
+        /// Registers for a callback that will be invoked when the entry has changed. <see cref="Microsoft.Extensions.Primitives.IChangeToken.HasChanged"/>
+        /// MUST be set before the callback is invoked.
+        /// </summary>
+        /// <param name="callback">The callback to invoke.</param>
+        /// <param name="state">State to be passed into the callback.</param>
+        /// <returns></returns>
```
wschin: > <returns></returns> [](start = 11, length = 20)

Empty returned value?

src/Microsoft.Extensions.ML/PoolLoader.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Primitives;
+using Microsoft.ML;
+
+namespace Microsoft.Extensions.ML
+{
+    internal class PoolLoader<TData, TPrediction>: IDisposable
+                        where TData : class
+                        where TPrediction : class, new()
+    {
```
wschin: Although this is an internal class, I'd still like to have some mild documentation. My intention is to treat devs like end users so that ML.NET can be more inclusive.

src/Microsoft.Extensions.ML/PredictionEnginePoolPolicy.cs
```diff
@@ -0,0 +1,53 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.ObjectPool;
+using Microsoft.ML;
+
+namespace Microsoft.Extensions.ML
+{
+    /// <summary>
+    /// <see cref="IPooledObjectPolicy{T}"/> for <see cref="PredictionEngine{TData, TPrediction}"/>.
```
wschin: >IPooledObjectPolicy [](start = 19, length = 19)

I am not sure what it is given an over-simplified official document ([https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.objectpool.ipooledobjectpolicy-1?view=aspnetcore-2.2](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.objectpool.ipooledobjectpolicy-1?view=aspnetcore-2.2)). Could you please elaborate?

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// </param>
+        /// <returns>
+        /// The <see cref="PredictionEnginePoolBuilder{TData, TPrediction}"/> that was added to the collection.
+        /// </returns>
+        public static PredictionEnginePoolBuilder<TData, TPrediction> AddPredictionEnginePool<TData, TPrediction>(
+            this IServiceCollection services)
+            where TData : class
+            where TPrediction : class, new()
+        {
+            services.AddPredictionEngineServices<TData, TPrediction>();
```
wschin: >AddPredictionEngineServices [](start = 21, length = 27)

Is this a function of `IServiceCollection`? this page ([https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.iservicecollection?view=aspnetcore-2.2](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.iservicecollection?view=aspnetcore-2.2)) doesn't mention `AddPredictionEngineServices`.

src/Microsoft.Extensions.ML/Builder/BuilderExtensions.cs
eerhardt: The APIs in this project are part of the public API. We have samples using them here:

https://github.com/dotnet/machinelearning-samples/tree/1131a9eb49f45cd0ff96e444ec584c78e77dc6c6/samples/csharp/end-to-end-apps/ScalableMLModelOnWebAPI-IntegrationPkg

These APIs are different from the "core" APIs in that they don't transform data or learn from data. But instead they are intended for bridging ML.NET and ASP.NET (and similar) technologies. So the same "sample" format for transforms and trainers won't work. We need a full-blown sample for using it end-to-end, like we have in machinelearning-sample.

src/Microsoft.Extensions.ML/Builder/PredictionEnginePoolBuilder.cs
eerhardt: > Am the a targeted audience of this function (I don't have background of Azure function and ASP.NET)?

Probably not. But when you start wanting to do more web service type of work, you will be :).

Check out https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection if you want to learn more about how the Dependency Injection system works in ASP.NET Core.

src/Microsoft.Extensions.ML/MLContextOptions.cs
eerhardt: That's a good question. @glennc - would this be possible?

src/Microsoft.Extensions.ML/ModelReloadToken.cs
eerhardt: Fixed.

src/Microsoft.Extensions.ML/PoolLoader.cs
eerhardt: Added a summary doc to this class.

src/Microsoft.Extensions.ML/PredictionEnginePoolPolicy.cs
eerhardt: I don't believe this class needs to be public. I have marked it as `internal`, since I don't think it will need to be used externally. @glennc - please raise any objections if I've misunderstood this class.

I have beef'd up the class summary doc here.

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
eerhardt: This is an extension method to `IServiceCollection`. It is declared directly below this method.

src/Microsoft.Extensions.ML/ModelLoaders/UriModelLoader.cs
```diff
@@ -0,0 +1,231 @@
...
+                    {
+                        _eTag = resp.Headers.GetValues(ETagHeader).First();
+                        return true;
+                    }
+                    return false;
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Error loading {uri}", _uri);
```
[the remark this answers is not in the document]
eerhardt: Done.

src/Microsoft.Extensions.ML/MLContextOptions.cs
glennc: Could you hook the logger a different way if you didn't have the PostOptionsConfigure step? I.e how will you register the func that can log with ILogger while you are adding services to the service collection?

wschin: I am not sure I am following. `MLContextOptions` is a container of `MLContext` and I don't see any new utility functions. Can we replace 
```
    public class PostMLContextOptionsConfiguration : IPostConfigureOptions<MLContextOptions>
```
with
```
    public class PostMLContextOptionsConfiguration : IPostConfigureOptions<MLContext>
```
?

test/Microsoft.Extensions.ML.Tests/FileLoaderTests.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            var services = new ServiceCollection()
+                .AddOptions()
+                .AddLogging();
+            var sp = services.BuildServiceProvider();
+
+            var loaderUnderTest = ActivatorUtilities.CreateInstance<FileModelLoader>(sp);
+            loaderUnderTest.Start(Path.Combine("TestModels", "SentimentModel.zip"), false);
+
+            var model = loaderUnderTest.GetModel();
```
wschin: Do you think it's reasonable to make a prediction using the loaded model?

eerhardt: Yes. Added.

src/Microsoft.Extensions.ML/MLContextOptions.cs
eerhardt: I looked into it, and you can't just use `MLContext` directly with `IOptions<T>` because:

```C#
public interface IOptions<out TOptions> where TOptions : class, new()
```
IOptions requires the class to have an empty constructor, which MLContext doesn't.

Also, if we did `IOptions<MLContext>`, we couldn't add new options in the future.

So instead, to solve this, I did the following:

1. Renamed `MLContextOptions` to just `MLOptions`. 
2. Changed the one option currently - the `MLContext` you want to use - to be lazy-loaded, which allows users to supply their own `MLContext`, if they want. So users can do something like:

```C#
            services.Configure<MLOptions>(options =>
            {
                options.MLContext = new MLContext(15);
            });
```

Which allows the user to specify the `seed` if they require it. Or they can configure the `ComponentCatalog` if they want.

I believe this alleviates the concern here. We have an `Options` class, which we can add more options to in the future. And one of those options is which `MLContext` to use.

PR #3831, opened 2019-06-05
Add default and missing value definitions for ML.NET types

### LINKED ISSUE #3443: Default and missing values for each type should be documented
In some of our documentation, like in #3424 we reference the "non-default" values or the "default" values of the parameters. Those values need to be documented together with the APIs using them.

shmoradims:
Fixes #3443.

This PR has two related parts:

1) Add default and missing values definitions. DataKind page, seems the best place to have all of that information in one place. The documented definitions come from [IsNa region](https://github.com/dotnet/machinelearning/blob/2960b273ee0eda1d3a285a4b46a58a4d6d7b6926/src/Microsoft.ML.Data/Data/Conversion.cs#L739) and [IsDefault region](https://github.com/dotnet/machinelearning/blob/2960b273ee0eda1d3a285a4b46a58a4d6d7b6926/src/Microsoft.ML.Data/Data/Conversion.cs#L749).

2) Fix the docs of CountFeatureSelectingEstimator to include missing value in addition to default value. Previously, we incorrectly thought only non-default values matter. But digging into [the code](https://github.com/dotnet/machinelearning/blob/5163413f3cfab430e49f6fb2a690c160fde46b58/src/Microsoft.ML.Transforms/CountFeatureSelection.cs#L433), it's clear that the check is for non-default AND non-missing.


src/Microsoft.ML.Core/Data/DataKind.cs
```diff
@@ -18,21 +42,24 @@ public enum DataKind : byte
...
-        /// <summary>8-byte usigned integer, type of <see cref="System.UInt64"/>.</summary>
+        /// <summary>8-byte unsigned integer, type of <see cref="System.UInt64"/>.</summary>
         UInt64 = 8,
         /// <summary>4-byte floating-point number, type of <see cref="System.Single"/>.</summary>
         Single = 9,
         /// <summary>8-byte floating-point number, type of <see cref="System.Double"/>.</summary>
         Double = 10,
-        /// <summary>string, type of <see cref="System.String"/>.</summary>
+        /// <summary>
+        /// string, type of <see cref="System.ReadOnlyMemory{T}"/>, where T is <see cref="char"/>.
```
eerhardt:
```suggestion
        /// string, type of <see cref="System.ReadOnlyMemory{char}"/>.
```

eerhardt: I see this doesn't actually work. I rescind my suggestion.

src/Microsoft.ML.Core/Data/DataKind.cs
```diff
@@ -9,6 +9,30 @@ namespace Microsoft.ML.Data
...
+    /// | <xref:Microsoft.ML.Data.DataKind.String> or [text](xref:Microsoft.ML.Data.TextDataViewType) | Empty or `null` string (both result in empty `System.ReadOnlyMemory<char>` | <xref:"System.ReadOnlyMemory`1.IsEmpty*> |
+    /// | [Key](xref:Microsoft.ML.Data.KeyDataViewType) type (supported by the unsigned integer types in `DataKind`) | Not defined | Always `false` |
+    /// | All other types | Default value of the corresponding system type as defined by .NET standard. In C#, default value expression `default(T)` provides that value. | Equality test with the default value |
+    ///
+    /// The table below shows the missing value definition for each of the data types.
+    ///
+    /// | Type | Missing Value | IsMissing Indicator |
+    /// | -- | -- | -- |
+    /// | <xref:Microsoft.ML.Data.DataKind.String> or [text](xref:Microsoft.ML.Data.TextDataViewType) | Not defined | Always `false` |
+    /// | [Key](xref:Microsoft.ML.Data.KeyDataViewType) type (supported by the unsigned integer types in `DataKind`) | `0` | Equality test with `0` |
```
wschin: Could you add a in-memory example to explain `Key`'s missing value? `0` is probably a key value only exists in ML.NET and user can't really see it by using `CreateEnumerable`.

src/Microsoft.ML.Transforms/CountFeatureSelection.cs
```diff
@@ -31,22 +31,23 @@ namespace Microsoft.ML.Transforms
...
     ///
-    /// This transform uses a set of aggregators to count the number of non-default values for each slot(vector element) and
-    /// instantiates a transformer to actually drop the slots that have the default values for the types
-    /// keeping at least 'count' non-default values.
+    /// This transform uses a set of aggregators to count the number of values for each slot (vector element)
+    /// that are non-default and non-missing (for the definitions of default and missing, refer to the remarks section
+    /// in <xref:Microsoft.ML.Data.DataKind>). If the count value is less than the provided count parameter, that slot is dropped.
     /// This transform is useful when applied together with a <xref:Microsoft.ML.Transforms.OneHotHashEncodingTransformer>.
-    /// The count feature selection can remove the features generated by the hash transform that have no data in the examples.
+    /// It can remove the features generated by the hash transform that have no data in the examples.
```
wschin: Could you add a reference to hash transfomer's API?

shmoradims: there's an xref link to OneHotHashEncodingTransformer in the line above. Is that what you mean?

---
In reply to: [290927868](https://github.com/dotnet/machinelearning/pull/3831#discussion_r290927868) [](ancestors = 290927868)

src/Microsoft.ML.Core/Data/DataKind.cs
shmoradims: I think this would be P2. I added it to backlog: [https://github.com/dotnet/machinelearning/issues/3832](https://github.com/dotnet/machinelearning/issues/3832)

---
In reply to: [290926731](https://github.com/dotnet/machinelearning/pull/3831#discussion_r290926731) [](ancestors = 290926731)

src/Microsoft.ML.Transforms/CountFeatureSelection.cs
wschin: I meant
```c#
            var pipeline = mlContext.Transforms.Categorical.OneHotHashEncoding(
                "EducationOneHotHashEncoded", "Education", numberOfBits: 3);
```

PR #3835, opened 2019-06-06
Explain MLContext seed parameter

### LINKED ISSUE #3048: MLContext class reference-doc should explain the seed parameter better and further introduction about the catalogs, etc.
This is the reference doc for MLContext constructor:

https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.mlcontext.-ctor?view=ml-dotnet#Microsoft_ML_MLContext__ctor_System_Nullable_System_Int32__System_Int32_ 

Currently, it just says the following about the 'seed' parameter:

_"Random seed. Set to null for a non-deterministic environment."_

The info provided here is too simple. This should be the place to explain in further details what the MLContext class is for.

Especially, the `seed `parameter should be properly explained here. Why do we have such a parameter in the constructor and when to use it (deterministic results, etc.) with further explanations and example cases.

Also, explain if the `seed ` parameter is applicable or impacts only when using MLContext for training (creating a model) or if `seed `parameter matters or not if you are just loading a model from a .ZIP file and scoring with it etc.

Finally, if it is recommended in most cases to have a seed, why don't we put any value as a by default value for the `seed ` parameter? For getting started users, the seed parameter feels really strange..

shmoradims:
Fixes #3048.

### DIFF
--- a/src/Microsoft.ML.Data/MLContext.cs
+++ b/src/Microsoft.ML.Data/MLContext.cs
@@ -90,7 +90,7 @@ namespace Microsoft.ML
         /// MLContext's random number generator is the global source of randomness for
         /// all of such random operations.
         ///
-        /// If a seed is provided by <paramref name="seed"/>, MLContext environment becomes
+        /// If a fixed seed is provided by <paramref name="seed"/>, MLContext environment becomes
         /// deterministic, meaning that the results are repeatable and will remain the same across multiple runs.
         /// For instance in many of ML.NET's API reference example code snippets, a seed is provided.
         /// That's because we want the users to get the same output as what's included in example comments,
--- a/src/Microsoft.ML.Data/MLContext.cs
+++ b/src/Microsoft.ML.Data/MLContext.cs
@@ -105,7 +105,7 @@ namespace Microsoft.ML
         /// The behavior of those operations are deterministic regardless of the seed value.
         ///
         /// Also ML.NET trainers don't use randomness *after* the training is finished.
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.Data/MLContext.cs
```diff
@@ -82,7 +82,31 @@ public sealed class MLContext : IHostEnvironment
...
-        /// <param name="seed">Random seed. Set to <c>null</c> for a non-deterministic environment.</param>
+        /// <param name="seed">Seed for MLContext's random number generator. See the remarks for more details.</param>
+        /// <remarks>
+        /// Many operations in ML.NET require randomness, such as
+        /// random data shuffling, random sampling, random parameter initialization,
+        /// random permutation, random feature selection, and many more.
+        /// MLContext's random number generator is the global source of randomness for
+        /// all of such random operations.
+        ///
+        /// If a seed is provided by <paramref name="seed"/>, MLContext environment becomes
```
wschin:
```suggestion
        /// If a fixed seed is provided by <paramref name="seed"/>, MLContext environment becomes
```

CESARDELATORRE: This is great! 👍 This is an explanation I asked for in this ISSUE: https://github.com/dotnet/machinelearning/issues/3048 

Adding @natke so specific documents at docs.microsoft.com will point to this explanation when first introducing the MLContext where seed is used. 

Thank you! 👍

PR #3840, opened 2019-06-07
Broken docs images for ranking metrics

justinormont:
Fixes two broken links to images at: https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.rankingmetrics?view=ml-dotnet-1.1.0.

The next doc release will show the fix.

Issues fixed by this PR:
* Linking to GitHub's webpage for the images, and not the images itself
* Linking to the master branch, which will change overtime, potentially breaking older versions of the docs

Issues NOT fixed by this PR:
* Linking at all to GitHub -- Do we have a way to have the docs make its own copy of the image? Currently the user's browser will pull the image from GitHub while on docs.microsoft.com. Perhaps this will be fixed when this PR fixes the link?
* NDCG.png has a transparent background; DCG.png one does not 
* We may want a check-in test which verifies document image links

Feel free to push to this PR if you can fix any of the additional above items.


src/Microsoft.ML.Data/Evaluators/Metrics/RankingMetrics.cs
```diff
@@ -15,7 +15,7 @@ public sealed class RankingMetrics
     {
         /// <summary>
         /// Array of normalized discounted cumulative gains where i-th element represent NDCG@i.
-        /// <image src="https://github.com/dotnet/machinelearning/tree/master/docs/images/NDCG.png"></image>
+        /// <image src="https://github.com/dotnet/machinelearning/blob/3cdd3c8b32705e91dcf46c429ee34196163af6da/docs/images/NDCG.png?raw=true"></image>
```
wschin: Is it ideal to point to a fixed commit? What if we change the figure but forget to update this link?

justinormont: I think we want to point to a fixed commit. If we point to /main/, our old docs will break if we move the image's location within the repo. 

We keep old versions of the docs for each release:
* https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.rankingmetrics?view=ml-dotnet-1.1.0
* https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.rankingmetrics?view=ml-dotnet
* https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.rankingmetrics?view=ml-dotnet-preview

The repo is expected to evolve over time. The better solution is having the docs keep track of its own images (which is the 1st item under 'ssues NOT fixed by this PR').

Hopefully the person modifying the images first looks in the code for where they are used. They should then see that they have to update the pinned commit.

codemzs:
@justinormont please sync to master before you merge.

PR #3842, opened 2019-06-07
Revert "Publish nugets to temporary myget feed (#3400)"

shmoradims:
This reverts commit 0a90bbb49da69765c2ff870078fdabfb20783a58.

shmoradi myget feed is downgraded to 200MB of space after a month and needs to be replaced with a permanent solution. It was created as a workaround before //build. Disabling it for now to prevent the failure in publishing step.

### DIFF
--- a/build/vsts-ci.yml
+++ b/build/vsts-ci.yml
@@ -182,7 +182,7 @@ phases:
     _SignType: real
     _UseEsrpSigning: true
     _TeamName: DotNetCore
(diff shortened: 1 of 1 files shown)


wschin:
Thanks @shmoradims a lot! How long does it need to implement the permanent solution? Is someone working on that?

shmoradims:
> Thanks @shmoradims a lot! How long does it need to implement the permanent solution? Is someone working on that?
@wschin: @abgoswam and  @artidoro are working on that. They can provide ETA.

PR #3850, opened 2019-06-10
Bugfix/hardwired sigmoid

### LINKED ISSUE #1422: Hardwired value of Sigmoid parameter used instead of value specified in parameters
https://github.com/dotnet/machinelearning/blob/5123aeedb91a435236e8b94ba0f17a6d15a20db5/src/Microsoft.ML.LightGBM/LightGbmBinaryTrainer.cs#L133

The hardwired value `0.5` in the line above should instead be the value of the `Sigmoid` parameter specified when the model was trained (the default value of which is 1, but used to be 0.5).

michaelgsharp:
Fixes #1422 

Fixes the Hardcoded Sigmoid value from -.5 to the value specified during training.

Let Microsoft.ML.Tests see internals of Microsoft.ML.Standard trainers to access trained model parameters for testing verification.


test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
```diff
@@ -15,6 +15,7 @@
 using Microsoft.ML.Trainers.FastTree;
 using Microsoft.ML.Transforms;
 using Xunit;
+using Microsoft.ML.Calibrators;
```
artidoro: Can you sort the usings?

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
```diff
@@ -64,6 +65,31 @@ public void LightGBMBinaryEstimator()
...
+                NumberOfLeaves = 10,
+                NumberOfThreads = 1,
+                MinimumExampleCountPerLeaf = 2,
+                Sigmoid = sigmoid
+            });
+
+            var transformedDataView = pipe.Fit(dataView).Transform(dataView);
+            var model = trainer.Fit(transformedDataView, transformedDataView);
+
+            //the slope in the model calibrator should be equal to the negative of the sigmoid passed into the trainer
```
artidoro: nit: usually the format for comments is:
// Comment starting after a space with a capital letter and ending in a period.

We try to follow this format as often as possible.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
```diff
@@ -251,6 +277,32 @@ public void LightGbmMulticlassEstimator()
...
+                Sigmoid = sigmoid
+            });
+
+            var pipe = pipeline.Append(trainer)
+                    .Append(new KeyToValueMappingEstimator(Env, "PredictedLabel"));
+
+            var transformedDataView = pipe.Fit(dataView).Transform(dataView);
+            var model = trainer.Fit(transformedDataView, transformedDataView);
+
+            //the slope in the all the calibrators should be equal to the negative of the sigmoid passed into the trainer
```
artidoro: nit: same suggestion regarding format of comment.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
michaelgsharp: Sounds good. I will update it.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
michaelgsharp: Sounds good. I will update it.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
michaelgsharp: Put the `Microsoft.ML.Calibrators` in the correct spot in the list.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
michaelgsharp: Fixed.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
michaelgsharp: Fixed.

wschin:
Thanks a lot for contributing this fix! Would you consider making a direct test between ML.NET and LightGBM using non-default sigmoid? We have a similar test [here](https://github.com/dotnet/machinelearning/blob/8b1b14f916428967dc0066b956485c35a0090fe3/test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs#L377).

michaelgsharp:
Thanks for your feedback @wschin ! I added a positive and negative test case for the case you mentioned.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
```diff
@@ -267,7 +319,7 @@ private class GbmExample
             public float[] Score;
         }
 
-        private void LightGbmHelper(bool useSoftmax, out string modelString, out List<GbmExample> mlnetPredictions, out double[] lgbmRawScores, out double[] lgbmProbabilities)
+        private void LightGbmHelper(bool useSoftmax, out string modelString, out List<GbmExample> mlnetPredictions, out double[] lgbmRawScores, out double[] lgbmProbabilities, double sigmoid = .5)
```
wschin: The parameter `sigmoid` can appear before those `out` variables.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
```diff
@@ -405,6 +458,82 @@ public void LightGbmMulticlassEstimatorCompareOva()
             Done();
         }
 
+        /// <summary>
+        /// LightGbmMulticlass UsingSigmoids test 
+        /// </summary>
+        [LightGBMFact]
+        public void LightGbmMulticlassEstimatorCompareOvaUsingSigmoids()
+        {
+            var sigmoid = .790;
```
wschin:
```suggestion
            var sigmoidScale = .790;
```
to match some inline comments below.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
```diff
@@ -405,6 +458,82 @@ public void LightGbmMulticlassEstimatorCompareOva()
             Done();
         }
 
+        /// <summary>
+        /// LightGbmMulticlass UsingSigmoids test
```
wschin:
```suggestion
        /// Test LightGBM's sigmoid parameter with a custom value. This test checks if ML.NET and LightGBM produce the same result.
```

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
```diff
@@ -405,6 +458,82 @@ public void LightGbmMulticlassEstimatorCompareOva()
...
+                    double prob = MathUtils.SigmoidSlow((float)sigmoid * (float)nativeResult1[j + i * _classNumber]);
+                    Assert.Equal(prob / sum, mlnetPredictions[i].Score[j], 6);
+                }
+            }
+
+            Done();
+        }
+
+        /// <summary>
+        /// LightGbmMulticlass UsingDifferentSigmoids test
```
wschin:
```suggestion
        /// Make sure different sigmoid parameters produce different scores. In this test, two LightGBM models are trained with two different sigmoid values.
```

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
```diff
@@ -300,7 +352,8 @@ private void LightGbmHelper(bool useSoftmax, out string modelString, out List<Gb
                     NumberOfIterations = numberOfTrainingIterations,
                     MinimumExampleCountPerGroup = 1,
                     MinimumExampleCountPerLeaf = 1,
-                    UseSoftmax = useSoftmax
+                    UseSoftmax = useSoftmax,
+                    Sigmoid = sigmoid
```
wschin:
```suggestion
                    Sigmoid = sigmoid // Custom sigmoid value.
```

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
michaelgsharp: Since `sigmoid` is an optional parameter it has to come after all the required parameters, including the out parameters. I could refactor the code to make it not optional and just pass in the default value from the other methods, but that is the only way to have it come before the `out` params in the list.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
michaelgsharp: I renamed it directly in VisualStudio so I could make sure I got all its references as well. Fixed.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
wschin: Explicitly specify it to 0.5 for all default cases can solve your concern maybe? As a test helper, I feel exposing all information explicitly is useful.

michaelgsharp: I changed it so its no longer a default value and is explicitly passed in from each method and moved it before the out params.

PR #3851, opened 2019-06-10
Add example for applying ONNX model to in-memory images

### LINKED ISSUE #3562: Making a prediction from ONNX model in ML.net
### System information

- **OS version: window 10
- Ml.Net verison: 0.11

### Issue
Hi, everyone. I am a newbie of using ML.net to implement my classification problem. I already trained a model from CIFA 10 datasets based on Keras library and that model has been transferred to ONNX model. My goal is to display the image's prediction task on WindowForm, so I am going to use ML.net integrated with ONNX model, but right now I am getting stuck on how to use the ONNX model on ML.net and how to define the label and get the prediction result by the model. Do you have any suggestion to me? Thank you a lot in advance.
### Source code / logs

Please paste or attach the code or logs or traces that would be helpful to diagnose the issue you are reporting.

wschin:
Fix #3562 by adding an sample.

### DIFF
--- a/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ApplyONNXModelWithInMemoryImages.cs
+++ b/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/ApplyONNXModelWithInMemoryImages.cs
@@ -50,8 +50,16 @@ namespace Samples.Dynamic
             foreach (var dataPoint in transformedDataPoints)
             {
                 var firstClassProb = dataPoint.Scores.First();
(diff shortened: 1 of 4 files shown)


CESARDELATORRE:
Here:
https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/end-to-end-apps/ObjectDetection-Onnx
An btw, in addition to using in-memory images in the ASP.NET Core web app, in that same solution we added a WPF app sample which is streaming video while using object detection from the ONNX model. :)

PR #3857, opened 2019-06-12
Early Draft specs doc for DatabaseLoader in ML.NET

CESARDELATORRE:
Early Draft specs doc for a `DatabaseLoader` component in ML.NET. 
Feel free to provide feedback. This specs document will be evolving significantly since it is in a very early draft state.

Main objective is to load data into an IDataView from relational databases (such as SQL Server, Azure SQL Database, PostgreSQL, MySQL, Oracle, etc.) in a very easy way, one line of code in most cases by simply specifying the connection string, the database `table/view/sql-sentence` and what database provider/connection type it is using.

Currently in ML.NET 1.0 or 1.1 we can only do the following:

- Load data from files by using `MLContext.Data.LoadFromTextFile()`, etc.
- Load data from an IEnumerable collection with `MLContext.Data.LoadFromEnumerable()`, usually coming from a database, but the developer/user is responsible for how to load/query the database. It is not straightforward such as when reading from a file.

**High level design goal:**
The intention of the new component is to make a lot easier to implement _'model training scenarios  pulling/streaming data from large database tables'_ while transparently solving complicated cases such as 'transient errors' in the cloud (database connections broken as a result) when using Azure SQL Database.

### DIFF
--- a/docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
+++ b/docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
@@ -162,10 +162,10 @@ The specific type (`SqlConnection`, `NpgsqlConnection`, `OracleConnection`, etc.
 
 **1. (Convenient 'must-have' method) Data loading from a database by specifying a SQL query sentence:**
 
-The signature of the `DataOperationsCatalog.LoadFromDatabaseSqlQuery` method looks like:
+The signature of the `DataOperationsCatalog.LoadFromDbSqlQuery` method looks like:
 
 ```C#
-public IDataView LoadFromDatabaseSqlQuery<TRow, DbConnection>(string connString, string sqlQuerySentence) where T : class;
+public IDataView LoadFromDbSqlQuery<TRow, DbConnection>(string connString, string sqlQuerySentence) where T : class;
 ```
 
 `TRow` is the model's input data class (Observation class) to be used by IDataView and ML.NET API.
@@ -176,18 +176,17 @@ Example code using it:
 MLContext mlContext = new MLContext();
 
 //Example loading from a SQL Server or SQL Azure database with a SQL query sentence
-IDataView trainingDataView = mlContext.Data.LoadFromDatabaseSqlQuery<ModelInputData, SqlConnection>(connString: myConnString,
-                                                                                                    sqlQuerySentence: "Select * from InputMLModelDataset where InputMLModelDataset.CompanyName = 'MSFT'"); 
+IDataView trainingDataView = mlContext.Data.LoadFromDbSqlQuery<ModelInputData, SqlConnection>(connString: myConnString, sqlQuerySentence: "Select * from InputMLModelDataset where InputMLModelDataset.CompanyName = 'MSFT'"); 
 ```
 
 **2. (Foundational method) Data loading from a database with an IDataReader object:**
 
 This is the foundational or pillar method which will be used by the rest of the higher level or convenient methods:
 
-The signature of the `DataOperationsCatalog.LoadFromDataReader` method looks like:
+The signature of the `DataOperationsCatalog.LoadFromDbDataReader` method looks like:
 
 ```C#
-public IDataView LoadFromDataReader<TRow>(Func<IDataReader> executeReader) where TRow : class;
+public IDataView LoadFromDbDataReader<TRow>(Func<IDataReader> executeReader) where TRow : class;
 ```
 
 `TRow` is the model's input data class (Observation class) to be used by IDataView and ML.NET API.
@@ -211,13 +210,13 @@ using (SqlConnection connection = new SqlConnection(conString))
 **IMPORTANT - Lambda with command.ExecuteReader() as parameter instead of a IDataReader as paramater:** It is important to highlight that because we want to enforce that each query will use a different data reader object (in order to avoid multi-threading issues), the `mlContext.Data.LoadFromDbDataReader<TRow>()` method is accepting a lambda expression with the `command.ExecuteReader()` line of code. That means the method only accepts a `Func<IDataReader>` delegate because the lambda expression has no input parameters and returns an IDataReader object (such as `SqlDataReader`).
 
 
-**3. (Nice to have) Data loading from a database table (Simplest approach):**
+**3. ('Nice to have') Data loading from a database table (Simplest approach):**
 
 This is mostly a 'sugar method', but can be a 'nice to have feature' for users.
-The signature of the `DataOperationsCatalog.LoadFromDatabaseTable` method looks like:
+The signature of the `DataOperationsCatalog.LoadFromDbTable` method looks like:
 
 ```C#
-public IDataView LoadFromDatabaseTable<TRow, DbConnection>(string connString, string tableName)  where T : class;
+public IDataView LoadFromDbTable<TRow, DbConnection>(string connString, string tableName)  where T : class;
 ```
 
 Example code using it:
@@ -226,17 +225,17 @@ Example code using it:
 MLContext mlContext = new MLContext();
 
 //Example loading from a SQL Server or SQL Azure database table
-IDataView trainingDataView = mlContext.Data.LoadFromDatabaseTable<ModelInputData, SqlConnection>(connString: myConnString, 
+IDataView trainingDataView = mlContext.Data.LoadFromDbTable<ModelInputData, SqlConnection>(connString: myConnString, 
                                                                                                  tableName: "TrainingDataTable"); 
 ```
 
-**4. (Nice to have) Data loading from a database view:**
+**4. ('Nice to have') Data loading from a database view:**
 
 This is mostly a 'sugar method', but can be a 'nice to have feature' for users.
-The signature of the `DataOperationsCatalog.LoadFromDatabaseView` method looks like:
+The signature of the `DataOperationsCatalog.LoadFromDbView` method looks like:
 
 ```C#
-public IDataView LoadFromDatabaseView<TRow, DbConnection>(string connString, string viewName) where T : class;
+public IDataView LoadFromDbView<TRow, DbConnection>(string connString, string viewName) where T : class;
 ```
 
 Example code using it:
@@ -245,13 +244,13 @@ Example code using it:
 MLContext mlContext = new MLContext();
 
 //Example loading from a SQL Server or SQL Azure database view
-IDataView trainingDataView = mlContext.Data.LoadFromDatabaseView<ModelInputData, SqlConnection>(connString: myConnString, 
+IDataView trainingDataView = mlContext.Data.LoadFromDbView<ModelInputData, SqlConnection>(connString: myConnString, 
                                                                                                 viewName: "TrainingDatabaseView"); 
 ```
 
-## Support to connect from .NET assemblies embeded into the RDBMS server
+## Support connectivity from .NET assemblies embedded into the RDBMS server
 
-As introduced, the database loader should not only support remote/network connection to the RDBMS server but also support from C# running within the RDBMS server such as [SQL Server CLR integration](https://docs.microsoft.com/en-us/sql/relational-databases/clr-integration/clr-integration-overview?view=sql-server-2017) (aka [SQL CLR](https://en.wikipedia.org/wiki/SQL_CLR)), [Oracle Database Extensions for .NET](https://docs.oracle.com/cd/B19306_01/win.102/b14306/intro.htm), etc.
+As introduced, the database loader should not only support remote/network connection to the RDBMS server but also support connectivity from C# running within the RDBMS server such as [SQL Server CLR integration](https://docs.microsoft.com/en-us/sql/relational-databases/clr-integration/clr-integration-overview?view=sql-server-2017) (aka [SQL CLR](https://en.wikipedia.org/wiki/SQL_CLR)), [Oracle Database Extensions for .NET](https://docs.oracle.com/cd/B19306_01/win.102/b14306/intro.htm), etc.
 
 The only difference is the way you define the connection string, which simply provides **'context' string** (instead of server name, user, etc. when using the network), such as:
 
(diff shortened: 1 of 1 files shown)


docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
```diff
@@ -0,0 +1,407 @@
...
+- The code to load from a database should be extremely easy, a single line of code in most cases.
+- Tooling (Model Builder in VS and the CLI) and AutoML API should also support this feature.  
+
+# Solution
+
+The solution is to create an ML.NET database loader classes supporting the above scenarios. 
+
+The main supported features are:
+
+- **Easily load data into IDataView from databases**: The main goal is to load data into IDataView from databases very easily (a single line of code as one of the approaches) and properly (with good performance when streaming data while training).
```
markusweimer: "Load" suggests that the data is physically copied from the DBMS to the machine running the training. While this is undoubtedly a common scenario, is it envisioned to be the only one?

docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
```diff
@@ -0,0 +1,407 @@
...
+
+- **Train ML.NET models with data streaming coming from a database**: As the ultimate goal, the main purpose of loading data into an IDataView is to usually train an ML.NET model with data coming from a database table (or view or SQL sentence).
+
+- **Evaluate ML.NET models with data streaming coming from a database**: As secondary goal is to evaluate the quality/accuracy of an ML.NET model with data coming from a database table (or view or SQL sentence) by comparing it with a set or predictions. This goal should support:
```
markusweimer: This includes the feature where I can score data in a DBMS using ML.NET, right?

docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
```diff
@@ -0,0 +1,407 @@
...
+--------------------------------------
+
+TBD - Include diagram with blocks of AutoML, CLI and Model Buidler using this component/loader
+
+--------------------------------------
+
+
+## Design criteria
+
+- **Supported frameworks in .NET**: The DatabaseLoader component should be supported for the [frameworks supported by ML.NET](https://github.com/dotnet/machinelearning#installation) which include:
```
markusweimer: Wouldn't it be easier to only target .NET Standard 2.0?

docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
```diff
@@ -0,0 +1,407 @@
...
+
+- `SqlConnection` (SQL Server)
+- `NpgsqlConnection` (PostgreSQL)
+- `OracleConnection` (Oracle)
+- `MySqlConnection` (MySql)
+- etc.
+
+The specific type (`SqlConnection`, `NpgsqlConnection`, `OracleConnection`, etc.) is specified as a generic parameter so references to the needed NuGet packages will be provided by the user when trying to compile.
+
+**1. Data loading from a database table (Simplest approach):**
```
markusweimer: Is the differentiation between tables and views etc. necessary in the API? Naively, I would have expected to provide a SQL Query which returns tuples of a schema matching the `IDataView` to create. Whether that query is `SELECT * FROM [TABLE]` or anything else would not matter then, right?

docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
CESARDELATORRE: For this feature spec that the scenario in scope since it is probably the most common scenario:
Training-machine <--network--> Database-in-RDBMS-server

However, note that usually data will be streamed as the training process requests more data, in a similar way we do from large files in ML.NET. "Load" doesn't mean we load the whole database table or query 'in-memory' in the training machine unless the pipeline specifies to use cache, which should be used only for small datasets.

That's the most common scenario, for instance having an Azure SQL Database in the cloud or accessing data to any relational database (not only SQL Server) such as PostgreSQL, Oracle, MySQL, etc.

**Do you have any other scenario in mind?** 

Probably running the ML.NET code within SQL Server you mean?: [SQL Server CLR integration](https://docs.microsoft.com/en-us/sql/relational-databases/clr-integration/clr-integration-overview?view=sql-server-2017) - 
That scenario is also very interesting when training a model since the data is closer to the training process. 

However, that is initially out of the scope of this spec doc since it'll be a particular scenario for SQL Server on-premises and Azure SQL Database managed instance different to other RDBMS or even not possible in some cases:  

- Regular Azure SQL Database (single databases and elastic pools): NOT POSSIBLE
- Other RDBMS might NOT support the .NET CLR running inside the RDBMS server or have a particular implementation different to SQL Server CLR integration such as [Oracle Database Extensions for .NET](https://docs.oracle.com/cd/B19306_01/win.102/b14306/intro.htm) or PostgreSQL extension.

My point is that the 'ML.NET integration with a SQL-CLR approach' should probably be a different component with a different architecture/design and a different implementation per RDBMS, hence I believe we should have a different specs docs and different priorities depending on usage (i.e. CLR integration is not something very common in .NET applications, even less for .NET apps using non SQL Server RDBMS).But certainly, I find it particularly useful for intense operations such as training a model. :)

docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
CESARDELATORRE: Not initially, as I explained in the previous comment I initially think that should be a different feature in a different specs docs.

I'm guessing you mean [SQL Server CLR integration](https://docs.microsoft.com/en-us/sql/relational-databases/clr-integration/clr-integration-overview?view=sql-server-2017), right?

The implementation of that approach is initially very different and very particular per RDBMS. Even regular Azure SQL Database (single databases and elastic pools) don't support that, only 'SQL Server on-premises' and 'Azure SQL Database managed instance'.

Then, for other RDBMS such a feature is either using a very different implementation or not supported for the CLR.

But, as mentioned, I find it particularly useful for intense operations that need to be 'close' to the data, such as when training a model. :)

docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
CESARDELATORRE: You are right, differentiation between tables and views etc. are not necessary in the API, they are just convenient simpler methods since it'll be very common to consolidate a prepared dataset in a single table. But I agree that having the SQL Query sentence method can also be enough.

The 'LoadFromDbTable()' or 'LoadFromDbView()' methods are just "sugar methods" re-using the other two foundational methods.

So, yeah, I find those methods optional and we could provide them or not depending on customer feedback.

docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
CESARDELATORRE: Sure, we'll be using/targeting .NET Standard 2.0 precisely, that was my conversation with Diego Vega. I just  mean that .NET Framework and .NET Core apps should be supported. I'll make that clear in the specs doc.

docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
CESARDELATORRE: However, I'm thinking we might be able to support 'CLR integration' at the same time if we are not really creating any database object but simply supporting to use the CLR within the database server.

The only difference is the way you define the connection string, such as:

//SQL Server
SqlConnection conn   = new SqlConnection("context connection=true")

//Oracle
OracleConnection con = new OracleConnection();
con.ConnectionString = "context connection=true"; 

So we can probably support both scenarios. 👍 

What would be a different implementation is if we were building concrete database objects such as CLR Scalar-Valued Functions or stored procedures which are different for SQL Server, Oracle, etc.

So, yeah, I think we can also support 'CLR integration' at the same time. 👍

docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
CESARDELATORRE: I updated the specs docs so it includes cases such as [SQL Server CLR integration](https://docs.microsoft.com/en-us/sql/relational-databases/clr-integration/clr-integration-overview?view=sql-server-2017) as part of the design.

docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
CESARDELATORRE: Update in the specs so .NET Standard is mentioned explicitly.

docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
```diff
@@ -0,0 +1,463 @@
...
+
+Whenever in the samples below it is using a `SqlConnection` type, it could also be using any other supported DbConnection in .NET providers, such as:
+
+- `SqlConnection` (SQL Server)
+- `NpgsqlConnection` (PostgreSQL)
+- `OracleConnection` (Oracle)
+- `MySqlConnection` (MySql)
+- etc.
+
+The specific type (`SqlConnection`, `NpgsqlConnection`, `OracleConnection`, etc.) is specified as a generic parameter so references to the needed NuGet packages will be provided by the user when trying to compile.
```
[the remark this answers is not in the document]
CESARDELATORRE: Those types would be specified in the user's code, not in ML.NET code. Our parameters are Interfaces or base classes.
We prefer the users to need to provide a type (such as `SqlConnection`, `NpgsqlConnection`, `OracleConnection`) in their code so at that time they'll need to provide the dependency to their specific chosen database provider, that's the only reason we're specifying to provide a DbConnection type instead of a string with the provider such as "SQLServer" but then they'll need to know what NuGet assembly to add, etc. IF not provided properly they'll get an execution exception. Instead, by providing a type (such as `NpgsqlConnection`) it won't compile until you add the needed NuGet package.

docs/specs/mlnet-database-loader/mlnet-database-loader-specs.md
```diff
@@ -0,0 +1,463 @@
...
+
+`TRow` is the model's input data class (Observation class) to be used by IDataView and ML.NET API.
+
+The following code is how the mlContext.Data.LoadFromDbDataReader() method is used:
+
+```C#
+MLContext mlContext = new MLContext();
+
+string conString = YOUR_CONNECTION_STRING;
+using (SqlConnection connection = new SqlConnection(conString))
```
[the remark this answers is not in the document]
CESARDELATORRE: This is very interesting. 👍 Not sure if we want to go that path, but we'll evaluate it, for sure.
It might be interesting for our internal implementation code so we wouldn't need to take direct dependency on specific ADO.NET providers (as opposed to user's code where it is safer to take dependency to catch errors in compilation time). Thanks for these pointers to Orleans project! 👍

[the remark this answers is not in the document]
veikkoeeva: @CESARDELATORRE No problem. To be clear, I don't mean you need to use that code or do something similar. Maybe more like give ideas how it could like for what I think is a fairly minimal way to create something that produces at the end strongly typed classes while supporting all the features the various databases can have.

I wrote that part of the code so I'm somewhat familiar with the problems. For instance, some drivers (or their versions) have bugs that they are not necessarily asynchronous. Or they just don't support streaming without some extra work. Maybe the largest thing here is that I indeed wouldn't like to see this tied to any specific ORM since what would happen is that I'd pipe plain SQL through it and use for serializing data and likely end up with some inefficiencies along the way (multiple resultsets?). I don't know if people want to bring their own ORMs or just write efficient SQL that likely involves analytic/windowing functions too.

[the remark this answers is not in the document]
endintiers: Hi Cesar,

I think that as a pre-requisite for this, we should fix IDataView LoadFromEnumerable so it works for all Enumerables. Developers can easily make Enumerables from almost anything - including any database.

Then we can create a set of helpers to create Enumerables from abstract 'Databases' (with concrete implementations). This sounds like the kind of problem that has been solved many times before? Once this is implemented then this can just feed the resulting Enumerables to LoadFromEnumerable... This is how AutoML would feed in training and test data (from databases or any kind of structured store?).

LoadFromEnumerable doesn't always work at the moment because code further on assumes that the Enumerable-backed IDataView is thread-safe - this is not always the case (ex. EF Core Context).  See: https://github.com/endintiers/Unearth.Demo.ML.FromDB 

I understand why limiting concurrency for trainers was a bad idea. We shouldn't. In the cases when training from huge amounts of data (in a DB for example) it is slooowww.

So I propose starting by adding a switch to LoadFromEnumerable 'singleThreaded=true' which would intercept IDataView cache update requests and force them onto a single thread. This may slow down the trainers a bit but nowhere near as much as single-threading them (where possible) and we can improve performance with eager loading.

Doing that first would remove a lot of complexity from the DatabaseLoader task (threading no longer an issue).

- Chris

veikkoeeva:
I add to the comments about some specific pain points I have had and seen others have:

1. Have a quick and easy way to get access to database.
2. Being able to write complex queries like [row pattern recognition](https://sqlperformance.com/2019/04/t-sql-queries/row-pattern-recognition-in-sql) or [graph/JSON](https://bytefish.de/blog/sql_server_2019_graph_database/).
3. Being able to insert data efficiently to the database. I mean something like [here](https://github.com/veikkoeeva/orleans/blob/adonetstample/Samples/AdoNetSample/src/AdotNetSample.Database/Scripts/BulkLoads.sql#L4) or [OPEN ROWSET](https://docs.microsoft.com/en-us/sql/t-sql/functions/openrowset-transact-sql?view=sql-server-2017) or something alike.
4. Ability to stream large blobs.

So something like
```
using(var db = new Database("connectionstring"))
{
    var res = db.QueryAsync<dynamic>("SELECT * FROM Table1; SELECT * FROM Table2;");
}
```
Instead of using `dynamic` could be a class too (since there's multiple result sets, one might want to provide a `Func` to construct the objects (see the already linked pieces)). This way there's a quick way without too much scafolding and thinking to get working with a database and one can transport the queries across the tools.

In that sense a small core and a set of extension methods seem to cut the problem and maybe lead one to the idea of extending the core (like having a heuristic on how to bulk load data). Not everyone likes composing with functions and extensions, however.

PR #3859, opened 2019-06-13
Change default EvaluationMetric for LightGbm trainers to conform to d…

### LINKED ISSUE #3822: LightGbm default evaluation metrics in ML.NET do not conform to standalone LightGbm
Found while investigating fix for #3761 

In ML.NET LightGbm wrapper, the default EvaluationMetric is set to EvaluateMetricType.Error for multiclass, EvaluationMetricType.LogLoss for binary, and so on. On the other hand, in standalone LightGbm, the default evaluation metric is "", which means that LightGbm will automatically select the default metric for the given objective function.

This leads to inconsistent behavior from the user's perspective: If a user specified EvaluationMetric = EvaluateMetricType.Default, the parameter passed to LightGbm would be the empty string "" but if they do not specify EvaluationMetric at all, the parameter passed to LightGbm would be Error for multiclass, LogLoss for binary, and so on.

We need to investigate whether these metrics are indeed the defaults for the respective objective functions in LightGbm, and if they are not, then change the defaults in ML.NET to conform to standalone LightGbm. Note that this would be a breaking change.

najeeb-kazmi:
…efault metric in standalone LightGbm

Fixes #3822 

In ML.NET, the default `EvaluationMetric` for LightGbm is set to `EvaluateMetricType.Error` for multiclass, `EvaluationMetricType.LogLoss` for binary, and so on. This leads to inconsistent behavior from the user's perspective: If a user specified `EvaluationMetric = EvaluateMetricType.Default`, the parameter passed to LightGbm would be the empty string "", which is the LightGbm default and means that the metric is selected based on the objective. However, if they do not specify `EvaluationMetric` at all, the parameter passed to LightGbm would be Error for multiclass, LogLoss for binary, and so on. 

We should make the experience for LightGbm in ML.NET consistent with what a user of standalone LightGbm experiences, and not expect them to dig through LightGbm docs and ML.NET docs to find this out.

This PR makes the user experience consistent with standalone LightGbm by by changing the default `EvaluationMetric` in ML.NET to `EvaluationMetricType.Default`.

[LightGbm metric parameters docs](https://lightgbm.readthedocs.io/en/latest/Parameters.html#metric-parameters)

### DIFF
--- a/src/Microsoft.ML.LightGbm/LightGbmBinaryTrainer.cs
+++ b/src/Microsoft.ML.LightGbm/LightGbmBinaryTrainer.cs
@@ -162,7 +162,7 @@ namespace Microsoft.ML.Trainers.LightGbm
             [Argument(ArgumentType.AtMostOnce,
                 HelpText = "Evaluation metrics.",
                 ShortName = "em")]
-            public EvaluateMetricType EvaluationMetric = EvaluateMetricType.Logloss;
+            public EvaluateMetricType EvaluationMetric = EvaluateMetricType.Default;
 
             static Options()
             {
--- a/src/Microsoft.ML.LightGbm/LightGbmMulticlassTrainer.cs
+++ b/src/Microsoft.ML.LightGbm/LightGbmMulticlassTrainer.cs
@@ -95,7 +95,7 @@ namespace Microsoft.ML.Trainers.LightGbm
             [Argument(ArgumentType.AtMostOnce,
                 HelpText = "Evaluation metrics.",
                 ShortName = "em")]
-            public EvaluateMetricType EvaluationMetric = EvaluateMetricType.Error;
+            public EvaluateMetricType EvaluationMetric = EvaluateMetricType.Default;
 
             static Options()
             {
--- a/src/Microsoft.ML.LightGbm/LightGbmRankingTrainer.cs
+++ b/src/Microsoft.ML.LightGbm/LightGbmRankingTrainer.cs
@@ -143,7 +143,7 @@ namespace Microsoft.ML.Trainers.LightGbm
             [Argument(ArgumentType.AtMostOnce,
                 HelpText = "Evaluation metrics.",
                 ShortName = "em")]
-            public EvaluateMetricType EvaluationMetric = EvaluateMetricType.NormalizedDiscountedCumulativeGain;
+            public EvaluateMetricType EvaluationMetric = EvaluateMetricType.Default;
 
             static Options()
             {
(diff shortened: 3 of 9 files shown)


src/Microsoft.ML.LightGbm/LightGbmBinaryTrainer.cs
```diff
@@ -162,7 +162,7 @@ public enum EvaluateMetricType
             [Argument(ArgumentType.AtMostOnce,
                 HelpText = "Evaluation metrics.",
                 ShortName = "em")]
-            public EvaluateMetricType EvaluationMetric = EvaluateMetricType.Logloss;
+            public EvaluateMetricType EvaluationMetric = EvaluateMetricType.Default;
```
yaeldekel: >Default [](start = 76, length = 7)

Isn't this a breaking change?
cc @eerhardt

najeeb-kazmi: Yes, but the other option is to have an inconsistent user experience. I talked to @ebarsoumMS about this. Let's discuss and reach a conclusion.

eerhardt: It's not an "API breaking change". I think it falls into the scenarios that @TomFinley listed here https://github.com/dotnet/machinelearning/issues/3602#issuecomment-487192476.

> However even many years later sometimes we still have somewhat troublesome defaults running around

Here, if there is a better default value, I think it is acceptable to change the default.

najeeb-kazmi: Also, this doesn't actually change training behavior, nor the metrics calculated by ML.NET evaluators. Just changes the metric that LightGbm calculates internally. 

In ML.NET, when we do the following (e.g. for binary classification)
```csharp
var transformedTestData = model.Transform(testData);
var metrics = mlContext.BinaryClassification.Evaluate(transformedTestData);
```
the evaluator computes all relevant metrics for binary classification regardless of what is specified by LightGbm's `EvaluationMetric` parameter.

justinormont: It may control LightGBM's early stopping, but otherwise I think this is a NOOP change. ML.NET doesn't relay the stdout from LightGBM to the user, and ML.NET uses its own evaluators for computing the final metrics.

Users could benefit from ML.NET relaying this info back to the user. This would allow a GUI to show the learning curves in real time (or as text output from a CLI):
[](https://github.com/catboost/catboost/tree/master/catboost/benchmarks/gpu_vs_cpu_training_speed)

PR #3860, opened 2019-06-13
[AutoML] add task agnostic wrappers for autofit calls

Dmitry-A:


justinormont:
@Dmitry-A: Rather late, though this PR could use an Issue tagged, and a description.

PR #3875, opened 2019-06-18
Add bindings for RowImpl in time series SequentialTransformerBase

### LINKED ISSUE #3874: Time series Sequential Transform RowImpl needs to have a binding mechanism
We need to keep track of bindings when GetGetter is called in the case of DataTransformer, these bindings will be used to map column id to relative column id of the transformer.

codemzs:
fixes #3874


codemzs:
Checking this in to unblock @ganik 's NimbusML integration PR.

PR #3882, opened 2019-06-19
[AutoML] bring AutoML API library to master

Dmitry-A:
This moves the AutoML API code from a feature branch to master. The CLI will move next.

Closes https://github.com/dotnet/machinelearning/issues/4008


codemzs:
Can you please create an issue and link to this PR?

pkg/Microsoft.ML.AutoML/Microsoft.ML.AutoML.nupkgproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk" DefaultTargets="Pack">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageDescription>ML.NET AutoML: Optimizes an ML pipeline for your dataset, by automatically locating the best feature engineering, model, and hyperparameters</PackageDescription>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ML" Version="1.0.0" />
```
eerhardt: My opinion here is that we should be using `<ProjectReference` instead of `PackageReference`.

The reasoning is:

1. That's the way all other libraries in this repo reference each other.
2. If AutoML wants to use a new feature (or needs a bug fix) in the core ML.NET, we will not be able to use it, since we are referencing an older core ML.NET version.
3. It can easily create version mis-match problems. For example, I am an app that consumes `ML.AutoML` NuGet package. By transitivity I also reference `Microsoft.ML` `v1.0.0` and `Microsoft.ML.LightGBM` `v1.0.0`. If I also add a PackageReference to `Microsoft.ML` `v1.1.0`, NuGet will [uplift the reference to the `v1.1.0` version](https://docs.microsoft.com/en-us/nuget/consume-packages/dependency-resolution#nearest-wins). However, no one uplifted `Microsoft.ML.LightGBM` to `v1.1.0`, so I will get `LightGBM` `v1.0.0` matched with a core `Microsoft.ML` `v1.1.0`, which may not be compatible because there are InternalsVisibleTo between `Microsoft.ML` and `Microsoft.ML.LightGBM`.

src/Microsoft.ML.AutoML/Utils/MLNetUtils/Hashing.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.ML.AutoML
+{
+    internal static class Hashing
```
eerhardt: We can't use the existing utility functions in ML.NET because they aren't public, right?

src/Microsoft.ML.AutoML/Utils/MLNetUtils/VectorUtils.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Microsoft.ML.AutoML
+{
+    internal static class VectorUtils
+    {
+        public static double GetMean(double[] vector)
+        {
+            double sum = 0;
+            for (int i = 0; i < vector.Length; i++)
+            {
+                sum += vector[i];
```
eerhardt: Do we want to use `CpuMath` here so this is vectorized?

test/Microsoft.ML.AutoML.Tests/Microsoft.ML.AutoML.Tests.csproj
```diff
@@ -0,0 +1,35 @@
+﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
+  <PropertyGroup>
+    <IsPackable>false</IsPackable>
+    <RootNamespace>Microsoft.ML.AutoML.Test</RootNamespace>
```
eerhardt: You shouldn't need `RootNamespace` or `ProjectGuid` in this .csproj.

`RootNamespace` gets defaulted to the project name, and since it already matches, this is redundant.
`ProjectGuid` is unnecessary in SDK-style .csproj.

test/Microsoft.ML.AutoML.Tests/Microsoft.ML.AutoML.Tests.csproj
```diff
@@ -0,0 +1,35 @@
...
+    <ProjectGuid>{EEC3ACCD-C8BC-4261-820F-EF1340B546DE}</ProjectGuid>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
+    <PackageReference Include="MSTest.TestAdapter" Version="1.3.2" />
```
eerhardt: We use XUnit (in all .NET Foundation projects). These tests should be converted from MSTest to XUnit.

test/Microsoft.ML.AutoML.Tests/Microsoft.ML.AutoML.Tests.csproj
```diff
@@ -0,0 +1,35 @@
...
+    <PackageReference Include="MSTest.TestFramework" Version="1.3.2" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.ML.AutoML\Microsoft.ML.AutoML.csproj">
+      <Project>{219D2E17-FEFD-4B03-BBA7-E4D4CEBA4BBB}</Project>
```
eerhardt: You don't need `<Project>` and `<Name>` here. You can just have:

```xml
 <ProjectReference Include="..\..\src\Microsoft.ML.AutoML\Microsoft.ML.AutoML.csproj" />
```

src/Microsoft.ML.AutoML/Microsoft.ML.AutoML.csproj
```diff
@@ -0,0 +1,32 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>7.3</LangVersion>
```
eerhardt: The root `Directory.Build.props` file defaults this to `latest`, so you shouldn't need to set this property.

You can also remove the `RootNamespace` element below.

src/Microsoft.ML.AutoML/Microsoft.ML.AutoML.csproj
```diff
@@ -0,0 +1,32 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ML" Version="$(MlDotNetPackageVersion)" />
+    <PackageReference Include="Microsoft.ML.LightGBM" Version="$(MlDotNetPackageVersion)" />
+    <PackageReference Include="Microsoft.ML.Mkl.Components" Version="$(MlDotNetPackageVersion)" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
eerhardt: You don't need `GenerateDocumentationFile` here, since it is defaulted to true in https://github.com/dotnet/machinelearning/blob/9a80b78d7aba8463859975119e697afb72f91c09/src/Directory.Build.props#L6.

src/Microsoft.ML.AutoML/Microsoft.ML.AutoML.csproj
```diff
@@ -0,0 +1,32 @@
...
+  </PropertyGroup>
+
+  <!-- here we place this assembly into CLI package dir so it gets included in CLI nuget -->
```
eerhardt: I think you can delete this commented out piece. We will need to rethink how we do the CLI packaging when it comes in. If we really need this again, we can get it from source control history.

Dmitry-A:
PR for #4008

eerhardt:
Any update on this PR? I didn't see responses to my comments.

src/Microsoft.ML.AutoML/Utils/MLNetUtils/Hashing.cs
Dmitry-A: they're not but now that I have taken project level dependency I can do BestFriends and nuke the private copies.

Dmitry-A: Yes we can.  I made it work for some of them but it requires some non-trivial changes to make them all go away completely, I filed a work item https://github.com/dotnet/machinelearning/issues/4064 for this.

pkg/Microsoft.ML.AutoML/Microsoft.ML.AutoML.nupkgproj
Dmitry-A: Yes agreed, and done.

src/Microsoft.ML.AutoML/Utils/MLNetUtils/VectorUtils.cs
Dmitry-A: Can't do it for this release due to test project dependency, should be done as part of https://github.com/dotnet/machinelearning/issues/4064

test/Microsoft.ML.AutoML.Tests/Microsoft.ML.AutoML.Tests.csproj
Dmitry-A: converted.

src/Microsoft.ML.AutoML/Microsoft.ML.AutoML.csproj
Dmitry-A: Removed.

pkg/Microsoft.ML.AutoML/Microsoft.ML.AutoML.nupkgproj
eerhardt: This doesn't appear to be done. Can you make this use `<ProjectReference`?

src/Microsoft.ML.AutoML/Assembly.cs
```diff
@@ -0,0 +1,13 @@
...
+[assembly: InternalsVisibleTo("mlnet, PublicKey=00240000048000009400000006020000002400005253413100040000010001004b86c4cb78549b34bab61a3b1800e23bfeb5b3ec390074041536a7e3cbd97f5f04cf0f857155a8928eaa29ebfd11cfbbad3ba70efea7bda3226c6a8d370a4cd303f714486b6ebc225985a638471e6ef571cc92a4613c00b8fa65d61ccee0cbe5f36330c9a01f4183559f1bef24cc2917c6d913e3a541333a1d05d9bed22b38cb")]
+[assembly: InternalsVisibleTo("mlnet.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010015c01ae1f50e8cc09ba9eac9147cf8fd9fce2cfe9f8dce4f7301c4132ca9fb50ce8cbf1df4dc18dd4d210e4345c744ecb3365ed327efdbc52603faa5e21daa11234c8c4a73e51f03bf192544581ebe107adee3a34928e39d04e524a9ce729d5090bfd7dad9d10c722c0def9ccc08ff0a03790e48bcd1f9b6c476063e1966a1c4")]
+[assembly: InternalsVisibleTo("Benchmark, PublicKey=00240000048000009400000006020000002400005253413100040000010001004b86c4cb78549b34bab61a3b1800e23bfeb5b3ec390074041536a7e3cbd97f5f04cf0f857155a8928eaa29ebfd11cfbbad3ba70efea7bda3226c6a8d370a4cd303f714486b6ebc225985a638471e6ef571cc92a4613c00b8fa65d61ccee0cbe5f36330c9a01f4183559f1bef24cc2917c6d913e3a541333a1d05d9bed22b38cb")]
```
eerhardt: What is this `Benchmark` project?

I think this should be removed if this project doesn't exist in this repo.

src/Microsoft.ML.AutoML/AutoMlUtils.cs
```diff
@@ -0,0 +1,23 @@
...
+using System;
+using System.Threading;
+
+namespace Microsoft.ML.AutoML
+{
+    internal static class AutoMlUtils
+    {
+        public static readonly ThreadLocal<Random> Random = new ThreadLocal<Random>(() => new Random());
+
+        public static void Assert(bool boolVal, string message = null)
```
eerhardt: A few problems with this method:

1. Typically a method named `Assert` should be removed on `Release` builds. Normal .NET devs expect asserts to be debug only.
2. ML.NET already has a mechanism for checking bool values, and either throwing or asserting. Since this code is a part of ML.NET, it should be using those mechanisms.

src/Microsoft.ML.AutoML/Microsoft.ML.AutoML.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <DefineConstants>CORECLR</DefineConstants>
```
eerhardt: Do you actually use any `#if CORECLR` code? I didn't see any in this review.

If not, you can remove this `DefineConstants` line.

If you want to keep this in, you should be doing it differently - it should be defined as 

```
<DefineConstants>$(DefineConstants);CORECLR</DefineConstants>
```

src/Microsoft.ML.AutoML/Microsoft.ML.AutoML.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <IncludeInPackage>Microsoft.ML.AutoML</IncludeInPackage>
+  </PropertyGroup>
+
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.ML.Core\Microsoft.ML.Core.csproj" />
```
eerhardt: I see you are using `ProjectReference` here, but `PackageReference` above in the .nupkgproj. This won't work because you will be compiling against the latest code, but the NuGet package is referencing old binaries. So any new methods you use from ML.NET won't be in the old binaries, and cause runtime errors.

src/Microsoft.ML.AutoML/Utils/Logger.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.ML.Runtime;
+
+namespace Microsoft.ML.AutoML
+{
+    internal class AutoMLLogger
```
eerhardt: (minor, doesn't need to be solved for this PR) -

I wonder if this object has enough value over just using the `IChannel` itself.

src/Microsoft.ML.AutoML/Utils/MLNetUtils/Contracts.cs
```diff
@@ -0,0 +1,94 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Diagnostics;
+using System.Globalization;
+using Microsoft.ML.Runtime;
+
+namespace Microsoft.ML.AutoML
+{
+    internal static class Contracts
```
eerhardt: ??? Why not use the existing `Contracts` from ML.NET and not duplicate the whole thing?

src/Microsoft.ML.AutoML/Utils/MLNetUtils/ArrayDataViewBuilder.cs
```diff
@@ -0,0 +1,468 @@
...
+using Microsoft.ML.Runtime;
+
+namespace Microsoft.ML.AutoML
+{
+    using BitArray = System.Collections.BitArray;
+
+    /// <summary>
+    /// This is a class for composing an in memory IDataView.
+    /// </summary>
+    internal sealed class ArrayDataViewBuilder
```
eerhardt: This appears to be duplicated from:

https://github.com/dotnet/machinelearning/blob/a802127883c0020516aa573436029cabed65d591/src/Microsoft.ML.Data/DataView/ArrayDataViewBuilder.cs#L14-L18

We shouldn't be doing that.

src/Microsoft.ML.Core/Properties/AssemblyInfo.cs
```diff
@@ -13,6 +13,7 @@
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.InferenceTesting" + PublicKey.TestValue)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.StaticPipelineTesting" + PublicKey.TestValue)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.OnnxTransformerTest" + PublicKey.TestValue)]
+[assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.AutoML.Tests" + PublicKey.TestValue)]
```
eerhardt: Why do the tests need IVT access to these assemblies?

test/Microsoft.ML.AutoML.Tests/Microsoft.ML.AutoML.Tests.csproj
```diff
@@ -0,0 +1,47 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <DefineConstants>CORECLR</DefineConstants>
```
eerhardt: Is this define actually used? I think it can be removed.

src/Microsoft.ML.AutoML/Assembly.cs
Dmitry-A: it exists in internal repo and needs access to some stuff for logging purposes, so I can't remove it.

src/Microsoft.ML.AutoML/Microsoft.ML.AutoML.csproj
Dmitry-A: Ok I think we just copied it from somewhere, will remove.

src/Microsoft.ML.AutoML/Assembly.cs
eerhardt: IMO - I think that is less than ideal. Is there a plan for moving that project to this repo? Usually having tests (even perf tests) along with the code in the same repo is a good idea.

Dmitry-A: Yes it's a tricky situation, benchmarking is an internal process, and it's all private and will remain private.  I could try to pull some of the code in but overall it will be strange to have it here since the stuff it doesn't make much sense without the context of our benchmarking infra.  It's a relatively thin shim between that infra and AutoML.NET.

eerhardt:
@Dmitry-A  Any update here?

src/Microsoft.ML.AutoML/Assembly.cs
Dmitry-A: Filed https://github.com/dotnet/machinelearning/issues/4114 on this.  This is tricky to fix and a good fix needs some cooperation from ML.NET pipelines.

src/Microsoft.ML.AutoML/AutoMlUtils.cs
Dmitry-A: Nuked this and there's another copy we had that's used here and there... what a mess, thanks for pointing it out.

src/Microsoft.ML.AutoML/Utils/Logger.cs
Dmitry-A: AFAIR this used to be a bit more elaborate but got trimmed down since, removed it.  I don't like how it hides MessageSensitivity and defaults it to None.  Thought about if it's reasonable to rename logger -> channel, decided to stick with logger since that's how it is really used.

test/Microsoft.ML.AutoML.Tests/Microsoft.ML.AutoML.Tests.csproj
Dmitry-A: yep

src/Microsoft.ML.Core/Properties/AssemblyInfo.cs
Dmitry-A: Ugh.  For using internal cursor object for a dummy DataView test fixture.  And the tests were all shooting in the air since they were expecting a certain exception for some specific conditions they were trying to test, but actually they all passed simply because the same exception is thrown if the dataset is empty (which it was).  Fixed test validation to be specific, fixed tests that weren't set up correctly so they produce the error they were looking to test.  And removed internalsvisible of course.

Dmitry-A: Hmm not so easy, we had a copy of ArrayDataViewBuilder in AutoML, removing that brings back the issue.

src/Microsoft.ML.AutoML/Utils/MLNetUtils/ArrayDataViewBuilder.cs
Dmitry-A: yes, removed this one too.  I filed https://github.com/dotnet/machinelearning/issues/4064 the first time I looked at it, this time I have a bit more time so looking into what else can I remove without much code churn.

src/Microsoft.ML.AutoML/Microsoft.ML.AutoML.csproj
Dmitry-A: done

src/Microsoft.ML.AutoML/Microsoft.ML.AutoML.csproj
Dmitry-A: done

src/Microsoft.ML.AutoML/Microsoft.ML.AutoML.csproj
Dmitry-A: replaced with nupkgproj references, checked what nuspec looks like in the final product and looks right to me.  Interesting that we only had to take a direct dependency on lgbm but Daniel says that's just a coincidence since others are pulled in from other dependencies.

  <dependencies>
      <group targetFramework=".NETStandard2.0">
        <dependency id="Microsoft.ML.LightGBM" version="1.3.0-preview-28005-0" exclude="Build,Analyzers" />
        <dependency id="Microsoft.ML.Mkl.Components" version="1.3.0-preview-28005-0" exclude="Build,Analyzers" />
        <dependency id="Microsoft.ML" version="1.3.0-preview-28005-0" exclude="Build,Analyzers" />
      </group>
    </dependencies>

This is what it used to be:

    <dependencies>
      <group targetFramework=".NETStandard2.0">
        <dependency id="Microsoft.ML" version="1.0.0-preview" include="All" />
        <dependency id="Microsoft.ML.LightGBM" version="1.0.0-preview" include="All" />
        <dependency id="Microsoft.ML.Mkl.Components" version="1.0.0-preview" include="All" />
      </group>
    </dependencies>

src/Microsoft.ML.AutoML/Utils/MLNetUtils/Contracts.cs
Dmitry-A: yeah, ugh. done

src/Microsoft.ML.AutoML/Microsoft.ML.AutoML.csproj
```diff
@@ -1,16 +1,21 @@
...
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.ML.Core\Microsoft.ML.Core.csproj" />
+    <ProjectReference Include="..\Microsoft.ML.CpuMath\Microsoft.ML.CpuMath.csproj" />
     <ProjectReference Include="..\Microsoft.ML.LightGbm\Microsoft.ML.LightGbm.csproj" />
     <ProjectReference Include="..\Microsoft.ML.Mkl.Components\Microsoft.ML.Mkl.Components.csproj" />
     <ProjectReference Include="..\Microsoft.ML.Transforms\Microsoft.ML.Transforms.csproj" />
   </ItemGroup>
 
+
```
Dmitry-A: oops removed it but somehow it didn't stick, trying again

src/Microsoft.ML.Data/Properties/AssemblyInfo.cs
```diff
@@ -13,6 +13,8 @@
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Predictor.Tests" + PublicKey.TestValue)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.TimeSeries.Tests" + PublicKey.TestValue)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Benchmarks" + PublicKey.TestValue)]
+[assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.StaticPipelineTesting" + PublicKey.TestValue)]
```
eerhardt: `Microsoft.ML.StaticPipelineTesting` is old and has been removed. You shouldn't be adding it back here.

test/Microsoft.ML.AutoML.Tests/Microsoft.ML.AutoML.Tests.csproj
```diff
@@ -0,0 +1,43 @@
...
+  <ItemGroup>
+    <NativeAssemblyReference Include="MklProxyNative" />
+    <NativeAssemblyReference Include="CpuMathNative" />
+    <NativeAssemblyReference Include="FastTreeNative" />
+    <NativeAssemblyReference Include="LdaNative" />
+    <NativeAssemblyReference Include="MklImports" />
+    <NativeAssemblyReference Condition="'$(OS)' == 'Windows_NT'" Include="libiomp5md" />
+  </ItemGroup>
+
+  <!-- TensorFlow is 64-bit only -->
```
eerhardt: Why would the AutoML tests be dependent on TensorFlow?

src/Microsoft.ML.Data/Properties/AssemblyInfo.cs
Dmitry-A: done

test/Microsoft.ML.AutoML.Tests/Microsoft.ML.AutoML.Tests.csproj
Dmitry-A: probably a cut&paste from core.tests, removed and tests pass just fine.

codemzs:
@Dmitry-A Can you please rebase to master and push to ensure the build isn't broke? We want to put these commits on top of master.

PR #3894, opened 2019-06-21
Fixed build errors resulting from upgrading to VS2019 compilers

### LINKED ISSUE #3893: Visual Studio 2019 Building Problem
I got a clean machine and installed the latest Visual Studio 2019. After calling `.\build.cmd`, I got the following error message.
```
  "Repos\machinelearning\bin\obj\x64.Debug\Native\INSTALL.vcxproj" (rebuild target) (1) ->
  "Repos\machinelearning\bin\obj\x64.Debug\Native\ALL_BUILD.vcxproj" (default target) (3:2) ->
  "Repos\machinelearning\bin\obj\x64.Debug\Native\SymSgdNative\SymSgdNative.vcxproj" (default target) (10:2) ->
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(429,5): error MSB6006: "CL.exe" exited with code 2. [Repos\machinelearning\bin\obj\x64.Debug\Native\SymSgdNative\SymSgdNative.vcxproj] [Repos\machinelearning\src\Native\build.proj]
```
Did I do something naïve? To resolve that bug, I remove one line below the it works.
```
diff --git a/src/Native/CMakeLists.txt b/src/Native/CMakeLists.txt
index a814277e..87b0606c 100644
--- a/src/Native/CMakeLists.txt
+++ b/src/Native/CMakeLists.txt
@@ -21,8 +21,6 @@ if(WIN32)
     add_compile_options($<$<CONFIG:Debug>:/MTd>) # /MT will static link the VC runtime library, so it doesn't need to be installed on the target machine
     add_compile_options($<$<CONFIG:Release>:/MT>)
     add_compile_options($<$<CONFIG:RelWithDebInfo>:/MT>)
-    add_compile_options(/guard:cf)
     add_compile_options(/d2Zi+) # make optimized builds debugging easier
     add_compile_options(/nologo) # Suppress Startup Banner
     add_compile_options(/W3) # set warning level to 3
     add_compile_options(/WX) # treat warnings as errors
@@ -46,9 +44,6 @@ if(WIN32)
```

harishsk:
Fixes #3893 

The default CMAKE_C_FLAG for debug configuration sets /ZI to generate a PDB capable of edit and continue. In the new compilers, this is incompatible with /guard:cf which we set for security reasons. So to fix, we need to go back to a regular pdb generated by the /Zi flag. 

Since this part of the default CMake rules, we need to override the full default set and update only that particular flag.


src/Native/CMakeLists.txt
```diff
@@ -8,6 +8,14 @@ set(RESOURCES)
 include_directories("${CMAKE_BINARY_DIR}/../../")
 
 if(WIN32)
+    # Clobber and reset the default C and CXX flags because
+    # CMake uses /ZI (Edit and Continue) for generating pdbs
+    # which is incompatible with the /guard:cf flag we set below
+    # for security. So we use the default flags set by CMake 
+    # and reset /ZI with /Zi
+    set(CMAKE_C_FLAGS_DEBUG "/MDd /Zi /Ob0 /Od /RTC1 /JMC")
```
wschin: Where do those flags `/MDd /Zi /Ob0 /Od /RTC1 /JMC` come from? Are they default flags in CMake? Could you please add more details for maintaining them in the future?

harishsk: Yes, those are the default flags in CMake. I will add a bit more detail and submit another PR soon.

harishsk:
Purely for educational purposes...these are the recent changes in Windows-MSVC.cmake that caused this issue:

```
# default value for Just My Code flag
set(_JMC "")
# default value for Debug Information Format flag
set(_DEBUGINFOFORMAT_DEBUG "/Zi")


if("x${CMAKE_C_COMPILER_ID}" STREQUAL "xMSVC" OR "x${CMAKE_CXX_COMPILER_ID}" STREQUAL "xMSVC" )
    # JMC only supported in MSVC
    if(MSVC_VERSION GREATER_EQUAL 1915)
      set(_JMC "/JMC")
    endif()
    # /ZI is supported by MSVC only, not by clang-cl for example.
    set(_DEBUGINFOFORMAT_DEBUG "/ZI")
  endif()

string(APPEND CMAKE_${lang}_FLAGS_DEBUG_INIT " /MDd ${_DEBUGINFOFORMAT_DEBUG} /Ob0 /Od ${_RTC1} ${_JMC}")
```

```

**Comment by janvorli** [comment]
@harishsk where have you found the stuff above? I've looked at the Windows-MSVC.cmake in all versions from 3.14 till 3.15 and none of this was there (looking via https://github.com/Kitware/CMake). And cmake 3.14.2 with VS2019 builds fine without your change (And the vcxproj files contain /Zi, not /ZI).

**Comment by eerhardt** [comment]
I am experiencing the same as @janvorli. I can build successfully on VS 2019 with CMake 3.14.5 without this change.

I sort of think this change should be reverted until we know more. I don't like the fact that we are clobbering the defaults.

**Comment by harishsk** [comment]
Where is your CMake located? And what is the version reported for CMake.exe --version

Mine is here: "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe"

And it reports: cmake version 3.14.19050301-MSVC_2

The changes I reported above are in this file:
c:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\share\cmake-3.14\Modules\Platform\Windows-MSVC.cmake

**Comment by eerhardt** [comment]
My CMake is located at `C:\Program Files\CMake` and I installed `3.14.5` from `https://cmake.org/download/` and added it to my `$PATH`. This is all documented on https://github.com/dotnet/machinelearning/blob/master/docs/building/windows-instructions.md#required-software.

@janvorli informed me there might be an issue with the CMake shipped by VS 2019:

> Comparing the Windows-MSVC.cmake of vanilla 3.14.2 and the 3.14 in VS2019, I can see that they removed this at one place:

```
# default value for Just My Code flag
set(_JMC "")
# default value for Debug Information Format flag
set(_DEBUGINFOFORMAT_DEBUG "/Zi")
```

> And added this at another

```
  if("x${CMAKE_C_COMPILER_ID}" STREQUAL "xMSVC" OR "x${CMAKE_CXX_COMPILER_ID}" STREQUAL "xMSVC" )
    # JMC only supported in MSVC
    if(MSVC_VERSION GREATER_EQUAL 1915)
      set(_JMC "/JMC")
    endif()
    # /ZI is supported by MSVC only, not by clang-cl for example.
    set(_DEBUGINFOFORMAT_DEBUG "/ZI")
  endif()
```

>    It seems as if they just made a mistake in the "i" case

@jkotas - do you happen to know anyone on the Visual Studio team who works on the CMake they ship? I'd like to find out if this is a bug in VS 2019.

**Comment by janvorli** [comment]
Btw, it seems that a cleaner way to fix the problem would be to use `string(REPLACE ....)` cmake command to just replace `/ZI` by `/Zi` in CMAKE_C_FLAGS_DEBUG instead of overwriting all the options with what we assume should be there. Like this:
```cmake
string(REPLACE "/ZI" "/Zi" CMAKE_C_FLAGS_DEBUG  ${CMAKE_C_FLAGS_DEBUG})
```

harishsk:
Yes, I have CMake installed in Program Files on my laptop and I don't see this bug. On my desktop, I don't have CMake installed and the one shipped in Visual Studio gets used, resulting in this bug. 

This is a bug in VS2019. And as my comment in the fix states, we can remove those lines after the defaults are fixed.

I agree what you propose is a cleaner fix. I have tested it and submitted a new [pull request](https://github.com/dotnet/machinelearning/pull/3896). 

@janvorli  and @eerhardt Can you please review and approve?

jkotas:
> do you happen to know anyone on the Visual Studio team who works on the CMake they ship? 

Take a look at https://github.com/microsoft/CMake . I believe it has the sources for the version that VS ships; and you can also find the people who work on it there.

PR #3896, opened 2019-06-21
Checked in a better fix based on code review

### LINKED ISSUE #3893: Visual Studio 2019 Building Problem

harishsk:
Fixes #3893 

The earlier fix was clobbering theCMake defaults. @janvorli  suggested a better fix to simply replace the offending flag. This checkin has that fix.


eerhardt:
Do we still want/need this change?

> add_compile_options(/Zo) # make optimized builds debugging easier. /Zo is the newer documented flag.

Or should that be reverted as well?

src/Native/CMakeLists.txt
```diff
@@ -8,16 +8,12 @@ set(RESOURCES)
...
-    # for security. So we use the default flags set by CMake 
-    # and reset /ZI with /Zi
+    # There seems to be a bug in the latest VS2019 
+    # which is adding /ZI (which conflicts with /guard:cf) instead of /Zi. 
 	message("CMAKE_C_FLAGS_DEBUG is ${CMAKE_C_FLAGS_DEBUG}")
-	message("CMAKE_CXX_FLAGS_DEBUG is ${CMAKE_CXX_FLAGS_DEBUG}")
-	message("In a future version, If the default compiler flags no longer contain the /ZI flag, delete this message block and the two lines below.")
-    set(CMAKE_C_FLAGS_DEBUG "/MDd /Zi /Ob0 /Od /RTC1 /JMC")
-    set(CMAKE_CXX_FLAGS_DEBUG "/MDd /Zi /Ob0 /Od /RTC1 /JMC")
+	message("In a future version, if the default compiler flags no longer contain the /ZI flag, delete this message block and the line below.")
```
eerhardt: (nit) `and the line below.` should be `and the two lines below.`

eerhardt: Instead of `if the default compiler flags no longer contain the /ZI flag`, can we say `if VS 2019's CMake no longer contain the /ZI flag` ?

harishsk: Yes, the /Zo is a good change. The behavior is the same, but /Zo is the documented flag.

harishsk:
@eerhardt I have updated the message. Please take a look.

PR #3881, opened 2019-06-19
ONNXTransform Upgrade to Enable Non-tensor Types

wschin:
The current ONNXTransform only operates on tensor types. As many ONNX models (especially classifiers) produce Seq<Map<T, float>> where T can be either int or string, we should remove that limitation.

This should fix #3900.

PR #3908, opened 2019-06-26
Move from phases to jobs and use bring your own cloud pool

safern:
This is needed to unblock: https://github.com/dotnet/machinelearning/pull/3881

Basically the hosted azure devops machines have limited hard drive (25gb) and since tests are executed on the build machine it is running out of space. BYOC pools have 512 gb so that will unblock this scenario.

However I do think, machinelearning repo should move to run tests on helix in the long term.

**Note:** *this is easier to review if omitting whitespaces from the diff*

cc: @codemzs @wschin

### DIFF
--- a/.vsts-dotnet-ci.yml
+++ b/.vsts-dotnet-ci.yml
@@ -10,11 +10,12 @@ resources:
   - container: UbuntuContainer
     image: mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-16.04-mlnet-207e097-20190312152303
 
-phases:
-- template: /build/ci/phase-template.yml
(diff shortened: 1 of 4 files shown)


wschin:
Just cherry-pick you changes to my branch and trigger a new build. Let's see if we can publish packages then. I will discard your changes on my side later so that we can have your changes from the right source. Thanks!

eerhardt:
@codemzs @wschin - I think moving from phases to jobs should happen in the official build as well: https://github.com/dotnet/machinelearning/blob/master/build/vsts-ci.yml.

I've opened #3926 for this.

PR #3912, opened 2019-06-26
Fix assignment of a variable to itself

flash2048:
The **invMap** variable is assigned to itself. Removed it.


codemzs:
thanks, @flash2048

Issue #3688, opened 2019-05-09
LightGBM: IsUnbalanced is relevant for OVA multi class


rauhs:
Version: 1.0

Setup:

- LightGBM Multiclass trainer
- NOT softmax (thus: "objective" = "mutliclassova").

Before 1.0 we had `UnbalancedSets` available for the three booster configs, but this option was removed it seems.

Though, that option very much had an effect on performance in this case.

You can see that the OVA objective just calls the binary one:

https://github.com/microsoft/LightGBM/blob/0a4a7a86f5a1d3146c36c7d8c082154a193d4893/src/objective/multiclass_objective.hpp#L173-L182

which will happily use the config parameter:

https://github.com/microsoft/LightGBM/blob/0a4a7a86f5a1d3146c36c7d8c082154a193d4893/src/objective/binary_objective.hpp#L20-L36

glebuk:
Perhaps we can expose this parameter in Multi-class LightGBM?

abgoswam:
I believe the bug was caused by the following [commit](https://github.com/dotnet/machinelearning/pull/2948/files#diff-8173688cd56a37cbcc92012f7d22a31dL152) 


- From what I can tell, the PR  tried to ensure that each trainer  had its own respective trainer class.

- In particular the `UnbalancedSets` parameter was  [removed](https://github.com/dotnet/machinelearning/pull/2948/files#diff-8173688cd56a37cbcc92012f7d22a31dL152) from the booster options.  
  - It was [added ](https://github.com/dotnet/machinelearning/pull/2948/files#diff-00ba5d9f5336b3d3d1f011472d6c6e4bR111) to binary trainer options.  
  - But  was not added to the multiclass  trainer.  As such  there is currently no way for  the user to  use  "multiclassova" with UnbalancedSets. 
  - this seems to be the root cause of the issue the customer reports in the bug description

**Proposal**:  Add UnbalancedSets option to lightgbm multiclass trainer 
- this would probably be similar to how it was [added ](https://github.com/dotnet/machinelearning/pull/2948/files#diff-00ba5d9f5336b3d3d1f011472d6c6e4bR111) to the binary trainer
- this would be an incremental change. non-breaking
- estimate : 2 days

PR #2948, opened 2019-03-13
Updating LightGBM Arguments

singlis:
This PR moves the LightGBM options from an all-in-one class to individual options for the binary, multiclass, regression and ranker trainers. 
- Options code that used to live in LightGbmArguments has moved to LightGbmBaseTrainer. Each trainer now has their own respective options class. 
- The booster options have changed a bit to hide the IBoosterFactory interface #2559

PR #3925, opened 2019-06-27
LightGBM Unbalanced Data Argument [Issue #3688 Fix]

RadicalRayan:
Fix for issue #3688. LightGBM Multiclass Trainer can now accept unbalanced data parameter as was previously possible in the Binary Trainer. An additional argument was added to the LightGBMBinaryEstimator test.


src/Microsoft.ML.LightGbm/LightGbmArguments.cs
```diff
@@ -58,6 +59,11 @@ public BoosterParameterBase(OptionsBase options)
         public abstract class OptionsBase : IBoosterParameterFactory
         {
             internal BoosterParameterBase GetBooster() { return null; }
+            /// <summary>
+            /// Whether training data is unbalanced. Used by <see cref="LightGbmBinaryTrainer"/>.
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Use for binary classification when training data is not balanced.", ShortName = "us")]
+            public bool UnbalancedSets = false;
```
wschin: We need two tests for `UnbalancedSets=true` and `UnbalancedSets=false`. In addition to simply run a model, we need more restrictive comparison like
```c#
        [LightGBMFact]
        public void LightGbmMulticlassEstimatorCompareOva()
        {
            float sigmoidScale = 0.5f; // Constant used train LightGBM. See gbmParams["sigmoid"] in the helper function.

            // Train ML.NET LightGBM and native LightGBM and apply the trained models to the training set.
            LightGbmHelper(useSoftmax: false, sigmoid: sigmoidScale, out string modelString, out List<GbmExample> mlnetPredictions, out double[] nativeResult1, out double[] nativeResult0);

            // The i-th predictor returned by LightGBM produces the raw score, denoted by z_i, of the i-th class.
            // Assume that we have n classes in total. The i-th class probability can be computed via
            // p_i = sigmoid(sigmoidScale * z_i) / (sigmoid(sigmoidScale * z_1) + ... + sigmoid(sigmoidScale * z_n)).
            Assert.True(modelString != null);
            // Compare native LightGBM's and ML.NET's LightGBM results example by example
            for (int i = 0; i < _rowNumber; ++i)
            {
                double sum = 0;
                for (int j = 0; j < _classNumber; ++j)
                {
                    Assert.Equal(nativeResult0[j + i * _classNumber], mlnetPredictions[i].Score[j], 6);
                    if (float.IsNaN((float)nativeResult1[j + i * _classNumber]))
                        continue;
                    sum += MathUtils.SigmoidSlow(sigmoidScale * (float)nativeResult1[j + i * _classNumber]);
                }
                for (int j = 0; j < _classNumber; ++j)
                {
                    double prob = MathUtils.SigmoidSlow(sigmoidScale * (float)nativeResult1[j + i * _classNumber]);
                    Assert.Equal(prob / sum, mlnetPredictions[i].Score[j], 6);
                }
            }

            Done();
        }
```
in `machinelearning2\test\Microsoft.ML.Tests\TrainerEstimators\TreeEstimators.cs`. Note that this test directly checks if ML.NET's LightGBM module and LightGBM's C API produce identical numbers.

RadicalRayan: Thank you. I will add and run these tests

RadicalRayan:
Needed to add argument to LightGbmMulticlassTrainer. Added new test for unbalanced binary tree, added two tests for unbalanced/balanced multiclass, added a test to compare multiclass unbalanced .NET with C. Revision to manifest file.

src/Microsoft.ML.LightGbm/LightGbmMulticlassTrainer.cs
```diff
@@ -75,6 +75,12 @@ public enum EvaluateMetricType
                 LogLoss,
             }
 
+            /// <summary>
+            /// Whether training data is unbalanced.
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Use for binary classification when training data is not balanced.", ShortName = "us")]
```
eerhardt: > Use for binary classification when training data is not balanced.

It's odd to say "use for binary classification" when this is the multiclass trainer.

src/Microsoft.ML.LightGbm/LightGbmArguments.cs
```diff
@@ -58,6 +59,11 @@ public BoosterParameterBase(OptionsBase options)
         public abstract class OptionsBase : IBoosterParameterFactory
         {
             internal BoosterParameterBase GetBooster() { return null; }
+            /// <summary>
+            /// Whether training data is unbalanced. Used by <see cref="LightGbmBinaryTrainer"/>.
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Use for binary classification when training data is not balanced.", ShortName = "us")]
```
eerhardt: > Use for binary classification when training data is not balanced.

Is putting this option on the `OptionsBase` class correct if it is only for binary classification?

src/Microsoft.ML.LightGbm/LightGbmArguments.cs
eerhardt: Where is this boolean being used? I don't see any code added using it.

src/Microsoft.ML.LightGbm/LightGbmArguments.cs
RadicalRayan: You're right. The cause of the problem was this [commit](https://github.com/dotnet/machinelearning/pull/2948/files#diff-8173688cd56a37cbcc92012f7d22a31dL152) which split the trainers in to their classes (binary and multiclass). is it better not to include it as an argument in this file?

src/Microsoft.ML.LightGbm/LightGbmArguments.cs
RadicalRayan: It's added to the dictionary, NameMapping on line 39:
{nameof(OptionsBase.UnbalancedSets),                 "is_unbalance"},

src/Microsoft.ML.LightGbm/LightGbmMulticlassTrainer.cs
RadicalRayan: Is it more helpful to just say "Use for multi-class classification when training data is not balanced."

src/Microsoft.ML.LightGbm/LightGbmArguments.cs
eerhardt: You are declaring 2 new boolean fields. I only see 1 being used.

src/Microsoft.ML.LightGbm/LightGbmMulticlassTrainer.cs
```diff
@@ -110,6 +116,7 @@ static Options()
             {
                 var res = base.ToDictionary(host);
 
+                res[GetOptionName(nameof(UnbalancedSets))] = UnbalancedSets;
```
codemzs: >res[GetOptionName(nameof(UnbalancedSets))] = UnbalancedSets; [](start = 16, length = 60)

how is getting mapped to is_unbalance 🔗︎, default = false, type = bool, aliases: unbalance, unbalanced_sets

codemzs: nevermind I see it now:

public static string GetOptionName(string name)
        {
            // Otherwise convert the name to the light gbm argument
            StringBuilder strBuf = new StringBuilder();
            bool first = true;
            foreach (char c in name)
            {
                if (char.IsUpper(c))
                {
                    if (first)
                        first = false;
                    else
                        strBuf.Append('_');
                    strBuf.Append(char.ToLower(c));
                }
                else
                    strBuf.Append(c);
            }
            return strBuf.ToString();
        }

---
In reply to: [299023864](https://github.com/dotnet/machinelearning/pull/3925#discussion_r299023864) [](ancestors = 299023864)

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
```diff
@@ -372,7 +432,9 @@ private void LightGbmHelper(bool useSoftmax, double sigmoid, out string modelStr
                     MinimumExampleCountPerGroup = 1,
                     MinimumExampleCountPerLeaf = 1,
                     UseSoftmax = useSoftmax,
-                    Sigmoid = sigmoid // Custom sigmoid value.
+                    Sigmoid = sigmoid, // Custom sigmoid value.
+                    UnbalancedSets = unbalancedSets // false by default
+
```
wschin: Empty line.

PR #3936, opened 2019-07-01
Internalize tensorflow API and sanity check APIs from Microsoft.ML.Tensorflow nuget

### LINKED ISSUE #3863: Sanity check APIs from Microsoft.ML.Tensorflow nuget
We plan to GA Microsoft.ML.Tensorflow nuget in 1.2 release. Lets do a final pass on the binary and make sure we are only exposing relevant API.

codemzs:
fixes #3863

### DIFF
--- a/src/Microsoft.ML.TensorFlow/TensorFlowModel.cs
+++ b/src/Microsoft.ML.TensorFlow/TensorFlowModel.cs
@@ -105,7 +105,7 @@ namespace Microsoft.ML.Transforms
         /// <remarks>
         /// The support for retraining is experimental.
         /// </remarks>
-        public TensorFlowEstimator RetrainTensorFlowModel(
+        internal TensorFlowEstimator RetrainTensorFlowModel(
             string[] outputColumnNames,
             string[] inputColumnNames,
             string labelColumnName,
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.TensorFlow/TensorFlowModel.cs
```diff
@@ -105,7 +105,7 @@ public TensorFlowEstimator ScoreTensorFlowModel(string[] outputColumnNames, stri
         /// <remarks>
         /// The support for retraining is experimental.
         /// </remarks>
-        public TensorFlowEstimator RetrainTensorFlowModel(
+        internal TensorFlowEstimator RetrainTensorFlowModel(
```
artidoro: That was a pretty useful feature to have: being able to retrain existing graphs. Why do you want to remove it?

codemzs: As per @eerhardt "TensorFlowModel contains the most important method of the component - ScoreTensorFlowModel. However, it also contains a RetrainTensorFlowModel whose documentation say it is "experimental:...I think it is confusing to have an "experimental" API on the same class as a stable API."

We can always add it back and we are switching to TF.NET where we will enable this functionality anyways. It is easier to add things than to add then remove and modify.

artidoro: Ok makes sense. Maybe we could either:

1. Write an issue about it.
2. Add it as an extension to a preview API nuget (such as the experimental nuget)

codemzs: sounds good, i'll open an issue.

PR #3785, opened 2019-05-29
Add AVX and FMA intrinsics in Factorization Machine

pkumar07:
Added AVX and FMA C++ intrinsics in factorizationmachinenative.dll which currently implements C++ SSE code as suggested in #3000.

Issue #3000, opened 2019-03-18
Perf: Optimization for FactorizationMachine


glebuk:
Benchmarking using VTune has found several bottlenecks in Factrization Machine training algorithm.

- Some phases training that are not parallelized. Consider adding parallel computation.
- Evaluate using AVX/AVX2 (C++ or C# instrinsics) in factorizationmachinenative.dll which currently implements C++ SSE code
- Consider optimizing the following hotspot

Function | Module | CPU Time
-- | -- | --
Microsoft::ML::Internal::Utilities::DoubleParser::TryParseCore | Microsoft.ML.Core.dll | 27.390s
CalculateGradientAndUpdateNative | factorizationmachinenative.dll | 22.609s
HelperImpl::FetchNextField | Microsoft.ML.Data.dll | 13.826s
CalculateIntermediateVariablesNative | factorizationmachinenative.dll | 12.201s
Microsoft::ML::Internal::Utilities::DoubleParser::TryParse | Microsoft.ML.Core.dll | 9.666s

frank-dong-ms:
Done in below PR, close the issue now:
https://github.com/dotnet/machinelearning/pull/3940

PR #3940, opened 2019-07-01
Add AVX and FMA intrinsics in Factorization Machine

pkumar07:
Added AVX and FMA C# intrinsics in Microsoft.ML.StandardTrainers for Factorization Machine algorithm  which currently implements C++ SSE code as suggested in #3000 and #3785

@wschin @eerhardt


pkumar07:
@dotnet-bot test this please

hanblee:
@dotnet-bot help

hanblee:
@dotnet-bot test this please

hanblee:
@wschin @eerhardt Would you please help trigger the tests again? They seem to have failed for reasons unrelated to the changes in this PR.

eerhardt:
> Would you please help trigger the tests again?

I have retried the failing tests. If you want to do this, in Azure Pipelines, sign in, and then at the top right is a `...` button - drop that down and click "Retry".

src/Microsoft.ML.StandardTrainers/Microsoft.ML.StandardTrainers.csproj
```diff
@@ -1,7 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework Condition="'$(UseIntrinsics)' != 'true'">netstandard2.0</TargetFramework>
+    <TargetFrameworks Condition="'$(UseIntrinsics)' == 'true'">netstandard2.0;netcoreapp3.0</TargetFrameworks>
```
eerhardt: This type of change isn't going to work for the same reasons as I outlined here: [https://github.com/dotnet/machinelearning/pull/3785#discussion_r289088948](https://github.com/dotnet/machinelearning/pull/3785#discussion_r289088948).

Instead, the intrinsics classes should be moved to `Microsoft.ML.CpuMath`, which is set up to multi-target for `netstandard2.0` and `netcoreapp3.0`.

src/Microsoft.ML.StandardTrainers/FactorizationMachine/FactorizationMachineInterface.netcoreapp.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Security;
+using Microsoft.ML.Internal.CpuMath;
+using Microsoft.ML.Runtime;
+
+namespace Microsoft.ML.Trainers
+{
+    internal static unsafe class FieldAwareFactorizationMachineInterface
+    {
+        internal const string NativePath = "FactorizationMachineNative";
+        public const int CbAlign = 16;
```
eerhardt: We shouldn't be duplicating all this code. Instead, the pattern is to have 3 files:

1. a `netcoreapp.cs` file with the code that only applies to .NET Core.
2. a `netstandard.cs` file with the code that only applies to .NET Standard.
3. A `FactorizationMachine.cs` file that contains the common code between them.

Then the class is marked `partial`, so multiple files can add members to the class.

src/Microsoft.ML.StandardTrainers/Microsoft.ML.StandardTrainers.csproj
pkumar07: > > Would you please help trigger the tests again?
> 
> I have retried the failing tests. If you want to do this, in Azure Pipelines, sign in, and then at the top right is a `...` button - drop that down and click "Retry".

Thanks Eric!

pkumar07: > > Would you please help trigger the tests again?
> 
> I have retried the failing tests. If you want to do this, in Azure Pipelines, sign in, and then at the top right is a `...` button - drop that down and click "Retry".

Thanks Eric!

pkumar07: > This type of change isn't going to work for the same reasons as I outlined here: [#3785 (comment)](https://github.com/dotnet/machinelearning/pull/3785#discussion_r289088948).
> 
> Instead, the intrinsics classes should be moved to `Microsoft.ML.CpuMath`, which is set up to multi-target for `netstandard2.0` and `netcoreapp3.0`.

Should I move the intrinsic methods in the existing `AvxIntrisics` class or should I keep them in a separate class in  `Microsoft.ML.CpuMath`?

eerhardt: The separate class(es) you made sound good to me. Keeping the logic separate makes sense.

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
```diff
@@ -0,0 +1,188 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.X86;
+
+namespace Microsoft.ML.Internal.CpuMath
+{
+    internal static class FactorizationMachineAvxIntrinsics
```
eerhardt: It is a best practice to keep the file name and the class name in sync.
Similarly, keeping the folder structure and the namespace in sync as well.

So you should either:

1. Keep the existing folder/file structure, and change the namespace to `Microsoft.ML.Internal.CpuMath.FactorizationMachine` and change this class name to `AvxIntrinsics`.
2. Change the folder/file structure to remove the `FactorizationMachine` folder and rename the file to `FactorizationMachineAvxIntrinsics.cs`.
3. Make some other change where the namespace and folder structures are aligned, and the class names and file names are aligned.

This makes it easier to navigate the folders/files in the repository.

Same for `FmaIntrinsics.cs`.

src/Microsoft.ML.CpuMath/FactorizationMachine/FactorizationMachineInterface.netcoreapp.cs
```diff
@@ -0,0 +1,80 @@
...
+                fixed (float* pv = &latentWeights.Items[0])
+                fixed (float* pq = &latentSum.Items[0])
+                fixed (float* pr = &response)
+                {
+                    if (Fma.IsSupported)
+                        FactorizationMachineFmaIntrinsics.CalculateIntermediateVariables(fieldCount, latentDim, count, pf, pi, px, pw, Ptr(latentWeights, pv), Ptr(latentSum, pq), pr);
+                    else if (Avx.IsSupported)
+                        FactorizationMachineAvxIntrinsics.CalculateIntermediateVariables(fieldCount, latentDim, count, pf, pi, px, pw, Ptr(latentWeights, pv), Ptr(latentSum, pq), pr);
+                    else
+                        CalculateIntermediateVariablesNative(fieldCount, latentDim, count, pf, pi, px, pw, Ptr(latentWeights, pv), Ptr(latentSum, pq), pr);
```
eerhardt: When Fma and Avx are not supported on the current machine, what do you think about not using native code, but instead falling back to a C# implementation?

The advantage of this is that we can support other architectures (like ARM) without having to compile directly to ARM code.

src/Microsoft.ML.CpuMath/FactorizationMachine/FactorizationMachineInterface.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System.Runtime.InteropServices;
+using System.Security;
+using Microsoft.ML.Internal.CpuMath.Core;
+
+namespace Microsoft.ML.Internal.CpuMath
+{
+    internal static unsafe partial class FieldAwareFactorizationMachineInterface
+    {
+        internal const string NativePath = "FactorizationMachineNative";
```
eerhardt: Unfortunately, now that we've moved this code into CpuMath, this dependency is now backwards.

The `FactorizationMachineNative` assembly ships in the `Microsoft.ML` nuget package, which has a dependency on the `Microsoft.ML.CpuMath` nuget pacakge. However, this is now creating a runtime dependency from CpuMath back to an assembly shipping in the `Microsoft.ML` nuget package.

I think the simplest fix for this would be to remove the `FactorizationMachineNative` native assembly all together, and put its 2 functions into the CpuMathNative assembly, which ships in this same nuget package.

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
pkumar07: I will keep the existing file structure and change the namespace to `Microsoft.ML.Internal.CpuMath.FactorizationMachine`

src/Microsoft.ML.CpuMath/FactorizationMachine/FactorizationMachineInterface.cs
```diff
@@ -0,0 +1,41 @@
...
+        private static unsafe float* Ptr(AlignedArray a, float* p)
+        {
+            Contracts.AssertValue(a);
+            float* q = p + a.GetBase((long)p);
+            Contracts.Assert(((long)q & (CbAlign - 1)) == 0);
+            return q;
+        }
+
+        [DllImport(NativePath), SuppressUnmanagedCodeSecurity]
+        public static extern void CalculateIntermediateVariablesNative(int fieldCount, int latentDim, int count, int* /*const*/ fieldIndices, int* /*const*/ featureIndices,
```
eerhardt: These two `DllImport` methods should be marked `private`. No one outside of this class should be able to call them.

src/Microsoft.ML.CpuMath/FactorizationMachine/FactorizationMachineInterface.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System.Runtime.InteropServices;
+using System.Security;
+using Microsoft.ML.Internal.CpuMath.Core;
+
+namespace Microsoft.ML.Internal.CpuMath.FactorizationMachine
+{
+    internal static unsafe partial class FieldAwareFactorizationMachineInterface
+    {
+        internal const string NativePath = "CpuMathNative";
```
eerhardt: These two `const` variables can be made `private`.

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
```diff
@@ -0,0 +1,188 @@
...
+                        Vector256<float> qfprimefBuffer = Avx.LoadVector256(qfprimef + k);
+                        y = Avx.Add(y, Avx.Multiply(qffprimeBuffer, qfprimefBuffer));
+                    }
+                }
+            }
+
+            y = Avx.Add(y, Avx.Multiply(Vector256.Create(0.5f), tmp));
+            tmp = Avx.Add(y, Avx.Permute2x128(y, y, 1));
+            tmp = Avx.HorizontalAdd(tmp, tmp);
+            y = Avx.HorizontalAdd(tmp, tmp);
```
eerhardt: Why the difference here from the C++ code?

In C++ we have:

```C++
_tmp = _mm_add_ps(_y, _mm_movehl_ps(_y, _y));
_y = _mm_add_ps(_tmp, _mm_shuffle_ps(_tmp, _tmp, 1)); // The lowest slot is the response value.
```

But here, we have:

```C#
tmp = Avx.Add(y, Avx.Permute2x128(y, y, 1));
tmp = Avx.HorizontalAdd(tmp, tmp);
y = Avx.HorizontalAdd(tmp, tmp);
```

Every other line in this method reads line-for-line identical to the C++ code (besides the SSE vs AVX differences).

src/Microsoft.ML.CpuMath/FactorizationMachine/FmaIntrinsics.cs
```diff
@@ -0,0 +1,188 @@
...
+                            gLatent = Fma.MultiplyAdd(sx, q, gLatent);
+                        else
+                            gLatent = Fma.MultiplyAdd(sx, Fma.MultiplyAddNegated(v, x, q), gLatent);
+                        gLatent = Avx.Multiply(wei, gLatent);
+
+                        // Accumulate the gradient of latent vectors.
+                        Vector256<float> h = Fma.MultiplyAdd(gLatent, gLatent, Avx.LoadVector256(hvjfprime + k));
+
+                        // Perform ADAGRAD update rule to adjust latent vector.
+                        v = Fma.MultiplyAddNegated(lr, Avx.Multiply(Avx.ReciprocalSqrt(h), gLatent), v);
```
eerhardt: For my knowledge (I'm not a SIMD expert), what is the difference between `MultiplyAddNegated` and `MultiplySubtract`?

src/Microsoft.ML.CpuMath/FactorizationMachine/FmaIntrinsics.cs
```diff
@@ -0,0 +1,188 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.X86;
+
+namespace Microsoft.ML.Internal.CpuMath.FactorizationMachine
+{
+    internal static class FmaIntrinsics
```
eerhardt: Since this class is nearly identical to the `AvxIntrinsics` class, would it make more sense to collapse these two classes together, and then just interleave `if (Fma.IsSupported)` inline in the methods when you want to switch use an `Fma` intrinsic?

That way there isn't so much duplicated code.

We already follow that pattern here: [https://github.com/dotnet/machinelearning/blob/fabdabf867dc68481d519740a96b981bbeb63180/src/Microsoft.ML.CpuMath/AvxIntrinsics.cs#L159-L171](https://github.com/dotnet/machinelearning/blob/fabdabf867dc68481d519740a96b981bbeb63180/src/Microsoft.ML.CpuMath/AvxIntrinsics.cs#L159-L171)

eerhardt:
Adding @tannergooding to give a 2nd pair of eyes on the Fma and Avx implementation.

pkumar07:
> This is really good work, @pkumar07! Thank you so much for it.
> 
> Do you happen to have any performance results to share for this change? I'd be very interested in seeing how switching between FMA, AVX, and the C++'s SSE implementations affect perf.

Thanks @eerhardt! I did some performance tests and I see 14x performance gain over Baseline(SSE) for AVX and 15x gain for FMA.

Summary:
OS=Windows 10.0.17763.55
Processor=Intel Core i7 8700K CPU 3.70GHz (Coffee Lake), 12 logical and 6 physical cores
.NET Core SDK=3.0.100-preview5-011568
Baseline: ML.NET Release 1.0

Command: .\dotnet.exe ..\Microsoft.ML.Console\netcoreapp3.0\MML.dll train data=higgs.idv loader=BinaryLoader{} trainer=ffm{} out=model_FM.zip
[Dataset](https://archive.ics.uci.edu/ml/datasets/HIGGS)

Baseline (Release 1.0) | AVX | AVX+FMA
-- | -- | --
75.47s | 66.08s | 65.89s

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
pkumar07: There is no AVX equivalent instruction for `_mm_movehl_ps(_y, _y)`.  Permuting and doing horizontal add on the two floating point elements (`tmp`) does the same computation.

src/Microsoft.ML.CpuMath/FactorizationMachine/FactorizationMachineInterface.netcoreapp.cs
pkumar07: I agree, it would be good to have a fallback implementation to support other architectures. I can open an issue for the same.

src/Microsoft.ML.CpuMath/FactorizationMachine/FmaIntrinsics.cs
pkumar07: The main difference is the order in which the vectors get multiplied and added/subtracted.

MultiplyAddNegated instruction translates to `__m128 _mm_fnmadd_ps (__m128 a, __m128 b, __m128 c)` instruction. The  result is computed as: -(a*b) + c

whereas `MultiplySubtract` instruction translates to `__m128 _mm_fmsub_ps (__m128 a, __m128 b, __m128 c) `.
The result for this is computed as: (a*b) - c

[More info](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=3847,4166,2697,2709&techs=FMA&text=%2520_mm_fnmadd_ps)

src/Microsoft.ML.CpuMath/FactorizationMachine/FmaIntrinsics.cs
pkumar07: I agree. We can interleave in the methods to avoid code duplication. I separated the implementation into two classes to make it more readable. I will update the code to reflect these changes.

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
```diff
@@ -0,0 +1,188 @@
...
+
+namespace Microsoft.ML.Internal.CpuMath.FactorizationMachine
+{
+    internal static class AvxIntrinsics
+    {
+        // This function implements Algorithm 1 in https://github.com/wschin/fast-ffm/blob/master/fast-ffm.pdf.
+        // Compute the output value of the field-aware factorization, as the sum of the linear part and the latent part.
+        // The linear part is the inner product of linearWeights and featureValues.
+        // The latent part is the sum of all intra-field interactions in one field f, for all fields possible
+        public static unsafe void CalculateIntermediateVariables(int fieldCount, int latentDim, int count, int* fieldIndices, int* featureIndices, float* featureValues,
```
tannergooding: Is there any perf concern around the number of parameters being passed here?

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
```diff
@@ -0,0 +1,188 @@
...
+            float* linearWeights, float* latentWeights, float* latentSum, float* response)
+        {
```
tannergooding: Might be nice to assert that AVX is supported, given the function is dependent on it.

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
```diff
@@ -0,0 +1,188 @@
...
+                    float* vjfprime = pv + vBias;
+                    float* qffprime = pq + qBias;
+
+                    // q_f,f' += v_j,f' * x
+                    for (int k = 0; k + 8 <= d; k += 8)
+                    {
+                        Vector256<float> vjfprimeBuffer = Avx.LoadVector256(vjfprime + k);
+                        Vector256<float> q = Avx.LoadVector256(qffprime + k);
+                        q = Avx.Add(q, Avx.Multiply(vjfprimeBuffer, x));
+                        Avx.Store(qffprime + k, q);
```
tannergooding: What's the codegen for this sequence... I would hope it folds both loads, but I've seen some codegen here sometimes.

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
```diff
@@ -0,0 +1,188 @@
...
+                    {
+                        // Inter-field interaction.
+                        Vector256<float> qffprimeBuffer = Avx.LoadVector256(qffprime + k);
+                        Vector256<float> qfprimefBuffer = Avx.LoadVector256(qfprimef + k);
+                        y = Avx.Add(y, Avx.Multiply(qffprimeBuffer, qfprimefBuffer));
+                    }
+                }
+            }
+
+            y = Avx.Add(y, Avx.Multiply(Vector256.Create(0.5f), tmp));
```
tannergooding: The `Vector256.Create(0.5f)` isn't optimized right now and this will broadcast each time. Might be better to manually hoist it to an readonly static?

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
```diff
@@ -0,0 +1,188 @@
...
+                    for (int k = 0; k + 8 <= d; k += 8)
+                    {
+                        Vector256<float> v = Avx.LoadVector256(vjfprime + k);
+                        Vector256<float> q = Avx.LoadVector256(qfprimef + k);
+
+                        // Calculate L2-norm regularization's gradient.
+                        Vector256<float> gLatent = Avx.Multiply(lambdav, v);
+
+                        // Calculate loss function's gradient.
+                        if (fprime != f)
```
tannergooding: is this better than creating a temporary to hold the multiplier and then doing `gLatent = Avx.Add(gLatent, Avx.Multiply(sx, tmp))` outside the if/else?

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
pkumar07: Codegen: 

```
movsxd rbx, eax
vmovups ymm3, ymmword ptr [r8+rbx*4]
movsxd rbx, eax
vmovups ymm4, ymmword ptr [r10+rbx*4]
vmulps ymm3, ymm3, ymm2
vaddps ymm4, ymm4, ymm3
movsxd rbx, eax
vmovups ymmword ptr [r10+rbx*4], ymm4
```

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
pkumar07: Can you elaborate more on this? For now, I have kept the parameters identical to the C++ code.

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
pkumar07: I tried this as suggested and I see a similar performance.

src/Microsoft.ML.CpuMath/FactorizationMachine/FactorizationMachineInterface.cs
```diff
@@ -0,0 +1,35 @@
+using System.Runtime.InteropServices;
```
eerhardt: All new code files need a copyright header. You can copy one from an existing .cs file in the repo.

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
tannergooding: C# isn't always as good about inlining as C/C+and there are more parameters being passed than will stay in registers (even for `fastcall`). Was wondering if that was a concern since these most of these will end up shadow copied to the stack and be double-indirections.

tannergooding: if it is, and if inlining isn't a valid option, reordering the parameters so the pointers are first would ensure they are the ones passed in register (keeping them single indirections).

pkumar07:
> This is really great! Thanks for the work, @pkumar07.

Thanks @eerhardt!

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
pkumar07: There is a very slight difference in the codegen. Changing the order of parameters reduces one instruction. I can reorder the parameters in the code.

pkumar07:
@eerhardt I have addressed your comments.

eerhardt:
@codemzs @wschin @artidoro - this change still needs another reviewer.

pkumar07:
@codemzs @wschin @artidoro Is there any update on this PR?

eerhardt:
@codemzs @wschin @artidoro - this change still needs another reviewer.

eerhardt:
@pkumar07 - can you sync up this PR to the latest code. Once it is synced up, and green, I will merge it into master.

pkumar07:
@eerhardt Could you please trigger the failing tests? I don't have access to Azure Pipelines to Retry the tests.

codemzs:
@pkumar07 done.

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
pkumar07: > What's the codegen for this sequence... I would hope it folds both loads, but I've seen some codegen here sometimes.

@tannergooding 
Should this be created as an issue in coreclr?

pkumar07:
Thanks @codemzs! :)

src/Microsoft.ML.CpuMath/FactorizationMachine/AvxIntrinsics.cs
tannergooding: I already logged one for it a while back: <https://github.com/dotnet/coreclr/issues/25008>

PR #3983, opened 2019-07-10
TF package size fix

harishsk:
The Tensorflow tar files contain libtensorflow.so, libtensorflow.so.$(MajorVersion) and libtensorflow.so.$(Version). Of these, the first two are symlinked to the third. When these files get copied over to the nuget package, they are copied as files and not as symlinks which causes the package size to almost triple. 
This fix reduces the copied files to only libtensorflow.so and libtensorflow_framework.so.$(MajorVersion).


src/Redist/Microsoft.ML.TensorFlow.Redist/Microsoft.ML.TensorFlow.Redist.proj
```diff
@@ -9,20 +9,12 @@
     <TensorFlowConfig Include="windows" FileExtension=".zip" FilesFromArchive="lib\tensorflow.dll;include\tensorflow\c\LICENSE" Runtime="win-x64"/>
     <TensorFlowConfig Include="linux" FileExtension=".tar.gz"
                       FilesFromArchive="lib\libtensorflow.so;
-                                        lib\libtensorflow.so.$(TensorFlowMajorVersion);
```
artidoro: Are all these dynamic libraries not necessary then?

src/Redist/Microsoft.ML.TensorFlow.Redist/Microsoft.ML.TensorFlow.Redist.proj
```diff
@@ -9,20 +9,12 @@
     <TensorFlowConfig Include="windows" FileExtension=".zip" FilesFromArchive="lib\tensorflow.dll;include\tensorflow\c\LICENSE" Runtime="win-x64"/>
```
artidoro: nit: could you format the windows config the same way you did below for linux and mac?
Not important if you are not doing another iteration.

src/Redist/Microsoft.ML.TensorFlow.Redist/Microsoft.ML.TensorFlow.Redist.proj
harishsk: The binaries deleted are identical to the ones remaining. They are symlinked in the tar file. But when they get to the nuget package, they become copies. 
This change keeps only the ones that are used.

Issue #3994, opened 2019-07-12
Docs show using VectorType instead of concatenating features


nicolehaugen:
Numerous places in the docs, we show to store features as a VectorType.  However, this isn't ideal because it doesn't allow you to easily do feature engineering where you pick\choose the most influential features to include when training a model.  Instead, to easily support feature engineering, it's recommended to concatenate your features as part of the pipeline. 

For example, here are a few places where we show using a VectorType:
1.) https://docs.microsoft.com/en-us/dotnet/machine-learning/how-to-guides/load-data-ml-net#create-the-data-model
2.) https://github.com/dotnet/machinelearning/blob/master/docs/code/MlNetCookBook.md#how-do-i-load-data-from-a-text-file

Instead, we should show feature concatenation and explain why this is a preferred approach - for example:
```
   IEstimator<ITransformer> dataPipeline = mlContext.Transforms.Concatenate(FeaturesVectorName, featureCols)
                .Append(mlContext.Transforms.Conversion.MapValueToKey(nameof(SearchResultData.Label)))
                .Append(mlContext.Transforms.Conversion.Hash(nameof(SearchResultData.GroupId), nameof(SearchResultData.GroupId), numberOfBits: 20));
```
Also, why is there a VectorType attribute?  Are there ever benefits to using this?  If not, we should consider removing.

codemzs:
Hi @nicolehaugen can you please point out the specific docs you are referring? VectorType is used to indicate the size of the vector at compile time.

nicolehaugen:
Hi @codemzs  -refer to the links that I provided above.  The first one shows use of the VectorType attribute, the second one shows loading features from a text file into a vector column.  Also, [this link](https://github.com/dotnet/machinelearning/blob/master/docs/code/MlNetCookBook.md#how-do-i-load-data-with-many-columns-from-a-csv) specifically says:

When the input file contains many columns of the same type, always intended to be used together, we recommend loading them as a vector column from the very start: this way the schema of the data is cleaner, and we don't incur unnecessary performance costs.

These docs led me to believe that I should be loading my features as a vector column as part of my data structure.  However, later I learned that this really isn't the preferred approach since it doesn't allow the user to easily apply feature engineering by including\excluding different columns within the selected features used for training.  It would be helpful to have the docs make this point clear by stating when it's beneficial to load features as a single vector column vs. loading the features as separate properties (and then doing a Concatenate) prior to training.

codemzs:
[...]

PR #4002, opened 2019-07-15
fixes #3992,  when inputColumnNames is not provided, it is set to new[] { outputColumnName }. and part of #3994(making SlotDroppingTransfomer exposed as public)

### LINKED ISSUE #3992: FeaturizeText should allow only outputColumnName to be defined
One of the extension methods for `FeaturizeText` needs both the `outputColumnName` and the `inputColumnNames` to be provided.

There is no compile error if `inputColumnNames` is not provided, only a runtime error.

We should fix the error so that when `inputColumnNames` is not provided, it is set to `new[] { outputColumnName }` as we do everywhere else in the code base.

https://github.com/dotnet/machinelearning/blob/c3bdaaa2a29f88a85dd91cde3fbb805001011903/src/Microsoft.ML.Transforms/Text/TextCatalog.cs#L60-L65

wangyems:
fixes #3992 and part of #3994(making SlotDroppingTransfomer exposed as public)

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/SlotsDroppingTransformer.cs
+++ b/src/Microsoft.ML.Data/Transforms/SlotsDroppingTransformer.cs
@@ -260,7 +260,7 @@ namespace Microsoft.ML.Transforms
         /// <param name="inputColumnName">Name of column to transform. If set to <see langword="null"/>, the value of the <paramref name="outputColumnName"/> will be used as source.</param>
         /// <param name="min">Specifies the lower bound of the range of slots to be dropped. The lower bound is inclusive. </param>
         /// <param name="max">Specifies the upper bound of the range of slots to be dropped. The upper bound is exclusive.</param>
-        internal SlotsDroppingTransformer(IHostEnvironment env, string outputColumnName, string inputColumnName = null, int min = default, int? max = null)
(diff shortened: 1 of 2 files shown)


harishsk:
This pull request contains fixes for two different bugs. Ideally you should create a separate pull request for each bug. But since you have fixed both bugs in the same branch (master) they will get accumulated into the same PR. 

The best way to handle this is by creating a separate branch for each bug and submitting a PR from that branch. That way, the review for each bug can be handled separately and the fixes merged separately. 
 
Would you mind abandoning this PR and creating a new branch and PR for each bug. (It is best not to submit PRs from master)

wangyems:
Sure!

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

codemzs:
@wangyems You will need to create an estimator API for slot dropper transform, document it, add it to its catalog, add tests and samples. Please see examples for estimator APIs such as  [KeyToVectorMappingTransformer](https://github.com/dotnet/machinelearning/blob/c3bdaaa2a29f88a85dd91cde3fbb805001011903/src/Microsoft.ML.Transforms/KeyToVectorMapping.cs#L445)

I will recommend it be its own PR.

Issue #3965, opened 2019-07-03
LightGbm Warning after upgrading to LightGBM v1.2 ([LightGBM] [Warning] Unknown parameter metric=)


bartczernicki:
### System information

- **OS version/distro**: Windows Server 2016
- **.NET Version (eg., dotnet --info)**: .NET Core 2.1

### Issue

- **What did you do?** Updated to ML.NET/LightGBM v1.2
- **What happened?** Warning started appearing.
- **What did you expect?** No warning

### Source code / logs

Source code is located here:
https://github.com/bartczernicki/MLDotNet-BaseballClassification

Fit method that throws the warning...
https://github.com/bartczernicki/MLDotNet-BaseballClassification/blob/28143bb71d962bcb919393db258e7e458480f071/MLDotNet-BaseballClassification/Program.cs#L113

_Error: [LightGBM] [Warning] Unknown parameter metric=_

codemzs:
@najeeb-kazmi this seems related to your metric change. Can you please take a look?

PR #4007, opened 2019-07-15
Stop LightGbm Warning for Default Metric Input [Issue #3965 Fix]

RadicalRayan:
Issue #3965 reported that a warning, "LightGBM] [Warning] Unknown parameter metric=" is produced when the default metric is used. This warning came after this [commit](https://github.com/dotnet/machinelearning/pull/3859) which aimed to provide a consistent user experience from an ML.NET implementation of LightGbm with standalone LightGbm. If a user were to set `EvaluationMetric = EvaluateMetricType.Default`, they might expect that this would set the EvaluationMetric to "" and assigned the metric based on the objective as shown in the LightGbm docs. When the correction was made, this warning began to appear when the metric parameter was set to "". which was also being produced in LightGbm alone. The only way to prevent this error would be to not assign a parameter to the metric at all.

This warning has not appeared in previous versions of ML.NET and can be prevent by assigning the correct metric based on the objective as was previously done.

To prevent this warning, the changes from this [commit](https://github.com/dotnet/machinelearning/pull/3859) were reverted.

### DIFF
--- a/src/Microsoft.ML.LightGbm/LightGbmBinaryTrainer.cs
+++ b/src/Microsoft.ML.LightGbm/LightGbmBinaryTrainer.cs
@@ -162,7 +162,7 @@ namespace Microsoft.ML.Trainers.LightGbm
             [Argument(ArgumentType.AtMostOnce,
                 HelpText = "Evaluation metrics.",
                 ShortName = "em")]
-            public EvaluateMetricType EvaluationMetric = EvaluateMetricType.Default;
+            public EvaluateMetricType EvaluationMetric = EvaluateMetricType.Logloss;
 
             static Options()
             {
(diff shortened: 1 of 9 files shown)


najeeb-kazmi:
If the only reason we are reverting the defaults is this warning that is not affecting training, we should weigh the two options:

1. Keep the defaults at the expense of having this warning, which might cause confusion for the user.

2. Revert defaults and avoid the warning at the expense of having inconsistent user experience compared to standalone LightGbm as described in #3822 and #3859, although this will also cause confusion for the user.

I have a preference for (1) from the point of view of a LightGbm user coming to ML.NET and finding that the defaults are different from what they are used to. That this warning is being printed when the correct LightGbm default i.e. "" is being passed to LightGbm seems to be an issue that should be fixed on the LightGbm side.

RadicalRayan:
> 
> 
> If the only reason we are reverting the defaults is this warning that is not affecting training, we should weigh the two options:
> 
>     1. Keep the defaults at the expense of having this warning, which might cause confusion for the user.
> 
>     2. Revert defaults and avoid the warning at the expense of having inconsistent user experience compared to standalone LightGbm as described in #3822 and #3859, although this will also cause confusion for the user.
> 
> 
> I have a preference for (1) from the point of view of a LightGbm user coming to ML.NET and finding that the defaults are different from what they are used to. That this warning is being printed when the correct LightGbm default i.e. "" is being passed to LightGbm seems to be an issue that should be fixed on the LightGbm side.

PR #4013, opened 2019-07-17
OneVersusAllModelParameters Strongly Typed

### LINKED ISSUE #2467: OvaModelParameters is not strongly-typed

michaelgsharp:
Fixes #2467

We used to remove all Type information when we constructed our `OneVersusAllModelParameters`. This prevented the users from access the inner model without run-time casting.

This PR makes the `OneVersusAllModelParameters` strongly typed, and adds a strongly typed version of the `OneVersusAllTrainer` as it uses the `OneVersusAllModelParameters`. This change is no longer a breaking change to the current public api.

### DIFF
--- a/src/Microsoft.ML.Core/Prediction/IPredictor.cs
+++ b/src/Microsoft.ML.Core/Prediction/IPredictor.cs
@@ -12,8 +12,7 @@ namespace Microsoft.ML
     /// <see cref="ITrainer"/> and <see cref="IPredictor"/> it is still useful, but for things based on
     /// the <see cref="IEstimator{TTransformer}"/> idiom, it is inappropriate.
     /// </summary>
-    [BestFriend]
-    internal enum PredictionKind
+    public enum PredictionKind
     {
         Unknown = 0,
         Custom = 1,
@@ -34,8 +33,7 @@ namespace Microsoft.ML
     /// <summary>
     /// Weakly typed version of IPredictor.
     /// </summary>
-    [BestFriend]
-    internal interface IPredictor
+    public interface IPredictor
     {
         /// <summary>
         /// Return the type of prediction task.
@@ -47,8 +45,7 @@ namespace Microsoft.ML
     /// A predictor the produces values of the indicated type.
     /// REVIEW: Determine whether this is just a temporary shim or long term solution.
     /// </summary>
-    [BestFriend]
-    internal interface IPredictorProducing<out TResult> : IPredictor
+    public interface IPredictorProducing<out TResult> : IPredictor
     {
     }
 
--- a/src/Microsoft.ML.Core/Prediction/IPredictor.cs
+++ b/src/Microsoft.ML.Core/Prediction/IPredictor.cs
@@ -33,7 +33,8 @@ namespace Microsoft.ML
     /// <summary>
     /// Weakly typed version of IPredictor.
     /// </summary>
-    public interface IPredictor
+    [BestFriend]
+    internal interface IPredictor
     {
         /// <summary>
         /// Return the type of prediction task.
@@ -45,7 +46,8 @@ namespace Microsoft.ML
     /// A predictor the produces values of the indicated type.
     /// REVIEW: Determine whether this is just a temporary shim or long term solution.
     /// </summary>
-    public interface IPredictorProducing<out TResult> : IPredictor
+    [BestFriend]
+    internal interface IPredictorProducing<out TResult> : IPredictor
     {
     }
 
--- a/src/Microsoft.ML.Data/Prediction/Calibrator.cs
+++ b/src/Microsoft.ML.Data/Prediction/Calibrator.cs
@@ -989,9 +989,7 @@ namespace Microsoft.ML.Calibrators
 
             var predWithFeatureScores = predictor as IPredictorWithFeatureWeights<float>;
             if (predWithFeatureScores != null && predictor is IParameterMixer<float> && cali is IParameterMixer)
-            {
                 return new ParameterMixingCalibratedModelParameters<TSubPredictor, TCalibrator>(env, predictor, cali);
-            }
 
             if (predictor is IValueMapper)
                 return new ValueMapperCalibratedModelParameters<TSubPredictor, TCalibrator>(env, predictor, cali);
--- a/src/Microsoft.ML.Data/Prediction/Calibrator.cs
+++ b/src/Microsoft.ML.Data/Prediction/Calibrator.cs
@@ -879,10 +879,12 @@ namespace Microsoft.ML.Calibrators
             return CreateCalibratedPredictor(env, (IPredictorProducing<float>)predictor, trainedCalibrator);
         }
 
-        public static CalibratedModelParametersBase<T, ICalibrator> GetCalibratedPredictor<T>(IHostEnvironment env, IChannel ch, ICalibratorTrainer caliTrainer,
-            T predictor, RoleMappedData data, int maxRows = _maxCalibrationExamples) where T: class
+        public static CalibratedModelParametersBase<TSubPredictor, TCalibrator> GetCalibratedPredictor<TSubPredictor, TCalibrator>(IHostEnvironment env, IChannel ch, ICalibratorTrainer caliTrainer,
+            TSubPredictor predictor, RoleMappedData data, int maxRows = _maxCalibrationExamples)
+            where TSubPredictor : class
+            where TCalibrator : class, ICalibrator
         {
-            var trainedCalibrator = TrainCalibrator(env, ch, caliTrainer, (IPredictorProducing<float>)predictor, data, maxRows);
+            var trainedCalibrator = TrainCalibrator(env, ch, caliTrainer, (IPredictorProducing<float>)predictor, data, maxRows) as TCalibrator;
             var cp = CreateCalibratedPredictor(env, predictor, trainedCalibrator);
             return cp;
         }
@@ -990,6 +992,7 @@ namespace Microsoft.ML.Calibrators
             if (predWithFeatureScores != null && predictor is IParameterMixer<float> && cali is IParameterMixer)
             {
                 var s = typeof(TSubPredictor);
+                var d = typeof(TCalibrator);
                 var pm = new ParameterMixingCalibratedModelParameters<TSubPredictor, TCalibrator>(env, predictor, cali);
                 return pm;
             }
--- a/src/Microsoft.ML.Data/Prediction/Calibrator.cs
+++ b/src/Microsoft.ML.Data/Prediction/Calibrator.cs
@@ -528,7 +528,7 @@ namespace Microsoft.ML.Calibrators
         where TSubModel : class
         where TCalibrator : class, ICalibrator
     {
-        private readonly IPredictorWithFeatureWeights<float> _featureWeights;
+        private readonly TSubModel _featureWeights;
 
         internal ParameterMixingCalibratedModelParameters(IHostEnvironment env, TSubModel predictor, TCalibrator calibrator)
             : base(env, RegistrationName, predictor, calibrator)
@@ -536,7 +536,7 @@ namespace Microsoft.ML.Calibrators
             Host.Check(predictor is IParameterMixer<float>, "Predictor does not implement " + nameof(IParameterMixer<float>));
             Host.Check(calibrator is IParameterMixer, "Calibrator does not implement " + nameof(IParameterMixer));
             Host.Assert(predictor is IPredictorWithFeatureWeights<float>);
-            _featureWeights = predictor as IPredictorWithFeatureWeights<float>;
+            _featureWeights = predictor;
         }
 
         internal const string LoaderSignature = "PMixCaliPredExec";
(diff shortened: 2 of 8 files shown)


src/Microsoft.ML.Core/Prediction/IPredictor.cs
```diff
@@ -12,8 +12,7 @@ namespace Microsoft.ML
     /// <see cref="ITrainer"/> and <see cref="IPredictor"/> it is still useful, but for things based on
     /// the <see cref="IEstimator{TTransformer}"/> idiom, it is inappropriate.
     /// </summary>
-    [BestFriend]
-    internal enum PredictionKind
+    public enum PredictionKind
```
eerhardt: These types aren't intended to be public. Can you keep them internal?

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/OneVersusAllTrainer.cs
```diff
@@ -227,37 +228,324 @@ public override MulticlassPredictionTransformer<OneVersusAllModelParameters> Fit
...
+                predictors[i] = (T)TrainOne(ch, Trainer, data, i).Model;
+            }
+            return OneVersusAllModelParametersBuilder.Create<T>(Host, TrainerOptions.UseProbabilities, predictors);
+        }
+    }
+
+    /// <summary>
+    /// Class that holds the static create methods for the <see cref="OneVersusAllModelParameters"/> classes.
+    /// </summary>
+    public class OneVersusAllModelParametersBuilder {
```
eerhardt: This class doesn't need to be public. It doesn't contain any public members.

Also, it can be marked as a `static class`.

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/OneVersusAllTrainer.cs
```diff
@@ -227,37 +228,324 @@ public override MulticlassPredictionTransformer<OneVersusAllModelParameters> Fit
...
+        }
+    }
+
+    /// <summary>
+    /// Strongly typed implementation of the <see cref="OneVersusAllTrainerBase{T}"/> where T is a <see cref="OneVersusAllModelParameters{T}"/>.  T can either be
+    /// a calibrated binary estimator of type <see cref="CalibratedModelParametersBase{TSubPredictor, TCalibrator}"/>, or a non calibrated binary estimary.
+    /// This cannot be used to turn a non calibrated binary classification estimator into its calibrated version. If that is required, use <see cref="OneVersusAllTrainerTyped{TSubPredictor, TCalibrator}"/> instead.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public sealed class OneVersusAllTrainerTyped<T> : OneVersusAllTrainerBase<OneVersusAllModelParametersBase<T>> where T : class
```
eerhardt: I'm not sure having `Typed` at the end of this class name is the best. Why can't it just be `OneVersusAllTrainer<T>`?

michaelgsharp: The problem is that the only difference between the original method and this one are the optional parameters. So if people are only specifying the required parameters, which is the most common case in our code base, then C# doesn't know which method to call. Isn't that a binary breaking change? So I chose to make this method have a different name. Do you have any other ideas on how to resolve it?

michaelgsharp: Same question about the other method I added in. I called it `OneVersusAllUnCalibratedToCalibratedTyped`, but since it has 2 type parameters that have to be manually specified, its name can be anything and it wont have conflicts. I just named it this to be very clear on what it does, though I am happy to change it as well if you have thoughts on this too.

src/Microsoft.ML.Core/Prediction/IPredictor.cs
michaelgsharp: Unfortunately not if we want the `OneVersusAllModelParameters` to be strongly typed and keep the backwards api compatibility. Since the original type was of `IPredictorProducing<float>`, once I type it that way then `IPredictorProducing<T>`, `IPredictor`, and `PredictionKind` have to all become public as well. The only other thought I had was that maybe I could change the type of the backwards compatible `OneVersusAllModelParameters` to be of type `object` instead of `IPredictorProducing<T>`. This will still keep the backwards API compatibility and would allow me to make those classes internal again.

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/OneVersusAllTrainer.cs
michaelgsharp: Good points. I will change those.

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/OneVersusAllTrainer.cs
eerhardt: This isn't a method, this is a class name. The class name and the method don't need to match 1-1.

If we need to suffix the method name with `Typed` or similar, that is fine. But the class doesn't need to have the suffix on it.

michaelgsharp: I see. I was thinking of the Trainers Catalog when I responded. I have fixed the class names and will push those changes. Thanks!

src/Microsoft.ML.Core/Prediction/IPredictor.cs
eerhardt: Another option would be to not have the base class be generic. Have `class OneVersusAllModelParametersBase` and `class OneVersusAllModelParameters : OneVersusAllModelParametersBase` and `class OneVersusAllModelParameters<T> : OneVersusAllModelParametersBase`.

These IPredictor types are not meant to be public. We explicitly made them internal in `1.0` because they shouldn't be exposed to the user. So we will have to find another way here.

michaelgsharp: I thought about that approach initially, but there is enough Strongly typed parameters in there I didn't think it would work. Let me test it and see what I can do there. If not, we can always change the type to `object` since that will work and not change anything about the api.

michaelgsharp: Alright, I have made the change following your suggestions, so now the interfaces are back to being internal. Thanks for your feedback!

src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
```diff
@@ -764,6 +764,91 @@ private static ICalibratorTrainer GetCalibratorTrainerOrThrow(IExceptionContext
...
+        /// <param name="imputeMissingLabelsAsNegative">Whether to treat missing labels as having negative labels, instead of keeping them missing.</param>
+        /// <param name="useProbabilities">Use probabilities (vs. raw outputs) to identify top-score category.</param>
+        /// <typeparam name="TModel">The type of the model. This type parameter will usually be inferred automatically from <paramref name="binaryEstimator"/>.</typeparam>
+        /// <example>
+        /// <format type="text/markdown">
+        /// <![CDATA[
+        ///  [!code-csharp[OneVersusAll](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/MulticlassClassification/OneVersusAll.cs)]
+        /// ]]></format>
+        /// </example>
+        public static OneVersusAllTrainer<TModel> OneVersusAllTyped<TModel>(this MulticlassClassificationCatalog.MulticlassClassificationTrainers catalog,
```
justinormont: Have you tested the new OneVersusAllTyped to see that you can match the scores of the existing one?

michaelgsharp: Yep! I added and modified the tests in the `OvaTest.cs` to show that the scores match exactly.

test/Microsoft.ML.Tests/Scenarios/OvaTest.cs
```diff
@@ -35,14 +36,24 @@ public void OvaLogisticRegression()
...
             // Metrics
             var metrics = mlContext.MulticlassClassification.Evaluate(predictions);
             Assert.True(metrics.MicroAccuracy > 0.94);
+
+            var metricsTyped = mlContext.MulticlassClassification.Evaluate(predictionsTyped);
+            Assert.True(metricsTyped.MicroAccuracy > 0.94);
+
+            Assert.Equal(metrics.MicroAccuracy, metricsTyped.MicroAccuracy);
         }
+                //.Append(ML.MulticlassClassification.Trainers.OneVersusAllUnCalibratedToCalibratedTyped<LinearBinaryModelParameters, PlattCalibrator>(sdcaTrainer))
```
justinormont: Extraneous comment.

src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
justinormont: The test which directly compares the score of the non-typed to the newly created typed version, is the perfect unit test. Thanks.

src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
```diff
@@ -764,6 +764,91 @@ private static ICalibratorTrainer GetCalibratorTrainerOrThrow(IExceptionContext
...
+                throw env.ExceptParam(nameof(binaryEstimator), "Trainer estimator does not appear to produce the right kind of model.");
+            return new OneVersusAllTrainer<TModel>(env, est, labelColumnName, imputeMissingLabelsAsNegative, useProbabilities);
+        }
+
+        /// <summary>
+        /// Create a <see cref="OneVersusAllTrainer{TModelIn, TCalibrator}"/>, which predicts a multiclass target using one-versus-all strategy with
+        /// the binary classification estimator specified by <paramref name="binaryEstimator"/>.This method works with binary classifiers that
+        /// are not calibrated and need to be calibrated before use. Due to the type of estimator changing (from uncalibrated to calibrated), you must manually
+        /// specify both the type of the model and the type of the calibrator. If your classifier is already calibrated or it does not need to be, use the
+        /// <see cref="OneVersusAllTyped{TModel}"/> method instead.
```
justinormont: May want to explicitly state:
If you want to retrieve the model parameters, you will want to use this strongly typed OneVersusAllTrainer.

And perhaps a note on the non-strongly typed which says similar.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -528,15 +528,15 @@ internal sealed class ParameterMixingCalibratedModelParameters<TSubModel, TCalib
...
     {
-        private readonly IPredictorWithFeatureWeights<float> _featureWeights;
+        private readonly TSubModel _featureWeights;
 
         internal ParameterMixingCalibratedModelParameters(IHostEnvironment env, TSubModel predictor, TCalibrator calibrator)
             : base(env, RegistrationName, predictor, calibrator)
         {
             Host.Check(predictor is IParameterMixer<float>, "Predictor does not implement " + nameof(IParameterMixer<float>));
             Host.Check(calibrator is IParameterMixer, "Calibrator does not implement " + nameof(IParameterMixer));
             Host.Assert(predictor is IPredictorWithFeatureWeights<float>);
```
yaeldekel: >Assert [](start = 17, length = 6)

Can this be a check instead of assert?

yaeldekel: On second thought, why does this class need to have a generic type?

---
In reply to: [310376340](https://github.com/dotnet/machinelearning/pull/4013#discussion_r310376340) [](ancestors = 310376340)

michaelgsharp: This class needs to be generic to get the typed predictor and calibrator out of it. Otherwise OVA would just return this and it wouldn't be useful at all as far as getting the model goes. And I didn't do anything with the assert, so I am not sure why they did it that way in the beginning.

yaeldekel: This specific class is internal anyway, so if OVA uses it users will not be able to get the typed calibrator anyway. But I don't think OVA uses this class, does it?

---
In reply to: [310687802](https://github.com/dotnet/machinelearning/pull/4013#discussion_r310687802) [](ancestors = 310687802)

michaelgsharp: So when OVA is returning the sub model parameters, if the model is calibrated it will return an instance of the `CalibratedModelParametersBase`. This class is public, and still allows access to the model parameters and the calibrator used for the training. Any changes made in this file were to facilitate this `CalibratedModelParametersBase` and returning the correct type, while still keeping the derived classes abstract.

src/Microsoft.ML.Core/Prediction/IPredictor.cs
eerhardt: `PredictionKind` should remain internal.

codemzs:
@michaelgsharp Closing this because this is not actively worked on.

PR #4021, opened 2019-07-18
DatabaseLoader specs: Update on NuGet and Class library design

CESARDELATORRE:
Minor update so we are more explicit about the NuGet and Class library design for this feature.


eerhardt:
cc @ebarsoumMS - FYI

eerhardt:
Only changing a `.md` file. The CI failures are just noise.  Merging.

PR #4035, opened 2019-07-23
Adding the initial prototype of a DatabaseLoader

tannergooding:
This adds the initial barebones prototype for DatabaseLoader to the Microsot.ML.Experimental project.


tannergooding:
CC. @eerhardt

src/Microsoft.ML.Experimental/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,10 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+using Microsoft.ML;
+
+[assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Tests" + PublicKey.TestValue)]
```
eerhardt: Instead of doing this, we should change the test to use the public APIs. Basically `new DatabaseLoader` => `mlContext.Data.CreateDatabaseLoader`

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoaderCursor.cs
```diff
@@ -0,0 +1,275 @@
...
+                }
+                else if (typeof(TValue) == typeof(sbyte))
+                {
+                    getterDelegate = CreateSByteGetterDelegate(colInfo);
+                }
+                else if (typeof(TValue) == typeof(float))
+                {
+                    getterDelegate = CreateSingleGetterDelegate(colInfo);
+                }
+                else if (typeof(TValue) == typeof(string))
```
eerhardt: For strings, this type is usually `typeof(ReadOnlyMemory<char>)`. ML.NET uses `ROM<char>` everywhere in order to support sub-strings and not having to allocate each individual string.

We should be sure to add a test for strings to cover this.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
```diff
@@ -0,0 +1,373 @@
...
+using Microsoft.ML.Runtime;
+
+[assembly: LoadableClass(DatabaseLoader.Summary, typeof(DatabaseLoader), null, typeof(SignatureLoadModel),
+    "Database Loader", DatabaseLoader.LoaderSignature)]
+
+namespace Microsoft.ML.Data
+{
+    public sealed partial class DatabaseLoader : IDataLoader<Func<DbDataReader>>
+    {
+        internal const string Summary = "Loads data from an DbDataReader.";
```
eerhardt: `an DbDataReader` should be `a DbDataReader`

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            : base(output)
+        {
+        }
+
+        [Fact]
+        public void TryItOut()
+        {
+            var mlContext = new MLContext(seed: 1);
+
+            var connectionString = @"Server=(localdb)\mssqllocaldb;Database=EFGetStarted.AspNetCore.NewDb;Trusted_Connection=True;ConnectRetryCount=0";
```
eerhardt: We will either need to disable this test, or change it so it doesn't rely on machine state/configuration.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoaderCursor.cs
tannergooding: `DbDataReader` does not support `Span` or `ReadOnlyMemory`.

eerhardt: Correct, but ML.NET's IDataView only supports `ReadOnlyMemory<char>` and not `string`. So `TValue` will be of type `ReadOnlyMemory<char>`, and you will need to set the `ref` to the result from `DbDataReader.GetString()`.

See https://github.com/dotnet/machinelearning/blob/45b6c43387294d19cad843a21e87539213e8d34c/src/Microsoft.ML.DataView/DataViewType.cs#L84-L87

When `IDataView` inspects string values, it uses `ReadOnlyMemory<char>`. So you will need to turn the `string` you get from DbDataReader into a `ReadOnlyMemory<char>` here.

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            using (SqlConnection connection = new SqlConnection(connectionString))
+            {
+                connection.Open();
+
+                using (SqlCommand command = new SqlCommand(
+                    "SELECT SepalLength, SepalWidth, PetalLength, PetalWidth, Label FROM IrisData",
+                    connection))
+                {
+                    DatabaseLoader loader = new DatabaseLoader(mlContext, new DatabaseLoader.Options()
```
CESARDELATORRE: In addition to that way using a Columns array, when possible, can we have another test (and implementation if not ready yet) that is using a convenient data-model class?

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
```diff
@@ -0,0 +1,95 @@
...
+                        {
+                            new DatabaseLoader.Column() { Name = "SepalLength", Type = DbType.Single },
+                            new DatabaseLoader.Column() { Name = "SepalWidth", Type = DbType.Single },
+                            new DatabaseLoader.Column() { Name = "PetalLength", Type = DbType.Single },
+                            new DatabaseLoader.Column() { Name = "PetalWidth", Type = DbType.Single },
+                            new DatabaseLoader.Column() { Name = "Label", Type = DbType.Int32 },
+                        }
+                    });
+
+                    IDataView trainingData = loader.Load(() => command.ExecuteReader());
```
CESARDELATORRE: Once we have this foundational API we also need to add it to the MLContext catalog, so having convenient methods such as:

`mlContext.Data.LoadFromDbDataReader<ModelInputData>(() => command.ExecuteReader());`

or

```
IDataView trainingDataView = mlContext.Data.LoadFromDbTable<ModelInputData, SqlConnection>(connString: myConnString, 
                                                                                                 tableName: "TrainingDataTable"); 
```

CESARDELATORRE:
I put a couple of comments but those asks don't need to be implemented before merging this first version. It could come later. :)

src/Microsoft.ML.Experimental/Properties/AssemblyInfo.cs
tannergooding: Fixed.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoaderCursor.cs
tannergooding: Fixed.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
tannergooding: Fixed.

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
tannergooding: I created a `MockDbConnection`, `MockDbCommand`, and `MockDbDataReader` class which wrap a `TextLoader`. I would guess this needs some cleanup to support more testing scenarios and to match w/e conventions ML.NET is currently using.

test/Microsoft.ML.Tests/Microsoft.ML.Tests.csproj
```diff
@@ -54,5 +54,6 @@
     <PackageReference Include="Microsoft.ML.TensorFlow.TestModels" Version="$(MicrosoftMLTensorFlowTestModelsVersion)" />
     <PackageReference Include="Microsoft.ML.Onnx.TestModels" Version="$(MicrosoftMLOnnxTestModelsVersion)" />
     <PackageReference Include="Microsoft.ML.TestModels" Version="$(MicrosoftMLTestModelsPackageVersion)" />
+    <PackageReference Include="System.Data.SqlClient" Version="4.6.1" />
```
eerhardt: We don't need this anymore, right?

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
```diff
@@ -0,0 +1,544 @@
...
+                new DatabaseLoader.Column() { Name = "SepalWidth", Type = DbType.Single },
+                new DatabaseLoader.Column() { Name = "PetalLength", Type = DbType.Single },
+                new DatabaseLoader.Column() { Name = "PetalWidth", Type = DbType.Single }
+            };
+
+            using (var connection = new MockConnection(mlContext, GetDataPath(TestDatasets.iris.trainFilename), loaderColumns))
+            {
+                connection.Open();
+
+                using (var command = new MockCommand(connection, "SepalLength", "SepalWidth", "PetalLength", "PetalWidth"))
```
eerhardt: You need to read the Label column, right? That's how the trainer is going to know what the expected value should be.

src/Microsoft.ML.Experimental/Properties/AssemblyInfo.cs
eerhardt: Do we still need this?

---
In reply to: [308308500](https://github.com/dotnet/machinelearning/pull/4035#discussion_r308308500) [](ancestors = 308308500)

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoaderCursor.cs
```diff
@@ -0,0 +1,276 @@
...
+            private ValueGetter<ReadOnlyMemory<char>> CreateStringGetterDelegate(ColInfo colInfo)
+            {
+                int columnIndex = GetColumnIndex(colInfo);
+                return (ref ReadOnlyMemory<char> value) => value = _input.GetString(columnIndex).AsMemory();
+            }
+
+            private ValueGetter<ushort> CreateUInt16GetterDelegate(ColInfo colInfo)
+            {
+                int columnIndex = GetColumnIndex(colInfo);
+                return (ref ushort value) => value = (ushort)_input.GetInt16(columnIndex);
```
eerhardt: Are there no such things as unsigned integers in DbDataReader?

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
tannergooding: The `MockCommand` doesn't actually do anything with these column names right now, I should remove them and/or update the mock to use them.

test/Microsoft.ML.Tests/Microsoft.ML.Tests.csproj
tannergooding: Right, will remove.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoaderCursor.cs
tannergooding: Not that I saw. I would guess `DbDataReader` was made back when people cared about `CLSCompliant`, which the unsigned types aren't.

test/Microsoft.ML.Tests/Microsoft.ML.Tests.csproj
tannergooding: Removed.

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
tannergooding: Updated to have the `MockDbDataReader` filter based on the input column names.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoaderCursor.cs
eerhardt: Hmm, it is odd that `System.Data.DbType` has unsigned integers, but `DbDataReader` doesn't seem to support them. Oh well, I guess this is the best we can do.

---
In reply to: [308375777](https://github.com/dotnet/machinelearning/pull/4035#discussion_r308375777) [](ancestors = 308375777)

src/Microsoft.ML.Experimental/Properties/AssemblyInfo.cs
eerhardt: I think this is the only comment I have left on this PR. If this is resolved, and all the builds pass, I think we will be ready to merge this.

tannergooding: Yes. (sorry forgot to update the comment)

This is still needed for converting from the DbType to the InternalKind to the public DataKind.

PR #4050, opened 2019-07-31
Exposing a new DatabaseReader type to track state and allow multi-threaded usage

tannergooding:
As per the title, this adds a new abstract `DatabaseReader` class that allows the various database types to more easily track and expose the state required. This allows easy and configurable state tracking and cleanup,

### DIFF
--- a/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseSource.cs
+++ b/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseSource.cs
@@ -7,9 +7,14 @@ using Microsoft.ML.Runtime;
 
 namespace Microsoft.ML.Data
 {
+    /// <summary>Exposes the data required for opening a database for reading.</summary>
     public sealed class DatabaseSource
     {
-        public DatabaseSource(DbProviderFactory providerFactory, string connectionString, string commandText = "")
+        /// <summary>Creates a new instance of the <see cref="DatabaseSource" /> class.</summary>
+        /// <param name="providerFactory">The factory used to create the <see cref="DbConnection"/>..</param>
+        /// <param name="connectionString">The string used to open the connection.</param>
+        /// <param name="commandText">The text command to run against the data source.</param>
+        public DatabaseSource(DbProviderFactory providerFactory, string connectionString, string commandText)
         {
             Contracts.CheckValue(providerFactory, nameof(providerFactory));
             Contracts.CheckValue(connectionString, nameof(connectionString));
@@ -20,10 +25,13 @@ namespace Microsoft.ML.Data
             CommandText = commandText;
         }
 
+        /// <summary>Gets the text command to run against the data source.</summary>
         public string CommandText { get; }
 
+        /// <summary>Gets the string used to open the connection.</summary>
         public string ConnectionString { get; }
 
+        /// <summary>Gets the factory used to create the <see cref="DbConnection"/>.</summary>
         public DbProviderFactory ProviderFactory { get; }
     }
 }
--- a/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseSource.cs
+++ b/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseSource.cs
@@ -1,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Data.Common;
+using Microsoft.ML.Runtime;
+
+namespace Microsoft.ML.Data
+{
+    public sealed class DatabaseSource
+    {
+        public DatabaseSource(DbProviderFactory providerFactory, string connectionString, string commandText = "")
+        {
+            Contracts.CheckValue(providerFactory, nameof(providerFactory));
+            Contracts.CheckValue(connectionString, nameof(connectionString));
+            Contracts.CheckValue(commandText, nameof(commandText));
+
+            ProviderFactory = providerFactory;
+            ConnectionString = connectionString;
+            CommandText = commandText;
+        }
+
+        public string CommandText { get; }
+
+        public string ConnectionString { get; }
+
+        public DbProviderFactory ProviderFactory { get; }
+    }
+}
--- a/test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
+++ b/test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
@@ -45,11 +45,9 @@ namespace Microsoft.ML.Tests
             var databaseSource = new DatabaseSource(mockProviderFactory, connectionString, commandText);
 
             var trainingData = loader.Load(databaseSource);
-            //trainingData = mlContext.Data.Cache(trainingData, "Label", "SepalLength", "SepalWidth", "PetalLength", "PetalWidth");
 
             var pipeline = mlContext.Transforms.Conversion.MapValueToKey("Label")
                 .Append(mlContext.Transforms.Concatenate("Features", "SepalLength", "SepalWidth", "PetalLength", "PetalWidth"))
-                //.AppendCacheCheckpoint(mlContext)
                 .Append(mlContext.MulticlassClassification.Trainers.LightGbm())
                 .Append(mlContext.Transforms.Conversion.MapKeyToValue("PredictedLabel"));
 
@@ -96,11 +94,9 @@ namespace Microsoft.ML.Tests
             var databaseSource = new DatabaseSource(mockProviderFactory, connectionString, commandText);
 
             var trainingData = loader.Load(databaseSource);
-            //trainingData = mlContext.Data.Cache(trainingData, "Label", "SepalLength", "SepalWidth", "PetalLength", "PetalWidth");
 
             var pipeline = mlContext.Transforms.Conversion.MapValueToKey("Label")
                 .Append(mlContext.Transforms.Concatenate("Features", "SepalLength", "SepalWidth", "PetalLength", "PetalWidth"))
-                //.AppendCacheCheckpoint(mlContext)
                 .Append(mlContext.MulticlassClassification.Trainers.SdcaMaximumEntropy())
                 .Append(mlContext.Transforms.Conversion.MapKeyToValue("PredictedLabel"));
 
--- a/test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
+++ b/test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
@@ -27,6 +27,8 @@ namespace Microsoft.ML.Tests
         public void IrisLightGbm()
         {
             var mlContext = new MLContext(seed: 1);
+            var connectionString = GetDataPath(TestDatasets.iris.trainFilename);
+            var commandText = "Label;SepalLength;SepalWidth;PetalLength;PetalWidth";
 
             var loaderColumns = new DatabaseLoader.Column[]
             {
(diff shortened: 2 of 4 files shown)


src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseReader.cs
```diff
@@ -0,0 +1,89 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Data.Common;
+
+namespace Microsoft.ML.Data
+{
+    public abstract class DatabaseReader : IDisposable
```
tannergooding: I ended up going with this layout after experimentation with several different models.

This, I believe, should be a good compromise over tracking the least amount of state and requiring the end user to implement the least amount of code.

src/Microsoft.ML.Experimental/Microsoft.ML.Experimental.csproj
```diff
@@ -5,6 +5,10 @@
     <IncludeInPackage>Microsoft.ML.Experimental</IncludeInPackage>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.Data.SqlClient" Version="4.6.1" />
```
eerhardt: We shouldn't be adding this dependency to the core logic. If we have APIs that are dependent on `System.Data.SqlClient`, those need to be split out into a separate nuget package.

src/Microsoft.ML.Experimental/DataLoadSave/Database/SqlDatabaseReader.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Microsoft.ML.Data
+{
+    public sealed class SqlDatabaseReader : DatabaseReader
+    {
+        public SqlDatabaseReader(string connectionString, string commandText) : base(connectionString, commandText)
+        {
+        }
+
+        protected override DbConnection CreateConnection()
```
tannergooding: In the simplest scenario, the user just needs to implement `CreateConnection`. The three steps can be overridden and customized as needed, however.

src/Microsoft.ML.Experimental/Microsoft.ML.Experimental.csproj
tannergooding: This is just in the experimental package. I don't feel its appropriate to have 3-4 different experimental packages (one for each database type) and we should probably split these out when they go into the main branch.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseReader.cs
eerhardt: I think we could iterate on the name here some. I don't think `DatabaseReader` is a good name, because this doesn't actually do the reading.

Maybe `DatabaseFactory` or similar?

Actually, as I type this, I now remember: [https://docs.microsoft.com/en-us/dotnet/api/system.data.common.dbproviderfactory.createconnection](https://docs.microsoft.com/en-us/dotnet/api/system.data.common.dbproviderfactory.createconnection).  What if we just took a `DbProviderFactory` instance instead of defining our own new type?

src/Microsoft.ML.Experimental/DataLoadSave/Database/SqlDatabaseReader.cs
```diff
@@ -0,0 +1,23 @@
...
+        {
+            var connection = new SqlConnection(ConnectionString);
+            connection.Open();
```
eerhardt: When we are explicitly opening connections, we will need to handle transient connection issues.

src/Microsoft.ML.Experimental/Microsoft.ML.Experimental.csproj
eerhardt: Just know that is undone work that will have to be done at some time.

eerhardt: Thinking about it, I believe if we use the DbProviderFactory approach suggested above, we probably won’t need a specific reference to the SQL client. Instead the user would pass us the SqlClientProviderFactory.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseReader.cs
CESARDELATORRE: I agree, I don't feel comfortable by creating our own new type for an abstraction or factory if there's something comparable available already in system.data such as DbProviderFactory.

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
```diff
@@ -27,6 +27,8 @@ public DatabaseLoaderTests(ITestOutputHelper output)
         public void IrisLightGbm()
         {
             var mlContext = new MLContext(seed: 1);
+            var connectionString = GetDataPath(TestDatasets.iris.trainFilename);
+            var commandText = "Label;SepalLength;SepalWidth;PetalLength;PetalWidth";
```
CESARDELATORRE: Why is the command text (`commandText`) having just a list of fields/columns? What if the user wants to run a join query against multiple tables?
Also, where the name of the database **table** would be specified in the sample if you are not accepting a real SQL statement?
I might be missing some context here.. sorry..

CESARDELATORRE:
When using different database providers, this is probably the code we want to use so we don't have direct dependencies in our code to explicit classes such as `System.Data.OracleClient` or `System.Data.SqlClient` when creating the higher level APIs:

```
string myProviderName = "MySql.Data.MySqlClient";  //It could be any other provider

//get the proper factory 
DbProviderFactory factory = 
  DbProviderFactories.GetFactory(connectionStringSettings.myProviderName);

//create a command of the proper type.
DbConnection conn = factory.CreateConnection();

//set the connection string
conn.ConnectionString = myConnectionString;

//open the connection, etc. 
conn.Open();
using (DbCommand dbcmd = dbcn.CreateCommand())
{
      cmd.CommandText = "SELECT Features.Label, Features.CustomerID, Customers.CustomerDetail, 
                                           Features.Col1, Features.Col2
                                           FROM Features
                                           INNER JOIN Features ON Features.CustomerID=Customers.CustomerID";
      
       //rest of the code...
```

That code above is 'standard code' from System.Data without using any class from an specific provider. :)

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
tannergooding: This is just for the MockCommand for testing purposes and is being done because we dont have a good way of hosting a database for the tests.

tannergooding: The MockCommand/Connection/Reader are themselves just a thin shim over the TextLoader

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
```diff
@@ -37,43 +39,84 @@ public void IrisLightGbm()
                 new DatabaseLoader.Column() { Name = "PetalWidth", Type = DbType.Single }
             };
 
-            using (var connection = new MockConnection(mlContext, GetDataPath(TestDatasets.iris.trainFilename), loaderColumns))
+            var loader = mlContext.Data.CreateDatabaseLoader(loaderColumns);
+
+            var trainingData = loader.Load(() => new MockDatabaseReader(mlContext, loaderColumns, connectionString, commandText));
```
CESARDELATORRE: I'm okay if we have units tests using a mock, but we should also have a real **integration test** using at least **SQL locadb**. Sometimes using mocks might have slightly differences that can impact us..

tannergooding: Yes, I agree 100%.

However, I don't believe we have anything setup for doing this in any of the repos, and I am not checking a database into git, so we need a solution for pulling that down and managing it (even if its an azure blob).

CESARDELATORRE: For the the .mdf and .ldf files we could have the same approach to what we do for test dataset files. We have NuGet packages with it since they are small datasets just for tests. We can probably do the same for the localdb files?

For tests, the database files should be kind of small. If testing with large databases those kind of tests shouldn't be running on localdb and our Azure Pipelines unless the database is published in an standalone Azure SQL Database so no large file is downloaded when running the tests.

I also have very large databases (up to 50GB with more than 100 million records) ready to be used for testing in an Azure SQL DB. But this is more for end-to-end sample apps testing.

src/Microsoft.ML.Experimental/Microsoft.ML.Experimental.csproj
tannergooding: @eerhardt, @CESARDELATORRE. How are we going to handle resource cleanup here? `DbProviderFactory` doesn't expose IDisposable. So are we just going to track the input and call `DbProviderFactory.CreateCommand()` and then dispose it always?

tannergooding: Also, how is it expected to work with users needing to specify a single `commandText` for all commands in something like the simple Sdca test in the PR?

If users are expected to follow a pattern like https://github.com/dotnet/machinelearning/pull/4050#issuecomment-517451687, then they will still need to manually track all of the state around the created connections/commands and shared state (like `commandText`, etc).

eerhardt: That would be my suggestion. It is how typical ADO.NET code is written.

Check out:

https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/dbconnection-dbcommand-and-dbexception#retrieving-data-example

and the older article:
https://docs.microsoft.com/en-us/previous-versions/dotnet/articles/ms971499(v=msdn.10)

eerhardt: Btw - you don't dispose the Factory.  You dispose the Connection, Command, and DataReader.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseSource.cs
```diff
@@ -0,0 +1,29 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Data.Common;
+using Microsoft.ML.Runtime;
+
+namespace Microsoft.ML.Data
+{
+    public sealed class DatabaseSource
+    {
+        public DatabaseSource(DbProviderFactory providerFactory, string connectionString, string commandText = "")
```
eerhardt: We probably don't want a default on `commandText` right? Normal operations wouldn't work without supplying a commandText.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseSource.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Data.Common;
+using Microsoft.ML.Runtime;
+
+namespace Microsoft.ML.Data
+{
+    public sealed class DatabaseSource
```
eerhardt: Can we slap a few quick xml summary comments to the public APIs?

Also, I assume we can unseal this later, if necessary. Should be fine to seal it for now.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseSource.cs
tannergooding: Yeah, that's probably right. It doesn't make sense to have a default if the user isn't overriding anything

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseSource.cs
tannergooding: Right. Unsealing is safe, sealing is not

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
```diff
@@ -37,43 +39,90 @@ public void IrisLightGbm()
...
             };
 
-            using (var connection = new MockConnection(mlContext, GetDataPath(TestDatasets.iris.trainFilename), loaderColumns))
+            var loader = mlContext.Data.CreateDatabaseLoader(loaderColumns);
+
+            var mockProviderFactory = new MockProviderFactory(mlContext, loaderColumns);
+            var databaseSource = new DatabaseSource(mockProviderFactory, connectionString, commandText);
+
+            var trainingData = loader.Load(databaseSource);
+            //trainingData = mlContext.Data.Cache(trainingData, "Label", "SepalLength", "SepalWidth", "PetalLength", "PetalWidth");
```
eerhardt: I think we can remove these commented out lines.

tannergooding: Fixed.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseSource.cs
tannergooding: Added comments.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseSource.cs
tannergooding: Fixed.

PR #4054, opened 2019-08-01
WIP: Preview of Torch support

### LINKED ISSUE #4053: Torch support in ML.NET
As ML.NET increases it's support for deep learning, we should add support for PyTorch models.
[TorchSharp](https://github.com/xamarin/TorchSharp) has bindings for LibTorch (PyTorch's C++ frontend) and offers the following functionality:

- Score models that were previously defined and trained in PyTorch's python API
- Define and train models directly in C#

We could have a scoring functionality similar to the TensorFlow scoring. We could also add training of Torch models defined in C#.

artidoro:
Fixes #4053.

I follow the API used in `TensorFlowTransformer` very closely to add support for Torch. We have `TorchModel` class that can be loaded through `mlContext.Model.LoadTorchModel("pathToModel")`. There is then a method on the `TorchModel` class that produces `TorchScoringEstimator` and that requires the usual names of input and output columns as well as the shape of the input vector. After calling `Fit()` on the pipeline we obtain a `TorchTransformer` that applies the Torch model to data in an `IDataView` column.

For training the only difference will be that instead of providing a path to a pretrained model, a `TorchModel` class will be created from a `TorchSharp.NN.Module` class. 

Support for Torch in ML.NET:
- [x] scoring
- [ ] training


codemzs:
@interesaaat Why are the builds failing?

eerhardt:
This PR is out of date. Closing for now.

If someone is interested in resolving the conflicts, go ahead.

PR #4057, opened 2019-08-02
Image classification using DNNs and Transfer Learning.

codemzs:
- Rewired Microsoft.ML.Tensorflow to use Tensorflow .NET C# bindings.
- Introduces Microsoft.ML.Dnn package that contains transfer learning based image classification API using resnet and inception V3 models. 
  - Contains API to retrain a DNN model using tensorflow. 

All tensorflow tests are passing. The current test failures have to do with ONNX tests that seem to be using a higher version of protobuf since TF.NET also uses protobuf.


docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/TransferLearning.cs
```diff
@@ -0,0 +1,140 @@
...
+                mlContext.Transforms.Conversion.MapValueToKey(
+                    nameof(TensorData.Label))
+                .Append(mlContext.Model.ImageClassification(
+                    nameof(TensorData.input),
+                    nameof(TensorData.Label), batchSize: 2,
+                        addBatchDimensionInput: true));
+
+            // Run the pipeline and get the transformed values.
+            var estimator = pipeline.Fit(idv);
+            var transformedValues = estimator.Transform(idv);
```
CESARDELATORRE: For this example and making batch predictions with Transform() I think we should provide a different TEST IDataView instead of using the same IDataView used for training too?

docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/TransferLearning.cs
```diff
@@ -0,0 +1,140 @@
...
+            [VectorType(imageHeight, imageWidth, numChannels)]
+            public float[] input { get; set; }
+
+            public Int64 Label { get; set; }
+        }
+
+        /// <summary>
+        /// Method to generate sample test data. Returns 2 sample rows.
+        /// </summary>
+        public static TensorData[] GetTensorData()
```
CESARDELATORRE: For example code like this I think it'd be much clearer for users to have a real image-set (TensorData) coming from real image files instead of auto-generated sample test data

src/Microsoft.ML.Dnn/DnnTransform.cs
```diff
@@ -0,0 +1,1737 @@
...
+        private class TensorValueGetter<T> : ITensorValueGetter
+        {
+            private readonly ValueGetter<T> _srcgetter;
+            private readonly T[] _bufferedData;
+            private readonly Int64[] _bufferedDataLong;
+            private readonly TensorShape _tfShape;
+            private int _position;
+            private readonly bool _keyType;
+
+            public TensorValueGetter(
```
CESARDELATORRE: InceptionV3 appears in the enum but it is not used or currently implemented yet as part of the ImageClassifierTrainer, right?
It'd be good to have InceptionV3 as soon as possible in addition to ResNet since most of our samples use Inception and it'll be good to be able to compare "apples with apples". If we compare with Resnet the results will be very different, logically.

CESARDELATORRE: If we're thinking about having Torch as another possibility for the ImageClassifier trainer we probably don't want to have the word TensorFlow hard-coded for the Label as `TensorFlowLabel` if the `Options `class will be generic for any DNN model, but maybe something generic like `DnnModelLabel`.

On the other hand, if the `Options `class will be specific for TensorFlow because the options/configurations are very different in TF compared to Torch, then we should probably call it something like `TensorFlowModelOptions` class and have another class in the future like `TorchModelOptions`.

CESARDELATORRE: This class name (`Options`) feels too generic. 

- If it is specific and exclusive about **TensorFlow** it should probably be named as 'TensorFlowOptions'. 

- If it'll be a common type shared among TF and Torch, then maybe call it `DnnModelOptions`.

If it is a shared/generic class, I see potential issues here. This type looks generic since it is part of the DnnEstimator constructor, however, the internal options look very specific to TensorFlow. I don't think you wil have properties such as `TensorFlowLabel` and not sure that all the rest of properties of Options would be applicable to Torch?

CESARDELATORRE: Same comment I did when using the `Options` class:

This class name (`Options`) feels too generic. 

- If it is specific and exclusive about **TensorFlow** it should probably be named as 'TensorFlowOptions'. 

- If it'll be a common type shared among TF and Torch, then maybe call it `DnnModelOptions`.

If it is a shared/generic class, I see potential issues here. This type looks generic since it is part of the DnnEstimator constructor, however, the internal options look very specific to TensorFlow. I don't think you wil have properties such as `TensorFlowLabel` and not sure that all the rest of properties of Options would be applicable to Torch?

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// <param name="arch"></param>
+        /// <param name="dnnFramework"></param>
+        /// <param name="epoch"></param>
+        /// <param name="batchSize"></param>
+        /// <param name="learningRate"></param>
+        /// <param name="addBatchDimensionInput"></param>
+        /// <remarks>
+        /// The support for image classification is under preview.
+        /// </remarks>
+        public static DnnEstimator ImageClassification(
```
CESARDELATORRE: `ImageClassification `should allow the user to provide their own DNN model (TF model until now) in their parameters, meaning that as an optional parameter the user should be able to provide the `Options `object including the 'ModelLocation` (either on a local drive path or an HTTP path pointing to the source model) and the additional TF model configuration (labelColumn, epoch, etc.) for the chose TF model.

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -0,0 +1,133 @@
...
+            Architecture arch = Architecture.ResnetV2101,
+            DnnFramework dnnFramework = DnnFramework.Tensorflow,
+            int epoch = 10,
+            int batchSize = 20,
+            float learningRate = 0.01f,
+            bool addBatchDimensionInput = false)
+        {
+            var options = new Options()
+            {
+                ModelLocation = arch == Architecture.ResnetV2101 ? @"DnnImageModels\Resnet101V2Tensorflow\resnet_v2_101_299.meta" : "",
```
CESARDELATORRE: The TensorFlow model location and rest of the base TF model's parameters are hard-coded in the ImageClassification method.
This should be configurable depending on the chosen base-model/`Architecture`.

Even if we currently have a single base-model/`Architecture` (ResnetV2101) it could be coded as if we had more additional models such as Inceptionv3, which we should have soon since most of our samples use Inceptionv3, so we can compare performance and quality.

src/Microsoft.ML.Dnn/DnnCatalog.cs
codemzs: No, the design of this API is to not allow the users to add their own models. We have a set if models and they get to pick. There will be a different API that will let you do transfer learning for any model.

src/Microsoft.ML.Dnn/DnnCatalog.cs
codemzs: They are not hardcoded, you can set them from API. @"DnnImageModels\Resnet101V2Tensorflow\resnet_v2_101_299.meta" will be downloaded to the bin location when you add the nuget for resnet.

src/Microsoft.ML.Dnn/DnnTransform.cs
codemzs: @CESARDELATORRE These are internal methods not public facing APIs.

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// <param name="metricOperation">The name of the operation in the TensorFlow graph to compute performance metric during training (Optional).</param>
+        /// <param name="learningRateOperation">The name of the operation in the TensorFlow graph which sets optimizer learning rate (Optional).</param>
+        /// <param name="learningRate">Learning rate to use during optimization (Optional).</param>
+        /// <param name="addBatchDimensionInput">Add a batch dimension to the input e.g. input = [224, 224, 3] => [-1, 224, 224, 3].
+        /// This parameter is used to deal with models that have unknown shape but the internal operators in the model require data to have batch dimension as well.</param>
+        /// <param name="dnnFramework"></param>
+        /// <remarks>
+        /// The support for retraining is under preview.
+        /// </remarks>
+        public static DnnEstimator RetrainDnnModel(
```
CESARDELATORRE: Since there's a DnnFramework parameter to discriminate between TF and Torch in the future, the rest of parameters should be shared across TF and Torch.. 

Are we certain that all those parameters are valid and used in Torch? (For sure the parameter tensorFlowLabel is not applicable for Torch, it should be renamed to dnnModelLabel or similar). But are all the rest applicable for both frameworks?

If not, it'd be better to have segregation between parameters that are valid for one or the other framework like by using config classes such as `TensorFlowOptions `and `TorchOptions` depending on the selected framework.

Another possibility is to have two different methods in the catalog instead of just `RetrainDnnModel()`:

- `RetrainTensorflowModel()`
- `RetrainTorchModel()`

src/Microsoft.ML.Dnn/DnnCatalog.cs
CESARDELATORRE: You're right, I only meant the Model location. But we might want to hide.
The only thing is that if the users chooses now  `Architecture.InceptionV3`, then the model location is blank, and it might get an exception.. 
But if we're adding InceptionV3 soon, I'm okay. I'll resolve this topic. 👍

src/Microsoft.ML.Dnn/DnnCatalog.cs
CESARDELATORRE: I understood that in the discussion the other day, that this API would allow additional models.
Why a different API if the 'external' model can also be about ImageClassification?
If it is something very different I understand, but being something pretty similar (another model for ImageClassification) why do we want a different API? We could have optional parameters, too.

src/Microsoft.ML.Dnn/DnnTransform.cs
CESARDELATORRE: Even if those methods are internal, it'd be better if the code is consistent, I think.
The code is also open source, so anyone can research it.

Oceania2018:
@CESARDELATORRE We will add more other Model Architectures like `MobileNet`, `ResNet` and `NASNet`, which could give you some extra precision. Hope it be true for ML.NET v1.4 release.

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -0,0 +1,166 @@
...
+                ScoreColumnName = scoreColumnName,
+                PredictedLabelColumnName = predictedLabelColumnName,
+                CheckpointName = checkpointName,
+                Arch = arch,
+                MeasureTrainAccuracy = measureTrainAccuracy
+            };
+
+            if (!File.Exists(options.ModelLocation))
+            {
+                if (options.Arch == Architecture.InceptionV3)
```
CESARDELATORRE: This is great, I see you added InceptionV3 over the weekend! 👍

src/Microsoft.ML.Dnn/DnnTransform.cs
CESARDELATORRE: I'm resolving this comment since I just saw you added INCEPTIONV3 to the code over the weekend. 👍

codemzs: These are all internal classes and right now backend is TF specific. I agree it will be nice to not have TF in the name, may be I can make that change before next release. 

---
In reply to: [310278510](https://github.com/dotnet/machinelearning/pull/4057#discussion_r310278510) [](ancestors = 310278510)

codemzs: Right now everything is using TF so its probably ok to leave there but before we GA we should change the names to be generic. I will suspect this code will change significantly in the next month as we add more scenarios.

---
In reply to: [310331337](https://github.com/dotnet/machinelearning/pull/4057#discussion_r310331337) [](ancestors = 310331337)

codemzs: replied earlier on this topic.

---
In reply to: [310295321](https://github.com/dotnet/machinelearning/pull/4057#discussion_r310295321) [](ancestors = 310295321)

src/Microsoft.ML.Dnn/DnnCatalog.cs
codemzs: ImageClassification is specific to ImageClassification scnearios. I want to control the models so that I can control how I edit the graph. If I allow the user to feed in whatever graph they want then I will need to also provide functionality for the user to do graph editing. The plan is we will have such a generic API before GA.

---
In reply to: [310331079](https://github.com/dotnet/machinelearning/pull/4057#discussion_r310331079) [](ancestors = 310331079)

docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/TransferLearning.cs
codemzs: Ideally but this is just a sample and we have done similar things in other samples too in the project. Its point is to illustrate the idea in a very simple manner.

---
In reply to: [310269964](https://github.com/dotnet/machinelearning/pull/4057#discussion_r310269964) [](ancestors = 310269964)

src/Microsoft.ML.Dnn/DnnCatalog.cs
codemzs: I think these parameters will be shared. They  are standard DNN parameters. Options class is internal. I will change tensorFlowLabel to dnnLabel

---
In reply to: [310329135](https://github.com/dotnet/machinelearning/pull/4057#discussion_r310329135) [](ancestors = 310329135)

codemzs:
@Oceania2018 This release we are not advertising GPU support. I want to first test it before we remove this dependency.

codemzs:
Thanks, @CESARDELATORRE for reviewing.

PR #4060, opened 2019-08-04
LDSVM trainer

### LINKED ISSUE #4031: Add a Local Deep Kernel Learning trainer
http://manikvarma.org/pubs/jose13.pdf

yaeldekel:
Add an LDSVM trainer.
Fixes #4031 .
I will add documentation and samples in the next iterations.


src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
```diff
@@ -873,5 +873,32 @@ private static ICalibratorTrainer GetCalibratorTrainerOrThrow(IExceptionContext
             Contracts.CheckValue(catalog, nameof(catalog));
             return new PriorTrainer(CatalogUtils.GetEnvironment(catalog), labelColumnName, exampleWeightColumnName);
         }
+
+        /// <summary>
+        ///
```
eerhardt: Are you going to fill these docs out?

src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
```diff
@@ -873,5 +873,32 @@ private static ICalibratorTrainer GetCalibratorTrainerOrThrow(IExceptionContext
...
+        /// </summary>
+        /// <param name="catalog"></param>
+        /// <param name="options"></param>
+        /// <returns></returns>
+        public static LdSvmTrainer LdSvm(this BinaryClassificationCatalog.BinaryClassificationTrainers catalog, LdSvmTrainer.Options options)
+            => new LdSvmTrainer(catalog.GetEnvironment(), options);
+
+        public static LdSvmTrainer LdSvm(this BinaryClassificationCatalog.BinaryClassificationTrainers catalog,
```
eerhardt: This public method should have docs too.

src/Microsoft.ML.StandardTrainers/StandardTrainersCatalog.cs
```diff
@@ -873,5 +873,32 @@ private static ICalibratorTrainer GetCalibratorTrainerOrThrow(IExceptionContext
...
+            string labelColumnName = DefaultColumnNames.Label,
```
eerhardt: Why aren't you using `labelColumnName`, `featureColumnName`, and `exampleWeightColumnName` in the method body?

src/Microsoft.ML.StandardTrainers/LdSvm/LdSvmModelParameters.cs
```diff
@@ -0,0 +1,282 @@
...
+
+[assembly: LoadableClass(typeof(LdSvmModelParameters), null, typeof(SignatureLoadModel), "LDSVM binary predictor", LdSvmModelParameters.LoaderSignature)]
+
+namespace Microsoft.ML.Trainers
+{
+    public sealed class LdSvmModelParameters : ModelParametersBase<float>,
+        IValueMapper,
+        ICanSaveModel
+    {
+        public const string LoaderSignature = "LDSVMBinaryPredictor";
```
eerhardt: Does this need to be public?

src/Microsoft.ML.StandardTrainers/LdSvm/LdSvmTrainer.cs
```diff
@@ -0,0 +1,489 @@
...
+            Host.CheckValue(options, nameof(options));
+            CheckOptions(Host, options);
+            _options = options;
+        }
+
+        // REVIEW: This does not need caching, but only because it's very
+        // badly written and the first thing is it just grabs all instances. If this
+        // ever changes, do review this return value to make sure it is still
+        // appropriate.
+        private static TrainerInfo _info = new TrainerInfo(calibration: true, caching: false);
```
eerhardt: should be `private readonly static` ?

src/Microsoft.ML.StandardTrainers/LdSvm/LdSvmTrainer.cs
```diff
@@ -0,0 +1,489 @@
...
+                  TrainerUtils.MakeBoolScalarLabel(options.LabelColumnName),
+                  TrainerUtils.MakeR4ScalarWeightColumn(options.ExampleWeightColumnName))
+        {
+            Host.CheckValue(options, nameof(options));
+            CheckOptions(Host, options);
+            _options = options;
+        }
+
+        // REVIEW: This does not need caching, but only because it's very
+        // badly written and the first thing is it just grabs all instances. If this
```
eerhardt: > `badly written and the first thing is it just grabs all instances`

Does this mean this trainer can't stream data, but instead it needs to load all the data into memory before it can train on it? Is that acceptable?

justinormont: When looking thru the code, it seemed like a sizable amount of work to bring this to a streaming mode. I didn't see an obvious method to be performant and streaming -- seems like a lots of passes over the dataset. I very well may be missing the obvious though.

src/Microsoft.ML.StandardTrainers/LdSvm/LdSvmTrainer.cs
```diff
@@ -0,0 +1,511 @@
...
+            [TlcModule.SweepableDiscreteParam("Sigma", new object[] { 1.0f, 0.1f, 0.01f })]
+            public float Sigma = Defaults.Sigma;
+
+            /// <summary>
+            /// Indicates if we should use Bias or not in our model.
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce, HelpText = "No bias", ShortName = "noBias")]
+            [TlcModule.SweepableDiscreteParam("NoBias", null, isBool: true)]
+            public bool NoBias = Defaults.NoBias;
+
```
harishsk: There seems to be a double negative in naming here. The NoBias term defaults to false. Would it be better to name this variable UseBias and default it to true?

src/Microsoft.ML.StandardTrainers/LdSvm/LdSvmTrainer.cs
yaeldekel: In the most recent iteration I added two code paths: one does streaming and one does caching. From running some experiments, the streaming version is around 2x slower. Could you take a look and let me know whether it's worth keeping both?

@justinormont, @harishsk

---
In reply to: [313606773](https://github.com/dotnet/machinelearning/pull/4060#discussion_r313606773) [](ancestors = 313606773)

harishsk: Is the streaming version 2x slower even in the case of large data? If there is a real tradeoff for large data sizes, then it is worth keeping both. 

---
In reply to: [368572412](https://github.com/dotnet/machinelearning/pull/4060#discussion_r368572412) [](ancestors = 368572412,313606773)

PR #4061, opened 2019-08-05
Release notes for v1.3.0

codemzs:


docs/release-notes/1.3.0/release-1.3.0.md
```diff
@@ -0,0 +1,97 @@
+# [ML.NET](http://dot.net/ml) 1.3.0
+
+## **New Features**
+- **Deep Neural Networks Training (PREVIEW)** ([#4057](https://github.com/dotnet/machinelearning/pull/4057))
```
CESARDELATORRE: I think we should be clearer about what's in 1.3 release and what's being released as PREVIEW in 0.15 which is not really 1.3.

In this case, DNN retrain with transfer learning is PREVIEW as part of 0.15 Preview, not really part of 1.3.0 packages, so probably in the 1.13 RELEASE notes we should start with what's in the GA/RTM release which in this case is mostly about bug fixing.

docs/release-notes/1.3.0/release-1.3.0.md
```diff
@@ -0,0 +1,97 @@
+# [ML.NET](http://dot.net/ml) 1.3.0
+
+## **New Features**
+- **Deep Neural Networks Training (PREVIEW)** ([#4057](https://github.com/dotnet/machinelearning/pull/4057))  
+  Introduces `Microsoft.ML.DNN` package that enables full DNN model retraining and transfer learning in .NET using C# bindings for tensorflow provided by Tensorflow .NET. The goal of this package is to allow high level DNN training and scoring tasks such as image classification, text classification, object detection, etc using simple yet powerful APIs that are framework agnostic but currently they only uses Tensorflow at the backend. The below APIs are in early preview and we hope to get customer feedback that we can incorporate in the next iteration.
+
+  ![DNN stack](dnn_stack.jpg)
+
+  ```cs
```
CESARDELATORRE: Do we need to show this amount of code in the RELEASE NOTES? In the past releases we usually never had this kind of code or signature code in the RELEASE NOTES. It feels cleaner and more concise if we don't put code as part of the release notes, I think. But links pointing to code examples for each feature is usually what we had. 👍

docs/release-notes/1.3.0/release-1.3.0.md
codemzs: @CESARDELATORRE I have put "PREVIEW" in heading that implies everything is preview, no?

docs/release-notes/1.3.0/release-1.3.0.md
codemzs: Pytorch and TF do it quite a bit and even show diagrams ...

docs/release-notes/1.3.0/release-1.3.0.md
```diff
@@ -0,0 +1,97 @@
...
+
+  [Image classification (Resnet V2 101) sample](https://github.com/dotnet/machinelearning/blob/b6146219156cb05e2e2cebd80fe4cda017f03470/docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearning.cs)
+
+- **PyTorch scoring support (PREVIEW)** ([#4054](https://github.com/dotnet/machinelearning/pull/4054))  
+  Introduces `Microsoft.ML.PyTorch` package that will enable scoring pytorch models in ML .NET. We plan to bring in PyTorch training next release.
+  ```cs
+  public static TorchModel LoadTorchModel(this ModelOperationsCatalog catalog, string modelLocation)
+  ```
+
+- **SQL Dataloader (PREVIEW)** ([#4035](https://github.com/dotnet/machinelearning/pull/4035))
```
CESARDELATORRE: It might be better if we don't call it "SQL Dataloader" but "Database Loader (PREVIEW)" since folks can think it only supports SQL Server if using SQL as part of the name/title. In the release notes we should say that this loader supports any relational database supported by System.Data in .NET Framework or .NET Core, meaning that you can use many RDBMS such as SQL Server, Azure SQL Database, Oracle, PostgreSQL, MySQL, etc. 
Also, the name is DatabaseLoader, we're not really using the acronym SQL as part of the name precisely to avoid confusion identifying it with SQL Server (Since in .NET System.Data.SqlClient targets only SQL Server family).

docs/release-notes/1.3.0/release-1.3.0.md
```diff
@@ -0,0 +1,97 @@
...
+  ```cs
+  public static TorchModel LoadTorchModel(this ModelOperationsCatalog catalog, string modelLocation)
+  ```
+
+- **SQL Dataloader (PREVIEW)** ([#4035](https://github.com/dotnet/machinelearning/pull/4035))  
+  Introduces SQL Dataloader that enables training on SQL tables. This feature is also in early preview and customer feedback will be much appreciated. 
+
+  [Design specification](https://github.com/dotnet/machinelearning/pull/3857)
+
+  ```cs
```
CESARDELATORRE: Until we have an end-to-end example, we can point in the RELEASE NOTES to this test instead of pointing to the specs doc because the code is not implementing 100% of the specs but only the foundational API. 
Sample code in test:
https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Tests/DatabaseLoaderTests.cs

docs/release-notes/1.3.0/release-1.3.0.md
```diff
@@ -0,0 +1,97 @@
...
+  public static DatabaseLoader CreateDatabaseLoader(this DataOperationsCatalog catalog,
+            params DatabaseLoader.Column[] columns)
+  ```    
+
+## **Bug Fixes**
+### Serious
+- **SaveOnnxCommand appears to ignore predictors when saving a model to ONNX format:** This broke export to ONNX functionality. ([3974][https://github.com/dotnet/machinelearning/issues/3974])
+
+- **Unable to use fasterrcnn onnx model** ([3963](https://github.com/dotnet/machinelearning/pull/3963))
+- **PredictedLabel is always true for Anomaly Detection:** This bug disabled scenarios like fraud detection using binary classification/PCA. ([#4039](https://github.com/dotnet/machinelearning/pull/4039))
```
CESARDELATORRE: We recently created this Fraud Detection using the anomaly detection PCA trainer. We still need to test it with this bug fix coming in 1.3. 
This is the sample that could be highlighted in the RELEASE notes, if needed: https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/getting-started/AnomalyDetection_CreditCardFraudDetection

docs/release-notes/1.3.0/release-1.3.0.md
CESARDELATORRE: Right, this point can be arguable, of course. But TF and PyTorch target a very different profile (it is low level DNN APIs mostly used by data scientists, not .NET developers) than ML.NET which is mostly about .NET developers.

See .NET Core release notes samples:
https://github.com/dotnet/machinelearning-samples/projects/1 

See previous ML.NET release notes samples where users would expect to continue to be consistent with being concise in the release notes, too:

https://github.com/dotnet/machinelearning/blob/master/docs/release-notes/1.0.0-preview/release-1.0.0-preview.md

I think diagrams are good, though. But code/signatures, it usually will expand the release notes too much. That kind of details should be explained in further provided links, I think.

docs/release-notes/1.3.0/release-1.3.0.md
CESARDELATORRE: I get that, but feels strange that the title is about the 1.3 release and it starts with PREVIEW features that are NOT part of 1.3 but part of PREVIEW 0.15.

If I read this as a customer without knowing about the existance of 0.15 PREVIEW I would thinkg that the features DNN, DatabaseLoader, etc. are actually part of 1.3 but somehow in PREVIEW? It is simply confusing because in reality, those features in PREVIEW will be part of upcoming GA releases, such as 1.4 or 1.5 depending on when we release it as GA.

Just saying that  we should be clearer on what's exactly coming as 1.3 release (GA), mostly bug fixing, then we can say that in the same timeframe we're releasing features in PREVIEW versioned as 0.15 packages. That's what we are really doing, right?

codemzs: You seem to be overthinking. Lets keep things simple, "its a 1.3 release YAY! and hey we have few preview features in these, these and these packages" ... if they search for those packages they will see ONLY preview version. Preview means Preview ... its not release, it can change anytime.

docs/release-notes/1.3.0/release-1.3.0.md
codemzs: It does not make sense to include sample for a feature whose bug we fixed. This section is just highlighting the bugs we fixed.

codemzs: I have added this under sample's section

PR #4081, opened 2019-08-07
Added RankingEvaluatorOptions and removed the truncation limit. 

### LINKED ISSUE #3993: Exception is thrown if NDCG > 10 is used with LightGbm for evaluating ranking
- Version: ML.NET 1.2.0

The current code in the RankingEvaluator.cs file has the MaxTruncationLevel for NDCG (Normalized Cumulative Gain Metric) set to 10.  Also, the code currently throws an exception if the NDCG is set to a value > 10.  This is a blocking issue for ranking because it prevents the ability to measure the quality of ranking with result sets > 10.  For example, if you were attempting to rank a group of 100 results, with the MaxTruncationLevel of 10, you could only measure whether the first 10 results were ranked correctly.

Here's the code:

```
         public RankingEvaluator(IHostEnvironment env, Arguments args)
            : base(env, LoadName)
        {
            // REVIEW: What kind of checking should be applied to labelGains?
            if (args.DcgTruncationLevel <= 0 || args.DcgTruncationLevel > Aggregator.Counters.MaxTruncationLevel)
                throw Host.ExceptUserArg(nameof(args.DcgTruncationLevel), "DCG Truncation Level must be between 1 and {0}", Aggregator.Counters.MaxTruncationLevel);
            Host.CheckUserArg(args.LabelGains != null, nameof(args.LabelGains), "Label gains cannot be null");
...
}
```
It appears from the //Review comment in the above code that this functionality hasn't been fully completed.  

While I'm unsure what the MaxTruncationLevel value should be, I have seen on a ranking contest\example on Kaggle.com where one contest was measuring NDCG with a truncation level of up to 38.  

I also noticed that in other parts of this file, the code indicates that a value between 0-100 should be allowed:

```
 public Transform(IHostEnvironment env, IDataView input, string labelCol, string scoreCol, string groupCol,
                int truncationLevel, Double[] labelGains)
                : base(env, input, labelCol, scoreCol, groupCol, RegistrationName)
            {
                Host.CheckParam(0 < truncationLevel && truncationLevel < 100, nameof(truncationLevel),
                    "Truncation level must be between 1 and 99");
...
}
```

Also, refer to the linked bug since it is related to this scenario: [Ranker Evaluate doesn't allow you specify metric parameters.] (https://github.com/dotnet/machinelearning/issues/2728)

harishsk:
Summary of changes:
- Added RankingEvaluatorOptions class to control the output of evaluation
- Removed hard coded truncation limit for the max truncation level
- Added corresponding unit tests and maml tests

Fixes #3993


src/Microsoft.ML.Data/Evaluators/RankingEvaluator.cs
```diff
@@ -271,7 +270,7 @@ public sealed class Aggregator : AggregatorBase
         {
             public sealed class Counters
             {
-                public const int MaxTruncationLevel = 10;
+                public const int MaxTruncationLevel = 100;
```
justinormont: Is there a technical need for any fixed limit? From the data science side, I don't know of one. 

Is there a ramification in our code for simply removing the limit?

harishsk: The original author of this code has assumed 100 to be a reasonable limit and that assumption is enforced in many places in the code. The MaxTruncationLevel of 10 set above made it a bit inconsistent and my fix just makes the max limit consistent. 

In theory it should be possible to remove the limit. But the way the code is structured makes it a bit more invasive than necessary.

Right now, in `RankingEvaluator.cs` the static class `RankingUtils` class has a static method `DiscountMap` which carries this assumption with a hard coded constant of 100 and also has a comment to reflect that.

`//RankingEvaluator.cs on line 954
`var discountMap = new Double[100]; //Hard to believe anyone would set truncation Level higher than 100`

The way the code is structured, this requires some well defined maximum. It should be possible to remove this but that would require a bigger code change. I would like input on whether or the original author's assumption that 100 is a reasonable limit is a good assumption. If it is really necessary to not have a limit at all, it is a bigger code change. 

Can we take this fix as is and open a new bug to remove the limit and we can triage that separately?

src/Microsoft.ML.Data/Evaluators/RankingEvaluator.cs
```diff
@@ -26,8 +26,7 @@
 
 namespace Microsoft.ML.Data
 {
-    [BestFriend]
-    internal sealed class RankingEvaluator : EvaluatorBase<RankingEvaluator.Aggregator>
+    public sealed class RankingEvaluator : EvaluatorBase<RankingEvaluator.Aggregator>
```
justinormont: Is exposing the four classes the right method?

/cc @eerhardt

eerhardt: No this is not the right approach. @harishsk - can you follow the same pattern we are following for the other evaluators? None of the other evaluator classes are public. There should be no reason why this is.

---
In reply to: [314017216](https://github.com/dotnet/machinelearning/pull/4081#discussion_r314017216) [](ancestors = 314017216)

harishsk:
If these are internal classes, what would be the right way to get at this functionality? @nicolehaugen wrote a sample and used reflection to access the evaluators and requested that this be made public. (See discussion on bug #3993) 
I am happy to revert them to internal classes, but it would help to give a better response on the associated bug.

eerhardt:
> what would be the right way to get at this functionality?

One approach would be to follow the same pattern the MulticlassClassification Evaluate does for `topKPredictionCount`:

https://github.com/dotnet/machinelearning/blob/398a978912b8b3f242bbecb3afe339c8cd9098f6/src/Microsoft.ML.Data/TrainCatalog.cs#L492-L506

However, I don't really like this approach because it doesn't scale well when you want to add more options in the future.

So another approach would be to create a new `public class MulticlassClassificationEvaluateOptions` class that could hold more options in the future. Then add a new overload to the public Evaluate method that takes in one of these Options. This is the pattern we follow for all the transforms and trainers in ML.NET.

In your case it would be a `RankingEvaluateOptions` class, and it could contain `DcgTruncationLevel` and `LabelGains`, and potentionally `OutputGroupSummary`.

harishsk:
If the new public class `RankingEvaluatorOptions` contains `DcgTruncationLevel`, `LabelGains` and `OutputGroupSummary` that class would become identical to the internal `RankingEvaluator.Arguments` class. 
Does it make sense then to make the `RankingEvaluator.Arguments` class public and rename it to `RankingEvaluator.Options`  and change the function signature of `RankingCatalog.Evaluate` to include the options with a default of null?

eerhardt:
> Does it make sense then to make the RankingEvaluator.Arguments class public 

No that doesn't make sense because `RankingEvaluator` isn't public. If you have a nested `public class` of an `internal class`, the nested class is still internal.

harishsk:
Sorry that wasn't clear. I should have phrased it differently. 

Let us say we create this new public `RankingEvaluatorOptions` class with the suggested members. This class will be identical to `RankingEvaluator.Arguments`. Then can we eliminate the `RankingEvaluator.Arguments` class and use `RankingEvaluatorOptions` in place `RankingEvaluator.Arguments`?
Is there any need to have to two classes with identical members and nearly identical function, except that one is public and one is private? Can we not use the public class for internal purposes as well?

eerhardt:
> Then can we eliminate the RankingEvaluator.Arguments class and use RankingEvaluatorOptions in place RankingEvaluator.Arguments?

I think that would be fine. If you took that approach, there are a few things you need to ensure you don't break:

* Entry Points
* The MAML command line

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -638,6 +638,21 @@ internal RankingTrainers(RankingCatalog catalog)
...
+        /// Evaluates scored ranking data.
+        /// </summary>
+        /// <param name="data">The scored data.</param>
+        /// <param name="options">Options to control the evaluation result.</param>
+        /// <param name="labelColumnName">The name of the label column in <paramref name="data"/>.</param>
+        /// <param name="rowGroupColumnName">The name of the groupId column in <paramref name="data"/>.</param>
+        /// <param name="scoreColumnName">The name of the score column in <paramref name="data"/>.</param>
+        /// <returns>The evaluation results for these calibrated outputs.</returns>
+        public RankingMetrics Evaluate(IDataView data,
+            RankingEvaluatorOptions options,
```
eerhardt: Can you add at least one new test for this new API?

justinormont: And ensure there is an existing unit test covering the existing MAML interface for `RankingEvaluator`.

It should look about like:
`CV tr=LightGBMRanking eval=RankingEvaluator{t=10} k=2 loader=TextLoader{col=Label:R4:0 col=GroupId:TX:1 col=Features:R4:2-138 header=+} data=MSLR-WEB10K_Fold1.TRAIN.SMALL_10k-rows.tsv xf=HashTransform{col=GroupId} xf=NAHandleTransform{col=Features}`

Specifically the `eval=RankingEvaluator{t=10}` part, where the DCG/NDCG threshold is set.

The output should end in a block of DCG/NDCG@N metrics:
```
OVERALL RESULTS
---------------------------------------
NDCG@1:             51.879160 (3.2943)
NDCG@2:             52.490449 (1.6035)
NDCG@3:             51.704020 (1.7012)
NDCG@4:             51.495099 (0.8726)
NDCG@5:             51.859874 (0.2759)
NDCG@6:             52.278430 (0.2561)
NDCG@7:             52.641604 (0.4342)
NDCG@8:             52.856661 (0.5255)
NDCG@9:             54.075498 (0.5233)
NDCG@10:            54.692374 (0.6690)
DCG@1:              10.778756 (0.1824)
DCG@2:              15.649059 (0.6480)
DCG@3:              18.580052 (0.5112)
DCG@4:              20.808287 (0.9611)
DCG@5:              22.819410 (1.1504)
DCG@6:              24.494010 (1.2360)
DCG@7:              26.108059 (1.6727)
DCG@8:              27.384487 (1.7825)
DCG@9:              29.079234 (1.8749)
DCG@10:             30.308760 (1.9971)

---------------------------------------
Physical memory usage(MB): 123
Virtual memory usage(MB): 5114
8/22/2019 4:56:55 PM	 Time elapsed(s): 5.534
``` 

And it should match the `@N` set in the RankingEvaluator's t=N.

src/Microsoft.ML.Data/Evaluators/RankingEvaluator.cs
```diff
@@ -26,21 +26,24 @@
...
-            [Argument(ArgumentType.AtMostOnce, HelpText = "Label relevance gains", ShortName = "gains")]
-            public string LabelGains = "0,3,7,15,31";
+        [Argument(ArgumentType.AtMostOnce, HelpText = "Label relevance gains", ShortName = "gains")]
+        public string LabelGains = "0,3,7,15,31";
 
-            [Argument(ArgumentType.AtMostOnce, HelpText = "Generate per-group (N)DCG", ShortName = "ogs")]
-            public bool OutputGroupSummary;
-        }
+        [Argument(ArgumentType.AtMostOnce, HelpText = "Generate per-group (N)DCG", ShortName = "ogs")]
+        public bool OutputGroupSummary;
```
eerhardt: From what I can tell by looking through the code, `OutputGroupSummary` has no affect on the output metrics that are returned on the public API.

I think we should either
1. Remove the public option
2. Expose the extra metrics (the group summary) this option controls

A hybrid approach would be to do number (1) now, and log a bug to do (2) later.

harishsk: By "Remove the public option" do you mean to make the OutputGroupSummary member private to the class? Won't that break the usage in the RankingEvaluator class? 
Can we keep it public for now and log a bug to do (2) later?


---
In reply to: [317249875](https://github.com/dotnet/machinelearning/pull/4081#discussion_r317249875) [](ancestors = 317249875)

eerhardt: I wouldn’t recommend adding a public API that doesn’t do anything. My preference would be to add the public API when it actually accomplishes something for the user.

harishsk: Is it acceptable to mark the field internal and leave the rest of the code intact? Or should we remove the variable entirely from the definition of RankingEvaluatorOptions?

---
In reply to: [323032537](https://github.com/dotnet/machinelearning/pull/4081#discussion_r323032537) [](ancestors = 323032537)

eerhardt: Marking it internal is totally acceptable.

src/Microsoft.ML.Data/Evaluators/RankingEvaluator.cs
```diff
@@ -946,23 +948,34 @@ private protected override IEnumerable<string> GetPerInstanceColumnsToSave(RoleM
...
             {
-                double[] result = _discountMap;
-                if (result == null)
+                return _discountMap;
+            }
+        }
+        /// <summary>
+        /// Reallocates discountMap for the largest truncationLevel seen so far
+        /// </summary>
+        public static void SetTruncationLevel(int truncationLevel)
```
eerhardt: This shouldn't be `static`. There is something wrong with the existing code here. Using a global static `_discountMap` array seems bad. What happens if multiple threads are training/evaluating ranking models?

Lynx1820: The way that it's implemented now is by locking the `_discountMap` and resizing if the `truncationLevel` is larger then the `maxTruncationLevel`. This ensures that the `_discountMap` is always large enough. Let me know if there's a better approach.

eerhardt: A better approach is to not use a static variable.

harishsk: RankingUtils is a static class so all its members will be static. I am guessing RankingUtils has been made static because the data in DiscountMap can be shared across threads once initialized. The implemented behavior right now will behave correctly even in a multi-threaded context.

---
In reply to: [327656304](https://github.com/dotnet/machinelearning/pull/4081#discussion_r327656304) [](ancestors = 327656304)

test/Microsoft.ML.TestFramework/Microsoft.ML.TestFramework.csproj
```diff
@@ -22,6 +23,6 @@
     <NativeAssemblyReference Include="MklProxyNative" />
     <NativeAssemblyReference Include="MatrixFactorizationNative" />
     <NativeAssemblyReference Include="MklImports" />
-    <NativeAssemblyReference Condition="'$(OS)' == 'Windows_NT'" Include="libiomp5md"/>
+    <NativeAssemblyReference Condition="'$(OS)' == 'Windows_NT'" Include="libiomp5md" />
   </ItemGroup>
```
harishsk: Minor nitpick, but this line is unchanged except for the space which probably was added mistakenly. Can you please delete the space and leave the line unchanged?

test/Microsoft.ML.TestFramework/TestCommandBase.cs
```diff
@@ -830,7 +846,6 @@ public void SaveMulticlassLrPredictorAsSummary()
             _step++;
             OutputPath summaryModel = CreateOutputPath("summary.txt");
             TestInCore("savepredictor", null, trainModel, null, summaryModel.Arg("sum"));
-
             Done();
```
harishsk: Nitpick. Can the blank like be added back?  Otherwise a high level view does not reflect true view of the changed areas.

codemzs:
@harishsk What is the status on this PR? It has been open for almost two months now, lets try to close on this PR soon.

harishsk:
Keren was planning to make one last change before it was ready to go in. She is at a conference this week.

---
In reply to: [537882908](https://github.com/dotnet/machinelearning/pull/4081#issuecomment-537882908) [](ancestors = 537882908)

harishsk:
@codemzs, @eerhardt, @justinormont 
I have updated the title to reflect the new changes based on your earlier suggestions. 
Can you please review the PR? 

Thanks @Lynx1820 for the recent updates and fixes.

src/Microsoft.ML.Data/Evaluators/RankingEvaluator.cs
```diff
@@ -26,21 +25,24 @@
...
+    ///  Options to control the output of the RankingEvaluator
+    /// </summary>
+    public sealed class RankingEvaluatorOptions
     {
-        public sealed class Arguments
-        {
-            [Argument(ArgumentType.AtMostOnce, HelpText = "Maximum truncation level for computing (N)DCG", ShortName = "t")]
-            public int DcgTruncationLevel = 3;
+        [Argument(ArgumentType.AtMostOnce, HelpText = "Maximum truncation level for computing (N)DCG", ShortName = "t")]
+        public int DcgTruncationLevel = 3;
```
eerhardt: These public properties should have XML doc comments

src/Microsoft.ML.Data/Evaluators/RankingEvaluator.cs
eerhardt: In general I find this pattern not following good practices. Maintaining mutable global state is hard to do right.

How large do we expect this `Double[] DiscountMap` to get in general? If it was limited to `100` before, maybe we could change this code a bit so we don't need to mutate shared state.

Instead, what if we had a method to get the `Double[] DiscountMap` like:

```C#
static double[] GetDiscountMap(int truncationLevel)
```

And inside that method we check if `truncationLevel <= some threshold (like 100)`, and if it is less than 100, then we return a cached, read-only `double[]` that was precomputed. That way most normal usages will get the cached array and won't allocate new memory. If  `truncationLevel` is bigger, than we allocate a new array, fill it up, and return it. But we don't cache it.

Thoughts?

harishsk: Let us say, two different callers set two different truncations levels of N and 2N. In the model you suggest, we will return two different arrays for each of them. But the first N entries of both those arrays would be identical and duplicated space. (In theory it is an actual savings, but it is just a bunch of floats even when the truncation limit is high. So the practical impact is minimal.)

When complex thread interactions and state is involved, I agree maintaining mutable global state is difficult. But this is a fairly simple case that is protected by a lock.  (So, in theory, it is a complex task, but right here this seems straightforward)

As I was writing this, I realized there is another approach. The value of the discount map entry is completely determined by the index of the entry (1 / Math.Log(2 + i)). So we don't need to maintain a statically computed map at all, and simply compute it dynamically each time it is accessed. We would be trading off a small gain in memory consumption for a small loss in runtime performance during evaluation. 

If we wanted to optimize that even further, we would try to maintain a sparse array where the value is calculated dynamically and cached.

The overall gains in any approach would be minimal. I think we have hit the point of diminishing returns and it is okay to leave it as is.


---
In reply to: [333065450](https://github.com/dotnet/machinelearning/pull/4081#discussion_r333065450) [](ancestors = 333065450)

eerhardt: > The overall gains in any approach would be minimal. 

I totally disagree with this statement. Following good programming practices is not a "minimal" gain IMO.

> Let us say, two different callers set two different truncations levels of N and 2N. In the model you suggest, we will return two different arrays for each of them.

This is only true when `2N > 100`, which wasn't even supported before this change. So the cases that actually would be doing this will not be the majority case.

harishsk: 100 was an arbitrarily chosen number. That argument wouldn't have been true if the previous chosen value was 10 instead of 100. Also the very idea of this fix is to support truncation levels > 100 and enable such cases in the future. 

I don't think of good programming practice as an absolute definition. There are three approaches here and there are tradeoffs with each approach.

Approach 1. Use a static class and protect state correctly with a lock.
- Pros: Uses the minimum memory and computation necessary 
- Cons: Has mutable global state (but is protected with a lock)

Approach 2: Create a new discount map for each caller that asks for a higher truncation level than current
- Pros: Eliminates mutable state
- Cons: Consumes more memory than necessary. The caller can in theory optimize this by determining the highest truncation level they will need in the lifetime of the process and set that truncation level first. But that would require the caller to know implementation specifics. 

Approach 3: Compute discount map value based on index dynamically
- Pros: No global state and consumes the least memory possible
- Cons: A minor calculation is done repeatedly during evaluation

As I mentioned above, I think we have hit the point of diminishing returns and am happy to leave it as is. But I am also looking to make progress on this PR. Since you feel strongly about this, we will make the change per your suggestion. 



---
In reply to: [333204539](https://github.com/dotnet/machinelearning/pull/4081#discussion_r333204539) [](ancestors = 333204539)

harishsk: I had another question. You proposed caching a precomputed array for a threshold of say 100. That once again special cases 100 for truncation level and the whole point of this fix was to remove that. Since you are okay with returning new arrays for all values greater than 100, why not return a new array for 100 too. 
Wouldn't that be a simpler implementation?

---
In reply to: [333234894](https://github.com/dotnet/machinelearning/pull/4081#discussion_r333234894) [](ancestors = 333234894,333204539)

eerhardt: It would be simpler, yes. But the reason I proposed it is to address your cons with Approach 2 above - `Consumes more memory than necessary`.  In my proposed approach, the caller will never need to care. They get the `truncationLevel` passed in through the Arguments class, and they just call `GetDiscountMap(truncationLevel)` when they need the DiscountMap. Inside of the `GetDiscountMap` it checks if the cached, pre-computed array is large enough to handle the request. If it is large enough (i.e. `truncationLevel <= 100`) it returns the cached array. If it is not large enough (`truncationLevel > 100`, which will barely ever happen), then it calculates a new array and hands it out.

> Since you are okay with returning new arrays for all values greater than 100, why not return a new array for 100 too.

The reason is because in the majority of the use cases, `100` will work. Even in the issue being addressed #3993, it says:

> While I'm unsure what the MaxTruncationLevel value should be, I have seen on a ranking contest\example on Kaggle.com where one contest was measuring NDCG with a truncation level of up to 38.

And `38` is a lot less than `100`.

harishsk:
@eric.erhardt@microsoft.com  Please Let me know if any further changes are necessary.

---
In reply to: [539703869](https://github.com/dotnet/machinelearning/pull/4081#issuecomment-539703869) [](ancestors = 539703869)

src/Microsoft.ML.Data/Evaluators/RankingEvaluator.cs
```diff
@@ -948,41 +953,41 @@ private protected override IEnumerable<string> GetPerInstanceColumnsToSave(RoleM
...
+        private static Double[] _discountMapFixed;
 
-        public static Double[] DiscountMap
+        private static Double[] GetDiscountMapCore(int truncationLevel)
         {
-            get
-            {
-                return _discountMap;
-            }
+            var discountMap = new Double[FixedDiscountMapSize];
```
eerhardt: This should use the truncationLevel passed into the function.

harishsk: Oops. Sorry. Fixed with the latest checkin. 

---
In reply to: [333276567](https://github.com/dotnet/machinelearning/pull/4081#discussion_r333276567) [](ancestors = 333276567)

src/Microsoft.ML.Data/Evaluators/RankingEvaluator.cs
```diff
@@ -948,41 +953,41 @@ private protected override IEnumerable<string> GetPerInstanceColumnsToSave(RoleM
...
-                        discountMap[i] = 1 / Math.Log(2 + i);
-                    }
-                    _discountMap = discountMap;
-                }
+        public static Double[] GetDiscountMap(int truncationLevel)
+        {
+            var discountMap = _discountMapFixed;
+            if (discountMap == null)
+            {
+                discountMap = GetDiscountMapCore(truncationLevel);
```
eerhardt: First check if the requested level is small enough for the “fixed” map, if it is small enough, use the cached one. If the cached one hasn’t been created yet, then create it using ‘FixedDiscountMapSize’.

PR #4091, opened 2019-08-07
Updating DatabaseLoader to support getting column info from a given .NET type.

tannergooding:
This adds limited support for determining the column info from a .NET type in order to match what TextLoader provides.

### DIFF
--- a/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
+++ b/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
@@ -123,18 +123,18 @@ namespace Microsoft.ML.Data
             for (int index = 0; index < memberInfos.Length; index++)
             {
                 var memberInfo = memberInfos[index];
-                var mappingAttr = memberInfo.GetCustomAttribute<LoadColumnAttribute>();
-
-                if (mappingAttr == null)
-                    throw host.Except($"{(memberInfo is FieldInfo ? "Field" : "Property")} '{memberInfo.Name}' is missing the {nameof(LoadColumnAttribute)} attribute");
-
                 var mappingAttrName = memberInfo.GetCustomAttribute<ColumnNameAttribute>();
 
                 var column = new Column();
                 column.Name = mappingAttrName?.Name ?? memberInfo.Name;
 
-                var source = mappingAttr.Sources.Select((source) => Range.FromTextLoaderRange(source)).ToArray();
-                column.Source = source.Single().Min;
+                var mappingAttr = memberInfo.GetCustomAttribute<LoadColumnAttribute>();
+
+                if (mappingAttr is object)
+                {
+                    var source = mappingAttr.Sources.Select((source) => Range.FromTextLoaderRange(source)).ToArray();
+                    column.Source = source.Single().Min;
+                }
 
                 InternalDataKind dk;
                 switch (memberInfo)
@@ -160,7 +160,8 @@ namespace Microsoft.ML.Data
                 columns.Add(column);
             }
 
-            var options = new Options {
+            var options = new Options
+            {
                 Columns = columns.ToArray()
             };
             return new DatabaseLoader(host, options);
@@ -244,29 +245,19 @@ namespace Microsoft.ML.Data
             /// The maximum index of the column, inclusive. If <see langword="null"/>
             /// indicates that the <see cref="TextLoader"/> should auto-detect the legnth
             /// of the lines, and read untill the end.
-            /// If max is specified, the fields <see cref="AutoEnd"/> and <see cref="VariableEnd"/> are ignored.
+            /// If <see cref="Max"/> is specified, the field <see cref="AutoEnd"/> is ignored.
             /// </summary>
             [Argument(ArgumentType.AtMostOnce, HelpText = "Last index in the range")]
             public int? Max;
 
             /// <summary>
             /// Whether this range extends to the end of the line, but should be a fixed number of items.
-            /// If <see cref="Max"/> is specified, the fields <see cref="AutoEnd"/> and <see cref="VariableEnd"/> are ignored.
+            /// If <see cref="Max"/> is specified, the field <see cref="AutoEnd"/> is ignored.
             /// </summary>
             [Argument(ArgumentType.AtMostOnce,
                 HelpText = "This range extends to the end of the line, but should be a fixed number of items",
                 ShortName = "auto")]
             public bool AutoEnd;
-
-            /// <summary>
-            /// Whether this range extends to the end of the line, which can vary from line to line.
-            /// If <see cref="Max"/> is specified, the fields <see cref="AutoEnd"/> and <see cref="VariableEnd"/> are ignored.
-            /// If <see cref="AutoEnd"/> is <see langword="true"/>, then <see cref="VariableEnd"/> is ignored.
-            /// </summary>
-            [Argument(ArgumentType.AtMostOnce,
-                HelpText = "This range extends to the end of the line, which can vary from line to line",
-                ShortName = "var")]
-            public bool VariableEnd;
 
             /// <summary>
             /// Whether this range includes only other indices not specified.
--- a/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
+++ b/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
@@ -132,8 +132,8 @@ namespace Microsoft.ML.Data
 
                 if (mappingAttr is object)
                 {
-                    var source = mappingAttr.Sources.Select((source) => Range.FromTextLoaderRange(source)).ToArray();
-                    column.Source = source.Single().Min;
+                    var sources = mappingAttr.Sources.Select((source) => Range.FromTextLoaderRange(source)).ToArray();
+                    column.Source = sources.Single().Min;
                 }
 
                 InternalDataKind dk;
--- a/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
+++ b/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
@@ -5,7 +5,11 @@
 using System;
 using System.Collections.Generic;
 using System.Data;
+using System.Data.Common;
 using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Text;
 using Microsoft.ML;
 using Microsoft.ML.CommandLine;
 using Microsoft.ML.Data;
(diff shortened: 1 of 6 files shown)


src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
```diff
@@ -98,6 +102,70 @@ void ICanSaveModel.Save(ModelSaveContext ctx)
...
+
+            var propertyInfos =
+                userType
+                .GetProperties(BindingFlags.Public | BindingFlags.Instance)
+                .Where(x => x.CanRead && x.GetGetMethod() != null && x.GetIndexParameters().Length == 0);
+
+            var memberInfos = (fieldInfos as IEnumerable<MemberInfo>).Concat(propertyInfos).ToArray();
+
+            if (memberInfos.Length == 0)
+                throw host.ExceptParam(nameof(TInput), $"Should define at least one public, readable field or property in {nameof(TInput)}.");
```
eerhardt: I wonder if this is a place we can deviate from TextLoader. The user shouldn’t be required to give a column index/order here. Instead, we can look the column index up given the column name. We already have support for that in the DatabaseLoader. We should simply be able to remove this exception and let that behavior kick in.

eerhardt: See 

https://github.com/dotnet/machinelearning/blob/e389752920ef0f58c5d3f6890d3609f7f8510df1/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoaderCursor.cs#L314-317

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
```diff
@@ -98,6 +102,70 @@ void ICanSaveModel.Save(ModelSaveContext ctx)
...
+
+            var columns = new List<Column>();
+
+            for (int index = 0; index < memberInfos.Length; index++)
+            {
+                var memberInfo = memberInfos[index];
+                var mappingAttr = memberInfo.GetCustomAttribute<LoadColumnAttribute>();
+
+                if (mappingAttr == null)
+                    throw host.Except($"{(memberInfo is FieldInfo ? "Field" : "Property")} '{memberInfo.Name}' is missing the {nameof(LoadColumnAttribute)} attribute");
```
eerhardt: I wonder if this is a place we can deviate from TextLoader. The user shouldn’t be required to give a column index/order here. Instead, we can look the column index up given the column name. We already have support for that in the DatabaseLoader. We should simply be able to remove this exception and let that behavior kick in. 

See [https://github.com/dotnet/machinelearning/blob/e389752920ef0f58c5d3f6890d3609f7f8510df1/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoaderCursor.cs#L314-L317](https://github.com/dotnet/machinelearning/blob/e389752920ef0f58c5d3f6890d3609f7f8510df1/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoaderCursor.cs#L314-L317)

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
```diff
@@ -128,6 +196,96 @@ public sealed class Column
...
+
+            /// <summary>
+            /// Whether this range extends to the end of the line, which can vary from line to line.
+            /// If <see cref="Max"/> is specified, the fields <see cref="AutoEnd"/> and <see cref="VariableEnd"/> are ignored.
+            /// If <see cref="AutoEnd"/> is <see langword="true"/>, then <see cref="VariableEnd"/> is ignored.
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce,
+                HelpText = "This range extends to the end of the line, which can vary from line to line",
+                ShortName = "var")]
+            public bool VariableEnd;
```
eerhardt: I think we should probably leave these options off until we are sure we will need them (or actually use them). I'm not certain `VariableEnd` makes sense for a DatabaseLoader.

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
```diff
@@ -137,45 +128,42 @@ public class IrisPrediction
         }
     }
 
-    internal sealed class MockProviderFactory : DbProviderFactory
+    internal class MockProviderFactory : DbProviderFactory
```
eerhardt: Why remove the `sealed`?

src/Microsoft.ML.Experimental/DataLoadSave/Database/DbExtensions.cs
```diff
@@ -66,5 +66,82 @@ public static Type ToType(this DbType dbType)
...
+                {
+                    return DbType.String;
+                }
+
+                case InternalDataKind.BL:
+                {
+                    return DbType.Boolean;
+                }
+
+                case InternalDataKind.DT:
```
eerhardt: What about DateTimeOffset (InternalDataKind.DZ) ?

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoaderCatalog.cs
```diff
@@ -11,20 +11,31 @@ namespace Microsoft.ML
...
-        /// </summary>
+        /// <summary>Create a database loader <see cref="DatabaseLoader"/>.</summary>
         /// <param name="catalog">The <see cref="DataOperationsCatalog"/> catalog.</param>
         /// <param name="columns">Array of columns <see cref="DatabaseLoader.Column"/> defining the schema.</param>
         public static DatabaseLoader CreateDatabaseLoader(this DataOperationsCatalog catalog,
             params DatabaseLoader.Column[] columns)
         {
-            var options = new DatabaseLoader.Options
-            {
+            var options = new DatabaseLoader.Options {
```
eerhardt: (nit) Allman style says braces should go on a new line. Reformatting this code goes against the style guideline.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
```diff
@@ -98,6 +102,70 @@ void ICanSaveModel.Save(ModelSaveContext ctx)
...
+                        Contracts.Assert(false);
+                        throw Contracts.ExceptNotSupp("Expected a FieldInfo or a PropertyInfo");
+                }
+
+                column.Type = dk.ToDbType();
+
+                columns.Add(column);
+            }
+
+            var options = new Options {
```
eerhardt: (nit) braces on a new line.

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
```diff
@@ -79,18 +70,9 @@ public void IrisSdcaMaximumEntropy()
...
-            {
-                new DatabaseLoader.Column() { Name = "Label", Type = DbType.Int32 },
-                new DatabaseLoader.Column() { Name = "SepalLength", Type = DbType.Single },
-                new DatabaseLoader.Column() { Name = "SepalWidth", Type = DbType.Single },
-                new DatabaseLoader.Column() { Name = "PetalLength", Type = DbType.Single },
-                new DatabaseLoader.Column() { Name = "PetalWidth", Type = DbType.Single }
-            };
+            var loader = mlContext.Data.CreateDatabaseLoader<IrisData>();
 
-            var loader = mlContext.Data.CreateDatabaseLoader(loaderColumns);
```
eerhardt: Should we have at least one test covering the other CreateDatabaseLoader public methods instead of moving them all to `CreateDatabaseLoader<T>()`?

src/Microsoft.ML.Experimental/DataLoadSave/Database/DbExtensions.cs
tannergooding: There is no equivalent for DbKind

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoaderCatalog.cs
tannergooding: This was my IDE, will fix. The .editorconfig should be updated to appropriately preference this.

It is not uncommon for object initializers and lambdas to use K&R while the rest of the code uses allman.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DbExtensions.cs
eerhardt: Even `DbType.DateTimeOffset` and `DbType.Time`?

https://docs.microsoft.com/en-us/dotnet/api/system.data.dbtype

tannergooding: The latter is the SQL time type, not an equivalent to the `System.TimeSpan` type; I'm not sure what the appropriate conversion is. I missed `DateTimeOffset`.

eerhardt: > The latter is the SQL time type, not an equivalent to the System.TimeSpan type;

According to https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql-server-data-type-mappings, it is the appropriate mapping:


SQL Server Database Engine type | .NET Framework type | SqlDbType enumeration | SqlDataReader SqlTypes typed accessor | DbType enumeration | SqlDataReader DbType typed accessor
-- | -- | -- | -- | -- | --
time(SQL Server 2008 and later) | TimeSpan | Time | none | Time | GetDateTime

tannergooding: Is that also the appropriate mapping for all database types, not just SQL?

Also, it looks like it isn't even the SQL `time` type, it is the `SQL` `datetime` type and there is a separate value in a separate enum that should be used for `SQL` `time`:

> | Time | 17 | A type representing a SQL Server DateTime value. If you want to use a SQL Server time value, use [Time](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqldbtype?view=netframework-4.8#System_Data_SqlDbType_Time). |

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
tannergooding: Removed.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
tannergooding: Updated.

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
tannergooding: Mistake. Fixed.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DbExtensions.cs
tannergooding: ah, right.
These weren't done because `DbDataReader` doesn't expose `Get` methods for them

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoaderCatalog.cs
tannergooding: Fixed.

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
tannergooding: Not sure it is worthwhile. `CreateDatabaseLoader<T>` just calls the non-generic version, so it is being tested regardless.

eerhardt: Having at least 1 test for a public API is always worthwhile IMO. It ensures people don't break the public API in the future.

tannergooding: Reverted `DatabaseLoaderTests.IrisLightGbm` to manually creating the columns.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
```diff
@@ -128,6 +197,86 @@ public sealed class Column
...
+            /// The maximum index of the column, inclusive. If <see langword="null"/>
+            /// indicates that the <see cref="TextLoader"/> should auto-detect the legnth
+            /// of the lines, and read untill the end.
+            /// If <see cref="Max"/> is specified, the field <see cref="AutoEnd"/> is ignored.
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Last index in the range")]
+            public int? Max;
+
+            /// <summary>
+            /// Whether this range extends to the end of the line, but should be a fixed number of items.
```
eerhardt: What does "to the end of the line" mean to a database loader?

I think we should remove these options until we decide we have data that says they are required. That way they don't "sneak" in to the product, and don't do anything.

tannergooding: In this case it would mean to the end of the table. I'm pretty sure it is used for vector support.

tannergooding:
@eerhardt, any other feedback here or can it be merged and I can start on getting the vector support up?

eerhardt:
> @eerhardt, any other feedback here or can it be merged

I don't believe that [this feedback](https://github.com/dotnet/machinelearning/pull/4091#discussion_r313057094) has been addressed:

> I think we should remove these options until we decide we have data that says they are required. 

We shouldn't be adding public options that aren't being used. You can easily add these new options if and when they get supported.

tannergooding:
> We shouldn't be adding public options that aren't being used. You can easily add these new options if and when they get supported.

They are going to be used in the next PR which adds the vector support. Is it really worthwhile to pull it out and retrigger CI just for it to be used and re-added in the following PR?

PR #4096, opened 2019-08-08
Enabling building ML.NET for .NET Core 3.0 in the official builds.

eerhardt:
This allows for the [C# hardware intrinsics code written last summer](https://github.com/dotnet/machinelearning/pulls?q=is%3Apr+author%3Abriancylui) to be used in official builds.

It also allows us to take advantage of other new APIs in .NET Core 3.0 in the future.


init-tools.cmd
```diff
@@ -46,25 +46,34 @@ if exist "%DotNetBuildToolsDir%" (
...
+set DOTNET_EXTRA_RUNTIME_LOCAL_PATH=%DOTNET_PATH%%DOTNET_EXTRA_RUNTIME_ZIP_NAME%
+echo Installing '%DOTNET_EXTRA_RUNTIME_REMOTE_PATH%' to '%DOTNET_EXTRA_RUNTIME_LOCAL_PATH%' >> "%INIT_TOOLS_LOG%"
+powershell -NoProfile -ExecutionPolicy unrestricted -Command "$retryCount = 0; $success = $false; $proxyCredentialsRequired = $false; do { try { $wc = New-Object Net.WebClient; if ($proxyCredentialsRequired) { [Net.WebRequest]::DefaultWebProxy.Credentials = [Net.CredentialCache]::DefaultNetworkCredentials; } $wc.DownloadFile('%DOTNET_EXTRA_RUNTIME_REMOTE_PATH%', '%DOTNET_EXTRA_RUNTIME_LOCAL_PATH%'); $success = $true; } catch { if ($retryCount -ge 6) { throw; } else { $we = $_.Exception.InnerException -as [Net.WebException]; $proxyCredentialsRequired = ($we -ne $null -and ([Net.HttpWebResponse]$we.Response).StatusCode -eq [Net.HttpStatusCode]::ProxyAuthenticationRequired); Start-Sleep -Seconds (5 * $retryCount); $retryCount++; } } } while ($success -eq $false); Expand-Archive '%DOTNET_EXTRA_RUNTIME_LOCAL_PATH%' '%DOTNET_PATH%';" >> "%INIT_TOOLS_LOG%"
```
safern: This is kind of hard to read, would it make sense to move both commands this one and the one to install the cli to a PS1 and just call that guy?

src/Native/build.cmd
```diff
@@ -18,10 +18,10 @@ set MKL_LIB_PATH=""
 :Arg_Loop
 if [%1] == [] goto :ToolsVersion
 if /i [%1] == [Release]     ( set CMAKE_BUILD_TYPE=Release&&shift&goto Arg_Loop)
-if /i [%1] == [Release-Intrinsics]     ( set CMAKE_BUILD_TYPE=Release-Intrinsics&&shift&goto Arg_Loop)
+if /i [%1] == [Release-netcoreapp3_0]     ( set CMAKE_BUILD_TYPE=Release-netcoreapp3_0&&shift&goto Arg_Loop)
```
safern: Do we really need 1 build for each configuration? Or can we just use the multi targets and do it all in one build?

eerhardt: We build the same product assemblies in each configuration.

But the different configurations change which TFM we run our tests for. We want to run the tests in parallel across the different machines in order to keep CI times down.

init-tools.cmd
eerhardt: 100% agreed this should be changed. But this will all go away when this repo moves to dotnet/arcade, so I didn't want to spend too much time re-writing this installation logic.

PR #4110, opened 2019-08-13
Load TransformerChain model file in ModelFileUtils

### LINKED ISSUE #4109: Enable loading TransformerChain models in EntryPoint graphs
Related to microsoft/NimbusML#201

The new `TransformerChain` model format cannot be used to score a dataset in NimbusML. This is because this model format cannot be loaded by `TransformModelImpl`, which is a field of `PredictorModelImpl`, which in turn is used in constructing the inputs for the entry point  graph passed by NimbusML.

 cc: @eerhardt @ganik

najeeb-kazmi:
Fixes #4109 

Related to microsoft/NimbusML#201

Enables loading TransformerChain models from entrypoints. This is needed so that models trained in ML.NET can be scored in NimbusML.

NimbusML models could be loaded and scored in ML.NET, but we never tested it the other way around.

### DIFF
--- a/src/Microsoft.ML.Data/Utilities/ModelFileUtils.cs
+++ b/src/Microsoft.ML.Data/Utilities/ModelFileUtils.cs
@@ -25,6 +25,8 @@ namespace Microsoft.ML.Model
     {
         public const string DirPredictor = "Predictor";
         public const string DirDataLoaderModel = "DataLoaderModel";
+        public const string DirTransformerChain = "TransformerChain";
+        public const string SchemaEntryName = "Schema";
         // ResultsProcessor needs access to this constant.
         public const string DirTrainingInfo = "TrainingInfo";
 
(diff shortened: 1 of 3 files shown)


eerhardt:
Can you add a test for this scenario?

src/Microsoft.ML.Data/Utilities/ModelFileUtils.cs
```diff
@@ -25,6 +25,8 @@ internal static class ModelFileUtils
     {
         public const string DirPredictor = "Predictor";
         public const string DirDataLoaderModel = "DataLoaderModel";
+        public const string DirTransformerChain = "TransformerChain";
```
eerhardt: (minor suggestion)

Instead of hard-coding these strings here, can you use/reference the existing const strings for these?

```
private const string DirTransformerChain = TransformerChain.LoaderSignature;
private const string SchemaEntryName =ModelOperationsCatalog.SchemaEntryName; 
```
Note that you will need to make ModelOperationsCatalog.SchemaEntryName internal, since it is private right now.

test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -612,6 +612,94 @@ public void EntryPointExecGraphCommand()
...
+                    'output_data': '$output_data'
+                  }}
+                }}", EscapePath(dataPath), EscapePath(modelPath));
+
+            var jsonPath = DeleteOutputPath("graph.json");
+            File.WriteAllLines(jsonPath, new[] { inputGraph });
+
+            var args = new ExecuteGraphCommand.Arguments() { GraphPath = jsonPath };
+            var cmd = new ExecuteGraphCommand(Env, args);
+            cmd.Run();
```
eerhardt: Can you verify the output is expected?

PR #4115, opened 2019-08-14
Update to .NET Core 3.0 preview9

eerhardt:
.NET Core preview 8 shipped yesterday. We should use it in our CI to ensure ML.NET runs correctly on the latest bits.


Directory.Build.props
```diff
@@ -29,7 +29,7 @@
     <PkgDir>$(RepoRoot)pkg/</PkgDir>
 
     <!-- Output directories -->
-    <BinDir Condition="'$(BinDir)'==''">$(RepoRoot)bin/</BinDir>
+    <BinDir Condition="'$(BinDir)'==''">$(RepoRoot)bin\</BinDir>
```
safern: Instead use NormalizeDirectory which normalizes the path for the host machine OS.

i.e:
```xml
$([MSBuild]::NormalizeDirectory('$(RepoRoot)', 'bin'))
```
that will add the correct trailing slash.

eerhardt: Good call. I'm not sure I knew about that function.
Fixed.

PR #4117, opened 2019-08-15
Rename LabelColumn to LabelColumnName for MutualInformationSelector

ganik:
To follow existing practice and match all other transforms need to rename LabelColumn to LabelColumnName. Currently in NimbusML parsing logic needs to handle LabelColumn for MI, while for rest of transformers form manifest its LabelColumnName

### DIFF
--- a/src/Microsoft.ML.Transforms/MutualInformationFeatureSelection.cs
+++ b/src/Microsoft.ML.Transforms/MutualInformationFeatureSelection.cs
@@ -80,7 +80,7 @@ namespace Microsoft.ML.Transforms
         [BestFriend]
         internal static class Defaults
         {
-            public const string LabelColumn = DefaultColumnNames.Label;
+            public const string LabelColumnName = DefaultColumnNames.Label;
             public const int SlotsInOutput = 1000;
             public const int NumBins = 256;
         }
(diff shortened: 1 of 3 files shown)


src/Microsoft.ML.Transforms/MutualInformationFeatureSelection.cs
```diff
@@ -105,13 +105,13 @@ internal sealed class Options : TransformInputBase
...
         private readonly (string outputColumnName, string inputColumnName)[] _columns;
-        private readonly string _labelColumn;
+        private readonly string _labelColumnName;
         private readonly int _slotsInOutput;
         private readonly int _numBins;
 
         /// <include file='doc.xml' path='doc/members/member[@name="MutualInformationFeatureSelection"]/*' />
         /// <param name="env">The environment to use.</param>
-        /// <param name="labelColumn">Name of the column to use for labels.</param>
+        /// <param name="labelColumnName">Name of the column to use for labels.</param>
```
justinormont: Will this be a breaking change for ML.NET users?

ganik: It will, re-compilation will do the trick. Already trained models will be fine. This is a missed change from previous push where LabelColumn was renamed LabelColumnName, WeightColumn was renamed to WeightColumnName and so on.Currently it acts as a technical debt in the sense that it needs a special treatment in NimbusML

---
In reply to: [314471870](https://github.com/dotnet/machinelearning/pull/4117#discussion_r314471870) [](ancestors = 314471870)

eerhardt: This is an internal API, so it isn't breaking external users.

justinormont: Awesome. Thanks for checking.

ganik: could one of you sign off :)

PR #4127, opened 2019-08-18
Full Model Retrain Sample 

sierralee51:
A sample using the full model retrain API with the MNIST model and data.


sierralee51:
working on the hard-coded parts that I didn't notice before right now

CESARDELATORRE:
As part of this PR, could you create a full-retrain sample using a more complex image dataset like the one we have in the sample below? Basically, photos of multiple classes, then measuring the real accuracy when performing predictions, etc. 

You could evolve that transfer-learning sample bewlow into a new sample using full-retrain: 

https://github.com/dotnet/machinelearning-samples/blob/migration/v1.3/samples/csharp/getting-started/DeepLearning_TensorFlow_TransferLearning/ImageClassification.Train/Program.cs

Issue #4134, opened 2019-08-20
Memory leak in the latest ML.NET 1.3.1


SnakyBeaky:
### System information

- **OS version/distro**: Windows 10 Pro 10.0.17763
- **.NET Version (eg., dotnet --info)**:  .NET Core 2.2.203 e5bab63eca

### Issue

- **What did you do?** Update the NuGet references to Microsoft.ML.* from `1.2.0` to `1.3.1`.
- **What happened?** Memory leaking
- **What did you expect?** Same behaviour as before

### Source code / logs

The pipeline consists of a ValueToKey mapping, loading image, resizing and extracting pixels, input them into an tensorflow inception model and finally a LbfgsMaximumEntropy trainer.

This is a run training several models in `1.2.0`:
![image](https://user-images.githubusercontent.com/8410877/63354071-0e906200-c364-11e9-8f98-e2a386efe7f9.png)

The same exact code, running in `1.3.1`:
![image](https://user-images.githubusercontent.com/8410877/63354132-2a940380-c364-11e9-86be-1193457a6caa.png)

Object count and heap size are more or less stable, my guess is something related to the tensorflow model not being cleared up.
![image](https://user-images.githubusercontent.com/8410877/63354150-3253a800-c364-11e9-98eb-9fe0428c8528.png)

codemzs:
Thanks! Looking at this.

Issue #4087, opened 2019-08-07
[ImageClassification Transfer Learning] Simplify/Hide steps and hyper-parameters in new ImageClassification - Transfer Learning 


CESARDELATORRE:
Since the new (early preview) 'ImageClassification - Transfer Learning' is aimed to provide a high-level API per scenario (image classifciation in this case), we should simplify the API even further.

Here's an example of a current training pipeline:

```
 var pipeline = mlContext.Transforms.Conversion.MapValueToKey("Label")
                    .Append(mlContext.Transforms.LoadImages("ImageObject", null,
                        "ImagePath"))
                    .Append(mlContext.Transforms.ResizeImages("Image",
                        inputColumnName: "ImageObject", imageWidth: 299,
                        imageHeight: 299))
                    .Append(mlContext.Transforms.ExtractPixels("Image",
                        interleavePixelColors: true))
                    .Append(mlContext.Model.ImageClassification("Image",
                        "Label", arch: DnnEstimator.Architecture.InceptionV3, 
                        epoch: 1, //an epoch is one learning cycle where the learner sees the whole training data set.
                        batchSize: 20));  // batchSize sets then number of images to feed the model at a time
```

Good points to simplify are image transformations and "technical" hyper-paramenters:

**1. Hide _image re-size_ step:** Since the image size is determined by the internal DNN model (TensorFlow model, currently) why don't we simply do it within the `ImageClassification` code?

**2. Hide extract pixels step:** Same thing here. The extract pixels step should be hidden as part of the `ImageClassification` code. This is surfacing too much details. I expect the interleavePixelColors could also be determined by the info in the TF model .pb file, ot in the worst case based on the Architecture metadata that we could have per architecture (InceptionV3, etc.).
 
**3. Hide EPOCH and BATCHSIZE:** These are also configuration details that, even when they currently have "by default values" (epoch=10 and batchSize=20), those values are not valid for *any* imageset and architecture and should vary depending on the specific image-set size and chosen architecture. I'd be good to have them initialized dynamically within the `ImageClassification` code depending on the context instead of "fixed default values" and parameters that surface to the user who might not know what is an EPOCH or BATCH.

After simplifying the above points, the example code could look like the following, which looks a lot simpler and clean: :)

```
 var pipeline = mlContext.Transforms.Conversion.MapValueToKey("Label")
                    .Append(mlContext.Transforms.LoadImages("ImageObject", null, "ImagePath"))
[...]

Issue #4084, opened 2019-08-07
[ImageClassification Transfer Learning] Higher level API supporting image predictions based on image file paths and in-memory images, alternatively


CESARDELATORRE:
This is probably more for the long-term API of our ImageClassification Transfer Learning once the foundational features for our 'ImageClassification Transfer Learning' are in place.

However, this is the point:

We aim to design and implement a high level API which will provide a very straightforward way of training (transfer learning) image classification and other capabilities such as object detection, etc. That also means, and is equally important, to be able to score/predict very easily according to normal scenarios which for images are two scenarios:

- A. Score based on a provided image file path.
- B. Score based on an in-memory image.

The second choice (B) is even more common for many apps, even more when moving to OBJECT DETECTION and LIVE VIDEO/IMAGES coming as streaming.

**CURRENT STATE IN ML.NET API**

The issue is with the current API design of ML.NET where the ML.NET model scoring API completely depends on the way you created its original pipeline and what schema you were using, meaning:

- If you trained an ML.NET model with image file paths, your ML.NET model expects an image  file path for scoring.

- If you trained an ML.NET model with in-memory images (not a common/straightforward approach), your ML.NET model expects an in-memory image (Bitmap) when scoring.  

However, that's not the typical scenario for a user where the easiest way for training is using image filepaths (that's the way you usually have image sets, right?), but then when scoring/predicting the user should be able to provide either a file path or an in-memory image (in-memory is more common for end-user apps vs. filepaths for batch processes) and anything else should be transparent for him.

**CURRENT WORKAROUND:**

Sure, you can always go deeper and deep dive in the TensorFlow .pb model that was also generated by our "ImageClassification Transfer Learning", find out the input and output tensor names, implement a different ML.NET pipeline that accepts in-memory images, create that ML.NET model by running that pipeline once (it is not really training, it is just creating the needed transformers for scoring later on) and finally writing a more specific and not very straightforward code for scoring with in-memory images.

That process is what I wrote in the second part of this BLOG POST and related sample app which is using an in-memory image coming through HTTP and was provided by the end-user for predicting its class:

https://devblogs.microsoft.com/cesardelatorre/run-with-ml-net-c-code-a-tensorflow-model-exported-from-azure-cognitive-services-custom-vision/

But that code for scoring a TF model with in-memory images is not straightforward to implement.

**INITIAL SOLUTION: Load images from files but convert them into in-memory image objects before loading it into the IDataView, so the schema would match for training and scoring**

[...]

PR #4151, opened 2019-08-28
Image classification preview 2.

### LINKED ISSUE #4090: [ImageClassification Transfer Learning] - Warning mesage: "Allocation of 553190400 exceeds 10% of system memory."
Why are we getting this warning message (I get it multiple times) when training with a small set of images?

"2019-08-07 09:49:52.642800: W tensorflow/core/framework/allocator.cc:107] Allocation of 553190400 exceeds 10% of system memory."

image

codemzs:
This change improves the in-preview image classification API further:

1. Increases DNN training speed by 10x.
2. Reduced and constant memory footprint.
3. Simplifies the API by not requiring the user to pre-process the image.
4. Introduces callback to provide metrics during training such as accuracy, cross-entropy.
5. Improved image classification sample.

fixes #4090, #4087, #4084, #4134


docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearning.cs
```diff
@@ -1,123 +0,0 @@
...
-            var pipeline = mlContext.Transforms.Conversion.MapValueToKey("Label")
-                .Append(mlContext.Transforms.LoadImages("ImageObject", null,
-                    "ImagePath"))
-                .Append(mlContext.Transforms.ResizeImages("Image",
-                    inputColumnName: "ImageObject", imageWidth: 299,
-                    imageHeight: 299))
-                .Append(mlContext.Transforms.ExtractPixels("Image",
-                    interleavePixelColors: true))
-                .Append(mlContext.Model.ImageClassification("Image",
-                    "Label", arch: DnnEstimator.Architecture.ResnetV2101, epoch: 4,
```
CESARDELATORRE: Maybe add a comment around here saying that just by changing/selecting InceptionV3 here instead of ResnetV2101 you can try a different architecture/pre-trained model. :)

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs
```diff
@@ -0,0 +1,265 @@
...
+            try
+            {
+
+                MLContext mlContext = new MLContext(seed: 1);
+
+                //Load all the original images info
+                IEnumerable<ImageData> images = LoadImagesFromDirectory(folder: fullImagesetFolderPath, useFolderNameasLabel: true);
+                IDataView shuffledFullImagesDataset = mlContext.Data.ShuffleRows(mlContext.Data.LoadFromEnumerable(images));
+                shuffledFullImagesDataset = mlContext.Transforms.Conversion.MapValueToKey("Label")
+                    .Fit(shuffledFullImagesDataset)
```
CESARDELATORRE: Let's keep it like this in any case. The only minor issue I see here is that if the user wants to remove the ValidationSet for the training pipeline and wants to simplify and do a single Fit() only for training, the changes in the code are more complex because he also has to realize that he needs to change the provided schema when saving the model. But it doesn't matter, I personally think you always will want to have the  ValidationSet for the training pipeline and see how it evolves while training... :)

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs
```diff
@@ -0,0 +1,265 @@
...
+                            batchSize: 10,
+                            learningRate: 0.01f,
+                            metricsCallback: (metrics) => Console.WriteLine(metrics),
+                            validationSet: testDataset);
+
+
+                Console.WriteLine("*** Training the image classification model with DNN Transfer Learning on top of the selected pre-trained model/architecture ***");
+
+                // Measuring training time
+                var watch = System.Diagnostics.Stopwatch.StartNew();
```
CESARDELATORRE: Measuring the time with watch makes the code a bit more verbose or difficult to read.
But to me it is useful, so we can keep it here, too. 

I'm removing it from the samples repo so the code is simpler, though.

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -85,33 +83,50 @@ public static class DnnCatalog
...
             string featuresColumnName,
             string labelColumnName,
-            string outputGraphPath = null,
             string scoreColumnName = "Score",
             string predictedLabelColumnName = "PredictedLabel",
-            string checkpointName = "_retrain_checkpoint",
-            Architecture arch = Architecture.ResnetV2101,
-            DnnFramework dnnFramework = DnnFramework.Tensorflow,
+            Architecture arch = Architecture.InceptionV3,
             int epoch = 10,
```
CESARDELATORRE: Why epoch=10 and batchSize=20 as defaults?
I find that for small/medium datasets (hundres of images) which is what initially folks are going to test, these defaults are what we were using:

epoch=100 
batchSize=10

But I'm open to different/better defaults. IT is simply that I think these defaults were defined in previous versions and we might want to change it?

Of course, I know that we cannot have good defaults for any scenario. We'd need to have some "AutoML" that would change dynamically depending on the input datasets type/size and even other hyper-parameters depending on the accuracy we're getting...

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -85,33 +83,50 @@ public static class DnnCatalog
...
             int batchSize = 20,
-            float learningRate = 0.01f)
+            float learningRate = 0.01f,
+            ImageClassificationMetricsCallback metricsCallback = null,
+            int statisticFrequency = 1,
+            DnnFramework framework = DnnFramework.Tensorflow,
```
CESARDELATORRE: I wonder if we have external hyper-parameters that only are applicable to TensorFlow but not to PyTorch?
Can you confirm that? - In any case, we'll be able to evolve it when implementing for PyTorch, let's not start building from the roof... ;) - Never mind about this comment, just an observation.

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -85,33 +83,50 @@ public static class DnnCatalog
...
+            string modelSavePath = null,
+            string checkpointName = "_retrain_checkpoint",
```
CESARDELATORRE: I think we need to change the name of the custom model generated for the user, I mean the .PB TensorFlow model file. Currently I think we have:

- `resnet_v2_101_299.meta_retrain_checkpoint.pb`
or if using Inception:
- `InceptionV3.meta_retrain_checkpoint.pb`

When the user sees those names, it'll feel like something related to an internal process (checkpoint, etc.).
Can you change it to something like the following?

- `custom.retrained.model.based.on.resnet_v2_101_299.pb`
or if using InceptionV3:
- `custom.retrained.model.based.on.InceptionV3.pb`

I'm open to comparable names, but using "checkpoint" will feel weird for a user. Feels like it is a checkpoint but still not finalized..

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -85,33 +83,50 @@ public static class DnnCatalog
...
+            IDataView validationSet = null,
+            bool testOnTrainSet = true,
+            bool reuseTrainSetBottleneckCachedValues = false,
```
CESARDELATORRE: Good, cahce defaults = false. 👍

docs/samples/Microsoft.ML.Samples/Program.cs
```diff
@@ -10,7 +10,7 @@ public static class Program
 
         internal static void RunAll()
         {
-            int samples = 0;
+            /*int samples = 0;
```
CESARDELATORRE: Why is this code commented? 
Shall we either remove it or keep it un-commented?

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -0,0 +1,1224 @@
...
+        private void CheckTrainingParameters(ImageClassificationEstimator.Options options)
+        {
+            Host.CheckNonWhiteSpace(options.LabelColumn, nameof(options.LabelColumn));
+            Host.CheckNonWhiteSpace(options.TensorFlowLabel, nameof(options.TensorFlowLabel));
+
+            if (_session.graph.OperationByName(options.TensorFlowLabel) == null)
+                throw Host.ExceptParam(nameof(options.TensorFlowLabel), $"'{options.TensorFlowLabel}' does not exist in the model");
+        }
+
+        private (Tensor, Tensor) AddJpegDecoding(int height, int width, int depth)
```
CESARDELATORRE: We'll need to support additional image formats, this is only for JPEG..

docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
```diff
@@ -71,6 +71,7 @@
       <DependentUpon>LbfgsLogisticRegressionWithOptions.tt</DependentUpon>
     </None>
     <PackageReference Include="Microsoft.ML.TensorFlow.Redist" Version="0.10.0" />
+    <PackageReference Include="SharpZipLib" Version="1.2.0" />
```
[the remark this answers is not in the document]
CESARDELATORRE: It's used only in the DNN ResNet sample to unzip the sample dataset, after downloading the .zip of the dataset. IT's used by a utility class that's why you might not see it used directly.

test/Microsoft.ML.Tests/ImageClassification.cs
```diff
@@ -0,0 +1,224 @@
...
+
+        public static IEnumerable<ImageData> LoadImagesFromDirectory(string folder,
+            bool useFolderNameasLabel = true)
+        {
+            var files = Directory.GetFiles(folder, "*",
+                searchOption: SearchOption.AllDirectories);
+
+            foreach (var file in files)
+            {
+                if (Path.GetExtension(file) != ".jpg")
```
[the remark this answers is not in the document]
CESARDELATORRE: I opened and issue for that:
https://github.com/dotnet/machinelearning/issues/4152

This is Zeeshan's answer: "The current tf.net version we are using does not seem to have bindings for all image types. But the latest version seems to have we will upgrade to them latest next month"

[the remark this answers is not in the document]
Nucs: We are planning to release the new TF.NET version in a day or two.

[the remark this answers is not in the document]
codemzs: @eerhardt @CESARDELATORRE The above answers your question :)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -0,0 +1,1224 @@
...
+        {
+            var labelColumn = input.Schema[labelColumnName];
+            var imagePathColumn = input.Schema[imagepathColumnName];
+            Runner runner = new Runner(_session);
+            runner.AddOutputs(outputTensorName);
+
+            using (TextWriter writer = File.CreateText(cacheFilePath))
+            using (var cursor = input.GetRowCursor(input.Schema.Where(c => c.Index == labelColumn.Index || c.Index == imagePathColumn.Index)))
+            {
+                var labelGetter = cursor.GetGetter<uint>(labelColumn);
```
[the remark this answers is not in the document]
codemzs: spoke offline, good to support all types but for now just throw a meaningful exception. UInt32 is default for value type.

---
In reply to: [319128148](https://github.com/dotnet/machinelearning/pull/4151#discussion_r319128148) [](ancestors = 319128148)

test/Microsoft.ML.Tests/ImageClassification.cs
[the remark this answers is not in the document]
Oceania2018: Correct, we have added a lot of APIs in this [commit](https://github.com/SciSharp/TensorFlow.NET/commit/9d90d74ddd85ed4e50c77463433567005355ea0f). And we can easily add more if needed.

docs/samples/Microsoft.ML.Samples/Microsoft.ML.Samples.csproj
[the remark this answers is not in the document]
Oceania2018: Why do we still use `Microsoft.ML.TensorFlow.Redist` ? Shoud refer to 
 [SciSharp.TensorFlow.Redist](https://www.nuget.org/packages/SciSharp.TensorFlow.Redist/)

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs
```diff
@@ -0,0 +1,307 @@
...
+            string fullImagesetFolderPath = Path.Combine(
+                imagesDownloadFolderPath, finalImagesFolderName);
+
+            try
+            {
+
+                MLContext mlContext = new MLContext(seed: 1);
+
+                //Load all the original images info
+                IEnumerable<ImageData> images = LoadImagesFromDirectory(
```
CESARDELATORRE: LoadImagesFromDirectory() is a method that is almost always going to be used by developers.
I think we should consider to have this method provided by the framework.
This is also something being requested by customers.

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1113,5 +1205,183 @@ public void TensorFlowStringTest()
...
+
+        public static string GetAbsolutePath(string relativePath) =>
+            Path.Combine(new FileInfo(typeof(
+                TensorFlowScenariosTests).Assembly.Location).Directory.FullName, relativePath);
+
+
+        public class ImageData
+        {
+            [LoadColumn(0)]
+            public string ImagePath;
```
CESARDELATORRE: This input data for the dataset is using FilePaths.
We need to test it with in-memory images. Is there code using and testing in-memory images somewhere else?

PR #4155, opened 2019-08-29
Implement ONNX conversion for TypeConverting transform

### LINKED ISSUE #4004: TypeConvertingTransformer requires "Experimental" onnx version.
When exporting a NimbusML pipeline which contains a predictor to ONNX, a `TypeConvertingTransformer` is sometimes inserted in the pipeline. The `TypeConvertingTransformer` currently requires the `Experimental` ONNX flag to be set in order for it to be converted to ONNX (see line 390 in `src\Microsoft.ML.Data\Transforms\TypeConverting.cs`). Can this be updated to no longer require the `Experimental` flag?

ganik:
fixes #4004 
Tests TBD.


codemzs:
@ganik do you want this to be part of preview release?

codemzs:
@ganik Please add tests that create a graph and output json file to compare against a baseline.

harishsk:
@codemzs, Can the test come in as part of another PR? I can use this PR for my work. Can this be merged once the tests pass?

ganik:
no, I am not in hurry, will add tests 

---
In reply to: [526321188](https://github.com/dotnet/machinelearning/pull/4155#issuecomment-526321188) [](ancestors = 526321188)

codemzs:
@harishsk Please clone @ganik 's repo locally and unblock yourself. Thanks, @ganik for adding tests, they should be pretty straightforward, we have few examples that you can replicate and feel free to ask me if you need any help as I have written all of ONNX converter infrastructure and even tests ... I'm not asking for tests that also run on ORT (though that would be great), just a json file test that compares against the baseline would suffice.

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -601,6 +601,83 @@ public void WordEmbeddingsTest()
...
+            for (int i = 0; i < supportedConversions.GetLength(0); i++)
+            {
+                var fromKind = supportedConversions[i, 0];
+                var toKind = supportedConversions[i, 1];
+
+                TextLoader.Column[] columns = new []
+                {
+                    new TextLoader.Column("Value", fromKind, 0, 0)
+                };
+                var dataView = mlContext.Data.LoadFromTextFile(filePath, columns);
```
eerhardt: Do you need to use a file just to contain a single value `3`? Why not use LoadFromEnumerable, and then you can delete the file?

harishsk: I strictly don't need to save a file. I can generate the file and delete it too.
But in this particular case, the text loader simplifies somethings like automatically converting that number to a range of data types that I can iterate through in my loop by simply changing the data kind.

eerhardt: Makes sense. I guess you could do the same with properties of different types, but this works.

harishsk:
I should add a note about the last commit I submitted. 
With the previous commit, a few other onnx export tests started failing. I looked into it and they were failing because the generated onnx wasn't equal the baseline onnx file. This is related to the change in OnnxUtils.cs where a type was wrong beingy translated. So these baselines had always been wrong and the tests started failing once the bug was fixed. 
So I had to update the baseline for these with the correct type translation and also had to a cast operator for the KMeans trainer.

harishsk:
@codemzs , @eerhardt , Can one of you please review and approve?

harishsk:
@codemzs / @eerhardt  Requesting review/approval

PR #4157, opened 2019-08-29
[WIP] Transformers for AutoML

michaelgsharp:
This is still a work in progress, but will be updated and finished over the next several hours. 

In order for AutoML to take a full dependency on ML.NET, several pieces of functionality need to be added. This PR is the first step in that process adding in 3 new transformers. The actual implementation of the transformers is done in C++. This code is currently in another repo, but will be turned in a Nuget package that ML.Net can take a binary dependency on. The code going into ML.Net is mostly wrapper code dealing with data interop and, since the binary is exposed via C api's with no overloading, code to call the correct methods based on the data type.

ToStringTransformer - Takes in a column and converts it into an appropriate string representation. It currently supports all integer types, float, double, bool, and string. This is fully done, and should be reviewed first.

CategoryImputer - Fill in missing values with the most common value in the column.  It currently supports all integer types, float, double, bool, and string. 0's are treated as missing for integer types. This is done except for a small piece of the c++ interop.

DateTimeTransformer - Splits a column into its appropriate date time components (20 columns total are added). Takes a Long that represents seconds since 1970. The c++ interop will be added for this tonight and can be reviewed last.

Until we get the binary added as an official dependency, the constructors of the estimator will throw an error if the correct .dll is not found. This allows us to do a private drop for now. The tests will also be disabled currently for that reason.


src/Microsoft.ML.Transforms/CategoryImputer.cs
```diff
@@ -0,0 +1,1829 @@
+﻿using System;
```
eerhardt: (nit) copyright on all code files.

src/Microsoft.ML.Transforms/CategoryImputer.cs
```diff
@@ -0,0 +1,1829 @@
...
+            {
+                _transformerHandler = CreateTransformerFromEstimatorBase(input);
+            }
+
+            [DllImport("Featurizers", EntryPoint = "CatImputerFeaturizer_int8_t_CreateTransformerFromSavedData", CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity]
+            private static extern bool CreateTransformerFromSavedDataCpp(IntPtr data, IntPtr bufferSize, out IntPtr transformer, out IntPtr errorHandle);
+            internal override void CreateTransformerFromSavedData(byte[] data)
+            {
+                bool result;
+                GCHandle handle = GCHandle.Alloc(data, GCHandleType.Pinned);
```
eerhardt: You can just use a `fixed` statement in C#. This is slightly faster than allocing and freeing a GCHandle. See the rest of our code (like in CpuMath) where we get a pointer for an array.

Here's an example, from one of our tests:

https://github.com/dotnet/machinelearning/blob/b861b5d64841cbe0f2c866ee7586872aac450a51/test/Microsoft.ML.CpuMath.PerformanceTests/NativePerformanceTests.cs#L27

But you get the idea.

src/Microsoft.ML.Transforms/CategoryImputer.cs
```diff
@@ -0,0 +1,1824 @@
...
+
+                }
+                finally
+                {
+                    handle.Free();
+                }
+            }
+
+            [DllImport("Featurizers", EntryPoint = "CatImputerFeaturizer_uint32_t_DestroyTransformer", CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity]
+            private static extern bool DestroyTransformerCpp(IntPtr transformer
```
harishsk: Don't these estimators need to be exported to onnx? Or is that functionality coming in later? 
We should chat about who does that work.

michaelgsharp: Per offline discussion, resolving.

src/Microsoft.ML.Transforms/CategoryImputer.cs
```diff
@@ -0,0 +1,1824 @@
...
+            }
+
+            protected override bool ReleaseHandle()
+            {
+                // Not sure what to do with error stuff here.  There shoudln't ever be one though.
+                return _destroySaveDataHandler(handle, _dataSize, out IntPtr errorHandle);
+            }
+        }
+
+        [DllImport("Featurizers", EntryPoint = "GetErrorInfoString", CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity]
```
eerhardt: These new transforms should be split out of the main `Microsoft.ML` nuget package. Since you are adding them to the main nuget package, EVERY application that uses ML.NET will have a dependency on these native assemblies, even if they don't use these new transforms.

See what we did with the transforms that depend on MKL - we split them into their own nuget package, so only people who needed those components would be forced to bring MKL along with their application.

https://www.nuget.org/packages/Microsoft.ML.Mkl.Components/

ganik:
>         [DllImport("Featurizers", EntryPoint = "GetErrorInfoString"), SuppressUnmanagedCodeSecurity]

NuGet dependency should be added to use this native assembly.


---
Refers to: src/Microsoft.ML.Transforms/ToStringTransformer.cs:313 in 260ff25. [](commit_id = 260ff25a403ce796c70c59dfd5f00e26e331abda, deletion_comment = False)

ganik:
>         [DllImport("Featurizers", EntryPoint = "DestroyErrorInfoString"), SuppressUnmanagedCodeSecurity]

As I am aware the NuGet for this native lib is not ready for various Linux flavors and Mac. I dont think it can go into master branch until the NuGet has them

---
Refers to: src/Microsoft.ML.Transforms/ToStringTransformer.cs:316 in 260ff25. [](commit_id = 260ff25a403ce796c70c59dfd5f00e26e331abda, deletion_comment = False)

ganik:
>     public sealed class ToStringTransformerEstimator : IEstimator<ToStringTransformer>

In general, this ToStringTransformer could be easily added as an add-on functionality to TypeConverting transform. And TypeConverting transform has SaveAsOnnx(..) functionality already

---
Refers to: src/Microsoft.ML.Transforms/ToStringTransformer.cs:73 in 260ff25. [](commit_id = 260ff25a403ce796c70c59dfd5f00e26e331abda, deletion_comment = False)

src/Microsoft.ML.Transforms/CategoryImputer.cs
```diff
@@ -0,0 +1,1876 @@
...
+
+        internal sealed class Options: TransformInputBase
+        {
+            [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "New column definition (optional form: name:src)",
+                Name = "Column", ShortName = "col", SortOrder = 1)]
+            public Column[] Columns;
+        }
+        #endregion
+
+        // Using this to confirm DLL exists. If does it will just return false since no parameters are being passed.
```
ebarsoumMS: Why do we need that? Isn't the DLL is part of the package?

src/Microsoft.ML.Transforms/CategoryImputer.cs
```diff
@@ -0,0 +1,1876 @@
...
+        #region ColumnInfo
+
+        // REVIEW: Since we can't do overloading on the native side due to the C style exports,
+        // this was the best way I could think handle it to allow for any conversions needed based on the data type.
+
+        #region BaseClass
+
+        internal delegate bool DestroyCppTransformerEstimator(IntPtr estimator, out IntPtr errorHandle);
+        internal delegate bool DestroyTransformerSaveData(IntPtr buffer, IntPtr bufferSize, out IntPtr errorHandle);
+        internal delegate bool DestroyTransformedDataCpp(IntPtr output, IntPtr outputSize, out IntPtr errorHandle);
```
ebarsoumMS: Use Native instead of Cpp

src/Microsoft.ML.Transforms/CategoryImputer.cs
```diff
@@ -0,0 +1,1876 @@
...
+                short* interopInput = input == 0 ? null : &input;
+                var success = TransformDataCpp(_transformerHandler, interopInput, out short output, out IntPtr errorHandle);
+                if (!success)
+                    throw new Exception(GetErrorDetailsAndFreeNativeMemory(errorHandle));
+
+                return output;
+            }
+
+            public override void Dispose()
+            {
```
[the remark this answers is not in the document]
michaelgsharp: I am using a SafeHandler for all handles themselves. This is for the main transformer class itself, and it just calls dispose on any SafeHandles it has.

src/Microsoft.ML.Transforms/CategoryImputer.cs
michaelgsharp: copyright has been added. Thanks for the reminder.

src/Microsoft.ML.Transforms/CategoryImputer.cs
michaelgsharp: Changed all occurrences to `fixed`.

src/Microsoft.ML.Transforms/CategoryImputer.cs
michaelgsharp: It will be a dependency we get through nuget, but until we get the nuget up and running this is in there just as a check. Once the nuget is ready we will remove this check.

src/Microsoft.ML.Transforms/CategoryImputer.cs
michaelgsharp: Done, changed all occurrences from `Cpp` to `Native`.

pkg/Microsoft.ML.AutoMLFeaturizers/Microsoft.ML.Mkl.Redist.nupkgproj
```diff
@@ -0,0 +1,19 @@
+<Project Sdk="Microsoft.NET.Sdk" DefaultTargets="Pack">
+
+  <PropertyGroup>
+    <Authors>Intel</Authors>
```
eerhardt: You need to rename this file to be the name of the NuGet package.

Also, remove this Authors property.

michaelgsharp: I fixed the files. It is creating separate nuget packages now in the bin/packages folder. It is not creating a folder under packages in the root though. Should it be creating them there as well?

eerhardt: It's a little confusing, but `$RepoRoot\packages` are the NuGet packages that our code depends on - what are downloaded when we do a restore.  `$RepoRoot\bin\packages` are the NuGet packages that our build produces.

src/Microsoft.ML.AutoMLFeaturizers/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,10 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+using Microsoft.ML;
+
+[assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Tests" + PublicKey.TestValue)]
```
eerhardt: What internals are the tests using? I think you should be able to test everything from the public API, right?

src/Microsoft.ML.AutoMLFeaturizers/Microsoft.ML.AutoMLFeaturizers.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <IncludeInPackage>Microsoft.ML.Featurizers</IncludeInPackage>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <DefineConstants>CORECLR</DefineConstants>
```
eerhardt: I don't think this `DefineConstant` is being used in the new assembly. It can be removed.

src/Microsoft.ML.AutoMLFeaturizers/Microsoft.ML.AutoMLFeaturizers.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <IncludeInPackage>Microsoft.ML.Featurizers</IncludeInPackage>
```
eerhardt: `IncludeInPackage` needs to match the name of the NuGet package. Or else this assembly won't be included in the right NuGet package.

pkg/Microsoft.ML.AutoMLFeaturizers/Microsoft.ML.AutoMLFeaturizers.nupkgproj
```diff
@@ -0,0 +1,16 @@
...
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageDescription>ML.NET additional featurizers for AutoML</PackageDescription>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../Microsoft.ML/Microsoft.ML.nupkgproj" />
+  </ItemGroup>
+
+  <ItemGroup> 
+    <Content Include="..\common\CommonPackage.props" Pack="true" PackagePath="build\netstandard2.0\$(MSBuildProjectName).props" />
```
eerhardt: This line isn't necessary. This CommonPackage.props file probably should have a better name, but it is needed for NuGet packages that contain native assets inside of them. This one doesn't, so it is unnecessary.

pkg/Microsoft.ML.AutoMLFeaturizers/Microsoft.ML.AutoMLFeaturizers.nupkgproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk" DefaultTargets="Pack">
```
eerhardt: I wonder if `AutoMLFeaturizers` is the best name for this package. Technically, anyone can use these featurizers without using AutoML.

What is going to be the name of the underlying NuGet package that contains the native assemblies? We should probably align with that name. So if that NuGet package is going to be named "AutoMLFeaturizers", then I guess that would be a good name here as well.

src/Microsoft.ML.AutoMLFeaturizers/ToStringTransformer.cs
```diff
@@ -0,0 +1,1763 @@
...
+            public Column[] Columns;
+        }
+        #endregion
+
+        // Using this to confirm DLL exists. If does it will just return false since no parameters are being passed.
+        // Once we have a binary dependency on the dll we can remove this code.
+        [DllImport("Featurizers", EntryPoint = "GetErrorInfoString"), SuppressUnmanagedCodeSecurity]
+        private static extern bool CheckIfDllExists(IntPtr error, out IntPtr errorHandleString, out IntPtr errorHandleStringSize);
+
+        public static ToStringTransformerEstimator Create(IHostEnvironment env, string outputColumnName, string inputColumnName)
```
eerhardt: These methods shouldn't be `public`. External users should only create the estimators from the Catalog APIs.

src/Microsoft.ML.AutoMLFeaturizers/ToStringTransformer.cs
```diff
@@ -0,0 +1,1763 @@
...
+            }
+
+            protected override bool ReleaseHandle()
+            {
+                // Not sure what to do with error stuff here.  There shoudln't ever be one though.
+                return DestroyTransformerSaveDataNative(handle, _dataSize, out IntPtr errorHandle);
+            }
+        }
+
+        internal class TransformedDataSafeHandler : SafeHandleZeroOrMinusOneIsInvalid
```
eerhardt: (minor) The .NET type is a `SafeHandle`, not `SafeHandler`. I think we should drop the `r` here and just be `TransformedDataSafeHandle`.

pkg/Microsoft.ML.AutoMLFeaturizers/Microsoft.ML.Mkl.Redist.nupkgproj
michaelgsharp: Ok, that makes sense. I have the `$RepoRoot\bin\packages` creating the NuGet package correctly and tested it with a local import into another solution. So we should be good then. Thanks for the instructions!

src/Microsoft.ML.Transforms/Properties/Resources.Designer.cs
```diff
@@ -8,7 +8,7 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
 
-namespace Microsoft.ML.Transforms.Properties
+namespace Microsoft.ML.AutoMLFeaturizers.Properties
```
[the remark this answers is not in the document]
michaelgsharp: Nope, good catch. Reverting that in next push.

src/Microsoft.ML.AutoMLFeaturizers/Properties/AssemblyInfo.cs
michaelgsharp: I was testing the constructor that took in the `Options` class, but looking at the other Transformers and their tests they are going about it differently. Let me fix it and remove that `VisibleTo` tag.

src/Microsoft.ML.AutoMLFeaturizers/Microsoft.ML.AutoMLFeaturizers.csproj
michaelgsharp: I am not sure why it is showing up like that. Last night I pushed a fix that had that change to the correct name. It looks like somehow the branch in my repo and this PR aren't syncing up correctly. The number of commits this PR shows is not the same as the number of commits in my branch itself? I will rebase and see if that fixes the issue.

src/Microsoft.ML.AutoMLFeaturizers/Microsoft.ML.AutoMLFeaturizers.csproj
michaelgsharp: Same comment as the prior one. This was removed last night, but is not showing up in the PR for some reason even though it is in my branch. Will rebase and check again.

src/Microsoft.ML.AutoMLFeaturizers/Properties/AssemblyInfo.cs
michaelgsharp: So I just double checked, and I was testing the entry point that AutoML uses and it takes in the `Options` class which is internal. Is it ok to leave everything as is?

pkg/Microsoft.ML.AutoMLFeaturizers/Microsoft.ML.AutoMLFeaturizers.nupkgproj
michaelgsharp: Sounds good. Removing in the next push.

pkg/Microsoft.ML.AutoMLFeaturizers/Microsoft.ML.AutoMLFeaturizers.nupkgproj
michaelgsharp: I will sync with my team today and figure that out. Part of the reason I picked this name is that these Transformers are added specifically for the AutoML team. So while anyone can use them, they are custom tailored for their use.

src/Microsoft.ML.AutoMLFeaturizers/ToStringTransformer.cs
michaelgsharp: Changed them all to internal. Fixed in next push.

src/Microsoft.ML.AutoMLFeaturizers/ToStringTransformer.cs
michaelgsharp: Removed the extra `r` from all the `SafeHandle` classes. Fixed in next push.

src/Microsoft.ML.AutoMLFeaturizers/Properties/AssemblyInfo.cs
eerhardt: Yeah, all our tests have IVT, so this is fine to leave. I was mostly just curious why it was needed here. IMO it is best to test without Internals Visible To, but if you need it don't worry.

pkg/Microsoft.ML.AutoMLFeaturizers/Microsoft.ML.AutoMLFeaturizers.nupkgproj
ganik: How about naming it NativeFeaturizers? Hinting that implementation is in native code

---
In reply to: [323317152](https://github.com/dotnet/machinelearning/pull/4157#discussion_r323317152) [](ancestors = 323317152)

michaelgsharp:
> CodeFlow for GitHub does not download executable files. If you trust this file and would like to run it, please download it yourself through GitHub.

This is temporary and will be removed before this PR is merged into master. This is just so that we can build using the nuget package.

---
Refers to: temp-nuget-folder/MicrosoftMLFeaturizers.0.1.0.nupkg:1 in efdf22a. [](commit_id = efdf22a65fa1b042e84039c07580b1d36e983ef3, deletion_comment = False)

pkg/Microsoft.ML.AutoMLFeaturizers/Microsoft.ML.AutoMLFeaturizers.nupkgproj
```diff
@@ -7,6 +7,8 @@
 
   <ItemGroup>
     <ProjectReference Include="../Microsoft.ML/Microsoft.ML.nupkgproj" />
+
+    <PackageReference Include="MicrosoftMLFeaturizers" Version="0.1.0" />
```
eerhardt: Will this be the final package name of the native featurizers? If yes, then I find it confusing that the ML.NET package has the word `Auto` in it, but the underlying native package doesn't. What about the ML.NET package makes it "AutoML", but the underlying native package isn't "AutoML"?

michaelgsharp: I will renaming the ML.NET package to be called NativeFeaturizers per Gani's suggestion. Or are you thinking since the native package is MicrosoftMLFeaturizers that I should name the ML.NET package the same thing?

---
In reply to: [326226514](https://github.com/dotnet/machinelearning/pull/4157#discussion_r326226514) [](ancestors = 326226514)

eerhardt: Totally just my opinion, but I don't really like the word `Native` in the public API. I don't think it conveys any real information, and is an implementation detail.

It almost feels like the native assemblies, and the managed assemblies belong in the same NuGet package. But if we need to split them, I think their names should align in an understandable way.

michaelgsharp: Sounds good to me. I'll rename the ML.NET package than to just be Microsoft.ML.Featurizers.

---
In reply to: [326238905](https://github.com/dotnet/machinelearning/pull/4157#discussion_r326238905) [](ancestors = 326238905)

eerhardt:
> CategoryImputer

Why are we creating a whole new transform here, when ML.NET already has [ReplaceMissingValues](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.extensionscatalog.replacemissingvalues) ?

I think this feature should just be an option to the existing ReplaceMissingValues transform. Today ML.NET supports:

https://github.com/dotnet/machinelearning/blob/6d192b691b8828afed52e64ff93e0e9728fdcb95/src/Microsoft.ML.Transforms/MissingValueReplacing.cs#L913-L934

We should just be adding a new value to that enum called `MostCommon` (or similar).

/cc @codemzs @ebarsoumMS

michaelgsharp:
Closing this as we are merging in them one featurizer at a time instead of all together.

PR #4158, opened 2019-08-30
Increment build version for 1.4-preview and 0.16-preview release.

codemzs:


build/BranchInfo.props
```diff
@@ -27,7 +27,7 @@
     <MajorVersion>1</MajorVersion>
     <MinorVersion>4</MinorVersion>
     <PatchVersion>0</PatchVersion>
-    <PreReleaseLabel>preview</PreReleaseLabel>
+    <PreReleaseLabel>preview1</PreReleaseLabel>
```
eerhardt: I think it should be preview2. Going from “preview” to “preview1” doesn’t really make sense IMO.

eerhardt: We also need to bump the 0.16 PreReleaseLabel below.

Issue #4173, opened 2019-09-03
Should use dll location instead of app domain in ResNet18/50/128 packages


LittleLittleCloud:
We find out that in  `Resnet18Extension.cs`, It uses `AppDomain.CurrentDomain` to get the DLL location, which could causes some unexpected behavior in Vsix packaging. (in Vsix AppDomain.CurrentDomain returns VS executables, which is `C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\Common7\IDE`, so it won't be able to find the right onnx file). It's better to use `GetAssembly.Loaction` to get the DLL location. You can see the following PR to see the suggest fix

#4174

eerhardt:
We used to use `Assembly.Location` but that was changed in #2411.  See the history of that file for all the changes that have gone into it.

You can easily call the overload that takes a `modelDir` inside a VSIX to point to the correct place where you are installing the onnx file. You won't need to change ML.NET in order to do that:

https://github.com/dotnet/machinelearning/blob/b0048716190b754eb994db310aaf46feddf2efca/src/Microsoft.ML.DnnImageFeaturizer.ResNet18/ResNet18Extension.cs#L31-L37

PR #4174, opened 2019-09-03
Using `GetAssembly` to get DLL loaction in ResNet18

LittleLittleCloud:


src/Microsoft.ML.DnnImageFeaturizer.ResNet18/ResNet18Extension.cs
```diff
@@ -25,7 +25,8 @@ public static class ResNet18Extension
         /// </summary>
         public static EstimatorChain<ColumnCopyingTransformer> ResNet18(this DnnImageModelSelector dnnModelContext, IHostEnvironment env, string outputColumnName, string inputColumnName)
         {
-            return ResNet18(dnnModelContext, env, outputColumnName, inputColumnName, Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "DnnImageModels"));
+            string fullPath = new DirectoryInfo(System.Reflection.Assembly.GetAssembly(typeof(DnnImageModelSelector)).Location).Parent.FullName;
```
eerhardt: Why are we just changing ResNet18, and not the other DnnImageFeaturizers?

src/Microsoft.ML.DnnImageFeaturizer.ResNet18/ResNet18Extension.cs
```diff
@@ -25,7 +25,8 @@ public static class ResNet18Extension
         /// </summary>
         public static EstimatorChain<ColumnCopyingTransformer> ResNet18(this DnnImageModelSelector dnnModelContext, IHostEnvironment env, string outputColumnName, string inputColumnName)
         {
-            return ResNet18(dnnModelContext, env, outputColumnName, inputColumnName, Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "DnnImageModels"));
```
eerhardt: We've gone through a few iterations of this code (see the git history on this file).

I don't see an issue being referenced for justifying this change. Can you add a scenario/reason why this is being changed?

LittleLittleCloud: Sure
#4173

src/Microsoft.ML.DnnImageFeaturizer.ResNet18/ResNet18Extension.cs
LittleLittleCloud: Just lazy.... I will change the other two after you think the change is necessary...

PR #4179, opened 2019-09-05
pack CodeGen into mlnet

LittleLittleCloud:


src/Microsoft.ML.CodeGen/Microsoft.ML.CodeGenerator.csproj
```diff
@@ -107,5 +107,12 @@
       <DependentUpon>ModelOutputClass.tt</DependentUpon>
     </Compile>
   </ItemGroup>
+  
+  <Import Project="..\mlnet\mlnet.Build.props" />
```
eerhardt: It might make sense to add a comment to why we are doing this. From AutoML.csproj:

```
<!-- here we place this assembly into CLI package dir so it gets included in CLI nuget -->
```

PR #4187, opened 2019-09-07
Modified how DataViewTypes are registered

### LINKED ISSUE #4121: Using OnnxSequenceType and ColumnName attributes together doesn't work
If you have both `OnnxSequenceType` and `ColumnName` attributes on a property, ML.NET gets confused and throws an exception.

For example, if you pull the following code: 

https://github.com/sethjuarez/SeeSharp/tree/c864892252880d4ef87006008823dee3b29adf9c

And change the ImagePrediction class to be:

```
    public class ImagePrediction
    {
        [ColumnName("classLabel")]
        [VectorType]
        public string[] Prediction;

        [ColumnName("loss")]
        [OnnxSequenceType(typeof(IDictionary<string, float>))]
        public IEnumerable<IDictionary<string, float>> Loss;
    }
```

(You will also need to change 1 place in the `Main` method to change from `output.loss.FirstOrDefault();` to `output.Loss.FirstOrDefault();` from the rename.)

You get an error:

```
Unhandled Exception: System.ArgumentOutOfRangeException: Could not determine an IDataView type for member Loss
Parameter name: rawType
   at Microsoft.ML.Data.InternalSchemaDefinition.GetVectorAndItemType(String name, Type rawType, IEnumerable`1 attributes, Boolean& isVector, Type& itemType)
   at Microsoft.ML.Data.SchemaDefinition.Create(Type userType, Direction direction)
   at Microsoft.ML.Data.TypedCursorable`1.Create(IHostEnvironment env, IDataView data, Boolean ignoreMissingColumns, SchemaDefinition schemaDefinition)
   at Microsoft.ML.PredictionEngineBase`2.PredictionEngineCore(IHostEnvironment env, InputRow`1 inputRow, IRowToRowMapper mapper, Boolean ignoreMissingColumns, SchemaDefinition outputSchemaDefinition, Action& disposer, IRowReadableAs`1& outputRow)
   at Microsoft.ML.PredictionEngineBase`2..ctor(IHostEnvironment env, ITransformer transformer, Boolean ignoreMissingColumns, SchemaDefinition inputSchemaDefinition, SchemaDefinition outputSchemaDefinition)
   at Microsoft.ML.PredictionEngineExtensions.CreatePredictionEngine[TSrc,TDst](ITransformer transformer, IHostEnvironment env, Boolean ignoreMissingColumns, SchemaDefinition inputSchemaDefinition, SchemaDefinition outputSchemaDefinition)
   at SeeSharp.Program.LoadModel(String onnxModelFilePath) in /Users/eerhardt/git/SeeSharp/SeeSharp/Program.cs:line 76
   at SeeSharp.Program.Main(String[] args) in /Users/eerhardt/git/SeeSharp/SeeSharp/Program.cs:line 39
```

This should work. The program works if you remove the `ColumnName` attribute.  And fails with the above as soon as you add the `ColumnName` attribute.

/cc @wschin

bpstark:
The DataViewManager registers DataViewTypes to determine how to
represent the the data within ML.Net. However, there was a bug in how
types were being queried. Types that were being registered via an
Attribute were being added to the manager with only that Attribute.
However, when queried from the manager all custom Attributes were being
passed.
To solve this we need to only pass custom Attributes that are relevant
to the Manager. All Type attributes now inherit from a single
TypeAttribute class such that we can filter custom attributes for
relevant types.

Fixes #4121


src/Microsoft.ML.DataView/DataViewType.cs
```diff
@@ -473,7 +473,7 @@ public override bool Equals(DataViewType other)
     /// in <see cref="IDataView"/> would be the associated <see cref="DataViewType"/>.
     /// </summary>
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property, AllowMultiple = false, Inherited = true)]
-    public abstract class DataViewTypeAttribute : Attribute, IEquatable<DataViewTypeAttribute>
+    public abstract class DataViewTypeAttribute : TypeAttribute, IEquatable<DataViewTypeAttribute>
```
yaeldekel: >DataViewTypeAttribute [](start = 26, length = 21)

I think this change should work without the intermediate `TypeAttribute` class, by using `DataViewTypeAttribute` in `DataViewTypeManager` instead.

src/Microsoft.ML.Data/Data/DataViewTypeManager.cs
```diff
@@ -53,6 +53,8 @@ public static class DataViewTypeManager
         /// </summary>
         internal static DataViewType GetDataViewType(Type type, IEnumerable<Attribute> typeAttributes = null)
         {
+            //Filter attributes as we only care about TypeAttributes
+            typeAttributes = typeAttributes == null ? typeAttributes : typeAttributes.Where(attr => attr.GetType().IsSubclassOf(typeof(TypeAttribute)));
```
yaeldekel: >Where [](start = 86, length = 5)

Can we instead pass an `IEnumerable<DataViewTypeAttribute>`?

src/Microsoft.ML.Data/Data/DataViewTypeManager.cs
```diff
@@ -73,6 +75,8 @@ internal static DataViewType GetDataViewType(Type type, IEnumerable<Attribute> t
         /// </summary>
         internal static bool Knows(Type type, IEnumerable<Attribute> typeAttributes = null)
         {
+            //Filter attributes as we only care about TypeAttributes
+            typeAttributes = typeAttributes == null ? typeAttributes : typeAttributes.Where(attr => attr.GetType().IsSubclassOf(typeof(TypeAttribute)));
```
yaeldekel: >Where [](start = 86, length = 5)

Same comment here, and in line 120.

eerhardt:
We should use the scenario from https://github.com/dotnet/machinelearning/issues/4121 to test whether this change fixes the problem. Use an `[OnnxSequenceType]` attribute with a `[ColumnName]` attribute and make sure you can execute the ONNX file correctly.

src/Microsoft.ML.DataView/DataViewType.cs
bpstark: So it looks like you are correct we can just use the DataViewTypeAttribute instead.

src/Microsoft.ML.Data/Data/DataViewTypeManager.cs
bpstark: We would then have to modify all callers of this function and the Knows function to do this same filtering logic as the way that they appear to be getting the list is by calling GetCustomAttributes. it seems cleaner instead to do the filtering logic here.

bpstark:
@eerhardt The tests added here do capture the issue of #4121, I have simply reduced the issue down to its simplest forms to make it a unit testable component.

eerhardt:
> I have simply reduced the issue down to its simplest forms

There are things that can go wrong when doing that.

1. The reduced scenario may not represent the original issue accurately.
2. There may be more issues that need fixing in the original scenario than just the reduced first issue hit.
3. In the future, someone may add/change code in the product that would break the original scenario, but not your reduced scenario test.

In a case like this, capturing the original scenario in a test case to ensure it is fixed, and prevent future regressions, is ideal.

test/Microsoft.ML.Tests/Microsoft.ML.Tests.csproj
```diff
@@ -54,4 +54,9 @@
     <PackageReference Include="Microsoft.ML.TestModels" Version="$(MicrosoftMLTestModelsPackageVersion)" />
     <PackageReference Include="System.Data.SqlClient" Version="$(SystemDataSqlClientVersion)" />
   </ItemGroup>
+  <ItemGroup>
+    <None Update="column_name_test\model.onnx">
```
eerhardt: We shouldn't add large files into the repo. Instead, we use https://github.com/dotnet/machinelearning-testdata, which builds NuGet packages that we consume in this repo.

There are a few ONNX models already in the `Microsoft.ML.Onnx.TestModels` nuget package, that you can reuse in your test. No need to add a new one.

test/Microsoft.ML.Tests/OnnxSequenceTypeWithAttributesTest.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System;
```
eerhardt: We need copyright headers on all `src` and `test` code files in the repo.

src/Microsoft.ML.Data/Data/DataViewTypeManager.cs
yaeldekel: Fair enough. For the `Register` method though I still think it would be better to pass the `DataViewTypeAttribute`s directly, since in all the calls I see to this method, the attributes that are passed are derived from `DataViewTypeAttribute` anyway.



---
In reply to: [322322602](https://github.com/dotnet/machinelearning/pull/4187#discussion_r322322602) [](ancestors = 322322602)

src/Microsoft.ML.Data/Data/DataViewTypeManager.cs
```diff
@@ -113,6 +117,16 @@ internal static bool Knows(DataViewType dataViewType)
         /// <param name="typeAttributes">The <see cref="Attribute"/>s attached to <paramref name="type"/>.</param>
         public static void Register(DataViewType dataViewType, Type type, IEnumerable<Attribute> typeAttributes = null)
```
yaeldekel: >IEnumerable [](start = 74, length = 11)

It's not a part of this change, but I'm just wondering why is this an enumerable, and not just one attribute? What would be the scenario for having multiple attributes?
In the `SchemaDefinition.Create` method (SchemaDefinition.cs line 388) it seems like it is not possible to have more than one `DataViewTypeAttribute`, or am I missing something?

Also, how is this method intended to be used? From the sample, it seems that it is only used for implementing the abstract method `Register()` of `DataViewTypeAttribute`. If this is the case, shouldn't `Register()` be implemented in the base class by calling this method and have abstract methods/properties to get the arguments needed for it? This way the `Register` method here can be internal, which would be less confusing (at least to me :)).

/cc @eerhardt .

src/Microsoft.ML.Data/Data/DataViewTypeManager.cs
bpstark: Yeah we can change to use a dataviewtype in the register function I think. however we should leave the Knows and Get calls as is.

src/Microsoft.ML.Data/Data/DataViewTypeManager.cs
eerhardt: @wschin - for the question. He originally added the API.

Since this is a public API, we probably can't change it since it has already shipped.

bpstark: I believe I have found the appropriate approach to how to address the API, please see the latest commit.

PR #4188, opened 2019-09-07
Added Onnx export functionality to PCATransformer

### LINKED ISSUE #4186: PCA Transformer does not support exporting to Onnx
- **What did you do?**
Created a pipeline from with PCATranfsform and tried to export it to Onnx with - ConvertToOnnxProtobuf.

- **What happened?**
The transform wasn't saved in the onnx graph

- **What did you expect?**
The transform should be saved in the onnx graph

harishsk:
Fixes #4186


test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -845,7 +886,14 @@ private void CompareSelectedR4VectorColumns(string leftColumnName, string rightC
...
                     Assert.Equal(expected.Length, actual.Length);
                     for (int i = 0; i < expected.Length; ++i)
-                        Assert.Equal(expected.GetItemOrDefault(i), actual.GetItemOrDefault(i), precision);
+                    {
+                        // We are using float values. But the Assert.Equal function takes doubles.
+                        // And sometimes the converted doubles are different in their precision.
+                        // So make sure we compare floats
+                        float exp = expected.GetItemOrDefault(i);
+                        float act = actual.GetItemOrDefault(i);
+                        Assert.True((Math.Abs(exp - act) < prec) || (float.IsNaN(exp) && float.IsNaN(act)));
```
eerhardt: There’s an overload to assert.equals that takes a precision. You don’t need to do this yourself.

harishsk: That is the problem (as described in the comments). The precision comparison happens after the conversion to double. And there seems to be some bug in the conversion to double where the precision isn't preserved during the conversion.

src/Microsoft.ML.PCA/PcaTransformer.cs
```diff
@@ -596,6 +597,78 @@ private static void TransformFeatures(IExceptionContext ectx, in VBuffer<float>
...
+
+                    if (!SaveAsOnnxCore(ctx, i, ctx.GetVariableName(inputColumnName),
+                        ctx.AddIntermediateVariable(transformInfo.OutputType, outputColumnName)))
+                    {
+                        ctx.RemoveColumn(outputColumnName, true);
+                    }
+                }
+            }
+
+            private bool SaveAsOnnxCore(OnnxContext ctx, int iinfo, string srcVariableName, string dstVariableName)
```
eerhardt: `bool` - When does this method ever return `false`?

test/Microsoft.ML.Tests/OnnxConversionTest.cs
eerhardt: In that case, then check out:

https://github.com/dotnet/machinelearning/blob/fabdabf867dc68481d519740a96b981bbeb63180/test/Microsoft.ML.TestFramework/BaseTestBaseline.cs#L572-L574

Which is how we handle this case for the rest of the tests. The advantage of using that method is that it will log the values of each in the failure. Where your test code above will just say "Failed expected `true` but got `false`", which doesn't help when investigating a failure.

src/Microsoft.ML.PCA/PcaTransformer.cs
harishsk: It doesn't in this case. I have tried to preserve the pattern used across other transformers in their implementation of ISaveAsOnnx.

eerhardt: That seems less than ideal, given that you now have dead code when calling this method.

You can still have a `SaveAsOnnxCore` method, but just make it return `void` and remove the `if` check above.

test/Microsoft.ML.Tests/OnnxConversionTest.cs
harishsk: Resolved with latest checkin.

src/Microsoft.ML.PCA/PcaTransformer.cs
harishsk: Resolved with latest checkin

src/Microsoft.ML.PCA/PcaTransformer.cs
```diff
@@ -596,6 +597,73 @@ private static void TransformFeatures(IExceptionContext ectx, in VBuffer<float>
...
+            {
+                Host.CheckValue(ctx, nameof(ctx));
+
+                for (int i = 0; i < _numColumns; i++)
+                {
+                    var colPair = _parent.ColumnPairs[i];
+                    var transformInfo = _parent._transformInfos[i];
+                    string inputColumnName = colPair.inputColumnName;
+                    string outputColumnName = colPair.outputColumnName;
+                    if (!ctx.ContainsColumn(inputColumnName))
```
yaeldekel: >!ctx.ContainsColumn(inputColumnName) [](start = 24, length = 36)

Can this happen?
If we just continue when this happens, doesn't it mean that the ONNX model will be different than the original model? Shouldn't we throw in that case?

harishsk: I think that logic exists precisely to allow the higher level code to drop specific columns that it doesn't want in the model. The `RemoveVariablesInPipelineTest` test in `test\Microsoft.ML.Tests\OnnxConversionTest.cs` exercises that functionality where it drops specific columns in its call to `SaveOnnxCommand.ConvertTransformListToOnnxModel`.

harishsk:
If all concerns have been resolved, can one of you please approve the review?

ganik:
Can you pls update comments that are resolved as such. 

---
In reply to: [530095935](https://github.com/dotnet/machinelearning/pull/4188#issuecomment-530095935) [](ancestors = 530095935)

PR #4190, opened 2019-09-08
Svmlight loader and saver

### LINKED ISSUE #4014: Add a loader/saver for SVMLight file format
Many datasets are saved in the SVMLight format, describe in this page: http://svmlight.joachims.org/.

It would be useful for ML.NET to support loading data directly from files in this format.

yaeldekel:
Fixes #4014 .


src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
```diff
@@ -0,0 +1,851 @@
...
+    /// due to the way we compose our feature vectors, but it will be most efficient if this policy
+    /// is still followed. If it is followed a sort will not be required.
+    ///
+    /// This loader has the special option through the <c>xf</c> option to specify a transform,
+    /// possibly trainable, to convert the raw text of the key values into the key value. The
+    /// transform, whatever it is, must in addition to user specified options accept an argument
+    /// of the form "column=Name" to identify a column to convert. Ideally there would be some
+    /// other way to specify this other than hacking arguments. The intent of this is to allow
+    /// things like string keys, a common variant of the format, but one emphatically not allowed
+    /// by the original format.
```
yaeldekel: This variation existed in the internal version of this loader, but I have not been able to find any references to it online. How important do you think it is to support this?

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoaderSaverCatalog.cs
```diff
@@ -0,0 +1,109 @@
...
+using System.Linq;
+using Microsoft.ML.Data;
+using Microsoft.ML.Runtime;
+
+namespace Microsoft.ML.Transforms
+{
+    public static class SvmLightLoaderSaverCatalog
+    {
+        /// <summary>
+        ///
```
yaeldekel: >/// [](start = 8, length = 3)

Comments coming soon.

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
```diff
@@ -0,0 +1,851 @@
+﻿//------------------------------------------------------------------------------
+// <copyright company="Microsoft Corporation">
+//     Copyright (c) Microsoft Corporation. All rights reserved.
+// </copyright>
+//------------------------------------------------------------------------------
```
codemzs: Please use the ML .NET header.

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoaderSaverCatalog.cs
```diff
@@ -0,0 +1,126 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.IO;
+using System.Linq;
+using Microsoft.ML.Data;
+using Microsoft.ML.Runtime;
+
+namespace Microsoft.ML.Transforms
+{
+    public static class SvmLightLoaderSaverCatalog
```
codemzs: >SvmLightLoaderSaverCatalog [](start = 24, length = 26)

Can we please add samples for all of these APIs?

codemzs:
@yaeldekel Can you please sync to master and push your changes so we can analyze the code coverage? thanks.

test/Microsoft.ML.Tests/SvmLightTests.cs
```diff
@@ -0,0 +1,193 @@
...
+            [KeyType(ulong.MaxValue - 1)]
+            public ulong GroupId = ulong.MaxValue;
+            public ReadOnlyMemory<char> Comment;
+            public VBuffer<float> Features;
+        }
+#pragma warning restore 0649
+
+        [Fact]
+        public void TestSvmLightLoaderAndSaver()
+        {
```
harishsk: It appears that this function is encapsulating several tests. Is it possible to refactor them into different tests?

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoaderSaverCatalog.cs
```diff
@@ -0,0 +1,126 @@
...
+                SvmLightLoader.FeatureIndices.ZeroBased : SvmLightLoader.FeatureIndices.OneBased }, dataSample);
+
+        /// <summary>
+        /// Creates a loader that loads SVM-light like files, where features are specified by their names.
+        /// </summary>
+        /// <param name="catalog">The <see cref="DataOperationsCatalog"/> catalog.</param>
+        /// <param name="numberOfRows">The number of rows from the sample to be used for determining the set of feature names.</param>
+        /// <param name="dataSample">A data sample to be used for determining the set of features names.</param>
+        public static SvmLightLoader CreateSvmLightLoaderWithFeatureNames(this DataOperationsCatalog catalog,
+            long? numberOfRows = null,
```
harishsk: CreateTextLoader two versions with different parameters. Is that a better practice to follow and call both functions CreateSvmLightLoader?

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoaderSaverCatalog.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        /// <summary>
+        /// Load a <see cref="IDataView"/> from a text file containing features specified by feature names,
+        /// using <see cref="SvmLightLoader"/>.
+        /// </summary>
+        /// <param name="catalog">The <see cref="DataOperationsCatalog"/> catalog.</param>
+        /// <param name="path">The path to the file.</param>
+        /// <param name="numberOfRows">The number of rows from the sample to be used for determining the set of feature names.</param>
+        public static IDataView LoadFromSvmLightFileWithFeatureNames(this DataOperationsCatalog catalog,
+            string path,
```
harishsk: Similar comment as above.

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
harishsk: As we discussed on the call, I think it is important to support this. Since SvmLightLoader is implemented as data loader, it is typically the first stage in the pipeline. If someone has data in this format, they will not be able to load the data into ML.NET if we don't support this.

---
In reply to: [322014712](https://github.com/dotnet/machinelearning/pull/4190#discussion_r322014712) [](ancestors = 322014712)

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
```diff
@@ -0,0 +1,851 @@
...
+        /// This class is used for mapping the vector of <see cref="ReadOnlyMemory{T}"/> of <see cref="char"/> field in an <see cref="IntermediateInput"/> object
+        /// to a numeric vector of indices. This class is used in case the indices in the input file are numeric. For the case where the input file contains
+        /// feature names, instead of a <see cref="CustomMappingTransformer{TSrc, TDst}"/>, we use a <see cref="ValueToKeyMappingTransformer"/>.
+        /// </summary>
+        private sealed class Indices
+        {
+            public VBuffer<int> FeatureKeys;
+
+            public static void ParseIndicesToOneBased(IntermediateInput input, Indices output)
+            {
```
harishsk: The implementations of the two functions in this class are nearly identical. Can they be refactored into a single function that takes as a parameter the offset to use?

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
```diff
@@ -0,0 +1,851 @@
...
+                    if (Conversions.Instance.TryParse(in inputValues[i], out int index) && index > 0)
+                        editor.Values[i] = index - 1;
+                    else
+                        editor.Values[i] = -1;
+                }
+                output.FeatureKeys = editor.Commit();
+            }
+
+            public static void ParseIndicesToZeroBased(IntermediateInput input, Indices output)
+            {
```
harishsk: What happens if the caller asks that the data be be parsed with a one based index, but the data has a zero in its list of keys?

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
```diff
@@ -0,0 +1,851 @@
...
+                var editor = VBufferEditor.Create(ref output.FeatureKeys, input.FeatureKeys.Length);
+                var inputValues = input.FeatureKeys.GetValues();
+                for (int i = 0; i < inputValues.Length; i++)
+                {
+                    if (Conversions.Instance.TryParse(in inputValues[i], out int index) && index >= 0)
+                        editor.Values[i] = index;
+                    else
```
harishsk: What is the desired behavior if the data is bad? E.g. if the data has duplicate keys? (Maybe throw an exception if the other data loaders throw an exception when the data is loaded?)

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
```diff
@@ -0,0 +1,851 @@
...
+
+        /// <summary>
+        /// This class and the <see cref="IntermediateOut"/> class are used by the <see cref="CustomMappingTransformer{TSrc, TDst}"/>
+        /// that maps a vector of indices and a vector of values into a single <see cref="VBuffer{T}"/> of values. When the indices
+        /// originate from the <see cref="ValueToKeyMappingTransformer"/> (in case features are specified by name), <see cref="IntermediateOutKeys"/>
+        /// is used, and when they originate from a <see cref="CustomMappingTransformer{TSrc, TDst}"/> that produces an <see cref="Indices"/>,
+        /// <see cref="IntermediateOut"/> is used.
+        /// </summary>
+        private sealed class IntermediateOutKeys
+        {
```
harishsk: Do we need these two nearly identical classes? Or can they be made into a generic class?

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
```diff
@@ -0,0 +1,851 @@
...
+            public VBuffer<uint> FeatureKeys;
+            public VBuffer<float> FeatureValues;
+        }
+
+        private sealed class IntermediateOut
+        {
```
harishsk: Or as we discussed, is it possible to have indices be uint always, since the indices into the array cannot be negative?

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
```diff
@@ -0,0 +1,851 @@
...
+            public VBuffer<int> FeatureKeys;
+            public VBuffer<float> FeatureValues;
+        }
+
+        private sealed class Output
+        {
+            public VBuffer<float> Features;
+        }
```
harishsk: I think you mentioned that this is a dense array. Would it be better to keep it as a sparse array so that the algorithm that is making use of this data can decide at a later time if it needs to convert this into a dense array?

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
```diff
@@ -0,0 +1,851 @@
...
+
+        private sealed class IntermediateOut
+        {
+            public VBuffer<int> FeatureKeys;
+            public VBuffer<float> FeatureValues;
+        }
+
+        private sealed class Output
+        {
```
harishsk: Is there a use case to (say after inference, or during evaluation) to map these indices back to keys or the original strings? In that case, does it make sense to provide a separate function/helper to extract the keys/strings back?

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
```diff
@@ -0,0 +1,851 @@
...
+                        _currReader = _parent._files.OpenTextReader(_fileIdx);
+                    }
+                }
+            }
+        }
+
+        private readonly IHost _host;
+        private readonly ITransformer _keyVectorsToIndexVectors;
+        private readonly FeatureIndices _indicesKind;
+        private readonly ulong _featureCount;
```
harishsk: >_featureCount [](start = 31, length = 13)

Is it useful to make this public?

codemzs:
@yaeldekel Thanks for syncing to master so that code coverage report is generated. Looking at code coverage report it seems there are branches in your code that is not covered by a test, can you please take a look here https://codecov.io/gh/dotnet/machinelearning/pull/4190/diff (on left side it will highlight lines in red) thanks!

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
yaeldekel: I am not sure there is a need to do that. You can get the length of the feature vector when you call `GetOutputSchema()` on the loader.

---
In reply to: [332212933](https://github.com/dotnet/machinelearning/pull/4190#discussion_r332212933) [](ancestors = 332212933)

yaeldekel:
Yes, I'll take a look. Question - we normally don't test all the cases where exceptions are thrown, do we? 

---
In reply to: [539612360](https://github.com/dotnet/machinelearning/pull/4190#issuecomment-539612360) [](ancestors = 539612360)

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoaderSaverCatalog.cs
yaeldekel: Do you think I should have one method that takes an `SvmLightLoader.FeatureIndices` as a parameter?
In that case, perhaps we could come up with a better name for that enum? Any suggestions?

---
In reply to: [332206307](https://github.com/dotnet/machinelearning/pull/4190#discussion_r332206307) [](ancestors = 332206307)

test/Microsoft.ML.Tests/SvmLightTests.cs
yaeldekel: Good suggestion, I'll do that in the next iteration.

---
In reply to: [332204210](https://github.com/dotnet/machinelearning/pull/4190#discussion_r332204210) [](ancestors = 332204210)

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
yaeldekel: The way it is written now, the feature will be ignored. (see line 390).
Do you think we should throw?

---
In reply to: [332208505](https://github.com/dotnet/machinelearning/pull/4190#discussion_r332208505) [](ancestors = 332208505)

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
yaeldekel: I added support for keeping track which indices exist in each row (see the `OutputMapper` class in line 359). Currently, I am not throwing in case of duplicate keys, just using the value in the first appearance.
Do you think I should throw instead?

---
In reply to: [332209060](https://github.com/dotnet/machinelearning/pull/4190#discussion_r332209060) [](ancestors = 332209060)

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
yaeldekel: I switched to using a `BufferBuilder` instead of the `VBufferEditor`, since it knows how to handle indices out of order. There might be perf implications though, not sure which is better to use in this case.

---
In reply to: [332210125](https://github.com/dotnet/machinelearning/pull/4190#discussion_r332210125) [](ancestors = 332210125)

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
harishsk: For the both the questions above, I would defer to precedence. If other data loaders throw when bad data is encountered, we should throw here as well.

---
In reply to: [333656937](https://github.com/dotnet/machinelearning/pull/4190#discussion_r333656937) [](ancestors = 333656937,332209060)

harishsk:
Code coverage as part of continuous integration has recently come in so I think we get to collectively define what the policy should be going forward and your thoughts and opinions on the issue would be most welcome. 

---
In reply to: [540696889](https://github.com/dotnet/machinelearning/pull/4190#issuecomment-540696889) [](ancestors = 540696889,539612360)

harishsk:
I am sure there are tests for negative cases in our code bases where exceptions are thrown. My question is, whether we have been doing it consistently throughout so far. And if not, is that the new policy you are proposing?

---
In reply to: [540708878](https://github.com/dotnet/machinelearning/pull/4190#issuecomment-540708878) [](ancestors = 540708878,540705504,540696889,539612360)

sharwell:
I generally encourage the creation of tests to cover these cases. Discretion should be used if there is a case that is not easy to hit (we don't want to spend more than a reasonable amount of time trying to find a way to hit it). Where possible, the negative tests should use the normal API entry points, as opposed to side-stepping the API for direct calls to internals/privates. This helps ensure that the error condition which leads to the negative result does not result in abnormal behavior prior to the expected error handling locations.

codemzs:
@sharwell Agreed. @yaeldekel lets try to add negative test.

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoaderSaverCatalog.cs
harishsk: Yes, I think having a method that takes FeatureIndices as a parameter would be helpful to distinguish the two cases. One option would be to just call the enum as `Features` and call the enum values as `ZeroBasedIndices`, `OneBaseIndices` and `Names`. 
Also, both the methods aboe have numberOfRows as a parameter, but they are in different positions. If the two functions have the same parameter, it would be good to keep them in the same position. 


---
In reply to: [333653705](https://github.com/dotnet/machinelearning/pull/4190#discussion_r333653705) [](ancestors = 333653705,332206307)

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
yaeldekel: In case of using string keys to load the data, the resulting `IDataView` schema will have slot names annotations for the Features column, so the feature names can be extracted this way.

---
In reply to: [332212446](https://github.com/dotnet/machinelearning/pull/4190#discussion_r332212446) [](ancestors = 332212446)

yaeldekel: (see in `CreateOutputTransformer` inside the `if (keyIndices)` part).

---
In reply to: [343596251](https://github.com/dotnet/machinelearning/pull/4190#discussion_r343596251) [](ancestors = 343596251,332212446)

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
yaeldekel: I'll mark this as resolved, if you think it's important to make public, please re-activate and let me know.

---
In reply to: [333650195](https://github.com/dotnet/machinelearning/pull/4190#discussion_r333650195) [](ancestors = 333650195,332212933)

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoaderSaverCatalog.cs
yaeldekel: I've addressed the parameter-order part of your comment, but I had second thoughts about having a single API that specifies whether the file contains feature indices or feature names - there is a parameter `inputSize` which is only relevant in the feature indices case and not in the feature names case, and I'm not sure that having a parameter that is sometimes ignored is the best thing to have - do you have any thoughts on that?

---
In reply to: [337250107](https://github.com/dotnet/machinelearning/pull/4190#discussion_r337250107) [](ancestors = 337250107,333653705,332206307)

harishsk: Based on your explanation, would reordering and renaming the params as follows work?
```
        // For namesCreateSvmLightLoader to use with indices
        public static SvmLightLoader CreateSvmLightLoader(
            this DataOperationsCatalog catalog,
            long? numberOfRows = null,
            IMultiStreamSource dataSample = null,
            Features features,
            int featureSize,
            );

        // For namesCreateSvmLightLoaderWithNames
        public static SvmLightLoader CreateSvmLightLoader( 
            this DataOperationsCatalog catalog,
            long? numberOfRows = null,
            IMultiStreamSource dataSample = null,
            Features features
            );
```
The other option is to keep only one version, (whichever you think would be the more common one and change the other one to have the SvmLightLoader.Options class directly as a param. That too has been a pattern in ML.NET.

In either case I think input from @eerhardt would be useful for the public API surface.



---
In reply to: [343611210](https://github.com/dotnet/machinelearning/pull/4190#discussion_r343611210) [](ancestors = 343611210,337250107,333653705,332206307)

eerhardt:
> // Licensed to the .NET Foundation under one or more agreements.

Any reason these are in `ML.Transforms` and not `ML.Data`? All our other "loaders" are in `.Data`, correct?

---
Refers to: src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs:1 in 7358266. [](commit_id = 73582665521114371f29927d3e13cb4c76df0f45, deletion_comment = False)

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
```diff
@@ -0,0 +1,819 @@
...
+            public int InputSize;
+
+            [Argument(ArgumentType.Multiple, HelpText = "Whether the features are indexed by numbers starting at 0, by numbers starting at 1, or by feature names.", ShortName = "indices")]
+            public FeatureIndices FeatureIndices = FeatureIndices.OneBased;
+
+            [Argument(ArgumentType.AtMostOnce, HelpText = "The number of rows used to train the feature name to index mapping transform. If unspecified, all rows will be used.", ShortName = "numxf")]
+            public long? NumberOfRows;
+        }
+
+#pragma warning disable 0649 // Disable warnings about unused members. They are used through reflection.
```
eerhardt: Can you just do this around the unused members instead of a huge block?

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
```diff
@@ -0,0 +1,819 @@
...
+    public sealed class SvmLightLoader : IDataLoader<IMultiStreamSource>
+    {
+        public enum FeatureIndices
+        {
+            ZeroBased,
+            OneBased,
+            Names
+        }
+
+        public sealed class Options
```
eerhardt: xml comments on all new public API

eerhardt: Do these types (FeatureIndices and Options) need to be public? I don't see them being used publicly.

---
In reply to: [356301711](https://github.com/dotnet/machinelearning/pull/4190#discussion_r356301711) [](ancestors = 356301711)

yaeldekel: Currently they don't need to be public. There was a discussion about whether there should be a public API that takes an `Options` parameter, or a `FeatureIndices` parameter (instead of having one called `CreateSvmLightLoader` and one called `CreateSvmLightLoaderWithFeatureNames`). Do you have a preference?

---
In reply to: [356304532](https://github.com/dotnet/machinelearning/pull/4190#discussion_r356304532) [](ancestors = 356304532,356301711)

yaeldekel:
> // Licensed to the .NET Foundation under one or more agreements.

The loader uses some components from `Microsoft.ML.Transforms` (specifically, `CustomMappingTransformer`). How critical is it do you think for `SvmLightLoader` not to be in the transformers assembly? The only other solution I can think of, is to add a new project that would reference both `ML.Data` and `ML.Transforms`.

---
In reply to: [564278518](https://github.com/dotnet/machinelearning/pull/4190#issuecomment-564278518) [](ancestors = 564278518)

---
Refers to: src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs:1 in 7358266. [](commit_id = 73582665521114371f29927d3e13cb4c76df0f45, deletion_comment = False)

eerhardt:
I don’t think the assembly it lives in is critical. It just struck me as odd. Thanks for answering my question.

src/Microsoft.ML.Transforms/SvmLight/SvmLightLoader.cs
eerhardt: Do all combinations of enums and options make sense together? Or do some options only make sense for certain values of the enum?

If they all make sense together then I would expose the Options. But if there are options that don’t make sense together, then I would keep it like you have it.

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/LoadingSvmLight.cs
```diff
@@ -0,0 +1,116 @@
...
+using Microsoft.ML.Data;
+using Microsoft.ML.Transforms;
+using Microsoft.VisualBasic.CompilerServices;
+using Tensorflow;
+
+namespace Samples.Dynamic.DataOperations
+{
+    public static class LoadingSvmLight
+    {
+        // This examples shows all the ways to load data with TextLoader.
```
codemzs: >TextLoader. [](start = 62, length = 11)

Text loader?

codemzs:
> using System;

header missing

---
Refers to: docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/LoadingSvmLight.cs:1 in d2619ed. [](commit_id = d2619eda0f318720bfd6d6f8b4863e1a89e739ed, deletion_comment = False)

Issue #4191, opened 2019-09-08
[DNN Training] Failed save model in Microsoft.ML.Transforms.DnnTransformer.SaveModel


yamachu:
### System information

- **OS version/distro**: maxOS 10.14.6
- **.NET Version (eg., dotnet --info)**: 3.0.100-preview9-014004

### Issue

- **What did you do?**

  * Use `RetrainDnnModel` for training regression-based deep-neural-network model
  * Try to save trained model

- **What happened?**

Application throw unhandled exception: System.ArgumentNullException and crash

- **What did you expect?**

We could save trained model

### Source code / logs

```
Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'value')
   at System.IO.BinaryWriter.Write(String value)
   at Microsoft.ML.Transforms.DnnTransformer.SaveModel(ModelSaveContext ctx)
   at Microsoft.ML.Data.RowToRowTransformerBase.Microsoft.ML.ICanSaveModel.Save(ModelSaveContext ctx)
   at Microsoft.ML.ModelSaveContext.SaveModel[T](RepositoryWriter rep, T value, String path)
   at Microsoft.ML.ModelOperationsCatalog.Save(ITransformer model, DataViewSchema inputSchema, Stream stream)
   at Microsoft.ML.ModelOperationsCatalog.Save(ITransformer model, DataViewSchema inputSchema, String filePath)
   at VoiceConversionStarter.Console.Program.Train(TrainMcapOptions opts) in /Users/yamachu/Projects/github.com/yamachu/VoiceConversionStarter/VoiceConversionStarter.Console/Program.cs:line 100
   at VoiceConversionStarter.Console.Program.<>c.<Main>b__2_0(TrainMcapOptions opts) in /Users/yamachu/Projects/github.com/yamachu/VoiceConversionStarter/VoiceConversionStarter.Console/Program.cs:line 118
   at CommandLine.ParserResultExtensions.MapResult[TSource,TResult](ParserResult`1 result, Func`2 parsedFunc, Func`2 notParsedFunc)
   at VoiceConversionStarter.Console.Program.Main(String[] args) in /Users/yamachu/Projects/github.com/yamachu/VoiceConversionStarter/VoiceConversionStarter.Console/Program.cs:line 116
```

I think this issue is not to check null in this line

[...]

PR #4197, opened 2019-09-10
Fix relation between IsSavedModel and isFrozen in DnnRetrainTransformer.SaveModel

yamachu:
Fix: https://github.com/dotnet/machinelearning/issues/4191

IsSavedModel returns true when loaded model is an-frozen model 
https://github.com/dotnet/machinelearning/blob/1503b0aa9cac997cff8b8bc7e2075eb23d61ad81/src/Microsoft.ML.Dnn/DnnUtils.cs#L137-L143

but now isFrozen variable is set true in spite of an-frozen model.

https://github.com/dotnet/machinelearning/blob/3c02da5d75534265223e56aee9c5d2da53ea4b99/src/Microsoft.ML.Dnn/DnnRetrainTransform.cs#L681

---

Checks

### DIFF
--- a/src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
+++ b/src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
@@ -678,7 +678,7 @@ namespace Microsoft.ML.Transforms
             // for each output column
             //   int: id of output column name
             // stream: tensorFlow model.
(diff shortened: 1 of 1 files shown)


codemzs:
Thanks @yamachu ! I recall fixing this bug in other DNN api but probably missed here. Good work investigating and fixing.

PR #4205, opened 2019-09-11
Upgrade TF.Net version from 0.10.10 to 0.11.3

### LINKED ISSUE #4204: Need to move from TF.NET version 0.10.10 to 0.11.3
- ML.Net currently runs with Tensorflow.Net version 0.10.10, but there is a newer stable version of Tensorflow.Net 0.11.3 with improvements :https://www.nuget.org/packages/TensorFlow.NET/0.11.3
- **What happened?**
There are build errors with TF.Net 0.11.3
- **What did you expect?**
Need the required API changes to make ML.Net compatible with TF.Net 0.11.3

ashbhandare:
Fixes #4204

Includes a few API changes necessary to upgrade TF.Net version.
Verified that ML.Samples  and Tests run.


docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/ImageClassification.cs
```diff
@@ -34,11 +34,13 @@ public static void Example()
             var idv = mlContext.Data.LoadFromEnumerable(data);
 
             // Create a ML pipeline.
+
```
eerhardt: I'm not sure if these changes are intended. Can they be reverted?

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
```diff
@@ -245,13 +245,14 @@ private void CheckTrainingParameters(DnnRetrainEstimator.Options options)
...
             {
-                var newShape = new int[tfInputShape.NDim];
-                newShape[0] = tfInputShape[0] == 0 || tfInputShape[0] == -1 ? batchSize : tfInputShape[0];
-
-                for (int j = 1; j < tfInputShape.NDim; j++)
-                    newShape[j] = tfInputShape[j];
+                var newShape = new int[tfInputShape.ndim];
+                var dims = tfInputShape.dims;
+                newShape[0] = dims[0] == 0 || dims[0] == -1 ? batchSize : dims[0];
+                //var v = tfInputShape[0].Equals(0);
```
eerhardt: Is this comment useful?

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
```diff
@@ -382,10 +383,10 @@ private void TrainCore(DnnRetrainEstimator.Options options, IDataView input, IDa
                 runner.AddInput(inputs[i], srcTensorGetters[i].GetBufferedBatchTensor());
 
             Tensor[] tensor = runner.Run();
-            var buffer = tensor[0].Data();
-            loss = tensor.Length > 0 && tensor[0] != IntPtr.Zero ? (float)tensor[0].Data<float>()[0] : 0.0f;
-            metric = tensor.Length > 1 && tensor[1] != IntPtr.Zero ? (float)tensor[1].Data<float>()[0] : 0.0f;
-            var b = tensor.Length > 2 && tensor[2] != IntPtr.Zero ? (float[])tensor[2].Data<float>() : null;
+            //var buffer = tensor[0].Data();
```
eerhardt: Is this comment useful?

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
bpstark: should remove unused code rather than commenting it out.

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
```diff
@@ -382,10 +383,10 @@ private void TrainCore(DnnRetrainEstimator.Options options, IDataView input, IDa
...
+            loss = tensor.Length > 0 && tensor[0] != IntPtr.Zero ? (float)tensor[0].ToArray<float>()[0] : 0.0f;
```
eerhardt: I haven't looked at the implementation of `ToArray`, but I assume it copies the data out into a new array instance. That isn't great for performance reasons (creating a new array instance, copying the values over, and then GC'ing the instance).

/cc @codemzs

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
bpstark: same here

codemzs:
Why are the below files added? They should not be added, please remove them.

image

docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/ImageClassification.cs
```diff
@@ -34,11 +34,13 @@ public static void Example()
...
             var pipeline = mlContext.Model.LoadTensorFlowModel(modelLocation)
```
codemzs: remove this new line.

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
```diff
@@ -811,18 +812,22 @@ private sealed class Mapper : MapperBase
                             throw Contracts.Except($"Input shape mismatch: Input '{_parent._inputs[i]}' has shape {originalShape.ToString()}, but input data is of length {typeValueCount}.");
 
                         // Fill in the unknown dimensions.
-                        var l = new int[originalShape.NDim];
-                        for (int ishape = 0; ishape < originalShape.NDim; ishape++)
-                            l[ishape] = originalShape[ishape] == -1 ? (int)d : originalShape[ishape];
+                        var l = new int[originalShape.ndim];
+                        for (int ishape = 0; ishape < originalShape.ndim; ishape++)
+                        {
+                            var dims = originalShape.dims;
```
eerhardt: I think this line can be lifted outside of the `for` loop. No reason to do it every time through the loop.

docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/ImageClassification.cs
```diff
@@ -34,11 +34,13 @@ public static void Example()
...
                 .ScoreTensorFlowModel(
                 new[] { nameof(OutputScores.output) },
                 new[] { nameof(TensorData.input) }, addBatchDimensionInput: true);
-
+            
+
```
codemzs: remove this line and please revert this file.

src/Native/out/build/x64-Debug/.cmake/api/v1/query/client-MicrosoftVS/query.json
```diff
@@ -0,0 +1 @@
+{"requests":[{"kind":"cache","version":2},{"kind":"cmakeFiles","version":1},{"kind":"codemodel","version":2}]}
```
eerhardt: I assume all changes in `src\Native` are not intended. Please revert.

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
```diff
@@ -245,13 +245,14 @@ private void CheckTrainingParameters(DnnRetrainEstimator.Options options)
 
                 tfInputShape = new TensorShape(colTypeDims);
             }
-            if (tfInputShape.NDim != -1)
+            if (tfInputShape.ndim != -1)
```
codemzs: >tfInputShape.ndim [](start = 16, length = 17)

cache this value.

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
codemzs: >//var buffer = tensor[0].Data(); [](start = 12, length = 32)

delete.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -243,7 +243,7 @@ public Tensor ProcessImage(string path)
                     var imageTensor = imageProcessor.ProcessImage(imagePathStr);
                     runner.AddInput(imageTensor, 0);
                     var featurizedImage = runner.Run()[0]; // Reuse memory?
-                    writer.WriteLine(label - 1 + "," + string.Join(",", featurizedImage.Data<float>()));
+                    writer.WriteLine(label - 1 + "," + string.Join(",", featurizedImage.ToArray<float>()));
```
eerhardt: The same comment as above about using `ToArray`. It is usually an indication of a performance problem.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -1023,7 +1018,7 @@ internal sealed class Options : TransformInputBase
             /// <summary>
             /// Location of the TensorFlow model.
             /// </summary>
-            [Argument(ArgumentType.Required, HelpText = "TensorFlow model used by the transform. Please see https://www.tensorflow.org/mobile/prepare_models for more details.", SortOrder = 0)]
+            [Argument(ArgumentType.Required, HelpText = "TensorFlow model used by the transform. Please see https://www.tf.org/mobile/prepare_models for more details.", SortOrder = 0)]
```
eerhardt: I can't access this new URL. Is it correct?

src/Microsoft.ML.Dnn/Microsoft.ML.Dnn.csproj
```diff
@@ -16,7 +16,7 @@
     <PackageReference Include="SciSharp.TensorFlow.Redist" Version="1.14.0" />
     <PackageReference Include="System.IO.FileSystem.AccessControl" Version="$(SystemIOFileSystemAccessControl)" />
     <PackageReference Include="System.Security.Principal.Windows" Version="$(SystemSecurityPrincipalWindows)" />
-    <PackageReference Include="TensorFlow.NET" Version="0.10.10" />
+    <PackageReference Include="TensorFlow.NET" Version="0.11.3" />
```
eerhardt: Don't forget to also update the file in the `pkg`.  Usually we put all the dependency versions in a single file - https://github.com/dotnet/machinelearning/blob/master/build/Dependencies.props. That way they only need to be updated in a single place.

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -428,7 +428,7 @@ private protected override void SaveModel(ModelSaveContext ctx)
                 var buffer = Session.graph.ToGraphDef(status);
                 ctx.SaveBinaryStream("TFModel", w =>
                 {
-                    w.WriteByteArray(buffer.Data);
+                    w.WriteByteArray(buffer.MemoryBlock.ToArray());
```
eerhardt: Same comment about `ToArray`

I'll stop commenting on them, and assume you will test the perf before and after these changes. And fix any perf degradations.

src/Microsoft.ML.Dnn/DnnUtils.cs
```diff
@@ -92,11 +92,14 @@ internal static Session LoadTFSession(IExceptionContext ectx, byte[] modelBytes,
...
-                tf.train.import_meta_graph(path);
-                return graph;
-            });
+
+            var graph = new Graph();
+            graph = graph.as_default();
+
+            tf.train.import_meta_graph(path);
+
+            return graph;
```
codemzs: remove new lines.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -530,14 +526,14 @@ private void VariableSummaries(RefVariable var)
         private (Operation, Tensor, Tensor, Tensor) AddFinalRetrainOps(int classCount, string labelColumn,
             string scoreColumnName, float learningRate, Tensor bottleneckTensor, bool isTraining)
         {
-            var (batch_size, bottleneck_tensor_size) = (bottleneckTensor.TensorShape.Dimensions[0], bottleneckTensor.TensorShape.Dimensions[1]);
+            var (batch_size, bottleneck_tensor_size) = (bottleneckTensor.TensorShape.dims[0], bottleneckTensor.TensorShape.dims[1]);
```
codemzs: >dims [](start = 85, length = 4)

cache

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -757,7 +752,7 @@ private protected override void SaveModel(ModelSaveContext ctx)
             var buffer = _session.graph.ToGraphDef(status);
             ctx.SaveBinaryStream("TFModel", w =>
             {
-                w.WriteByteArray(buffer.Data);
+                w.WriteByteArray(buffer.MemoryBlock.ToArray());
```
codemzs: >w.WriteByteArray(buffer.MemoryBlock.ToArray()); [](start = 16, length = 47)

can you please do a diff between the graph with your change and master and see if there are any changes? I know the graph size increased and I want to know why.

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -380,7 +380,7 @@ internal static (TF_DataType[] tfOutputTypes, DataViewType[] outputTypes, (Opera
                 // i.e. the first dimension (if unknown) is assumed to be batch dimension.
                 // If there are other dimension that are unknown the transformer will return a variable length vector.
                 // This is the work around in absence of reshape transformer.
-                int[] dims = shape.NDim > 0 ? shape.Dimensions.Skip(shape[0] == -1 ? 1 : 0).ToArray() : new[] { 0 };
+                int[] dims = shape.ndim > 0 ? shape.dims.Skip(shape.dims[0] == -1 ? 1 : 0).ToArray() : new[] { 0 };
```
codemzs: >dims [](start = 52, length = 4)

cache

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -534,18 +534,18 @@ private sealed class Mapper : MapperBase
                             throw Contracts.Except($"Input shape mismatch: Input '{_parent.Inputs[i]}' has shape {originalShape.ToString()}, but input data is of length {typeValueCount}.");
 
                         // Fill in the unknown dimensions.
-                        var l = new int[originalShape.NDim];
-                        for (int ishape = 0; ishape < originalShape.NDim; ishape++)
-                            l[ishape] = originalShape[ishape] == -1 ? (int)d : originalShape[ishape];
+                        var l = new int[originalShape.ndim];
+                        for (int ishape = 0; ishape < originalShape.ndim; ishape++)
```
codemzs: >ndim [](start = 68, length = 4)

cache

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -534,18 +534,18 @@ private sealed class Mapper : MapperBase
...
+                            l[ishape] = originalShape.dims[ishape] == -1 ? (int)d : originalShape.dims[ishape];
```
codemzs: >dims [](start = 98, length = 4)

cache

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -635,6 +635,7 @@ private void UpdateCacheIfNeeded(long position, ITensorValueGetter[] srcTensorGe
             {
                 if (outputCache.Position != position)
                 {
+                    _parent.Session.graph.as_default();
```
codemzs: >_parent.Session.graph.as_default(); [](start = 20, length = 35)

This line is setting the graph as the default graph every time the cursor moves to the new row. We just need to do this once in the thread of the cursor. Please fix, this is call is an interop call and can reduce performance at scoring.

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
```diff
@@ -382,10 +385,9 @@ private void TrainCore(DnnRetrainEstimator.Options options, IDataView input, IDa
                 runner.AddInput(inputs[i], srcTensorGetters[i].GetBufferedBatchTensor());
 
             Tensor[] tensor = runner.Run();
-            var buffer = tensor[0].Data();
-            loss = tensor.Length > 0 && tensor[0] != IntPtr.Zero ? (float)tensor[0].Data<float>()[0] : 0.0f;
-            metric = tensor.Length > 1 && tensor[1] != IntPtr.Zero ? (float)tensor[1].Data<float>()[0] : 0.0f;
-            var b = tensor.Length > 2 && tensor[2] != IntPtr.Zero ? (float[])tensor[2].Data<float>() : null;
+            loss = tensor.Length > 0 && tensor[0] != IntPtr.Zero ? (float)tensor[0].ToArray<float>()[0] : 0.0f;
+            metric = tensor.Length > 1 && tensor[1] != IntPtr.Zero ? (float)tensor[1].ToArray<float>()[0] : 0.0f;
+            var b = tensor.Length > 2 && tensor[2] != IntPtr.Zero ? (float[])tensor[2].ToArray<float>() : null;
```
eerhardt: I don't see `b` being used anywhere. Can it be removed?

docs/samples/Microsoft.ML.Samples/Dynamic/TensorFlow/ImageClassification.cs
```diff
@@ -38,7 +38,7 @@ public static void Example()
                 .ScoreTensorFlowModel(
                 new[] { nameof(OutputScores.output) },
                 new[] { nameof(TensorData.input) }, addBatchDimensionInput: true);
-
+
```
eerhardt: Can you revert this file? Changing files unnecessarily bloats the history.

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
ashbhandare: The API change is from Data() to ToArray(). The Data<T>() already did a memcopy, so the performance would be unchanged with ToArray()

---
In reply to: [323474780](https://github.com/dotnet/machinelearning/pull/4205#discussion_r323474780) [](ancestors = 323474780)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: Replied there.

---
In reply to: [323476010](https://github.com/dotnet/machinelearning/pull/4205#discussion_r323476010) [](ancestors = 323476010)

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
```diff
@@ -857,7 +861,7 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
                 return Utils.MarshalInvoke(MakeGetter<int>, type, input, iinfo, srcTensorGetters, activeOutputColNames, outputCache);
             }
 
-            private Delegate MakeGetter<T>(DataViewRow input, int iinfo, ITensorValueGetter[] srcTensorGetters, string[] activeOutputColNames, OutputCache outputCache)
+            private Delegate MakeGetter<T>(DataViewRow input, int iinfo, ITensorValueGetter[] srcTensorGetters, string[] activeOutputColNames, OutputCache outputCache) where T: unmanaged
```
eerhardt: `where T: unmanaged`

Can this ever be used with a `string` type? If so, this is going to fail at runtime since we use reflection to call this method.

src/Microsoft.ML.TensorFlow/Microsoft.ML.TensorFlow.csproj
```diff
@@ -10,7 +10,7 @@
   <ItemGroup>
     <PackageReference Include="System.IO.FileSystem.AccessControl" Version="$(SystemIOFileSystemAccessControl)" />
     <PackageReference Include="System.Security.Principal.Windows" Version="$(SystemSecurityPrincipalWindows)" />
-    <PackageReference Include="TensorFlow.NET" Version="0.10.10" />
+    <PackageReference Include="TensorFlow.NET" Version="0.11.3" />
```
eerhardt: This line can actually be removed. Since this project references `Microsoft.ML.Dnn`, and `Microsoft.ML.Dnn` references TF.NET, we don't need to say that this project references TF.NET. That way one duplicated version number can be removed.

eerhardt:
You still need to update

https://github.com/dotnet/machinelearning/blob/abfd76dd0c045e20b5d52141b06ae6afba9c4bf3/pkg/Microsoft.ML.Dnn/Microsoft.ML.Dnn.nupkgproj#L12

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
```diff
@@ -18,6 +18,7 @@
 using NumSharp;
 using Tensorflow;
 using static Microsoft.ML.Transforms.Dnn.DnnUtils;
```
yaeldekel: >Microsoft.ML.Transforms.Dnn.DnnUtils [](start = 13, length = 36)

This should be ordered alphabetically.

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
```diff
@@ -811,18 +812,20 @@ private sealed class Mapper : MapperBase
...
 
                         // Fill in the unknown dimensions.
-                        var l = new int[originalShape.NDim];
-                        for (int ishape = 0; ishape < originalShape.NDim; ishape++)
-                            l[ishape] = originalShape[ishape] == -1 ? (int)d : originalShape[ishape];
+                        var originalShapeDims = originalShape.dims;
+                        var originalShapeNdim = originalShape.ndim;
+                        var l = new int[originalShapeNdim];
+                        for (int ishape = 0; ishape < originalShapeNdim; ishape++)
+                            l[ishape] = originalShapeDims[ishape] == -1 ? (int)d : originalShapeDims[ishape];
```
yaeldekel: >(int)d [](start = 74, length = 6)

Should this cast also be done outside the loop?

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
```diff
@@ -912,6 +915,8 @@ private void UpdateCacheIfNeeded(long position, ITensorValueGetter[] srcTensorGe
             {
                 if (outputCache.Position != position)
                 {
+                    if (_parent.Graph.graph_key != tf.get_default_graph().graph_key)
+                        _parent._session.graph.as_default();
```
yaeldekel: What do these lines do?

src/Microsoft.ML.Dnn/DnnUtils.cs
```diff
@@ -92,11 +92,10 @@ internal static Session LoadTFSession(IExceptionContext ectx, byte[] modelBytes,
...
         internal static Graph LoadMetaGraph(string path)
         {
-            return tf_with(tf.Graph().as_default(), graph =>
-            {
-                tf.train.import_meta_graph(path);
-                return graph;
-            });
+            var graph = new Graph();
+            graph = graph.as_default();
+            tf.train.import_meta_graph(path);
```
yaeldekel: >tf.train.import_meta_graph(path); [](start = 12, length = 33)

It seems like there is no connection between `graph` and `tf`, does this line do anything to `graph`?

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -170,7 +170,7 @@ private void CheckTrainingParameters(ImageClassificationEstimator.Options option
             Host.CheckNonWhiteSpace(options.LabelColumn, nameof(options.LabelColumn));
             Host.CheckNonWhiteSpace(options.TensorFlowLabel, nameof(options.TensorFlowLabel));
 
-            if (_session.graph.OperationByName(options.TensorFlowLabel) == null)
+            if (_session.graph.OperationByName(_labelTensor.name.Split(':')[0]) == null)
```
yaeldekel: >_labelTensor.name.Split(':')[0] [](start = 47, length = 31)

If this is the name of the operation you are searching for in the graph, then the error message should mention the correct name. In this case, the options argument is not needed here.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -458,17 +458,13 @@ private IDataView GetShuffledData(string path)
             Tensor evaluationStep = null;
             Tensor prediction = null;
             Tensor bottleneckTensor = evalGraph.OperationByName(_bottleneckOperationName);
-
-            tf_with(evalGraph.as_default(), graph =>
-            {
-                var (_, _, groundTruthInput, finalTensor) = AddFinalRetrainOps(classCount, options.LabelColumn,
+            evalGraph.as_default();
```
yaeldekel: >evalGraph.as_default(); [](start = 12, length = 23)

Does this mutate `evalGraph`? In DnnUtils.cs line 96 you wrote `graph = graph.as_default();`.

yaeldekel: In line 842 in this file there is also a call to `as_default()` without assignment.

---
In reply to: [324450796](https://github.com/dotnet/machinelearning/pull/4205#discussion_r324450796) [](ancestors = 324450796)

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
ashbhandare: VS shows a warning if I reorder. I am guessing the 'using static' derivatives should appear at the end, and they are already ordered alphabetically. 

---
In reply to: [324450008](https://github.com/dotnet/machinelearning/pull/4205#discussion_r324450008) [](ancestors = 324450008)

src/Microsoft.ML.Dnn/DnnUtils.cs
ashbhandare: This creates a new graph and the 'as_default()' sets this as the tf default graph [https://github.com/SciSharp/TensorFlow.NET/blob/41432600c8263cf972b13930c31479ffb412fa65/src/TensorFlowNET.Core/Graphs/Graph.cs#L134](https://github.com/SciSharp/TensorFlow.NET/blob/41432600c8263cf972b13930c31479ffb412fa65/src/TensorFlowNET.Core/Graphs/Graph.cs#L134). The 'graph' is where the meta graph is loaded and returned. 

---
In reply to: [324450631](https://github.com/dotnet/machinelearning/pull/4205#discussion_r324450631) [](ancestors = 324450631)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: the options argument is not used anymore. The _labelTensor has the correct name that the error message would need

---
In reply to: [324450724](https://github.com/dotnet/machinelearning/pull/4205#discussion_r324450724) [](ancestors = 324450724)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: This sets the tensorflow default graph to the evalGraph. This does not mutate the evalGraph.

---
In reply to: [324450796](https://github.com/dotnet/machinelearning/pull/4205#discussion_r324450796) [](ancestors = 324450796)

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
ashbhandare: if the current graph is not the tf default graph, this line sets it as the default graph.

---
In reply to: [324450509](https://github.com/dotnet/machinelearning/pull/4205#discussion_r324450509) [](ancestors = 324450509)

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
ashbhandare: This is not an addition in this PR. If required, I can do this change

---
In reply to: [324450404](https://github.com/dotnet/machinelearning/pull/4205#discussion_r324450404) [](ancestors = 324450404)

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
ashbhandare: I have made a separate function for when T is string to avoid this failure.

---
In reply to: [324317321](https://github.com/dotnet/machinelearning/pull/4205#discussion_r324317321) [](ancestors = 324317321)

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
```diff
@@ -235,7 +236,8 @@ private void CheckTrainingParameters(DnnRetrainEstimator.Options options)
                 inputTensor.InputType(index);
             var tfInputShape = ((Tensor)inputTensor).TensorShape;
 
-            if (isInputVector && (tfInputShape == null || (tfInputShape.NDim == 0)))
+            var numInputDims = tfInputShape.ndim;
```
codemzs: >var numInputDims = tfInputShape.ndim; [](start = 12, length = 37)

It seems tfInputShape can be null. We should set numInputDims to -1 if tfInputShape is null otherwise assign numInputDims = tfInputShape.ndim;

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -843,6 +839,7 @@ public void UpdateCacheIfNeeded()
             protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, bool> activeOutput, out Action disposer)
             {
                 disposer = null;
+                _parent._session.graph.as_default();
```
codemzs: >_parent._session.graph.as_default(); [](start = 16, length = 36)

not needed.

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -580,7 +585,7 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
                 return Utils.MarshalInvoke(MakeGetter<int>, type, input, iinfo, srcTensorGetters, activeOutputColNames, outputCache);
             }
 
-            private Delegate MakeGetter<T>(DataViewRow input, int iinfo, ITensorValueGetter[] srcTensorGetters, string[] activeOutputColNames, OutputCache outputCache)
+            private Delegate MakeGetter<T>(DataViewRow input, int iinfo, ITensorValueGetter[] srcTensorGetters, string[] activeOutputColNames, OutputCache outputCache) where T:unmanaged
             {
```
codemzs: refactor code for unmanged as done in ImageClassificationTranform.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: Required for the Microsoft.ML.Scenarios.TensorFlowScenariosTests.TensorFlowImageClassification

---
In reply to: [324890773](https://github.com/dotnet/machinelearning/pull/4205#discussion_r324890773) [](ancestors = 324890773)

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
ashbhandare: I did the refactor and looks like one of the tests was hitting this code path for string. This means that the original MakeGetter was not erroring out for the type string even after 'where T: unmanaged' was specified. Do we still want to refactor the code?

---
In reply to: [324894953](https://github.com/dotnet/machinelearning/pull/4205#discussion_r324894953) [](ancestors = 324894953)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
codemzs: >_parent._session.graph.as_default(); [](start = 16, length = 36)

is this needed?

ashbhandare: Yes, test Microsoft.ML.Scenarios.TensorFlowScenariosTests.TensorFlowImageClassification is failing without this

---
In reply to: [324938993](https://github.com/dotnet/machinelearning/pull/4205#discussion_r324938993) [](ancestors = 324938993)

PR #4211, opened 2019-09-13
Fixes #3992 and corner cases of inputColumnNames on FeaturizeText

### LINKED ISSUE #3992: FeaturizeText should allow only outputColumnName to be defined

antoniovs1029:
Fixes #3992 , where FeaturizeText is used with an options object, but no inputColumnNames is provided, thus expecting the inputColumnName to be defaulted to the outputColumnName.

It also covers the corner cases where the user uses FeaturizeText with an options object along with a 'null' or an empty string as inputColumnNames.

Three different tests are provided to cover those 3 cases. It is verified that the text is featurized correctly and the features are saved in a column with the same name as the input column. They are all based on the TextFeaturizerWithWordFeatureExtractorTest, but without using a PredictionEngine because the output column hides the input column, and thus instead getting the rows of the dataview to verify their values.

A small fix to the documentation of FeaturizeText is added.


src/Microsoft.ML.Transforms/Text/TextCatalog.cs
```diff
@@ -62,7 +62,10 @@ public static class TextCatalog
             TextFeaturizingEstimator.Options options,
             params string[] inputColumnNames)
             => new TextFeaturizingEstimator(Contracts.CheckRef(catalog, nameof(catalog)).GetEnvironment(),
-                outputColumnName, inputColumnNames, options);
+                outputColumnName,
+                inputColumnNames == null || inputColumnNames.Length == 0 || (inputColumnNames.Length == 1 && inputColumnNames[0] == "")  ?
```
yaeldekel: >inputColumnNames == null || inputColumnNames.Length == 0 [](start = 16, length = 56)

These two conditions can be replaced by `Utils.Size(inputColumnNames) == 0`.

yaeldekel: >inputColumnNames[0] == "" [](start = 109, length = 25)

This should check for `!string.IsNullOrWhiteSpace` instead, but I am actually not sure if we should replace `inputColumnNames` with `new[] {outputColumnName}` in this case, might be better to throw (The constructor of `TextFeaturizingEstimator` will throw in this case, so no need to check for this here).

test/Microsoft.ML.Tests/Transformers/TextFeaturizerTests.cs
```diff
@@ -80,6 +86,133 @@ public void TextFeaturizerWithWordFeatureExtractorTest()
...
+
+            var options = new TextFeaturizingEstimator.Options()
+            {
+                WordFeatureExtractor = new WordBagEstimator.Options() { NgramLength = 1 },
+                CharFeatureExtractor = null,
+                Norm = TextFeaturizingEstimator.NormFunction.None,
+                OutputTokensColumnName = "OutputTokens"
+            };
+
+            var pipeline = ML.Transforms.Text.FeaturizeText("Features", options, null);
```
yaeldekel: >, null [](start = 79, length = 6)

Since it is a `params` argument, you can omit the argument altogether: `ML.Transforms.Text.FeaturizeText("Features", options)'`.

antoniovs1029: Thanks for the feedback. I already provided another test to test when no inputColumnNames is provided (TextFeaturizerWithWordFeatureExtractorTest**WithNoInputNames**). Here in "TextFeaturizerWithWordFeatureExtractorTest**WithNullInputNames**" I provide the case when a null reference is passed as InputColumnNames.

I do this since the [documentation ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.textcatalog.featurizetext?view=ml-dotnet
) explicitly states that "[when the inputColumnNames is] set to null, the value of the outputColumnName will be used as source.", whereas the issue #3992 refers to the case where no inputColumnName is provided.

src/Microsoft.ML.Transforms/Text/TextCatalog.cs
antoniovs1029: @codemzs advised me to also cover the case when an empty string is passed as the only argument for inputColumnNames.

If it is not necessary to cover it, then I can simply remove the code and the test related to this case.

antoniovs1029: Thank you for the feedback, I will change that in my next update.

test/Microsoft.ML.Tests/Transformers/TextFeaturizerTests.cs
antoniovs1029: So, should I leave my pull request as it is regarding this comment, or should I also remove the code and the test case that cover when an user passes a Null as columnInputNames? @yaeldekel

src/Microsoft.ML.Transforms/Text/TextCatalog.cs
yaeldekel: I don't think this case is needed - we are talking about the case where a user explicitly passes an empty string with the intention of having the input column name be identical to the output column name? It doesn't seem necessary, and it doesn't say anywhere in the documentation that this can be done.

---
In reply to: [324773767](https://github.com/dotnet/machinelearning/pull/4211#discussion_r324773767) [](ancestors = 324773767)

test/Microsoft.ML.Tests/Transformers/TextFeaturizerTests.cs
yaeldekel: Hi Antonio, thanks for the explanation. Technically, this test (with the explicit null) checks the `inputColumnNames == null` code path, and the other one (with nothing provided) checks the `inputColumnNames.Length == 0` code path, so you can leave them both if you prefer, I don't have a strong preference.

---
In reply to: [324824556](https://github.com/dotnet/machinelearning/pull/4211#discussion_r324824556) [](ancestors = 324824556)

src/Microsoft.ML.Transforms/Text/TextCatalog.cs
antoniovs1029: Ok. I will remove the code related to this case. Thanks for the feedback.

test/Microsoft.ML.Tests/Transformers/TextFeaturizerTests.cs
antoniovs1029: Ok, I will leave them both. Thanks for the feedback.

Issue #3995, opened 2019-07-12
Exception messages need to be richer\clearer


nicolehaugen:
Many exception messages thrown are unclear - as a result, when an exception occurs, it's challenging to identify whether the issue in with the ML.NET code, with the underlying data, with how the algorithm is being applied, etc.  Often it takes stepping through the ML.NET fwk in attempt to get further context.

I logged this as a single issue because I think there would be benefit in looking at all places where exceptions are being thrown\rethrown to ensure that default exception messages aren't provided and that the messages are as clear\rich as possible.  Let me know if you would like these broken into separate issues rather than having them combined in one.  

Here are some specific examples:

  | Trainer | Scenario | Actual Message | Suggested   Message
-- | -- | -- | -- | --
|1. | N/A | Occurs when invalid field index is provided to the LoadColumn   attribute.         For example:<br>``` [LoadColumn(100)]   public uint Label { get; set; }```<br>In the above code, the value of 100 is an invalid index value since   the underlying data has less than 100 columns. | System.ArgumentNullException: 'Value cannot be null.   Parameter   name: items' | Message should indicate which column has the issue; the reference to   parameter ‘items’ is unclear.|
|2. | N/A | Occurs when Feature column is of some other type than float\single.        For example: <br>``` [ColumnName("Test"), LoadColumn(135)]   public uint Test { get; set; }``` | System.InvalidOperationException:   'Column ‘Test’ has values of UInt32, which is not the same as earlier   observed type of Single.' | It’s unclear what “same as earlier observed type” means.  Consider rewording to state that the   Feature columns must all be of a certain type (e.g. Single).
[...]

PR #4214, opened 2019-09-13
fixing scenerio 3 of issue #3995

Lynx1820:
This fixes the issue #3995 scenario 3 
I'm waiting to see what can be done about scenario 1(desired behavior not decided yet) and 2( can't replicate). Scenario 4 is not part of our code base. 

In this commit, I have made the error comment more clear and switched the exception from an System.ArgumentOutOfRangeException to System.ArgumentNullException. ArgumentOutOfRangeException are reserved for signaling bad function parameters and thus doesn't make sense to use for signaling a missing column value. 

### DIFF
--- a/src/Microsoft.ML.LightGbm/LightGbmRankingTrainer.cs
+++ b/src/Microsoft.ML.LightGbm/LightGbmRankingTrainer.cs
@@ -222,7 +222,8 @@ namespace Microsoft.ML.Trainers.LightGbm
                     $"Label column '{labelCol.Name}' is of type '{labelType.RawType}', but must be Key or Single.");
             }
             // Check group types.
-            ch.CheckParam(data.Schema.Group.HasValue, nameof(data), "Need a group column.");
+            if(!data.Schema.Group.HasValue)
+                throw ch.ExceptValue(nameof(data.Schema.Group), "Group column is missing.") ;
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.LightGbm/LightGbmRankingTrainer.cs
```diff
@@ -222,7 +222,8 @@ private protected override void CheckDataValid(IChannel ch, RoleMappedData data)
                     $"Label column '{labelCol.Name}' is of type '{labelType.RawType}', but must be Key or Single.");
             }
             // Check group types.
-            ch.CheckParam(data.Schema.Group.HasValue, nameof(data), "Need a group column.");
+            if(!data.Schema.Group.HasValue)
+                throw ch.ExceptValue(nameof(data.Schema.Group), "Group column is missing.") ;
```
yaeldekel:
```
if (!condition)
    throw ...
```
is the same as doing a check on `condition`.

Lynx1820: Hi Yael! Checkparam throws an ArgumentOutOfRangeException, I switched to checking manually, so I can throw a different type of exception. There's also a checkvalue function, but it only works for checking classes.

PR #4219, opened 2019-09-16
Added onnx export support for CopyColumns

### LINKED ISSUE #4218: CopyColumns does not support saving to Onnx
- **What did you do?**
Created a pipeline from with CopyColumns and tried to export it to Onnx with ConvertToOnnxProtobuf.

- **What happened?**
The transform wasn't saved in the onnx graph

- **What did you expect?**
The transform should be saved in the onnx graph

harishsk:
Fixes #4218


harishsk:
@codemzs / @gani: Can I please get a code review on this PR?

src/Microsoft.ML.Data/Transforms/ColumnCopying.cs
```diff
@@ -233,15 +233,21 @@ protected override DataViewSchema.DetachedColumn[] GetOutputColumnsCore()
...
                 {
-                    var srcVariableName = ctx.GetVariableName(column.inputColumnName);
-                    _schema.TryGetColumnIndex(column.inputColumnName, out int colIndex);
+                    var inputColumnName = column.inputColumnName;
+                    if (!ctx.ContainsColumn(inputColumnName))
+                    {
+                        ctx.RemoveColumn(inputColumnName, false);
+                        continue;
+                    }
+
```
codemzs: Was this branch tested? It seems not because this code seems buggy to me as RemoveColumn tries to remove inputColumnName from  columnNameMap but ctx.ContainsColumn(inputColumnName) tells us this key is not present if it is false. It would also be nice if you synced latest to master so that we can see code coverage on this. 

Remove Column is defined as:
        public override void RemoveColumn(string colName, bool removeVariable)
        {
            _host.CheckNonEmpty(colName, nameof(colName));

            if (removeVariable)
            {
                foreach (var val in _intermediateValues)
                {
                    if (val.Name == _columnNameMap[colName])
                    {
                        _intermediateValues.Remove(val);
                        break;
                    }
                }
            }
            _columnNameMap.Remove(colName);
        }

and ContainsColumn(string colName) => _columnNameMap.ContainsKey(colName);

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -846,6 +846,39 @@ void IndicateMissingValuesOnnxConversionTest()
...
+                // Evaluate the saved ONNX model using the data used to train the ML.NET pipeline.
+                string[] inputNames = onnxModel.Graph.Input.Select(valueInfoProto => valueInfoProto.Name).ToArray();
+                string[] outputNames = onnxModel.Graph.Output.Select(valueInfoProto => valueInfoProto.Name).ToArray();
+                var onnxEstimator = mlContext.Transforms.ApplyOnnxModel(outputNames, inputNames, onnxModelPath);
+                var onnxTransformer = onnxEstimator.Fit(dataView);
+                var onnxResult = onnxTransformer.Transform(dataView);
+                CompareSelectedR4ScalarColumns(model.ColumnPairs[0].outputColumnName, outputNames[2], transformedData, onnxResult);
+            }
+            Done();
+        }
```
codemzs: Please write a test where input column is not present.

harishsk: A test that where the column is not present does not trigger line that wasnt covered. Neither does DropColumns. A way to trigger that code path is a contrived code path through an internal API. That line can only be triggered if you call CopyColumns and then also use the internal API to drop that column. But then that leads to a different bug with the output missing from the final model. 
I have restored the earlier behavior which is the correct behavior. 


---
In reply to: [330345862](https://github.com/dotnet/machinelearning/pull/4219#discussion_r330345862) [](ancestors = 330345862)

src/Microsoft.ML.Data/Transforms/ColumnCopying.cs
```diff
@@ -233,15 +233,15 @@ protected override DataViewSchema.DetachedColumn[] GetOutputColumnsCore()
...
-                var opType = "CSharp";
+                var opType = "Identity";
 
                 foreach (var column in _columns)
                 {
-                    var srcVariableName = ctx.GetVariableName(column.inputColumnName);
-                    _schema.TryGetColumnIndex(column.inputColumnName, out int colIndex);
+                    var inputColumnName = column.inputColumnName;
+                    var srcVariableName = ctx.GetVariableName(inputColumnName);
+                    _schema.TryGetColumnIndex(inputColumnName, out int colIndex);
```
codemzs: This code seems suspicious, what if inputColumnName is not present in _columnNameMap (because a previous transformer removed it as it might have been an input that transformer and that transformer was not convertible to ONNX)? and if that is the case then you will error out 242-243. This is why we need to test for negative cases.

harishsk: The onnx converter walks the transformer chain backwards and if it encounters a transformer that cannot export to Onnx, it stops processing and returns only those set of transformers it has processed so far. (That is broken behavior in my mind). 

I have tried several different ways of configuring a pipeline (including DropColumns before and after the CopyColumns stage, and using the private API to explicitly include omitted columns in the call) and I have not found a way to craft a working pipeline where that case is triggered. 

Let us say the input schema contains columns [A, B, C] and I have a pipeline to drop column B and copy C to D. Here B would get dropped from the onnx context, but the ColumnCopyingTransformer doesn't care. Its column map will only contain C->D and it will perform that copy. 

The only way to trigger that case is to have the dropped column be part of the column map for the CopyColumn transformer. That results in a weird pipeline where you are trying to drop a column and copy it at the same time. Even if we accept that weird behavior, it results in an exception because the CopyColumnTransformer won't be able to find the column to copy.

It is possible that I have got some of this analysis wrong. I would greatly appreciate seeing a pipeline that would trigger this case.


---
In reply to: [331255780](https://github.com/dotnet/machinelearning/pull/4219#discussion_r331255780) [](ancestors = 331255780)

harishsk: With a schema of columns [A|B] and a pipeline consisting of DropColumns(A) + CopyColumns(C, B) resulting in [B|C] will export to Onnx. But it still doesn't trigger that condition. As we discussed, for that condition to trigger, we need a Transform that returns true for CanSaveAsOnnx but in the call to SaveAsOnnx fails and drops columns from the onnx context. I have not found any such transform that has partial failure. Please let me know if there is such a transform.

---
In reply to: [331290041](https://github.com/dotnet/machinelearning/pull/4219#discussion_r331290041) [](ancestors = 331290041,331255780)

PR #4220, opened 2019-09-17
Adding code generation logic to skip generating properties for ignored columns

tbombach:
Adding code generation logic to skip generating properties for ignored columns in the Input Model. Ignored columns are specified by the code gen caller in the `ColumnInferenceResults.ColumnInformation` that is passed to the CodeGenerator.

- Adding unit tests to verify that ignored columns do not generate properties.

**Note - I also verified that this change works as expected with a fully generated project.

Fixes https://github.com/dotnet/machinelearning/issues/4224

### DIFF
--- a/src/Microsoft.ML.CodeGen/CodeGenerator/CSharp/CodeGenerator.cs
+++ b/src/Microsoft.ML.CodeGen/CodeGenerator/CSharp/CodeGenerator.cs
@@ -218,6 +218,10 @@ namespace Microsoft.ML.CodeGenerator.CSharp
             IList<string> result = new List<string>();
             foreach (var column in _columnInferenceResult.TextLoaderOptions.Columns)
             {
+                if (this.columnInferenceResult.ColumnInformation.IgnoredColumnNames.Contains(column.Name))
+                {
+                    continue;
+                }
                 StringBuilder sb = new StringBuilder();
                 int range = (column.Source[0].Max - column.Source[0].Min).Value;
                 bool isArray = range > 0;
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.CodeGen/CodeGenerator/CSharp/CodeGenerator.cs
```diff
@@ -218,6 +218,10 @@ internal IList<string> GenerateClassLabels()
             IList<string> result = new List<string>();
             foreach (var column in _columnInferenceResult.TextLoaderOptions.Columns)
             {
+                if (IsColumnIgnored(column.Name))
+                {
+                    continue;
```
justinormont: --summary of non-github discussion--

I'd recommend leaving the generation of the unused (Ignore) column in place. I think what you're purposing is "ignore the entire existence of the column" vs. the current "just don't use it for featurization".

Generally column are ignored by the AutoML heuristics. These, while generally good, are imperfect. This leads to AutoML ignore column which could be useful.

Removing the column from the input class makes the column no longer appear in the IDataView, which means users won't be able to easily add back the missing column. 

When the user explicitly ignores a column, it's often done due to the column leaking the label. When ignoring a column from the CLI, after CodeGen, the user can handle this column without leaking.

eerhardt: From a user's perspective, I find it surprising that these columns still make it into the code generation when I told the tool to exclude this column. My opinion is we should be removing those columns.

Maybe it would be best to have an option to include/exclude these columns or not. That way the tool or user can decide whether they want these columns generated or not.

PR #4223, opened 2019-09-17
Fix memory leak in TensorflowTransform

### LINKED ISSUE #4134: Memory leak in the latest ML.NET 1.3.1

KsenijaS:
This PR fixes two memory leaks:
1. After executing the graph, input tensor are not being freed
2. When running multiple pipelines (sessions), graphs are not being freed.

This fixes #4134

### DIFF
--- a/src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
+++ b/src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
@@ -645,10 +645,12 @@ namespace Microsoft.ML.Transforms
                     Runner runner = new Runner(_parent.Session);
 
                     // Feed inputs to the graph.
+                    List<Tensor> inputTensors = new List<Tensor>();
                     for (int i = 0; i < _parent.Inputs.Length; i++)
                     {
                         var tensor = srcTensorGetters[i].GetTensor();
                         runner.AddInput(_parent.Inputs[i], tensor);
+                        inputTensors.Add(tensor);
                     }
 
                     // Add outputs.
@@ -657,6 +659,11 @@ namespace Microsoft.ML.Transforms
 
                     // Execute the graph.
                     var tensors = runner.Run();
+
+                    foreach (Tensor inputTensor in inputTensors)
+                    {
+                        inputTensor.Dispose();
+                    }
 
                     Contracts.Assert(tensors.Length > 0);
 
--- a/src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
+++ b/src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
@@ -463,6 +463,11 @@ namespace Microsoft.ML.Transforms
                 {
                     Session.close(); // invoked Dispose()
                 }
+
+                if (Session != null && Session.graph != IntPtr.Zero)
+                {
+                    Session.graph.Dispose();
+                }
             }
             finally
             {
--- a/src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
+++ b/src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
@@ -650,12 +650,10 @@ namespace Microsoft.ML.Transforms
                     Runner runner = new Runner(_parent.Session);
 
                     // Feed inputs to the graph.
-                    List<Tensor> inputTensors = new List<Tensor>();
-                    for (int i = 0; i < _parent.Inputs.Length; i++)
+                     for (int i = 0; i < _parent.Inputs.Length; i++)
                     {
                         var tensor = srcTensorGetters[i].GetTensor();
                         runner.AddInput(_parent.Inputs[i], tensor);
-                        inputTensors.Add(tensor);
                     }
 
                     // Add outputs.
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -463,6 +463,11 @@ private void Dispose(bool disposing)
                 {
                     Session.close(); // invoked Dispose()
                 }
+
+                if (Session != null && Session.graph != IntPtr.Zero)
+                {
+                    Session.graph.Dispose();
```
eerhardt: Check out the comment at the top of this method, and this article: [https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#the-disposeboolean-overload](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#the-disposeboolean-overload)

I'm not sure we should be disposing other managed objects if this object is being finalized (i.e. disposing is `false`). That object might have already been finalized, which may lead to unintended errors.

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -645,10 +650,12 @@ private void UpdateCacheIfNeeded(long position, ITensorValueGetter[] srcTensorGe
                     Runner runner = new Runner(_parent.Session);
 
                     // Feed inputs to the graph.
+                    List<Tensor> inputTensors = new List<Tensor>();
                     for (int i = 0; i < _parent.Inputs.Length; i++)
                     {
                         var tensor = srcTensorGetters[i].GetTensor();
                         runner.AddInput(_parent.Inputs[i], tensor);
+                        inputTensors.Add(tensor);
```
eerhardt: Instead of adding these to another list - would it be better to just get the inputs from the `runner` and dispose of them?

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
codemzs: @eric.erhardt@microsoft.com We need to call dispose explicitly on the graph because it is not disposed by the destructor. TF .NET disposes the graph if you use their session layer which we don't. 

---
In reply to: [325323057](https://github.com/dotnet/machinelearning/pull/4223#discussion_r325323057) [](ancestors = 325323057)

codemzs: @eerhardt  We need to call dispose explicitly on the graph because it is not disposed by the destructor. TF .NET disposes the graph if you use their session layer which we don't. 
CC: @Oceania2018

eerhardt: You shouldn't be touching other objects if `disposing` is `false`.

Please read 
https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#the-disposeboolean-overload 

and 

https://ericlippert.com/2015/05/21/when-everything-you-know-is-wrong-part-two/.  "Myth: An object being finalized can safely access another object."

> This myth follows directly from the previous. If you have a tree of objects and you are finalizing the root, then the children are still alive — because the root is alive, because it is on the finalization queue, and so the children have a living reference — but the children may have already been finalized, and are in no particularly good state to have their methods or data accessed.

___________

If the GC is finalizing this object, the `Session` may have gotten GC'd already, along with it's `.graph`. If that happens, and you try to dereference it here and call methods on it, this code will throw an exception and bring down the process. As stated in both those links, you can't access other objects in your finalizer (i.e. when `disposing` is `false`).

The `graph` object here should also implement a finalizer, and clean up unmanaged resources when it is finalized.

codemzs: I have read the article and also understand your point. Graph does not seem to define a destructor but that is ok, I can look into making this transform disposable so that when pipeline goes out of scope it calls dispose on this where we dispose the graph. @oceania2018 can we please have destructor in Graph that frees unmanaged memory?

Oceania2018: @codemzs `Graph` inherits from `DisposableObject`, so it has destructor. You just need to call `Graph.Dispose()` when you don't need it, that's it.

Issue #4169, opened 2019-09-02
[Image Classification DNN Transfer Learning] - Use PredictedLabel as String/Value instead of an UInt32 (Index))


CESARDELATORRE:
I think it'd be simpler for users and also consistent with other ML.NET APIs if the returned PredictedLabel is a string/value (categorial value) instead UInt32 (Index).

That way, simply by having a .MapKeyToValue() at the end of the pipeline, the user will have it as the value being looked for instead of an index which is what we currently get:

```
//pipeline code
.Append(mlContext.Transforms.Conversion.MapKeyToValue(outputColumnName: "Label" , inputColumnName: "LabelAsKey"));
```

As currently implemented, when the user is simply predicting in an end-user app, you need to find out the real label value by using code based on the schema API, like this:

```
var prediction = predictionEngine.Predict(imageToPredict);

var index = prediction.PredictedLabel;

// Obtain the original label names to map through the predicted label-index
VBuffer<ReadOnlyMemory<char>> keys = default;
predictionEngine.OutputSchema["LabelAsKey"].GetKeyValues(ref keys);
var originalLabels = keys.DenseValues().ToArray();

Console.WriteLine($"ImageFile : [{Path.GetFileName(imageToPredict.ImagePath)}], " +
                    $"Scores : [{string.Join(",", prediction.Score)}], " +
                    $"Predicted Label : {originalLabels[index]}");
```

If the PredictedLabel was already the string/value, the user won't usually need to use the additional `VBuffer `and `OutputSchema `APIs above.

Also, this way would be consistent with other multi-class classification algorithms in ML.NET.

codemzs:
@harshithapv is working on this.

PR #4228, opened 2019-09-18
Added categorical value support for PredictedLabel for the Image Classification Transfer Learning example. Addresses Issue #4169

mstfbl:
Attempt on Issue #4169

- Added the necessary estimator pipeline for a KeyDataViewType string/value binding
- Changed the Output Column types of the ImageClassificationEstimator and  the Mapper for future compatibility with KeyType composite values
- Added test cases to test features added above


src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -878,7 +878,7 @@ protected override DataViewSchema.DetachedColumn[] GetOutputColumnsCore()
             {
                 var info = new DataViewSchema.DetachedColumn[_parent._outputs.Length];
                 info[0] = new DataViewSchema.DetachedColumn(_parent._outputs[0], new VectorDataViewType(NumberDataViewType.Single, _parent._classCount), null);
-                info[1] = new DataViewSchema.DetachedColumn(_parent._outputs[1], NumberDataViewType.UInt32, null);
+                info[1] = new DataViewSchema.DetachedColumn(_parent._outputs[1], new KeyDataViewType(typeof(uint), _parent._classCount), ((DataViewSchema.Column)InputSchema.GetColumnOrNull("Label")).Annotations);
```
codemzs: >Label [](start = 190, length = 5)

You are making an assumption that label column will always be called "Label". This assumption is incorrect.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -1166,7 +1166,7 @@ internal ImageClassificationEstimator(IHostEnvironment env, Options options, Dnn
             _options = options;
             _dnnModel = dnnModel;
             _tfInputTypes = new[] { TF_DataType.TF_STRING };
-            _outputTypes = new[] { new VectorDataViewType(NumberDataViewType.Single), NumberDataViewType.UInt32.GetItemType() };
+            _outputTypes = new DataViewType[] { new VectorDataViewType(NumberDataViewType.Single), new KeyDataViewType(typeof(uint), 5) };
```
codemzs: >5 [](start = 133, length = 1)

You are hard coding the number of classes as 5, this will vary from scenario to scenario.

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs
```diff
@@ -46,9 +46,13 @@ public static void Example()
...
-                shuffledFullImagesDataset = mlContext.Transforms.Conversion
-                    .MapValueToKey("Label")
-                    .Fit(shuffledFullImagesDataset)
+                var estimator = mlContext.Transforms.Conversion
+                    .MapValueToKey("Label");
+                var estimatorWithKeyType = estimator.Append(
+                    mlContext.Transforms.Conversion.MapKeyToValue(
+                        outputColumnName: "LabelAsKey", inputColumnName: "Label"));
+
+                shuffledFullImagesDataset = estimatorWithKeyType.Fit(shuffledFullImagesDataset)
```
codemzs: There shouldn't a need to modify this part of code. You will just need to add a transform after ImageClassification that gets Key mappings from PredictedLabel column by doing MapKeyToValue.

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs
```diff
@@ -93,15 +97,16 @@ public static void Example()
                 DataViewSchema schema;
                 using (var file = File.OpenRead("model.zip"))
                     loadedModel = mlContext.Model.Load(file, out schema);
+                // the schema in line 99 and the shuffledFullImagesDataset.Schema in line 93 don't have
+                // the same annotations.
```
codemzs: why are you adding these comments here?

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs
```diff
@@ -93,15 +97,16 @@ public static void Example()
...
 
                 EvaluateModel(mlContext, testDataset, loadedModel);
 
                 VBuffer<ReadOnlyMemory<char>> keys = default;
                 loadedModel.GetOutputSchema(schema)["Label"].GetKeyValues(ref keys);
```
codemzs: shouldn't need to do all this if you convert predicted label column from Key to Value. It will output string type that will contain class names that corresponds to key indices.

codemzs:
>             public UInt32 PredictedLabel;

this needs to be string. 

---
Refers to: docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs:308 in 8d07401. [](commit_id = 8d07401eeb8726a3fa975da661acec55b6cbddf1, deletion_comment = False)

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs
```diff
@@ -135,6 +139,7 @@ public static void Example()
             };
 
             var prediction = predictionEngine.Predict(imageToPredict);
+            var predictedLabelsKeyType = ((DataViewSchema.Column)schema.GetColumnOrNull("Label")).Annotations;
             var index = prediction.PredictedLabel;
```
codemzs: No need for all this.

codemzs:
>                 $"Predicted Label : {originalLabels[index]}");

prediction.PredictedLabel. 

---
Refers to: docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs:148 in 8d07401. [](commit_id = 8d07401eeb8726a3fa975da661acec55b6cbddf1, deletion_comment = False)

yaeldekel:
>                     : SchemaShape.Column.VectorKind.VariableVector, _outputTypes[i].GetItemType(), false);

This is wrong. Given how `_outputTypes` is initialized in line 1169, the outcome of the condition here is always `false`: `_outputType[0]` is a variable length vector, and `_outputTypes[1]` is a scalar. This would make the result always be of kind `SchemaShape.Column.VectorKind.VariableVector`, which is not the case.

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTransform.cs:1203 in 8d07401. [](commit_id = 8d07401eeb8726a3fa975da661acec55b6cbddf1, deletion_comment = False)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
yaeldekel: >_outputTypes [](start = 12, length = 12)

`_outputTypes` doesn't need to be an array of `DataViewType`. All the estimator needs to know is whether the length of the score vector is variable or not (seems to me that it is typically not variable, or at least this information can be inferred from the `DnnModel`).
If you get rid of this field, you will not have to guess the size of the key (which is also not needed by the estimator).

yaeldekel:
>             for (var i = 0; i < _options.OutputColumns.Length; i++)

How does this work?
It seems like there is an assumption here that the output columns array is always of length 2, and always contains the score column and the predicted label column, in that order. If this is true, then this argument needs to be replaced with two `string` arguments for the score column name and the predicted label column name.

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTransform.cs:1199 in 8d07401. [](commit_id = 8d07401eeb8726a3fa975da661acec55b6cbddf1, deletion_comment = False)

yaeldekel:
>                     : SchemaShape.Column.VectorKind.VariableVector, _outputTypes[i].GetItemType(), false);

This needs to output exactly two `SchemaShape.Column`s: The first with the score column name, of kind `VectorKind.Vector`, and the other with the predicted label column name, of kind `VectorKind.Scalar`, and with `isKey` set to true.

---
In reply to: [533174369](https://github.com/dotnet/machinelearning/pull/4228#issuecomment-533174369) [](ancestors = 533174369)

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTransform.cs:1203 in 8d07401. [](commit_id = 8d07401eeb8726a3fa975da661acec55b6cbddf1, deletion_comment = False)

yaeldekel:
>             _labelColumnName = labelColumnName;

This is not used anywhere but at serialization time. Why does this need to be serialized?

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTransform.cs:667 in 8d07401. [](commit_id = 8d07401eeb8726a3fa975da661acec55b6cbddf1, deletion_comment = False)

yaeldekel:
>             bool? addBatchDimensionInput, int batchSize, string labelColumnName, string finalModelPrefix, Architecture arch,

This is where we have both the label column name and the training data schema, so this is where we need to extract the key values (if they exist). They will need to be serialized with the transformer, because at load time we don't have access to the training data.

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTransform.cs:652 in 8d07401. [](commit_id = 8d07401eeb8726a3fa975da661acec55b6cbddf1, deletion_comment = False)

yaeldekel:
>             bool? addBatchDimensionInput, int batchSize, string labelColumnName, string finalModelPrefix, Architecture arch,

I think this argument is not used here.

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTransform.cs:652 in 8d07401. [](commit_id = 8d07401eeb8726a3fa975da661acec55b6cbddf1, deletion_comment = False)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
mstfbl: This is a good catch, thank you! Quick question, the Image Classification example ResnetV2101TransferLearningTrainTestSplit.cs runs perfectly well when I don't further define _outputTypes, i,e, delete line 69. Does this also mean than the estimator isn't using this field?

mstfbl: Actually, I've found that the DataviewType field is needed for the pipeline in ResnetV2101TransferLearningTrainTestSplit.cs to fit properly.

mstfbl:
With my latest commit, I have isolated the issue to the annotations of KeyType not copying over to the SchemaShapes that are being created in line 194 of IEstimator.cs and subsequently in line 61 of EstimatorChains.cs, which are part of the Fitting process in line 81 of ResnetV2101TransferLearningTrainTestSplit.cs . If we can incorporate a way of preserving the original KeyType annotations after the process, we can achieve a KeyType implementation when the Prediction Engine is run in the TrySinglePrediction method of ResnetV2101TransferLearningTrainTestSplit.cs.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
mstfbl: It was giving a different error when I removed _outputTypes, but as I now see that was not the bottleneck problem I was having while implementing this KeyType solution.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -878,7 +878,7 @@ protected override DataViewSchema.DetachedColumn[] GetOutputColumnsCore()
             {
                 var info = new DataViewSchema.DetachedColumn[_parent._outputs.Length];
                 info[0] = new DataViewSchema.DetachedColumn(_parent._outputs[0], new VectorDataViewType(NumberDataViewType.Single, _parent._classCount), null);
-                info[1] = new DataViewSchema.DetachedColumn(_parent._outputs[1], NumberDataViewType.UInt32, null);
+                info[1] = new DataViewSchema.DetachedColumn(_parent._outputs[1], new KeyDataViewType(typeof(uint), _parent._classCount), ((DataViewSchema.Column)InputSchema.GetColumnOrNull(_parent._labelColumnName)).Annotations);
```
yaeldekel: >_labelColumnName [](start = 197, length = 16)

We cannot rely on the `_labelColumnName` column in the transformer, since it may not be present in `InputSchema`.
The way this should work is by getting the key values at the time we have the training data (which is required to contain the label column), which is in the constructor of the transformer. The key values should be serialized as part of saving the transformer, so that they are available even when the model is loaded from disk.
You should also be able to remove the `_labelColumnName` field from the transformer, it is not needed.

src/Microsoft.ML.Data/DataLoadSave/EstimatorChain.cs
```diff
@@ -57,7 +57,8 @@ public EstimatorChain()
         public TransformerChain<TLastTransformer> Fit(IDataView input)
         {
             // Before fitting, run schema propagation.
-            GetOutputSchema(SchemaShape.Create(input.Schema));
+            SchemaShape schemaShape = SchemaShape.Create(input.Schema);
+            GetOutputSchema(schemaShape);
```
yaeldekel: Why is this changed?

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
yaeldekel: >_outputs [](start = 68, length = 8)

Please replace this field with two fields containing the score column name and the predicted label column name, and do the same for the argument in the `Options` class.
It is implicitly assumed throughout this code that there are exactly two output columns, the first one being the score column and the second one being the predicted label column - for example, this method (`GetOutputColumnsCore`), `GetOutputSchema` in the estimator, `GetOutputColumnsCore` in the mapper.

src/Microsoft.ML.Data/DataLoadSave/EstimatorChain.cs
mstfbl: I had changed this line to see the difference in the stored annotations between input.Schema and schemaShape, and experimented in ways to store the annotations at this step in the code.

yaeldekel:
`SchemaShape` is a class that holds some partial information about the schema that the estimators can provide. Estimators are not able to give the exact schema, since they do not see any data until `Fit` is called. For this reason, they are only able to give information such as:
* The column that is created is a fixed length vector (what is the length? the estimator doesn't know, it can know only when it sees the data).
* The column that is created is a key type, and it has an annotation of key values that is a fixed length vector of text (what is the length of the key values vector and what are the actual key values? There is no way to know until we see the data).

Each estimator is in charge of creating the `SchemaShape` of its output. The `EstimatorChain` cannot know that it needs to copy annotations from a different column (from the label column to the predicted label column), the `ImageClassificationEstimator` is the component that has to do that, in its implementation of `GetOutputSchema`. Same with the transformer - the `ImageClassificationTransformer` needs to store the key values from the label column when it sees the training data, and pass them to the predicted label column.

---
In reply to: [533441383](https://github.com/dotnet/machinelearning/pull/4228#issuecomment-533441383) [](ancestors = 533441383)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
yaeldekel: I just noticed that there is the same problem with the input column names - on one hand the argument in the `Options` class is an array, and in other place it is assumed that it contains exactly one element which is the features column.

---
In reply to: [326619105](https://github.com/dotnet/machinelearning/pull/4228#discussion_r326619105) [](ancestors = 326619105)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -650,7 +657,7 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Dat
         internal ImageClassificationTransformer(IHostEnvironment env, Session session, string[] outputColumnNames,
             string[] inputColumnNames, string modelLocation,
             bool? addBatchDimensionInput, int batchSize, string labelColumnName, string finalModelPrefix, Architecture arch,
-            string scoreColumnName, string predictedLabelColumnName, float learningRate, DataViewSchema inputSchema, int? classCount = null, bool loadModel = false,
+            string scoreColumnName, string predictedLabelColumnName, float learningRate, DataViewSchema inputSchema, int? classCount = null, string[] keyValueAnnotations = null, bool loadModel = false,
```
yaeldekel: >keyValueAnnotations [](start = 150, length = 19)

The transformer can be created in one of two ways: Either by calling `Fit` on the estimator, or by deserializing from file. In the second case, the key values come from reading them from the model as you have written. In the first case, you need to get them from `inputSchema` by getting the key value annotations of the training label column.

yaeldekel: Notice that it is possible that the label column does not contain key values annotations, which means that the PredictedLabel column will not have them either. You have to account for that at serialization/deserialization time (for example, by serialiazing the number of key value strings, and only then serializing the strings them selves).

---
In reply to: [327481050](https://github.com/dotnet/machinelearning/pull/4228#discussion_r327481050) [](ancestors = 327481050)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -684,6 +692,9 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Dat
                     throw Host.ExceptSchemaMismatch(nameof(inputSchema), "label", (string)labelColumn.Name, "Key", (string)labelType.ToString());
 
                 _classCount = labelCount == 1 ? 2 : (int)labelCount;
+
+                //Temporary string array for key values, will replace soon
+                _keyValueAnnotations = new string[] { "sunflowers", "roses", "dandelion", "daisy", "tulips"};
```
yaeldekel: >_keyValueAnnotations [](start = 16, length = 20)

It seems that the `loadModel` argument specifies whether we are training the model or loading it from a file, inside the if in line 718 is where you should extract the key values from the input schema.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -684,6 +692,9 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Dat
                     throw Host.ExceptSchemaMismatch(nameof(inputSchema), "label", (string)labelColumn.Name, "Key", (string)labelType.ToString());
 
                 _classCount = labelCount == 1 ? 2 : (int)labelCount;
```
yaeldekel: >labelCount == 1 ? 2 : (int)labelCount [](start = 30, length = 37)

Why?
/cc @codemzs

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -1196,12 +1217,16 @@ public SchemaShape GetOutputSchema(SchemaShape inputSchema)
...
-                resultDic[_options.OutputColumns[i]] = new SchemaShape.Column(_options.OutputColumns[i],
-                    _outputTypes[i].IsKnownSizeVector() ? SchemaShape.Column.VectorKind.Vector
-                    : SchemaShape.Column.VectorKind.VariableVector, _outputTypes[i].GetItemType(), false);
-            }
+
+            resultDic[_options.OutputColumns[0]] = new SchemaShape.Column(_options.OutputColumns[0],
+                    SchemaShape.Column.VectorKind.Vector, _outputTypes[0].GetItemType(), false);
+
+            var metadata = new List<SchemaShape.Column>();
+            metadata.Add(new SchemaShape.Column(AnnotationUtils.Kinds.KeyValues, SchemaShape.Column.VectorKind.Vector, TextDataViewType.Instance, false, inputSchema[1].Annotations));
```
yaeldekel: >, inputSchema[1].Annotations [](start = 151, length = 28)

This is not needed.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -1038,6 +1053,12 @@ internal sealed class Options : TransformInputBase
             [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "The name of the outputs", ShortName = "outputs", SortOrder = 2)]
             public string[] OutputColumns;
 
+            /// <summary>
+            /// The names of the requested key value annotations.
+            /// </summary>
+            [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "The name of the key value annotations", ShortName = "annotations", SortOrder = 3)]
+            public string[] KeyValueAnnotations;
```
yaeldekel: >KeyValueAnnotations [](start = 28, length = 19)

This should not be a user specified argument, it needs to be extracted from the input schema.

yaeldekel:
>                 _transformer = new ImageClassificationTransformer(_host, _options, _dnnModel, input);

Why do we train only once? Throughout the code base `IEstimator`s are stateless, and they train a new transformer each time they are given data. Why is this one different?

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTransform.cs:1240 in 7b239b0. [](commit_id = 7b239b0e9f878113fef6def6579b98fb25325c6c, deletion_comment = False)

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs
```diff
@@ -68,8 +68,8 @@ public static void Example()
                     batchSize: 10,
                     learningRate: 0.01f,
                     metricsCallback: (metrics) => Console.WriteLine(metrics),
-                    validationSet: testDataset);
-
+                    validationSet: testDataset)
+                    .Append(mlContext.Transforms.Conversion.MapKeyToValue(outputColumnName: "PredictedLabel", inputColumnName: "Label"));
```
yaeldekel: >outputColumnName: "PredictedLabel", inputColumnName: "Label") [](start = 74, length = 61)

This is incorrect. What it will do is create a new column called `PredictedLabel` (overwriting the one created by the `ImageClassificationTransformer`), that is the key to value transformer applied to the `Label` column.

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs
```diff
@@ -119,12 +115,11 @@ public static void Example()
...
 
         private static void TrySinglePrediction(string imagesForPredictions,
-            MLContext mlContext, ITransformer trainedModel,
-            ReadOnlyMemory<char>[] originalLabels)
+            MLContext mlContext, ITransformer trainedModel)
         {
             // Create prediction function to try one prediction
             var predictionEngine = mlContext.Model
-                .CreatePredictionEngine<ImageData, ImagePrediction>(trainedModel);
+                .CreatePredictionEngine<ImageData, ImagePrediction>(trainedModel, true);
```
yaeldekel: Why is this changing?

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -720,6 +727,15 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Dat
                 (_evaluationStep, _) = AddEvaluationStep(_softMaxTensor, _labelTensor);
                 _softmaxTensorName = _softMaxTensor.name;
                 _predictionTensorName = _prediction.name;
+
+                VBuffer<ReadOnlyMemory<char>> keysVBuffer = default;
+                inputSchema[labelColumnName].GetKeyValues(ref keysVBuffer);
```
yaeldekel: >GetKeyValues [](start = 45, length = 12)

You need to check whether the label column has key values first.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -720,6 +727,15 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Dat
...
+                _keyValueAnnotations = keysVBuffer.DenseValues().ToArray();
+            }
+            else
```
yaeldekel: >else [](start = 12, length = 4)

I am assuming that this is meant for the case where you are loading the model from file. In that case, you need to use the key values that are deserialized from the model.

In general, I think the way the constructors for this class work is kind of confusing - I think there should be one constructor that is given all the information it needs, either from the deserialization `Create` method, or from the estimator `Fit` method. So everything inside `if (loadModel == false)` would be done in the estimator, as well as figuring out the class count. And the `Create` method for the `SignatureDataTransform` should not create a transformer directly, but instead create an estimator, then call `Fit` and `Transform`.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -749,6 +765,10 @@ private protected override void SaveModel(ModelSaveContext ctx)
             ctx.Writer.Write(_predictedLabelColumnName);
             ctx.Writer.Write(_learningRate);
             ctx.Writer.Write(_classCount);
+
+            for (int j = 0; j < _classCount; j++)
+                ctx.SaveNonEmptyString(_keyValueAnnotations[j]);
```
yaeldekel: This should be done only if `_keyValueAnnotations` is not null (it is possible that the training label column does not contain key values), and you need to serialize a byte before it indicating whether there are key values or not.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -749,6 +765,10 @@ private protected override void SaveModel(ModelSaveContext ctx)
             ctx.Writer.Write(_predictedLabelColumnName);
             ctx.Writer.Write(_learningRate);
             ctx.Writer.Write(_classCount);
+
+            for (int j = 0; j < _classCount; j++)
```
yaeldekel: >_classCount [](start = 32, length = 11)

Can you add an assert here that the length of `_keyValueAnnotations` is equal to `_classCount` or that it is null?

yaeldekel:
>             // stream: tensorFlow model.

Please add a comment about the key value annotations, and also copy the entire comment to `SaveModel`.

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTransform.cs:109 in 33e71ae. [](commit_id = 33e71aeaf08cd9424b00495bb90a519ea2bb9927, deletion_comment = False)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -1211,8 +1288,7 @@ public SchemaShape GetOutputSchema(SchemaShape inputSchema)
         public ImageClassificationTransformer Fit(IDataView input)
         {
             _host.CheckValue(input, nameof(input));
-            if (_transformer == null)
-                _transformer = new ImageClassificationTransformer(_host, _options, _dnnModel, input);
+            _transformer = new ImageClassificationTransformer(_host, _options, _dnnModel, input);
```
yaeldekel: >_transformer [](start = 12, length = 12)

The `_transformer` field is not needed, you can just have:
`return new ImageClassificationTransformer(...)`.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
yaeldekel: Since it is possible that `_keyValueAnnotations` will be null, you need to account for that at serialization/deserialization time.

---
In reply to: [328528099](https://github.com/dotnet/machinelearning/pull/4228#discussion_r328528099) [](ancestors = 328528099)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
yaeldekel: By "this", I mean passing `inputSchema[1].Annotations` to the `SchemaShape.Column` constructor. The annotations of a `SchemaShape.Column` are stored themselves as a `SchemaShape` (since they are similar, in that they have a name and a type). The column you are instantiating here is a column in the `SchemaShape` of the annotations of the `PredictedLabel` column, it does not have annotations of its own.

---
In reply to: [327485460](https://github.com/dotnet/machinelearning/pull/4228#discussion_r327485460) [](ancestors = 327485460)

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs
yaeldekel: It is enough to just pass one string: "PredictedLabel" (when `inputColumnName` is not passed, `outputColumnName` is used for both output and input).

---
In reply to: [327501425](https://github.com/dotnet/machinelearning/pull/4228#discussion_r327501425) [](ancestors = 327501425)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -749,6 +807,12 @@ private protected override void SaveModel(ModelSaveContext ctx)
             ctx.Writer.Write(_predictedLabelColumnName);
             ctx.Writer.Write(_learningRate);
             ctx.Writer.Write(_classCount);
+
+            Host.AssertNonEmpty(_keyValueAnnotations);
+            Host.Assert(_keyValueAnnotations.Length == _classCount);
+            for (int j = 0; j < _classCount; j++)
```
eerhardt: Just a minor FYI - 

In C#, when you index into an array, the index is bounds-checked at runtime. The runtime checks if the index is non-negative, and less than the length of the array. If it is outside those bounds, an exception is thrown. While these bounds checks aren't very expensive, they can add up.
The C# compiler is smart enough to notice this pattern:

```C#
int[] myArray = ...;
for (int i = 0; i < myArray.Length; i++)
{
    int element = myArray[i];
}
```

It sees this pattern, and knows that `i` will never be outside the bounds of `myArray`. So the compiler removes the bounds checks as a performance improvement.

The take away here is, when you have 2 variables to pick from to use in your `for` loop break condition, and one of them is the `.Length` of an array you are indexing into, and one isn't. Choose the `array.Length` variable, so the bounds checks will be removed.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
mstfbl: Yup, in my commit before I fixed that.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
mstfbl: I see, that makes sense, thank you Eric! It's pretty cool that the C# compiler is able to recognize the traversing of an array and modify the bound checks as it goes.

codemzs:
>             _labelColumnName = labelColumnName;

If it is not used anywhere then we should remove it but please confirm, it have been used before and hence it was being serialized. 

---
In reply to: [533181203](https://github.com/dotnet/machinelearning/pull/4228#issuecomment-533181203) [](ancestors = 533181203)

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTransform.cs:667 in 8d07401. [](commit_id = 8d07401eeb8726a3fa975da661acec55b6cbddf1, deletion_comment = False)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
codemzs: Label column can be KeyType, uint32, long but for now just put an assert that label column is keytype and open an issue to make label column be keytype, uint32, long.

---
In reply to: [329080695](https://github.com/dotnet/machinelearning/pull/4228#discussion_r329080695) [](ancestors = 329080695,328528099)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
codemzs: Please open a separate issue for making two constructors, this suggestion is out of the scope of this change. 

---
In reply to: [328535147](https://github.com/dotnet/machinelearning/pull/4228#discussion_r328535147) [](ancestors = 328535147)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
codemzs: To account for binary classification case where in the train set all you have seen is just one class :-). 

---
In reply to: [327483161](https://github.com/dotnet/machinelearning/pull/4228#discussion_r327483161) [](ancestors = 327483161)

yaeldekel:
>             _labelColumnName = labelColumnName;

After deserialization there is no use for the training label column since it might not even exist in new data being passed to the transformer, so if it is being used then it is an error.

---
In reply to: [536237534](https://github.com/dotnet/machinelearning/pull/4228#issuecomment-536237534) [](ancestors = 536237534,533181203)

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTransform.cs:667 in 8d07401. [](commit_id = 8d07401eeb8726a3fa975da661acec55b6cbddf1, deletion_comment = False)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
yaeldekel: Key values annotations can be null even if the column is a key type.

---
In reply to: [329333010](https://github.com/dotnet/machinelearning/pull/4228#discussion_r329333010) [](ancestors = 329333010,329080695,328528099)

mstfbl:
> > ```
> >         _labelColumnName = labelColumnName;
> > ```
> 
> If it is not used anywhere then we should remove it but please confirm, it have been used before and hence it was being serialized.
> 
> In reply to: [533181203](https://github.com/dotnet/machinelearning/pull/4228#issuecomment-533181203) [](ancestors = 533181203)
> 
> Refers to: src/Microsoft.ML.Dnn/ImageClassificationTransform.cs:667 in [8d07401](https://github.com/dotnet/machinelearning/commit/8d07401eeb8726a3fa975da661acec55b6cbddf1). [](commit_id = [8d07401](https://github.com/dotnet/machinelearning/commit/8d07401eeb8726a3fa975da661acec55b6cbddf1), deletion_comment = False)

If I'm not mistaken, _labelColumnName needs to be serialized here, as GetModelInfo(env, ctx, ...) in ImageClassificationTransform.cs outputs _labelColumn_, which is one of the variables given to the constructor of ImageClassificationTransformer during de-serialization. 

Also, in the constructor of ImageClassificationTransformer, all code that accesses the training label column itself exists under if statements that check whether we just loaded from an already-trained model or not, and the training label column is accessed only when we aren't loading from a model. As a result, there won't be an instance where we are trying to access a non-existent training label column. 

Lastly, I think that the potential removal of _labelColumnName is outside the scope of this PR, and should be done under a different issue/pull request.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -753,6 +811,12 @@ private protected override void SaveModel(ModelSaveContext ctx)
             ctx.Writer.Write(_predictedLabelColumnName);
             ctx.Writer.Write(_learningRate);
             ctx.Writer.Write(_classCount);
+
+            Host.AssertNonEmpty(_keyValueAnnotations);
```
yaeldekel: >AssertNonEmpty [](start = 17, length = 14)

You cannot assert this, since it is possible that `_keyValueAnnotations` is null: You are initializing it above in line 750 only if `inputSchema[labelColumnName].HasKeyValues()`. So you need to serialize a byte indicating whether it is null or not, followed by `_classCount` strings if it is not null. And same at deserialization time (read a byte, and then read `_classCount` strings only if that byte is `true`).

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
yaeldekel: The field `_outputTypes` can be deleted. The only place where it is used is in line 1281 where you have `_outputTypes[0].GetItemType()`. Instead of that you can write `NumberDataViewType.Single`, and then the field is not needed.




---
In reply to: [326229209](https://github.com/dotnet/machinelearning/pull/4228#discussion_r326229209) [](ancestors = 326229209)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
yaeldekel: Reactivating, since line 1284 still instantiates the new `SchemaShape.Column` incorrectly in iteration 16.

---
In reply to: [329082925](https://github.com/dotnet/machinelearning/pull/4228#discussion_r329082925) [](ancestors = 329082925,327485460)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
mstfbl: I missed this, thank you for the catch!

mstfbl:
All builds pass, ready for commit!

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1281,6 +1293,49 @@ public void TensorFlowImageClassification()
...
+            var labelTypeFirst = labelColumnFirst.Type;
+            var labelCountFirst = labelTypeFirst.GetKeyCount();
+            var labelColumnLast = schema.GetColumnOrNull("Label").Value;
+            var labelTypeLast = labelColumnLast.Type;
+            var labelCountLast = labelTypeLast.GetKeyCount();
+
+            Assert.Equal((int)labelCountFirst, predictionFirst.Score.Length);
+            Assert.Equal((int)labelCountLast, predictionLast.Score.Length);
+            Assert.True(Array.IndexOf(labels, predictionFirst.PredictedLabel) > -1);
+            Assert.True(Array.IndexOf(labels, predictionLast.PredictedLabel) > -1);
```
codemzs: Can you please add an assert that predictionFirst and predictionLast have the same value? and also put an assert on the value? (it should remain the same since your seed is fixed)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -731,13 +751,51 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Dat
...
+                if (inputSchema[labelColumnName].HasKeyValues())
+                {
+                    inputSchema[labelColumnName].GetKeyValues(ref keysVBuffer);
+                    _keyValueAnnotations = keysVBuffer.DenseValues().ToArray();
+                }
+            }
+            else
+            {
+                // Load annotations as key values, if they exist
+                if (!(labelAnnotations is null))
```
codemzs: >if (!(labelAnnotations is null)) [](start = 16, length = 32)

if(labelAnnotations != null)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -760,6 +818,12 @@ private protected override void SaveModel(ModelSaveContext ctx)
             ctx.Writer.Write(_predictedLabelColumnName);
             ctx.Writer.Write(_learningRate);
             ctx.Writer.Write(_classCount);
+
+            Host.AssertNonEmpty(_keyValueAnnotations);
+            Host.Assert(_keyValueAnnotations.Length == _classCount);
+            for (int j = 0; j < _classCount; j++)
+                ctx.SaveNonEmptyString(_keyValueAnnotations[j]);
```
codemzs: Is it really guaranteed _keyValueAnnotations is not empty? refer to below lines of code that you have written in the constructor.

VBuffer<ReadOnlyMemory<char>> keysVBuffer = default;
                if (inputSchema[labelColumnName].HasKeyValues())
                {
                    inputSchema[labelColumnName].GetKeyValues(ref keysVBuffer);
                    _keyValueAnnotations = keysVBuffer.DenseValues().ToArray();
                }

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -731,13 +751,51 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Dat
...
                 _softmaxTensorName = _softMaxTensor.name;
                 _predictionTensorName = _prediction.name;
+
+                // Add annotations as key values, if they exist.
+                VBuffer<ReadOnlyMemory<char>> keysVBuffer = default;
+                if (inputSchema[labelColumnName].HasKeyValues())
+                {
+                    inputSchema[labelColumnName].GetKeyValues(ref keysVBuffer);
+                    _keyValueAnnotations = keysVBuffer.DenseValues().ToArray();
+                }
```
codemzs: If it does not have key values then just add numeric values (0 - classCount -1) as string to have something in there

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
mstfbl: Adjusted code with latest push so that _keyValueAnnotations will not be null, thanks to Zeeshan for the suggestion

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
mstfbl: With your suggested changes below, it is now guaranteed that _keyValueAnnotations will not be empty.

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
codemzs: I'm sorry I meant can you please assert on the value for predictionFirst.PredictedLabel and predictionLast.PredictedLabel ... instead of just checking its greater than -1

---
In reply to: [330779542](https://github.com/dotnet/machinelearning/pull/4228#discussion_r330779542) [](ancestors = 330779542)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -731,13 +751,55 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Dat
...
+                else
+                {
+                    _keyValueAnnotations = Enumerable.Range(0, _classCount).ToArray().Select(x => x.ToString().AsMemory()).ToArray();
+                }
```
codemzs: Enumerable.Range(0, _classCount).Select(x => x.ToString().AsMemory()).ToArray();

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1281,6 +1293,54 @@ public void TensorFlowImageClassification()
...
+            var labelTypeLast = labelColumnLast.Type;
+            var labelCountLast = labelTypeLast.GetKeyCount();
+
+            Assert.Equal((int)labelCountFirst, predictionFirst.Score.Length);
+            Assert.Equal((int)labelCountLast, predictionLast.Score.Length);
+            Assert.Equal(predictionFirst.PredictedLabel, predictionLast.PredictedLabel);
+            for (int j = 0; j < predictionFirst.Score.Length; j++)
+            {
+                predictionFirst.Score[j] = predictionLast.Score[j];
+            }
```
codemzs: I'm sorry what is this for?

mstfbl: My apologies, it's a mistake.

codemzs:
@mstfbl How come the tests are now failing?

codemzs:
It seems your tests are failing because of the below error:

tarting test: Microsoft.ML.Scenarios.TensorFlowScenariosTests.TensorFlowSentimentClassificationTest
[xUnit.net 00:09:54.21]     Microsoft.ML.Scenarios.TensorFlowScenariosTests.TensorFlowImageClassification [FAIL]
  X Microsoft.ML.Scenarios.TensorFlowScenariosTests.TensorFlowImageClassification [1m 14s]
  Error Message:
   Assert.Equal() Failure
          ↓ (pos 0)
Expected: dandelion
Actual:   sunflowers
          ↑ (pos 0)


Can you please make the tests deterministic by always passing the same image?

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1281,6 +1293,56 @@ public void TensorFlowImageClassification()
...
+            IEnumerable<ImageData> testImages = LoadImagesFromDirectory(
+                fullImagesetFolderPath, true);
+
+            string[] directories = Directory.GetDirectories(fullImagesetFolderPath);
+            string[] labels = new string[directories.Length];
+            for(int j = 0; j < labels.Length; j++)
+            {
+                var dir = new DirectoryInfo(directories[j]);
+                labels[j] = dir.Name;
+            }
```
codemzs: this is problematic, you cannot assume the ValueToKey transform will see the directories as labels in this order. Instead get keys from the IDataview like we were doing in the sample before 

VBuffer<ReadOnlyMemory<char>> keys = default;
                loadedModel.GetOutputSchema(schema)["Label"].GetKeyValues(ref keys);

mstfbl:
I just manually tested for predicting two known images for their labels using the master, unmerged code. I tested predicting an image of a daisy and an image of a rose. The predictions here are correct, which means the error is in my additions pertaining to my annotations, and not in the trained model. I have attached a screenshot of this below.

ah

@yaeldekel @eerhardt We've determined that the trained model is performing correctly without the current additions of categorical values to PredictedLabel. Something is going on with my annotation additions, and I can't find the bug. Would you guys be able to help in diagnosing where the bug is in my additions with fresh new eyes?

codemzs:
@mstfbl Your bug is in predictedLabel value in output cache. In KeyValue type Keys start from 1 but in Tensorflow predictedLabels start from 0. In file: ImageClassificationTransform.cs after Line 915 outputTensor[1].ToScalar<long>(ref _predictedLabel); add _predictedLabel += 1;

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs
```diff
@@ -63,14 +63,19 @@ public static void Example()
...
+                    // Uncomment reuseTrainSetBottleneckCachedValues and
+                    // reuseValidationSetBottleneckCachedValues to reuse trained model
+                    // for faster debugging.
+                    arch: ImageClassificationEstimator.Architecture.ResnetV2101,
                     epoch: 50,
                     batchSize: 10,
                     learningRate: 0.01f,
                     metricsCallback: (metrics) => Console.WriteLine(metrics),
+                    // reuseTrainSetBottleneckCachedValues: true,
+                    // reuseValidationSetBottleneckCachedValues: true,
```
codemzs: Please remove this

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs
```diff
@@ -63,14 +63,19 @@ public static void Example()
                     // Just by changing/selecting InceptionV3 here instead of 
                     // ResnetV2101 you can try a different architecture/pre-trained 
                     // model. 
-                    arch: ImageClassificationEstimator.Architecture.ResnetV2101, 
+                    // Uncomment reuseTrainSetBottleneckCachedValues and
+                    // reuseValidationSetBottleneckCachedValues to reuse trained model
+                    // for faster debugging.
```
codemzs: Please remove this

PR #4232, opened 2019-09-20
Add entrypoint for PFI

### LINKED ISSUE #4231: Add Entry Point for Permutation Feature Importance
Needed to add PFI to NimbusML.

microsoft/NimbusML#92

najeeb-kazmi:
Fixes #4231 

Entrypoint is needed to add PFI to NimbusML
microsoft/NimbusML#92


src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
```diff
@@ -0,0 +1,342 @@
+﻿using System;
```
eerhardt: copyright

src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
```diff
@@ -0,0 +1,342 @@
...
+        public double NegativePrecision { get; set; }
+
+        public double NegativeRecall { get; set; }
+
+        public double F1Score { get; set; }
+
+        public double AreaUnderPrecisionRecallCurve { get; set; }
+    }
+
+    internal class MulticlassMetrics
```
eerhardt: Can we not use the metrics classes that are already in the product?

test/BaselineOutput/Common/EntryPoints/core_manifest.json
```diff
@@ -21710,6 +21710,103 @@
         "ITransformOutput"
       ]
     },
+    {
```
eerhardt: Can we add a test actually invoking the entry points for each supported ml task?

src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
najeeb-kazmi: These classes are actually being used as the schema of the `IDataView` that I want to return. The metrics returned by PFI are of type `MetricStatistics`, on which I cannot directly create an `IDataView` as `MetricStatistics` is not a supported type in `IDataView`. I want to extract the means and standard error of means for each metric i.e. AUC, RSquared, NDCG, etc.

Also, I cannot use `BinaryClassificationMetrics` or the others, as the fields cannot be set in that class.

src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
```diff
@@ -0,0 +1,411 @@
...
+                                              data,
+                                              labelColumnName: input.LabelColumnName,
+                                              useFeatureWeightFilter: input.UseFeatureWeightFilter,
+                                              numberOfExamplesToUse: input.NumberOfExamplesToUse,
+                                              permutationCount: input.PermutationCount);
+
+            Contracts.Assert(slotNames.Length == permutationMetrics.Length,
+                "Mismatch between number of feature slots and number of features permuted.");
+
+            IEnumerable<MulticlassMetrics> metrics = Enumerable.Empty<MulticlassMetrics>();
```
eerhardt: You should use a `List<MulticlassMetrics>` instead. Appending to Enumerable.Empty is quite inefficient.

src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
```diff
@@ -0,0 +1,411 @@
...
+            Contracts.CheckValue(env, nameof(env));
+            var host = env.Register("Pfi");
+            host.CheckValue(input, nameof(input));
+            EntryPointUtils.CheckInputArgs(host, input);
+
+            var mlContext = new MLContext();
+
+            var model = mlContext.Model.Load(input.ModelPath.OpenReadStream(), out DataViewSchema schema);
+            var chain = model as TransformerChain<ITransformer>;
+            var predictor = chain.LastTransformer as ISingleFeaturePredictionTransformer<object>;
```
eerhardt: Is assuming the `LastTransformer` is the predictor always correct? What about the cases where someone adds something afterwards - like a KeyToValueMapping?

src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
```diff
@@ -0,0 +1,411 @@
...
+    {
+        [TlcModule.EntryPoint(Name = "Transforms.PermutationFeatureImportance", Desc = "Permutation Feature Importance (PFI)", UserName = "PFI", ShortName = "PFI")]
+        public static PermutationFeatureImportanceOutput PermutationFeatureImportance(IHostEnvironment env, PermutationFeatureImportanceArguments input)
+        {
+            Contracts.CheckValue(env, nameof(env));
+            var host = env.Register("Pfi");
+            host.CheckValue(input, nameof(input));
+            EntryPointUtils.CheckInputArgs(host, input);
+
+            var mlContext = new MLContext();
```
eerhardt: You shouldn't be creating a new MLContext here. You should use the environment being passed into the method.

src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
eerhardt: Maybe in this case then, let's make these as `private` nested classes. That should signify that they are only to be used inside of this class.

src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
najeeb-kazmi: `IHostEnvironment` doesn't have model loading that I need here. Casting it to `MLContext` in the method body returns null due to inheritance direction. Also, changing the type in the method signature from `IHostEnvironment` to `MLContext` doesn't work as the first parameter is required to be `IHostEnvironment` here:
https://github.com/dotnet/machinelearning/blob/edfd10f131abe97a7f7bb25c0070843a67b8b660/src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs#L310-L311

What do you suggest here? Change the component catalog code above to accept both `IHostEnvironment` and `MLContext`, and change the type in the method signature ?

src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
najeeb-kazmi: PFI expects transformed data and a predictor. Right now, PFI is not supported for the case where there is something added after a predictor. Reason being that PFI needs the data that the predictor trained on, so transforming the data with something else that came after the predictor is not correct.

If we had the ability to extract part of the `TransformerChain` up to the predictor, we could apply the transforms up to that point and extract the predictor to pass to PFI. But this raises the question: what happens if there are multiple predictors in the `TransformerChain`, e.g. a regressor followed by a classifier training on the output of the regressor. This can get messy really quickly, so the suggested usage is to have the predictor one wants to compute PFI for at the end of the `TransformerChain`, then add whatever else one wants after PFI has been calculated.

I'll change the error message to clarify this.

justinormont: Isn't `KeyToValueMapping` rather needed for most multi-class problems, as the user needs to know which input class the term transform mapped to class 0/1/2/3/4/..? 

Perhaps you could walk the transformer chain backwards from the end until you find a trainer?

src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
eerhardt: > What do you suggest here? 

I think one of the following approaches would work just fine:

1. Refactor `ModelOperationsCatalog.Load` to be an internal static method that takes in the `IHostEnvironment` along with the `Stream`. Call it here and in the original `ModelOperationsCatalog.Load` method.
2. Create a `new ModelOperationsCatalog` instance with the `IHostEnvironment` you have, and call `Load` using it. No need to create a new `MLContext`.

We can't be creating new environments in the middle of a library function. We need to use the environment coming from the caller.

src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
ganik: this wont happen in NimbusML case, the pipeline has to end with the Predictor. Mapping from predicted key to actual values is handled without adding the KeyToValueMapping transformer in NimbusML code

---
In reply to: [326823195](https://github.com/dotnet/machinelearning/pull/4232#discussion_r326823195) [](ancestors = 326823195)

ganik: Nope its not needed. Mapping from predicted key to actual values is handled without adding the KeyToValueMapping transformer in NimbusML code

---
In reply to: [326848473](https://github.com/dotnet/machinelearning/pull/4232#discussion_r326848473) [](ancestors = 326848473)

ganik: Why do you need to split up predictor and transformers ? Can you calculate PFI on whole chain transformer + predictor. User can decide whether to supply you just a Predictor model with featurized data or whole Trasnformer + Predictor chain with raw features.

---
In reply to: [326842514](https://github.com/dotnet/machinelearning/pull/4232#discussion_r326842514) [](ancestors = 326842514)

eerhardt: I was under the impression that EntryPoints were going to be used in more places than just Python. But I could have misunderstood.

ganik: ideally it should be general enough, still python is the only use case currently and in foreseeable future. Anyway in this case we should not split the model chain to Transformer and Predictor but use the whole chain as a Predictor

---
In reply to: [327301376](https://github.com/dotnet/machinelearning/pull/4232#discussion_r327301376) [](ancestors = 327301376)

najeeb-kazmi: @eerhardt @justinormont The last commit adds a method in `TransformerChain` that returns the `TransformerChain` up to the last predictor, and throws if there is no predictor in the `TransformerChain` at all.

src/Microsoft.ML.Data/DataLoadSave/TransformerChain.cs
```diff
@@ -128,6 +128,25 @@ public DataViewSchema GetOutputSchema(DataViewSchema inputSchema)
...
+            {
+                var current = _transformers[i - 1];
+                if (current is IPredictionTransformer<IPredictorProducing<float>>
+                    || current is IPredictionTransformer<IPredictorProducing<VBuffer<float>>>)
+                {
+                    lastPredictorIndex = i;
+                    break;
+                }
+            }
+            Contracts.Assert(lastPredictorIndex != 0, "No predictor found in the model.");
```
yaeldekel: >Assert [](start = 22, length = 6)

If your intention is to throw if there is no predictor, then this needs to be a `Check` instead of `Assert`. Otherwise, callers of this method need to make sure there is a predictor in the chain before calling it.

src/Microsoft.ML.Data/DataLoadSave/TransformerChain.cs
```diff
@@ -128,6 +128,25 @@ public DataViewSchema GetOutputSchema(DataViewSchema inputSchema)
             return s;
         }
 
+        [BestFriend]
+        internal TransformerChain<ITransformer> RewindToLastPredictionTransformer()
+        {
+            int lastPredictorIndex = 0;
```
yaeldekel: >lastPredictorIndex [](start = 16, length = 18)

nit: technically, this isn't the index of the last predictor, it's the index+1... Might be less confusing to rename.

src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
```diff
@@ -0,0 +1,420 @@
...
+    internal sealed class PermutationFeatureImportanceOutput
+    {
+        [TlcModule.Output(Desc = "The PFI metrics")]
+        public IDataView Metrics;
+    }
+
+    internal sealed class PermutationFeatureImportanceArguments : TransformInputBase
+    {
+        [Argument(ArgumentType.Required, HelpText = "The path to the model file", ShortName = "path", Visibility = ArgumentAttribute.VisibilityType.EntryPointsOnly)]
+        public IFileHandle ModelPath;
```
yaeldekel: >IFileHandle [](start = 15, length = 11)

In most of the entrypoints, models are passed in and out as objects of type `PredictorModel`. I think it would make more sense to have a `PredictorModel` input here as well, that well the entrypoint can be added to an entrypoint graph.

najeeb-kazmi: I did try this. The PFI implementation here ([https://github.com/dotnet/machinelearning/blob/edfd10f131abe97a7f7bb25c0070843a67b8b660/src/Microsoft.ML.Transforms/PermutationFeatureImportance.cs#L20-L31](https://github.com/dotnet/machinelearning/blob/edfd10f131abe97a7f7bb25c0070843a67b8b660/src/Microsoft.ML.Transforms/PermutationFeatureImportance.cs#L20-L31)) expects an `IPredictionTransformer`, while `PredictorModel` has an `IPredictor` that cannot be cast to `IPredictionTransformer`.

---
In reply to: [328094891](https://github.com/dotnet/machinelearning/pull/4232#discussion_r328094891) [](ancestors = 328094891)

yaeldekel: `PredictorModel` has a method called `PrepareData` which gives back the `IPredictor` and a `RoleMappedData` (see an example of how it's used in SummarizePredictor.cs line 41). Then you can use the `IPredictor` and the `RoleMappedSchema` to create the correct `ISingleFeaturePredictionTransformer` as is done in the method `LoadFromLegacy` in TransformerChain.cs line 253.
Having the input be an `IFileHandle` requires two separate entrypoint graphs (like you have in your test), and requires serialization and deserialization of the model.

---
In reply to: [328344099](https://github.com/dotnet/machinelearning/pull/4232#discussion_r328344099) [](ancestors = 328344099,328094891)

najeeb-kazmi: I see what you mean by adding it to the graph. Thanks for these pointers - I'll look into the code and do another iteration.

---
In reply to: [328499170](https://github.com/dotnet/machinelearning/pull/4232#discussion_r328499170) [](ancestors = 328499170,328344099,328094891)

src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
najeeb-kazmi: Changed code to load model as `PredictorModel` instead of `IFileHandle`, like most entrypoints. The transforms are stored in `PredistorModel.TransformModel` and the predictor in `PredictorModel.Predictor`, so this is now irrelevant. 

Marking this as resolved.

src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
```diff
@@ -0,0 +1,435 @@
...
+            Contracts.CheckValue(env, nameof(env));
+            var host = env.Register("Pfi");
+            host.CheckValue(input, nameof(input));
+            EntryPointUtils.CheckInputArgs(host, input);
+
+            input.PredictorModel.PrepareData(env, input.Data, out RoleMappedData roleMappedData, out IPredictor predictor);
+            Contracts.Assert(predictor != null, "No predictor found in model");
+            var transformedData = input.PredictorModel.TransformModel.Apply(env, input.Data);
+            IDataView result = PermutationFeatureImportanceUtils.GetMetrics(env, predictor, roleMappedData.Schema, transformedData, input);
+            return new PermutationFeatureImportanceOutput { Metrics = result };
```
ganik: Awesome change!

test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -612,11 +612,746 @@ public void EntryPointExecGraphCommand()
...
+
+            var args = new ExecuteGraphCommand.Arguments() { GraphPath = jsonPath };
+            var cmd = new ExecuteGraphCommand(Env, args);
+            cmd.Run();
+
+            var mlContext = new MLContext();
+            var loadedData = mlContext.Data.LoadFromBinary(outputDataPath);
+            Assert.NotNull(loadedData.Schema.GetColumnOrNull("FeatureName"));
+            Assert.NotNull(loadedData.Schema.GetColumnOrNull("AreaUnderRocCurve"));
+            Assert.NotNull(loadedData.Schema.GetColumnOrNull("Accuracy"));
```
eerhardt: Do we want to verify that the numbers that come back here are correct?

najeeb-kazmi: We would verify this by loading the same predictor model and data that goes into the PFI node, and applying the same steps to them that the PFI node does, so I don't see the point of doing this.

src/Microsoft.ML.EntryPoints/PermutationFeatureImportance.cs
```diff
@@ -0,0 +1,429 @@
...
+        }
+
+        private static IDataView GetBinaryMetrics(
+            IHostEnvironment env,
+            IPredictor predictor,
+            RoleMappedData roleMappedData,
+            PermutationFeatureImportanceArguments input)
+        {
+            var roles = roleMappedData.Schema.GetColumnRoleNames();
+            var featureColumnName = roles.Where(x => x.Key.Value == RoleMappedSchema.ColumnRole.Feature.Value).First().Value;
```
yaeldekel: >featureColumnName [](start = 16, length = 17)

All these special roles - features, label, group etc. have properties in `RoleMappedSchema`, but you need to check first that they are not null. (here and in the other `Get***Metrics` methods as well).

PR #4237, opened 2019-09-20
Adding Early stopping feature in ImageClassification (WIP)

### LINKED ISSUE #4236: Need Early Stopping feature in Image Classification
### Issue
Need the Early stopping feature in ML .NET so that number of epochs do not need to be set/predefined for Image Classification training.
Said feature available as a Keras callback: https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/EarlyStopping

ashbhandare:
Fixes #4236

Modeled after https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/EarlyStopping


src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -338,7 +338,13 @@ private IDataView GetShuffledData(string path)
...
             ImageClassificationMetrics metrics = new ImageClassificationMetrics();
             metrics.Train = new TrainMetrics();
-            for (int epoch = 0; epoch < epochs; epoch += 1)
+            //Early stopping variables
+            bool earlyStop = false;
+            int wait = 0;
+            var history = new TrainMetrics();
+            history.Accuracy = 0;
+
+            for (int epoch = 0; epoch < epochs & !earlyStop; epoch += 1)
```
eerhardt: `&` => `&&`

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -113,6 +116,9 @@ public static class DnnCatalog
             int epoch = 100,
             int batchSize = 10,
             float learningRate = 0.01f,
+            bool enableEarlyStopping = true,
+            float earlyStoppingminDelta = 0.01f,
```
eerhardt: `earlyStoppingminDelta` - I assume `Min` is a new word and should be capitalized.

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -113,6 +116,9 @@ public static class DnnCatalog
             int epoch = 100,
             int batchSize = 10,
             float learningRate = 0.01f,
+            bool enableEarlyStopping = true,
```
eerhardt: Do we really need a `enableEarlyStopping` parameter? What if we instead used nullable `earlyStoppingminDelta` and `earlyStoppingPatience` parameters, whose default value is `null`. If the user doesn't supply those values, then early stopping isn't enabled.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -445,6 +451,24 @@ private IDataView GetShuffledData(string path)
...
+                    {
+                        history.Accuracy = metrics.Train.Accuracy;
+                        wait = 0;
+                    }
+                    else
+                    {
+                        wait += 1;
+                        if (wait >= options.Patience)
+                        {
+                            Console.WriteLine("*** Early Stopping at epoch " + epoch.ToString());
```
eerhardt: Please don't `Console.Write` inside of library code.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -1065,6 +1089,24 @@ internal sealed class Options : TransformInputBase
             [Argument(ArgumentType.AtMostOnce, HelpText = "Learning rate to use during optimization.", SortOrder = 12)]
             public float LearningRate = 0.01f;
 
+            /// <summary>
+            /// Whether early stopping technique should be used when accuracy stops improving.
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Whether early stopping technique should be used when accuracy stops improving.", SortOrder = 15)]
+            public bool EnableEarlyStopping = true;
```
eerhardt: Why is the default for this and `MinDelta` different than what it is in the public API? That is confusing.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -338,7 +338,13 @@ private IDataView GetShuffledData(string path)
 
             ImageClassificationMetrics metrics = new ImageClassificationMetrics();
             metrics.Train = new TrainMetrics();
-            for (int epoch = 0; epoch < epochs; epoch += 1)
+            //Early stopping variables
+            bool earlyStop = false;
+            int wait = 0;
+            var history = new TrainMetrics();
+            history.Accuracy = 0;
+
```
bpstark: why are you using a TrainMetrics object when you only care about the history of the accuracy. you can simply use a float here instead.

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ResnetV2101TransferLearningTrainTestSplit.cs
```diff
@@ -67,6 +67,7 @@ public static void Example()
                     epoch: 50,
                     batchSize: 10,
                     learningRate: 0.01f,
+                    enableEarlyStopping: true,
```
codemzs: I will prefer you create a new sample for early stopping.

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -89,6 +89,9 @@ public static class DnnCatalog
         /// <param name="epoch">Number of training iterations. Each iteration/epoch refers to one pass over the dataset.</param>
         /// <param name="batchSize">The batch size for training.</param>
         /// <param name="learningRate">The learning rate for training.</param>
+        /// <param name="enableEarlyStopping">Whether early stopping technique should be used when accuracy stops improving.</param>
+        /// <param name="earlyStoppingminDelta">Minimum change in accuracy to qualify as improvement.</param>
+        /// <param name="earlyStoppingPatience">Number of epochs to wait after no improvement is observed before early stopping.</param>
```
codemzs: Make this option a class.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -445,6 +451,24 @@ private IDataView GetShuffledData(string path)
                         statisticsCallback(metrics);
                     }
                 }
+                // early stopping check
```
codemzs: >// early stopping check [](start = 16, length = 23)

New line and capital E

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -445,6 +451,24 @@ private IDataView GetShuffledData(string path)
...
+                    else
+                    {
+                        wait += 1;
+                        if (wait >= options.Patience)
+                        {
+                            Console.WriteLine("*** Early Stopping at epoch " + epoch.ToString());
+                            earlyStop = true;
+                        }
+                    }
+                }
```
codemzs: Please add comments here that document this technique and also add relevant links

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
codemzs: >Console.WriteLine("*** Early Stopping at epoch " + epoch.ToString()); [](start = 28, length = 69)

Use message channels for logging.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -338,7 +338,13 @@ private IDataView GetShuffledData(string path)
 
             ImageClassificationMetrics metrics = new ImageClassificationMetrics();
             metrics.Train = new TrainMetrics();
-            for (int epoch = 0; epoch < epochs; epoch += 1)
+            //Early stopping variables
+            bool earlyStop = false;
+            int wait = 0;
+            var history = new TrainMetrics();
+            history.Accuracy = 0;
```
codemzs: Why create the a new class when you just want a variable to store accuracy? just create a float variable "lastSeenAccuracy"

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -445,6 +451,24 @@ private IDataView GetShuffledData(string path)
...
+                    }
+                    else
+                    {
+                        wait += 1;
+                        if (wait >= options.Patience)
+                        {
+                            Console.WriteLine("*** Early Stopping at epoch " + epoch.ToString());
+                            earlyStop = true;
+                        }
+                    }
```
codemzs: Please do not put a PR without a unit-test, if you do, please mark it as Draft PR or WIP.

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -136,6 +142,9 @@ public static class DnnCatalog
                 Epoch = epoch,
                 LearningRate = learningRate,
                 BatchSize = batchSize,
+                EnableEarlyStopping = enableEarlyStopping,
+                MinDelta = earlyStoppingminDelta,
+                Patience = earlyStoppingPatience,
```
codemzs: Since you are taking this technique from [https://github.com/tensorflow/tensorflow/blob/r1.14/tensorflow/python/keras/callbacks.py#L1143-L1260](https://github.com/tensorflow/tensorflow/blob/r1.14/tensorflow/python/keras/callbacks.py#L1143-L1260) 

were you also planning to add the "modes", i.e min, max, auto?

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -445,6 +451,24 @@ private IDataView GetShuffledData(string path)
...
                     }
                 }
+                // early stopping check
+                if (options.EnableEarlyStopping)
+                {
+                    if (metrics.Train.Accuracy - options.MinDelta > history.Accuracy)
+                    {
+                        history.Accuracy = metrics.Train.Accuracy;
+                        wait = 0;
+                    }
```
codemzs: This is incorrect. It needs to be absolute change as documented here: [https://github.com/tensorflow/tensorflow/blob/r1.14/tensorflow/python/keras/callbacks.py#L1143-L1260](https://github.com/tensorflow/tensorflow/blob/r1.14/tensorflow/python/keras/callbacks.py#L1143-L1260)

min_delta: Minimum change in the monitored quantity
          to qualify as an improvement, i.e. an absolute
          change of less than min_delta, will count as no
          improvement.

Looking at the code:

 if mode == 'min':
      self.monitor_op = np.less
    elif mode == 'max':
      self.monitor_op = np.greater
    else:
      if 'acc' in self.monitor:
        self.monitor_op = np.greater
      else:
        self.monitor_op = np.less

    if self.monitor_op == np.greater:
      self.min_delta *= 1
    else:
      self.min_delta *= -1

The last 4 lines change the sign of the delta and that takes care of absolute difference in below function:

def on_epoch_end(self, epoch, logs=None):
    current = self.get_monitor_value(logs)
    if current is None:
      return
    if self.monitor_op(current - self.min_delta, self.best):
      self.best = current
      self.wait = 0
      if self.restore_best_weights:
        self.best_weights = self.model.get_weights()
    else:
      self.wait += 1
      if self.wait >= self.patience:
        self.stopped_epoch = epoch
        self.model.stop_training = True
        if self.restore_best_weights:
          if self.verbose > 0:
            print('Restoring model weights from the end of the best epoch.')
          self.model.set_weights(self.best_weights)

src/Microsoft.ML.Dnn/DnnCatalog.cs
ashbhandare: According to @codemzs , we do not want the user to have to set any of these parameters, and want the default values to work well in most of the cases. If the default is set to null, the users would have to appropirately set these values to make use of this feature.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: This was with a potential of possibly using other train metrics as well for the criteria for early stopping. I will refactor the code and this will change.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: addresssed here: https://github.com/dotnet/machinelearning/pull/4237#discussion_r328833518

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: As we discussed offline, the change in sign of the min_delta doesn't handle taking the absolute value. Even though it is mentioned in the comment that they take absolute value of the change, it is not implemented that way in the code. However, we want to consider absolute value and I will make that change.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: Similar behaviour exists for the Epoch option. Basically, the MinDelta can be 0.0, but a value of 0.01 works well in most cases, and hence the ImageClassificationEstimator is initialized with 0.01f but the option itself is set to 0. I can make both values same(0.01f) if it is indeed confusing to the users.

src/Microsoft.ML.Dnn/DnnCatalog.cs
ashbhandare: According to @Zeeshan Siddiqui , we do not want the user to have to set any of these parameters, and want the default values to work well in most of the cases. If the default is set to null, the users would have to appropirately set these values to make use of this feature.  

---
In reply to: [327289504](https://github.com/dotnet/machinelearning/pull/4237#discussion_r327289504) [](ancestors = 327289504)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: Similar behaviour exists for the Epoch option. Basically, the MinDelta can be 0.0, but a value of 0.01 works well in most cases, and hence the ImageClassificationEstimator is initialized with 0.01f but the option itself is set to 0. I can make both values same(0.01f) if it is indeed confusing to the users.

---
In reply to: [327290199](https://github.com/dotnet/machinelearning/pull/4237#discussion_r327290199) [](ancestors = 327290199)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: As we discussed offline, the change in sign of the min_delta doesn't handle taking the absolute value. Even though it is mentioned in the comment that they take absolute value of the change, it is not implemented that way in the code. However, we want to consider absolute value and I will make that change.

---
In reply to: [328248944](https://github.com/dotnet/machinelearning/pull/4237#discussion_r328248944) [](ancestors = 328248944)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: addresssed here: [https://github.com/dotnet/machinelearning/pull/4237#discussion_r328833518](https://github.com/dotnet/machinelearning/pull/4237#discussion_r328833518)

---
In reply to: [328229850](https://github.com/dotnet/machinelearning/pull/4237#discussion_r328229850) [](ancestors = 328229850)

src/Microsoft.ML.Dnn/DnnCatalog.cs
eerhardt: I'm not convinced this is the best API to enable this. What happens if we want to enable a different stopping criteria in the future?

It feels like we should consider a different API to enable this. Check out the EarlyStoppingCriteria in FastTree. That seems like more of an extensible/future-proof API.

https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.FastTree/Training/EarlyStoppingCriteria.cs#L34-L38

codemzs: @eerhardt @ashbhandare We spoke about this on Friday and I suggested we have a class object that defines early stopping criteria, this class should extend an interface that defines bool ShouldStop(...). The API parameter should be a reference to this interface EarlyStoping and if it is set to null then we don't apply early stopping but by default it can be set to XYXEarlyStoping ...

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -445,6 +447,12 @@ private IDataView GetShuffledData(string path)
                         statisticsCallback(metrics);
                     }
                 }
+
+                //Early stopping check
+                if (options.EarlyStopper != null)
+                {
+                    earlyStop = options.EarlyStopper.ShouldStop(metrics.Train);
+                }
```
codemzs: why not just break out? it will save you a variable

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -981,6 +998,108 @@ public sealed class BottleneckMetrics
...
+        public sealed class EarlyStopping
+        {
+            /// <summary>
+            /// Best value of metric seen so far.
+            /// </summary>
+            private float _bestMetricValue;
+
+            /// <summary>
+            /// Current number of epochs where there has been no improvement.
+            /// Stop training when wait >=patience.
```
codemzs: >wait [](start = 35, length = 4)

please use param ref to refer variables.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -981,6 +998,108 @@ public sealed class BottleneckMetrics
...
+            /// Any change  in the value of metric of less than 'minDelta' is not considered a change.
+            /// </summary>
+            public bool ShouldStop(TrainMetrics currentMetrics)
+            {
+                float currentMetricValue;
+                if (_metric == EarlyStoppingMetric.Accuracy)
+                    currentMetricValue = currentMetrics.Accuracy;
+                else
+                    currentMetricValue = currentMetrics.CrossEntropy;
+                if(CheckIncreasing)
```
codemzs: >if(CheckIncreasing) [](start = 16, length = 19)

new line

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -981,6 +998,108 @@ public sealed class BottleneckMetrics
...
+            /// within 'patience' number of epochs.
+            /// Any change  in the value of metric of less than 'minDelta' is not considered a change.
+            /// </summary>
+            public bool ShouldStop(TrainMetrics currentMetrics)
+            {
+                float currentMetricValue;
+                if (_metric == EarlyStoppingMetric.Accuracy)
+                    currentMetricValue = currentMetrics.Accuracy;
+                else
+                    currentMetricValue = currentMetrics.CrossEntropy;
```
codemzs: currentMetricValue = _metric == EarlyStoppingMetric.Accuracy ? currentMetrics.Accuracy : currentMetrics.CrossEntropy

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -981,6 +998,108 @@ public sealed class BottleneckMetrics
...
+                if(CheckIncreasing)
+                {
+                    if((currentMetricValue- _bestMetricValue) < MinDelta)
+                    {
+                        _wait += 1;
+                        if(_wait >= Patience)
+                            return (true);
```
codemzs: >(true); [](start = 35, length = 7)

why the brackets? just "return true;"

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -981,6 +998,108 @@ public sealed class BottleneckMetrics
...
+                if (_metric == EarlyStoppingMetric.Accuracy)
+                    currentMetricValue = currentMetrics.Accuracy;
+                else
+                    currentMetricValue = currentMetrics.CrossEntropy;
+                if(CheckIncreasing)
+                {
+                    if((currentMetricValue- _bestMetricValue) < MinDelta)
+                    {
+                        _wait += 1;
+                        if(_wait >= Patience)
```
codemzs: >if(_wait >= Patience) [](start = 24, length = 21)

Can _wait ever be greater than Patience?

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -981,6 +998,108 @@ public sealed class BottleneckMetrics
...
+                        _bestMetricValue = currentMetricValue;
+                    }
+                }
+                else
+                {
+                    if ((_bestMetricValue - currentMetricValue) < MinDelta)
+                    {
+                        _wait += 1;
+                        if (_wait >= Patience)
+                            return (true);
```
codemzs: >return (true); [](start = 28, length = 14)

return true

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -981,6 +998,108 @@ public sealed class BottleneckMetrics
...
+                    }
+                    else
+                    {
+                        _wait = 0;
+                        _bestMetricValue = currentMetricValue;
+                    }
+                }
+                return (false);
```
codemzs: > return (false); [](start = 15, length = 16)

return false

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -1065,6 +1184,12 @@ internal sealed class Options : TransformInputBase
             [Argument(ArgumentType.AtMostOnce, HelpText = "Learning rate to use during optimization.", SortOrder = 12)]
             public float LearningRate = 0.01f;
 
+            /// <summary>
+            /// Early Stopping technique to stop training when accuracy stops improving.
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Early Stopping technique to stop training when accuracy stops improving.", SortOrder = 15)]
+            public EarlyStopping EarlyStopper;
```
codemzs: >EarlyStopper [](start = 33, length = 12)

We generally refer this as EarlyStoppingCriteria

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1281,6 +1282,168 @@ public void TensorFlowImageClassification()
...
+            IDataView testDataset = trainTestData.TestSet;
+
+            int lastEpoch = 0;
+            var pipeline = mlContext.Model.ImageClassification(
+                "ImagePath", "Label",
+                arch: ImageClassificationEstimator.Architecture.ResnetV2101,
+                epoch: 50,
+                batchSize: 5,
+                learningRate: 0.01f,
+                earlyStopping: new ImageClassificationEstimator.EarlyStopping(),
```
codemzs: >earlyStopping: new ImageClassificationEstimator.EarlyStopping(), [](start = 16, length = 64)

isn't this the case by default? How is this test different from the above test? May be in the above test disable early stopping so we get that case covered and here enable but also verify the epoch at which it stops via metrics callback

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1281,6 +1282,168 @@ public void TensorFlowImageClassification()
...
+                metricsCallback: (metrics) => { Console.WriteLine(metrics); lastEpoch = metrics.Train != null ? metrics.Train.Epoch : 0; },
```
codemzs: >(metrics) => { Console.WriteLine(metrics); lastEpoch = metrics.Train != null ? metrics.Train.Epoch : 0; }, [](start = 33, length = 106)

Printing isn't useful we need to add COntract.Assert on he lastEpoch

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1281,6 +1282,168 @@ public void TensorFlowImageClassification()
...
+            mlContext.Model.Save(trainedModel, shuffledFullImagesDataset.Schema,
+                "model.zip");
+
+            ITransformer loadedModel;
+            DataViewSchema schema;
+            using (var file = File.OpenRead("model.zip"))
+                loadedModel = mlContext.Model.Load(file, out schema);
+
+            IDataView predictions = trainedModel.Transform(testDataset);
+            var metrics = mlContext.MulticlassClassification.Evaluate(predictions);
```
codemzs: same feedback as the above test.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: since patience is an int, user might supply a negative value. in that case, it is better to check >= instead of ==.

---
In reply to: [330317534](https://github.com/dotnet/machinelearning/pull/4237#discussion_r330317534) [](ancestors = 330317534)

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
ashbhandare: this already exists in this test in line 1363

---
In reply to: [330318727](https://github.com/dotnet/machinelearning/pull/4237#discussion_r330318727) [](ancestors = 330318727)

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
ashbhandare: done

---
In reply to: [330318498](https://github.com/dotnet/machinelearning/pull/4237#discussion_r330318498) [](ancestors = 330318498)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: updated description to not use variables. Couldn't use <paramref>  as these variables are not parameters to this particular variable(i guess).

---
In reply to: [330316879](https://github.com/dotnet/machinelearning/pull/4237#discussion_r330316879) [](ancestors = 330316879)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: done

---
In reply to: [330316698](https://github.com/dotnet/machinelearning/pull/4237#discussion_r330316698) [](ancestors = 330316698)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: renamed

---
In reply to: [330318020](https://github.com/dotnet/machinelearning/pull/4237#discussion_r330318020) [](ancestors = 330318020)

PR #4246, opened 2019-09-25
AutoML Add Recommendation Task

LittleLittleCloud:
What's already be done in this PR
- [x] added `Recommendation` task and experiment in `AutoML`
- [x] added `MatrixFactorization` as `MatrixFactorizationExtension`
- [x] added a new Column Purpose (`LabelFeature`) and it's corresponding TransformerExtension (`LabelCategorical`) so that `AutoML` can construct the pre-process pipeline for `MatrixFactorizationExtension` correctly
- [x] added a new recommendation example (with rating only) in `AutoML.Example`, and you can play with that!

What's need to be done (Feel Free to CRUD)
- [ ] figuring out how to accelerate and properly presenting the training process. Seems that `MatrixFactorization` requires more time to train a round, and the algorithm for sweeping params requires to train many rounds to find out the best parameter. It's time costy and customers might not like that.
- [ ] Corresponding `CodeGen` part
- [x] Test case!
- [x] Better Naming and code style
- [x] Enable support for multiple feature trainers in `AutoML` (it requires some refactor works and shouldn't be done in this PR. But it's important)


src/Microsoft.ML.Recommender/MatrixFactorizationPredictor.cs
```diff
@@ -422,14 +422,18 @@ DataViewRow ISchemaBoundRowMapper.GetRow(DataViewRow input, IEnumerable<DataView
...
-    public sealed class MatrixFactorizationPredictionTransformer : PredictionTransformerBase<MatrixFactorizationModelParameters>
+    public sealed class MatrixFactorizationPredictionTransformer : PredictionTransformerBase<MatrixFactorizationModelParameters>, ISingleFeaturePredictionTransformer<MatrixFactorizationModelParameters>
     {
         internal const string LoaderSignature = "MaFactPredXf";
         internal string MatrixColumnIndexColumnName { get; }
         internal string MatrixRowIndexColumnName { get; }
         internal DataViewType MatrixColumnIndexColumnType { get; }
         internal DataViewType MatrixRowIndexColumnType { get; }
 
+        public string FeatureColumnName => throw new NotImplementedException();
```
eerhardt: Are you really a `ISingleFeaturePredictionTransformer` if you just throw for all the members of `ISingleFeaturePredictionTransformer` ?

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
```diff
@@ -137,7 +137,7 @@ public enum LossFunctionType
         /// <summary>
         /// Options for the <see cref="MatrixFactorizationTrainer"/> as used in [MatrixFactorization(Options)](xref:Microsoft.ML.RecommendationCatalog.RecommendationTrainers.MatrixFactorization(Microsoft.ML.Trainers.MatrixFactorizationTrainer.Options)).
         /// </summary>
-        public sealed class Options
+        public sealed class Options : TrainerInputBaseWithLabel
```
eerhardt: This now adds a `FeatureColumnName` property to this class. I don't it gets used, correct? That may be confusing to users.

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
```diff
@@ -151,12 +151,6 @@ public sealed class Options
             /// </summary>
             public string MatrixRowIndexColumnName;
 
-            /// <summary>
-            /// The name variable (i.e., column in a <see cref="IDataView"/> type system) used as matrix's element value.
-            /// The column data must be <see cref="Microsoft.ML.Data.KeyDataViewType"/>.
-            /// </summary>
-            public string LabelColumnName;
```
eerhardt: I _think_ moving this is a binary breaking change.

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
```diff
@@ -331,9 +325,8 @@ internal static class Defaults
         /// <summary>
         /// The <see cref="TrainerInfo"/> contains general parameters for this trainer.
         /// </summary>
-        TrainerInfo ITrainer.Info => _info;
 
-        private readonly TrainerInfo _info;
+        public TrainerInfo Info { get; set; }
```
eerhardt: This definitely shouldn't be publicly settable.

Also - what is the reason for this change?

src/Microsoft.ML.AutoML/API/AutoCatalog.cs
```diff
@@ -13,6 +14,11 @@ public sealed class AutoCatalog
     {
         private readonly MLContext _context;
 
+        /// <summary>
+        /// save some intermediate value
+        /// </summary>
+        public static Dictionary<string, object> ValuePairs { get; set; } = new Dictionary<string, object>();
```
eerhardt: What is this for? This doesn't seem like a good API, we should stay away from static variables in our API.

src/Microsoft.ML.AutoML/API/AutoCatalog.cs
```diff
@@ -123,6 +129,11 @@ public MulticlassClassificationExperiment CreateMulticlassClassificationExperime
             return new MulticlassClassificationExperiment(_context, experimentSettings);
         }
 
+        public RecommendationExperiment CreateRecommendationExperiment(RecommendationExperimentSettings experimentSettings)
```
eerhardt: All public API needs xml comments.

src/Microsoft.ML.AutoML/ColumnInference/ColumnPurpose.cs
```diff
@@ -13,6 +13,7 @@ internal enum ColumnPurpose
         TextFeature = 4,
         Weight = 5,
         ImagePath = 6,
-        SamplingKey = 7
+        SamplingKey = 7,
+        LabelFeature,   // CategoricalFeature that requires ValueToKey converter, better naming?
```
eerhardt: Should follow the same pattern as above and explicitly give this a value.

eerhardt:
> using System;

All product code needs a copyright.

---
Refers to: src/Microsoft.ML.AutoML/API/RecommendationExperiment.cs:1 in 9695ffe. [](commit_id = 9695ffe2136d3db3b97533df8ffab42671572dc8, deletion_comment = False)

eerhardt:

This cannot go into master yet.

---
Refers to: src/mlnet/mlnet.csproj:1 in 9695ffe. [](commit_id = 9695ffe2136d3db3b97533df8ffab42671572dc8, deletion_comment = False)

src/Microsoft.ML.Recommender/MatrixFactorizationPredictor.cs
LittleLittleCloud: Good question!! That brings to another topic ( In `ITrainerEsitmator<ITransformer, IModel>` Does `ITransformer` has to be a `ISingleFeaturePredictionTransformer`?). I feel like the answer is no here, and `MatrixFactorizationPredictionTransformer` shouldn't implement that `ISingleFeaturePredictionTransformer` interface. So although it seems to implement `ISingleFeaturePredictionTransformer`, it actually not. And we need to change that

The reason why currently it implement that interface is because now `AutoML` requires all it's trainers to be `ITrainerEsitmator<ITransformer, IModel>`. So in order to add `MatrixFactorization` to AutoML, it has to be a `ISingleFeaturePredictionTransformer`

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
LittleLittleCloud: `TrainerInputBaseWithLabel` adds `LabelColumnName` to class, which is already exists in `Options` class

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
LittleLittleCloud: I will remove the `set` property

To implement `ITrainerEstimater` interface, see [this](https://github.com/dotnet/machinelearning/blob/edfd10f131abe97a7f7bb25c0070843a67b8b660/src/Microsoft.ML.Data/Training/ITrainerEstimator.cs#L12)

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
eerhardt: It _also_ adds a `FeatureColumnName` because it inherits from `TrainerInputBase`

https://github.com/dotnet/machinelearning/blob/edfd10f131abe97a7f7bb25c0070843a67b8b660/src/Microsoft.ML.Data/Training/TrainerInputBase.cs#L29-L33

src/Microsoft.ML.Recommender/MatrixFactorizationPredictor.cs
eerhardt: > AutoML requires all it's trainers to be ITrainerEsitmator<ITransformer, IModel>

Maybe AutoML needs to change then.

@Dmitry-A @srsaggam @vinodshanbhag @justinormont

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
eerhardt: To implement ITrainerEstimater you can use the same pattern as above - 

```
        /// <summary>
        /// The <see cref="TrainerInfo"/> contains general parameters for this trainer.
        /// </summary>
        TrainerInfo ITrainer.Info => _info;
```

src/Microsoft.ML.AutoML/API/AutoCatalog.cs
LittleLittleCloud: That's the shortcut to make it functionally working... Because `MatrixFactorizationTrainer` accepts both `MatrixColumnIndexColumnName` and `MatrixRowIndexColumnName` as input feature columns, but currently `AutoML` requires it's trainers only accept one `feature` columns as input. To change that we need to do some refactor work on `AutoML` side, so that's just the quick way to make it work

src/Microsoft.ML.AutoML/ColumnInference/ColumnPurpose.cs
justinormont: I think we'd want two new explicit column purposes of: 
* ColumnPurpose.UserId
* ColumnPurpose.ItemId

I don't know that we have a method of telling them apart otherwise. Feel free to correct me if wrong.

justinormont: I think we'd want to use `HashToKey` (name be off) instead of the mentioned `ValueToKey` as the `ValueToKey `will map future unseen values to NA in your test dataset; and as a lesser issue is slow by taking a full pass of the dataset.

justinormont: Agreed. 

For other places in the ML.NET codebase, this helps with backwards compatibility. The numeric value gets serialized in the model, then read by the newer code, which may have additional values inserted or old values removed from the enum. Without the numeric form, the enum will get deserialized to the wrong value.

src/Microsoft.ML.AutoML/TransformInference/TransformInference.cs
```diff
@@ -330,12 +356,14 @@ public static SuggestedTransform[] InferTransforms(MLContext context, TaskKind t
                 suggestedTransforms.AddRange(suggestions);
             }
 
-            var finalFeaturesConcatTransform = BuildFinalFeaturesConcatTransform(context, suggestedTransforms, intermediateCols);
-            if (finalFeaturesConcatTransform != null)
+            if(task != TaskKind.Recommendation)
```
justinormont: Space:
```suggestion
            if (task != TaskKind.Recommendation)
```

test/Microsoft.ML.AutoML.Tests/TrainerExtensionsTests.cs
```diff
@@ -21,6 +21,11 @@ public void TrainerExtensionInstanceTests()
                 .Except(new[] { TrainerName.Ova });
             foreach (var trainerName in trainerNames)
             {
+                if(trainerName == TrainerName.MatrixFactorization)
```
justinormont: Space:
```suggestion
                if (trainerName == TrainerName.MatrixFactorization)
```

docs/samples/Microsoft.ML.AutoML.Samples/RecommendationExperiment.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            // STEP 6: Save the best model for later deployment and inferencing
+            using (FileStream fs = File.Create(ModelPath))
+                mlContext.Model.Save(bestRun.Model, trainDataView.Schema, fs);
+
+            // STEP 7: Create prediction engine from the best trained model
+            var predictionEngine = mlContext.Model.CreatePredictionEngine<Movie, TaxiTripFarePrediction>(bestRun.Model);
+
+            // STEP 8: Initialize a new test taxi trip, and get the predicted fare
+            var testTaxiTrip = new Movie
```
justinormont: Various references to 'taxi' in the example:
```suggestion
            var testRecommendation = new Movie
```

src/Microsoft.ML.AutoML/API/AutoCatalog.cs
justinormont: See my feedback about creating two explicit ColumnPurposes for these two.

src/Microsoft.ML.AutoML/API/RecommendationExperiment.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Linq;
+using System.Text;
+using Microsoft.ML.Data;
+using Microsoft.ML.Trainers;
+
+namespace Microsoft.ML.AutoML
+{
+    public enum RecommendationExperimentScenario
+    {
+        MF,
```
justinormont: Is this enum to support the three styles of recommendation?

Also, should number enums.

src/Microsoft.ML.AutoML/API/RecommendationExperiment.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        // We can use RegressionMetric as evaluation Metric
+        public RegressionMetric OptimizingMetric { get; set; }
+
+        public ICollection<RecommendationTrainer> Trainers { get; }
+
+        public RecommendationExperimentSettings(RecommendationExperimentScenario scenario, string columnIndexName, string rowIndexName)
+            : this()
+        {
+            if(scenario == RecommendationExperimentScenario.MF)
```
justinormont: Space:
```suggestion
            if (scenario == RecommendationExperimentScenario.MF)
```

src/Microsoft.ML.AutoML/API/RecommendationExperiment.cs
```diff
@@ -0,0 +1,74 @@
...
+                AutoCatalog.ValuePairs[nameof(MatrixFactorizationTrainer.Options.MatrixColumnIndexColumnName)] = columnIndexName;
+                AutoCatalog.ValuePairs[nameof(MatrixFactorizationTrainer.Options.MatrixRowIndexColumnName)] = rowIndexName;
+                return;
+            }
+            throw new NotImplementedException();
+        }
+
+        private RecommendationExperimentSettings()
+        {
+            OptimizingMetric = RegressionMetric.RSquared;
```
justinormont: I think this depends on which style of recommendation we are using. If I recall, we had a method which produces binary classification metrics.

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
```diff
@@ -115,6 +117,15 @@ public static IEnumerable<SweepableParam> BuildLightGbmParams()
             };
         }
 
+        public static IEnumerable<SweepableParam> BuildMatrixFactorizationParmas()
+        {
+            return new SweepableParam[]
+            {
+                new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.NumberOfIterations), new object[] { 10, 20 }),
+                new SweepableFloatParam(nameof(MatrixFactorizationTrainer.Options.LearningRate), 0.025f, 0.04f, numSteps:2),
+                new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.ApproximationRank), new object[] { 10, 20 }),
```
justinormont: Please copy sweep ranges & hyperparameters from:
https://github.com/dotnet/machinelearning/blob/edfd10f131abe97a7f7bb25c0070843a67b8b660/src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs#L173-L250

Ideally, in the future, we should access them directly from the learner instead of having our copy in AutoML.

src/Microsoft.ML.AutoML/TransformInference/TransformInference.cs
```diff
@@ -180,6 +183,29 @@ public override IEnumerable<SuggestedTransform> Apply(IntermediateColumn[] colum
...
+                            continue;
+                        }
+
+                        if (!column.Type.IsKey())
+                        {
+                            yield return ValueToKeyMappingExtension.CreateSuggestedTransform(Context, column.ColumnName, column.ColumnName);
+                        }
+                    }
+                }
+            }
```
justinormont: Somewhere earlier, we should push the MatrixColumnIndexColumnName in to the Columns.SamplingKey, then handle it nearby here.

Specifically, this causes the dataset to split on the the UserId when creating either the CV or TrainTest splits. Data leakage other occurs as part of the same user's data is in the train set and part is in the test dataset. We should test on distinct users not seen in the training set.

If no splitting occurs in the AutoML code, for instance if a train/validate dataset is explicitly provided, this added part will have no impact (which is good).

justinormont: We should also allow the user to set a different `SamplingKey`, or empty it should they want to. 

For instance, they may want to enforce that their model can be predictive across geographic boundaries, so they may pick their `State` column as the `SamplingKey` which would put all users from the same state in the same CrossValidation fold.

src/Microsoft.ML.Recommender/MatrixFactorizationPredictor.cs
justinormont: @daholste (once he's back) may also know why it is a `ITrainerEstimator<ITransformer, IModel>`; and the impact of changing it.

src/Microsoft.ML.AutoML/API/RecommendationExperiment.cs
LittleLittleCloud: yup and sure

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
LittleLittleCloud: Seems that automl will try 5*4*3*3*... set of params on MatrixFactorization, will that cost too much time?

justinormont: tldr; SMAC focuses our search efforts, and we are bounded by our total runtime.

----

This is expected (and should be greater). The Bayesian hyperparameter optimization (SMAC) focuses on the useful areas of the search space. 

For selected trainers, we first do 20 iterations of random sweeping to warm up the search space. Then SMAC uses the found results of those 20 iterations to predict which areas of the hyperparameter space is best to explore next. The choice of what to explore next is based on which areas areas are doing best and which areas are unexplored. 

Beyond SMAC we have additional hyperparameter optimization algos we should be using like [KDO](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Sweeper/Algorithms/KdoSweeper.cs).

src/Microsoft.ML.AutoML/API/RecommendationExperiment.cs
justinormont: Should this enum be public?

We could spell-out the Scenarios (MatrixFactorization instead of MF). We may want to make them generically named instead of named after the trainers; hopefully in the future, more trainers will be added for each of the three scenarios.

Perhaps we could fill in the other two styles. This can make it easier for the next coder. Though it's debatable, so up to you.

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
justinormont: For your comment in the PR description:
>  figuring out how to do sweepParams for MatrixFactorization, the current one seems to have some bugs and it's never stop! (maybe SMACSweeper does not converge?)

Is AutoML not stopping at the specified timeout? If you set it to 60s of runtime, it should stop soon after this limit.

AutoML is designed to do round robin between three trainers culled from 8-11 depending on the task. Since this has only one trainer to choose from, perhaps the AutoML code needs updating?

LittleLittleCloud: Turns out that I forget to set the experiment time so it is defaulted to 16400... My Bad

src/Microsoft.ML.AutoML/API/RecommendationExperiment.cs
maryamariyan: As instructed above I will rename MF to MatrixFactorization

What would be the name for the other two?

codemzs:
@wschin Can you please review the recommended part of this PR in ML .NET and confirm the change is ok?

docs/samples/Microsoft.ML.AutoML.Samples/RecommendationExperiment.cs
```diff
@@ -0,0 +1,85 @@
...
+using Microsoft.ML.AutoML.Samples.DataStructures;
+using Microsoft.ML.Data;
+
+namespace Microsoft.ML.AutoML.Samples
+{
+    public static class RecommendationExperiment
+    {
+        private static string TrainDataPath = @"C:\Users\xiaoyuz\Desktop\machinelearning-samples\datasets\recommendation-ratings-train.csv";
+        private static string TestDataPath = @"C:\Users\xiaoyuz\Desktop\machinelearning-samples\datasets\recommendation-ratings-test.csv";
+        private static string ModelPath = @"C:\Users\xiaoyuz\source\test\recommendation.zip";
```
maryamariyan: TODO: change to

```
        private static string TrainDataPath = "<Path to your train dataset goes here>";
        private static string TestDataPath = "<Path to your test dataset goes here>";
        private static string ModelPath = @"<Desired model output directory goes here>\TaxiFareModel.zip";
```

src/Microsoft.ML.AutoML/ColumnInference/ColumnPurpose.cs
justinormont: #resolved.

src/Microsoft.ML.AutoML/API/RecommendationExperiment.cs
```diff
@@ -0,0 +1,76 @@
...
+namespace Microsoft.ML.AutoML
+{
+    public enum RecommendationExperimentScenario
+    {
+        MatrixFactorization = 0,
+    }
+
+    public sealed class RecommendationExperimentSettings : ExperimentSettings
+    {
+        public RecommendationExperimentScenario Scenerio { get; set; }
```
pgovind: Typo in "Scenerio"?

src/Microsoft.ML.AutoML/API/RecommendationExperiment.cs
maryamariyan: cc @justinormont @eerhardt  ^ question above

docs/samples/Microsoft.ML.AutoML.Samples/RecommendationExperiment.cs
```diff
@@ -0,0 +1,85 @@
...
+            inputColumnInformation.LabelColumnName = "Rating";
+
+            // STEP 2: Run AutoML experiment
+            Console.WriteLine($"Running AutoML recommendation experiment for {ExperimentTime} seconds...");
+            ExperimentResult<RegressionMetrics> experimentResult = mlContext.Auto()
+                .CreateRecommendationExperiment(settings)
+                .Execute(trainDataView, inputColumnInformation);
+
+            // STEP 3: Print metric from best model
+            RunDetail<RegressionMetrics> bestRun = experimentResult.BestRun;
```
maryamariyan: @justinormont @eerhardt 
Are we OK with using RegressionMetrics for Recommendation tasks? (assuming no)
Or should I create a new RecommendationMetrics task? (assuming yes)

docs/samples/Microsoft.ML.AutoML.Samples/RecommendationExperiment.cs
```diff
@@ -0,0 +1,85 @@
...
+            Console.WriteLine($"Total models produced: {experimentResult.RunDetails.Count()}");
+            Console.WriteLine($"Best model's trainer: {bestRun.TrainerName}");
+            Console.WriteLine($"Metrics of best model from validation data --");
+            PrintMetrics(bestRun.ValidationMetrics);
+
+            // STEP 5: Evaluate test data
+            IDataView testDataViewWithBestScore = bestRun.Model.Transform(testDataView);
+            RegressionMetrics testMetrics = mlContext.Regression.Evaluate(testDataViewWithBestScore, labelColumnName: LabelColumnName);
```
maryamariyan: same question here, MLContext does not have a `RecommendationCatalog Recommendation` property. Perhaps I should add this as well.

cc: @eerhardt @justinormont

src/Microsoft.ML.AutoML/API/RecommendationExperiment.cs
```diff
@@ -0,0 +1,76 @@
...
+            Trainers = Enum.GetValues(typeof(RecommendationTrainer)).OfType<RecommendationTrainer>().ToList();
+        }
+    }
+
+    public enum RecommendationTrainer
+    {
+        MatrixFactorization,
+    }
+
+    public sealed class RecommendationExperiment : ExperimentBase<RegressionMetrics, RecommendationExperimentSettings>
```
maryamariyan: Observation: for RecommendationExperiment we are using RegressionMetrics.

docs/samples/Microsoft.ML.AutoML.Samples/RecommendationExperiment.cs
maryamariyan: This comment answers the question for RegressionMetrics: https://github.com/dotnet/machinelearning/pull/4246/files#r328404787

Seems like it's OK for now, until there are more trainers added

eerhardt: The `Recommendation` evaluation in ML.NET uses `RegressionMetrics`, so I think we are safe.

https://github.com/dotnet/machinelearning/blob/3f98485e887177f2082f027c22f46fd64ebbe246/src/Microsoft.ML.Recommender/RecommenderCatalog.cs#L101-L117

docs/samples/Microsoft.ML.AutoML.Samples/RecommendationExperiment.cs
eerhardt: It does - but `Recommendation()` is an extension method on `MLContext`, and not a direct property.

https://github.com/dotnet/machinelearning/blob/3f98485e887177f2082f027c22f46fd64ebbe246/src/Microsoft.ML.Recommender/RecommenderCatalog.cs#L16-L19

https://github.com/dotnet/machinelearning/blob/3f98485e887177f2082f027c22f46fd64ebbe246/src/Microsoft.ML.Recommender/RecommenderCatalog.cs#L101-L117

maryamariyan: I found out why MLContext doesn't have RecommendationCatalog. Because there is this static helper used instead:
https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Recommender/RecommenderCatalog.cs#L19

It's been used throughout the repo. So I'll keep it as is.

maryamariyan: Thanks @eerhardt just saw your reply. I will use Recommendation() instead.

test/Microsoft.ML.AutoML.Tests/Microsoft.ML.AutoML.Tests.csproj
```diff
@@ -1,8 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <ItemGroup>
+    <ProjectReference Include="..\..\docs\samples\Microsoft.ML.AutoML.Samples\Microsoft.ML.AutoML.Samples.csproj" />
```
maryamariyan: remove

docs/samples/Microsoft.ML.AutoML.Samples/RecommendationExperiment.cs
```diff
@@ -0,0 +1,84 @@
...
+            // STEP 2: Run AutoML experiment
+            Console.WriteLine($"Running AutoML recommendation experiment for {ExperimentTime} seconds...");
+            ExperimentResult<RegressionMetrics> experimentResult = mlContext.Auto()
+                .CreateRecommendationExperiment(ExperimentTime)
+                .Execute(trainDataView, testDataView,
+                    new ColumnInformation()
+                    {
+                        LabelColumnName = LabelColumnName,
+                        MatrixColumnIndexColumnName = UserColumnName,
+                        MatrixRowIndexColumnName = ItemColumnName
```
maryamariyan: User and Item ids, instead.

maryamariyan: For now I named them
MatrixColumnIndexColumnName and MatrixRowIndexColumnName.

Could rename it to: UserIdColumnName and ItemIdColumnName

cc: @wschin

src/Microsoft.ML.AutoML/API/RecommendationExperiment.cs
```diff
@@ -0,0 +1,52 @@
...
+            : base(context,
+                  new RegressionMetricsAgent(context, settings.OptimizingMetric),
+                  new OptimizingMetricInfo(settings.OptimizingMetric),
+                  settings,
+                  TaskKind.Recommendation,
+                  TrainerExtensionUtil.GetTrainerNames(settings.Trainers))
+        {
+        }
+
+        private protected override CrossValidationRunDetail<RegressionMetrics> GetBestCrossValRun(IEnumerable<CrossValidationRunDetail<RegressionMetrics>> results)
```
maryamariyan: I haven't experimented with SamplingKey and CrossValidation yet. 

cc: @justinormont

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
```diff
@@ -115,6 +117,15 @@ public static IEnumerable<SweepableParam> BuildLightGbmParams()
             };
         }
 
+        public static IEnumerable<SweepableParam> BuildMatrixFactorizationParams()
+        {
+            return new SweepableParam[]
+            {
+                new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.NumberOfIterations), new object[] { 10, 20, 40 }),
+                new SweepableFloatParam(nameof(MatrixFactorizationTrainer.Options.LearningRate), 0.001f, 0.1f, numSteps: 2),
+                new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.ApproximationRank), new object[] { 8, 16, 64, 128 }),
```
maryamariyan: I took the values here from MatrixFactorizationTrainer
@justinormont could you please cross check if this is OK?

test/Microsoft.ML.AutoML.Tests/UserInputValidationTests.cs
```diff
@@ -33,10 +33,13 @@ public void ValidateExperimentExecuteNullLabel()
         [Fact]
         public void ValidateExperimentExecuteLabelNotInTrain()
         {
-            var ex = Assert.Throws<ArgumentException>(() => UserInputValidationUtil.ValidateExperimentExecuteArgs(Data,
-                new ColumnInformation() { LabelColumnName = "L" }, null, TaskKind.Regression));
+            foreach (var task in new[] { TaskKind.Recommendation, TaskKind.Regression })
```
maryamariyan: I would have passed down TaskKind as arg using Theory/InlineData but because it is an internal type, I went against that idea.

src/Microsoft.ML.Data/Training/ITrainerEstimator.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.ML.Trainers
     /// <typeparam name="TTransformer">The type of the transformer returned by the estimator.</typeparam>
     /// <typeparam name="TModel">The type of the model parameters.</typeparam>
     public interface ITrainerEstimator<out TTransformer, out TModel> : IEstimator<TTransformer>
-        where TTransformer : ISingleFeaturePredictionTransformer<TModel>
```
maryamariyan: Also need to confirm if there is a better way than loosening this restriction here, since we don't want ISingleFeaturePredictionTransformer for MatrixFactorization

test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
```diff
@@ -60,5 +63,92 @@ public void AutoFitRegressionTest()
...
+                        MatrixRowIndexColumnName = itemColumnName
+                    });
+
+            RunDetail<RegressionMetrics> bestRun = experimentResult.BestRun;
+            Assert.True(experimentResult.RunDetails.Count() > 1);
+            Assert.NotNull(bestRun.ValidationMetrics);
+            Assert.True(experimentResult.RunDetails.Max(i => i.ValidationMetrics.RSquared != 0));
+
+            var outputSchema = bestRun.Model.GetOutputSchema(trainDataView.Schema);
+            var expectedOutputNames = new string[] { labelColumnName, userColumnName, userColumnName, itemColumnName, itemColumnName, scoreColumnName };
```
maryamariyan: - [ ] revisit MatrixColumn.Apply method which applies ValueToKey mapping

Given test data with three columns: 
```
Label	Row	Column
...
1	14	35
1	14	36
1	14	37
1	14	38
1	14	39
3	15	0
3	15	1
3	15	3
3	15	4
3	15	5
3	15	6
...
```

As shown in line 98, we will get output schema with 5 columns. That is due to line 204 in the diffs of TransformInference.cs

I'm not sure if this is OK.
It looks odd compared to test MatrixFactorizationSimpleTrainAndPredict in MatrixFactorizationTests where
```
            var expectedOutputNames = new string[] { labelColumnName, userColumnName, itemColumnName, scoreColumnName };
```

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
```diff
@@ -110,7 +110,7 @@ namespace Microsoft.ML.Trainers
     /// <seealso cref="Microsoft.ML.RecommendationCatalog.RecommendationTrainers.MatrixFactorization(MatrixFactorizationTrainer.Options)"/>
     /// <seealso cref="Options"/>
     public sealed class MatrixFactorizationTrainer : ITrainer<MatrixFactorizationModelParameters>,
-        IEstimator<MatrixFactorizationPredictionTransformer>
+        IEstimator<MatrixFactorizationPredictionTransformer>, ITrainerEstimator<MatrixFactorizationPredictionTransformer, MatrixFactorizationModelParameters>
```
eerhardt: (nit) you don't need both `IEstimator` and `ITrainerEstimator`.  Since `ITrainerEstimator` inherits from `IEstimator`, you can just put `ITrainerEstimator<MatrixFactorizationPredictionTransformer, MatrixFactorizationModelParameters>` here, and remove `IEstimator<MatrixFactorizationPredictionTransformer>` for simplicity.

src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs
```diff
@@ -302,6 +302,13 @@ internal static class Defaults
 
         PredictionKind ITrainer.PredictionKind => PredictionKind.Recommendation;
 
+        /// <summary>
+        /// The <see cref="TrainerInfo"/> contains general parameters for this trainer.
+        /// </summary>
+        TrainerInfo ITrainer.Info => _info;
```
eerhardt: You can remove this. It is no longer needed now that we expose the `public TrainerInfo Info`.

eerhardt: But put the XML doc comment above on the public property below.

---
In reply to: [335631831](https://github.com/dotnet/machinelearning/pull/4246#discussion_r335631831) [](ancestors = 335631831)

src/Microsoft.ML.AutoML/Microsoft.ML.AutoML.csproj
```diff
@@ -10,6 +10,7 @@
     <ProjectReference Include="..\Microsoft.ML.CpuMath\Microsoft.ML.CpuMath.csproj" />
     <ProjectReference Include="..\Microsoft.ML.LightGbm\Microsoft.ML.LightGbm.csproj" />
     <ProjectReference Include="..\Microsoft.ML.Mkl.Components\Microsoft.ML.Mkl.Components.csproj" />
+    <ProjectReference Include="..\Microsoft.ML.Recommender\Microsoft.ML.Recommender.csproj" />
```
eerhardt: You will also need to update the AutoML nupkgproj as well:


[https://github.com/dotnet/machinelearning/blob/461b7d38c06d64e53c76e29770665c4b0bb44e72/pkg/Microsoft.ML.AutoML/Microsoft.ML.AutoML.nupkgproj#L9-L11](https://github.com/dotnet/machinelearning/blob/461b7d38c06d64e53c76e29770665c4b0bb44e72/pkg/Microsoft.ML.AutoML/Microsoft.ML.AutoML.nupkgproj#L9-L11)

src/Microsoft.ML.AutoML/Assembly.cs
```diff
@@ -11,3 +11,5 @@
 [assembly: InternalsVisibleTo("Benchmark, PublicKey=00240000048000009400000006020000002400005253413100040000010001004b86c4cb78549b34bab61a3b1800e23bfeb5b3ec390074041536a7e3cbd97f5f04cf0f857155a8928eaa29ebfd11cfbbad3ba70efea7bda3226c6a8d370a4cd303f714486b6ebc225985a638471e6ef571cc92a4613c00b8fa65d61ccee0cbe5f36330c9a01f4183559f1bef24cc2917c6d913e3a541333a1d05d9bed22b38cb")]
 
 [assembly: WantsToBeBestFriends]
+[assembly: InternalsVisibleTo("Microsoft.ML.CodeGenerator, PublicKey=00240000048000009400000006020000002400005253413100040000010001004b86c4cb78549b34bab61a3b1800e23bfeb5b3ec390074041536a7e3cbd97f5f04cf0f857155a8928eaa29ebfd11cfbbad3ba70efea7bda3226c6a8d370a4cd303f714486b6ebc225985a638471e6ef571cc92a4613c00b8fa65d61ccee0cbe5f36330c9a01f4183559f1bef24cc2917c6d913e3a541333a1d05d9bed22b38cb")]
```
eerhardt: I believe these can be removed. If we didn't need them before, we shouldn't need them now.

src/Microsoft.ML.AutoML/Utils/UserInputValidationUtil.cs
```diff
@@ -20,6 +20,8 @@ internal static class UserInputValidationUtil
         private const string TextColumnPurposeName = "text";
         private const string IgnoredColumnPurposeName = "ignored";
         private const string SamplingKeyColumnPurposeName = "sampling key";
+        private const string MatrixColumnIndexColumnName = "matrix column index";
```
eerhardt: I thought we were going to rename these to "User ID" and "Item ID"?

eerhardt: I see, MatrixFactorization calls these columns those names.  Feel free to ignore this.

---
In reply to: [335649246](https://github.com/dotnet/machinelearning/pull/4246#discussion_r335649246) [](ancestors = 335649246)

maryamariyan: Please refer to https://github.com/dotnet/machinelearning/pull/4246#discussion_r335316079

I want to make sure nobody is against the UserId and ItemId naming before I finalize rename.
@justinormont would you be OK with using UserId/ItemId?

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
```diff
@@ -115,6 +117,15 @@ public static IEnumerable<SweepableParam> BuildLightGbmParams()
             };
         }
 
+        public static IEnumerable<SweepableParam> BuildMatrixFactorizationParams()
+        {
+            return new SweepableParam[]
+            {
+                new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.NumberOfIterations), new object[] { 10, 20, 40 }),
+                new SweepableFloatParam(nameof(MatrixFactorizationTrainer.Options.LearningRate), 0.001f, 0.1f, numSteps: 2),
```
eerhardt: How come this is different than:

[https://github.com/dotnet/machinelearning/blob/461b7d38c06d64e53c76e29770665c4b0bb44e72/src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs#L216-L218](https://github.com/dotnet/machinelearning/blob/461b7d38c06d64e53c76e29770665c4b0bb44e72/src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs#L216-L218)

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
```diff
@@ -115,6 +117,15 @@ public static IEnumerable<SweepableParam> BuildLightGbmParams()
             };
         }
 
+        public static IEnumerable<SweepableParam> BuildMatrixFactorizationParams()
+        {
+            return new SweepableParam[]
```
eerhardt: What about the other sweepable params in MatrixFactorizationTrainer.cs?

* LossFunction
* Lambda
* Alpha
* C

Do we want to sweep over them?

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
maryamariyan: OK changing 
```
new SweepableFloatParam(nameof(MatrixFactorizationTrainer.Options.LearningRate), 0.001f, 0.1f, numSteps: 2),
```
to 
```
new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.LearningRate), new object[] { 0.001f, 0.01f, 0.1f }),
```
in next commit

src/Microsoft.ML.AutoML/API/RecommendationExperiment.cs
```diff
@@ -0,0 +1,52 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.ML.Data;
+
+namespace Microsoft.ML.AutoML
+{
+    public sealed class RecommendationExperimentSettings : ExperimentSettings
```
eerhardt: All these new public APIs should have XML doc comments on them.

src/Microsoft.ML.AutoML/API/ColumnInference.cs
```diff
@@ -53,6 +53,16 @@ public sealed class ColumnInformation
         /// <value>The default value is "Label".</value>
         public string LabelColumnName { get; set; }
 
+        /// <summary>
+        /// The dataset column to use as a matrix column for computation.
+        /// </summary>
+        public string MatrixColumnIndexColumnName { get; set; }
```
eerhardt: Should these public APIs be "User ID" and "Item ID"?

docs/samples/Microsoft.ML.AutoML.Samples/RecommendationExperiment.cs
```diff
@@ -0,0 +1,93 @@
...
+            PrintMetrics(testMetrics);
+
+            // STEP 6: Save the best model for later deployment and inferencing
+            using (FileStream fs = File.Create(ModelPath))
+                mlContext.Model.Save(bestRun.Model, trainDataView.Schema, fs);
+
+            // STEP 7: Create prediction engine from the best trained model
+            var predictionEngine = mlContext.Model.CreatePredictionEngine<Movie, MovieRatingPrediction>(bestRun.Model);
+
+            // STEP 8: Initialize a new test, and get the predicted fare
```
eerhardt: `predicted fare` ?

docs/samples/Microsoft.ML.AutoML.Samples/RecommendationExperiment.cs
```diff
@@ -0,0 +1,93 @@
...
+            PrintMetrics(bestRun.ValidationMetrics);
+
+            // STEP 5: Evaluate test data
+            IDataView testDataViewWithBestScore = bestRun.Model.Transform(testDataView);
+            RegressionMetrics testMetrics = mlContext.Recommendation().Evaluate(testDataViewWithBestScore, labelColumnName: LabelColumnName);
+            Console.WriteLine($"Metrics of best model on test data --");
+            PrintMetrics(testMetrics);
+
+            // STEP 6: Save the best model for later deployment and inferencing
+            using (FileStream fs = File.Create(ModelPath))
```
eerhardt: There is an overload to `mlContext.Model.Save` that just takes a path instead of a stream. You can use that instead of creating the FileStream here.

src/Microsoft.ML.Data/Training/ITrainerEstimator.cs
maryamariyan: A separate conversation is ongoing in terms of finding a better way to do this. It doesn't block merging this PR.
If further change ends up needing to be done, I can submit in a follow up PR.

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
maryamariyan: Added more sweepable params.

docs/samples/Microsoft.ML.AutoML.Samples/RecommendationExperiment.cs
```diff
@@ -0,0 +1,91 @@
...
+            // STEP 8: Initialize a new test, and get the prediction
+            var testMovie = new Movie
+            {
+                UserId = "1",
+                MovieId = "1097",
+            };
+            var prediction = predictionEngine.Predict(testMovie);
+            Console.WriteLine($"Predicted rating for: {prediction.Rating}");
+
+            // Note: MatrixFactorization only can predict for existing users
```
daholste: nit: // Note: only predict for existing users
(maybe avoid mentioning a single algorithm in general recommendation sample. not critical ofc)

src/Microsoft.ML.AutoML/Experiment/SuggestedTrainer.cs
```diff
@@ -41,7 +41,7 @@ public SuggestedTrainer Clone()
             return new SuggestedTrainer(_mlContext, _trainerExtension, _columnInfo, HyperParamSet?.Clone());
         }
 
-        public ITrainerEstimator<ISingleFeaturePredictionTransformer<object>, object> BuildTrainer()
+        public ITrainerEstimator<IPredictionTransformer<object>, object> BuildTrainer()
```
daholste: nice!

docs/samples/Microsoft.ML.AutoML.Samples/RecommendationExperiment.cs
```diff
@@ -0,0 +1,91 @@
...
+            IDataView testDataView = mlContext.Data.LoadFromTextFile<Movie>(TestDataPath, hasHeader: true, separatorChar: ',');
+
+            // STEP 2: Run AutoML experiment
+            Console.WriteLine($"Running AutoML recommendation experiment for {ExperimentTime} seconds...");
+            ExperimentResult<RegressionMetrics> experimentResult = mlContext.Auto()
+                .CreateRecommendationExperiment(new RecommendationExperimentSettings() { MaxExperimentTimeInSeconds = ExperimentTime })
+                .Execute(trainDataView, testDataView,
+                    new ColumnInformation()
+                    {
+                        UserIdColumnName = UserColumnName,
```
eerhardt: Don't you need to set the "Label" column here?

test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
```diff
@@ -60,5 +62,91 @@ public void AutoFitRegressionTest()
...
+            // STEP 1: Load data
+            var reader = new TextLoader(mlContext, GetLoaderArgs(labelColumnName, userColumnName, itemColumnName));
+            var trainDataView = reader.Load(new MultiFileSource(GetDataPath(TestDatasets.trivialMatrixFactorization.trainFilename)));
+            var testDataView = reader.Load(new MultiFileSource(GetDataPath(TestDatasets.trivialMatrixFactorization.testFilename)));
+
+            // STEP 2: Run AutoML experiment
+            ExperimentResult<RegressionMetrics> experimentResult = mlContext.Auto()
+                .CreateRecommendationExperiment(5)
+                .Execute(trainDataView, testDataView,
+                    new ColumnInformation() {
```
eerhardt: Don't you need to set the Label column here? Or is it inferred because it is explicitly named `Label`?

Do we throw an exception of the Label column is not found?

maryamariyan: The default column name for label is "Label" and is implicitly specified.
We would throw if Label column is not found. That is why this test does not fail.

I will add the LabelColumnName specifically on the next commit.

src/Microsoft.ML.AutoML/TransformInference/TransformInference.cs
```diff
@@ -183,9 +183,9 @@ public override IEnumerable<SuggestedTransform> Apply(IntermediateColumn[] colum
                 }
             }
 
-            internal sealed class MatrixColumns : TransformInferenceExpertBase
+            internal sealed class UserIdPairs : TransformInferenceExpertBase
```
eerhardt: I'm not crazy about this name. How about `RecommendationColumns`?

PR #4247, opened 2019-09-25
Fix NgramExtractingTransformer GetSlotNames to not allocate a new delegate on every invoke.

eerhardt:
This was showing up as allocating a huge number of delegates for no reason.

I removed the delegate all together and just pass the VBuffer in.


eerhardt:
Using @ganik's [NimbusML PR](https://github.com/microsoft/NimbusML/pull/278) as a test benchmark, I'm seeing the following results.

Before this change:
image

After:
image

PR #4260, opened 2019-09-27
Tensor extensions

KsenijaS:
Output tensors from graph are being copied from unamanaged to managed memory and every time a new buffer is created. Performance will be better if we reuse buffer when it's possible.
Improve performance by replacing ToArray method in Tensor class with extension methods: ToScalar, ToSpan and ToArray


src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
```diff
@@ -903,7 +907,7 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
 
                             var editor = VBufferEditor.Create(ref dst, (int)tensorSize);
 
-                            DnnUtils.FetchData<T>(tensor.ToArray<T>(), editor.Values);
+                            tensor.ToSpan<T>(editor.Values);
```
eerhardt: I would name this method `CopyTo`

```C#
tensor.CopyTo(editor.Values);
```

Note that you can drop the generic on `CopyTo<T>` since it can be inferred from the parameter being passed in.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+using Tensorflow;
+
+namespace Microsoft.ML.Transforms
+{
+    [BestFriend]
+    internal static class TensorTypeExtensions
+    {
+        public static void ToScalar<T>(this Tensor tensor, ref T dst) where T : unmanaged
+        {
+            if (typeof(T).as_dtype() == tensor.dtype)
```
eerhardt: is it expected to just no-op when the types mis-match? Would it be better to throw an exception?

eerhardt:
> using System;

copyright header is needed on all code

---
Refers to: src/Microsoft.ML.Dnn/TensorTypeExtensions.cs:1 in 0b2c97e. [](commit_id = 0b2c97eb34128b30443920521c0c2809c4d77576, deletion_comment = False)

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                unsafe
+                {
+                    var len = (long)tensor.size;
+                    fixed (T* dst = values)
+                    {
+                        var src = (T*)tensor.buffer;
+                        len *= ((long)tensor.itemsize);
+                        System.Buffer.MemoryCopy(src, dst, len, len);
```
eerhardt: Doing it this way is dangerous because we don't know if the `values` span is big enough. So copying a too big buffer into it will trash memory.

Instead, I would write this the other way. Turn the `tensor.buffer` into a `Span<T>`, and then use `Span.CopyTo(Span<T> destination)` to do the copy. That will throw if there isn't enough space.

Note:

The way you create a `Span<T>` is to use this constructor:

```
        public Span(void* pointer, int length);
```

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public static void ToArray<T>(this Tensor tensor, ref T[] array) where T : unmanaged
+        {
+            ulong arrayLen = 0;
+            if (array != null)
+                arrayLen = (ulong)array.Length;
+
+            if (array == null || arrayLen == 0 || arrayLen < tensor.size)
+            {
+                array = new T[tensor.size];
```
eerhardt: We have a utility for this. We should use that instead.

`Utils.EnsureSize`.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+                arrayLen = tensor.size;
+            }
+
+            if (typeof(T).as_dtype() == tensor.dtype)
```
eerhardt: Similar to above - should we throw if it isn't? And if not, then we probably shouldn't be allocating an array above and then not doing anything with it.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+        public static void ToArray<T>(this Tensor tensor, ref T[] array) where T : unmanaged
+        {
+            ulong arrayLen = 0;
+            if (array != null)
+                arrayLen = (ulong)array.Length;
+
+            if (array == null || arrayLen == 0 || arrayLen < tensor.size)
+            {
+                array = new T[tensor.size];
+                arrayLen = tensor.size;
```
eerhardt: We don't use `arrayLen` later, why are we changing it here?

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+
+            if (typeof(T).as_dtype() == tensor.dtype)
+            {
+                unsafe
+                {
+                    var len = (long)tensor.size;
+                    fixed (T* dst = array)
```
eerhardt: Once you have an array that is big enough, you can just call the overload above that takes in a Span, and pass the array into it. No need to copy the logic below.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
KsenijaS: When I use CopyTo method I get a test failure:
InvalidOperationException : Splitter/consolidator worker encountered exception while consuming source data
    ---- System.ArgumentException : Destination is too short.

eerhardt: That's perfect! You caught a bug. You should see why your destination is too short.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public static void CopyTo<T>(this Tensor tensor, Span<T> values) where T: unmanaged
+        {
+            if (typeof(T).as_dtype() != tensor.dtype)
+                throw new NotSupportedException();
+
+            unsafe
+            {
+                var len = (long)tensor.size;
+                fixed (T* dst = values)
```
eerhardt: You don't need to pin the `values` span. You can remove this line.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+                {
+                    var src = (T*)tensor.buffer;
+                    var span = new Span<T>(src, (int)len);
```
eerhardt: Just to be safe, we should add a `checked` keyword around this cast. That way if `len` is more than int.MaxValue, an exception will occur.

See [https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/checked](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/checked)

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+                    span.CopyTo(values);
+                }
+            }
+        }
+
+        public static void ToArray<T>(this Tensor tensor, ref T[] array) where T : unmanaged
+        {
+            Utils.EnsureSize(ref array, (int)tensor.size, (int)tensor.size, true);
```
eerhardt: You actually want to pass in `keepOld: false` here. `keepOld` will retain the old values inside of `array` when it increases its length (by calling `Array.Resize`). You have no need for the old values, since you will be overwriting them in the `CopyTo` call below. So there is no need to copy the old values to the new memory.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -235,15 +235,17 @@ public Tensor ProcessImage(string path)
...
                     labelGetter(ref label);
                     imagePathGetter(ref imagePath);
                     var imagePathStr = imagePath.ToString();
                     var imageTensor = imageProcessor.ProcessImage(imagePathStr);
                     runner.AddInput(imageTensor, 0);
-                    var featurizedImage = runner.Run()[0]; // Reuse memory?
-                    writer.WriteLine(label - 1 + "," + string.Join(",", featurizedImage.ToArray<float>()));
+                    var featurizedImage = runner.Run()[0]; // Reuse memory
+                    featurizedImage.ToArray<float>(ref imageArray);
+                    writer.WriteLine(label - 1 + "," + string.Join(",", imageArray));
```
eerhardt: This is potentially wrong because `imageArray.Length` has the potential to be bigger than the actual values stored in `featurizedImage`.

Image 2 rows. One has 1000 elements. The second has 500 elements. When the first one runs, it will create an array of size `1000` and copy all the values into it. On the next pass through the loop, when you get to the row that has `500` elements in it, you will reuse the `1000`-length array from the first loop and only fill up 500 elements in it with the 2nd row of data. But here you are writing out all `1000` elements because you are using every element in the array.

However, if you are guaranteed that every `featurizedImage` through the loop will have the same size, then you should probably add an assert here to catch any bugs.

KsenijaS: Can I use instead an assert, string join method like this:
public static string Join (string separator, string[] value, int startIndex, int count);

src/Microsoft.ML.TensorFlow/TensorTypeExtensions.cs
```diff
@@ -0,0 +1,121 @@
...
+                    }
+                }
+            }
+            else
+            {
+                if (tensor.NDims == 0 && size == 1) //is it a scalar?
+                {
+                    unsafe
+                    {
+#if _REGEN
```
Nucs: I suggest to add a comment mentioning that this is a template written in Regen with a url:
https://github.com/Nucs/Regen

src/Microsoft.ML.TensorFlow/TensorTypeExtensions.cs
```diff
@@ -0,0 +1,121 @@
...
+                        #region Compute
+		                switch (dtype.as_numpy_dtype().GetTypeCode())
+		                {
+			                %foreach supported_dtypes,supported_dtypes_lowercase%
```
Nucs: These variables are stored [here](https://github.com/SciSharp/NumSharp/blob/master/src/NumSharp.Core/globals.regen)

I would copy them locally, the template will look like this:
```C#
%supported_dtypes = ["Boolean","Byte","Int16","UInt16","Int32","UInt32","Int64","UInt64","Char","Double","Single","Decimal"]
%supported_dtypes_lowercase = ["bool","byte","short","ushort","int","uint","long","ulong","char","double","float","decimal"]
%foreach supported_dtypes,supported_dtypes_lowercase%
```

src/Microsoft.ML.TensorFlow/TensorTypeExtensions.cs
Nucs: Nevermind, I just realized it was removed later on #resolved

Nucs:
After this PR will be merged, If needed - I'll add performant support for conversion between types instead of throwing `NotSupportedException`.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
eerhardt: You can't use that method because the array is a `string[]` in that method, and you have a `float[]`.

KsenijaS: I know, but I didn't get any errors and all tests passed when I used that method.

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
```diff
@@ -903,7 +907,7 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
 
                             var editor = VBufferEditor.Create(ref dst, (int)tensorSize);
 
-                            DnnUtils.FetchData<T>(tensor.ToArray<T>(), editor.Values);
```
eerhardt: Is the method `DnnUtils.FetchData` used anymore with this change? I believe it can now be removed.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        public static void CopyTo<T>(this Tensor tensor, Span<T> values) where T: unmanaged
+        {
+            if (typeof(T).as_dtype() != tensor.dtype)
+                throw new NotSupportedException();
+
+            unsafe
+            {
+                var len = (long)tensor.size;
```
eerhardt: (nit) you might as well cast to `int` right away, instead of casting to `long` and then casting to `int`.

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
KsenijaS: No, it's not used. I will remove it.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -33,9 +33,9 @@ internal static class TensorTypeExtensions
 
             unsafe
             {
-                var len = (long)tensor.size;
```
eerhardt: the `checked` should happen here.

src/Microsoft.ML.Dnn/DnnRetrainTransform.cs
```diff
@@ -379,14 +379,18 @@ private void TrainCore(DnnRetrainEstimator.Options options, IDataView input, IDa
...
                 runner.AddInput(inputs[i], srcTensorGetters[i].GetBufferedBatchTensor());
 
             Tensor[] tensor = runner.Run();
-            loss = tensor.Length > 0 && tensor[0] != IntPtr.Zero ? (float)tensor[0].ToArray<float>()[0] : 0.0f;
-            metric = tensor.Length > 1 && tensor[1] != IntPtr.Zero ? (float)tensor[1].ToArray<float>()[0] : 0.0f;
+            if (tensor.Length > 0 && tensor[0] != IntPtr.Zero)
+                tensor[0].ToScalar<float>(ref loss);
+
+            if (tensor.Length > 1 && tensor[1] != IntPtr.Zero)
+                tensor[1].ToScalar<float>(ref metric);
```
codemzs: In this case it is guaranteed that you will have two tensors, see line 354-357 please remove the if condition check:
tensor[0].ToScalar<float>(ref loss);
tensor[1].ToScalar<float>(ref metric);

Issue #3976, opened 2019-07-09
It is not possible to use PermutationFeatureImportance from a model loaded from disk in F#


fwaris:
I am trying to use PermutationFeatureImportance (PFI)  with F# but the F# type system is not resolving ITransformer to ISingleFeaturePredictionTransformer - which is required by PFI.

I believe it is due to  IPredictorProducing (and related interfaces) being marked as "internal".

F# supports explicit interfaces and maybe that is the reason for this issue.

Here is a snippet of code that shows what I am trying to do
(I am using the latest bits - v 1.2.0 at the time of this post)

```F#
let mutable schema = null
let mdl = ctx.Model.Load(@"F:\fwaris\data\t\analysis\model_cv_LightGbmBinary.bin", &schema) 
let mdlt =  mdl :?> TransformerChain<ITransformer>
let m1 =  mdlt.LastTransformer //debugger shows it is Microsoft.ML.Data.BinaryPredictionTransformer<Microsoft.ML.IPredictorProducing<float>>
let scored = mdl.Transform(trainView)
scored.Preview()
ctx.BinaryClassification.PermutationFeatureImportance(m1 :?> _,scored)
```

@dsyme

fwaris:
work around for now is to use a C# helper given below but really if an interface (IPredictorProducing) is going to be exposed via another public interface, it should not really be marked internal.

```C#
        public static System.Collections.Immutable.ImmutableArray<BinaryClassificationMetricsStatistics> PFI_BinaryClassification
             
             (
                MLContext ctx,
                ITransformer model,
                IDataView data,
                string labelColumnName = "Label",
                bool useFeatureWeightFilter = false,
                int? numberOfExamplesToUse = null,
                int permutationCount = 1
            )
        {
            
            var m = ctx.BinaryClassification.PermutationFeatureImportance(
                    model as ISingleFeaturePredictionTransformer<T>, 
                    data, 
                    labelColumnName : labelColumnName, 
                    useFeatureWeightFilter : useFeatureWeightFilter, 
                    numberOfExamplesToUse : numberOfExamplesToUse, 
                    permutationCount : permutationCount
                    );
            return m;
        }
        
    }
```

eerhardt:
@fwaris - I just ran into this issue as well. I don't understand how your workaround works. What `T` is getting passed into `MLHelper<T>`?

[...]

PR #4262, opened 2019-09-28
Addresses #3976 about using PFI with a model loaded from disk

antoniovs1029:
With this pull request it is now possible to use PFI with some models loaded from disk. **This is not yet a final solution to the problem**, as described in the last section below.

### The Problem
As explained in [this comment](https://github.com/dotnet/machinelearning/issues/3976#issuecomment-517862076) of issue #3976 it was not possible to use PFI with a model loaded from disk, because the last transformer of the loaded model was not of the appropriate type to use it with the `PermutationFeatureImportance `method.

Specifically the problem occurred because when loading the last transformer, a 'create' method would be called and it would assign an inappropriate type to the last transformer, making it unusable for PFI. For example, if a model had `RegressionPredictionTransformer<OlsModelParameters>` as last transformer, and it was saved to disk, later on when loading it from disk the last transformer would be of type `RegressionPredictionTransformer<IPredictorProducing<float>>.` This would have happened because the [Create method ](https://github.com/dotnet/machinelearning/blob/bb00e07b30e9626b3578ff1934b86dad0d1d1ce9/src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs#L615) that is called when loading a `RegressionPredictionTransformer ` would always return a `RegressionPredictionTransformer<IPredictorProducing<float>>` regardless of the actual `TModel `that should be loaded. In this case, it would be necessary to load the last transformer as `RegressionPredictionTransformer<OlsModelParameters>` in order to use it with PFI.

This was a problem also in the BinaryClassification, MulticlassClassification and Ranking prediction transformers which implemented a similar Create method. All of these classes are used with PFI.

### The approach of the solution
The main obstacle was that the appropriate type `TModel` to be used when loading the last transformer couldn't be known at compile time, only at runtime once the last transformer was being loaded.

So to solve the problem, it was necessary to load first the internal model (e.g. the `OlsModelParameters `object) in the Create method of the prediction transformer, get its type to be used as `TModel`, make a generic type at runtime for the prediction transformer using the actual `TModel`, and instantiate that type with a constructor that would receive the internal model (previously loaded) to add it to the last transformer; this constructor would then continue to load the prediction transformer.

### Changes implemented

- The create method of the prediction transformers (binary, multiclass, ranking and regression) was modified to solve the problem. Different overloads of constructors were created to receive the internal model once it was loaded and then continue loading the prediction transformer.

- **Samples were added based on the original PFI samples**, but now using a model that is saved and loaded from disk directly in the sample. This is provided for the multiclass, ranking and regression, but NOT for the binary classification case (see the final section below).

- **Test cases based on the original PFI test cases are also provided**, but now using a model that is saved and loaded from disk directly in the test case. Again, this was done for multiclass, ranking and regression, but not for binary classification.

- **Changes were made to 2 tests in the `LbfgsTests.cs`** file where PFI is not used, but prediction transformers are loaded from disk. The changes regard casts that were used in the tests, but can no longer be used. For example, `as MulticlassPredictionTransformer<IPredictorProducing<VBuffer<float>>>` was replaced for `as MulticlassPredictionTransformer<MaximumEntropyModelParameters>` in one test. This is done because now the create method returns the appropriate TModel type instead of the `IPredictorProducing<>` type. Notice that it is invalid to cast from `MulticlassPredictionTransformer<MaximumEntropyModelParameters>` to `MulticlassPredictionTransformer<IPredictorProducing<VBuffer<float>>>`... so those tests fail without the changes I made to the tests. Also notice that since `IPredictorProducing<>` is internal, a regular user using a nugget wouldn't be able to do the casts that were done in those tests.

### Further problems
In this pull request I provide working samples and tests for regression, multiclass and ranking. Still, I've been unable to provide them for binary classification.

The [existing sample for PFI with binary classification](https://github.com/dotnet/machinelearning/blob/2942ca4e02d354b48767a1b01017f57cdc3fe44c/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/BinaryClassification/PermutationFeatureImportance.cs#L27) uses a last transformer of type `BinaryPredictionTransformer<CalibratedModelParameterBase<LinearBinaryModelParameters,PlattCalibrator>>` which is the type necessary to use it in the PFI method. When saving and then loading that model from disk, the last transformer was of type `BinaryPredictionTransformer<IPredictorProducing<float>>`; with the changes I made to the binary transformer, it now loads a last transformer of type `BinaryPredictionTransformer<ParameterMixingCalibratorModelParameters<IPredictorProducing<float>, ICalibrator>>`, which is still not the type necessary to use PFI.

The problem is that the [Create method of ParameterMixingCalibratorModelParameters](https://github.com/dotnet/machinelearning/blob/2942ca4e02d354b48767a1b01017f57cdc3fe44c/src/Microsoft.ML.Data/Prediction/Calibrator.cs#L564) returns a `CalibratedModelParameterBase` object. This is similar to the problem that the prediction transformers had, but there's a key difference in that `ParameterMixingCalibratorModelParameters ` is internal, and its create method returns an object of its public superclass `CalibratedModelParameterBase`. This key difference has stopped me from solving the problem using the same approach used to fix the prediction transformers. Once this pull request is accepted, I will open another issue with this specific use case, explaining it in more detail.

Nonetheless, notice that the problem is not in the `BinaryPredictionTransformer<TModel>` but rather in its `TModel` for this specific case. Having changed the `BinaryPredictionTransformer<TModel>` actually works as expected, in that it no longer returns a fixed `<IPredictorProducing<float>>` as TModel, and so the problem is actually in the `ParameterMixingCalibratorModelParameters `class.

Also notice that this is a signal that there might be other generic classes where the Create method returns an object with fixed type parameters that aren't the ones actually being used. This might become a problem for users trying to use PFI with a model that uses one of such classes.


test/Microsoft.ML.Functional.Tests/Explainability.cs
```diff
@@ -49,6 +51,43 @@ public void GlobalFeatureImportanceWithPermutationFeatureImportance()
...
+
+            var modelAndSchemaPath = GetOutputPath("TestFunctionalTestPFI.zip");
+            mlContext.Model.Save(model, data.Schema, modelAndSchemaPath);
+
+            ITransformer loadedModel;
+            using (var fs = File.OpenRead(modelAndSchemaPath))
+                loadedModel = mlContext.Model.Load(modelAndSchemaPath, out var schema);
+
+            var transformedData = loadedModel.Transform(data);
+            var linearPredictor = (loadedModel as TransformerChain<ITransformer>).LastTransformer as RegressionPredictionTransformer<LinearRegressionModelParameters>;
```
yaeldekel: >RegressionPredictionTransformer<LinearRegressionModelParameters> [](start = 101, length = 64)

Wouldn't this work if we write `RegressionPredictionTransformer<object>`?

antoniovs1029: It doesn't work since the cast to `RegressionPredictionTransformer<object>` returns null, and having a null linearPredictor makes it impossible to use the PFI method.

I've just tested this, and this happens both with or without the changes I've made to the Prediction Transformers in this pull request.

Notice that originally the `.LastTransformer ` would return a `RegressionPredictionTransformer<IPredictorProducing<float>>` object, but with the changes in this pull request it now returns a `RegressionPredictionTransformer<LinearRegressionModelParameters>` object. So, as I've stated, in both cases a cast to `RegressionPredictionTransformer<object>` returns null.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -2,22 +2,24 @@
...
+using System;
 using System.IO;
+using System.Reflection;
 using Microsoft.ML;
 using Microsoft.ML.Data;
 using Microsoft.ML.Data.IO;
 using Microsoft.ML.Runtime;
 
-[assembly: LoadableClass(typeof(BinaryPredictionTransformer<IPredictorProducing<float>>), typeof(BinaryPredictionTransformer), null, typeof(SignatureLoadModel),
+[assembly: LoadableClass(typeof(object), typeof(BinaryPredictionTransformer), null, typeof(SignatureLoadModel),
```
eerhardt: This part seems a little concerning. What are the drawbacks of putting `object` here?

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -109,6 +110,30 @@ private protected PredictionTransformerBase(IHost host, ModelLoadContext ctx)
...
+
+            // *** Binary format ***
+            // model: prediction model.
+            // stream: empty data view that contains train schema.
+            // id of string: feature column.
+
+            Model = model;
+
+            // Clone the stream with the schema into memory.
+            var ms = new MemoryStream();
```
eerhardt: Each new constructor here seems to have a decent amount of duplicated code with the existing constructor. Can it be refactored to share initialize logic?

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -595,33 +671,125 @@ private static VersionInfo GetVersionInfo()
...
     internal static class RegressionPredictionTransformer
     {
         public const string LoaderSignature = "RegressionPredXfer";
+        private const string DirModel = "Model"; // This should match "DirModel" in PredictionTransformerBase<>
+
+        public static object Create(IHostEnvironment env, ModelLoadContext ctx)
+        {
+            // Load internal model to be used as TModel of RegressionPredictionTransformer<TModel>
+            var host = Contracts.CheckRef(env, nameof(env)).Register(nameof(RegressionPredictionTransformer<IPredictorProducing<float>>));
+            ctx.LoadModel<IPredictorProducing<float>, SignatureLoadModel>(host, out IPredictorProducing<float> model, DirModel); // MYTODO: don't hardcode the DirModel
```
eerhardt: What is the plan for addressing this `MYTODO` ?

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -595,33 +671,125 @@ private static VersionInfo GetVersionInfo()
     internal static class BinaryPredictionTransformer
     {
         public const string LoaderSignature = "BinaryPredXfer";
+        private const string DirModel = "Model"; // This should match "DirModel" in PredictionTransformerBase<>
+
+        public static object Create(IHostEnvironment env, ModelLoadContext ctx)
+        {
+            // Load internal model to be used as TModel of BinaryPredictionTransformer<TModel>
```
eerhardt: These methods appears to be duplicated, can they be refactored to avoid so much duplication?

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -595,33 +671,125 @@ private static VersionInfo GetVersionInfo()
     internal static class BinaryPredictionTransformer
     {
         public const string LoaderSignature = "BinaryPredXfer";
+        private const string DirModel = "Model"; // This should match "DirModel" in PredictionTransformerBase<>
```
eerhardt: Can the existing `const string` be reused instead?

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
antoniovs1029: I believe that it has to be 'object' in there, since the Create method now returns an object. I was working with @codemzs when we decided to put 'object' in there to make it work. I am not sure about the drawbacks.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
antoniovs1029: I forgot to remove that comment. I should remove it, since it is no longer a problem there.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
antoniovs1029: I believe it can't without creating an instance of `PredictionTransformerBase<TModel>`. I tried other approaches but they didn't work. Do you know of a way to reuse the existing one in `PredictionTransformeBase<>` without making an instance?

antoniovs1029: I also thought of creating a new class just to hold the DirModel string, and reuse it in all of the other classes (including the existing one in PredictionTransformerBase). But I don't know if creating a whole other class only for that would be ok.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
antoniovs1029: I think it can be done. I just didn't want to modify the existing constructors. But I will change this.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
antoniovs1029: I can create a new class (perhaps named 'CreatePredictionTransformer') to hold an static Create Method which all the other Create methods call.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
eerhardt: You don't need to create an instance to reference a `const` variable. Rather, the issue here is that since `PredictionTransformerBase<TModel>` is generic, you need to fill out the generic type in order to reference the `const`.

Making a new `static class` that is not generic is probably the best approach for this.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
antoniovs1029: In my latest update I've created initialize logic methods, except for the following classes:
- `RegressionPredictionTransformer<>` and `RankingPredictionTransformer<>`, because their constructors were already one liners, and didn't see much use in refactoring them. If requested I can change them as well.
- `SingleFeaturePredictionTransformerBase<>`, because in its constructors there are several assignments to readonly properties, and I believe it wouldn't be possible to assign to them outside of the constructors (in an initialize logic method) without making other changes to the properties themselves. So I am unsure if should also go on and refactor this class.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
antoniovs1029: I've refactored this into the CreatePredictionTransformer static class I talked about.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
antoniovs1029: In my latest update I've created initialize logic methods, except for the following classes:
-` RegressionPredictionTransformer<>` and `RankingPredictionTransformer<>`, because their constructors were already one liners, and didn't see much use in refactoring them. If requested I can change them as well.
- `SingleFeaturePredictionTransformerBase<>`, because in its constructors there are several assignments to readonly properties, and I believe it wouldn't be possible to assign to them outside of the constructors (e.g. in an initialize logic method) without making other changes to the properties themselves. So I am unsure if should also go on and refactor this class.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
antoniovs1029: In my latest update I've refactored this.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
antoniovs1029: In my latest update I've created an internal static class `PredictionTransformerBase ` only to hold this string.

test/Microsoft.ML.Functional.Tests/Explainability.cs
yaeldekel: Oh, I see. Would it work with `ISingleFeaturePredictionTransformer<object>`? There is a test in the functional tests project that loads a model and extracts the predictor:
[https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Functional.Tests/ModelFiles.cs#L193](https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Functional.Tests/ModelFiles.cs#L193)

Would something similar work here?

---
In reply to: [329679240](https://github.com/dotnet/machinelearning/pull/4262#discussion_r329679240) [](ancestors = 329679240)

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
eerhardt: I think the current refactoring is enough. Thanks for cleaning it up a bit.

test/Microsoft.ML.Functional.Tests/Explainability.cs
antoniovs1029: Yes, a cast to `ISingleFeaturePredictionTransformer<object>` works in this test.

I've tested this with the changes I've made in this pull request, and without them. And it works in both cases.

antoniovs1029: So, should I change the cast in here to `ISingleFeaturePredictionTransformer<object>` for some reason?

Or are there other casts you want me to try out?

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
antoniovs1029: So we agree that it's ok to leave object in the LoadableClassAttribute, right? @codemzs @eerhardt

eerhardt: I took 20 minutes to look through the code. The only drawback I could find is when someone internally is trying to create one of these objects through the ComponentCatalog, they will have to ask for it using `typeof(object)` and then cast it down.

However, I didn't find any usages like this in the code when I searched. Since PredictionTransformer is a newer type (newer than the internal TLC code), there is a good chance no one is doing this internally. Also, I believe this only gets called when loading a serialized model, so I think we can make this change.

test/Microsoft.ML.Functional.Tests/Explainability.cs
```diff
@@ -49,6 +51,43 @@ public void GlobalFeatureImportanceWithPermutationFeatureImportance()
...
+                .Append(mlContext.Regression.Trainers.Sdca());
+
+            // Fit the pipeline and transform the data.
+            var model = pipeline.Fit(data);
+
+            var modelAndSchemaPath = GetOutputPath("TestFunctionalTestPFI.zip");
+            mlContext.Model.Save(model, data.Schema, modelAndSchemaPath);
+
+            ITransformer loadedModel;
+            using (var fs = File.OpenRead(modelAndSchemaPath))
```
eerhardt: You aren't using `fs` here. This can be removed.

test/Microsoft.ML.Functional.Tests/Explainability.cs
```diff
@@ -49,6 +51,43 @@ public void GlobalFeatureImportanceWithPermutationFeatureImportance()
                 Common.AssertMetricsStatistics(metricsStatistics);
         }
 
+        /// <summary>
+        /// GlobalFeatureImportance: PFI can be used to compute global feature importance. Here it is used with a model loaded from disk.
+        /// </summary>
+        [Fact]
```
eerhardt: You could probably turn the above method into a `[Theory]` and then pass in `true` and `false` for a `bool saveModel` parameter. That way you can reuse all the code, and only change what you do with the model in between fitting and transforming the data.

test/Microsoft.ML.Functional.Tests/Explainability.cs
yaeldekel: If it works without the change, then why do we need this change at all?

Is it because we don't expect the user to know which version of `PermutationFeatureImportance` to call (regression/classification)?

---
In reply to: [330194364](https://github.com/dotnet/machinelearning/pull/4262#discussion_r330194364) [](ancestors = 330194364)

test/Microsoft.ML.Tests/PermutationFeatureImportanceTests.cs
```diff
@@ -55,6 +57,48 @@ public void TestPfiRegressionOnDenseFeatures()
...
+        public void TestPfiRegressionOnDenseFeaturesWithLoadedModel()
+        {
+            var data = GetDenseDataset();
+            var model = ML.Regression.Trainers.OnlineGradientDescent().Fit(data);
+
+            var modelAndSchemaPath = GetOutputPath("TestPfiRegressionOnDenseFeatures.zip");
+            ML.Model.Save(model, data.Schema, modelAndSchemaPath);
+
+            ITransformer loadedModel;
+            using (var fs = File.OpenRead(modelAndSchemaPath))
```
eerhardt: You aren't using `fs` here. This can be removed.

test/Microsoft.ML.Tests/PermutationFeatureImportanceTests.cs
```diff
@@ -55,6 +57,48 @@ public void TestPfiRegressionOnDenseFeatures()
             Done();
         }
 
+        /// <summary>
+        /// Test PFI Regression for Dense Features with a model loaded from disk
+        /// </summary>
+        [Fact]
```
eerhardt: Similar to the functional tests - this could be a `[Theory]` instead, to not have to duplicate so much logic.

docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/MulticlassClassification/PermutationFeatureImportanceLoadFromDisk.cs
```diff
@@ -0,0 +1,138 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.ML;
+using Microsoft.ML.Calibrators;
+using Microsoft.ML.Data;
+using Microsoft.ML.Trainers;
+
+namespace Samples.Dynamic.Trainers.MulticlassClassification
+{
+    public static class PermutationFeatureImportance2
```
eerhardt: `PermutationFeatureImportance2`? Should this instead match the name of the file?

test/Microsoft.ML.Functional.Tests/Explainability.cs
eerhardt: > If it works without the change, then why do we need this change at all?

I don't believe users should be using `ISingleFeaturePredictionTransformer<object>` in a mainline scenario, like using PFI. They should be able to use the exact same type as what they got back when they created the model in the first place.

```C#
TransformerChain<RegressionPredictionTransformer<OlsModelParameters>> model = pipeline.Fit(data);
```

When I save and then load the above model, I should get back a `TransformerChain<RegressionPredictionTransformer<OlsModelParameters>>` object. But with the current code, the loaded object is a TransformerChain<RegressionPredictionTransformer<IPredictorProducing<float>>>

The root issue here is that the type of object that we get back when loading a model from disk is not the same as the type of object that is created in the first place. Thus, certain casts that work when you haven't saved the model to disk will work, and then fail for the same model after it is saved and then loaded.

antoniovs1029: I hadn't tried to cast to `ISingleFeaturePredictionTransformer<object>` until you've mentioned it in here, so I didn't know PFI would work that way, even without my changes. So now I am unsure if the changes I've donde in this pull request are actually necessary.

There is still a problem that isn't present in the test cases, but that [arises in the sample case of PFI with Regression](https://github.com/antoniovs1029/machinelearning/blob/8e190e480ae2270da6ee0f587f1a9cb7a52976fe/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/Regression/PermutationFeatureImportanceLoadFromDisk.cs#L79). Specifically what happens is that the code tries to acces `linearPredictor.Model.Weights[i]` and if `linearPredictor ` was created using the cast you are proposing:
`var linearPredictor =  (model as TransformerChain<ITransformer>).LastTransformer as ISingleFeaturePredictionTransformer<object>`

then linearPredictor doesn't have access to the Model.Weights member, and the compiler throws an error.

If I change the cast to `ISingleFeaturePredictionTransformer<OlsModelParameters>` then the compiler doesn't complain, and it works with the changes I've made in this pull request... but it doesn't work without them (the cast returns a null reference).

So, in short, it seems to me that if the user casts the last transformer to `ISingleFeaturePredictionTransformer<object>` then s/he will be able to use PFI even without my changes, but s/he might not be able to access the members of the parameters.

I don't know if there's a work around I am not seeing for this. But I will keep on trying other casts and see if I find one that works.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
eerhardt: Given @yaeldekel's other comment, we could change this to be `typeof(ISingleFeaturePredictionTransformer<object>)` instead, to give this at least some semblance of strong typing.

test/Microsoft.ML.Functional.Tests/Explainability.cs
antoniovs1029: > I don't believe users should be using ISingleFeaturePredictionTransformer<object> in a mainline scenario, like using PFI. They should be able to use the exact same type as what they got back when they created the model in the first place.

So I understand that the way to go is to keep my changes and not expect the user to use the  `ISingleFeaturePredictionTransformer<object>` cast, right?

Should I also add a test where loading a model from disk and casting it to  `ISingleFeaturePredictionTransformer<object>` doesn't work? Such as the situation I've described in my last comment in here?

docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/MulticlassClassification/PermutationFeatureImportanceLoadFromDisk.cs
antoniovs1029: You're right. I've fixed this in my last commit.

test/Microsoft.ML.Tests/PermutationFeatureImportanceTests.cs
antoniovs1029: Fixed in my last commit.

test/Microsoft.ML.Tests/PermutationFeatureImportanceTests.cs
antoniovs1029: Fixed in my last commit.

test/Microsoft.ML.Functional.Tests/Explainability.cs
antoniovs1029: Thanks for the feedback. I didn't know about [Theory]'s. I've used them in my latest commit, in all the tests I had provided.

test/Microsoft.ML.Functional.Tests/Explainability.cs
antoniovs1029: Fixed in my last commit.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
antoniovs1029: Thanks very much for taking the time to look at this.

In my latest commit I have updated the `LoadableClassAttribute`s with the `typeof(ISingleFeaturePredictionTransformer<object>)` you requested.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -91,12 +97,27 @@ private protected PredictionTransformerBase(IHost host, ModelLoadContext ctx)
...
+
+        [BestFriend]
+        private protected PredictionTransformerBase(IHost host, ModelLoadContext ctx, TModel model)
+        {
+            Host = host;
+            Model = model; // prediction model
+            InitializeLogic(host, ctx);
+        }
+
+        private protected void InitializeLogic(IHost host, ModelLoadContext ctx)
```
eerhardt: (minor) This (and other InitializeLogic methods) should just be `private`. No one outside of this class needs to call it.

antoniovs1029: Now fixed in my latest commit.

test/Microsoft.ML.Functional.Tests/Explainability.cs
codemzs: Agree on "They should be able to use the exact same type as what they got back when they created the model in the first place.". I think we should be expecting users to cast it to the same type they saved it not some generic type with object ... 

---
In reply to: [330212148](https://github.com/dotnet/machinelearning/pull/4262#discussion_r330212148) [](ancestors = 330212148)

PR #4272, opened 2019-10-01
Issue 4120, add reasonable exception when user try to use OnnxSequenceType attribute without specify sequence type

### LINKED ISSUE #4120: Using default OnnxSequenceType attribute throws meaningless exception
Pull the code at this commit and run the application - https://github.com/sethjuarez/SeeSharp/tree/b6a7e8a0094b99af05bc72528a2c6e306abbf7cd

The code of interest is:

```C# 
    public class ImagePrediction
    {
        [ColumnName("classLabel")]
        [VectorType]
        public string[] Prediction;


        [ColumnName("loss")]
        [OnnxSequenceType]
        public List<float> Loss;
    }
```

You get the following error:

```
Unhandled Exception: System.ArgumentNullException: Value cannot be null.
   at System.RuntimeType.MakeGenericType(Type[] instantiation)
   at Microsoft.ML.Transforms.Onnx.OnnxSequenceTypeAttribute.Register()
   at Microsoft.ML.Data.SchemaDefinition.Create(Type userType, Direction direction)
   at Microsoft.ML.Data.TypedCursorable`1.Create(IHostEnvironment env, IDataView data, Boolean ignoreMissingColumns, SchemaDefinition schemaDefinition)
   at Microsoft.ML.PredictionEngineBase`2.PredictionEngineCore(IHostEnvironment env, InputRow`1 inputRow, IRowToRowMapper mapper, Boolean ignoreMissingColumns, SchemaDefinition outputSchemaDefinition, Action& disposer, IRowReadableAs`1& outputRow)
   at Microsoft.ML.PredictionEngineBase`2..ctor(IHostEnvironment env, ITransformer transformer, Boolean ignoreMissingColumns, SchemaDefinition inputSchemaDefinition, SchemaDefinition outputSchemaDefinition)
   at Microsoft.ML.PredictionEngineExtensions.CreatePredictionEngine[TSrc,TDst](ITransformer transformer, IHostEnvironment env, Boolean ignoreMissingColumns, SchemaDefinition inputSchemaDefinition, SchemaDefinition outputSchemaDefinition)
   at SeeSharp.Program.LoadModel(String onnxModelFilePath) in /Users/eerhardt/git/SeeSharp/SeeSharp/Program.cs:line 70
   at SeeSharp.Program.Main(String[] args) in /Users/eerhardt/git/SeeSharp/SeeSharp/Program.cs:line 39
```

It looks like the code here:

https://github.com/dotnet/machinelearning/blob/0c0789f8eba1e47f6fa759de01f99588e560935c/src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs#L95-L100

Is expecting `_elemType` to never be null.  However, that is exactly what happens when you use the default attribute:

https://github.com/dotnet/machinelearning/blob/0c0789f8eba1e47f6fa759de01f99588e560935c/src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs#L57-L64

I don't think having a default constructor is right with this attribute - you can't use this sequence attribute without specifying an element type.

/cc @wschin

frank-dong-ms:
Fixes #4120 

The issue is: when user use OnnxSequenceType attribute directly without specify sequence type like [OnnxSequenceType], user will hit run time exception when try to use it.

The ideal way to fix this issue is to remove the default constructor of OnnxSequenceTypeAttribute and user will get compiler error when he/she try to use [OnnxSequenceType]. 
But this can be a break change in Public API. After discuss with @eerhardt and @ericstj , we decide to obsolete the default constructor for now. Then we will remove the default constructor in next major release.

### DIFF
--- a/src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs
+++ b/src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs
@@ -57,7 +57,8 @@ namespace Microsoft.ML.Transforms.Onnx
         private Type _elemType;
 
         // Comment out default constructor
-        // Use default contructor will left the _elemType field empty and cause exception in methods using _elemType
+        // Use default constructor will left the _elemType field empty and cause exception in methods using _elemType
+        // Comment out default constructor will give user compile error when user try to use [OnnxSequenceType] attribute directly without specify sequence type
         //public OnnxSequenceTypeAttribute()
         //{
         //}
--- a/src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs
+++ b/src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs
@@ -56,12 +56,13 @@ namespace Microsoft.ML.Transforms.Onnx
     {
         private Type _elemType;
 
-        // Comment out default constructor
-        // Use default constructor will left the _elemType field empty and cause exception in methods using _elemType
-        // Comment out default constructor will give user compile error when user try to use [OnnxSequenceType] attribute directly without specify sequence type
-        //public OnnxSequenceTypeAttribute()
-        //{
-        //}
+        // Make default constructor obsolete.
+        // Use default constructor will left the _elemType field empty and cause exception in methods using _elemType.
+        // User will receive compile error when try to use [OnnxSequenceType] attribute directly without specify sequence type
+        [Obsolete("Please specify sequence type when use OnnxSequenceType Attribute", true)]
+        public OnnxSequenceTypeAttribute()
+        {
+        }
 
         /// <summary>
         /// Create a <paramref name="elemType"/>-sequence type.
--- a/src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs
+++ b/src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs
@@ -94,6 +94,12 @@ namespace Microsoft.ML.Transforms.Onnx
         /// </summary>
         public override void Register()
         {
+            // this happens when use OnnxSequenceType attribute without specify sequence type
+            if (_elemType == null)
+            {
+                throw new Exception("Please specify sequence type when use OnnxSequenceType Attribute.");
+            }
+
             var enumerableType = typeof(IEnumerable<>);
             var type = enumerableType.MakeGenericType(_elemType);
             DataViewTypeManager.Register(new OnnxSequenceType(_elemType), type, this);
--- a/src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs
+++ b/src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs
@@ -56,12 +56,11 @@ namespace Microsoft.ML.Transforms.Onnx
     {
         private Type _elemType;
 
-        /// <summary>
-        /// Create a sequence type.
-        /// </summary>
-        public OnnxSequenceTypeAttribute()
-        {
-        }
+        // Comment out default constructor
+        // Use default contructor will left the _elemType field empty and cause exception in methods using _elemType
+        //public OnnxSequenceTypeAttribute()
+        //{
+        //}
 
         /// <summary>
         /// Create a <paramref name="elemType"/>-sequence type.
@@ -94,10 +93,6 @@ namespace Microsoft.ML.Transforms.Onnx
         /// </summary>
         public override void Register()
         {
-            // this happens when use OnnxSequenceType attribute without specify sequence type
-            if (_elemType == null)
-                throw new InvalidOperationException("Please specify sequence type when use OnnxSequenceType Attribute.");
-
             var enumerableType = typeof(IEnumerable<>);
             var type = enumerableType.MakeGenericType(_elemType);
             DataViewTypeManager.Register(new OnnxSequenceType(_elemType), type, this);
--- a/src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs
+++ b/src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs
@@ -96,9 +96,7 @@ namespace Microsoft.ML.Transforms.Onnx
         {
             // this happens when use OnnxSequenceType attribute without specify sequence type
             if (_elemType == null)
-            {
-                throw new Exception("Please specify sequence type when use OnnxSequenceType Attribute.");
-            }
+                throw new InvalidOperationException("Please specify sequence type when use OnnxSequenceType Attribute.");
 
             var enumerableType = typeof(IEnumerable<>);
             var type = enumerableType.MakeGenericType(_elemType);
--- a/test/Microsoft.ML.Tests/OnnxSequenceTypeTest.cs
+++ b/test/Microsoft.ML.Tests/OnnxSequenceTypeTest.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
 using System.Drawing;
 using Microsoft.ML.Data;
@@ -13,7 +14,7 @@ using Xunit.Abstractions;
 using System.Linq;
 using System.IO;
 using Microsoft.ML.TestFramework.Attributes;
-using System;
+
 
 namespace Microsoft.ML.Tests
 {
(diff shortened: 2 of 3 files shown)


src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs
```diff
@@ -94,6 +94,12 @@ public override int GetHashCode()
         /// </summary>
         public override void Register()
         {
+            // this happens when use OnnxSequenceType attribute without specify sequence type
+            if (_elemType == null)
+            {
+                throw new Exception("Please specify sequence type when use OnnxSequenceType Attribute.");
+            }
```
codemzs: please do not use braces for single line if condition

test/Microsoft.ML.Tests/OnnxSequenceTypeTest.cs
```diff
@@ -13,37 +13,55 @@
 using System.Linq;
 using System.IO;
 using Microsoft.ML.TestFramework.Attributes;
+using System;
```
codemzs: Please place "using System" at the top and sort alphabetical order as done in source files. System libraries take preference over non-system libraries like Microsoft.ML.*

src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs
harishsk: Perhaps it is better to throw ArgumentNullException here?

test/Microsoft.ML.Tests/OnnxSequenceTypeTest.cs
```diff
@@ -64,5 +82,21 @@ public void OnnxSequenceTypeWithColumnNameAttributeTest()
...
+
+            var ctx = new MLContext();
+            var model = PrepareModel(onnxModelFilePath, ctx);
+            return ctx.Model.CreatePredictionEngine<FloatInput, ProblematicOutputObj>(model);
+        }
+
+        [OnnxFact]
+        public void OnnxSequenceTypeWithouSpecifySequenceTypeTest()
+        {
+            Exception ex = Assert.Throws<Exception>(() => CreatePredictorWithPronlematicOutputObj());
```
harishsk: >CreatePredictorWithPronlematicOutputObj [](start = 58, length = 39)

A small typo here with Problematic

test/Microsoft.ML.Tests/OnnxSequenceTypeTest.cs
```diff
@@ -64,5 +82,21 @@ public void OnnxSequenceTypeWithColumnNameAttributeTest()
...
+            Assert.Equal("Please specify sequence type when use OnnxSequenceType Attribute.", ex.Message);
+        }
```
harishsk: Would it be better to assert on the type of exception rather than the specific message?

src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs
eerhardt: Throwing just a plain `Exception` is not recommended.

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/using-standard-exception-types

X DO NOT throw System.Exception or System.SystemException.

However, ArgumentNullException is not correct either, since there is no argument to this method.

✓ DO throw ArgumentException or one of its subtypes if bad arguments are passed to a member.

In this case, InvalidOperationException is the most correct:

✓ DO throw an InvalidOperationException if the object is in an inappropriate state.

eerhardt: IMO - it the best approach would be to remove the empty constructor and verify the `Type elemType` in the other constructor is never null.

test/Microsoft.ML.Tests/OnnxSequenceTypeTest.cs
frank-dong-ms: yes, will do

---
In reply to: [330344198](https://github.com/dotnet/machinelearning/pull/4272#discussion_r330344198) [](ancestors = 330344198)

test/Microsoft.ML.Tests/OnnxSequenceTypeTest.cs
frank-dong-ms: yes, will do, thanks

---
In reply to: [330650375](https://github.com/dotnet/machinelearning/pull/4272#discussion_r330650375) [](ancestors = 330650375)

test/Microsoft.ML.Tests/OnnxSequenceTypeTest.cs
frank-dong-ms: yes, the exception type is assert in previous line Assert.Throws<Exception>, after change to use ArgumentNullException, it should look like Assert.Throws<ArgumentNullException>(....)

---
In reply to: [330651709](https://github.com/dotnet/machinelearning/pull/4272#discussion_r330651709) [](ancestors = 330651709)

src/Microsoft.ML.OnnxTransformer/OnnxSequenceType.cs
frank-dong-ms: Yes, will do

---
In reply to: [330649635](https://github.com/dotnet/machinelearning/pull/4272#discussion_r330649635) [](ancestors = 330649635)

frank-dong-ms: Actually this is best practice to always use braces for if condition, in my previous team it is also suggested to always use braces for if condition to avoid further problem.

There is also some discussion online, please check out and see whether this make sense to you: [https://softwareengineering.stackexchange.com/questions/16528/single-statement-if-block-braces-or-no](https://softwareengineering.stackexchange.com/questions/16528/single-statement-if-block-braces-or-no) 



---
In reply to: [330344007](https://github.com/dotnet/machinelearning/pull/4272#discussion_r330344007) [](ancestors = 330344007)

frank-dong-ms: Offline discussed with Zeeshan, this can be debatable, I will just follow the coding style in this team, not a big issue

---
In reply to: [330672086](https://github.com/dotnet/machinelearning/pull/4272#discussion_r330672086) [](ancestors = 330672086,330344007)

frank-dong-ms: Thanks, very good suggestion, InvalidOperationException is the most proper one to use here

---
In reply to: [330652629](https://github.com/dotnet/machinelearning/pull/4272#discussion_r330652629) [](ancestors = 330652629)

frank-dong-ms: Yes, I have a similar discussion with Harish about this yesterday, we agreed that remove empty constructor seems to be more ambiguous because when our customer use [OnnxSequenceType] attribute directly he/she will get error message like "method is inaccessible due to protection level". By throwing exception we can give user more specific error message. Make sense to you?

---
In reply to: [330653643](https://github.com/dotnet/machinelearning/pull/4272#discussion_r330653643) [](ancestors = 330653643)

frank-dong-ms: I would like to take Eric's suggestion and use InvalidOperationException here.

---
In reply to: [330649635](https://github.com/dotnet/machinelearning/pull/4272#discussion_r330649635) [](ancestors = 330649635)

eerhardt: >  when our customer use [OnnxSequenceType] attribute directly he/she will get error message like "method is inaccessible due to protection level"

That's a compile time error - which is exactly what we want. We want to raise the error as early as possible, so the user knows they are doing something wrong.

Also, when I get a different error message than what you are displaying:

```C#
    class Program
    {
        [My]
        private int foo;
    }

    class MyAttribute : Attribute
    {
        public MyAttribute(int required)
        { }
    }
```

```
Error	CS7036	There is no argument given that corresponds to the required formal parameter 'required' of 'MyAttribute.MyAttribute(int)'	ConsoleApp80	C:\Users\eerhardt\source\repos\ConsoleApp80\Program.cs	13	Active
```

Pretty obvious what to do to fix it.

harishsk: I think Eric's suggestion is the better approach here. Please modify it as per his suggestion.

frank-dong-ms: Sounds good to me, will do

---
In reply to: [330736190](https://github.com/dotnet/machinelearning/pull/4272#discussion_r330736190) [](ancestors = 330736190)

PR #4273, opened 2019-10-01
Add DateTime to DateTime standard conversion.

pieths:
This fixes an error which is caused by a missing DateTime to DateTime conversion when outputting DateTime columns in NimbusML. NimbusML calls in to `RowCursorUtils.GetGetterAsCore` (indirectly) which tries to find a conversion from DateTime to DateTime and fails with the following error:

```console
Error: *** System.InvalidOperationException: 'No standard conversion from 'DateTime' to 'DateTime'
```

### DIFF
--- a/src/Microsoft.ML.Data/Data/Conversion.cs
+++ b/src/Microsoft.ML.Data/Data/Conversion.cs
@@ -231,6 +231,7 @@ namespace Microsoft.ML.Data.Conversion
             AddStd<DT, I8>(Convert);
             AddStd<DT, R4>(Convert);
             AddStd<DT, R8>(Convert);
+            AddStd<DT, DT>(Convert);
             AddAux<DT, SB>(Convert);
 
             AddStd<DZ, I8>(Convert);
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Data/Data/Conversion.cs
```diff
@@ -1673,5 +1674,9 @@ public void Convert(in TX src, ref SB dst)
         public void Convert(in BL src, ref R8 dst) => dst = System.Convert.ToDouble(src);
         public void Convert(in BL src, ref BL dst) => dst = src;
         #endregion FromBL
+
+        #region ToDT
+        public void Convert(in DT src, ref DT dst) => dst = src;
```
codemzs: >public void Convert(in DT src, ref DT dst) => dst = src; [](start = 8, length = 56)

Does not have test coverage as per [https://codecov.io/gh/dotnet/machinelearning/pull/4273/diff](https://codecov.io/gh/dotnet/machinelearning/pull/4273/diff)

pieths: Added a unit test for the new DateTime to DateTime conversion method.

---
In reply to: [330341048](https://github.com/dotnet/machinelearning/pull/4273#discussion_r330341048) [](ancestors = 330341048)

codemzs: Thanks, PH!

---
In reply to: [330650008](https://github.com/dotnet/machinelearning/pull/4273#discussion_r330650008) [](ancestors = 330650008,330341048)

PR #4290, opened 2019-10-03
TensorTypeExtensions: Added conversion between Tensor to primitive C# types instead of throwing NotSupportedException

Nucs:
- Added conversion between Tensor to native C# types instead of throwing NotSupportedException
- Fixed proper reading of TF_STRING
- Added overloads that support TF_STRING (string does not meet T generic constraint `unmanaged`


src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -3,49 +3,477 @@
...
+                    case TF_DataType.TF_COMPLEX64:
+                    case TF_DataType.TF_COMPLEX128:
+                    default:
+                        throw new NotSupportedException();
+                }
+            }
         }
 
-        public static void CopyTo<T>(this Tensor tensor, Span<T> values) where T: unmanaged
+        public static void CopyTo(this Tensor tensor, Span<string> destination)
```
eerhardt: Who is calling this method? We shouldn't have dead code in the repo.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -3,49 +3,477 @@
...
+
+                    case TF_DataType.TF_COMPLEX64:
+                    case TF_DataType.TF_COMPLEX128:
+                    default:
+                        throw new NotSupportedException();
+                }
             }
+        }
 
+        public static void ToScalar(this Tensor tensor, ref string dst)
```
eerhardt: I don't believe any code is calling this method. Can it be removed?

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -3,49 +3,477 @@
...
-                var len = checked((int)tensor.size);
-                var src = (T*)tensor.buffer;
-                var span = new Span<T>(src, len);
-                span.CopyTo(values);
+                var len = checked((int) tensor.size);
+                //perform regular CopyTo using Span.CopyTo.
+                if (typeof(T).as_dtype() == tensor.dtype && tensor.dtype != TF_DataType.TF_STRING) //T can't be a string but tensor can.
+                {
+                    var src = (T*) tensor.buffer;
+                    var srcspan = new Span<T>(src, len);
```
eerhardt: (nit) - variables should be camelCased. `srcspan` is not one word - it should be `srcSpan`.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -3,49 +3,477 @@
...
+                    var dst = dst_;
+                    switch (tensor.dtype)
+                    {
+#if _REGEN
+                        %foreach supported_numericals_TF_DataType,supported_numericals,supported_numericals_lowercase%
+                        case TF_DataType.#1:
+                        {
+                            var converter = Converts.FindConverter<#3, T>();
+                            var src = (#3*) tensor.buffer;
+                            Parallel.For(0, len, i => *(dst + i) = converter(unchecked(*(src + i))));
```
eerhardt: Have you measured the perf of this? In the past, I've seen a tight loop inside a `Parallel.For` take longer than just a single loop. This is because `Parallel.For` needs to start new threads, and do coordination between the threads. When the work inside the loop is small, adding threading just makes this slower.

Nucs: Tensorflow.NET and NumSharp use `Parallel.For` extensively. Entirely up to you.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
Nucs: Right, `TensorTypeExtensions` is internal. My bad.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
Nucs: Right, `TensorTypeExtensions` is internal. My bad.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
eerhardt: I would want to see a perf comparison that proves `Parallel.For` is faster before I would use it in a tight loop.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -3,49 +3,477 @@
...
+%supported_numericals_TF_DataType_full = ["TF_DataType.TF_BOOL","TF_DataType.TF_UINT8","TF_DataType.TF_INT16","TF_DataType.TF_UINT16","TF_DataType.TF_INT32","TF_DataType.TF_UINT32","TF_DataType.TF_INT64","TF_DataType.TF_UINT64","TF_DataType.TF_DOUBLE","TF_DataType.TF_FLOAT"]
+#endif
+
 namespace Microsoft.ML.Transforms
 {
     [BestFriend]
     internal static class TensorTypeExtensions
     {
         public static void ToScalar<T>(this Tensor tensor, ref T dst) where T : unmanaged
         {
```
codemzs: Can this handle string tensors?

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
codemzs: Eric, I'm curious if this method is for string tensors? If it is we should keep it as it is one of the types we may need to use when extracting output tensors. 

---
In reply to: [331242079](https://github.com/dotnet/machinelearning/pull/4290#discussion_r331242079) [](ancestors = 331242079)

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
Nucs: No because string is not unmanaged so it can't be used in a `T : unmanaged` constraint. This is why I provided a string-speicifc overloads which were requested for removal (and removed at https://github.com/dotnet/machinelearning/pull/4290/commits/effdc8aefb44862167e72e00996cf9d1307f4dfc) by @eerhardt

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
```diff
@@ -3,49 +3,477 @@
...
+                        }
+#endif
+                        default:
+                            throw new NotSupportedException();
+                    }
+                }
+            }
+        }
+
+        public static void CopyTo<T>(this Tensor tensor, Span<T> destination) where T : unmanaged
```
codemzs: >public static void CopyTo<T>(this Tensor tensor, Span<T> destination) where T : unmanaged [](start = 8, length = 89)

There is a lot of unsafe code in here that does memory copy/manipulation, have you considered using Span<T>.CopyTo(Span) and Span<T>.Slice(int,int) to achieve the same?

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
Nucs: I was sure I have conversion of `String-Tensor -> T`, Adding that one.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
Nucs: I don't see a point, I perform all the checks necessary beforehand. The code is safe. Any use of Span will result in unnecessary overhead.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
Nucs: You are right, changing it to a for loop.
Benchmark: https://pastebin.com/raw/iALC8XAT

Oceania2018: Kidding? Linear always win?

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
eerhardt: If/when we have a caller we can add that support.

As it exists, we could never use this method in ML.NET because we never have `Span<string>` variables. We always code to `ReadOnlyMemory<char>`.

Nucs: In tensorflow, string are stored encoded (usually UTF8) with a header of `ulong length`.
If you are willing to provide me with an example/explanation towards how strings are handled inside ML.NET, I'm willing to write a conversion function from Tensor to ReadOnlyMemory<char> or what ever you use.

src/Microsoft.ML.Dnn/TensorTypeExtensions.cs
Nucs: @codemzs I can replace:
`var sdst = (bool*) Unsafe.AsPointer(ref destination.GetPinnableReference());`
With:
`fixed (byte* sdst = MemoryMarshal.Cast<T, byte>(destination))`

They do a similar logic just with more overhead.

codemzs:
@Nucs Thanks for the change but I'm curious why are we making this change? Where exactly will it be used in the code base? This is an internal class btw.

Nucs:
@codemzs, I assumed at [this](https://github.com/dotnet/machinelearning/pull/4260#issuecomment-537106945) comment that the upvote was an answer to my question there. Should've asked for a more explicit answer 😅.
Assuming conversion between `Tensor of type X` to `T of type Y` is not needed then this PR can be closed. Currently this case throw `NotSupportedException`.
ToScalar has a `unmanaged` constraint but If you need conversion from Tensor to System.String then the implementation is [here](https://github.com/dotnet/machinelearning/blob/1e979f1db79697b73388a7ac2c5ef73349b5ad48/src/Microsoft.ML.Dnn/TensorTypeExtensions.cs#L81).

Eitherway I'll probably add this code to `tensorflow.Tensor`.

PR #4297, opened 2019-10-04
Issue 4120

frank-dong-ms:
Follow up on Issue #4120 

1. Use EditorBrowsable attribute to hide the default constructor from intellisense.
2. Use Obsolete("XX", false) instead of Obsolete("XX", true) to be safe.


frank-dong-ms:
@ericstj is added to the review. #Closed

PR #4300, opened 2019-10-04
Fixes #4299 related to ML.Samples causing problems in ML.Samples.GPU

### LINKED ISSUE #4299: Problems caused by ML.Samples and ML.Samples.GPU sharing the same Program.cs file
### Issue

**What did you do?**: I modified the Program.cs file in the ML.Samples project to run a sample that exists in the Samples.Dynamic.Trainers.Regression namespace.

**What happened?**:
Since ML.Samples and ML.Samples.GPU share the same Program.cs file, my changes in the Program.cs under ML.Samples had effects on the ML.Samples.GPU project.

An error appeared in the Program.cs file inside the ML.Samples.GPU project saying that the namespace Samples.Dynamic didn't have a name 'Trainers' and that I might have been missing an assembly reference. Also, while editing the Program.cs inside ML.Samples I received warning tooltips making reference to the errors in ML.Samples.GPU's Program.cs.

Although this didn't prevent me to run the sample, it was odd to see that error messages. Also, later, when trying to run tests inside Visual Studio, I got a compiler errors because of this, and I couldn't run tests. I didn't use to have these compiler errors before, because ML.Samples.GPU was introduced in a recent commit to the master branch.

**How to solve the problem?**
Even ML.Samples has the correct settings to access the sample I was trying to run, ML.Samples.GPU doesn't. This would happen in other samples as well.

So to simply solve this problem, a new Program.cs file needs to be added to the ML.Samples.GPU project, that is independent to the one in ML.Samples.

antoniovs1029:
A simple change to fix #4299 . Simply added a new Program.cs file to the ML.Samples.GPU, independent of the one in ML.Samples.

### DIFF
--- a/docs/samples/Microsoft.ML.Samples.GPU/Program.cs
+++ b/docs/samples/Microsoft.ML.Samples.GPU/Program.cs
@@ -1,0 +1,29 @@
+﻿using System;
+using System.Reflection;
+using Samples.Dynamic;
+
+namespace Microsoft.ML.Samples
+{
+    public static class Program
+    {
+        public static void Main(string[] args) => RunAll();
+
+        internal static void RunAll()
+        {
+            int samples = 0;
+            foreach (var type in Assembly.GetExecutingAssembly().GetTypes())
+            {
+                var sample = type.GetMethod("Example", BindingFlags.Public | BindingFlags.Static | BindingFlags.FlattenHierarchy);
+
+                if (sample != null)
+                {
+                    Console.WriteLine(type.Name);
+                    sample.Invoke(null, null);
+                    samples++;
+                }
+            }
+
+            Console.WriteLine("Number of samples that ran without any exception: " + samples);
+        }
+    }
+}
--- a/docs/samples/Microsoft.ML.Samples.GPU/Program.cs
+++ b/docs/samples/Microsoft.ML.Samples.GPU/Program.cs
@@ -10,20 +10,7 @@ namespace Microsoft.ML.Samples
 
         internal static void RunAll()
         {
-            int samples = 0;
-            foreach (var type in Assembly.GetExecutingAssembly().GetTypes())
-            {
-                var sample = type.GetMethod("Example", BindingFlags.Public | BindingFlags.Static | BindingFlags.FlattenHierarchy);
-
-                if (sample != null)
-                {
-                    Console.WriteLine(type.Name);
-                    sample.Invoke(null, null);
-                    samples++;
-                }
-            }
-
-            Console.WriteLine("Number of samples that ran without any exception: " + samples);
+            
         }
     }
 }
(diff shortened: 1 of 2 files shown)


docs/samples/Microsoft.ML.Samples.GPU/Program.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Reflection;
+using Samples.Dynamic;
+
+namespace Microsoft.ML.Samples
+{
+    public static class Program
+    {
+        public static void Main(string[] args) => RunAll();
+
+        internal static void RunAll()
```
codemzs: Why is this empty?

antoniovs1029: As I wrote in [the issue](https://github.com/dotnet/machinelearning/issues/4299#issuecomment-538587038), I got runtime errors if I tried to run it with the original content of Program.cs. So I removed that.

Should I put that back into the method?

yaeldekel: What were the errors? Did you run the samples on a machine with a GPU?
If there are errors, then the samples that throw them should be fixed.

---
In reply to: [331726596](https://github.com/dotnet/machinelearning/pull/4300#discussion_r331726596) [](ancestors = 331726596)

antoniovs1029: My machine doesn't have a dedicated GPU. And now I realize that the sample didn't run because I don't have tensorflow installed either.

I will find a machine that has those to run the samples, and if it works then I will return the content of Program.cs.

Thanks for the feedback.

antoniovs1029: So I had checked this on another machine, and the samples run with the original content of the Program.cs file, so I've put that back in here.

So, please, approve this PR. Thanks!

antoniovs1029: So today I got my machine with a GPU running. And I checked that this works as expected; everything went correctly. Please, approve this PR. Thanks!

yaeldekel: Thanks for verifying that the samples run correctly!
Since they are, is this PR still needed? This file is identical to the one in Microsoft.ML.Samples, isn't it?

---
In reply to: [337241467](https://github.com/dotnet/machinelearning/pull/4300#discussion_r337241467) [](ancestors = 337241467)

antoniovs1029: Yes, they are identical. The problem was that both projects shared the same file, and when editing the other file to run a specific sample, then some error tooltips appeared when editing it or when trying to rebuild the solution after making those changes, because ML.Samples.GPU doesn't have access to the same things that ML.Samples.

So these errors didn't appear before, and after talking about it with @codemzs we agreed that I should add another Program.cs file in ML.Samples.GPU with the same content of the one in ML.Samples.

PR #4306, opened 2019-10-07
Fixes #4292 about using PFI with BPT and CMPB

### LINKED ISSUE #4292: Use PFI with Binary Prediction Transformer and CalibratedModelParametersBase loaded from disk
In my last accepted pull request (#4262 ) I addressed issue #3976 and was able to provide working samples and tests for using PFI with models loaded from disk except for the case of Binary Prediction Transformer. Here I open this issue about that specific problem.

### Problem
In the [sample using PFI with binary classification](https://github.com/dotnet/machinelearning/blob/e19369b407e8630cfaa82b07f81f0576c9bbd145/docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/BinaryClassification/PermutationFeatureImportance.cs#L40) the last transformer of the model (i.e. the linearPredictor) is of type `BinaryPredictionTransformer<CalibratedModelParametersBase<LinearBinaryModelParameters, PlattCalibrator>>`.

Problem is that when saving and then loading that model from disk, a null reference is returned when trying to access the last transformer by casting it to the original type.
```
// linearPredictor is null:
var linearPredictor = (loadedmodel as TransformerChain<ITransformer>).LastTransformer as BinaryPredictionTransformer<CalibratedModelParametersBase<LinearBinaryModelParameters, PlattCalibrator>>; 
```

Having a null linearPredictor makes it unusable with PFI.

In version 1.3 of ML.Net the last transformer of the loaded model would actually be of type `BinaryPredictionTransformer<IPredictorProducing<float>>`

With the changes I made in my last PR (which will be available in version 1.4.0 preview 2) the loaded model's last transformer would be of type `BinaryPredictionTransformer<ParameterMixingCalibratedModelParameters<IPredictorProducing<float>, ICalibrator>>` which is a step forward in solving the problem, but is not yet enough.

As stated, in both cases, a cast to the original type would return null. In general, it would be expected that the user tries to make that cast in order to use PFI, failing to accomplish it.

This problem would be solved if the loaded model actually had a lastTransformer of the original type, or something castable to it.

### Workaround
Based on [this comment](https://github.com/dotnet/machinelearning/pull/4262#discussion_r330175774) made by @yaeldekel I've just made [this working sample of using PFI with a binary prediction transformer loaded from disk](https://gist.github.com/antoniovs1029/e5fdd86d5b7c8b6adf34cb5481ee20dd). It is pretty much the same as the original sample, only that it works with a model loaded from disk.

The key of the workaround is that the user should cast the lastTransformer not into a binary prediction transformer but rather into a `ISingleFeaturePredictionTransformer<object>`, and then do a series of casts to get whatever other object s/he may want to get from inside the lastTransformer.

In the sample I've just provided it works pretty much in this way:
```
var linearPredictor = (loadedmodel as TransformerChain<ITransformer>).LastTransformer as ISingleFeaturePredictionTransformer<object>;
var predictorModel = linearPredictor.Model as CalibratedModelParametersBase;
var predictorSubModel = predictorModel.SubModel as LinearBinaryModelParameters;
```

Notice that this workaround worked even in ML.Net 1.3, and also works with the changes that I introduced in 1.4.0 preview 2.

Notice that a similar workaround might help a user that tries to use PFI with any kind of prediction transformer loaded from disk. This would come useful if the user, for whatever reason, can not extract the linearPredictor by casting to the same type used in the original model.

### Cause of the Problem
There are 3 main points that are related to the cause of this problem, all of which pertain the `Calibrator.cs ` file and aren't related to the binary prediction transformer itself:
1. Unexpectedly, when loading a `ParameterMixingCalibratedModelParameters<>` its [Create method](https://github.com/dotnet/machinelearning/blob/e19369b407e8630cfaa82b07f81f0576c9bbd145/src/Microsoft.ML.Data/Prediction/Calibrator.cs#L564) isn't called. I discovered this while debugging, and what actually happens is that, during loading, inside the [CreateInstanceCore method](https://github.com/dotnet/machinelearning/blob/7c067854b564275b0d6387ca59c0ec83e8fc91b9/src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs#L197), it first looks for a constructor, and so it calls the constructor of `ParameterMixingCalibratedModelParameters<>` instead of the Create method.
2. Currently, when loading a `ParameterMixingCalibratedModelParameters<>` model, a `ParameterMixingCalibratedModelParameters<IPredictorProducing<float>, ICalibrator>` is always loaded, no matter what the actual submodel and calibrator are. This doesn't change by fixing point 1). This point is similar to the [original problem found on the prediction transformers](https://github.com/dotnet/machinelearning/issues/3976#issuecomment-517862076), which I fixed in [my last pull request](https://github.com/dotnet/machinelearning/pull/4262); using a similar approach in this case would fix this point... that is, loading first the submodel and calibrator to then create a generic type at runtime with the correct parameter types.
3. When fiting the model (i.e. before even saving it or loading it) the [SdcaLogisticRegressionBinaryTrainer ](https://github.com/dotnet/machinelearning/blob/46ede664e8cff585c450a7005c7069b64df5a6f1/src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs#L1606)creates a predictor of type `ParameterMixingCalibratedModelParameters<LinearBinaryModelParameters, PlattCalibrator>` (which I will now refer to as "PMCMP") but returns it as a `CalibratedModelParametersBase<LinearBinaryModelParameters, PlattCalibrator>` (let's call it "CMPB") this then is what makes the last transformer of the model to be a `BinaryPredictionTransformer<CMPB>` whereas the internal model of the last transformer is actually a PMCMP. When saving it to disk, it's saved as a PMCMP (i.e. it's saved using a LoaderSignature of "PMixCaliPredExec"), so when loading occurs, it calls the constructor of PMCMP but it doesn't cast it to a CMPB. This is different from the problem fixed in my last pull request; there, if a Regression prediction transformer was saved, then we expected to load a regression prediction transformer... whereas in here if a BPT<PMCMP> is saved we actually want to load a PMCMP with the correct type parameters, but actually create a BPT<CMPB> where the CMPB should also have the correct type parameters.

### Trying to solve the problem
So far I've been able to solve problems 1) and 2) described above, but after trying out different approaches I haven't been able to solve problem 3). To solve those problems I've changed different things in the `Calibrator.cs` file. My attempt to solve this problem can be found in PR #4306

With those changes (along with the ones of my last PR), the loaded model's last transformer becomes a `BinaryPredictionTransformer<ParameterMixingCalibratedModelParameters<LinearBinaryModelParameters, PlattCalibrator>>`. Notice that even here a cast to `BPT<CMPB>` would be null, so it doesn't solve the problem. Also notice that since PMCMP is an internal class the user wouldn't be able to cast the last transformer to `BPT<PMCMP>` either, since s/he wouldn't have access to that class.

### Further problems
Here I've explained the specific case of loading a `BPT<CMPB>` with the specific problems that arise in CMPB and PMCMP classes because that is what is used in the sample of PFI with BPT, and in the tests of PFI with BPT. It could be possible that the problems here described are also present in other classes (for example in the other classes of [Calibrator.cs](https://github.com/dotnet/machinelearning/blob/18394c4f9e45b9c5ff41dfbabd30e31132ae4cdc/src/Microsoft.ML.Data/Prediction/Calibrator.cs#L1)) but they might not become a problem unless the user tries to access the last transformer of a model loaded from disk. In such a case the described workaround might help.

antoniovs1029:
Fixes #4292 about using PFI with a `BinaryPredictionTransformer<>` and classes derived from `CalibratedModelParametersBase<>`.

### Approach of the solution
Regarding the 3 problems listed in #4292 the solutions were the following:
1. **To call the correct 'create' method**: as discussed [here](https://github.com/dotnet/machinelearning/pull/4306#discussion_r336139379), I removed the constructor with overload `(IHostEnvironment env, ModelLoadContext ctx)` from the `PMCMP<,> ` class, since its not needed after the changes on this PR. Without that constructor, the 'create' method gets called directly by `CreateInstanceCore` without changing anything else.
2. **To create a `PMCMP` object with the correct parameter types**: I used a similar approach as the one explained in my previous pull request #4262. That is, the calibrator and submodel of the `PMCMP<>` are loaded first, and then they are used to create a generic type at runtime for `PMCMP<>` but now using the correct parameter types.
3. **To create a `BPT<CMPB<>>` object whereas a `PMCMP<>` was loaded**: After trying different approaches that didn't work, and following the suggestions of @eerhardt the best way to fix this was to create a new attribute called `PredictionTransformerLoadTypeAttribute`, that would help the prediction transformer to know what's the correct type to use as a TModel when loading a model from disk. This attribute is applied to the `PMCMP<>` class, and it's used in `PredictionTransformer.cs` to decide the correct type for the BPT.

Also, after discussing it with @eerhardt , I also fixed these 3 problems for the classes that inherit from `CalibratedModelParametersBase<,>`; those are: `ValueMapperCalibratedModelParameters<,>`, and `FeatureWeightsCalibratedModelParameters<,>`. Notice that  `SchemaBindableCalibratedModelParameters<,>` also inherits from CMPB<,>, but, as discussed [here](https://github.com/dotnet/machinelearning/pull/4306#issuecomment-545562898), it was decided not to fix it.

### Changes implemented
* Changes in `PredictionTransformer.cs` and `Calibrator.cs` to implement the approach previously described
* Added a working sample for using PFI with BPT and CMPB while loading a model from disk. This is based entirely in the original sample.
* Added file `CalibratedModelParametersTests.cs ` with tests that the CMPs modified in this PR are now being correctly loaded from disk.
* Changed a couple of tests in `LbfgsTests.cs` that failed because they used casts that now return 'null'. Notice that those casts involved using classes such as `IPredictorProducing ` and `PMCMP `which are internal classes, so a regular user wouldn't have been able to use those casts. They have now been replaced with casts to the public classes  `BPT<CMPB<>>` and `CMPB<>` using the correct type parameters of the model.


src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -769,6 +769,13 @@ internal static object Create(IHostEnvironment env, ModelLoadContext ctx, IHost
 
             return genericInstance;
         }
+
+        internal static Type GetLoadType(Type modelType)
+        {
+            var y = modelType.GetField("LoadType", BindingFlags.NonPublic | BindingFlags.Static);
```
eerhardt: This approach is very easily broken, since people won't know that the unused field you are adding is only being used through reflection.

Instead, I think you should just inspect the type using normal Reflection APIs. Check out these methods

* [GetGenericArguments()](https://docs.microsoft.com/en-us/dotnet/api/system.type.getgenericarguments?view=netframework-4.8)
* [GetGenericTypeDefinition()](https://docs.microsoft.com/en-us/dotnet/api/system.type.getgenerictypedefinition?view=netframework-4.8)

antoniovs1029: Thanks for the feedback. But I am not sure how to follow your suggestion.

Since the problem is that when loading a `PMCMP<TSubmodel, TCalibrator>` inside a `BinaryPredictionTransformer<>`, what actually needs to be loaded is not a `BPT<PMCMP<>>` but rather a `BPT<CMPB<>>`, then are you suggesting something similar to the following pseudocode?:

```
// Inside the CreatePredictionTransformer class
internal static Type GetLoadType(Type modelType)
{
    if(modelType is PMCMP)
    {
         use GetGenericArguments() and GetGenericTypeDefinition() to get the correct
        parameter types for the submodel and calibrator, and return type CMPB with the
        the correct type parameters.
    }
     
     // if the type of the model isn't PMCMP then simply return the original type
     return modelType;
}
```

In short, is your suggestion to hardcode the case of PMCMP directly in the method that is called when loading a BPT, so that it gets managed differently to other cases??

Thanks.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -561,12 +563,50 @@ private ParameterMixingCalibratedModelParameters(IHostEnvironment env, ModelLoad
...
+        // ORIGINAL CREATE METHOD
+        //private static CalibratedModelParametersBase Create(IHostEnvironment env, ModelLoadContext ctx)
+        //{
+        //    Contracts.CheckValue(env, nameof(env));
+        //    env.CheckValue(ctx, nameof(ctx));
+        //    ctx.CheckAtModel(GetVersionInfo());
+        //    return new ParameterMixingCalibratedModelParameters<TSubModel, TCalibrator>(env, ctx);
+        //}
+
+        internal static object Create(IHostEnvironment env, ModelLoadContext ctx)
```
eerhardt: (nit) this should return `CalibratedModelParametersBase`, not `object`

eerhardt: Do we really need this method and `ParameterMixingCalibratedModelParameters.Create` below? Can't this code just move into `ParameterMixingCalibratedModelParameters.Create`?

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -604,6 +644,17 @@ IParameterMixer<float> IParameterMixer<float>.CombineParameters(IList<IParameter
         }
     }
 
+    internal static class ParameterMixingCalibratedModelParameters
+    {
+        internal const string LoaderSignature = "PMixCaliPredExec";
```
eerhardt: Probably don't want to duplicate these strings. The other strings should be updated to these.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -6,6 +6,7 @@
 using System.IO;
 using System.Reflection;
 using Microsoft.ML;
+using Microsoft.ML.Calibrators;
```
eerhardt: This `using` shouldn't be necessary anymore.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -769,6 +770,37 @@ internal static object Create(IHostEnvironment env, ModelLoadContext ctx, IHost
...
+                    return genericType.MakeGenericType(typeArguments);
+                }
+            }
+
+            return modelType;
+        }
+    }
+
+    [AttributeUsage(AttributeTargets.Class)]
+    internal class PredictionTransformerLoadType : Attribute
```
eerhardt: Attribute classes should have the suffix `Attribute`. So this should be changed to `internal class PredictionTransformerLoadTypeAttribute : Attribute`

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -769,6 +770,37 @@ internal static object Create(IHostEnvironment env, ModelLoadContext ctx, IHost
...
+    {
+        internal Type LoadType;
```
eerhardt: Use a property instead of a field.

```C#
internal Type LoadType { get; }
```

src/Microsoft.ML.Data/Prediction/Calibrator.cs
antoniovs1029: Well, problem was that this create method calls `GetPredictor()` and `GetCalibrator()` which are static methods of generic class `CMPB<,>`, and it also calls `GetVersionInfo()` which is static method of generic class `PMCMP<,>`.

But yeah, I actually can get rid of this create method, by calling those other methods through `ParameterMixingCalibratedModelParameters<IPredictorProducing<float>, ICalibrator>` inside the `ParameterMixingCalibratedModelParameters.Create` method. I just think it would look uglier, but it's doable.

So I will do this asap. Thanks for the feedback!

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
antoniovs1029: I will change this asap. I just wanted to ask why is it better to use a property in this case? I wonder so that I can better choose between properties and fields in the future. Thanks!

eerhardt: In general, if you are exposing a value outside of a class, use a property. The benefits are:

1. You can disallow others from changing the value. Fields are settable (like yours above).
2. You can intercept the call, if necessary. For example, validate a value being set is non-null. Or you can lazy-load the property if it isn't already initialized.
3. It provides more flexibility in the future for public APIs. If you ship a public field in a public API you cannot change it to a property later.

In your current code, you are allowing anyone else in the assembly to change this value (whether they intended to or made a mistake).

It is a good habit to get into to always use properties for non-private instance data.

See https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/field for more info.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
antoniovs1029: So I tried to do this as I said, and problem was that GetPredictor(), GetCalibrator() and GetVersionInfo() are all private methods.

There are many ways to refactor this, though... specially since I still have to make changes to the other classes in Calibrator.cs, so I will address this particular comment in my next iteration when I change those other classes.

---
In reply to: [334058843](https://github.com/dotnet/machinelearning/pull/4306#discussion_r334058843) [](ancestors = 334058843)

eerhardt: The reason I proposed the above is because of the following code in `ParameterMixingCalibratedModelParameters.Create`:

```C#
public static CalibratedModelParametersBase Create(IHostEnvironment env, ModelLoadContext ctx)
        {
            return (CalibratedModelParametersBase) ParameterMixingCalibratedModelParameters<IPredictorProducing<float>, ICalibrator>.Create(env, ctx);
        }
```

Here we are still using `IPredictorProducing<float>` and `ICalibrator` for the generic types, but I was thinking we were ignoring those types.  I was wrong because we are using those types in `GetPredictor()` and `GetCalibrator()`, as you mention above. So my question was answered. The two separate methods are still necessary. Thanks for checking.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
antoniovs1029: So I will resolve this conversation, because, as we discussed offline, it's better to use Attributes in this case, and I've already implemented that. Thanks.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -430,8 +432,13 @@ private static VersionInfo GetVersionInfoBulk()
                 loaderAssemblyName: typeof(ValueMapperCalibratedModelParameters<TSubModel, TCalibrator>).Assembly.FullName);
         }
 
-        private ValueMapperCalibratedModelParameters(IHostEnvironment env, ModelLoadContext ctx)
-            : base(env, RegistrationName, GetPredictor(env, ctx), GetCalibrator(env, ctx))
+        //private ValueMapperCalibratedModelParameters(IHostEnvironment env, ModelLoadContext ctx)
+        //    : base(env, RegistrationName, GetPredictor(env, ctx), GetCalibrator(env, ctx))
```
antoniovs1029: @eerhardt @yaeldekel I have a question about this. As I pointed out in point 1) of the problems on issue #4292 , the Create method of ParameterMixingCalibratedModelParameters (PMCMP) wasn't called when loading that model, and instead its constructor with parameters (IHostEnvironment env, ModelLoadContext ctx) was called directly without calling the Create Method. This happens because the [CreateInstanceCoreMethod ](https://github.com/dotnet/machinelearning/blob/7c067854b564275b0d6387ca59c0ec83e8fc91b9/src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs#L197) looks for a constructor before a create method. To solve that I created an static version of `PMCMP` with a create method that calls the create method of generic `PMCMP<,>`

This problem actually happens also in the other `CalibratedModelParameters `classes, but I have just realized that it is actually not necessary to create another class to bypass this. As I show here in `ValueMapperCalibratedModelParameters`, I can solve this problem by actually removing the constructor `VMCMP(env,ctx)` so that `Create(env, ctx)` of VMCMP actually gets called.

Removing this constructor seems safe to me, since it was only referenced by the `Create(env, ctx) ` method that was being ignored anyway, and which I've updated, and now doesn't call that constructor, but calls constructor `VMCMP(env, ctx, predictor, calibrator)`. I've done this also in the `FeatureWeightsCalibratedModelParameters` class. And all the tests pass as expected.

So I just wanted to double check with you if it's safe to remove that constructor (I don't know if it might be, perhaps, called through reflection somewhere else, or something).

If it's safe, then I will also remove it in PMCMP and creating the static PMCMP wouldn't be necessary either.

Thanks.

test/Microsoft.ML.Tests/CalibratedModelParametersTests.cs
```diff
@@ -0,0 +1,163 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
antoniovs1029: @eerhardt @yaeldekel I didn't know where to put these tests, so I created this file. If you think they should go elsewhere, please instruct me to do so.

Btw, I thought of putting them in the [CalibratorEstimators.cs](https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Tests/TrainerEstimators/CalibratorEstimators.cs) tests, but it seems to me that file is more about testing the calibrator classes, whereas in here I am focusing in the CalibratedModelParameters classes.

test/Microsoft.ML.Tests/CalibratedModelParametersTests.cs
```diff
@@ -0,0 +1,163 @@
...
+            var castedModel = loadedModel as BinaryPredictionTransformer<CalibratedModelParametersBase<FastTreeBinaryModelParameters, PlattCalibrator>>;
+
+            Assert.NotNull(castedModel);
+
+            Type expectedInternalType = typeof(FeatureWeightsCalibratedModelParameters<FastTreeBinaryModelParameters, PlattCalibrator>);
+            Assert.Equal(expectedInternalType, castedModel.Model.GetType());
+            Assert.Equal(model.Model.GetType(), castedModel.Model.GetType());
+            Done();
+        }
+
```
antoniovs1029: @yaeldekel I still need to add a test for `SchemaBindableCalibratedModelParameters` similar to the others. I haven't found how to make a model that actually uses that class. I am working on this, and I hope to find one ASAP. But if you happen to know how to do this, I would greatly appreciate your advice. Thanks.

antoniovs1029: So I've been trying to find a way to create a model that actually uses SBCMP, but so far I've been unsuccessful. If someone could tell me how to do it, it would be fantastic.

Still, I've noticed something quiet odd about how SBCMP is used in the existing codebase, and I would love if someone could clarify this to me. (perhaps, @eerhardt @yaeldekel  ? ).

If I look for the references of SBCMP outside the `Calibrator.cs` file, there are only four references. **They're all on the `Create` methods of the following classes, where the method returns a new SBMCP object**:

1. [LinearBinaryModelParameters](https://github.com/dotnet/machinelearning/blob/d531ea801a34a3018b8fa2a2f352902eb703cd25/src/Microsoft.ML.StandardTrainers/Standard/LinearModelParameters.cs#L485)

2. [GamBinaryModelParameters](https://github.com/dotnet/machinelearning/blob/d531ea801a34a3018b8fa2a2f352902eb703cd25/src/Microsoft.ML.FastTree/GamClassification.cs#L244)

3. [FastTreeBinaryModelParameters](https://github.com/dotnet/machinelearning/blob/d531ea801a34a3018b8fa2a2f352902eb703cd25/src/Microsoft.ML.FastTree/FastTreeClassification.cs#L97)

4. [FastForestBinaryModelParameters](https://github.com/dotnet/machinelearning/blob/d531ea801a34a3018b8fa2a2f352902eb703cd25/src/Microsoft.ML.FastTree/RandomForestClassification.cs#L112)

So below I write the 3 main questions that I've been asking myself while looking at all of the above references.

**If I get a working example of a model that actually uses SBCMP I might be able to answer myself question 2 and 3, but I might still want to discuss question 1, since it seems to be present in several different classes...**:

### Question 1: Is it right that these create methods aren't being called when loading?
As a matter of fact, when loading those 4 parameter classes I've listed, they don't even call their Create(env, ctx) method, but rather they call their constructor overload with (env, ctx) parameters. This situation, of ignoring the Create method, and going directly to the constructor, is something I've noticed in different classes, and I've mentioned it inside other comments of this PR and in issue #4292 point 1) of the "Problems" subheader.

**So I do wonder if I am missing something important, and maybe this is how it is supposed to work (for these classes, and also the other classes where I've seen this behavior)??**.

Notice that these 4 create methods I've listed have "0 references" when checking it with VisualStudio's CodeLense tool, so it seems to me that they are not used directly anywhere in the code... and so I am unable to find a working example of where in the codebase a SBCMP object gets actually instatiated.

### Question 2: Why are these create methods returning a SBCMP object? Are those BinaryModelParameters supposed to be loaded as SBCMP objects?
_If question 1 has a negative answer (i.e. if those create methods are supposed to be called when loading the BinaryModelParameters classes) then I have this question._

So far I've been assuming that if a class X has a 'Create(env, ctx)' method, it's because it is going to be called when loading an X object. For example, the create method of BinaryPredictionTransformer loads and returns a BinaryPredictionTransformer object. So, when loading a `LinearBinaryModelParameter` am I supposed to expect a SBCMP object?.


### Question 3: Is SBCMP supposed to work differently from the other CMPs classes?
_If question 2 has a positive answer (i.e. those create methods are supposed to be called and return a SBCMP object) then I have this question._

I don't really know how SBCMP is supposed to work, or be used, but I assume that it should be similar to the other classes that also inherit from `CalibratedModelParametersBase<,>`.

So it is also strange to me that those other classes are instantiated when fitting/training _trainer classes_ (e.g. PMCMP on [ScaLogisticRegressionBinaryTrainer](https://github.com/dotnet/machinelearning/blob/46ede664e8cff585c450a7005c7069b64df5a6f1/src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs#L1610) class, VMCMP on [GamBinaryTrainer ](https://github.com/dotnet/machinelearning/blob/d531ea801a34a3018b8fa2a2f352902eb703cd25/src/Microsoft.ML.FastTree/GamClassification.cs#L142)class, and FWCMP on [FastTreeBinaryTrainer](https://github.com/dotnet/machinelearning/blob/d531ea801a34a3018b8fa2a2f352902eb703cd25/src/Microsoft.ML.FastTree/FastTreeClassification.cs#L212)) whereas SBCMP only have references inside _parameter_ classes. 

Notice that the first 3 parameter classes I've listed are actually used as the predictors (submodels) of PMCMP, VMCMP and FWCMP respectively inside the tests I've added in this PR... making this whole thing somewhat more confusing to me, because if question 2 has an affirmative answer, then it would mean that SBCMP could be used as the predictor of PMCMP/VMCMP/FWCMP, right? Which doesn't make sense to me so far...

src/Microsoft.ML.Data/Prediction/Calibrator.cs
eerhardt: My understanding is that it is safe to remove the constructor as well - because we have the `Create` method instead in its place. I _think_ the only place that looks for it is in the `ComponentCatalog`. Doing a quick search for `ConstructorInfo` across the code, that appears to be the case.

So I would go ahead and remove the constructor, and use the static `Create` method instead.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -443,7 +450,16 @@ private static CalibratedModelParametersBase Create(IHostEnvironment env, ModelL
             var ver2 = GetVersionInfoBulk();
             var ver = ctx.Header.ModelSignature == ver2.ModelSignature ? ver2 : ver1;
             ctx.CheckAtModel(ver);
-            return new ValueMapperCalibratedModelParameters<TSubModel, TCalibrator>(env, ctx);
+
```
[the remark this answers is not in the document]
eerhardt: I guess since they are identical model formats, my thought is that the only difference is the string. So it probably isn't worth it to try to test the `BCALPRED` path. You wouldn't be able to make one of these models - so you would have to somehow find an old one that was made this way. It probably isn't worth it.

test/Microsoft.ML.Tests/CalibratedModelParametersTests.cs
eerhardt: I think adding a new test file is reasonable.

test/Microsoft.ML.Tests/CalibratedModelParametersTests.cs
```diff
@@ -0,0 +1,163 @@
...
+        #region Helpers
```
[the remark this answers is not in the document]
eerhardt: It's only a little bit of duplication, in test code. At least this way the methods can evolve separately as needed for the different tests. I think what you have here is fine.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
antoniovs1029: Ok. So I've deleted the constructor(env,ctx) from PMCMP, VMCMP and FWCMP. I've also deleted the non-generic PMCMP class that I had previously added on this PR, since it's no longer necessary.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -553,20 +576,29 @@ private static VersionInfo GetVersionInfo()
...
+        private ParameterMixingCalibratedModelParameters(IHostEnvironment env, ModelLoadContext ctx, TSubModel predictor, TCalibrator calibrator)
+            : base(env, RegistrationName, predictor, calibrator)
         {
             Host.Check(SubModel is IParameterMixer<float>, "Predictor does not implement " + nameof(IParameterMixer));
             Host.Check(SubModel is IPredictorWithFeatureWeights<float>, "Predictor does not implement " + nameof(IPredictorWithFeatureWeights<float>));
             _featureWeights = SubModel as IPredictorWithFeatureWeights<float>;
         }
 
-        private static CalibratedModelParametersBase Create(IHostEnvironment env, ModelLoadContext ctx)
+        internal static CalibratedModelParametersBase Create(IHostEnvironment env, ModelLoadContext ctx)
```
yaeldekel: >internal [](start = 8, length = 8)

Does this need to be internal?

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -487,8 +499,9 @@ private static VersionInfo GetVersionInfo()
                 loaderAssemblyName: typeof(FeatureWeightsCalibratedModelParameters<TSubModel, TCalibrator>).Assembly.FullName);
         }
 
-        private FeatureWeightsCalibratedModelParameters(IHostEnvironment env, ModelLoadContext ctx)
-            : base(env, RegistrationName, GetPredictor(env, ctx), GetCalibrator(env, ctx))
+        internal FeatureWeightsCalibratedModelParameters(IHostEnvironment env, ModelLoadContext ctx,
```
yaeldekel: >internal [](start = 8, length = 8)

Does this need to be internal?

test/Microsoft.ML.Tests/CalibratedModelParametersTests.cs
yaeldekel: Thanks for the thorough investigation! 

Answer 1: the reason the create method isn't being called, as you mentioned in the other comment, is because the `ComponentCatalog` is searching first for a constructor, and then for a `Create` method. Since the constructor and the `Create` methods both have the signature that `ComponentCatalog` is looking for, it is ignoring the `Create` method. I think this duplication is happening because a long time ago, and before the official release of ML.NET, `ComponentCatalog` only searched for public constructors/methods, so the constructor (which is typically private) was not found, and the `Create` method was called. Since we tried to internalize as many APIs as possible, and these `Create` methods didn't need to be part of the public API, `ComponentCatalog` was modified to search for internal and private members as well. From looking at the four `Create` methods you mentioned, this is what they do before actually calling the constructor:
```
            Contracts.CheckValue(env, nameof(env));
            env.CheckValue(ctx, nameof(ctx));
            ctx.CheckAtModel(GetVersionInfo());
```
I think these checks are worth keeping, but not sure how we should solve this - we can either change the order that `ComponentCatalog` searches to look for the `Create` method first, or change the signature of the constructor so that `ComponentCatalog` doesn't find it (change the order of the parameters?). @eerhardt, what do you think?

Answer 2: The `Create` methods that you mentioned seem very wrong to me. When a `FastTree` (or any of the others in your list) model is created from the `Fit()` method (like [here](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.FastTree/FastTreeClassification.cs#L183)), the model that is returned is not a `FastTreeBinaryModelParameters`, it is a `FeatureWeightsCalibratedModelParameters<...>`. So, when it is serialized, it is serialized as a `FeatureWeightsCalibratedModelParameters<,>`, and this is also how it is deserialized. At deserialization time the `FeatureWeightsCalibratedModelParameters<,>` constructor takes care of creating the predictor, not the other way around, so as far as I can tell, [this](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.FastTree/FastTreeClassification.cs#L95) line should always return true.
I tried looking at the history of these files as far as I could find, but I couldn't find where this code was introduced. I can only guess, that this is because years ago, the calibrators used to be a field of the predictor, so the predictors were responsible for loading them at deserialization time, and when calibrators were extracted out of the predictors, this was not handled properly for these four cases.

Answer 3: The SBCMP exists so that we can support binary predictors that are not `IValueMapper`s. I.e., predictors that use input that is not necessarily a single numeric vector of features. It was introduced when the `PipelineEnsemble` class was introduced, which is a predictor that can be created only from an entrypoint, and it doesn't have an estimator that creates it, or a public API. The only other predictor that I know of in ML.NET that is not an `IValueMapper` is `FieldAwareFactorizationMachineModelParameters`, which has its own `PredictionTransformer` and doesn't use the `BinaryPredictionTransformer`, for the exact reason that `BinaryPredictionTransformer` derives from `SingleFeaturePredictionTransformerBase`. So the bottom line is that there should never exist a `BinaryPredictionTransformer` where the type of model is `SchemaBindableCalibratedModelParameters`. You can still add a test that tests saving and loading of a `SchemaBindableCalibratedModelParameters`, you can see an example of how to create one in the test `EntryPointPipelineEnsemble`. The variable [binaryEnsembleCalibrated](https://github.com/dotnet/machinelearning/blob/master/test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs#L1582) is a `PredictorModel` with a `Predictor` of type `SchemaBindableCalibratedModelParameters`.


---
In reply to: [336265838](https://github.com/dotnet/machinelearning/pull/4306#discussion_r336265838) [](ancestors = 336265838)

antoniovs1029:
So, as discussed offline, I will open other separate issues to address the problems mentioned on Question 1 and Question 2 of [this comment](https://github.com/dotnet/machinelearning/pull/4306#discussion_r336265838)

Regarding Question 3 on that comment, and [the answers](https://github.com/dotnet/machinelearning/pull/4306#discussion_r337981820) that @yaeldekel provided, as well as suggestions made by @eerhardt offline, it will be better to not make any change on the `SchemaBindableCalibratedModelParameters<,>` class, since it is not used in the same way as the other CMPs (i.e. it's not used inside binary prediction transformer), and it is used in ways that might not present the same problem as the other CMPs (i.e. that they're loaded from disk using incorrect types for its internal predictor and calibrator). If in the future SBCMP presents this problem, then it might be appropriate to change it in a similar manner as the other CMPs are being modified in this PR.

So there are some minor fixes that I will want to make on this PR, and after that I will remove the Draft status, and I would appreciate for people to review this and approve it.Thanks.

test/Microsoft.ML.Tests/CalibratedModelParametersTests.cs
antoniovs1029: Thank you very much for all your feedback.

Since we've also discussed this offline, I'm marking this conversation as resolved. I've also written down our conclusions of the offline discussions [here](https://github.com/dotnet/machinelearning/pull/4306#issuecomment-545562898).

Thanks.

antoniovs1029: Thank you very much for all your feedback.

Since we've also discussed this offline, I'm marking this conversation as resolved. I've also written down our conclusions of the offline discussions [here](https://github.com/dotnet/machinelearning/pull/4306#issuecomment-545562898).

Thanks.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
antoniovs1029: No. This was a mistake. I've fixed this.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
antoniovs1029: No. This was a mistake. I've fixed this.

yaeldekel:
Thank you! Can you please add a link to these new issues in this PR?

---
In reply to: [545562898](https://github.com/dotnet/machinelearning/pull/4306#issuecomment-545562898) [](ancestors = 545562898)

antoniovs1029:
> Thank you! Can you please add a link to these new issues in this PR?

@yaeldekel Yes, I've opened the following issues and made a reference there to this PR:
- Issue #4381 
- Issue #4385

Issue #4195, opened 2019-09-09
[DatabaseLoader] Error when using attributes (i.e ColumnName)


luisquintanilla:
### System information

- **OS version/distro**: Windows 10 18362
- **.NET Version (eg., dotnet --info)**:  .NET Core 2.2

### Issue

- **What did you do?**

Tried to use `ColumnName` attribute in class that defines IDataView schema.  

- **What happened?**

When operating on the IDataView, I received the following error

```text
System.Reflection.TargetInvocationException: 'Exception has been thrown by the target of an invocation.'
 
Inner Exception
IndexOutOfRangeException: Features
```

- **What did you expect?**

No error.

### Source code / logs

Given the following data stored in a SQL Server DB

![image](https://user-images.githubusercontent.com/46974588/64536954-fb7c0c80-d2e7-11e9-8a5b-531ddd267aa8.png)

The data schema is defined as such

```csharp
public class HouseData
{
    public float Size { get; set; }

    public float Price { get; set; }
}
```

The following code works:

```csharp
string connectionString = @"Connection-String";

string sqlCommand = "SELECT Size,Price FROM House";

MLContext mlContext = new MLContext();

DatabaseLoader loader = mlContext.Data.CreateDatabaseLoader<HouseData>();

DatabaseSource dbSource = new DatabaseSource(SqlClientFactory.Instance,connectionString,sqlCommand);

IDataView data = loader.Load(dbSource);

// Test Code
IEnumerable<HouseData> housingData = mlContext.Data.CreateEnumerable<HouseData>(data, reuseRowObject: true);

foreach (HouseData house in housingData)
{
    Console.WriteLine($"{house.Size} costs {house.Price}");
}

Console.ReadKey();
```

However, when attributes are added to the schema class, it produces an error.

```csharp
public class HouseData
{
    [ColumnName("Features")]
    public float Size { get; set; }

    [ColumnName("Label")]
    public float Price { get; set; }
}
```

CESARDELATORRE:
Agree. This needs to be fixed.
Basically, the point is, when using a file, the source of truth is the index in LoadColumnAttribute, whereas when using a database the source of truth should be the C# property name that has to match the database column. Then, in order to be consistent across both scenarios, the ColumnNameAttribute allows to change the IDataView schema and change any column name for instance from "HousePrice" to "Label".
​
I think it should be that way so we are consistent with how the ColumnNameAttribute works when using files:

[...]

PR #4308, opened 2019-10-07
Adding a LoadColumnNameAttribute

tannergooding:
This resolves https://github.com/dotnet/machinelearning/issues/4195

### DIFF
--- a/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
+++ b/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
@@ -237,7 +237,7 @@ namespace Microsoft.ML.Data
             public DbType Type = DbType.Single;
 
             /// <summary>
-            /// Source index range(s) of the column.
+            /// Source index or name range(s) of the column.
             /// </summary>
             [Argument(ArgumentType.Multiple, HelpText = "Source index range(s) of the column", ShortName = "src")]
             public Range[] Source;
@@ -250,7 +250,7 @@ namespace Microsoft.ML.Data
         }
 
         /// <summary>
-        /// Specifies the range of indices of input columns that should be mapped to an output column.
+        /// Specifies the range of indices or names of input columns that should be mapped to an output column.
         /// </summary>
         public sealed class Range
         {
@@ -265,6 +265,7 @@ namespace Microsoft.ML.Data
                 Contracts.CheckParam(index >= 0, nameof(index), "Must be non-negative");
                 Min = index;
                 Max = index;
+                Name = null;
             }
 
             /// <summary>
@@ -274,6 +275,8 @@ namespace Microsoft.ML.Data
             public Range(string name)
             {
                 Contracts.CheckValue(name, nameof(name));
+                Min = -1;
+                Max = -1;
                 Name = name;
             }
 
@@ -297,15 +300,27 @@ namespace Microsoft.ML.Data
             /// <summary>
             ///  The minimum index of the column, inclusive.
             /// </summary>
+            /// <remarks>
+            /// This is <c>-1</c> if the range represents a column name.
+            /// </remarks>
             [Argument(ArgumentType.Required, HelpText = "First index in the range")]
             public int Min;
 
             /// <summary>
             /// The maximum index of the column, inclusive.
             /// </summary>
+            /// <remarks>
+            /// This is <c>-1</c> if the range represents a column name.
+            /// </remarks>
             [Argument(ArgumentType.AtMostOnce, HelpText = "Last index in the range")]
             public int Max;
 
+            /// <summary>
+            /// The name of the input column.
+            /// </summary>
+            /// <remarks>
+            /// This is <c>null</c> if the range represents an index.
+            /// </remarks>
             [Argument(ArgumentType.AtMostOnce, HelpText = "Name of the column")]
             public string Name;
 
@@ -358,8 +373,8 @@ namespace Microsoft.ML.Data
             {
                 Contracts.Assert(name != null);
                 Name = name;
-                Min = 0;
-                Lim = 0;
+                Min = -1;
+                Lim = -1;
                 ForceVector = forceVector;
             }
         }
@@ -535,7 +550,7 @@ namespace Microsoft.ML.Data
                 //     ulong: count for key range
                 //   int: number of segments
                 //   foreach segment:
-                //     int: name
+                //     string id: name
                 //     int: min
                 //     int: lim
                 //     byte: force vector (verWrittenCur: verIsVectorSupported)
@@ -610,7 +625,7 @@ namespace Microsoft.ML.Data
                 //     ulong: count for key range
                 //   int: number of segments
                 //   foreach segment:
-                //     int: name
+                //     string id: name
                 //     int: min
                 //     int: lim
                 //     byte: force vector (verWrittenCur: verIsVectorSupported)
--- a/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
+++ b/src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
@@ -125,12 +125,21 @@ namespace Microsoft.ML.Data
                 var column = new Column();
                 column.Name = mappingAttrName?.Name ?? memberInfo.Name;
 
-                var mappingAttr = memberInfo.GetCustomAttribute<LoadColumnAttribute>();
+                var indexMappingAttr = memberInfo.GetCustomAttribute<LoadColumnAttribute>();
+                var nameMappingAttr = memberInfo.GetCustomAttribute<LoadColumnNameAttribute>();
 
-                if (mappingAttr is object)
+                if (indexMappingAttr is object)
                 {
-                    var sources = mappingAttr.Sources.Select((source) => Range.FromTextLoaderRange(source)).ToArray();
-                    column.Source = sources;
+                    if (nameMappingAttr is object)
+                    {
+                        throw Contracts.Except($"Cannot specify both {nameof(LoadColumnAttribute)} and {nameof(LoadColumnNameAttribute)}");
+                    }
+
+                    column.Source = indexMappingAttr.Sources.Select((source) => Range.FromTextLoaderRange(source)).ToArray();
+                }
+                else if (nameMappingAttr is object)
+                {
+                    column.Source = nameMappingAttr.Sources.Select((source) => new Range(source)).ToArray();
                 }
 
                 InternalDataKind dk;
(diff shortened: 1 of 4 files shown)


tannergooding:
CC. @eerhardt, @CESARDELATORRE, @luisquintanilla

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
```diff
@@ -202,6 +305,19 @@ public class IrisVectorData
             public float[] PetalInfo;
         }
 
+        public class IrisVectorDataWithLoadColumnName
+        {
+            public int Label;
+
+            [LoadColumnName("SepalLength", "SepalWidth")]
```
CESARDELATORRE: Not sure I understand this. What if you have hundreds of columns to be loaded as a single vector. Do you need to specify all the column names? Can't you specify a column index range?

tannergooding: What would it mean to have a name and range? Why not just use the more efficient index based LoadColumn attribute at that point?

CESARDELATORRE: Sorry, that's what I mean, to use the LoadColumn attribute for those cases. I was thinking about the `LoadColumn` attribute. I didn't realize that here you are using `LoadColumnName`, not the same attribute, ok. ;)

src/Microsoft.ML.Experimental/DataLoadSave/Database/LoadColumnNameAttribute.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// Maps member to set of fields in database.
+        /// </summary>
+        /// <param name="fieldNames">Distinct database field names to load as part of this column.</param>
+        public LoadColumnNameAttribute(params string[] fieldNames)
+        {
+            Sources = new List<string>(fieldNames);
+        }
+
+        [BestFriend]
+        internal List<string> Sources;
```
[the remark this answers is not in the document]
tannergooding: I just followed what `LoadColumnAttribute` was doing here. I'm fine with changing.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
```diff
@@ -490,6 +535,7 @@ public Bindings(ModelLoadContext ctx, DatabaseLoader parent)
                 //     ulong: count for key range
                 //   int: number of segments
                 //   foreach segment:
+                //     int: name
```
eerhardt: `int` => `string`

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
```diff
@@ -368,19 +401,23 @@ public static ColInfo Create(string name, PrimitiveDataViewType itemType, Segmen
                 if (segs != null)
                 {
                     var order = Utils.GetIdentityPermutation(segs.Length);
-                    Array.Sort(order, (x, y) => segs[x].Min.CompareTo(segs[y].Min));
 
-                    // Check that the segments are disjoint.
-                    for (int i = 1; i < order.Length; i++)
+                    if (segs[0].Name is null)
```
eerhardt: I don't understand this check. Why is `segs[0]` special here?

Also- what happens if `segs` is non-null, but is empty?

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
```diff
@@ -318,17 +340,28 @@ public sealed class Options
...
                 Min = min;
                 Lim = lim;
                 ForceVector = forceVector;
             }
+
+            public Segment(string name, bool forceVector)
+            {
+                Contracts.Assert(name != null);
+                Name = name;
+                Min = 0;
```
eerhardt: Maybe `-1` would be better for these? Since that is an invalid value. Rather than using `0`, which could possibly work.

eerhardt:
>             /// Source index range(s) of the column.

This comment is now a bit out-of-date.

`Source index range(s) of the column.`

`Range` can now be more than just `index` ranges.

---
Refers to: src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs:240 in 90316ec. [](commit_id = 90316ec3518890eeeaedab10f6ed1ad1b2b78853, deletion_comment = False)

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
```diff
@@ -287,6 +306,9 @@ public Range(int min, int max)
             [Argument(ArgumentType.AtMostOnce, HelpText = "Last index in the range")]
             public int Max;
 
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Name of the column")]
+            public string Name;
```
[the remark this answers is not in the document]
eerhardt: Maybe `Name` is OK, since the XML class doc for Range says it represents input columns. Maybe if we just update the help text here to be `Name of the input column`?

---
In reply to: [332550925](https://github.com/dotnet/machinelearning/pull/4308#discussion_r332550925) [](ancestors = 332550925)

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoaderCursor.cs
```diff
@@ -441,9 +465,17 @@ private ValueGetter<VBuffer<double>> CreateVBufferDoubleGetterDelegate(ColInfo c
 
                     foreach (var seg in segs)
                     {
-                        for (int columnIndex = seg.Min; columnIndex < seg.Lim; columnIndex++, i++)
+                        if (seg.Name is null)
+                        {
+                            for (int columnIndex = seg.Min; columnIndex < seg.Lim; columnIndex++, i++)
+                            {
+                                editor.Values[i] = DataReader.IsDBNull(columnIndex) ? default : DataReader.GetDouble(columnIndex);
```
eerhardt: Needs to stay `double.NaN`.

Same for the `else` code below.

Maybe we should write a test to ensure we don't break this and `float`.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
tannergooding: I think it should be `int: id of name`, to match what it is doing above: https://github.com/dotnet/machinelearning/pull/4308/files/90316ec3518890eeeaedab10f6ed1ad1b2b78853#diff-31c601a2df378d0ae48fc3007c11c5edR531

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
```diff
@@ -189,6 +277,21 @@ public class IrisData
             public float PetalWidth;
         }
 
+        public class IrisDataWithLoadColumnName
+        {
+            [LoadColumnName("Label")]
```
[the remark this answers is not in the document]
tannergooding: Changed.

src/Microsoft.ML.Experimental/DataLoadSave/Database/LoadColumnNameAttribute.cs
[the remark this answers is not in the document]
tannergooding: Changed.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
tannergooding: I've changed to `string id: name`

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
tannergooding: We don't support mixing index based and name based segments, so if the first segment has a name, all of them must be names.

As for `non-null`, but empty, is that supported; why wouldn't we just normalize that to `null`?

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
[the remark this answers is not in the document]
tannergooding: Changed.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoaderCursor.cs
tannergooding: Fixed.

Adding tests will require a new database with null column types.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
eerhardt: > We don't support mixing index based and name based segments, so if the first segment has a name, all of them must be names.

I think that would be a good comment to stick in here. It isn't immediately obvious to readers.

> As for non-null, but empty, is that supported

Walking backward through the code, I see that Column.Source is a publicly settable array. An external caller could set that to an empty array. In the `Bindings` constructor, if Column.Source is an empty array, it would cause `segs` here to be an empty array.

---
In reply to: [332593763](https://github.com/dotnet/machinelearning/pull/4308#discussion_r332593763) [](ancestors = 332593763)

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
```diff
@@ -278,15 +300,30 @@ public Range(int min, int max)
             /// <summary>
             ///  The minimum index of the column, inclusive.
             /// </summary>
+            /// <remarks>
+            /// This is <c>-1</c> if the range represents a column name.
```
eerhardt: Since these fields are publicly settable, this doesn't always hold true. The following code is possible:

```C#
Range r = new Range();
r.Name = "foo";
r.Min = 3;
```

Maybe we should somehow relay in the XML docs that `Name` will override `Min` and `Max`.

src/Microsoft.ML.Experimental/DataLoadSave/Database/LoadColumnNameAttribute.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// Maps member to set of fields in database.
+        /// </summary>
+        /// <param name="fieldNames">Distinct database field names to load as part of this column.</param>
+        public LoadColumnNameAttribute(params string[] fieldNames)
+        {
+            Sources = new List<string>(fieldNames);
+        }
+
+        [BestFriend]
+        internal IReadOnlyList<string> Sources;
```
eerhardt: This is still using an `internal`ly settable field. Can we use a readonly property instead?

tannergooding: Fixed.

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
tannergooding: Changed to:
```
/// This value is ignored if <see cref="Name" /> is not <c>null</c>.
```

and
```
/// This value, if non-<c>null</c>, overrides <see cref="Min" /> and <see cref="Max" />.
```

src/Microsoft.ML.Experimental/DataLoadSave/Database/DatabaseLoader.cs
tannergooding: Updated to handle an empty segment.

PR #4321, opened 2019-10-09
Addresses #4226 . Fixes problem when loading NormalizerTransformer from disk.

### LINKED ISSUE #4226: `predictionEngine` breaks after saving/loading a Model
### System information

- **win 10**:
- **1.3.1**: 

I was trying to create a PredictEngine using a saved model. I found out that if I directly use the `ITransformer` retrieve from `Pipeline.Fit`, the `CreatePredictionEngine` works well. But after I save/reload it, then it will give the following error
image

The code for the pipeline is like this

```
public static IEstimator<ITransformer> BuildTrainingPipeline(MLContext mlContext)
        {
            // Data process configuration with pipeline data transformations 
            var dataProcessPipeline = mlContext.Transforms.Conversion.MapValueToKey("Label", "Label")
                                      .Append(mlContext.Transforms.LoadImages("ImagePath_featurized", @"C:\Users\xiaoyuz\Desktop\machinelearning-samples\datasets\images", "ImagePath"))
                                      .Append(mlContext.Transforms.ResizeImages("ImagePath_featurized", 224, 224, "ImagePath_featurized"))
                                      .Append(mlContext.Transforms.ExtractPixels("ImagePath_featurized", "ImagePath_featurized"))
                                      .Append(mlContext.Transforms.DnnFeaturizeImage("ImagePath_featurized", m => m.ModelSelector.ResNet18(mlContext, m.OutputColumn, m.InputColumn), "ImagePath_featurized"))
                                      .Append(mlContext.Transforms.Concatenate("Features", new[] { "ImagePath_featurized" }))
                                      .Append(mlContext.Transforms.NormalizeMinMax("Features", "Features"))
                                      .AppendCacheCheckpoint(mlContext);
            // Set the training algorithm 
            var trainer = mlContext.MulticlassClassification.Trainers.OneVersusAll(mlContext.BinaryClassification.Trainers.AveragedPerceptron(labelColumnName: "Label", numberOfIterations: 10, featureColumnName: "Features"), labelColumnName: "Label")
                                      .Append(mlContext.Transforms.Conversion.MapKeyToValue("PredictedLabel", "PredictedLabel"));
            var trainingPipeline = dataProcessPipeline.Append(trainer);

            return trainingPipeline;
        }
```

And `ModelInput` and `ModelOutput` class is like this
```
    public class ModelInput
    {
        [ColumnName("Label"), LoadColumn(0)]


        [ColumnName("Title"), LoadColumn(1)]
        public string Title { get; set; }


        [ColumnName("Url"), LoadColumn(2)]
        public string Url { get; set; }


        [ColumnName("ImagePath"), LoadColumn(3)]
        public string ImagePath { get; set; }


    }
```

```
public class ModelOutput
    {
        // ColumnName attribute is used to change the column name from
        // its default value, which is the name of the field.
        [ColumnName("PredictedLabel")]
        public String Prediction { get; set; }
        public float[] Score { get; set; }
    }
```

It's really wield though. And my description may not be that detailed. If you need further information, please let me know

antoniovs1029:
Addresses and actually solves issue #4226 by fixing [the problem with loading a NormalizerTransformer from disk](https://github.com/dotnet/machinelearning/issues/4226#issuecomment-532902543), where the NormalizerTransformer is supposed to work with multidimensional vectors as input.

Although this PR actually fixes that issue, another problem [is described in one comment](https://github.com/dotnet/machinelearning/issues/4226#issuecomment-533772062) of the issue. This other problem is related to Resnet18, and is independent of the problem here fixed. Solving any of these two problems actually solve this issue, but perhaps the issue shall remain open until both problems are fixed.

### The solution
- I changed the way NormalizerTransformers are saved and loaded, so that it included the information of the dimensions of the vector it has as input. This changes maintain backward compatibility in that it is still possible to load models that were saved with previous binary format . But notice that the original problem described in the issue will persist for models that were saved before the changes made in this PR... it would be necessary to save the model with the updated save method in order to avoid that specific problem when working with multidimensional vectors.

- I added a 3 test cases: one reproducing a very similar problem as in the original issue, another that tests a simplified version of the original problem, and another that checks that backward compatibility is still maintained.


src/Microsoft.ML.Data/Transforms/Normalizer.cs
```diff
@@ -387,32 +388,70 @@ internal static DataViewType LoadType(ModelLoadContext ctx)
...
-                //   - int: vector size
-                //   - byte: ItemKind of input column (only R4 and R8 are valid)
                 bool isVector = ctx.Reader.ReadBoolean();
-                int vectorSize = ctx.Reader.ReadInt32();
-                Contracts.CheckDecode(vectorSize >= 0);
-                Contracts.CheckDecode(vectorSize > 0 || !isVector);
+
+                if (ctx.Header.ModelVerWritten < 0x00010002)
+                {
+                    // *** Previous Binary format ***
```
eerhardt: Generally, it is good to add a test case for the "previous" format, so we know it still loads correctly.

In the past, we've added saved models here: https://github.com/dotnet/machinelearning-testdata/tree/master/Microsoft.ML.TestModels/TestModels

test/Microsoft.ML.Tests/Transformers/NormalizerTests.cs
```diff
@@ -886,5 +887,50 @@ void TestNormalizeLogMeanVarianceFixZeroVec()
...
+
+                return new TensorData[]
+                {
+                    new TensorData() { input = tensor1},
+                    new TensorData() { input = tensor2}
+                };
+            }
+        }
+
+        [Fact]
```
eerhardt: When fixing a bug, it is generally useful to add the repro scenario as a test case. That way it ensures the full scenario is actually fixed, and won't regress again in the future.

antoniovs1029: So, do you mean that I should remove this test case and replace it with the problem described in the original issue (including a small dataset of images, and using the resnet18)?

src/Microsoft.ML.Data/Transforms/Normalizer.cs
antoniovs1029: Good idea. Thanks for the feedback.

There's [another test](https://github.com/dotnet/machinelearning/blob/d3dca2e952f25e5d445dda0383d2b9765bf6625b/test/Microsoft.ML.Tests/Transformers/NormalizerTests.cs#L778) that checks for backward compatibility with an even older version of the NormalizerTransformer. I noticed that in that test they actually have the model with the 'previous older' format inside this other folder: https://github.com/dotnet/machinelearning/tree/master/test/data/backcompat

So maybe its better to use that folder instead, for this specific case?

test/Microsoft.ML.Tests/Transformers/NormalizerTests.cs
eerhardt: I’m not necessarily saying remove this test. It can still provide value by itself testing at a lower level.

But I’m saying it is good to have the complete end-to-end scenario tested as well. 

We should already have an image set to use from other image tests.

src/Microsoft.ML.Data/Transforms/Normalizer.cs
eerhardt: Checking in binary files into our code repo becomes a problem by bloating our git history. That’s why we made the testdata repo.

src/Microsoft.ML.Data/Transforms/Normalizer.cs
```diff
@@ -387,32 +388,70 @@ internal static DataViewType LoadType(ModelLoadContext ctx)
...
+                    InternalDataKind itemKindOld = (InternalDataKind)ctx.Reader.ReadByte();
+                    Contracts.CheckDecode(itemKindOld == InternalDataKind.R4 || itemKindOld == InternalDataKind.R8);
+                    var itemTypeOld = ColumnTypeExtensions.PrimitiveTypeFromKind(itemKindOld);
+                    return isVector ? (DataViewType)(new VectorDataViewType(itemTypeOld, vectorSize)) : itemTypeOld;
+                }
+
+                // *** New Binary format ***
+                //   - int: number of dimensions
+                //   - ints: as many as dimensions, each one represent the size of each dimension
+                //   - byte: ItemKind of input column (only R4 and R8 are valid)
```
yaeldekel: >byte [](start = 23, length = 4)

If you read (and write) this byte first, then in the `!isVector` case you can return `itemType` without reading anything else.

src/Microsoft.ML.Data/Transforms/Normalizer.cs
```diff
@@ -385,39 +386,77 @@ internal ColumnOptions(string name, string inputColumnName, DataViewType inputTy
...
                 Contracts.CheckDecode(itemKind == InternalDataKind.R4 || itemKind == InternalDataKind.R8);
-
                 var itemType = ColumnTypeExtensions.PrimitiveTypeFromKind(itemKind);
-                return isVector ? (DataViewType)(new VectorDataViewType(itemType, vectorSize)) : itemType;
+
+                if (!isVector)
+                    return itemType;
+
+                int ndimensions = ctx.Reader.ReadInt32();
+                Contracts.CheckDecode(ndimensions > 0);
```
antoniovs1029: I am assuming that if isVector is true, then ndimensions should be bigger than 0. This assumption is correct, right?

eerhardt: https://github.com/dotnet/machinelearning/blob/60b15835f3228e809967daa3506ce5c2422bd455/src/Microsoft.ML.DataView/VectorType.cs#L20-L35

src/Microsoft.ML.Data/Transforms/Normalizer.cs
```diff
@@ -385,39 +386,77 @@ internal ColumnOptions(string name, string inputColumnName, DataViewType inputTy
...
-                ctx.Writer.Write(vectorType?.Size ?? 0);
-
                 DataViewType itemType = vectorType?.ItemType ?? type;
                 itemType.RawType.TryGetDataKind(out InternalDataKind itemKind);
                 Contracts.Assert(itemKind == InternalDataKind.R4 || itemKind == InternalDataKind.R8);
                 ctx.Writer.Write((byte)itemKind);
+
+                if(vectorType != null)
+                {
+                    Contracts.Assert(vectorType == null || vectorType.IsKnownSize);
```
antoniovs1029: Should I move this assert before the 'if' block it is in?

antoniovs1029: ok, I've moved it

src/Microsoft.ML.Data/Transforms/Normalizer.cs
```diff
@@ -385,39 +386,74 @@ internal ColumnOptions(string name, string inputColumnName, DataViewType inputTy
...
                 Contracts.Assert(itemKind == InternalDataKind.R4 || itemKind == InternalDataKind.R8);
                 ctx.Writer.Write((byte)itemKind);
+
+                Contracts.Assert(vectorType == null || vectorType.IsKnownSize);
+                if (vectorType != null)
+                {
+                    var dims = vectorType.Dimensions;
+                    ctx.Writer.Write(dims.Length);
+                    for (int i = 0; i < dims.Length; i++)
+                            ctx.Writer.Write(dims[i]);
```
eerhardt: (nit) this is too far indented.

test/Microsoft.ML.Tests/Transformers/NormalizerTests.cs
```diff
@@ -886,5 +887,93 @@ void TestNormalizeLogMeanVarianceFixZeroVec()
...
+            var normalizer = ML.Model.Load(modelPath, out var schema);
+
+            var outputPath = GetOutputPath("NormalizerEstimator", "normalized2.tsv");
+            using (var ch = Env.Start("save"))
+            {
+                var saver = new TextSaver(Env, new TextSaver.Arguments { Silent = true });
+                using (var fs = File.Create(outputPath))
+                {
+                    var transformedData = normalizer.Transform(data);
+                    var dataView = ML.Transforms.DropColumns(new[] { "float0" }).Fit(transformedData).Transform(transformedData);
```
eerhardt: Why are you dropping `float0` here?

codemzs:
Can you please remove “WIP” from the title?

test/Microsoft.ML.Tests/Transformers/NormalizerTests.cs
antoniovs1029: I based this test on the [NormalizerWorkout test](https://github.com/dotnet/machinelearning/blob/461b7d38c06d64e53c76e29770665c4b0bb44e72/test/Microsoft.ML.Tests/Transformers/NormalizerTests.cs#L30), where that column is dropped so to be able to compare the transformed data with the file on `normalized.tsv` (which is also done in this test). I've just realized that the `float0 `column is created on that test to work with some things that I don't do in this test; i.e. the normalizer transformer I load from disk doesn't actually do anything with that column.

So I've fixed this by simply removing the part of the `TextLoader `where `float0 `is created, and removed this line of code you've referred since its no longer needed to drop it.

---
In reply to: [334561513](https://github.com/dotnet/machinelearning/pull/4321#discussion_r334561513) [](ancestors = 334561513)

src/Microsoft.ML.Data/Transforms/Normalizer.cs
antoniovs1029: I've done this. Please review if it was what you were expecting. Thanks!

test/Microsoft.ML.OnnxTransformerTest/DnnImageFeaturizerTest.cs
```diff
@@ -173,5 +174,61 @@ public void TestOldSavingAndLoading()
...
+            // ColumnName attribute is used to change the column name from
+            // its default value, which is the name of the field.
+            [ColumnName("PredictedLabel")]
+            public String Prediction { get; set; }
+            public float[] Score { get; set; }
+        }
+
+        [Fact]
+        public void TestLoadFromDiskAndPredictionEngine()
+        {
```
antoniovs1029: I wasn't sure if it was okay to add this test in here, in the DnnImageFeaturizerTest.

Originally I intended to add this as a Functional Test, but I believe I couldn't do that because I couldn't access ResNet18 unless, I believe, I changed the dependencies there. So I added the test in here instead.

Also, in order to make this test to run in here I had to modify the project `OnnxTransformerTest.csproj ` to include `CpuMathNative`, because if I didn't do that, then the test threw an error.

In short, I wasn't sure where to put this test; if this isn't the place, please tell me so I can make the appropriate changes. Thanks. @codemzs

test/Microsoft.ML.Tests/Transformers/NormalizerTests.cs
antoniovs1029: I've added this test in my latest iteration. Thanks.

test/Microsoft.ML.OnnxTransformerTest/DnnImageFeaturizerTest.cs
antoniovs1029: Also, I wonder if it should use [OnnxFact] instead of [Fact]??

antoniovs1029: Also, I don't know if to include CpuMathNative into the OnnxTransformerTest.csproj the only thing I had to do was to add a line of code (which I did). I wonder if I had to do something else.

test/Microsoft.ML.OnnxTransformerTest/Microsoft.ML.OnnxTransformerTest.csproj
```diff
@@ -27,6 +27,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <NativeAssemblyReference Include="CpuMathNative" />
```
antoniovs1029: I don't know if to include CpuMathNative into the OnnxTransformerTest.csproj the only thing I had to do was to add this line of code (which I did manually). I wonder if I had to do something else.

test/Microsoft.ML.OnnxTransformerTest/DnnImageFeaturizerTest.cs
antoniovs1029: Before changing [Fact] to [OnnxFact] some tests were failing, but now all of them pass. So I guess it is correct to make that change... but I still wonder if this is the right place for this test.

antoniovs1029: So I discussed this offline with @codemzs and all of these is fine. So I'll mark this as resolved.

test/Microsoft.ML.OnnxTransformerTest/Microsoft.ML.OnnxTransformerTest.csproj
antoniovs1029: So I discussed this offline with @codemzs and all of these is fine. So I'll mark this as resolved.

src/Microsoft.ML.Data/Transforms/Normalizer.cs
yaeldekel: Yes! Thanks!

---
In reply to: [334615811](https://github.com/dotnet/machinelearning/pull/4321#discussion_r334615811) [](ancestors = 334615811)

justinormont:
@eerhardt: Is this ready for merge? The CodeCoverage is failing; does that matter?

eerhardt:
> Is this ready for merge? 

I believe so.

> The CodeCoverage is failing; does that matter?

No.

src/Microsoft.ML.Data/Transforms/Normalizer.cs
```diff
@@ -385,39 +386,74 @@ internal ColumnOptions(string name, string inputColumnName, DataViewType inputTy
...
+
+                if (!isVector)
+                    return itemType;
+
+                int ndimensions = ctx.Reader.ReadInt32();
+                Contracts.CheckDecode(ndimensions > 0);
+
+                var dimensions = new int[ndimensions];
+                for (int i = 0; i < ndimensions; i++)
+                    dimensions[i] = ctx.Reader.ReadInt32();
```
codemzs: Please use ctx.Reader.ReadIntArray() function, its a util extension function that is designed for reading arrays.

src/Microsoft.ML.Data/Transforms/Normalizer.cs
```diff
@@ -385,39 +386,74 @@ internal ColumnOptions(string name, string inputColumnName, DataViewType inputTy
...
                 Contracts.Assert(itemKind == InternalDataKind.R4 || itemKind == InternalDataKind.R8);
                 ctx.Writer.Write((byte)itemKind);
+
+                Contracts.Assert(vectorType == null || vectorType.IsKnownSize);
+                if (vectorType != null)
+                {
+                    var dims = vectorType.Dimensions;
+                    ctx.Writer.Write(dims.Length);
+                    for (int i = 0; i < dims.Length; i++)
+                        ctx.Writer.Write(dims[i]);
```
codemzs: Please use ctx.Writer.WriteIntArray() function, its a util extension function that is designed for writing arrays.

codemzs:
@justinormont @eerhardt It's not ready to be merged I have few comments that need to be addressed but it should be merged EOD.

PR #4324, opened 2019-10-11
Fix nuget dependency for Dnn

### LINKED ISSUE #4325: [Image Classification DNN based] Cannot use GPU with NuGet 0.16.0-preview2
The NuGet package 'Microsoft.ML.Dnn 0.16.0-preview2' is including a dependency on the CPU-based `SciSharp.TensorFlow.Redist` package: 

image

Therefore the user cannot reference and use the `SciSharp.TensorFlow.Redist-Windows-GPU` package because the CPU version has preference, afaik.  

The NuGet package `Microsoft.ML.Dnn 0.16.0-preview2` must not reference any of those, so depending on what SciSharp TensorFlow redist (CPU vs. GPU) the user is referencing from his code, it'll use CPU or GPU.

I know this is being fixed in the ML.NET source code repo with this PR after my heads-up:
https://github.com/dotnet/machinelearning/pull/4324

**But users mostly use the NuGet packages so we probably need to publish a new fix-patch-release for that package like the following?**:

`Microsoft.ML.Dnn 0.16.1-preview2` ?

Any other solution so users can try Image Classification with GPU when using the NuGet packages?

bpstark:
With change to allow users to support GPU, need to remove the dependency
on TF redist from nuget. It will now be up to the user to add a
dependency on the correct nuget, such that they can choose to use either
the GPU or the CPU.

fixes #4325

### DIFF
--- a/pkg/Microsoft.ML.Dnn/Microsoft.ML.Dnn.nupkgproj
+++ b/pkg/Microsoft.ML.Dnn/Microsoft.ML.Dnn.nupkgproj
@@ -6,7 +6,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="SciSharp.TensorFlow.Redist" Version="1.14.0" />
     <PackageReference Include="System.IO.FileSystem.AccessControl" Version="$(SystemIOFileSystemAccessControl)" />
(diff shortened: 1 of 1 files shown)


codemzs:
Have you tested this change by consuming the nuget?

bpstark:
@codemzs Yes tested this using the nuget package, and confirmed using the GPU version of TF redist now.

PR #4334, opened 2019-10-14
Add non-generic IEstimator interface to be more SOLID

NeoXtreem:
This is a fairly trivial change, but makes the `IEstimator` interface comply with SOLID more, particularly interface segregation.  It is useful in that code need only declare a variable of `IEstimator` type (instead of `IEstimator<ITransformer>`) if they wish only to access `GetOutputSchema` which does not require the generic type parameter `TTransformer`.  Or where they only need to cast to an `IEstimator` where the generic is unimportant such as the usage of the null-coalescing operator in the following example: 
```cs
IEstimator<ITransformer> pipeline = null;

foreach (var label in new[] {"Foo1", "Foo2"})
{
    var estimator = mlContext.Transforms.CopyColumns("Label", label)
        .Append(mlContext.Transforms.Concatenate("Features", "Baz", "Bar"))
        .Append(mlContext.Regression.Trainers.FastTree());

    pipeline = pipeline?.Append(estimator) ?? (IEstimator)estimator;
}
```

Also fixed a couple of typos in `EstimatorChain` that I noticed.


src/Microsoft.ML.Core/Data/IEstimator.cs
```diff
@@ -297,16 +297,22 @@ public interface ITransformer : ICanSaveModel
     /// <summary>
     /// The estimator (in Spark terminology) is an 'untrained transformer'. It needs to 'fit' on the data to manufacture
     /// a transformer.
-    /// It also provides the 'schema propagation' like transformers do, but over <see cref="SchemaShape"/> instead of <see cref="DataViewSchema"/>.
+    /// <inheritdoc />
     /// </summary>
-    public interface IEstimator<out TTransformer>
+    public interface IEstimator<out TTransformer> : IEstimator
```
eerhardt: This change isn't possible to make at this point because it is an API breaking change. That is the reason all the test legs are failing in CI:

https://dev.azure.com/dnceng/public/_build/results?buildId=387546&view=logs&jobId=ac97a0ee-ee90-5a55-af17-a5a589fa545f

```
/__w/1/s/packages/microsoft.dotnet.apicompat/1.0.0-beta.19225.5/build/Microsoft.DotNet.ApiCompat.targets(72,5): error : InterfacesShouldHaveSameMembers : Interface member 'Microsoft.ML.IEstimator<TTransformer>.GetOutputSchema(Microsoft.ML.SchemaShape)' is present in the contract but not in the implementation. [/__w/1/s/src/Microsoft.ML.Core/Microsoft.ML.Core.csproj]
```

Since we have shipped `1.0.0` a few months ago, we are no longer able to make API breaking changes, as this will break any consumers of ML.NET.

src/Microsoft.ML.Data/DataLoadSave/EstimatorChain.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.ML.Data
     public sealed class EstimatorChain<TLastTransformer> : IEstimator<TransformerChain<TLastTransformer>>
         where TLastTransformer : class, ITransformer
     {
-        // Host is not null iff there is any 'true' values in _needCacheAfter (in this case, we need to create an instance of
+        // Host is not null if there are any 'true' values in _needCacheAfter (in this case, we need to create an instance of
```
eerhardt: `iff` is not a type-o. It is intended. See "Entry 1" on https://www.merriam-webster.com/dictionary/iff.

eerhardt:
Thanks @NeoXtreem for the contribution! Unfortunately we won't be able to take a change like this since the API is now considered "stable".

For the cases you list, consumers are able to use `IEstimator<ITransformer>`. This type can be used to reference any `IEstimator` - and it works with null-coalescing as well, if you write your code as:

```C#
IEstimator<ITransformer> pipeline = null;

foreach (var label in new[] {"Foo1", "Foo2"})
{
    IEstimator<ITransformer>estimator = mlContext.Transforms.CopyColumns("Label", label)
        .Append(mlContext.Transforms.Concatenate("Features", "Baz", "Bar"))
        .Append(mlContext.Regression.Trainers.FastTree());

    pipeline = pipeline?.Append(estimator) ?? estimator;
}
```

src/Microsoft.ML.Core/Data/IEstimator.cs
NeoXtreem: Users of `IEstimator<ITransformer>` would not see a difference because the inherited `IEstimator` is public by default and therefore its members visible, right? Or is this because changes in the signature of public interfaces cannot be changed (e.g., if reflection is used to query the interface methods)?

NeoXtreem:
> Thanks @NeoXtreem for the contribution! Unfortunately we won't be able to take a change like this since the API is now considered "stable".
> 
> For the cases you list, consumers are able to use `IEstimator<ITransformer>`. This type can be used to reference any `IEstimator` - and it works with null-coalescing as well, if you write your code as:
> 
> ```cs
> IEstimator<ITransformer> pipeline = null;
> 
> foreach (var label in new[] {"Foo1", "Foo2"})
> {
>     IEstimator<ITransformer> estimator = mlContext.Transforms.CopyColumns("Label", label)
>         .Append(mlContext.Transforms.Concatenate("Features", "Baz", "Bar"))
>         .Append(mlContext.Regression.Trainers.FastTree());
> 
>     pipeline = pipeline?.Append(estimator) ?? estimator;
> }
> ```

@eerhardt Thanks for the reply. Yes, I'm aware that I can use `IEstimator<ITransformer>` which is the reason I suggested this change - to avoid having to (a little less client code, and more SOLID to use just `IEstimator`).  In your example, `estimator` could be declared as just `IEstimator`.

But no worries if this can't be pushed through.  Not a showstopper. :)

src/Microsoft.ML.Data/DataLoadSave/EstimatorChain.cs
NeoXtreem: OK, thanks. Have reverted and also correct other small typo in the same sentence. `is` should be `are` though I believe.

src/Microsoft.ML.Core/Data/IEstimator.cs
eerhardt: See https://stackoverflow.com/a/1522698/3680432

> Refactoring interface members into a base interface

It looks like this is exactly the same scenario.

NeoXtreem: Ah, so it breaks _explicit_ implementations. OK, I guess this will have to be left out then if breaking changes are not permitted. Is it something that could go in `2.0.0`?

eerhardt: There is some debate on whether we should make breaking changes, even on a major version release. The problem is, anyone building a library on top of `Microsoft.ML v1.0.0` could be broken if used in an application that referenced `Microsoft.ML v2.0.0`.

NeoXtreem: Of course, there will always be a divide in the development community, just like in anything. From experience, breaking changes are now considered part and parcel of software application major version releases (recently-released .NET Core 3.0 has many for example including no longer using Newtonsoft Json.NET, but added instructions for handling this [here](https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-3.0&tabs=visual-studio&WT.mc_id=DT-MVP-5002999#jsonnet-support) as well as other migration instructions on the [same page](https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-3.0&tabs=visual-studio)). Migration or upgrade instructions such as this tend to accompany major or even sometimes minor releases to deal with breaking changes and I imagine ML.NET 2.0.0 will be no different (I'm sure there will be many other breaking changes).

So my suggestion would be to keep this PR open for `2.0.0` as I'm fairly certain in all probability that this will be just one of many other breaking changes when the time comes.

NeoXtreem: @sharwell This isn't about a code style change, this is about making the code more SOLID, particularly with regard to ISP. Currently, I am forced in my client code to depend on an interface (`ITransformer`) that I do not care for when only using `GetOutputSchema` or less.

I have no problem adding `IEstimator` to all concrete classes, but there is a long list of them (see below) which breaks DRY as well as SRP (all those classes will have an additional reason to change, i.e., both `IEstimator<T>` and `IEstimator`).

Your example of `IReadOnlyList<T>` and `IList<T>` is not analogous to `IEstimate<T>` and `IEstimator`. `IReadOnlyList<T>` and `IReadOnlyCollection<T>` were added as convenience interfaces to restrict explicit usage of those types to only provide indexing or a count. Any client code can still cast to a `List<T>` and have full write-access to the collection. A more comparable example is `IEnumerable<T>` and `IEnumerable` which is almost exactly the same scenario as `IEstimate<T>` and `IEstimator`. The non-generic version of each offers one method that doesn't require `T` while all methods on the generic versions _do_ require `T`.

I would suggest your own argument applies in this case in that a breaking change such as this applies as the code is broken as many would regarding a framework breaking ISP causing more brittle client code as broken (or at least fragile).

**List of concrete classes that inherit `IEstimator<T>`:**
```
1. EstimatorChain<TLastTransformer> : IEstimator<TransformerChain<TLastTransformer>>
2. DelegateEstimator<TTransformer> : IEstimator<TTransformer>
3. TrainedWrapperEstimatorBase : IEstimator<TransformWrapper>
4. TrivialEstimator<TTransformer> : IEstimator<TTransformer>
5. CalibratorEstimatorBase<TICalibrator> : IEstimator<CalibratorTransformer<TICalibrator>>, IHaveCalibratorTrainer
6. ITrainerEstimator<out TTransformer, out TModel> : IEstimator<TTransformer>
7. ColumnConcatenatingEstimator : IEstimator<ColumnConcatenatingTransformer>
8. HashingEstimator : IEstimator<HashingTransformer>
9. NormalizingEstimator : IEstimator<NormalizingTransformer>
10. ValueToKeyMappingEstimator : IEstimator<ValueToKeyMappingTransformer>
11. internal sealed class DnnRetrainEstimator : IEstimator<DnnRetrainTransformer>
12. TreeEnsembleFeaturizationEstimatorBase : IEstimator<TreeEnsembleFeaturizationTransformer>
13. VectorWhiteningEstimator : IEstimator<VectorWhiteningTransformer>
14. DnnImageFeaturizerEstimator : IEstimator<TransformerChain<ColumnCopyingTransformer>>
15. PrincipalComponentAnalyzer : IEstimator<PrincipalComponentAnalysisTransformer>
16. TensorFlowEstimator : IEstimator<TensorFlowTransformer>
17. SsaChangePointEstimator : IEstimator<SsaChangePointDetector>
18. SsaForecastingEstimator : IEstimator<SsaForecastingTransformer>
19. SsaSpikeEstimator : IEstimator<SsaSpikeDetector>
20. CountFeatureSelectingEstimator : IEstimator<ITransformer>
21. MissingValueReplacingEstimator : IEstimator<MissingValueReplacingTransformer>
22. MutualInformationFeatureSelectingEstimator : IEstimator<ITransformer>
23. OneHotEncodingEstimator : IEstimator<OneHotEncodingTransformer>
24. OneHotHashEncodingEstimator : IEstimator<OneHotHashEncodingTransformer>
25. ApproximatedKernelMappingEstimator : IEstimator<ApproximatedKernelTransformer>
26. LatentDirichletAllocationEstimator : IEstimator<LatentDirichletAllocationTransformer>
27. NgramHashingEstimator : IEstimator<NgramHashingTransformer>
28. NgramExtractingEstimator : IEstimator<NgramExtractingTransformer>
29. TextFeaturizingEstimator : IEstimator<ITransformer>
30. WordEmbeddingEstimator : IEstimator<WordEmbeddingTransformer>
31. WordBagEstimator : IEstimator<ITransformer>
32. WordHashBagEstimator : IEstimator<TTransformer>
```

sharwell: I used `IReadOnlyList<T>` for my example because it was added after `IList<T>` shipped. `IEnumerable<T>` implemented `IEnumerable` from the start; if this change was proposed at a similar point in time this entire discussion would be very different.

I was referring to code style at a slightly higher level, which was intended to include behavior-preserving refactorings in consideration of code organization principles such as SOLID and DRY.

NeoXtreem: It's open for debate then. I agree it's different in ALM terms to `IEnumerable<T>` and `IEnumerable`, but it's the same in technical terms. And vice versa for `IList<T>` and `IReadOnlyList<T>`. So it comes down to the definition of 'broken', and for such a young library where these core types will be used for a long time, I'd argue that the threshold of 'broken' in terms of SOLID and DRY outweighs the breaking change aspect which I'd guess would be small pain (can you see from telemetry?) for a lot more long-term gain.

eerhardt: > for a lot more long-term gain.

I'm not sure I agree that this change will provide a lot of gain. `IEstimator` doesn't provide much value over `IEstimator<ITransformer>` in my opinion.

Your proposed `IEstimator` interface isn't capable of doing the "main" job that an estimator does - which is `Fit` some data. From a "Single responsibility principal" standpoint, the "single responsibility" of an estimator is to `Fit` data and produce a Transformer. See https://github.com/dotnet/machinelearning/blob/master/docs/code/MlNetHighLevelConcepts.md#list-of-high-level-concepts for what an Estimator is supposed to do.

In `IEnumerable` and `IEnumerable<T>`, they are both capable of doing the "main" job - enumerate a collection of objects. Here, `IEstimator` is no longer capable of doing what an "estimator" can do.

NeoXtreem: Then this is just a semantic issue then. The naming is not the point here. There is no escaping that I am forced to use `ITransformer` in places I don't want it, and that certainly breaks ISP. `IEstimator` technically doesn't break SRP, but the naming may be misleading in terms of the link you referenced (I'm only thinking in SOLID terms here, not ML terms). In that case, it could easily be named something else such as `ISchema` or anything else.

The gain may seem small, but it's a gain nonetheless in SOLID terms (I doubt that can be argued against even given the concurring replies on this thread). The magnitude of that gain is subjective, and it has already caused me problems in my code where I am having to use casting or an explicit polymorphic variable declaration (cannot use `var`, and possible small performance hit).

You've already seen the effect on `List<T>` with `IReadOnlyList<T>` being shoehorned in later. `IEnumerable` got it right from the start. It's still early days with ML.NET, so get it right now before it becomes more difficult. The use case for this is a form of imputation as you mentioned in [your answer to my SO question](https://stackoverflow.com/a/58543375/197591) recently, which resulted in this issue. If this becomes a common use case (and I see no reason for it not to as it could apply in most predictions involving multiple properties on an object), then this could become an annoying niggle for people due to the casting or polymorphic variable declaration required, both of which can be strong code smells in SOLID coding.

eerhardt: > There is no escaping that I am forced to use ITransformer in places I don't want it,

Can you give a concrete example of where you want to use estimators, but don’t want to use transformers?

The only reason estimators exist is to Fit data to create transformers. So I don’t believe there are real-world use cases where you’d want to use an estimator, but not eventually get a transformer from it.

> and that certainly breaks ISP.

There are 2 methods on IEstimator: Fit and GetSchemaShape. These two things go together. Fit will return a transformer that transforms an input dataset into an output dataset. GetSchemaShape describes what shape that output data will have given an input shape. There is no reason to separate these two things. Just like there is no reason to separate “Add” and “Remove” into separate interfaces from ICollection. They go together.

NeoXtreem: The example in the issue description is the concrete example from my own application where I have an `EstimatorChain`. In this case, I only want to call `Fit` on the chain, not the individual estimators that make it up. My code has some isolated logic which creates a single estimator and assigns it to a variable only used in that logic which it then appends to a wider-scope estimator pipeline variable or, if the pipeline doesn't exist yet (first in the chain), makes it the first estimator in a new pipeline by _assigning_ it to that wider-scope variable.

\<aside\>
This is in fact a design problem I noticed with `EstimatorChain` in that the following two methods result in very different pipelines:

1. Create an empty `EstimatorChain` then append an estimator then append another one to the first estimator.
1. Create an estimator then append another one to it thereby implicitly creating an `EstimatorChain`.

\<\/aside\>

As requested, to show my real-world code in its entirety (using some metasyntactics for brevity) to achieve imputation:
```
var mlContext = new MLContext(0);
IEstimator<ITransformer> pipeline = null;

foreach (var (label, score) in new[] {("Bar", "BarPrediction"), ("Baz", "BazPrediction")})
{
    IEstimator<ITransformer> estimator = CreateEstimators()
        .Aggregate((IEstimator<ITransformer>)mlContext.Transforms.CopyColumns("Label", label), (c, e) => c.Append(e))
        .Append(mlContext.Transforms.CopyColumns(score, "Score"));

    // Appends to the pipeline only after the first iteration.
    pipeline = pipeline?.Append(estimator) ?? estimator;
}

IEstimator<ITransformer>[] CreateEstimators()
{
    return new IEstimator<ITransformer>[]
    {
        mlContext.Transforms.Concatenate("Features", new[] {"Feature1", "Feature2"}),
        mlContext.Regression.Trainers.FastTree()
    };
}

var trainTestData = mlContext.Data.TrainTestSplit(mlContext.Data.LoadFromEnumerable(items));
var model = pipeline.Fit(trainTestData.TrainSet);
```
...where `items` is an `IEnumerable` of my input data with `Bar`, `Baz`, `Feature1` and `Feature2` properties.

As you can see, the `estimator` variable in the `foreach` doesn't need to be the stronger `IEstimator<ITransformer>` type but is covariantly forced to be so, but its job is still fulfilled later as part of a chain that calls `Fit` holistically and not on each estimator in the chain.

Hope this helps you see the issue more clearly.

PS: I updated my last comment slightly to be more accurate (replaced 'polymorphic' with 'covariant').

eerhardt: Have you tried re-writing your above code with your proposed change to see if you achieve the desired result? I don't believe you will be able to with the proposed change.

> the `estimator` variable in the `foreach` doesn't need to be the stronger `IEstimator<ITransformer>` type

That is not correct. 

```
// Appends to the pipeline only after the first iteration.
pipeline = pipeline?.Append(estimator) ?? estimator;
```

Here you are assigning `estimator` to `pipeline`.  And `pipeline` is of type `IEstimator<ITransformer>`, thus `estimator` must be assignable to `IEstimator<ITransformer>`.

Also, there is no `Append` method that takes a plain `IEstimator`. But there is one that takes `IEstimator<T>`. So that part of the line won't compile either.

NeoXtreem: Alas, of course you are correct. I was too focussed on the need to cast or have a covariant assignment to allow the code to work that I overlooked the assignment of `estimator` to the outer variable (`pipeline`), and the signature of `Append`. It was (and still is) a code smell, but I completely went down the wrong solution path. Sorry if I wasted anyone's time.

The silver lining is that my original instinct that I didn't pursue via this PR, and only alluded to as an 'aside' in my last comment, and which led to me raising _this_ PR about `IEstimator`, may be more correct.

So I've gone back to my original attempt where I started with an empty `EstimatorChain`. This is the code:
```cs
var mlContext = new MLContext(0);
IEstimator<ITransformer> pipeline = new EstimatorChain<ITransformer>();

foreach (var (label, score) in new[] {("Bar", "BarPrediction"), ("Baz", "BazPrediction")})
{
    IEstimator<ITransformer> estimator = CreateEstimators()
        .Aggregate((IEstimator<ITransformer>)mlContext.Transforms.CopyColumns("Label", label), (c, e) => c.Append(e))
        .Append(mlContext.Transforms.CopyColumns(score, "Score"));

    pipeline = pipeline.Append(estimator);
}

IEstimator<ITransformer>[] CreateEstimators()
{
    return new IEstimator<ITransformer>[]
    {
        mlContext.Transforms.Concatenate("Features", "Feature1", "Feature2"),
        mlContext.Regression.Trainers.FastTree()
    };
}
```
This is of course slightly better as `pipeline` is initialised upon declaration, so there is no need for a null-coalescing operator when it is assigned in the `foreach` (which was being done to ensure it was initialised on the first loop). I still keep the declaration of `estimator` covariant, though.

At the end of this code, `pipeline` looks like this:
image

Two sibling estimators each with four child estimators.

However, if I make one seemingly innocuous change to the code by making the declaration of `estimator` implicit using `var`, then I suddenly get a very different result:
image

I believe this breaks LSP as the behaviour of the subsystem should not change just because the type of the passed-in variable is different, especially if it's the same object.

However, this is more likely a completely separate matter, and not related to `Estimator` at all, so if you agree this is an issue, then I can raise an issue for it.

eerhardt: > Sorry if I wasted anyone's time.

I think we all learned something in this discussion, so I don't think anyone's time was wasted.

> However, this is more likely a completely separate matter, and not related to Estimator at all, so if you agree this is an issue, then I can raise an issue for it.

I definitely don't want to dissuade you from opening issues that you see. If you believe it is an issue, feel free to log it.

Are you seeing any observable behavior difference between the two? For example, is an exception being thrown in one case, or is the data pipeline changed in a way such that the two pipelines behave differently? If yes, then it is definitely a bug that needs to be fixed.

NeoXtreem: @eerhardt I've done some testing with four different ways of creating the pipeline, namely all four combinations of:

```cs
IEstimator<ITransformer> pipeline = new EstimatorChain<ITransformer>();
...
pipeline = pipeline.Append(estimator);
```

vs

```cs
IEstimator<ITransformer> pipeline = null;
...
pipeline = pipeline?.Append(estimator) ?? estimator;
```

And...

```cs
IEstimator<ITransformer> estimator = CreateEstimators()...
```
vs
```cs
var estimator = CreateEstimators()...
```

In each case, I got a different tree of estimators, but the prediction was exactly the same.  Should the difference in estimators tree be considered a bug or design?  And if the prediction is the same in all cases, should it be fixed if it's a bug?

eerhardt: My personal opinion would be that as long as the end result is the same, I wouldn't think it is a bug. But if you feel it is an issue, feel free to log it as a separate issue.

NeoXtreem: OK, I trust your personal opinion in this case. I'm not familiar enough with the way ML estimator chains work to make a considered judgment on it, so I'll leave as is. If it's a real issue, I'm sure it will be raised in due course. Thanks for all your help on this.

Issue #97, opened 2018-05-09
Support Windows x86 builds


eerhardt:
### System information

- **OS version/distro**:  Windows
- **.NET Version (eg., dotnet --info)**:   All - desktop, .net core

### Issue

- **What did you do?**  Try to use ML.NET in an x86 process
- **What happened?**  It doesn't work because CpuMathNative can't be loaded into an x86 process
- **What did you expect?**  I expected it to work in a Windows x86 process.

### Notes

There are cases where developers are forced to use x86 processes.  For example, if their hosting environment only supports x86 processes.  Or if they are using other native libraries that are only available for x86, and they don't want to (or can't) spin up multiple processes (x64 for ML.NET and x86 for other native libraries).

Ivanidzo4ka:
Is this good place for discussion?
From C# point of view, we currently have SseUtils, AvxUtils classes which we call directly. We need to change our code to use something like CpuUtils which will perform platform validation and instruction availability and fallback on proper implementation.

From native point of view, is it possible to have x64 and x86 code in one library or we need to compile two version? 

Or shall we get rid of native code and switch to  hardware Intrinsics in .core 2.1?

fiigii:
> Or shall we get rid of native code and switch to hardware Intrinsics in .core 2.1?

Hardware intrinsic in .NET Core 2.1 supports 64-bit and 32-bit both, so switching to use HW intrinsic will solve the problem for .NET Core build at least. We can keep the native code for other platform builds (e.g., .NET framework, Mono, etc.).

dan-drews:
Adding another +1 to this. I looked to add this to an older application that is set to be x86. Will look to change that to be 64-bit (I'm assuming without issue) in the near future in order to use this.

I could spin up a new web service to handle the ML predictions, but that'll take a while to get the traction. Thankfully, our application is not complicated, so converting to x64 should go off without a hitch, but I can only assume that others will expect this support.

Anipik:
Another +1 for x86 . I was trying to use ML.net in azure functions. But Azure Functions only support x86 processes. And for training as well predicting the value we require CpuMathNative.
The error occurs while calculating L2Norm
```C#
Microsoft.ML.Runtime.Data.LpNormNormalizerTransform.GetGetterCore.AnonymousMethod__5(ref Microsoft.ML.Runtime.Data.VBuffer<float> dst) Line 434
[...]

PR #4337, opened 2019-10-15
Update OnnxRuntime to latest version.

eerhardt:
OnnxRuntime supports win-x86, so start running tests on the win-x86 leg.


test/Microsoft.ML.TestFramework/Attributes/OnnxFactAttribute.cs
```diff
@@ -12,14 +12,15 @@ namespace Microsoft.ML.TestFramework.Attributes
...
     public sealed class OnnxFactAttribute : EnvironmentSpecificFactAttribute
     {
-        public OnnxFactAttribute() : base("Onnx is 64-bit Windows only")
+        public OnnxFactAttribute() : base("Onnx is not supported on Linux with libc < v2.23")
         {
         }
 
         /// <inheritdoc />
         protected override bool IsEnvironmentSupported()
         {
```
harishsk: If we are going ahead with this change right now, can you please make the corresponding change in Microsoft.ML.Tests\OnnxConversionTest.cs in the function IsOnnxRuntimeSupported?

This release doesn't add any extra capabilities that ML.NET needs (other than x86 support - Do we have any numbers for x86 consumers of ML.NET? ). But a major release of ORT is forthcoming very soon and that has a lot more fixes and features we need.  I would prefer to wait for that release to make the next major upgrade and announce that along with all the new transformers that would support exporting to Onnx.

But if you think this is important to get through right now, please make the change above and I will sign off on the PR.

eerhardt: > I would prefer to wait for that release to make the next major upgrade

Why would we wait? If there is a new version now - why not get the test coverage now? Especially since the work to upgrade now is very minimal - changing a version number.

> Do we have any numbers for x86 consumers of ML.NET? 

This came up in the context of using ML.NET in Azure Functions. Someone pointed out me to https://github.com/dotnet/machinelearning#operating-systems-and-processor-architectures-supported-by-mlnet

> 32 bit is supported on Windows, except for TensorFlow, LightGBM, and ONNX related functionality.

ONNX functionality supports x86 (even in the current `0.4.0` version). This reminds me, I should update that line in the README in this PR as well.

eerhardt: > can you please make the corresponding change in Microsoft.ML.Tests\OnnxConversionTest.cs in the function IsOnnxRuntimeSupported?

Done. I consolidated the logic into a single place.

eerhardt: > My impression is that deployments of 32bit processors and 32 bit Windows is steadily shrinking.

While that is true, there are still applications/processes that are 32-bit. Take, for example, Visual Studio. Even though your OS and machine are 64-bit, the Visual Studio process itself is 32-bit. (Visual Studio is just an example here - there are plenty of other applications that are 32-bit. See #97.)

> Do you know where we can get the numbers from?

I don't.

eerhardt:
It looks like with this change our CI machines are bumping against the storage limit:

```
D:\a\1\s\Tools\dotnetcli\sdk\3.0.100-preview9-014004\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(198,5): error : There is not enough space on the disk. : 'D:\a\1\s\bin\packages\Microsoft.ML.symbols.1.4.0-preview3-28215-0.nupkg' [D:\a\1\s\pkg\Microsoft.ML\Microsoft.ML.symbols.nupkgproj]
```

@chcosta - is there any way to request more storage on the hosted machines?

I guess another option would be to clean up some space after we publish the test results, and before we build the NuGet packages.

PR #4338, opened 2019-10-15
Remove TensorFlow.Redist since we no longer use it.

eerhardt:


eerhardt:
This is green and ready to be reviewed.

PR #4340, opened 2019-10-16
Added LearningRateScheduler functionality for Image Classification

harshithapv:
1. Added LearningRateScheduler abstract class which will be implemented for various learning rate scheduling algorithms. 
2.  Has the lastest Tensorflow nuget which allows Gradient Descent Optimizer to use learning rate input as tensor


src/Microsoft.ML.Dnn/GradientDescentOptimizerTensor.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            _lr_t = learningRateTensor;
+        }
+
+        public override void _prepare()
+        {
+        }
+
+    }
+}
```
codemzs: I have asked Haiping to add this to the TF .NET library and do a patch release.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -382,12 +385,18 @@ private IDataView GetShuffledData(string path)
             metrics.Train = new TrainMetrics();
             float accuracy = 0;
             float crossentropy = 0;
+            TrainState trainstate = new TrainState();
+            trainstate.BatchSize = options.BatchSize;
+            trainstate.BatchesPerEpoch = (int)(trainingSet.GetRowCount() ?? 0 / options.BatchSize);
```
codemzs: >trainstate.BatchesPerEpoch = (int)(trainingSet.GetRowCount() ?? 0 / options.BatchSize); [](start = 12, length = 87)

GetRowCount() can be null even when there are rows, so this line will result in an incorrect number of batches per epoch. If total number of examples/rows isn't available anywhere then you can simple calculate that in CacheFeaturizedImagesToDisk as we are featuring every image by keeping a counter and then passing that counter to this method.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -407,15 +416,18 @@ private IDataView GetShuffledData(string path)
 
                         labelBatch[batchIndex] = label;
                         batchIndex += 1;
+                        trainstate.CurrentBatchIndex = batchIndex;
                         // Train.
                         if (batchIndex == batchSize)
                         {
+
+                            learningRate = options.LearningRateScheduler.GetLearningRate(trainstate);
```
codemzs: >learningRate = options.LearningRateScheduler.GetLearningRate(trainstate); [](start = 28, length = 73)

please only do this if scheduling is being used otherwise use constant learning as it was being used before. Passing LR in each run is just inefficient and does not make sense if LR is going to be constant.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -658,10 +672,11 @@ private void VariableSummaries(RefVariable var)
                 {
                     _bottleneckInput = tf.placeholder_with_default(
                         bottleneckTensor,
-                        shape: bottleneckTensorDims ,
+                        shape: bottleneckTensorDims,
                         name: "BottleneckInputPlaceholder");
-                }
+                    _learningRateInput = tf.placeholder(tf.float64, null, name: "learningRateInputPlaceholder");
```
codemzs: >_learningRateInput = tf.placeholder(tf.float64, null, name: "learningRateInputPlaceholder"); [](start = 20, length = 92)

This should only be added when scheduling is needed,.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -710,7 +725,7 @@ private void VariableSummaries(RefVariable var)
 
             tf_with(tf.name_scope("train"), delegate
             {
-                var optimizer = tf.train.GradientDescentOptimizer(learningRate);
+                var optimizer = new GradientDescentOptimizerTensor(_learningRateInput);
```
codemzs: > var optimizer = new GradientDescentOptimizerTensor(_learningRateInput); [](start = 15, length = 72)

please pass LR tensor when scheduling is needed.

src/Microsoft.ML.Dnn/GradientDescentOptimizerTensor.cs
Oceania2018: Related issue: https://github.com/SciSharp/TensorFlow.NET/issues/422

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
Oceania2018: Please upgrade to https://www.nuget.org/packages/TensorFlow.NET/0.11.7

codemzs: Thanks Haiping for quickly adding this constructor!!

src/Microsoft.ML.Dnn/GradientDescentOptimizerTensor.cs
harshithapv: I am aware of this. I am the one who added the issue on Tensorflow.NET. Thanks.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -674,7 +688,8 @@ private void VariableSummaries(RefVariable var)
                         bottleneckTensor,
                         shape: bottleneckTensorDims,
                         name: "BottleneckInputPlaceholder");
-                    _learningRateInput = tf.placeholder(tf.float64, null, name: "learningRateInputPlaceholder");
+                    if(useLearningRateScheduler)
+                        _learningRateInput = tf.placeholder(tf.float64, null, name: "learningRateInputPlaceholder");
```
Oceania2018: I think we should use `tf.float32`.

harshithapv: Thanks for the suggestion. I realize we don't need a high precision to store learning rate. Will change that.

build/Dependencies.props
```diff
@@ -16,15 +16,15 @@
     <GoogleProtobufPackageVersion>3.5.1</GoogleProtobufPackageVersion>
     <LightGBMPackageVersion>2.2.3</LightGBMPackageVersion>
     <MicrosoftExtensionsPackageVersion>2.1.0</MicrosoftExtensionsPackageVersion>
-    <MicrosoftMLOnnxRuntimePackageVersion>0.5.1</MicrosoftMLOnnxRuntimePackageVersion>
+    <MicrosoftMLOnnxRuntimePackageVersion>0.4.0</MicrosoftMLOnnxRuntimePackageVersion>
```
eerhardt: Please revert this change.

src/Microsoft.ML.Dnn/LearningRateScheduler.cs
```diff
@@ -0,0 +1,105 @@
...
+        internal int CurrentBatchIndex;
+        internal int CurrentEpoch;
+        internal int BatchSize;
+        internal int BatchesPerEpoch;
+    }
+
+    /// <summary>
+    /// This interface defines a learning rate scheduler.
+    /// </summary>
+    public interface ILearningRateScheduler
```
eerhardt: Instead of an interface, let's use an abstract class here. That way we can add more members to it in the future.

src/Microsoft.ML.Dnn/LearningRateScheduler.cs
```diff
@@ -0,0 +1,105 @@
...
+    {
+        internal float GetLearningRate(TrainState options);
```
eerhardt: this should not be `internal`.

src/Microsoft.ML.Dnn/LearningRateScheduler.cs
```diff
@@ -0,0 +1,156 @@
...
+        internal int CurrentEpoch;
+        internal int BatchSize;
+        internal int BatchesPerEpoch;
+    }
+
+    /// <summary>
+    /// This interface defines a learning rate scheduler.
+    /// </summary>
+    public abstract class LearningRateScheduler
+    {
```
eerhardt: Make an `internal` constructor on `LearningRateScheduler`, so no one outside of this assembly can derive from it.

src/Microsoft.ML.Dnn/LearningRateScheduler.cs
```diff
@@ -0,0 +1,156 @@
...
+        internal abstract float GetLearningRate(TrainState options);
+    };
```
eerhardt: This `;` is not necessary.

src/Microsoft.ML.Dnn/LearningRateScheduler.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// <summary>
+        /// Learning rate is scaled at epoch boundaries provided in LrSchedule to corresponding multiplier in the LrSchedule.
+        /// Format for LrSchedule: {epoch, scaling factor}
+        /// </summary>
+        public readonly float[,] LrSchedule;
+
+        /// <summary>
+        /// Base Learning rate to start off with.
+        /// </summary>
+        public float BaseLearningRate;
```
eerhardt: `readonly`?

src/Microsoft.ML.Dnn/LearningRateScheduler.cs
```diff
@@ -0,0 +1,156 @@
...
+
+        /// <summary>
+        /// Learning rate decay factor.
+        /// </summary>
+        public float DecayRate;
+
+        /// <summary>
+        /// If Staircase is True the learning rate decays at discrete intervals and the decayed learning rate follows a staircase function.
+        /// </summary>
+        public bool Staircase { get; }
```
eerhardt: Why is this a property, but everything else in this class is a field?

IMO - we should only be exposing properties publicly in APIs.

src/Microsoft.ML.Dnn/LearningRateScheduler.cs
```diff
@@ -0,0 +1,156 @@
...
+    /// Implementation adopted from RESNET-CIFAR benchmark test in Tensorflow slim.
+    /// https://github.com/tensorflow/models/blob/b974c3f95a37acedcc3c58566834c78fcae4b214/official/vision/image_classification/resnet_cifar_main.py
+    /// </summary>
+    public sealed class LsrDecay : LearningRateScheduler
+    {
+        /// <summary>
+        /// Learning rate is scaled at epoch boundaries provided in LrSchedule to corresponding multiplier in the LrSchedule.
+        /// Format for LrSchedule: {epoch, scaling factor}
+        /// </summary>
+        public readonly float[,] LrSchedule;
```
eerhardt: A thing with having public array fields is that people can change the elements of the array. Even though it is `readonly`, that only means that someone can't change this reference to a different array. But the array itself can be changed.

eerhardt: Also having a multi-dimensional array `[,]` in a public API is a little bit confusing. If you used an `IReadOnlyList` of some new `struct` type, then you can name each argument:

```
public readonly struct LearningDecayItem
{
    public readonly float Epoch;  // actually should this be float or int?
    public readonly float ScalingFactor;
}
```

Then people know which is which easily.

codemzs:
@harshithapv Can you please update the title and description of the PR?

src/Microsoft.ML.Dnn/LearningRateScheduler.cs
harshithapv: What remedy do you suggest? Basically I want it to be set only in the constructor.

src/Microsoft.ML.Dnn/LearningRateScheduler.cs
harshithapv: Thanks changed it. It doesn't have to be a property. I added that by mistake.

src/Microsoft.ML.Dnn/LearningRateScheduler.cs
harshithapv: Removed

src/Microsoft.ML.Dnn/LearningRateScheduler.cs
harshithapv: Thanks will change that.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -317,11 +329,18 @@ private IDataView GetShuffledData(string path)
                     .Load(new MultiFileSource(path)));
         }
 
+        private int GetNumSamples(string path)
+        {
+            using (var reader = File.OpenText(path))
+                return int.Parse(reader.ReadLine());
```
codemzs: >return int.Parse(reader.ReadLine()); [](start = 16, length = 36)

align with line 334

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -759,6 +805,7 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Dat
             scoreColumnName = ctx.Reader.ReadString();
             predictedColumnName = ctx.Reader.ReadString();
             learningRate = ctx.Reader.ReadFloat();
+            useLearningRateScheduling = ctx.Reader.ReadBoolByte();
```
codemzs: >useLearningRateScheduling [](start = 12, length = 25)

why are we serializing this? It is only needed at training and not inferencing, right?

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -1172,13 +1222,18 @@ public sealed class TrainMetrics
...
+            public float LearningRate { get; set; }
+
             /// <summary>
             /// String representation of the metrics.
             /// </summary>
             public override string ToString()
             {
                 if (DatasetUsed == ImageClassificationMetrics.Dataset.Train)
-                    return $"Phase: Training, Dataset used: {DatasetUsed.ToString(),10}, Batch Processed Count: {BatchProcessedCount,3}, " +
+                    return $"Phase: Training, Dataset used: {DatasetUsed.ToString(),10}, Batch Processed Count: {BatchProcessedCount,3}, learning rate: {LearningRate,10} " +
```
codemzs: >learning rate [](start = 137, length = 13)

Learning Rate

src/Microsoft.ML.Dnn/LearningRateScheduler.cs
```diff
@@ -0,0 +1,160 @@
...
+    internal class TrainState
+    {
+        internal int CurrentBatchIndex;
+        internal int CurrentEpoch;
+        internal int BatchSize;
+        internal int BatchesPerEpoch;
+    }
+
+    /// <summary>
+    /// This interface defines a learning rate scheduler.
```
codemzs: >interface [](start = 13, length = 9)

abstract class

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
harshithapv: We should not be serializing this. This was discussed with Zeeshan and will be addressed by him in his PR.

---
In reply to: [337825750](https://github.com/dotnet/machinelearning/pull/4340#discussion_r337825750) [](ancestors = 337825750)

Issue #4323, opened 2019-10-10
Move DatabaseLoader from the Experimental package


eerhardt:
We want to mark the DatabaseLoader as a "stable API" when we ship 1.4. However, it is currently in the Experimental package, which will never be marked as stable.

We should move the DatabaseLoader to a stable package.

If I get a vote, my vote would be to add it to the core `Microsoft.ML` package, next to TextLoader. DatabaseLoader adds no extra dependencies, so I think it is completely fine to add it to the `Microsoft.ML` package.

eerhardt:
fyi - @codemzs @ebarsoumMS

tannergooding:
If we can confirm where it should go, I'm happy to move it there.

eerhardt:
If I was to do it, I would move it next to the `TextLoader`. I don't feel adding a new package for this is valuable since it doesn't add any new dependencies.

@CESARDELATORRE @codemzs @yaeldekel - thoughts? We need this in order to ship a stable `1.4` package with the DatabaseLoader in it.

CESARDELATORRE:
I agree. It should be near the TextLoader type if ti is not adding any new dependencies.
It’d be a very small package if adding a new one.

tannergooding:
Any other input I should wait for, or am I good to move the types and put up the PR?

tannergooding:
Move the type into a new assembly and add a type forward at that point?

codemzs:
Ok, sounds good. Let’s move it out of Experimental nuget. Thanks @eerhardt and @tannergooding

PR #4342, opened 2019-10-16
Moving the DatabaseLoader support into Microsoft.ML.Data

tannergooding:
This resolves https://github.com/dotnet/machinelearning/issues/4323


tannergooding:
CC. @eerhardt, @CESARDELATORRE

eerhardt:
I believe we can remove this file now:

https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Experimental/Properties/AssemblyInfo.cs

It was only added for the DatabaseLoader change.

Also, I believe we can revert remove this line as well:

https://github.com/dotnet/machinelearning/blob/461b7d38c06d64e53c76e29770665c4b0bb44e72/src/Microsoft.ML.Core/Properties/AssemblyInfo.cs#L25

Basically, we can undo the extra workarounds in #4035.

It looks like there are 2 workarounds that can be undone from #4091. (Adding 2 `[BestFriend]` attributes.)

PR #4346, opened 2019-10-17
Extract TestFrameworkCommon Project, remove dependency of TestFramework from FT

frank-dong-ms:
1. Extract TestFrameworkCommon from TestFramework
   a. Extract CommonUtilities method, move some utilities from BaseTestClass to CommonUtilities
   b. Move some test attributes from TestFramework to TestFrameworkCommon
   c. Move TestDataset from TestFramework to TestFrameworkCommon and fix corresponding 
reference
   d. Move some schema compare method from TestDataPipeBase to CommonUtilities
   e. TestFrameworkCommon only have public dependency, no internal class dependency
2. Remove dependency of TestFramework from FT
   a. Remove dependency of TestFramework from FT, now FT dependent on TestFrameworkCommon
   b. Remove unnecessary dependency of Maml from FT
3. Remove duplicate CompareVec method, also fix bug in CompareVec method
   a. Remove duplicate CompareVec method from CoreBaseTestClass, TestDataPipeBase and 
CopyColumnEstimatorTests, move CompareVec method to CommonUtilities and fix reference
   b. Fix bug in CompareVec method from CoreBaseTestClass class, original line 125 seems problematic.


test/Microsoft.ML.Core.Tests/UnitTests/CoreBaseTestClass.cs
```diff
@@ -85,75 +76,10 @@ protected Func<bool> GetComparerVec<T>(DataViewRow r1, DataViewRow r2, int col,
...
-            }
-
-            Contracts.Assert(!v1.IsDense || !v2.IsDense);
-            int iiv1 = 0;
-            int iiv2 = 0;
-            var v1Indices = v1.GetIndices();
-            var v2Indices = v2.GetIndices();
-            for (; ; )
-            {
-                int iv1 = v1.IsDense ? iiv1 : iiv1 < v2Indices.Length ? v1Indices[iiv1] : v1.Length;
```
frank-dong-ms: >v2Indices [](start = 53, length = 9)

Seems like a bug to me, should be v1Indices

harishsk: Good catch! Yes, it does appear to be a bug. Can you please check the callers who are using this function and how those tests are passing?

---
In reply to: [336252686](https://github.com/dotnet/machinelearning/pull/4346#discussion_r336252686) [](ancestors = 336252686)

frank-dong-ms: TestEntryPoints class is using this method, these tests will pass as long as v1 and v2 has same indices length, this happens to be true in these tests

---
In reply to: [336584593](https://github.com/dotnet/machinelearning/pull/4346#discussion_r336584593) [](ancestors = 336584593,336252686)

test/Microsoft.ML.TestFrameworkCommon/Microsoft.ML.TestFrameworkCommon.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!--<PackageReference Include="Microsoft.ML" Version="1.3.1" />
+    <PackageReference Include="Microsoft.ML.Ensemble" Version="0.15.1" />
+    <PackageReference Include="Microsoft.ML.FastTree" Version="1.3.1" />-->
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.8.0" />
```
harishsk: I am guessing this is no longer needed? If not, can you please delete it?

test/Microsoft.ML.Tests/Transformers/NormalizerTests.cs
```diff
@@ -410,10 +411,10 @@ public void SimpleConstructorsAndExtensions()
             var data4 = est4.Fit(data).Transform(data);
             var data5 = est5.Fit(data).Transform(data);
 
-            CheckSameSchemas(data1.Schema, data2.Schema);
-            CheckSameSchemas(data1.Schema, data3.Schema);
-            CheckSameSchemas(data1.Schema, data4.Schema);
-            CheckSameSchemas(data1.Schema, data5.Schema);
+            CommonUtilities.CheckSameSchemas(data1.Schema, data2.Schema);
```
harishsk: >CommonUtilities [](start = 12, length = 15)

Does CommonUtilities sounds a bit too generic as a name? How about TestCommon or TestUtilities? I would appreciate it if another reviewer can chime in on this on whether this needs to be changed or okay as is.

test/Microsoft.ML.TestFrameworkCommon/Microsoft.ML.TestFrameworkCommon.csproj
frank-dong-ms: Yes, will remove these

---
In reply to: [337122073](https://github.com/dotnet/machinelearning/pull/4346#discussion_r337122073) [](ancestors = 337122073)

test/Microsoft.ML.TestFrameworkCommon/CommonUtilities.cs
```diff
@@ -0,0 +1,366 @@
...
+        }
+
+        private static void Fail(string fmt, params object[] args)
+        {
+            Fail(false, fmt, args);
+        }
+
+        private static void Fail(bool relax, string fmt, params object[] args)
+        {
+            Log("*** Failure: " + fmt, args);
```
eerhardt: This won't fail the test. Instead of `Fail` we should use the `Assert` class in xunit.

test/Microsoft.ML.TestFrameworkCommon/CommonUtilities.cs
```diff
@@ -0,0 +1,366 @@
...
+        }
+
+        private static bool Check(bool f, string msg)
+        {
+            if (!f)
+                Fail(msg);
+            return f;
+        }
+
+        [Conditional("DEBUG")]
```
eerhardt: I don't understand why we have these methods. These are tests, you should use the `Assert` class in xunit. If any conditions fail, it should fail the test, no matter if this is DEBUG or not.

frank-dong-ms: This function is copied from Contracts Utilities class and this is also widely used in our test project like below, so are these also abuse and should be fixed?

protected bool EqualTypes(DataViewType type1, DataViewType type2, bool exactTypes)
        {
            Contracts.AssertValue(type1);
            Contracts.AssertValue(type2);

            return exactTypes ? type1.Equals(type2) : type1.SameSizeAndItemType(type2);
        }



---
In reply to: [337248885](https://github.com/dotnet/machinelearning/pull/4346#discussion_r337248885) [](ancestors = 337248885)

test/Microsoft.ML.Tests/Transformers/NormalizerTests.cs
frank-dong-ms: Yes, this is good suggestion, will modify this

---
In reply to: [337126027](https://github.com/dotnet/machinelearning/pull/4346#discussion_r337126027) [](ancestors = 337126027)

test/Microsoft.ML.TestFrameworkCommon/CommonUtilities.cs
eerhardt: Just `Assert.IsNotNull` instead of `Contracts.AssertValue`

test/Microsoft.ML.TestFrameworkCommon/Microsoft.ML.TestFrameworkCommon.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <ProjectReference Include="..\..\src\Microsoft.ML.Data\Microsoft.ML.Data.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.ML.Ensemble\Microsoft.ML.Ensemble.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.ML.EntryPoints\Microsoft.ML.EntryPoints.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.ML.FastTree\Microsoft.ML.FastTree.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.ML.KMeansClustering\Microsoft.ML.KMeansClustering.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.ML.PCA\Microsoft.ML.PCA.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <NativeAssemblyReference Include="LdaNative" />
```
eerhardt: I don't believe these add any value, since we never actually execute the TestFrameworkCommon project. These `NativeAssemblyReference` are only necessary for projects that get executed (like real test projects).

You can remove this whole ItemGroup.

test/Microsoft.ML.TestFrameworkCommon/TestCommon.cs
```diff
@@ -0,0 +1,285 @@
...
+using System.Diagnostics;
+using System.IO;
+using Microsoft.ML.Data;
+using Xunit;
+
+namespace Microsoft.ML.TestFrameworkCommon
+{
+    public static class TestCommon
+    {
+        public static string GetOutputPath(string name, string outDir)
```
eerhardt: This seems backwards to me. Shouldn't `outDir` go before `name`? Just like how `Path.Combine(outDir, name)` does?

test/Microsoft.ML.TestFrameworkCommon/Microsoft.ML.TestFrameworkCommon.csproj
frank-dong-ms: I see, thanks, will remove these.
 Copied these from TestFramework project, then I realized there are some tests in TestFramework and TestFramework project needs these, but sounds like TestFramework project shouldn't have any real test cases.

---
In reply to: [337707632](https://github.com/dotnet/machinelearning/pull/4346#discussion_r337707632) [](ancestors = 337707632)

codemzs:
@frank-dong-ms Thanks and looks good to me!

PR #4347, opened 2019-10-17
Devproperties

harishsk:
@eerhardt I am creating this draft pull request to initiate a discussion about this. 

Michael and I are running into similar issues where we would like to do our daily development against unstable and unreleased nugets from other sources and then merge our daily working branches into master when our dependent nugets are stable. 

The current approach is to modify Directory.Build.Props and Dependencies.props appropriately and keep those files checked out. 

What do you think of optionally importing DevProperties.props from Directory.Build.props which would override those variables? (The DevProperties.props file would never be checked in and would be added to the .gitignore file if you are okay with this approach).

@michaelgsharp I have verified that this approach works. Let me know if this works for you as well.


eerhardt:
> we would like to do our daily development against unstable and unreleased nugets from other sources and then merge our daily working branches into master when our dependent nugets are stable.

The way I solve this is to add the unstable/unreleased source to the Directory.Build.props file, and commit it to your private branch. Then when the nugets move to stable and move to official feeds, then update your Dependencies.props and Directory.Build.props as necessary to the new version, and to remove the unnecessary feed. People can easily catch adding unnecessary feeds during the code review when you actually do want to merge into master.

The reasons I'd recommend that approach instead is because if you are collaborating with someone else on the same branch. You add a dependency on `PackageA, v1.2.3` and commit that into your branch, but you don't add the feed to where that package/version can be found. When anyone else pulls the branch and tries to build, they will fail to find the package and need to manually update their "DevProperties.props" file. Another reason is if someone `git clean`s their repo, the "DevProperties.props" file will be deleted, because it isn't under source control. Thus, they would need to manually create it again.

In general, if you are committing code that depends on `PackageA, v1.2.3` into a branch, you should also be committing the source of that package as well. That way the branch stays self-describing.

harishsk:
We are already adding the unstable/unreleased source to the Directory.Build.props file and committing to the private branch. If that is the best approach available, I will go ahead and close this PR. 

---
In reply to: [543887368](https://github.com/dotnet/machinelearning/pull/4347#issuecomment-543887368) [](ancestors = 543887368)

PR #4349, opened 2019-10-18
Added support for resnet50 architecture for image classification

bpstark:
Created a sample to show usage of resnet50 V2.
Tested against Tensorflow we can see there is a definitive discrepancy
in accuracy.

Training for ~136 epochs (TF uses steps rather than epochs) we are able
to achieve an accuracy of 78.7% in Tensorflow, compared to our accuracy
which was 54.6%. This discrepancy can be accounted for due to the way in
which TF adjusts the learning rate over time. We have already begun to
make those changes to our code, and will be added in a separate change.


eerhardt:
Do you want to add a test for the new functionality? The CI isn't executing any of your new code.

bpstark:
@eerhardt Will add the unit test as soon as #4351 Is complete as I need to rely on the newly added theory to avoid duplication.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -132,7 +132,7 @@ private static ImageClassificationTransformer Create(IHostEnvironment env, Model
 
             return new ImageClassificationTransformer(env, DnnUtils.LoadTFSession(env, modelBytes), outputs, inputs,
                 null, addBatchDimensionInput, 1, labelColumn, checkpointName, arch,
-                scoreColumnName, predictedColumnName, learningRate, null, classCount, true, predictionTensorName,
+                scoreColumnName, predictedColumnName, learningRate,null, null, classCount, true, predictionTensorName,
```
codemzs: >e,null [](start = 65, length = 6)

space, ctrl + K + D

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -833,6 +833,11 @@ private static IRowMapper Create(IHostEnvironment env, ModelLoadContext ctx, Dat
                 _bottleneckOperationName = "import/MobilenetV2/Logits/Squeeze";
                 _inputTensorName = "import/input";
             }
+			else if (arch == ImageClassificationEstimator.Architecture.ResnetV250)
```
codemzs: >			 [](start = 0, length = 3)

spaces not tabs

test/Microsoft.ML.Benchmarks/ImageClassificationBench.cs
```diff
@@ -0,0 +1,225 @@
...
+
+            // Split the data 90:10 into train and test sets, train and
+            // evaluate.
+            TrainTestData trainTestData = mlContext.Data.TrainTestSplit(
+                shuffledFullImagesDataset, testFraction: 0.1, seed: 1);
+
+            trainDataset = trainTestData.TrainSet;
+            testDataset = trainTestData.TestSet;
+
+        }
```
codemzs: new line

test/Microsoft.ML.Benchmarks/ImageClassificationBench.cs
```diff
@@ -0,0 +1,225 @@
...
+        [Benchmark]
+        public TransformerChain<KeyToValueMappingTransformer> Train()
```
codemzs: >Train [](start = 62, length = 5)

TrainResnetV250

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -1108,7 +1115,8 @@ public enum Architecture
         {
             { Architecture.ResnetV2101, new Tuple<int, int>(299,299) },
             { Architecture.InceptionV3, new Tuple<int, int>(299,299) },
-            { Architecture.MobilenetV2, new Tuple<int, int>(224,224) }
+            { Architecture.MobilenetV2, new Tuple<int, int>(224,224) },
+            { Architecture.ResnetV250, new Tuple<int, int>(299,299) }
```
codemzs: >9,2 [](start = 61, length = 3)

ctrl + K + D?

codemzs:
builds seem to be failing because of /__w/1/s/Tools/dotnetcli/sdk/3.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets(234,5): error NETSDK1004: Assets file '/__w/1/s/bin/obj/AnyCPU.Release-netcoreapp3_0/Microsoft.ML.Benchmarks/project.assets.json' not found. Run a NuGet package restore to generate this file. [/__w/1/s/test/Microsoft.ML.Benchmarks/Microsoft.ML.Benchmarks.csproj]
/__w/1/s/dir.traversal.targets(25,5): error : Build failed. See earlier errors. [/__w/1/s/build.proj]

eerhardt:
> using System;

This needs a copyright header

---
Refers to: test/Microsoft.ML.Benchmarks/ImageClassificationBench.cs:1 in d0ffa47. [](commit_id = d0ffa472dc632f6f7ef322af2e1617e178d41b31, deletion_comment = False)

test/Microsoft.ML.Benchmarks/ImageClassificationBench.cs
```diff
@@ -0,0 +1,226 @@
...
+        [Benchmark]
+        public TransformerChain<KeyToValueMappingTransformer> TrainResnetV250()
+        {
+            var pipeline = mlContext.Model.ImageClassification(
+                "Image", "Label",
+                arch: ImageClassificationEstimator.Architecture.ResnetV250,
+                epoch: 50,
+                batchSize: 10,
+                learningRate: 0.01f,
+                metricsCallback: (metrics) => Console.WriteLine(metrics),
```
eerhardt: Do we really want to write out to the Console during a benchmark?

test/Microsoft.ML.Benchmarks/ImageClassificationBench.cs
```diff
@@ -0,0 +1,226 @@
...
+
+            string relativeFilePath = Path.Combine(destDir, destFileName);
+
+            if (File.Exists(relativeFilePath))
+            {
+                Console.WriteLine($"{relativeFilePath} already exists.");
+                return false;
+            }
+
+            var wc = new WebClient();
```
eerhardt: Don't use WebClient. Instead use HttpClient.

[https://docs.microsoft.com/en-us/dotnet/api/system.net.webclient?view=netframework-4.8#remarks](https://docs.microsoft.com/en-us/dotnet/api/system.net.webclient?view=netframework-4.8#remarks)

> Important
> We don't recommend that you use the WebClient class for new development. Instead, use the System.Net.Http.HttpClient class.

test/Microsoft.ML.Benchmarks/ImageClassificationBench.cs
```diff
@@ -0,0 +1,226 @@
...
+            Console.WriteLine($"Downloading {relativeFilePath}");
+            var download = Task.Run(() => wc.DownloadFile(url, relativeFilePath));
```
eerhardt: Don't use Task.Run, and then Thread.Sleep. This is an anti-pattern.

Instead, this task should be `async` and then you can `await` the download to complete.

Note that a `[GlobalSetup]` method can return a `Task`, and BenchmarkDotnet will wait for that task's completion before running the benchmarks.

test/Microsoft.ML.Benchmarks/ImageClassificationBench.cs
```diff
@@ -0,0 +1,226 @@
...
+        {
+            var flag = gzArchiveName.Split(Path.DirectorySeparatorChar)
+                .Last()
+                .Split('.')
+                .First() + ".bin";
+
+            if (File.Exists(Path.Combine(destFolder, flag))) return;
+
+            Console.WriteLine($"Extracting.");
+            var task = Task.Run(() =>
```
eerhardt: Similar comment here. Don't use `Task.Run`. Instead just call `ZipFile.ExtractToDirectory` inline in the method. No need to mess with different threads.

Issue #4307, opened 2019-10-07
DnnCatalog methods should use a public Options class


eerhardt:
The two methods in `DnnCatalog`:

https://github.com/dotnet/machinelearning/blob/c9f616fdc5a144250535949fc8c8f9c971cbab88/src/Microsoft.ML.Dnn/DnnCatalog.cs#L49-L63

and

https://github.com/dotnet/machinelearning/blob/c9f616fdc5a144250535949fc8c8f9c971cbab88/src/Microsoft.ML.Dnn/DnnCatalog.cs#L120-L143

have a lot of optional parameters, and users may get confused on which ones are important, and which aren't.

I've seen comments from @terrajobst in the past saying:

> In UX studies we have seen that many developers struggle with methods that have many optional arguments.

In ML.NET, the pattern we have established is that we have 2 overloads:

1. An overload that takes the required parameters, and optionally the most important/common parameters to a method.
2. An overload that takes an `Options` object, which contains all the options to the method - simple and advanced.

We should follow this pattern with these DNN APIs as well. See the discussions and related PRs to https://github.com/dotnet/machinelearning/issues/1798.

/cc @ebarsoumMS @codemzs

codemzs:
@eerhardt We are aware and already tracking this issue and it will be implemented before we go GA.

ashbhandare:
I will be looking into this.

codemzs:
Thanks @ashbhandare

PR #4351, opened 2019-10-18
Add support for Mobilenet v2 in Image Classification transfer learning

ashbhandare:
The base MobileNetV2 model is from https://github.com/tensorflow/models/tree/master/research/slim/nets/mobilenet . Checkpoint file used: https://storage.googleapis.com/mobilenet_v2/checkpoints/mobilenet_v2_1.4_224.tgz

Benchmarked transfer learning on the Cats V/s Dogs dataset at:
https://www.microsoft.com/en-us/download/details.aspx?id=54765

Trained last layer for 10 epochs with 0.0001f learning rate with batch size 32.
Final accuracy on eval set:
 Micro-accuracy: 0.980959936533122,macro-accuracy = 0.980874311779911
Comparable Tensorflow acuracy for same task: 
https://www.tensorflow.org/tutorials/images/transfer_learning#train_the_model


test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1249,7 +1249,7 @@ public void TensorFlowImageClassification()
             var pipeline = mlContext.Transforms.LoadImages("Image", fullImagesetFolderPath, false, "ImagePath") // false indicates we want the image as a VBuffer<byte>
                 .Append(mlContext.Model.ImageClassification(
                     "Image", "Label",
-                    // Just by changing/selecting InceptionV3 here instead of 
+                    // Just by changing/selecting InceptionV3/MobilenetV2 here instead of
```
eerhardt: Do you want to add a test for the new functionality? The CI isn't executing any of your new code.

ashbhandare: Added unit test.

---
In reply to: [336686355](https://github.com/dotnet/machinelearning/pull/4351#discussion_r336686355) [](ancestors = 336686355)

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -144,7 +144,7 @@ public static class DnnCatalog
         {
             var options = new ImageClassificationEstimator.Options()
             {
-                ModelLocation = arch == Architecture.ResnetV2101 ? @"resnet_v2_101_299.meta" : @"InceptionV3.meta",
+                ModelLocation = ModelLocation[arch],
```
codemzs: >ModelLocation = ModelLocation[arch], [](start = 16, length = 36)

Not your fault but since we are here might as well correct it. ModelLocation is not something that we need as an input from the user hence it should not be part of the options class. PLease remove this and move it as a property of ImageClassificationTransformer but not need to serialize it. This would mean you will need to move download code from lines 172 to 205 into ImageClassification constructor and specifically when we are not loading the model from disk, i,.e we are training the model for the first i.e within if (loadModel == false) case. This also means we can move ImagePreprocessingSize and ModelLocation constants into ImageClassificationTransformer since they are not really a property that needs to be set for initializing ImageClassification Estimator or something we need to expose publically.

ashbhandare: Since this change is not realted to the current PR adding MobilenetV2 support, I will make this change in the PR fixing [https://github.com/dotnet/machinelearning/issues/4307](https://github.com/dotnet/machinelearning/issues/4307).

---
In reply to: [337218608](https://github.com/dotnet/machinelearning/pull/4351#discussion_r337218608) [](ancestors = 337218608)

codemzs: ok, thanks @ashbhandare !

PR #4314, opened 2019-10-08
Use resource manager to download meta files. Fixes #4234

ashbhandare:
Previously, when the ImageClassification pipeline is run for the first time, the meta graph of the model (ResnetV2101 or InceptionV3) is downloaded, and in the subsequent runs, it is reused. If the run is interrupted while the download is in progress(eg.: by stopping), the protobuff is partially downloaded. This throws an error when this incomplete graph is attempted to be read in the subsequent runs.

This changes uses resource manager to download the meta file. Resource Manager downloads the file to a temporary file path(GUID) and then atomically moves the temporary file path to the actual path. Hence once a file path is created it is guaranteed to have all the contents (unless the user or someone goes and manipulates the file)

Fixes #4234

PR #4362, opened 2019-10-22
Redesign DnnCatalog methods API for ease of use and consistency.

### LINKED ISSUE #4307: DnnCatalog methods should use a public Options class

ashbhandare:
Fixes #4307

Initially the methods in the DnnCatalog took a lot of optional arguments which would confuse users and adding features to the ImageClassification API would require breaking changes in the future.

To be consistent with the rest of ML .Net, this change follows the pattern where we have 2 overloads:
1)An overload that takes the required parameters, and optionally the most important/common parameters to a method.
2)An overload that takes an Options object, which contains all the options to the method 

This changes also removes a couple of unused options for the ImageClassification API, while making RetrainDnnModel internal to enable further testing before releasing.

WIP to get initial feedback/add necessary tests.


eerhardt:
@ashbhandare - please don't use `WIP` in the title anymore on PRs. If the PR isn't ready for review and merge, then open a `Draft` PR.

https://github.blog/2019-02-14-introducing-draft-pull-requests/

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -83,6 +83,120 @@ public static class DnnCatalog
             return new DnnRetrainEstimator(env, options, DnnUtils.LoadDnnModel(env, modelPath, true));
         }
 
+        /// <summary>
+        /// The options for the <see cref="ImageClassificationTransformer"/>.
+        /// </summary>
+        public sealed class Options
```
eerhardt: We don't need 2 "Options" classes. We only need one.

This appears to be a duplicate of the `ImageClassificationEstimator.Options` class. One of these should be removed.

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -123,89 +221,62 @@ public static class DnnCatalog
...
+                TestOnTrainSet = true,
+                TrainSetBottleneckCachedValuesFilePath = "trainSetBottleneckFile.csv",
+                ValidationSetBottleneckCachedValuesFilePath = "validationSetBottleneckFile.csv",
+                ReuseTrainSetBottleneckCachedValues = false,
+                ReuseValidationSetBottleneckCachedValues = false
             };
 
-            if (!File.Exists(options.ModelLocation))
+            var env = CatalogUtils.GetEnvironment(catalog);
+            return new ImageClassificationEstimator(env, options, DnnUtils.LoadDnnModel(env, options.Arch, true));
```
eerhardt: This method should just call into the `ImageClassification` overload below that takes an `Options` class.

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -83,6 +83,120 @@ public static class DnnCatalog
             return new DnnRetrainEstimator(env, options, DnnUtils.LoadDnnModel(env, modelPath, true));
         }
 
+        /// <summary>
+        /// The options for the <see cref="ImageClassificationTransformer"/>.
```
eerhardt: This should say it is the options for the Estimator, not the Transformer. These options are used to create the Estimator.

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -97,23 +211,7 @@ public static class DnnCatalog
...
-        /// <param name="finalModelPrefix">The name of the prefix for the final mode and checkpoint files.</param>
         /// <param name="validationSet">Validation set.</param>
-        /// <param name="testOnTrainSet">Indicates to evaluate the model on train set after every epoch.</param>
-        /// <param name="reuseTrainSetBottleneckCachedValues">Indicates to not re-compute cached trainset bottleneck values if already available in the bin folder.</param>
-        /// <param name="reuseValidationSetBottleneckCachedValues">Indicates to not re-compute validataionset cached bottleneck validationset values if already available in the bin folder.</param>
-        /// <param name="trainSetBottleneckCachedValuesFilePath">Indicates the file path to store trainset bottleneck values for caching.</param>
-        /// <param name="validationSetBottleneckCachedValuesFilePath">Indicates the file path to store validationset bottleneck values for caching.</param>
         /// <remarks>
         /// The support for image classification is under preview.
```
eerhardt: @codemzs - When should this `remark` get removed? I'm assuming we don't want to ship with this in the documentation.

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -83,6 +83,120 @@ public static class DnnCatalog
...
 
+        /// <summary>
+        /// The options for the <see cref="ImageClassificationTransformer"/>.
+        /// </summary>
+        public sealed class Options
+        {
+            /// <summary>
+            /// The names of the model inputs.
+            /// </summary>
+            [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "The names of the model inputs", ShortName = "inputs", SortOrder = 1)]
```
eerhardt: Why is this marked with `ArgumentType.Multiple`? We only support a single `Features` column name, correct?

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -83,6 +83,120 @@ public static class DnnCatalog
...
+            /// Name of the tensor that will contain the predicted label from output scores of the last layer when transfer learning is done.
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Argmax tensor of the last layer in transfer learning.", SortOrder = 15)]
+            public string PredictedLabelColumnName = "PredictedLabel";
+
+            /// <summary>
+            /// Validation set.
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Validation set.", SortOrder = 15)]
+            public IDataView ValidationSet = null;
```
eerhardt: Reference types are `null` by default. You don't need to explicitly put `= null` at the end here, and below.

src/Microsoft.ML.Dnn/DnnCatalog.cs
eerhardt: What should happen is the nested `ImageClassificationEstimator.Options` class should become public, and then this one can be deleted.

---
In reply to: [337563694](https://github.com/dotnet/machinelearning/pull/4362#discussion_r337563694) [](ancestors = 337563694)

docs/samples/Microsoft.ML.Samples/Program.cs
```diff
@@ -15,7 +15,7 @@ internal static void RunAll()
             {
                 var sample = type.GetMethod("Example", BindingFlags.Public | BindingFlags.Static | BindingFlags.FlattenHierarchy);
 
-                if (sample != null)
+                if (sample != null && type.Name == "ResnetV2101TransferLearningTrainTestSplit")
                 {
```
harishsk: Did this get accidentally left behind?

src/Microsoft.ML.Dnn/DnnCatalog.cs
ashbhandare: The two classes have most members in common, except that DnnCatalog.Options has featuresColumnName, and  ImageClassificationEstimator.Options() takes in InputColumns and OutputColumns. Using ImageClassificationEstimator.Options directly would change the necessary arguments that ImageClassification() would take. I can make that change if that is acceptable. 


---
In reply to: [337568602](https://github.com/dotnet/machinelearning/pull/4362#discussion_r337568602) [](ancestors = 337568602,337563694)

docs/samples/Microsoft.ML.Samples/Program.cs
ashbhandare: That's right, thank you!

---
In reply to: [337586213](https://github.com/dotnet/machinelearning/pull/4362#discussion_r337586213) [](ancestors = 337586213)

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -46,7 +46,7 @@ public static class DnnCatalog
         /// <remarks>
         /// The support for retraining is under preview.
         /// </remarks>
-        public static DnnRetrainEstimator RetrainDnnModel(
+        internal static DnnRetrainEstimator RetrainDnnModel(
```
eerhardt: Should the `DnnRetrainTransformer` and `DnnRetrainEstimator` classes also be marked as `internal`?

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -1341,31 +1340,31 @@ public enum Dataset
...
             /// </summary>
-            [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "The name of the outputs", ShortName = "outputs", SortOrder = 2)]
+            [Argument(ArgumentType.Multiple , HelpText = "The name of the outputs", ShortName = "outputs", SortOrder = 2)]
             public string[] OutputColumns;
 
+            /// <summary>
+            /// The names of the model input features.
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce | ArgumentType.Required, HelpText = "The names of the model inputs", ShortName = "features", SortOrder = 1)]
+            public string FeaturesColumnName;
```
eerhardt: Why do we have both `InputColumns` and `FeaturesColumnName`? We should only have 1.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -1341,31 +1340,31 @@ public enum Dataset
         /// <summary>
         /// The options for the <see cref="ImageClassificationTransformer"/>.
         /// </summary>
-        internal sealed class Options : TransformInputBase
+        public sealed class Options : TransformInputBase
```
eerhardt: This shouldn't inherit from `TransformInputBase`. That class is for data transforms, and this is a trainer.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
eerhardt: The same applies for `LabelColumnName`.

And also for `OutputColumns` and `ScoreColumnName`.

---
In reply to: [337694012](https://github.com/dotnet/machinelearning/pull/4362#discussion_r337694012) [](ancestors = 337694012)

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ImageClassificationDefault.cs
```diff
@@ -0,0 +1,338 @@
...
+
+            string relativeFilePath = Path.Combine(destDir, destFileName);
+
+            if (File.Exists(relativeFilePath))
+            {
+                Console.WriteLine($"{relativeFilePath} already exists.");
+                return false;
+            }
+
+            var wc = new WebClient();
```
eerhardt: We need to fix this pattern. This is becoming pervasive in our repo.

It is not recommended to use WebClient. We should use HttpClient instead.

[https://docs.microsoft.com/en-us/dotnet/api/system.net.webclient?view=netframework-4.8#remarks](https://docs.microsoft.com/en-us/dotnet/api/system.net.webclient?view=netframework-4.8#remarks)

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ImageClassificationDefault.cs
```diff
@@ -0,0 +1,338 @@
...
+        {
+            var flag = gzArchiveName.Split(Path.DirectorySeparatorChar)
+                .Last()
+                .Split('.')
+                .First() + ".bin";
+
+            if (File.Exists(Path.Combine(destFolder, flag))) return;
+
+            Console.WriteLine($"Extracting.");
+            var task = Task.Run(() =>
```
eerhardt: There is no reason to do a `Task.Run` here, and then Thread.Sleep later. Just call `ZipFile.ExtractToDirectory` inline in the method.

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -6,11 +6,11 @@
 using System.IO;
 using System.IO.Compression;
 using System.Net;
+using Microsoft.ML.CommandLine;
```
eerhardt: Do we still need this using?

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -123,89 +107,32 @@ public static class DnnCatalog
...
-                else if(options.Arch == Architecture.MobilenetV2)
-                {
-                    var baseGitPath = @"https://tlcresources.blob.core.windows.net/image/MobileNetV2TensorFlow/mobilenet_v2.meta";
-                    using (WebClient client = new WebClient())
-                    {
-                        client.DownloadFile(new Uri($"{baseGitPath}"), @"mobilenet_v2.meta");
-                    }
-                }
-            }
+        public static ImageClassificationEstimator ImageClassification(
```
eerhardt: This should have XML docs since it is a public method.

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -1341,31 +1340,31 @@ public enum Dataset
...
+            [Argument(ArgumentType.AtMostOnce | ArgumentType.Required, HelpText = "The names of the model inputs", ShortName = "features", SortOrder = 1)]
+            public string FeaturesColumnName;
+
             /// <summary>
             /// The name of the label column in <see cref="IDataView"/> that will be mapped to label node in TensorFlow model.
             /// </summary>
-            [Argument(ArgumentType.AtMostOnce, HelpText = "Training labels.", ShortName = "label", SortOrder = 4)]
-            public string LabelColumn;
+            [Argument(ArgumentType.AtMostOnce | ArgumentType.Required, HelpText = "Training labels.", ShortName = "label", SortOrder = 4)]
+            public string LabelColumnName;
```
eerhardt: What's the difference between this and `TensorFlowLabel`?

docs/samples/Microsoft.ML.Samples/Dynamic/ImageClassification/ImageClassificationDefault.cs
ashbhandare: This change is included in another PR [https://github.com/dotnet/machinelearning/pull/4314](https://github.com/dotnet/machinelearning/pull/4314), which is failing the CI test intermittently. I would like to make that change separate from this if that's okay.

---
In reply to: [337696066](https://github.com/dotnet/machinelearning/pull/4362#discussion_r337696066) [](ancestors = 337696066)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: As far as I can understand, TensorFlowLabel is the name of the TF node in the model that contains the label, while the LabelColumn is the name of the column in the IDataView. They are assigned the same value, so working on removing one of them.

---
In reply to: [337697817](https://github.com/dotnet/machinelearning/pull/4362#discussion_r337697817) [](ancestors = 337697817)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: OutputColumns is an array of ScoreColumn and PredictedLabelColumn. Users can have a choice of setting either one or both of these names. I could make OutputColumns internal if that's cleaner.

---
In reply to: [337694812](https://github.com/dotnet/machinelearning/pull/4362#discussion_r337694812) [](ancestors = 337694812,337694012)

ashbhandare: Would this API take more than one input at any point, as InputColumn is an array? I could either make InputColumn internal or get rid of one of InputColumns and FeaturesColumnName.

---
In reply to: [337694012](https://github.com/dotnet/machinelearning/pull/4362#discussion_r337694012) [](ancestors = 337694012)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: done.

---
In reply to: [337694597](https://github.com/dotnet/machinelearning/pull/4362#discussion_r337694597) [](ancestors = 337694597)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
ashbhandare: Making InputColumns and OutputColumns internal after talking with @eric.erhardt@microsoft.com offline.

---
In reply to: [337712848](https://github.com/dotnet/machinelearning/pull/4362#discussion_r337712848) [](ancestors = 337712848,337694012)

src/Microsoft.ML.Dnn/DnnCatalog.cs
ashbhandare: Removing the remark after talking to @codemzs offline. 

---
In reply to: [337567053](https://github.com/dotnet/machinelearning/pull/4362#discussion_r337567053) [](ancestors = 337567053)

src/Microsoft.ML.Dnn/ImageClassificationTransform.cs
```diff
@@ -1341,37 +1339,31 @@ public enum Dataset
...
-            [Argument(ArgumentType.Required, HelpText = "TensorFlow model used by the transform. Please see https://www.tensorflow.org/mobile/prepare_models for more details.", SortOrder = 0)]
-            public string ModelLocation;
-
             /// <summary>
             /// The names of the model inputs.
             /// </summary>
-            [Argument(ArgumentType.Multiple | ArgumentType.Required, HelpText = "The names of the model inputs", ShortName = "inputs", SortOrder = 1)]
-            public string[] InputColumns;
+            [Argument(ArgumentType.Multiple , HelpText = "The names of the model inputs", ShortName = "inputs", SortOrder = 1)]
+            internal string[] InputColumns;
```
codemzs: > internal string[] InputColumns; [](start = 11, length = 32)

probably don't need this or Output coliumns but I will remove this in my PR where I will refactor all this code and convert to ITrainerEstimator

codemzs:
@eerhardt I need this change for my AutoML integration change and also the change where I convert this API from IEstimator to ITrainerEstimator, hence I will try to check-in as it mostly looks good and if there is any other feedback from you I will take it as part of my PR.

eerhardt:
>         public enum DnnFramework

@codemzs - do we still need this public enum? I don't think it is used anymore.

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTransform.cs:1115 in f956a07. [](commit_id = f956a07fdd81392db4483752e1abb7bf10d57ed5, deletion_comment = False)

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -97,115 +92,46 @@ public static class DnnCatalog
...
+        /// [!include[io](~/../docs/samples/docs/api-reference/tensorflow-usage.md)]
+        /// ]]>
+        /// </format>
+        /// </summary>
+        /// <param name="catalog"></param>
+        /// <param name="options">An <see cref="Options"/> object specifying advanced options for <see cref="ImageClassificationEstimator"/>.</param>
+        public static ImageClassificationEstimator ImageClassification(
+            this ModelOperationsCatalog catalog, Options options)
+        {
+            options.EarlyStoppingCriteria = options.DisableEarlyStopping ? null : options.EarlyStoppingCriteria ?? new EarlyStopping();
```
eerhardt: I still think this acts in a surprising behavior. What if we changed it to the following?

* Remove `options.DisableEarlyStopping`.
* Default `options.EarlyStoppingCriteria = new EarlyStopping();`

Then, by default the EarlyStopping is enabled to the default EarlyStopping. If someone wants to disable EarlyStopping, they set `options.EarlyStoppingCriteria = null;` when calling this API. If they want to change EarlyStopping to a different criteria, they can just set  `options.EarlyStoppingCriteria = new EarlyStopping(metric: EarlyStoppingMetric.Loss)` (or whatever criteria they need).

This feels like a better API design choice because then the caller doesn't have 2 properties to deal with - `DisableEarlyStopping` and `EarlyStoppingCriteria`. There is only 1 to consider, that either they leave to the default, or set it as appropriate. If they want to disable EarlyStopping - they just set it to `null`.

eerhardt:
> if there is any other feedback from you I will take it as part of my PR.

I have left the above 2 pieces of feedback. Can you address them in your PR? Thanks.

PR #4368, opened 2019-10-23
Fix warnings/errors while buliding NuGet packages.

eerhardt:
- A new warning that PackageIconUrl is deprecated, and use PackageIcon instead.
- Mkl.Redist has `build/netstandard2.0` assets, but no `lib` or `ref` assets, so this caused a warning. The fix is to put a placeholder file in the `lib/netstandard2.0` folder.
- Microsoft.ML.symbols had a warning that it contains a `build` targets file, but not for the current package name. This was suppressed because we only use the symbols packages for publishing symbols.


pkg/Directory.Build.props
```diff
@@ -34,6 +34,8 @@
     <Content Include="$(PackageAssetsPath)$(PackageIdFolderName)\analyzers\**\*" Pack="true" PackagePath="analyzers" />
     <Content Include="$(PackageAssetsPath)$(PackageIdFolderName)\runtimes\**\*" Pack="true" PackagePath="runtimes" />
     <Content Include="$(PackageAssetsPath)$(PackageIdFolderName)\tools\**\*" Pack="true" PackagePath="tools" />
+
+    <Content Include="$(RepoRoot)\pkg\mlnetlogo.png" Pack="true" PackagePath=""/>
```
safern: Nit: space before `/>`

PR #4370, opened 2019-10-23
Created Polynomial Learning Rate scheduler

bpstark:
Created the polynomial learning rate scheduler based upon Tensorflow's
implementation of polynomial lr scheduler. which can be found
https://www.tensorflow.org/api_docs/python/tf/compat/v1/train/polynomial_decay

### DIFF
--- a/src/Microsoft.ML.Dnn/LearningRateScheduler.cs
+++ b/src/Microsoft.ML.Dnn/LearningRateScheduler.cs
@@ -195,4 +195,68 @@ namespace Microsoft.ML.Dnn
         }
 
     }
+    /// <summary>
+    /// This class implements polynomial Learning rate decay.
+    /// Implemented from the tensorflow documentation.
+    /// Source: https://www.tensorflow.org/api_docs/python/tf/compat/v1/train/polynomial_decay
+    /// Default values and implementation of learning rate is from Tensorflow Slim model tests.
+    /// Source : https://github.com/tensorflow/models/blob/master/research/slim/train_image_classifier.py
+    /// </summary>
+    public sealed class PolynomialLRDecay : LearningRateScheduler
+    {
+        /// <summary>
+        /// Initial learning rate.
+        /// </summary>
+        public readonly float LearningRate;
+
+        /// <summary>
+        /// The minimal end learning rate.
+        /// </summary>
+        public readonly float EndLearningRate;
+
+        /// <summary>
+        /// The power of the polynomial
+        /// </summary>
+        public readonly float Power;
+
+        /// <summary>
+        /// whether or not it should cycle once decay has been reached
+        /// </summary>
+        public readonly bool Cycle;
+
+        /// <summary>
+        /// Number of epochs after which learning rate decays.
+        /// </summary>
+        public readonly float NumEpochsPerDecay;
+
+        public PolynomialLRDecay(float learningRate = 0.01f, float numEpochsPerDecay = 2.0f, float endLearningRate = 0.0001f, float power = 1.0f, bool cycle = false)
+        {
+            LearningRate = learningRate;
+            NumEpochsPerDecay = numEpochsPerDecay;
+            EndLearningRate = endLearningRate;
+            Power = power;
+            Cycle = cycle;
+        }
+
+        internal override float GetLearningRate(TrainState trainstate)
+        {
+            int numSamplesPerEpoch = trainstate.BatchSize * trainstate.BatchesPerEpoch;
+            int decaySteps = (int) (numSamplesPerEpoch * NumEpochsPerDecay / trainstate.BatchSize);
+            int globalStep = (trainstate.CurrentEpoch) *(trainstate.BatchesPerEpoch) + trainstate.CurrentBatchIndex;
+
+            float decayedLearningRate;
+            if (Cycle && globalStep > decaySteps)
+            {
+                float calculatedStep = (float)decaySteps * (float)Math.Ceiling((double)globalStep/(double)decaySteps);
+                decayedLearningRate = (LearningRate - EndLearningRate) * ((float) Math.Pow((1 - (float)globalStep / calculatedStep), Power)) + EndLearningRate;
+            }
+            else
+            {
+                float calculatedStep = Math.Min(globalStep, decaySteps);
+                decayedLearningRate = (LearningRate - EndLearningRate) * ((float) Math.Pow((1 - calculatedStep / (float)decaySteps), Power)) + EndLearningRate;
+            }
+            return decayedLearningRate;
+        }
+
+    }
 }
--- a/test/Microsoft.ML.Tests/LearningRateSchedulerTest.cs
+++ b/test/Microsoft.ML.Tests/LearningRateSchedulerTest.cs
@@ -1,0 +1,54 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.ML.RunTests;
+using Microsoft.ML.Transforms;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.ML.Tests
+{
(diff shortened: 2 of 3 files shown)


src/Microsoft.ML.Dnn/LearningRateScheduler.cs
```diff
@@ -192,4 +192,52 @@ internal override float GetLearningRate(TrainState trainstate)
...
+    /// <summary>
+    /// This class implements polynomial Learning rate decay.
+    /// Implemented from the tensorflow documentation.
+    /// Source: https://www.tensorflow.org/api_docs/python/tf/compat/v1/train/polynomial_decay
+    /// Default values and implementation of learning rate is from Tensorflow Slim model tests.
+    /// Source : https://github.com/tensorflow/models/blob/master/research/slim/train_image_classifier.py
+    /// </summary>
+    public sealed class PolynomialLRDecay : LearningRateScheduler
+    {
+        public float LearningRate;
```
eerhardt: These should all be `readonly`. You don't want people changing these values once the object is created.

src/Microsoft.ML.Dnn/LearningRateScheduler.cs
```diff
@@ -192,4 +192,52 @@ internal override float GetLearningRate(TrainState trainstate)
...
+        public int DecaySteps;
+        public float EndLearningRate;
+        public float Power;
+        public bool Cycle;
+        public float NumEpochsPerDecay;
+
+        public int GlobalStep;
```
eerhardt: Should `GlobalStep` be public? What would a caller need to use it?

Actually, why is it a instance variable at all? Why isn't it just a local variable in `GetLearningRate`?

test/Microsoft.ML.Tests/LearningRateSchedulerTest.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
```
eerhardt: All product and test code files need a copyright header.

src/Microsoft.ML.Dnn/LearningRateScheduler.cs
```diff
@@ -195,4 +195,68 @@ internal override float GetLearningRate(TrainState trainstate)
         }
 
     }
```
codemzs: new line

test/Microsoft.ML.Tests/LearningRateSchedulerTest.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+        }
+
+        [Fact]
+        public void TestPolynomialDecayNoCycle()
+        {
+            //Values obtained by running tf.compat.v1.train.polynomial_decay on TF 1.14
+            float[] expectedValues = new float[] { 0.1f, 0.091f, 0.082f, 0.073f, 0.064f, 0.055f, 0.045999996f, 0.037f, 0.027999999f, 0.019000001f, 0.01f, 0.01f, 0.01f, 0.01f, 0.01f };
+            TestPolynomialDecay(expectedValues, false);
+        }
```
codemzs: new line

test/Microsoft.ML.Tests/LearningRateSchedulerTest.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.ML.RunTests;
+using Microsoft.ML.Transforms;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.ML.Tests
+{
+    public sealed class LearningRateSchedulerTest : TestDataPipeBase
```
codemzs: >LearningRateSchedulerTest [](start = 24, length = 25)

Can we also put a test with a ImageClassification API? I know it will increase the build time but I will prefer we have that test.

bpstark: Actually I disagree there is already a test that uses a LR scheduler which tests that the classification API can work with an LR scheduler we should not be duplicating that test to test with this scheduler, instead this is a much more correct unit test

---
In reply to: [339268046](https://github.com/dotnet/machinelearning/pull/4370#discussion_r339268046) [](ancestors = 339268046)

codemzs: Your code passes a learning rate to Image Classification API that is passed to TF graph runner and by adding an image classification test you would have added an e2e test. I appreciate your isolated test but I will prefer we have a unit-test that also test image classification API with this LR since our users will also be using it this way.

---
In reply to: [339268678](https://github.com/dotnet/machinelearning/pull/4370#discussion_r339268678) [](ancestors = 339268678,339268046)

bpstark: [https://github.com/dotnet/machinelearning/blob/b1788c36268999e66df00356417e604f08d2a568/test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs#L1485](https://github.com/dotnet/machinelearning/blob/b1788c36268999e66df00356417e604f08d2a568/test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs#L1485)

Already tests that a learning rate when passed to the image classification API is properly handled by the API, we should not be running the same test with a different LR scheduler that is not a useful test instead we need to confirm that this scheduler is producing the results we expect

---
In reply to: [339271171](https://github.com/dotnet/machinelearning/pull/4370#discussion_r339271171) [](ancestors = 339271171,339268678,339268046)

codemzs:
@bpstark My feelings on this: 
https://cdn-images-1.medium.com/max/1600/1*UtZzMT32fRMnSN-HmgiSVQ.gif

test/Microsoft.ML.Tests/LearningRateSchedulerTest.cs
codemzs: @eerhardt to comment further. My feelings: https://cdn-images-1.medium.com/max/1600/1*UtZzMT32fRMnSN-HmgiSVQ.gif

eerhardt: My opinion is that both levels of testing are important, and there are tradeoffs on which level you write your tests.

1. If all of your tests are unit tests, you can't tell if the components work when used together.
2. If all of your tests are functional tests, it is a lot harder to test different permutations of inputs/settings. Both trying to set all those settings set at the functional level, and in the amount of time it is going to take to execute end-to-end tests.

Having a good balance is crucial. It ensures that a component works correctly with all different inputs, but our tests don't become so long to run that they are no longer maintainable.

In this case, you can view the whole `LearningRateScheduler` system as the "component" or each individual derived class as the "component". This is why I think you two seem to be having a disagreement. It appears @codemzs views each individual derived class as the "component" that should have a functional test to ensure it works with the rest of the system. And @bpstark views the whole `LearningRateScheduler` system as the "component" that already does have a functional test, and so there is no need testing each individual subclass.

If I were doing this myself, I would see how long the existing functional test takes. If it takes so long that doubling the time would be unacceptable, I wouldn't add a new functional test. But if the existing functional test is fast enough that adding a second one wouldn't be noticeable, I would add it.

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1424,7 +1424,18 @@ public void TensorFlowImageClassification(ImageClassificationTrainer.Architectur
...
+            TensorFlowImageClassificationWithLRScheduling(new ExponentialLRDecay());
+        }
+
+        [TensorFlowFact]
+        public void TensorFlowImageClassificationWithPolynomialLRScheduling()
+        {
+            TensorFlowImageClassificationWithLRScheduling(new PolynomialLRDecay());
+        }
+
+        internal void TensorFlowImageClassificationWithLRScheduling(LearningRateScheduler  learningRateScheduler)
```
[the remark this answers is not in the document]
bpstark: talked offline theory is really for constants not classes, so it does not really work well for this case. will keep as is

PR #4372, opened 2019-10-24
Make ImageClassification API an ITrainerEstimator and refactor code.

### LINKED ISSUE #4276: Issues with ImageClassificationTransformer
- The Options class contains two string array fields: one for input column names and one for output column names. However, it always uses exactly two input columns: a label column and a features column, and it always outputs exactly two output columns: score and predicted label. This discrepancy should be addressed.
- The transformer has a field called `_labelColumnName` that is serialized and deserialized and otherwise does nothing. Since after deserialization we cannot rely on having data that has a label column anyway, this field should be removed.
- `GetOutputSchema()` in the estimator does not check that the type of the label column is correct (what other types, if any, should be allowed other than key type?), and neither does the constructor (it doesn't check the type of the features column either).
- I think the constructor of the transformer should not have a boolean flag to indicate whether the transformer is being instantiated from file or from the estimator. Instead, the estimator should do the required training (including figuring out the class count), and there should be one constructor for the transformer, that is called both from the estimator's `Fit()` method, and from the transformer's `Create(ModelLoadContext)` method.
- The estimator has a `_transformer` field that is not needed.
- The transformer ctor has an unused argument: `batchSize` - don't know if the error is in not using it or in not needing it.

codemzs:
In addition to converting ImageClassification API to ITrainerEstimator it also fixes #4276


src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -80,59 +81,53 @@ public static class DnnCatalog
...
         {
-            var options = new ImageClassificationEstimator.Options()
-            {
-                FeaturesColumnName = featuresColumnName,
-                LabelColumnName = labelColumnName,
-                ScoreColumnName = scoreColumnName,
-                PredictedLabelColumnName = predictedLabelColumnName,
-                ValidationSet = validationSet
-            };
+            options.EarlyStoppingCriteria = options.DisableEarlyStopping ? null : options.EarlyStoppingCriteria ?? new ImageClassificationTrainer.EarlyStopping();
```
eerhardt: I made a comment on the previous PR that you said you would address in this PR.

[https://github.com/dotnet/machinelearning/pull/4362#discussion_r338160348](https://github.com/dotnet/machinelearning/pull/4362#discussion_r338160348)

src/Microsoft.ML.Dnn/DnnCatalog.cs
```diff
@@ -80,59 +81,53 @@ public static class DnnCatalog
...
 
-            var env = CatalogUtils.GetEnvironment(catalog);
-            return new ImageClassificationEstimator(env, options, DnnUtils.LoadDnnModel(env, options.Arch, true));
+        public static ImageClassificationTrainer ImageClassification(
+            this MulticlassClassificationCatalog.MulticlassClassificationTrainers catalog,
+            string labelColumnName = DefaultColumnNames.Label,
+            string featureColumnName = DefaultColumnNames.Features,
+            string scoreColumnName = DefaultColumnNames.Score,
+            string predictedLabelColumnName = DefaultColumnNames.PredictedLabel,
+            IDataView validationSet = null)
```
eerhardt: I thought you were going to move the `validationSet` to be during the `Fit` method?

eerhardt:
>     public class ImageClassificationTrainer :

public `sealed` class ?

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs:68 in 9a01968. [](commit_id = 9a019684c5ab6216d623c1cb4254b0c56d35f0de, deletion_comment = False)

eerhardt:
>         public delegate void ImageClassificationMetricsCallback(ImageClassificationMetrics metrics);

Do we need a new delegate here? Or can we instead use `Action<ImageClassificationMetrics>` ?

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs:124 in 9a01968. [](commit_id = 9a019684c5ab6216d623c1cb4254b0c56d35f0de, deletion_comment = False)

eerhardt:
>         public sealed class Options : TrainerInputBaseWithLabel

(nit) xml doc comments on the class.

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs:336 in 9a01968. [](commit_id = 9a019684c5ab6216d623c1cb4254b0c56d35f0de, deletion_comment = False)

eerhardt:
>             public IDataView ValidationSet;

Same comment here about the ValidationSet being passed into `Fit`, like the other algorithms in ML.NET.

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs:426 in 9a01968. [](commit_id = 9a019684c5ab6216d623c1cb4254b0c56d35f0de, deletion_comment = False)

src/Microsoft.ML.Dnn/DnnCatalog.cs
codemzs: I have and it works well when the trainer is the last item in the pipeline but if it is not then the pipeline becomes nasty and you will need to save the model twice. Most people would want to add KeyToValue Transform after this trainer to convert predictedLabel to string type for example .

---
In reply to: [338737502](https://github.com/dotnet/machinelearning/pull/4372#discussion_r338737502) [](ancestors = 338737502)

codemzs:
>             public IDataView ValidationSet;

see my previous comment. We have an api for Fit that takes both train and validation set.

---
In reply to: [546056311](https://github.com/dotnet/machinelearning/pull/4372#issuecomment-546056311) [](ancestors = 546056311)

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTransform.cs:426 in 9a01968. [](commit_id = 9a019684c5ab6216d623c1cb4254b0c56d35f0de, deletion_comment = False)

src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs
```diff
@@ -499,7 +490,8 @@ public sealed class Options : TrainerInputBaseWithLabel
                 LabelColumnName = labelColumn,
                 ScoreColumnName = scoreColumn,
                 PredictedLabelColumnName = predictedLabelColumn,
-                ValidationSet = validationSet
+                ValidationSet = validationSet,
+                EarlyStoppingCriteria = new EarlyStopping()
```
eerhardt: I think the default value of `EarlyStoppingCriteria` on any `Options` instance should be a `new EarlyStopping()` object. Can you move this to be always the default on `Options`?

src/Microsoft.ML.Dnn/DnnUtils.cs
```diff
@@ -13,6 +13,7 @@
 using Microsoft.ML.Data;
 using Microsoft.ML.Runtime;
 using Tensorflow;
+using static Microsoft.ML.Dnn.ImageClassificationTrainer;
 using static Tensorflow.Binding;
 
 namespace Microsoft.ML.Transforms.Dnn
```
eerhardt: (nit) This namespace seems wrong since we have `Microsoft.ML.Dnn` now .

eerhardt: We have other public types in this assembly that we should re-evaluate their namespaces as well.

`Microsoft.ML.Transforms.LearningRateScheduler`.
`Microsoft.ML.Transforms.DnnModel`.

---
In reply to: [339194479](https://github.com/dotnet/machinelearning/pull/4372#discussion_r339194479) [](ancestors = 339194479)

eerhardt: I think `DnnModel` should be made `internal`. I don't see it exposed anywhere.

---
In reply to: [339195151](https://github.com/dotnet/machinelearning/pull/4372#discussion_r339195151) [](ancestors = 339195151,339194479)

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -500,13 +517,15 @@ private protected override void SaveCore(ModelSaveContext ctx)
...
+            ctx.SaveStringOrNull(_scoreColumn);
+            ctx.SaveStringOrNull(_predictedLabelColumn);
         }
 
         private static VersionInfo GetVersionInfo()
         {
             return new VersionInfo(
                 modelSignature: "MC  PRED",
-                verWrittenCur: 0x00010001, // Initial
+                verWrittenCur: 0x00010002, // Initial
```
eerhardt: It would be good to ensure we have a test explicitly testing the old format.

If necessary, this can be logged as an issue and added in a follow-up PR.

eerhardt: Typically the way this is written is like the following:

[https://github.com/dotnet/machinelearning/blob/7cb98932523774cf988c88a825c516a11c216444/src/Microsoft.ML.Transforms/Text/TokenizingByCharacters.cs#L79-L80](https://github.com/dotnet/machinelearning/blob/7cb98932523774cf988c88a825c516a11c216444/src/Microsoft.ML.Transforms/Text/TokenizingByCharacters.cs#L79-L80)

That way we see each version, and what was changed in each version.

eerhardt:
>             };

(nit) unnecessary semi-colon.

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs:311 in 72e0eee. [](commit_id = 72e0eeeadff30e722b10d24ba337fa0e17823263, deletion_comment = False)

eerhardt:
>                 new Span<float>(classifier.ClassProbabilities, 0, _classCount).CopyTo(editor.Values);

I think this should be changed to pass the `editor.Values` into the `classifier.Score` method. Then we only need to copy the values a single time - from TensorFlow's native memory directly into the VBuffer.

The way we have it today, we are holding this extra `float[] _classProbabilities` as an intermediate step, which isn't necessary.

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs:1427 in 72e0eee. [](commit_id = 72e0eeeadff30e722b10d24ba337fa0e17823263, deletion_comment = False)

eerhardt:
>             public void Score(VBuffer<byte> image)

(nit) should be `in VBuffer<byte> image`.

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs:1407 in 72e0eee. [](commit_id = 72e0eeeadff30e722b10d24ba337fa0e17823263, deletion_comment = False)

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
codemzs: [https://github.com/dotnet/machinelearning/issues/4389](https://github.com/dotnet/machinelearning/issues/4389)

---
In reply to: [339196504](https://github.com/dotnet/machinelearning/pull/4372#discussion_r339196504) [](ancestors = 339196504)

src/Microsoft.ML.Dnn/DnnUtils.cs
codemzs: Done. I was planning on doing a final API surface review and changing the namespace to Microsoft.ML.Vision in the end. 

---
In reply to: [339195669](https://github.com/dotnet/machinelearning/pull/4372#discussion_r339195669) [](ancestors = 339195669,339195151,339194479)

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -500,13 +517,16 @@ private protected override void SaveCore(ModelSaveContext ctx)
...
+            ctx.SaveStringOrNull(_scoreColumn);
+            ctx.SaveStringOrNull(_predictedLabelColumn);
         }
 
         private static VersionInfo GetVersionInfo()
         {
             return new VersionInfo(
                 modelSignature: "MC  PRED",
-                verWrittenCur: 0x00010001, // Initial
+                //verWrittenCur: 0x00010002, // Initial
```
eerhardt: (nit)

```
//verWrittenCur: 0x00010001, // Initial
```

codemzs: yep, I just addressed that in my latest commit.

---
In reply to: [339233591](https://github.com/dotnet/machinelearning/pull/4372#discussion_r339233591) [](ancestors = 339233591)

eerhardt:
>     public sealed class ImageClassificationModelParameters : ModelParametersBase<VBuffer<float>>, IValueMapper

xml docs on public types.

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs:1281 in ea79b6f. [](commit_id = ea79b6f0e84fd85f13e6c454914eee5a214fa0f6, deletion_comment = False)

eerhardt:
>         public DataViewType InputType => _inputType;

Do `InputType` and `OutputType` need to be public?

---
Refers to: src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs:1328 in ea79b6f. [](commit_id = ea79b6f0e84fd85f13e6c454914eee5a214fa0f6, deletion_comment = False)

PR #4374, opened 2019-10-24
Add SampleBase abstract class. 

Oceania2018:
1. Add SampleBase abstract class.
2. Make GetAbsolutePath() return be true absolute path.
image

image

Please let me know if you want other Samples to inherit from `SampleBase`.


codemzs:
@Oceania2018 thanks for your PR but we prefer our samples to be standalone, i.e users should be able to copy-paste code from the sample and run but with your change they will also need to copy the base class. Our samples appear in documentation hence we have this policy to make it easier to run samples, therefore you see a lot of repeated code in the samples. Appreciate your PR though.

Oceania2018:
OK, make sense.

PR #4375, opened 2019-10-24
Move windows build machines to NetCorePublic-Pool.

codemzs:
Builds seem to be failing because it runs out of disk space.

### DIFF
--- a/.vsts-dotnet-ci.yml
+++ b/.vsts-dotnet-ci.yml
@@ -66,6 +66,7 @@ jobs:
     buildScript: build.cmd
     pool:
       name: NetCorePublic-Pool
+      queue: buildpool.windows.10.amd64.vs2017.open
 
 - template: /build/ci/job-template.yml
   parameters:
@@ -82,6 +83,7 @@ jobs:
         _includeBenchmarkData: false
     pool:
       name: NetCorePublic-Pool
+      queue: buildpool.windows.10.amd64.vs2017.open
 
 - template: /build/ci/job-template.yml
   parameters:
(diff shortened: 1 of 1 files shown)


.vsts-dotnet-ci.yml
```diff
@@ -65,7 +65,7 @@ jobs:
     name: Windows_x64_NetCoreApp21
     buildScript: build.cmd
     pool:
-      name: Hosted VS2017
+      name: NetCorePublic-Pool
```
eerhardt: Do we need to say

```
queue: buildpool.windows.10.amd64.vs2017.open
```

like we have above?

codemzs: what about for x86?

---
In reply to: [338675654](https://github.com/dotnet/machinelearning/pull/4375#discussion_r338675654) [](ancestors = 338675654)

eerhardt: We run our `x86` build on a 64-bit machine/OS. We just use a 32-bit `dotnet.exe` to do the build and run the tests.

codemzs: thanks. I have the made the change.

.vsts-dotnet-ci.yml
```diff
@@ -65,7 +65,8 @@ jobs:
...
+      queue: buildpool.windows.10.amd64.vs2017.open
```
safern: Should we instead build on VS2019?

eerhardt: It probably doesn't matter either way. The only thing we use the VS on the machine for is the C++ compiler.

But if we wanted to update all of them to vs2019, that would be OK with me. But I don't think it is 100% required.

PR #4377, opened 2019-10-24
Delete build files before packaging and increase build timeout

codemzs:
This reverts commit c922529e669d1c4dcb4d2bf8157a579b10a60cee and deletes build files before packaging to get rid of out of memory error. It also increases the build timeout to 60 minutes.

### DIFF
--- a/build/ci/job-template.yml
+++ b/build/ci/job-template.yml
@@ -10,7 +10,7 @@ parameters:
 jobs:
   - job: ${{ parameters.name }}
     ${{ if eq(parameters.codeCoverage, 'false') }}:
-      timeoutInMinutes: 40
+      timeoutInMinutes: 60
     ${{ if eq(parameters.codeCoverage, 'true') }}:
       timeoutInMinutes: 60
     strategy:
--- a/build/ci/job-template.yml
+++ b/build/ci/job-template.yml
@@ -77,5 +77,35 @@ jobs:
         pathToPublish: $(Build.ArtifactStagingDirectory)
         artifactName: ${{ parameters.name }} $(_config_short)
         artifactType: container
+    - ${{ if eq(parameters.buildScript, 'build.cmd') }}:
+      - powershell: |
+                  $taskListOutput = & TaskList /FI "IMAGENAME eq dotnet.exe" | Out-String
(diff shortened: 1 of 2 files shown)


eerhardt:
Do we know what the issue is here?

safern:
>  we package that deletes files no longer need to free up disk space this was suggested by @safern

I just added a commit here https://github.com/dotnet/machinelearning/pull/4377/commits/d3742e4cce8b96e46f46c4a0f0f208226d7d3f40 to add a step to clean-up before building packages on Windows. Later we can add the step for Unix as the script would be different and didn't want to invest much time on it now.

I did some local investigation. After building, downloading test data and running tests the repo size grows to 16GB, after the clean-up it goes down to 1.5GBs which I believe is enough to build packages without hitting the space issue.

Also I did a test build and it didn't repro anymore, however there where 2 timeouts... I think bumping the timeout to 50mins shouldn't hurt.

build/ci/job-template.yml
```diff
@@ -10,7 +10,7 @@ parameters:
 jobs:
   - job: ${{ parameters.name }}
     ${{ if eq(parameters.codeCoverage, 'false') }}:
-      timeoutInMinutes: 40
+      timeoutInMinutes: 60
```
eerhardt: We should really be careful about letting our builds take so long. This isn't a huge project, and having an hour CI time doesn't seem great. Maybe someone should investigate what is taking so long in our builds.

codemzs: Eric, I will revert this back after the release. We have a huge backlog of PRs that is threatening the release.

---
In reply to: [339138790](https://github.com/dotnet/machinelearning/pull/4377#discussion_r339138790) [](ancestors = 339138790)

eerhardt: Agreed - we need to get unblocked in the short term. I meant to say that someone should investigate it after this gets us unblocked.

codemzs: I suspect the reason our build times went up is because of newly added TensorFlow tests that each take easily a minute to run. We can try to reduce the dataset size but obviously after this release.

CC: @ashbhandare @harshithapv @bpstark

PR #4384, opened 2019-10-25
Integrate Image Classification API in AutoML

codemzs:
This change also contains ImageClassification API change to convert the API to ITrainerEstimator as it has not been checked-in yet but it is only temporary and I will remove it before checking this in change into master branch. Code gen changes will come in a different PR.


src/Microsoft.ML.AutoML/API/ExperimentBase.cs
```diff
@@ -102,16 +102,17 @@ public abstract class ExperimentBase<TMetrics, TExperimentSettings>
             const int crossValRowCountThreshold = 15000;
 
             var rowCount = DatasetDimensionsUtil.CountRows(trainData, crossValRowCountThreshold);
+            bool isNotImageClassificationTask = columnInformation.ImagePathColumnNames.Count != 1;
 
-            if (rowCount < crossValRowCountThreshold)
+            if (rowCount < crossValRowCountThreshold && isNotImageClassificationTask)
```
justinormont: You'll need CrossVal when your count of images is low. Otherwise noise will dominate differences in your model metrics.

I assume ImageClassification will work in CV and you're just worried about runtime. Perhaps there's a more runtime sensitive heuristic that we could make.

src/Microsoft.ML.AutoML/EstimatorExtensions/EstimatorExtensions.cs
```diff
@@ -256,17 +257,45 @@ public IEstimator<ITransformer> CreateInstance(MLContext context, PipelineNode p
...
         {
             var pipelineNode = new PipelineNode(EstimatorName.ValueToKeyMapping.ToString(),
                 PipelineNodeType.Transform, inColumn, outColumn);
+            var estimator = CreateInstance(context, inColumn, outColumn, sort);
+            return new SuggestedTransform(pipelineNode, estimator);
+        }
+
+        private static IEstimator<ITransformer> CreateInstance(MLContext context, string inColumn, string outColumn, bool sort = false)
+        {
+            if(sort)
```
justinormont: nit: space
```suggestion
            if (sort)
```

src/Microsoft.ML.AutoML/TrainerExtensions/TrainerExtensionCatalog.cs
```diff
@@ -58,14 +59,21 @@ public static ITrainerExtension GetTrainerExtension(TrainerName trainerName)
...
-            IEnumerable<TrainerName> whitelist)
+            IEnumerable<TrainerName> whitelist, ColumnInformation columnInfo)
         {
             IEnumerable<ITrainerExtension> trainers;
             if (task == TaskKind.BinaryClassification)
             {
                 trainers = GetBinaryLearners();
             }
-            else if (task == TaskKind.MulticlassClassification)
+            else if (task == TaskKind.MulticlassClassification && columnInfo.ImagePathColumnNames.Count == 1)
```
justinormont: Should explicitly check others are empty:
```suggestion
            else if (task == TaskKind.MulticlassClassification && columnInfo.ImagePathColumnNames.Count == 1 &&
                columnInfo.CategoricalColumnNames.Count == 0 && columnInfo.NumericColumnNames.Count == 0 &&
                columnInfo.TextColumnNames.Count == 0)
```

This more specifically checks what the ImageClassifier can take in. The fall through is normal multi-class. We should add image handling there. 

I'd recommend two options for the fall-thru case: (when there are images+text/numerics/categoricals)
1. Returning the penultimate layer from the `ImageClassification` trainer (running as a featurizer)
2. Running as the 1st model in a stack ensemble (returning the `score` column as features)

Other tasks, like regression, etc still need image support. Option 1, above, won't cover it as `ImageClassification` requires a classification label. Perhaps a label-less, no fine-tuning, featurizer mode? Basically a replacement for DNNImageFeaturizer.

src/Microsoft.ML.AutoML/TrainerExtensions/TrainerExtensionCatalog.cs
```diff
@@ -58,14 +59,21 @@ public static ITrainerExtension GetTrainerExtension(TrainerName trainerName)
...
+            {
+                // Image Classification case where all you have is a label column and image column.
+                // This trainer takes features column of vector of bytes which will not work with any
+                // other trainer.
+                return new List<ITrainerExtension>() { new ImageClassificationExtension() };
+            }
+            else if(task == TaskKind.MulticlassClassification)
```
justinormont: nit: space
```suggestion
            else if (task == TaskKind.MulticlassClassification)
```

src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
```diff
@@ -47,9 +47,12 @@ internal static class SplitUtil
         /// Split the data into a single train/test split.
         /// </summary>
         public static (IDataView trainData, IDataView validationData) TrainValidateSplit(MLContext context, IDataView trainData,
-            string samplingKeyColumn)
+            string samplingKeyColumn, bool shuffle = false)
         {
             var originalColumnNames = trainData.Schema.Select(c => c.Name);
+            if (shuffle)
+                trainData = context.Data.ShuffleRows(trainData);
```
justinormont: Any particular reason to shuffle here? Seems like an odd place to add a shuffle. If for no other reason you're also shuffling the validation split. There's another code path for the dataset splits where the user can provide the train/validation split directly; in this case your shuffle will not run.

I assume you're looking for the ImageClassification trainer to receive shuffled data. It's likely you'll want the shuffle right before your trainer. Alternatively it should be part of the trainer itself (adding it itself) w/ the normal `Shuffle = true` hyperparameter. Perhaps a `RowShufflingTransformer` just before sending data to TF.

src/Microsoft.ML.Dnn/DnnUtils.cs
```diff
@@ -266,19 +267,12 @@ internal static DnnModel LoadDnnModel(IHostEnvironment env, string modelPath, bo
...
+        internal static DnnModel LoadDnnModel(IHostEnvironment env, Architecture arch, bool metaGraph = false)
         {
-            var modelPath = ImageClassificationEstimator.ModelLocation[arch];
+            var modelPath = ModelLocation[arch];
             if (!File.Exists(modelPath))
             {
-                if (arch == ImageClassificationEstimator.Architecture.InceptionV3)
+                if (arch == Architecture.InceptionV3)
                 {
                     var baseGitPath = @"https://raw.githubusercontent.com/SciSharp/TensorFlow.NET/master/graph/InceptionV3.meta";
```
justinormont: Without pinning to a specific git commit, the referenced file can be move/changed causing a breakage.  
```suggestion
                    var baseGitPath = @"https://raw.githubusercontent.com/SciSharp/TensorFlow.NET/547c4e6bf48fceb8266d0cb88d3a13ec4171606a/graph/InceptionV3.meta";
```

src/Microsoft.ML.Dnn/DnnUtils.cs
```diff
@@ -293,23 +287,23 @@ internal static DnnModel LoadDnnModel(IHostEnvironment env, ImageClassificationE
                         ZipFile.ExtractToDirectory(Path.Combine(Directory.GetCurrentDirectory(), @"tfhub_modules.zip"), @"tfhub_modules");
                     }
                 }
-                else if (arch == ImageClassificationEstimator.Architecture.ResnetV2101)
+                else if (arch == Architecture.ResnetV2101)
                 {
                     var baseGitPath = @"https://aka.ms/mlnet-resources/image/ResNet101Tensorflow/resnet_v2_101_299.meta";
                     using (WebClient client = new WebClient())
```
justinormont: Recommend using Yael's ResourceManager. 

Example from the WordEmbedding transform:
https://github.com/dotnet/machinelearning/blob/5163413f3cfab430e49f6fb2a690c160fde46b58/src/Microsoft.ML.Transforms/Text/WordEmbeddingsExtractor.cs#L631-L642

src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs
```diff
@@ -11,1598 +7,1456 @@
...
-            File.WriteAllBytes(_checkpointPath + ".pb", outputGraphDef.ToByteArray());
-            _session.graph.Dispose();
-            _session.Dispose();
-            _session = LoadTFSessionByModelFilePath(_env, frozenModelPath, false);
-        }
+            /// <summary>
+            /// Number of training iterations.
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Number of training iterations.", SortOrder = 10)]
+            public int Epoch = 100;
```
justinormont: You may want to align  with naming conventions of ML.NET instead of TF. This is called `numberOfIterations` in other ML.NET trainers.

src/Microsoft.ML.AutoML/EstimatorExtensions/EstimatorExtensions.cs
```diff
@@ -256,17 +257,45 @@ public IEstimator<ITransformer> CreateInstance(MLContext context, PipelineNode p
...
+            };
+            var pipelineNode = new PipelineNode(EstimatorName.ImageLoading.ToString(), PipelineNodeType.Transform, inColumn, outColumn, pipelineNodeProperty);
             var estimator = CreateInstance(context, inColumn, outColumn);
             return new SuggestedTransform(pipelineNode, estimator);
         }
 
         private static IEstimator<ITransformer> CreateInstance(MLContext context, string inColumn, string outColumn)
         {
-            return context.Transforms.Conversion.MapValueToKey(outColumn, inColumn);
+            return context.Transforms.LoadImages(outColumn, ImageFolder, false, inColumn);
```
eerhardt: suggestion: When passing hard-coded boolean parameters to a method, it is nice to use a named parameter, so the reader can tell what `false` means here.

```C#
return context.Transforms.LoadImages(outColumn, ImageFolder, useImageType: false, inColumn);
```

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -35,6 +37,126 @@ public static IDataView GetUciAdultDataView()
...
+                file.WriteLine("Label,ImagePath");
+                foreach (var image in images)
+                    file.WriteLine(image.Label + "," + image.ImagePath);
+            }
+
+            return datasetName;
+        }
+
+        public static IEnumerable<ImageData> LoadImagesFromDirectory(string folder,
+            bool useFolderNameAsLabel = true)
```
eerhardt: You can remove `useFolderNameAsLabel` since it is always `true`.

src/Microsoft.ML.AutoML/EstimatorExtensions/EstimatorExtensions.cs
```diff
@@ -256,17 +257,45 @@ public IEstimator<ITransformer> CreateInstance(MLContext context, PipelineNode p
...
+            if(sort)
+                return context.Transforms.Conversion.MapValueToKey(outColumn, inColumn, keyOrdinality:ValueToKeyMappingEstimator.KeyOrdinality.ByValue);
+
+            return context.Transforms.Conversion.MapValueToKey(outColumn, inColumn);
+        }
+    }
+
+    internal class ImageLoadingExtension : IEstimatorExtension
+    {
+        public static string ImageFolder { private get; set; }
```
LittleLittleCloud: Where do we Initialize `ImageFolder`

Or the image dataset has to be full_path?

docs/samples/Microsoft.ML.Samples/Dynamic/Trainers/MulticlassClassification/ImageClassification/ImageClassificationDefault.cs
```diff
@@ -60,12 +59,11 @@ public static void Example()
                 IDataView trainDataset = trainTestData.TrainSet;
                 IDataView testDataset = trainTestData.TestSet;
 
-                var pipeline = mlContext.Model.ImageClassification("Image", "Label", validationSet: testDataset)
-                    .Append(mlContext.Transforms.Conversion.MapKeyToValue(
-                        outputColumnName: "PredictedLabel", 
+                var pipeline = mlContext.MulticlassClassification.Trainers
+                        .ImageClassification(featureColumnName:"Image", validationSet:testDataset)
+                    .Append(mlContext.Transforms.Conversion.MapKeyToValue(outputColumnName: "PredictedLabel",
```
daholste: nit: indent

src/Microsoft.ML.AutoML/API/ExperimentBase.cs
```diff
@@ -102,16 +102,17 @@ public abstract class ExperimentBase<TMetrics, TExperimentSettings>
             const int crossValRowCountThreshold = 15000;
 
             var rowCount = DatasetDimensionsUtil.CountRows(trainData, crossValRowCountThreshold);
+            bool isNotImageClassificationTask = columnInformation.ImagePathColumnNames.Count != 1;
```
daholste: nit: isImageClassificationTask may be a simpler name?

daholste: or hasImagePathColumns. because I suppose a classification task could have image path columns

src/Microsoft.ML.AutoML/API/ExperimentBase.cs
daholste: also, could you leave a comment about the rationale for the decision?

src/Microsoft.ML.AutoML/TransformInference/TransformInference.cs
```diff
@@ -338,6 +342,32 @@ public override IEnumerable<SuggestedTransform> Apply(IntermediateColumn[] colum
...
+                {
+                    var featureCols = new List<string>();
+
+                    foreach (var column in columns)
+                    {
+                        if (!column.Type.GetItemType().IsText() || column.Purpose != ColumnPurpose.ImagePath)
+                            continue;
+
+                        var columnDestSuffix = "_featurized";
+                        var columnNameSafe = column.ColumnName;
```
daholste: why is this column name safe?

src/Microsoft.ML.AutoML/TransformInference/TransformInference.cs
```diff
@@ -338,6 +342,32 @@ public override IEnumerable<SuggestedTransform> Apply(IntermediateColumn[] colum
...
+
+                        string columnDestRenamed = $"{columnNameSafe}{columnDestSuffix}";
+
+                        featureCols.Add(columnDestRenamed);
```
daholste: where is featureCols used?

src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
```diff
@@ -47,9 +47,12 @@ internal static class SplitUtil
         /// Split the data into a single train/test split.
         /// </summary>
         public static (IDataView trainData, IDataView validationData) TrainValidateSplit(MLContext context, IDataView trainData,
-            string samplingKeyColumn)
+            string samplingKeyColumn, bool shuffle = false)
```
daholste: Perhaps conceptually 'shuffle' belongs outside of the train/valid split API ?
If desired (and only desired for images right now), you could shuffle the results of the this API
For instance, I don't think it's part of the ML.NET train/valid split API

test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
```diff
@@ -45,6 +46,22 @@ public void AutoFitMultiTest()
             Assert.Equal(NumberDataViewType.Single, scoredData.Schema[DefaultColumnNames.PredictedLabel].Type);
         }
 
+        [TensorFlowFact]
+        public void AutoFitImageClassificationTest()
```
daholste: awesome!

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -35,6 +37,126 @@ public static IDataView GetUciAdultDataView()
...
+            //Download the image set and unzip
+            string finalImagesFolderName = DownloadImageSet(
+                imagesDownloadFolderPath);
+
+            string fullImagesetFolderPath = Path.Combine(
+                imagesDownloadFolderPath, finalImagesFolderName);
+
+            var images = LoadImagesFromDirectory(
+                folder: fullImagesetFolderPath, useFolderNameAsLabel: true);
+
```
daholste: nit: two empty lines

src/Microsoft.ML.AutoML/EstimatorExtensions/EstimatorExtensions.cs
```diff
@@ -256,17 +257,45 @@ public IEstimator<ITransformer> CreateInstance(MLContext context, PipelineNode p
             return CreateInstance(context, pipelineNode.InColumns[0], pipelineNode.OutColumns[0]);
         }
 
-        public static SuggestedTransform CreateSuggestedTransform(MLContext context, string inColumn, string outColumn)
+        public static SuggestedTransform CreateSuggestedTransform(MLContext context, string inColumn, string outColumn, bool sort = false)
```
daholste: out of curiosity, why sort for images?

src/Microsoft.ML.AutoML/TrainerExtensions/MultiTrainerExtensions.cs
```diff
@@ -229,4 +229,22 @@ public PipelineNode CreatePipelineNode(IEnumerable<SweepableParam> sweepParams,
                 columnInfo.LabelColumnName, columnInfo.ExampleWeightColumnName);
         }
     }
+
+    internal class ImageClassificationExtension : ITrainerExtension
```
daholste: you may want to make sure that the CLI still runs / doesn't break if someone has image paths in their column names?

src/Microsoft.ML.AutoML/API/ExperimentBase.cs
codemzs: Hey Justin, it is not about runtime, I just haven't seen anywhere on Tensorflow examples where they used CV for image classification and my other rational was that with CV is that it does not shuffle the data which means you could have all the images of cats, then all the images of dogs and with CV you might train on just cats and then test on dogs. I will try to add shuffling to CV as well so that it trains on all classes in every split.  

---
In reply to: [338960267](https://github.com/dotnet/machinelearning/pull/4384#discussion_r338960267) [](ancestors = 338960267)

src/Microsoft.ML.AutoML/EstimatorExtensions/EstimatorExtensions.cs
codemzs: great question, when you split the dataset and then apply ValueToKeyMapping after the split then it could result in different key to value mapping for the same class. I saw some cases where this was being done in the code base so for image classification I just want to be sure the mappings is consistent else it can degrade the accuracy.

---
In reply to: [339251823](https://github.com/dotnet/machinelearning/pull/4384#discussion_r339251823) [](ancestors = 339251823)

src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
codemzs: We need to shuffle otherwise you can train on set of images and test on set of images that you did not train on, images need to be shuffled before they are fed into the trainer. Think of a dataset that has images of cats followed by images of dogs, now whether you do CV or TrainTest, you will end up training on cats and testing on dogs and get a very poor accuracy.

---
In reply to: [339248471](https://github.com/dotnet/machinelearning/pull/4384#discussion_r339248471) [](ancestors = 339248471)

src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
codemzs: We need to shuffle otherwise you can train on set of images and test on another set of images that you did not train on, images need to be shuffled before they are fed into the trainer. Think of a dataset that has images of cats followed by images of dogs, now whether you do CV or TrainTest, you will end up training on cats and testing on dogs and get a very poor accuracy.

---
In reply to: [338993257](https://github.com/dotnet/machinelearning/pull/4384#discussion_r338993257) [](ancestors = 338993257)

src/Microsoft.ML.Dnn/DnnUtils.cs
codemzs: I have removed this change as part of this PR but in this case we plan to move this to Azure Blob

---
In reply to: [338996402](https://github.com/dotnet/machinelearning/pull/4384#discussion_r338996402) [](ancestors = 338996402)

src/Microsoft.ML.Dnn/DnnUtils.cs
codemzs: removed this change as part of PR but good recommendation will add that later.

---
In reply to: [338999752](https://github.com/dotnet/machinelearning/pull/4384#discussion_r338999752) [](ancestors = 338999752)

src/Microsoft.ML.AutoML/API/ExperimentBase.cs
justinormont: Shuffling is not required for CV to work properly. And CV is required when your number of rows is small, to get a useful measure of how good your model performs.

I think you're visualizing the CV folds as created as `folds =  [ rows 0 ... 99, rows 100...199, rows 200..299 ]`. Instead of this method, the rows are split pseudorandomly.

CV splits the dataset with each row going into one of K folds pseudorandomly. Then each of the K folds are chosen one at a time as the test dataset, and the other K-1 folds are used as the training set. 

The cats and dogs will be roughly equally split among the CV folds. That said, you may want a shuffle as part of your trainer, to lessen overfitting.

src/Microsoft.ML.AutoML/EstimatorExtensions/EstimatorExtensions.cs
justinormont: Why does the mapping being inconsistent affect the model's accuracy? 

You'll run into inconsistent label mappings rather often, specifically the mapping can ends up different each time it runs on a different split of the dataset.

For instance every model you create, the dataset will be split slightly differently in TrainValidate mode, or within CV. One ValueToKeyMapping run will see 100 classes, another will see 99.

src/Microsoft.ML.AutoML/Microsoft.ML.AutoML.csproj
```diff
@@ -8,10 +8,14 @@
...
+    <ProjectReference Include="..\Microsoft.ML.Dnn\Microsoft.ML.Dnn.csproj" />
+    <ProjectReference Include="..\Microsoft.ML.ImageAnalytics\Microsoft.ML.ImageAnalytics.csproj" />
     <ProjectReference Include="..\Microsoft.ML.LightGbm\Microsoft.ML.LightGbm.csproj" />
     <ProjectReference Include="..\Microsoft.ML.Mkl.Components\Microsoft.ML.Mkl.Components.csproj" />
     <ProjectReference Include="..\Microsoft.ML.Recommender\Microsoft.ML.Recommender.csproj" />
     <ProjectReference Include="..\Microsoft.ML.Transforms\Microsoft.ML.Transforms.csproj" />
   </ItemGroup>
-
+  <ItemGroup>
+    <PackageReference Include="SciSharp.TensorFlow.Redist" Version="$(TensorFlowVersion)" />
```
justinormont: Would it be possible for the ImageClassifier to have a GPU setting? 

`UseGPU = { Auto, On, Off }`

Options.
* `Auto` will try the GPU and fail over to the CPU (default)
* `On` will use the GPU the throw if not possible. 
* `Off` will use only the CPU.

src/Microsoft.ML.AutoML/EstimatorExtensions/EstimatorExtensions.cs
codemzs: I’ll explain offline.

src/Microsoft.ML.AutoML/TrainerExtensions/MultiTrainerExtensions.cs
justinormont: The main issue I expect is needing to add the right package references to the CLI. Otherwise the CLI builds, and runs correctly from VS, but the packaged/shipped CLI fails at runtime.

Need to add the new packages used to `mlnet.nupkgproj`. 

Thanks to @eerhardt for his help in debugging the CLI when I was building this.

src/Microsoft.ML.AutoML/TrainerExtensions/MultiTrainerExtensions.cs
```diff
@@ -229,4 +229,22 @@ public PipelineNode CreatePipelineNode(IEnumerable<SweepableParam> sweepParams,
...
+    {
+        public IEnumerable<SweepableParam> GetHyperparamSweepRanges() => null;
```
justinormont: Should at minimum sweep over the model architectures and possibly additional impactful hyperparameters.

Possible useful sweep targets:
* Model architecture (or [domain adapted models](https://en.wikipedia.org/wiki/Domain_adaptation))
* Layer training schedule, like only last layer, last N layers, or 100 epochs of last layer and 10 epochs of all layers
* NumIterations (epochs)
* Learning rate schedule

All of the above should have an Auto option to keep the current heuristics.

src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
justinormont: See comment above on the non-need of shuffling for TrainValidate/CV splits: https://github.com/dotnet/machinelearning/pull/4384/files#r339313039

src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
justinormont: See comment above on the non-need of shuffling for TrainValidate/CV splits: https://github.com/dotnet/machinelearning/pull/4384/files#r339313039

codemzs:
@daholste @justinormont @eerhardt I'm closing this PR and opening a new PR with code gen change and comments addressed from this PR. Please review #4395

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -35,6 +37,126 @@ public static IDataView GetUciAdultDataView()
...
+        }
+
+        public static IEnumerable<ImageData> LoadImagesFromDirectory(string folder,
+            bool useFolderNameAsLabel = true)
+        {
+            var files = Directory.GetFiles(folder, "*",
+                searchOption: SearchOption.AllDirectories);
+            foreach (var file in files)
+            {
+                if (Path.GetExtension(file) != ".jpg")
```
justinormont: Incase someone sees and replicates this code...

Recommend: .jpg, .jpeg, .png, .gif. 
And case insensitive. 
```suggestion
                var extension = Path.GetExtension(file).ToLower();
                if (extension != ".jpg" &&
                    extension != ".jpeg" &&
                    extension != ".png" &&
                    extension != ".gif"
                )
```

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -35,6 +37,126 @@ public static IDataView GetUciAdultDataView()
...
+                else
+                {
+                    for (int index = 0; index < label.Length; index++)
+                    {
+                        if (!char.IsLetter(label[index]))
+                        {
+                            label = label.Substring(0, index);
+                            break;
+                        }
+                    }
```
justinormont: nit: (as it's only part of the test setup code)

I wouldn't recommend supporting auto-labels from the filenames. If auto-inferring the labels, I'd stick with only the folder names.

I think this is trying to catch a dataset named as `cat_image101.jpg`, `dog_image341.jpg` into:
```
Label,ImagePath
cat,c:\path\cat_image101.jpg
dog,c:\path\dog_iamge341.jpg
```

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -35,6 +37,126 @@ public static IDataView GetUciAdultDataView()
...
+            }
+        }
+
+        public static string DownloadImageSet(string imagesDownloadFolder)
+        {
+            string fileName = "flower_photos_tiny_set_for_unit_tests.zip";
+            string url = $"https://mlnetfilestorage.file.core.windows.net/imagesets" +
+                $"/flower_images/flower_photos_tiny_set_for_unit_tests.zip?st=2019" +
+                $"-08-29T00%3A07%3A21Z&se=2030-08-30T00%3A07%3A00Z&sp=rl&sv=2018" +
+                $"-03-28&sr=f&sig=N8HbLziTcT61kstprNLmn%2BDC0JoMrNwo6yRWb3hLLag%3D";
```
justinormont: Recommend linking to aka.ms links, which point to the CDN, which point to public folders in the blobstore.

Linking to aka.ms allows us to redirect the links in the future incase we decide to move them (without breaking old versions. Linking to the CDN gives us speed, and hides the actual blobstore backing the CDN.

As above, Yael's downloader class will help with the downloading process.

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -35,6 +37,126 @@ public static IDataView GetUciAdultDataView()
...
+            Download(url, imagesDownloadFolder, fileName);
+            UnZip(Path.Combine(imagesDownloadFolder, fileName), imagesDownloadFolder);
+
+            return Path.GetFileNameWithoutExtension(fileName);
+        }
+
+        private static bool Download(string url, string destDir, string destFileName)
+        {
+            if (destFileName == null)
+                destFileName = url.Split(Path.DirectorySeparatorChar).Last();
```
justinormont: This will fail on Windows (and succeed on Linux/macOS) as the Separator char is '\' instead of '/'. Also should handle parsing the '?' and '#' of URLs, along with unescaping chars.

This should work:
```suggestion
                destFileName = Path.GetFileName(new Uri(url).AbsolutePath);
```

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -35,6 +37,126 @@ public static IDataView GetUciAdultDataView()
...
+
+            Directory.CreateDirectory(destDir);
+
+            string relativeFilePath = Path.Combine(destDir, destFileName);
+
+            if (File.Exists(relativeFilePath))
+                return false;
+
+            new WebClient().DownloadFile(url, relativeFilePath);
```
justinormont: When writing a downloader for large files, it's helps to download to a temp filename and do an atomic move once finished. Otherwise partially downloaded files will never be retried.

We worked thru similar issues in @yaeldekel's downloader class. I'd recommend it.

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -35,6 +37,126 @@ public static IDataView GetUciAdultDataView()
...
+            return true;
+        }
+
+        private static void UnZip(String gzArchiveName, String destFolder)
+        {
+            var flag = gzArchiveName.Split(Path.DirectorySeparatorChar)
```
justinormont: Path. GetFileNameWithoutExtension is made for this. 
Instead of `gzArchiveName.Split(Path.DirectorySeparatorChar).Last().Split('.').First() + ".bin"`
```suggestion
            var flag = Path.GetFileNameWithoutExtension(gzArchiveName) + ".bin"
```

justinormont: Also, `flag` is a bit ambiguous of a name. Perhaps `completionIndicatorFileName`.

src/Microsoft.ML.AutoML/API/ExperimentBase.cs
justinormont: Also, the `ShuffleRows()` does not do the expected style of shuffling. 

Most people would assume `ShuffleRows()` shuffles the dataset. Instead it does a block shuffle, where continuous segments of the dataset are shuffled only within that block/partition. Aka, rows at the top of the dataset will remain near the top; and rows at the bottom will remain near the bottom. Shuffling a dataset normally implies a row will have equal likelihood of ending anywhere.

The `ShuffleRows()` block shuffle allows for a streaming mode by taking only the block size (~1k rows) of memory at a time. There's no general way support full shuffling of dataset in a streaming pipeline.

justinormont:
> @daholste @justinormont @eerhardt I'm closing this PR and opening a new PR with code gen change and comments addressed from this PR. Please review #4395

Ahh great. Saw this a bit late. See a few new review comments above.

src/Microsoft.ML.AutoML/TrainerExtensions/TrainerExtensionCatalog.cs
justinormont: Thanks for the change. I see it in the new PR.

PR #4393, opened 2019-10-26
Fixes #4392 Add AddPredictionEnginePool overload for implementation factory

### LINKED ISSUE #4392: Cannot use middleware dependencies when using custom ModelLoader on AddPredictionEnginePool
If a custom `ModelLoader` is used for the `PredictionEnginePool` that has a dependency on a configured service, there is no possibility to add this dependency via `AddPredictionEnginePool` as it does not offer an overload that accepts an implementation factory, unlike `AddSingleton`, `AddScoped` and `AddTransient`.

This is how it should be expected to happen in `Startup`:

```
services.AddPredictionEnginePool<Foo, Bar>(serviceProvider =>
{
    services.AddOptions<PredictionEnginePoolOptions<Foo, Bar>>().Configure(options =>
    {
        options.ModelLoader = new MyModelLoader(serviceProvider.GetService<IMyService>());
    });
    return new PredictionEnginePool<Foo, Bar>();
});
```

Since this is not possible with the current code, the creation of a model via middleware is therefore very difficult.  The only option is to use `ServiceProvider.BuildServiceProvider` which causes a warning as it results in an additional copy of singleton services being created:

```
services.AddPredictionEnginePool<Foo, Bar>()
services.AddOptions<PredictionEnginePoolOptions<Foo, Bar>>().Configure(options =>
{
    var serviceProvider = services.BuildServiceProvider();
    options.ModelLoader = new MyModelLoader(serviceProvider.GetService<IMyService>());
});
```

A new `AddPredictionEnginePool` overload should be added that accepts an implementation factory delegate to allow middleware services to be accessed via the `IServiceProvider`.

NeoXtreem:
This is an additional overload for `AddPredictionEnginePool` to allow middleware dependencies to be used in a custom `ModelLoader`.  This will allow the following code to be used in `Startup` in a similar manner to the corresponding overloads on `AddSingleton`, `AddScoped` and `AddTransient`:

```cs
services.AddPredictionEnginePool<Foo, Bar>(serviceProvider =>
{
    services.AddOptions<PredictionEnginePoolOptions<Foo, Bar>>().Configure(options =>
    {
        options.ModelLoader = new MyModelLoader(serviceProvider.GetService<IMyService>());
    });
    return new PredictionEnginePool<Foo, Bar>();
});
```

This fixes issue #4392, and is related to a [answer to a question](https://stackoverflow.com/a/57698799/197591) on Stack Overflow given by @eerhardt.


src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
```diff
@@ -19,27 +20,50 @@ public static class ServiceCollectionExtensions
         /// <param name="services">
         /// The <see cref="IServiceCollection "/> to add services to.
         /// </param>
+        /// <param name="implementationFactory">
+        /// The factory that creates the service.
```
eerhardt: (nit) "the service" is a little vague here. Would be better to say `that creates the <see cref="PredictionEnginePoolBuilder{TData, TPrediction}"/>`.

NeoXtreem: I actually just copied this from the corresponding parameter in `AddSingleton` for consistency. Take a look there, and let me know if you still think it should be changed.

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
```diff
@@ -19,27 +20,50 @@ public static class ServiceCollectionExtensions
...
-            services.AddOptions();
-            services.TryAddEnumerable(ServiceDescriptor.Singleton<IPostConfigureOptions<MLOptions>, PostMLContextOptionsConfiguration>());
-            services.AddSingleton<PredictionEnginePool<TData, TPrediction>, PredictionEnginePool<TData, TPrediction>>();
+            services
+                .AddPrerequisiteServices()
+                .AddSingleton<PredictionEnginePool<TData, TPrediction>, PredictionEnginePool<TData, TPrediction>>();
+            return new PredictionEnginePoolBuilder<TData, TPrediction>(services);
+        }
+
+        private static IServiceCollection AddPrerequisiteServices(this IServiceCollection services)
```
NeoXtreem: @eerhardt I'm considering making this method public and perhaps renaming it to something like `AddPredictionEnginePoolServices` or `AddPredictionEnginePoolPrerequisites` or `AddPredictionEnginePoolPrerequisiteServices` (or whatever you think is appropriate) because I'm in a situation where I need finer control on the registration.

Essentially, I want to create two `PredictionEnginePool`s and each one is using a custom `ModelLoader` as per the [SO question](https://stackoverflow.com/a/57698799/197591) this PR is based on.  However, the custom `ModelLoader.GetModel` method is costly as it performs a full train of the model.  This method gets called when the `PredictionEnginePool` is resolved during constructor injection.

So, I basically have a controller that injects multiple services as a collection:
```
public FooController(IEnumerable<IFooService> fooServices)
```

Each implementation of `IFooService` injects a `PredictionEnginePool`:
```
public BarService(PredictionEnginePool<Foo, Bar> predictionEnginePool)

public BazService(PredictionEnginePool<Foo, Baz> predictionEnginePool)
```

Of course the problem with this is that the models get trained as soon as the controller is created as everything is instantiated at that point.  However, I want to make the `PredictionEnginePool` services lazy so that they are only resolved when I first call a method on them.

I found a solution [here](https://stackoverflow.com/a/45775657/197591) to do that.  This is the code:
```
class LazyService<T> : Lazy<T> where T : class
{
    public LazyService(IServiceProvider serviceProvider) : base(serviceProvider.GetRequiredService<T>) {}
}
```

This means my service DI constructors become this:
```
public BarService(LazyService<PredictionEnginePool<Foo, Bar>> lazyPredictionEnginePool)

public BazService(LazyService<PredictionEnginePool<Foo, Baz>> lazyPredictionEnginePool)
```

And I just add this to my `Startup.ConfigureServices`:
```
services.AddPredictionEnginePool<Foo, Bar>();
services.AddPredictionEnginePool<Foo, Baz>();
services.AddSingleton<LazyService<PredictionEnginePool<Foo, Bar>>>();
services.AddSingleton<LazyService<PredictionEnginePool<Foo, Baz>>>();
```

However, this means I now have unnecessary registrations of `PredictionEnginePool<Foo, Bar>` and `PredictionEnginePool<Foo, Baz>` caused by the convenience method of `AddPredictionEnginePool` which I must call for all the other prerequisites, unless I call them myself, but this would then bloat my code a little.

By making the "prerequisites" method public, I can do the following instead which would not result in any unnecessary additional registrations, or indeed having all those prerequisites being called twice (!):
```
services.AddPredictionEnginePoolServices();
services.AddSingleton<LazyService<PredictionEnginePool<Foo, Bar>>>();
services.AddSingleton<LazyService<PredictionEnginePool<Foo, Baz>>>();
```

Let me know your thoughts on this.  It's a minor change and I think would be very helpful for users wanting a little more control on the registration of the `PredictionEnginePool`, but it would show as an additional exposed method on `services`. Perhaps then it could be in another extension namespace, or perhaps it's not a big issue at all.

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
eerhardt: In the context of `AddSingleton`, saying `the service` makes sense because it can be any kind of service. But here, we are in the context of `AddPredictionEnginePool`, so `the service` doesn't really fit. As a reader, I found myself asking the question "what service?", which is why I think replacing `service` with `PredictionEnginePoolBuilder` makes more sense. We are adding a `PredictionEnginePoolBuilder`, and the `implementationFactory` is the thing that makes it.

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
eerhardt: I think I'd need to see the full code to get my head around this scenario. Is it possible to share the full thing?

NeoXtreem: I've created a temporary repo [here](https://github.com/NeoXtreem/prediction-engine-pool-sample-app) which is a stripped down version of my application to show just the relevant parts. I've left in code where necessary to show intent and to add context such as the custom `ModelLoader`, `IModelService` and `IPredictionService` concrete classes, etc. I've put a `TODO` comment in `Startup` to indicate where the effect of my proposed change would be.

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
NeoXtreem: OK, have updated.

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
eerhardt: Thanks for taking the time to explain it, @NeoXtreem. I now understand why you want this. I think it makes sense to expose it. I'll comment on the new code inline.

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
```diff
@@ -19,27 +20,50 @@ public static class ServiceCollectionExtensions
...
-            services.AddOptions();
-            services.TryAddEnumerable(ServiceDescriptor.Singleton<IPostConfigureOptions<MLOptions>, PostMLContextOptionsConfiguration>());
-            services.AddSingleton<PredictionEnginePool<TData, TPrediction>, PredictionEnginePool<TData, TPrediction>>();
+            services
+                .AddPrerequisiteServices()
+                .AddSingleton<PredictionEnginePool<TData, TPrediction>, PredictionEnginePool<TData, TPrediction>>();
+            return new PredictionEnginePoolBuilder<TData, TPrediction>(services);
+        }
+
+        public static IServiceCollection AddPrerequisiteServices(this IServiceCollection services)
```
eerhardt: This needs XML doc since it is now public.

eerhardt: `AddPrerequisiteServices` is too vague, since this is an extension on `IServiceCollection`, people won't know what kind of pre-reqs are being added.

What do you think of renaming it to `AddPredictionEnginePoolPrerequisites`?

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
```diff
@@ -19,27 +20,50 @@ public static class ServiceCollectionExtensions
...
+        {
+            services
+                .AddLogging()
+                .AddOptions()
+                .TryAddEnumerable(ServiceDescriptor.Singleton<IPostConfigureOptions<MLOptions>, PostMLContextOptionsConfiguration>());
             return services;
```
eerhardt: You could change this to simply:

```C#

return services
                .AddLogging()
                .AddOptions()
                .TryAddEnumerable(ServiceDescriptor.Singleton<IPostConfigureOptions<MLOptions>, PostMLContextOptionsConfiguration>());
```

NeoXtreem: Unfortunately, `TryAddEnumerable` returns void, so this isn't possible in this case.

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
NeoXtreem: I felt a little uncomfortable with that name as it could imply to the user that they must call this method _before_ calling `AddPredictionEnginePool`.  I wanted to think of a name that indicates that it _only_ does something _less_ than `AddPredictionEnginePool`.  So I came up with `AddPredictionEnginePoolConfigOnly` given that the services being added appear to be configuration required for `PredictionEnginePool` (i.e., logging, options, post-configure options).

Grammatically, it makes sense also.  In `AddPredictionEnginePool`, "PredictionEnginePool" is the noun. In `AddPredictionEnginePoolConfigOnly`, "PredictionEnginePool" becomes an adjective while "Config" becomes the noun. "Only" is an adverb in this case.  "Add" in both cases is of course the verb.

Let me know what you think.  I want to be careful with the naming on this as it will be there forever once set. 🙂

NeoXtreem: Have added in my latest commit.  I also changed the summary for the other methods to clarify that the "required config services" will also be added.  This ensures the user knows that `AddPredictionEnginePoolConfigOnly` is a subroutine of `AddPredictionEnginePool` in case it wasn't clear from the method names themselves.

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
NeoXtreem: > I think this change looks good. Thanks for the contribution @NeoXtreem!
> 
> Once you fix up the doc comments, and the CI is green, I can merge.

@eerhardt I've updated the code to be more DRY by putting the common functionality of the two `AddPredictionEnginePool` methods into a private method called by each.  They each pass a callback delegate to do the specific thing which is to call the appropriate overload of `AddSingleton`.

NeoXtreem: @eerhardt I've updated the code to be more DRY by putting the common functionality of the two AddPredictionEnginePool methods into a private method called by each. They each pass a callback delegate to do the specific thing which is to call the appropriate overload of AddSingleton.

NeoXtreem: @eerhardt I've updated the code to be more DRY by putting the common functionality of the two `AddPredictionEnginePool` methods into a private method called by each. They each pass a callback delegate to do the specific thing which is to call the appropriate overload of `AddSingleton`.

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
```diff
@@ -27,19 +28,60 @@ public static class ServiceCollectionExtensions
...
+        /// <summary>
+        /// Adds only the required config services for <see cref="PredictionEnginePool{TData, TPrediction}"/> to the service collection.
+        /// </summary>
+        /// <param name="services">
+        /// The <see cref="IServiceCollection "/> to add services to.
+        /// </param>
+        /// <returns>
+        /// A reference to this instance after the operation has completed.
+        /// </returns>
+        public static IServiceCollection AddPredictionEnginePoolConfigOnly(this IServiceCollection services)
```
eerhardt: I'm not certain `ConfigOnly` is a great name here. How about `AddRequiredPredictionEnginePoolServices`?

eerhardt: or

`AddPredictionEnginePoolRequiredServices`

NeoXtreem: The _only issue_ (so to speak) with those is the same as with `AddPredictionEnginePoolPrerequisites` in that they imply to the user that they must call this method _before_ calling `AddPredictionEnginePool`.  Did you see [my reply](https://github.com/dotnet/machinelearning/pull/4393#discussion_r344935260) to our existing conversation regarding this, particularly the reasoning behind "ConfigOnly"?

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
```diff
@@ -14,7 +15,7 @@ namespace Microsoft.Extensions.ML
     public static class ServiceCollectionExtensions
     {
         /// <summary>
-        /// Adds a <see cref="PredictionEnginePoolBuilder{TData, TPrediction}"/> to the service collection.
+        /// Adds a <see cref="PredictionEnginePool{TData, TPrediction}"/> and required config services to the service collection.
```
eerhardt: This should stay `PredictionEnginePoolBuilder`, that is what is added to the service collection.

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
eerhardt: I wouldn't use the terms "Add...Config" because it sounds like it has to do with [`Microsoft.Extensions.Configuration`](https://www.nuget.org/packages/Microsoft.Extensions.Configuration), but it doesn't.

How about `ConfigurePredictionEnginePoolServices`? To me it doesn't imply that it must be called first. With good XML docs and good samples, it will be fairly obvious to users that they can just call `AddPredictionEnginePool`, which follows the same pattern as other `Microsoft.Extensions` APIs.

NeoXtreem: I think keeping the beginning part `Add` is better as it does indeed add services that are also added with `Add` methods. So, based on your reasoning about good XML and good samples, then I would say your original idea `AddPredictionEnginePoolPrerequisites`, or your second idea `AddRequiredPredictionEnginePoolServices` are the best options. If you're happy with one of those, let me know, and I'll make the change.

eerhardt: I'd be happy with either of those names as well.

If I was forced to pick between them, I would shy away from the term "Prerequisites". It doesn't appear we use that term often in .NET APIs. So I guess I would pick `AddRequiredPredictionEnginePoolServices`

NeoXtreem: OK, I'll go with that one. However, before I do, I need to ask your opinion on something first. I originally came up with this use case because I needed to create multiple `PredictionEnginePool`s, one for a Regression model, and one for a Time Series model. However, I learned from your colleague when he [replied here](https://github.com/dotnet/machinelearning-samples/issues/739#issuecomment-552004422) (also mentioning you) to an issue I raised that `PredictionEnginePool` cannot be used for Time Series model. Therefore, in my case, I only need one `PredictionEnginePool` for now. That's not to say the use case is void as I guess multiple `PredictionEnginePool`s would still be useful for different models that use `PredictionEnginePool`, or even the same model (Regression for instance) where the prediction (output) POCO is different. If you agree, then yes, this change is worthwhile. Let me know please, then I'll make the change.

eerhardt: I _think_ this change is still valid, even if you just have 1 `PredictionEnginePool`. In your case you are using a `LazyService<PredictionEnginePool...` service, right? Even though you only need 1 PredictionEnginePool service, but you aren't calling through our normal `AddPredictionEnginePool` method to create it, you'd still want these "Required Services" added. Having to call through `AddPredictionEnginePool` would just add an unnecessary service that you aren't going to use.

NeoXtreem: Exactly right, although the reason I was using `LazyService<PredictionEnginePool...` was because I was using multiple `PredictionEnginePool`s and only wanted one to be fully instantiated (when required). Now I'm only using one (because Time Series doesn't use `PredictionEnginePool`), it's not as crucial for it to be instantiated lazily because it will always definitely be used at some point after startup.

But the extra control is definitely useful for anyone wanting to instantiate lazily for whatever reason, or if they want additional control on adding the `PredictionEnginePool` for some other reason (maybe `AddScoped` instead of `AddSingleton`).

So I'll go ahead and make the change to the method name as discussed. :)

src/Microsoft.Extensions.ML/ServiceCollectionExtensions.cs
NeoXtreem: Are you sure? The code seems pretty clear in that it adds a `PredictionEnginePool<TData, TPrediction>` object:

```cs
services.AddSingleton<PredictionEnginePool<TData, TPrediction>
```

What is returned is a `PredictionEnginePoolBuilder<TData, TPrediction>` object which simply wraps `IServiceCollection`. Speaking of which, this class doesn't do anything with the generic arguments which could probably even be removed?

eerhardt: > Speaking of which, this class doesn't do anything with the generic arguments which could probably even be removed?

You can't change the generic arguments because this class is public and has shipped as a stable API.

Also, the generic arguments are used by the extension methods.

harishsk:
@NeoXtreem Sorry for the delay in merging your PR. We have had a bit of instability in our CI builds that we are trying to work through. We have checked in a temporary fix yesterday that should improve the situation a bit. Can you please fetch from master, rebase your PR branch and resubmit your PR. We should be able to merge your PR then.

gvashishtha:
@NeoXtreem any success in rebasing?

iluveu28:
Is this available in 1.4 already? Can I get the complete code samples?

gvashishtha:
These changes will not be available until the pull request is formally merged, but there are still some test failures that are preventing that.

iluveu28:
Thanks for your swift response gvashishtha.

This has been sitting there since Feb 19. In the mean time, is there an alternative way to load a newly trained model into the prediction engine pool past the startup ? If so, can you make a sample that checks a folder for a new model and load it maybe from a controller or middleware?

iluveu28:
Guys - can we prioritize this? Correct me if I'm wrong but with this change in place, I can load a newly trained model into an existing prediction engine pool during runtime(post StartUp.cs) e.g. inside a Web API controller....yes?

gvashishtha:
@natke to follow up as I have since moved teams

sanketshah11:
Fellow Microsoft FTE here, It will be very helpful if this change is made available. I will try to send email internally to get some update.

frank-dong-ms:
This is PR has passed CI and merged into master, thanks.

iluveu28:
Thanks a lot for making good progress guys! 

In which version will this be released with? Will there be sample codes on how I can add/update/remove a model from the engine pool inside a web api controller or something?

frank-dong-ms:
@iluveu28  This PR will be released in new 1.5.2 version I think, add @natke and @harishsk  to comment on release date.

I don't find any existing sample code on add/update/remove a model from engine pool but these can be really easy.We have file change monitor that keep monitoring the model file, so if any changes to model file ML.NET will automatically load the new model into memory and the new model will be ready to use. Below is sample usage from one of our customer that shows how to update a model: https://github.com/alexandermujirishvili/DotnetMLWatchForChanges

iluveu28:
Yes I know but when a new model is trained, I need to restart the service in order for it to be loaded into the engine pool. I need to be able to add a new model into the pool outside of the startup.cs. Try to create a web api Controller that accepts a training file to train then immediately predict after that and you’ll see what I mean.

iluveu28:
Note that the new model uploaded by users will have different Model names every time. We don’t overwrite the same file as we have multiple models for different use cases. Thanks again.

frank-dong-ms:
@iluveu28 I see what you mean now. What you ask is more of a ASP.NET question instead of a ML.NET question to me. 
Check out the so thread below on adding service outside startup class:
https://stackoverflow.com/questions/42433321/add-service-outside-startup

iluveu28:
@frank-dong-ms yes that might work but this PR is to make the pool accessible through the middleware which is a better solution correct?

Also on your note about the watchForChanges - https://github.com/alexandermujirishvili/DotnetMLWatchForChanges
Is this issue resolved? I did have the same issue whereby the retrained model is not refreshed until the service restarts although the watchForChanges flag is set to true. The temporary workaround I did is when the model is stale, I create a new predict engine instead of using the pool until the service is restarted then the IsStale flag is set to false in the DB. I think I can solve this by reloading the model into the pool once this PR is released? Or is there a bug in the watchForChanges implementations that you guys will fix soon?

Thanks a lot for the tips!

frank-dong-ms:
@iluveu28 Yes, check out Eric's reply on so for how to define custom model loader:
https://stackoverflow.com/questions/57682393/how-to-add-model-to-predictionenginepool-in-middleware-ml-net/57698799#57698799
https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/end-to-end-apps/DeepLearning_ImageClassification_TensorFlow

For file watch bug, yes we already have fixed the issue and this fix should come out with this PR at next release.

iluveu28:
@frank-dong-ms Sure. This PR is still needed for the custom model loader to work due to below reply from @NeoXtreem in https://stackoverflow.com/questions/57682393/how-to-add-model-to-predictionenginepool-in-middleware-ml-net/57698799#57698799

"I tried using a custom ModelBuilder but found I'm unable to access middleware services during Startup.ConfigureServices, and AddPredictionEnginePool doesn't offer an overload that accepts an implementation factory where I would normally be able to access IServiceProvider as I would with AddAddSingleton, AddScoped and AddTransient. Using the method I mentioned before (ServiceProvider.BuildServiceProvider) results in a warning that an additional copy of singleton services would be created. I've fixed this and raise a PR here. – Neo Oct 26 '19 at 4:01"

Thanks for making ML.NET better and I can't wait to try out the next version 👍

iluveu28:
Hi,

Is this released already? v1.5.2 ?

harishsk:
We are about to release v1.5.2. (The nugets have just been uploaded, so in theory you can make use of them. The rest of the release process should be complete by Monday)

iluveu28:
Thanks a lot guys.

Is there a method that I can call or override to load a newly trained model into the prediction engine pool? 

I have 2 microservices (training and prediction). My users can upload a training data into the training microservice and should be able to predict right after training is completed by calling the prediction micro service without having to restart the service. The model name can be new and different from what's already loaded up into the engine pool during StartUp.cs ConfigureServices method. I hope there is something like AddPredictionEnginePool() accessible outside the StartUp.cs e.g. in the API controller.

A sample code would be appreciated.

PR #4395, opened 2019-10-27
Integrate Image Classification API in AutoML.

codemzs:
- Integrates Image Classification API in AutoML
- Code generation for Image Classification API.
- Unit-tests
- Increases build timeout to 75 minutes.
- Changes Image Classification defaults.

addressed comments from #4384


docs/samples/Microsoft.ML.AutoML.Samples/MulticlassClassificationExperiment.cs
```diff
@@ -46,7 +56,7 @@ public static void Run()
                 mlContext.Model.Save(bestRun.Model, trainDataView.Schema, fs);
 
             // STEP 6: Create prediction engine from the best trained model
-            var predictionEngine = mlContext.Model.CreatePredictionEngine<PixelData, PixelPrediction>(bestRun.Model);
+             var predictionEngine = mlContext.Model.CreatePredictionEngine<PixelData, PixelPrediction>(bestRun.Model);
```
LittleLittleCloud: I'm not sure if we want changes in sample get checked in

codemzs: please look at iteration #4

---
In reply to: [339325057](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339325057) [](ancestors = 339325057)

src/Microsoft.ML.CodeGenerator/CodeGenerator/CSharp/TransformGenerators.cs
```diff
@@ -330,4 +330,26 @@ public override string GenerateTransformer()
...
+        {
+        }
+
+        internal override string MethodName => "LoadRawImageBytes";
+
+        public override string GenerateTransformer()
+        {
+            string inputColumn = InputColumns.Count() == 1 ? InputColumns[0] : throw new Exception($"output columns for the suggested transform: {MethodName} are null");
+            string outputColumn = OutputColumns.Count() == 1 ? OutputColumns[0] : throw new Exception($"output columns for the suggested transform: {MethodName} are null");
+            string inputFolder = Properties["imageFolder"] as string;
```
LittleLittleCloud: why dont we just use `pipelineNode.property`, the refactor in `TransformGeneratorBase` seems to be unnecessary

src/Microsoft.ML.CodeGenerator/CodeGenerator/CSharp/TransformGenerators.cs
```diff
@@ -330,4 +330,26 @@ public override string GenerateTransformer()
...
+            if (inputFolder == null)
+            {
+                throw new Exception($"miss imageFolder property");
```
LittleLittleCloud: And in modelBuilder, we don't require ImageFolder to be set (we use full path), so maybe instead of throwing an error, giving a hint msg might be better choice?

codemzs: good idea I will remove it. I prefer full paths any day

---
In reply to: [339325224](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339325224) [](ancestors = 339325224)

src/Microsoft.ML.AutoML/API/ExperimentBase.cs
```diff
@@ -102,16 +102,16 @@ public abstract class ExperimentBase<TMetrics, TExperimentSettings>
             const int crossValRowCountThreshold = 15000;
 
             var rowCount = DatasetDimensionsUtil.CountRows(trainData, crossValRowCountThreshold);
-
-            if (rowCount < crossValRowCountThreshold)
+            bool isImageClassificationTask = columnInformation.ImagePathColumnNames.Count != 1;
+            if (!isImageClassificationTask && rowCount < crossValRowCountThreshold)
```
justinormont: As a continuation of https://github.com/dotnet/machinelearning/pull/4384#discussion_r338960267

---

Shuffling is not required for CV to work properly. And CV is required when your number of rows is small, to get a useful measure of how good your model performs.

I think you're visualizing the CV folds as created as `folds =  [ rows 0...99, rows 100...199, rows 200..299 ]`. Instead of this method, the rows are split pseudorandomly.

CV splits the dataset with each row going into one of K folds pseudorandomly. Then each of the K folds are chosen one at a time as the test dataset, and the other K-1 folds are used as the training set. 

The cats and dogs in your example will be roughly equally split among the CV folds. That said, you may want a shuffle as part of your trainer, to lessen overfitting.

```suggestion
            if (rowCount < crossValRowCountThreshold)
```

src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
```diff
@@ -47,16 +47,18 @@ internal static class SplitUtil
         /// Split the data into a single train/test split.
         /// </summary>
         public static (IDataView trainData, IDataView validationData) TrainValidateSplit(MLContext context, IDataView trainData,
-            string samplingKeyColumn)
+            string samplingKeyColumn, bool shuffle = false)
```
justinormont: As a continuation of previous comments https://github.com/dotnet/machinelearning/pull/4384#pullrequestreview-307464610

---

Shuffling is helpful as part of the trainer options. No need to have a shuffle here.

```suggestion
            string samplingKeyColumn)
```

src/Microsoft.ML.AutoML/TrainerExtensions/TrainerExtensionCatalog.cs
```diff
@@ -58,13 +59,24 @@ public static ITrainerExtension GetTrainerExtension(TrainerName trainerName)
...
+                columnInfo.CategoricalColumnNames.Count == 0 &&
+                columnInfo.NumericColumnNames.Count == 0 &&
+                columnInfo.TextColumnNames.Count == 0)
+            {
+                // Image Classification case where all you have is a label column and image column.
+                // This trainer takes features column of vector of bytes which will not work with any
+                // other trainer.
+                return new List<ITrainerExtension>() { new ImageClassificationExtension() };
+            }
             else if (task == TaskKind.MulticlassClassification)
```
justinormont: The fall through is normal multi-class. We should add image handling there. 

I'd recommend two options for the fall-thru case: (for when there are images+text/numerics/categoricals)
1. Returning the penultimate layer from the `ImageClassification` trainer (running as a featurizer)
2. Running as the 1st model in a stack ensemble (returning the `score` column as features)

Other tasks, like regression/clustering, etc still need image support. Option 1, above, won't cover it as `ImageClassification` requires a classification label. Perhaps a label-less, no fine-tuning, featurizer mode? Basically using your `ImageClassification` as a replacement for DNNImageFeaturizer.

src/Microsoft.ML.AutoML/EstimatorExtensions/EstimatorExtensions.cs
```diff
@@ -256,17 +257,45 @@ public IEstimator<ITransformer> CreateInstance(MLContext context, PipelineNode p
             return CreateInstance(context, pipelineNode.InColumns[0], pipelineNode.OutColumns[0]);
         }
 
-        public static SuggestedTransform CreateSuggestedTransform(MLContext context, string inColumn, string outColumn)
+        public static SuggestedTransform CreateSuggestedTransform(MLContext context, string inColumn, string outColumn, bool sort = false)
```
justinormont: Continuation of https://github.com/dotnet/machinelearning/pull/4384#pullrequestreview-307469017

---


> From: @daholste, edited by codemzs 
> out of curiosity, why sort for images? #Resolved
>
> ---
>
>    From: @codemzs
> great question, when you split the dataset and then apply ValueToKeyMapping after the split then it could result in different key to value mapping for the same class. I saw some cases where this was being done in the code base so for image classification I just want to be sure the mappings is consistent else it can degrade the accuracy.
>  
> In reply to: 339251823 [](ancestors = 339251823)
>
> ---
>
>    From: @justinormont
> Why does the mapping being inconsistent affect the model's accuracy?
> 
> You'll run into inconsistent label mappings rather often, specifically the mapping can ends up different each time it runs on a different split of the dataset.
> 
> For instance every model you create, the dataset will be split slightly differently in TrainValidate mode, or within CV. One ValueToKeyMapping run will see 100 classes, another will see 99.
>  
> ---
>
>    From: @codemzs
> I’ll explain offline.

Offline discussion sounds great. Thanks.

src/Microsoft.ML.AutoML/TrainerExtensions/MultiTrainerExtensions.cs
```diff
@@ -229,4 +229,22 @@ public PipelineNode CreatePipelineNode(IEnumerable<SweepableParam> sweepParams,
                 columnInfo.LabelColumnName, columnInfo.ExampleWeightColumnName);
         }
     }
+
+    internal class ImageClassificationExtension : ITrainerExtension
```
justinormont: Continuation of https://github.com/dotnet/machinelearning/pull/4384#pullrequestreview-307470047

---

> From: @daholste
> you may want to make sure that the CLI still runs / doesn't break if someone has image paths in their column names?
> 
> ---
>    
> From: @justinormont
> The main issue I expect is needing to add the right package references to the CLI. Otherwise the CLI builds, and runs correctly from VS, but the packaged/shipped CLI fails at runtime.
> 
> Need to add the new packages used to mlnet.nupkgproj.
> 
> Thanks to @eerhardt for his help in debugging the CLI when I was building this.

src/Microsoft.ML.AutoML/TrainerExtensions/MultiTrainerExtensions.cs
```diff
@@ -229,4 +229,22 @@ public PipelineNode CreatePipelineNode(IEnumerable<SweepableParam> sweepParams,
...
+    {
+        public IEnumerable<SweepableParam> GetHyperparamSweepRanges() => null;
```
justinormont: Should at minimum sweep over the model architectures and possibly additional impactful hyperparameters.

Possible useful sweep targets:
* Model architecture (or [domain adapted models](https://en.wikipedia.org/wiki/Domain_adaptation))
* Layer training schedule, like only last layer, last N layers, or 100 epochs of last layer and 10 epochs of all layers
* NumIterations (epochs)
* Learning rate schedule

All of the above should have an Auto option to keep the current heuristics.

src/Microsoft.ML.AutoML/TrainerExtensions/MultiTrainerExtensions.cs
```diff
@@ -229,4 +229,22 @@ public PipelineNode CreatePipelineNode(IEnumerable<SweepableParam> sweepParams,
...
+
+        public ITrainerEstimator CreateInstance(MLContext mlContext, IEnumerable<SweepableParam> sweepParams,
+            ColumnInformation columnInfo)
+        {
+            var options = TrainerExtensionUtil.CreateOptions<Dnn.ImageClassificationTrainer.Options>(null, columnInfo.LabelColumnName);
```
justinormont: Would it be possible for the ImageClassifier to have a GPU setting? 

`UseGPU = { Auto, On, Off }`

Options.
* `Auto` will try the GPU and fail over to the CPU (default)
* `On` will use the GPU the throw if not possible. 
* `Off` will use only the CPU.

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -82,5 +84,124 @@ private static string DownloadIfNotExists(string baseGitPath, string dataFile)
...
+        }
+
+        public static IEnumerable<ImageData> LoadImagesFromDirectory(string folder,
+            bool useFolderNameAsLabel)
+        {
+            var files = Directory.GetFiles(folder, "*",
+                searchOption: SearchOption.AllDirectories);
+            foreach (var file in files)
+            {
+                if (Path.GetExtension(file) != ".jpg")
```
justinormont: Incase someone sees and replicates this code...

Recommend: .jpg, .jpeg, .png, .gif. 
And case insensitive. 
```suggestion
                var extension = Path.GetExtension(file).ToLower();
                if (extension != ".jpg" &&
                    extension != ".jpeg" &&
                    extension != ".png" &&
                    extension != ".gif"
                )
```

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -82,5 +84,124 @@ private static string DownloadIfNotExists(string baseGitPath, string dataFile)
...
+                else
+                {
+                    for (int index = 0; index < label.Length; index++)
+                    {
+                        if (!char.IsLetter(label[index]))
+                        {
+                            label = label.Substring(0, index);
+                            break;
+                        }
+                    }
```
justinormont: nit: (as it's only part of the test setup code)

I wouldn't recommend supporting auto-labels from the filenames. If auto-inferring the labels, I'd stick with only the folder names.

I think this is trying to catch a dataset named as `cat_image101.jpg`, `dog_image341.jpg` into:
```
Label,ImagePath
cat,c:\path\cat_image101.jpg
dog,c:\path\dog_image341.jpg
```

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -82,5 +84,124 @@ private static string DownloadIfNotExists(string baseGitPath, string dataFile)
...
+            }
+        }
+
+        public static string DownloadImageSet(string imagesDownloadFolder)
+        {
+            string fileName = "flower_photos_tiny_set_for_unit_tests.zip";
+            string url = $"https://mlnetfilestorage.file.core.windows.net/imagesets" +
+                $"/flower_images/flower_photos_tiny_set_for_unit_tests.zip?st=2019" +
+                $"-08-29T00%3A07%3A21Z&se=2030-08-30T00%3A07%3A00Z&sp=rl&sv=2018" +
+                $"-03-28&sr=f&sig=N8HbLziTcT61kstprNLmn%2BDC0JoMrNwo6yRWb3hLLag%3D";
```
justinormont: Recommend linking to aka.ms links, which point to the CDN, which point to public folders in the blobstore.

Linking to aka.ms allows us to redirect the links in the future incase we decide to move them (without breaking old versions. Linking to the CDN gives us speed, and hides the actual blobstore backing the CDN.

As above, Yael's downloader class will help with the downloading process.

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -82,5 +84,124 @@ private static string DownloadIfNotExists(string baseGitPath, string dataFile)
...
+            Download(url, imagesDownloadFolder, fileName);
+            UnZip(Path.Combine(imagesDownloadFolder, fileName), imagesDownloadFolder);
+
+            return Path.GetFileNameWithoutExtension(fileName);
+        }
+
+        private static void Download(string url, string destDir, string destFileName)
+        {
+            if (destFileName == null)
+                destFileName = url.Split(Path.DirectorySeparatorChar).Last();
```
justinormont: This will fail on Windows (and succeed on Linux/macOS) as the Separator char is '\\' instead of '/'. Also should handle parsing the '?' and '#' of URLs, along with unescaping chars.

This should work:
```suggestion
                destFileName = Path.GetFileName(new Uri(url).AbsolutePath);
```

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -82,5 +84,124 @@ private static string DownloadIfNotExists(string baseGitPath, string dataFile)
...
+
+            Directory.CreateDirectory(destDir);
+
+            string relativeFilePath = Path.Combine(destDir, destFileName);
+
+            if (File.Exists(relativeFilePath))
+                return;
+
+            new WebClient().DownloadFile(url, relativeFilePath);
```
justinormont: When writing a downloader for large files, it's helps to download to a temp filename and do an atomic move once finished. Otherwise partially downloaded files will never be retried.

We worked thru similar issues in @yaeldekel's downloader class. I'd recommend using her class.

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -82,5 +84,124 @@ private static string DownloadIfNotExists(string baseGitPath, string dataFile)
...
+            return;
+        }
+
+        private static void UnZip(String gzArchiveName, String destFolder)
+        {
+            var flag = gzArchiveName.Split(Path.DirectorySeparatorChar)
+                .Last()
+                .Split('.')
+                .First() + ".bin";
```
justinormont: `Path.GetFileNameWithoutExtension()` is made for this. 

Instead of the four lines of `gzArchiveName.Split(Path.DirectorySeparatorChar).Last().Split('.').First() + ".bin";`

```suggestion
            var flag = Path.GetFileNameWithoutExtension(gzArchiveName) + ".completed"
```


Also, `flag` is a bit ambiguous of a name. Perhaps `completionIndicatorFileName`.

src/Microsoft.ML.AutoML/EstimatorExtensions/EstimatorExtensions.cs
```diff
@@ -256,17 +257,45 @@ public IEstimator<ITransformer> CreateInstance(MLContext context, PipelineNode p
...
+            if (sort)
+                return context.Transforms.Conversion.MapValueToKey(outColumn, inColumn, keyOrdinality: ValueToKeyMappingEstimator.KeyOrdinality.ByValue);
+
+            return context.Transforms.Conversion.MapValueToKey(outColumn, inColumn);
+        }
+    }
+
+    internal class ImageLoadingExtension : IEstimatorExtension
+    {
+        public static string ImageFolder { private get; set; }
```
justinormont: Continuation of https://github.com/dotnet/machinelearning/pull/4384#pullrequestreview-307343740

---

> From: @LittleLittleCloud
> Where do we Initialize ImageFolder
> 
> Or the image dataset has to be full_path?

justinormont: The most common dataset type has a relative path to the image: 
```csv
Label,SplitColumn,Weight,ImagePath
Pneumonia,Patient001,10,images/file97234.png
Pneumonia,Patient001,10,images/file09712.png
NoPneumonia,Patient002,1,images/file82193.png
NoPneumonia,Patient002,1,images/file34387.png
```

The images are then relative to the CSV/TSV dataset provided. Can you ensure this still works? The code is within the other image PR. The base path is assumed to be the CSV/TSV dataset's folder and can be overridden.

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
justinormont: The new FlowerPhotosTiny dataset likely should have an entry added to the [test data README.md](https://github.com/dotnet/machinelearning/blob/056c60479304a3b5dbdf129c9bc6e853322bb090/test/data/README.md). 

While we are not redistributing the dataset from the repo, we are still redistributing the dataset from our stores (as opposed to downloading from a 3rd party). We don't otherwise have a place to mention the dataset source and the modifications we made to the dataset.

test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
```diff
@@ -45,6 +48,36 @@ public void AutoFitMultiTest()
...
+
+            //Known issue, where on Ubuntu there is degradation in accuracy.
+            if (!(RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ||
+                (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))))
+            {
+                Assert.Equal(0.889, result.BestRun.ValidationMetrics.MicroAccuracy, 3);
+            }
+            else
+            {
+                Assert.Equal(0.444, result.BestRun.ValidationMetrics.MicroAccuracy, 3);
```
justinormont: What's causing the `ImageClassificationTrainer` (or another component) to fail on Linux? What's the timeline for a fix?

@rustd, @CESARDELATORRE -- If we have to, is shipping local image classification without Linux support ok?

src/Microsoft.ML.AutoML/API/ExperimentBase.cs
codemzs: Hi Justin, 

I don't want our ImageClassificationTrainer to be used on CV. For the initial release it will be disabled until we figured out there is value in doing CV. Contrary to what you had suggested, using train-test on smaller datasets is actually give great accuracy, please see my tests, it works on trainset of 50 images. 

---
In reply to: [339329519](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339329519) [](ancestors = 339329519)

src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
codemzs: I disagree here, I have seen accuracy go down with shuffling turned off, this has been my experience after benchmarking several image datasets on ImageClassificationTrainer. 

---
In reply to: [339329734](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339329734) [](ancestors = 339329734)

src/Microsoft.ML.AutoML/TrainerExtensions/TrainerExtensionCatalog.cs
codemzs: We are not handling Image handling here because ImageClassificationTrainer does not support just returning featurized outputs and your approach actually causes a crash in the code, as there is a bug that fails to load the saved model @bigmiao.zhang@gmail.com can comment more. It is not within the scope of this PR to add your approach or handle cases other than where you just have images and labels, the scope is to integrate ImageClassificationTrainer API into AutoML.

---
In reply to: [339329937](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339329937) [](ancestors = 339329937)

src/Microsoft.ML.AutoML/EstimatorExtensions/EstimatorExtensions.cs
codemzs: @bigmiao.zhang@gmail.com says the dataset passed will contain full paths, we are good here.

---
In reply to: [339334860](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339334860) [](ancestors = 339334860)

src/Microsoft.ML.AutoML/TrainerExtensions/MultiTrainerExtensions.cs
codemzs: Not in the current release. You will need to add GPU nuget at compile time to use it. There are ways around to auto detection if the user has GPU, CUDA and CUDANN installed and then copying binaries around but it will be after November release. Please also note tensorflow also works the same way as us currently.

---
In reply to: [339330427](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339330427) [](ancestors = 339330427)

test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
codemzs: First, its not failing on Linux :) we see not so good accuracy on Linux for *some* datasets (mostly smaller ones .. our samples seem to work great) but on Windows and MacOS accuracy is consistent. We haven't investigated why for some smaller test datasets Linux gives not so good accuracy but may be after release we can do that.

---
In reply to: [339356839](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339356839) [](ancestors = 339356839)

src/Microsoft.ML.AutoML/API/ExperimentBase.cs
justinormont: The current transition point for TrainTest vs. CV was chosen based on benchmarking on over 100 datasets in the AutoML team. We are very confident on its importance. 

Disabling CV gives you a 10x speed up, but makes the accuracy metric too noisy on a dataset with only 50 images for a unit test. At a seeded-random 80/20 split only 9 images are the validation split. Hence the accuracy metric will jump by 11% for each image being right or wrong. 

This gives a random value for the output accuracy metric. It's not  a _great_ accuracy, instead it's roughly _unknown_.

We're trying to measure the value of a weighted coin (percent of correct predictions). The actual model accuracy is perhaps 0.7890, but given only 9 datapoint in the unit test's measurement set we will know the accuracy only with enormous error bars.

The point of using TrainTest mode or CV mode is to estimate how well the user's model will do in production. Currently your model's accuracy is somewhere between 55% to 91% (0.95 confidence interval for a Bernoulli distribution w/ 8 of 9 correct predictions)<sup>[[1]](http://epitools.ausvet.com.au/content.php?page=CIProportion&SampleSize=9&Positive=8&Conf=0.95&method=6&Digits=4)</sup>. Bringing in CV mode makes your error bars significantly smaller, which is its utility.

CV is how you get a useful metric on a small dataset. Currently the code returns a non-useful metric for small datasets.

codemzs: Thanks, @justinormont. Another topic for offline discussion :)

src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
justinormont: Similar to https://github.com/dotnet/machinelearning/pull/4395#discussion_r339375063, you may be seeing noisy numbers, which makes it difficult to tell if the shuffling is impactful or not. 

For a NN, I would expect if your learning rate is high, or if you have a very low number of epochs, then shuffling can be helpful. Due to the block shuffle style in ML.NET, you won't see gains when you have above ~1k images per class.

If the new trainer prefers shuffled data, we should align with the rest of trainers which offer a hyperparameter for shuffling. 

As [noted](https://github.com/dotnet/machinelearning/pull/4384#discussion_r339248471) by @daholste, the shuffling is currently out of place:
> Perhaps conceptually 'shuffle' belongs outside of the train/valid split API ?
> If desired (and only desired for images right now), you could shuffle the results of the this API
> For instance, I don't think it's part of the ML.NET train/valid split API

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
codemzs: @justinormont You are correct. Downloading to temp file names makes sense otherwise you can end up with partially downloaded files that will be treated as fully downloaded files and can cause unknown exceptions.

src/Microsoft.ML.AutoML/EstimatorExtensions/EstimatorExtensions.cs
justinormont: Why mark as resolved without the discussion?

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
codemzs: It is just an option, not on by default. We use it in some cases but you are right in  most cases labels are determined from the folder name.

src/Microsoft.ML.AutoML/TrainerExtensions/MultiTrainerExtensions.cs
justinormont: Ok. Thanks. 

Being AutoML, we should handle this for the user. We can handle either at the `ImageClassificationTrainer` level, or at the AutoML level. Either will work, though the former has benefits for users directly using the `ImageClassificationTrainer`.

src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
codemzs: thanks, we can discuss this offline. To begin with, we should be asking the user for a trainset and a validation set, this how other APIs such as TensorFlow SLIM's image classification API is designed. Our ImageClassification API is also designed the same way but unfortunately the Model Builder UI does not accept a validation set so we are left with creating a validation set from the train set, this is troublesome in scenarios like image classification where users give a folder to all their images and within this folder you have sub folders that contain images for various classes, now if we don't shuffle we could end up training on some classes and then validating on classes that we didn;'t even train on! hence we shuffle before we create train and validation set, and FYI, Image classification API does also shuffle the train set it receives to keep data points independent. 

I can chat offline and close on this. thanks all.

src/Microsoft.ML.AutoML/TrainerExtensions/MultiTrainerExtensions.cs
codemzs: thanks, as discussed for the first release we DO NOT want ANY sort of sweeping. Image Classification changes the learning rate after every epoch to reduce loss and it auto stops when it feels model has reached optimal accuracy. We plan to use RESNET 50 as the default architecture as it works well in most scenario and is also very popular.

src/Microsoft.ML.AutoML/Microsoft.ML.AutoML.csproj
```diff
@@ -8,6 +8,8 @@
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.ML.Core\Microsoft.ML.Core.csproj" />
     <ProjectReference Include="..\Microsoft.ML.CpuMath\Microsoft.ML.CpuMath.csproj" />
+    <ProjectReference Include="..\Microsoft.ML.Dnn\Microsoft.ML.Dnn.csproj" />
```
eerhardt: Need to update the `.nupkgproj` as well.

src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs
```diff
@@ -363,7 +363,7 @@ public sealed class Options : TrainerInputBaseWithLabel
             /// Specifies the model architecture to be used in the case of image classification training using transfer learning.
             /// </summary>
             [Argument(ArgumentType.AtMostOnce, HelpText = "Model architecture to be used in transfer learning for image classification.", SortOrder = 15)]
-            public Architecture Arch = Architecture.InceptionV3;
+            public Architecture Arch = Architecture.ResnetV250;
```
eerhardt: Why this change? This PR seems like an odd place to make this change.

test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
```diff
@@ -45,6 +48,57 @@ public void AutoFitMultiTest()
...
+            var trainData = context.Data.ShuffleRows(textLoader.Load(datasetPath), seed: 1);
+            var originalColumnNames = trainData.Schema.Select(c => c.Name);
+            TrainTestData trainTestData = context.Data.TrainTestSplit(trainData, testFraction: 0.2, seed: 1);
+            IDataView trainDataset = SplitUtil.DropAllColumnsExcept(context, trainTestData.TrainSet, originalColumnNames);
+            IDataView testDataset = SplitUtil.DropAllColumnsExcept(context, trainTestData.TestSet, originalColumnNames);
+            var result = context.Auto()
+                            .CreateMulticlassClassificationExperiment(0)
+                            .Execute(trainDataset, testDataset, columnInference.ColumnInformation);
+
+            //Known issue, where on Ubuntu there is degradation in accuracy.
```
eerhardt: Can you link to the issue here? This seems like something that needs to be addressed.

test/Microsoft.ML.CodeGenerator.Tests/ApprovalTests/ConsoleCodeGeneratorTests.cs
```diff
@@ -41,7 +41,8 @@ public void ConsoleAppModelBuilderCSFileContentOvaTest()
                 LabelName = "Label",
                 ModelPath = "x:\\models\\model.zip"
             });
-            var result = consoleCodeGen.GenerateConsoleAppProjectContents(namespaceValue, typeof(float), true, true, false);
+            var result = consoleCodeGen.GenerateConsoleAppProjectContents(namespaceValue, typeof(float), true, true,
```
eerhardt: All these booleans make it super hard to read what is happening here. I suggest to use named parameters to make it easy to read:

` consoleCodeGen.GenerateConsoleAppProjectContents(namespaceValue, typeof(float), includeLightGbmPackage: true, includeMklComponentsPackage: true, ...`

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -82,5 +84,124 @@ private static string DownloadIfNotExists(string baseGitPath, string dataFile)
...
+            {
+                file.WriteLine("Label,ImagePath");
+                foreach (var image in images)
+                    file.WriteLine(image.Label + "," + image.ImagePath);
+            }
+
+            return datasetName;
+        }
+
+        public static IEnumerable<ImageData> LoadImagesFromDirectory(string folder,
```
eerhardt: This method is already duplicated twice in our tests. It is time to extract it to a common helper method.

src/Microsoft.ML.CodeGenerator/Templates/Console/PredictProject.tt
```diff
@@ -20,6 +20,12 @@
 <#}#>
 <# if (IncludeFastTreePackage){ #>
     <PackageReference Include="Microsoft.ML.FastTree" Version="1.3.1" />
+<#}#>
+<# if (IncludeImageTransformerPackage){ #>
+    <PackageReference Include="Microsoft.ML.ImageAnalytics" Version="1.4.0" />
```
eerhardt: This code isn't going to work until we ship `1.4.0`.

We may need a way to update these numbers during the build.

src/Microsoft.ML.CodeGenerator/Templates/Console/ConsumeModel.cs
```diff
@@ -18,7 +18,7 @@ namespace Microsoft.ML.CodeGenerator.Templates.Console
     /// Class to produce the template output
     /// </summary>
     
-    #line 1 "C:\CodeHub\machinelearning\src\Microsoft.ML.CodeGenerator\Templates\Console\ConsumeModel.tt"
```
eerhardt: This file should be reverted.

src/Microsoft.ML.CodeGenerator/CodeGenerator/CSharp/TransformGenerators.cs
```diff
@@ -330,4 +330,21 @@ public override string GenerateTransformer()
...
+    {
+        public ImageLoading(PipelineNode node) : base(node)
+        {
+        }
+
+        internal override string MethodName => "LoadRawImageBytes";
+
+        public override string GenerateTransformer()
+        {
+            string inputColumn = InputColumns.Count() == 1 ? InputColumns[0] : throw new Exception($"output columns for the suggested transform: {MethodName} are null");
```
eerhardt: This exception message doesn't make sense for the check it is performing. If the `InputColumns.Count == 2` the message will say $"output columns for the suggested transform: {MethodName} are null".

1. "Input" instead of "output".
2. "are null" doesn't make sense when there are 2 of them.

src/Microsoft.ML.AutoML/EstimatorExtensions/EstimatorExtensions.cs
justinormont: Again, why mark as resolved without the discussion?

src/Microsoft.ML.CodeGenerator/Templates/Console/PredictProject.tt
justinormont: Is there a cleaner way to include the build numbers as variables as we do elsewhere?

test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
justinormont: Related comment -- https://github.com/dotnet/machinelearning/pull/4395#discussion_r339356839

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
justinormont: Thanks. I see there's another PR out which starts using Yeal's downloader class for the `ImageClassifierTrainer` models. Very nice.

codemzs: @justinormont yea we were basically trying to solve the same problem there but I pushed the commit to use resource manager and merged it.

src/Microsoft.ML.AutoML/TrainerExtensions/MultiTrainerExtensions.cs
justinormont: AutoML is a trade-off where compute time is used as a replacement for human time. The goal is to go after a high accuracy using the time the user hands us. If you train only one model we are not doing what the user asked of us. This misses the point of AutoML.

The current goal is to integrate support for images, while fitting in with the current AutoML platform.

src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
justinormont: > ... now if we don't shuffle we could end up training on some classes and then validating on classes that we didn't even train on! hence we shuffle before we create train and validation set

This is an incorrect understanding of how our TrainTest and CV splitting works. More detail here: https://github.com/dotnet/machinelearning/pull/4395#discussion_r339329519

src/Microsoft.ML.AutoML/TrainerExtensions/MultiTrainerExtensions.cs
codemzs: Hi @justinormont I'm not adding any sweeping because this was the initial plan and I will stick to the plan. I'm not saying we should not add sweeping, just that it will not be added in the first release.

test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
codemzs: @eerhardt There is no issue for this because we haven;t fully investigated the cause or even understand the problem, right now my suspicion is bad images format for Linux since it only seems to happen on the smaller version of flowers dataset that has 50 images. It seems there are two images that cause this kind of drop. I will have someone investigate this later.

src/Microsoft.ML.CodeGenerator/Templates/Console/PredictProject.tt
codemzs: @eerhardt this is how code generator is structured, they hardcode version numbers. May be they can make this better as a separate change.

src/Microsoft.ML.CodeGenerator/Templates/Console/ConsumeModel.cs
codemzs: why?

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
codemzs: I would ideally but because I'm short on time I can open an issue and extract later.

---
In reply to: [339654988](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339654988) [](ancestors = 339654988)

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
codemzs: thanks and I agree we should make short links but because I'm short on time and this is not a pressing change in a test code so I can open an issue and resolve this later.

---
In reply to: [339330533](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339330533) [](ancestors = 339330533)

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
codemzs: agreed, just short on time and this is test code so will make this in a separate change,

---
In reply to: [339330556](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339330556) [](ancestors = 339330556)

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
codemzs: thanks but what I have works perfectly well,.

---
In reply to: [339330693](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339330693) [](ancestors = 339330693)

src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs
codemzs: yep, I will describe in the PR that in addition it also does this this and this.

---
In reply to: [339649528](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339649528) [](ancestors = 339649528)

test/Microsoft.ML.CodeGenerator.Tests/ApprovalTests/ConsoleCodeGeneratorTests.cs
codemzs: this is a test file and this is how it was before as well.

---
In reply to: [339651958](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339651958) [](ancestors = 339651958)

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
justinormont: Ok. Great. Can you make a tracking issue for these?

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
justinormont: Nit: Very minor corner case if the file to unzip has two periods: My.File.zip will be "My.bin" instead of "My.File.bin".

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
codemzs: yep

---
In reply to: [339927017](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339927017) [](ancestors = 339927017)

src/Microsoft.ML.AutoML/Experiment/SuggestedPipeline.cs
```diff
@@ -118,8 +119,13 @@ public IEstimator<ITransformer> ToEstimator()
                 }
             }
 
+            if(validationSet != null)
```
justinormont: Nit: space

src/Microsoft.ML.AutoML/Experiment/SuggestedPipeline.cs
```diff
@@ -118,8 +119,13 @@ public IEstimator<ITransformer> ToEstimator()
...
+            {
+                validationSet = pipeline.Fit(trainset).Transform(trainset);
```
justinormont: I was tracking down how you were creating a validation set, to ensure we're not overfitting the validation dataset. We'd have to work out the impact of re-using the existing validation dataset (as it serves the purpose of a test set in this code). The actual test set is kept separately besides for codegen.

I read the current code as: If there is a validation dataset, replace it with a featurized version of the training set. If there is no validation dataset, keep it empty. (bug?)

Creating another validation set by featurizing the trainset is not bad from a re-use of splits reason, but you may not get what you're looking for. I think you're using it within your trainer for early stopping. 

If you used the actual validation dataset for early stopping, this would skew the user's metrics by over-representing the quality of the model.

I'd recommend creating a new TrainTest split to create an independent dataset split for the early stopping. You may want this as a feature of your trainer itself.

codemzs: Nope, the validation set in AutoML code is unchanged, this is just a validation set that is passed to the image classification trainer for its own internal calculations. 

---
In reply to: [339943032](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339943032) [](ancestors = 339943032)

src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs
```diff
@@ -705,18 +705,20 @@ public ImageProcessor(Session session, string jpegDataTensorName, string resizeI
...
-                        throw;
+                    try
+                    {
+                        return _imagePreprocessingRunner.AddInput(imageTensor, 0).Run()[0];
+                    }
+                    catch (TensorflowException e)
+                    {
+                        //catch the exception for images of unknown format
+                        if (e.HResult == -2146233088)
+                            return null;
```
justinormont: Is there a way to notify users which image fails? What occurs if all images don't load? 

Is there an enum that we can pull the "-2146233088" magic number from? Seems fragile. Something like `TensorflowExceptions.UnknownImageFormat`.

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
justinormont: Thanks for fixing.

src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs
codemzs: Yep, TF prints to the console.

---
In reply to: [339946888](https://github.com/dotnet/machinelearning/pull/4395#discussion_r339946888) [](ancestors = 339946888)

test/data/README.md
```diff
@@ -62,6 +62,14 @@ Redistributing the dataset "taxi-fare-test.csv", "taxi-fare-train.csv" with attr
 > 
 > The dataset is provided under terms provided by City of New York: https://opendata.cityofnewyork.us/overview/#termsofuse.
 
+### Flowers dataset.
+
+Redistributing the dataset "tflower_photos_tiny_set_for_unit_tests.zip" with attribution:
+
+> Original source: http://download.tensorflow.org/example_images/flower_photos.tgz
+>
```
justinormont: Should also list modifications made:
```suggestion
> 
> Dataset was modified to reduce the number of images per class to ten.
>
```

src/Microsoft.ML.AutoML/Experiment/SuggestedPipeline.cs
codemzs: P.S: early stopping is disabled by default.

src/Microsoft.ML.CodeGenerator/Templates/Console/PredictProject.tt
justinormont: Would it be possible to change the other package version numbers to next releases' numbers than? Otherwise the codegen (if released like current) will pull in old packages. This is step needed before release anyway.

test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
eerhardt: Can you log an issue for this? That is how work gets tracked. Just because it hasn't been fully investigated doesn't mean there shouldn't be an issue logged.

src/Microsoft.ML.CodeGenerator/Templates/Console/PredictProject.tt
eerhardt: `Version="1.4.0"` doesn't exist yet. So the users project isn't going to build until we publish `1.4.0`.

test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
eerhardt: Can you add a link here to the issue?

build/ci/job-template.yml
```diff
@@ -10,7 +10,7 @@ parameters:
 jobs:
   - job: ${{ parameters.name }}
     ${{ if eq(parameters.codeCoverage, 'false') }}:
-      timeoutInMinutes: 60
+      timeoutInMinutes: 75
```
eerhardt: We really need to get a handle on this situation. Our build times have doubled in the past week.

src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs
eerhardt: I didn't see a description of why this was changed.

test/Microsoft.ML.CodeGenerator.Tests/ApprovalTests/ConsoleCodeGeneratorTests.cs
eerhardt: You are adding to the problem here. When making an existing problem even worse, it makes a lot of sense to address the problem at that time.

src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs
eerhardt: > TF prints to the console.

That's not going to help users who are in a GUI application - like Visual Studio Model Builder.

src/Microsoft.ML.AutoML/Utils/SplitUtil.cs
daholste: also, can we please move shuffle outside of the train/test split API? Isn't it semantically separate from a train/test split? At the very least, since shuffle is only used in the images path, why not have a single line of code shuffling the training data before passing it in to train/test split?

src/Microsoft.ML.AutoML/Experiment/SuggestedPipeline.cs
justinormont: Since I saw this has moved from `validationSet = pipeline.Fit(trainset).Transform(trainset)` to `validationSet = pipeline.Fit(validationSet).Transform(validationSet)`. 

This now uses what the code is calling a validationSet, though it is currently the test dataset as used here. We shouldn't be reusing the dataset for this purpose. This hits on my earlier comment of:
> If you used the actual validation dataset for early stopping, this would skew the user's metrics by over-representing the quality of the model.

Even if not used currently for early stopping, other uses of this dataset will have similar effects.

As mentioned before, we should split off the training set:
> I'd recommend creating a new TrainTest split to create an independent dataset split for the early stopping. You may want this as a feature of your trainer itself.

src/Microsoft.ML.Dnn/ImageClassificationTrainer.cs
codemzs: @eerhardt model builder receives logging from image classification api through callback that is attached to mlcontext.Log.

eerhardt: And how does the "which image fails" information get through the callback if TF prints it to the console?

Where is the `TF prints to the console` code? Will that information flow the callback?

src/Microsoft.ML.AutoML/Experiment/SuggestedPipeline.cs
codemzs: Define reuse? Can you please walk me through how this is being used anywhere in the trainer, I have put it for a reason that I can explain.

PR #4403, opened 2019-10-28
Updating the DatabaseLoader to not force vector for single element columns

tannergooding:
This resolves https://github.com/dotnet/machinelearning-samples/issues/722

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseLoader.cs
@@ -333,7 +333,14 @@ namespace Microsoft.ML.Data
             internal static Range FromTextLoaderRange(TextLoader.Range range)
             {
                 Contracts.Assert(range.Max.HasValue);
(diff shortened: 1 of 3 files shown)


tannergooding:
CC. @eerhardt, @CESARDELATORRE

src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseLoader.cs
```diff
@@ -333,7 +333,14 @@ public Range(int min, int max)
             internal static Range FromTextLoaderRange(TextLoader.Range range)
             {
                 Contracts.Assert(range.Max.HasValue);
-                return new Range(range.Min, range.Max.Value);
+                if ((range.Max.Value == range.Min) && !range.ForceVector)
```
eerhardt: Another option might be:

```C#
Range result = new Range();
result.Min = range.Min;
result.Max = range.Max.Value;
result.ForceVector = range.ForceVector;
return result;
```

It feels a little more straight forward than having to parse this `if` condition, and now which constructors will set ForceVector and which won't.

PR #4406, opened 2019-10-29
Fixes #4405 About publishing nugets to public feed

### LINKED ISSUE #4405: NuGet packages not being published to public feed
It's desirable to have NuGets being published to a public feed as soon as the master branch of this repo gets updated.

In the past, this was tried to be done by publishing them to MyGet, but currently the lines of code that do this are commented out. https://github.com/dotnet/machinelearning/blob/f0fb7203d7597d2779a298662970b1f74b47f4b8/build/vsts-ci.yml#L237-L242

As discussed offline with @codemzs it would be better to instead publish them into this other public feed in Azure DevOps:
https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=MachineLearning

antoniovs1029:
Fixes #4405 about publishing nugets to this public feed:
https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=MachineLearning

This will be executed by the AzureDevOps build pipeline whenever a new commit is added to the master branch of this repo. Notice that sometimes there are some problems on the side of Azure DevOps, and it might fail when executing the programs in the pipeline that should lead to publishing the nugets; the solution is to rerun the build manually.

Worked this out following @safern instructions.

### DIFF
--- a/build/vsts-ci.yml
+++ b/build/vsts-ci.yml
@@ -174,13 +174,13 @@ phases:
   variables:
     BuildConfig: Release
     OfficialBuildId: $(BUILD.BUILDNUMBER)
-    DOTNET_CLI_TELEMETRY_OPTOUT: 1
+    DOTNET_CLI_TELEMETRY_OUTPUT: 1
     DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
     DOTNET_MULTILEVEL_LOOKUP: 0
     _SignType: real
     _UseEsrpSigning: true
     _TeamName: DotNetCore
-    _NuGetFeedUrl: https://dotnet.myget.org/F/dotnet-core/api/v2/package
+    _AzureDevopsFeedUrl: https://pkgs.dev.azure.com/dnceng/public/_packaging/MachineLearning/nuget/v3/index.json
     _SymwebSymbolServerPath: https://microsoft.artifacts.visualstudio.com/DefaultCollection
     _MsdlSymbolServerPath: https://microsoftpublicsymbols.artifacts.visualstudio.com/DefaultCollection
   queue:
@@ -226,20 +226,19 @@ phases:
       msbuildVersion: 15.0
     continueOnError: false
 
-  - task: NuGetCommand@2 
-    displayName: Publish Packages to VSTS Feed 
-    inputs:
-      command: push 
-      packagesToPush: $(Build.SourcesDirectory)/bin/packages/**/*.nupkg;!$(Build.SourcesDirectory)/bin/packages/**/*.symbols.nupkg 
-      nuGetFeedType: internal 
-      feedPublish: MachineLearning
-      
   # - task: MSBuild@1
   #   displayName: Publish Packages to MyGet Feed
   #   inputs:
   #     solution: build/publish.proj
   #     msbuildArguments: /t:PublishPackages /p:NuGetFeedUrl=$(_NuGetFeedUrl) /p:NuGetApiKey=$(dotnet-myget-org-api-key)
   #     msbuildVersion: 15.0
+
+  - task: NuGetAuthenticate@0
+    inputs:
+      nuGetServiceConnections: machinelearning-dnceng-public-feed # To allow publishing to a feed of another organization
+
+  - script: Tools\dotnetcli\dotnet msbuild build\publish.proj /t:PublishPackages /p:NuGetFeedUrl=$(_AzureDevopsFeedUrl) /p:NuGetApiKey=AzureArtifacts
+    displayName: Publish Packages to AzureDevOps Feed
 
   - task: MSBuild@1
     displayName: Publish Symbols to SymWeb Symbol Server
--- a/build/vsts-ci.yml
+++ b/build/vsts-ci.yml
@@ -256,6 +256,14 @@ phases:
       msbuildVersion: 15.0
     continueOnError: true
 
+  - task: NuGetCommand@2 
+    displayName: Publish Packages to private VSTS Feed 
+    inputs:
+      command: push 
+      packagesToPush: $(Build.SourcesDirectory)/bin/packages/**/*.nupkg;!$(Build.SourcesDirectory)/bin/packages/**/*.symbols.nupkg 
+      nuGetFeedType: internal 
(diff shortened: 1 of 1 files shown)


build/vsts-ci.yml
```diff
@@ -174,13 +174,13 @@ phases:
   variables:
     BuildConfig: Release
     OfficialBuildId: $(BUILD.BUILDNUMBER)
-    DOTNET_CLI_TELEMETRY_OPTOUT: 1
+    DOTNET_CLI_TELEMETRY_OUTPUT: 1
```
antoniovs1029: @safern told me this was a typo that needed to be fixed. I believe this change doesn't affects directly the issue that is being fixed in this PR.

build/vsts-ci.yml
```diff
@@ -226,21 +226,20 @@ phases:
...
 
-  - task: NuGetCommand@2 
-    displayName: Publish Packages to VSTS Feed 
-    inputs:
-      command: push 
-      packagesToPush: $(Build.SourcesDirectory)/bin/packages/**/*.nupkg;!$(Build.SourcesDirectory)/bin/packages/**/*.symbols.nupkg 
-      nuGetFeedType: internal 
-      feedPublish: MachineLearning
-      
   # - task: MSBuild@1
```
antoniovs1029: I wonder if I should delete this, since I believe this won't be used in the future?

build/vsts-ci.yml
```diff
@@ -226,21 +226,20 @@ phases:
       msbuildVersion: 15.0
     continueOnError: false
 
-  - task: NuGetCommand@2
```
antoniovs1029: I asked @codemzs and he told me he was okay with me removing this task that publishes the nugets into the private feed. Still, he was concerned about not knowing if this would "break any dependencies". It seems it doesn't, but if anyone has a reason for not removing this task, please tell me. (@eerhardt ?).

antoniovs1029: So after I mentioned to @codemzs that publishing to the public feed might not always automatically work because of problems on the side of AzureDevOps, he recommended me to put back the task that publishes to the private feed.

Still, I decided to put it back after the other publishing tasks, since if the task of publishing to the private feed fails it would interrupt the other publishing tasks.

If you want to see it before the task of publishing to the public feed, please tell me.

safern: > So after I mentioned to @codemzs that publishing to the public feed might not always automatically work because of problems on the side of AzureDevOps

This is not publishing to the public feed itself, the build definition already has that flakiness because of network connection issues or other unrelated issues to publishing... I think we shouldn't be publishing to both sides as it just slows down builds, it's expensive and nobody will be looking at the internal feed.

build/vsts-ci.yml
safern: Yes, please remove.

build/vsts-ci.yml
```diff
@@ -257,6 +256,14 @@ phases:
       msbuildVersion: 15.0
     continueOnError: true
 
+  - task: NuGetCommand@2
```
safern: [As I mentioned here](https://github.com/dotnet/machinelearning/pull/4406/files#r339876679), I don't think it is worth keeping this.

build/vsts-ci.yml
antoniovs1029: I agree that the problem isn't in the publishing step itself.

I've looked into the builds that were executed in the private pipelines before this PR, and it seems to me that in many cases the nugets weren't even published to the private feed because something went wrong with the pipeline.

After this PR got merged, the automatic build failed at one of the steps that runs before even trying to publish the nugets to the public feed. After Maryam's latest PR got merged an hour later, the automatic build failed while installing a plugin even before building the Windows x86 and Windows x64 versions. I've also tried to manually rerun the builds, but I also get similar problems, and even in one occasion the build failed because it spent over an hour trying to build and sign the Windows x64 binaries. So, so far, the things I've introduced in this PR haven't been actually executed in Azure DevOps in the master branch (I've only tested my changes by running them from a branch in the private repo, and that worked as expected and published the nugets to the feed, but even then it required me to run a couple of manual builds because of the same problems I've already described).

I am somewhat worried about all of these, because it seems that other people will rely on the public feed to be always up-to-date (like @frank-dong-ms nightly build tests). But I wouldn't know if there could be other options, or workarounds to all of this.

antoniovs1029: Ok, so in the last two builds (one manual, and one triggered automatically after @codemzs latest PR), the nugets were published successfully to the public feed, but the pipeline actually failed when publishing to the private feed.

build/vsts-ci.yml
eerhardt: Changing this was a mistake. If you search for `DOTNET_CLI_TELEMETRY_OUTPUT` in all of github, you will find this is the only place this string is used. No one reads an `OUTPUT` environment variable.

But if you search for `DOTNET_CLI_TELEMETRY_OPTOUT` you will see it is used in hundreds of places.

Note: `OPTOUT` stands for "opt out" - as in "don't log CLI telemetry".

safern: You’re completely right I really got confused by reading output and just seems wrong and I recommended to change, my bad @antoniovs1029

build/vsts-ci.yml
antoniovs1029: About that last point, it actually seems that the nugets were correctly published to the feed, but the feed decides (for whatever reason) to select previous published nugets and show them directly in the feed instead of the latest ones. If I go to the 'versions' tab of an specific nugget, I actually can see that it got updated on the builds the last builds, but it is selecting to show the one built 8 hours ago: 
https://dev.azure.com/dnceng/public/_packaging?_a=package&feed=MachineLearning&package=Microsoft.Extensions.ML&protocolType=NuGet&version=1.4.0-preview3-28229-9&view=versions

I am guessing this is, for some reason, the expected behavior, and that I was wrong when saying that the Feed takes time to reflect the changes... it just doesn't display the latest version on the feed.

eerhardt: Let's only have a single feed - the public feed. Having 2 feeds will only increase confusion, storage, and increase the chance of failure.

PR #4407, opened 2019-10-29
SEAL Homomorphic Encryption support

jeremytieman:


eerhardt:
> using System;

all product and test code needs a copyright header

---
Refers to: src/Microsoft.ML.DataView/CiphertextDataViewType.cs:1 in 56f9a75. [](commit_id = 56f9a7517cd18920a8924dc6894d1e9d128a855c, deletion_comment = False)

src/Microsoft.ML.DataView/CiphertextDataViewType.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
+using System.Diagnostics;
+using System.Threading;
+using Microsoft.ML.Data;
+using Microsoft.Research.SEAL;
+
+namespace Microsoft.ML.DataView
+{
+    public class CiphertextDataViewType : StructuredDataViewType
```
eerhardt: All public types and members need xml doc comments

src/Microsoft.ML.DataView/Microsoft.ML.DataView.csproj
```diff
@@ -5,7 +5,12 @@
     <IncludeInPackage>Microsoft.ML.DataView</IncludeInPackage>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <RestoreSources>$(RestoreSources);https://pkgs.dev.azure.com/foundry99/_packaging/SEAL/nuget/v3/index.json</RestoreSources>
+  </PropertyGroup>
+
   <ItemGroup>
+    <PackageReference Include="Microsoft.Research.SEALNet" Version="3.3.1" />
```
eerhardt: We shouldn't be adding new dependencies on this project. People should be able to opt into optional functionality.

The beauty of the IDataView type system is that it is open/extensible. You can add `CiphertextDataViewType` in the `Microsoft.ML.SEAL` project instead. See the Onnx types, and the ImageDataViewType classes for examples.

src/Microsoft.ML.DataView/Microsoft.ML.DataView.csproj
```diff
@@ -5,7 +5,12 @@
     <IncludeInPackage>Microsoft.ML.DataView</IncludeInPackage>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <RestoreSources>$(RestoreSources);https://pkgs.dev.azure.com/foundry99/_packaging/SEAL/nuget/v3/index.json</RestoreSources>
```
eerhardt: All of ML.NET's dependencies need to be on nuget.org. Our users can't use our packages if the dependency graph isn't available to them.

src/Microsoft.ML.Data/Data/SchemaDefinition.cs
```diff
@@ -405,40 +407,44 @@ public static SchemaDefinition Create(Type userType, Direction direction = Direc
...
-                            columnType = new VectorDataViewType(itemType, dims);
+                            columnType = itemType;
                     }
                     else
-                        columnType = itemType;
+                        columnType = DataViewTypeManager.GetDataViewType(dataType, customAttributes);
                 }
-                else
-                    columnType = DataViewTypeManager.GetDataViewType(dataType, customAttributes);
+                else columnType = CiphertextDataViewType.Instance;
```
eerhardt: You shouldn't be adding this hard-coded type here. Instead, look at how we use the `DataViewTypeManager` above to allow for extension types.

src/Microsoft.ML.Data/DataView/TypedCursor.cs
```diff
@@ -288,7 +296,8 @@ private Action<TRow> GenerateSetter(DataViewRow input, int index, InternalSchema
                         Ch.Assert(colType.GetItemType().RawType == Nullable.GetUnderlyingType(fieldType.GetElementType()));
                     else
                         Ch.Assert(colType.GetItemType().RawType == fieldType.GetElementType());
-                    del = CreateDirectVBufferSetter<int>;
+                    if (fieldType.GetElementType() == typeof(Ciphertext)) del = CreateDirectVBufferSetter<Ciphertext>;
```
eerhardt: I don't think this change is necessary. It can be reverted.

eerhardt:
We need to add tests for these new features. We can't add new functionality that isn't being tested.

src/Microsoft.ML.DataView/Microsoft.ML.DataView.csproj
jeremytieman: Unless something has changed since October 7th, the type system in not extensible in the manner needed for this extension. I have shown Zeeshan that I had to make these changes internally to get it to work. All of these changes were initially made in the Microsoft.ML.SEAL project and only moved to an internal project when visibility issues (largely the "internal" keyword on the vast majority of the ML.NET API) prevented access.

src/Microsoft.ML.DataView/Microsoft.ML.DataView.csproj
jeremytieman: Absolutely, I'm just pending the package getting published on Nuget. I had some issues and delays getting the permissions set correctly, but SEAL should be published tonight or tomorrow morning and I will update the PR.

src/Microsoft.ML.DataView/Microsoft.ML.DataView.csproj
eerhardt: > the type system in not extensible in the manner needed for this extension

We should sync on this then. If we need to add more extensibility mechanisms we can to solve this.

src/Microsoft.ML.DataView/CiphertextDataViewType.cs
jeremytieman: Done.

src/Microsoft.ML.SEAL/EncryptedLinearModelParameters.cs
```diff
@@ -0,0 +1,604 @@
...
+
+        private Evaluator _evaluator;
+
+        private CKKSEncoder _ckksEncoder;
+
+        private readonly DataViewType _inputType;
+
+        bool ICanSavePfa.CanSavePfa => true;
+
+        bool ICanSaveOnnx.CanSaveOnnx(OnnxContext ctx) => true;
```
codemzs: >CanSaveOnnx [](start = 26, length = 11)

are you sure this model can also be saved as ONNX?

jeremytieman: There is certainly no reason it can't be saved as a standard SDCA ONNX model. The key to keep in mind about this change is that it only affects evaluation, so all the data is unencrypted and shouldn't have any issue being written out as an ONNX model. Is there something specific you are concerned over?

src/Microsoft.ML.StandardTrainers/IValueMapper.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace Microsoft.ML.Data
+{
+    [BestFriend]
+    internal delegate void ValueMapperEncrypted<TVal1, TVal2, TVal3>(in TVal1 val1, in TVal2 val2, ref TVal3 val3);
+
+    [BestFriend]
+    internal interface IValueMapperEncrypted : IValueMapper
+    {
+        ValueMapperEncrypted<TSrc, TKey, TDst> GetMapper<TSrc, TKey, TDst>();
+    }
```
codemzs: I would put this next to other mapper interfaces in machinelearning\src\Microsoft.ML.Core\Data\IValueMapper.cs. This is really one to two mapper, the name needs to be generic enough that others can also use this and should follow the pattern, so please remove "encrypted.".

test/Microsoft.ML.SEAL/SEALTests.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                keygen.SecretKey.Save(fs);
+            }
+
+            using (var fs2 = File.Open("public.key", FileMode.OpenOrCreate))
+            {
+                keygen.PublicKey.Save(fs2);
+            }
+
+            var encryptPipeline = Microsoft.ML.SEAL.SealCatalog.EncryptFeatures(true,
```
codemzs: >SealCatalog [](start = 52, length = 11)

the word catalog should not be there, please add a new catalog entry. It needs to be something like context.Transforms.Encryption.ABC(....), refer [https://github.com/dotnet/machinelearning/tree/master/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms](https://github.com/dotnet/machinelearning/tree/master/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms)

src/Microsoft.ML.StandardTrainers/IValueMapper.cs
jeremytieman: OK, but it will have to be named IValueMapperOneToTwo (or another name) to avoid a name conflict. And really, isn't it more a two inputs to one output mapper? Should it be IValueMapperTwoToOne?

test/Microsoft.ML.SEAL/SEALTests.cs
```diff
@@ -0,0 +1,72 @@
...
+                    18014398492704769,
+                    18014398491918337
+                },
+                "plaintext",
+                "ciphertext"));
+
+            var data = new[] { 1.1, 2.2, 3.3, 4.4 };
+            var dataView = ML.Data.LoadFromEnumerable(data);
+            var model = encryptPipeline.Fit(dataView);
+            var engine = model.CreatePredictionEngine<TestClass, TestClass>(ML);
```
codemzs: >var engine = model.CreatePredictionEngine<TestClass, TestClass>(ML); [](start = 12, length = 68)

can you please predict some values and add asserts to ensure decrypted values is the same as pre-encrypted value,

test/Microsoft.ML.SEAL/SEALTests.cs
```diff
@@ -0,0 +1,72 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.ML.RunTests;
+using Microsoft.Research.SEALNet;
+using Xunit;
+
+namespace Microsoft.ML.Tests.SEAL
+{
+    public sealed class SEALTests : TestDataPipeBase
```
codemzs: > public sealed class SEALTests : TestDataPipeBase [](start = 3, length = 49)

Add a test that mimics real world scenario:

Pipeline A takes plain data and encrypts it.
Pipeline B accepts encrypted data and performs inference on it.
Pipeline C accepts encrypted inferred data and decrypts it. 

In the samples I hope the above is done using multiple processes and IPC.

src/Microsoft.ML.StandardTrainers/IValueMapper.cs
codemzs: sorry, yes, Two to One, that is what I had meant :)

src/Microsoft.ML.StandardTrainers/Microsoft.ML.StandardTrainers.csproj
```diff
@@ -6,6 +6,10 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Research.SEALNet" Version="$(MicrosoftResearchSealPackageVersion)" />
```
eerhardt: The new types in this assembly should be moved to the `Microsoft.ML.SEAL` assembly as well. We shouldn't add any new dependencies to the core Microsoft.ML package (which the StandardTrainers assembly is a part of).

PR #4408, opened 2019-10-29
Rename Microsoft.ML.Dnn to Microsoft.ML.Vision and reverse dependency between Microsoft.ML.TensorFlow and Microsoft.ML.Dnn

### LINKED ISSUE #4305: ML.TensorFlow has a dependency on ML.Dnn
I don't think the dependency from `ML.TensorFlow` to `ML.Dnn` is correct. My understanding of `ML.Dnn` is that it is a higher-level abstraction for deep neural networks, of which TensorFlow is the first one we are using. However, in the future, we may want to add more neural network libraries to `ML.Dnn` - for example, PyTorch. If that ever happens, now the `ML.TensorFlow` library will also be dependent on PyTorch, since it will have a transitive reference from `ML.Dnn` to PyTorch.

We should rethink the dependency structure here. If the only reason `ML.TensorFlow` has a dependency on `ML.Dnn` is for sharing internal code, we should come up with a different code sharing mechanism. For example, the way we solve this in dotnet/corefx is we use share the same source code, and compile it directly into multiple assemblies using source links in the .csprojs.

codemzs:
- Refactors Microsoft.ML.TensorFlow and Microsoft.ML.Dnn such that Microsoft.ML.Dnn depends on Microsoft.ML.TensorFlow and not vice-versa and fixes #4305.

- Renames Microsoft.ML.Dnn nuget and namespace to Microsoft.ML.Vision.


src/Microsoft.ML.Vision/VisionCatalog.cs
```diff
@@ -3,13 +3,13 @@
...
-using Microsoft.ML.Dnn;
 using Microsoft.ML.Runtime;
 using Microsoft.ML.Transforms;
+using Microsoft.ML.Vision;
+using static Microsoft.ML.TensorFlow.TensorFlowUtils;
 
 namespace Microsoft.ML
 {
-    /// <include file='doc.xml' path='doc/members/member[@name="DnnRetrainTransformer"]/*' />
     public static class DnnCatalog
```
eerhardt: Should this be renamed to `VisionCatalog`?

codemzs: oh sorry, my bad, I thought I had renamed it.

codemzs: It shows I have renamed it on my side!

codemzs:
>     internal sealed class DnnRetrainTransformer : RowToRowTransformerBase

agreed, that is what I was thinking as well.

---
In reply to: [547494674](https://github.com/dotnet/machinelearning/pull/4408#issuecomment-547494674) [](ancestors = 547494674)

---
Refers to: src/Microsoft.ML.Dnn/DnnRetrainTransform.cs:40 in 9af64bb. [](commit_id = 9af64bb44a126ecf992221aeaa6d1fd644528b7a, deletion_comment = False)

src/Microsoft.ML.TensorFlow/TensorflowUtils.cs
```diff
@@ -98,6 +103,138 @@ internal static DataViewSchema GetModelSchema(IHostEnvironment env, string model
...
+        {
+            var graph = new Graph();
+            try
+            {
+                graph.Import(modelBytes, "");
+            }
+            catch (Exception ex)
+            {
+                if (!string.IsNullOrEmpty(modelFile))
+                    throw ectx.Except($"TensorFlow exception triggered while loading model from '{modelFile}'");
```
eerhardt: This should have the `ex` as an inner exception. So that information doesn't get lost.

src/Microsoft.ML.Vision/VisionCatalog.cs
eerhardt: It is not showing as renamed in the review. You renamed the file, but not the class.

---
In reply to: [340166729](https://github.com/dotnet/machinelearning/pull/4408#discussion_r340166729) [](ancestors = 340166729)

eerhardt:
>     public abstract class LearningRateScheduler

It doesn't feel like `LearningRateScheduler` belongs in the `Microsoft.ML.Vision` namespace or package. It may make sense to put it in the `Microsoft.ML.Trainers` namespace.

I'm not sure on the right package. Maybe in `Microsoft.ML.TensorFlow` or even maybe putting it in the core `Microsoft.ML` package? It seems like something we would want to use in other tasks besides Vision - like Audio or Text.

---
Refers to: src/Microsoft.ML.Vision/LearningRateScheduler.cs:24 in 9af64bb. [](commit_id = 9af64bb44a126ecf992221aeaa6d1fd644528b7a, deletion_comment = False)

eerhardt:
>     public abstract class LearningRateScheduler

One idea is that we can put it in the `Microsoft.ML.Trainers` namespace now (the namespace can never change), but leave it in the `Microsoft.ML.Vision` package/assembly.  The type can move to a lower assembly later if we use the [TypeForwardedToAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.typeforwardedtoattribute) in the future. So we should be able to extract it later, if necessary.

However, the namespace needs to be decided upon now.

---
In reply to: [547516624](https://github.com/dotnet/machinelearning/pull/4408#issuecomment-547516624) [](ancestors = 547516624)

---
Refers to: src/Microsoft.ML.Dnn/LearningRateScheduler.cs:24 in 9af64bb. [](commit_id = 9af64bb44a126ecf992221aeaa6d1fd644528b7a, deletion_comment = False)

src/Microsoft.ML.TensorFlow/TensorflowUtils.cs
codemzs: thanks!

---
In reply to: [340221260](https://github.com/dotnet/machinelearning/pull/4408#discussion_r340221260) [](ancestors = 340221260)

src/Microsoft.ML.StandardTrainers/Properties/AssemblyInfo.cs
```diff
@@ -16,3 +16,6 @@
 [assembly: InternalsVisibleTo(assemblyName: "ParameterMixer" + InternalPublicKey.Value)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Runtime.SequencePrediction" + InternalPublicKey.Value)]
 [assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Runtime.Sar" + InternalPublicKey.Value)]
+[assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Vision" + PublicKey.Value)]
+[assembly: InternalsVisibleTo(assemblyName: "Microsoft.ML.Samples" + PublicKey.Value)]
```
eerhardt: This shouldn't happen. We can't have IVT our samples.

PR #4413, opened 2019-10-29
Initial featurizers project

michaelgsharp:
This is the initial featurizers PR to split apart PR #4157 and get all the common code into master.
Once this code goes in, I will create separate PR's for each of the 5 featurizers.

This is the common code for the featurizers. Its mostly changes to project files and solution files. Common.cs is for all the shared code for the featurizers. The change in Utils.cs is to allow `Marshal.Invoke` with multiple type parameters. The RowToRowMapperTransform.cs change has it create a new mapper when possible. This helps with thread safety/local caching of the mappers when run in a multi threaded approach.


pkg/Microsoft.ML.Featurizers/Microsoft.ML.Featurizers.nupkgproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk" DefaultTargets="Pack">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageDescription>Additional ML.NET featurizers for AutoML</PackageDescription>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ML" Version="1.4.0-preview3*" />
```
eerhardt: This should be a `ProjectReference`. See our other .nupkgproj files for an example.

pkg/Microsoft.ML.Featurizers/Microsoft.ML.Featurizers.nupkgproj
```diff
@@ -0,0 +1,15 @@
...
+    <PackageReference Include="Microsoft.ML.DataView" Version="1.4.0-preview3*" />
```
eerhardt: This can be removed since it is brought in transitively from referencing Microsoft.ML.

Directory.Build.props
```diff
@@ -19,6 +19,7 @@
       https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json;
       https://dotnet.myget.org/F/dotnet-core/api/v3/index.json;
       https://dotnet.myget.org/F/roslyn-analyzers/api/v3/index.json;
+      https://pkgs.dev.azure.com/dnceng/public/_packaging/MachineLearning/nuget/v3/index.json;
```
eerhardt: This should be removed.

src/Microsoft.ML.Featurizers/Common.cs
```diff
@@ -0,0 +1,189 @@
...
+
+    internal enum FitResult : byte
+    {
+        Complete = 1, Continue, ResetAndContinue
+    }
+
+    // Not all these types are currently supported. This is so the ordering will allign with the native code.
+    internal enum TypeId : uint
+    {
+        String = 1, SByte, Short, Int, Long, Byte, UShort,
```
eerhardt: Typical C# code has these be on separate lines.

src/Microsoft.ML.Featurizers/Common.cs
```diff
@@ -0,0 +1,189 @@
...
+    {
+        public byte* Data;
+        public IntPtr DataSize;
+    }
+
+    #region SafeHandles
+
+    internal class ErrorInfoSafeHandle : SafeHandleZeroOrMinusOneIsInvalid
+    {
+        [DllImport("Featurizers", EntryPoint = "DestroyErrorInfo", CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity]
```
eerhardt: `Featurizers` seems like a pretty generic name for this native assembly. Is that the final name? Since it comes from the `Microsoft.MLFeaturizers` package, it may make sense to use `MLFeaturizers` as the native assembly name.

pkg/Microsoft.ML.Featurizers/Microsoft.ML.Featurizers.nupkgproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk" DefaultTargets="Pack">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageDescription>Additional ML.NET featurizers for AutoML</PackageDescription>
```
eerhardt: Is `for AutoML` important enough to use in the description of what this package is?

Directory.Build.props
michaelgsharp: This goes along with question about project/package reference. I added this after the email we had yesterday, but it sounds like this should only exist in the feature branch and the code that gets checked into master wont have this. Is that a correct understanding?

---
In reply to: [340734033](https://github.com/dotnet/machinelearning/pull/4413#discussion_r340734033) [](ancestors = 340734033)

pkg/Microsoft.ML.Featurizers/Microsoft.ML.Featurizers.nupkgproj
michaelgsharp: Sounds good. Set it back to the `ProjectReference`.

---
In reply to: [340748512](https://github.com/dotnet/machinelearning/pull/4413#discussion_r340748512) [](ancestors = 340748512)

Directory.Build.props
michaelgsharp: Per the comment to my other question, I will revert this here as well and only have it in the feature branch.

---
In reply to: [340748790](https://github.com/dotnet/machinelearning/pull/4413#discussion_r340748790) [](ancestors = 340748790,340734033)

src/Microsoft.ML.Featurizers/Common.cs
michaelgsharp: That is the name as it currently stands. Closing as `ByDesign` for now but I will sync with the team and change this if the name gets updated.

---
In reply to: [340737413](https://github.com/dotnet/machinelearning/pull/4413#discussion_r340737413) [](ancestors = 340737413)

src/Microsoft.ML.Featurizers/Common.cs
justinormont: Should we have a int next to each? This will assist with long term compatibility by allowing for inserting new values, and in case someone decides to sort the enum differently later (more important if serialized in a model).

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -966,6 +966,14 @@ private static MethodInfo MarshalInvokeCheckAndCreate<TRet>(Type genArg, Delegat
             return meth;
         }
 
+        private static MethodInfo MarshalInvokeCheckAndCreate<TRet>(Delegate func, Type[] genArgs)
+        {
+            var meth = MarshalActionInvokeCheckAndCreate(func, genArgs);
+            if (meth.ReturnType != typeof(TRet))
+                throw Contracts.ExceptParam(nameof(func), "Cannot be generic on return type");
+            return meth;
```
justinormont: I don't think `method` is a reserved keyword in C#. Recommend the longer name.
```suggestion
            return method;
```

pkg/Microsoft.ML.Featurizers/Microsoft.ML.Featurizers.nupkgproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk" DefaultTargets="Pack">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageDescription>Additional ML.NET featurizers</PackageDescription>
```
codemzs: >Additional ML.NET featurizers [](start = 24, length = 29)

ML .NET featurizers with native code implementation.

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -966,6 +966,14 @@ private static MethodInfo MarshalInvokeCheckAndCreate<TRet>(Type genArg, Delegat
             return meth;
         }
 
+        private static MethodInfo MarshalInvokeCheckAndCreate<TRet>(Delegate func, Type[] genArgs)
```
codemzs: >Delegate func, Type[] genArgs) [](start = 68, length = 30)

lets keep the order consistent types then delegate

src/Microsoft.ML.Core/Utilities/Utils.cs
```diff
@@ -1104,6 +1134,18 @@ private static MethodInfo MarshalActionInvokeCheckAndCreate(Type genArg, Delegat
             return meth;
         }
 
+        private static MethodInfo MarshalActionInvokeCheckAndCreate(Delegate func, params Type[] typeArguments)
```
codemzs: >(Delegate func, params Type[] typeArguments) [](start = 67, length = 44)

same here, lets keep the order consistent, types then delegate

src/Microsoft.ML.Featurizers/Common.cs
```diff
@@ -0,0 +1,204 @@
...
+
+namespace Microsoft.ML.Featurizers
+{
+    #region Native Function Declarations
+
+    #endregion
+
+    internal enum FitResult : byte
+    {
+        Complete = 1, Continue, ResetAndContinue
```
codemzs: >Complete = 1, Continue, ResetAndContinue [](start = 8, length = 40)

put each on its own line and explicitly define values.

src/Microsoft.ML.Featurizers/Common.cs
```diff
@@ -0,0 +1,204 @@
...
+        Timepoint,
+        Duration,
+
+        LastStaticValue,
+        Tensor = 0x1001 | LastStaticValue + 1,
+        SparseTensor = 0x1001 | LastStaticValue + 2,
+        Tabular = 0x1001 | LastStaticValue + 3,
+        Nullable = 0x1001 | LastStaticValue + 4,
+        Vector = 0x1001 | LastStaticValue + 5,
+        MapId = 0x1002 | LastStaticValue + 6
```
codemzs: >MapId = 0x1002 | LastStaticValue + 6 [](start = 8, length = 36)

please add comments for this

src/Microsoft.ML.Featurizers/Common.cs
```diff
@@ -0,0 +1,204 @@
...
+                GetErrorInfoString(errorHandle, out IntPtr errorHandleString, out IntPtr errorHandleStringSize);
+                using (var errorString = new ErrorInfoStringSafeHandle(errorHandleString, errorHandleStringSize))
+                {
+                    byte[] buffer = new byte[errorHandleStringSize.ToInt32()];
+                    Marshal.Copy(errorHandleString, buffer, 0, buffer.Length);
+
+                    return Encoding.UTF8.GetString(buffer);
+                }
+            }
+        }
```
codemzs: >} [](start = 8, length = 1)

new line

src/Microsoft.ML.Featurizers/Common.cs
```diff
@@ -0,0 +1,204 @@
...
+        {
+            if (type == typeof(byte))
+                return TypeId.Byte;
+            else if (type == typeof(short))
+                return TypeId.Short;
+            else if (type == typeof(int))
+                return TypeId.Int;
+            else if (type == typeof(long))
+                return TypeId.Long;
+            else if (type == typeof(byte))
```
codemzs: >(byte)) [](start = 35, length = 7)

sbyte

src/Microsoft.ML.Featurizers/Common.cs
```diff
@@ -0,0 +1,204 @@
...
+                return TypeId.Byte;
```
codemzs: >Byte [](start = 30, length = 4)

SByte

codemzs:
> 

Can you please add comments for these functions? and how they are used?

---
Refers to: src/Microsoft.ML.Featurizers/Common.cs:205 in 49f3798. [](commit_id = 49f379801928f9468e794ae9a532092ed295efb6, deletion_comment = False)

src/Microsoft.ML.Data/DataView/RowToRowMapperTransform.cs
```diff
@@ -367,7 +367,7 @@ public Cursor(IChannelProvider provider, DataViewRowCursor input, RowToRowMapper
                 : base(provider, input)
             {
                 var pred = parent.GetActiveOutputColumns(active);
-                _getters = parent._mapper.CreateGetters(input, pred, out _disposer);
```
codemzs: >_getters = parent._mapper.CreateGetters(input, pred, out _disposer); [](start = 16, length = 68)

lets revert this and use a thread safe dictionary for now. I want to think a little more on this.

src/Microsoft.ML.Core/Utilities/Utils.cs
michaelgsharp: I kept it as `meth` because that is consistent with all the other methods here that do this same thing. I think if we want it to be `method` then we need to change them all.

---
In reply to: [347619466](https://github.com/dotnet/machinelearning/pull/4413#discussion_r347619466) [](ancestors = 347619466)

src/Microsoft.ML.Featurizers/Common.cs
michaelgsharp: Yeah, you are correct. I have added an int next to each one now.

justinormont: Thanks

src/Microsoft.ML.Featurizers/Common.cs
justinormont: Why does this add `LastStaticValue`? That shifts all of the lower items by 1 everytime a new type is inserted.

Why the math? Seems more simple to directly map to the values.

Why is the base of the last one `0x1002`? That skips a value.

michaelgsharp: All of these are taken directly as is from the Native code base. They are here just so we have a correct mapping on our side.

`LastStaticValue` is supposed to represent the last stand alone type. All the types that follow after it are compound types. There is probably a more appropriate name, but since I want to keep this identical to the Native code I left it as is.

The math is there just to keep it identical to the native code. I could directly map the values, but then I would have to map the values every time I need to check if it has changed in the native code. Leaving it like this makes comparisons much easier.

The number at the end of `0x100N` represents how many types come after it that are associated to it. So `Tabular`, `Nullable`, and `Vector`, end in `1` meaning only 1 type follows it. The `Map` ends in `2` because we need types for the Keys and for the Values.

---
In reply to: [348115118](https://github.com/dotnet/machinelearning/pull/4413#discussion_r348115118) [](ancestors = 348115118)

codemzs:
@michaelgsharp Can you please sync to master?

PR #4421, opened 2019-10-30
Cleanup: makes Recommendation sample easier to use

maryamariyan:
This PR makes the Recommendation sample under docs easier to use, by pointing to an existing sample in machinelearning-samples repo:

Using the same datasets under [https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/getting-started/MatrixFactorization_MovieRecommendation/Data](https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/getting-started/MatrixFactorization_MovieRecommendation/Data)

The model generated using instructions in [here](https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/getting-started/MatrixFactorization_MovieRecommendation), produces metrics below:
```
MeanAbsoluteError: 0.619940823978848
MeanSquaredError: 0.947787365931922
RootMeanSquaredError: 0.973543715470406
RSquared: 0.43654475860194
```

Whereas using AutoML for Recommendation task with just 540 seconds training time (as seen in this PR), we can get a little over 100 models generated out of which the best produced has the improved metrics below:
```
Metrics of best model on test data --
MeanAbsoluteError: 0.578644904825423
MeanSquaredError: 0.747918695002957
RootMeanSquaredError: 0.864822926964218
RSquared: 0.555365766640444
```
--- 

Additionally in this PR I updated the CodeGeneratorTests a bit so that, the code generated in 
- CodeGeneratorTests.Recommendation_GenerateConsoleAppProjectContents_VerifyPredictProgram
- CodeGeneratorTests.Recommendation_GenerateConsoleAppProjectContents_VerifyPredictProject

can be used to test the model generated from running the Recommendation experiment in the doc\samples folder.

cc: @CESARDELATORRE, @eerhardt @LittleLittleCloud

### DIFF
--- a/docs/samples/Microsoft.ML.AutoML.Samples/DataStructures/Movie.cs
+++ b/docs/samples/Microsoft.ML.AutoML.Samples/DataStructures/Movie.cs
@@ -8,13 +8,19 @@ namespace Microsoft.ML.AutoML.Samples.DataStructures
 {
     public class Movie
     {
-        [LoadColumn(0)]
-        public string UserId;
+        [ColumnName("userId"), LoadColumn(0)]
+        public float UserId { get; set; }
 
-        [LoadColumn(1)]
-        public string MovieId;
 
-        [LoadColumn(2)]
-        public float Rating;
+        [ColumnName("movieId"), LoadColumn(1)]
+        public float MovieId { get; set; }
+
+
+        [ColumnName("rating"), LoadColumn(2)]
+        public float Rating { get; set; }
+
+
+        [ColumnName("timestamp"), LoadColumn(3)]
+        public float Timestamp { get; set; }
     }
 }
--- a/test/Microsoft.ML.CodeGenerator.Tests/ApprovalTests/ConsoleCodeGeneratorTests.cs
+++ b/test/Microsoft.ML.CodeGenerator.Tests/ApprovalTests/ConsoleCodeGeneratorTests.cs
@@ -286,7 +286,7 @@ namespace mlnet.Tests
             (Pipeline pipeline,
                        ColumnInferenceResults columnInference) = GetMockedRecommendationPipelineAndInference();
 
-            var consoleCodeGen = new CodeGenerator(pipeline, columnInference, CreateCodeGeneratorSettingsFor(TaskKind.Recommendation));
+            var consoleCodeGen = new CodeGenerator(pipeline, columnInference, CreateCodeGeneratorSettingsFor(TaskKind.Recommendation, columnInference.ColumnInformation.LabelColumnName));
             return consoleCodeGen;
         }
 
@@ -297,7 +297,7 @@ namespace mlnet.Tests
                 MLContext context = new MLContext();
 
                 var trainer1 = new SuggestedTrainer(context, new MatrixFactorizationExtension(), new ColumnInformation() {
-                    LabelColumnName = "Label",
+                    LabelColumnName = "rating",
                     UserIdColumnName = "userId",
                     ItemIdColumnName = "movieId",
                 }, hyperParamSet: null);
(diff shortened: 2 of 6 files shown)


test/Microsoft.ML.CodeGenerator.Tests/ApprovalTests/ConsoleCodeGeneratorTests.cs
```diff
@@ -308,9 +308,10 @@ private CodeGenerator PrepareForRecommendationTask()
                 var textLoaderArgs = new TextLoader.Options()
                 {
                     Columns = new[] {
-                        new TextLoader.Column("Label", DataKind.String, 0),
-                        new TextLoader.Column("userId", DataKind.String, 1),
-                        new TextLoader.Column("movieId", DataKind.String, 2),
+                        new TextLoader.Column("userId", DataKind.Single, 0),
+                        new TextLoader.Column("movieId", DataKind.Single, 1),
+                        new TextLoader.Column("rating", DataKind.Single, 2),
+                        new TextLoader.Column("timestamp", DataKind.Single, 3),
```
eerhardt: Why load this column here, only to ignore it later?

docs/samples/Microsoft.ML.AutoML.Samples/DataStructures/Movie.cs
```diff
@@ -8,13 +8,19 @@ namespace Microsoft.ML.AutoML.Samples.DataStructures
...
+        [ColumnName("movieId"), LoadColumn(1)]
+        public float MovieId { get; set; }
+
+
+        [ColumnName("rating"), LoadColumn(2)]
+        public float Rating { get; set; }
+
+
+        [ColumnName("timestamp"), LoadColumn(3)]
+        public float Timestamp { get; set; }
```
eerhardt: Can this just be removed?

maryamariyan: It could, but also from in the CodeGeneratorTests you can see that column is added to IgnoredColumnNames collection in ColumnInformation, to make it hidden,

I am working with an existing dataset which already contains timestamp, this sample shows how to ignore it.

The benefit of keeping it this way is at a later time when we add FFM for recommendation, we could show how that column can be used as a user feature.

test/Microsoft.ML.CodeGenerator.Tests/ApprovalTests/ConsoleCodeGeneratorTests.cs
maryamariyan: Only because the sample in machinelearning-samples repo already has that column and I want to show it can be used and ignored at the same time.

I would have not included this if the existing dataset did not come with that column. It's added just for convenience of not wanting to add a whole new dataset to just make this work. (easier to later compare with FFM and to currently compare with the sample in machinelearning-samples repo), since they all use the same dataset.

docs/samples/Microsoft.ML.AutoML.Samples/DataStructures/Movie.cs
maryamariyan: This is potentially another use case for having DataFrame; If we had DataFrame, we could have loaded the dataset as IDataView without needing it to match the Movie class in the statement below:
```
IDataView trainDataView = mlContext.Data.LoadFromTextFile<Movie>(TrainDataPath, hasHeader: true, separatorChar: ',');
```
Then the Movie class wouldn't have needed Timestamp field.

cc: @pgovind

maryamariyan:
CI failure is unrelated to my change:

```
[xUnit.net 00:00:04.05]     Microsoft.Extensions.ML.UriLoaderTests.can_reload_model [FAIL]
  X Microsoft.Extensions.ML.UriLoaderTests.can_reload_model [490ms]
  Error Message:
   Assert.True() Failure
Expected: True
Actual:   False
```

I won't be able to hit the merge button until all CI checks succeed. 
@eerhardt could we just merge this as is?

PR #4424, opened 2019-10-31
Modified how data is saved to disk

bpstark:
pre-trained meta files are now stored in one location always, this
allows multiple runs to re-use the same meta file without having to
redownload.

Additionally added the ability to cleanup the temporary workspace used
to train the model. This should prevent issues of running out of disk
space when running multiple training session sequentially.

### DIFF
--- a/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
+++ b/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
@@ -1144,7 +1144,7 @@ namespace Microsoft.ML.Vision
                 {
                     Directory.Delete(_options.WorkspacePath, true);
                 }
-                catch (Exception e)
+                catch (Exception)
                 {
                     //We do not want to stop pipeline due to failed cleanup.
                 }
--- a/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
+++ b/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
@@ -1154,7 +1154,6 @@ namespace Microsoft.ML.Vision
             }
         }
 
-
         private (Session, Tensor, Tensor, Tensor) BuildEvaluationSession(int classCount)
         {
             var evalGraph = LoadMetaGraph(Path.Combine(Path.GetTempPath(), ResourceDir, ModelFileName[_options.Arch]));
--- a/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
+++ b/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
@@ -1319,9 +1319,10 @@ namespace Microsoft.ML.Vision
             DownloadIfNeeded(env, modelFileName, _resourcePath, modelFileName, timeout);
             if (arch == Architecture.InceptionV3)
             {
+                var extrasPath = Path.Combine(_resourcePath, @"tfhub_modules");
                 DownloadIfNeeded(env, @"tfhub_modules.zip", _resourcePath, @"tfhub_modules.zip", timeout);
-                if (!Directory.Exists(@"tfhub_modules"))
-                    ZipFile.ExtractToDirectory(Path.Combine(_resourcePath, @"tfhub_modules.zip"), Path.Combine(_resourcePath, @"tfhub_modules"));
+                if (!Directory.Exists(extrasPath))
+                    ZipFile.ExtractToDirectory(Path.Combine(_resourcePath, @"tfhub_modules.zip"), extrasPath);
             }
 
             return new TensorFlowSessionWrapper(GetSession(env, modelFilePath, true), modelFilePath);
--- a/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
+++ b/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
@@ -475,6 +475,7 @@ namespace Microsoft.ML.Vision
         private int _classCount;
         private Graph Graph => _session.graph;
         private static readonly string _resourcePath = Path.Combine(Path.GetTempPath(), "MLNET");
+        private readonly string _sizeFile;
 
         /// <summary>
         /// Initializes a new instance of <see cref="ImageClassificationTrainer"/>
@@ -544,6 +545,7 @@ namespace Microsoft.ML.Vision
             _useLRScheduling = _options.LearningRateScheduler != null;
             _checkpointPath = Path.Combine(_options.WorkspacePath, _options.FinalModelPrefix +
                     ModelFileName[_options.Arch]);
+            _sizeFile = Path.Combine(_options.WorkspacePath, "TrainingSetSize.txt");
 
             // Configure bottleneck tensor based on the model.
             var arch = _options.Arch;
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -519,6 +527,17 @@ internal ImageClassificationTrainer(IHostEnvironment env, Options options)
             {
                 options.WorkspacePath = GetTemporaryDirectory();
             }
+            else
+            {
+                //Since this is not a tmp workspace, but one provided by the user, do not allow deletion.
+                options.CleanTmpWorkspace = false;
+            }
+
+            string resourcePath = Path.Combine(Path.GetTempPath(), ResourceDir);
```
codemzs: >string resourcePath [](start = 12, length = 19)

make it a readonly property of the class.

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -78,6 +78,7 @@ public sealed class ImageClassificationTrainer :
         internal const string UserName = "Image Classification Trainer";
         internal const string ShortName = "IMGCLSS";
         internal const string Summary = "Trains a DNN model to classify images.";
+        internal const string ResourceDir = "MLNET_VISION";
```
codemzs: >MLNET_VISION [](start = 45, length = 12)

MLNET

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -438,6 +439,13 @@ public sealed class Options : TrainerInputBaseWithLabel
             /// </summary>
             [Argument(ArgumentType.AtMostOnce, HelpText = "A class that performs learning rate scheduling.", SortOrder = 15)]
             public LearningRateScheduler LearningRateScheduler = new LsrDecay();
+
+            /// <summary>
+            /// Indicates if the temporary workspace should be deleted on cleanup
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Indicates if the temporary workspace should be deleted on cleanup", SortOrder = 15)]
+            public bool CleanTmpWorkspace = true;
```
codemzs: remove.

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -519,6 +527,17 @@ internal ImageClassificationTrainer(IHostEnvironment env, Options options)
             {
                 options.WorkspacePath = GetTemporaryDirectory();
             }
+            else
+            {
+                //Since this is not a tmp workspace, but one provided by the user, do not allow deletion.
+                options.CleanTmpWorkspace = false;
```
codemzs: add a readonly private variable and set it to true indicating we need to cleanup workspace in the event user does not provide one.

codemzs:
> using System.Collections.Generic;

add/modify existing tests to check for below:
1) meta files exist in temp dir/mlnet 
2) the file that contains number of examples also exists in the workspace directory after the run is completed (this is when user provides a workspace).
3) we need to think how we can test the scenario when workspace is not passed by the user and we create a temp workspace in temp dir and we later clean it.... we need to verify the clean step but I understand getting hold of this temp directory is challenging. 

---
Refers to: src/Microsoft.ML.Vision/ImageClassificationTrainer.cs:6 in 9b1d5b1. [](commit_id = 9b1d5b116c8566202bdbfa095235eb1749333314, deletion_comment = False)

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -1123,11 +1138,19 @@ private int GetNumSamples(string path)
 
             trainSaver.save(_session, _checkpointPath);
             UpdateTransferLearningModelOnDisk(_classCount);
+            CleanUpTmpWorkspace();
+        }
```
codemzs: newline.

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -1123,11 +1138,19 @@ private int GetNumSamples(string path)
...
+        private void CleanUpTmpWorkspace()
```
codemzs: >CleanUpTmpWorkspace() [](start = 21, length = 21)

CleanupTemporaryWorkspace

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -471,9 +471,12 @@ public sealed class Options : TrainerInputBaseWithLabel
         private readonly string _checkpointPath;
         private readonly string _bottleneckOperationName;
         private readonly bool _useLRScheduling;
+        private readonly bool _cleanupWorkspace;
         private int _classCount;
         private Graph Graph => _session.graph;
 
+        private static readonly string _resourcePath = Path.Combine(Path.GetTempPath(), "MLNET");
```
codemzs: why is there a newline between this and the previous field?

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -518,6 +521,16 @@ internal ImageClassificationTrainer(IHostEnvironment env, Options options)
             if (string.IsNullOrEmpty(options.WorkspacePath))
             {
                 options.WorkspacePath = GetTemporaryDirectory();
+                _cleanupWorkspace = true;
+            }
+            else
+            {
+                _cleanupWorkspace = false;
+            }
```
codemzs: No need for this, by default in C# boolean variables are initialized to false.

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -630,8 +638,9 @@ private protected override ImageClassificationModelParameters TrainModelCore(Tra
                     _inputTensorName, _bottleneckTensor.name, trainSetBottleneckCachedValuesFilePath,
                     ImageClassificationMetrics.Dataset.Train, _options.MetricsCallback);
 
+                string sizeFile = Path.Combine(_options.WorkspacePath, "TrainingSetSize.txt");
```
eerhardt: (minor) you may want to extract this into a helper method so it isn't duplicated where this file is, and what its name is.

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -1123,11 +1133,20 @@ private int GetNumSamples(string path)
...
             trainSaver.save(_session, _checkpointPath);
             UpdateTransferLearningModelOnDisk(_classCount);
+            TryCleanupTemporaryWorkspace();
+        }
+
+        private void TryCleanupTemporaryWorkspace()
+        {
+            if (_cleanupWorkspace && Directory.Exists(_options.WorkspacePath))
+            {
+                Directory.Delete(_options.WorkspacePath, true);
```
eerhardt: Do you want a try-catch here? Or if this throws an exception, do you want it to kill the training process?

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -1123,11 +1133,27 @@ private int GetNumSamples(string path)
 
             trainSaver.save(_session, _checkpointPath);
             UpdateTransferLearningModelOnDisk(_classCount);
+            TryCleanupTemporaryWorkspace();
+        }
+
+        private void TryCleanupTemporaryWorkspace()
+        {
+            if (_cleanupWorkspace && Directory.Exists(_options.WorkspacePath))
```
justinormont: Assuming you hold the file pointers open, an elegant method is using the `FileOptions.DeleteOnClose` when you open the temporary files. 

This then causes the OS (or filesystem when on macOS/Linux) to cleanup the files when they are closed, or if your program crashes or is killed by the user.

https://docs.microsoft.com/en-us/dotnet/api/system.io.fileoptions?view=netcore-3.0

bpstark:
Note this change will continue to fail until such time as we update the inception meta file in storage

This will fix #4426

PR #4438, opened 2019-11-04
Recreate workspace directory when fit() is called, fix documentation, free up unmanaged memory.

codemzs:


src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -392,7 +393,7 @@ public sealed class Options : TrainerInputBaseWithLabel
             public Action<ImageClassificationMetrics> MetricsCallback = null;
 
             /// <summary>
-            /// Indicates the path where the models get downloaded to and cache files saved, default is a new temporary directory
+            /// Indicates the path where the image bottleneck cache files are saved, default is a new temporary directory
```
ashbhandare: > image bottleneck [](start = 44, length = 17)

This is also the path where the trained model is saved, so we could add that here.

codemzs: Done.

PR #4443, opened 2019-11-05
Hash Transform API that takes in advanced options.

### LINKED ISSUE #4422: ML.net doesn't exposes useOrderedHashing input parameter for Hash transformation
### System information

- **OS version/distro**:

### Issue

- **What did you do?**
Using Conversion Hash transform.
https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.conversionsextensionscatalog.hash?view=ml-dotnet-1.0.0

- **What happened?**

It doesn't exposes property useOrderedHashing 

- **What did you expect?**

Expose useOrderedHashing  property as this is blocking us to move to ML.net.

### Source code / logs
See link .
https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.conversionsextensionscatalog.hash?view=ml-dotnet-1.0.0

codemzs:
fixes #4422

Needed by a 1P customer.

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
+++ b/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
@@ -47,8 +47,8 @@ namespace Microsoft.ML
             => new HashingEstimator(CatalogUtils.GetEnvironment(catalog), outputColumnName, inputColumnName, numberOfBits, maximumNumberOfInverts);
 
         /// <summary>
-        /// Create a <see cref="HashingEstimator"/>, which hashes the input column's data type <see cref="InputOutputColumnPair.InputColumnName" />
-        /// to a new column: <see cref="InputOutputColumnPair.OutputColumnName" />.
+        /// Create a <see cref="HashingEstimator"/>, which hashes the input column's data type <see cref="ColumnOptions.InputColumnName" />
+        /// to a new column: <see cref="ColumnOptions.Name" />.
         /// </summary>
         /// <remarks>This transform can operate over several columns.</remarks>
         /// <param name="catalog">The transform's catalog.</param>
--- a/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
+++ b/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
@@ -10,6 +10,7 @@ using Microsoft.ML.Transforms;
 
 namespace Microsoft.ML
 {
+    using static Microsoft.ML.Transforms.HashingEstimator;
     using ConvertDefaults = TypeConvertingEstimator.Defaults;
     using HashDefaults = HashingEstimator.Defaults;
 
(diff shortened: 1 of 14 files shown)


src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
```diff
@@ -51,13 +51,18 @@ public static class ConversionsExtensionsCatalog
...
-        [BestFriend]
-        internal static HashingEstimator Hash(this TransformsCatalog.ConversionTransforms catalog, params HashingEstimator.ColumnOptions[] columns)
-            => new HashingEstimator(CatalogUtils.GetEnvironment(catalog), columns);
+        ///  <example>
+        /// <format type="text/markdown">
+        /// <![CDATA[
+        ///  [!code-csharp[Hash](~/../docs/samples/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/Conversion/HashWithOptions.cs)]
+        /// ]]></format>
+        /// </example>
+        public static HashingEstimator Hash(this TransformsCatalog.ConversionTransforms catalog, params HashingEstimator.Options[] options)
```
yaeldekel: >options [](start = 131, length = 7)

Wouldn't it be simpler to add the missing options `Seed` and `UseOrderedHashing` as parameters to the API above?
From looking over the transforms, I didn't see any that expose a ColumnOptions/Options class (I may have missed some though). For estimators that are trainable sometimes we expose an API for defining multiple columns using an array of `InputOutputColumnPair`, but the other options are usually common for all the columns.
Was there a specific ask for defining multiple columns with different options in a single estimator? (the hashing estimator is trainable in case of `maximumNumberOfInverts > 0`, so there is an advantage to exposing an API for multiple columns, I'm just not sure whether we need to support defining different options for each column).

codemzs: We follow the pattern as for each component there is a convenience API and an advanced API that takes in all the options.

codemzs: @eerhardt what are your thoughts?

eerhardt: See the discussion on https://github.com/dotnet/machinelearning/issues/1798.

eerhardt: Also see the discussion on https://github.com/dotnet/machinelearning/issues/2884. Renaming `ColumnOptions` to `Options` and leaving it as-is may not be the best choice here. See Artidoro, Senja, and Tom's comments.

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -1132,11 +1133,69 @@ internal static class Defaults
...
+            /// <summary>
+            /// During hashing we constuct mappings between original values and the produced hash values.
+            /// Text representation of original values are stored in the slot names of the annotations for the new column.Hashing, as such, can map many initial values to one.
+            /// <see cref="MaximumNumberOfInverts"/> specifies the upper bound of the number of distinct input values mapping to a hash that should be retained.
+            /// <value>0</value> does not retain any input values. <value>-1</value> retains all input values mapping to each hash.
+            /// </summary>
+            public int MaximumNumberOfInverts { get; set; }
+
+            /// <summary>
+            /// Describes how the transformer handles one column pair.
```
eerhardt: (nit) this should say "Initializes a new instance of the ColumnOptions class"

eerhardt:
Do you want to add a new functional test that uses the new advanced options? That way we ensure the actual API works as users will use it.

codemzs:
@eerhardt I have added the functional test you asked for (even though I had a unit-test called HashWorkout that exercised this routine). Since the API looks good to you so I will check-in this PR (because I need to provide the nuget today) but if you have more comments please make here and I will include them in a separate PR.

PR #4444, opened 2019-11-06
nightly build pipeline

frank-dong-ms:
new nightly build pipeline:

1. add new nightly build pipeline project, disable project build from solution
2. add NuGet package version updater project
3. add new Azure nightly build pipeline and template file
4. TestFrameworkCommon project use conditional reference to source code:
    a. when reference from functional test, use project reference
    b. when reference from nightly build test, use package reference
5. process of nightly build pipeline:
    a. get latest NuGet package version from public NuGet feed
    b. update version to .props file
    c. build nightly build project
    d. run nightly build tests, which is functional test for now
    e. output test results
6. a sample test pipeline can be seems here: https://dev.azure.com/dnceng/public/_build?definitionId=644&_a=summary


build/nightly-build/job-template.yml
```diff
@@ -0,0 +1,116 @@
...
+        targetFolder: $(Build.ArtifactStagingDirectory)
+    - task: PublishBuildArtifacts@1
+      displayName: Publish build and test logs
+      condition: not(succeeded())
+      inputs:
+        pathToPublish: $(Build.ArtifactStagingDirectory)
+        artifactName: parameters.name(_config_short)
+        artifactType: container
+    - ${{ if eq(parameters.buildScript, 'build.cmd') }}:
+      - powershell: |
```
safern: This step to clean-up test data is not needed if you're not running any steps after running tests... the reason why we have it on CI builds is because we build packages after running tests and we need to cleanup space disk before we can do that to avoid failing because the machine ran out of space.

build/nightly-build/job-template.yml
```diff
@@ -0,0 +1,116 @@
+parameters:
+  name: ''
```
safern: This file is very similar to the `job-template.yml` that we already have for CI builds. Would it make sense to instead use that template, and then define variables and steps based on a parameter passed down to the template? I.e: add a parameter called... isNightlyBuild: false (defaults to false)... then when you reference the template from the nightly build you pass that parameter down as `true`. Then in the template when `isNightlyBuild == true` you run the build and restore steps differently. That way you only maintain one base template file.

test/NightlyBuildDependency.props
```diff
@@ -0,0 +1,14 @@
...
+    <PackageReference Include="Microsoft.ML.EntryPoints" Version="0.17.0-preview-28305-3" />
+    <PackageReference Include="Microsoft.ML.FastTree" Version="1.5.0-preview-28305-3" />
+    <PackageReference Include="Microsoft.ML.LightGbm" Version="1.5.0-preview-28305-3" />
+    <PackageReference Include="Microsoft.ML.Mkl.Components" Version="1.5.0-preview-28305-3" />
+    <PackageReference Include="Microsoft.ML.OnnxConverter" Version="0.17.0-preview-28305-3" />
+    <PackageReference Include="Microsoft.ML.OnnxTransformer" Version="1.5.0-preview-28305-3" />
+    <PackageReference Include="Microsoft.ML.Parquet" Version="0.17.0-preview-28305-3" />
+    <PackageReference Include="Microsoft.ML.Recommender" Version="0.17.0-preview-28305-3" />
+  </ItemGroup>
+</Project>
```
safern: Nit: extra line in EOF in all these files.

test/NightlyBuildDependency.props
```diff
@@ -0,0 +1,14 @@
+<Project>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ML" Version="1.5.0-preview-28305-3" />
```
safern: What's the plan for updating these versions? I see only 2 package versions here, should we define a property for them and then use those properties here instead of hard-coding the versions?

test/TestFrameworkDependency.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <ItemGroup Condition="'$(ReferenceTypeForTestFramework)'=='' OR '$(ReferenceTypeForTestFramework)'=='Project'">
+    <ProjectReference Include="..\..\src\Microsoft.ML.Data\Microsoft.ML.Data.csproj" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(ReferenceTypeForTestFramework)'=='NuGet'">
+    <PackageReference Include="Microsoft.ML" Version="1.5.0-preview-28305-3" />
```
safern: You could use the shared property for the package version here as I suggested on my previous comment.

test/TestFrameworkDependency.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <ItemGroup Condition="'$(ReferenceTypeForTestFramework)'=='' OR '$(ReferenceTypeForTestFramework)'=='Project'">
+    <ProjectReference Include="..\..\src\Microsoft.ML.Data\Microsoft.ML.Data.csproj" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(ReferenceTypeForTestFramework)'=='NuGet'">
```
safern: Nit: space before and after `==`

test/TestFrameworkDependency.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <ItemGroup Condition="'$(ReferenceTypeForTestFramework)'=='' OR '$(ReferenceTypeForTestFramework)'=='Project'">
```
safern: Nit: space before and after `==`

build/nightly-build/job-template.yml
frank-dong-ms: I see, thanks for explain, will remove

---
In reply to: [343775233](https://github.com/dotnet/machinelearning/pull/4444#discussion_r343775233) [](ancestors = 343775233)

test/NightlyBuildDependency.props
frank-dong-ms: These versions are only place holder, when running pipeline, the NugetPackageVersionUpdater will update all these out of date version number

---
In reply to: [343777962](https://github.com/dotnet/machinelearning/pull/4444#discussion_r343777962) [](ancestors = 343777962)

test/TestFrameworkDependency.props
frank-dong-ms: same as pervious

---
In reply to: [343778600](https://github.com/dotnet/machinelearning/pull/4444#discussion_r343778600) [](ancestors = 343778600)

build/nightly-build/job-template.yml
frank-dong-ms: Yes, I saw what was done for codeCoverage, for nightly build, the logic difference is more as you can see there are 5 steps that is nightly build only, and there are several steps that nightly build don't need. So I use separate template here to keep the logic simple but I don't have strong preference here, let's discuss offline.

---
In reply to: [343777088](https://github.com/dotnet/machinelearning/pull/4444#discussion_r343777088) [](ancestors = 343777088)

test/NightlyBuildDependency.props
safern: Would it make sense to instead use a placeholder? Rather than a hard-coded version? I think it deserves a comment explaining what happens so that people understand that we update it on the fly.

build/nightly-build/job-template.yml
```diff
@@ -0,0 +1,116 @@
...
+      container: ${{ parameters.container }}
+
+    steps:
+    - ${{ if eq(parameters.pool.name, 'Hosted macOS High Sierra') }}:
+      - script: brew update && brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/f5b1ac99a7fba27c19cee0bc4f036775c889b359/Formula/libomp.rb && brew install mono-libgdiplus gettext && brew link gettext --force && brew link libomp --force
+        displayName: Install build dependencies
+    - ${{ if eq(parameters.name, 'Ubuntu_x64_NetCoreApp21') }}:
+      - bash: echo "##vso[task.setvariable variable=LD_LIBRARY_PATH](nightlyBuildRunPath):LD_LIBRARY_PATH"
+        displayName: Set LD_LIBRARY_PATH for Ubuntu to locate Native shared library
+    - script: parameters.buildScript−(_configuration) -buildArch=${{ parameters.architecture }}
```
safern: Please prepend variables with `$` to be consistent... `(configuration)` -> `$(configuration)`

Also in the rest of this file.

frank-dong-ms: Yes, will do

---
In reply to: [343901037](https://github.com/dotnet/machinelearning/pull/4444#discussion_r343901037) [](ancestors = 343901037)

test/NightlyBuildDependency.props
frank-dong-ms: Yes, I think this is good suggestion, I will sue placeholder to reduce confusion. 

---
In reply to: [343900725](https://github.com/dotnet/machinelearning/pull/4444#discussion_r343900725) [](ancestors = 343900725)

build/nightly-build/job-template.yml
frank-dong-ms: Offline discussed with Santiago, will merge these template file into

---
In reply to: [343828680](https://github.com/dotnet/machinelearning/pull/4444#discussion_r343828680) [](ancestors = 343828680,343777088)

build/ci/job-template.yml
```diff
@@ -35,12 +45,32 @@ jobs:
...
+      condition: eq(${{ parameters.nightlyBuild }}, true)
+    - script: ${{ parameters.buildScript }} -$(_configuration) -runnightlybuildtests
+      displayName: Run Nightly Build Tests
+      condition: and(succeeded(), eq(${{ parameters.nightlyBuild }}, true))
     - script: ${{ parameters.buildScript }} -- /t:DownloadExternalTestFiles /p:IncludeBenchmarkData=$(_includeBenchmarkData)
       displayName: Download Benchmark Data
+      condition: eq(${{ parameters.nightlyBuild }}, false)
     - script: ${{ parameters.buildScript }} -$(_configuration) -runtests -coverage=${{ parameters.codeCoverage }}
       displayName: Run Tests.
+      condition: eq(${{ parameters.nightlyBuild }}, false)
```
safern: Instead of using runtime conditions, could you use compile time conditions? That way when you're running on nightly build, the UI will only show the "real" steps and when in CI, the "real" CI steps. The way currently is, it will show a bunch of disabled steps, which makes harder to read the logs.

Compile time would be `${{ if eq(parameters.nightlyBuild, 'false') }}`

frank-dong-ms: I see, will do, good suggestions

---
In reply to: [344278680](https://github.com/dotnet/machinelearning/pull/4444#discussion_r344278680) [](ancestors = 344278680)

build/ci/job-template.yml
```diff
@@ -35,12 +45,32 @@ jobs:
...
+        displayName: List latest package versions
+    - ${{ if eq(parameters.nightlyBuild, 'true') }}:
+      - script: $(dotnetPath) run --project $(packageUpdaterProjPath)
+        displayName: Update package version
+    - ${{ if eq(parameters.nightlyBuild, 'true') }}:
+      - script: $(dotnetPath) msbuild -restore $(nightlyBuildProjPath) /p:ReferenceTypeForTestFramework="Nuget" /p:Configuration=$(_configuration) /p:TargetArchitecture=${{ parameters.architecture }}
+        displayName: Build Nightly Build Project with latest package versions
+    - ${{ if eq(parameters.nightlyBuild, 'true') }}:
+      - script: ${{ parameters.buildScript }} -$(_configuration) -runnightlybuildtests
+        displayName: Run Nightly Build Tests
```
safern: You should only use 1 if for all these lines. An if can have multiple tasks/blocks under it, it is just a matter of indentation. 

Same for the `false` block.

frank-dong-ms: I see, will do, thanks

---
In reply to: [344337822](https://github.com/dotnet/machinelearning/pull/4444#discussion_r344337822) [](ancestors = 344337822)

PR #4446, opened 2019-11-06
Stabilize the LR test

bpstark:
Found issue with how we were using random for our
ImageClassificationTrainer. This caused instability in our unit test, as
we were not able to control the random seed. Modified the code to now
use the same random object throughout, the trainer, thus allowing us to
control the seed and therefor have predictable output.

### DIFF
--- a/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
+++ b/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
@@ -937,7 +937,7 @@ namespace Microsoft.ML.Vision
                 metrics.Train.LearningRate = learningRate;
                 // Update train state.
                 trainstate.CurrentEpoch = epoch;
-                using (var cursor = trainingSet.GetRowCursor(trainingSet.Schema.ToArray(), new Random()))
+                using (var cursor = trainingSet.GetRowCursor(trainingSet.Schema.ToArray(), Host.Rand))
                 {
                     var labelGetter = cursor.GetGetter<long>(trainingSet.Schema[0]);
                     var featuresGetter = cursor.GetGetter<VBuffer<float>>(featureColumn);
@@ -1069,7 +1069,7 @@ namespace Microsoft.ML.Vision
                 metrics.Train.BatchProcessedCount = 0;
                 metrics.Train.Accuracy = 0;
                 metrics.Train.CrossEntropy = 0;
-                using (var cursor = validationSet.GetRowCursor(validationSet.Schema.ToArray(), new Random()))
+                using (var cursor = validationSet.GetRowCursor(validationSet.Schema.ToArray(), Host.Rand))
                 {
                     var labelGetter = cursor.GetGetter<long>(validationSet.Schema[0]);
                     var featuresGetter = cursor.GetGetter<VBuffer<float>>(featureColumn);
--- a/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
+++ b/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
@@ -937,7 +937,7 @@ namespace Microsoft.ML.Vision
                 metrics.Train.LearningRate = learningRate;
                 // Update train state.
                 trainstate.CurrentEpoch = epoch;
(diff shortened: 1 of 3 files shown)


codemzs:
@bpstark The polynomial LR test still seems to be failing, see below snapshot from [test results](https://dev.azure.com/dnceng/public/_build/results?buildId=418264&view=ms.vss-test-web.build-test-results-tab).

image

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -937,7 +937,7 @@ private int GetNumSamples(string path)
                 metrics.Train.LearningRate = learningRate;
                 // Update train state.
                 trainstate.CurrentEpoch = epoch;
-                using (var cursor = trainingSet.GetRowCursor(trainingSet.Schema.ToArray(), new Random()))
+                using (var cursor = trainingSet.GetRowCursor(trainingSet.Schema.ToArray(), Host.Rand))
```
codemzs: >trainingSet.GetRowCursor(trainingSet.Schema.ToArray(), Host.Rand)) [](start = 36, length = 66)

You don't need to pass anything, look at the code for ShuffleRowTransformer:

            if (_forceShuffle || _forceShuffleSource)
                _forceShuffleSeed = options.ForceShuffleSeed ?? Host.Rand.NextSigned();

bpstark: While you are correct that if we do not pass anything then it should still be stable, RowShuffling uses host to create a seed to then create a new random, this seems unnecessary, as we would be creating a new object when we can just use an existing random which is more efficient.

---
In reply to: [344297118](https://github.com/dotnet/machinelearning/pull/4446#discussion_r344297118) [](ancestors = 344297118)

codemzs: I discussed offline with @bpstark and we agreed that we should not be passing in Host.Rand object (as I suggested earlier) because it is not thread safe. As I explained, the shuffling transformer code will create a seed from Host.Rand object (that it has access internally) to create an local/internal Rand object and this will ensure stability and also remove concerns around thread safety.

codemzs:
@bpstark Thanks! Can you please confirm samples run without any change in accuracy? Otherwise the change looks great.

bpstark:
Actually accuracy did change, it improved, it just happens that this random number is slightly better than the previously used random number, and since the dataset is so small it made an actual impact.

---
In reply to: [551956033](https://github.com/dotnet/machinelearning/pull/4446#issuecomment-551956033) [](ancestors = 551956033)

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1429,16 +1429,23 @@ public void TensorFlowImageClassification(ImageClassificationTrainer.Architectur
...
-        [Fact(Skip ="Very unstable tests, causing many build failures.")]
+        [TensorFlowFact]
         public void TensorFlowImageClassificationWithPolynomialLRScheduling()
         {
-            TensorFlowImageClassificationWithLRScheduling(new PolynomialLRDecay());
+
+            /*
+             * Due to an issue with Nix based os performance is not as good,
+             * as such increase the number of epochs to produce a better model.
+             */
```
codemzs: I think this is because of some images that seem to have different pixel format on nix than windows, can we please try changing the dataset by subsampling another dataset like cats vs dogs?

bpstark: Discussed offline will look into this and potentially replace the dataset in a separate PR, or potentially replace just the bad images if possible.

---
In reply to: [344365787](https://github.com/dotnet/machinelearning/pull/4446#discussion_r344365787) [](ancestors = 344365787)

codemzs: thanks

PR #4448, opened 2019-11-06
Fixes onnx exports for regression trainers

Lynx1820:
This PR fixes the issues with exporting the following regression trainers to onnx:
- LbfgsPoissonRegression
- FastTreeTweedie
- FastForest

And adds onnx conversion tests for the following regression trainers: 
- FastTree
- OnlineGradientDescent
- Ols
- LightGbm
- Sdca


src/Microsoft.ML.FastTree/FastTree.cs
```diff
@@ -3132,6 +3131,10 @@ bool ISingleCanSaveOnnx.SaveAsOnnx(OnnxContext ctx, string[] outputNames, string
 
             return true;
         }
```
codemzs: > } [](start = 7, length = 2)

newline

src/Microsoft.ML.StandardTrainers/Standard/LinearModelParameters.cs
```diff
@@ -132,6 +132,19 @@ internal LinearModelParameters(IHostEnvironment env, string name, in VBuffer<flo
             else
                 _weightsDenseLock = new object();
         }
```
codemzs: >} [](start = 8, length = 1)

newline

src/Microsoft.ML.FastTree/RandomForestRegression.cs
```diff
@@ -209,6 +211,17 @@ private static FastForestRegressionModelParameters Create(IHostEnvironment env,
 
         private protected override PredictionKind PredictionKind => PredictionKind.Regression;
 
+        bool ISingleCanSaveOnnx.SaveAsOnnx(OnnxContext ctx, string[] outputNames, string featureColumn)
+        {
+            // Mapping score to prediction
+            var fastTreeOutput = ctx.AddIntermediateVariable(null, "FastTreeOutput", true);
+            var numTrees = ctx.AddInitializer((float)TrainedEnsemble.NumTrees, "NumTrees");
+            base.SaveAsOnnx(ctx, new[] { fastTreeOutput }, featureColumn);
```
harishsk: NumTrees is an integer variable. And AddInitializer supports long as a parameter. Can you please explain why it is necessary to cast it to float?

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -187,6 +187,59 @@ public void KmeansOnnxConversionTest()
             Done();
         }
 
+        [LightGBMFact]
```
harishsk: If we add LightGBMFact to the whole test, this test won't run on x86 builds at all even for the trainers that are supported on x86. A better approach would be to do runtime checking like IsOnnxRuntimeSupported. 
For an example, see MulticlassTrainersOnnxConversionTest() in my pull request for multiclass classifiers. [https://github.com/dotnet/machinelearning/pull/4462](https://github.com/dotnet/machinelearning/pull/4462)

src/Microsoft.ML.FastTree/RandomForestRegression.cs
Lynx1820: The onnx "Div" operation requires that both operands are of the same type T, and FastTreeOutput has to be a float.

test/Microsoft.ML.Tests/OnnxConversionTest.cs
Lynx1820: Thanks! Made the change.

PR #4454, opened 2019-11-07
Added onnx export support for OptionalColumnTransform 

harishsk:
Optional columns in ML.NET are supported through initializers in Onnx runtime, support for which was added in the v1 release of ORT. This PR adds support for integrating initializers in ML.NET and adding the corresponding OptionalColumnTransform support in ML.NET.

### DIFF
--- a/src/Microsoft.ML.Transforms/OptionalColumnTransform.cs
+++ b/src/Microsoft.ML.Transforms/OptionalColumnTransform.cs
@@ -536,7 +536,8 @@ namespace Microsoft.ML.Transforms
 
             // REVIEW:
             // AddInitializer only supports long, float and string.
-            // Is it correct to cast double to float and ulong to long?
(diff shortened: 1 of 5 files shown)


src/Microsoft.ML.Transforms/OptionalColumnTransform.cs
```diff
@@ -498,6 +499,57 @@ private Delegate MakeGetterVec<T>(int length)
...
+
+            int size;
+            if (columnType is VectorDataViewType && columnType.IsKnownSizeVector())
+                size = columnType.GetVectorSize();
+            else
+                size = 1;
+
+            // REVIEW:
+            // AddInitializer only supports long, float and string.
+            // Is it correct to cast double to float and ulong to long?
```
ganik: > // Is it correct to cast double to float and ulong to long [](start = 11, length = 59)

State this as a known issue for future

PR #4456, opened 2019-11-07
Reduce time taken for first prediction, Fixes #4428

### LINKED ISSUE #4428: [Image Classification] Very long time to warm-up when doing the first prediction
I'd like to know if we can do anything to improve the first prediction's needed time when using the new Image Classification model based on DNN (TensorFlow).

This behavior/times happen when using the default DNN architecture which is **ResnetV250**.

When using the CPU, the first prediction takes something in between 7 to 12 seconds depending on the model and environment.
Then, upcoming predictions using the same PredictionEngine only need around 200 mlSecs if using CPU.

image

When using a GPU the difference is even larger. Around 15 secs for the first prediction, then a lot less for the next predictions (in this case, something in between 40 mlsecs and 100 mlSecs)

image

Basically, after the first prediction, it behaves good, with CPU and even better with GPU, but the first prediction needs a huge amount of time to probably initialize internally?

Could that initialization be improved or happen before calling .Predict()?
I'd like to know if we can do anything to improve the behavior/perf of the first prediction like initializing in advanced when creating the prediction engine instead of when predicting the first time?

@codemzs - Thoughts?

ashbhandare:
Fixes #4428 
As reported in the issue, the first prediction takes a lot longer than the subsequent predictions.
This is due to the tensorflow initialization/graph optimizations happenning on at the time of the first prediction. 
In order to mask the time within the PredictionEngine creation, this change runs the prediction 
 on the tensorflow graph with a dummy tensor of the same size as an Imagenet image, of zeros.

### DIFF
--- a/src/Microsoft.Extensions.ML/Microsoft.Extensions.ML.csproj
+++ b/src/Microsoft.Extensions.ML/Microsoft.Extensions.ML.csproj
@@ -3,7 +3,6 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <IncludeInPackage>Microsoft.Extensions.ML</IncludeInPackage>
-    <RunApiCompat>false</RunApiCompat>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Microsoft.ML.Vision/Microsoft.ML.Vision.csproj
+++ b/src/Microsoft.ML.Vision/Microsoft.ML.Vision.csproj
@@ -5,7 +5,6 @@
     <IncludeInPackage>Microsoft.ML.Vision</IncludeInPackage>
     <DefineConstants>CORECLR</DefineConstants>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <RunApiCompat>false</RunApiCompat>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
--- a/tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
+++ b/tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj
@@ -6,16 +6,18 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.ML" Version="1.2.0" />
-    <PackageReference Include="Microsoft.ML.DataView" Version="1.2.0" />
-    <PackageReference Include="Microsoft.ML.CpuMath" Version="1.2.0" />
-    <PackageReference Include="Microsoft.ML.FastTree" Version="1.2.0" />
-    <PackageReference Include="Microsoft.ML.LightGbm" Version="1.2.0" />
-    <PackageReference Include="Microsoft.ML.ImageAnalytics" Version="1.2.0" />
-    <PackageReference Include="Microsoft.ML.Mkl.Components" Version="1.2.0" />
-    <PackageReference Include="Microsoft.ML.TimeSeries" Version="1.2.0" />
-    <PackageReference Include="Microsoft.ML.TensorFlow" Version="1.2.0" />
-    <PackageReference Include="Microsoft.ML.OnnxTransformer" Version="1.2.0" />
+    <PackageReference Include="Microsoft.Extensions.ML" Version="1.4.0" />
+    <PackageReference Include="Microsoft.ML" Version="1.4.0" />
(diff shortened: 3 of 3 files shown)


src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -1496,6 +1497,28 @@ public Classifier(ImageClassificationModelParameters model)
...
+                {
+                    //Create a fake input and sess.run to mask the tensorflow graph initialization/
+                    // optimization time within Classifier creation, so that the first prediction using
+                    // Classifier.Score takes the same amount of time as the subsequent predictions.
+                    float[] fakeBatch = new float[224 * 224 * 3];
+                    long[] fakeTensorShape = { 1, 224, 224, 3 };
+                    var fakeBatchHandle = GCHandle.Alloc(fakeBatch, GCHandleType.Pinned);
+                    IntPtr fakeBatchPtr = fakeBatchHandle.AddrOfPinnedObject();
+                    int fakeBatchSizeInBytes = sizeof(float) * fakeBatch.Length;
+                    var fakeOutput = _runner.AddInput(new Tensor(fakeBatchPtr, fakeTensorShape, TF_DataType.TF_FLOAT, fakeBatchSizeInBytes), 0).Run();
```
eerhardt: Are we sure doing this is a good idea overall?

I can think of the following places where this is actually worse for a user:

1. The total time take is now longer because we are making this "fake" prediction.
2. The user creates multiple PredictionEngines in their application - we will be issuing "fake" requests even for the 2nd, 3rd, etc.
3. The case where they create a PredictionEngine, but never make a prediction. (not very common, but still calling it out)

If some customers want this, they can "warm up" TensorFlow themselves. We shouldn't make every user pay.

I think the intention of the issue was to profile and investigate if there is something unnecessary happening, or if we could do something another way to speed it up. I'm not sure "masking" the time taken is a good solution.

codemzs: I agree with you @eerhardt. I haze been looking if TF has ways to warm things up but the one thing to try is to save graph in SaveModel format and see if that fixes this issue. Thanks for the investigation @ashbhandare but like @eerhardt says the point was to root cause the delay which you did but let’s try to not put hacks in.

ashbhandare: I agree that if there is a better way to do this, that should go in instead, but the issue specifically asked for masking this time as an alternative, which is why this PR exists. 
The fastest way of inferring in tensorflow from what I found is to freeze the model, apply certain graph optimizations like constant folding, remove unused nodes from the graph. This can be done by using the graph transform tools ([https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/graph_transforms](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/graph_transforms)). I did not find a straightforward way of using these in ML .Net when we save our trained model.  

---
In reply to: [344231882](https://github.com/dotnet/machinelearning/pull/4456#discussion_r344231882) [](ancestors = 344231882)

codemzs: I would try with SaveModel format first before spending too much cycles on anything too fancy.

ashbhandare: SavedModel format is not supported by TF .Net. The one way that is supported is tf.train.saver.save(), which saves the checkpoint. Since we are freezing our trained model already(in this process, converting variable nodes to constants, and extracting the inference graph), I doubt that the saved model format would give much benefit(since the primary intution of using the saved_model format was that one can directly load the inference meta-graph https://www.tensorflow.org/guide/saved_model .)

Oceania2018: Run the sessions in a long-running process is a better for end-user. 
But there is another solution provided by TFX: https://www.tensorflow.org/tfx/serving/saved_model_warmup @codemzs @eerhardt

codemzs: @oceania2018 Thanks! @ashbhandare this is what I was talking about and yes, SaveModel is supported in TF .NET code, we have tests that exercise that code path, please check retrain tests.

ashbhandare: @codemzs I tried and failed to find a way of saving using the tensorflow-like api : https://www.tensorflow.org/api_docs/python/tf/saved_model/save, which produces a SavedModel with the following directory structure: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/saved_model/README.md#components. I might be missing something.

PR #4458, opened 2019-11-07
Fix a flaky Extensions.ML test.

eerhardt:
Make the reload model tests more resistant to timing changes.

### DIFF
--- a/test/Microsoft.Extensions.ML.Tests/FileLoaderTests.cs
+++ b/test/Microsoft.Extensions.ML.Tests/FileLoaderTests.cs
@@ -60,7 +60,7 @@ namespace Microsoft.Extensions.ML
             loaderUnderTest.Start("testdata.txt", true);
 
             using AutoResetEvent changed = new AutoResetEvent(false);
-            var changeTokenRegistration = ChangeToken.OnChange(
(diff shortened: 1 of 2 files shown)


eerhardt:
@codemzs - thoughts on this? This test doesn't fail super often, but I've seen it fail enough to fix it here.

test/Microsoft.Extensions.ML.Tests/FileLoaderTests.cs
```diff
@@ -61,16 +59,14 @@ public async Task can_reload_model()
             var loaderUnderTest = ActivatorUtilities.CreateInstance<FileLoaderMock>(sp);
             loaderUnderTest.Start("testdata.txt", true);
 
-            var changed = false;
+            using AutoResetEvent changed = new AutoResetEvent(false);
             var changeTokenRegistration = ChangeToken.OnChange(
```
codemzs: >changeTokenRegistration [](start = 16, length = 23)

where is this being used?

eerhardt: `changeTokenRegistration` is an `IDisposable`, so we should be disposing it. Fixing.

PR #4462, opened 2019-11-08
Added onnx export support for several multiclass classifiers

harishsk:
This PR adds Onnx export support + tests for the following multiclass classifiers:
* LbfgsMaximumEntropyMulticlassTrainer
* LightGbmMulticlassTrainer
* LightGbmMulticlassTrainer with SoftMax
* OneVersusAllTrainer
* SdcaMaximumEntropyMulticlassTrainer
* SdcaNonCalibratedMulticlassTrainer

There is a question marked with a REVIEW comment in the PR below. Please comment if you know of a better way to do what is done there.

### DIFF
--- a/src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/OneVersusAllTrainer.cs
+++ b/src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/OneVersusAllTrainer.cs
@@ -16,6 +16,7 @@ using Microsoft.ML.EntryPoints;
 using Microsoft.ML.Internal.Internallearn;
 using Microsoft.ML.Internal.Utilities;
 using Microsoft.ML.Model;
+using Microsoft.ML.Model.OnnxConverter;
 using Microsoft.ML.Model.Pfa;
 using Microsoft.ML.Runtime;
 using Microsoft.ML.Trainers;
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/OneVersusAllTrainer.cs
```diff
@@ -499,6 +505,10 @@ private abstract class ImplBase : ISingleCanSavePfa
             public abstract ValueMapper<VBuffer<float>, VBuffer<float>> GetMapper();
             public abstract JToken SaveAsPfa(BoundPfaContext ctx, JToken input);
 
+            public bool CanSaveOnnx(OnnxContext ctx) => Predictors.All(pred => (pred as ICanSaveOnnx)?.CanSaveOnnx(ctx) == true);
```
ganik: >(pred as ICanSaveOnnx)? [](start = 79, length = 23)

What happens if it cant cast?

harishsk: If it can't cast, it will return a null. And the question mark after the ')' ensures makes it a null conditional operator and will not execute further and the overall function will return false.

---
In reply to: [348211703](https://github.com/dotnet/machinelearning/pull/4462#discussion_r348211703) [](ancestors = 348211703)

PR #4472, opened 2019-11-12
Fixed model saving and loading of OneVersusAllTrainer to include SoftMax

### LINKED ISSUE #3647: LightGBM Save/Load round trip loses Softmax OneVersusAllModelParameters
version: 0.11

Related: #1424

When training a softmax Multi classifier with LightGBM a save/load will lose the softmax it seems:

Also inspecting the model we can see it doesn't use `ImplSoftmax` but `ImplRaw`.

Reproduce:

```C#
    public class GenericSample
    {
      public string A { get; set; }
    }
    public static void ReproduceLightGbmPersistanceBug()
    {
      var data = Enumerable.Range(1, 100).Select(x => new GenericSample { A = $"{x % 20}", Label = $"{x % 10}" });
      var ctx = new MLContext();

      var options = new Options {
        UseSoftmax = true,
      };
      var pipe = ctx.Transforms.Categorical.OneHotEncoding("A")
        .Append(ctx.Transforms.Concatenate("Features", "A"))
        .Append(ctx.Transforms.Conversion.MapValueToKey("Label"))
        .Append(ctx.MulticlassClassification.Trainers.LightGbm(options));
      var dataView = ctx.Data.LoadFromEnumerable(data);
      ITransformer model = pipe.Fit(dataView);
      var scores = model.Transform(dataView).GetColumn<float[]>(ctx,"Score");

      Console.WriteLine($"Min: {scores.Select(x => x.Min()).Min()}");
      Console.WriteLine($"Max: {scores.Select(x => x.Max()).Max()}");

      var memoryStream = new MemoryStream();
      ctx.Model.Save(model, memoryStream);
      model = ctx.Model.Load(memoryStream);

      scores = model.Transform(dataView).GetColumn<float[]>(ctx,"Score");
      Console.WriteLine($"Min: {scores.Select(x => x.Min()).Min()}");
      Console.WriteLine($"Max: {scores.Select(x => x.Max()).Max()}");
    }
```

Output:

```
Min: 0.001027671
Max: 0.9907509
Min: -4.843706
Max: 2.027462
```

### LINKED ISSUE #4051: LightGBM is producing different multiclass scores after loading saved model
### System information

Windows 10
Microsoft.ML (1.2.0)
Microsoft.ML.LightGbm (1.2.0)
.NET Core 2.2

### Issue

After training a lightgbm model, the model is producing multiclass scores between [0, 1] which totals 1, as expected. 

However, after saving the model, then loading it into a new trainedModel object - the scores are now not probabilities, but decimal values. 

I have tested the saving and loading with other model types and I cannot replicate the results. It is only the case with the lightgbm model.

Please advise. I am now attempting to rollback library versions to see if it's still an issue

### Source code / logs

**Before saving model...**
0.003305528
0.01293249
0.01907223
0.9646355
5.421485E-05
3.556848E-08

**After saving model...**
-3.623514
-2.259367
-1.870877
2.05264
-7.733911
-15.06316

image

Source code

mlContext.Model.Save(trainedModel, dataView.Schema, _modelPath);

            // Save Data Prep transformer
            //mlContext.Model.Save(pipeline, dataView.Schema, "data_preparation_pipeline.zip");

            schema = dataView.Schema;

            Console.WriteLine("Before saving model...");
            TestModelOutput(mlContext, trainedModel);

            // Load trained model
            trainedModel = mlContext.Model.Load(_modelPath, out schema);
            //trainedModel = mlContext.Model.LoadWithDataLoader()

            Console.WriteLine("After saving model...");
            TestModelOutput(mlContext, trainedModel);

 private static void TestModelOutput(MLContext mlContext, ITransformer model)
        {
IDataView batchData = mlContext.Data.LoadFromEnumerable(testActions);

            IDataView predictions = model.Transform(batchData);

            IEnumerable<PredictionData> predictedResults = mlContext.Data
                .CreateEnumerable<PredictionData>(predictions, reuseRowObject: false);

            foreach (var item in predictedResults)
            {
                foreach (var score in item.Score)
                {
                    Console.WriteLine(score);
                }
                
            }
}

### LINKED ISSUE #4450: LightGBMMulticlass doesn't normalize its Score column
when I try AutoML v0.16.0 multi-classification, sometimes I get prediction score <0, and the summation of all scores doesn't equal to 1. which is unexpected. Maybe some trainer's output doesn't normalize.

image

I publish a mini-reproducible project in this [repo](https://github.com/LittleLittleCloud/AutoMLMultiClassificationBug) for facilitating debug.

harishsk:
When the model parameters from OnVersusAllTrainer are persisted, it currently writes a single bool byte to indicate what type of output formula to use during prediction. If that byte is false, it outputs raw values else it outputs probability values normalized to one. 
The sample code in this bug uses AutoML to request a softmax output. The code currently only checks whether the output type is ImplDist which excludes softmax output.

This PR changes the saved model format in a backward compatible way by continuing to write a single byte for output type and continues to use 0 for raw output, 1 for probability normalization but also adds 2 to indicate SoftMax. These values are already in alignment with the OutputFormula enum.

Fixes #4450 
Fixes #3647
Fixes #4051

### DIFF
--- a/src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/OneVersusAllTrainer.cs
+++ b/src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/OneVersusAllTrainer.cs
@@ -364,23 +364,29 @@ namespace Microsoft.ML.Trainers
                 : base(env, RegistrationName, ctx)
         {
             // *** Binary format ***
-            // bool: useDist
+            // byte: OutputFormula as byte
             // int: predictor count
-            bool useDist = ctx.Reader.ReadBoolByte();
+            OutputFormula outputFormula = (OutputFormula)ctx.Reader.ReadByte();
             int len = ctx.Reader.ReadInt32();
             Host.CheckDecode(len > 0);
 
-            if (useDist)
+            if (outputFormula == OutputFormula.Raw)
+            {
+                var predictors = new TScalarPredictor[len];
+                LoadPredictors(Host, predictors, ctx);
+                _impl = new ImplRaw(predictors);
+            }
+            else if (outputFormula == OutputFormula.ProbabilityNormalization)
             {
                 var predictors = new IValueMapperDist[len];
                 LoadPredictors(Host, predictors, ctx);
                 _impl = new ImplDist(predictors);
             }
-            else
+            else if (outputFormula == OutputFormula.Softmax)
             {
                 var predictors = new TScalarPredictor[len];
                 LoadPredictors(Host, predictors, ctx);
-                _impl = new ImplRaw(predictors);
+                _impl = new ImplSoftmax(predictors);
             }
 
             DistType = new VectorDataViewType(NumberDataViewType.Single, _impl.Predictors.Length);
@@ -409,9 +415,10 @@ namespace Microsoft.ML.Trainers
             var preds = _impl.Predictors;
 
             // *** Binary format ***
-            // bool: useDist
+            // byte: _impl.OutputFormula as byte
             // int: predictor count
-            ctx.Writer.WriteBoolByte(_impl is ImplDist);
+            byte[] outputFormula = { (byte)_impl.OutputFormula };
+            ctx.Writer.WriteBytesNoCount(outputFormula, 1);
             ctx.Writer.Write(preds.Length);
 
             // Save other streams.
@@ -485,6 +492,7 @@ namespace Microsoft.ML.Trainers
 
         private abstract class ImplBase : ISingleCanSavePfa
         {
+            public OutputFormula OutputFormula;
             public abstract DataViewType InputType { get; }
             public abstract IValueMapper[] Predictors { get; }
             public abstract bool CanSavePfa { get; }
@@ -536,6 +544,7 @@ namespace Microsoft.ML.Trainers
                 CanSavePfa = Predictors.All(m => (m as ISingleCanSavePfa)?.CanSavePfa == true);
                 Contracts.AssertValue(inputType);
                 InputType = inputType;
+                OutputFormula = OutputFormula.Raw;
             }
 
             public override ValueMapper<VBuffer<float>, VBuffer<float>> GetMapper()
@@ -601,6 +610,7 @@ namespace Microsoft.ML.Trainers
                 CanSavePfa = Predictors.All(m => (m as IDistCanSavePfa)?.CanSavePfa == true);
                 Contracts.AssertValue(inputType);
                 InputType = inputType;
+                OutputFormula = OutputFormula.ProbabilityNormalization;
             }
 
             private bool IsValid(IValueMapperDist mapper, ref VectorDataViewType inputType)
@@ -712,6 +722,7 @@ namespace Microsoft.ML.Trainers
                 CanSavePfa = false;
                 Contracts.AssertValue(inputType);
                 InputType = inputType;
+                OutputFormula = OutputFormula.Softmax;
             }
 
             public override ValueMapper<VBuffer<float>, VBuffer<float>> GetMapper()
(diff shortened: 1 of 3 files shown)


src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/OneVersusAllTrainer.cs
```diff
@@ -364,23 +364,29 @@ private OneVersusAllModelParameters(IHostEnvironment env, ModelLoadContext ctx)
                 : base(env, RegistrationName, ctx)
         {
             // *** Binary format ***
-            // bool: useDist
+            // byte: OutputFormula as byte
             // int: predictor count
-            bool useDist = ctx.Reader.ReadBoolByte();
+            OutputFormula outputFormula = (OutputFormula)ctx.Reader.ReadByte();
```
justinormont: Very elegant reuse of the byte.

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
```diff
@@ -308,7 +308,9 @@ public void LightGbmMulticlassEstimatorWithOptions()
         {
             var options = new LightGbmMulticlassTrainer.Options
             {
-                EvaluationMetric = LightGbmMulticlassTrainer.Options.EvaluateMetricType.Default
+                EvaluationMetric = LightGbmMulticlassTrainer.Options.EvaluateMetricType.Default,
+                UseSoftmax = true
```
justinormont: You may want two unit tests total; both setting it explicitly as `true` and `false`. I believe the hyerparameter defaults to `<auto>`, where LightGBM decides for you.

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/OneVersusAllTrainer.cs
harishsk: Thank you !

---
In reply to: [345483370](https://github.com/dotnet/machinelearning/pull/4472#discussion_r345483370) [](ancestors = 345483370)

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/OneVersusAllTrainer.cs
```diff
@@ -364,23 +364,29 @@ private OneVersusAllModelParameters(IHostEnvironment env, ModelLoadContext ctx)
                 : base(env, RegistrationName, ctx)
         {
             // *** Binary format ***
-            // bool: useDist
+            // byte: OutputFormula as byte
```
ganik: You are changing how model is saved / loaded from disk. For backward compatibility you will need to use versioning

test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs
harishsk: Fixed. Thanks!

---
In reply to: [345950381](https://github.com/dotnet/machinelearning/pull/4472#discussion_r345950381) [](ancestors = 345950381)

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/OneVersusAllTrainer.cs
harishsk: Backward compatibility isnt affected. In the old models, it stored a bool value as a byte. (0 = false = RawScores and 1 = true = Probability Normalization)
All I have done here is to add 2 = SoftMax. 

The old models will continue to be read and function as before. 


---
In reply to: [346003757](https://github.com/dotnet/machinelearning/pull/4472#discussion_r346003757) [](ancestors = 346003757)

PR #4475, opened 2019-11-13
Change the dataset used for unit test to fix issue on Linux

bpstark:
There is an issue with how some of the images were saved in the dataset
such that they were not being read properly on linux. As such accuracy
was dramatically reduced given the small size of the dataset to begin
with. Have corrected the images and made a new dataset such that this
issue should now be resolved.

### DIFF
--- a/test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
+++ b/test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
@@ -1455,7 +1455,14 @@ namespace Microsoft.ML.Scenarios
         [TensorFlowFact]
         public void TensorFlowImageClassificationWithPolynomialLRScheduling()
         {
-            TensorFlowImageClassificationWithLRScheduling(new PolynomialLRDecay(), 50);
+
+            /*
+             * Due to an issue with Nix based os performance is not as good,
+             * as such increase the number of epochs to produce a better model.
+             */
+            bool isNix = (!(RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ||
+                (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))));
+            TensorFlowImageClassificationWithLRScheduling(new PolynomialLRDecay(), isNix ? 75: 50);
         }
 
         internal void TensorFlowImageClassificationWithLRScheduling(LearningRateScheduler  learningRateScheduler, int epoch)
--- a/test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
+++ b/test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
@@ -1286,22 +1286,8 @@ namespace Microsoft.ML.Scenarios
             IDataView predictions = trainedModel.Transform(testDataset);
             var metrics = mlContext.MulticlassClassification.Evaluate(predictions);
 
-            // On Ubuntu the results seem to vary quite a bit but they can probably be 
-            // controlled by training more epochs, however that will slow the 
-            // build down. Accuracy values seen were 0.33, 0.66, 0.70+. The model
-            // seems to be unstable, there could be many reasons, will need to 
-            // investigate this further.
-            if (!(RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ||
-                (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))))
-            {
-                Assert.InRange(metrics.MicroAccuracy, 0.2, 1);
-                Assert.InRange(metrics.MacroAccuracy, 0.2, 1);
-            }
-            else
-            {
-                Assert.InRange(metrics.MicroAccuracy, 0.8, 1);
-                Assert.InRange(metrics.MacroAccuracy, 0.8, 1);
-            }
+            Assert.InRange(metrics.MicroAccuracy, 0.8, 1);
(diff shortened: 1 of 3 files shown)


codemzs:
>             }

what about here?

---
Refers to: test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs:1375 in 800721d. [](commit_id = 800721d13fb7ca228fea022477e8f8aa1b622897, deletion_comment = False)

codemzs:
>             }

and here?

---
Refers to: test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs:1279 in 800721d. [](commit_id = 800721d13fb7ca228fea022477e8f8aa1b622897, deletion_comment = False)

codemzs:
>             // investigate this further.

can we now remove the special if condition for nix? and then get rid of this comment?

---
Refers to: test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs:1522 in 800721d. [](commit_id = 800721d13fb7ca228fea022477e8f8aa1b622897, deletion_comment = False)

codemzs:
@bpstark This change does not fix the issue for lower accuracy completely. Please investigate why we are seeing lower accuracy on Linux.

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1857,6 +1781,12 @@ public void TensorFlowImageClassificationBadImages()
         {
             var files = Directory.GetFiles(folder, "*",
                 searchOption: SearchOption.AllDirectories);
+            /*
+             * This is only needed as Linux can produce files in a different 
+             * order than other OSs. As this is a test case we want to maintain
+             * consisten accuracy across all OSs, so we sort to remove this discrepency.
```
codemzs: >consisten [](start = 15, length = 9)

consistent

codemzs: >OSs [](start = 45, length = 3)

typo?

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1266,22 +1266,8 @@ public void TensorFlowImageClassificationDefault()
...
-            // build down. Accuracy values seen were 0.33, 0.66, 0.70+. The model
-            // seems to be unstable, there could be many reasons, will need to 
-            // investigate this further.
-            if (!(RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ||
-                (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))))
-            {
-                Assert.InRange(metrics.MicroAccuracy, 0.2, 1);
-                Assert.InRange(metrics.MacroAccuracy, 0.2, 1);
-            }
-            else
```
codemzs: also remove in AutoFitImageClassificationTrainTest

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1889,10 +1819,7 @@ public void TensorFlowImageClassificationBadImages()
         public static string DownloadImageSet(string imagesDownloadFolder)
         {
             string fileName = "flower_photos_tiny_set_for_unit_tests.zip";
-            string url = $"https://mlnetfilestorage.file.core.windows.net/imagesets" +
-                $"/flower_images/flower_photos_tiny_set_for_unit_tests.zip?st=2019" +
-                $"-08-29T00%3A07%3A21Z&se=2030-08-30T00%3A07%3A00Z&sp=rl&sv=2018" +
-                $"-03-28&sr=f&sig=N8HbLziTcT61kstprNLmn%2BDC0JoMrNwo6yRWb3hLLag%3D";
+            string url = "https://tlcresources.blob.core.windows.net/datasets/flower_photos_tiny_set_for_unit_test.zip";
```
justinormont: Please move to an aka.ms link pointing to the CDN. Recommend using the resource downloader class for your downloads.

justinormont: What kind of images are not loading? AutoML should detect this issue and notify the user. Just swapping out the images means users will still run into this issue.

bpstark: The issue ended up not being the images, but the ordering. I have changed the dataset, so we use a short link now, but the images were not bad, just had some odd metadata, that I removed as I thought that may have contributed to the issue initially, however that ended up not being the issue.

---
In reply to: [347016790](https://github.com/dotnet/machinelearning/pull/4475#discussion_r347016790) [](ancestors = 347016790)

justinormont: Can we fix the loader such that a user doesn't run into the same issue?

Does the fix address the underlying issue? Or does fixing the dataset just avoid the issue for our unit test?

justinormont: Ahh... the order of the dataset rows differed between Windows/Unix, which lead to different dataset splits, which lead to differing metrics.

If we haven't, we may want to modify our example code where we load a folder of images.

PR #4479, opened 2019-11-14
Ensure BufferBlocks are completed and empty in RowShufflingTransformer.

### LINKED ISSUE #4399: Memory leak in text classification pipeline
### System information

- **OS version/distro**: Microsoft Windows [Version 10.0.18362.418]
- **.NET Version (eg., dotnet --info)**: .NET Core 3.0.100 04339c3a26

### Issue

- **What did you do?** Detected potential memory leak in production application, so I wrote simple application to see if problem persists.
- **What happened?** Microsoft.ML 1.3.1 and 1.4.0-preview2 both appear to be leaking memory

### Source code / logs

Minimal code that consistently reproduces this problem:
```
    class ModelInput
    {
        [ColumnName("TextColumn"), LoadColumn(0)]
        public string ItemDescription { get; set; }
        [ColumnName("Label"), LoadColumn(1)]
        public int ItemId { get; set; }
    }

    class Program
    {
        {
            Console.WriteLine("Sleeping for 5 seconds. Collect initial memory snapshot...");
            while(true)
            {
                Thread.Sleep(5000);
                BuildAndTrainModel();
                Console.WriteLine("Training done. Collect memory snapshot...");
            }
        }

        static void BuildAndTrainModel()
        {
            MLContext context = new MLContext(seed: 1);

            var dataView = context.Data.LoadFromTextFile<ModelInput>("input.csv", separatorChar: ',');

            var trainingPipeline = context.Transforms.Conversion.MapValueToKey(new[] { new InputOutputColumnPair("Label", "Label") })
                .Append(context.Transforms.Text.FeaturizeText("TextColumn", "TextColumn"))
                .Append(context.MulticlassClassification.Trainers.SdcaMaximumEntropy(labelColumnName: "Label", featureColumnName: "TextColumn", maximumNumberOfIterations: 1));

            var model = trainingPipeline.Fit(dataView);
        }
    }
```
After each call to BuildAndTrainModel following objects appear to leak:
image

Amount of objects leaked appears to correlate with number of iterations (increasing maximumNumberOfIterations increases leaked object count).

eerhardt:
If BufferBlock doesn't get completed and drained of its items, it will have non-completed Tasks. When debugging in VS, this will appear to be a memory leak because VS adds all running Tasks to a static Dictionary, and then removes them when the Task is complete. If the Task doesn't get completed, it won't be removed from the Dictionary - thus it looks like a leak.

Note there is no leak when the VS Debugger isn't attached because the non-completed Tasks don't get added to the static Dictionary.

Fix #4399

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/RowShufflingTransformer.cs
+++ b/src/Microsoft.ML.Data/Transforms/RowShufflingTransformer.cs
@@ -591,7 +591,7 @@ namespace Microsoft.ML.Transforms
                 try
                 {
                     int circularIndex = 0;
-                    while (await _toProduce.OutputAvailableAsync())
+                    while (await _toProduce.OutputAvailableAsync().ConfigureAwait(false))
                     {
                         int requested;
                         if (!_toProduce.TryReceive(out requested))
--- a/src/Microsoft.ML.Data/Transforms/RowShufflingTransformer.cs
+++ b/src/Microsoft.ML.Data/Transforms/RowShufflingTransformer.cs
@@ -553,11 +553,24 @@ namespace Microsoft.ML.Transforms
             {
                 if (_disposed)
                     return;
-                if (disposing && _producerTask.Status == TaskStatus.Running)
+
+                if (disposing)
                 {
-                    _toProduce.Post(0);
+                    _toProduce.Complete();
                     _producerTask.Wait();
+
+                    // Complete the consumer after the producerTask has finished, since producerTask could
+                    // have posted more items to _toConsume.
+                    _toConsume.Complete();
+
+                    // Drain both BufferBlocks - this prevents what appears to be memory leaks when using the VS Debugger
+                    // because if a BufferBlock still contains items, its underlying Tasks are not getting completed.
+                    // See https://github.com/dotnet/corefx/issues/30582 for the VS Debugger issue.
+                    // See also https://github.com/dotnet/machinelearning/issues/4399
+                    _toProduce.TryReceiveAll(out _);
+                    _toConsume.TryReceiveAll(out _);
                 }
+
                 _disposed = true;
                 base.Dispose(disposing);
             }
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.Data/Transforms/RowShufflingTransformer.cs
```diff
@@ -553,11 +553,24 @@ protected override void Dispose(bool disposing)
...
+                    // Complete the consumer after the producerTask has finished, since producerTask could
+                    // have posted more items to _toConsume.
+                    _toConsume.Complete();
+
+                    // Drain both BufferBlocks - this prevents what appears to be memory leaks when using the VS Debugger
+                    // because if a BufferBlock still contains items, its underlying Tasks are not getting completed.
+                    // See https://github.com/dotnet/corefx/issues/30582 for the VS Debugger issue.
+                    // See also https://github.com/dotnet/machinelearning/issues/4399
+                    _toProduce.TryReceiveAll(out _);
+                    _toConsume.TryReceiveAll(out _);
```
stephentoub: In what situation does the LoopProducerWorker exit without consuming everything?  I see the `if (numRows < requested)` condition, but I'm wondering if that's required, or if removing that and these TryReceiveAlls would be better.

src/Microsoft.ML.Data/Transforms/RowShufflingTransformer.cs
```diff
@@ -578,15 +591,16 @@ private async Task LoopProducerWorker()
                 try
                 {
                     int circularIndex = 0;
-                    for (; ; )
+                    while (await _toProduce.OutputAvailableAsync())
```
stephentoub: I'm not sure what policy ML.NET has used for async in general, but generally we recommend libraries use `.ConfigureAwait(false)` on all awaits.

src/Microsoft.ML.Data/Transforms/RowShufflingTransformer.cs
eerhardt: The `_toConsume` can easily get in a situation where everything isn't consumed. All that would need to happen is for the caller to create one of these Cursors, enumerate over a few rows and then exit early without reading all the way to the end of the cursor. In that situation, `_toConsume` can still have items ready to consume, but no one got them.

The `_toProduce` gets into this situation exactly where you pointed out - the `if (numRows < requested)` it exits early and doesn't drain the `_toProduce` buffer.

We could change LoopProducerWorker to not exit early and always consume everything in `_toProduce`, but we would still need to leave `_toConsume.TryReceiveAll` here.

Thoughts?

src/Microsoft.ML.Data/Transforms/RowShufflingTransformer.cs
eerhardt: Good call. This was a mistake.

src/Microsoft.ML.Data/Transforms/RowShufflingTransformer.cs
stephentoub: What you have is fine.

(As an aside, I expect if you were to switch to using Channel instead of BufferBlock, you wouldn't see this debugger-induced leak, even without consuming everything from the block.)

eerhardt: Ok, I will leave this as-is for now. But I've logged https://github.com/dotnet/machinelearning/issues/4482 to replace the BufferBlock usage with Channel.

PR #4481, opened 2019-11-15
Improve key value mapping api docs

natke:
We received many comments that the documentation for value to key mapping and the reverse is not clear, including this issue: https://github.com/dotnet/docs/issues/13590.

This PR attempts to address those concerns.

Please review with a focus in clarity for the user of this API:
- what does this API do?
- when would I use each of the different methods?
- how does this API fit into the larger ML.NET framework?

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
+++ b/src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
@@ -214,7 +214,7 @@ namespace Microsoft.ML
         }
 
         /// <summary>
-        /// Create a <see cref="ValueToKeyMappingEstimator"/>, which converts categorical values into keys.
+        /// Create a <see cref="ValueToKeyMappingEstimator"/>, which converts categorical values into numerical keys.
         /// </summary>
         /// <param name="catalog">The conversion transform's catalog.</param>
         /// <param name="outputColumnName">Name of the column containing the keys.</param>
@@ -225,7 +225,9 @@ namespace Microsoft.ML
         /// <param name="keyOrdinality">The order in which keys are assigned.
         /// If set to <see cref="ValueToKeyMappingEstimator.KeyOrdinality.ByOccurrence"/>, keys are assigned in the order encountered.
         /// If set to <see cref="ValueToKeyMappingEstimator.KeyOrdinality.ByValue"/>, values are sorted, and keys are assigned based on the sort order.</param>
-        /// <param name="addKeyValueAnnotationsAsText">Whether key value annotations should be text, regardless of the actual input type.</param>
+        /// <param name="addKeyValueAnnotationsAsText">If set to true, use text type
+        /// for values, regardless of the actual input type. When doing the reverse
+        /// mapping, the values are text rather than the original input type.</param>
         /// <param name="keyData">Use a pre-defined mapping between values and keys, instead of building
         /// the mapping from the input data during training. If specified, this should be a single column <see cref="IDataView"/> containing the values.
         /// The keys are allocated based on the value of keyOrdinality.</param>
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.Data/Transforms/ValueToKeyMappingEstimator.cs
```diff
@@ -22,13 +25,16 @@ namespace Microsoft.ML.Transforms
...
-    /// The ValueToKeyMappingEstimator builds up term vocabularies(dictionaries) mapping the input values to the keys on the dictionary.
-    /// If multiple columns are used, each column builds/uses exactly one vocabulary.
-    /// The output columns are KeyDataViewType-valued.
-    /// The Key value is the one-based index of the item in the dictionary.
-    /// If the key is not found in the dictionary, it is assigned the missing value indicator.
-    /// This dictionary mapping values to keys is most commonly learnt from the unique values in input data,
-    /// but can be defined through other means: either with the mapping defined, or as loaded from an external file.
+    /// The ValueToKeyMappingEstimator maps the input values to keys using a
+    /// dictionary that is built during training. The dictionary mapping values to
+    /// keys is most commonly learnt from the unique values in input data,
```
codemzs: >most commonly [](start = 16, length = 13)

is this needed?

natke: Do you mean the whole sentence, or the word "most"?

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingEstimator.cs
```diff
@@ -108,9 +114,12 @@ internal sealed class ColumnOptions : ColumnOptionsBase
...
-            /// <param name="keyOrdinality">How items should be ordered when vectorized. If <see cref="KeyOrdinality.ByOccurrence"/> choosen they will be in the order encountered.
-            /// If <see cref="KeyOrdinality.ByValue"/>, items are sorted according to their default comparison, for example, text sorting will be case sensitive (for example, 'A' then 'Z' then 'a').</param>
-            /// <param name="addKeyValueAnnotationsAsText">Whether key value annotations should be text, regardless of the actual input type.</param>
+            /// <param name="keyOrdinality">The order in which keys are assigned.
+            /// If set to <see cref="ValueToKeyMappingEstimator.KeyOrdinality.ByOccurrence"/>, keys are assigned in the order encountered.
+            /// If set to <see cref="ValueToKeyMappingEstimator.KeyOrdinality.ByValue"/>, values are sorted, and keys are assigned based on the sort order.</param>
+            /// <param name="addKeyValueAnnotationsAsText">If set to true, use text type
+            /// for values, regardless of the actual input type. When doing the reverse
+            /// mapping, the values are text rather than the original input type.</param>
```
codemzs: great! :)

src/Microsoft.ML.DataView/KeyDataViewType.cs
```diff
@@ -9,24 +9,27 @@
...
+    /// The underlying .NET type is one of the unsigned integer types. The default is
+    /// <see cref="uint"/>, but it can also be <see cref="byte"/>,
+    /// <see cref="ushort"/>, or <see cref="ulong"/>.
+    /// Despite keys being numerical types, the information is not inherently numeric,
+    /// so typically, arithmetic is not meaningful.
     ///
-    /// Note that for data of this type, a value of 0, being the default value of the representation type, indicates
-    /// the missing value since it would not be sensible for the default value to correspond to any one particular specific
-    /// value of the set. The first non-missing value for the enumeration of the set is always <c>1</c>.
+    /// Missing values are mapped to 0.
```
codemzs: such great documentation

src/Microsoft.ML.Data/Transforms/ConversionsExtensionsCatalog.cs
```diff
@@ -246,20 +249,23 @@ public static KeyToValueMappingEstimator MapKeyToValue(this TransformsCatalog.Co
...
+        /// <param name="maximumNumberOfKeys">Maximum number of keys to keep per column when training.</param>
+        /// <param name="keyOrdinality">The order in which keys are assigned.
+        /// If set to <see cref="ValueToKeyMappingEstimator.KeyOrdinality.ByOccurrence"/>, keys are assigned in the order encountered.
+        /// If set to <see cref="ValueToKeyMappingEstimator.KeyOrdinality.ByValue"/>, values are sorted, and keys are assigned based on the sort order.</param>
+        /// <param name="addKeyValueAnnotationsAsText">If set to true, use text type
+        /// for values, regardless of the actual input type. When doing the reverse
+        /// mapping, the values are text rather than the original input type.</param>
+        /// <param name="keyData">Use a pre-defined mapping between values and keys, instead of building
+        /// the mapping from the input data during training. If specified, this should be a single column <see cref="IDataView"/> containing the values.
+        /// The keys are allocated based on the value of keyOrdinality.</param>
```
codemzs: I wish we had a sample for this because it is useful in cases where train set and test set may not have exactly the same labels but you still want the mapping to be consistent (sorting the labels does not take care of this situation).

natke: Right. That's a great use case for this. I can add it to these docs as an example

PR #4485, opened 2019-11-18
Fixes #4385 about calling the Create methods when loading models from disk

### LINKED ISSUE #4385: Create methods not being called when loading models from disk
There are some classes that define a Create method which is supposed to be called when loading a model from disk, but the problem is that the method is not being called at all.

For example, I've noticed this unexpected behavior in the following classes:
1. [LinearBinaryModelParameters](https://github.com/dotnet/machinelearning/blob/d531ea801a34a3018b8fa2a2f352902eb703cd25/src/Microsoft.ML.StandardTrainers/Standard/LinearModelParameters.cs#L473)
2. [GamBinaryModelParameters](https://github.com/dotnet/machinelearning/blob/d531ea801a34a3018b8fa2a2f352902eb703cd25/src/Microsoft.ML.FastTree/GamClassification.cs#L233)
3. [FastTreeBinaryModelParameters](https://github.com/dotnet/machinelearning/blob/d531ea801a34a3018b8fa2a2f352902eb703cd25/src/Microsoft.ML.FastTree/FastTreeClassification.cs#L87)
4. [FastForestBinaryModelParameters](https://github.com/dotnet/machinelearning/blob/d531ea801a34a3018b8fa2a2f352902eb703cd25/src/Microsoft.ML.FastTree/RandomForestClassification.cs#L102)
5. [PlattCalibrator](https://github.com/dotnet/machinelearning/blob/c922529e669d1c4dcb4d2bf8157a579b10a60cee/src/Microsoft.ML.Data/Prediction/Calibrator.cs#L1569)

when loading a model that uses any of these classes, their Create method is expected to be called, but, as stated, the method is not being called.

I also noticed this behavior in 3 other classes of the [Calibrator.cs](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Prediction/Calibrator.cs#L1569) file (namely, the `ParameterMixingCalibratedModelParameters`, `ValueMapperCalibratedModelParameters`, and `FeatureWeightsCalibratedModelParameters` classes), but I've fixed this problem for those specific classes in my recent PR #4306 (which, as of the moment of writing this, is still waiting to get approved). It was while working on that PR that I noticed this problem on these classes, and I commented about it [there](https://github.com/dotnet/machinelearning/pull/4306#discussion_r336139379)... but it is appropriate to open this separate issue to better document this, since it is a problem that affects different classes across different files.

In fact, as I will describe below, there's a certain code pattern that is related to this problem, and I've seen this pattern in other classes of the Calibrator.cs file as well. So, the problem I describe here might be affecting even more classes than the ones I've mentioned.

### Cause of the problem
The [CreateInstanceCore](https://github.com/dotnet/machinelearning/blob/7c067854b564275b0d6387ca59c0ec83e8fc91b9/src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs#L186) method in the `ComponentCatalog.cs` file is responsible to try to call the `Create` method of any class when loading a model from disk.

The current implementation of the method actually checks first if the class has a constructor with parameters `(IHostEnvironment env, ModelLoadContext ctx)` and invokes that constructor through reflection. If the class doesn't have such a constructor, then it checks if it has a `Create((IHostEnvironment env, ModelLoadContext ctx)` method, and it gets invoked through reflection.

This behavior is not desired for the classes I've mentioned (and potentially other classes), since they define both a constructor and a Create method with those parameters, but in these cases it's actually expected that the Create method gets called instead of the constructor. Thus, if a class follows the pattern of having a private or internal constructor (with the `(env, ctx)` parameters) and also has a Create method, then this problem might also be affecting that class.

Since the Create method typically only runs some security checks before calling the constructor, it turns out that the overall process of loading models doesn't seem affected by this issue. But the problem remains that these security checks are being missed along with whatever behavior the Create method adds to the process.

As explained by @yaeldekel in [her comment](https://github.com/dotnet/machinelearning/pull/4306#discussion_r337981820) on my recent PR #4306 (under "Answer 1"), this problem might had been introduced before the official release of ML.net, when the `ComponentCatalog` method was modified in a way that permitted the `CreateInstanceCore` method to use private and internal constructors, which didn't use to happen... so before those changes were made, classes could have private or internal constructors and a Create method, and the latter would appropriately be called. But now the constructor gets called, and this is the case of all the classes mentioned in this issue.

Since these changes were made while trying to internalize as many APIs as possible before the ML.net official release, many constructors where also made private or internal, and thus the changes in ComponentCatalog that permit using those constructors are also necessary.

Because of these, further investigation is needed to know for sure which classes are being affected by this problem, so to better find a way to fix this problem without affecting all of the other classes that doesn't present this situation.

antoniovs1029:
Fixes #4385 

As concluded in the discussion there, if a class has both a constructor and a create method that matches the parameter types that the `ComponentCatalog `is looking for, then it should use the one which is public. If both are non-public, then it should use the internal one. If both have the same visibility then it should throw an exception.

For this to happen, I modified the `ComponentCatalog ` to work as described. I also had to change the visibility of several methods and constructors to work as described. Since, as per @yaeldekel 's instructions, the create method should be called instead of the constructor in all cases, then I did the following:
1. For most classes, the case was that both the conflictive constructor and create method were private, so I simply made the create method internal.
2. For LabelIndicatorTransform and SkipTakeFilter the conflictive constructors and create methods were public, so I made the constructors internal.
3. For the following classes the constructor was internal, and the create method was private. In these cases I made the constructor private, and the create methods internal.

```
Microsoft.ML.Trainers.FieldAwareFactorizationMachinePredictionTransformer
Microsoft.ML.Transforms.TimeSeries.IidChangePointDetector
Microsoft.ML.Transforms.TimeSeries.IidSpikeDetector
Microsoft.ML.Transforms.TimeSeries.SrCnnAnomalyDetector
Microsoft.ML.Transforms.TimeSeries.SsaChangePointDetector
Microsoft.ML.Transforms.TimeSeries.SsaForecastingTransformer
Microsoft.ML.Transforms.TimeSeries.SsaSpikeDetector
```


src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
```diff
@@ -414,21 +414,78 @@ internal ComponentInfo(Type interfaceType, string kind, Type argumentType, TlcMo
...
+                create = FindCreateMethod(instType, loaderType, parmTypesWithEnv ?? Type.EmptyTypes);
+                if (create != null)
+                    requireEnvironmentCreate = true;
+            }
+
+            // If both 'ctor' and 'create' methods were found
+            // Choose the one that is public, if both are non-public choose the one that isn't private
+            // if they have the same visibility, then throw an exception, since this shouldn't happen.
+            if(ctor != null && create != null)
+            {
```
antoniovs1029: I don't really like this nested if's, I am not sure if they're legible enough, and it is ambiguous what should happen if there's a 'protected' create or constructor method (which, I believe, never happens in the codebase...). Still, this gets the job done.

I can think of a couple of ways of making this, but not sure if they would be more legible. Please, let me know if I should rewrite this in another way.

src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
```diff
@@ -414,21 +414,78 @@ internal ComponentInfo(Type interfaceType, string kind, Type argumentType, TlcMo
...
+                if (ctor.IsPublic == create.IsPublic)
+                {
+                    if(ctor.IsPublic || ctor.IsPrivate == create.IsPrivate)
+                    {
+                        throw Contracts.Except($"Can't load type {instType}, because it has both create and constructor methods with the same visibility. Please open an issue for this to be fixed.");
```
antoniovs1029: I didn't know what text to put in the Exception. Please let me know if I should change it.

src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
```diff
@@ -414,21 +462,64 @@ internal ComponentInfo(Type interfaceType, string kind, Type argumentType, TlcMo
...
+            }
+
+            if (ctor != null && create != null)
+            {
+                // If both 'ctor' and 'create' methods were found
+                // Choose the one that is 'more' public
+                // If they have the same visibility, then throw an exception, since this shouldn't happen.
+
+                if (ctor.Accessmodifier() == create.Accessmodifier())
+                {
```
antoniovs1029: So I changed the nested if's for this other solution using the Accessmodifier() extension method. Although I think this one is more legible, I wouldn't be sure if it's worth it to create the extension method. So let me know your opinions, Thanks!

src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
antoniovs1029: So I've changed this nested if's in a new iteration ([see this comment](https://github.com/dotnet/machinelearning/pull/4485#discussion_r347688260)) but I am not sure if I prefer the nested of's or the new solution.

src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
yaeldekel: I would change "please open an issue..." to something like: "Please indicate which one should be used by changing either the signature or the visibility of one of them".

---
In reply to: [347643643](https://github.com/dotnet/machinelearning/pull/4485#discussion_r347643643) [](ancestors = 347643643)

src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
yaeldekel: To me this seems cleaner.

A couple of more ways you can decrease the amount of "if/else"s in the code:
1. you can put the assignment inside the if condition, like this:
```
if ((ctor = loaderType.GetConstructor(...)) == null)
```
2. If you throw or return inside the "if", then you don't need the "else":
```
if (ctor.Accessmodifier() == create.Accessmodifier())
    throw ...
if (ctor.Accessmodifier() > create.Accessmodifier())
{
    ...
    return true
}
if (ctor.Accessmodifier() < create.Accessmodifier())
...
etc.
```

---
In reply to: [347688260](https://github.com/dotnet/machinelearning/pull/4485#discussion_r347688260) [](ancestors = 347688260)

PR #4490, opened 2019-11-20
Ignore hidden columns in AutoML schema checks of validation data

### LINKED ISSUE #4491: Ignore hidden columns in AutoML schema checks of validation data
When the AutoML API consumes data, it validates schema consistency between the train and validation data.

There are two bugs in this logic:
1. The API asserts that the count of columns in the train and validation data must be equal. This throws an exception if the two data views have the same number of active columns but a different number of hidden columns. This should be updated to assert that the # of active (not hidden) columns in the train and validation data are equal.

2. If either the train or validation data has a hidden column with a type that differs from an active column of the same name, an exception is thrown. Type consistency checks should be restricted to active columns only.

daholste:
Closes #4491

When the AutoML API consumes data, it validates schema consistency between the train and validation data.

There are two bugs in this logic:

1. The API asserts that the count of columns in the train and validation data must be equal. This throws an exception if the two data views have the same number of active columns but a different number of hidden columns. This PR updates to assert that the # of active (not hidden) columns in the train and validation data are equal.

2. If either the train or validation data has a hidden column with a type that differs from an active column of the same name, an exception is thrown. This PR restricts type consistency checks to active columns only.

### DIFF
--- a/src/Microsoft.ML.AutoML/Utils/UserInputValidationUtil.cs
+++ b/src/Microsoft.ML.AutoML/Utils/UserInputValidationUtil.cs
@@ -183,7 +183,7 @@ namespace Microsoft.ML.AutoML
 
             const string schemaMismatchError = "Training data and validation data schemas do not match.";
 
-            if (trainData.Schema.Count != validationData.Schema.Count)
+            if (trainData.Schema.Count(c => !c.IsHidden) != validationData.Schema.Count(c => !c.IsHidden))
             {
                 throw new ArgumentException($"{schemaMismatchError} Train data has '{trainData.Schema.Count}' columns," +
                     $"and validation data has '{validationData.Schema.Count}' columns.", nameof(validationData));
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.AutoML/Utils/UserInputValidationUtil.cs
```diff
@@ -183,14 +183,19 @@ private static void ValidateValidationData(IDataView trainData, IDataView valida
...
             const string schemaMismatchError = "Training data and validation data schemas do not match.";
 
-            if (trainData.Schema.Count != validationData.Schema.Count)
+            if (trainData.Schema.Count(c => !c.IsHidden) != validationData.Schema.Count(c => !c.IsHidden))
             {
                 throw new ArgumentException($"{schemaMismatchError} Train data has '{trainData.Schema.Count}' columns," +
                     $"and validation data has '{validationData.Schema.Count}' columns.", nameof(validationData));
             }
 
             foreach (var trainCol in trainData.Schema)
```
justinormont: Could use a comment...
```suggestion
            // Also indirectly checks for new columns in the validation datasets as we above enforce the column counts are equal
            foreach (var trainCol in trainData.Schema)
```

I was otherwise going to suggest we check the reverse direction. A comment helps future readers to not have to think through it.

PR #4494, opened 2019-11-21
Add SQL command timeout option to database loader.

### LINKED ISSUE #4484: Command Timeouts While Using The Database Loader
### System information

- **Windows 10**:
- **.NET Core 3.0**: 

### Issue

- Loaded training data from SQL Server (1m rows)
- SqlException Execution timeout in less than a minute
- The query should complete even if it's slow (based on volume). There should also be a way to modify the sql command timeout.

### Source code / logs

`          string connectionString = _config["Database.ConnectionString"];

            string sqlCommand = await GetSql("OrderHistory.sql");

            DatabaseSource dbSource = new DatabaseSource(SqlClientFactory.Instance, connectionString, sqlCommand);

            return dbSource;`

ashbhandare:
Fixes #4484 

This change adds a way to specify the SQL command timeout https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlcommand.commandtimeout?view=netframework-4.8
while creating the DatabaseSource.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseSource.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseSource.cs
@@ -11,7 +11,7 @@ namespace Microsoft.ML.Data
     public sealed class DatabaseSource
     {
         /// <summary>Creates a new instance of the <see cref="DatabaseSource" /> class.</summary>
-        /// <param name="providerFactory">The factory used to create the <see cref="DbConnection"/>..</param>
+        /// <param name="providerFactory">The factory used to create the <see cref="DbConnection"/>.</param>
         /// <param name="connectionString">The string used to open the connection.</param>
         /// <param name="commandText">The text command to run against the data source.</param>
         public DatabaseSource(DbProviderFactory providerFactory, string connectionString, string commandText)
@@ -26,7 +26,7 @@ namespace Microsoft.ML.Data
         }
 
         /// <summary>Creates a new instance of the <see cref="DatabaseSource" /> class.</summary>
-        /// <param name="providerFactory">The factory used to create the <see cref="DbConnection"/>..</param>
+        /// <param name="providerFactory">The factory used to create the <see cref="DbConnection"/>.</param>
         /// <param name="connectionString">The string used to open the connection.</param>
         /// <param name="commandText">The text command to run against the data source.</param>
         /// <param name="commandTimeoutInSeconds">The time in seconds to wait for the command to execute.</param>
--- a/src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseSource.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseSource.cs
@@ -14,8 +14,23 @@ namespace Microsoft.ML.Data
         /// <param name="providerFactory">The factory used to create the <see cref="DbConnection"/>..</param>
         /// <param name="connectionString">The string used to open the connection.</param>
         /// <param name="commandText">The text command to run against the data source.</param>
-        /// <param name="commandTimeout">The time in seconds to wait for the command to execute. Default is 30 sec.</param>
-        public DatabaseSource(DbProviderFactory providerFactory, string connectionString, string commandText, int commandTimeout = 30)
+        public DatabaseSource(DbProviderFactory providerFactory, string connectionString, string commandText)
+        {
+            Contracts.CheckValue(providerFactory, nameof(providerFactory));
+            Contracts.CheckValue(connectionString, nameof(connectionString));
+            Contracts.CheckValue(commandText, nameof(commandText));
+
+            ProviderFactory = providerFactory;
+            ConnectionString = connectionString;
+            CommandText = commandText;
+        }
+
+        /// <summary>Creates a new instance of the <see cref="DatabaseSource" /> class.</summary>
+        /// <param name="providerFactory">The factory used to create the <see cref="DbConnection"/>..</param>
+        /// <param name="connectionString">The string used to open the connection.</param>
+        /// <param name="commandText">The text command to run against the data source.</param>
+        /// <param name="commandTimeout">The time in seconds to wait for the command to execute.</param>
+        public DatabaseSource(DbProviderFactory providerFactory, string connectionString, string commandText, int commandTimeout)
         {
             Contracts.CheckValue(providerFactory, nameof(providerFactory));
             Contracts.CheckValue(connectionString, nameof(connectionString));
--- a/src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseSource.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseSource.cs
@@ -29,8 +29,8 @@ namespace Microsoft.ML.Data
         /// <param name="providerFactory">The factory used to create the <see cref="DbConnection"/>..</param>
         /// <param name="connectionString">The string used to open the connection.</param>
         /// <param name="commandText">The text command to run against the data source.</param>
-        /// <param name="commandTimeout">The time in seconds to wait for the command to execute.</param>
-        public DatabaseSource(DbProviderFactory providerFactory, string connectionString, string commandText, int commandTimeout)
+        /// <param name="commandTimeoutInSeconds">The time in seconds to wait for the command to execute.</param>
+        public DatabaseSource(DbProviderFactory providerFactory, string connectionString, string commandText, int commandTimeoutInSeconds)
         {
             Contracts.CheckValue(providerFactory, nameof(providerFactory));
             Contracts.CheckValue(connectionString, nameof(connectionString));
@@ -39,14 +39,14 @@ namespace Microsoft.ML.Data
             ProviderFactory = providerFactory;
             ConnectionString = connectionString;
             CommandText = commandText;
-            CommandTimeout = commandTimeout;
+            CommandTimeoutInSeconds = commandTimeoutInSeconds;
         }
 
         /// <summary>Gets the text command to run against the data source.</summary>
         public string CommandText { get; }
 
         /// <summary>Gets the command timeout.</summary>
-        public int CommandTimeout { get; }
+        public int CommandTimeoutInSeconds { get; }
 
         /// <summary>Gets the string used to open the connection.</summary>
         public string ConnectionString { get; }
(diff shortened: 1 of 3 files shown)


src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseSource.cs
```diff
@@ -14,7 +14,8 @@ public sealed class DatabaseSource
         /// <param name="providerFactory">The factory used to create the <see cref="DbConnection"/>..</param>
         /// <param name="connectionString">The string used to open the connection.</param>
         /// <param name="commandText">The text command to run against the data source.</param>
-        public DatabaseSource(DbProviderFactory providerFactory, string connectionString, string commandText)
+        /// <param name="commandTimeout">The time in seconds to wait for the command to execute. Default is 30 sec.</param>
+        public DatabaseSource(DbProviderFactory providerFactory, string connectionString, string commandText, int commandTimeout = 30)
```
eerhardt: This is a breaking change. You can't add a parameter (even an optional parameter) to an already shipped method.

You will need to create a new overload to add this new property.

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
```diff
@@ -79,6 +79,63 @@ public void IrisLightGbm()
             }).PredictedLabel);
         }
 
+        [LightGBMFact]
+        public void IrisLightGbmConnectionTimeout()
```
eerhardt: How is this testing the timeout?

src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseSource.cs
tannergooding: It's likely worth making it explicitly clear that this is in seconds.

ashbhandare: changing

---
In reply to: [349252694](https://github.com/dotnet/machinelearning/pull/4494#discussion_r349252694) [](ancestors = 349252694)

ashbhandare: The XML description of the parameter does state that the time is in seconds already 

---
In reply to: [349254631](https://github.com/dotnet/machinelearning/pull/4494#discussion_r349254631) [](ancestors = 349254631)

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
ashbhandare: what would you recommend that the test look like?

---
In reply to: [349252966](https://github.com/dotnet/machinelearning/pull/4494#discussion_r349252966) [](ancestors = 349252966)

eerhardt: I'm not exactly sure. But my initial thinking is doing something that would make the command timeout, and ensuring the correct behavior occurs.

ashbhandare: added test where SQL query has delay added, and command timeout is exercised.

---
In reply to: [349721564](https://github.com/dotnet/machinelearning/pull/4494#discussion_r349721564) [](ancestors = 349721564)

src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseSource.cs
veikkoeeva: If I may, as a bit of an outsider, for a reason or another it's not always easy to see XML comments. So being explicit like `commandTimeoutInSeconds = 30` is easier to catch. I believe it's sort of a pattern used quite commonly too.

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
veikkoeeva: Should you also test for user initiated cancellation? I wonder if this way of testing introduces test failures on the infra or problems with other databases than SQL Server?

As for an example, here is one another way: https://github.com/dotnet/orleans/blob/585680f3c06cb82f22e07b65330b769d333a8aae/test/Extensions/TesterAdoNet/StorageTests/RelationalStoreTests.cs#L179. (For context, this is actually testing timeouts while streaming data. This piece isn't the best possible otherwise, but should give perhaps food for thoughts in the future.)

src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseSource.cs
ashbhandare: Wouldn't matching the name of the actual property that is being set be more consistent. See [https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlcommand.commandtimeout?view=netcore-3.0](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlcommand.commandtimeout?view=netcore-3.0). 

---
In reply to: [351692404](https://github.com/dotnet/machinelearning/pull/4494#discussion_r351692404) [](ancestors = 351692404)

tannergooding: The ML API here is not, strictly speaking, tied to SQL. It is meant to be usable with any database provider. It is more "tied" to `DbCommand.CommandTimeout`: https://docs.microsoft.com/en-us/dotnet/api/system.data.common.dbcommand.commandtimeout?view=netframework-4.8#System_Data_Common_DbCommand_CommandTimeout

Adding `InSeconds` just helps clarify the unit of measurement that the timeout is (which can be particularly helpful as .NET APIs frequently deal in milliseconds instead).

ashbhandare: changing.

---
In reply to: [352740161](https://github.com/dotnet/machinelearning/pull/4494#discussion_r352740161) [](ancestors = 352740161)

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
ashbhandare: > Should you also test for user initiated cancellation? I wonder if this way of testing introduces test failures on the infra or problems with other databases than SQL Server?
> 
> As for an example, here is one another way: https://github.com/dotnet/orleans/blob/585680f3c06cb82f22e07b65330b769d333a8aae/test/Extensions/TesterAdoNet/StorageTests/RelationalStoreTests.cs#L179. (For context, this is actually testing timeouts while streaming data. This piece isn't the best possible otherwise, but should give perhaps food for thoughts in the future.)

Could you elaborate on what 'user-initiated cancellation' would mean in this context?

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
```diff
@@ -79,6 +80,59 @@ public void IrisLightGbm()
...
+
+                IEstimator<ITransformer> pipeline = mlContext.Transforms.Conversion.MapValueToKey("Label")
+                .Append(mlContext.Transforms.Concatenate("Features", "SepalLength", "SepalWidth", "PetalLength", "PetalWidth"))
+                .AppendCacheCheckpoint(mlContext)
+                .Append(mlContext.MulticlassClassification.Trainers.LightGbm())
+                .Append(mlContext.Transforms.Conversion.MapKeyToValue("PredictedLabel"));
+
+                var model = pipeline.Fit(trainingData);
+            }
+            catch(Exception e)
```
eerhardt: You can use `Assert.Throws` instead.

Also - you should ensure the correct type of exception is thrown.

src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseSource.cs
```diff
@@ -25,9 +25,29 @@ public DatabaseSource(DbProviderFactory providerFactory, string connectionString
             CommandText = commandText;
         }
 
+        /// <summary>Creates a new instance of the <see cref="DatabaseSource" /> class.</summary>
+        /// <param name="providerFactory">The factory used to create the <see cref="DbConnection"/>..</param>
```
eerhardt: Why the double ending `..` ?

eerhardt:
@tannergooding - any further comments on this?

src/Microsoft.ML.Data/DataLoadSave/Database/DatabaseSource.cs
ashbhandare: changing here and in the above constructor.

---
In reply to: [355678098](https://github.com/dotnet/machinelearning/pull/4494#discussion_r355678098) [](ancestors = 355678098)

test/Microsoft.ML.Tests/DatabaseLoaderTests.cs
veikkoeeva: @ashbhandare I was thinking that why fix the command timeout at https://github.com/dotnet/machinelearning/pull/4494/files#diff-6bd895b22ba4c7e741ba7526ac214e5aR33 when it could be per query? While maybe also providing an overload to a user supplied cancellation token and then writing a test that excercises that token and timeout. But maybe I misunderstood this and `DatabaseSource` is intended to use per query or this is the behaviour hoped.

Sorry for the delay, I lost the plot for a while.

thtemme:
I'm voting strongly for this bugfix!
In a big data pipeline long running SQL Statements are normal. So it is not clear why ML.NET is restricted to 30 Seconds timeout in SQL Statement.
Please fix this!

mstfbl:
Hi @thtemme, what is the upper bound you've seen for the timeout of SQL statements, and where have you seen them? We can update our timeouts if it is indeed the case that 30 seconds isn't enough to load big datasets in ML .NET. Thanks!

codemzs:
@mstfbl You don't want to update timeouts statically as your response seems to imply but we must instead add an option that allows changing timeouts dynamically which is exactly what this change is _trying_  to do.

CC: @harishsk

mstfbl:
I understand @codemzs , I implied that if @thtemme can provide examples on where 30 seconds isn't enough to load in datasets, then we can update our timeouts through this PR.

codemzs:
@mstfbl This PR is NOT updating the timeout. It is providing an option to update the timeout at runtime. The user can specify whatever timeout they want once this PR is merged. You do not need to update the timeout as you say "then we can update our timeouts through this PR.".

PR #4496, opened 2019-11-22
Adds mlnet src and test projects from feature branch

maryamariyan:
Adds mlnet src and test projects from features/automl into master branch

- [x] pack mlnet

cc: @LittleLittleCloud @eerhardt


src/mlnet/mlnet.csproj
```diff
@@ -0,0 +1,55 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
LittleLittleCloud: Can we upgrade this to net core 3.0? it should be simply a one - line change @maryamariyan @JakeRadMSFT 

I know that the current recommend way from dotnet is to downgrade to netcore 2.1. But we continuously received issues from customers complaining about cli can't run on netcore 3.0. Even after hint message about downgrading was shown.

src/mlnet/Assembly.cs
```diff
@@ -0,0 +1,7 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("mlnet.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010015c01ae1f50e8cc09ba9eac9147cf8fd9fce2cfe9f8dce4f7301c4132ca9fb50ce8cbf1df4dc18dd4d210e4345c744ecb3365ed327efdbc52603faa5e21daa11234c8c4a73e51f03bf192544581ebe107adee3a34928e39d04e524a9ce729d5090bfd7dad9d10c722c0def9ccc08ff0a03790e48bcd1f9b6c476063e1966a1c4")]
```
LittleLittleCloud: Do we need to do this explicitly after merging to master?

pkg/mlnet/mlnet.nupkgproj
```diff
@@ -0,0 +1,27 @@
...
+
+  <ItemGroup>
+    <PackageReference Include="NLog" Version="$(NLogVersion)" />
+    <PackageReference Include="System.CommandLine.Experimental" Version="$(SystemCommandLineExperimentalVersion)" />
+    <PackageReference Include="System.Text.Encoding.CodePages" Version="$(SystemTextEncodingCodePagesVersion)" />
+    <PackageReference Include="Microsoft.ApplicationInsights" Version="$(MicrosoftApplicationInsightsVersion)" />
+    <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="$(MicrosoftDotNetPlatformAbstractionsVersion)" />
+  </ItemGroup>
+
+  <Target Name="CreateManifestResourceNames" />
```
LittleLittleCloud: What's this empty task for...

src/mlnet/mlnet.csproj
eerhardt: Let's do that in a separate change from porting the code. Can you please log an issue for it?

The drawback is that once you upgrade to `3.0`, you are requiring users to install `3.0`. But `2.1` is the current long-term support version.

pkg/mlnet/mlnet.nupkgproj
eerhardt: It is required by the Common targets in MSBuild, but not used here.

See https://github.com/search?p=2&q=org%3Adotnet+CreateManifestResourceNames&type=Code for others doing the same.

Let's add a comment here, like in the other places.

src/mlnet/mlnet.csproj
JakeRadMSFT: @eerhardt Can we support both somehow?

eerhardt: No matter what, I feel this is a separate discussion from merging the branch into `master`.

I'm not sure how to support both. It may be possible, but that may involve duplicating all the assemblies and doubling the size of the NuGet package.

justinormont: As Eric says, this can be a separate discussion.

The example on https://docs.microsoft.com/en-us/dotnet/standard/frameworks may work.

> The following library project file targets APIs of .NET Standard (netstandard1.4) and APIs of the .NET Framework (net40 and net45). Use the plural **TargetFrameworks** element with multiple target frameworks. Note how the Condition attributes include implementation-specific packages when the library is compiled for the two .NET Framework TFMs:
> 
> ```xml
> 
>   <PropertyGroup>
>     <TargetFrameworks>netstandard1.4;net40;net45</TargetFrameworks>
>   </PropertyGroup>
> 
>   
>   <ItemGroup Condition=" '$(TargetFramework)' == 'net40' ">
>     <Reference Include="System.Net" />
>   </ItemGroup>
> 
>   
>   <ItemGroup Condition=" '$(TargetFramework)' == 'net45' ">
>     <Reference Include="System.Net.Http" />
>     <Reference Include="System.Threading.Tasks" />
>   </ItemGroup>
> 
> </Project>
> ```
> Within your library or app, you write conditional code to compile for each target framework:
> ```c#
> public class MyClass
> {
>     static void Main()
>     {
> #if NET40
>         Console.WriteLine("Target framework: .NET Framework 4.0");
> #elif NET45  
>         Console.WriteLine("Target framework: .NET Framework 4.5");
> #else
>         Console.WriteLine("Target framework: .NET Standard 1.4");
> #endif
>     }
> }
> ```

I think ours would be:
```suggestion
    <TargetFrameworks>netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
```

maryamariyan:
moving mlnet to machinelearning-tools repo instead

PR #4497, opened 2019-11-22
wip - CodeGen for Azure Attach Image

LittleLittleCloud:
Transformer added:
- ExtractPixel
- NormalizeMapping
- ResizeImage
- ApplyOnnxModel

Notify that those are transformers that not exist in AutoML


src/Microsoft.ML.CodeGenerator/CodeGenerator/CSharp/CodeGenerator.cs
```diff
@@ -89,6 +90,53 @@ public void GenerateOutput()
             Utils.AddProjectsToSolution(modelprojectDir, modelProjectName, consoleAppProjectDir, consoleAppProjectName, solutionPath);
         }
 
+        public void GenerateAzureRemoteImageOutput()
```
LittleLittleCloud: Ideal way is we don't differentiate pipelines from AutoML / Azure AutoML, However there's already too many if-else conditions and some function's parameter list are so long to be bearable. So I feel like make it a special case can be a better choice

src/Microsoft.ML.CodeGenerator/CodeGenerator/CSharp/TransformGenerators.cs
```diff
@@ -336,7 +336,7 @@ public ImageLoading(PipelineNode node) : base(node)
         {
         }
 
-        internal override string MethodName => "LoadRawImageBytes";
```
LittleLittleCloud: this transformer's name should be something like ImageLoadingRawBytes

codemzs:
@LittleLittleCloud Please do not use WIP tag instead use DRAFT PR. Thanks!

PR #4500, opened 2019-11-23
Reenabling MacOS tests that had an issue with libgdiplus

antoniovs1029:
### Edited First Post:
PR #4492 disabled some tests that threw a `System.DllNotFoundException : Unable to load DLL 'libgdiplus'`, when trying to run them on our MacOS CI Pipeline.

As explained below in this thread, the problem was on homebrew's side, and this PR was meant to give a provitional solution for the problem. Nonetheless, this problem has just been fixed on homebrew's side, so now this PR simply reenables those tests.

-----------------------------

### Original First Post:
Solves the problem with MacOS Pipeline where certain tests threw a
`System.DllNotFoundException : Unable to load DLL 'libgdiplus': The specified module could not be found`

The issue was caused somewhere in homebrew's installation, and one way to solve this issue is to actually make homebrew to build the libgdiplus library after downloading it (instead of downloading a prebuilt version from [https://homebrew.bintray.com/bottles/mono-libgdiplus-6.0.4.high_sierra.bottle.tar.gz](https://homebrew.bintray.com/bottles/mono-libgdiplus-6.0.4.high_sierra.bottle.tar.gz) which is where it gets libgdiplus if homebrew doesn't build it). For some reason that's still unknown to me, this prebuilt version isn't working as expected, and the above exception is thrown.

### DIFF
--- a/build/ci/job-template.yml
+++ b/build/ci/job-template.yml
@@ -43,7 +43,7 @@ jobs:
 
     steps:
     - ${{ if eq(parameters.pool.name, 'Hosted macOS High Sierra') }}:
-      - script: brew update && brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/f5b1ac99a7fba27c19cee0bc4f036775c889b359/Formula/libomp.rb && brew install mono-libgdiplus gettext && brew link gettext --force && brew link libomp --force
+      - script: brew update && brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/f5b1ac99a7fba27c19cee0bc4f036775c889b359/Formula/libomp.rb && brew install --build-from-source mono-libgdiplus && brew install gettext && brew link gettext --force && brew link libomp --force
         displayName: Install build dependencies
     - ${{ if and( eq(parameters.nightlyBuild, 'true'), eq(parameters.name, 'Ubuntu_x64_NetCoreApp21')) }}:
       - bash: echo "##vso[task.setvariable variable=LD_LIBRARY_PATH]$(nightlyBuildRunPath):$LD_LIBRARY_PATH"
--- a/build/ci/job-template.yml
+++ b/build/ci/job-template.yml
@@ -43,7 +43,7 @@ jobs:
 
     steps:
     - ${{ if eq(parameters.pool.name, 'Hosted macOS High Sierra') }}:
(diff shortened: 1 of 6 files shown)


safern:
Cc @joshields as it seems pretty odd to me that libgdiplus is not found when installed from the latest stable package but it does when it’s built from source.

safern:
Cc @directhex as it seems pretty odd to me that libgdiplus is not found when installed from the latest stable package but it does when it’s built from source.

antoniovs1029:
> Cc @directhex as it seems pretty odd to me that libgdiplus is not found when installed from the latest stable package but it does when it’s built from source.

@safern It seems this problem is not only affecting us.

For instance, I created a simple Hello World app on a MacBook with High Sierra, where I made use of System.Drawing's Bitmap, and it threw the same DllNotFoundException if I installed mono-libgdiplus with

`brew install mono-libgdiplus`

... but it got fixed after I reinstalled mono-libgdiplus with
`brew install --build-from-source mono-libgdiplus`

In fact, it seems that other homewbrew's contributors have become aware that there's an issue with mono-libgdiplus bottle, as pointed [here ](https://github.com/Homebrew/homebrew-core/pull/46991#issuecomment-557939127) in this thread on homebrew's repo. And it _seems_ they are working on this, although I am not sure if it would be a good idea to open there an issue for this problem.

directhex:
We have ~no involvement with Mono on Homebrew

antoniovs1029:
As kindly explained by @Bo98 on [this comment](https://github.com/Homebrew/homebrew-core/pull/47141#issuecomment-558347373) on the homebrew-core repo, there were some problems with the libgdiplus prebuilt library that was downloaded when using `brew install mono-libgdiplus`. In that PR, Bo98 has fixed the problem by adding a new revision to the homebrew formula, so that now it downloads the new, fixed, prebuilt version of the library from:

`https://homebrew.bintray.com/bottles/mono-libgdiplus-6.0.4_1.high_sierra.bottle.tar.gz`

(notice the 6.0.4_1 which is used now, against the previous, problematic 6.0.4)

With those changes, it is no longer necessary to use the `--build-from-source` flag that I proposed as a solution for this issue.

That is, the problem was solved by Bo98 contributions on hombrew-core, and now this PR would simply reenable the tests that were failing on MacOS pipeline.

Bo98:
Apologies for the disruption everyone. We know Homebrew is important for many people's workflows.

Although the precise cause was a rather unusual one, I will try improve the tests of `mono-libgdiplus` to avoid this happening again.

antoniovs1029:
No worries, @Bo98 . Thank you very much for your contributions that solved this problem.

PR #4503, opened 2019-11-26
improve test stability

frank-dong-ms:
1. add datatime when start/finish tests
2. add timeout for thread waiting to prevent infinite wait
3. fix cancel token cancel twice issue

### DIFF
--- a/src/Microsoft.Extensions.ML/ModelReloadToken.cs
+++ b/src/Microsoft.Extensions.ML/ModelReloadToken.cs
@@ -44,6 +44,12 @@ namespace Microsoft.Extensions.ML
         /// <summary>
         /// Used to trigger the change token when a reload occurs.
         /// </summary>
-        public void OnReload() => _cts.Cancel();
+        public void OnReload()
+        {
+            if (!_cts.IsCancellationRequested)
+                _cts.Cancel();
+            else
+                Console.WriteLine("ModelReloadToken has already been Canceled.");
+        }
     }
 }
--- a/src/Microsoft.ML.Core/Utilities/OrderedWaiter.cs
+++ b/src/Microsoft.ML.Core/Utilities/OrderedWaiter.cs
@@ -90,7 +90,7 @@ namespace Microsoft.ML.Internal.Utilities
                 ev = new WaitStats(position);
                 _waiters.Add(ev);
             }
-            ev.Event.Wait(token);
+            ev.Event.Wait(5*60*1000, token);
             if (_ex != null)
                 throw Contracts.Except(_ex, "Event we were waiting on was subject to an exception");
         }
--- a/src/Microsoft.ML.Data/Data/DataViewUtils.cs
+++ b/src/Microsoft.ML.Data/Data/DataViewUtils.cs
@@ -402,7 +402,7 @@ namespace Microsoft.ML.Data
                                             // The waiter event should never be null since this is only
                                             // called after a point where waiter.Register has been called.
                                             ch.AssertValue(waiterEvent);
-                                        waiterEvent.Wait();
+                                        waiterEvent.Wait(5*60*1000);
                                         waiterEvent = null;
                                         toConsume.Add(batch);
                                     }
--- a/test/Microsoft.ML.TestFramework/BaseTestClass.cs
+++ b/test/Microsoft.ML.TestFramework/BaseTestClass.cs
@@ -49,14 +49,14 @@ namespace Microsoft.ML.TestFramework
             TestName = test.TestCase.TestMethod.Method.Name;
 
             // write to the console when a test starts and stops so we can identify any test hangs/deadlocks in CI
-            Console.WriteLine($"Starting test: {FullTestName}");
+            Console.WriteLine($"[{DateTime.Now}] Starting test: {FullTestName}");
             Initialize();
         }
 
         void IDisposable.Dispose()
         {
             Cleanup();
-            Console.WriteLine($"Finished test: {FullTestName}");
+            Console.WriteLine($"[{DateTime.Now}] Finished test: {FullTestName}");
         }
 
         protected virtual void Initialize()


src/Microsoft.ML.Core/Utilities/OrderedWaiter.cs
```diff
@@ -90,7 +90,7 @@ public void Wait(long position, CancellationToken token = default(CancellationTo
                 ev = new WaitStats(position);
                 _waiters.Add(ev);
             }
-            ev.Event.Wait(token);
+            ev.Event.Wait(5*60*1000, token);
             if (_ex != null)
```
harishsk: This is a five minute wait. Why five minutes? Can you please give more background behind this choice?

harishsk: Can this be a constant that will apply to all use cases? Or does this need to be a parameter to the function? If it is a constant, can you please give it a name and use that instead?

src/Microsoft.ML.Data/Data/DataViewUtils.cs
```diff
@@ -402,7 +402,7 @@ private static DataViewRowCursor ConsolidateCore(IChannelProvider provider, Data
                                             // The waiter event should never be null since this is only
                                             // called after a point where waiter.Register has been called.
                                             ch.AssertValue(waiterEvent);
-                                        waiterEvent.Wait();
+                                        waiterEvent.Wait(5*60*1000);
                                         waiterEvent = null;
```
harishsk: Similar comment as the one in OrderedWaiter.cs. If the same constant applies everywhere we should use a name.

src/Microsoft.Extensions.ML/ModelReloadToken.cs
```diff
@@ -44,6 +44,12 @@ public ModelReloadToken()
         /// <summary>
         /// Used to trigger the change token when a reload occurs.
         /// </summary>
-        public void OnReload() => _cts.Cancel();
+        public void OnReload()
+        {
+            if (!_cts.IsCancellationRequested)
+                _cts.Cancel();
```
harishsk: The code seems to be saying if cancellation is not requested, call Cancel(). Am I reading that correctly? Can you please describe the root cause and the fix a bit more thoroughly in the pull request?

src/Microsoft.Extensions.ML/ModelReloadToken.cs
```diff
@@ -44,6 +44,12 @@ public ModelReloadToken()
...
+            else
+                Console.WriteLine("ModelReloadToken has already been Canceled.");
+        }
```
harishsk: ML.NET is a library that can be used in environments where there is no console. I don't think we should be using Console.WriteLine here. 
What is the desired behavior? Is this for information only to the user or is this an error condition?

src/Microsoft.Extensions.ML/ModelReloadToken.cs
frank-dong-ms: This means if the cancellation has not already been requested, then cancel the token.
we should always add this check as calling cancel twice will crash the process. This is one of the reason that test process is crashing.

---
In reply to: [350532410](https://github.com/dotnet/machinelearning/pull/4503#discussion_r350532410) [](ancestors = 350532410)

src/Microsoft.Extensions.ML/ModelReloadToken.cs
frank-dong-ms: This is just to output some info into test command, this is not for user but rather a error info, I just output some info to see whether this is common test crash case.

---
In reply to: [350532971](https://github.com/dotnet/machinelearning/pull/4503#discussion_r350532971) [](ancestors = 350532971)

src/Microsoft.ML.Core/Utilities/OrderedWaiter.cs
frank-dong-ms: This caused most of our test hanging according to my local test result.
Some processor is waiting for signal to start but seems in some case it never can get the signal to run and the wait hangs the test process. This happens in multiple test cases so I just putting a pretty large timeout here, maybe we need discuss more to understand the root cause

---
In reply to: [350531489](https://github.com/dotnet/machinelearning/pull/4503#discussion_r350531489) [](ancestors = 350531489)

frank-dong-ms: Yes, this should, this number is not reliable now so just put the number directly, let's discuss more in person.

---
In reply to: [350531730](https://github.com/dotnet/machinelearning/pull/4503#discussion_r350531730) [](ancestors = 350531730)

src/Microsoft.Extensions.ML/ModelReloadToken.cs
eerhardt: > This is one of the reason that test process is crashing.

Can you provide a test log that shows this is where the test process is crashing? I don't see `.Cancel()` causing a process crash.

I made a console app that calls Cancel multiple times, and it doesn't even throw an exception.

src/Microsoft.Extensions.ML/ModelReloadToken.cs
eerhardt: We should not be writing to the console inside of our libraries.

src/Microsoft.Extensions.ML/ModelReloadToken.cs
frank-dong-ms: Below is the call stack I have reproduced in my local env:

The active test run was aborted. Reason: Test host process crashed : Unhandled Exception: System.AggregateException: One or more errors occurred. (Safe handle has been closed) ---> System.ObjectDisposedException: Safe handle has been closed
   at System.Runtime.InteropServices.SafeHandle.DangerousAddRef(Boolean& success)
   at System.StubHelpers.StubHelpers.SafeHandleAddRef(SafeHandle pHandle, Boolean& success)
   at Microsoft.Win32.Win32Native.SetEvent(SafeWaitHandle handle)
   at System.Threading.EventWaitHandle.Set()
   at Microsoft.Extensions.ML.FileLoaderTests.<>c__DisplayClass2_0.<can_reload_model>b__1() in C:\work\machinelearning\test\Microsoft.Extensions.ML.Tests\FileLoaderTests.cs:line 65
   at Microsoft.Extensions.Primitives.ChangeToken.<>c__DisplayClass0_0.<OnChange>b__0(Object s)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.CancellationTokenSource.ExecuteCallbackHandlers(Boolean throwOnFirstException)
   --- End of inner exception stack trace ---
   at System.Threading.CancellationTokenSource.ExecuteCallbackHandlers(Boolean throwOnFirstException)
   at Microsoft.Extensions.ML.ModelReloadToken.OnReload() in C:\work\machinelearning\src\Microsoft.Extensions.ML\ModelReloadToken.cs:line 47
   at Microsoft.Extensions.ML.FileModelLoader.WatcherChanged(Object sender, FileSystemEventArgs e) in C:\work\machinelearning\src\Microsoft.Extensions.ML\ModelLoaders\FileModelLoader.cs:line 78
   at System.IO.FileSystemWatcher.NotifyFileSystemEventArgs(WatcherChangeTypes changeType, String name)
   at System.IO.FileSystemWatcher.ParseEventBufferAndNotifyForEach(Byte[] buffer)
   at System.IO.FileSystemWatcher.ReadDirectoryChangesCallback(UInt32 errorCode, UInt32 numBytes, NativeOverlapped* overlappedPointer)
   at System.Threading._IOCompletionCallback.PerformIOCompletionCallback(UInt32 errorCode, UInt32 numBytes, NativeOverlapped* pOVERLAP)


---
In reply to: [350906884](https://github.com/dotnet/machinelearning/pull/4503#discussion_r350906884) [](ancestors = 350906884)

sharwell: > we should always add this check as calling cancel twice will crash the process. This is one of the reason that test process is crashing.

This is not correct. Cancellation a second time is a NOP, and `Cancel` itself is guarded by a check of `IsCancellationRequested`.

src/Microsoft.Extensions.ML/ModelReloadToken.cs
```diff
@@ -44,6 +44,12 @@ public ModelReloadToken()
         /// <summary>
         /// Used to trigger the change token when a reload occurs.
         /// </summary>
-        public void OnReload() => _cts.Cancel();
+        public void OnReload()
+        {
+            if (!_cts.IsCancellationRequested)
```
sharwell: 📝 The changes in this file should be reverted. It duplicates a cancellation check that appears here:

https://github.com/dotnet/runtime/blob/b2073bd41e7ba0606000e5950e3d960d1d2eb36d/src/libraries/System.Private.CoreLib/src/System/Threading/CancellationTokenSource.cs#L582

src/Microsoft.ML.Core/Utilities/OrderedWaiter.cs
```diff
@@ -90,7 +90,7 @@ public void Wait(long position, CancellationToken token = default(CancellationTo
                 ev = new WaitStats(position);
                 _waiters.Add(ev);
             }
-            ev.Event.Wait(token);
+            ev.Event.Wait(5*60*1000, token);
```
sharwell: 📝 This change should be reverted. This is production code, not test code. If you want to add a delay, update the tests to pass a `CancellationToken` that will cancel after a period of time.

src/Microsoft.ML.Data/Data/DataViewUtils.cs
```diff
@@ -402,7 +402,7 @@ private static DataViewRowCursor ConsolidateCore(IChannelProvider provider, Data
                                             // The waiter event should never be null since this is only
                                             // called after a point where waiter.Register has been called.
                                             ch.AssertValue(waiterEvent);
-                                        waiterEvent.Wait();
+                                        waiterEvent.Wait(5*60*1000);
```
sharwell: 📝 This change should be reverted for the same reason I mentioned for **OrderedWaiter.cs**.

frank-dong-ms:
Thanks all for information, close this PR now and working on new one

PR #4510, opened 2019-11-28
add fail retry for failed tests

frank-dong-ms:
Add MLNETFactAttribute and replace default Fact Attribute:
1. every test case will retry for 3 times at max by default if the test fails
2. log flaky tests to a sql db to better estimate which tests are fail often
3. add default timeout for every tests
4. refine logging


test/Microsoft.Extensions.ML.Tests/FileLoaderTests.cs
```diff
@@ -10,13 +10,14 @@
...
 using Microsoft.ML.Data;
+using Microsoft.ML.TestFrameworkCommon.Attributes;
 using Xunit;
 
 namespace Microsoft.Extensions.ML
 {
     public class FileLoaderTests
     {
-        [Fact]
+        [MLNETFact]
```
sharwell: IMO this should only be applied to tests that are known to be flaky.

test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -1442,7 +1443,7 @@ public void ScoreTransformerChainModel()
             Assert.NotNull(loadedData.Schema.GetColumnOrNull("Score"));
         }
 
-        //[Fact]
+        //[RetryFact]
```
sharwell: 📝 Stale comment.

💡 Tests should be disabled with a `Skip` attribute, not by commenting out the code.

test/Microsoft.ML.TestFrameworkCommon/Attributes/MLNETFactAttribute.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// ML.NET facts that will retry several flaky test cases, use default timeout settings
+    /// </summary>
+    [XunitTestCaseDiscoverer("Microsoft.ML.TestFrameworkCommon.MLNETFactDiscoverer", "Microsoft.ML.TestFrameworkCommon")]
+    public class MLNETFactAttribute : FactAttribute
+    {
+        /// <summary>
+        /// Number of retries allowed for a failed test. If unset (or set less than 1), will
+        /// default to 3 attempts.
```
sharwell: Two issues here:

1. If this attribute is used for all tests, it needs to default to 1
2. If this attribute is used only for tests known to be flaky, the maximum acceptable default is 2

test/Microsoft.ML.TestFrameworkCommon/Centrallizedlogger.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
+using System.Data.SqlClient;
+
+namespace Microsoft.ML.TestFrameworkCommon
+{
+    public static class Centrallizedlogger
```
sharwell: ❔ Is this used by local developer builds or only on CI?

test/Microsoft.ML.Functional.Tests/FunctionalTestBaseClass.cs
```diff
@@ -50,15 +50,12 @@ public FunctionalTestBaseClass(ITestOutputHelper output)
             FullTestName = test.TestCase.TestMethod.TestClass.Class.Name + "." + test.TestCase.TestMethod.Method.Name;
             TestName = test.TestCase.TestMethod.Method.Name;
 
-            // write to the console when a test starts and stops so we can identify any test hangs/deadlocks in CI
-            Console.WriteLine($"Starting test: {FullTestName}");
```
sharwell: ❔ Are these getting logged anywhere else?

test/Microsoft.Extensions.ML.Tests/FileLoaderTests.cs
eerhardt: +💯

harishsk: I agree. And I think that was the second proposal. To apply this only to tests known to be flaky and if we still see failures, to retry the whole configuration. 
Can you please amend the pull request as per your second proposal?

safern: I agree. Actually in the dotnet/runtime repo we kept it simple instead of adding a `FactAttribute` a `RetryHelper` class which forces people to use that when a test is flaky and it has to be justified. 

Also, that way you can also use it for theory tests not only facts.

https://github.com/dotnet/runtime/blob/master/src/libraries/Common/tests/CoreFx.Private.TestUtilities/System/RetryHelper.cs#L11

safern: Also, if we stick with an attribute I believe the name should express what the attribute does.

For example: `RetryFactAttribute` or `RetryAbleFactAttribute`.

test/Microsoft.ML.Functional.Tests/FunctionalTestBaseClass.cs
frank-dong-ms: Yes, logged in MLNETTestCase class, not all test class inherinet from this base test class but we can change all tests to use MLNETFact to logging start and end of test

---
In reply to: [354511751](https://github.com/dotnet/machinelearning/pull/4510#discussion_r354511751) [](ancestors = 354511751)

test/Microsoft.ML.TestFrameworkCommon/Centrallizedlogger.cs
frank-dong-ms: both

---
In reply to: [354509412](https://github.com/dotnet/machinelearning/pull/4510#discussion_r354509412) [](ancestors = 354509412)

test/Microsoft.Extensions.ML.Tests/FileLoaderTests.cs
```diff
@@ -10,13 +10,14 @@
...
 using Microsoft.Extensions.Primitives;
 using Microsoft.ML.Data;
+using Microsoft.ML.TestFrameworkCommon.Attributes;
 using Xunit;
 
 namespace Microsoft.Extensions.ML
 {
     public class FileLoaderTests
     {
-        [Fact]
```
frank-dong-ms: This Fact can do more than just retry failed tests, example: better place to log start/end test, set default timeout for every test, exception handling etc...
we can filter tests known to be flaky and only retry them in MLNETTestCase Class but I found some test case are randomly failing no obvious pattern except some few tests are failing a lot. So what I do here is retry every failed tests and log all fails into a db even the test success after retry. This can be more reliable way to tell which tests are flaky.

test/Microsoft.ML.TestFrameworkCommon/Attributes/MLNETFactAttribute.cs
frank-dong-ms: We intend to retry all failed test cases and log all the failing for investigate. All retry is to unblock every one from passing their PR build for now as failing is random and frequent and investigating flaky test can be time consuming. 

---
In reply to: [354508994](https://github.com/dotnet/machinelearning/pull/4510#discussion_r354508994) [](ancestors = 354508994)

PR #4512, opened 2019-11-29
Fixes #4505 Remove reliance on getting product version for model.zip/version.txt from FileVersionInfo and replace with using assembly custom attributes

### LINKED ISSUE #4505: Version.txt should use file version from assembly custom attributes not FileVersionInfo
The change made in issue https://github.com/dotnet/machinelearning/issues/3132 have introduced a bug due to relying on a physical location for the assembly. If the Microsoft.ML.Core assembly is loaded from memory the assembly.location will be empty. Using **FileVersionInfo.GetVersionInfo** relies on a physical path - so an argument exception is thrown inside System.IO.Path since the supplied path is empty.

https://github.com/dotnet/machinelearning/blob/37ed3361f330ed439c9bd477abcfe5a27e06c0a1/src/Microsoft.ML.Core/Data/Repository.cs#L308-L311

Instead of using FileVersionInfo.GetVersionInfo the assembly custom attributes should be used, for example:
`var productVersion = assembly.CustomAttributes.FirstOrDefault(a => a.AttributeType == typeof(AssemblyFileVersionAttribute)).ConstructorArguments.First();`

This will return the same product string version as the FileVersionInfo.GetVersionInfo does.

r0ss88:
Refer to original issue for further details of the reasons for this change.

I have used AssemblyInformationalVersionAttribute to match the same behaviour as was introduced with the change to reliance of FileVersionInfo.GetVersionInfo. There is already a test covering -ModelFiles.DetermineNugetVersionFromModel in Microsoft.ML.FunctionalTests.

### DIFF
--- a/src/Microsoft.ML.Core/Data/Repository.cs
+++ b/src/Microsoft.ML.Core/Data/Repository.cs
@@ -7,6 +7,8 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
+using System.Linq;
+using System.Reflection;
 using Microsoft.ML.Internal.Utilities;
 using Microsoft.ML.Runtime;
 
@@ -305,10 +307,10 @@ namespace Microsoft.ML
             Contracts.CheckValueOrNull(ectx);
             ectx.CheckValue(stream, nameof(stream));
             var rep = new RepositoryWriter(stream, ectx, useFileSystem);
-            var versionInfo = FileVersionInfo.GetVersionInfo(typeof(RepositoryWriter).Assembly.Location);
+
             using (var ent = rep.CreateEntry(DirTrainingInfo, "Version.txt"))
             using (var writer = Utils.OpenWriter(ent.Stream))
-                writer.WriteLine(versionInfo.ProductVersion);
+                writer.WriteLine(GetProductVersion());
             return rep;
         }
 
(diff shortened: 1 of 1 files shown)


r0ss88:
Any chance this could be PR'd? the change is pretty minor.

src/Microsoft.ML.Core/Data/Repository.cs
```diff
@@ -432,6 +434,24 @@ public void Commit()
             Flush();
             Dispose(true);
         }
+
+        private static string GetProductVersion()
+        {
+            var assembly = typeof(RepositoryWriter).Assembly;
+
+            var assemblyInternationalVersionAttribute = assembly.CustomAttributes.FirstOrDefault(a =>
```
yaeldekel: >assemblyInternationalVersionAttribute [](start = 16, length = 37)

Nit: Can this be renamed `assemblyInformationalVersionAttribute`?

src/Microsoft.ML.Core/Data/Repository.cs
```diff
@@ -432,6 +434,24 @@ public void Commit()
...
+                    a.AttributeType == typeof(AssemblyInformationalVersionAttribute));
+
+            if (assemblyInternationalVersionAttribute == null)
+            {
+                throw new ApplicationException($"Cannot determine product version from assembly {assembly.FullName}.");
```
yaeldekel: >throw [](start = 16, length = 5)

Not sure we should throw in this case. Perhaps we should save `0.0.0.0` in this case? Or, alternatively, not include `Version.txt` in the model?

@eerhardt, @harishsk

r0ss88: This was only to keep consistency with how it was behaving before - I prefer the approach you have suggested though it should not fail because of this.

src/Microsoft.ML.Core/Data/Repository.cs
r0ss88: OK sure

src/Microsoft.ML.Core/Data/Repository.cs
r0ss88: Please confirm @eerhardt @harishsk @yaeldekel  and I'll make the change.

michaelgsharp:
Hey @r0ss88, is still still something you would like to see merged in? You mention to refer to the original issue, but I don't see one linked. Which issue is it?

michaelgsharp:
Nevermind, I see it in the title not the description. Let me take a look at it real fast.

PR #4514, opened 2019-12-02
Learning with counts (Dracula) transformer

### LINKED ISSUE #4016: Add a "learning with counts" transformer
This transformation is a simple way to handle categorical features, and is explained here: https://blogs.technet.microsoft.com/machinelearning/2015/02/17/big-learning-made-easy-with-counts/

yaeldekel:
Fixes #4016.


src/Microsoft.ML.Transforms/HashJoiningTransform.cs
```diff
@@ -188,9 +187,9 @@ private static VersionInfo GetVersionInfo()
             IDataView input,
             string name,
             string source = null,
-             bool join = Defaults.Join,
+             bool join = Defaults.Combine,
```
eerhardt: `bool join` => `bool combine`. And update the xml docs

src/Microsoft.ML.Transforms/Dracula/DraculaTransform.cs
```diff
@@ -0,0 +1,518 @@
...
+            inputColumnName = string.IsNullOrEmpty(inputColumnName) ? outputColumnName : inputColumnName;
+            builder = builder ?? new CMCountTableBuilder();
+
+            return new DraculaEstimator(env, labelColumn,
+                new[] { new CountTableEstimator.ColumnOptions(outputColumnName, inputColumnName, builder, priorCoefficient, laplaceScale) },
+                countsPath, numberOfBits, combine, hashingSeed);
+        }
+    }
+
+    public static class Dracula
```
eerhardt: No need for this to be public, it only has an internal method.

src/Microsoft.ML.Transforms/Dracula/CountTableTransformer.cs
```diff
@@ -0,0 +1,711 @@
...
+                        }
+                    }
+
+                    dst = editor.Commit();
+                };
+            }
+        }
+    }
+
+    public static class CountTable
```
eerhardt: should be internal

src/Microsoft.ML.Transforms/Dracula/CountTable.cs
```diff
@@ -0,0 +1,150 @@
...
+        IReadOnlyCollection<float> GarbageCounts { get; }
+        ReadOnlySpan<double> PriorFrequencies { get; }
+
+        int AppendRows(List<int> hashIds, List<ulong> hashValues, List<float[]> counts);
+    }
+
+    /// <summary>
+    /// Signature for CountTableBuilder.
+    /// </summary>
+    public delegate void SignatureCountTableBuilder();
```
eerhardt: Does this need to be public?

codemzs:
@yaeldekel Shall we give it a better name as opposed to calling it "Dracula"? Seems a little informal to me.

src/Microsoft.ML.Transforms/Dracula/CMCountTable.cs
```diff
@@ -0,0 +1,340 @@
...
+namespace Microsoft.ML.Transforms
+{
+    internal sealed class CMCountTable : CountTableBase
+    {
+        public const string LoaderSignature = "CMCountTable";
+
+        private static VersionInfo GetVersionInfo()
+        {
+            return new VersionInfo(
+                modelSignature: "CM    CT",
```
codemzs: >"CM    CT", [](start = 32, length = 11)

what kind of model signature is this? Why is there a tab in between?

src/Microsoft.ML.Transforms/Dracula/DraculaTransform.cs
```diff
@@ -0,0 +1,518 @@
...
+    DraculaTransformer.UserName, DraculaTransformer.LoaderSignature, "Dracula")]
+
+[assembly: LoadableClass(DraculaTransformer.Summary, typeof(DraculaTransformer), typeof(DraculaEstimator), null, typeof(SignatureLoadModel),
+    DraculaTransformer.UserName, DraculaTransformer.LoaderSignature)]
+
+[assembly: EntryPointModule(typeof(Dracula))]
+
+namespace Microsoft.ML.Transforms
+{
+    public class DraculaEstimator : IEstimator<DraculaTransformer>
```
justinormont: Moving @codemzs' [comment](https://github.com/dotnet/machinelearning/pull/4514#issuecomment-560502623) to a thread:

> @yaeldekel Shall we give it a better name as opposed to calling it "Dracula"? Seems a little informal to me.

justinormont: If `Dracula` is not used, `CountTargetEncoder` might be suitable.

Our old blog post lists DRACULA as "**D**istributed **R**obust **A**lgorithm for **C**ount-based **L**e**a**rning": https://blogs.technet.microsoft.com/machinelearning/2015/02/17/big-learning-made-easy-with-counts/

As an aside, there's also public slides: 
https://www.slideshare.net/SessionsEvents/misha-bilenko-principal-researcher-microsoft

yaeldekel: Good suggestion. Changing the name to `CountTargetEncoder`, unless someone has a different idea?

sharwell:
📝 The OutOfMemoryException test failures appear to be true failures

src/Microsoft.ML.Transforms/Dracula/CountTableTransformer.cs
```diff
@@ -0,0 +1,734 @@
...
+    CountTableTransformer.UserName, CountTableTransformer.LoaderSignature)]
+
+[assembly: LoadableClass(typeof(IRowMapper), typeof(CountTableTransformer), null, typeof(SignatureLoadRowMapper),
+   CountTableTransformer.UserName, CountTableTransformer.LoaderSignature)]
+
+[assembly: EntryPointModule(typeof(CountTable))]
+
+namespace Microsoft.ML.Transforms
+{
+    public sealed class CountTableEstimator : IEstimator<CountTableTransformer>
```
eerhardt: Does this class need to be `public`? How can an external user use this class?

eerhardt: Also, xml comments on all public surface area.

---
In reply to: [356307263](https://github.com/dotnet/machinelearning/pull/4514#discussion_r356307263) [](ancestors = 356307263)

src/Microsoft.ML.Transforms/HashJoiningTransform.cs
eerhardt: Actually, I am rethinking this comment. What's weird to me is that this is the "hash joining transform", and we are calling this behavior "combine". That seems inconsistent. We should use the same term to mean the same thing. Either always use "join" or always use "combine".

src/Microsoft.ML.Transforms/Dracula/Featurizer.cs
```diff
@@ -0,0 +1,224 @@
...
+            GenerateLogOdds(iCol, countTable, countsBuffer, features.Slice(_labelBinCount, _logOddsCount), sum);
+            AssertValidOutput(features);
+
+            // Add the last feature: an indicator for isGarbage.
+            features[NumFeatures - 1] = isGarbage ? 1 : 0;
+        }
+
+        // Assumes features are filled with the respective counts.
+        // Adds laplacian noise if set and returns the sum of the counts.
+        private float AddLaplacianNoisePerLabel(int iCol, Random rand, Span<float> counts)
```
justinormont: Is the noise still added once the ML model is fully created?

@davidbrownellWork was mentioning that there is added noise at the prediction time. This complicates the ONNX conversion. The AutoML team is looking to use Dracula for its target encoding and we need it in ML.NET, NimbusML, and all the way ONNX.

src/Microsoft.ML.Transforms/HashJoiningTransform.cs
yaeldekel: I was actually thinking that the name `HashJoiningTransform` is not very informative. What do you think about `VectorHashingTransform`?

---
In reply to: [356308668](https://github.com/dotnet/machinelearning/pull/4514#discussion_r356308668) [](ancestors = 356308668)

src/Microsoft.ML.Transforms/Dracula/Featurizer.cs
yaeldekel: The noise is only used at prediction time. From your experience using this transform, do you think it is needed?

---
In reply to: [356322013](https://github.com/dotnet/machinelearning/pull/4514#discussion_r356322013) [](ancestors = 356322013)

src/Microsoft.ML.Transforms/Dracula/DraculaTransform.cs
```diff
@@ -0,0 +1,553 @@
...
+                }
+            }
+
+            var inputData = new EmptyDataView(_host, inputSchemaBuilder.ToSchema());
+            var hashJoinSchema = CreateHashJoiningTransform(inputData).OutputSchema;
+            return SchemaShape.Create(hashJoinSchema);
+        }
+    }
+
+    public sealed class CountTargetEncodingTransformer : ITransformer
```
codemzs: >CountTargetEncodingTransformer [](start = 24, length = 30)

documentation?

src/Microsoft.ML.Transforms/Dracula/DraculaTransform.cs
```diff
@@ -0,0 +1,553 @@
...
+    CountTargetEncodingTransformer.UserName, CountTargetEncodingTransformer.LoaderSignature, "Dracula")]
+
+[assembly: LoadableClass(CountTargetEncodingTransformer.Summary, typeof(CountTargetEncodingTransformer), typeof(CountTargetEncodingEstimator), null, typeof(SignatureLoadModel),
+    CountTargetEncodingTransformer.UserName, CountTargetEncodingTransformer.LoaderSignature)]
+
+[assembly: EntryPointModule(typeof(CountTargetEncoder))]
+
+namespace Microsoft.ML.Transforms
+{
+    public class CountTargetEncodingEstimator : IEstimator<CountTargetEncodingTransformer>
```
codemzs: >CountTargetEncodingEstimator [](start = 17, length = 28)

Documentation?

src/Microsoft.ML.Transforms/Dracula/DraculaTransform.cs
```diff
@@ -0,0 +1,553 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
codemzs: Please rename the final to more relevant name. If we are calling it CountTragetEncoding then we should not be seeing Dracula in the codebase.

src/Microsoft.ML.Transforms/Dracula/Featurizer.cs
yaeldekel: I looked at the EstimatorChain class, and it seems to me that it would be possible to add a method similar to AppendCacheCheckPointthat would add noise to a requested column. Here is the code that is currently in the Fit method of EstimatorChain:
```
            IDataView current = input;
            var xfs = new ITransformer[_estimators.Length];
            for (int i = 0; i < _estimators.Length; i++)
            {
                var est = _estimators[i];
                xfs[i] = est.Fit(current);
                current = xfs[i].Transform(current);
                if (_needCacheAfter[i] && i < _estimators.Length - 1)
                {
                    Contracts.AssertValue(_host);
                    current = new CacheDataView(_host, current, null);
                }
            }
```

If we add, similarly to how we have `current = new CacheDataView(...)`, something like `current = new NoiseDataView(...)`, then we can let the user writes something like
```
var pipeline = mlContext.Transforms.CountTargetEncoding()
    .AppendNoise("Features").Append(mlContext.Binary.Sdca());
```

then the SDCA component in the pipeline will see the `Features` column with noise added to it when training, however, the resulting `TransformerChain` will not have noise as part of the model.
 
Together with good samples and documentation emphasizing the recommended way to use the `CountTargetEncodingEstimator` (which is either to train the transformer in a separate pipeline than the trainer and on different data, or use the AppendNoise API), I think that may be a better solution to having a noise option as part of the estimator constructor. 

Please let me know what you think of this solution and whether I should go ahead and implement that.


---
In reply to: [357085107](https://github.com/dotnet/machinelearning/pull/4514#discussion_r357085107) [](ancestors = 357085107,356322013)

gvashishtha: Hi Yael, what is the name of this transform in TLC? I went to our TLC telemetry to look for internal users of the dracula transform but I did not see anyone.

antoniovs1029: In the TLC GUI it's called "Dracula Transform" (under "Featurizers: Supervised") and in the TLC code it is called `DraculaTransform`… but I wouldn't know if other names are used in the telemetry.

justinormont: If I recall, the log-odds are returned _approximately_ as:
```c#
CountPositve = CountPositive + LaplacianNoise(); // Add noise to pos count
CountNegative = CountNegative + LaplacianNoise();  // Add noise to neg count
logOdds = ln( (CountPositive + PriorCount) / (CountPositive + CountNegative + 2*PriorCount) );  // Includes 2x noise, plus prior as smoothing
```

I don't think we'll be able to just add noise outside of the Dracula transform and end up with the same result.

My recommendation is keeping the noise computation as is, and disable noise for the ONNX unit tests, or add a window  (e.g. +/- 10%)  of acceptable values when checking for (near) equality of the outputs.

yaeldekel: The two possible solutions as I see it are:
* Leaving the computation as it is right now.
  * Pros: easy solution for fitting a pipeline that has a trainer after the Dracula transform without having to split into two pipelines and training each part on a different dataset.
  * Cons: if the ONNX export of this transform will not include the noise part (which it likely won't), then this will result in a model that gives different results in the ML.NET format and in the ONNX format.
* Making the transformer apply the noise, but not serialize the noise parameter, i.e. after deserialization it won't add noise.
  * Pros: consistency with the ONNX version of the model. Also, the noise is really not needed at prediction time, it is only needed to train the next component in the pipeline.
  * Cons: if you call `Transform` on the pipeline with the same data right after it is trained, and after serializing and deserializing the model, it will give different results.

I don't particularly like either of these solutions, I think they are both problematic, but currently I can't think of a solution that will not have any of the cons listed above...  Which solution do you think I should go with?

cc @justinormont, @eerhardt


---
In reply to: [365470116](https://github.com/dotnet/machinelearning/pull/4514#discussion_r365470116) [](ancestors = 365470116)

davidbrownellWork: The introduction of noise is a transformation step that is only used when the featurizer is included in a pipeline that is used during training time. Because ONNX exports are only used for inference, it seems OK that the export doesn't include noise, as the usage in ONNX will be the same as when this featurizer is used in an inferencing ML.Net pipeline.

It seems that the first point is the solution, as long as the code is written such that noise is not introduced when the featurizer is part of an inferencing pipeline.

harishsk:
>     {

This class doesn't follow the typical pattern of using a Mapper. Any particular reason why?


---
Refers to: src/Microsoft.ML.Transforms/HashJoiningTransform.cs:35 in c3df0e0. [](commit_id = c3df0e0a5808f1d12e022b1d537b87ad2dd6a89a, deletion_comment = False)

harishsk:
>     {

How is this different from HashingEstimator? 

---
Refers to: src/Microsoft.ML.Transforms/HashJoiningTransform.cs:35 in c3df0e0. [](commit_id = c3df0e0a5808f1d12e022b1d537b87ad2dd6a89a, deletion_comment = False)

harishsk:
>                 };

Internally this too uses the MurmurHash which is used by HashingTransformer. Is it possible to rationalize this transform the HashingTransformer?

---
Refers to: src/Microsoft.ML.Transforms/HashJoiningTransform.cs:647 in c3df0e0. [](commit_id = c3df0e0a5808f1d12e022b1d537b87ad2dd6a89a, deletion_comment = False)

harishsk:
>                     return Hashing.MurmurHash(seed, sb, 0, sb.Length);

We have had a number of backward compatibility issues to address with MurmurHash when trying to export to onnx. Can we defer making this public until we have implemented the onnx export functionality for this transformer?

---
Refers to: src/Microsoft.ML.Transforms/HashJoiningTransform.cs:646 in c3df0e0. [](commit_id = c3df0e0a5808f1d12e022b1d537b87ad2dd6a89a, deletion_comment = False)

yaeldekel:
>     {

This class was never converted to `IEstimator`/`ITransformer` because it was never a priority. It is currently only used in the entry points for train-test and cv split. The cv command and the public train-test and cv APIs use all use the `HashingEstimator`.
There are a couple of differences between `HashJoiningTransform` and `HashingEstimator`:
* `HashJoiningTransform` hashes floats/doubles using murmur hash, and the rest of the types by converting the value to string, and then hashing. `HashingEstimator` has a hash function the simple data view types: floats, doubles, all the integer types, all the unsigned integer types, all the key types, bool and text. I think that it was implemented that way for simplicity, not because of some kind of advantage of hashing strings over other types.
* `HashJoiningTransform` has an option to hash a vector of values into a single hash value, as opposed to the `HashingEstimator`, that always handles vectors by hashing each slot separately and producing a vector of hashes. This is an advantage that is useful for cv and train-test splits, since these splits apply a `RangeFilter` on top of the output of `HashingTransformer`, which would cause an exception if given a vector column.
* `HashJoiningTransform` has an option to specify subsets of the vector input that should be hashed together. For example, the user can specity that they wish to hash slots 1,3,5 into a single hash value, and slots 0,2,4 into a single hash value (this option is where the name of the transform came from :-)). I think that this capability was added specifically for the dracula transform, although I don't know how useful it is (@justinormont, can you say anything about that?).


---
In reply to: [582084074](https://github.com/dotnet/machinelearning/pull/4514#issuecomment-582084074) [](ancestors = 582084074)

---
Refers to: src/Microsoft.ML.Transforms/HashJoiningTransform.cs:35 in c3df0e0. [](commit_id = c3df0e0a5808f1d12e022b1d537b87ad2dd6a89a, deletion_comment = False)

yaeldekel:
>                 };

I think that if we add the option to hash a vector into a single hash to `HashingTransformer` then we can switch the `CountTargetEncodingTransformer` to use it instead of `HashJoiningTransform`, but I would like to wait for @justinormont to comment about the importance of the `Combine` option first.

---
In reply to: [582085386](https://github.com/dotnet/machinelearning/pull/4514#issuecomment-582085386) [](ancestors = 582085386)

---
Refers to: src/Microsoft.ML.Transforms/HashJoiningTransform.cs:647 in c3df0e0. [](commit_id = c3df0e0a5808f1d12e022b1d537b87ad2dd6a89a, deletion_comment = False)

yaeldekel:
>                     return Hashing.MurmurHash(seed, sb, 0, sb.Length);

Is there a plan to switch `HashingTransformer` to use something other than Murmur hash?

---
In reply to: [582086748](https://github.com/dotnet/machinelearning/pull/4514#issuecomment-582086748) [](ancestors = 582086748)

---
Refers to: src/Microsoft.ML.Transforms/HashJoiningTransform.cs:646 in c3df0e0. [](commit_id = c3df0e0a5808f1d12e022b1d537b87ad2dd6a89a, deletion_comment = False)

harishsk:
>                     return Hashing.MurmurHash(seed, sb, 0, sb.Length);

No, there is no plan to switch to something other MurmurHash. But we have had a number of discussions on how to keep compatibility between the version of MurmurHash in ML.NET and the one in Onnx. We have had to increment the version of the HashingEstimator to make them compatible. 

Since the HashJoiningTransform also uses MurmurHash we should discuss its Onnx export mechanism and implement it before making this public so that if any changes are required in the model versions or on the onnx side, we make it only once. 

---
In reply to: [582428008](https://github.com/dotnet/machinelearning/pull/4514#issuecomment-582428008) [](ancestors = 582428008,582086748)

---
Refers to: src/Microsoft.ML.Transforms/HashJoiningTransform.cs:646 in c3df0e0. [](commit_id = c3df0e0a5808f1d12e022b1d537b87ad2dd6a89a, deletion_comment = False)

harishsk:
>                 };

It would be great if the CountTargetEncodingTransformer also used the HashingTransformer.

---
In reply to: [582427616](https://github.com/dotnet/machinelearning/pull/4514#issuecomment-582427616) [](ancestors = 582427616,582085386)

---
Refers to: src/Microsoft.ML.Transforms/HashJoiningTransform.cs:647 in c3df0e0. [](commit_id = c3df0e0a5808f1d12e022b1d537b87ad2dd6a89a, deletion_comment = False)

src/Microsoft.ML.Transforms/HashJoiningTransform.cs
harishsk: VectoHashingTransform implies each element is going to be hashed separately (which is the default behavior). With respect to @eerhardt's comment that we should user either Join or Combine consistently, a better name might be HashCombiningTransform.

---
In reply to: [356607566](https://github.com/dotnet/machinelearning/pull/4514#discussion_r356607566) [](ancestors = 356607566,356308668)

test/Microsoft.ML.Tests/Transformers/DraculaTests.cs
```diff
@@ -0,0 +1,176 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
```
harishsk, line 5: Can you please rename this file to not call it Dracula?

src/Microsoft.ML.Transforms/HashJoiningTransform.cs
yaeldekel: I have reverted the changes in this file, since it is no longer being used in this PR. This transform is now not being used directly anywhere in the code, and just being kept around for backward compatibility of maml, and for its entry point.

---
In reply to: [427561266](https://github.com/dotnet/machinelearning/pull/4514#discussion_r427561266) [](ancestors = 427561266,356607566,356308668)

PR #4515, opened 2019-12-03
Adds PriorTrainer Onnx conversion

Lynx1820:
1. Adding Onnx support for PriorTrainer 
       To follow the onnx construction pattern of other binary classifiers, this trainer uses the probability to predict labels, instead of score

### DIFF
--- a/src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
+++ b/src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
@@ -197,16 +197,9 @@ namespace Microsoft.ML.Data
         public string FeatureColumnName { get; }
 
         /// <summary>
-        /// The name of the label column used by the prediction transformer.
-        /// </summary>
-        public string LabelColumnName { get; }
-
-        /// <summary>
         /// The type of the prediction transformer
         /// </summary>
         public DataViewType FeatureColumnType { get; }
-
-        public DataViewType LabelColumnType { get; }
 
         /// <summary>
         /// Initializes a new reference of <see cref="SingleFeaturePredictionTransformerBase{TModel}"/>.
@@ -225,20 +218,6 @@ namespace Microsoft.ML.Data
                 throw Host.ExceptSchemaMismatch(nameof(featureColumn), "feature", featureColumn);
             else
                 FeatureColumnType = trainSchema[col].Type;
-
-            BindableMapper = ScoreUtils.GetSchemaBindableMapper(Host, ModelAsPredictor);
-        }
-
-        private protected SingleFeaturePredictionTransformerBase(IHost host, TModel model, DataViewSchema trainSchema, string featureColumn, string labelColumn)
-            : base(host, model, trainSchema)
-        {
-            LabelColumnName = labelColumn;
-            if (labelColumn == null)
-                LabelColumnType = null;
-            else if (!trainSchema.TryGetColumnIndex(labelColumn, out int col))
-                throw Host.ExceptSchemaMismatch(nameof(labelColumn), "label", labelColumn);
-            else
-                LabelColumnType = trainSchema[col].Type;
 
             BindableMapper = ScoreUtils.GetSchemaBindableMapper(Host, ModelAsPredictor);
         }
@@ -391,6 +370,7 @@ namespace Microsoft.ML.Data
     {
         internal readonly string ThresholdColumn;
         internal readonly float Threshold;
+        internal readonly string LabelColumnName;
 
         [BestFriend]
         internal BinaryPredictionTransformer(IHostEnvironment env, TModel model, DataViewSchema inputSchema, string featureColumn,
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.Data/Scorers/SchemaBindablePredictorWrapper.cs
```diff
@@ -426,9 +426,15 @@ private protected override bool SaveAsOnnxCore(OnnxContext ctx, RoleMappedSchema
 
             var mapper = ValueMapper as ISingleCanSaveOnnx;
             Contracts.CheckValue(mapper, nameof(mapper));
-            Contracts.Assert(schema.Feature.HasValue);
             Contracts.Assert(Utils.Size(outputNames) == 3); // Predicted Label, Score and Probablity.
 
+            // Prior doesn't have a feature column and uses the training label column to determine predicted labels
+            if (mapper.ToString().EndsWith("PriorModelParameters")) {
```
harishsk: I know I had asked you to explore the option of checking the mapper type here but I didn't mean check it as a string. I meant check the type directly (using typeof()). But now I don't think that is good idea either because we might introduce a bad dependency. 
I think the better approach is to explicitly handle two cases (if the feature column is present and otherwise).

src/Microsoft.ML.Data/Scorers/SchemaBindablePredictorWrapper.cs
```diff
@@ -426,9 +426,15 @@ private protected override bool SaveAsOnnxCore(OnnxContext ctx, RoleMappedSchema
...
+                var labelColumnName = schema.Schema[0].Name;
+                return mapper.SaveAsOnnx(ctx, outputNames, ctx.GetVariableName(labelColumnName));
```
harishsk: Can you use the same pattern as schema.Feature.HasValue? (schema.Label.HasValue)?

Lynx1820: I can check whether the name is null, but the schema itself doesn't have the label name initialized. BinaryPredictionTransformer only populates the schema's feature column.

src/Microsoft.ML.Data/Scorers/PredictionTransformer.cs
```diff
@@ -382,7 +383,17 @@ public sealed class BinaryPredictionTransformer<TModel> : SingleFeaturePredictio
 
             SetScorer();
         }
+        internal BinaryPredictionTransformer(IHostEnvironment env, TModel model, DataViewSchema inputSchema, string featureColumn, string labelColumn,
```
harishsk: Nit: Please add a separating line between functions. (before and after)

src/Microsoft.ML.Data/Scorers/SchemaBindablePredictorWrapper.cs
```diff
@@ -426,9 +426,16 @@ private protected override bool SaveAsOnnxCore(OnnxContext ctx, RoleMappedSchema
...
             Contracts.Assert(Utils.Size(outputNames) == 3); // Predicted Label, Score and Probablity.
 
+            // Prior doesn't have a feature column and uses the training label column to determine predicted labels
+            if (!schema.Feature.HasValue) {
+                Contracts.Assert(schema.Label.HasValue);
+                var labelColumnName = schema.Label.Value.Name;
+                return mapper.SaveAsOnnx(ctx, outputNames, ctx.GetVariableName(labelColumnName));
+            }
+
+            Contracts.Assert(schema.Feature.HasValue);
```
harishsk: Strictly speaking, the assert is unnecessary here because the if block has a return embedded in it.

src/Microsoft.ML.Data/Scorers/SchemaBindablePredictorWrapper.cs
```diff
@@ -426,9 +426,16 @@ private protected override bool SaveAsOnnxCore(OnnxContext ctx, RoleMappedSchema
 
             var mapper = ValueMapper as ISingleCanSaveOnnx;
             Contracts.CheckValue(mapper, nameof(mapper));
-            Contracts.Assert(schema.Feature.HasValue);
             Contracts.Assert(Utils.Size(outputNames) == 3); // Predicted Label, Score and Probablity.
 
+            // Prior doesn't have a feature column and uses the training label column to determine predicted labels
+            if (!schema.Feature.HasValue) {
+                Contracts.Assert(schema.Label.HasValue);
```
harishsk: The convention in dotnet is to have to { on the following line. Not the same line. Please fix.

src/Microsoft.ML.StandardTrainers/Standard/Simple/SimpleTrainers.cs
```diff
@@ -396,6 +398,37 @@ private protected override void SaveCore(ModelSaveContext ctx)
             Contracts.Assert(!float.IsNaN(_prob));
             ctx.Writer.Write(_prob);
         }
+        bool ISingleCanSaveOnnx.SaveAsOnnx(OnnxContext ctx, string[] outputs, string labelColumn)
+        {
```
harishsk: Nit: Newline before function

Issue #906, opened 2018-09-13
TensorFlowTransform does not implement IDisposable.


abgoswam:
TensorFlowTransform does not implement IDisposable. 

Instead the Finalizer directly calls the private Dispose() method to free resources. We need to understand the reasoning behind ML.NET Transforms not implementing IDisposable

[Discussion](https://github.com/dotnet/machinelearning/pull/853#issuecomment-423236224)

codemzs:
I don't see this as an issues as the same applies for every other transform. Please add more details to why you believe this is an issue and we can reopen.

PR #4521, opened 2019-12-04
DateTimeTransformer featurizer

michaelgsharp:
This change adds in the DateTimeTransformer into the new Featurizers project. It is the first of a series of PR's that will go in. The DateTimeTransformer is implemented in native code, so this is mostly just a wrapper around that with the appropriate entrypoints for NimbusML as well. Since all the new estimator/transformers follow the same patterns, once this one is reviewed and checked in I will create the other PR's.


src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,782 @@
...
+    ///
+    /// The <xref:Microsoft.ML.Transforms.CategoryImputerEstimator> is a trivial estimator and does not need training.
+    ///
+    ///
+    /// ]]>
+    /// </format>
+    /// </remarks>
+    /// <seealso cref="DateTimeTransformerExtensionClass.DateTimeTransformer(TransformsCatalog, string, string, DateTimeTransformerEstimator.ColumnsProduced[])"/>
+    /// <seealso cref="DateTimeTransformerExtensionClass.DateTimeTransformer(TransformsCatalog, string, string, DateTimeTransformerEstimator.ColumnsProduced[], DateTimeTransformerEstimator.Countries)"/>
+    public sealed class DateTimeTransformerEstimator : IEstimator<DateTimeTransformer>
```
eerhardt: The term `TransformerEstimator` is confusing. An object is either a "transformer" or it is an "estimator". There isn't such a thing as a "TransformerEstimator".

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,782 @@
...
+            [Argument(ArgumentType.MultipleUnique, HelpText = "Columns to drop after the DateTime Expansion", Name = "ColumnsToDrop", ShortName = "drop", SortOrder = 3)]
+            public ColumnsProduced[] ColumnsToDrop;
+
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Country to get holidays for. Defaults to none if not passed", Name = "Country", ShortName = "ctry", SortOrder = 4)]
+            public Countries Country = Countries.None;
+        }
+
+        #endregion
+
+        public DateTimeTransformerEstimator(IHostEnvironment env, string inputColumnName, string columnPrefix, ColumnsProduced[] columnsToDrop, Countries country = Countries.None)
```
eerhardt: The constructors should all be `internal`. The way these objects are instantiated is through the Catalog methods.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,782 @@
...
+
+        #region Enums
+        public enum ColumnsProduced : byte
+        {
+            Year = 1, Month, Day, Hour, Minute, Second, AmPm, Hour12, DayOfWeek, DayOfQuarter, DayOfYear,
+            WeekOfMonth, QuarterOfYear, HalfOfYear, WeekIso, YearIso, MonthLabel, AmPmLabel, DayOfWeekLabel,
+            HolidayName, IsPaidTimeOff
+        };
+
+        public enum Countries : byte
```
eerhardt: Is using an `enum` for countries the best route here? And if so, do you want this to be a `[Flags]` enum so you can request multiple countries?

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,782 @@
...
+        internal const string LoadName = "DateTimeTransform";
+        internal const string LoaderSignature = "DateTimeTransform";
+        private DateTimeTypedColumn _column;
+
+        private DateTimeTransformerEstimator.ColumnsProduced[] _columnsToDrop;
+        private byte[] _activeColumnMapping;
+
+        #endregion
+
+        public DateTimeTransformer(IHostEnvironment env, string inputColumnName, string columnPrefix, DateTimeTransformerEstimator.ColumnsProduced[] columnsToDrop, DateTimeTransformerEstimator.Countries country) :
```
eerhardt: This constructor shouldn't be public.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,782 @@
...
+    public static class DateTimeTransformerExtensionClass
+    {
+        /// <summary>
+        /// Create a <see cref="DateTimeTransformerEstimator"/>, which splits up the input column specified by <paramref name="inputColumnName"/>
+        /// into all its individual datetime components. Input column must be of type Int64 representing the number of seconds since the unix epoc.
+        /// This transformer will append the <paramref name="columnPrefix"/> to all the output columns. If <paramref name="columnsToDrop"/> is empty,
+        /// then all the columns are returned. Otherwise, the columns listed in the array will be dropped from the return value.
+        /// </summary>
+        /// <param name="catalog">Transform catalog</param>
+        /// <param name="inputColumnName">Input column name</param>
```
eerhardt: Is it possible to name the output columns? ML.NET typically allows the user to name the output column.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,782 @@
...
+            return new SchemaShape(columns.Values);
+        }
+
+        #region Enums
+        public enum ColumnsProduced : byte
+        {
+            Year = 1, Month, Day, Hour, Minute, Second, AmPm, Hour12, DayOfWeek, DayOfQuarter, DayOfYear,
+            WeekOfMonth, QuarterOfYear, HalfOfYear, WeekIso, YearIso, MonthLabel, AmPmLabel, DayOfWeekLabel,
+            HolidayName, IsPaidTimeOff
+        };
```
justinormont: Recommend explicitly numbering each enum name. This lets us insert new values later without affecting the enum numeric value. 

This is more important if the enum is serialized within the model, which I expect you're doing since the model needs to know which features to produce.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,782 @@
...
+        /// <summary>
+        /// Create a <see cref="DateTimeTransformerEstimator"/>, which splits up the input column specified by <paramref name="inputColumnName"/>
+        /// into all its individual datetime components. Input column must be of type Int64 representing the number of seconds since the unix epoc.
+        /// This transformer will append the <paramref name="columnPrefix"/> to all the output columns. If <paramref name="columnsToDrop"/> is empty,
+        /// then all the columns are returned. Otherwise, the columns listed in the array will be dropped from the return value.
+        /// </summary>
+        /// <param name="catalog">Transform catalog</param>
+        /// <param name="inputColumnName">Input column name</param>
+        /// <param name="columnPrefix">Prefix to add to the generated columns</param>
+        /// <param name="columnsToDrop">List of columns to drop, if any</param>
```
eerhardt: One drawback to using `columnsToDrop` is that if a user is only interested in the `DayOfYear` and `DayOfWeekLabel`, they need to explicitly exclude all the rest. And then if we ever add a new output column in a future release, they would start getting it, and would need to add the new column to the "to drop" list.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
justinormont: May want a UINT given that there are ~200 current countries (and they keep changing), and in the future we may a more specific break down, like `UnitedStatesBankingHolidays` vs. `UnitedStatesFederalHolidays`

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,782 @@
...
+
+            // This transformer adds columns
+            [Argument(ArgumentType.Required, HelpText = "Output column prefix", Name = "Prefix", ShortName = "pre", SortOrder = 2)]
+            public string Prefix;
+
+            [Argument(ArgumentType.MultipleUnique, HelpText = "Columns to drop after the DateTime Expansion", Name = "ColumnsToDrop", ShortName = "drop", SortOrder = 3)]
+            public ColumnsProduced[] ColumnsToDrop;
+
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Country to get holidays for. Defaults to none if not passed", Name = "Country", ShortName = "ctry", SortOrder = 4)]
+            public Countries Country = Countries.None;
```
justinormont: `Countries` may be a bit restrictive of a term for the future. Perhaps name as `HolidayList`. 

Perhaps in the future we'll want `UnitedStatesCaliforniaEducationHolidays` (see: [list](https://en.wikipedia.org/wiki/Holidays_with_paid_time_off_in_the_United_States#California_education_holidays)). We may also want more specific lists, like `UnitedStatesMajorHolidays`. Or localized lists like `GermanyBavaria` (see: [list](https://en.wikipedia.org/wiki/Public_holidays_in_Germany)).

Or perhaps in the future we'll want to version the holiday list `Germany2019` (as holidays are declared and dropped). Versioning helps old models run in current versions of ML.NET.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
justinormont: I might recommend a bitwise use of an enum (flags) to say which columns to include.

Default would be `DateTimeTransformerEstimator.ColumnsProduced.All = 0xFFFFFFFFFFFFFFFF` (uint64).

Then users can bitwise `or`/`and-not` to select the columns they want to keep/drop.

## Example usage
All output except `DayOfYear`: (subtractive column selection)
```c#
var pipeline = mlContext.Transforms.DateTimeTransformer("Date", "DTC", 
    DateTimeTransformerEstimator.ColumnsProduced.All & ~DateTimeTransformerEstimator.ColumnsProduced.DayOfYear)
```

Only `DayOfYear` and `DayOfWeekLabel`: (additive column selection)
```c#
var pipeline = mlContext.Transforms.DateTimeTransformer("Date", "DTC", 
    DateTimeTransformerEstimator.ColumnsProduced.DayOfYear | DateTimeTransformerEstimator.ColumnsProduced.DayOfWeekLabel)
```

justinormont: Apparently, there's a `[Flags]` attribute in C# -- https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=netcore-3.0

```c#
[Flags] public enum ColumnsProduced : uint64
{
    Year = 1 << 0,
    Month = 1 << 1,
    Day = 1 << 2,
    Hour = 1 << 3, 
    ...
}
             WeekOfMonth, QuarterOfYear,
```

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/Featurizers/DateTimeTransformerDropColumns.cs
```diff
@@ -0,0 +1,80 @@
...
+            public byte DTCDayOfWeek { get; set; }
+            public byte DTCDayOfQuarter { get; set; }
+            public ushort DTCDayOfYear { get; set; }
+            public ushort DTCWeekOfMonth { get; set; }
+            public byte DTCHalfOfYear { get; set; }
+            public byte DTCWeekIso { get; set; }
+            public int DTCYearIso { get; set; }
+            public string DTCMonthLabel { get; set; }
+            public string DTCAmPmLabel { get; set; }
+            public string DTCDayOfWeekLabel { get; set; }
```
justinormont: It would be needed to be added to the code being wrapped here, but when featurizing DateTime fields, I prefer polar (Sin/Cos) transforms (along with the others listed). 

Clocks and times/dates are cyclic. The current feature set doesn't include this fact.

The main gain of polar date transforms is that it places times like `11:50PM` and `12:10AM` right next to each other (both mapping to `~0.999` in the cosine transform of percent of day). This lets a model make a decision boundary at `CosTimeOfDay > 0.9` to treat these two times as very similar. 

Contrast this to the `DTCHour` feature which will map these to `23` and `0` (issue: non-continuous range, and not possible to group both the `0` and `23` in a single split point, which to a human is similar).

Info:
* https://ianlondon.github.io/blog/encoding-cyclical-features-24hour-time/
* https://medium.com/ai%C2%B3-theory-practice-business/top-6-errors-novice-machine-learning-engineers-make-e82273d394db#56ae
* https://blog.davidkaleko.com/feature-engineering-cyclical-features.html
* https://www.kaggle.com/avanwyk/encoding-cyclical-features-for-deep-learning
* https://medium.com/open-machine-learning-course/open-machine-learning-course-topic-6-feature-engineering-and-feature-selection-8b94f870706a#c36f


Code example I use for DateTime feature engineering:
```c#
DateTime dt;
var invalidDate = !DateTime.TryParse(input.Date, out dt);
var percentOfDay = (invalidDate ? Single.NaN : (float)(new TimeSpan(dt.Hour, dt.Minute, 0).TotalMinutes) / 1440 f);
var percentOfWeek = (invalidDate ? Single.NaN : ((float) dt.DayOfWeek + percentOfDay) / 7 f);
var percentOfMonth = (invalidDate ? Single.NaN : (float)(dt.Day + percentOfDay) / (float) DateTime.DaysInMonth(dt.Year, dt.Month));
var percentOfYear = (invalidDate ? Single.NaN : (float)((dt.DayOfYear + percentOfDay) / (new DateTime(dt.Year + 1, 1, 1) - new DateTime(dt.Year, 1, 1)).TotalDays));

// Time features like the current ones
output.DateWeekday = (invalidDate ? Single.NaN : (float) dt.DayOfWeek);
output.DateMonth = (invalidDate ? Single.NaN : (float) dt.Month);
output.DateHour = (invalidDate ? Single.NaN : dt.Hour);
output.DatePartOfDay = (invalidDate ? Single.NaN : (float) Math.Round(dt.AddHours(-3).Hour / 6 f));
output.InvalidDate = (invalidDate ? 1.0f : 0.0f);

// Polar date/time transforms for percent of day & week & year 
output.CosTimeOfDay = (float) Math.Cos(2 * Math.PI * percentOfDay);
output.SinTimeOfDay = (float) Math.Sin(2 * Math.PI * percentOfDay);
output.CosTimeOfWeek = (float) Math.Cos(2 * Math.PI * percentOfWeek);
output.SinTimeOfWeek = (float) Math.Sin(2 * Math.PI * percentOfWeek);
output.CosTimeOfMonth = (float) Math.Cos(2 * Math.PI * percentOfMonth);
output.SinTimeOfMonth = (float) Math.Sin(2 * Math.PI * percentOfMonth);
output.CosTimeOfYear = (float) Math.Cos(2 * Math.PI * percentOfYear);
output.SinTimeOfYear = (float) Math.Sin(2 * Math.PI * percentOfYear);
```

justinormont: /cc @davidbrownellWork

davidbrownellWork: Let's add this as a potential feature for Mn.

gvashishtha: You mean add to the ML.NET plan? Any estimate on how much additional dev time this would take?

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Changed to just be `DateTimeEstimator`.

---
In reply to: [353904119](https://github.com/dotnet/machinelearning/pull/4521#discussion_r353904119) [](ancestors = 353904119)

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Changed to internal.

---
In reply to: [353908126](https://github.com/dotnet/machinelearning/pull/4521#discussion_r353908126) [](ancestors = 353908126)

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Changed to internal.

---
In reply to: [353904541](https://github.com/dotnet/machinelearning/pull/4521#discussion_r353904541) [](ancestors = 353904541)

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Makes sense. I have renamed it to `HolidayList`, but we can revisit it if needed in the future as well.

---
In reply to: [353920532](https://github.com/dotnet/machinelearning/pull/4521#discussion_r353920532) [](ancestors = 353920532)

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Good point. They are now explicitly numbered. The `Countries` enum (now `HolidayList`) is also numbered now.

---
In reply to: [353910235](https://github.com/dotnet/machinelearning/pull/4521#discussion_r353910235) [](ancestors = 353910235)

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: So currently you can only request the holidays for one country at a time, since we can only return one holiday name and one paid time off value, so I think that an enum is fine for that.

---
In reply to: [353907829](https://github.com/dotnet/machinelearning/pull/4521#discussion_r353907829) [](ancestors = 353907829)

michaelgsharp: That is a good point. I have changed the enum to be `uint` as suggested. 

---
In reply to: [353916304](https://github.com/dotnet/machinelearning/pull/4521#discussion_r353916304) [](ancestors = 353916304)

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: This Transformer returns 21 different columns. Instead of specifying an output name for each column, the  columns already have a predefined name (Year, Month, Date, Etc), and the user can just specify a prefix to add to each column.

---
In reply to: [353910044](https://github.com/dotnet/machinelearning/pull/4521#discussion_r353910044) [](ancestors = 353910044)

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Do you have a way you think would be better? We had also thought about not even including that ability here and just letting the user use the `DropColumns` transformer to remove any columns they didn't want. That would keep this code cleaner and not duplicate that type of functionality. Thoughts on that?

---
In reply to: [353911746](https://github.com/dotnet/machinelearning/pull/4521#discussion_r353911746) [](ancestors = 353911746)

michaelgsharp: I dont think that doing flags is a great idea personally. I think enums are more clear (if more verbose) than flags usually. As I mentioned to Eric in this same thread though, we can remove this capability from here and just let the user use the `DropColumn` transformer to remove anything they dont want. What is your opinion on that?

---
In reply to: [353968394](https://github.com/dotnet/machinelearning/pull/4521#discussion_r353968394) [](ancestors = 353968394)

eerhardt: > We had also thought about not even including that ability here and just letting the user use the DropColumns transformer to remove any columns they didn't want.

The way [IDataView works](https://github.com/dotnet/machinelearning/blob/master/docs/code/IDataViewDesignPrinciples.md#lazy-computation-and-active-columns) in ML.NET is that if a column isn't used downstream, it isn't even materialized. So if a data transform produces 100 columns, but only 10 columns are used by the learner, the other 90 columns don't get materialized. See the linked doc for more info.

So if you are open to not even including the `drop` ability here, I think that works and the perf should be the same.

michaelgsharp: I am open to that. I already synced with my team on it and they are fine with that as well. I will go ahead then and remove it.

---
In reply to: [354479281](https://github.com/dotnet/machinelearning/pull/4521#discussion_r354479281) [](ancestors = 354479281)

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,834 @@
...
+
+                return columns.ToArray();
+            }
+
+            private Delegate MakeGetter<T>(DataViewRow input, int iinfo)
+            {
+                ValueGetter<T> result = (ref T dst) =>
+                {
+                    long dateTime = default;
+                    var getter = input.GetGetter<long>(input.Schema[_parent._column.Source]);
```
eerhardt: The `GetGetter` method should be called outside of this delegate and cached. Doing this every time a value is requested isn't great for performance.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,834 @@
...
+                    }
+
+                return columns.ToArray();
+            }
+
+            private Delegate MakeGetter<T>(DataViewRow input, int iinfo)
+            {
+                ValueGetter<T> result = (ref T dst) =>
+                {
+                    long dateTime = default;
```
eerhardt: Why is the input `dateTime` column a `long` and not a `DateTime`?

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,834 @@
...
+                else // 64 bit machine
+                    buffer = new byte[*(ulong*)(rawData + intPtrSize)];
+
+                if (buffer.Length == 0)
+                {
+                    rawData += intPtrSize * 2;
+                    return string.Empty;
+                }
+
+                Marshal.Copy(new IntPtr(*(int**)rawData), buffer, 0, buffer.Length);
```
eerhardt: If you are using .NET Core 2.1 and above, you can save this buffer and copy and use `Encoding.UTF8.GetString(ReadOnlySpan<byte>)`, where the Span points directly to native memory. It is one less alloc and copy.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,834 @@
...
+                foreach (var toDrop in _columnsToDrop)
+                    ctx.Writer.Write((byte)toDrop);
+            }
+
+            var data = _column.CreateTransformerSaveData();
+            ctx.Writer.Write(data.Length);
+            ctx.Writer.Write(data);
+        }
+
+        public void Dispose()
```
eerhardt: Who calls this Dispose method? We have a problem with tranformers that need to be disposed. See [https://github.com/dotnet/machinelearning/issues/906](https://github.com/dotnet/machinelearning/issues/906) and my comment on [https://github.com/dotnet/machinelearning/pull/4223](https://github.com/dotnet/machinelearning/pull/4223)

cc @codemsz

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,834 @@
...
+            public byte HalfOfYear;
+            public byte WeekIso;
+            public int YearIso;
+            public string MonthLabel;
+            public string AmPmLabel;
+            public string DayOfWeekLabel;
+            public string HolidayName;
+            public byte IsPaidTimeOff;
+
+            internal unsafe TimePoint(byte* rawData)
```
eerhardt: (suggestion) you could make this safe code by using `ReadOnlySpan<byte>` instead of `byte*`.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: `long` to match a unix timestamp. We will be adding in the ability to use C# `DateTime` and `string` in an ISO format in the future, but the initial implementation is just the unix timestamp.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: It should just be called automatically when the pipeline teardown happens as mentioned in that comment chain. I am not manually calling it anywhere. Is it happening during pipeline teardown ok? Or is there more desirable behavior?

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: I will make that change. How does the performance compare between the raw pointer and the span?

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
eerhardt: > It should just be called automatically when the pipeline teardown happens as mentioned in that comment chain.

There is no call of it when the pipeline teardown happens. That is why that bug was opened. I'm still not certain on why it was closed. Maybe @codemzs  can comment.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
eerhardt: There should be no measurable performance difference between them.

eerhardt: There should be no measurable performance difference between them.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
eerhardt: My concern here is that .NET developers expect to have their date time data using `System.DateTime`. That is the natural type in .NET. Forcing the user to convert from `System.DateTime` to a long unix timestamp doesn't seem like a good experience. It would make more sense if the input column was a `System.DateTime`.

eerhardt: Why not do the conversion yourself here? The user's input column can be `DateTime`, and then you convert it to a long unix timestamp before passing it to the C++ code?

davidbrownellWork: We want to ensure that the process of converting from a string to "Date time" is consistent across frameworks - especially considering that a DateTime in and of itself isn't sufficient in some scenarios, as we lose (potentially important) time zone information in the process.

eerhardt: I'm not asking for string. I'm asking for the ML.NET data type that a user would feed into this transform to be the canonical `System.DateTime` type and not a `long` type.

Specifically, this transform should require the input column to be of this type:

https://github.com/dotnet/machinelearning/blob/9fc8f1cd323d572672a5460914f927acfef310dc/src/Microsoft.ML.DataView/DataViewType.cs#L361-L365

It doesn't make sense for a user to have to convert a `DateTime` to a `long` before calling this transform.

eerhardt: Specifically, look at the sample code being checked in under `samples`:

```C#
            // Create a small dataset as an IEnumerable.
            // Future Date - 2025 June 30
            var samples = new[] { new DateTimeInput() { Date = 1751241600 } };
```

Instead, that sample code should be:

```C#
            // Create a small dataset as an IEnumerable.
            var samples = new[] { new DateTimeInput() { Date = new DateTime(2025, 6, 30) } };
```

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/Featurizers/DateTimeTransformerDropColumns.cs
justinormont: Adding the new features is pretty simple. I expect more of the work is in testing.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
justinormont: > The way IDataView works in ML.NET is that if a column isn't used downstream, it isn't even materialized. So if a data transform produces 100 columns, but only 10 columns are used by the learner, the other 90 columns don't get materialized. See the linked doc for more info.

While true in general, I expect this featurizer isn't aware of which columns are being requested of it, and will produce all columns.

harishsk: I think the better approach is to use the DropColumns facility that is already present in ML.NET. 

---
In reply to: [354480672](https://github.com/dotnet/machinelearning/pull/4521#discussion_r354480672) [](ancestors = 354480672,354479281)

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: I have changed it to use spans. Because of a double pointer issue I still need unsafe, but now the constructor is safe and just internal things are unsafe.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,834 @@
...
+            AmPmLabel = 18,
+            DayOfWeekLabel = 19,
+            HolidayName = 20,
+            IsPaidTimeOff = 21
+        };
+
+        public enum HolidayList : uint
+        {
+            None = 1,
+            Argentina = 2,
```
harishsk: What about choosing the ISO 3166 country codes for the enum values? 
[https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes)

michaelgsharp: The name of the country needs to be passed to the native code, so I use C# to turn the name of the enum into a string, and then pass it in that way.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
eerhardt: >  I expect this featurizer isn't aware of which columns are being requested of it

Maybe it should be made aware?

justinormont: I like the usability of the flag method. 

There is a user need of not including all outputs as overfitting occurs mainly on the larger time scales, like year number. Including the year in a classification or regression model often leads to the model memorizing the label value specifically for a date range. 

For instance the model will memorize that 2017-03-01 to 2017-04-05, the label value was high. This information is not valuable to memorize when predicting future data. Instead we want to ensure it memorizes that traffic volumes (the label) are higher on weekends.

Hence it nice to encourage users (as part of the timedate featurizer API itself) to think about which columns to include/exclude. This is as opposed to a user needing a separate DropColumns step in their pipeline.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
justinormont: What is the purpose of returning a holiday name (string)? Shouldn't we return a boolean of `IsHoliday`? Or perhaps the boolean and a one-hot encoded version of which holiday? The slot names of the one-hot encoded field can be the exact holiday name.

In the end, all of the created features need to be turned into a float32, as the model takes in a float32 array as a feature vector.

You may also want to include `TimeUntilNextHoliday` and `TimeLastLastHoliday`. @EricWrightAtWork, which DateTime features are needed for AutoML forecasting?

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: How can I tell if its using .NET Core 2.1 or above? I thought I was, but I dont have the method signature you listed. I only have the one that takes a `byte[]`. All I am seeing is that the target framework is `netstandard2.0`.

eerhardt: You need to cross-target the project for multiple frameworks. Today the project only builds for `netstandard2.0`. 

https://github.com/dotnet/machinelearning/blob/9fc8f1cd323d572672a5460914f927acfef310dc/src/Microsoft.ML.Featurizers/Microsoft.ML.Featurizers.csproj#L4

But if you change that to

```xml
<TargetFrameworks>netstandard2.0;netcoreapp2.1</TargetFrameworks>
```

Then the project builds 2 dlls - one for netstandard 2.0 and one for netcoreapp2.1.

And then in your code you can use a `#if` to detect which one you are building for:

```C#
#if NETSTANDARD2_0
// current code
#else
// use Encoding.UTF8.GetString(ReadOnlySpan<byte>)
#endif
```

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
davidbrownellWork: Agreed - eventually a C# developer will be able to pass `System:DateTime`, `string`, or `long`.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: I moved the `GetGetter` outside the delegate following the pattern that the other transformers follow. Is .net able to optimize it if its done that way?

michaelgsharp:
Going to close and reopen the PR to reset the stuck builds.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
eerhardt: See https://github.com/dotnet/machinelearning/blob/master/docs/code/IDataViewImplementation.md#why-getgetter

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,794 @@
...
+        /// </summary>
+        /// <param name="catalog">Transform catalog</param>
+        /// <param name="inputColumnName">Input column name</param>
+        /// <param name="columnPrefix">Prefix to add to the generated columns</param>
+        /// <returns><see cref="DateTimeEstimator"/></returns>
+        public static DateTimeEstimator DateTimeTransformer(this TransformsCatalog catalog, string inputColumnName, string columnPrefix)
+            => new DateTimeEstimator(CatalogUtils.GetEnvironment(catalog), inputColumnName, columnPrefix);
+
+        /// <summary>
+        /// Create a <see cref="DateTimeEstimator"/>, which splits up the input column specified by <paramref name="inputColumnName"/>
```
natke: Distinguish early in the description the variant that this method offers. In fact, is this method necessary. Could the country be an optional parameter of the first method?

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,794 @@
...
+    /// <format type="text/markdown"><![CDATA[
+    ///
+    /// ###  Estimator Characteristics
+    /// |  |  |
+    /// | -- | -- |
+    /// | Does this estimator need to look at the data to train its parameters? | No |
+    /// | Input column data type | Int64 |
+    /// | Output column data type | Columns and types listed in the summary |
+    ///
+    /// The <xref:Microsoft.ML.Transforms.CategoryImputerEstimator> is a trivial estimator and does not need training.
```
natke: Why is this reference here?

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,794 @@
...
+            DayOfWeek = 9,
+            DayOfQuarter = 10,
+            DayOfYear = 11,
+            WeekOfMonth = 12,
+            QuarterOfYear = 13,
+            HalfOfYear = 14,
+            WeekIso = 15,
+            YearIso = 16,
+            MonthLabel = 17,
+            AmPmLabel = 18,
```
natke: If we do AmPm label, would it make sense to also do weekend/weekday? (Often used in feature engineering)

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
natke: Strongly agree on using System.DateTime from the beginning. This will make the transform consistent with other parts of the API, as well as more usable to .NET developers

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,794 @@
...
+        /// <summary>
+        /// Create a <see cref="DateTimeEstimator"/>, which splits up the input column specified by <paramref name="inputColumnName"/>
+        /// into all its individual datetime components. Input column must be of type Int64 representing the number of seconds since the unix epoc.
+        /// This transformer will append the <paramref name="columnPrefix"/> to all the output columns.
+        /// </summary>
+        /// <param name="catalog">Transform catalog</param>
+        /// <param name="inputColumnName">Input column name</param>
+        /// <param name="columnPrefix">Prefix to add to the generated columns</param>
+        /// <returns><see cref="DateTimeEstimator"/></returns>
+        public static DateTimeEstimator DateTimeTransformer(this TransformsCatalog catalog, string inputColumnName, string columnPrefix)
```
natke: The naming of the extension methods are not consistent with the other transformers (I'll leave it to others to judge  whether the original scheme makes sense :-), but consistency is probably a more important consideration now)

If you look at the other transforms they mostly have the form Verb-Object e.g. CopyColumns, MapValueToKey etc. This makes sense when you are chaining them together in a pipeline.

Suggestions for this one: FeaturizeDateTime, SplitDateTime, TokenizeDateTime, ExtractDateTime ...

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: So so far we have had 3 suggestions.
1 - Dont support dropping columns as a native part of this transformer. 
2 - Use Enums.
3 - Use Flags.

Can we finalize on the decision? I personally agree with Harish that since we already have a drop columns transformer that we can just use that and not have that functionality be present here.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Yep. I will go ahead and combine them.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Good catch. Renamed it to DateTimeTransformer.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: I think thats not a bad idea, but the initial ask we got was just for these columns. We will start with just these and then can add in more if the requirement arises.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Of the ones you listed I would vote `FeaturizeDateTime`. Anyone have any objections if I rename the extension method to that?

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
justinormont: I'd recommend a way to proactively select the output columns to create during the setup of the transform. Instead of dropping later, just don't create them.

Generally, we should assume users won't think (or know) to drop later. For usability, placing it in the constructor exposes users to the idea, and forces them to consider which columns will be beneficial.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -0,0 +1,794 @@
...
+                case DateTimeEstimator.ColumnsProduced.Year:
+                case DateTimeEstimator.ColumnsProduced.YearIso:
+                    return typeof(int);
+                case DateTimeEstimator.ColumnsProduced.DayOfYear:
+                case DateTimeEstimator.ColumnsProduced.WeekOfMonth:
+                    return typeof(ushort);
+                case DateTimeEstimator.ColumnsProduced.MonthLabel:
+                case DateTimeEstimator.ColumnsProduced.AmPmLabel:
+                case DateTimeEstimator.ColumnsProduced.DayOfWeekLabel:
+                case DateTimeEstimator.ColumnsProduced.HolidayName:
```
justinormont: Usability question, why are we producing multiple output columns of instead of a single vector type columns?

The next step after this DateTime transform is:
* **Convert** all the INTs/USHORTs/BYTEs columns to Float32
* **One-hot encode** the STRING columns (`"Friday"` to `[ 0, 0,  0, 0,  0, 1, 0 ]`)
* **Concatenate** all of these feature to single vector of Float32

Why aren't we just outputting the final vector type column? Then the user can directly use its output, like our other transforms.

justinormont: If there's a usecase for the individual columns, perhaps just a boolean parameter `produceRawColumns`, which then produces the individual outputs instead of a singular vector column?

michaelgsharp: We did it this way just because thats how the original python code works. It splits everything up into different columns and doesn't combine them into a vector.

natke: Having a single vector is consistent with `FeaturizeText`.

michaelgsharp: The original code we were bringing to ML.NET returned all individual columns. Because of that, for now, we will be leaving them as individual columns. If we learn that we never need individual columns and only use this as a vector then we can revisit this then.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: That is on our todo list and will be added in in the future. We will eventually be supporting `DateTime` and `string` as well, but initial support will just be for `long`.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: For now, since this functionality already exists inside of ML.NET as a separate transformer we will be leaving this as is. We can revisit this later as the need arises. Even if this transformer produces all the columns, if no downstream transformers request them then they will never be realized. It shouldn't cause any performance impact this way.

src/Microsoft.ML.Featurizers/Common.cs
```diff
@@ -219,5 +220,31 @@ internal static TypeId GetNativeTypeIdFromType(this Type type)
 
             throw new InvalidOperationException($"Unsupported type {type}");
         }
+
+        // The Native Featurizers do not currently support CentOS7, this method checks the OS and returns true if it is CentOS7.
```
eerhardt: What about RedHat 7?

src/Microsoft.ML.Featurizers/Common.cs
```diff
@@ -219,5 +220,31 @@ internal static TypeId GetNativeTypeIdFromType(this Type type)
...
+        internal static bool OsIsCentOS7()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                using (Process process = new Process())
```
eerhardt: A more performant way of doing this check would be to read from the file directly. Starting a new process just to get this info is too heavy-weight IMO.

See [https://github.com/dotnet/runtime/blob/6f445d6dc237f59d730e0df47f8630b18887d776/src/installer/managed/Microsoft.DotNet.PlatformAbstractions/Native/PlatformApis.cs#L140-L142](https://github.com/dotnet/runtime/blob/6f445d6dc237f59d730e0df47f8630b18887d776/src/installer/managed/Microsoft.DotNet.PlatformAbstractions/Native/PlatformApis.cs#L140-L142)

PR #4533, opened 2019-12-06
Onnx conversion theory

sharwell:
Use `[Theory]` to break up tests in OnnxConversionTest.

📝 Builds on #4532 and #4545

💡 Recommended review strategy is commit-by-commit. The final commit is easier to review with whitespace-only changes hidden.

### DIFF
--- a/test/Microsoft.ML.Tests/OnnxConversionTest.cs
+++ b/test/Microsoft.ML.Tests/OnnxConversionTest.cs
@@ -314,8 +314,11 @@ namespace Microsoft.ML.Tests
             public float[] Features { get; set; }
         }
 
-        [Fact]
-        public void LpNormOnnxConversionTest()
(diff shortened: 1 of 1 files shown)


test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -312,8 +312,11 @@ private class DataPoint
             public float[] Features { get; set; }
         }
 
-        [Fact]
-        void LpNormOnnxConversionTest()
+        [Theory]
+        [CombinatorialData]
+        public void LpNormOnnxConversionTest(
+            bool ensureZeroMean,
+            LpNormNormalizingEstimatorBase.NormFunction norm)
```
codemzs: I assume it will infer all possible values in this case since they are discrete?

PR #4535, opened 2019-12-06
Prevent exceptions from escaping FileSystemWatcher events

sharwell:
Exceptions in the handler of FileSystemWatcher events will terminate the process. This pull request updates the event handler to use a try/catch block that logs the error instead of crashing the process.

This fixes one known cause of test process failures.

### DIFF
--- a/src/Microsoft.Extensions.ML/ModelLoaders/FileModelLoader.cs
+++ b/src/Microsoft.Extensions.ML/ModelLoaders/FileModelLoader.cs
@@ -64,20 +64,32 @@ namespace Microsoft.Extensions.ML
         private void WatcherChanged(object sender, FileSystemEventArgs e)
         {
             var timer = Stopwatch.StartNew();
-            Logger.FileReloadBegin(_logger, _filePath);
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.ML/ModelLoaders/FileModelLoader.cs
```diff
@@ -116,6 +128,7 @@ internal static class EventIds
             public static readonly EventId FileReloadBegin = new EventId(100, "FileReloadBegin");
             public static readonly EventId FileReloadEnd = new EventId(101, "FileReloadEnd");
             public static readonly EventId FileReload = new EventId(102, "FileReload");
+            public static readonly EventId FileReloadError = new EventId(103, nameof(FileReloadError));
```
sharwell: 📝 I wanted to make this have the same order as `UriModelLoader`, but I was not sure if these event IDs need to be stable over time. Since I wasn't sure, I opted for keeping the current IDs unchanged and just added the new one.

PR #4536, opened 2019-12-06
Update Microsoft.Extensions.* to 3.0.1

sharwell:
The 3.0 release includes a fix to cases where `ChangeToken.OnChange` would not unregister callbacks even when disposed.

See aspnet/Extensions#558, aspnet/Extensions#869

### DIFF
--- a/build/Dependencies.props
+++ b/build/Dependencies.props
@@ -52,6 +52,7 @@
   <PropertyGroup>
     <BenchmarkDotNetVersion>0.11.3</BenchmarkDotNetVersion>
     <MicrosoftCodeAnalysisTestingVersion>1.0.1-beta1.20080.1</MicrosoftCodeAnalysisTestingVersion>
+    <MicrosoftExtensionsTestPackageVersion>3.0.1</MicrosoftExtensionsTestPackageVersion>
     <MicrosoftMLTestDatabasesPackageVersion>0.0.5-test</MicrosoftMLTestDatabasesPackageVersion>
     <MicrosoftMLTestModelsPackageVersion>0.0.6-test</MicrosoftMLTestModelsPackageVersion>
     <MicrosoftMLTensorFlowTestModelsVersion>0.0.11-test</MicrosoftMLTensorFlowTestModelsVersion>
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -38,6 +38,14 @@
   </ItemGroup>
 
   <ItemGroup>
+    <!-- Make sure to use the updated version of Microsoft.Extensions.* in test projects -->
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="$(MicrosoftExtensionsTestPackageVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsTestPackageVersion)" />
+    <PackageReference Include="Microsoft.Extensions.ObjectPool" Version="$(MicrosoftExtensionsTestPackageVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Options" Version="$(MicrosoftExtensionsTestPackageVersion)" />
+  </ItemGroup>
+
+  <ItemGroup>
     <ProjectReference
       Condition="'$(UseMLCodeAnalyzer)' != 'false' and '$(MSBuildProjectExtension)' == '.csproj'"
       Include="$(MSBuildThisFileDirectory)\..\tools-local\Microsoft.ML.InternalCodeAnalyzer\Microsoft.ML.InternalCodeAnalyzer.csproj">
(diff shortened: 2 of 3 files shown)


build/Dependencies.props
```diff
@@ -15,7 +15,7 @@
   <PropertyGroup>
     <GoogleProtobufPackageVersion>3.10.1</GoogleProtobufPackageVersion>
     <LightGBMPackageVersion>2.2.3</LightGBMPackageVersion>
-    <MicrosoftExtensionsPackageVersion>2.1.0</MicrosoftExtensionsPackageVersion>
+    <MicrosoftExtensionsPackageVersion>3.0.1</MicrosoftExtensionsPackageVersion>
```
eerhardt: How will this affect users who are targeting .NET Core 2.1 in their web apps? Is it valid for us to pull in a higher `Microsoft.Extensions.*` assemblies? I believe it will mean that these assemblies will now be published along with the application, and no longer used from the shared framework, which means larger apps. It could also mean slower start up times because the assemblies won't be crossgen'd.

/cc @glennc

eerhardt: @glennc @davidfowl @Pilchie - is the `OnChange` bug fix for dotnet/extensions#558 something that should be ported to `2.1`? I'm concerned here that anyone using `2.1` + ML.NET would be forced to update `Microsoft.Extensions.*` to `3.0.1`. If this is a serious enough bug to cause ML.NET to force everyone to `3.0`, maybe it should be back-ported to `2.1.` instead.

Pilchie: In general, we don't really recommend using 3.x NuGet packages in a 2.1 app.  It might work for this one, but there are a significant number of breaking changes in ASP.NET all-up that makes it problematic.  We can consider back-porting a fix to 2.1, but the usual bar is that there is a large customer blocked by the issue.

Pilchie: What is the impact of this bug on ML.NET Customer?

harishsk: @sharwell I thought you said you noticed this issue in the LdaTransform. Can you please help identify the overall impact?

eerhardt: > What is the impact of this bug on ML.NET Customer?

@sharwell - can you answer this?

eerhardt: > What is the impact of this bug on ML.NET Customer?

@sharwell - can you answer this?

sharwell: The problem was observed here:

https://github.com/dotnet/machinelearning/blob/f3e0f6b3a52f7e3998978239a0f52f35fd6197d8/test/Microsoft.Extensions.ML.Tests/UriLoaderTests.cs#L44-L48

Disposing of the registration doesn't do anything. It's been a while since I reviewed this to figure out all the details. What about conditioning this dependency on the build target, so we have a dependency to 2.1 for netcoreapp2.1, but a dependency on 3.0.1 when targeting netcoreapp3.0+?

eerhardt: If this is just a test concern, I think we should update the test's .csproj to reference `Microsoft.Extensions.*` version `3.0.*` as necessary.

Users who need this bug fix can reference `3.0` themselves, if necessary. But at least we won't force any users on `2.1` to be in a non-recommended situation.

harishsk: @sharwell Are there other cases where this bug surfaces? If not, is Eric's suggestion an acceptable fix?

harishsk:
@sharwell Is it possible to have a separate property for just the test project and apply it only to the test project?

sharwell:
@harishsk Done!

build/Dependencies.props
harishsk: @eerhardt Does the update PR address your concerns? If so, can you please approve?

test/Directory.Build.props
```diff
@@ -37,6 +37,14 @@
     <PackageReference Include="coverlet.msbuild" Version="$(CoverletVersion)" />
   </ItemGroup>
 
+  <ItemGroup>
+    <!-- Make sure to use the updated version of Microsoft.Extensions.* in test projects -->
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="$(MicrosoftExtensionsTestPackageVersion)" />
```
eerhardt: Why does every test project have to reference these packages? Most of the test projects don't need these references.

sharwell: It didn't seem to hurt anything, and makes it easier to ensure that if the packages are needed, they use the correct version.

eerhardt: > It didn't seem to hurt anything

That doesn't seem like a good reason.

Adding unnecessary dependencies causes people to start using them when they shouldn't be.

sharwell: If it's helpful for test code, does it matter?

eerhardt: There’s only 1 project that needs these dependencies right? I think it would be different if there were multiple projects that needed these.

It also sets a pattern for the next person who needs to add a dependency to a project to just add it here and add it for all.

sharwell: It's hard for me to tell because of the transient nature of dependencies today. Otherwise I could just use this:

```xml
<PackageReference Update="Microsoft.Extensions.DependencyInjection.Abstractions" Version="$(MicrosoftExtensionsTestPackageVersion)" />
```

I'm not sure of a way to verify these are correct _without_ adding the dependencies to all test projects except by reviewing the full restore graph for each project.

PR #4537, opened 2019-12-06
Enable VSTestBlame to show details for crashes

sharwell:


test/run-tests.proj
```diff
@@ -10,7 +10,7 @@
     <Message Importance="High" Text="Running tests ..." />
     <MSBuild Targets="VSTest"
              Projects="@(Project)"
-             Properties="VSTestNoBuild=true" />
+             Properties="VSTestNoBuild=true;VSTestBlame=true" />
```
codemzs: >Properties="VSTestNoBuild=true;VSTestBlame=true" /> [](start = 13, length = 51)

seems like a great change for debugging test failures.

sharwell: It's not the be-all-end-all, but it does seem to produce reasonable output. Marked as ready for review.

PR #4538, opened 2019-12-07
Convert LdaEngine to a SafeHandle

sharwell:
This pull request follows the test failure observed in https://dev.azure.com/dnceng/public/_build/results?buildId=449759&view=ms.vss-test-web.build-test-results-tab&runId=14347330&resultId=100770&paneView=debug. Converting this object to a SafeHandle has two primary advantages:

1. The object will not be disposed while there is an ongoing native call
2. Misuse of the object by trying to pass an invalid instance to a native call will throw an exception at the point of the call

### DIFF
--- a/src/Microsoft.ML.Transforms/Text/LdaSingleBox.cs
+++ b/src/Microsoft.ML.Transforms/Text/LdaSingleBox.cs
@@ -8,83 +8,93 @@ using System.Linq;
 using System.Runtime.InteropServices;
 using System.Security;
 using Microsoft.ML.Runtime;
+using Microsoft.Win32.SafeHandles;
 
 namespace Microsoft.ML.TextAnalytics
 {
 
     internal static class LdaInterface
     {
-        public struct LdaEngine
+        public sealed class SafeLdaEngineHandle : SafeHandleZeroOrMinusOneIsInvalid
         {
-            public IntPtr Ptr;
+            private SafeLdaEngineHandle()
+                : base(true)
+            {
+            }
+
+            protected override bool ReleaseHandle()
+            {
+                DestroyEngine(handle);
+                return true;
+            }
         }
 
         private const string NativePath = "LdaNative";
         [DllImport(NativePath), SuppressUnmanagedCodeSecurity]
-        internal static extern LdaEngine CreateEngine(int numTopic, int numVocab, float alphaSum, float beta, int numIter,
+        internal static extern SafeLdaEngineHandle CreateEngine(int numTopic, int numVocab, float alphaSum, float beta, int numIter,
             int likelihoodInterval, int numThread, int mhstep, int maxDocToken);
 
         [DllImport(NativePath), SuppressUnmanagedCodeSecurity]
-        internal static extern void AllocateModelMemory(LdaEngine engine, int numTopic, int numVocab, long tableSize, long aliasTableSize);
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.Transforms/Text/LdaTransform.cs
```diff
@@ -892,11 +877,10 @@ private static List<VBuffer<ReadOnlyMemory<char>>> Train(IHostEnvironment env, I
             // Initialize all LDA states
             for (int i = 0; i < columns.Length; i++)
             {
-                var state = new LdaState(env, columns[i], numVocabs[i]);
-
                 if (numDocArray[i] == 0 || corpusSize[i] == 0)
                     throw ch.Except("The specified documents are all empty in column '{0}'.", columns[i].InputColumnName);
 
+                var state = new LdaState(env, columns[i], numVocabs[i]);
```
sharwell: 📝 This is a fairly straightforward change. We avoid creating an instance of `LdaState` which would result in an exception during dispose, for the case where an exception is thrown two lines above.

src/Microsoft.ML.Transforms/Text/LdaTransform.cs
```diff
@@ -693,25 +693,10 @@ internal static LatentDirichletAllocationTransformer TrainLdaTransformer(IHostEn
...
-            if (disposing)
-                GC.SuppressFinalize(this);
-        }
-
         public void Dispose()
         {
-            Dispose(true);
-        }
-
-        ~LatentDirichletAllocationTransformer()
```
sharwell: 📝 This finalizer was unnecessary. The type is sealed and does not directly own any unmanaged resources.

src/Microsoft.ML.Transforms/Text/LdaTransform.cs
```diff
@@ -49,7 +49,7 @@ namespace Microsoft.ML.Transforms.Text
     /// <summary>
     /// <see cref="ITransformer"/> resulting from fitting a <see cref="LatentDirichletAllocationEstimator"/>.
     /// </summary>
-    public sealed class LatentDirichletAllocationTransformer : OneToOneTransformerBase
+    public sealed class LatentDirichletAllocationTransformer : OneToOneTransformerBase, IDisposable
```
sharwell: ⚠️ This is the "ugly" part of this change for me. If someone is upcasting this to `ITransformer`, how will they know to dispose of it?

sharwell:
Due to a bug in the native `LdaEngine` code, I am treating this pull request as blocked on #4544.

src/Microsoft.ML.Transforms/Text/LdaTransform.cs
sharwell: This change is no longer necessary after #4551

src/Microsoft.ML.Transforms/Text/LdaTransform.cs
sharwell: I'm leaving this change for a separate pull request.

src/Microsoft.ML.Transforms/Text/LdaTransform.cs
sharwell: I'm leaving this change for a separate pull request.

sharwell:
With changes to the native library now merged (#4551), I reduced this pull request to the minimum necessary to stabilize the use of this particular handle.

PR #4539, opened 2019-12-07
Create SafeBoosterHandle and SafeDataSetHandle

sharwell:
These safe handles were created based on the crash data in https://dev.azure.com/dnceng/public/_build/results?buildId=449745&view=logs.

### DIFF
--- a/src/Microsoft.ML.LightGbm/WrappedLightGbmBooster.cs
+++ b/src/Microsoft.ML.LightGbm/WrappedLightGbmBooster.cs
@@ -19,14 +19,13 @@ namespace Microsoft.ML.Trainers.LightGbm
         private readonly bool _hasValid;
         private readonly bool _hasMetric;
 
-        public IntPtr Handle { get; private set; }
+        public WrappedLightGbmInterface.SafeBoosterHandle Handle { get; private set; }
         public int BestIteration { get; set; }
 
         public Booster(Dictionary<string, object> parameters, Dataset trainset, Dataset validset = null)
         {
             var param = LightGbmInterfaceUtils.JoinParameters(parameters);
-            var handle = IntPtr.Zero;
-            LightGbmInterfaceUtils.Check(WrappedLightGbmInterface.BoosterCreate(trainset.Handle, param, ref handle));
+            LightGbmInterfaceUtils.Check(WrappedLightGbmInterface.BoosterCreate(trainset.Handle, param, out var handle));
             Handle = handle;
             if (validset != null)
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.ML.LightGbm/WrappedLightGbmBooster.cs
```diff
@@ -19,14 +19,13 @@ internal sealed class Booster : IDisposable
...
-        public IntPtr Handle { get; private set; }
+        public WrappedLightGbmInterface.SafeBoosterHandle Handle { get; private set; }
         public int BestIteration { get; set; }
 
         public Booster(Dictionary<string, object> parameters, Dataset trainset, Dataset validset = null)
         {
             var param = LightGbmInterfaceUtils.JoinParameters(parameters);
-            var handle = IntPtr.Zero;
-            LightGbmInterfaceUtils.Check(WrappedLightGbmInterface.BoosterCreate(trainset.Handle, param, ref handle));
+            LightGbmInterfaceUtils.Check(WrappedLightGbmInterface.BoosterCreate(trainset.Handle, param, out var handle));
```
sharwell: 📝 The change to SafeHandle means that the finalizer for `SafeBoosterHandle` will now be invoked at some point if the call to `BoosterCreate` succeeds but an exception is thrown later in this constructor (i.e. an exception prevents the caller from being able to use `Booster.Dispose()`). Ideally, code which constructs disposable objects should be robust in ensuring created objects are disposed if a failure would prevent those objects from being disposed by the caller. I'm not ready to submit the change to ensure all the callers dispose of objects deterministically on exceptional paths, but I'd like to merge this one in the meantime.

eerhardt: Doing a quick scan of the usages of the Booster and Dataset classes, they are always disposed of in the product code (one in a `using` and the other in a `try-finally`. The only case I saw where they weren't be disposed in in the tests.

https://github.com/dotnet/machinelearning/blob/8221dace027292a875922a93f2689ba1797f9411/test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs#L494

https://github.com/dotnet/machinelearning/blob/8221dace027292a875922a93f2689ba1797f9411/test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs#L475

sharwell: The `Booster` class is disposed if the constructor returns without throwing an exception. However, if `LightGbmInterfaceUtils.Check` throws an exception from the constructor itself, the `SafeBoosterHandle` will have been created by the call to `BoosterCreate` but never stored in an object that is later disposed. Prior to the use of a `SafeHandle`, an instance in this scenario would never be cleaned up. Now it will be cleaned up by the finalizer.

PR #4543, opened 2019-12-07
Update code coverage integration

sharwell:
* Simplifies coverage integration
* Removes the need to define CODECOV_TOKEN during builds


sharwell:
This build appears to have the expected results (no coverage changes).

PR #4544, opened 2019-12-08
WIP Safe pointers

sharwell:
Locally I converted most owned native objects to `std::unique_ptr` and `std::shared_ptr`. This pull request represents a relatively easy-to-review subset of the changes with minimal overall fan-out in the code.

* Pass by reference when null is not expected (submitted separately as #4546)
* Use `std::unique_ptr` for objects in `LdaEngine` (submitted separately as #4547)
* Use `std::unique_ptr` for `samplers_` and `likelihood_in_iter_`

### DIFF
--- a/src/Native/LdaNative/lda_engine.cpp
+++ b/src/Native/LdaNative/lda_engine.cpp
@@ -167,7 +167,7 @@ namespace lda {
         CTimer tmDebug(true);
         CheckFunction(0, tmDebug, "enter initializeBeforeTrain", false);
         //allocate model memory from the data preloaded
-        AllocateModelMemory(data_block_);
+        AllocateModelMemory(*data_block_);
         CheckFunction(0, tmDebug, "allocate model memory", false);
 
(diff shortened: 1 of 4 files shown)


src/Native/LdaNative/lda_engine.cpp
```diff
@@ -105,44 +101,30 @@ namespace lda {
...
+        samplerQueue_.reset(nullptr);
 
-        delete samplerQueue_;
-        samplerQueue_ = nullptr;
-
-        for (int i = 0; i < num_threads_; ++i)
-        {
-            delete samplers_[i];
-        }
-        delete[] samplers_;
```
sharwell: 📝 This block was responsible for some of the crashes observed in CI, and especially the crashes prior to 0cbfa54b641663627b194775de30bac3ad35f803. The destructor is missing a null check for `samplers_`.

sharwell:
Superseded by #4551.

PR #4546, opened 2019-12-08
Pass by reference when null is not expected

sharwell:


codemzs:
Thanks but can you please the rationale for passing by reference instead of pointers?

sharwell:
The method uses the object without checking it for null, so it's clear that it never expects a null pointer as input. References are an idiomatic way of expressing intent in C++ that the input is not expected to be null, and is used by the method but not stored by the method for later use (i.e. the scope of use extends to the return from the method, and no further).

PR #4547, opened 2019-12-08
Use std::unique_ptr for objects in LdaEngine

sharwell:
Builds on #4546

### DIFF
--- a/src/Native/LdaNative/lda_engine.cpp
+++ b/src/Native/LdaNative/lda_engine.cpp
@@ -68,11 +68,11 @@ namespace lda {
         printf("using %d thread(s) to do train/test\n", num_threads_);
 
         bAlphaSumMultiplied = false;
-        atomic_stats_ = new LDAEngineAtomics();
-        model_block_ = new LDAModelBlock();
-        data_block_ = new LDADataBlock(num_threads_);
-        process_barrier_ = new SimpleBarrier(num_threads_);
-        samplerQueue_ = new CBlockedIntQueue();
+        atomic_stats_.reset(new LDAEngineAtomics());
+        model_block_.reset(new LDAModelBlock());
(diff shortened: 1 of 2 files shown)


src/Native/LdaNative/lda_engine.cpp
```diff
@@ -123,20 +123,11 @@ namespace lda {
...
-        delete model_block_;
-        model_block_ = nullptr;
-
-        delete samplerQueue_;
-        samplerQueue_ = nullptr;
+        process_barrier_.reset(nullptr);
+        data_block_.reset(nullptr);
+        atomic_stats_.reset(nullptr);
+        model_block_.reset(nullptr);
+        samplerQueue_.reset(nullptr);
```
sharwell: 📝 These calls to `reset` are _likely_ not required, but by including them it ensures the objects are destroyed in the same order the old code destroyed them.

src/Native/LdaNative/lda_engine.cpp
```diff
@@ -68,11 +68,11 @@ namespace lda {
...
-        atomic_stats_ = new LDAEngineAtomics();
-        model_block_ = new LDAModelBlock();
-        data_block_ = new LDADataBlock(num_threads_);
-        process_barrier_ = new SimpleBarrier(num_threads_);
-        samplerQueue_ = new CBlockedIntQueue();
+        atomic_stats_.reset(new LDAEngineAtomics());
+        model_block_.reset(new LDAModelBlock());
+        data_block_.reset(new LDADataBlock(num_threads_));
+        process_barrier_.reset(new SimpleBarrier(num_threads_));
+        samplerQueue_.reset(new CBlockedIntQueue());
```
sharwell: 📝 Using `reset` instead of `std::make_shared` because some of the builds only have C++ 11 (the latter is C++ 14).

PR #4551, opened 2019-12-08
Use std::unique_ptr for samplers_ and likelihood_in_iter_

sharwell:
Fixes an access violation in native code if `LdaEngine` is deleted before `samplers_` is initialized.

### DIFF
--- a/src/Native/LdaNative/lda_engine.cpp
+++ b/src/Native/LdaNative/lda_engine.cpp
@@ -53,7 +53,6 @@ namespace lda {
         mh_step_(mhstep),
         alpha_sum_(alphaSum),
         maxDocToken_(maxDocToken),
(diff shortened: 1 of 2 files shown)


src/Native/LdaNative/lda_engine.cpp
```diff
@@ -129,11 +124,7 @@ namespace lda {
         model_block_.reset(nullptr);
         samplerQueue_.reset(nullptr);
 
-        for (int i = 0; i < num_threads_; ++i)
-        {
-            delete samplers_[i];
-        }
-        delete[] samplers_;
```
sharwell: 📝 This block was responsible for some of the crashes observed in CI, and especially the crashes prior to 0cbfa54. The destructor is missing a null check for samplers_.

PR #4552, opened 2019-12-08
WIP Log exceptions thrown within finalizers

sharwell:
Exceptions thrown from a finalizer will terminate the process.

### DIFF
--- a/src/Microsoft.ML.Core/Data/Repository.cs
+++ b/src/Microsoft.ML.Core/Data/Repository.cs
@@ -145,8 +145,16 @@ namespace Microsoft.ML
 
         ~Repository()
         {
-            if (!Disposed)
-                Dispose(false);
+            try
+            {
+                if (!Disposed)
+                    Dispose(false);
+            }
+            catch (Exception ex) when (!Environment.HasShutdownStarted)
+            {
+                Console.WriteLine($"Exception thrown in call to {GetType()}.Finalize:{Environment.NewLine}{ex}");
+                throw;
+            }
         }
 
         public void Dispose()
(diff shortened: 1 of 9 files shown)


src/Microsoft.ML.Core/Data/Repository.cs
```diff
@@ -145,8 +145,14 @@ private static bool EnsureDirectory(string path)
...
         ~Repository()
         {
-            if (!Disposed)
+            try
+            {
                 Dispose(false);
+            }
+            catch (Exception ex) when (!Environment.HasShutdownStarted)
+            {
+                Console.Error.WriteLine($"Exception thrown in call to {GetType()}.Finalize:{Environment.NewLine}{ex}");
```
harishsk: @eerhardt had suggested that we shouldn't be writing to the console from libraries. Is there a better way we can log the errors? Maybe using the Channel.Trace or similar functions?

sharwell: If we log this message, the process is already in the process of crashing, which is arguably worse than logging from the library.

sharwell: 📝 In the first iteration of this pull request, the new `catch` blocks did not rethrow the exception. I have since updated the pull request in two ways:

* Use `Console.WriteLine` instead of `Console.Error.WriteLine`. Only the former is captured by whatever strategy we are using for running tests in CI, which is where we really need to be able to see the messages.
* I added `throw;` statements to rethrow the exception. This means the pull request will not add stability in cases where these exceptions were already being thrown, but instead merely aims to give us data necessary to identify and resolve the original source of the exception.

sharwell:
Marking as 🚧 to see if we can instead enable stderr logging in the builds, which should already contain this information.

harishsk:
@sharwell Can you please fix the merge conflicts and resubmit?

sharwell:
@frank-dong-ms Are heap dumps working if the test process crashes? If so this change is not necessary.

frank-dong-ms:
Yes, it is working but only on windows.

---
In reply to: [582936269](https://github.com/dotnet/machinelearning/pull/4552#issuecomment-582936269) [](ancestors = 582936269)

PR #4560, opened 2019-12-10
Add netcoreapp3.0 output for SsaForecast test

sharwell:


test/Microsoft.ML.TimeSeries.Tests/TimeSeriesDirectApi.cs
```diff
@@ -361,9 +361,15 @@ public void SsaForecast()
             // Get predictions
             var enumerator = env.Data.CreateEnumerable<ForecastPrediction>(output, true).GetEnumerator();
             ForecastPrediction row = null;
+#if NETCOREAPP3_0
```
eerhardt: What happens when we want to run the tests on .NET Core 5?

eerhardt: No, I guess when we start running on 5, we would just need to add `|| NETCOREAPP5_0` here.

sharwell: I could invert the condition if we think newer .NET will match the behavior of netcoreapp3.0.

eerhardt: At first I thought this was the formatting change we made in 3.0 - https://devblogs.microsoft.com/dotnet/floating-point-parsing-and-formatting-improvements-in-net-core-3-0/. But looking at it deeper, this doesn't appear to be the cause.

@tannergooding - do you happen to know of a change in 3.0 that causes this difference?

tannergooding: What math functions does ssa use under the covers?

harishsk:
@sharwell:This test has been fixed by @frank-dong-ms  and includes the different results for netcoreapp3.0. So I am closing this PR. Please reopen if necessary.

PR #4561, opened 2019-12-10
Add IterationDataAttribute

sharwell:


test/Microsoft.ML.TestFrameworkCommon/Attributes/IterationDataAttribute.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Reflection;
+using Xunit.Sdk;
+
+namespace Microsoft.ML.TestFrameworkCommon.Attributes
```
sharwell: We should consider consolidating the namespaces Microsoft.ML.TestFramework and Microsoft.ML.TestFrameworkCommon, but I left that for another change.

PR #1008, opened 2018-09-24
Move towards being able to build for x86

tannergooding:
This is work towards https://github.com/dotnet/machinelearning/issues/97

It allows you to build using `build.cmd -buildArch=x86` and, if you manually set the correct testhost, it allows the tests to pass.

PR #624, opened 2018-08-01
Port SymSGD trainer

codemzs:
This change adds parallel SGD trainer but disables its multi-threading capabilities because of the lack of OpenMP support by the Clang compiler on linux and macOS build systems and MKL library. It also Supersedes PR #556 by squashing all the commits in one and rebasing with master.

fixes #623

PR #4569, opened 2019-12-12
Enable Conditional Numerical Reproducibility for tests

sharwell:


test/BaselineOutput/SingleDebug/MulticlassLogisticRegression/MulticlassLogisticRegression-TrainTest-iris-tree-featurized-out.txt
```diff
@@ -51,19 +51,19 @@ Virtual memory usage(MB): %Number%
...
-[6] (%Time%)	0 iterations	Loss: 1.0986123085022
-[6] (%Time%)	1 iterations	Loss: 0.62896740436554	Improvement: 0.4696
-[6] (%Time%)	2 iterations	Loss: 0.213765218853951	Improvement: 0.4261
-[6] (%Time%)	3 iterations	Loss: 0.144495338201523	Improvement: 0.1542
-[6] (%Time%)	4 iterations	Loss: 0.106832779943943	Improvement: 0.06646
-[6] (%Time%)	5 iterations	Loss: 0.0936193987727165	Improvement: 0.02649
-[6] (%Time%)	6 iterations	Loss: 0.083323560655117	Improvement: 0.01434
-[6] (%Time%)	7 iterations	Loss: 0.0787383615970612	Improvement: 0.007024
-[6] (%Time%)	8 iterations	Loss: 0.0768212750554085	Improvement: 0.003194
+[6] (%Time%)	0 iterations	Loss: 1.0986123085021973
```
sharwell: 💭 This pull request is only still marked as a draft because I'm not sure how I feel about changing this baseline. I'm going to try and change only the SymSGD baseline and see if the tests still pass. The one here is known to be flaky even with my changes.

sharwell: This change was subsequently removed from the pull request.

sharwell:
I'm going to submit some parts of this as separate pull requests, and then rebase to focus it and fix conflicts.

test/Microsoft.ML.Predictor.Tests/TestPredictors.cs
```diff
@@ -283,6 +283,11 @@ public void BinaryClassifierLogisticRegressionTest()
         [Trait("Category", "SkipInCI")]
```
antoniovs1029: ❓  I guess I am missing something... but shouldn't this line be removed in order to check that the updated baselines are actually correct?

sharwell: These baselines haven't changed in the master branch since I first implemented this pull request. I'm assuming the new values (which were correct then) are still correct now. They would be validated as part of #4722.

sharwell:
Based on the results of #4722 as a validation, the current plan is:

1. Merge this PR, since the baselines here are a clear statistical improvement
2. As a separate PR, add baselines for Windows x86, Centos (and Ubuntu separately if necessary), and MacOS
3. Update #4722 to determine the new status

antoniovs1029:
For the record, the validation @sharwell has just mentioned, consisted in running 100 times the `BinaryClassifierSymSgdTest` test on _every_ build configuration. This was to validate the changes he made to the baseline in here, which is used [exclusively?] on that test which is (at the moment of writing) disabled.

It passed on every Windows x64 build, except for one, where it failed only 1 out of the 100 iterations.

It failed on every other build configuration, but it is thought to be because CNR works differently on each OS and configuration, and then it would be necessary to create a new baseline for each one of these. Notice that that test was disabled recently because it was very flaky on most of our CI builds, and that it was previously disabled on MacOS and x86 since #624 and #1008 respectively. So when doing the baselines for the other build configurations, it would be to keep in mind this, and see if that now that CNR is enabled, these pending test scenarios are now resolved.

antoniovs1029:
After briefly discussing it with @harishsk , I will only run all the tests (including all the disabled ones) with this changes, before letting this PR in. Just to double check. 😄

antoniovs1029:
@harishsk has expressed that he doesn't know if it's ok to have a different baselines for each build configuration, and would like @eerhardt , @justinormont , and @tannergooding opinion on this regard, before accepting enabling the CNR.

antoniovs1029:
So I have just ran this again on #4722 , and actually now even the Windows x64 are failing, it fails in all 100 iterations of the `BinaryClassifierSymSgdTest` test. It doesn't seem to me that any update on master during the weekend could have changed the baseline here again, so I am not sure why this is happening.

sharwell:
@antoniovs1029 the baselines were altered in #4710

antoniovs1029:
> @antoniovs1029 the baselines were altered in #4710

You're right. On that PR it's added that the BinaryClassifierSymSGD should also print its bias and weights into the output. So it will only be necessary to add a couple of lines to the baseline.

After running it on the CI, I believe the updated baseline (for Windows x64) should be

_on SymSGD-CV-breast-cancer-out.txt_
```
Bias: -467.9297, Weights: [5.415065,76.39395,22.351555,-11.988394,-28.264456,44.58415,22.720116,11.132539,2.851256]
Not training a calibrator because it is not needed.
Not adding a normalizer.
Data fully loaded into memory.
Initial learning rate is tuned to 100.000000
Bias: -484.28625, Weights: [-12.787148,140.42905,121.938194,37.527378,-129.81407,70.90587,-89.37068,81.64309,-32.32781]
```

On SymSGD-TrainTest-breast-cancer-out.txt
`Bias: -473.82773, Weights: [1.8576517,52.924484,9.490602,14.593927,-17.657146,10.126332,-8.042431,1.9634984,26.753149]`

------------------------

By the way, I've gotten baselines for the other platforms and made a couple of experiments with them. So far it seems that enabling CNR does yield more consistent results from the test, although different baselines would indeed be necessary for each platform. I still need to do further experiments on this, though.

sharwell:
@antoniovs1029 baselines have been updated

Issue #4528, opened 2019-12-06
PredictionEnginePool.GetPredictionEngine is not thread safe


thomasrosdahl:
### System information

- **OS version/distro**: Windows 10
- **.NET Version (eg., dotnet --info)**: Azure Functions Runtime 2.7.1948

### Issue

- **What did you do?**
Invoked `PredictionEnginePool.Predict("MyModelName", example)` from multiple threads.

- **What happened?**
`System.ArgumentException: 'An item with the same key has already been added. Key: MyModelName'`

- **What did you expect?**
Method is thread safe so that multiple named pools cannot be created simultaneously.

### Source code / logs

   at System.ThrowHelper.ThrowAddingDuplicateWithKeyArgumentException[T](T key)
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at Microsoft.Extensions.ML.PredictionEnginePool`2.GetPredictionEngine(String modelName)
   at Microsoft.Extensions.ML.PredictionEnginePoolExtensions.Predict[TData,TPrediction](PredictionEnginePool`2 predictionEnginePool, String modelName, TData example)

gvashishtha:
@frank-dong-ms

frank-dong-ms:
Fix has been checked in.

PR #4570, opened 2019-12-12
fix issue 4528, use thread safe ConcurrentDictionary instead of Dictionary

frank-dong-ms:
fix issue #4528

### DIFF
--- a/src/Microsoft.Extensions.ML/PredictionEnginePool.cs
+++ b/src/Microsoft.Extensions.ML/PredictionEnginePool.cs
@@ -80,9 +80,11 @@ namespace Microsoft.Extensions.ML
         /// </param>
         public PredictionEngine<TData, TPrediction> GetPredictionEngine(string modelName)
         {
-            if (_namedPools.ContainsKey(modelName))
+            PoolLoader<TData, TPrediction> existingPool = null;
+            _namedPools.TryGetValue(modelName, out existingPool);
+            if (_namedPools.TryGetValue(modelName, out existingPool))
             {
-                return _namedPools[modelName].PredictionEnginePool.Get();
+                return existingPool.PredictionEnginePool.Get();
             }
 
             //This is the case where someone has used string.Empty to get the default model.
@@ -101,7 +103,7 @@ namespace Microsoft.Extensions.ML
             //Here we are in the world of named models where the model hasn't been built yet.
             var options = _predictionEngineOptions.Create(modelName);
             var pool = new PoolLoader<TData, TPrediction>(_serviceProvider, options);
-            _namedPools.TryAdd(modelName, pool);
+            pool = _namedPools.GetOrAdd(modelName, pool);
             return pool.PredictionEnginePool.Get();
         }
 
--- a/src/Microsoft.Extensions.ML/PredictionEnginePool.cs
+++ b/src/Microsoft.Extensions.ML/PredictionEnginePool.cs
@@ -80,8 +80,7 @@ namespace Microsoft.Extensions.ML
         /// </param>
         public PredictionEngine<TData, TPrediction> GetPredictionEngine(string modelName)
         {
-            PoolLoader<TData, TPrediction> existingPool = null;
-            if (_namedPools.TryGetValue(modelName, out existingPool))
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.ML/PredictionEnginePool.cs
```diff
@@ -100,7 +101,7 @@ public ITransformer GetModel()
             //Here we are in the world of named models where the model hasn't been built yet.
             var options = _predictionEngineOptions.Create(modelName);
             var pool = new PoolLoader<TData, TPrediction>(_serviceProvider, options);
-            _namedPools.Add(modelName, pool);
+            _namedPools.TryAdd(modelName, pool);
```
sharwell: 📝 This method should use `_namedPools.TryGetValue(modelName, out existingPool)` at the start of the method, rather than `ContainsKey` followed by indexing.

sharwell: ⚠️  This implementation can end up using multiple pools for the same key. If you don't want that to happen, use `GetOrAdd` instead of `TryAdd`.

```suggestion
            pool = _namedPools.GetOrAdd(modelName, pool);
```

frank-dong-ms: Can you explain in detail the reason?

---
In reply to: [357425080](https://github.com/dotnet/machinelearning/pull/4570#discussion_r357425080) [](ancestors = 357425080)

frank-dong-ms: - This implementation can end up using multiple pools for the same key
How can this happen? 
What's the difference between TryAdd and GetOrAdd?

---
In reply to: [357425395](https://github.com/dotnet/machinelearning/pull/4570#discussion_r357425395) [](ancestors = 357425395)

sharwell: > How can this happen?

Two threads call `TryAdd` at the same time with different pools. Only one will end up in `_namedPools`, yet each thread uses the pool it created itself.

> What's the difference between TryAdd and GetOrAdd?

In the presence of multiple callers, `TryAdd` will use the value from the first and ignore the others. `GetOrAdd` will atomically add the value or return the existing value.

sharwell: In this case, it's just more efficient. `ContainsKey` is doing a lookup, but the lookup value is ignored and then the indexer does exactly the same thing.

frank-dong-ms: I see, thanks for explain!

---
In reply to: [357434533](https://github.com/dotnet/machinelearning/pull/4570#discussion_r357434533) [](ancestors = 357434533)

frank-dong-ms: Got it, thanks

---
In reply to: [357434839](https://github.com/dotnet/machinelearning/pull/4570#discussion_r357434839) [](ancestors = 357434839)

src/Microsoft.Extensions.ML/PredictionEnginePool.cs
```diff
@@ -79,9 +80,11 @@ public ITransformer GetModel()
         /// </param>
         public PredictionEngine<TData, TPrediction> GetPredictionEngine(string modelName)
         {
-            if (_namedPools.ContainsKey(modelName))
+            PoolLoader<TData, TPrediction> existingPool = null;
+            _namedPools.TryGetValue(modelName, out existingPool);
+            if (_namedPools.TryGetValue(modelName, out existingPool))
```
sharwell: 💡 All three of these lines can be reduced to just one:

```csharp
if (_namedPools.TryGetValue(modelName, out var existingPool))
```

frank-dong-ms: Good to know, thanks Sam

---
In reply to: [357439017](https://github.com/dotnet/machinelearning/pull/4570#discussion_r357439017) [](ancestors = 357439017)

PR #4576, opened 2019-12-13
Fixes #4571. About memory leak when using FeaturizeText.

### LINKED ISSUE #4571: Memory leak when featurizing text with the default settings
When featurizing text with the default settings, references to the entire dataset rows are kept around

antoniovs1029:
### Updated first post:
Originally this PR included different proposals to solve issue #4571 , which are described below. From all of them, proposal 3 was chosen, and now this PR only includes the changes of that Proposal.

----------------------------------

### Original first post:
In this PR I include 3 proposals to solve issue #4571:
1. Modifying NormStr constructor
- OR -
2. Modifying TextNormalizer
- OR -
3. Modifying NormStr.Pool Get Method

(the proposals are mutually exclusive, in that we are supposed to choose one of those, but I've included the three in this PR so that we can discuss about them, since @harishsk and I are unsure about the overall impact of each one, please let us know your opinion, and if there's anything else I should test)

For reference: Offline, @daholste provided me with a dataset which I can't share in here because of privacy reasons. It had 100k rows and 2.5k columns. When featurizing 22 of those columns, he got 22 NormStr Pools which, together, had an inclusive size of 2.25 GB, because many entire rows where held in memory, mainly because of the reason I described in [this comment](https://github.com/dotnet/machinelearning/issues/4571). The text that follows assume experiments based on that dataset, finding n-grams of size 1.

When using Proposal 1 the inclusive size of the NormStr Pools went down from 2.25GB to 22MB. When using Proposal 2 it went down to 30MB. With Proposal 3 it went down to 22MB. With any of the proposals, or without them, the time for fitting was around 40 seconds on my machine.

All of the proposals are based in the idea of creating a `ReadOnlyMemory<char>` that doesn't have an `_object` member holding a reference to the string of the whole row.

### Proposal 1 - Modifying NormStr constructor
[Link to proposal](https://github.com/dotnet/machinelearning/pull/4576#discussion_r357827889)
- Each NormStr has _object that only contains the text that NormStr is representing, and it's saved as a char[]
- With @daholste's dataset 260k NormStrs are created, and each one of those holds a reference to the corresponding char[]
- When fitting with this proposal, this was the Process Memory, notice the oscillations:
image

@harishsk  and I think the oscillations in memory usage are caused by the strings that are created with this implementation, but that are then garbage collected every once in a while, because no actual reference to them is kept.
- All the references to entire rows were gone

### Proposal 2 - Modifying Text Normalizer
[Link to proposal](https://github.com/dotnet/machinelearning/pull/4576#discussion_r357827979)
- _In general_, each NormStr has a _object of type string, which holds the content of the entire column where the NormStr was found.
- With the dataset, 260k NormStrs are created, and less than 120k strings are held in memory.
- When fitting with this proposal, this was the Process Memory graph, notice no oscillations:
image

- ~~Some references to entire rows were still there (I am still figuring out why). But there were only a few of them, compared to when there was no fix; as it is shown in the decrease in size of the NormStr pools I mentioned (from 2.2GB to 30MB), this proposal is still useful, at least for @daholste case.~~ All the references to strings containing entire rows are gone.

### Proposal 3 - Modifying NormStr.Pool Get Method
[Link to proposal](https://github.com/dotnet/machinelearning/pull/4576#discussion_r357871260)
- Each NormStr has _object that only contains the text of the string that NormStr is representing
- With the dataset, 260k NormStr were created, each one holding a reference to their corresponding string.
- When fitting with this proposal, this was the Process Memory, notice the oscillations:
image
- All the references to entire rows were gone


src/Microsoft.ML.Core/Utilities/NormStr.cs
```diff
@@ -29,7 +29,7 @@ internal sealed class NormStr
         private NormStr(ReadOnlyMemory<char> str, int id, uint hash)
         {
             Contracts.Assert(id >= 0 || id == -1 && str.IsEmpty);
-            Value = str;
+            Value = new ReadOnlyMemory<char>(str.ToString().ToCharArray());
```
antoniovs1029: This is Proposal 1

src/Microsoft.ML.Transforms/Text/TextNormalizing.cs
```diff
@@ -398,7 +398,8 @@ private void NormalizeSrc(in ReadOnlyMemory<char> src, ref ReadOnlyMemory<char>
                 if (min == 0)
                 {
                     Host.Assert(src.Length == len);
-                    dst = src;
+                    buffer.AppendSpan(span);
+                    dst = buffer.ToString().AsMemory();
                 }
```
antoniovs1029: This is Proposal 2

src/Microsoft.ML.Core/Utilities/NormStr.cs
antoniovs1029: I am just realizing that one downside of this proposal would be that when finding n-grams bigger than 1, or char-grams, this proposal would take a lot of memory unnecessarily repeating strings.

sharwell: 💭 The call to `ToCharArray()` is unnecessary right?

https://github.com/dotnet/runtime/blob/71d7e6cd72e9d4154f595cc43adac4b66ca53e48/src/libraries/System.Private.CoreLib/src/System/MemoryExtensions.cs#L164-L173

sharwell:
I would offer proposal 3: only reconstruct strings when they are actually getting added to the pool. We can still use sub-arrays (spans of larger strings) to avoid allocations when looking things up in the pool. To implement proposal 3, you only need to change `str` to `str.ToString().AsMemory()` on this line:

https://github.com/dotnet/machinelearning/blob/fa62c8196e345de82918469b56c2c6c3b16321c1/src/Microsoft.ML.Core/Utilities/NormStr.cs#L139

src/Microsoft.ML.Core/Utilities/NormStr.cs
antoniovs1029: Thanks for the comment @sharwell 

In fact I am just realizing that it might be better to simply use
`Value = str.ToString().AsMemory();`

antoniovs1029: So I've just had a conversation with @justinormont and he does prefer this proposal instead of the other one, since the other one would only work for pipelines that use TextNormalizer, but not every pipeline does that.

He does have some other comments on how this proposal should be done... The most relevant which is that this "str.ToString().AsMemory()" (i.e., creating a new ReadOnlyMemory<char> so that the reference to the whole row is lost) should be left as an option for the user. This would imply adding a flag to several methods, up to the FeturizeText method, and adding the corresponding documentation about the option. So what do people think?

sharwell: Can you review my proposal 3 below too?

sharwell:
@antoniovs1029 I would suggest running the test again on proposal 3. Since it only allocates on a cache _miss_, it may substantially lower the allocation churn you observed when testing proposal 1.

src/Microsoft.ML.Core/Utilities/NormStr.cs
antoniovs1029: Yeah, so I would replace proposal 1 with proposal 3, since @justinormont also thought what is done in proposal 3 would be better.

Still, the question remains about if we should make it optional what is done in proposal 3, having to add a new flag up to FeaturizeText.

antoniovs1029:
> @antoniovs1029 I would suggest running the test again on proposal 3. Since it only allocates on a cache miss, it may substantially lower the allocation churn you observed when testing proposal 1.

I've done it. But it seems to show the same behavior as Proposal 1. Still I've updated the first post with those findings, and made some corrections to the results of Proposal 2.

src/Microsoft.ML.Core/Utilities/NormStr.cs
```diff
@@ -136,7 +136,7 @@ public NormStr Get(ReadOnlyMemory<char> str, bool add = false)
                 }
                 Contracts.Assert(ins == -1);
 
-                return add ? AddCore(str, hash) : null;
+                return add ? AddCore(str.ToString().AsMemory(), hash) : null;
```
antoniovs1029: This is Proposal 3

src/Microsoft.ML.Core/Utilities/NormStr.cs
antoniovs1029: You can find Proposal 3 [here](https://github.com/dotnet/machinelearning/pull/4576/files#r357871260). And I've also updated the first post of this issue with the results of that proposal (which were pretty much the same as Proposal 1 for this specific case).

src/Microsoft.ML.Core/Utilities/NormStr.cs
antoniovs1029: @sharwell and @justinormont proposed this. @justinormont also suggests adding a flag so that this conversion is optional, and that flag should be added to different methods up to FeaturizeText, where it should become an option for the user.

justinormont: `NormStr` being an internal class wouldn't expose the flag to users, only to the ML.NET code. Giving the ML.NET coders the choice to re-allocate the specific substring, or continue (potentially) saving memory and allocation time by using the original `NormStr`.

sharwell: I'm happy with not adding an option here, provided we do add a benchmark test.

harishsk: @sharwell's idea sounds the best to me. Please proceed with this and amend the PR to include only the proposal 3 changes.

antoniovs1029: > NormStr being an internal class wouldn't expose the flag to users, only to the ML.NET code. Giving the ML.NET coders the choice to re-allocate the specific substring, or continue (potentially) saving memory and allocation time by using the original NormStr.

@justinormont My bad, I guess I misunderstood you. I thought you wanted to add this option all the way up to FeaturizeText, which is a public method, thus exposing the option to users. Since that wasn't what you suggested, then how up should this option go? I'd think it'd be okay for now to only create the flag `createNewStr` in the Get method, and default it to `true`, and only leave that option there, unless in the future we see the necessity to add this option to whatever methods call the `Get `method.

src/Microsoft.ML.Core/Utilities/NormStr.cs
```diff
@@ -136,6 +136,14 @@ public NormStr Get(ReadOnlyMemory<char> str, bool add = false)
                 }
                 Contracts.Assert(ins == -1);
 
+                if(createNewStr)
```
justinormont:
```suggestion
                if (createNewStr)
```

src/Microsoft.ML.Core/Utilities/NormStr.cs
```diff
@@ -116,7 +116,7 @@ public NormStr Get(string str, bool add = false)
                 return add ? AddCore(str.AsMemory(), hash) : null;
             }
 
-            public NormStr Get(ReadOnlyMemory<char> str, bool add = false)
+            public NormStr Get(ReadOnlyMemory<char> str, bool add = false, bool createNewStr = true)
```
justinormont: You'll want to expose the same flag in the Add() methods, and any other callers of this function within the NormStr class.

Speaking of, you may want to investigate which outside components call this interface to see if their memory usage / runtime will increase when defaulting to `true`.

src/Microsoft.ML.Core/Utilities/NormStr.cs
```diff
@@ -136,6 +136,14 @@ public NormStr Get(ReadOnlyMemory<char> str, bool add = false)
                 }
                 Contracts.Assert(ins == -1);
 
+                if(createNewStr)
+                {
+                    // To avoid the case where 'str' actually stores a string with the
+                    // content of a whole row in the dataset, a new 'str' is created
+                    // See issue #4571 and PR #4576
```
justinormont: The shortlink issue/PR number won't survive repo forks, and the unrelenting march of time. You may want full links.
```suggestion
                    // See issue https://github.com/dotnet/machinelearning/issues/4571 
                    // and PR https://github.com/dotnet/machinelearning/pull/4576
```

src/Microsoft.ML.Core/Utilities/NormStr.cs
justinormont: nit -- Naming: I can foresee confusion between the two flags, `add` and `createNewStr`. Since they could both be interpreted as "if my key is missing, insert (add/create) it".

We want to convey that the string will be reallocated and no longer hold open a pointer to the original string. Perhaps as `duplicateStr` or invert as `keepAsReference`.

```suggestion
            public NormStr Get(ReadOnlyMemory<char> str, bool add = false, bool duplicateStr = true)
```

src/Microsoft.ML.Core/Utilities/NormStr.cs
antoniovs1029: So Proposal 3 was chosen. So I will resolve this thread.

src/Microsoft.ML.Transforms/Text/TextNormalizing.cs
antoniovs1029: So Proposal 3 was chosen. So I will resolve this thread.

test/Microsoft.ML.Benchmarks/FeaturizeTextBench.cs
```diff
@@ -0,0 +1,136 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
antoniovs1029: @sharwell Please, review this benchmark test and tell me if this was what you hand in mind when you requested to add one

test/Microsoft.ML.Benchmarks/FeaturizeTextBench.cs
```diff
@@ -0,0 +1,139 @@
...
+            IEstimator<ITransformer>  pipeline = featurizers.First();
+            foreach (var featurizer in featurizers.Skip(1))
+            {
+                pipeline = pipeline.Append(featurizer);
+            }
+
+            var model = pipeline.Fit(dataset);
+            var memoryUsage = GC.GetTotalMemory(true);
+            Console.WriteLine($"Memory Used: {memoryUsage/1000000:0,0.00}MB");
+            Assert.True(memoryUsage < 1000000000, $"This benchmark should use less than 1GB of memory, but it's using {memoryUsage/1000000:0,0.00}MB"); // Memory usage should be less than 1GB after PR #4576
```
antoniovs1029: @harishsk this is the assertion you asked me to add to the benchmark. In my machine, without the changes made in this PR, the value of `memoryUsage` is over 2.5GB... with the changes of this PR it's less than 200MB. So I didn't know where to put the threshold, and I simply chose 1GB. Don't know if that's ok to you. Thanks.

harishsk: 1GB maybe a bit high. How about 2x the memory usage you are seeing on your machine? If memory usage on a different machine exceeds two times what you are seeing on your machine, then it is reason to investigate what is happening.

antoniovs1029: Ok, so I will change it into 400MB. Thanks!

src/Microsoft.ML.Core/Utilities/NormStr.cs
antoniovs1029: The only method inside the `NormStr` class that calls `Get()` with `add = true` is the Add() method you mention. So that's going to be the only other method where I will add the flag.

About your other suggestion, I will discuss with @harishsk about where to investigate. But it seems to me that the Add() method is _only_ used in 2 other classes:
1. The `ValueToKeyMapperTransformer.Builder.TextImpl` , which is the class that holds the _pool that I've been working with in this issue. It might be worth exploring cases where the VTKMT would have a worse memory usage with the changes introduced in this PR. I've tried with a couple of different datasets, but I haven't seen this being an issue, and I don't know which kind of cases to explore.
2. The `ModelSaveContext`, where strings are serialized to disk as indices, and they're added to a NormStr.Pool that is later serialized. In this case I don't see why the changes in this PR could cause a problem in its memory usage/runtime (because the changes of this PR would only execute once, when the `str` is not already in the pool).

justinormont: Thanks for checking for other uses which would be affected.

test/Microsoft.ML.Benchmarks/FeaturizeTextBench.cs
antoniovs1029: Sorry, when running the test on my machine memoryUsage is actually, at most 120MB, so I will fix the threshold into 240MB.

sharwell: @adamsitnik Is there a way to make assertions about the stabilized outcome of the benchmark?

test/Microsoft.ML.Benchmarks/FeaturizeTextBench.cs
```diff
@@ -0,0 +1,139 @@
...
+                Columns = columns.ToArray(),
+                HasHeader = false,
+                Separators = new char[] { ',' }
+            });
+
+            dataset = textLoader.Load(path);
+        }
+
+        [Benchmark]
+        public ITransformer TrainFeaturizeText()
```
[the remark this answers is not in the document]
antoniovs1029: Well, varying the number of rows generated for the dataset is the first thing I can think of. Would you recommend adding more benchmarks for different sizes?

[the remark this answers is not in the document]
antoniovs1029: Not sure what the sample output should look like, since in this sample I am actually not generating an output, only a transformer that featurizes the text.

I could add an step where I actually use the transformer to transform some input data (perhaps a single row of the dataset). But then again, the output of the featurizer would be a _big_ vector of floats representing random strings.

So I wouldn't think it's worth it to do this.

PR #4577, opened 2019-12-14
Added onnx export support for VectorWhitening

Lynx1820:

### DIFF
--- a/src/Microsoft.ML.Mkl.Components/VectorWhitening.cs
+++ b/src/Microsoft.ML.Mkl.Components/VectorWhitening.cs
@@ -649,22 +649,12 @@ namespace Microsoft.ML.Transforms
                 Host.Assert(parameters.Kind == WhiteningKind.PrincipalComponentAnalysis || parameters.Kind == WhiteningKind.ZeroPhaseComponentAnalysis);
 
                 int rank = (parameters.Kind == WhiteningKind.PrincipalComponentAnalysis && parameters.Rank > 0) ? parameters.Rank : dimension;
+                Host.CheckParam(rank <= dimension, nameof(rank), "Rank must be at most the dimension of untransformed data.");
+
                 long[] modelDimension = { rank, dimension };
 
-                if (rank != dimension)
-                {
-                    float[] principalComponents = new float[rank * dimension];
-                    for (int i = 0; i < parameters.Rank; i++)
-                    {
-                        if (i >= dimension)
-                            break;
-                        Array.Copy(model, i * dimension, principalComponents, i * dimension, dimension);
-                    }
-                    model = principalComponents;
-                }
-
                 var opType = "Gemm";
-                var modelName = ctx.AddInitializer(model, modelDimension, "model");
+                var modelName = ctx.AddInitializer(model.Take(rank * dimension), modelDimension, "model");
                 var zeroValueName = ctx.AddInitializer((float)0);
 
                 var node = ctx.CreateNode(opType, new[] { modelName, srcVariableName, zeroValueName }, new[] { dstVariableName }, ctx.GetNodeName(opType), "");
--- a/src/Microsoft.ML.Mkl.Components/VectorWhitening.cs
+++ b/src/Microsoft.ML.Mkl.Components/VectorWhitening.cs
@@ -3,7 +3,6 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Security;
@@ -620,12 +619,13 @@ namespace Microsoft.ML.Transforms
                     };
                 return del;
             }
+
             public bool CanSaveOnnx(OnnxContext ctx) => true;
 
             public void SaveAsOnnx(OnnxContext ctx)
             {
                 Host.CheckValue(ctx, nameof(ctx));
-                int numColumns = _parent.ColumnPairs.Length; //why are the multiple columns
+                int numColumns = _parent.ColumnPairs.Length;
                 for (int iinfo = 0; iinfo < numColumns; ++iinfo)
                 {
                     string inputColumnName = _parent.ColumnPairs[iinfo].inputColumnName;
@@ -637,8 +637,8 @@ namespace Microsoft.ML.Transforms
                     string dstVariableName = ctx.AddIntermediateVariable(_srcTypes[iinfo], outputColumnName, true);
                     SaveAsOnnxCore(ctx, iinfo, srcVariableName, dstVariableName);
                 }
-
             }
+
             private void SaveAsOnnxCore(OnnxContext ctx, int iinfo, string srcVariableName, string dstVariableName)
             {
                 var model = _parent._models[iinfo];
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Mkl.Components/VectorWhitening.cs
```diff
@@ -617,6 +620,57 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
                     };
                 return del;
             }
+            public bool CanSaveOnnx(OnnxContext ctx) => true;
```
yaeldekel: >public [](start = 12, length = 6)

nit: could you add an empty line between methods?

src/Microsoft.ML.Mkl.Components/VectorWhitening.cs
```diff
@@ -617,6 +620,57 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
...
+                        continue;
+
+                    string outputColumnName = _parent.ColumnPairs[iinfo].outputColumnName;
+                    string srcVariableName = ctx.GetVariableName(inputColumnName);
+                    string dstVariableName = ctx.AddIntermediateVariable(_srcTypes[iinfo], outputColumnName, true);
+                    SaveAsOnnxCore(ctx, iinfo, srcVariableName, dstVariableName);
+                }
+
+            }
+            private void SaveAsOnnxCore(OnnxContext ctx, int iinfo, string srcVariableName, string dstVariableName)
```
yaeldekel: here too.

src/Microsoft.ML.Mkl.Components/VectorWhitening.cs
```diff
@@ -617,6 +620,57 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
                     };
                 return del;
             }
+            public bool CanSaveOnnx(OnnxContext ctx) => true;
+
+            public void SaveAsOnnx(OnnxContext ctx)
+            {
+                Host.CheckValue(ctx, nameof(ctx));
+                int numColumns = _parent.ColumnPairs.Length; //why are the multiple columns
```
yaeldekel: >why are the multiple columns [](start = 63, length = 28)

This comment seems partial.

src/Microsoft.ML.Mkl.Components/VectorWhitening.cs
```diff
@@ -617,6 +620,57 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
...
+                    for (int i = 0; i < parameters.Rank; i++)
+                    {
+                        if (i >= dimension)
+                            break;
+                        Array.Copy(model, i*dimension, principalComponents, i*dimension, dimension);
+                    }
+                    model = principalComponents;
+                }
+
+                var opType = "Gemm";
```
yaeldekel: >Gemm [](start = 30, length = 4)

Just out of curiosity: The `Mapper` of this class calls `Mkl.Gemv`, as opposed to `Gemm`. Is the ONNX operation `Gemm` equivalent to the MKL operation `Gemv`?

src/Microsoft.ML.Mkl.Components/VectorWhitening.cs
```diff
@@ -617,6 +620,57 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
...
+                long[] modelDimension = { rank, dimension };
+
+                if (rank != dimension)
+                {
+                    float[] principalComponents = new float[rank * dimension];
+                    for (int i = 0; i < parameters.Rank; i++)
+                    {
+                        if (i >= dimension)
+                            break;
+                        Array.Copy(model, i*dimension, principalComponents, i*dimension, dimension);
```
yaeldekel: >i*dimension [](start = 42, length = 11)

Ctrl-k-d.

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -26,6 +26,7 @@
 using Newtonsoft.Json;
 using Xunit;
 using Xunit.Abstractions;
+using System.Diagnostics;
```
yaeldekel: >Diagnostics [](start = 13, length = 11)

Is this used? If it is, please move it up where the rest of the System.* usings are.

src/Microsoft.ML.Mkl.Components/VectorWhitening.cs
```diff
@@ -617,6 +620,57 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
...
+                var parameters = _parent._columns[iinfo];
+                Host.Assert(parameters.Kind == WhiteningKind.PrincipalComponentAnalysis || parameters.Kind == WhiteningKind.ZeroPhaseComponentAnalysis);
+
+                int rank = (parameters.Kind == WhiteningKind.PrincipalComponentAnalysis && parameters.Rank > 0) ? parameters.Rank : dimension;
+                long[] modelDimension = { rank, dimension };
+
+                if (rank != dimension)
+                {
+                    float[] principalComponents = new float[rank * dimension];
+                    for (int i = 0; i < parameters.Rank; i++)
```
yaeldekel: >for (int i = 0; i < parameters.Rank; i++) [](start = 20, length = 41)

If I understand correctly, this loop copies the first `rank*dimension` elements from `model` to `principalComponents` in the same order. Do we really need to copy?
I noticed that `ctx.AddInitializer` takes an `IEnumerable`, perhaps it would be enough to call it with `model.Take(rank*dimension)`?

Lynx1820: I think this problem with Take is that when the rank is higher than the dimension, ML.NET simply treats the extra eigenvectors as rows of zeros. For ONNX, the model tensor has to be resized. I can introduce a check for both ML.NET and ONNX to check that the rank < dimension?

src/Microsoft.ML.Mkl.Components/VectorWhitening.cs
Lynx1820: Yes, they perform the same operation, with slightly different options.

src/Microsoft.ML.Mkl.Components/VectorWhitening.cs
```diff
@@ -657,8 +657,12 @@ private void SaveAsOnnxCore(OnnxContext ctx, int iinfo, string srcVariableName,
                 var modelName = ctx.AddInitializer(model.Take(rank * dimension), modelDimension, "model");
                 var zeroValueName = ctx.AddInitializer((float)0);
 
-                var node = ctx.CreateNode(opType, new[] { modelName, srcVariableName, zeroValueName }, new[] { dstVariableName }, ctx.GetNodeName(opType), "");
+                var gemmOutput = ctx.AddIntermediateVariable(null, "GemmOutput", true);
```
harishsk: Can you please help me understand the rationale for this?

Lynx1820: The model dimensions are (r x m), where r is the user specified rank. After the gemm operation, you get an r x 1 tensor. I added a transpose to match the expected 1 x r tensor.

Issue #4579, opened 2019-12-16
In CodeGenerator, Recommendation scenario Training cause null reference Exception


LittleLittleCloud:
Just dig in a bit. The null reference exception is because, in the AutoML generated pipeline, one of the fields in the trainer node's property dictionary is null, which crashes CodeGenerator.

LittleLittleCloud:
The root cause for that is in CodeGenerator, it only handle cases when trainer option's value type is built-in type. see the following code link.
https://github.com/dotnet/machinelearning/blob/f1f8942a8272a9c87373d11bc89467461c8ecad1/src/Microsoft.ML.CodeGenerator/CodeGenerator/CSharp/TrainerGeneratorBase.cs#L56

While in this case, the null-value field's type is a `MatrixFactorizationTrainer.LossFunctionType`. So it's been ignored and assigned a null value.

A quick fix can be fairly simple: just pass the string value of `MatrixFactorizationTrainer.LossFunctionType` in sweepable parameters. 
Or we can use reflection to get the right type.

PR #4580, opened 2019-12-16
A quick fix for Recommendation Null Reference bug

LittleLittleCloud:
#4579

### DIFF
--- a/src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
+++ b/src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
@@ -124,7 +124,7 @@ namespace Microsoft.ML.AutoML
                 new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.LearningRate), new object[] { 0.001f, 0.01f, 0.1f }),
                 new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.ApproximationRank), new object[] { 8, 16, 64, 128 }),
                 new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.Lambda), new object[] { 0.01f, 0.05f, 0.1f, 0.5f, 1f }),
-                new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.LossFunction), new object[] { MatrixFactorizationTrainer.LossFunctionType.SquareLossRegression, MatrixFactorizationTrainer.LossFunctionType.SquareLossOneClass }),
+                new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.LossFunction), new object[] { "MatrixFactorizationTrainer.LossFunctionType.SquareLossRegression", "MatrixFactorizationTrainer.LossFunctionType.SquareLossOneClass" }),
                 new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.Alpha), new object[] { 1f, 0.01f, 0.0001f, 0.000001f }),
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
```diff
@@ -124,7 +124,7 @@ public static IEnumerable<SweepableParam> BuildMatrixFactorizationParams()
                 new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.LearningRate), new object[] { 0.001f, 0.01f, 0.1f }),
                 new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.ApproximationRank), new object[] { 8, 16, 64, 128 }),
                 new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.Lambda), new object[] { 0.01f, 0.05f, 0.1f, 0.5f, 1f }),
-                new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.LossFunction), new object[] { MatrixFactorizationTrainer.LossFunctionType.SquareLossRegression, MatrixFactorizationTrainer.LossFunctionType.SquareLossOneClass }),
+                new SweepableDiscreteParam(nameof(MatrixFactorizationTrainer.Options.LossFunction), new object[] { "MatrixFactorizationTrainer.LossFunctionType.SquareLossRegression", "MatrixFactorizationTrainer.LossFunctionType.SquareLossOneClass" }),
```
justinormont: > @LittleLittleCloud [Dec 16, 2019, 3:28 PM PST](https://github.com/dotnet/machinelearning/issues/4579#issuecomment-566296343)
>
> -----
> The root cause for that is in CodeGenerator, it only handle cases when trainer option's value type is built-in type. see the following code link.
> https://github.com/dotnet/machinelearning/blob/f1f8942a8272a9c87373d11bc89467461c8ecad1/src/Microsoft.ML.CodeGenerator/CodeGenerator/CSharp/TrainerGeneratorBase.cs#L56
> 
> While in this case, the null-value field's type is a `MatrixFactorizationTrainer.LossFunctionType`. So it's been ignored and assigned a null value.
> 
> A quick fix can be fairly simple: just pass the string value of `MatrixFactorizationTrainer.LossFunctionType` in sweepable parameters. 
> Or we can use reflection to get the right type.

Will changing this to a string affect our ability to sweep over the values? I think we need the ability pass the native objects for sweeping.

Dmitry-A: can you do a nameof() on those?  It's a bit odd design that I don't think I fully understand why it is done this way but otherwise the fix looks reasonable to me, the whole point of the code that is currently crashing is to convert these to string anyway.

LittleLittleCloud: Oh right it should fail. But when testing it runs pretty well... Now I'm even more confusing. I will dig into more...

LittleLittleCloud:
The way to update one trainer's option is mild so if the type of sweep option value not match, it simply falls back to the default value. instead of throwing an error
https://github.com/dotnet/machinelearning/blob/7628d6cd06a0cf8294ddeebbb9613aaec72ab4e7/src/Microsoft.ML.AutoML/TrainerExtensions/TrainerExtensionUtil.cs#L65

So It's not the right way to fix the bug. Maybe we should fix that in CodeGenerator

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
LittleLittleCloud: Thanks for pointing out, it will cause bug when sweeping parameter, but it doesn't crash pipeline.

Issue #4518, opened 2019-12-03
System.ArgumentOutOfRangeException in NimbusML's NGramFeaturizer2.py caused by CustomStopWordsRemover/Feature_extraction.text.stopwords in ML.NET


mstfbl:
As addressed in this [issue ](https://github.com/microsoft/NimbusML/issues/365) in the NimbusML repo, there is a bug with the ML.NET bindings of CustomStopWordsRemover, which is StopWordsRemovingTransformer.cs. 

The specific error message is: 
RuntimeError: Error: *** System.ArgumentOutOfRangeException: 'dataFile is empty
The full error log can be found [here](https://dev.azure.com/aifx/public/_build/results?buildId=2089&view=logs&j=d0e8f4b8-2f67-5548-290c-4d6f15a1cbca&t=e5950cbb-cc45-5e54-16a6-7bfce2ad073d&l=8236).

gvashishtha:
@frank-dong-ms

frank-dong-ms:
Stop words are not passed properly to ml.net class - CustomStopWordsRemovingTransformer, Piet is helping take a look from NimbusML side

PR #4581, opened 2019-12-17
fix issue-4518

frank-dong-ms:
fix issue #4518 , fix System.ArgumentOutOfRangeException in CustomStopWordsRemover


justinormont:
nit: This could use a better title, like `System.ArgumentOutOfRangeException in CustomStopWordsRemover`, or the longer version of the referenced issue.

Issue #4555, opened 2019-12-09
AutoML Nuget Package won't train against TSV


KiransHub:
### System information

- **OS version/distro**: Win10 64bit
- **.NET Version (eg., dotnet --info)**: 
dotnet core 3.0

### Issue

- **What did you do?** 
Installed AutoML nuget package. Setup sentiment analysis training. Selected correct labels and features columns. 

Went to train and... 

- **What happened?**

I got a message almost immediately "Failed. See more in Output pane". 

The output pane for "Machine Learning" is completely empty! 

- **What did you expect?**
At least an error... 

screenshot:

![image](https://user-images.githubusercontent.com/4026553/70447947-814e5600-1a97-11ea-8515-6c53bd2c77c7.png)

Here's the input data pane: 

![image](https://user-images.githubusercontent.com/4026553/70448245-fe79cb00-1a97-11ea-8040-79b9deeccb83.png)

### Source code / logs

Please paste or attach the code or logs or traces that would be helpful to diagnose the issue you are reporting.

gvashishtha:
@briacht any ideas on this?

KiransHub:
I'm happy to help in any way that I can to get this resolved. I think it may have something to do with hidden characters.

Perhaps I can try with just one row.. I could duplicate it over and over so the content is the same. If that fails we know the problem is likely to be with that row. 

I can then test it with a simple string to prove that works. 

Then I can strip all email addresses and confidential data from the email contents and I can provide that data from you to look at. 

How does that sound?

justinormont:
It's likely an escaping issue which is not handled by ML.NET's TextLoader. Most common is quote (") or newline (\r|\n) character within a string.

See: https://github.com/dotnet/machinelearning/issues/4460

KiransHub:
> It's likely an escaping issue which is not handled by ML.NET's TextLoader. Most common is quote (") or newline (\r|\n) character within a string.

@justinormont thanks for getting back to me! 

That is more than likely... Do you think I could encode the text to numerical values for each char to solve this issue?

justinormont:
@KiransHub : I wouldn't recommend pre-featurizing as this can cause stealthily treacherous data leakage. 

I would pre-clean the TSV file by ensuring no newlines or quotes are in the text columns. 

[...]

Issue #4460, opened 2019-11-08
[AutoML v0.16.0] InferColumn doesn't work on tricky csv file


LittleLittleCloud:
For some csv file that contains double quotes in it's field, the `inferColumn` API can't work properly. It's probably because when guessing delimiter, AutoML takes the candidates inside double quote into consideration, which should be neglect. (Or when splitting lines, it uses \n inside double quote)

steps to reproduce:
download this [dataset](http://data.insideairbnb.com/the-netherlands/north-holland/amsterdam/2019-09-14/visualisations/listings.csv)

```
MLContext mlContext = new MLContext();
var inputColumnInformation = new ColumnInformation();
inputColumnInformation.LabelColumnName = @"review_scores_rating";
var train = mlContext.Auto().InferColumns(TrainDataPath, inputColumnInformation);
```

#### Updated
The dataset actually works for latest AutoML/ModelBuilder, To reproduce the error, please uses this dataset:

[jigsaw.txt](https://github.com/dotnet/machinelearning/files/4631129/jigsaw.txt)

justinormont:
This is a issue of the TextLoader in ML.NET. It does not currently support escaped quotes in a quoted field.

The TextLoader has a rather limited support for TSV/CSV files. 

The issue is noted in the old repo: https://github.com/dotnet/machinelearning-automl/issues/193 ("Infercolumn fails to parse new lines inside quoted text")

> 
> ---
> 
> > @vinodshanbhag : 
> > Wikidetox fails in benchmarking because of this.
> > Tools like Excel are able to handle this.
> 
> ---
> 
> 
> > @justinormont :
> > @CESARDELATORRE: what do you think about writing an example of converting a dataset from CSV/TSV to IDV?
> > 
> > The `TextLoader` can not handle many CSV/TSV files. Using a more general reader and outputting to IDV would the allow the AutoML code to read the IDV format. 
> > 
> > Basic example:
> > * General (C# CSV/TSV) reader which returns an `IEnumerable` (maybe [CsvHelper](https://joshclose.github.io/CsvHelper/))
> > * ML.NET creates an `IDataView` from the `IEnumerable` https://github.com/dotnet/machinelearning/blob/d6a4b1dd79e13c9a749ad9fdeb1c6ea2f4f547a6/docs/code/SchemaComprehension.md#streaming-data-views
> > * ML.NET writes to an IDV/TDV file -- `mlContext.Data.SaveAsBinary("pathToFile")`
> > 
[...]

PR #4584, opened 2019-12-17
fix TextLoader bug when there's newline between quotes 

LittleLittleCloud:
A quick workaround for this Issue
#4460 
perhaps also this one
#4555


src/Microsoft.ML.Data/Utilities/StreamUtils.cs
```diff
@@ -174,5 +174,40 @@ private static string[] Expand(string pattern)
             return matchList.ToArray();
         }
 #endif
+
+        public static string ReadEntry(this TextReader sr)
+        {
+            string strReturn = string.Empty;
```
JakeRadMSFT: string entry

src/Microsoft.ML.Data/Utilities/StreamUtils.cs
```diff
@@ -174,5 +174,40 @@ private static string[] Expand(string pattern)
...
+
+            // get first bit
+            strReturn += sr.ReadLine();
+
+            // And get more lines until the number of quotes is even
+            while (strReturn.GetNumberOf("\"").IsOdd())
+            {
+                string strNow = sr.ReadLine();
```
JakeRadMSFT: string line

src/Microsoft.ML.Data/Utilities/StreamUtils.cs
```diff
@@ -174,5 +174,40 @@ private static string[] Expand(string pattern)
...
+            {
+                return null;
+            }
+            else
+            {
+                return strReturn;
+            }
+        }
+
+        public static int GetNumberOf(this string s, string strSearchString)
```
JakeRadMSFT: I would just leave this as a static method instead of an extension method. It's easier to read that way.

src/Microsoft.ML.Data/Utilities/StreamUtils.cs
```diff
@@ -174,5 +174,40 @@ private static string[] Expand(string pattern)
...
+        {
+            return s.Length - s.Replace(strSearchString, string.Empty).Length;
```
JakeRadMSFT: Interesting approach :)

src/Microsoft.ML.Data/Utilities/StreamUtils.cs
sharwell: I would agree, especially with completion for unimported extension methods coming from Roslyn. Specialized extension methods for common types will be uncomfortably noisy during development.

src/Microsoft.ML.Data/Utilities/StreamUtils.cs
LittleLittleCloud: Actually there's a bug.... It should divided by the length of `strSearchString` to get the correct result. also if I do that, it won't work on string.Empty (div by 0)... so hard to write the correct code!

src/Microsoft.ML.Data/Utilities/StreamUtils.cs
```diff
@@ -174,5 +174,39 @@ private static string[] Expand(string pattern)
...
+
+        public static string ReadEntry(this TextReader sr)
+        {
+            string entry = string.Empty;
+
+            // get first bit
+            entry += sr.ReadLine();
+
+            // And get more lines until the number of quotes is even
+            while (GetNumberOf(entry, "\"") % 2 != 0 )
```
justinormont: Need to first check that that quoting is enabled, and need to check for escaped quotes.

Some tests:
* `True,"asdf,123` is perfectly ok without quoting enabled.
* `True,"hi\"asdf",123` has an escaped quote.

src/Microsoft.ML.Data/Utilities/StreamUtils.cs
```diff
@@ -174,5 +174,39 @@ private static string[] Expand(string pattern)
...
+            }
+            else
+            {
+                return entry;
+            }
+        }
+
+        public static int GetNumberOf(string s, string strSearchString)
+        {
+            if(strSearchString.Length == 0 || s.Length == 0)
```
justinormont:
```suggestion
            if (strSearchString.Length == 0 || s.Length == 0)
```

src/Microsoft.ML.Data/Utilities/StreamUtils.cs
```diff
@@ -174,5 +174,39 @@ private static string[] Expand(string pattern)
...
+            {
+                return 0;
+            }
+            return (s.Length - s.Replace(strSearchString, string.Empty).Length) / strSearchString.Length;
```
justinormont: `String.Replace()` is not efficient as it reallocates.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderCursor.cs
```diff
@@ -514,7 +514,7 @@ private void ThreadProc()
                                     if (_abort)
                                         return;
 
-                                    text = rdr.ReadLine();
+                                    text = rdr.ReadEntry();
```
justinormont: Should check if moving to multi-line reading will interfere with multi-threaded reading.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderCursor.cs
```diff
@@ -487,7 +487,7 @@ private void ThreadProc()
                                     // REVIEW: Avoid allocating a string for every line. This would probably require
                                     // introducing a CharSpan type (similar to ReadOnlyMemory but based on char[] or StringBuilder)
                                     // and implementing all the necessary conversion functionality on it. See task 3871.
-                                    text = rdr.ReadLine();
+                                    text = rdr.ReadEntry();
```
justinormont: Recommend a flag for multi-line support.

src/Microsoft.ML.Data/Utilities/StreamUtils.cs
```diff
@@ -174,5 +174,39 @@ private static string[] Expand(string pattern)
...
+            string entry = string.Empty;
+
+            // get first bit
+            entry += sr.ReadLine();
+
+            // And get more lines until the number of quotes is even
+            while (GetNumberOf(entry, "\"") % 2 != 0 )
+            {
+                string line = sr.ReadLine();
+                entry += line;
```
justinormont: Likely want to begin a string builder instead of a `+=`.

src/Microsoft.ML.Data/Utilities/StreamUtils.cs
justinormont: Also need to not append to lines at the start of the file which begin with `\\` or `#`: 
```c#
if (entry[0] == '#' ||  entry[0] == '/' && entry[1] == '/') { ... }
```

src/Microsoft.ML.Data/Utilities/StreamUtils.cs
justinormont: You'll want to push thru a newline into the string builder. `ReadLine()` doesn't include the `\n` or `\r` or `\r\n` [[ref](https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.readline?view=netcore-3.0#remarks)].

Otherwise:
```csv
a,b,"c my missing
space",e,f
``` 

Would be read as:
```csv
a,b,"c my missingspace",e,f
```

Instead of:
```csv
a,b,"c my missing\nspace",e,f
```

It likely is not important, but it may be better to match the `\n` or `\r` or `\r\n` of the last line read. This would only be important if the user read the file w/ `TextLoader` but fed the model with an IEnumerable.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderCursor.cs
LittleLittleCloud: How to check that, could you expand a bit?

src/Microsoft.ML.Data/Utilities/StreamUtils.cs
LittleLittleCloud: So try to repeat the logic here

if line starts with '#' or starts with '\\' => return line    # comments
if quote is disabled: => use the existing logic in TextLoader
else
if ( line has only odd# real quote ) => read next line until sum of real quote is even  and return line # real quote means `"` and it's not one of [ `""`, `\"` ` "`]
else return line
    

And BTW it seems the existing enable quote logic doesn't check if it's real quote or not. ( for example in lines: "blablabla""blablabla" will cause an error in `FetchNextField` function, but in Excel it's legal) Do we have a plan for adding support for that?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderCursor.cs
LittleLittleCloud: I feel it can share the same flag with EnableQuote, If we allow quote, then multi-line support is necessary to make sure TextLoader function well

PR #4587, opened 2019-12-18
Fix PFI issue in binary classification

### LINKED ISSUE #4517: PFI doesn't work with uncalibrated binary classifiers
Some binary classification estimators automatically return a calibrated model (for example, FastTree and LogisticRegression), but some don't - for example, FastForest. When trying to pass such a model to PFI, there is an exception thrown saying that the probability column was not found.
```
var ml = new MLContext();
var ff = ml.BinaryClassification.Trainers.FastForest();
var data = ml.Data.LoadFromTextFile(@"breast-cancer.txt",
                new[] { new TextLoader.Column("Label", DataKind.Boolean, 0),
                            new TextLoader.Column("Features", DataKind.Single, 1, 9) });
var model = ff.Fit(data);
var pfi = ml.BinaryClassification.PermutationFeatureImportance(model, data);
```

There are actually two issues here: The first is what I described above, and the second is that there is no workaround for this problem. I tried adding a calibrator manually:

```
var ff = ml.BinaryClassification.Trainers.FastForest();
var ffmodel = ff.Fit(data);
var calibrator = ml.BinaryClassification.Calibrators.Platt();
var calibratormodel = calibrator.Fit(ffmodel.Transform(data));
var pfi = ml.BinaryClassification.PermutationFeatureImportance(calibratormodel, ffmodel.Transform(data));
```
The reason I could not train these two models as a pipeline is because the resulting model is of type `TransformerChain` so I cannot pass it to PFI. However, this code doesn't work either, because even though `calibratormodel` is indeed an `ISingleFeaturePredictionTransformer`, the features column of `calibratormodel`, is the score column of the output of `ffmodel`, so PFI doesn't do the right thing. As far as I can tell, there is no way to pass a model where the calibrator was trained separately to PFI. It might be worth opening a separate issue for this, not sure.

yaeldekel:
This change adds support for running PFI on binary classification models that do not contain a calibrator. Fixes #4517 .

### DIFF
--- a/src/Microsoft.ML.Data/Evaluators/Metrics/CalibratedBinaryClassificationMetrics.cs
+++ b/src/Microsoft.ML.Data/Evaluators/Metrics/CalibratedBinaryClassificationMetrics.cs
@@ -49,5 +49,15 @@ namespace Microsoft.ML.Data
             LogLossReduction = Fetch(BinaryClassifierEvaluator.LogLossReduction);
             Entropy = Fetch(BinaryClassifierEvaluator.Entropy);
         }
+
+        [BestFriend]
+        internal CalibratedBinaryClassificationMetrics(double auc, double accuracy, double positivePrecision, double positiveRecall,
+            double negativePrecision, double negativeRecall, double f1Score, double auprc, double logLoss, double logLossReduction, double entropy)
+            : base(auc, accuracy, positivePrecision, positiveRecall, negativePrecision, negativeRecall, f1Score, auprc)
+        {
+            LogLoss = logLoss;
+            LogLossReduction = logLossReduction;
+            Entropy = entropy;
+        }
     }
 }
--- a/src/Microsoft.ML.Data/Evaluators/Metrics/CalibratedBinaryClassificationMetrics.cs
+++ b/src/Microsoft.ML.Data/Evaluators/Metrics/CalibratedBinaryClassificationMetrics.cs
@@ -49,15 +49,5 @@ namespace Microsoft.ML.Data
             LogLossReduction = Fetch(BinaryClassifierEvaluator.LogLossReduction);
             Entropy = Fetch(BinaryClassifierEvaluator.Entropy);
         }
-
-        [BestFriend]
-        internal CalibratedBinaryClassificationMetrics(double auc, double accuracy, double positivePrecision, double positiveRecall,
-            double negativePrecision, double negativeRecall, double f1Score, double auprc, double logLoss, double logLossReduction, double entropy)
-            : base(auc, accuracy, positivePrecision, positiveRecall, negativePrecision, negativeRecall, f1Score, auprc)
-        {
-            LogLoss = logLoss;
-            LogLossReduction = logLossReduction;
-            Entropy = entropy;
-        }
     }
 }
--- a/src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
+++ b/src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
@@ -4,7 +4,9 @@
 
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using Microsoft.ML.Calibrators;
 using Microsoft.ML.Data;
+using Microsoft.ML.Internal.Utilities;
 using Microsoft.ML.Runtime;
 using Microsoft.ML.Transforms;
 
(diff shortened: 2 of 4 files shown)


src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
```diff
@@ -171,6 +199,23 @@ public static class PermutationFeatureImportanceExtensions
                 auprc: a.AreaUnderPrecisionRecallCurve - b.AreaUnderPrecisionRecallCurve);
         }
 
+        private static CalibratedBinaryClassificationMetrics CalibratedBinaryClassifierDelta(
+            CalibratedBinaryClassificationMetrics a, CalibratedBinaryClassificationMetrics b)
```
antoniovs1029: If I open this PR in Visual Studio I can see that this new method has 0 references, is this intended?

Particularly, notice that the `BinaryClassifierDelta ` (which already existed) gets a reference in `PermutationFeatureImportance<TModel>` weather `isCalibratedModel` is true or not.

Further more, I quickly ran another existing test for PFI for Binary Classification, with a calibrated model, and it still called the BinaryClassifierDelta instead of the new CalibratedBinaryClassifierDelta. So I wouldn't know if this is intended.

gvashishtha:
So based on the [ML.NET documentation on PFI](https://docs.microsoft.com/en-us/dotnet/machine-learning/how-to-guides/explain-machine-learning-model-permutation-feature-importance-ml-net), it seems that if you were to take the more time-intensive route of exposing the LogLoss, LogLossReduction, and Entropy metrics, there would be an additional way for users to compare the importance of features (rather than simply R^2)? 

What's a situation where R^2 would not be sufficient for doing PFI well?

antoniovs1029:
> What's a situation where R^2 would not be sufficient for doing PFI well?

@gvashishtha 

For the record, R^2 is not included as a metric inside `BinaryClassificationMetrics` ([link to code](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Evaluators/Metrics/BinaryClassificationMetrics.cs)) and so it doesn't exist either in `BinaryClassificationMetricsStatistics` ([link](https://github.com/dotnet/machinelearning/blob/44ed77266daaae3fa52234d6d4096b9cb9987df4/src/Microsoft.ML.Transforms/MetricStatistics.cs#L142)). In the tutorial you linked, Regression is used, and in that case R^2 is included inside its metrics ([link](https://github.com/dotnet/machinelearning/blob/44ed77266daaae3fa52234d6d4096b9cb9987df4/src/Microsoft.ML.Transforms/MetricStatistics.cs#L113)).

So the PR here would only affect Binary Classification. Currently `BinaryClassificationMetricsStatistics `only offers metrics for AreaUnderRocCurve, Accuracy, PositivePrecision, PositiveRecall, NegativePrecision, NegativeRecall, F1Score, and AreaUnderPrecisionRecallCurve. When performing PFI on a Binary Classification model, the change in the metrics is computed for all those metrics, and for every feature. So a user could evaluate the PFI results with any of those metrics (though, in practice, I wouldn't know which metrics would actually be used to choose the most important feature, I _guess_ it depends on the specific case the user is using PFI and there's no objective way of saying if they're "sufficient" or not...).

So, the question remains if it's worth it to make breaking changes in the API to also include LogLoss, LogLossReduction, and Entropy metrics when using PFI with a Binary Classification Calibrated model, by returning CalibratedBinaryClassificationMetrics instead. The real question would be if those metrics would be valuable to users.... and since no one has opened an issue about them, I would tend to think that it's better to not make the braking changes in the API.

src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
```diff
@@ -144,17 +146,43 @@ public static class PermutationFeatureImportanceExtensions
...
-                            BinaryClassifierDelta,
-                            predictionTransformer.FeatureColumnName,
-                            permutationCount,
-                            useFeatureWeightFilter,
-                            numberOfExamplesToUse);
+                catalog.GetEnvironment(),
+                predictionTransformer,
+                data,
+                () => new BinaryClassificationMetricsStatistics(),
+                idv => catalog.EvaluateNonCalibrated(idv, labelColumnName),
```
harishsk: Small Nit:
The only difference between the `if (isCalibratedModel)` and the else case is the idv parameter. Is it possible to make this a bit more readable by factoring out just that line and using a single call to the PermutationFeatureImportance constructor?

test/Microsoft.ML.Tests/PermutationFeatureImportanceTests.cs
```diff
@@ -305,6 +305,18 @@ public void TestPfiBinaryClassificationOnSparseFeatures(bool saveModel)
 
             Done();
         }
+
+        [Fact]
+        public void TestBinaryClassificationWithoutCalibrator()
```
harishsk: The test does not Assert anything. Can you please include Asserts for the relevant results that this test is supposed to verify?

harishsk:
Breaking the API has a high bar and I think this does not meet that bar. I would suggest leaving it as is or adding a new API that returns the calibrated metrics.

yaeldekel:
I will leave it as is for now, and make a change that ensures the existing API doesn't break only in case we get customer asks.

---
In reply to: [570431913](https://github.com/dotnet/machinelearning/pull/4587#issuecomment-570431913) [](ancestors = 570431913)

src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
yaeldekel: Resolving this comment, since `isCalibratedModel` has been removed.

---
In reply to: [362685943](https://github.com/dotnet/machinelearning/pull/4587#discussion_r362685943) [](ancestors = 362685943)

src/Microsoft.ML.Transforms/PermutationFeatureImportanceExtensions.cs
yaeldekel: I have removed it (I added it initially before we decided not to add calibrated metrics to PFI).

---
In reply to: [359531828](https://github.com/dotnet/machinelearning/pull/4587#discussion_r359531828) [](ancestors = 359531828)

src/Microsoft.ML.Data/Evaluators/Metrics/CalibratedBinaryClassificationMetrics.cs
```diff
@@ -49,5 +49,15 @@ internal CalibratedBinaryClassificationMetrics(IHost host, DataViewRow overallRe
             LogLossReduction = Fetch(BinaryClassifierEvaluator.LogLossReduction);
             Entropy = Fetch(BinaryClassifierEvaluator.Entropy);
         }
+
+        [BestFriend]
+        internal CalibratedBinaryClassificationMetrics(double auc, double accuracy, double positivePrecision, double positiveRecall,
+            double negativePrecision, double negativeRecall, double f1Score, double auprc, double logLoss, double logLossReduction, double entropy)
```
antoniovs1029: So I believe this constructor is no longer used (after you removed the other code handling the calibrated case). Is there a reason the keep this constructor?

PR #4594, opened 2019-12-18
Add Category Imputer Featurizer

michaelgsharp:
This change adds in the CategoryImputer into the new Featurizers project. It is the second of a series of PR's that will go in. The CategoryImputer is implemented in native code, so this is mostly just a wrapper around that with the appropriate entrypoints for NimbusML as well.

The CategoryImputer imputes missing values in a column based on the most frequent value seen during training. Due to not having support for nullable types in ML.NET, this only supports Float, Double, and Strings (with empty string being treated as "missing").

This code is auto generated. The functionality will be migrated into the MissingValueReplacer over the next several weeks and this file will be removed, but its going in as is to unblock other development work.


src/Microsoft.ML.Featurizers/CategoryImputer.cs
```diff
@@ -0,0 +1,766 @@
...
+
+    public sealed class CategoryImputerTransformer : RowToRowTransformerBase, IDisposable
+    {
+        #region Class data members
+
+        internal const string Summary = "Fills in missing values in a column based on the most frequent value";
+        internal const string UserName = "CategoryImputer";
+        internal const string ShortName = "CategoryImputer";
+        internal const string LoadName = "CategoryImputer";
+        internal const string LoaderSignature = "CategoryImputer";
```
justinormont: Should this be named `CategoricalImputer`?

src/Microsoft.ML.Featurizers/CategoryImputer.cs
```diff
@@ -0,0 +1,766 @@
...
+            private static extern bool DestroyTransformerNative(IntPtr transformer, out IntPtr errorHandle);
+
+            [DllImport("Featurizers", EntryPoint = "CatImputerFeaturizer_string_Transform", CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity]
+            private static unsafe extern bool TransformDataNative(TransformerEstimatorSafeHandle transformer, byte* input, out IntPtr output, out IntPtr outputSize, out IntPtr errorHandle);
+            [DllImport("Featurizers", EntryPoint = "CatImputerFeaturizer_string_DestroyTransformedData", CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity]
+            private static extern bool DestroyTransformedDataNative(IntPtr output, IntPtr outputSize, out IntPtr errorHandle);
+            internal unsafe override ReadOnlyMemory<char> Transform(ReadOnlyMemory<char> input)
+            {
+                var inputAsString = input.ToString();
+                fixed (byte* interopInput = string.IsNullOrEmpty(inputAsString) ? null : Encoding.UTF8.GetBytes(inputAsString + char.MinValue))
```
justinormont: Is `Transform()` called for each row to apply the, already learned, imputer? If so, isn't this quite inefficient to reallocate the string for each row of data?

Shouldn't this be roughly:
```C#
return ((!input.IsEmpty || input.Length == 0) ? ReplacementValue : input);
```

src/Microsoft.ML.Featurizers/CategoryImputer.cs
michaelgsharp: I am fine with either name. It was named `CategoryImputer` in the original python code, but if we think `CategoricalImputer` sounds better I am fine with renaming it.

src/Microsoft.ML.Featurizers/CategoryImputer.cs
michaelgsharp: I need the input string as a byte array so that I can pass it to the native code. Currently, we are not getting the `Replacement` value back from the native code, though that is an interesting optimization that we could potentially look at in the future.

src/Microsoft.ML.Featurizers/CategoryImputer.cs
```diff
@@ -0,0 +1,766 @@
...
+
+[assembly: EntryPointModule(typeof(CategoryImputerEntrypoint))]
+
+namespace Microsoft.ML.Featurizers
+{
+    public static class CategoryImputerExtensionClass
+    {
+        /// <summary>
+        /// Create a <see cref="CategoryImputerEstimator"/>, which fills in the missing values in a column with the most frequent value. Supports Floats, Doubles, and Strings.
+        /// A string is assumed "missing" if it is empty.
```
harishsk: ML.NET already has a MissingValueReplacingEstimator with different replacement modes. What seems to be missing is a replacement mode for most frequent value. Wouldn't it be simpler to just add that mode to MissingValueReplacingEstimator?

src/Microsoft.ML.Featurizers/CategoryImputer.cs
justinormont: I'd follow the naming conventions of ML.NET (and my suggestion may be imperfect too). 

Side topic: I wouldn't perfectly replicate the existing Python code as it itself is just fulfilling a user data science need; in this case imputing categorical features with the most common value. The general goal is to fulfill the same user need; the exact methods can be a bit fungible.

src/Microsoft.ML.Featurizers/CategoryImputer.cs
justinormont: Yes, I agree this route of extending the existing transform. 

This is cleaner for the user as they don't have to learn a new transform, and there isn't two places to accomplish similar tasks. One transform increases discoverability and encourages the user to explore alternative imputation methods.

src/Microsoft.ML.Featurizers/CategoryImputer.cs
justinormont: I think we're relying on the ONNX code a bit much in that case. This is a lot of setup code to do a really small amount of work (the one line above). This seems to be a bit of the tail-wagging-the-dog style, where the `ML.NET` dependency, ONNX, is driving the `ML.NET` code instead of the other way around. If we need ONNX to take  `ReadOnlyMemory<char>`, it should (and without allocating any new memory in its code). Specifically, there should be no allocations in this function, or the ONNX portion.

We can mitigate the overhead:
The replacement value is static once the transform is trained, and it's trained before hitting this section of the code. 

How about we just cache the value after calling the native ONNX code _once_? 

```C#
// Early out if (1) if a replacement isn't needed or (2) the cached value is known.
// Otherwise continue with existing code and cache the inputation value at the end.
if (!(input.IsEmpty || input.Length == 0))
  return input; // No need to impute
else {
  if (ReplacementValue != null) // Check if the cached replacement value is known
    return ReplacementValue;
}

// Fall through to continue with the existing function
//...

ReplacementValue = valueFromOnnx;
```

Where `ReplacementValue`  is thread local to the transform (to avoid using a lock for race condition in initializing the value; though all threads will fill it with the same value); or more simply just store the value after the transform is trained or initialized.

src/Microsoft.ML.Featurizers/CategoryImputer.cs
michaelgsharp: That wont work unfortunately. To use this code you have to take a new dependency on the featurizers binary. The main ML.NET nuget cannot have any new dependencies added to it, so this has to be in a separate class/nuget package.

src/Microsoft.ML.Featurizers/CategoryImputer.cs
michaelgsharp: If `CategoricalImputer` would better match ML.NET naming schemes then I will update it. I am still not very good at figuring out what matches well and what doesn't.

src/Microsoft.ML.Featurizers/CategoryImputer.cs
michaelgsharp: I can easily fix it so that if its not null it just returns the input. I will add that in shortly. The caching of the replacement value though I am not sure how easy/hard that would be to update as the native code itself doens't currently have a way to return that to me. Is that something you are willing to punt on if I fix it so it only calls the native code if there isn't a value in the input?

src/Microsoft.ML.Featurizers/CategoryImputer.cs
harishsk: ML.NET already has most of the plumbing in this specific case. There is already a MeanAggregatorOne class in MissingValueReplacingUtils.cs. Wouldn't it be simpler to add a ModeAggregratorOne and related classes and use the featurizer code only for the Onnx functionality?

---
In reply to: [363476040](https://github.com/dotnet/machinelearning/pull/4594#discussion_r363476040) [](ancestors = 363476040)

src/Microsoft.ML.Featurizers/CategoryImputer.cs
justinormont: I believe the native code returns the same value each time (for each row with a missing value). Hence all empty/null inputs, for a single feature column, gets filled with the same static value. 

I believe we just have to cache this one value after the 1st (and only) call to the native code -- call it with the 1st missing value, and store the result. I expect there's no changes to the native code.

src/Microsoft.ML.Featurizers/CategoryImputer.cs
michaelgsharp: Renamed to CategoricalImputer

src/Microsoft.ML.Featurizers/CategoryImputer.cs
michaelgsharp: Made the changes you suggest. Returning the input value if its not null/empty, and caching the result after the first call to transform.

src/Microsoft.ML.Featurizers/CategoryImputer.cs
```diff
@@ -0,0 +1,799 @@
...
+    /// <remarks>
+    /// <format type="text/markdown"><![CDATA[
+    ///
+    /// ###  Estimator Characteristics
+    /// |  |  |
+    /// | -- | -- |
+    /// | Does this estimator need to look at the data to train its parameters? | Yes |
+    /// | Input column data type | Float, Double, String |
+    /// | Output column data type | Same as input column |
+    ///
```
michaelgsharp: /// | Exportable to ONNX | Yes |

src/Microsoft.ML.Featurizers/CategoryImputer.cs
```diff
@@ -0,0 +1,799 @@
...
+                    return savedData;
+                }
+            }
+
+            internal unsafe void CreateTransformerFromSavedData(byte[] data)
+            {
+                fixed (byte* rawData = data)
+                {
+                    IntPtr dataSize = new IntPtr(data.Count());
+                    CreateTransformerFromSavedDataHelper(rawData, dataSize);
```
michaelgsharp: Its not clear where the transformer is being returned

src/Microsoft.ML.Featurizers/CategoryImputer.cs
```diff
@@ -0,0 +1,799 @@
...
+                using (var estimatorHandler = new TransformerEstimatorSafeHandle(estimator, DestroyEstimatorHelper))
+                {
+                    var fitResult = FitResult.Continue;
+                    while (fitResult != FitResult.Complete)
+                    {
+                        using (var data = input.GetColumn<T>(Source).GetEnumerator())
+                        {
+                            while (fitResult == FitResult.Continue && data.MoveNext())
+                            {
+                                {
```
michaelgsharp: remove

src/Microsoft.ML.Featurizers/CategoryImputer.cs
```diff
@@ -0,0 +1,799 @@
...
+            private static unsafe extern bool TransformDataNative(TransformerEstimatorSafeHandle transformer, in float input, out float output, out IntPtr errorHandle);
+            internal override float Transform(float input)
+            {
+                if (!float.IsNaN(input))
+                {
+                    return input;
+                }
+                else if (float.IsNaN(Result))
+                {
+                    var success = TransformDataNative(_transformerHandler, input, out float output, out IntPtr errorHandle);
```
michaelgsharp: Move to constructor, make result readonly

src/Microsoft.ML.Featurizers/CategoryImputer.cs
```diff
@@ -0,0 +1,799 @@
...
+                CreateTransformerSaveDataNative(_transformerHandler, out buffer, out bufferSize, out errorHandle);
+        }
+
+        #endregion
+
+        #region  ReadOnlyCharColumn
+
+        internal sealed class ReadOnlyCharTypedColumn : TypedColumn<ReadOnlyMemory<char>>
+        {
+            private TransformerEstimatorSafeHandle _transformerHandler;
```
michaelgsharp: move to the base class

src/Microsoft.ML.Featurizers/CategoryImputer.cs
```diff
@@ -0,0 +1,799 @@
...
+            private static extern bool DestroyTransformedDataNative(IntPtr output, IntPtr outputSize, out IntPtr errorHandle);
+            internal unsafe override ReadOnlyMemory<char> Transform(ReadOnlyMemory<char> input)
+            {
+                if (!input.IsEmpty)
+                {
+                    return input;
+                }
+                else if (Result.IsEmpty)
+                {
+                    var inputAsString = input.ToString();
```
michaelgsharp: please do not send input to the transform

src/Microsoft.ML.Featurizers/CategoryImputer.cs
```diff
@@ -0,0 +1,766 @@
...
+    /// <remarks>
+    /// <format type="text/markdown"><![CDATA[
+    ///
+    /// ###  Estimator Characteristics
+    /// |  |  |
+    /// | -- | -- |
+    /// | Does this estimator need to look at the data to train its parameters? | Yes |
+    /// | Input column data type | Float, Double, String |
+    /// | Output column data type | Same as input column |
+    ///
```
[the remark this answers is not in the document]
michaelgsharp: Added

src/Microsoft.ML.Featurizers/CategoryImputer.cs
michaelgsharp: Changed to return transformer as suggested.

src/Microsoft.ML.Featurizers/CategoryImputer.cs
michaelgsharp: As discussed offline this code is auto generated. The functionality will be migrated into the `MissingValueReplacingTransformer` over the next several weeks and this file will be removed, but its going in as is to unblock other development work.

src/Microsoft.ML.Featurizers/CategoryImputer.cs
justinormont: Thanks for making the change! 

Your new function looks fast and svelte:
https://github.com/dotnet/machinelearning/blob/795559d951366dd126b6d7b092342155459c25d8/src/Microsoft.ML.Featurizers/CategoricalImputer.cs#L686-L692

PR #4596, opened 2019-12-19
ONNX Conversion for Hashing Transformer

KsenijaS:
Here is the draft for Onnx Conversion for Hashing transformer.
Please not that some test will not pass because onnx only support string and uint for murmur hash3
and also needs to add mask and useOrderedHashing to be compatible with ML.NET
Further more there is a bug in ORT regarding UTF 8 encoding, (they fixed it in this PR: https://github.com/microsoft/onnxruntime/pull/2697), but I think the next release is in about 2 months


src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -122,8 +123,9 @@ private static VersionInfo GetVersionInfo()
             return new VersionInfo(
                 modelSignature: "HASHTRNS",
                 // verWrittenCur: 0x00010001, // Initial
-                verWrittenCur: 0x00010002, // Invert hash key values, hash fix
-                verReadableCur: 0x00010002,
+                //verWrittenCur: 0x00010002, // Invert hash key values, hash fix
+                verWrittenCur: 0x00010003,
```
yaeldekel: >0x00010003 [](start = 31, length = 10)

Please add a comment explaining why the version is changing (similar to the line above).

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -245,9 +247,15 @@ private Delegate GetGetterCore(DataViewRow input, int iinfo, out Action disposer
             disposer = null;
             input.Schema.TryGetColumnIndex(_columns[iinfo].InputColumnName, out int srcCol);
             var srcType = input.Schema[srcCol].Type;
-            if (!(srcType is VectorDataViewType vectorType))
-                return ComposeGetterOne(input, iinfo, srcCol, srcType);
-            return ComposeGetterVec(input, iinfo, srcCol, vectorType);
+            if (GetVersionInfo().VerWrittenCur == 0x00010002)
```
yaeldekel: >GetVersionInfo [](start = 16, length = 14)

This will always return `0x00010003`.

yaeldekel: I think what you need is to add a field to `HashingTransformer` that has the version. That field will be initialized to be `0x00010003` in the constructors in lines 166 and 177, and according to the `ModelLoadContext`'s version info in the constructor in line 267.

---
In reply to: [387066236](https://github.com/dotnet/machinelearning/pull/4596#discussion_r387066236) [](ancestors = 387066236)

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -834,6 +1083,66 @@ private void AddMetaKeyValues(int i, DataViewSchema.Annotations.Builder builder)
             }
 
             protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, bool> activeOutput, out Action disposer) => _parent.GetGetterCore(input, iinfo, out disposer);
+
+            private bool SaveAsOnnxCore(OnnxContext ctx, int iinfo, string srcVariable, string dstVariable)
```
yaeldekel: >SaveAsOnnxCore [](start = 25, length = 14)

Should `NumberOfBits` and `UseOrderedHashing` influence the output here?

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -1087,6 +1087,98 @@ public void WordTokenizerOnnxConversionTest()
...
+                new HashData {Education = "alibaba".AsMemory()},
+                new HashData {Education = "baba".AsMemory()},
+                new HashData {Education = "U+123".AsMemory()},
+                new HashData {Education = "djldaoiejffjauhglehdlgh".AsMemory()},
+                new HashData {Education = "~".AsMemory()},
+            };
+
+            IDataView data = mlContext.Data.LoadFromEnumerable(samples);
+
+            var hashEstimator = new HashingEstimator(Env, "Education");
```
yaeldekel: >HashingEstimator [](start = 36, length = 16)

Please add tests that test `HashingEstimator` with non-default arguments.

src/Microsoft.ML.Data/Transforms/Hashing.cs
KsenijaS: yes, mask is dependant on NumberofBits, and we aslo need UseOrderedHashing, but we can get that info from _parent._columns[iinfo]

yaeldekel: But I don't see this information being used anywhere in `SaveOnnxCore`.

---
In reply to: [391310537](https://github.com/dotnet/machinelearning/pull/4596#discussion_r391310537) [](ancestors = 391310537)

src/Microsoft.ML.Data/Transforms/Hashing.cs
KsenijaS: field GetVersionInfo().VerWrittenCur will always return 0x00010003 for newly created models, this was meant for already existed models and those models will not be able to be converted to ONNX

yaeldekel: `GetVersionInfo()` calls the method in line 121, so it will return `0x00010003` even if the transformer was created from a model with a different version. When the `GetGetterCore` method is called you no longer have access to the model that created the transformer (or, it might not have been created from a serialized model at all), which is why you need to add a field to store the version, and initialize it in all the constructors. In the 

---
In reply to: [396068943](https://github.com/dotnet/machinelearning/pull/4596#discussion_r396068943) [](ancestors = 396068943)

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -1087,6 +1087,98 @@ public void WordTokenizerOnnxConversionTest()
...
+            var onnxModel = mlContext.Model.ConvertToOnnxProtobuf(model, data);
+
+            var onnxFileName = "MurmurHashV2.onnx";
+            var onnxTextName = "MurmurHashV2.txt";
+            var onnxModelPath = GetOutputPath(onnxFileName);
+            var onnxTextPath = GetOutputPath(onnxTextName);
+
+            SaveOnnxModel(onnxModel, onnxModelPath, onnxTextPath);
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && Environment.Is64BitProcess)
```
Lynx1820: can you use IsOnnxRuntimeSupported()? Or will this only be available in Windows?

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -834,6 +1083,66 @@ private void AddMetaKeyValues(int i, DataViewSchema.Annotations.Builder builder)
...
+            {
+                string opType;
+                OnnxNode murmurNode;
+
+                opType = "MurmurHash3";
+                if (_types[iinfo].RawType == typeof(KeyDataViewType))
```
Lynx1820, line 1093: It seems like the output for this transformer is always uint32 for the tests in your branch. Do you have a test where this code get's called?

KsenijaS: No, I don't have.

harishsk:
This was replaced by #5013. Closing.

PR #4608, opened 2019-12-31
Fixes #3878. About calling Fit more than once on Multiclass LightGBM trainer.

### LINKED ISSUE #3878: Multiclass LightGBM bug
LightGBM trainer has two non-readonly fields called `_numClass` and `_tlcNumClass`. The second one is used to determine the number of predictors in the OVA predictor. However, the value of `_tlcNumClass` is only updated once, so if `Fit` is called again on the same estimator, it might give the wrong number of classes.

antoniovs1029:
Fixes #3878, by initializing the values of _`tlcNumClass` and `_numClass` everytime `Fit()` is called on a Multiclass LightGBM trainer. This is done under the assumption that trainers should behave as if they were stateless, and every call to `Fit()` should re-initialize those values.

As discussed offline with @yaeldekel , `_tlcNumClass` and `_numClass` hold the same value if there were no NaN labels on the dataset used for training. If there were NaN labels then `_tlcNumClass = _numClass - 1`. Mantaining both fields is necessary, because [here](https://github.com/dotnet/machinelearning/blob/aa5b3be14eeab1160dca2cc08561190141199f4f/src/Microsoft.ML.LightGbm/LightGbmMulticlassTrainer.cs#L219) NaN labels are replaced to be a new class, and then, [here](https://github.com/dotnet/machinelearning/blob/aa5b3be14eeab1160dca2cc08561190141199f4f/src/Microsoft.ML.LightGbm/LightGbmTrainerBase.cs#L620), when training the `WrappedLightGbmTraining` it is as if NaN labels were an extra class, but then [here](https://github.com/dotnet/machinelearning/blob/aa5b3be14eeab1160dca2cc08561190141199f4f/src/Microsoft.ML.LightGbm/LightGbmMulticlassTrainer.cs#L185), when creating the Predictors, only `_tlcNumClass` predictors are created. So, for example, if I have a dataset with 3 classes (0-2), but some rows have NaN values on their labels, then NaN values get converted to "3", `_numClass` is equal to "4" and `WrappedLightGbmTraining` trains as if there were truly 4 classes... but `_tlcNumClass` is equal to "3" and when creating the Predictors, only 3 predictors are created (one for each of the original classes ignoring the "fake NaN class").

The above wasn't documented in the code, but after doing some tests it seems that is how it is supposed to behave, and so I added some comments explaining this, as @yaeldekel  and I agree that the above isn't really clear directly from reading the code, and the names `_tlcNumClass` and `_numClass` are somewhat obscure.

I also added a test.


src/Microsoft.ML.LightGbm/LightGbmMulticlassTrainer.cs
```diff
@@ -255,6 +266,13 @@ private protected override void ConvertNaNLabels(IChannel ch, RoleMappedData dat
                     _tlcNumClass = (int)maxLabel + 1;
                 }
             }
+
+            // If there are NaN labels, they are converted to be equal to _tlcNumClass (i.e. _numClass - 1).
```
antoniovs1029: I am not sure if this comment I added is clear enough, or if I should have added it somewhere else, or if I shouldn't add it anyway.

Problem is that @yaeldekel and I think that having both _numClass and _tlcNumClass variables  is somewhat confusing and it's not clear why they exist... and simply adding a comment somewhere (explaining their difference) might be useful.

src/Microsoft.ML.LightGbm/LightGbmTrainerBase.cs
```diff
@@ -371,6 +373,9 @@ private protected override TModel TrainModelCore(TrainContext context)
             return CreatePredictor();
         }
 
+        private protected virtual void InitializeBeforeTraining()
+        { return; } // MYTODO: Is there a better way to avoid having to do this? An abstract method?
```
antoniovs1029: So, originally the idea that @yaeldekel suggested to me offline was to create an abstract method `InitializeBeforeTraining`, and add the implementation in the `LightGbmMulticlassTrainer` child class to always initialize `_tlcNumClass` and `_numClass` when fitting.

Problem with that is that then I would need to provide an implementation also in the other children of `LightGbmTrainerBase`, but as far as issue #3878 concerns, there isn't any problem in those classes. So instead of making an abstract method, I made this virtual empty method that is overridden in the multiclass LightGBM trainer. Still, I am unsure if this is the best way to do this, as it seems ugly to me... but I wouldn't know better.

src/Microsoft.ML.LightGbm/LightGbmMulticlassTrainer.cs
```diff
@@ -216,9 +216,20 @@ private protected override void CheckDataValid(IChannel ch, RoleMappedData data)
...
         }
 
+        private protected override void InitializeBeforeTraining()
+        {
+            _numClass = -1;
+            _tlcNumClass = 0;
+
+            //MYTODO: Include more initializations, of TrainedEnsemble, for example?
+            //For example:
+            //TrainedEnsemble = null;
```
antoniovs1029: So here I re-initialize the `_numClass` and `_tlcNumClass` fields.

But I wonder if there's any reason to also re-initialize any other field, such as `TrainedEnsemble`, since trainers are supposed to be stateless.

After looking into the code I don't find any reason to re-initialize any other variable... but I would still appreciate any feedback on this regard.

yaeldekel: I think you are right. `TrainedEnsemble` and `FeatureCount` get assigned new values anyway each time `Fit()` is called so they don't need to be initialized in the new method you are introducing. I looked at the other fields of `LightGbmTrainerBase`, and I have additional thoughts about the two fields `GbmOptions` and `ParallelTraining`. I will write them down in a separate comment.

---
In reply to: [362128849](https://github.com/dotnet/machinelearning/pull/4608#discussion_r362128849) [](ancestors = 362128849)

src/Microsoft.ML.LightGbm/LightGbmTrainerBase.cs
yaeldekel: A virtual method is good. You don't need to have `return;` in there though, you can just have empty braces: `{ }`.

---
In reply to: [362128656](https://github.com/dotnet/machinelearning/pull/4608#discussion_r362128656) [](ancestors = 362128656)

yaeldekel:
>         private protected IParallel ParallelTraining;

This should be `readonly`.

---
Refers to: src/Microsoft.ML.LightGbm/LightGbmTrainerBase.cs:291 in 1aaa77f. [](commit_id = 1aaa77f73f237483cfc355a3b0e08ec845fff028, deletion_comment = False)

yaeldekel:
>         private protected Dictionary<string, object> GbmOptions;

This is one of the trainer fields that are not `readonly`, but I see no reason why it shouldn't be `readonly`. I was actually wondering whether we need this field at all - it is used to create a string of options passed to the LightGbm trainer, perhaps this string can be created using the `LightGbmTrainerOptions` field and output of other methods called in `LightGbmTrainerBase.TrainModelCore`?

---
Refers to: src/Microsoft.ML.LightGbm/LightGbmTrainerBase.cs:289 in 1aaa77f. [](commit_id = 1aaa77f73f237483cfc355a3b0e08ec845fff028, deletion_comment = False)

src/Microsoft.ML.LightGbm/LightGbmMulticlassTrainer.cs
yaeldekel: I would put this comment where the fields are defined instead.
Also, maybe we should rename the fields?
`_tlcNumClass` -> `_numberOfClasses`
`_numClass` => `_numberOfClassesIncludingNaN` (or `_lightGbmNumberOfClasses` - this name in addition to your comment explaining that LightGBM considers NaN to be an additional class might be clear enough).

---
In reply to: [362128258](https://github.com/dotnet/machinelearning/pull/4608#discussion_r362128258) [](ancestors = 362128258)

yaeldekel:
>                 var gbmNative = WrappedLightGbmTraining.Train(ch, pch, gbmParams, gbmDataSet, numIteration: numberOfTrainingIterations);

Unrelated: this is `IDisposable`, should be `using ...`.

---
Refers to: test/Microsoft.ML.Tests/TrainerEstimators/TreeEstimators.cs:495 in 1aaa77f. [](commit_id = 1aaa77f73f237483cfc355a3b0e08ec845fff028, deletion_comment = False)

antoniovs1029:
>         private protected IParallel ParallelTraining;

It seems to me the only reason `ParallelTraining` and `GbmOptions` were not `readonly`, was because they weren't initialized in the constructor, but in a helper method called by the constructor (i.e. `InitParallelTraining`). So in order to make them `readonly`, I also had to move their initialization to the constructor of `LightGbmTrainerBase`.

---
In reply to: [570035614](https://github.com/dotnet/machinelearning/pull/4608#issuecomment-570035614) [](ancestors = 570035614)

---
Refers to: src/Microsoft.ML.LightGbm/LightGbmTrainerBase.cs:291 in 1aaa77f. [](commit_id = 1aaa77f73f237483cfc355a3b0e08ec845fff028, deletion_comment = False)

src/Microsoft.ML.LightGbm/LightGbmMulticlassTrainer.cs
```diff
@@ -216,10 +221,21 @@ private protected override void CheckDataValid(IChannel ch, RoleMappedData data)
...
         }
 
+        private protected override void InitializeBeforeTraining()
+        {
+            _numberOfClassesIncludingNan = -1;
+            _numberOfClasses = 0;
+
+            //MYTODO: Include more initializations, of TrainedEnsemble, for example?
+            //For example:
+            //TrainedEnsemble = null;
```
yaeldekel: I think these comments can be safely deleted.

PR #4613, opened 2020-01-01
refine TensorFlowTransformer to avoid early dispose

frank-dong-ms:
from build: https://dev.azure.com/dnceng/public/_build/results?buildId=468639&view=logs&j=d1af5113-e574-5a31-f7f3-02fc40ea7b26&t=167d9e7d-b609-5b0a-7efa-d26b0dafb88f

below unhandled exception:

```
Unhandled exception: System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'The ThreadLocal object has been disposed.'.
at System.Threading.ThreadLocal`1.GetValueSlow()
at Tensorflow.Graph.DisposeManagedResources()
at Tensorflow.DisposableObject.internal_dispose(Boolean disposing)
at Tensorflow.DisposableObject.Dispose()
at Microsoft.ML.Transforms.TensorFlowTransformer.Dispose(Boolean disposing) in D:\a\1\s\src\Microsoft.ML.TensorFlow\TensorflowTransform.cs:line 467
at Microsoft.ML.Transforms.TensorFlowTransformer.Finalize() in D:\a\1\s\src\Microsoft.ML.TensorFlow\TensorflowTransform.cs:line 450
```

### DIFF
--- a/src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
+++ b/src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
@@ -37,8 +37,10 @@ using static Tensorflow.Binding;
 
 namespace Microsoft.ML.Transforms
 {
-    public sealed class TensorFlowTransformer : RowToRowTransformerBase
+    public sealed class TensorFlowTransformer : RowToRowTransformerBase, IDisposable
     {
+        private bool _isDisposed;
+
         private readonly string _savedModelPath;
         private readonly bool _isTemporarySavedModel;
         private readonly bool _addBatchDimensionInput;
@@ -52,7 +54,6 @@ namespace Microsoft.ML.Transforms
         internal readonly (Operation, int)[] TFOutputOperations;
         internal TF_Output[] TFInputNodes;
         internal TF_Output[] TFOutputNodes;
-        internal IntPtr[] TFOperations;
         internal Graph Graph => Session.graph;
 
         internal readonly string[] Inputs;
@@ -444,17 +445,11 @@ namespace Microsoft.ML.Transforms
                 ctx.SaveNonEmptyString(colName);
         }
 
-        ~TensorFlowTransformer()
+        public void Dispose()
         {
-            Dispose(false);
-        }
+            if (_isDisposed)
+                return;
 
-        private void Dispose(bool disposing)
-        {
-            // Ensure that the Session is not null and it's handle is not Zero, as it may have already been disposed/finalized.
-            // Technically we shouldn't be calling this if disposing == false, since we're running in finalizer
-            // and the GC doesn't guarantee ordering of finalization of managed objects, but we have to make sure
-            // that the Session is closed before deleting our temporary directory.
             try
             {
                 if (Session != null && Session.graph != IntPtr.Zero)
@@ -467,19 +462,14 @@ namespace Microsoft.ML.Transforms
                     Session.close(); // invoked Dispose()
                 }
             }
-            catch (ObjectDisposedException ex)
-            {
-                using (var channel = Host.Start("TensorflowTransform"))
-                {
-                    channel.Warning($"Caught: {ex.Message} during TensorflowTransform dispose.");
-                }
-            }
             finally
             {
                 if (!string.IsNullOrEmpty(_savedModelPath) && _isTemporarySavedModel)
                 {
                     DeleteFolderWithRetries(Host, _savedModelPath);
                 }
+
+                _isDisposed = true;
             }
         }
 
--- a/src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
+++ b/src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
@@ -42,6 +42,7 @@ namespace Microsoft.ML.Transforms
         private readonly string _savedModelPath;
         private readonly bool _isTemporarySavedModel;
         private readonly bool _addBatchDimensionInput;
+        private readonly IHostEnvironment _env;
         internal readonly Session Session;
         internal readonly Runner Runner;
         internal readonly DataViewType[] OutputTypes;
@@ -195,6 +196,8 @@ namespace Microsoft.ML.Transforms
         {
             Contracts.CheckValue(env, nameof(env));
             env.CheckValue(options, nameof(options));
+
+            _env = env;
         }
 
         private static ITensorValueGetter CreateTensorValueGetter<T>(DataViewRow input, bool isVector, int colIndex, TensorShape tfShape)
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -42,6 +42,7 @@ public sealed class TensorFlowTransformer : RowToRowTransformerBase
         private readonly string _savedModelPath;
         private readonly bool _isTemporarySavedModel;
         private readonly bool _addBatchDimensionInput;
+        private readonly IHostEnvironment _env;
```
yaeldekel: >_env [](start = 42, length = 4)

The base class `RowToRowTransformerBase` has this field:
```
        private protected readonly IHost Host;
```
Can it be used instead of adding a new field?

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -457,14 +462,21 @@ private void Dispose(bool disposing)
...
-
-                if (Session != null && Session.graph != IntPtr.Zero)
+            }
+            catch (ObjectDisposedException ex)
+            {
+                using (var channel = _env.Start("TensorflowTransform"))
                 {
-                    Session.graph.Dispose();
+                    channel.Warning($"Caught: {ex.Message} during TensorflowTransform dispose.");
                 }
```
sharwell: :grey_exclamation: Managed resources cannot be used when `disposing` is false. Logging from this location is essentially impossible.

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -457,14 +462,21 @@ private void Dispose(bool disposing)
             // that the Session is closed before deleting our temporary directory.
             try
             {
+                if (Session != null && Session.graph != IntPtr.Zero)
+                {
+                    Session.graph.Dispose();
```
sharwell: 📝 This call is _not valid_ when `disposing` is false.

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -457,14 +462,21 @@ private void Dispose(bool disposing)
...
+                }
+
                 if (Session != null && Session != IntPtr.Zero)
                 {
                     Session.close(); // invoked Dispose()
```
sharwell: 📝 This call is _not valid_ when `disposing` is false.

sharwell:
@frank-dong-ms The easiest solution here is removing `~TensorFlowTransformer()`. This type does not (and must not) directly own any unmanaged resources, so the finalizer is not necessary. Instead, use the new `IDisposable` pattern where `IDisposable.Dispose()` is implemented directly and the `bool disposing` parameter is removed.

frank-dong-ms:
Check out below comments from original writer, seems this is done in purpose:

            // Ensure that the Session is not null and it's handle is not Zero, as it may have already been disposed/finalized.
            // Technically we shouldn't be calling this if disposing == false, since we're running in finalizer
            // and the GC doesn't guarantee ordering of finalization of managed objects, but we have to make sure
            // that the Session is closed before deleting our temporary directory.

---
In reply to: [337479664](https://github.com/dotnet/machinelearning/pull/4613#issuecomment-337479664) [](ancestors = 337479664)

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
frank-dong-ms: Yes, that is good point, I will modify the code as you suggested, thanks.

---
In reply to: [362313339](https://github.com/dotnet/machinelearning/pull/4613#discussion_r362313339) [](ancestors = 362313339)

frank-dong-ms:
>         private void Dispose(bool disposing)

we are cleaning managed resources even when disposing == false due to below comments:

            // Technically we shouldn't be calling this if disposing == false, since we're running in finalizer
            // and the GC doesn't guarantee ordering of finalization of managed objects, but we have to make sure
            // that the Session is closed before deleting our temporary directory.


---
Refers to: src/Microsoft.ML.TensorFlow/TensorflowTransform.cs:457 in 5c48a7a. [](commit_id = 5c48a7a7eaa2880d4e70b9c11f17af3036fa11ee, deletion_comment = False)

frank-dong-ms:
Thanks, I will go with your suggestions

---
In reply to: [570039229](https://github.com/dotnet/machinelearning/pull/4613#issuecomment-570039229) [](ancestors = 570039229)

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -444,27 +445,21 @@ private protected override void SaveModel(ModelSaveContext ctx)
...
-        private void Dispose(bool disposing)
-        {
-            // Ensure that the Session is not null and it's handle is not Zero, as it may have already been disposed/finalized.
-            // Technically we shouldn't be calling this if disposing == false, since we're running in finalizer
-            // and the GC doesn't guarantee ordering of finalization of managed objects, but we have to make sure
-            // that the Session is closed before deleting our temporary directory.
             try
             {
-                if (Session != null && Session != IntPtr.Zero)
+                if (Session != null && Session.graph != IntPtr.Zero)
```
sharwell: 💡 This can be simplified

```suggestion
                if (Session?.graph != IntPtr.Zero)
```

src/Microsoft.ML.TensorFlow/TensorflowTransform.cs
```diff
@@ -444,27 +445,21 @@ private protected override void SaveModel(ModelSaveContext ctx)
                 ctx.SaveNonEmptyString(colName);
         }
 
-        ~TensorFlowTransformer()
+        public void Dispose()
         {
-            Dispose(false);
-        }
+            if (_isDisposed)
+                return;
```
sharwell: 💭 I would have liked to see `Session` have a more idiomatic dispose behavior so this state tracking wasn't necessary, but I wasn't able to prove it to be the case.

sharwell:
@frank-dong-ms The title of this pull request should be updated to reflect the new behavior.

codemzs:
There is a similar dispose situation in ImageClassification as well

frank-dong-ms:
Yes, done, thanks

---
In reply to: [570624023](https://github.com/dotnet/machinelearning/pull/4613#issuecomment-570624023) [](ancestors = 570624023)

frank-dong-ms:
Cool, I will take a look as well

---
In reply to: [570625344](https://github.com/dotnet/machinelearning/pull/4613#issuecomment-570625344) [](ancestors = 570625344)

Issue #4615, opened 2020-01-01
Support saving as ONNX for ExpressionTransformer


yaeldekel:
This is needed, as mentioned in [this comment](https://github.com/dotnet/machinelearning/pull/4548#pullrequestreview-335990337).

matouskozak:
Triage: Closing as this work is tracked in https://github.com/dotnet/machinelearning/issues/6810

PR #4622, opened 2020-01-03
ToString Featurizer added

michaelgsharp:
This change adds in the ToStringTransformer into the new Featurizers project. It is the fourth of a series of PR's that will go in. The ToStringTransformer is implemented in native code, so this is mostly just a wrapper around that with the appropriate entrypoints for NimbusML as well.

The ToStringTransformer converts values into the appropriate string representations.

This code is auto generated. The functionality will be migrated into the existing type conversion over the next several weeks and this file will be removed, but its going in as is to unblock other development work.


src/Microsoft.ML.Featurizers/ToStringTransformer.cs
```diff
@@ -0,0 +1,1533 @@
...
+    /// <remarks>
+    /// <format type="text/markdown"><![CDATA[
+    ///
+    /// ###  Estimator Characteristics
+    /// |  |  |
+    /// | -- | -- |
+    /// | Does this estimator need to look at the data to train its parameters? | No |
+    /// | Input column data type | Scalar of numeric, boolean, [text](xref:Microsoft.ML.Data.TextDataViewType) |
+    /// | Output column data type | Scalar of [text](xref:Microsoft.ML.Data.TextDataViewType) type. |
+    ///
```
michaelgsharp: add onnx export

src/Microsoft.ML.Featurizers/ToStringTransformer.cs
```diff
@@ -0,0 +1,1533 @@
...
+        }
+    }
+
+    public sealed class ToStringTransformer : RowToRowTransformerBase, IDisposable
+    {
+        #region Class data members
+
+        internal const string Summary = "Turns the given column into a column of its string representation";
+        internal const string UserName = "ToString Transform";
+        internal const string ShortName = "ToStringTransform";
```
michaelgsharp: Fix short name.

src/Microsoft.ML.Featurizers/ToStringTransformer.cs
```diff
@@ -0,0 +1,1533 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
```
michaelgsharp: add comments about code generation.

michaelgsharp:
This code is auto generated. The functionality will be migrated into the existing type conversion over the next several weeks and this file will be removed, but its going in as is to unblock other development work.

src/Microsoft.ML.Featurizers/ToStringTransformer.cs
```diff
@@ -0,0 +1,1537 @@
...
+using Microsoft.ML.EntryPoints;
+using Microsoft.ML.Featurizers;
+using Microsoft.ML.Internal.Utilities;
+using Microsoft.ML.Runtime;
+using Microsoft.ML.Transforms;
+using Microsoft.Win32.SafeHandles;
+using static Microsoft.ML.Featurizers.CommonExtensions;
+
+[assembly: LoadableClass(typeof(ToStringTransformer), null, typeof(SignatureLoadModel),
+    ToStringTransformer.UserName, ToStringTransformer.LoaderSignature)]
```
justinormont: Shouldn't `ToStringTransformer` be part of the `ConvertTransform` ([info](https://github.com/dotnet/machinelearning/blob/0a5514b8d1a8a597976aecec6f463e4471b73d6d/docs/code/IDataViewTypeSystem.md#standard-conversions), [info](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.conversionsextensionscatalog.converttype?view=ml-dotnet), [code](https://github.com/dotnet/machinelearning/blob/e19369b407e8630cfaa82b07f81f0576c9bbd145/src/Microsoft.ML.Data/Data/Conversion.cs))?

The Convert transform's purpose is to convert the type of columns, for instance, convert a String to a Float, or an Int32 to Int64.

If I recall, the `ConvertTransform` didn't allow conversion to the String type from numeric; I'm unsure the reasoning, but it could be added.


Historically, the Expression transform does make the ToString() conversion easy: `xf=Expression{ col=ColAsString:MyNumericColumn expr={(x) : text(x)} }`, and this is [available in ML.NET](https://github.com/dotnet/machinelearning/blob/354798dc2d9431f5a6a9f27a46b2f50a7f2e3f57/docs/api-reference/expression-estimator.md#functions) (see: "​convert to text").

michaelgsharp: You are correct Justin. This is going in as is for now to unblock the work on ONNX export, and then over the next several weeks I will be converting this, as well as PRs 4594 and 4597, to be in the correct ML.NET transformer, and then removing these classes.

michaelgsharp: Does the Expression transform export to ONNX?

justinormont: There's an issue logged for ONNX for the Expression transform: https://github.com/dotnet/machinelearning/issues/4615

Hopefully the ONNX support is added. The particular AutoML ask was for basic math functions { `log(a)`, `exp(a)`, `a-b` }. With ONNX output for the Expression transform, it would cover those needs. Otherwise the discussion was making independent transforms for each of log/exp/subtract, which would end up with an endless parade of trivial transforms.

justinormont: /cc @davidbrownellWork : My recommendation for implementing the requested `log(a)`, `exp(a)`, and `a-b` transforms is the provide ONNX export for the Expression transform.

PR #4623, opened 2020-01-03
TimeSeriesImputer featurizer added

michaelgsharp:
This change adds in the TimeSeriesImputer into the new TimeSeriesImputer project. It is the final of a series of PR's that will go in. The TimeSeriesImputer is implemented in native code, so this is mostly just a wrapper around that with the appropriate entrypoints for NimbusML as well.

The TimeSeriesImputer imputes rows and columns based on the time series given. This is the first transformer that imputes rows in ML.NET.


src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -0,0 +1,637 @@
...
+        {
+            [Argument(ArgumentType.Required, HelpText = "Column representing the time", Name = "TimeSeriesColumn", ShortName = "time", SortOrder = 1)]
+            public string TimeSeriesColumn;
+
+            [Argument((ArgumentType.MultipleUnique | ArgumentType.Required), HelpText = "List of grain columns", Name = "GrainColumns", ShortName = "grains", SortOrder = 2)]
+            public string[] GrainColumns;
+
+            // This transformer adds columns
+            [Argument(ArgumentType.MultipleUnique, HelpText = "Columns to filter", Name = "FilterColumns", ShortName = "filters", SortOrder = 2)]
+            public string[] FilterColumns;
```
justinormont: What's the use case for filtering? Are more columns produced besides `IsImputed`? Does the filtering operate only on the newly produced columns, or can I remove any pre-existing column?

michaelgsharp: If a column is filtered, whenever a row is imputed the `default` value for that row is used to fill in the value. Using the `FilterMode`, you can either use this list as an include or exclude list.

justinormont: > If a column is filtered, whenever a row is imputed the default value for that row is used to fill in the value.

I'm not sure I'm parsing this correctly, and I think I'm missing something obvious. If a column is filtered, why do we fill in a value and then drop the column? Are there other columns produced besides `IsImputed`? What other columns would a user use the filtering to keep/remove besides `IsImpted`?

michaelgsharp: We aren't dropping any columns in this process. If a column is excluded from being imputed, the default value is provided when a new row is generated. The biggest use case for this is dealing with columns that aren't supported for imputation. You can exclude that column from being imputed, but still have the default value filled in automatically.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -0,0 +1,637 @@
...
+            public bool SupressTypeErrors = false;
+        }
+
+        #endregion
+
+        #region Class Enums
+
+        public enum ImputationStrategy : byte
+        {
+            ForwardFill = 1, BackFill, Median, Interpolate
```
michaelgsharp: Separate lines.
Hard coded numbers.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -0,0 +1,637 @@
...
+    public static class TimeSeriesImputerExtensionClass
+    {
+        /// <summary>
+        /// Create a <see cref="TimeSeriesImputerEstimator"/>, Imputes missing rows and column data per grain. Operates on all columns in the IDataView. Currently only numeric columns are supported.
+        /// </summary>
+        /// <param name="catalog">The transform catalog.</param>
+        /// <param name="timeSeriesColumn">Column representing the time series. Should be of type <see cref="long"/></param>
+        /// <param name="grainColumns">List of columns to use as grains</param>
+        /// <param name="imputeMode">Mode of imputation for missing values in column. If not passed defaults to forward fill</param>
+        public static TimeSeriesImputerEstimator TimeSeriesImputer(this TransformsCatalog catalog, string timeSeriesColumn, string[] grainColumns, TimeSeriesImputerEstimator.ImputationStrategy imputeMode = TimeSeriesImputerEstimator.ImputationStrategy.ForwardFill) =>
```
michaelgsharp: Change name to verb.
ReplaceMissingTimeSeriesValues?
ImputeTimeSeriesValues?

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -0,0 +1,637 @@
...
+    /// <remarks>
+    /// <format type="text/markdown"><![CDATA[
+    ///
+    /// ###  Estimator Characteristics
+    /// |  |  |
+    /// | -- | -- |
+    /// | Does this estimator need to look at the data to train its parameters? | Yes |
+    /// | Input column data type | All Types |
+    /// | Output column data type | All Types |
+    ///
```
michaelgsharp: Add onnx export comment.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -0,0 +1,637 @@
...
+
+        /// <summary>
+        /// Create a <see cref="TimeSeriesImputerEstimator"/>, Imputes missing rows and column data per grain. Operates on a filtered list of columns in the IDataView.
+        /// If a column is not imputed but rows are added then it will be filled with the default value for that data type. Currently only numeric columns are supported for imputation.
+        /// </summary>
+        /// <param name="catalog">The transform catalog.</param>
+        /// <param name="timeSeriesColumn">Column representing the time series. Should be of type <see cref="long"/></param>
+        /// <param name="grainColumns">List of columns to use as grains</param>
+        /// <param name="filterColumns">List of columns to filter. If <paramref name="filterMode"/> is <see cref="TimeSeriesImputerEstimator.FilterMode.Exclude"/> than columns in the list will be ignored.
+        /// If <paramref name="filterMode"/> is <see cref="TimeSeriesImputerEstimator.FilterMode.Include"/> than values in the list are the only columns imputed.</param>
```
michaelgsharp: Remove IF

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -0,0 +1,637 @@
...
+   TimeSeriesImputerTransformer.UserName, TimeSeriesImputerTransformer.LoaderSignature)]
+
+[assembly: EntryPointModule(typeof(TimeSeriesTransformerEntrypoint))]
+
+namespace Microsoft.ML.Featurizers
+{
+    public static class TimeSeriesImputerExtensionClass
+    {
+        /// <summary>
+        /// Create a <see cref="TimeSeriesImputerEstimator"/>, Imputes missing rows and column data per grain. Operates on all columns in the IDataView. Currently only numeric columns are supported.
```
michaelgsharp: Add comment about chained estimator issues.

michaelgsharp: and specifics around strings/suppress type errors.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -0,0 +1,637 @@
...
+        }
+    }
+
+    public sealed class TimeSeriesImputerTransformer : ITransformer, IDisposable
+    {
+        #region Class data members
+
+        internal const string Summary = "Fills in missing row and values";
+        internal const string UserName = "TimeSeriesImputer";
+        internal const string ShortName = "TimeSeriesImputer";
```
michaelgsharp: Change to TSI

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -0,0 +1,637 @@
...
+        private readonly string[] _dataColumns;
+        private readonly string[] _allColumnNames;
+        private readonly bool _suppressTypeErrors;
+        private readonly TimeSeriesImputerEstimator.ImputationStrategy _imputeMode;
+        internal TransformerEstimatorSafeHandle TransformerHandle;
+
+        #endregion
+
+        // Normal constructor.
+        internal TimeSeriesImputerTransformer(IHostEnvironment host, string timeSeriesColumn, string[] grainColumns, string[] filterColumns, TimeSeriesImputerEstimator.FilterMode filterMode, TimeSeriesImputerEstimator.ImputationStrategy imputeMode, bool suppressTypeErrors, IDataView input)
```
michaelgsharp: Change parameters to just pass in the options class.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -0,0 +1,637 @@
...
+                _dataColumns[i] = ctx.Reader.ReadString();
+
+            _imputeMode = (TimeSeriesImputerEstimator.ImputationStrategy)ctx.Reader.ReadByte();
+
+            _allColumnNames = new string[1 + _grainColumns.Length + _dataColumns.Length];
+            _allColumnNames[0] = _timeSeriesColumn;
+            Array.Copy(_grainColumns, 0, _allColumnNames, 1, _grainColumns.Length);
+            Array.Copy(_dataColumns, 0, _allColumnNames, 1 + _grainColumns.Length, _dataColumns.Length);
+
+            var nativeData = ctx.Reader.ReadByteArray();
```
michaelgsharp: rename to something with state.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -0,0 +1,637 @@
...
+            TransformerHandle = CreateTransformerFromSavedData(nativeData);
+        }
+
+        private unsafe TransformerEstimatorSafeHandle CreateTransformerFromSavedData(byte[] data)
```
michaelgsharp: rename param to state.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -0,0 +1,637 @@
...
+        private unsafe TransformerEstimatorSafeHandle CreateTransformerFromEstimator(IDataView input)
+        {
+            IntPtr estimator;
+            IntPtr errorHandle;
+            bool success;
+
+            var allColumns = input.Schema.Where(x => _allColumnNames.Contains(x.Name)).Select(x => TypedColumn.CreateTypedColumn(x, _dataColumns)).ToDictionary(x => x.Column.Name);
+
+            // Create buffer to hold binary data
+            var columnBuffer = new byte[1024];
```
michaelgsharp: Bump up to 4k

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -0,0 +1,637 @@
...
+
+                success = CreateTransformerFromEstimatorNative(estimatorHandler, out IntPtr transformer, out errorHandle);
+                if (!success)
+                    throw new Exception(GetErrorDetailsAndFreeNativeMemory(errorHandle));
+
+                return new TransformerEstimatorSafeHandle(transformer, DestroyTransformerNative);
+            }
+        }
+
+        private void BuildColumnByteArray(Dictionary<string, TypedColumn> allColumns, ref byte[] columnByteBuffer, out int bufferLength)
```
michaelgsharp: Change to something with serialized data size.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -0,0 +1,637 @@
...
+        /// </summary>
+        /// <param name="catalog">The transform catalog.</param>
+        /// <param name="timeSeriesColumn">Column representing the time series. Should be of type <see cref="long"/></param>
+        /// <param name="grainColumns">List of columns to use as grains</param>
+        /// <param name="filterColumns">List of columns to filter. If <paramref name="filterMode"/> is <see cref="TimeSeriesImputerEstimator.FilterMode.Exclude"/> than columns in the list will be ignored.
+        /// If <paramref name="filterMode"/> is <see cref="TimeSeriesImputerEstimator.FilterMode.Include"/> than values in the list are the only columns imputed.</param>
+        /// <param name="filterMode">Whether the list <paramref name="filterColumns"/> should include or exclude those columns.</param>
+        /// <param name="imputeMode">Mode of imputation for missing values in column. If not passed defaults to forward fill</param>
+        /// <param name="suppressTypeErrors">Supress the errors that would occur if a column and impute mode are imcompatible. If true, will skip the column. If false, will stop and throw an error.</param>
```
justinormont: Spelling:
```suggestion
        /// <param name="suppressTypeErrors">Suppress the errors that would occur if a column and impute mode are imcompatible. If true, will skip the column. If false, will stop and throw an error.</param>
```

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
justinormont: We may want to rename it from `FilterColumns` to perhaps `ColumnsToImpute` or `ColumnsToFillDefault`.  

The naming suggestion is to disambiguate from the existing concept in `ML.NET` of _filtering_, which removes rows of data from the IDataView ([info](https://docs.microsoft.com/en-us/dotnet/machine-learning/how-to-guides/prepare-data-ml-net#filter-data)).

Would another way of handling the non-supported column types be to automatically fill w/ default when needed? Seems we could auto-handle this use case. Assuming there aren't other use cases, this would remove the need of the `FilterColumns` field.

michaelgsharp: I think renaming is a good idea. Its currently setup to where you can either list the columns to fill with default or columns to impute based on the `FilterMode` paramenter. So the name needs to somehow show that it can do both based on that parameter.

We could automatically fill in default values when needed. I didn't do that as this could potentially mask errors if you wanted to convert something prior and forgot. If you don't think thats a big deal though and think it would be better then I am not opposed to switching to that.

justinormont: Auto-filling w/ default for unsupported types is likely good from the data science side. Then we note in the docs "TimeSeriesImputer supports these data types (...) and will fill with the [default value](https://github.com/dotnet/machinelearning/blob/0a5514b8d1a8a597976aecec6f463e4471b73d6d/docs/code/IDataViewTypeSystem.md#default-value) for columns of unsupported data types".

@EricWrightAtWork -- since you know more about the needs of time-series forecasting, any thoughts? Is auto-filling unsupported datatypes with their default values reasonable?

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
justinormont: Thanks
#resolved

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -0,0 +1,642 @@
...
+
+            internal override byte[] GetSerializedValue()
+            {
+                dynamic value = GetValue();
+                byte[] bytes;
+                if (value.GetType() == typeof(byte))
+                    bytes = new byte[1] { value };
+                if (BitConverter.IsLittleEndian)
+                    bytes = BitConverter.GetBytes(value);
+                // Will need to enable this when Jin's pr goes in. return ((IEnumerable<byte>)BitConverter.GetBytes(value)).Reverse().ToArray();
```
michaelgsharp: Remove comment.

src/Microsoft.ML.Featurizers/TimeSeriesImputerDataView.cs
```diff
@@ -0,0 +1,761 @@
...
+{
+
+    internal sealed class TimeSeriesImputerDataView : IDataTransform
+    {
+        #region Typed Columns
+        private TimeSeriesImputerTransformer _parent;
+        public class SharedColumnState
+        {
+            public bool SourceCanMoveNext { get; set; }
+            public int TransformedDataPosition { get; set; }
```
michaelgsharp: Change to long

src/Microsoft.ML.Featurizers/TimeSeriesImputerDataView.cs
```diff
@@ -0,0 +1,761 @@
...
+            public NativeBinaryArchiveData[] TransformedData { get; set; }
```
michaelgsharp: Add note why this is array and how it works.

src/Microsoft.ML.Featurizers/TimeSeriesImputerDataView.cs
```diff
@@ -0,0 +1,761 @@
...
+                        if (!success)
+                            throw new Exception(GetErrorDetailsAndFreeNativeMemory(errorHandle));
+
+                        if (outputDataSize.ToInt32() > 0)
+                        {
+                            SharedState.TransformedDataHandler = new TransformedDataSafeHandle((IntPtr)outputData, outputDataSize);
+                            SharedState.TransformedData = new NativeBinaryArchiveData[outputDataSize.ToInt32()];
+                            for (int i = 0; i < outputDataSize.ToInt32(); i++)
+                            {
+                                SharedState.TransformedData[i] = *(outputData + i);
```
michaelgsharp: Change to span

src/Microsoft.ML.Featurizers/TimeSeriesImputerDataView.cs
```diff
@@ -0,0 +1,761 @@
...
+                            }
+                            SharedState.TransformedDataPosition = 0;
+                        }
+                    }
+
+                    // Base case where we didn't impute anything
+                    if (!allImputedColumnNames.Contains(Column.Name))
+                    {
+                        var imputedData = SharedState.TransformedData[SharedState.TransformedDataPosition];
+                        if (BoolTypedColumn.GetBoolFromNativeBinaryArchiveData(imputedData.Data, 0))
```
michaelgsharp: Add comment explaining this.

src/Microsoft.ML.Featurizers/TimeSeriesImputerDataView.cs
```diff
@@ -0,0 +1,761 @@
...
+                IsNullable = isNullable;
+            }
+
+            internal override byte[] GetSerializedValue()
+            {
+                dynamic value = GetSourceValue();
+                byte[] bytes;
+                if (value.GetType() == typeof(byte))
+                    bytes = new byte[1] { value };
+                if (BitConverter.IsLittleEndian)
```
michaelgsharp: Remove this check.

michaelgsharp: and throw exception if not little endian.

src/Microsoft.ML.Featurizers/TimeSeriesImputerDataView.cs
```diff
@@ -0,0 +1,761 @@
...
+                           val = new DataViewRowId((ulong)Position, 0);
+                       };
+            }
+
+            public sealed override DataViewSchema Schema => _schema;
+
+            /// <summary>
+            /// Since rows will be generated all columns are active
+            /// </summary>
+            public override bool IsColumnActive(DataViewSchema.Column column)
```
michaelgsharp: use lambda

PR #4624, opened 2020-01-03
remove unnecessary finalizer from imageclassification trainer

frank-dong-ms:
1. remove unnecessary finalizer from imageclassification trainer
2. simplify conditional check

### DIFF
--- a/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
+++ b/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
@@ -77,8 +77,10 @@ namespace Microsoft.ML.Vision
     /// </remarks>
     public sealed class ImageClassificationTrainer :
         TrainerEstimatorBase<MulticlassPredictionTransformer<ImageClassificationModelParameters>,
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -1490,26 +1490,22 @@ public void Score(in VBuffer<byte> image, Span<float> classProbabilities)
             return (ValueMapper<TSrc, TDst>)(Delegate)del;
         }
 
-        ~ImageClassificationModelParameters()
+        public void Dispose()
```
codemzs: >Dispose [](start = 20, length = 7)

I understand this is the right thing to do but who will call Dispose? Currently the framework does not support disposing transforms, in the past by not freeing the session we have seen memory leak in the scenario where a user creates multiple pipelines.

sharwell: `Session` and `Graph` both have finalizers.

Issue #3436, opened 2019-04-19
Normalizers multicolumn samples


sfilipi:
All transform extension APIs have been documented as part of #1209. 
The only samples left to be created are the ones about normalizer extensions operating on multiple columns. 

Replicate PR #3435 for the other normalizer APIs

Akshive:
I would like to work on this, but I need help to start as I am new to open source world.

jwood803:
From what I can tell these are all the normalizers and if they are have a multi column sample:

| Normalizer| Has Multi Column Sample? |
| ----------------------------------- | ----------------------- |
| Normalize Binning             | &check;                 |
| Normalize Global Contrast | |
| Normalize Log Mean Variance | |
| Normalize LP Norm | |
| Normalize Mean Variance | |
| Normalize MinMax | &check; |
| Normalize Supervised Binning | |

The `NormalizeGlobalContrast` and `NormalizeLpNorm` normalizers don't seem to have an overload to take in an array of `InputOutputColumnPair`.

PR #4644, opened 2020-01-11
NormalizeMinMax Multicolumn example

jwood803:
Update for #3436 for the NormalizeMinMax transform.

@sfilipi Took a shot at adding a multicolumn sample. If this is on the right track, I can help with some of the others.


docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/NormalizeMinMaxMulticolumn.cs
```diff
@@ -0,0 +1,133 @@
...
+using Microsoft.ML.Data;
+using static Microsoft.ML.Transforms.NormalizingTransformer;
+
+namespace Samples.Dynamic
+{
+    class NormalizeMinMaxMulticolumn
+    {
+        public static void Example()
+        {
+            // Create a new ML context, for ML.NET operations. It can be used for
```
natke: I don't think we need this comment here, as we are not demonstrating usage of MLContext

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/NormalizeMinMaxMulticolumn.cs
```diff
@@ -0,0 +1,133 @@
...
+                string.Join(", ", column2FixZero[i].Select(x => x.ToString("f4"))));
+
+            // Expected output:
+            // Features                                Features2  
+            // 0.5000, 0.5000, 1.0000, 0.0000          0.1111, 0.2857, 0.3750
+            // 1.0000, 1.0000, 0.6667, 0.0000          0.3333, 0.5714, 0.6250
+            // 0.0000, 0.0000, 0.3333, 0.0000          0.6667, 1.0000, 1.0000
+            // -0.5000, -0.5000, -0.3333, 1.0000       1.0000, 0.0000, 0.5000
+
+            // Let's get transformation parameters. Since we work with only one
```
natke: Let's get -> Get

natke: Since we work with only one -> Since there is only one

natke: Don't we have two columns here? Features and Features2?

jwood803: Updated the wording of the comment, but I'll need to see how the formula changes since there are two columns instead of one.

docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/NormalizeMinMaxMulticolumn.cs
```diff
@@ -0,0 +1,135 @@
...
+                    new InputOutputColumnPair("Features"),
+                    new InputOutputColumnPair("Features2")
+                },
+                fixZero: false);
+
+            var columnPair = new[]
+            {
+                new InputOutputColumnPair("Features"),
+                new InputOutputColumnPair("Features2")
+            };
```
najeeb-kazmi: Why not re-use this array in the fixZero: false transform above?

najeeb-kazmi:
@jwood803 please pull master, so that you have [this commit](https://github.com/dotnet/machinelearning/commit/97c4533df05314da44f3b784fde08bd8379488c9). Then the tests that fail randomly should pass, so we can merge your PR.

najeeb-kazmi:
@natke does this look good to you? If so, let's try to merge this before release so that the example shows up in the docs.

PR #4645, opened 2020-01-11
Use a GUID when creating the temp path

### LINKED ISSUE #568: Temporary filename collision
### System information

- **OS version/distro**: Windows Server 2016

### Issue

When running simultaneous sweeps (multiple processes), one process tried to open a new temp file which already exists. 

`Unexpected exception: The file 'C:\Users\jormont\AppData\Local\Temp\2\TLC_25553D77\0' already exists`

- **What did you do?**
Running an AutoML macro sweep w/ four processes on the same machine. 

Perhaps our temporary path names aren't unique enough? 

https://github.com/dotnet/machinelearning/blob/211c0439aa7bec33747536a273f90b1718ead24a/src/Microsoft.ML.Data/Model/Repository.cs#L507

Unsure which call is used to create the `TLC_...` folder, but could be this one:
https://github.com/dotnet/machinelearning/blob/c023727b76970ab913ec1ce38276508835c17bcf/src/Microsoft.ML.Core/Environment/TlcEnvironment.cs#L398

### Source code / logs

Error:
```
Unexpected exception: The file 'C:\Users\jormont\AppData\Local\Temp\2\TLC_25553D77\0' already exists., 'System.IO.IOException'
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
   at System.IO.Compression.ZipFileExtensions.ExtractToFile(ZipArchiveEntry source, String destinationFileName, Boolean overwrite)
   at Microsoft.ML.Runtime.Model.RepositoryReader.OpenEntryOrNull(String dir, String name) in D:\TLC\OpenSource\src\Microsoft.ML.Data\Model\Repository.cs:line 507
   at Microsoft.ML.Runtime.Model.ModelLoadContext.LoadModelOrNull[TRes,TSig](IHostEnvironment env, TRes& result, RepositoryReader rep, String dir, Object[] extra) in D:\TLC\OpenSource\src\Microsoft.ML.Data\Model\ModelLoading.cs:line 55
   at Microsoft.ML.Runtime.Model.ModelLoadContext.LoadModel[TRes,TSig](IHostEnvironment env, TRes& result, RepositoryReader rep, String dir, Object[] extra) in D:\TLC\OpenSource\src\Microsoft.ML.Data\Model\ModelLoading.cs:line 89
   at Microsoft.ML.Runtime.Data.ApplyTransformUtils.ApplyTransformToData(IHostEnvironment env, IDataTransform transform, IDataView newSource) in D:\TLC\OpenSource\src\Microsoft.ML.Data\Utilities\ApplyTransformUtils.cs:line 47
   at Microsoft.ML.Runtime.Data.ApplyTransformUtils.ApplyAllTransformsToData(IHostEnvironment env, IDataView chain, IDataView newSource, IDataView oldSource) in D:\TLC\OpenSource\src\Microsoft.ML.Data\Utilities\ApplyTransformUtils.cs:line 101
   at Microsoft.ML.Runtime.EntryPoints.TransformModel..ctor(IHostEnvironment env, IDataView result, IDataView input) in D:\TLC\OpenSource\src\Microsoft.ML.Data\EntryPoints\TransformModel.cs:line 62
   at Microsoft.ML.Runtime.Data.TextAnalytics.TextTransform(IHostEnvironment env, Arguments input) in D:\TLC\source\TextAnalytics\TextTransform.cs:line 549
```

jwood803:
Addresses #568.

@justinormont Not sure if the placement of it is the best, but it should still prevent the collisions.

Edit: Fixes #568 (added so that issue #568 will close automatically when this PR is successfully merged.)


src/Microsoft.ML.Core/Data/Repository.cs
```diff
@@ -256,7 +256,7 @@ protected void GetPath(out string pathEnt, out string pathTemp, string dir, stri
             string root = Path.GetFullPath(DirTemp ?? @"x:\dummy");
             string entityPath = Path.Combine(root, dir ?? "", name);
             entityPath = Path.GetFullPath(entityPath);
-            string tempPath = Path.Combine(root, PathMap.Count.ToString());
+            string tempPath = Path.Combine(root, Guid.NewGuid().ToString());
```
harishsk: Can you please help me understand the problem you are trying to fix?

src/Microsoft.ML.Core/Data/Repository.cs
```diff
@@ -340,7 +340,9 @@ public Entry CreateEntry(string dir, string name)
 
             Stream stream;
             if (pathTemp != null)
+            {
```
harishsk: This is probably counter to other coding guidelines. But it seems that in this repository we are not using braces for single line blocks. Can you please revert this to stay aligned with that?

src/Microsoft.ML.Core/Data/Repository.cs
justinormont: Currently, there's a race condition in creating temporary files, causing a naming collision. 

The paths are created as `C:\Users\jormont\AppData\Local\Temp\2\TLC_25553D77\0`, where:
1. `TLC_` is fixed
2. `25553D77` is a seeded random in hex (bad)
3. `0` is the count [0..N] of folders existing in the parent folder (also bad).

The prefix `TLC_` (1) is fine, though should be updated to the current name, perhaps to `ML_NET_`. Or `Microsoft_ML_`, though I'd worry about creating too long of paths under Windows ~260 chars, especially as a we move towards a GUID. Note users can have long usernames, plus the names of the files placed in this path.

First issue -- The seeded random number (2) is an issue since running parallelly (or rerunning) with the same seed creates the same "random" folder name. This part is a deterministic collision.

Second issue -- The race condition then occurs when counting how many folders (3) are in the parent folder. When multiple processes/threads check the contents of the folder, they all receive the same number.  The purposed fix replaces this count [0..N] with a GUID.

The need is to have unique folders for creating temporary files, otherwise they clobber each other. This causing a crash and has the possibility of silent failures, especially on macOS/Linux where you can delete and overwrite files with open handles.

justinormont: Pushing deeper into the code, the random number (2) is not seeded, but uses the system seed which is set from the current time. I expect the collision occurs due the remark in `System.Random` docs, "On most Windows systems, Random objects created within 15 milliseconds of one another are likely to have identical seed values." ([src](https://docs.microsoft.com/en-us/dotnet/api/system.random?view=netcore-3.1)). Hence another race condition in (2).

Hopefully the .NET GUIDs don't have the same 15ms race condition as the RNG.

src/Microsoft.ML.Core/Data/Repository.cs
jwood803: Reverted the braces

src/Microsoft.ML.Core/Data/Repository.cs
harishsk: In the current code, it looks like we are using our own random number generation, which seems odd. 
It is okay to use GUIDs. But they seem a bit of overkill. 

[Path.GetRandomFileName](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getrandomfilename?view=netframework-4.8) seems to be intended for this kind of scenario. 
Would that work better?

justinormont: @harishsk: You're right, `Path.GetRandomFileName()` seems more purpose built than a GUID.

Code behind this function -- https://github.com/dotnet/runtime/blob/f57d49bb711fda8ac0bd231229d91919f4fca8e8/src/libraries/System.Private.CoreLib/src/System/IO/Path.cs#L249-L265

jwood803: @harishsk I had no idea that method existed. Thanks for letting me know! Updated to use it.

src/Microsoft.ML.Core/Data/Repository.cs
```diff
@@ -256,7 +256,7 @@ protected void GetPath(out string pathEnt, out string pathTemp, string dir, stri
             string root = Path.GetFullPath(DirTemp ?? @"x:\dummy");
             string entityPath = Path.Combine(root, dir ?? "", name);
             entityPath = Path.GetFullPath(entityPath);
-            string tempPath = Path.Combine(root, PathMap.Count.ToString());
+            string tempPath = Path.Combine(root, Path.GetRandomFileName());
             tempPath = Path.GetFullPath(tempPath);
 
             string parent = Path.GetDirectoryName(entityPath);
```
justinormont: The current function here, `GetPath()`, creates the path. It's called below in `CreateEntry()`:
https://github.com/dotnet/machinelearning/blob/c29a00b5f6089fb5eba17b26f826b791e63236c4/src/Microsoft.ML.Core/Data/Repository.cs#L335-L339

There another minor race condition in how the new path is being checked/added to the dictionary. It's a nit, and minor enough to ignore if you want.

If we want to handle in this PR, we could fix by either adding a lock around the IF/ELSE, or by moving to a [ConcurrentDictionary](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2) and using its `TryAdd()` to decide if to throw.

_(I tagged this on a line of code to allow for a threaded discussion)_

jwood803: Went the `ConcurrentDictionary` route, but let me know if that's not the correct implementation of it.

src/Microsoft.ML.Core/Data/Repository.cs
```diff
@@ -333,10 +334,8 @@ public Entry CreateEntry(string dir, string name)
             string pathEnt;
             string pathTemp;
             GetPath(out pathEnt, out pathTemp, dir, name, true);
-            if (PathMap.ContainsKey(pathEnt))
+            if (PathMap.TryAdd(pathEnt, pathTemp))
```
justinormont: Assuming I'm reading the docs right, `TryAdd()` returns `false` when it doesn't add ([docs](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.tryadd?view=netcore-3.0#examples)). 

Hence need to invert the return value.
```suggestion
            if (!PathMap.TryAdd(pathEnt, pathTemp))
```

src/Microsoft.ML.Core/Data/Repository.cs
```diff
@@ -525,7 +524,7 @@ public Entry OpenEntryOrNull(string dir, string name)
                     // Extract to a temporary file.
                     Directory.CreateDirectory(Path.GetDirectoryName(pathTemp));
                     entry.ExtractToFile(pathTemp);
-                    PathMap.Add(pathLower, pathTemp);
+                    PathMap.TryAdd(pathLower, pathTemp);
```
justinormont: To maintain similar to the previous behavior, we'll have to throw if the `TryAdd()` returns false.

The previous `Dictionary.Add()` throws an `ArgumentException` if the value being added already exists ([docs](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.add?view=netcore-3.0#exceptions)).

```suggestion
                    if (!PathMap.TryAdd(pathLower, pathTemp))
                        throw ExceptionContext.ExceptParam(nameof(name), "Duplicate entry: '{0}'", pathLower); 
```

src/Microsoft.ML.Core/Data/Repository.cs
jwood803: Whoops! I thought I added that, but I must have missed it. Thanks for catching!

src/Microsoft.ML.Core/Data/Repository.cs
jwood803: Updated to throw exception.

justinormont:
I restarted the failed tests, hopefully they succeed.

mstfbl:
Restarting failed tests one more time

justinormont:
@jwood803: Can you pull in changes from the master branch? This should allow the CI tests to pass (various flaky unit tests were temporarily disabled).

PR #4647, opened 2020-01-13
Fixed xref formatting

### LINKED ISSUE #4611: Broken cross references in estimator transformer docs
See https://review.docs.microsoft.com/en-us/dotnet/api/microsoft.ml.transforms.expressionestimator?view=ml-dotnet&branch=smoke-test

The xrefs should render as hyperlinks to the relevant API docs.

Xrefs not rendering:
- <xref:Microsoft.ML.Data.ExpressionTransformer> This one might be because it is new
- <xref:Microsof.ML.Data.VectorDataViewType> This one should work

mstfbl:
Fixes #4611 

Updated cross-references of ML.Data.ExpressionTransformer and ML.Data.VectorDataViewType to the correct format.

### DIFF
--- a/src/Microsoft.ML.Transforms/ExpressionTransformer.cs
+++ b/src/Microsoft.ML.Transforms/ExpressionTransformer.cs
@@ -29,29 +29,29 @@ using Microsoft.ML.Transforms;
 
 namespace Microsoft.ML.Transforms
 {
-    /// <summary>
-    /// This estimator applies a user provided expression (specified as a string) to input column values to produce new output column values.
-    /// </summary>
-    /// <remarks>
-    /// <format type="text/markdown"><![CDATA[
-    ///
-    /// ###  Estimator Characteristics
-    /// |  |  |
-    /// | -- | -- |
-    /// | Does this estimator need to look at the data to train its parameters? | No |
-    /// | Input column data type | float, double, int, long, bool or text.  |
-    /// | Output column data type | Can be float, double, int, long, bool or text, depending on the expression. |
-    ///
-    /// The resulting <xref:Microsoft.ML.Data.ExpressionTransformer> creates a new column,
-    /// named as specified in the output column name parameters, where the expression is applied to the input values.
-    /// At most one of the input columns can be of type <xref:Microsof.ML.Data.VectorDataViewType>, and when the input contains a vector column, the expression
-    /// is computed independently on each element of the vector, to create a vector output with the same length as that input.
-    ///
-    /// [!include[algorithm](~/../docs/samples/docs/api-reference/expression-estimator.md)]
-    /// ]]></format>
-    /// </remarks>
-    /// <seealso cref="ExpressionCatalog.Expression(TransformsCatalog, string, string, string[])"/>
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.Transforms/ExpressionTransformer.cs
```diff
@@ -29,29 +29,29 @@
...
+	/// <format type="text/markdown"><![CDATA[
+	///
+	/// ###  Estimator Characteristics
+	/// |  |  |
+	/// | -- | -- |
+	/// | Does this estimator need to look at the data to train its parameters? | No |
+	/// | Input column data type | float, double, int, long, bool or text.  |
+	/// | Output column data type | Can be float, double, int, long, bool or text, depending on the expression. |
+	///
+	/// The resulting [ExpressionTransformer](xref:Microsoft.ML.Data.ExpressionTransformer) creates a new column,
```
sharwell:
```suggestion
	/// The resulting <see cref="ExpressionTransformer"/> creates a new column,
```

src/Microsoft.ML.Transforms/ExpressionTransformer.cs
```diff
@@ -29,29 +29,29 @@
...
+	/// named as specified in the output column name parameters, where the expression is applied to the input values.
+	/// At most one of the input columns can be of type [VectorDataViewType](xref:Microsof.ML.Data.VectorDataViewType), and when the input contains a vector column, the expression
```
sharwell:
```suggestion
	/// At most one of the input columns can be of type <see cref="VectorDataViewType"/>, and when the input contains a vector column, the expression
```

mstfbl:
@sharwell Silly Visual Studio setting, fixed, thanks!

src/Microsoft.ML.Transforms/ExpressionTransformer.cs
natke: Actually this is within a markdown block, so it won't be resolved properly in the API docs.

natke: The reason that it is in a markdown block is for the table. We could not discover any way of rendering a table without markdown

sharwell: The compiler will convert this to the full form. I would expect the markdown processor to still be able to handle this case:

```
<see cref="T:Microsoft.ML.Data.VectorDataViewType"/>
```

Since it's trivial to convert this to the Markdown xref form and treat it the same way. If it's not already possible, it would still be better to update the Markdown conversion step so we get true validation on both sides instead of assuming the form in code is correct (which right now, it is not).

sharwell: > We could not discover any way of rendering a table without markdown

Well I can help with that!
http://www.blackwasp.co.uk/DocumentationLists_2.aspx

If you want, you can install [Documentation Analyzers](https://www.nuget.org/packages/DotNetAnalyzers.DocumentationAnalyzers/1.0.0-beta.59) and you'll have a refactoring available to convert Markdown to XML documentation. It's a one-way conversion, but helpful for first-time comment writing and if you forget.

PR #4650, opened 2020-01-13
Cancellation in Image Classification (fixes #4632)

### LINKED ISSUE #4632: image classification needs cancel
Hi, is there any plan to add cancel support for image classification API so that we can cancel an image classification training by passing in a cancellation token? 

[related issue](https://app.zenhub.com/workspaces/mlnet-tools-5cde1c97e3106e39e8ae08fc/issues/dotnet/machinelearning-modelbuilder/413)

antoniovs1029:
Adds support for cancellation to the Image Classification trainer in a similar manner as done in #3062 (and [other PRs](https://github.com/dotnet/machinelearning/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+cancellation++)) by adding cancellation checkpoints to the train method.

I've tested it by running the sample related to this trainer. Since the other PR's that included checkpoints for cancellation don't include unit tests, I also didn't include any in here.

Fixes #4632 .


antoniovs1029:
I don't know if I should also add a `.CheckAlive()` chekpoint inside the `CacheFeaturizedImagesToDisk` method of Image Classification Trainer, as that method can take a couple of minutes, but once the method is over, the trainer will anyway end hitting the checkpoint I've already added in `TrainAndEvaluateClassificationLayer`.

Also, if anyone has other opinions as to where to put more checkpoints, please, let me know!

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -992,6 +995,7 @@ public Tensor ProcessImage(in VBuffer<byte> imageBuffer)
 
                 for (int epoch = 0; epoch < epochs; epoch += 1)
                 {
+                    Host.CheckAlive();
```
codemzs: >Host.CheckAlive(); [](start = 20, length = 18)

I would just put the check in this loop and in the CreateFeaturizedCacheFile. Please also report numbers in perf differences before and after. Please remove CheckAlive from everywhere else as its not very significant and only pollutes the code. You also need to call TryCleanupTemporaryWorkspace for a graceful termination.

codemzs:
CacheFeaturizedImagesToDisk  can take significant time, we must add there. 

---
In reply to: [573918512](https://github.com/dotnet/machinelearning/pull/4650#issuecomment-573918512) [](ancestors = 573918512)

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
antoniovs1029: I have added a new method "CheckAlive" to the ImageClassification trainer, with a try...catch to call TryCleanupTemporaryWorkspace when it's needed.

Also changed the places where I added the checkpoints.

I will see how to get the perf difference now.

antoniovs1029: So I ran the ImageClassificationBench.TrainResnetV250 benchmark, with and without the changes of this PR, and they both behaved in pretty much the same way.

Without the changes this was the summary output of the benchmark:
```
          Method |    Mean |   Error |   StdDev | Extra Metric |
---------------- |--------:|--------:|---------:|-------------:|
 TrainResnetV250 | 41.55 s | 5.580 s | 0.3058 s |            - |
```

And with the changes, the summary was:
```
          Method |    Mean |   Error |   StdDev | Extra Metric |
---------------- |--------:|--------:|---------:|-------------:|
 TrainResnetV250 | 40.10 s | 2.723 s | 0.1493 s |            - |
```

So on average the version with the changes was reported to ran faster.

In any case, the CheckAlive() method is simply doing if-statements evaluations, so I don't think it can introduce meaningful performance difference (given that image classification is a task expected to take meaningful amount of time anyway).

antoniovs1029: So, as suggested online by @codemzs I have reran the benchmarks, but using the CIFAR-10 dataset. 

Without the changes introduced in the PR the summary is as follows:
```
          Method |    Mean |   Error |   StdDev | Extra Metric |
---------------- |--------:|--------:|---------:|-------------:|
 TrainResnetV250 | 79.29 m | 4.850 m | 0.2658 m |            - |
```

With the changes:
```
          Method |    Mean |   Error |  StdDev | Extra Metric |
---------------- |--------:|--------:|--------:|-------------:|
 TrainResnetV250 | 78.82 m | 21.71 m | 1.190 m |            - |
```

So, again, my understanding is that there's some variability in the time it takes to train this model (and that's why the benchmark with the changes ran a little bit faster), and the introduction of the CheckAlive() method doesn't really have an impact on the performance of this.

Issue #4646, opened 2020-01-13
Exception on loading ONNX model


dasMulli:
I am unable to use an ONNX model for a transformation that happily works when used directly from M.OnnxRuntime using `InferenceSession`. The model was created using a RandomForestClassifier from SciKitLearn in python.

Repro: [onnx-issue-repro.zip](https://github.com/dotnet/machinelearning/files/4054732/onnx-issue-repro.zip)

```csharp
    class Program
    {
        static void Main(string[] args)
        {
            var mlContext = new MLContext();
            var transform = mlContext.Transforms.ApplyOnnxModel("rf_iris.onnx");
        }
    }
```

```
> cd TestMLNet
> dotnet run
 Unhandled exception. System.ArgumentOutOfRangeException: dim (Parameter 'Dimension { } in ONNX tensor cannot exceed the maximum of 32-bit signed integer.')
Actual value was 0.
   at Microsoft.ML.Transforms.Onnx.OnnxTypeParser.GetDimValue(Dimension dim)
   at Microsoft.ML.Transforms.Onnx.OnnxTypeParser.GetTensorDims(TensorShapeProto tensorShapeProto)
   at Microsoft.ML.Transforms.Onnx.OnnxTypeParser.GetDataViewType(TypeProto typeProto)
   at Microsoft.ML.Transforms.Onnx.OnnxModel..ctor(String modelFile, Nullable`1 gpuDeviceId, Boolean fallbackToCpu, Boolean ownModelFile, IDictionary`2 shapeDictionary)
   at Microsoft.ML.Transforms.Onnx.OnnxTransformer..ctor(IHostEnvironment env, Options options, Byte[] modelBytes)
   at Microsoft.ML.Transforms.Onnx.OnnxTransformer..ctor(IHostEnvironment env, String[] outputColumnNames, String[] inputColumnNames, String modelFile, Nullable`1 gpuDeviceId, Boolean fallbackToCpu, IDictionary`2 shapeDictionary)
   at Microsoft.ML.Transforms.Onnx.OnnxScoringEstimator..ctor(IHostEnvironment env, String modelFile, Nullable`1 gpuDeviceId, Boolean fallbackToCpu, IDictionary`2 shapeDictionary)
   at Microsoft.ML.OnnxCatalog.ApplyOnnxModel(TransformsCatalog catalog, String modelFile, Nullable`1 gpuDeviceId, Boolean fallbackToCpu)
[...]

PR #4652, opened 2020-01-14
Fixes None dimension bug

Lynx1820:
Bug: #4646 

ML.Net considers empty dimensions invalid. However, the "None" dimension is valid in OnnxRuntime.

### DIFF
--- a/src/Microsoft.ML.OnnxTransformer/OnnxTypeParser.cs
+++ b/src/Microsoft.ML.OnnxTransformer/OnnxTypeParser.cs
@@ -178,7 +178,7 @@ namespace Microsoft.ML.Transforms.Onnx
                 case OnnxCSharpToProtoWrapper.TensorShapeProto.Types.Dimension.ValueOneofCase.DimParam:
                     // Variable-length dimension is translated to 0.
                     break;
-                case 0:
+                case OnnxCSharpToProtoWrapper.TensorShapeProto.Types.Dimension.ValueOneofCase.None:
                     // Empty dimension is translated to 0.
                     break;
                 default:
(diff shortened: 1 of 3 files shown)


src/Microsoft.ML.OnnxTransformer/OnnxTypeParser.cs
```diff
@@ -177,10 +177,12 @@ public static int GetDimValue(OnnxCSharpToProtoWrapper.TensorShapeProto.Types.Di
                     break;
                 case OnnxCSharpToProtoWrapper.TensorShapeProto.Types.Dimension.ValueOneofCase.DimParam:
                     // Variable-length dimension is translated to 0.
-                    value = 0;
+                    break;
+                case 0:
```
harishsk: The switch statement is on an enum. Can you please change the case 0: to use ValueOneCase.None?

dasMulli:
Quick question: Would an error message saying that ML.Net doesn't support `None` dimension be more actionable than calling the dimension "invalid" in the exception text? At least this indicates that the onnx model isn't faulty but simply that using it is unsupported..

Lynx1820:
> Quick question: Would an error message saying that ML.Net doesn't support `None` dimension be more actionable than calling the dimension "invalid" in the exception text? At least this indicates that the onnx model isn't faulty but simply that using it is unsupported..

Actually, I added another case for the None value. ML.NET should be able to load the model without an exception.

najeeb-kazmi:
@Lynx1820 please pull master so that this PR can skip the randomly failing tests. Then we can merge this.

Also, does this fix #4646 or is there more work for that bug?

PR #4655, opened 2020-01-15
Added documentation regarding TextLoader's hasHeader field

### LINKED ISSUE #3053: Defining a TextLoader with `hasHeader=true` and no data sample results in no slot names
Related to issue #2663. We need to document the fact that if we create a `TextLoader` specifying `hasHeader=true` but without a data sample, then we are actually just letting the loader know that it needs to skip the first line when it eventually loads from a text file. The output schema of that `TextLoader` will not contain slot names, since the schema is created when the loader is created, and not when `Load` is called.

mstfbl:
Fixes #3053

Added note regarding TextLoader's hasHeader field per @yaeldekel 's suggestion in #3053.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -479,7 +479,7 @@ namespace Microsoft.ML.Data
 
             /// <summary>
             /// Whether the data file has a header with feature names.
-            /// Note: If a TextLoader is created with hasHeader=true but without a dataSample, then the TextLoader will not contain slot names, because the output schema is made when the TextLoader is made, and not when Load is called.
+            /// Note: If a TextLoader is created with HasHeader=true but without a dataSample, then the TextLoader will not contain slot (columns that are chosen for manipulation) names, because the output schema is made when the TextLoader is made, and not when TextLoader.Load(IMultiStreamSource source) is called.
             /// </summary>
             [Argument(ArgumentType.AtMostOnce, ShortName = "header",
                 HelpText = "Data file has header with feature names. Header is read only if options 'hs' and 'hf' are not specified.")]
@@ -679,6 +679,9 @@ namespace Microsoft.ML.Data
                         if (!needInputSize && col.Source.Any(r => r.AutoEnd && r.Max == null))
                             needInputSize = true;
                     }
+
+                    if (parent.HasHeader && dataSample == null)
+                        throw ch.ExceptNotSupp("Data sample cannot be null if the HasHeader flag is set to True.");
 
                     int inputSize = parent._inputSize;
                     ch.Assert(0 <= inputSize & inputSize < SrcLim);
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System;
+using System; 
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System; 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -479,9 +479,9 @@ namespace Microsoft.ML.Data
 
             /// <summary>
             /// Whether the data file has a header with feature names.
-            /// Note: If a TextLoader is created with HasHeader = true but without a dataSample, then vector columns made by TextLoader will not contain slot name
+            /// Note: If a TextLoader is created with hasHeader = true but without a dataSample, then vector columns made by TextLoader will not contain slot name
             /// annotations (slots being the elements of the given vector column), because the output schema is made when the TextLoader is made, and not when
-            /// TextLoader.Load(IMultiStreamSource source) is called. In addition, the case where dataSample = null and HasHeader = true indicates to the
+            /// TextLoader.Load(IMultiStreamSource source) is called. In addition, the case where dataSample = null and hasHeader = true indicates to the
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -479,6 +479,7 @@ public class Options
 
             /// <summary>
             /// Whether the data file has a header with feature names.
+            /// Note: If a TextLoader is created with hasHeader=true but without a dataSample, then the TextLoader will not contain slot names, because the output schema is made when the TextLoader is made, and not when Load is called.
```
harishsk: Nit: The field name is HasHeader (Upper case H)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -479,6 +479,7 @@ public class Options
...
             /// </summary>
```
harishsk: Can you please explain the reference to slots? Does it mean column names?
Is it possible to throw an exception instead under these conditions?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
yaeldekel: >Note [](start = 16, length = 4)

Please document the public APIs as well (in TextLoaderSaverCatalog.cs).

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
mstfbl: Of course. Yes, the term "slots" here refers to columns that are chosen for data manipulation. In my latest commit I have added the case where a situation where hasHeader = true and null dataSample results in a Not Supported exception.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -21,7 +21,9 @@ public static class TextLoaderSaverCatalog
         /// <param name="catalog">The <see cref="DataOperationsCatalog"/> catalog.</param>
         /// <param name="columns">Array of columns <see cref="TextLoader.Column"/> defining the schema.</param>
         /// <param name="separatorChar">The character used as separator between data points in a row. By default the tab character is used as separator.</param>
-        /// <param name="hasHeader">Whether the file has a header.</param>
+        /// <param name="hasHeader">Whether the file has a header with feature names. Note: If a TextLoader is created with HasHeader=true but without a dataSample,
+        /// then the TextLoader will not contain slot (columns that are chosen for manipulation) names, because the output schema is made when the TextLoader is made,
```
yaeldekel: >columns that are chosen for manipulation [](start = 55, length = 40)

the term "slots" is how we refer to the elements of a vector column. In `TextLoader`, these correspond to actual columns (in the "Excel" sense) in the text file. For example, if a user defines `TextLoader.Column("Features", 3, 14)`, that would create a vector column called `Features`, with 12 slots, and values taken from columns 3 thru 14 in the text file. If the text loader is given a `dataSample` and `hasHeader` is true, then the column `Features` will have slot names annotations, with values that will be taken from the first line, columns 3 thru 14 in the `dataSample` file.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -679,6 +680,9 @@ public Bindings(TextLoader parent, Column[] cols, IMultiStreamSource headerFile,
                             needInputSize = true;
                     }
 
+                    if (parent.HasHeader && dataSample == null)
+                        throw ch.ExceptNotSupp("Data sample cannot be null if the HasHeader flag is set to True.");
```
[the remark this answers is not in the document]
mstfbl: Got it, I have added this case as a comment to the HasHeader field summary.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
mstfbl: Got it, I've updated the summary here.

yaeldekel: The comment in the parentheses here (and in the other public APIs) also needs to be updated.

---
In reply to: [370074523](https://github.com/dotnet/machinelearning/pull/4655#discussion_r370074523) [](ancestors = 370074523)

yaeldekel: Can you add the part about skipping the first line to the documentation in this file as well? Thanks.

---
In reply to: [370987713](https://github.com/dotnet/machinelearning/pull/4655#discussion_r370987713) [](ancestors = 370987713,370074523)

mstfbl: Hi Yael, done and done!

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -21,7 +21,11 @@ public static class TextLoaderSaverCatalog
         /// <param name="catalog">The <see cref="DataOperationsCatalog"/> catalog.</param>
         /// <param name="columns">Array of columns <see cref="TextLoader.Column"/> defining the schema.</param>
         /// <param name="separatorChar">The character used as separator between data points in a row. By default the tab character is used as separator.</param>
-        /// <param name="hasHeader">Whether the file has a header.</param>
+        /// <param name="hasHeader">Whether the file has a header with feature names. Note: If a TextLoader is created with HasHeader = true but without a
```
harishsk: >HasHeader [](start = 124, length = 9)

Nit: Please match the case of the parameter 'hasHeader' in all instances in the updated docs.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -21,7 +21,11 @@ public static class TextLoaderSaverCatalog
...
+        /// dataSample, then vector columns made by TextLoader will not contain slot name annotations (slots being the elements of the given vector column),
+        /// because the output schema is made when the TextLoader is made, and not when TextLoader.Load(IMultiStreamSource source) is called.
```
yaeldekel: >Load [](start = 99, length = 4)

This can use <see ...>.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
yaeldekel: Or use <paramref ...> both for `hasHeader` and for `dataSample`.

---
In reply to: [373803799](https://github.com/dotnet/machinelearning/pull/4655#discussion_r373803799) [](ancestors = 373803799)

PR #4659, opened 2020-01-15
Testing changing the behavior of baseline Fail function

harishsk:
Draft PR to try to get callstacks at the point of failure.

### DIFF
--- a/test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
+++ b/test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
@@ -168,6 +168,8 @@ namespace Microsoft.ML.RunTests
                 _passed = false;
 
             Log("*** Failure: " + fmt, args);
+
+            Assert.True(_passed);
         }
 
         protected void Log(string msg)


sharwell:
I was going to propose something like 78792a0366b90704914c29c8a48c2b7815f92af2

harishsk:
I was trying to implement the idea you suggested in mail. The idea I was going with was that we should catch all failures early and stop right away. I think we should be treating Fail the same as failed test assertion.  
Would your model give us anything different or more?
(I agree with deleting DoNotEverUseInvertPass)

With the PR I have submitted, I am seeing a few incorrectly written tests. I am in the process of fixing those. Once I have all the tests passing locally, I was planning to update the PR and see what sort of traces we get from the build servers.

sharwell:
> With the PR I have submitted, I am seeing a few incorrectly written tests. I am in the process of fixing those. 

Yes, there are many tests which do not make any assertions about whether the actual results match the expected results.

> Would your model give us anything different or more?

There are two cases where failing on first exception via throwing can be problematic:

1. If the exception is caught, it could result in a test passing that should have failed
2. You can only see one failure each time a test runs, so if a test has multiple failures in sequence it can take a large number of iterations to fix all the issues with a test

I can't make a determination on how much these matter for your case.

eerhardt:
> You can only see one failure each time a test runs, so if a test has multiple failures in sequence it can take a large number of iterations to fix all the issues with a test

My understanding was this was the reason why the original test code went with the current approach. A single test will output/compare against multiple baseline files, so if a developer needs to update one of those tests, they would need to run the test multiple times to get all the updated outputs.

harishsk:
That makes sense. @sharwell, We can adopt your method. Do you have a PR open already for it?

PR #4661, opened 2020-01-15
added in support for System.DateTime type for the DateTimeTransformer

michaelgsharp:
Added in support to the `DateTimeTransformer` to use columns of `DateTime` type.

Removed the thread safe caching as its faster to just query the native code multiple times then to deal with the thread safe cache.

### DIFF
--- a/src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
+++ b/src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
@@ -148,7 +148,7 @@ namespace Microsoft.ML.Featurizers
 
         public DateTimeTransformer Fit(IDataView input)
         {
-            return new DateTimeTransformer(_host, _options.Source, _options.Prefix, _options.Country, input);
+            return new DateTimeTransformer(_host, _options.Source, _options.Prefix, _options.Country, input.Schema);
         }
 
         public SchemaShape GetOutputSchema(SchemaShape inputSchema)
@@ -250,12 +250,12 @@ namespace Microsoft.ML.Featurizers
 
         #endregion
 
-        internal DateTimeTransformer(IHostEnvironment host, string inputColumnName, string columnPrefix, DateTimeEstimator.HolidayList country, IDataView input) :
+        internal DateTimeTransformer(IHostEnvironment host, string inputColumnName, string columnPrefix, DateTimeEstimator.HolidayList country, DataViewSchema schema) :
             base(host.Register(nameof(DateTimeTransformer)))
         {
             host.Check(!CommonExtensions.OsIsCentOS7(), "CentOS7 is not supported");
 
-            _schema = input.Schema;
+            _schema = schema;
             if (_schema[inputColumnName].Type.RawType != typeof(long) &&
                 _schema[inputColumnName].Type.RawType != typeof(DateTime))
             {
@@ -650,20 +650,19 @@ namespace Microsoft.ML.Featurizers
         private sealed class Mapper : MapperBase
         {
 
-#region Class data members
+            #region Class data members
+            private static readonly DateTime _unixEpoch = new DateTime(1970, 1, 1);
 
             private readonly DateTimeTransformer _parent;
             private ConcurrentDictionary<long, TimePoint> _cache;
             private ConcurrentQueue<long> _oldestKeys;
-            private readonly DateTime _unixEpoc;
 
-#endregion
+            #endregion
 
             public Mapper(DateTimeTransformer parent, DataViewSchema inputSchema) :
                 base(parent.Host.Register(nameof(Mapper)), inputSchema, parent)
             {
                 _parent = parent;
-                _unixEpoc = new DateTime(1970, 1, 1);
                 _cache = new ConcurrentDictionary<long, TimePoint>();
                 _oldestKeys = new ConcurrentQueue<long>();
             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -455,11 +465,17 @@ internal abstract class TypedColumn : IDisposable
...
+            internal readonly int StructSize;
 
             internal TypedColumn(string source, string prefix)
             {
                 Source = source;
                 Prefix = prefix;
+                IntPtrSize = IntPtr.Size;
+
+                // The native struct is 25 bytes + 8 size_t.
+                StructSize = 25 + (IntPtrSize * 8);
```
harishsk: Is it possible to avoid hardcoded numbers? Would Marshal.Sizeof on the native struct work?

test/Microsoft.ML.Tests/Transformers/DateTimeTransformerTests.cs
```diff
@@ -133,7 +133,50 @@ public void CanUseDateFromColumn()
             Done();
         }
 
-        [NotCentOS7FactAttribute]
+        [NotCentOS7Fact]
+        public void CanUseDateFromColumnDateTimeType()
+        {
+            // Future Date - 2025 June 30
```
harishsk: If this code survives this date...will it still be the future? ;-)

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -643,13 +655,15 @@ private sealed class Mapper : MapperBase
...
             private ConcurrentQueue<long> _oldestKeys;
+            private readonly DateTime _unixEpoc;
 
 #endregion
 
             public Mapper(DateTimeTransformer parent, DataViewSchema inputSchema) :
                 base(parent.Host.Register(nameof(Mapper)), inputSchema, parent)
             {
                 _parent = parent;
+                _unixEpoc = new DateTime(1970, 1, 1);
```
harishsk: Nit: Typo for Epoch

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -667,74 +681,68 @@ protected override DataViewSchema.DetachedColumn[] GetOutputColumnsCore()
...
+                var getter = input.GetGetter<TInput>(input.Schema[_parent._column.Source]);
+
+                ValueGetter<TTransformed> result = (ref TTransformed dst) =>
                 {
-                    long dateTime = default;
-                    getter(ref dateTime);
+                    TInput dateTimeColumn = default;
+                    getter(ref dateTimeColumn);
 
-                    if (!_cache.TryGetValue(dateTime, out TimePoint timePoint))
```
harishsk: I am curious about this deleted code. What was the cache for? Why is it no longer needed?

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -148,7 +148,7 @@ internal DateTimeEstimator(IHostEnvironment env, Options options)
 
         public DateTimeTransformer Fit(IDataView input)
         {
-            return new DateTimeTransformer(_host, _options.Source, _options.Prefix, _options.Country);
+            return new DateTimeTransformer(_host, _options.Source, _options.Prefix, _options.Country, input);
         }
```
harishsk: Nit: Since you are using only Schema in your class, would it be better to pass in the schema instead of input?

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -745,7 +753,13 @@ protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, b
                 disposer = null;
 
                 // Have to add 1 to iinfo since the enum starts at 1
-                return Utils.MarshalInvoke(MakeGetter<int>, ((DateTimeEstimator.ColumnsProduced)iinfo + 1).GetRawColumnType(), input, iinfo);
+                // If input type is long, Posix time
+                if (input.Schema[_parent._column.Source].Type.RawType == typeof(long))
+                    return Utils.MarshalInvoke(MakeGetter<int, int>, new Type[] { typeof(long), ((DateTimeEstimator.ColumnsProduced)iinfo + 1).GetRawColumnType() }, input, iinfo);
+                // If input types is System.DateTime
+                else
```
harishsk: Would an assert be better than a comment?

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -643,13 +655,15 @@ private sealed class Mapper : MapperBase
             private readonly DateTimeTransformer _parent;
             private ConcurrentDictionary<long, TimePoint> _cache;
             private ConcurrentQueue<long> _oldestKeys;
+            private readonly DateTime _unixEpoc;
```
eerhardt: This should be a `static readonly` field. No need to new one up every time.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -667,74 +681,68 @@ protected override DataViewSchema.DetachedColumn[] GetOutputColumnsCore()
...
-                                long key;
-                                while (!_oldestKeys.TryDequeue(out key)) { }
-                                while (!_cache.TryRemove(key, out TimePoint removedValue)) { }
-                            }
-                        }
-                    }
+                    long dateTime;
+
+                    // Type already in Posix format
+                    if(typeof(TInput) == typeof(long))
```
eerhardt: This type checking should happen outside of the getter delegate. There is no need to pay this price every time this method is called. See other transformer implementations for examples.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: If I remember correctly I tried that and had issues with it, but let me try again.

michaelgsharp: Oh, the issue was that the native struct returns pointers to a string buffer and my struct uses strings directly. I can use Marshal.SizeOf, but then I need to define a second struct that exactly matches the native code. Would you prefer that approach?

test/Microsoft.ML.Tests/Transformers/DateTimeTransformerTests.cs
michaelgsharp: Good point. I'll remove the "Future" comment.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: I was trying to optimize by caching the retrieved value. Since this code can run in parallel though, the cache was thread safe. After doing some testing during this change it turned out in this case the thread safe cache was slower then just calling the native code extra times.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Good idea. Changed it to take just the schema.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: I was just trying to document the code execution path. If the type is not a supported type then an error will be thrown before it gets to this point.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Changed it as suggested.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Good catch. Fixed.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Refactored my MakeGetter and moved the type checking out of the delegate.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
harishsk: Yes, I would prefer that approach. The presence of a struct that matches the native code makes the contract explicit and if something were to change on the native side, the corresponding change would be easier to identify.

---
In reply to: [368129495](https://github.com/dotnet/machinelearning/pull/4661#discussion_r368129495) [](ancestors = 368129495)

michaelgsharp: Sounds good. I have added in another struct that exactly matches the native code and am using that to get the size of the struct.

PR #4662, opened 2020-01-15
Throw exception in ImageClassificationTrainer when dataset contains only 1 class

### LINKED ISSUE #4660: Using an ImageClassificationTrainer with a dataset that only contains one class of images
### Issue

- **What did you do?**
I tried to train a model that uses the `ImageClassificationTrainer`, using a dataset that only contains images labeled as 'dog'.

- **What happened?**
I got a `System.InvalidOperationException: 'Metadata KeyValues does not exist'` while fitting the model, which isn't very informative about the problem. If I didn't know my dataset only contained one label, or if I were to split the dataset in a way that the training set had only one label, then getting that exception wouldn't help to fix the problem.

Also notice that the exception is thrown _after_ training is done, so an user would have already invested time on it before noticing that something is wrong. 

- **What did you expect?**
Probably an exception which tells me that there is only one class in my dataset, as done by other multiclass trainers such as the Multiclass LightGBM which throws a `"System.InvalidOperationException: 'LightGBM Error, code is -1, error message is 'Number of classes should be specified and greater than 1 for multiclass training'.'"`, or the `LinearMulticlassModelParametersBase` which throws a `System.ArgumentOutOfRangeException: 'Must be at least 2. Parameter name: numClasses'` in [here](https://github.com/dotnet/machinelearning/blob/c4e4263188dccf16903b8f3fea7e65213a69c6e3/src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/MulticlassLogisticRegression.cs#L505).

If I changed my dataset to include other labels, then the exception is gone and it works as expected.

### Source code / logs
[dataset.zip](https://github.com/dotnet/machinelearning/files/4067127/dataset.zip)

```C#
using Microsoft.ML.Data;
using Microsoft.ML.Vision;

{
    public class ModelInput
    {
        [ColumnName("Label"), LoadColumn(0)]


        [ColumnName("ImageSource"), LoadColumn(1)]
        public string ImageSource { get; set; }
    }

    public static class Program
    {
        // private static string TRAIN_DATA_FILEPATH = @"C:\Users\anvelazq\Desktop\issue19\dogs-cats-horses.tsv"; // This one works because it has multiple classes
        private static string TRAIN_DATA_FILEPATH = @"C:\Users\anvelazq\Desktop\issue19\only-dogs.tsv"; // This one doesn't work because it has only one class

        private static MLContext mlContext = new MLContext(seed: 1);

        public static void Main()
        {
            IDataView trainingDataView = mlContext.Data.LoadFromTextFile<ModelInput>(
                                            path: TRAIN_DATA_FILEPATH,
                                            hasHeader: true,
                                            separatorChar: '\t',
                                            allowQuoting: true,
                                            allowSparse: false);

            var dataProcessPipeline = mlContext.Transforms.Conversion.MapValueToKey("Label", "Label")
                                      .Append(mlContext.Transforms.LoadRawImageBytes("ImageSource_featurized", null, "ImageSource"))
                                      .Append(mlContext.Transforms.CopyColumns("Features", "ImageSource_featurized"));

            var trainer = mlContext.MulticlassClassification.Trainers.ImageClassification(new ImageClassificationTrainer.Options() { LabelColumnName = "Label", FeatureColumnName = "Features" })
                                      .Append(mlContext.Transforms.Conversion.MapKeyToValue("PredictedLabel", "PredictedLabel"));
            var trainingPipeline = dataProcessPipeline.Append(trainer);

            var model = trainingPipeline.Fit(trainingDataView); // System.InvalidOperationException: 'Metadata KeyValues does not exist'

            var transformed = model.Transform(trainingDataView);
            var transformedPreview = transformed.Preview();
        }
    }
}
```

### Nugets used:
image

antoniovs1029:
Throw exception in ImageClassificationTrainer when dataset contains only 1 class

Fixes #4660

### DIFF
--- a/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
+++ b/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
@@ -603,7 +603,8 @@ namespace Microsoft.ML.Vision
                     (string)labelType.ToString());
             }
 
-            Contracts.CheckParam(labelCount > 1, nameof(labelCount), "ImageClassificationTrainer requires more than 1 class in the training dataset");
+            var msg = $"Only one class found in {_options.LabelColumnName} column. To build a multiclass classification model, the number of classes needs to be 2 or greater";
+            Contracts.CheckParam(labelCount > 1, nameof(labelCount), msg);
 
             _classCount = (int)labelCount;
             var imageSize = ImagePreprocessingSize[_options.Arch];
--- a/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
+++ b/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
@@ -603,7 +603,10 @@ namespace Microsoft.ML.Vision
                     (string)labelType.ToString());
             }
 
-            _classCount = labelCount == 1 ? 2 : (int)labelCount;
+            if (labelCount == 1)
+                throw new InvalidOperationException("Dataset contains only 1 class. ImageClassificationTrainer requires more than 1");
+
+            _classCount = (int)labelCount;
             var imageSize = ImagePreprocessingSize[_options.Arch];
             _session = LoadTensorFlowSessionFromMetaGraph(Host, _options.Arch).Session;
             _session.graph.as_default();
--- a/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
+++ b/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
@@ -603,8 +603,7 @@ namespace Microsoft.ML.Vision
                     (string)labelType.ToString());
             }
 
-            if (labelCount == 1)
-                throw new InvalidOperationException("Dataset contains only 1 class. ImageClassificationTrainer requires more than 1");
+            Contracts.CheckParam(labelCount > 1, nameof(labelCount), "ImageClassificationTrainer requires more than 1 class in the training dataset");
 
             _classCount = (int)labelCount;
             var imageSize = ImagePreprocessingSize[_options.Arch];
--- a/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
+++ b/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
@@ -603,7 +603,7 @@ namespace Microsoft.ML.Vision
                     (string)labelType.ToString());
             }
 
-            var msg = $"Only one class found in {_options.LabelColumnName} column. To build a multiclass classification model, the number of classes needs to be 2 or greater";
+            var msg = $"Only one class found in the {_options.LabelColumnName} column. To build a multiclass classification model, the number of classes needs to be 2 or greater";
             Contracts.CheckParam(labelCount > 1, nameof(labelCount), msg);
 
             _classCount = (int)labelCount;


src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -603,7 +603,10 @@ private void InitializeTrainingGraph(IDataView input)
                     (string)labelType.ToString());
             }
 
-            _classCount = labelCount == 1 ? 2 : (int)labelCount;
+            if (labelCount == 1)
+                throw new InvalidOperationException("Dataset contains only 1 class. ImageClassificationTrainer requires more than 1");
+
```
harishsk: If I understand it correctly, this happens when the supplied data is bad (i.e. it has only one class of labels)
Would it be better to throw an ArgumentException?

justinormont: We throw a similar error in other places: 
https://github.com/dotnet/machinelearning/blob/1a1a62ad3d582e1ab99fe900c73b908da94af647/src/Microsoft.ML.Data/Evaluators/AucAggregator.cs#L115-L116

Producing:
`System.ArgumentOutOfRangeException: AUC is not definied 
when there is no positive class in the data`

antoniovs1029: Yes, @harishsk , your understanding is correct about this case. I will change this asap.

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -603,7 +603,10 @@ private void InitializeTrainingGraph(IDataView input)
                     (string)labelType.ToString());
             }
 
-            _classCount = labelCount == 1 ? 2 : (int)labelCount;
+            if (labelCount == 1)
+                throw new InvalidOperationException("Dataset contains only 1 class. ImageClassificationTrainer requires more than 1");
```
ganik: >ImageClassificationTrainer [](start = 84, length = 26)

Could you make the message more generic. Its not only ImageClassification that requires more than 1 class. Pls see what other trainers throw and align the message with those

antoniovs1029: So, for example, when fitting a `SdcaMaximumEntropy` trainer with a dataset with 1 label then `LinearMulticlassModelParametersBase` will throw this exception:
`System.ArgumentOutOfRangeException: 'Must be at least 2. Parameter name: numClasses'`

Because of this line of code:
https://github.com/dotnet/machinelearning/blob/c4e4263188dccf16903b8f3fea7e65213a69c6e3/src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/MulticlassLogisticRegression.cs#L505

So I could change the exception in here to have the same output. But I didn't do that because, in general, I've seen users mentioning that our Exception messages aren't very readable. If I were to use the same exception message in here, it would be something like this:
`System.ArgumentOutOfRangeException: 'Must be at least 2. Parameter name: labelCount`

Which, to be honest, I think would be very unreadable for a user (specially one who is simply using the nugets, and wouldn't want to delve into the code to figure out what "labelCount" is and where is it used).

That's why I preferred to use the specific exception message I wrote, and also because it's an exception thrown in the ImageClassificationTrainer because of a requirement of the trainer itself, which throws before training.

Please let me know if you still think the exception message should be changed. If anything I guess something like
"Dataset contains only 1 class. More than 1 class is required. Parameter name: labelCount"

harishsk: @antoniovs1029 Good idea to make the exception message more readable. I think @ganik's point was that all classifiers need more than one class. But in this particular case, labelCount as a parameter would not mean anything to the user, since it is not a parameter that they supplied.

A good error message would also tell me what needs to be done to fix the problem. (Perhaps @natke can weigh in here about best practices) How about something like:

"Dataset contains only one class. Please use a dataset with more than one labeled class"

natke: I agree that consistency across all of the multiclass trainers would be good. But not at the expense of clarity if the existing messages are poorly written.

I had a look through the other exceptions and this is what I found:
- Linear trainers such as  SdcaMaximumEntropy: Unhandled Exception: System.ArgumentOutOfRangeException: Must be at least 2.
- LightGbm: "Number of classes should be specified and greater than 1 for multiclass training"
- Meta trainers such as OneVersusAll and Pairwise coupling don't even check for the condition, and let you build a model that always predicts the same class

To @harishsk's point about `labelCount` not being a parameter that the user supplied, can we reference the label column name from the schema, as that is what the user has supplied?

Something like: "Only one class found in XXX column. To build a multiclass classification model, the number of classes needs to be 2 or greater."

antoniovs1029: Yes, we can reference the label column name to output that message. I will update this PR asap with those changes.

antoniovs1029: So I've updated the PR with the exception message that @natke suggested.

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -603,7 +603,8 @@ private void InitializeTrainingGraph(IDataView input)
                     (string)labelType.ToString());
             }
 
-            Contracts.CheckParam(labelCount > 1, nameof(labelCount), "ImageClassificationTrainer requires more than 1 class in the training dataset");
+            var msg = $"Only one class found in {_options.LabelColumnName} column. To build a multiclass classification model, the number of classes needs to be 2 or greater";
```
natke: Slight reword to my original: var msg = $"Only one class found in *the* {_options.LabelColumnName} column.

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -603,7 +603,10 @@ private void InitializeTrainingGraph(IDataView input)
                     (string)labelType.ToString());
             }
 
-            _classCount = labelCount == 1 ? 2 : (int)labelCount;
+            var msg = $"Only one class found in {_options.LabelColumnName} column. To build a multiclass classification model, the number of classes needs to be 2 or greater";
+            Contracts.CheckParam(labelCount > 1, nameof(labelCount), msg);
```
antoniovs1029: I've implemented the message that @natke suggested.

I had to use the "var msg" to store the string, because I couldn't use string interpolation inside the CheckParam() argument, because of this:

https://github.com/dotnet/machinelearning/blob/fa75f4fa93a252d15f05893ddbd1a22b9983cbf4/tools-local/Microsoft.ML.InternalCodeAnalyzer/Descriptions.Designer.cs#L64-L71

So the suggestion there is to use Contracts.Except() instead of Contracts.Param() if I want to use string interpolation. Problem is Contracts.Except() throws an `InvalidOperationException` (which @justinormont and @harishsk [had suggested](https://github.com/dotnet/machinelearning/pull/4662#discussion_r367778352) I shouldn't use). So I guess it's best to simply create the "msg" string before passing it to the Contracts.CheckParam()...

PR #4665, opened 2020-01-16
TextLoader now checks for presence of source file with clear error message

### LINKED ISSUE #2461: The method Read of TextLoader does not check for the presence of the file.
### System information

- **OS version/distro**:Windows 10
- **.NET Version (eg., dotnet --info)**: .NET 4.6

### Issue

- **What did you do?**
- **What happened?**
- **What did you expect?**

### Source code / logs
The method Read of the class TextLoader does not validate the presence of the input file. I understand that there is an element of lazy loading here. But, it would help to get feedback earlier in the form of System.IO.FileNotFoundException

mstfbl:
Fixed #2461 

TextLoader now checks whether the given file in path exists. If the file is missing, a clear exception stating that the file is missing is given, instead of stating that some field is missing some unknown attribute.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
@@ -114,7 +114,7 @@ namespace Microsoft.ML
             Contracts.CheckNonEmpty(path, nameof(path));
             if (!File.Exists(path))
             {
-                throw Contracts.ExceptDecode("File does not exist at path: {0}", path);
+                throw Contracts.ExceptParam("File does not exist at path: {0}", path);
             }
 
             var options = new TextLoader.Options
@@ -160,7 +160,7 @@ namespace Microsoft.ML
             Contracts.CheckNonEmpty(path, nameof(path));
             if (!File.Exists(path))
             {
-                throw Contracts.ExceptDecode("File does not exist at path: {0}", path);
+                throw Contracts.ExceptParam("File does not exist at path: {0}", path);
             }
 
             // REVIEW: it is almost always a mistake to have a 'trainable' text loader here.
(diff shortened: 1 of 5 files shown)


src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -154,6 +158,10 @@ public static class TextLoaderSaverCatalog
             bool allowSparse = TextLoader.Defaults.AllowSparse)
         {
             Contracts.CheckNonEmpty(path, nameof(path));
+            if (!File.Exists(path))
+            {
+                throw Contracts.ExceptDecode("File does not exist at path: {0}", path);
```
yaeldekel: >ExceptDecode [](start = 32, length = 12)

`ExceptDecode` is used for errors in deserialization of models. I think `ExceptParam` would be more natural in this case.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -112,6 +112,10 @@ public static class TextLoaderSaverCatalog
             bool allowSparse = TextLoader.Defaults.AllowSparse)
         {
             Contracts.CheckNonEmpty(path, nameof(path));
+            if (!File.Exists(path))
```
yaeldekel: >!File.Exists(path) [](start = 16, length = 18)

Consider adding this check to the `BinaryLoader` and `SvmLightLoader` APIs.

mstfbl: Good point, I just did!

mstfbl: Good point, I just did!

PR #4666, opened 2020-01-16
collect crash dump, upload dump and pdb to artifact

frank-dong-ms:
1. collect crash dump when test host process crashes
2. upload dump and pdb to artifact for further investigate

### DIFF
--- a/init-tools.cmd
+++ b/init-tools.cmd
@@ -60,10 +60,12 @@ if exist "%DOTNET_CMD%" goto :afterdotnetrestore
 if NOT exist "%DOTNET_PATH%" mkdir "%DOTNET_PATH%"
 
 :: set registry to take dump automatically when test process crashes
-powershell -Command "New-Item -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting' -Name 'LocalDumps' -Force"
-powershell -Command "New-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps' -Name 'DumpFolder' -Value '%~dp0CrashDumps' -PropertyType 'ExpandString' -Force"
-:: take full dump
-powershell -Command "New-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps' -Name 'DumpType' -Value 2 -PropertyType DWord -Force"
+if NOT [%AGENT_ID%] == [] (
+  reg add "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps"
+  reg add "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps" /v DumpType /t REG_DWORD /d 2
+  reg add "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps" /v DumpCount /t REG_DWORD /d 2
+  reg add "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps" /v DumpFolder /t REG_SZ /d "%~dp0CrashDumps"
+)
 
 :: install the extra runtime first, so the SDK install will overwrite the root dotnet executable
 echo Installing dotnet runtime %DOTNET_EXTRA_RUNTIME_VERSION%...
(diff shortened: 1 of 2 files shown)


init-tools.cmd
```diff
@@ -59,6 +59,12 @@ if exist "%DOTNET_CMD%" goto :afterdotnetrestore
 
 if NOT exist "%DOTNET_PATH%" mkdir "%DOTNET_PATH%"
 
+:: set registry to take dump automatically when test process crashes
```
safern: This is going to change people environment. Should we change it to just be done when in CI?

We can detect that via env vars that azure devops sets.

sharwell: Agreed this should only run on CI.

init-tools.cmd
```diff
@@ -59,6 +59,12 @@ if exist "%DOTNET_CMD%" goto :afterdotnetrestore
...
+powershell -Command "New-Item -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting' -Name 'LocalDumps' -Force"
+powershell -Command "New-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps' -Name 'DumpFolder' -Value '%~dp0CrashDumps' -PropertyType 'ExpandString' -Force"
+:: take full dump
+powershell -Command "New-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps' -Name 'DumpType' -Value 2 -PropertyType DWord -Force"
```
sharwell:
```
reg add "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps"
reg add "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps" /v DumpType /t REG_DWORD /d 2
reg add "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps" /v DumpCount /t REG_DWORD /d 2
reg add "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps" /v DumpFolder /t REG_SZ /d "%~dp0CrashDumps"
```

sharwell: ❔ Normally these registry entries are added for one or more specific process names. Since it's on CI, it seems reasonable to not restrict it in this manner. However, are we sure that the entries will work when used without specified a process name?

frank-dong-ms: Yes, this works on my test in both local and CI

---
In reply to: [367624922](https://github.com/dotnet/machinelearning/pull/4666#discussion_r367624922) [](ancestors = 367624922)

init-tools.cmd
frank-dong-ms: good point, will fix this

---
In reply to: [367621559](https://github.com/dotnet/machinelearning/pull/4666#discussion_r367621559) [](ancestors = 367621559)

init-tools.cmd
```diff
@@ -59,6 +59,14 @@ if exist "%DOTNET_CMD%" goto :afterdotnetrestore
 
 if NOT exist "%DOTNET_PATH%" mkdir "%DOTNET_PATH%"
 
+:: set registry to take dump automatically when test process crashes
+if NOT [%AGENT_ID%] == [] (
+  reg add "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps"
+  reg add "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps" /v DumpType /t REG_DWORD /d 2
+  reg add "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps" /v DumpCount /t REG_DWORD /d 2
+  reg add "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps" /v DumpFolder /t REG_SZ /d "%~dp0CrashDumps"
+)
```
sharwell: I've never seen this syntax before, but assuming it works then 👍

frank-dong-ms: Below is predefined variables in azure devops:
[https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#agent-variables](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#agent-variables)

The one variable I'm using is: Agent.Id

These variables can also be accessed through environment with name conversion:
[https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch)

---
In reply to: [367687379](https://github.com/dotnet/machinelearning/pull/4666#discussion_r367687379) [](ancestors = 367687379)

PR #4669, opened 2020-01-17
retry and disable tests to get stable CI

frank-dong-ms:
Due to test in private agent pool, these 2 tests seems crashing test host process, try to disable them.

https://dev.azure.com/dnceng/public/_build?definitionId=712&_a=summary


sharwell:
Can we use https://github.com/dotnet/machinelearning/pull/4666 instead of this?

frank-dong-ms:
I am not intend to check this in, disable test is bad thing, this is only for test to see whether we see less crash or not

---
In reply to: [575671268](https://github.com/dotnet/machinelearning/pull/4669#issuecomment-575671268) [](ancestors = 575671268)

PR #4674, opened 2020-01-18
F1-score to return 0.0 instead of NaN

### LINKED ISSUE #4664: F1-score to return 0.0 instead of NaN
Our `F1` metric can currently return a `NaN` value when both the precision and recall are zero, as `F1 = 2 * precision * recall / (precision + recall)`, which turns into `0 / 0`. Precision/recall being zero occurs for a trivial model which always guesses the majority class zero.

We should return `F1=0.0` instead of `NaN`  -- this is done in sklearn ([code](https://github.com/scikit-learn/scikit-learn/blob/4b82c4450b759f4efd0dbad2840739ffc416e533/sklearn/metrics/classification.py#L927-L929)), and others ([note](https://github.com/dice-group/gerbil/wiki/Precision,-Recall-and-F1-measure#dividing-by-0)) 
<br/>Our `F1` calculation code:
https://github.com/dotnet/machinelearning/blob/610ffcb67083c2e5e6e1a14884ba24b1da0384c7/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs#L483

See background: https://github.com/dotnet/machinelearning/issues/4648#issuecomment-574886743

Lynx1820:
Fixes #4664

### DIFF
--- a/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
+++ b/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
@@ -480,7 +480,15 @@ namespace Microsoft.ML.Data
                     }
                 }
 
-                public Double F1 { get { return 2 * PrecisionPos * RecallPos / (PrecisionPos + RecallPos); } }
+                public Double F1
+                {
+                    get
+                    {
+                        if (PrecisionPos + RecallPos == 0)
+                            return 0;
+                        return 2 * PrecisionPos * RecallPos / (PrecisionPos + RecallPos);
+                    }
+                }
 
                 public Counters(bool useRaw, Single threshold)
                 {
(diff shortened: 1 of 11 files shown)


src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs
```diff
@@ -480,7 +480,15 @@ public Double LogLossReduction
                     }
                 }
 
-                public Double F1 { get { return 2 * PrecisionPos * RecallPos / (PrecisionPos + RecallPos); } }
+                public Double F1
+                {
+                    get
+                    {
+                        if (PrecisionPos + RecallPos == 0)
```
harishsk: Nit: Maybe we can avoid computing the value of (PrecisionPos + RecalPos) twice?

Issue #3988, opened 2019-07-11
Troubles with CustomMappingEstimator after save/load because InputSchemaDefinition is not being saved.


SkinnyMan32:
### System information

- **ML.Net v1.2.0**

### Issue
I try to use CustomMapping with specified column names. It works fine, but after save/load model I get exception:
**System.ArgumentOutOfRangeException: "Could not find  column 'Words'".** 
'Words' - the original name of property, not specified by me.

What am i doing wrong?

I can fix it by using only original names, it is not comfortable in some cases.

### Source code
```C#
// --- Test method ---
var ml = new MLContext();
var descriptions = new[]
{
	new { Description = "Painted, Painting, Painter" }
};
var dataView = ml.Data.LoadFromEnumerable(descriptions);

var pipeline = ml.Transforms.Text.NormalizeText("Normalized", "Description")
	.Append(ml.Transforms.Text.TokenizeIntoWords("Tokens", "Normalized"))
     // (Extension method) CustomMapping with specified column names
	.Append(ml.Transforms.StemText("Stemmed", "Tokens"));

var model = pipeline.Fit(dataView);
var preview = model.Transform(dataView).Preview();  // everything is ok

// Save model
MemoryStream stream = new MemoryStream();
ml.Model.Save(model, dataView.Schema, stream);
stream.Position = 0;

// Load model in the new context
var ml2 = new MLContext();
// Register custom action
ml2.ComponentCatalog.RegisterAssembly(typeof(StemmerCustomAction).Assembly);
var loadedModel = ml2.Model.Load(stream, out var schema);

// Exception:
// System.ArgumentOutOfRangeException: "Could not find  column 'Words'
var preview2 = loadedModel.Transform(dataView).Preview();


//--- Classes ---

public class StemmerInput
{
	public string[] Words { get; set; }
}

public class StemmerOutput
{
	public string[] Stemmed { get; set; }
}

[CustomMappingFactoryAttribute("StemText")]
public class StemmerCustomAction : CustomMappingFactory<StemmerInput, StemmerOutput>
{
	public static void StemAction(StemmerInput input, StemmerOutput output)
	{
		var stemmer = new EnglishStemmer();
		output.Stemmed = new string[input.Words.Length];
		for (int i = 0; i < input.Words.Length; i++)
		{
			output.Stemmed[i] = stemmer.Stem(input.Words[i]);
		}
	}

	public override Action<StemmerInput, StemmerOutput> GetMapping() => StemAction;
}

[...]

PR #4676, opened 2020-01-19
Add support for stateful custom mappers, and for custom filters.

### LINKED ISSUE #4675: Add support for stateful custom mappers, and for custom filters

yaeldekel:
Fixes #4675 .


test/Microsoft.ML.Tests/Transformers/CustomMappingTests.cs
```diff
@@ -121,5 +122,128 @@ public void TestSchemaPropagation()
...
+                new MyFilterInput() { Counter = 5, Value = 4 },
+                new MyFilterInput() { Counter = 6, Value = 1 },
+                new MyFilterInput() { Counter = 7, Value = 1 },
+                new MyFilterInput() { Counter = 8, Value = 2 },
+            });
+
+            var filteredData = ML.Data.StatefulCustomFilter<MyFilterInput, MyFilterState>(data,
+                (input, state) =>
+                {
+                    if (state.Count++ % 2 == 0)
```
justinormont: Will this run on multiple threads? If so, I'm not sure how to get the row number deterministically.

src/Microsoft.ML.Transforms/CustomMappingFactory.cs
```diff
@@ -43,4 +43,34 @@ ITransformer ICustomMappingFactory.CreateTransformer(IHostEnvironment env, strin
...
     }
+
+    /// <summary>
+    /// The base type for stateful custom mapping factories.
+    /// </summary>
+    /// <typeparam name="TSrc">The type that describes what 'source' columns are consumed from the input <see cref="IDataView"/>.</typeparam>
+    /// <typeparam name="TState">The type that describes the state object the mapping uses.</typeparam>
+    /// <typeparam name="TDst">The type that describes what new columns are added by this transform.</typeparam>
+    public abstract class StatefulCustomMappingFactory<TSrc, TState, TDst> : ICustomMappingFactory
+        where TSrc : class, new()
```
harishsk: Tiny nit: CustomMappingFactory uses <TSrc, TDst>. Would it be more consistent to order these as <TSrc, TDst, TState> or <TState, TSrc, TDst> ?

src/Microsoft.ML.Transforms/CustomMappingCatalog.cs
```diff
@@ -46,5 +46,64 @@ public static class CustomMappingCatalog
             where TSrc : class, new()
             where TDst : class, new()
             => new CustomMappingEstimator<TSrc, TDst>(catalog.GetEnvironment(), mapAction, contractName, inputSchemaDefinition, outputSchemaDefinition);
+
+        /// <summary>
+        /// Create a <see cref="StatefulCustomMappingEstimator{TSrc, TState, TDst}"/>, which applies a custom mapping of input columns to output columns,
+        /// while allowing a per-cursor state.
```
harishsk: Are we adding this more for TLC parity or is there a customer ask for this? (Not opposed to the former, just curious)

src/Microsoft.ML.Transforms/CustomMappingCatalog.cs
```diff
@@ -46,5 +46,64 @@ public static class CustomMappingCatalog
...
+        /// Drop rows where a specified predicate returns true.
+        /// </summary>
+        /// <typeparam name="TSrc">The class defining which columns to take from the incoming data.</typeparam>
+        /// <param name="catalog">The data operations catalog.</param>
+        /// <param name="input">The input data.</param>
+        /// <param name="filterPredicate">A predicate, that takes an input of type <typeparamref name="TSrc"/> and returns true if the row should be filtered (dropped) and false otherwise.</param>
+        /// <param name="inputSchemaDefinition">Additional parameters for schema mapping between <typeparamref name="TSrc"/> and input data. Useful
+        /// when dealing with annotations.</param>
+        public static IDataView CustomFilter<TSrc>(this DataOperationsCatalog catalog, IDataView input, Func<TSrc, bool> filterPredicate, SchemaDefinition inputSchemaDefinition = null)
+            where TSrc : class, new()
```
harishsk: Do we need samples illustrating these?

src/Microsoft.ML.Transforms/CustomMappingCatalog.cs
```diff
@@ -46,5 +46,64 @@ public static class CustomMappingCatalog
...
+        /// <param name="stateInitAction">The action to initialize the state object, that is called once before the cursor is initialized.</param>
+        /// <param name="contractName">The contract name, used by ML.NET for loading the model.
+        /// If <see langword="null"/> is specified, resulting transformer would not be save-able.</param>
+        /// <param name="inputSchemaDefinition">Additional parameters for schema mapping between <typeparamref name="TSrc"/> and input data.
+        /// Useful when dealing with annotations.</param>
+        /// <param name="outputSchemaDefinition">Additional parameters for schema mapping between <typeparamref name="TDst"/> and output data.
+        /// Useful when dealing with annotations.</param>
+        public static StatefulCustomMappingEstimator<TSrc, TState, TDst> StatefulCustomMapping<TSrc, TState, TDst>(this TransformsCatalog catalog, Action<TSrc, TState, TDst> mapAction,
+            Action<TState> stateInitAction, string contractName, SchemaDefinition inputSchemaDefinition = null, SchemaDefinition outputSchemaDefinition = null)
```
harishsk: Can you please add Estimator characteristics (like the others) and add Exportable to Onnx: No?
(I am guessing this will never be exportable to onnx...is that correct?)

test/Microsoft.ML.Tests/Transformers/CustomMappingTests.cs
yaeldekel: The stateful mapping transformer and filter don't use threads, exactly for this reason.

---
In reply to: [397438593](https://github.com/dotnet/machinelearning/pull/4676#discussion_r397438593) [](ancestors = 397438593)

src/Microsoft.ML.Transforms/CustomMappingCatalog.cs
yaeldekel: Yes, good idea.

---
In reply to: [399050773](https://github.com/dotnet/machinelearning/pull/4676#discussion_r399050773) [](ancestors = 399050773)

src/Microsoft.ML.Transforms/CustomMappingCatalog.cs
yaeldekel: This is for parity, I don't know about customer asks.

---
In reply to: [399043928](https://github.com/dotnet/machinelearning/pull/4676#discussion_r399043928) [](ancestors = 399043928)

test/Microsoft.ML.Tests/Transformers/CustomMappingTests.cs
```diff
@@ -121,5 +122,128 @@ public void TestSchemaPropagation()
...
+                new TextLoader.Column("Value", DataKind.Single, 2),
+            });
+            var data = loader.Load(source);
+
+            // We create a temporary environment to instantiate the custom transformer. This is to ensure that we don't need the same
+            // environment for saving and loading.
+            var tempoEnv = new MLContext();
+            var customEst = tempoEnv.Transforms.StatefulCustomMapping<MyStatefulInput, MyStatefulOutput, MyState>(MyStatefulLambda.MyStatefulAction, MyStatefulLambda.MyStateInit, nameof(MyStatefulLambda));
+
+            ML.ComponentCatalog.RegisterAssembly(typeof(MyStatefulLambda).Assembly);
```
antoniovs1029, line 183: Just a heads up that registering the assembly won't be necessary if this PR gets merged:
https://github.com/dotnet/machinelearning/pull/4989

Since it's modifying the `LambdaTransform` Save and Create methods, it will also be necessary to update `StatefulCustomMappingTransformer` in here to also save the assembly name. My understanding is that there shouldn't be any problem in that PR or in this PR to do this change.

src/Microsoft.ML.Transforms/CustomMappingCatalog.cs
```diff
@@ -46,5 +46,64 @@ public static class CustomMappingCatalog
...
+        /// <param name="stateInitAction">The action to initialize the state object, that is called once before the cursor is initialized.</param>
+        /// <param name="contractName">The contract name, used by ML.NET for loading the model.
+        /// If <see langword="null"/> is specified, resulting transformer would not be save-able.</param>
+        /// <param name="inputSchemaDefinition">Additional parameters for schema mapping between <typeparamref name="TSrc"/> and input data.
+        /// Useful when dealing with annotations.</param>
+        /// <param name="outputSchemaDefinition">Additional parameters for schema mapping between <typeparamref name="TDst"/> and output data.
+        /// Useful when dealing with annotations.</param>
+        public static StatefulCustomMappingEstimator<TSrc, TDst, TState> StatefulCustomMapping<TSrc, TDst, TState>(this TransformsCatalog catalog, Action<TSrc, TDst, TState> mapAction,
+            Action<TState> stateInitAction, string contractName, SchemaDefinition inputSchemaDefinition = null, SchemaDefinition outputSchemaDefinition = null)
```
antoniovs1029, line 72: So I guess this is right, but there's something that concerns me about letting the user to change the input and output SchemaDefinition of the transformer.

Some time ago there was this issue where the behavior of CustomMappingTransformer was different between saving the model to disk, and after loading it. My conclusion on the issue was that when saving the model, the information of the SchemaDefinition wasn't saved into disk, and so when loading back the model that information got lost.

https://github.com/dotnet/machinelearning/issues/3988

The issue remains unfixed, but I think it was closed because it's not in our plans to fix it soon and there existed a workaround for that particular issue.

In fact, I am not 100% sure if my conclusion there was correct (so perhaps having input from @yaeldekel would help), but if I was right then I'd think it's better to not let the user manipulate the SchemaDefinitions so that they don't encounter this issue when loading back their models.

On the other hand, I am also not completely sure in which case would the end user require to modify the SchemaDefinition. There's actually only 1 test (and no samples) where the SchemaDefinitions of a CustomMappingTransformer are modified, and I am still not clear about what use case would that test correspond to.

Anyway, it's just a minor concern I have.

test/Microsoft.ML.Tests/Transformers/CustomMappingTests.cs
yaeldekel: Thanks for the heads up. I guess if PR 4989 gets merged before this one, the code of `StatefulCustomMappingTransformer` won't build, so that will remind me to update this :).

---
In reply to: [401497684](https://github.com/dotnet/machinelearning/pull/4676#discussion_r401497684) [](ancestors = 401497684)

src/Microsoft.ML.Transforms/CustomMappingCatalog.cs
yaeldekel: Interesting. It seems to me that the fix for this would be to serialize both the input and the output `SchemaDefinition`s in `LambdaTransform.SaveCustomTransformer`, and then have the `ICustomMappingFactory.CreateTransformer` method take an optional input/output `SchemaDefinition`. 

Since the issue has been closed (because of low priority I guess?), should I disallow these input parameters in this new class? We cannot remove them from the existing `CustomMappingEstimator` class since that would be a breaking change, but we can at least not introduce the new class with a bug (even if it's just an edge case usage). What do you think? I'm good either way.

---
In reply to: [401514473](https://github.com/dotnet/machinelearning/pull/4676#discussion_r401514473) [](ancestors = 401514473)

antoniovs1029: I'd think it's better not to expose in here the input and output SchemaDefinition to the user. And only expose them if in the future someone requests it or if we actually get to solve the problem of serializing/deserializing SchemaDefinitions.

As for this issue on `CustomMappingTransformer`, I understand that we can't remove the SchemaDefinition parameters from the API. So I will reopen that issue, and tag it as 'breaking-api' so that it gets reviewed in the future if ML.NET 2.0 ever gets planned.

test/Microsoft.ML.Tests/Transformers/CustomMappingTests.cs
harishsk: PR 4989 is merged. 

---
In reply to: [401643916](https://github.com/dotnet/machinelearning/pull/4676#discussion_r401643916) [](ancestors = 401643916,401497684)

src/Microsoft.ML.Transforms/StatefulCustomMappingTransformer.cs
```diff
@@ -0,0 +1,398 @@
...
+
+                _bindings = new ColumnBindings(input.Schema, cols);
+
+                _typedSrc = TypedCursorable<TSrc>.Create(Host, input, false, _parent.InputSchemaDefinition);
+            }
+
+            public override DataViewRowCursor[] GetRowCursorSet(IEnumerable<DataViewSchema.Column> columnsNeeded, int n, Random rand = null)
+            {
+                return new[] { GetRowCursorCore(columnsNeeded, rand) };
+            }
```
antoniovs1029, line 139: Just curious. So I'm guessing that by simply doing this we're avoiding this Transform to use multiple threads, right? is there anything else that needs to be changed to stop this transform from using multiple threads?

antoniovs1029, line 139: So I have a question.

My impression, is that this code in here is like this so to avoid this transform to use more than one thread, is that right?

Still, on the `ShouldUseParallelCursors` it returns null if the predicate there only wants columns that were already on the input of this transformer. 

So my question is, shouldn't this method here actually return multiple cursors if `ShouldUseParallelCursors` returns null??

To be honest, I don't know if what I am asking makes sense, but for instance there's the case of `GenerateNumberTransform`, where its `GetRowCursorSet` actually checks the `ShouldUseParallelCursors` to know if it returns 1 or more cursors.

https://github.com/dotnet/machinelearning/blob/4b3cc7e2253ac195f01533459f2fc8e98d3b0cea/src/Microsoft.ML.Data/Transforms/GenerateNumberTransform.cs#L361-L378

src/Microsoft.ML.Transforms/StatefulCustomMappingTransformer.cs
```diff
@@ -0,0 +1,398 @@
...
+        private readonly string _contractName;
+
+        internal InternalSchemaDefinition AddedSchema { get; }
+        internal SchemaDefinition InputSchemaDefinition { get; }
+
+        /// <summary>
+        /// Whether a call to <see cref="ITransformer.GetRowToRowMapper(DataViewSchema)"/> should succeed, on an
+        /// appropriate schema.
+        /// </summary>
+        bool ITransformer.IsRowToRowMapper => true;
```
antoniovs1029, line 37: So 2 questions:

1) I see that `StatefulCustomMappingTransformer` doesn't implements the `IRowToRowMapper` interface (because it inherits from `RowToRowTransformBase`), so it doesn't have a `GetRow` method nor a `Row` class. According to what we chatted the other day, this would also mean that this transformer can't be used with `PredictionEngine`, is this right? I just want to make sure I am getting this correctly. I guess it makes sense since I guess `PredictionEngine` should only make predictions based on the contents of each row, and shouldn't have the ability to have a state.

2) Is it correct that this `ITransformer` is a `RowToRowMapper` whereas it returns a _Transform_ that actually doesn't implement IRowToRowMapper?

antoniovs1029, line 37: I have one question: If this is `RowToRowMapper`, then, according to what we chatted the other day, this can actually be used with `PredictionEngine`, right? But my understanding was that `PredictionEngine` process 1 row at a time. So I wonder: what will happen to the "state" on this transform? Will it actually update the state accross rows or will it restart the state on every row with the init function?

harishsk:
@yaeldekel I think this is almost ready for merging. Can you please address the pending comments?
- Adding the Exportable to onnx entry for the estimator
- Pull, rebase and test now that #4989 is merged
- Disallow SchemaDefinition parameters in StatefulCustomMapping

src/Microsoft.ML.Transforms/CustomMappingCatalog.cs
yaeldekel: Correct, it cannot be exportable to ONNX, since it is based on custom C# code specified by the user. The estimator documentation regarding this is in StatefulCustomMappingTransform.cs in line 332.

---
In reply to: [399067853](https://github.com/dotnet/machinelearning/pull/4676#discussion_r399067853) [](ancestors = 399067853)

src/Microsoft.ML.Transforms/StatefulCustomMappingTransformer.cs
yaeldekel: Yes, you're right. `ShouldUseParallelCursors` is used in the base class `TransformBase`, inside the `IDataView.GetRowCursor` method. This method gets the cursor recursively from the `Source` of each `IDataTransform`, and in order to know whether to do that using `GetRowCursor` or `GetRowCursorSet` it calls `ShouldUseParallelCursors`. If this transform (`StatefulCustomMappingTransformer.RowToRowMapper`) is somewhere in the middle of the pipeline, and none of the transform's added columns are active, then it can safely produce multiple cursors from its `Source`.

---
In reply to: [401983989](https://github.com/dotnet/machinelearning/pull/4676#discussion_r401983989) [](ancestors = 401983989)

src/Microsoft.ML.Transforms/StatefulCustomMappingTransformer.cs
yaeldekel: The init function gets called once when the getter is created, and since it is created only once when `PredictionEngine` is created, then it will use the same state object, that changes from prediction to prediction.

---
In reply to: [401998841](https://github.com/dotnet/machinelearning/pull/4676#discussion_r401998841) [](ancestors = 401998841)

harishsk:
@yaeldekel Thanks for the update. It appears only the minor addition of Onnx exportable is left before we can merge. Please update it and let me know.

harishsk:
I was looking in the wrong place. You have already added it. All the comments have been addressed. Please merge the PR if you think it is ready from your end as well. Thanks.

PR #4687, opened 2020-01-22
Updated langversion to 8.0

### LINKED ISSUE #3786: Replace LangVersion=latest with a specific version
https://github.com/dotnet/machinelearning/blob/9a80b78d7aba8463859975119e697afb72f91c09/Directory.Build.props#L94

This property means "vary the build environment for each individual developer, and break teams when possible". It should be avoided in all circumstances outside of the dotnet/roslyn test suite. Currently this project appears to use C# 7.3, so I would recommend replacing the current value with:

```xml
<LangVersion>7.3</LangVersion>
```

📝 While the compile-time behavior can be locked to a specific version by referencing the Microsoft.Net.Compilers NuGet package, the IDE features triggered by specific language versions will not be able to correctly determine the effective language version in all cases and the experience will still vary across users. Using a specific version resolves all issues and ensures the experience is consistent.

harishsk:
Fixes #3786

We cannot use 7.3 because Utilities\ThreadUtils.cs uses static local functions which is available only in 8.0. I also had to set the langVersion to 4.7 for the F# project.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -92,7 +92,8 @@
  
   <!-- Language configuration -->
   <PropertyGroup>
-    <LangVersion>8.0</LangVersion>
+    <LangVersion Condition="'$(MSBuildProjectExtension)'=='.csproj'">8.0</LangVersion>
+    <LangVersion Condition="'$(MSBuildProjectExtension)'=='.fsproj'">4.7</LangVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -92,7 +92,7 @@
  
   <!-- Language configuration -->
   <PropertyGroup>
-    <LangVersion>latest</LangVersion> <!-- default to allowing all language features -->
+    <LangVersion>8.0</LangVersion> <!-- default to allowing all language features -->
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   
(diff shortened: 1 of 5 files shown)


Directory.Build.props
```diff
@@ -92,7 +92,7 @@
  
   <!-- Language configuration -->
   <PropertyGroup>
-    <LangVersion>latest</LangVersion> <!-- default to allowing all language features -->
+    <LangVersion>8.0</LangVersion> <!-- default to allowing all language features -->
```
eerhardt: > 

This comment no longer makes sense.

sharwell:
📝 There are three additional locations where `LangVersion` can now be removed, allowing it to instead inherit from the root **Directory.Build.props**.

harishsk:
@sharwell / @eerhardt Can one of you please review and approve?

test/Microsoft.ML.FSharp.Tests/Microsoft.ML.FSharp.Tests.fsproj
```diff
@@ -5,6 +5,7 @@
     <PublicSign>false</PublicSign>
     <SourceLink></SourceLink>
     <PlatformTarget>$(TargetArchitecture)</PlatformTarget>
+    <LangVersion>4.7</LangVersion>
```
safern: Why not put this in Directory.Build.props and condition LangVersion based on the project extension? 4.7 for fsproj and 8.0 for csproj, that way if for some reason a new fsproj is added, it gets the right LangVersion and we have less duplication.

harishsk: Fixed with the latest update.

PR #4688, opened 2020-01-22
separate build pipelines

frank-dong-ms:
1. new outer loop pipeline that runs all tests
2. remove flaky tests from CI

### DIFF
--- a/test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
+++ b/test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
@@ -2122,7 +2122,11 @@ namespace Microsoft.ML.RunTests
             }
         }
 
+#if INNER_LOOP
+        [Fact(Skip = "skip due to flaky")]
+#else
         [LessThanNetCore30OrNotNetCoreFact("netcoreapp3.0 output differs from Baseline")]
+#endif
         public void EntryPointPipelineEnsembleGetSummary()
         {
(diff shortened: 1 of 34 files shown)


frank-dong-ms:
This is the CI build for latest iteration:  [https://dev.azure.com/dnceng/public/_build/results?buildId=493381&view=results](https://dev.azure.com/dnceng/public/_build/results?buildId=493381&view=results)
Success at first attempt

test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -3846,7 +3858,11 @@ public void EntryPointChainedTrainTestMacros()
             validateAuc(metrics);
         }
 
-        [RetryFact]
+#if INNER_LOOP
+        [Fact(Skip = "skip due to flaky")]
```
harishsk: Can you please change the message to something like "Skipping test temporarily. This test will be re-enabled once the cause of failures has been determined"

test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -5495,7 +5511,11 @@ public void TestCrossValidationMacroMulticlassWithWarnings()
             }
         }
 
+#if INNER_LOOP
```
harishsk: Can you please investigate whether is possible to skip the test at runtime instead of disabling it at compile time?

PR #4690, opened 2020-01-22
Make SVM Light loader create the correct schema in case of indices greater than int.MaxValue

### LINKED ISSUE #4689: SVM Light loader bug
When SVM Light loader encounters a feature index greater than `int.MaxValue` it ignores the value with this index, but it creates a schema where the length of the Features column is `int.MaxValue`. However, if the file is saved with the SVM Light saver, and then loaded again with SVM Light loader, the feature with that index will not be there, so the new loader will have a different length for the Features column. 
Saving and reloading should give the same data and the same schema, so SVM Light loader should not assign `int.MaxValue` as the length of the Features column, but instead it should assign the maximum valid index in the file.

yaeldekel:
Fixes #4689 .

### DIFF
--- a/test/Microsoft.ML.Tests/SvmLightTests.cs
+++ b/test/Microsoft.ML.Tests/SvmLightTests.cs
@@ -51,6 +51,8 @@ namespace Microsoft.ML.Tests
             data = ML.Data.LoadFromSvmLightFile(savingPath, inputSize: inputSize, zeroBased: zeroBased);
             CheckSameValues(ColumnSelectingTransformer.CreateDrop(Env, data, "Comment"),
                 ColumnSelectingTransformer.CreateDrop(Env, expectedData, "Comment"), checkId: false);
+
+            Done();
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


test/Microsoft.ML.Tests/SvmLightTests.cs
```diff
@@ -495,7 +497,7 @@ public void TestSvmLightLoaderLongIndex()
                 new SvmLightOutput() { Label = -1, Weight = 1, Features = new VBuffer<float>(6, 1, new[] { 5f }, new[] { 1 }) }
             }, schemaDef);
             var savingPath = DeleteOutputPath(TestName + "-saved-data.txt");
-            TestSvmLight(path, savingPath, 0, int.MaxValue, false, expectedData);
+            TestSvmLight(path, savingPath, 0, 6, false, expectedData);
```
harishsk: Earlier you had explained to me that the right fix was to change the size of the VBuffer to be int.MaxValue. Can you please explain why this line had to change? Does this implies that there is now no test for key with int.MaxValue or we don't support keys with int.MaxValue?

yaeldekel: We support the same scenarios as we did before, except now without a bug :-).

The old behavior was that if while loading we encounter an index greater than `int.MaxValue`, then we define the length of the Features column to be `int.MaxValue`. The change this PR is making is to define the length of Features column to be the maximum encountered index that is at most `int.MaxValue`, because the old behavior is what caused the issue that when you save and reload, the length of the Features column might change.


---
In reply to: [369680210](https://github.com/dotnet/machinelearning/pull/4690#discussion_r369680210) [](ancestors = 369680210)

PR #4691, opened 2020-01-22
Fixes cases of invalid image folder path and input column name

### LINKED ISSUE #4429: LoadImages not warning that input column is empty and ignoring imageFolder parameter in such a case
I don't know if this is actually an issue, but it's something that I noticed when working with the LoadImages method, and perhaps it is necessary to warn the user that this could happen, whether at runtime, or in the documentation.

### Issue
As I show in the source code I provide below, if all the values of the input column of a LoadImages transform are empty when fitting a pipeline, then the code will still run and not give any warning whatsoever, even though no image is actually loaded to train the model. The transform would also appear to work when transforming an input Data View which uses an empty column as input of the LoadImages, and, in the example I provide, the pipeline would still assign a predicted label to each row of the input data view bein transformed, even if no image was actually loaded.

I show this by exemplifying two main cases in which this could happen:
1. When the user loads its data through a method such as LoadFromTextFile, with an input file that only has 2 columns, but the user specifies that the ImagePath column (to be used as input column of the LoadImages method) is the 3rd column inside of the file.  This kind of scenario could happen if the user makes a typo in the ModelInput class, or if the user (perhaps mistakenly) passes an input file that doesn't contain an image path column.

2. When the user loads its data through a method such as LoadFromEnumerable from an array where all of the objects provide either a null or an empty string value to the ImagePath column.

Also notice in my code that in both cases the LoadImages transform **also ignores whatever is passed as the imageFolder parameter**, since because there are actually no ImagePaths, it will never _try_ to load images. If there was at least one ImagePath in the input dataview, then LoadImages would actually try to load that image using the imageFolder parameter, and an exception is correctly thrown if the folder doesn't exist.

### **Why is this a problem?**
- I would understand this behavior happening if the input data view doesn't provide an image path for _some_ of the rows, specially when working with big datasets. But I think it becomes a problem if there's actually no image loaded, and the whole thing appears to work without a warning, like in the example I provided. If a user unknowingly makes a mistake that leads to this problem, then s/he might believe that the model was actually correctly trained with actual images, or that it actually transformed an input where no imagePath was provided. This problem might be harder to spot in more complex pipelines or input files.

- Also the fact that the imageFolder parameter gets ignored in this case seems odd to me, as I would have expected an exception to be thrown if a user passes an inexistent folder path to the LoadImages transformer, regardless of the content of the ImagePath column.

### Source code and input file
Download solution:
[solution.zip](https://github.com/dotnet/machinelearning/files/4100381/solution.zip)

Or look into the code in here:
https://gist.github.com/antoniovs1029/997ca183411f173e81a131f09722b092

mstfbl:
Fixes #4429 

Input column names and image folders can no longer be empty or null.

### DIFF
--- a/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
+++ b/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
@@ -89,14 +89,8 @@ namespace Microsoft.ML.Data
         /// <param name="imageFolder">Folder where to look for images.</param>
         /// <param name="columns">Names of input and output columns.</param>
         internal ImageLoadingTransformer(IHostEnvironment env, string imageFolder = null, params (string outputColumnName, string inputColumnName)[] columns)
-            : base(Contracts.CheckRef(env, nameof(env)).Register(nameof(ImageLoadingTransformer)), columns)
+            : this (env, imageFolder, type: true, columns)
         {
-            // Throws ArgumentException if given imageFolder path is invalid or empty. Note: imageFolder may be null in this case.
-            if (imageFolder != null)
-                ImageFolder = Path.GetFullPath(imageFolder);
-            else
-                ImageFolder = null;
-            _useImageType = true;
         }
 
         /// <summary>
--- a/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
+++ b/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
@@ -109,11 +109,8 @@ namespace Microsoft.ML.Data
         internal ImageLoadingTransformer(IHostEnvironment env, string imageFolder, bool type = true, params (string outputColumnName, string inputColumnName)[] columns)
             : base(Contracts.CheckRef(env, nameof(env)).Register(nameof(ImageLoadingTransformer)), columns)
         {
-            if (String.IsNullOrWhiteSpace(imageFolder) || !File.Exists(imageFolder))
-            {
-                throw new ArgumentException("The provided ImageFolder path {0} is invalid.", imageFolder);
-            }
-            ImageFolder = imageFolder;
+            // Throws ArgumentException if given path is invalid/empty, ArgumentNullException if given path is null
+            ImageFolder = Path.GetFullPath(imageFolder);
             _useImageType = type;
         }
 
--- a/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
+++ b/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
@@ -88,9 +88,13 @@ namespace Microsoft.ML.Data
         /// <param name="env">The host environment.</param>
         /// <param name="imageFolder">Folder where to look for images.</param>
         /// <param name="columns">Names of input and output columns.</param>
-        internal ImageLoadingTransformer(IHostEnvironment env, string imageFolder = null, params (string outputColumnName, string inputColumnName)[] columns)
+        internal ImageLoadingTransformer(IHostEnvironment env, string imageFolder, params (string outputColumnName, string inputColumnName)[] columns)
             : base(Contracts.CheckRef(env, nameof(env)).Register(nameof(ImageLoadingTransformer)), columns)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
```diff
@@ -88,9 +88,13 @@ internal sealed class Options : TransformInputBase
...
         /// <param name="imageFolder">Folder where to look for images.</param>
         /// <param name="columns">Names of input and output columns.</param>
-        internal ImageLoadingTransformer(IHostEnvironment env, string imageFolder = null, params (string outputColumnName, string inputColumnName)[] columns)
+        internal ImageLoadingTransformer(IHostEnvironment env, string imageFolder, params (string outputColumnName, string inputColumnName)[] columns)
             : base(Contracts.CheckRef(env, nameof(env)).Register(nameof(ImageLoadingTransformer)), columns)
         {
+            if (String.IsNullOrWhiteSpace(imageFolder) || !File.Exists(imageFolder))
+            {
+                throw new ArgumentException("The provided ImageFolder path {0} is invalid.", imageFolder);
+            }
```
antoniovs1029: I wonder: why did you write all of this in here but not in the other ImageLoadingTransformer overload were you actually used `Path.GetFullPath(imageFolder);`? Is it not possible to use GetFullPath in here as well, in order to avoid having to write this 'if' statement?

antoniovs1029:
Hi, @mstfbl 

About the exception thrown if the inputFolder doesn't exist, that's generally ok. If anything, I left some comments there.

About the other changes in `ImageAnalytics/ExtensionsCatalog.cs` they're not ok. I am afraid I didn't explain myself clearly on issue #4429 , so I am thinking about updating that issue in order to make it more clear. The issue isn't that `inputColumnName` can be `null`, the problem I described there is that the `inputColumn` can be empty for all rows (i.e. that content of the column, for every row, is empty) and the trainer would still train and classify every row.

antoniovs1029:
Also, it seems that all of the pipelines are failing because of the changes on the imageFolder parameter. It seems it's supposed to be an optional parameter, so it should actually support being "null". In that case I guess you should also remove the changes you made to that.

src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
mstfbl: Good point, after I saw I can use `Path.GetFullPath(imageFolder);` I forgot to update it for this case, thank you!

mstfbl:
Thanks @antoniovs1029 for your review. I made new changes after reviewing your input. I'll make a new PR to implement the other changes once you clarify your issue.

antoniovs1029:
> Thanks @antoniovs1029 for your review. I made new changes after reviewing your input. I'll make a new PR to implement the other changes once you clarify your issue.

Hi, I had added my clarifications [on the issue.](https://github.com/dotnet/machinelearning/issues/4429#issuecomment-577414270) Idk if it's clear enough, though, let me know if you need further explanations. Thanks!

src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
```diff
@@ -91,7 +91,11 @@ internal sealed class Options : TransformInputBase
         internal ImageLoadingTransformer(IHostEnvironment env, string imageFolder = null, params (string outputColumnName, string inputColumnName)[] columns)
             : base(Contracts.CheckRef(env, nameof(env)).Register(nameof(ImageLoadingTransformer)), columns)
         {
-            ImageFolder = imageFolder;
+            // Throws ArgumentException if given imageFolder path is invalid or empty. Note: imageFolder may be null in this case.
+            if (imageFolder != null)
+                ImageFolder = Path.GetFullPath(imageFolder);
+            else
```
harishsk: The two functions are almost identical. Can you have the code inside the function in only one place and have one constructor call the other?

mstfbl: Good suggestion, done!

PR #4695, opened 2020-01-23
Updated handling of missing values with LightGBM, and added ability to use (0) as missing value

### LINKED ISSUE #4681: LightGBM trainer filters out rows with NaN features
LightGBM (binary trainer) will filter out NaN (feature)-values even though we have the option of `HandleMissingValue` which allows LightGBM to properly deal with missing values:

https://lightgbm.readthedocs.io/en/latest/Advanced-Topics.html

https://github.com/dotnet/machinelearning/blob/7a4372e1dda8f5d5c070ffce6bd1929216939d90/src/Microsoft.ML.LightGbm/LightGbmTrainerBase.cs#L178

I think this:

https://github.com/dotnet/machinelearning/blob/7a4372e1dda8f5d5c070ffce6bd1929216939d90/src/Microsoft.ML.LightGbm/LightGbmTrainerBase.cs#L439

should also specify the flag "AllFeatures" so they're allowed through.

mstfbl:
Fixes #4681 , and also adds the ability to use the numerical value (0) as missing value with LightGBM. In addition, I've also updated the baseline results for LightGbm as it is normal for them to change with `CursOpt.AllFeatures`, and all added a unit test checking that a previously trained LightGBM model with the `CursOpt.Features` flag produces the baselines it did before.


src/Microsoft.ML.LightGbm/LightGbmTrainerBase.cs
```diff
@@ -436,7 +444,7 @@ private protected virtual void GetDefaultParameters(IChannel ch, int numRow, boo
 
         private FloatLabelCursor.Factory CreateCursorFactory(RoleMappedData data)
         {
-            var loadFlags = CursOpt.AllLabels | CursOpt.Features;
+            var loadFlags = CursOpt.AllLabels | CursOpt.Features | CursOpt.AllFeatures;
```
rauhs: Slightly weird. Why not use just `CursOpt.AllLabels  | CursOpt.AllFeatures` ?

src/Microsoft.ML.LightGbm/LightGbmTrainerBase.cs
```diff
@@ -436,7 +445,7 @@ private protected virtual void GetDefaultParameters(IChannel ch, int numRow, boo
 
         private FloatLabelCursor.Factory CreateCursorFactory(RoleMappedData data)
         {
-            var loadFlags = CursOpt.AllLabels | CursOpt.Features;
+            var loadFlags = CursOpt.AllLabels | CursOpt.AllFeatures;
```
yaeldekel: >AllFeatures [](start = 56, length = 11)

What does LightGBM do if `HandleMissingValues` is false? If it ignores examples with missing values, perhaps we should filter them on our end in that case?

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
```diff
@@ -105,6 +105,7 @@ public static IEnumerable<SweepableParam> BuildLightGbmParams()
                 new SweepableDiscreteParam("MinimumExampleCountPerLeaf", new object[] { 1, 10, 20, 50 }),
                 new SweepableDiscreteParam("UseCategoricalSplit", new object[] { true, false }),
                 new SweepableDiscreteParam("HandleMissingValue", new object[] { true, false }),
+                new SweepableDiscreteParam("UseZeroAsMissingValue", new object[] { true, false }),
```
yaeldekel: >UseZeroAsMissingValue [](start = 44, length = 21)

I am not sure that this parameter should be sweepable. Isn't it just to indicate what the file format is?

justinormont: > ... Isn't it just to indicate what the file format is?

Along w/ the original file using zero to indicate missing values, `UseZeroAsMissingValue` would also indicate that we/the user have filled the missing values with `0`, which is the default imputation value in ML․NET.

> I am not sure that this parameter should be sweepable. ...

My thinking is that sweeping this is going to have minimal impact either positively or negatively. 

The main positive impact (to sweeping `UseZeroAsMissingValue`) is when the raw dataset stored its missing values as `0`, the TextLoader mapped missing to zero, or we/the user imputes them to `0` ... and if LightGBM's handling of missing values is more beneficial than treating as an ordinary value (likely, but unknown to me).

The main potential negative impact (to sweeping `UseZeroAsMissingValue`) is that additional sweep options lead to multiplicative combinations for the SMAC sweeper to figure out. The dimensionality of the hyperparameter sweep space increases. The `HandleMissingValue` interacts with `UseZeroAsMissingValue` with `HandleMissingValue` taking precedent and if false, ignores `UseZeroAsMissingValue` ([code](https://github.com/microsoft/LightGBM/blob/c7e903932a28a90ee3477fbd99d6c36219090381/src/io/bin.cpp#L336-L347)); this dependent dimension takes more examples for the sweeper to figure out.

Benchmarks showing `UseZeroAsMissingValue` improving single datasets would be helpful (demonstrating the positive impact). AutoML CLI runs showing the sweeper still gets to the same metric values on a variety of datasets without missing values (demonstrating the potential negative impact) would also be helpful. Though without the CLI currently in master, it's hard to quickly assess this change (@JakeRadMSFT, @eerhardt).

Without benchmarks, my assumption is that it's beneficial to add this sweeping.

---

Missing value handling in LightGBM: ([code](https://github.com/microsoft/LightGBM/blob/c7e903932a28a90ee3477fbd99d6c36219090381/src/io/bin.cpp#L394-L422)), ([code](https://github.com/microsoft/LightGBM/blob/d6c41fa1b287cf1f19fbd694ae6b7423840a014f/src/io/dense_bin.hpp#L151-L229)), ([code](https://github.com/microsoft/LightGBM/blob/d7f8aa53b8be5212887e97e474c90bc5d8beac19/include/LightGBM/tree.h#L221-L241)), ([issue](https://github.com/microsoft/LightGBM/issues/122))

src/Microsoft.ML.LightGbm/LightGbmTrainerBase.cs
mstfbl: I believe it does ignore examples with missing values when HandleMissingValues is false. When it is true, the missing values are imputed.

mstfbl:
The current CI tests are failing due to the current generated netcoreapp core_manifest.json not exactly matching with the generated baseline core_manifest.json. The difference here is caused by the newly added `UseZeroAsMissingValue` variable. I am currently investigating how this baseline is generated, so that I can update the generated file to match the correct core_manifest.json. 

> Output matches baseline: '..\Common\EntryPoints\core_ep-list.tsv'
> *** Failure: Output and baseline mismatch at line 11652, expected '          "Name": "MinimumExampleCountPerGroup",' but got '          "Name": "UseZeroAsMissingValue",' : '..\Common\EntryPoints\core_manifest.json'
> Test EntryPointCatalog: completed normally: failed

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
yaeldekel: Wait, I'm still confused - is this argument used to tell LightGBM how to load data from a file? If this is true, then isn't it meaningless when running in ML.NET?

---
In reply to: [371024932](https://github.com/dotnet/machinelearning/pull/4695#discussion_r371024932) [](ancestors = 371024932)

justinormont:
@mstfbl: We may want to document how to update the `core_manifest.json` and `core_ep-list.tsv` in the [developer-guide.md](https://github.com/dotnet/machinelearning/blob/master/docs/project-docs/developer-guide.md).

I think the method is to unskip the `RegenerateEntryPointCatalog` test, run it, and push changes to the two files.
https://github.com/dotnet/machinelearning/blob/97c4533df05314da44f3b784fde08bd8379488c9/test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs#L230-L256

src/Microsoft.ML.LightGbm/LightGbmTrainerBase.cs
yaeldekel: If LightGBM is going to ignore these examples anyway, we might as well not pass them to LightGBM in the first place.

---
In reply to: [371399038](https://github.com/dotnet/machinelearning/pull/4695#discussion_r371399038) [](ancestors = 371399038)

yaeldekel:
Another option is to run `EntryPointCatalog` locally, and then copy the files created in the bin directory to test\BaselineOutput\Common\EntryPoints.

---
In reply to: [579335552](https://github.com/dotnet/machinelearning/pull/4695#issuecomment-579335552) [](ancestors = 579335552)

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
justinormont: The parameter is used to tell LightGBM how we are representing its missing values in the feature vector we send it. No files involved.

When `UseZeroAsMissingValue == true`, LightGBM will treat feature values of `0.0` as a missing value ([code](https://github.com/microsoft/LightGBM/blob/c7e903932a28a90ee3477fbd99d6c36219090381/src/io/bin.cpp#L336-L347)).

LightGBM then dynamically places missing values in the side of the tree split which maximizes the split gain ([issue](https://github.com/microsoft/LightGBM/issues/122)).

src/Microsoft.ML.LightGbm/LightGbmTrainerBase.cs
justinormont: LightGBM's missing value handling won't cause the entire row to be ignored. 

When not specifically handling a missing value, (I believe) the one missing value in the row is not counted in the histogram for the feature. The non-missing values in the row are still be counted in the histograms for their respective features.

src/Microsoft.ML.AutoML/TrainerExtensions/SweepableParams.cs
yaeldekel: I see. So if I understand it correctly now, when running in ML.NET that option would indicate whether the file contains empty strings for missing values, or some other string (like `?` or `nan`) for missing values (since `TextLoader` returns zero when it reads an empty string, and `NaN` when it reads a `?`).

Although, if there are any featurizers in the pipeline, that can "corrupt" what `TextLoader` outputs, can't it? (for instance, change zero to non-zero, or change something to `NaN`?).


---
In reply to: [371927729](https://github.com/dotnet/machinelearning/pull/4695#discussion_r371927729) [](ancestors = 371927729)

justinormont: No files to consider here (or their content). 

Only part to consider is what's passed in the final `Features` column to LightGBM by ML․NET. It's indicating whether we are representing missing values as `0.0` when passed to LightGBM. Otherwise `0.0` is treated as a standard float, like any other feature value.

Generally, users will fill missing values in ML․NET w/ `0.0` since other trainers require imputing, and `0.0` is the `ReplacementMode.DefaultValue` for a float. This is done with:

```c#
pipeline = pipeline.append(mlContext.Transforms.ReplaceMissingValues(new[] {
                new InputOutputColumnPair("FeaturesImputedWithZeros", "FeaturesWithMissingValues")
            },
            MissingValueReplacingEstimator.ReplacementMode.DefaultValue));
```

src/Microsoft.ML.LightGbm/LightGbmTrainerBase.cs
justinormont: Interesting side effect of moving to `CursOpt.AllFeatures` is that feature values of +/-Infinity will also be now be allowed to flow to LightGBM.

How we use the `CursOpt.AllowBadLabels` (called `_keepBad` here) to select rows to keep:
https://github.com/dotnet/machinelearning/blob/af4ca7ce217ca7deac938d2ed4ea1b84d0028b31/src/Microsoft.ML.Data/Training/TrainerUtils.cs#L849-L863

This is good since LightGBM handles infinity nicely. Though we may want a unit test taking in a `Single.PositiveInfinity` as a feature value to LightGBM to train and to score, and seeing that it (or our wrapper code) doesn't throw to test this newfound ability.

ganik:
Pls try to run old LightGBM model (trained without your changes) with your new code. Baseline numbers should be the same.

mstfbl:
These builds are currently failing because the required baselines for `TestDatasets.breastCancerPipeMissing` have not yet been merged into the master branch. They will build succesfully when these baselines and their unit tests have been merged in PR #4785.

test/BaselineOutput/Common/LightGBMBinary/LightGBM-Test-breast-cancer-out.txt
```diff
@@ -1,4 +1,4 @@
-maml.exe Test dout=%Output% loader=Text{sparse- col=Attr:TX:6 col=Label:0 col=Features:1-5,6,7-9} data=%Data% in=../../../../test/BaselineOutput/Common/LightGBMBinary/LightGBM-Train-breast-cancer-model.zip seed=1
+maml.exe Test dout=%Output% loader=Text{sparse- col=Attr:TX:6 col=Label:0 col=Features:1-5,6,7-9} data=%Data% in=C:\Users\mubal\Documents\GitHub\machinelearning\test/Microsoft.ML.Predictor.Tests/PreviousBaselineModel\LightGBM-Train-breast-cancer-model.zip seed=1
```
justinormont: Likely want the relative path for the model

test/Microsoft.ML.TestFrameworkCommon/Datasets.cs
```diff
@@ -98,6 +98,19 @@ public static class TestDatasets
...
+        /// Modified version of breastCancerPipe without mamlExtraSettings.
+        /// </summary>
+        public static TestDataset breastCancerPipeWithoutMamlExtraSettings = new TestDataset
+        {
+            name = "breast-cancer",
+            trainFilename = "breast-cancer.txt",
+            testFilename = "breast-cancer.txt",
+            // Using "col=Features:1-5,6,7-9" improves code coverage. Same with "col=Attr:TX:6".
+            loaderSettings = "loader=Text{sparse- col=Attr:TX:6 col=Label:0 col=Features:1-5,6,7-9}",
+            extraSettings = "/cacheinst- /inst Pipe{loader=Text{sparse- col=Attr:TX:6 col=Label:0 col=Features:1-5,6,7-9} lab=Label feat=Features}"
```
justinormont: Should be able to remove (background: https://github.com/dotnet/machinelearning/issues/4742)

```suggestion
```

Issue #4693, opened 2020-01-22
SelectFeaturesBasedOn* Transformers are not outputing vectors of known size.


nsingal:
### System information

- **OS version/distro**: Windows
- **.NET Version (eg., dotnet --info)**: 

**Code that failed:**

Context.Transforms.Text.TokenizeIntoWords("TextFeature")                                   .Append(Context.Transforms.FeatureSelection.SelectFeaturesBasedOnCount("TextFeature", "TextFeature", 10)).
Append(Context.Transforms.FeatureSelection.SelectFeaturesBasedOnMutualInformation("TextFeature", "TextFeature", "Label", 5000))
.Append(Context.Transforms.NormalizeLpNorm("TextFeature", "TextFeature", Microsoft.ML.Transforms.LpNormNormalizingEstimatorBase.NormFunction.L2, true));

**Error:**
At NormalizeLpNorm, it fails with
Schema mismatch for input column 'TextFeature': expected Expected Single or known-size vector of Single, got VarVector<Single>

yaeldekel:
After investigating, I discovered that the bug is actually in the `WordBagEstimator`. In PR #4702 I am adding early checks for the input types in the feature selection estimators.

mstfbl:
This issue is fixed by #4702 .

PR #4696, opened 2020-01-23
Fix bug in WordBagEstimator when training on empty data

### LINKED ISSUE #969: TestEstimatorCore fails when training on EmptyDataView for the listed transforms.
`TestEstimatorCore` method tests the schema of estimator that of transformer. It trains estimator on EmptyDataView and get the schema. Then, it fits the estimator on actual data, and get the schema from transformer.

#### Ngram
It then verifies of both schema are same. However for the case of ngram transform, schema learnt from EmptyDataView is different from actually learned from data. The actual cause is size of ngram is zero in case of training on EmptyDataView which causes output vector type to be variable length. Additionally, some other meta are not filled properly in this case.

#### PCA Transform
Can not applied on EmptyDataView. Exception is thrown regrading empty data view.

Because of this following tests are failing right now.

WordBagWorkout
NgramWorkout
PcaWorkout

Need to think how to fix it.

yaeldekel:
Fixes #969 .
Related to issue #4693 .

### DIFF
--- a/src/Microsoft.ML.Transforms/MissingValueDroppingTransformer.cs
+++ b/src/Microsoft.ML.Transforms/MissingValueDroppingTransformer.cs
@@ -28,6 +28,31 @@ using Microsoft.ML.Transforms;
 
 namespace Microsoft.ML.Transforms
 {
+    internal sealed class MissingValueDroppingEstimator : TrivialEstimator<MissingValueDroppingTransformer>
+    {
+        public MissingValueDroppingEstimator(IHostEnvironment env, params (string outputColumnName, string inputColumnName)[] columns)
+            : base(Contracts.CheckRef(env, nameof(env)).Register(nameof(MissingValueDroppingEstimator)), new MissingValueDroppingTransformer(env, columns))
+        {
+        }
+
+        public override SchemaShape GetOutputSchema(SchemaShape inputSchema)
+        {
+            Host.CheckValue(inputSchema, nameof(inputSchema));
+
+            var resultDic = inputSchema.ToDictionary(x => x.Name);
+            foreach (var (outputColumnName, inputColumnName) in Transformer.Columns)
+            {
+                if (!inputSchema.TryFindColumn(inputColumnName, out var originalColumn))
+                    throw Host.ExceptSchemaMismatch(nameof(inputSchema), "input", inputColumnName);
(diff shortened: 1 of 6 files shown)


src/Microsoft.ML.Transforms/MissingValueDroppingTransformer.cs
```diff
@@ -28,6 +28,31 @@
 
 namespace Microsoft.ML.Transforms
 {
+    internal sealed class MissingValueDroppingEstimator : TrivialEstimator<MissingValueDroppingTransformer>
```
harishsk: Is this useful in other cases? Do we need to add this to the catalog and make it public?

src/Microsoft.ML.Transforms/Text/WordBagTransform.cs
```diff
@@ -149,17 +148,14 @@ internal static ITransformer CreateTransfomer(IHostEnvironment env, Options opti
...
-            return new TransformerChain<ITransformer>(new[] { t0, t1, t2 });
+            IEstimator<ITransformer> estimator = NgramExtractionUtils.GetConcatEstimator(h, options.Columns);
+            estimator = estimator.Append(new WordTokenizingEstimator(env, tokenizeColumns));
+            estimator = estimator.Append(NgramExtractorTransform.CreateEstimator(h, extractorArgs, estimator.GetOutputSchema(inputSchema)));
+            return estimator;
         }
 
         internal static IDataTransform Create(IHostEnvironment env, Options options, IDataView input) =>
-            (IDataTransform)CreateTransfomer(env, options, input).Transform(input);
+            (IDataTransform)CreateEstimator(env, options, SchemaShape.Create(input.Schema)).Fit(input).Transform(input)/* CreateTransfomer(env, options, input).Transform(input)*/;
```
harishsk: Nit: Can you please remove the commented code?

src/Microsoft.ML.Transforms/MissingValueDroppingTransformer.cs
yaeldekel: I don't think it is - seems to be something very specific to Ngram extraction.

---
In reply to: [370393234](https://github.com/dotnet/machinelearning/pull/4696#discussion_r370393234) [](ancestors = 370393234)

PR #4698, opened 2020-01-23
Fix for OneVersusAll Multiclass trainer

Lynx1820:
Also adding tests for OVA with FastForest, LinearSVM, and AveragedPerceptron.

### DIFF
--- a/src/Microsoft.ML.Data/Prediction/Calibrator.cs
+++ b/src/Microsoft.ML.Data/Prediction/Calibrator.cs
@@ -1750,7 +1750,7 @@ namespace Microsoft.ML.Calibrators
             var node = ctx.CreateNode(opType, new[] { scoreProbablityColumnNames[0], slopVar }, new[] { mulNodeOutput }, ctx.GetNodeName(opType), "");
 
             opType = "Add";
-            var betaVar = ctx.AddInitializer(-0.0000001f, "Slope");
+            var betaVar = ctx.AddInitializer((float)(-Offset), "Offset");
             var linearOutput = ctx.AddIntermediateVariable(null, "linearOutput", true);
             node = ctx.CreateNode(opType, new[] { mulNodeOutput, betaVar }, new[] { linearOutput }, ctx.GetNodeName(opType), "");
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.ML.StandardTrainers/Standard/LinearModelParameters.cs
```diff
@@ -138,7 +138,7 @@ private protected virtual bool SaveAsOnnx(OnnxContext ctx, string[] outputs, str
             Host.CheckValue(ctx, nameof(ctx));
             Host.Check(Utils.Size(outputs) >= 1);
             string opType = "LinearRegressor";
-            string scoreVarName = (Utils.Size(outputs) == 2) ? outputs[1] : outputs[0]; // Get Score from PredictedLabel and/or Score columns
+            string scoreVarName = (Utils.Size(outputs) >= 2) ? outputs[1] : outputs[0]; // Get Score from PredictedLabel and/or Score columns
```
harishsk: What cases result in the output size being greater than 2?

Lynx1820: When a calibrator is used, the probability column is added.  More specifically, the OVA trainer always creates a probability output column, so the output count will be 3. An alternative would be to distinguish between trainers that need the probability column and those that don't inside OVA, but  I think that would be more involved.

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -1750,7 +1750,7 @@ bool ISingleCanSaveOnnx.SaveAsOnnx(OnnxContext ctx, string[] scoreProbablityColu
             var node = ctx.CreateNode(opType, new[] { scoreProbablityColumnNames[0], slopVar }, new[] { mulNodeOutput }, ctx.GetNodeName(opType), "");
 
             opType = "Add";
-            var betaVar = ctx.AddInitializer(-0.0000001f, "Slope");
+            var betaVar = ctx.AddInitializer((float)(-Offset), "Slope");
             var linearOutput = ctx.AddIntermediateVariable(null, "linearOutput", true);
```
antoniovs1029: So I've updated this line in my last PR #4699 which just has been merged. You should sync this to the master so you get the updated version. Thanks!

PR #4699, opened 2020-01-24
Added Onnx Export to PlattCalibratorTransformer

antoniovs1029:
`PlattCalibrator` already had a `SaveAsOnnx` method ([link](https://github.com/dotnet/machinelearning/blob/2267f8d709ad053a5db5867abb396c960173d6ef/src/Microsoft.ML.Data/Prediction/Calibrator.cs#L1739)) which was called when saving to Onnx a `PlattCalibrator` through a `CalibratedModelParameter` class (such as in [here](https://github.com/dotnet/machinelearning/blob/2267f8d709ad053a5db5867abb396c960173d6ef/src/Microsoft.ML.Data/Prediction/Calibrator.cs#L391)). This would happen when saving a model produced by a calibrated binary classifier.

Besides being part of calibrated binary classifiers, a PlattCalibrator can also be used independently, through a `PlattCalibratorTransformer`. So in this PR I add the necessary code so to also make it possible to save as Onnx a model that used a `PlattCalibratorTransformer`.

I added 2 tests where a PlattCalibratorTransformer is added at the end of the model (in one test it's added on top of binary classifiers, in the other test no binary classifiers were used).

I also fixed a bug in the SaveAsOnnx method of PlattCalibrator. For some reason, it was hardcoded to use "-0.0000001f" as the value of the Offset, ignoring the actual offset that the calibrator had. This worked with the existing tests because in them the offset was actually "0", but that isn't always the case, and so in the tests that I am adding the offset is not 0.


antoniovs1029:
By the way, I tried to add more tests using the optional parameters of the PlattCalibratorEstimator, such as scoreColumnName, but it seems that API is broken. I've opened issue #4700 about this. If that issue gets fixed, then it would be a good idea to add more OnnxConversion tests to cover those cases and see if the onnx conversion still works when using "non-default" names.

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -309,6 +309,40 @@ public void BinaryClassificationTrainersOnnxConversionTest()
             Done();
         }
 
+        [Fact]
```
harishsk: This shouldn't be part of your changes. Maybe you did your merge steps wrong?

antoniovs1029: Yeah, I think I messed up, but it's fixed now

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -343,6 +343,112 @@ public void TestVectorWhiteningOnnxConversionTest()
...
+            if (Environment.Is64BitProcess)
+            {
+                estimators.Add(mlContext.BinaryClassification.Trainers.LightGbm());
+            }
+
+            var initialPipeline = mlContext.Transforms.ReplaceMissingValues("Features").
+                Append(mlContext.Transforms.NormalizeMinMax("Features"));
+            foreach (var estimator in estimators)
+            {
+                // var pipeline = initialPipeline.Append(estimator);
```
Lynx1820: Seems like you forgot to uncomment.

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -343,6 +343,112 @@ public void TestVectorWhiteningOnnxConversionTest()
...
+                string[] inputNames = onnxModel.Graph.Input.Select(valueInfoProto => valueInfoProto.Name).ToArray();
+                string[] outputNames = onnxModel.Graph.Output.Select(valueInfoProto => valueInfoProto.Name).ToArray();
+                var onnxEstimator = mlContext.Transforms.ApplyOnnxModel(outputNames, inputNames, onnxModelPath);
+                var onnxTransformer = onnxEstimator.Fit(data);
+                var onnxResult = onnxTransformer.Transform(data);
+                CompareSelectedR4ScalarColumns(transformedData.Schema.Last().Name, outputNames.Last(), transformedData, onnxResult, 3); //compare probabilities
+            }
+
+            Done();
+
```
Lynx1820: could you get rid of the extra spaces?

test/Microsoft.ML.Tests/OnnxConversionTest.cs
harishsk: I think that line needs to be deleted. 
If it is uncommented, the next line can get rid of the first Apend call. (Maybe the latter is the preferred option for readability)

PR #4701, opened 2020-01-24
Changed type of CustomGains to double[] for consistency

### LINKED ISSUE #3708: Custom gain argument for ranking: double[] in trainer, string in evaluator
The argument was recently changed to double[] in the ranking trainers, but is still a string in the ranking evaluator. We should be consistent.

mstfbl:
Fixes #3708

### DIFF
--- a/test/Microsoft.ML.TestFramework/TestCommandBase.cs
+++ b/test/Microsoft.ML.TestFramework/TestCommandBase.cs
@@ -1271,7 +1271,7 @@ namespace Microsoft.ML.RunTests
             var dataPath = CreateOutputPath("Traindata.txt");
             File.WriteAllLines(dataPath.Path, new string[] {
                 "0,0,1,2", "3,0,4,5", "4,0,5,6", "0,1,1,2", "3,1,4,5", "5,1,5,6" });
-            const string customGains = "0,3,7,15,35,63";
+            double[] customGains = new double[] { 0, 3, 7, 15, 35, 63 };
 
             RunMTAThread(() =>
             {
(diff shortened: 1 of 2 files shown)


eerhardt:
The build is breaking because this is an API breaking change. This change will break anyone's code who is using it.

test/Microsoft.ML.TestFramework/TestCommandBase.cs
```diff
@@ -1271,7 +1271,7 @@ public void CommandTrainTestMoreGains()
             var dataPath = CreateOutputPath("Traindata.txt");
             File.WriteAllLines(dataPath.Path, new string[] {
                 "0,0,1,2", "3,0,4,5", "4,0,5,6", "0,1,1,2", "3,1,4,5", "5,1,5,6" });
-            const string customGains = "0,3,7,15,35,63";
+            double[] customGains = new double[] { 0, 3, 7, 15, 35, 63 };
```
yaeldekel: >customGains [](start = 21, length = 11)

This is only used to format the string in `extraArgs` below, so it should stay a string.

yaeldekel:
The original issue (#3708) was about the `CustomGains` argument in the evaluator, which was a string, with the evaluator doing the parsing into a `double` array. That argument has since been made public (as a string).
@eerhardt, is there anything we can do about it at this point? 

---
In reply to: [578164471](https://github.com/dotnet/machinelearning/pull/4701#issuecomment-578164471) [](ancestors = 578164471)

eerhardt:
> @eerhardt, is there anything we can do about it at this point?

If you ship a public API, you can't change that API without a breaking change. The only thing you can do is add new APIs.

mstfbl:
Hi everyone, in this case I'm closing this PR as new API additions will fall outside this context.

PR #4702, opened 2020-01-24
Disallow bad input types in FeatureSelection estimator's GetOutputSchema method

yaeldekel:
Related to #4693 .


yaeldekel:
>         public void FeatureSelectionWorkout()

This test is now failing because of issue #969, so this PR should wait until PR #4696 is merged before it can be merged.

---
Refers to: test/Microsoft.ML.Tests/Transformers/FeatureSelectionTests.cs:27 in aae82c5. [](commit_id = aae82c5c17258f77b36817d5029f53ab9389c11d, deletion_comment = False)

Issue #3920, opened 2019-06-27
Stuck on evaluate


agerewines:
### System information

- **OS version/distro**: Windows 10
- **.NET Version (eg., dotnet --info)**: ML.NET 1.1.0 

### Issue

- **What did you do?**
Created ML model with Model Builder, given 1000 seconds, it was able to evaluate 2 models and returned the best one.
Afterwards I tried to rerun the modelbuilder class to recreate the model.
The csv from where I take the data is around 8 MB, with 70k rows
- **What happened?**
It has been stuck on Evaluate for so long, ~48 mins~. EDIT: now 82 mins.
- **What did you expect?**
Have some feedback on the evaluation process or create the new model.

My suggestion is to have some logs on the screen while it is evaluating.
### Source code / logs
![Proof](https://user-images.githubusercontent.com/20186579/60268902-c4ec4380-98ed-11e9-9ab5-910ee5583629.png)

justinormont:
Related issue: https://github.com/dotnet/machinelearning-modelbuilder/issues/126

> Have some feedback on the evaluation process or create the new model.
> My suggestion is to have some logs on the screen while it is evaluating.

Your suggestion sounds good to me. Though ML.NET currently has the option of a "firehose" trace log, or nothing at all. 

To quote an earlier [issue comment](https://github.com/dotnet/machinelearning/issues/3752#issuecomment-494612683):

> As mentioned in https://github.com/dotnet/machinelearning/issues/3235, `MLContext.Log()` doesn't have a verbosity selection, so it's more of a firehose. 
> 
> If a verbosity argument is added to `MLContext.Log()`, the log output from there should be human readable to see general progress. 
> 
> I believe it's still hidden within the firehose of output and once the verbosity is scaled down, you should see messages like:
> ```
> LightGBM objective=multiclassova
> [7] 'Loading data for LightGBM' finished in 00:00:15.6600468.
> [8] 'Training with LightGBM' started.
> ..................................................(00:30.58)	0/200 iterations
> ..................................................(01:00.9)	1/200 iterations
> ..................................................(01:31.2)	2/200 iterations
> ..................................................(02:01.4)	2/200 iterations
> ..................................................(02:31.9)	3/200 iterations
> ..................................................(03:02.5)	4/200 iterations
> ..................................................(03:32.9)	4/200 iterations
> ..................................................(04:03.6)	5/200 iterations
> ..................................................(04:34.4)	5/200 iterations
[...]

Issue #3235, opened 2019-04-08
Log Verbosity Handling in MLContext


daholste:
When subscribing to logs from MLContext, there may not be an easy / future-proof way to know the verbosity level of each returned log statement.
Also, when subscribing to logs, for one pipeline I ran, the logger returned a total of about 255 KB of text (lots of 'Trace'-level logs). Perhaps it could be convenient for a user, when subscribing to logs, to be able to specify a minimum log severity level if desired. (For instance, a user could subscribe to 'Warning'-level logs and above.)
Thanks for reading

eerhardt:
@CESARDELATORRE @glennc and I were just discussing this. When spitting out the Log in an ASP.NET service, it really slows down the service because of how verbose the logs are. We were thinking that 

https://github.com/dotnet/machinelearning/blob/1724da898d8cff88543fcf9e7b356ef0989b7bf7/src/Microsoft.ML.Data/LoggingEventArgs.cs#L12-L27

Should have some sort of category/level/verbosity/etc on it.

rosebyte:
Log severity is now exposed through `LoggingEventArgs.Kind`. The remaining request to configure minimum verbosity is tracked in #1852, so I’m closing this as a duplicate.

PR #4710, opened 2020-01-26
Draft modification to redirect logs to test output

harishsk:
Currently when running tests the channel output doesn't appear in the test logs. This is a draft change to enable that functionality in order to get better logging to debug tests better.

There are two changes here. 
The first is a simple adds a handler for MLContext's Log event and logs the message to test output. The second adds a TextWriter derived class to ConsoleEnvironment that makes ConsoleEnvironment also log to the test console if the code is running a test environment.

I am putting this PR up for review for two purposes. 

The first is for general verification of the code and concept. And while the code reviews are in progress I am also using this to try and debug the failure in BinaryClassifierSymSgdTest in the CI builds.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -26,3 +26,52 @@ dotnet_diagnostic.MSML_ParameterLocalVarName.severity = none
 
 # MSML_TypeParamName: Type parameter name not standard
 dotnet_diagnostic.MSML_TypeParamName.severity = none
+
+[test/Microsoft.ML.CodeAnalyzer.Tests/**.cs]
+# BaseTestClass does not apply for analyzer testing.
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
+
+[test/Microsoft.Extensions.ML.Tests/**.cs]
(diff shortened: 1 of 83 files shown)


justinormont:
Related issues on having an output verbosity level besides `zero` & `firehose`:
* https://github.com/dotnet/machinelearning/issues/3235
* https://github.com/dotnet/machinelearning/issues/3920

harishsk:
I understand the issue about the all or nothing logs. It would be good to add additional params to control the verbosity but that will impact the public API. I want the full logs right now to debug test failures. Can we defer the issue of adding additional control params?

justinormont:
> Can we defer the issue of adding additional control params?

Oh certainly. I was hoping/dreaming that both the unit test and user level logging could be solved in the same way.

harishsk:
I have added additional params to LoggingEventArgs

sharwell:
Can we not log to a file that gets attached as a build artifact?

harishsk:
I think it is better in the default test log. It gives more flexibility that if I am running a single test locally on the console, I can see the output on the console or redirect to a file as I see fit. And on the build server, it gets logged to a file anyway. It it still a file whether it is in the artifacts or the default log file.

PR #4714, opened 2020-01-27
Updated BinaryLoader & Removed RowIndexName and its functionality

### LINKED ISSUE #3707: BinaryLoader throws exception when a dataset contains 0 rows
BinaryLoader tries to create a `KeyDataViewType` with count equal to the number of rows [here](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs#L1208), but with the recent change to `KeyDataViewType` `count=0` is not allowed.

mstfbl:
Fixes #3707

Removed adding column containing the row index as it is no longer necessary and IDataViews have RowIds. Also updated BinaryLoader and its version.


yaeldekel:
>         private TableOfContentsEntry CreateRowIndexEntry(string rowIndexName)

It seems to me that this is adding a column that outputs the row index, which I am not sure is needed at all given that we have the `GetIdGetter` method. Also, this seems to be used only when `_generatedRowIndexName` is not null, which happens only if a non-null value is specified in the `Arguments` object, and that case is not exposed in any API.

---
Refers to: src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs:1201 in b9fae47. [](commit_id = b9fae47aa33f6c777b7defed1a9f89126d3fa905, deletion_comment = False)

harishsk:
>         private TableOfContentsEntry CreateRowIndexEntry(string rowIndexName)

@yaeldekel This is for bug [#3707](https://github.com/dotnet/machinelearning/issues/3707) Do we have a repro for that bug? 
Can you please explain your comment a bit more?

Do you mean we don't need to fix that bug? And do you also mean that we don't need to generate a column to hold the row id and we can delete the code associated with generating the row id?

---
In reply to: [578799056](https://github.com/dotnet/machinelearning/pull/4714#issuecomment-578799056) [](ancestors = 578799056)

---
Refers to: src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs:1201 in cc0f4d2. [](commit_id = cc0f4d2ef6e68a1a5c20350c120ffc64e6dd2c93, deletion_comment = False)

yaeldekel:
>             // int: Id of the generated row index name (can be null)

Delete this line (the comment in the `Save` function should be identical to the comment in the `Load` function).

---
Refers to: src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs:838 in e063f9e. [](commit_id = e063f9ec263f9d6fe0817c0bd76f3e29aa0c1b8f, deletion_comment = False)

src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs
```diff
@@ -852,8 +850,11 @@ private BinaryLoader(IHost host, ModelLoadContext ctx, Stream stream)
                         _threads = Math.Max(1, Environment.ProcessorCount / 2);
                     }
 
-                    _generatedRowIndexName = ctx.LoadStringOrNull();
-                    ch.CheckDecode(_generatedRowIndexName == null || !string.IsNullOrWhiteSpace(_generatedRowIndexName));
+                    if (ctx.Header.ModelVerWritten == 0x00010002 || ctx.Header.ModelVerWritten == 0x00010003)
+                    {
+                        _generatedRowIndexName = ctx.LoadStringOrNull();
```
yaeldekel: >_generatedRowIndexName [](start = 24, length = 22)

This field is not needed anymore. You can load the string into a temporary variable, or even just call `ctx.LoadStringOrNull()` without assigning it.

src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs
```diff
@@ -959,18 +960,15 @@ void ICanSaveModel.Save(ModelSaveContext ctx)
...
         /// loader, only default parameters.
         /// </summary>
-        private static void SaveParameters(ModelSaveContext ctx, int threads, string generatedRowIndexName, Double shuffleBlocks)
+        private static void SaveParameters(ModelSaveContext ctx, int threads, Double shuffleBlocks)
         {
             // *** Binary format **
             // int: Number of threads if explicitly defined, or 0 if the
             //      number of threads was automatically determined
-            // int: Id of the generated row index name (can be null)
             // Double: The randomness coefficient.
```
yaeldekel: >Double: The randomness coefficient. [](start = 15, length = 35)

I noticed that this line doesn't appear in the load method, can you please add it?

src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs
```diff
@@ -755,8 +752,9 @@ private static VersionInfo GetVersionInfo()
                 modelSignature: "BINLOADR",
                 //verWrittenCur: 0x00010001, // Initial
                 //verWrittenCur: 0x00010002, // Generated row index column
-                verWrittenCur: 0x00010003, // Number of blocks to put in the shuffle pool
-                verReadableCur: 0x00010003,
+                //verWrittenCur: 0x00010003, // Number of blocks to put in the shuffle pool
+                verWrittenCur: 0x00010004, // Row index column no longer being generated
```
yaeldekel: >0x00010004 [](start = 31, length = 10)

Please add a unit test loading an old model to see that the backwards compatibility works.

mstfbl: Hi @yaeldekel , I added a unit test where I mirrored EntryPointPipelineEnsemble(), but instead of saving and then loading that model, I load a model saved with BinaryLoader.GetVersionInfo().verWrittenCur = 0x00010003. Let me know what you think!

mstfbl: Hi @yaeldekel , there is indeed already a test that is checking for compatibility with BinaryLoader.GetVersionInfo().verWrittenCur = 0x00010003, which is the unit test `LoadOriginalBinaryLoaderModel`. As I am updating the BinaryLoader's written version from 0x00010003 to 0x00010004, I don't see the need to add a complicated new test for backwards compatibility, since a simpler version of a test like this already exists. :D

src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs
```diff
@@ -852,13 +848,12 @@ private BinaryLoader(IHost host, ModelLoadContext ctx, Stream stream)
                         _threads = Math.Max(1, Environment.ProcessorCount / 2);
                     }
 
-                    _generatedRowIndexName = ctx.LoadStringOrNull();
-                    ch.CheckDecode(_generatedRowIndexName == null || !string.IsNullOrWhiteSpace(_generatedRowIndexName));
+                    if (ctx.Header.ModelVerWritten == 0x00010002 || ctx.Header.ModelVerWritten == 0x00010003)
+                        ctx.LoadStringOrNull(); // for for _generatedRowIndexName in previous model versions
```
yaeldekel: Extra "for"

src/Microsoft.ML.Data/DataLoadSave/Binary/BinaryLoader.cs
harishsk: I am not sure I follow this. Since you are deleting a variable from the saved model, don't you need a test to validate that?

---
In reply to: [398764732](https://github.com/dotnet/machinelearning/pull/4714#discussion_r398764732) [](ancestors = 398764732)

yaeldekel: The test that Mustafa mentioned uses a model that was pre-generated using version 0x00010003, which has the row index column saved in the model, and it validates that this model can be used with the current code (version 0x00010004). Other tests (I hope they exist :-)) that create a binary loader, serialize and deserialize it test that the version 0x00010004 serialization is valid. I would be very surprised if there is no test that does this, but @mstfbl, can you verify this, just in case?

---
In reply to: [403399348](https://github.com/dotnet/machinelearning/pull/4714#discussion_r403399348) [](ancestors = 403399348,398764732)

mstfbl: Hi @yaeldekel and @harishsk , yup there are tests such as `EntryPointPipelineEnsemble` and `EntryPointMulticlassPipelineEnsemble` that do test BinaryLoader for the current 0x00010004 version.

mstfbl:
In this PR I also updated `LoadOriginalBinaryLoaderModel` to reflect that it is loading with version 3 (0x00010003) of BinaryLoader.

PR #4715, opened 2020-01-27
crash test host process when test fails to take memory dump

frank-dong-ms:
1. try crash the test host process to take memory dump when test fails
2. enable all test running on CI to investigate


frank-dong-ms:
This is one is to show how to crash test host process when investigate test issue, close for now

PR #4718, opened 2020-01-28
only run spefic test in CI for test

frank-dong-ms:
Sample PR with ability to only run specific tests on CI.


frank-dong-ms:
This just shows the idea of how to run specific tests only in CI, close as we will not check this in.

PR #4722, opened 2020-01-29
[Test Stability] PR to Test BinaryClassifierSymSgdTest

antoniovs1029:
PR to Test BinaryClassifierSymSgdTest


sharwell:
> BinaryClassifierSymSgdTest

This is the test most impacted by CNR (#4569). It will not be possible to stabilize this test without one of the following:

1. Enabling conditional numerical reproducibility
2. Internally switching to aligned data vectors before calling into the MKL BLAS implementation (this will fix some sources of flakiness, but perhaps not all of them)

sharwell:
@antoniovs1029 I have now updated #4569; you should be able to test this again after it is merged.

antoniovs1029:
So I've updated this with the different baselines for each platform, and enabling CNR, and it has passed on pretty much all builds (except for one, which was cancelled because of a timeout while downloading the data for benchmarks). This experiment consisted of running this test 100 times along with all the other enabled tests.

Now I am going to rerun this, but now running it along all the other tests (including the disabled ones), just to double check.

PR #4726, opened 2020-01-29
skip more test

frank-dong-ms:
skip below tests as they are crashing, hanging or failing the CI:

SavePipeSlidingWindowW2L1 
EntryPointSdcaBinary
MulticlassLRTest 
AutoFitImageClassificationTrainTest 
TensorFlowImageClassification
IrisVectorLightGbmWithLoadColumnName
TensorFlowTransformMNISTConvTrainingTest 
TestEstimatorHogwildSGD 
PlattCalibratorOnnxConversionTest
TensorFlowImageClassificationEarlyStopping

### DIFF
--- a/test/Microsoft.ML.Tests/OnnxConversionTest.cs
+++ b/test/Microsoft.ML.Tests/OnnxConversionTest.cs
@@ -346,6 +346,8 @@ namespace Microsoft.ML.Tests
         }
 
         [Fact]
+        //Skipping test temporarily. This test will be re-enabled once the cause of failures has been determined
+        [Trait("Category", "SkipInCI")]
         public void PlattCalibratorOnnxConversionTest()
         {
             var mlContext = new MLContext(seed: 1);
--- a/test/Microsoft.ML.TimeSeries.Tests/TimeSeries.cs
+++ b/test/Microsoft.ML.TimeSeries.Tests/TimeSeries.cs
@@ -124,6 +124,8 @@ namespace Microsoft.ML.RunTests
         }
 
         [Fact]
+        //Skipping test temporarily. This test will be re-enabled once the cause of failures has been determined
+        [Trait("Category", "SkipInCI")]
         public void SavePipeSlidingWindowW2L1()
         {
             TestCore(null, true,
--- a/test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
+++ b/test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
@@ -50,6 +50,8 @@ namespace Microsoft.ML.AutoML.Test
         }
 
         [TensorFlowFact]
+        //Skipping test temporarily. This test will be re-enabled once the cause of failures has been determined
+        [Trait("Category", "SkipInCI")]
         public void AutoFitImageClassificationTrainTest()
         {
             var context = new MLContext(seed: 1);
--- a/test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
+++ b/test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
@@ -2752,6 +2752,8 @@ namespace Microsoft.ML.RunTests
         }
 
         [Fact]
+        //Skipping test temporarily. This test will be re-enabled once the cause of failures has been determined
+        [Trait("Category", "SkipInCI")]
         public void EntryPointSdcaBinary()
         {
             TestEntryPointRoutine("breast-cancer.txt", "Trainers.StochasticDualCoordinateAscentBinaryClassifier");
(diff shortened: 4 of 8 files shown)


test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -346,6 +346,8 @@ public void TestVectorWhiteningOnnxConversionTest()
         }
 
         [Fact]
+        //Skipping test temporarily. This test will be re-enabled once the cause of failures has been determined
+        [Trait("Category", "SkipInCI")]
```
harishsk: @antoniovs1029 This is a recent test you have added. Please take a look

test/Microsoft.ML.TimeSeries.Tests/TimeSeries.cs
```diff
@@ -124,6 +124,8 @@ public void SavePipeSlidingWindowW1L1()
         }
 
         [Fact]
+        //Skipping test temporarily. This test will be re-enabled once the cause of failures has been determined
+        [Trait("Category", "SkipInCI")]
         public void SavePipeSlidingWindowW2L1()
```
justinormont: This test is from 2016.

Seems pretty odd that it would suddenly turn flaky. Is there some common cause for the tests being unstable? I fully anticipate new ones like the TF tests would be suspect (as they are less battle hardened), but old tests, I would expect to be pretty stable.

test/Microsoft.ML.TimeSeries.Tests/TimeSeries.cs
```diff
@@ -124,6 +124,8 @@ public void SavePipeSlidingWindowW1L1()
...
         {
             TestCore(null, true,
```
justinormont: I assume folks already know (and are working from it) that Azure DevOps gives us a list of the tests that filed in the last 30 days -- https://dev.azure.com/dnceng/public/_test/analytics?definitionId=104&contextType=build

image

All of those are either disabled in the last PR or this one, besides: `Microsoft.ML.RunTests.TestEntryPoints.EntryPointSDCAMulticlass`.

test/Microsoft.ML.TimeSeries.Tests/TimeSeries.cs
frank-dong-ms: Thanks Justin, we are not fully understand all the flakyness now but we will investigate one by one.

---
In reply to: [372261206](https://github.com/dotnet/machinelearning/pull/4726#discussion_r372261206) [](ancestors = 372261206)

test/Microsoft.ML.TimeSeries.Tests/TimeSeries.cs
frank-dong-ms: Yes, we are aware of this, thanks. This list itself is not accurate as this list will skip any test that failed but success after retry. 

---
In reply to: [372291720](https://github.com/dotnet/machinelearning/pull/4726#discussion_r372291720) [](ancestors = 372291720)

test/Microsoft.ML.TimeSeries.Tests/TimeSeries.cs
antoniovs1029: @justinormont I do wonder: you said this tests exists since 2016, and it wasn't flaky. Was it the case that you had to run the tests on the Azure CI Pipelines?

On my computer pretty much all the tests pass on the first run, and it's only on Azure pipelines that they become flaky, that's why I am asking.

justinormont: >  I do wonder: you said this tests exists since 2016, and it wasn't flaky. Was it the case that you had to run the tests on the Azure CI Pipelines?

This test was running on the Azure CI Pipelines since the ML.NET repo started using it. As the 2016 date implies, the unit test pre-exists the ML.NET repo, and originates in TLC.

> On my computer pretty much all the tests pass on the first run, and it's only on Azure pipelines that they become flaky, that's why I am asking.

Is there something interesting about the Azure CI Pipelines which makes unit tests unstable, yet stable in local runs? 

E.g. build server clears the build cache between builds/tests (starting from a clean slate); build server runs everything in parallel; has 32-cores vs. local machine has 4-cores; some oddity causing all unit tests to run in the same process but in a different order (causing different random numbers to be generated for each segment of the run); or odd math differences like one environment having gcc's `--ffast-math` enabled/disabled (`/fp:precise`  in VS) causing small round-off errors to accumulate.

antoniovs1029: @frank-dong-ms has mentioned to me that some tests pass when running them in isolation on the pipelines, but they tend to be flaky when ran along with all the other tests. So it seems that tests do have an effect on each other.

About the other things you mention, it might also be worth of investigation.

PR #4736, opened 2020-01-29
Changed all MLContext creation to include a fixed seed

harishsk:
Many tests are creating MLContext without a seed. Since we are seeing a number of random failures, this PR attempts to standardize all MLContext creation to include a fixed seed. 

Note to reviewers: If you know if cases where a fixed seed is not desired, please flag those cases.

### DIFF
--- a/test/Microsoft.ML.TimeSeries.Tests/TimeSeriesSimpleApiTests.cs
+++ b/test/Microsoft.ML.TimeSeries.Tests/TimeSeriesSimpleApiTests.cs
@@ -37,7 +37,7 @@ namespace Microsoft.ML.Tests
         [Fact]
         public void ChangeDetection()
         {
-            var env = new MLContext();
+            var env = new MLContext(1);
             const int Size = 10;
             var data = new List<Data>(Size);
             var dataView = env.Data.LoadFromEnumerable(data);
@@ -75,7 +75,7 @@ namespace Microsoft.ML.Tests
         [Fact]
         public void ChangePointDetectionWithSeasonality()
         {
-            var env = new MLContext();
+            var env = new MLContext(1);
             const int ChangeHistorySize = 10;
             const int SeasonalitySize = 10;
             const int NumberOfSeasonsInTraining = 5;
@@ -119,7 +119,7 @@ namespace Microsoft.ML.Tests
         [Fact]
         public void SpikeDetection()
         {
-            var env = new MLContext();
+            var env = new MLContext(1);
             const int Size = 10;
             const int PvalHistoryLength = Size / 4;
 
@@ -169,7 +169,7 @@ namespace Microsoft.ML.Tests
         [Fact]
         public void SsaSpikeDetection()
         {
-            var env = new MLContext();
+            var env = new MLContext(1);
             const int Size = 16;
             const int ChangeHistoryLength = Size / 4;
             const int TrainingWindowSize = Size / 2;
--- a/test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
+++ b/test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
@@ -20,7 +20,7 @@ namespace Microsoft.ML.AutoML.Test
         [Fact]
         public void AutoFitBinaryTest()
         {
-            var context = new MLContext();
+            var context = new MLContext(1);
             var dataPath = DatasetUtil.DownloadUciAdultDataset();
             var columnInference = context.Auto().InferColumns(dataPath, DatasetUtil.UciAdultLabel);
             var textLoader = context.Data.CreateTextLoader(columnInference.TextLoaderOptions);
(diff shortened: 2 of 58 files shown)


test/Microsoft.ML.TimeSeries.Tests/TimeSeriesSimpleApiTests.cs
```diff
@@ -169,7 +169,7 @@ public void SpikeDetection()
         [Fact]
         public void SsaSpikeDetection()
         {
-            var env = new MLContext();
+            var env = new MLContext(1);
```
justinormont: As discussion topic, does any know of a reason to not have a fixed seed for unit tests?

najeeb-kazmi:
>                     new SdcaMaximumEntropyMulticlassTrainer.Options { MaximumNumberOfIterations = 100, Shuffle = true, NumberOfThreads = 1, }));

Shuffle = false for SDCA will give more/full determinism as Justin suggests

---
Refers to: test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs:34 in 4492586. [](commit_id = 44925863ed727b66f344ac7c4cd90bd07528d0ee, deletion_comment = False)

najeeb-kazmi:
>                     new SdcaMaximumEntropyMulticlassTrainer.Options { MaximumNumberOfIterations = 100, Shuffle = true, NumberOfThreads = 1 }))

Same here

---
Refers to: test/Microsoft.ML.Tests/Scenarios/Api/Estimators/Extensibility.cs:45 in 4492586. [](commit_id = 44925863ed727b66f344ac7c4cd90bd07528d0ee, deletion_comment = False)

harishsk:
> Same here, and everywhere else SDCA shows up in tests.
> 

What about cases where Shuffle has been set to true explicitly? E.g.

https://github.com/harishsk/machinelearning/blob/44925863ed727b66f344ac7c4cd90bd07528d0ee/test/Microsoft.ML.Tests/Scenarios/Api/Estimators/DecomposableTrainAndPredict.cs#L36

https://github.com/harishsk/machinelearning/blob/44925863ed727b66f344ac7c4cd90bd07528d0ee/test/Microsoft.ML.Tests/Scenarios/Api/Estimators/Extensibility.cs#L45

https://github.com/harishsk/machinelearning/blob/44925863ed727b66f344ac7c4cd90bd07528d0ee/test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs#L34

https://github.com/harishsk/machinelearning/blob/44925863ed727b66f344ac7c4cd90bd07528d0ee/test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs#L83

najeeb-kazmi:
> > Same here, and everywhere else SDCA shows up in tests.
> 
> What about cases where Shuffle has been set to true explicitly? E.g.
> 
> https://github.com/harishsk/machinelearning/blob/44925863ed727b66f344ac7c4cd90bd07528d0ee/test/Microsoft.ML.Tests/Scenarios/Api/Estimators/DecomposableTrainAndPredict.cs#L36
> 
> https://github.com/harishsk/machinelearning/blob/44925863ed727b66f344ac7c4cd90bd07528d0ee/test/Microsoft.ML.Tests/Scenarios/Api/Estimators/Extensibility.cs#L45
> 
> https://github.com/harishsk/machinelearning/blob/44925863ed727b66f344ac7c4cd90bd07528d0ee/test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs#L34
> 
> https://github.com/harishsk/machinelearning/blob/44925863ed727b66f344ac7c4cd90bd07528d0ee/test/Microsoft.ML.Tests/Scenarios/Api/Estimators/PredictAndMetadata.cs#L83

The default is true, so I don't why it has been set explicitly. I don't see anything in the tests that would require shuffling either.

harishsk:
Would you recommend changing those to false as well in that case?

justinormont:
> Would you recommend changing those to false as well in that case?

I'm unsure if disabling shuffling with cause the unit tests to be more stable. The shuffling and TrainTest/CV splits (hopefully) listen to seed, and would then be deterministic. Can verify by inspecting the one of the TrainTest splits, or repeatedly dumping the iDataView to a file (expecting with a fixed seed they stable perfectly the same, and changing the seed gets you a different set and ordering of rows).

harishsk:
How about we do this in two passes then? Can we merge these changes in first and then do the shuffle=false in a second pass?

frank-dong-ms:
Do you plan to do some test whether this will improve stability?

najeeb-kazmi:
> How about we do this in two passes then? Can we merge these changes in first and then do the shuffle=false in a second pass?

Sounds good to me. Should we also add "seed=1" to the predictor tests in another PR or do you want to take care of this here?

harishsk:
> 
> 
> Do you plan to do some test whether this will improve stability?

By all consensus, seeded unit tests are the right thing to do. The tests passed here on first attempt. So I don't think this has made it any worse. I plan to run full tests against these also.

PR #4737, opened 2020-01-30
fix path in test that not working in linux and mac os

frank-dong-ms:
Path in TestAPI is not working in linux and mac os and cause publish to artifacts fails:

https://dev.azure.com/dnceng/public/_build/results?buildId=501206&view=logs&j=5aa5c7df-492a-5eaf-973a-62b7b0f0ee3b

with error message:

Fail to upload '/home/vsts/work/1/a/AnyCPU.Release/Microsoft.ML.Tests/netcoreapp2.1/TestOutput/..\Common\Api/lambda-output.tsv' due to 'The following path is not valid: Ubuntu_x64_NetCoreApp21 R/AnyCPU.Release/Microsoft.ML.Tests/netcoreapp2.1/TestOutput/../Common/Api/lambda-output.tsv. Path segments must contain characters other than period and whitespace.'.

After fix, artifacts upload is success in all platform, below build is with this fix but an on purpose test fail, all builds fails and successfully uploads artifacts:
https://dev.azure.com/dnceng/public/_build/results?buildId=501376&view=results

### DIFF
--- a/test/Microsoft.ML.Tests/Scenarios/Api/TestApi.cs
+++ b/test/Microsoft.ML.Tests/Scenarios/Api/TestApi.cs
@@ -20,7 +20,7 @@ namespace Microsoft.ML.Tests.Scenarios.Api
 {
     public partial class TestApi : BaseTestClass
     {
-        private const string OutputRelativePath = @"..\Common\Api";
+        private const string OutputRelativePath = @"../Common/Api";
 
         public TestApi(ITestOutputHelper output) : base(output)
         {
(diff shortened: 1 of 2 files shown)


test/Microsoft.ML.Tests/Scenarios/Api/TestApi.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.ML.Tests.Scenarios.Api
 {
     public partial class TestApi : BaseTestClass
     {
-        private const string OutputRelativePath = @"..\Common\Api";
+        private const string OutputRelativePath = @"../Common/Api";
```
eerhardt: Note: Typically it is best to use `Path.Combine` to build up paths when you want the code to work cross-platform.

```C#
private const string OutputRelativePath = Path.Combine("..", "Common", "Api");
```

PR #4744, opened 2020-01-30
add run specific test capability and CrashTestHostProcessorHelper

frank-dong-ms:
2 purpose for this PR:
1. add capability of run specific test only on CI by:
   a. set innerLoop to false to certain configuration
   b. set runSpecific to true to certain configuration
   c. add [Trait("Category", "RunSpecificTests")] to the test methods wish to run
2. add CrashTestHostProcessorHelper to crash process and take memory dump at any place


test/Microsoft.ML.TestFrameworkCommon/CrashTestHostProcessorHelper.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
```
harishsk: I know I had asked you to add this, but on second thoughts, it doesn't seem to add much. It is simply a rename over an existing function. Maybe we can delete this and just use Environment.FailFast directly wherever we would have called CrashTestHostProcessHelper.CrashTestHostProcess()
Maybe we should delete this.

frank-dong-ms: Yes, it is just one line function I'm okay with delete this.

---
In reply to: [373198033](https://github.com/dotnet/machinelearning/pull/4744#discussion_r373198033) [](ancestors = 373198033)

PR #4745, opened 2020-01-30
Double cast to float for some onnx estimators 

Lynx1820:


src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs
```diff
@@ -800,6 +801,16 @@ private bool SaveAsOnnxCore(OnnxContext ctx, int iinfo, ColInfo info, string src
...
+                {
+                    var castOutput = ctx.AddIntermediateVariable(null, "castOutput", true);
+                    castNode = ctx.CreateNode("Cast", srcVariableName, castOutput, ctx.GetNodeName(opType), "");
+                    var t2 = InternalDataKindExtensions.ToInternalDataKind(DataKind.Single).ToType();
+                    castNode.AddAttribute("to", t2);
+                    node = ctx.CreateNode(opType, castOutput, labelEncoderOutput, ctx.GetNodeName(opType));
+                    var terms = GetTermsAndIds<double>(iinfo, out termIds);
+                    node.AddAttribute("keys_floats", terms);
+                }
                 else
```
ganik: > else [](start = 15, length = 5)

Please add also support for int64. LabelEncoder supports int64 type

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs
```diff
@@ -811,6 +811,16 @@ private bool SaveAsOnnxCore(OnnxContext ctx, int iinfo, ColInfo info, string src
                     var terms = GetTermsAndIds<double>(iinfo, out termIds);
                     node.AddAttribute("keys_floats", terms);
                 }
+                else if (info.TypeSrc.GetItemType().Equals(NumberDataViewType.Int64))
```
harishsk: I think this is different enough that it warrants some test. Can you please add one?

PR #4746, opened 2020-01-30
Implement MSML_ExtendBaseTestClass (Test classes should be derived from BaseTestClass)

sharwell:
All current violations are excluded from analysis via **.editorconfig**. After this is merged, someone can enable the diagnostic in one project at a time and ensure there are no violations.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -26,3 +26,52 @@ dotnet_diagnostic.MSML_ParameterLocalVarName.severity = none
 
 # MSML_TypeParamName: Type parameter name not standard
 dotnet_diagnostic.MSML_TypeParamName.severity = none
+
+[test/Microsoft.ML.CodeAnalyzer.Tests/**.cs]
+# BaseTestClass does not apply for analyzer testing.
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
+
+[test/Microsoft.Extensions.ML.Tests/**.cs]
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
+
+[test/Microsoft.ML.Tests/**.cs]
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
+
+[test/Microsoft.ML.AutoML.Tests/**.cs]
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
+
+[test/Microsoft.ML.Benchmarks.Tests/**.cs]
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
+
+[test/Microsoft.ML.CodeGenerator.Tests/**.cs]
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
+
+[test/Microsoft.ML.Core.Tests/**.cs]
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
+
+[test/Microsoft.ML.CpuMath.UnitTests/**.cs]
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
+
+[test/Microsoft.ML.Functional.Tests/**.cs]
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
+
+[test/Microsoft.ML.Predictor.Tests/**.cs]
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
+
+[test/Microsoft.ML.Sweeper.Tests/**.cs]
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
+
+[test/Microsoft.ML.TimeSeries.Tests/**.cs]
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
--- a/tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
+++ b/tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
@@ -1,0 +1,134 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Concurrent;
+using System.Collections.Immutable;
+using System.Diagnostics;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.ML.InternalCodeAnalyzer
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class BaseTestClassAnalyzer : DiagnosticAnalyzer
+    {
+        private const string Category = "Test";
+        internal const string DiagnosticId = "MSML_ExtendBaseTestClass";
+
+        private const string Title = "Test classes should be derived from BaseTestClass";
+        private const string Format = "Test class '{0}' should extend BaseTestClass.";
+        private const string Description =
+            "Test classes should be derived from BaseTestClass.";
+
+        private static DiagnosticDescriptor Rule =
+            new DiagnosticDescriptor(DiagnosticId, Title, Format, Category,
+                DiagnosticSeverity.Warning, isEnabledByDefault: true, description: Description);
+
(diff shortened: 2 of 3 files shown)


tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.EnableConcurrentExecution();
+
+            context.RegisterCompilationStartAction(AnalyzeCompilation);
+        }
+
+        private void AnalyzeCompilation(CompilationStartAnalysisContext context)
```
sharwell: This method sets up the analysis for a compilation, which is the compilation of one assembly (normally one project, but can occur more than once for a project if you multi-target).

tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            if (!(context.Compilation.GetTypeByMetadataName("Xunit.FactAttribute") is { } factAttribute))
+            {
+                return;
+            }
```
sharwell: If the project doesn't reference `Xunit.FactAttribute`, then return early because there are no test classes of interest.

tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+
+            var analyzerImpl = new AnalyzerImpl(context.Compilation, factAttribute);
+            context.RegisterSymbolAction(analyzerImpl.AnalyzeNamedType, SymbolKind.NamedType);
```
sharwell: `analyzerImpl` is a state object in the context of one `Compilation`. Analyzers which require state (in this case to cache the symbol lookups of `FactAttribute` and `BaseTestClass`) can do so in one of the `*StartAnalysisContext` callbacks.

tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+                _compilation = compilation;
+                _factAttribute = factAttribute;
+                _baseTestClass = _compilation.GetTypeByMetadataName("Microsoft.ML.TestFramework.BaseTestClass");
+            }
+
+            public void AnalyzeNamedType(SymbolAnalysisContext context)
+            {
+                var namedType = (INamedTypeSymbol)context.Symbol;
+                if (namedType.TypeKind != TypeKind.Class)
+                    return;
```
sharwell: This method is called for all named types. We know the symbol is a `INamedTypeSymbol`, but we don't know what kind of named type it is. Since we only want to analyze classes, we check that and return immediately for any other type kind.

tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+
+                if (ExtendsBaseTestClass(namedType))
+                    return;
```
sharwell: Avoid looking at all the members if the class already extends `BaseTestClass`.

tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+                {
+                    if (member is IMethodSymbol method && IsTestMethod(method))
+                    {
+                        hasTestMethod = true;
+                        break;
+                    }
+                }
+
+                if (!hasTestMethod)
+                    return;
```
sharwell: If the class does not have any test methods, it does not need to extend `BaseTestClass`.

tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+
+                context.ReportDiagnostic(Diagnostic.Create(Rule, namedType.Locations[0], namedType));
```
sharwell: At this point, we have a class with test methods, but the class did not extend `BaseTestClass`. We report a diagnostic instance, using the descriptor `Rule` defined above. The location is the identifier for the class, and the third parameter (`namedType`) is the message argument for the `{0}` in the message format.

tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            private bool IsTestMethod(IMethodSymbol method)
```
sharwell: A test method is a method with a test attribute

tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+                foreach (var attribute in method.GetAttributes())
+                {
+                    if (IsTestAttribute(attribute.AttributeClass))
+                        return true;
+                }
+
+                return false;
+            }
+
+            private bool IsTestAttribute(INamedTypeSymbol attributeClass)
```
sharwell: A test attribute is any attribute that extends `Xunit.FactAttribute`

tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+            {
+                if (_knownTestAttributes.TryGetValue(attributeClass, out var isTest))
```
sharwell: We cache the "is test attribute" result for each attribute we encounter, since a few types of attributes tend to be used in many locations.

tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            private bool ExtendsBaseTestClass(INamedTypeSymbol namedType)
+            {
+                if (_baseTestClass is null)
+                    return false;
+
+                for (var current = namedType; current is object; current = current.BaseType)
+                {
+                    if (Equals(current, _baseTestClass))
```
sharwell: Symbols need to be compared with `Equals` (as opposed to `==`). We have an analyzer for this to help developers avoid mistakes.

tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        private static DiagnosticDescriptor Rule =
+            new DiagnosticDescriptor(DiagnosticId, Title, Format, Category,
+                DiagnosticSeverity.Warning, isEnabledByDefault: true, description: Description);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
```
sharwell: `None` is the normal behavior for analyzers, but it is not the *default* behavior. It is a combination of two things:

* This analyzer does not need to analyze generated code (i.e. it will not receive Named Type symbol callbacks for types defined in generated code files, such as **\*.Designer.cs** files). Some analyzers (e.g. analysis of unused code) will not be accurate if generated code is skipped, but this analyzer is not impacted.
* This analyzer does not report diagnostics in generated code.

tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+            context.EnableConcurrentExecution();
```
sharwell: This analyzer is safe for concurrent execution. In particular, `AnalyzerImpl.AnalyzeNamedType` can be called concurrently for different types.

.editorconfig
```diff
@@ -26,3 +26,52 @@ dotnet_diagnostic.MSML_ParameterLocalVarName.severity = none
...
 dotnet_diagnostic.MSML_TypeParamName.severity = none
+
+[test/Microsoft.ML.CodeAnalyzer.Tests/**.cs]
+# BaseTestClass does not apply for analyzer testing.
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
+
+[test/Microsoft.Extensions.ML.Tests/**.cs]
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
```
sharwell: This block disables the `MSML_ExtendBaseTestClass` rule for source files under **test/Microsoft.Extensions.ML.Tests**. This is required because one or more classes in this folder is considered a test class by this analyzer but does not extend `BaseTestClass`. However, this pull request only implements the analyzer; a future pull request will remove these lines from the configuration file and update the test classes to correct the violations.

.editorconfig
```diff
@@ -26,3 +26,52 @@ dotnet_diagnostic.MSML_ParameterLocalVarName.severity = none
 
 # MSML_TypeParamName: Type parameter name not standard
 dotnet_diagnostic.MSML_TypeParamName.severity = none
+
+[test/Microsoft.ML.CodeAnalyzer.Tests/**.cs]
+# BaseTestClass does not apply for analyzer testing.
+# MSML_ExtendBaseTestClass: Test classes should be derived from BaseTestClass
+dotnet_diagnostic.MSML_ExtendBaseTestClass.severity = none
```
sharwell: We never want to report `BaseTestClass` violations in the code analyzer tests. These are "meta" tests for the analyzer itself, and thus follows the patterns for analyzer testing instead of the patterns for ML.NET testing.

PR #4750, opened 2020-01-31
Added Shuffle=false for all the Sdca related tests

harishsk:
See discussion [here](https://github.com/dotnet/machinelearning/pull/4736)

Reviewers please check if any of them need to have to Shuffle=true.

### DIFF
--- a/test/Microsoft.ML.Benchmarks/PredictionEngineBench.cs
+++ b/test/Microsoft.ML.Benchmarks/PredictionEngineBench.cs
@@ -58,7 +58,7 @@ namespace Microsoft.ML.Benchmarks
             var pipeline = new ColumnConcatenatingEstimator(env, "Features", new[] { "SepalLength", "SepalWidth", "PetalLength", "PetalWidth" })
                 .Append(env.Transforms.Conversion.MapValueToKey("Label"))
                 .Append(env.MulticlassClassification.Trainers.SdcaMaximumEntropy(
-                    new SdcaMaximumEntropyMulticlassTrainer.Options { NumberOfThreads = 1, ConvergenceTolerance = 1e-2f, }));
+                    new SdcaMaximumEntropyMulticlassTrainer.Options { NumberOfThreads = 1, ConvergenceTolerance = 1e-2f, Shuffle = false }));
 
             var model = pipeline.Fit(data);
 
@@ -93,7 +93,7 @@ namespace Microsoft.ML.Benchmarks
 
             var pipeline = mlContext.Transforms.Text.FeaturizeText("Features", "SentimentText")
                 .Append(mlContext.BinaryClassification.Trainers.SdcaNonCalibrated(
-                    new SdcaNonCalibratedBinaryTrainer.Options { NumberOfThreads = 1, ConvergenceTolerance = 1e-2f, }));
+                    new SdcaNonCalibratedBinaryTrainer.Options { NumberOfThreads = 1, ConvergenceTolerance = 1e-2f, Shuffle = false}));
 
             var model = pipeline.Fit(data);
 
@@ -127,7 +127,7 @@ namespace Microsoft.ML.Benchmarks
             IDataView data = loader.Load(_breastCancerDataPath);
 
             var pipeline = env.BinaryClassification.Trainers.SdcaNonCalibrated(
-                new SdcaNonCalibratedBinaryTrainer.Options { NumberOfThreads = 1, ConvergenceTolerance = 1e-2f, });
+                new SdcaNonCalibratedBinaryTrainer.Options { NumberOfThreads = 1, ConvergenceTolerance = 1e-2f, Shuffle = false});
 
             var model = pipeline.Fit(data);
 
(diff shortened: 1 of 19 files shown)


test/Microsoft.ML.Benchmarks/PredictionEngineBench.cs
```diff
@@ -58,7 +58,7 @@ public void SetupIrisPipeline()
             var pipeline = new ColumnConcatenatingEstimator(env, "Features", new[] { "SepalLength", "SepalWidth", "PetalLength", "PetalWidth" })
                 .Append(env.Transforms.Conversion.MapValueToKey("Label"))
                 .Append(env.MulticlassClassification.Trainers.SdcaMaximumEntropy(
-                    new SdcaMaximumEntropyMulticlassTrainer.Options { NumberOfThreads = 1, ConvergenceTolerance = 1e-2f, }));
+                    new SdcaMaximumEntropyMulticlassTrainer.Options { NumberOfThreads = 1, ConvergenceTolerance = 1e-2f, Shuffle = false }));
```
justinormont: I'm unsure if disabling shuffling with cause the unit tests to be more stable. The shuffling and TrainTest/CV splits (hopefully) listen to seed, and would then be deterministic. 

Can verify by inspecting the one of the TrainTest splits, or repeatedly dumping the iDataView to a file (expecting with a fixed seed they stable perfectly the same, and changing the seed gets you a different set and ordering of rows).

Specifically, I would first test if shuffling is already stable given a seed, before making this change. Shuffling is beneficial to have turned on.

Copied from: https://github.com/dotnet/machinelearning/pull/4736#issuecomment-580515353

codemzs: I agree with Justin. We just need to check the final value to be within certain range. Shuffling is important for SDCA.

najeeb-kazmi: So shuffle seed is indeed coming from `MLContext`:
https://github.com/dotnet/machinelearning/blob/24c827416d23a6a74f16f71ec9123ca15e0e12fd/src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs#L555-L557

`Host` is being registered from the environment here:
https://github.com/dotnet/machinelearning/blob/24c827416d23a6a74f16f71ec9123ca15e0e12fd/src/Microsoft.ML.StandardTrainers/Standard/SdcaBinary.cs#L300-L302
https://github.com/dotnet/machinelearning/blob/24c827416d23a6a74f16f71ec9123ca15e0e12fd/src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs#L385-L392

and `_rand` is getting its seed from `MLContext`:
https://github.com/dotnet/machinelearning/blob/24c827416d23a6a74f16f71ec9123ca15e0e12fd/src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs#L350-L355
https://github.com/dotnet/machinelearning/blob/24c827416d23a6a74f16f71ec9123ca15e0e12fd/src/Microsoft.ML.Data/Utilities/LocalEnvironment.cs#L49-L50
https://github.com/dotnet/machinelearning/blob/24c827416d23a6a74f16f71ec9123ca15e0e12fd/src/Microsoft.ML.Data/MLContext.cs#L110-L115

najeeb-kazmi: `TrainTestSplit`, `CrossValidationSplit`, and `CrossValidate` have a seed, but when not specified, the seed doesn't come from `MLContext`. Instead, `GenerateNumberTransform` and `HashingEstimator` use their own default seeds. See #4752

test/Microsoft.ML.Benchmarks/PredictionEngineBench.cs
```diff
@@ -127,7 +127,7 @@ public void SetupBreastCancerPipeline()
             IDataView data = loader.Load(_breastCancerDataPath);
 
             var pipeline = env.BinaryClassification.Trainers.SdcaNonCalibrated(
-                new SdcaNonCalibratedBinaryTrainer.Options { NumberOfThreads = 1, ConvergenceTolerance = 1e-2f, });
+                new SdcaNonCalibratedBinaryTrainer.Options { NumberOfThreads = 1, ConvergenceTolerance = 1e-2f, Shuffle = false});
```
frank-dong-ms: It seems we better create different tests to test againt Shuffle=true and Shuffle=false.
  1. when Shuffle = false, the result should be deterministic and we can assert value
  2. when Shuffle = true, the result should lie in certain range and we can assert range

justinormont: With `shuffle = false` you won't be fully deterministic as there's always race conditions in due to threading.

Easy example of a four core machine with four threads training a model slightly differently than a two core machine. 

Lets assume there's either (1) a single shared PRNG in a component where N threads pull random numbers at a slightly different ordering each run, or (2) each thread has its own PRNG where the number of threads affect the random numbers generated.

frank-dong-ms: I see, thanks for your explain.
So in that case, we can use a way (for example mock the parts that is not deterministic) to gain deterministic and be able to do test specific part of code is correct, this is what we usually do for unit test (mock operation that is not deterministic or have outer dependency).
In integration test, we can assert on range to ensure whole scenario is correct.


---
In reply to: [373736059](https://github.com/dotnet/machinelearning/pull/4750#discussion_r373736059) [](ancestors = 373736059)

harishsk:
For all the reasons cited above, I am abandoning this PR.

PR #4755, opened 2020-01-31
Check exception status even if TF_SessionRun throws an exception

sharwell:
Extracted from #4569


sharwell:
📝 Note that this is nice in general, but for the exception case we hit there is still no additional information about the crash. 😦

sharwell:
@antoniovs1029 I'm removing this **test** label since this is a production code change.

PR #4756, opened 2020-01-31
Update test baseline for TimeSeries tests

sharwell:
Extracted from #4569

### DIFF
--- a/test/Microsoft.ML.TimeSeries.Tests/TimeSeriesDirectApi.cs
+++ b/test/Microsoft.ML.TimeSeries.Tests/TimeSeriesDirectApi.cs
@@ -5,15 +5,21 @@
 using System.Collections.Generic;
 using System.IO;
 using Microsoft.ML.Data;
+using Microsoft.ML.TestFramework;
 using Microsoft.ML.TestFramework.Attributes;
 using Microsoft.ML.TestFrameworkCommon.Attributes;
 using Microsoft.ML.Transforms.TimeSeries;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace Microsoft.ML.Tests
 {
-    public sealed class TimeSeries
+    public sealed class TimeSeries : BaseTestClass
     {
+        public TimeSeries(ITestOutputHelper output)
+            : base(output)
+        {
+        }
 
         private sealed class Prediction
         {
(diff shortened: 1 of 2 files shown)


test/Microsoft.ML.TimeSeries.Tests/TimeSeriesDirectApi.cs
```diff
@@ -163,8 +169,13 @@ public void ChangePointDetectionWithSeasonality()
             // Get predictions
             var enumerator = env.Data.CreateEnumerable<Prediction>(output, true).GetEnumerator();
             Prediction row = null;
+#if NETCOREAPP3_0
```
frank-dong-ms: Could you please help me understand the root cause of why netcore 3.0 generates different result?

antoniovs1029: I am curious to know as well :)

sharwell: @tannergooding ?

harishsk: @tannergooding / @sharwell Any updates on this? It would be really good to know why netcore3.0 would generate different results. It may help with other investigations?

harishsk: I appears that previously the tests werent running when .net version was less than 3.0. And @sharwell's fix improves that by getting the tests to run on both again but with different baselines. 

@frank-dong-ms: Can we accept this PR and re-investigate the issue of different baselines later?

@sharwell Can you please amend this PR with an // TODO: [TEST_STABILITY] comment? We are using that tag to track issues in code that still need to be addressed beyond outright test failures?

harishsk:
@sharwell Can you please fix the merge conflicts and address the small change requested in my last comment so we can merge this PR?

harishsk:
@sharwell Can you please update the PR according to the note above?

harishsk:
@frank-dong-ms : Can you please review this PR and close it if you have already incorporated all these changes into your prior PRs?

frank-dong-ms:
This is done through our work of all legacy tests. We created separate baseline/expected result for some tests that generates different result for netcore app 3.1. The reason of these difference comes from CPUMath library, for netcoreapp 3.1 we are using AVX instruction set and for netcoreapp2.1 and netfx we are using SSE instruction set and these 2 instruction set generates slightly different results.

PR #4762, opened 2020-01-31
Enable MSML_TypeParamName for the full solution

sharwell:


test/Microsoft.ML.AutoML.Tests/UserInputValidationTests.cs
```diff
@@ -346,7 +346,7 @@ private static void ValidateLabelTypeTestCore<LabelRawType>(TaskKind task, Primi
             }
             else
             {
-                dataViewBuilder.AddColumn(DefaultColumnNames.Label, labelType, Activator.CreateInstance<LabelRawType>());
+                dataViewBuilder.AddColumn(DefaultColumnNames.Label, labelType, Activator.CreateInstance<TLabelRawType>());
```
justinormont: What is a `TLabelRawType`?

sharwell: I'm not sure, but the .NET convention for type parameters is starting with `T`.

justinormont: Ahh, just a simple generic type defined above. Was thinking I was missing something obvious ... I was.

#resolved

PR #4764, opened 2020-02-01
Data splits to default to MLContext seed when not specified

### LINKED ISSUE #4752: Default seed is not propagated from MLContext
In theory, the seed set in `MLContext` is intended to provide the global seed for all components and operations requiring randomness, e.g. sampling, permutation, etc. In practice, this doesn't always hold true.

`TrainTestSplit`, `CrossValidationSplit`, and `CrossValidate` all have a user specified seed and call `EnsureGroupPreservationColumn`, which in turn uses `GenerateNumberTransform` and `HashingEstimator`.

When the seed is not specified by the user, it is not derived from `MLContext`. Instead, `GenerateNumberTransform` and `HashingEstimator` use their own defaults, so that if a user doesn't specify a seed to `TrainTestSplit`, `CrossValidationSplit`, or `CrossValidate`, they will always get a deterministic split regardless of the seed in `MLContext`.

https://github.com/dotnet/machinelearning/blob/24c827416d23a6a74f16f71ec9123ca15e0e12fd/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs#L496-L505
https://github.com/dotnet/machinelearning/blob/24c827416d23a6a74f16f71ec9123ca15e0e12fd/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs#L521-L525

cc: @harishsk @justinormont

najeeb-kazmi:
Fixes #4752

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
@@ -495,13 +495,14 @@ namespace Microsoft.ML
         /// </summary>
         internal static void EnsureGroupPreservationColumn(IHostEnvironment env, ref IDataView data, ref string samplingKeyColumn, int? seed = null)
         {
+            Contracts.CheckValue(env, nameof(env));
+            var host = env.Register("rand");
             // We need to handle two cases: if samplingKeyColumn is provided, we use hashJoin to
             // build a single hash of it. If it is not, we generate a random number.
-
             if (samplingKeyColumn == null)
             {
                 samplingKeyColumn = data.Schema.GetTempColumnName("SamplingKeyColumn");
-                data = new GenerateNumberTransform(env, data, samplingKeyColumn, (uint?)seed);
+                data = new GenerateNumberTransform(env, data, samplingKeyColumn, (uint?)(seed ?? host.Rand.Next()));
             }
             else
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
```diff
@@ -495,13 +495,14 @@ internal static IEnumerable<TrainTestData> CrossValidationSplit(IHostEnvironment
...
+            Contracts.CheckValue(env, nameof(env));
+            var host = env.Register("rand");
             // We need to handle two cases: if samplingKeyColumn is provided, we use hashJoin to
             // build a single hash of it. If it is not, we generate a random number.
-
             if (samplingKeyColumn == null)
             {
                 samplingKeyColumn = data.Schema.GetTempColumnName("SamplingKeyColumn");
-                data = new GenerateNumberTransform(env, data, samplingKeyColumn, (uint?)seed);
+                data = new GenerateNumberTransform(env, data, samplingKeyColumn, (uint?)(seed ?? host.Rand.Next()));
```
justinormont: Why fallback to `host.Rand.Next()` instead of the fixed seed given to the MLContext?

I think this is doing (1), then (4) from: https://github.com/dotnet/machinelearning/issues/4752#issuecomment-580865994

`TrainTestSplit` etc should listen to `MLContext` seed if not given by the user for the specific component.

> I think the precedence for PRNG seeds should be:
> 1. User seed specified for the component instance (or from parent component if hiding behind another)
> 2. User seed specified in `MLContext`
> 3. Default seed in component (unsure if we should remove these, might be a component-by-component discussion)
> 4. Random (and more random than current -- see https://github.com/dotnet/machinelearning/pull/4645#discussion_r365552609)

najeeb-kazmi: MLContext seed isn't exposed, i.e. I cannot directly use mlContext.Seed. The seed for Host.Rand is set from MLContext though.

PR #4765, opened 2020-02-01
Enable MSML_SingleVariableDeclaration for the full solution

sharwell:


test/Microsoft.ML.AutoML.Tests/UserInputValidationTests.cs
```diff
@@ -346,7 +346,7 @@ private static void ValidateLabelTypeTestCore<LabelRawType>(TaskKind task, Primi
             }
             else
             {
-                dataViewBuilder.AddColumn(DefaultColumnNames.Label, labelType, Activator.CreateInstance<LabelRawType>());
+                dataViewBuilder.AddColumn(DefaultColumnNames.Label, labelType, Activator.CreateInstance<TLabelRawType>());
```
justinormont: Your branch includes the older commit https://github.com/dotnet/machinelearning/commit/0500b02de58ae430d60cfac22a96d041ab8f8e02, which already made it to master (its merge commit https://github.com/dotnet/machinelearning/commit/be4b42d2ed308712f990bebda0e52a62ffe43e54 has a new commit id, hence this still shows here). 

Pretty minor as it's just a few extra lines of code showing in this PR's diff. A git guru would know how to fix, though it's probably not worth fixing. Building a new branch off of master will avoid next time.

sharwell: I avoid this problem by always using merge commits when merging pull requests. The diff here is just an artifact of a non-standard repository configuration that doesn't work for stacked PRs.

PR #4766, opened 2020-02-02
Branch to test current state of outer loop builds

harishsk:
This is a test branch to test the current state of builds when all test run including the disabled test.

### DIFF
--- a/.vsts-dotnet-ci.yml
+++ b/.vsts-dotnet-ci.yml
@@ -25,7 +25,7 @@ jobs:
         _configuration: Release-netcoreapp3_0
         _config_short: RI
         _includeBenchmarkData: true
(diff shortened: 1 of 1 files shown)


antoniovs1029:
What's the difference between the results of the CI on this branch and the MachineLearning-Full CI, which runs all the tests (including the disabled ones):
https://dev.azure.com/dnceng/public/_build?definitionId=707&_a=summary

??

harishsk:
MachineLearning-Full runs all tests by default. But it is only able to run them with commits merged into a specific branch on the dotnet/machinelearning repo. I don't know of a way to run it against a specific PR. 
If we want to separate our work on test stability, we may need to move into a separate branch on dotnet/machinelearning

PR #4770, opened 2020-02-03
try enable test parallel between assembly

frank-dong-ms:
RT.


frank-dong-ms:
No much difference in test running performance, better idea might be use test collection for parallelism

frank-dong-ms:
close for now, will go back to this later when we need to improve test performance

PR #4774, opened 2020-02-03
fix recommendation Bug

LittleLittleCloud:
add code for handle enum case in CodeGen

### DIFF
--- a/src/Microsoft.ML.CodeGenerator/CodeGenerator/CSharp/CodeGenerator.cs
+++ b/src/Microsoft.ML.CodeGenerator/CodeGenerator/CSharp/CodeGenerator.cs
@@ -329,18 +329,11 @@ namespace Microsoft.ML.CodeGenerator.CSharp
         {
             if (_pipeline == null)
                 throw new ArgumentNullException(nameof(_pipeline));
-            try
-            {
-                var node = _pipeline.Nodes.Where(t => t.NodeType == PipelineNodeType.Trainer).First();
-                ITrainerGenerator generator = TrainerGeneratorFactory.GetInstance(node);
-                var trainerString = generator.GenerateTrainer();
-                var trainerUsings = generator.GenerateUsings();
-                return (trainerString, trainerUsings);
-            }
-            catch (Exception)
-            {
-                return (string.Empty, new string[0]);
-            }
+            var node = _pipeline.Nodes.Where(t => t.NodeType == PipelineNodeType.Trainer).First();
+            ITrainerGenerator generator = TrainerGeneratorFactory.GetInstance(node);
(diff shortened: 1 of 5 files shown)


LittleLittleCloud:
#4579

test/Microsoft.ML.CodeGenerator.Tests/TrainerGeneratorTests.cs
```diff
@@ -295,14 +296,21 @@ public void MatrixFactorizationAdvancedTest()
                 {"MatrixColumnIndexColumnName","userId" },
                 {"MatrixRowIndexColumnName","movieId" },
                 {"LabelColumnName","rating" },
+                {nameof(MatrixFactorizationTrainer.Options.NumberOfIterations), 10 },
+                {nameof(MatrixFactorizationTrainer.Options.LearningRate), 0.01f },
+                {nameof(MatrixFactorizationTrainer.Options.ApproximationRank), 8 },
+                {nameof(MatrixFactorizationTrainer.Options.Lambda), 0.01f },
+                {nameof(MatrixFactorizationTrainer.Options.LossFunction), MatrixFactorizationTrainer.LossFunctionType.SquareLossRegression },
+                {nameof(MatrixFactorizationTrainer.Options.Alpha), 1f },
+                {nameof(MatrixFactorizationTrainer.Options.C), 0.00001f },
```
justinormont: Thanks for adding the MF options to a test.

src/Microsoft.ML.CodeGenerator/CodeGenerator/CSharp/CodeGenerator.cs
```diff
@@ -329,18 +329,11 @@ internal IList<(string, string[])> GenerateTransformsAndUsings(IEnumerable<Pipel
...
-            }
-            catch (Exception)
-            {
-                return (string.Empty, new string[0]);
-            }
+            var node = _pipeline.Nodes.Where(t => t.NodeType == PipelineNodeType.Trainer).First();
+            ITrainerGenerator generator = TrainerGeneratorFactory.GetInstance(node);
+            var trainerString = generator.GenerateTrainer();
+            var trainerUsings = generator.GenerateUsings();
+            return (trainerString, trainerUsings);
```
justinormont: I'm not sure why the try/catch was needed. If no one else knows, we may want drill down in the code, or run a variety of datasets thru the CLI to show that nothing is expected to throw.

The old catch seem to eat the error without a hint of what went wrong.

LittleLittleCloud: Yeah, that's why I removed that part of code

PR #4775, opened 2020-02-04
Add Seed property to MLContext and use as default for data splits

### LINKED ISSUE #4752: Default seed is not propagated from MLContext

najeeb-kazmi:
Fixes #4752 

Addresses leftover feedback from #4764

### DIFF
--- a/src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
+++ b/src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
@@ -355,7 +355,7 @@ namespace Microsoft.ML.Runtime
                 var parameters = method.GetParameters();
                 if (parameters.Length != 2 && parameters.Length != 3)
                     throw Contracts.Except("Method '{0}' has {1} parameters, but must have 2 or 3", method.Name, parameters.Length);
-                if (parameters[0].ParameterType != typeof(IHostEnvironment))
+                if (parameters[0].ParameterType != typeof(IHostEnvironment) && parameters[0].ParameterType != typeof(ISeededEnvironment))
                     throw Contracts.Except("Method '{0}', 1st parameter is {1}, but must be IHostEnvironment", method.Name, parameters[0].ParameterType);
                 InputType = parameters[1].ParameterType;
                 var outputType = method.ReturnType;
--- a/src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
+++ b/src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
@@ -355,7 +355,7 @@ namespace Microsoft.ML.Runtime
                 var parameters = method.GetParameters();
                 if (parameters.Length != 2 && parameters.Length != 3)
                     throw Contracts.Except("Method '{0}' has {1} parameters, but must have 2 or 3", method.Name, parameters.Length);
-                if (parameters[0].ParameterType != typeof(IHostEnvironment) && parameters[0].ParameterType != typeof(ISeededEnvironment))
+                if (parameters[0].ParameterType != typeof(IHostEnvironment))
                     throw Contracts.Except("Method '{0}', 1st parameter is {1}, but must be IHostEnvironment", method.Name, parameters[0].ParameterType);
                 InputType = parameters[1].ParameterType;
                 var outputType = method.ReturnType;
--- a/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
+++ b/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
@@ -66,6 +66,11 @@ namespace Microsoft.ML.Runtime
         /// The catalog of loadable components (<see cref="LoadableClassAttribute"/>) that are available in this host.
         /// </summary>
         ComponentCatalog ComponentCatalog { get; }
+
+        /// <summary>
+        /// The seed property that, if assigned, makes components requiring randomness behave deterministically.
+        /// </summary>
+        int? Seed { get; }
     }
 
     [BestFriend]
--- a/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
+++ b/src/Microsoft.ML.Core/Data/IHostEnvironment.cs
@@ -66,11 +66,6 @@ namespace Microsoft.ML.Runtime
         /// The catalog of loadable components (<see cref="LoadableClassAttribute"/>) that are available in this host.
         /// </summary>
         ComponentCatalog ComponentCatalog { get; }
-
-        /// <summary>
-        /// The seed property that, if assigned, makes components requiring randomness behave deterministically.
-        /// </summary>
-        int? Seed { get; }
     }
 
     [BestFriend]
@@ -85,6 +80,15 @@ namespace Microsoft.ML.Runtime
         /// Flag which indicates host execution has been stopped.
         /// </summary>
         bool IsCanceled { get; }
+    }
+
+    [BestFriend]
+    internal interface ISeededEnvironment : IHostEnvironment
+    {
+        /// <summary>
+        /// The seed property that, if assigned, makes components requiring randomness behave deterministically.
+        /// </summary>
+        int? Seed { get; }
     }
 
     /// <summary>
--- a/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
@@ -17,7 +17,7 @@ namespace Microsoft.ML
     public sealed class DataOperationsCatalog : IInternalCatalog
     {
         IHostEnvironment IInternalCatalog.Environment => _env;
-        private readonly IHostEnvironment _env;
+        private readonly ISeededEnvironment _env;
 
         /// <summary>
         /// A pair of datasets, for the train and test set.
@@ -44,7 +44,7 @@ namespace Microsoft.ML
             }
         }
 
-        internal DataOperationsCatalog(IHostEnvironment env)
+        internal DataOperationsCatalog(ISeededEnvironment env)
         {
             Contracts.AssertValue(env);
             _env = env;
@@ -493,7 +493,7 @@ namespace Microsoft.ML
         /// <summary>
         /// Ensures the provided <paramref name="samplingKeyColumn"/> is valid for <see cref="RangeFilter"/>, hashing it if necessary, or creates a new column <paramref name="samplingKeyColumn"/> is null.
         /// </summary>
-        internal static void EnsureGroupPreservationColumn(IHostEnvironment env, ref IDataView data, ref string samplingKeyColumn, int? seed = null)
+        internal static void EnsureGroupPreservationColumn(ISeededEnvironment env, ref IDataView data, ref string samplingKeyColumn, int? seed = null)
         {
             Contracts.CheckValue(env, nameof(env));
             // We need to handle two cases: if samplingKeyColumn is provided, we use hashJoin to
--- a/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
@@ -496,13 +496,12 @@ namespace Microsoft.ML
         internal static void EnsureGroupPreservationColumn(IHostEnvironment env, ref IDataView data, ref string samplingKeyColumn, int? seed = null)
         {
             Contracts.CheckValue(env, nameof(env));
-            var host = env.Register("rand");
             // We need to handle two cases: if samplingKeyColumn is provided, we use hashJoin to
             // build a single hash of it. If it is not, we generate a random number.
             if (samplingKeyColumn == null)
             {
                 samplingKeyColumn = data.Schema.GetTempColumnName("SamplingKeyColumn");
-                data = new GenerateNumberTransform(env, data, samplingKeyColumn, (uint?)(seed ?? host.Rand.Next()));
+                data = new GenerateNumberTransform(env, data, samplingKeyColumn, (uint?)(seed ?? env.Seed));
             }
             else
             {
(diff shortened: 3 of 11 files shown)


najeeb-kazmi:
>                 Random rand = (seed.HasValue) ? RandomUtils.Create(seed.Value) : RandomUtils.Create(_rand);

Should this just use `_rand` here rather than creating another `TauswortheHybrid` from `_rand`?

---
Refers to: src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs:394 in 9bbccaf. [](commit_id = 9bbccafe5f8701e4e53b7e7758e8c352ea99df99, deletion_comment = False)

najeeb-kazmi:
>                     Random rand = (seed.HasValue) ? RandomUtils.Create(seed.Value) : RandomUtils.Create(_rand);

Should `HostBase` use the same `TauswortheHybrid` as the parent environment or create another one using the parent `_rand` as seed?

---
Refers to: src/Microsoft.ML.Core/Environment/HostEnvironmentBase.cs:136 in 9bbccaf. [](commit_id = 9bbccafe5f8701e4e53b7e7758e8c352ea99df99, deletion_comment = False)

eerhardt:
> The CI builds are giving the following errors:

>> 2020-02-04T02:22:45.6483640Z /__w/1/s/packages/microsoft.dotnet.apicompat/1.0.0-beta.19225.5/build/Microsoft.DotNet.ApiCompat.targets(72,5): error : InterfacesShouldHaveSameMembers : Interface member 'Microsoft.ML.Runtime.IHostEnvironment.Seed' is present in the implementation but not in the contract. [/__w/1/s/src/Microsoft.ML.Core/Microsoft.ML.Core.csproj]

This is because you are adding a member to an interface, which is a breaking change. Interfaces, once shipped, can't be changed.

src/Microsoft.ML.Core/Data/IHostEnvironment.cs
```diff
@@ -66,6 +66,11 @@ public interface IHostEnvironment : IChannelProvider, IProgressChannelProvider
         /// The catalog of loadable components (<see cref="LoadableClassAttribute"/>) that are available in this host.
         /// </summary>
         ComponentCatalog ComponentCatalog { get; }
+
+        /// <summary>
+        /// The seed property that, if assigned, makes components requiring randomness behave deterministically.
+        /// </summary>
+        int? Seed { get; }
```
justinormont: > @eerhardt commented [2020-02-04T16:32:54Z](https://github.com/dotnet/machinelearning/pull/4775#issuecomment-581996250)
> ...
> This is because you are adding a member to an interface, which is a breaking change. Interfaces, once shipped, can't be changed.

No need to expose a getter on the seed value. It's nice to be able to get it back, but the user sets the value and should have the value they gave.

> Should `HostBase` use the same `TauswortheHybrid` as the parent environment or create another one using the parent `_rand` as seed?

I assume `TauswortheHybrid` is not thread safe. So sharing can be an issue. Note: I haven't read the code for `HostBase`, so don't know the background.

src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
```diff
@@ -496,13 +496,12 @@ internal static IEnumerable<TrainTestData> CrossValidationSplit(IHostEnvironment
...
         {
             Contracts.CheckValue(env, nameof(env));
-            var host = env.Register("rand");
             // We need to handle two cases: if samplingKeyColumn is provided, we use hashJoin to
             // build a single hash of it. If it is not, we generate a random number.
             if (samplingKeyColumn == null)
             {
                 samplingKeyColumn = data.Schema.GetTempColumnName("SamplingKeyColumn");
-                data = new GenerateNumberTransform(env, data, samplingKeyColumn, (uint?)(seed ?? host.Rand.Next()));
+                data = new GenerateNumberTransform(env, data, samplingKeyColumn, (uint?)(seed ?? env.Seed));
```
najeeb-kazmi: @justinormont this is where I need `IHostEnvironment.Seed` (this is not public in `MLContext`). Any suggestion on how to get seed without adding it to the interface?

justinormont: Would making it non-public work? Or does that still count as modifying the interface?

@eerhardt, do you see how to let other (internal) ML․NET components access the seed value without modifying the interface?

najeeb-kazmi: No, the interface members must be public. It can be accessed from `MLContext` after casting it to `IHostEnvironment`, but not without the cast.

One thought is to add an interface that inherits from `IHostEnvironment` and adds a seed. Then, we pass this interface through the code path to `EnsureGroupPreservationColumn` instead of `IHostEnvironment`. Wondering if there is a better way to do this.

eerhardt: You can add a new interface. One option is to make that new interface internal. See the usages of

https://github.com/dotnet/machinelearning/blob/de5d48a959a789553c337a7d096e58ef0a2410bb/src/Microsoft.ML.Core/Data/IHostEnvironment.cs#L71-L83

for an example.

najeeb-kazmi: @justinormont @eerhardt does the latest commit look good?

test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
```diff
@@ -37,7 +37,7 @@ public void AutoFitBinaryTest()
         [Fact]
         public void AutoFitMultiTest()
         {
-            var context = new MLContext(1);
+            var context = new MLContext(42);
```
najeeb-kazmi: @justinormont before I changed the seed in `EnsureGroupPreservationColum` to fall back to the `MLContext` seed, the `GenerateNumverTransform` was using its default `seed=42`. With the change, when it defaulted to `seed=1` from `MLContext`, the cross validation microaccuracy for this test fell from ~0.71 to ~0.57. Changing the `MLContext` seed to 42 gets the micro accuracy back to ~0.57. Very curious to see so much delta in micro accuracy based solely on the data splits.

justinormont: `42` is a great default seed :)

The [dataset ](https://github.com/dotnet/machinelearning/blob/553e089259b228c29403fd71bd006d7fdef3e0db/test/Microsoft.ML.AutoML.Tests/TestData/TrivialMulticlassDataset.txt) being 180 rows, I would expect swings in a model accuracy. Though I would expect our use of 10-fold cross-validation to keep the value pretty stable, even for small datasets.

Can you check that cross-validation still gets independent folds (and doesn't keep recreating the exact same fold due to the fixed seed)? We may need to modify the CV fold creation, perhaps by using `seed = seed + foldNum`.

justinormont: > Changing the `MLContext` seed to 42 gets the micro accuracy back to ~0.71. 

If changing the seed back to 42 fixes the accuracy, it's likely not the folds being recreated; though still useful to check. 

Perhaps trying a few seed values will tell us the normal range for the model accuracy. 

Side note: Technically the AutoML returns the metrics of the fold [closest to the average](https://github.com/dotnet/machinelearning/blob/e50c4d20012e0d62852f404ae443afca7dad043e/src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs#L75-L78) for CV instead of the average, which is a bug we should fix. Since AutoML will automatically choose CV or TrainTest and can have only one return type, and AutoML couldn't at the time create a metric with its own values, CV also returned the singular metric class instead of the CV metric stats class. Should switch to always return `MetricStatistics` instance.

src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
```diff
@@ -518,7 +517,7 @@ internal static void EnsureGroupPreservationColumn(IHostEnvironment env, ref IDa
                     // instead of having two hash transformations.
                     var origStratCol = samplingKeyColumn;
                     samplingKeyColumn = data.Schema.GetTempColumnName(samplingKeyColumn);
-                    var columnOptions = new HashingEstimator.ColumnOptionsInternal(samplingKeyColumn, origStratCol, 30, (uint)(seed ?? host.Rand.Next()));
+                    var columnOptions = new HashingEstimator.ColumnOptionsInternal(samplingKeyColumn, origStratCol, 30, (uint)(seed ?? env.Seed));
```
yaeldekel: >env.Seed [](start = 135, length = 8)

This can also be null, can't it?

src/Microsoft.ML.Core/ComponentModel/ComponentCatalog.cs
```diff
@@ -355,7 +355,7 @@ internal sealed class EntryPointInfo
                 var parameters = method.GetParameters();
                 if (parameters.Length != 2 && parameters.Length != 3)
                     throw Contracts.Except("Method '{0}' has {1} parameters, but must have 2 or 3", method.Name, parameters.Length);
-                if (parameters[0].ParameterType != typeof(IHostEnvironment))
+                if (parameters[0].ParameterType != typeof(IHostEnvironment) && parameters[0].ParameterType != typeof(ISeededEnvironment))
```
eerhardt: I wouldn't do it this way.

Instead, I would keep `DataOperationsCatalog` taking an `IHostEnvironment` as a parameter, and casting it down to `ISeededEnvironment` when you need to grab the `Seed` property.

src/Microsoft.ML.Data/MLContext.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.ML
     /// create components for data preparation, feature enginering, training, prediction, model evaluation.
     /// It also allows logging, execution control, and the ability set repeatable random numbers.
     /// </summary>
-    public sealed class MLContext : IHostEnvironment
+    public sealed class MLContext : ISeededEnvironment, IHostEnvironment
```
yaeldekel: >IHostEnvironment [](start = 56, length = 16)

`ISeededEnvironment` is an `IHostEnvironment`.

src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
justinormont: @eerhardt would be the expert here. I would bow to his opinion, though looks good to me.

From @eerhardt's `ICancelable` example I was envisioning a multi-inheritance of having `IHostEnvironment` inherit from a new `ISeeded`, though this may count as modifying `IHostEnvironment`; hence you created a new interface to inherit from `IHostEnvironment`.

New interface:
https://github.com/dotnet/machinelearning/blob/9fd29aca95da8f4359c1ef551ce4f20050327da6/src/Microsoft.ML.Core/Data/IHostEnvironment.cs#L85-L92

eerhardt: > I was envisioning a multi-inheritance of having IHostEnvironment inherit from a new ISeeded, though this may count as modifying IHostEnvironment;

Correct - that would count as modifying IHostEnvironment, which is a breaking change.

src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
najeeb-kazmi: Yes, if `env.Seed` is also null, it will fall back to the `HashingEstimator` default seed:
https://github.com/dotnet/machinelearning/blob/46e7dc6269ef4a55c0e5788f0c327ef2029ee2f2/src/Microsoft.ML.Data/Transforms/Hashing.cs#L1129-L1132

test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
najeeb-kazmi: Yes, the folds are independent. Changing the seed only affects how the `SamplingKeyColumn` in the `IDataView data` coming into this method is randomized. The folds are then created in the same way as before here:
https://github.com/dotnet/machinelearning/blob/46e7dc6269ef4a55c0e5788f0c327ef2029ee2f2/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs#L465-L473

justinormont: Thanks @najeeb-kazmi for looking, and you're completely correct.

src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
yaeldekel: But you are casting it to `uint` here. If it's null it will throw an exception.

---
In reply to: [375521058](https://github.com/dotnet/machinelearning/pull/4775#discussion_r375521058) [](ancestors = 375521058)

najeeb-kazmi: Ah yes, you're right. Thanks for catching that @yaeldekel. I've made it nullable like in `GenerateNumberTransform` above this.

PR #4776, opened 2020-02-04
build and test failure due to download resource fail

frank-dong-ms:
facing build and test failure due to download resource fail:
https://microsoft.sharepoint.com/teams/ML.NET/_layouts/OneNote.aspx?id=%2Fteams%2FML.NET%2FSiteAssets%2FML.NET%20Notebook&wd=target%28Tests.one%7C5E713C2E-6DCF-4AA3-9071-69CBBAF48985%2FTest%20Network%20Issue%7CDBE36B97-0518-43C2-8C3F-A7D2A461FBDE%2F%29

this fix including:
1. add retry and delay for resource downloading when run test
2. disable parallel when restore nuget package, this can prevent restore consume too much connections from vm and cause the following download to hanging, restore itself has retry already
3. add retry for download resource from Internet when build, if disable parallel when restore can prevent the network glitch this can possibility be unnecessary


build.proj
```diff
@@ -43,7 +43,7 @@
     <Message Importance="High" Text="Restoring all projects..." />
     <MSBuild Projects="@(Project)"
              Targets="Restore" 
-             Properties="MSBuildWarningsAsMessages=NU1503" />
+             Properties="MSBuildWarningsAsMessages=NU1503;RestoreDisableParallel=true" />
```
harishsk: Are we seeing connection issues due to parallel connections on all platforms or only on non-windows platforms? Do we want to consider disabling this only on specific platforms?

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -98,7 +98,7 @@ public static string GetUrl(string suffix)
         /// <param name="timeout">An integer indicating the number of milliseconds to wait before timing out while downloading a resource.</param>
         /// <returns>The download results, containing the file path where the resources was (or should have been) downloaded to, and an error message
         /// (or null if there was no error).</returns>
-        public async Task<ResourceDownloadResults> EnsureResource(IHostEnvironment env, IChannel ch, string relativeUrl, string fileName, string dir, int timeout)
```
harishsk: Is there a particular reason to no change the interface from returning a Task to making a synchronous call?

frank-dong-ms: change to synchronous call to retry on download, we need to wait for the last result of download (success or fail) then determine whether to return result or continue retry download.
this is not public API we only use this method as internal and not intended to let our user use this

---
In reply to: [375020245](https://github.com/dotnet/machinelearning/pull/4776#discussion_r375020245) [](ancestors = 375020245)

build.proj
frank-dong-ms: this issue is observed on windows as well and disable parallel has minor impact on build performance

so I want to change this to see if the download issue is indeed on insufficient connections

---
In reply to: [375017713](https://github.com/dotnet/machinelearning/pull/4776#discussion_r375017713) [](ancestors = 375017713)

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
harishsk: Even if you have to wait for the last download to complete and examine the status before deciding whether to continue retrying, you should still use the async/await pattern. If you don't you might block the calling thread.

---
In reply to: [375064673](https://github.com/dotnet/machinelearning/pull/4776#discussion_r375064673) [](ancestors = 375064673,375020245)

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -110,7 +110,28 @@ public async Task<ResourceDownloadResults> EnsureResource(IHostEnvironment env,
...
+            int timeout, string filePath, int retryTimes = 5)
+        {
+            var downloadResult = "";
+
+            for (int i = 0; i < retryTimes; ++i)
+            {
+                var thisDownloadResult = DownloadFromUrl(env, ch, url, fileName, timeout, filePath).Result;
+
+                if (string.IsNullOrEmpty(thisDownloadResult))
+                    return thisDownloadResult;
```
harishsk: You should be using an await here. I am not sure this will work reliably.

build.proj
sharwell: > this issue is observed on windows as well

The issue on Windows is not likely to be caused by parallel connections. I'm concerned that this change confuses the issue, and leaves the repository in an inferior state for unclear reasons.

/cc @jaredpar

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -110,7 +110,28 @@ public async Task<ResourceDownloadResults> EnsureResource(IHostEnvironment env,
...
+            {
+                var thisDownloadResult = DownloadFromUrl(env, ch, url, fileName, timeout, filePath);
+
+                if (string.IsNullOrEmpty(await thisDownloadResult))
+                    return await thisDownloadResult;
+                else
+                    downloadResult += thisDownloadResult.Result + @"\n";
+
+                Random random = new Random();
+                Thread.Sleep(random.Next(1, 60) * 1000);
```
sharwell: ❗️ Don't call `Thread.Sleep` from an asynchronous method. Use `await Task.Delay` instead.

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -110,7 +110,28 @@ public async Task<ResourceDownloadResults> EnsureResource(IHostEnvironment env,
                     $"Could not create a valid URI from the base URI '{MlNetResourcesUrl}' and the relative URI '{relativeUrl}'");
             }
             return new ResourceDownloadResults(filePath,
-                await DownloadFromUrl(env, ch, absoluteUrl.AbsoluteUri, fileName, timeout, filePath), absoluteUrl.AbsoluteUri);
+                await DownloadFromUrlWithRetry(env, ch, absoluteUrl.AbsoluteUri, fileName, timeout, filePath), absoluteUrl.AbsoluteUri);
+        }
+
+        private async Task<string> DownloadFromUrlWithRetry(IHostEnvironment env, IChannel ch, string url, string fileName,
```
sharwell:
```suggestion
        private async Task<string> DownloadFromUrlWithRetryAsync(IHostEnvironment env, IChannel ch, string url, string fileName,
```

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
sharwell: 💭 This is an extremely unusual delay strategy. I would expect constant, progressive backoff, or exponential backoff.

build.proj
frank-dong-ms: I'm not sure about whether parallel connections cause issue on windows and best way to verify is to merge this to master and observe the result, we can't tell the difference from PR build as this issue occurs quite rare.
And disable parallelism has minor impact to build performance so I would like say we give it a try.

---
In reply to: [375543129](https://github.com/dotnet/machinelearning/pull/4776#discussion_r375543129) [](ancestors = 375543129)

harishsk: @frank-dong-ms: @sharwell has a valid concern. But I see your point that we should at least give it a try. Can you please create a P1 bug to track this issue to try and revert it back for Windows at a later date if this PR fixes the other issues?   


---
In reply to: [375566985](https://github.com/dotnet/machinelearning/pull/4776#discussion_r375566985) [](ancestors = 375566985,375543129)

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -110,7 +110,27 @@ public async Task<ResourceDownloadResults> EnsureResource(IHostEnvironment env,
...
+        {
+            var downloadResult = "";
+
+            for (int i = 0; i < retryTimes; ++i)
+            {
+                var thisDownloadResult = DownloadFromUrl(env, ch, url, fileName, timeout, filePath);
+
+                if (string.IsNullOrEmpty(await thisDownloadResult))
+                    return await thisDownloadResult;
+                else
```
harishsk: There shouldnt be a need to await twice on the same async task. I think it would be cleaner if there was a single await call and then use the cached results.

frank-dong-ms: yes, thanks, fixed

---
In reply to: [375625182](https://github.com/dotnet/machinelearning/pull/4776#discussion_r375625182) [](ancestors = 375625182)

build.proj
frank-dong-ms: Sure, issue below created:
[https://github.com/dotnet/machinelearning/issues/4788](https://github.com/dotnet/machinelearning/issues/4788) 

---
In reply to: [375622405](https://github.com/dotnet/machinelearning/pull/4776#discussion_r375622405) [](ancestors = 375622405,375566985,375543129)

frank-dong-ms:
yes, done

---
In reply to: [354197547](https://github.com/dotnet/machinelearning/pull/4776#issuecomment-354197547) [](ancestors = 354197547)

PR #4781, opened 2020-02-04
TextNormalizing export to Onnx 

Lynx1820:
- can, at the moment, only handle case changing options

### DIFF
--- a/build/BranchInfo.props
+++ b/build/BranchInfo.props
@@ -31,12 +31,12 @@
     <MajorVersion>1</MajorVersion>
     <MinorVersion>5</MinorVersion>
     <PatchVersion>0</PatchVersion>
-    <PreReleaseLabel>preview5</PreReleaseLabel>
+    <PreReleaseLabel>preview2</PreReleaseLabel>
   </PropertyGroup>
   <PropertyGroup Condition="'$(IsStableProject)' != 'true'">
     <MajorVersion>0</MajorVersion>
     <MinorVersion>17</MinorVersion>
     <PatchVersion>0</PatchVersion>
-    <PreReleaseLabel>preview5</PreReleaseLabel>
+    <PreReleaseLabel>preview2</PreReleaseLabel>
   </PropertyGroup>
 </Project>
--- a/build/BranchInfo.props
+++ b/build/BranchInfo.props
@@ -31,12 +31,12 @@
     <MajorVersion>1</MajorVersion>
     <MinorVersion>5</MinorVersion>
     <PatchVersion>0</PatchVersion>
-    <PreReleaseLabel>preview2</PreReleaseLabel>
+    <PreReleaseLabel>preview5</PreReleaseLabel>
   </PropertyGroup>
   <PropertyGroup Condition="'$(IsStableProject)' != 'true'">
     <MajorVersion>0</MajorVersion>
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.Transforms/Text/TextNormalizing.cs
```diff
@@ -212,6 +213,31 @@ public Mapper(TextNormalizingTransformer parent, DataViewSchema inputSchema)
...
+
+                    string outputColumnName = _parent.ColumnPairs[iinfo].outputColumnName;
+                    string srcVariableName = ctx.GetVariableName(inputColumnName);
+                    string dstVariableName = ctx.AddIntermediateVariable(_types[iinfo], outputColumnName, true);
+                    SaveAsOnnxCore(ctx, srcVariableName, dstVariableName);
+                }
+            }
+
+            private void SaveAsOnnxCore(OnnxContext ctx, string srcVariableName, string dstVariableName)
+            {
```
harishsk: The normalizer has two other options for keepDiacritics and keepPunctuations. What should happen if either of these options are specified in the pipeline? As things stand, the output of ML.NET and ORT will be different.

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -448,6 +448,41 @@ public void PlattCalibratorOnnxConversionTest2()
...
+        public void TextNormalizingOnnxConversionTest()
+        {
+            var mlContext = new MLContext(seed: 1);
+            var dataPath = GetDataPath("wikipedia-detox-250-line-test.tsv");
+            var dataView = ML.Data.LoadFromTextFile(dataPath, new[] {
+                new TextLoader.Column("label", DataKind.Boolean, 0),
+                new TextLoader.Column("text", DataKind.String, 1)
+            }, hasHeader: true);
+            var pipeline = new TextNormalizingEstimator(mlContext, keepDiacritics: true, columns: new[] { ("NormText", "text") }).Append(
+                new TextNormalizingEstimator(mlContext, keepDiacritics: true, caseMode: TextNormalizingEstimator.CaseMode.Upper, columns: new[] { ("UpperText", "text") })).Append(
```
harishsk: Adding keepDiacritics only to the test code won't fix the issue. What happens if somebody creates a pipeline with a different option? I think this has to be handled in the SaveAsOnnx function for TextNormalizing

Lynx1820: I also qualified CanSaveOnnx(OnnxContext ctx) to return false for the other options. Is that not the expected behavior?

harishsk: Sorry, I missed that line. That sounds good.

build/BranchInfo.props
```diff
@@ -31,12 +31,12 @@
     <MajorVersion>1</MajorVersion>
     <MinorVersion>5</MinorVersion>
     <PatchVersion>0</PatchVersion>
-    <PreReleaseLabel>preview2</PreReleaseLabel>
+    <PreReleaseLabel>preview5</PreReleaseLabel>
```
harishsk: Please revert this change.

PR #4782, opened 2020-02-05
fix some Baselinetest and add missing Done()

frank-dong-ms:
1. add missing baseline file for compare
2. use allow mismatch in certain tests that only intend compare value and return result but not fail test

### DIFF
--- a/test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
+++ b/test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
@@ -523,7 +523,7 @@ namespace Microsoft.ML.RunTests
                     float f1 = float.Parse(firstCollection[i].ToString());
                     float f2 = float.Parse(secondCollection[i].ToString());
 
-                    if (!CompareNumbersAndLogErrors(f1, f2, i, digitsOfPrecision))
+                    if (!CompareNumbersWithTolerance(f1, f2, i, digitsOfPrecision))
                     {
                         return false;
                     }
@@ -536,7 +536,7 @@ namespace Microsoft.ML.RunTests
                     double f1 = double.Parse(firstCollection[i].ToString());
                     double f2 = double.Parse(secondCollection[i].ToString());
 
-                    if (!CompareNumbersAndLogErrors(f1, f2, i, digitsOfPrecision))
+                    if (!CompareNumbersWithTolerance(f1, f2, i, digitsOfPrecision))
                     {
                         return false;
                     }
@@ -550,22 +550,11 @@ namespace Microsoft.ML.RunTests
             return true;
         }
 
-        public bool CompareNumbersAndLogErrors(double expected, double actual,
-            int? iterationOnCollection = null, int digitsOfPrecision = DigitsOfPrecision)
-        {
-            var compareResult = CompareNumbers(expected, actual, iterationOnCollection, digitsOfPrecision);
-
-            if (!compareResult.Item1)
-                Fail(compareResult.Item2);
-
-            return compareResult.Item1;
-        }
-
-        public (bool, string) CompareNumbers(double expected, double actual, 
-            int? iterationOnCollection = null, int digitsOfPrecision = DigitsOfPrecision, bool allowMismatch = false)
+        public bool CompareNumbersWithTolerance(double expected, double actual, int? iterationOnCollection = null, 
+            int digitsOfPrecision = DigitsOfPrecision, bool logFailure = true)
         {
             if (double.IsNaN(expected) && double.IsNaN(actual))
-                return (true, "");
+                return true;
 
             // this follows the IEEE recommendations for how to compare floating point numbers
             double allowedVariance = Math.Pow(10, -digitsOfPrecision);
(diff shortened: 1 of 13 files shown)


test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -193,6 +193,9 @@ protected void Fail(bool relax, string fmt, params object[] args)
             {
                 _failures.Add(ex);
             }
+            catch when (relax)
```
harishsk: Why would we not want to report the callstack when relax is true? (More importantly what is the point of relax at all?)

test/Microsoft.ML.Tests/TrainerEstimators/SdcaTests.cs
```diff
@@ -140,13 +141,15 @@ public void SdcaLogisticRegressionWithWeight()
             bool sameScores = true;
             for (int i = 0; i < scores1.Length; i++)
             {
-                if(!CompareNumbersWithTolerance(scores1[i], scores2[i]))
+                if(!CompareNumbersWithTolerance(scores1[i], scores2[i], allowMismatch: true))
                 {
```
harishsk: Do we need allowMismatch or will specifying the digitsOfPrecision work?

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
frank-dong-ms: For these case, we only want to do deep compare of 2 score values and intend these 2 scores to be different.
when set this relax to true (from outside, it is allowMismatch), we want to compare values but not to fail the test (that is to say, after add Done() check).

---
In reply to: [375022748](https://github.com/dotnet/machinelearning/pull/4782#discussion_r375022748) [](ancestors = 375022748)

test/Microsoft.ML.Tests/TrainerEstimators/SdcaTests.cs
frank-dong-ms: We need this allowMismatch, these test are intend to get the conclusion of these 2 scores are different.

---
In reply to: [375022988](https://github.com/dotnet/machinelearning/pull/4782#discussion_r375022988) [](ancestors = 375022988)

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
harishsk: Fail with a relax parameter seems really odd. This whole thing should be refactored differently.  (More notes below near the CompareNumbersWithTolerance function)

---
In reply to: [375064007](https://github.com/dotnet/machinelearning/pull/4782#discussion_r375064007) [](ancestors = 375064007,375022748)

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -473,7 +446,7 @@ protected bool CheckOutputIsSuffix(string basePath, string outPath, int skip = 0
         }
 
         protected bool CheckEqualityFromPathsCore(string relPath, string basePath, string outPath, int skip = 0,
-            int digitsOfPrecision = DigitsOfPrecision, NumberParseOption parseOption = NumberParseOption.Default)
+            int digitsOfPrecision = DigitsOfPrecision, NumberParseOption parseOption = NumberParseOption.Default, bool allowMismatch = false)
```
harishsk: This function is named CheckEqualityFromPathsCore. This function shouldn't be called by functions that need to check for unequal values. So this function shouldn't take an allowMismatch parameter.

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -527,7 +500,7 @@ protected bool CheckOutputIsSuffix(string basePath, string outPath, int skip = 0
                         if (line1 == null || line2 == null)
                             Fail("Output and baseline different lengths: '{0}'", relPath);
                         else
-                            Fail(_allowMismatch, "Output and baseline mismatch at line {1}, expected '{2}' but got '{3}' : '{0}'", relPath, count, line1, line2);
```
harishsk: If we are checking for equality, we should always call out mismatches. Especially in this function when we are doing file comparison.

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -590,7 +563,8 @@ protected bool CheckOutputIsSuffix(string basePath, string outPath, int skip = 0
             return true;
         }
 
-        public bool CompareNumbersWithTolerance(double expected, double actual, int? iterationOnCollection = null, int digitsOfPrecision = DigitsOfPrecision)
```
harishsk: This function is doing two jobs: 1) Comparison and 2) Calling fail if the comparison fails.  

I agree with your point above that there are cases where we want to compare unequal values and make sure they are unequal. Therefore this function should be split into two different functions. One for only doing comparison, and a second one that does the comparison and calls fails depending on whether the comparison was intended to succeed or fail.

test/Microsoft.ML.Tests/TextLoaderTests.cs
```diff
@@ -81,6 +81,8 @@ public void TestTextLoaderDataTypes()
 
                 Assert.Equal(i, sByteTargets.Length);
             }
+
+            Done();
```
sharwell: 💭 I thought we were going to move this call to the base test class?

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
sharwell: Yes, super strange

test/Microsoft.ML.Tests/TextLoaderTests.cs
frank-dong-ms: Yes, ideally but that should be other PR, this PR is intended to fix these few tests and make proper assertion

---
In reply to: [375545013](https://github.com/dotnet/machinelearning/pull/4782#discussion_r375545013) [](ancestors = 375545013)

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -590,10 +550,22 @@ protected bool CheckOutputIsSuffix(string basePath, string outPath, int skip = 0
...
+
+            if (!compareResult.Item1)
+                Fail(compareResult.Item2);
+
+            return compareResult.Item1;
+        }
+
+        public (bool, string) CompareNumbers(double expected, double actual, 
+            int? iterationOnCollection = null, int digitsOfPrecision = DigitsOfPrecision, bool allowMismatch = false)
         {
```
harishsk: Sorry for changing my mind on this. As I read through your current implementation I see that it is resulting in a lot of changes. I am now thinking it may be better to add a default parameter instead. Instead of the previous allowMismatch param you added, would it be better to add a logFailure parameter that defaults to true and retain the same function name? That will eliminate the changes in the rest of the code and the only changes should be in places where we are comparing unequal numbers.

PR #4783, opened 2020-02-05
Onnx Export change to allow for multiple rows

michaelgsharp:
Changed ONNX export from always exporting the dimensions as 1 to -1. This lets ONNX Runtime determine the dimension when the data is passed to it, allowing for batching to be done if desired. ML.NET doesn't support batching, but this allows the model to be run directly in ORT using batching while still supporting the streaming approach that ML.NET uses.

### DIFF
--- a/src/Microsoft.ML.OnnxConverter/OnnxUtils.cs
+++ b/src/Microsoft.ML.OnnxConverter/OnnxUtils.cs
@@ -375,7 +375,12 @@ namespace Microsoft.ML.Model.OnnxConverter
                         dimsLocal.Add(vec.Dimensions[i]);
                 }
             }
-            //batch size.
+            // Set batch size to -1. The ONNX docs, https://github.com/onnx/onnx/blob/master/docs/IR.md#static-tensor-shapes, state that if
+            // dim_param is used instead of dim_value, that the size of the dimension "is not statically constrained to a particular number"
+            // "This is useful for declaring the interfaces that care about the number of dimensions, but not the exact size of each dimension"
+            // This file, https://github.com/onnx/onnx/blob/master/onnx/tools/update_model_dims.py, explains that if the dim value is negative
+            // than it treats that as a dim_param instead of a dim_value. This allows ML.NET to run 1 row at a time in a streaming fassion,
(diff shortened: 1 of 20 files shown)


src/Microsoft.ML.OnnxConverter/OnnxUtils.cs
```diff
@@ -376,7 +376,7 @@ public ModelArgs(string name, TensorProto.Types.DataType dataType, List<long> di
                 }
             }
             //batch size.
-            dimsLocal?.Insert(0, 1);
```
harishsk: Have you tested that this works? That is, can you take a model exported from ML.NET with this fix and run it with batch size > 1 outside of ML.NET?

harishsk: We should also explore whether ML.NET can run the models with batch sizes > 1

---
In reply to: [375025535](https://github.com/dotnet/machinelearning/pull/4783#discussion_r375025535) [](ancestors = 375025535)

src/Microsoft.ML.OnnxConverter/OnnxUtils.cs
```diff
@@ -376,7 +376,7 @@ public ModelArgs(string name, TensorProto.Types.DataType dataType, List<long> di
...
+            dimsLocal?.Insert(0, -1);
```
ganik: >dimsLocal?.Insert(0, -1); [](start = 12, length = 25)

PLs add comment describing this value -1 with reference to original ONNX doc on web

src/Microsoft.ML.OnnxConverter/OnnxUtils.cs
michaelgsharp: As discussed offline, I have taken an ML.NET model and ran it using the onnxruntime python bindings with a batch of 150 rows and streaming it 1 row at a time. Both work as expected.

src/Microsoft.ML.OnnxConverter/OnnxUtils.cs
michaelgsharp: added more comments.

PR #4791, opened 2020-02-06
Enable VSTHRD100 (Avoid async void methods)

sharwell:


src/Microsoft.Extensions.ML/ModelLoaders/UriModelLoader.cs
```diff
@@ -53,13 +53,16 @@ internal void Start(Uri uri, TimeSpan period)
             _started = true;
         }
 
-        private async void ReloadTimerTick(object state)
+        private void ReloadTimerTick(object state)
         {
-            StopReloadTimer();
+            _ = Task.Run(async () =>
```
eerhardt: I'm concerned about the behavior change here. As I can tell this is doing 2 things that I don't feel are correct:

1. We are already on a background thread, and here we are jumping to another background thread for seemingly no benefit.
2. Previously, if an exception was thrown in this method, it would bubble up to the calling synchronization context, which is the thread pool since this is a Timer.Tick handler. Previously, an unhandled exception would crash the process. With this change, unhandled exceptions are being ignored.

I think we should change this back to `async void` and suppress the rule here. `async void` event handlers are reasonable things to do.

PR #4792, opened 2020-02-06
Enable VSTHRD103 (Call async methods when in an async method)

sharwell:


sharwell:
@harishsk @frank-dong-ms for reviews

PR #4794, opened 2020-02-06
Enable VSTHRD200 (Use "Async" suffix for async methods)

sharwell:
This rule relies on a suppression analyzer (#4803) to avoid requiring test methods to be renamed.

### DIFF
--- a/src/Microsoft.ML.Sweeper/AsyncSweeper.cs
+++ b/src/Microsoft.ML.Sweeper/AsyncSweeper.cs
@@ -47,7 +47,7 @@ namespace Microsoft.ML.Sweeper
         /// Propose a <see cref="ParameterSet"/>.
         /// </summary>
         /// <returns>A future <see cref="ParameterSet"/> and its id. Null if unavailable or cancelled.</returns>
-        Task<ParameterSetWithId> Propose();
+        Task<ParameterSetWithId> ProposeAsync();
 
         /// <summary>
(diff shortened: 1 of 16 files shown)


src/Microsoft.ML.Sweeper/AsyncSweeper.cs
```diff
@@ -47,7 +47,7 @@ public interface IAsyncSweeper
         /// Propose a <see cref="ParameterSet"/>.
         /// </summary>
         /// <returns>A future <see cref="ParameterSet"/> and its id. Null if unavailable or cancelled.</returns>
-        Task<ParameterSetWithId> Propose();
+        Task<ParameterSetWithId> ProposeAsync();
```
sharwell: 📝 This is the only public API change in this pull request. It's not clear to me if/when this API shipped or whether it's allowed to be renamed at this point. If we can't rename it, I'll revert the change to this file and use a suppression instead.

sharwell:
@harishsk This is now ready for review. Please make specific note of the indicated public API change.

src/Microsoft.ML.Sweeper/AsyncSweeper.cs
eerhardt: `Microsoft.ML.Sweeper` hasn't shipped in any nuget package, so I think we are safe to rename this.

harishsk:
The changes have been approved. Can you please rebase and merge?

PR #4805, opened 2020-02-06
TokenizingByCharacters export to Onnx

Lynx1820:
- Transformer that tokenizes by character and returns the characters (as uint16)
- Since there's not a comparable onnx operator, a label encoder is used to map a string token to it's corresponding character value.  This will unfortunately make the model much larger, since 65535 values have to be saved as a mapping guide for label encoder.

### DIFF
--- a/test/Microsoft.ML.Tests/OnnxConversionTest.cs
+++ b/test/Microsoft.ML.Tests/OnnxConversionTest.cs
@@ -934,7 +934,6 @@ namespace Microsoft.ML.Tests
             var model = pipeline.Fit(dataView);
             var transformedData = model.Transform(dataView);
             var onnxModel = mlContext.Model.ConvertToOnnxProtobuf(model, dataView);
-            //CompareSelectedVectorColumns<UInt16>(transformedData.Schema[2].Name, transformedData.Schema[2].Name, transformedData, transformedData);
             // Compare model scores produced by ML.NET and ONNX's runtime. 
             if (IsOnnxRuntimeSupported())
             {
@@ -948,7 +947,6 @@ namespace Microsoft.ML.Tests
                 var onnxTransformer = onnxEstimator.Fit(dataView);
                 var onnxResult = onnxTransformer.Transform(dataView);
                 CompareSelectedVectorColumns<UInt16>(transformedData.Schema[2].Name, outputNames[2], transformedData, onnxResult); //compare scores
-
             }
             Done();
         }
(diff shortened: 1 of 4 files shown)


harishsk:
>     /// | Exportable to ONNX | No |

Please change this line whenever you add new support

---
Refers to: src/Microsoft.ML.Transforms/Text/TokenizingByCharacters.cs:610 in 0532330. [](commit_id = 05323306729624a34cc53579eac0ee6a993017f7, deletion_comment = False)

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -921,6 +921,36 @@ public void WordEmbeddingsTest()
...
+        public void TokenizingByCharactersOnnxConversionTest()
+        {
+            var mlContext = new MLContext(seed: 1);
+            var dataPath = GetDataPath("wikipedia-detox-250-line-test.tsv");
+            var dataView = ML.Data.LoadFromTextFile(dataPath, new[] {
+                new TextLoader.Column("label", DataKind.Boolean, 0),
+                new TextLoader.Column("text", DataKind.String, 1)
+            }, hasHeader: true);
+            var pipeline = new TokenizingByCharactersEstimator(mlContext, columns: new[] { ("TokenizedText", "text") });
+            var model = pipeline.Fit(dataView);
```
harishsk: Can you please add a test with UseMarkerChars?

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -921,6 +921,36 @@ public void WordEmbeddingsTest()
             Done();
         }
 
+        [Fact]
+        public void TokenizingByCharactersOnnxConversionTest()
+        {
+            var mlContext = new MLContext(seed: 1);
+            var dataPath = GetDataPath("wikipedia-detox-250-line-test.tsv");
+            var dataView = ML.Data.LoadFromTextFile(dataPath, new[] {
```
harishsk: Are there no issues with utf-8 encoding here?

Lynx1820: yes, the issues happen with the wikipedia-detox-250-line-data.tsv file

harishsk: Does the test run successfully? If not is there a different file you can use? Or just used fixed inline data.

---
In reply to: [376146626](https://github.com/dotnet/machinelearning/pull/4805#discussion_r376146626) [](ancestors = 376146626)

Lynx1820: Yes, I'm using wikipedia-detox-250-line-**test**.tsv instead

harishsk: Ah..ok. Sounds good.

PR #4810, opened 2020-02-07
modify base test class analyzer, allow to extend from ft base test class

frank-dong-ms:
We have separate base test class for all functional tests: FunctionalTestBaseClass
The reason for that is FT is also runned on NightlyBuild pipeline which not take project dependency directly but take ML.NET related Nuget package as dependency and we don't want FT to take dependency from TestFramework.

### DIFF
--- a/tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
+++ b/tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
@@ -16,10 +16,10 @@ namespace Microsoft.ML.InternalCodeAnalyzer
         private const string Category = "Test";
         internal const string DiagnosticId = "MSML_ExtendBaseTestClass";
 
-        private const string Title = "Test classes should be derived from BaseTestClass";
-        private const string Format = "Test class '{0}' should extend BaseTestClass.";
+        private const string Title = "Test classes should be derived from BaseTestClass or FunctionalTestBaseClass";
+        private const string Format = "Test class '{0}' should extend BaseTestClass or FunctionalTestBaseClass.";
         private const string Description =
-            "Test classes should be derived from BaseTestClass.";
+            "Test classes should be derived from BaseTestClass or FunctionalTestBaseClass.";
 
         private static DiagnosticDescriptor Rule =
             new DiagnosticDescriptor(DiagnosticId, Title, Format, Category,
@@ -51,6 +51,7 @@ namespace Microsoft.ML.InternalCodeAnalyzer
             private readonly Compilation _compilation;
             private readonly INamedTypeSymbol _factAttribute;
             private readonly INamedTypeSymbol _baseTestClass;
+            private readonly INamedTypeSymbol _FTbaseTestClass;
             private readonly ConcurrentDictionary<INamedTypeSymbol, bool> _knownTestAttributes = new ConcurrentDictionary<INamedTypeSymbol, bool>();
 
             public AnalyzerImpl(Compilation compilation, INamedTypeSymbol factAttribute)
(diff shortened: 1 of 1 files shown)


tools-local/Microsoft.ML.InternalCodeAnalyzer/BaseTestClassAnalyzer.cs
```diff
@@ -51,13 +51,15 @@ private sealed class AnalyzerImpl
...
             private readonly INamedTypeSymbol _baseTestClass;
+            private readonly INamedTypeSymbol _FTbaseTestClass;
             private readonly ConcurrentDictionary<INamedTypeSymbol, bool> _knownTestAttributes = new ConcurrentDictionary<INamedTypeSymbol, bool>();
 
             public AnalyzerImpl(Compilation compilation, INamedTypeSymbol factAttribute)
             {
                 _compilation = compilation;
                 _factAttribute = factAttribute;
                 _baseTestClass = _compilation.GetTypeByMetadataName("Microsoft.ML.TestFramework.BaseTestClass");
+                _FTbaseTestClass = _compilation.GetTypeByMetadataName("Microsoft.ML.Functional.Tests.FunctionalTestBaseClass");
```
harishsk: Why do we have two base test classes? I know you have investigated this before. Can you please help me understand why we can't have a single BaseTestClass and have FunctionalTestBaseClass derive from BaseTestClass? (From what I understand, the only difference for functional tests is that they only use public API)

frank-dong-ms: Yes, please reference previous PR for this: [https://github.com/dotnet/machinelearning/pull/4346](https://github.com/dotnet/machinelearning/pull/4346)

Basically, we want FT to be able to dependent on public nuget package so we need to break FT from any direct or indirect project reference. TestFramework project itself has lots of project dependency and also been marked as BestFriend from some projects from source folder which allows TestFramework peoject to access some non public code. So, we extracted TestFrameworkCommon test project which only have public dependency. Previously FTs are inherited from BaseTestBaseline which has lots of BestFriend thing and can't be moved to TestFrameworkCommon project so after discuss we decide to write one base class for FT only with only public dependency.
So in short we have separate base class for FT because we don't want to take project dependency from our source code to let FT able to run in NightlyBuild

---
In reply to: [376581212](https://github.com/dotnet/machinelearning/pull/4810#discussion_r376581212) [](ancestors = 376581212)

harishsk: You are right! That sounds good. In that case, please make sure the functionality between the two base classes are in sync.  I recently added some code to extract the LogMessageKind attribute from the test and store it in the base class. I also changed the derived class to append to the logs correctly. Can you please make sure that the important parts from this PR are integrated into the the FunctionalTestBaseClass: [https://github.com/dotnet/machinelearning/pull/4710](https://github.com/dotnet/machinelearning/pull/4710)

---
In reply to: [376616140](https://github.com/dotnet/machinelearning/pull/4810#discussion_r376616140) [](ancestors = 376616140,376581212)

harishsk: The above mentioned work need not be part of this PR. I am approving this. Please use a different PR for that work.

harishsk: I am approving this. You can do the above mentioned work in a different PR.

frank-dong-ms: Sure, I will open a separate PR if we need certain fix

---
In reply to: [376622761](https://github.com/dotnet/machinelearning/pull/4810#discussion_r376622761) [](ancestors = 376622761)

PR #4812, opened 2020-02-07
Added in DateTime type support for TimeSeriesImputer

michaelgsharp:
Added in System.DateTime support for the Date column in TimeSeriesImputer.

Internally it converts it to POSIX time before it sends it to the native implementation, and then converts it back to System.DateTime for the IDataView.

Also added more comments and comments for the Enums.

### DIFF
--- a/src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
+++ b/src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
@@ -32,7 +32,7 @@ namespace Microsoft.ML.Featurizers
         /// purpose of this estimator. Other column types will have the default value placed if a row is imputed.
         /// </summary>
         /// <param name="catalog">The transform catalog.</param>
-        /// <param name="timeSeriesColumn">Column representing the time series. Should be of type <see cref="long"/></param>
+        /// <param name="timeSeriesColumn">Column representing the time series. Should be of type <see cref="long"/> or <see cref="System.DateTime"/></param>
         /// <param name="grainColumns">List of columns to use as grains</param>
         /// <param name="imputeMode">Mode of imputation for missing values in column. If not passed defaults to forward fill</param>
         public static TimeSeriesImputerEstimator ReplaceMissingTimeSeriesValues(this TransformsCatalog catalog, string timeSeriesColumn, string[] grainColumns,
@@ -46,7 +46,7 @@ namespace Microsoft.ML.Featurizers
         /// purpose of this estimator.
         /// </summary>
         /// <param name="catalog">The transform catalog.</param>
-        /// <param name="timeSeriesColumn">Column representing the time series. Should be of type <see cref="long"/></param>
+        /// <param name="timeSeriesColumn">Column representing the time series. Should be of type <see cref="long"/> or <see cref="System.DateTime"/></param>
         /// <param name="grainColumns">List of columns to use as grains</param>
         /// <param name="filterColumns">List of columns to filter. If <paramref name="filterMode"/> is <see cref="TimeSeriesImputerEstimator.FilterMode.Exclude"/> than columns in the list will be ignored.
         /// If <paramref name="filterMode"/> is <see cref="TimeSeriesImputerEstimator.FilterMode.Include"/> than values in the list are the only columns imputed.</param>
@@ -83,6 +83,16 @@ namespace Microsoft.ML.Featurizers
     /// | Output column data type | All Types |
     /// | Exportable to ONNX | No |
     ///
+    /// The TimeSeriesImputer imputes missing rows and column data per grain (category), based on the dates in the date column. This operation needs to happen to every column in the IDataView,
+    /// If you "filter" a column using the filterColumns and filterMode parameters, if a row is imputed the default value for that type will be used.
+    /// Currently only float/double/string columns are supported for imputation strategies, and an empty string is considered "missing" for the
+    /// purpose of this estimator. A new column is added to the schema after this operation is run. The column is called "IsRowImputed" and is a
+    /// boolean value representing if the row was created as a result of this operation or not.
+    ///
+    /// NOTE: It is not recommended to chain this multiple times. If a column is filtered, the default value is placed when a row is imputed, and the
+    /// default value is not null. Thus any other TimeSeriesImputers will not be able to replace those values anymore causing essentially a very
+    /// computationally expensive NO-OP.
+    ///
     /// The <xref:Microsoft.ML.Transforms.TimeSeriesImputerEstimator> is not a trivial estimator and needs training.
     ///
     ///
@@ -98,7 +108,7 @@ namespace Microsoft.ML.Featurizers
 
         private readonly IHost _host;
         private static readonly List<Type> _currentSupportedTypes = new List<Type> { typeof(sbyte), typeof(byte), typeof(short), typeof(ushort), typeof(int), typeof(uint),
-            typeof(long), typeof(ulong), typeof(float), typeof(double), typeof(string), typeof(ReadOnlyMemory<char>)};
+            typeof(long), typeof(ulong), typeof(float), typeof(double), typeof(string), typeof(ReadOnlyMemory<char>), typeof(DateTime)};
 
         #region Options
         internal sealed class Options : TransformInputBase
@@ -127,6 +137,12 @@ namespace Microsoft.ML.Featurizers
 
         #region Class Enums
 
+        /// <summary>
+        /// This is the representation of which Imputation Strategy to use.
+        /// ForwardFill takes the value from the last good row and propagates it forward anytime a row is imputer or a missing value is found.
+        /// BackFill is the same as ForwardFill, expect it takes from the next good row and propagates backwards.
+        /// Median only supports float/double, takes the median value found during training and uses that to replace missing values
+        /// </summary>
         public enum ImputationStrategy : byte
         {
             ForwardFill = 1,
@@ -135,6 +151,12 @@ namespace Microsoft.ML.Featurizers
             // Interpolate = 4, interpolate not currently supported in the native code.
         };
 
+        /// <summary>
+        /// What the filter strategy used is.
+        /// NoFilter takes all of the columns so you dont have to specify anything.
+        /// Include only does the specified ImputationStrategy on the columns you specify. The other columns will get a default value.
+        /// Exclude is the exact opposite of Include, and does the ImputationStrategy on all columns but the ones you specify, which will get the default value.
+        /// </summary>
         public enum FilterMode : byte
         {
             NoFilter = 1,
@@ -559,6 +581,8 @@ namespace Microsoft.ML.Featurizers
                     return new NumericTypedColumn<double>(column, optionalColumns.Contains(column.Name));
                 else if (type == typeof(ReadOnlyMemory<char>).ToString())
                     return new StringTypedColumn(column, optionalColumns.Contains(column.Name));
+                else if (type == typeof(DateTime).ToString())
+                    return new DateTimeTypedColumn(column, optionalColumns.Contains(column.Name));
 
                 throw new InvalidOperationException($"Unsupported type {type}");
             }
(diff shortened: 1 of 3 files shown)


src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -83,6 +83,16 @@ public static class TimeSeriesImputerExtensionClass
...
     ///
+    /// The TimeSeriesImputer imputes missing rows and column data per grain (category), based on the dates in the date column. This operation needs to happen to every column in the IDataView,
+    /// If you "filter" a column using the filterColumns and filterMode parameters, if a row is imputed the default value for that type will be used.
+    /// Currently only float/double/string columns are supported for imputation strategies, and an empty string is considered "missing" for the
+    /// purpose of this estimator. A new column is added to the schema after this operation is run. The column is called "IsRowImputed" and is a
+    /// boolean value representing if the row was created as a result of this operation or not.
+    ///
+    /// NOTE: It is not recommended to chain this multiple times. If a column is filtered, the default value is placed when a row is imputed, and the
+    /// default value is not null. Thus any other TimeSeriesImputers will not be able to replace those values anymore causing essentially a very
+    /// computationally expensive NO-OP.
```
yaeldekel: Is this intentionally duplicated?

yaeldekel:
>     /// Currently only float/double/string columns are supported for imputation strategies, and an empty string is considered "missing" for the

This should be updated. Also, it should match the types in the table below.

---
Refers to: src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs:66 in 99a5eb3. [](commit_id = 99a5eb31ba8cd686c301c2f39e724f398f42c9f0, deletion_comment = False)

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -83,6 +83,16 @@ public static class TimeSeriesImputerExtensionClass
     /// | Output column data type | All Types |
     /// | Exportable to ONNX | No |
     ///
+    /// The TimeSeriesImputer imputes missing rows and column data per grain (category), based on the dates in the date column. This operation needs to happen to every column in the IDataView,
```
natke: Restructure summary and remarks so that the summary is just that - a one line description of what the class and when you would use it (if appropriate). Have a look at this page to see what the descriptions look like once they are rendered as HTML: https://review.docs.microsoft.com/en-us/dotnet/api/microsoft.ml.featurizers?view=ml-dotnet-preview-latest&branch=smoke-test-preview

Put the rest of the information into the remarks section. This is what a developer will look at when they are using the class rather than comparing transforms to see which one they need to use: https://review.docs.microsoft.com/en-us/dotnet/api/microsoft.ml.featurizers.timeseriesimputerestimator?view=ml-dotnet-preview-latest&branch=smoke-test-preview

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -83,6 +83,16 @@ public static class TimeSeriesImputerExtensionClass
...
+    /// If you "filter" a column using the filterColumns and filterMode parameters, if a row is imputed the default value for that type will be used.
+    /// Currently only float/double/string columns are supported for imputation strategies, and an empty string is considered "missing" for the
+    /// purpose of this estimator. A new column is added to the schema after this operation is run. The column is called "IsRowImputed" and is a
+    /// boolean value representing if the row was created as a result of this operation or not.
+    ///
+    /// NOTE: It is not recommended to chain this multiple times. If a column is filtered, the default value is placed when a row is imputed, and the
+    /// default value is not null. Thus any other TimeSeriesImputers will not be able to replace those values anymore causing essentially a very
+    /// computationally expensive NO-OP.
+    ///
     /// The <xref:Microsoft.ML.Transforms.TimeSeriesImputerEstimator> is not a trivial estimator and needs training.
```
natke: This xref is not resolving. Do we even need this line? We have the information in the table already

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -127,6 +137,12 @@ internal sealed class Options : TransformInputBase
 
         #region Class Enums
 
+        /// <summary>
+        /// This is the representation of which Imputation Strategy to use.
```
natke: Suggest the following re-word:

///&lt;summary&gt;
/// Method by which values are imputed.
/// &lt;/summary&gt;
///&lt;remarks&gt;
/// Add the extra information here ...

It would be even better if the descriptions of each value could be annotated on the value itself. I have not seen examples of this but I will check

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -135,6 +151,12 @@ public enum ImputationStrategy : byte
             // Interpolate = 4, interpolate not currently supported in the native code.
         };
 
+        /// <summary>
+        /// What the filter strategy used is.
```
natke: Method by which columns are selected for imputing values

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
michaelgsharp: Left the first line in the summary section as its a pretty good summary. Moved the rest down into the remarks section.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
michaelgsharp: We don't need that line. I'll go ahead and remove it.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
michaelgsharp: Let me know if you find a way to do it. I'll skip this for now until I hear back.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
michaelgsharp: Nope. I didn't mean to leave it in the summary as well. Fixed now.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
michaelgsharp: Changed to your wording.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
natke: We should at least put them in remarks.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -61,15 +61,7 @@ public static class TimeSeriesImputerExtensionClass
...
-    /// Imputes missing rows and column data per grain, based on the dates in the date column. This operation needs to happen to every column in the IDataView,
-    /// If you "filter" a column using the filterColumns and filterMode parameters, if a row is imputed the default value for that type will be used.
-    /// Currently only float/double/string columns are supported for imputation strategies, and an empty string is considered "missing" for the
-    /// purpose of this estimator. A new column is added to the schema after this operation is run. The column is called "IsRowImputed" and is a
-    /// boolean value representing if the row was created as a result of this operation or not.
-    ///
-    /// NOTE: It is not recommended to chain this multiple times. If a column is filtered, the default value is placed when a row is imputed, and the
-    /// default value is not null. Thus any other TimeSeriesImputers will not be able to replace those values anymore causing essentially a very
-    /// computationally expensive NO-OP.
+    /// Imputes missing rows and column data per grain, based on the dates in the date column.
```
natke: Looking much better! What does "grain" mean in this context?

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
natke: Found out that adding summaries for individual enum values *is* possible. See source example here: https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.AutoML/API/BinaryClassificationExperiment.cs#L45 rendered in the table here: https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.automl.binaryclassificationmetric?view=ml-dotnet-preview

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -559,6 +570,8 @@ internal static TypedColumn CreateTypedColumn(DataViewSchema.Column column, stri
                     return new NumericTypedColumn<double>(column, optionalColumns.Contains(column.Name));
                 else if (type == typeof(ReadOnlyMemory<char>).ToString())
                     return new StringTypedColumn(column, optionalColumns.Contains(column.Name));
+                else if (type == typeof(DateTime).ToString())
```
eerhardt: This `if` clause seems like it should be re-written to not call `ToString()` and compare strings.

```C#
var type = column.Type.RawType;
if (type == typeof(sbyte))
{
...
}
else if (type == typeof(short))
{
...
```

That is a much more performant way of doing the same checks.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
```diff
@@ -637,6 +650,33 @@ internal override byte[] GetSerializedValue()
...
+            }
+
+            internal override byte[] GetSerializedValue()
+            {
+                var dateTime = GetValue();
+                byte[] bytes;
+
+                var value = dateTime.Subtract(_unixEpoch).Ticks / TimeSpan.TicksPerSecond;
+
+                bytes = BitConverter.GetBytes(value);
```
eerhardt: This is going to allocate a new, very small, `byte[]` every time it is called. That is going to make the garbage collector do a lot of extra work to clean these small arrays up. Have we done perf testing on this?

Maybe a better design here would be to ask the Column for the size it needs, and then pass an appropriately sized `Span<byte>` into the Column, telling it to write to that Span. This would use as few allocations as necessary.

eerhardt: It would also avoid having to then copy this data into a different buffer that gets passed into native code. Instead, the `TypedColumn` could just write to the "final" place directly.

---
In reply to: [377914670](https://github.com/dotnet/machinelearning/pull/4812#discussion_r377914670) [](ancestors = 377914670)

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
michaelgsharp: Grain is essentially the category. Not sure why its called a "grain", but thats the terminology they use.

eerhardt:
Any chance you can update the samples to use a `DateTime` value instead of the `long` values it uses now? I think that makes a better sample for users.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
natke: Who are "they"? :-) Is the term relevant to the user of this API? If not, I would just change it to `category`

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
michaelgsharp: This check happens only once for each column, so its not a big performance hit. It also makes saving and loading much easier due to namespace issues. I had an issue in the past with this when I was comparing types directly. If you think its a big enough deal I can refactor it, but I think the amount of perf we will gain will be negligible since its only once per column.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
michaelgsharp: Alright, I changed the code to use a `MemoryStream` and then I reuse the underlying `byte[]` stream for each call. The `MemoryStream` handles the resizing if needed so I removed that from my code, and now it should only be creating the `byte[]` once directly using the `MemoryStream`.

src/Microsoft.ML.Featurizers/TimeSeriesImputer.cs
michaelgsharp: Alright, I have added in the summaries on the individual enum values.

PR #4816, opened 2020-02-08
Fixed bug in tensorflow tests due to long paths for working folders

harishsk:
TF tests are relying on IClassFixture to carry forward state from test to test. This by default creates excessively long paths that result in file system errors when there is a longer prefix for the source code. I have fixed it to take temp paths with shorter path lengths.

### DIFF
--- a/test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
+++ b/test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
@@ -29,37 +29,33 @@ namespace Microsoft.ML.Scenarios
 
     internal sealed class TensorFlowScenariosTestsFixture : IDisposable
     {
+        public static string tempFolder;
         public static string parentWorkspacePath;
         public static string assetsPath;
-        internal void CreateParentWorkspacePathForImageClassification()
+        internal static void CreateParentWorkspacePathForImageClassification()
         {
-            string assetsRelativePath = @"assets";
-            assetsPath = GetAbsolutePath(assetsRelativePath);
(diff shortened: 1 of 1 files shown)


test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -29,37 +29,33 @@ namespace Microsoft.ML.Scenarios
...
+        public static string tempFolder;
         public static string parentWorkspacePath;
         public static string assetsPath;
-        internal void CreateParentWorkspacePathForImageClassification()
+        internal static void CreateParentWorkspacePathForImageClassification()
         {
-            string assetsRelativePath = @"assets";
-            assetsPath = GetAbsolutePath(assetsRelativePath);
-            string workspacePath = Path.Combine(assetsPath, "cached");
+            tempFolder = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
```
yaeldekel: >Path.GetTempPath(), Path.GetRandomFileName() [](start = 38, length = 44)

Are these guaranteed to be shorter?

harishsk: Yes, these are guaranteed to be shorter. Path.GetRandomeFile() only produces a 12 character random filename and in combination with Path.GetTempPath() (which is typically %USERPROFILE%\AppData\Local\Temp) the overall path will be significantly shorter than what what produced earlier. 

---
In reply to: [376950593](https://github.com/dotnet/machinelearning/pull/4816#discussion_r376950593) [](ancestors = 376950593)

PR #4817, opened 2020-02-08
WIP: Make sure Done() is called in cleanup if it hasn't been called

harishsk:
This makes sure call stack is collected for intermediate failures.


src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/LogisticRegression.cs
```diff
@@ -455,7 +455,7 @@ private protected override VBuffer<float> InitializeWeightsFromPredictor(IPredic
             CurrentWeights.CopyTo(ref weights, 1, CurrentWeights.Length - 1);
             return new ParameterMixingCalibratedModelParameters<LinearBinaryModelParameters, PlattCalibrator>(Host,
                 new LinearBinaryModelParameters(Host, in weights, bias, _stats),
-                new PlattCalibrator(Host, -1, 0));
+                new PlattCalibrator(Host, 1, 0));
```
yaeldekel: It seems that the bug is actually in the `PlattCalibrator` itself:

While in the comment it says this:

```
    /// The Platt calibrator calculates the probability following:
    /// P(x) = 1 / (1 + exp(-<see cref="PlattCalibrator.Slope"/> * x + <see cref="PlattCalibrator.Offset"/>)
```
The actual calculation does this:

```
        public float PredictProbability(float output)
        {
            if (float.IsNaN(output))
                return output;
            return PredictProbability(output, Slope, Offset);
        }

        internal static float PredictProbability(float output, Double a, Double b)
        {
            return (float)(1 / (1 + Math.Exp(a * output + b)));
        }
```

The value here should be changed to 1 only if we change the computation in `PlattCalibrator`. (looking at the baselines, the computation without the change looks better: label 1 examples give higher probabilities, and label 0 examples give lower probabilities.)

frank-dong-ms:
I remember you said this PR is still in progress, right?
If so, you can use draft PR before it is ready

harishsk:
I forgot to add a draft flag earlier. I have changed the title to include WIP now. I will most likely close this PR and reopen a new one.

---
In reply to: [584371281](https://github.com/dotnet/machinelearning/pull/4817#issuecomment-584371281) [](ancestors = 584371281)

harishsk:
This PR is superseded by #4823 . Closing this one.

PR #4818, opened 2020-02-08
Changed Binarizer node to be cast to the type of the predicted label …

harishsk:
…column's data type

In BinaryClassifierScorer's SaveAsOnnxCore function we were always casting the output of the Binarizer to a bool. But in some cases BinaryClassifierScorer can output a key value (uint) and in this case we should cast the output to a uint. This fix changes the cast to be dependent on the output type of the predicted label.

### DIFF
--- a/src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
+++ b/src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Diagnostics;
 using Microsoft.ML;
 using Microsoft.ML.Data;
 using Microsoft.ML.Internal.Utilities;
@@ -202,7 +203,15 @@ namespace Microsoft.ML.Data
             OnnxNode node;
             var binarizerOutput = ctx.AddIntermediateVariable(null, "BinarizerOutput", true);
 
-            node = ctx.CreateNode(opType, ctx.GetVariableName(outColumnNames[1]), binarizerOutput, ctx.GetNodeName(opType));
+            string scoreColumn;
+            if (Bindings.RowMapper.OutputSchema[Bindings.ScoreColumnIndex].Name == "Score")
+                scoreColumn = outColumnNames[1];
+            else
+            {
+                Host.Assert(Bindings.InfoCount >= 3);
+                scoreColumn = outColumnNames[2];
+            }
+            node = ctx.CreateNode(opType, ctx.GetVariableName(scoreColumn), binarizerOutput, ctx.GetNodeName(opType));
             node.AddAttribute("threshold", _threshold);
 
             opType = "Cast";
--- a/src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
+++ b/src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
@@ -216,10 +216,11 @@ namespace Microsoft.ML.Data
                 node = ctx.CreateNode(opType, ctx.GetVariableName(outColumnNames[1]), binarizerOutput, ctx.GetNodeName(opType));
                 node.AddAttribute("threshold", 0.0);
             }
+
             opType = "Cast";
             node = ctx.CreateNode(opType, binarizerOutput, ctx.GetVariableName(outColumnNames[0]), ctx.GetNodeName(opType), "");
-            var t = InternalDataKindExtensions.ToInternalDataKind(DataKind.Boolean).ToType();
-            node.AddAttribute("to", t);
+            var predictedLabelCol = OutputSchema.GetColumnOrNull(outColumnNames[0]);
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
```diff
@@ -216,10 +216,11 @@ private protected override void SaveAsOnnxCore(OnnxContext ctx)
                 node = ctx.CreateNode(opType, ctx.GetVariableName(outColumnNames[1]), binarizerOutput, ctx.GetNodeName(opType));
                 node.AddAttribute("threshold", 0.0);
             }
+
             opType = "Cast";
             node = ctx.CreateNode(opType, binarizerOutput, ctx.GetVariableName(outColumnNames[0]), ctx.GetNodeName(opType), "");
-            var t = InternalDataKindExtensions.ToInternalDataKind(DataKind.Boolean).ToType();
-            node.AddAttribute("to", t);
+            var predictedLabelCol = OutputSchema.GetColumnOrNull(outColumnNames[0]);
+            node.AddAttribute("to", predictedLabelCol.HasValue ? predictedLabelCol.Value.Type.RawType : typeof(bool));
```
yaeldekel: >predictedLabelCol.HasValue [](start = 36, length = 26)

Doesn't it always have a value?
Or, if it doesn't - should we be adding the ONNX node at all?

yaeldekel:
>              */

I know this isn't related to the change in this PR, but is this correct? The binary classifier scorer has a `_threshold` field, and in the constructor you can specify which column (score/probability) you want to apply the threshold to.

While I could not find a way in the public API to change which column is used (by default it is actually the score column, even if the probability column exists), there is a public API to change the threshold value: [`ChangeModelThreshold`](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/TrainCatalog.cs#L268-L274)

---
Refers to: src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs:203 in fc1925c. [](commit_id = fc1925ceaef44845d10d7788c0a1f544a5219fba, deletion_comment = False)

ganik:
>              */

Yep, this is not correct. Predicted label has to be based off threshold

---
In reply to: [584023200](https://github.com/dotnet/machinelearning/pull/4818#issuecomment-584023200) [](ancestors = 584023200)

---
Refers to: src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs:203 in fc1925c. [](commit_id = fc1925ceaef44845d10d7788c0a1f544a5219fba, deletion_comment = False)

harishsk:
>              */

You are right! I have fixed it in the next commit. 

---
In reply to: [584286522](https://github.com/dotnet/machinelearning/pull/4818#issuecomment-584286522) [](ancestors = 584286522,584023200)

---
Refers to: src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs:203 in fc1925c. [](commit_id = fc1925ceaef44845d10d7788c0a1f544a5219fba, deletion_comment = False)

src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
harishsk: I have added an Assert to capture that and fixed the next line.

---
In reply to: [376929923](https://github.com/dotnet/machinelearning/pull/4818#discussion_r376929923) [](ancestors = 376929923)

src/Microsoft.ML.Data/Scorers/BinaryClassifierScorer.cs
```diff
@@ -198,28 +198,18 @@ private protected override void SaveAsOnnxCore(OnnxContext ctx)
...
-                Host.Assert(ctx.ContainsColumn(outColumnNames[2]));
-                node = ctx.CreateNode(opType, ctx.GetVariableName(outColumnNames[2]), binarizerOutput, ctx.GetNodeName(opType));
-                node.AddAttribute("threshold", 0.5);
-            }
-            else
-            {
-                node = ctx.CreateNode(opType, ctx.GetVariableName(outColumnNames[1]), binarizerOutput, ctx.GetNodeName(opType));
-                node.AddAttribute("threshold", 0.0);
-            }
+            node = ctx.CreateNode(opType, ctx.GetVariableName(outColumnNames[1]), binarizerOutput, ctx.GetNodeName(opType));
```
yaeldekel: >outColumnNames[1] [](start = 62, length = 17)

We should still account for the case where the threshold is applied to the probability column. It can be done this way:
```
    string scoreCol = Bindings.RowMapper.OutputSchema[Bindings.ScoreColumnIndex].Name;
    if (scoreCol == "Score")
        // use outColumnNames[1]
    else
        // use outColumnNames[2]
```

harishsk: Thanks! Fixed.

---
In reply to: [377516305](https://github.com/dotnet/machinelearning/pull/4818#discussion_r377516305) [](ancestors = 377516305)

Issue #4143, opened 2019-08-23
LDA. Seed topic-words matrix


IvanAntipov:
The widely used practice of LDA is the seeding of the matrix of topic-word

ex.
https://github.com/vi3k6i5/GuidedLDA

This possibility required to:
1. Have stable clustering (topics number remain if data is not changed significantly) when re-training the model
2. Have semi-supervised model (when user define some words for topics)

This is a very common business requirement

codemzs:
Thanks but we no plans to implement this in the short-term. Since enhancement tag has been added we can pull this up when we triage the requested features.

PR #4823, opened 2020-02-10
Added Done() call in BaseTestBaseline.Cleanup and added related fixes

harishsk:
This PR is similar to #4817 but standardizes the default value for Slope in PlattCalibrator to -1 instead of 1 since -1 seems to be used everywhere in the code. (If this PR is approved, I will be closing the other PR)

### DIFF
--- a/src/Microsoft.ML.Data/Prediction/Calibrator.cs
+++ b/src/Microsoft.ML.Data/Prediction/Calibrator.cs
@@ -1581,10 +1581,10 @@ namespace Microsoft.ML.Calibrators
         [TlcModule.Component(Name = "FixedPlattCalibrator", FriendlyName = "Fixed Platt Calibrator", Aliases = new[] { "FixedPlatt", "FixedSigmoid" })]
         public sealed class Arguments : ICalibratorTrainerFactory
         {
-            [Argument(ArgumentType.LastOccurrenceWins, HelpText = "The slope parameter of f(x) = 1 / (1 + exp(-slope * x + offset)", ShortName = "a")]
-            public Double Slope = 1;
+            [Argument(ArgumentType.LastOccurrenceWins, HelpText = "The slope parameter of f(x) = 1 / (1 + exp(slope * x + offset)", ShortName = "a")]
+            public Double Slope = -1;
 
-            [Argument(ArgumentType.LastOccurrenceWins, HelpText = "The offset parameter of f(x) = 1 / (1 + exp(-slope * x + offset)", ShortName = "b")]
+            [Argument(ArgumentType.LastOccurrenceWins, HelpText = "The offset parameter of f(x) = 1 / (1 + exp(slope * x + offset)", ShortName = "b")]
             public Double Offset = 0;
 
             public ICalibratorTrainer CreateComponent(IHostEnvironment env)
@@ -1618,7 +1618,7 @@ namespace Microsoft.ML.Calibrators
 
     ///<summary>
     /// The Platt calibrator calculates the probability following:
-    /// P(x) = 1 / (1 + exp(-<see cref="PlattCalibrator.Slope"/> * x + <see cref="PlattCalibrator.Offset"/>)
+    /// P(x) = 1 / (1 + exp(<see cref="PlattCalibrator.Slope"/> * x + <see cref="PlattCalibrator.Offset"/>)
     /// </summary>.
     public sealed class PlattCalibrator : ICalibrator, IParameterMixer, ICanSaveModel, ISingleCanSavePfa, ISingleCanSaveOnnx
     {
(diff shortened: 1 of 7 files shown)


harishsk:
This test passes locally but is failing remotely. It seems to be one of the flaky tests.

---
In reply to: [356336770](https://github.com/dotnet/machinelearning/pull/4823#issuecomment-356336770) [](ancestors = 356336770)

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -2086,7 +2086,7 @@ public sealed class NoArgumentsInput : CalibrateInputBase
         public sealed class FixedPlattInput : CalibrateInputBase
         {
             [Argument(ArgumentType.AtMostOnce, ShortName = "slope", HelpText = "The slope parameter of the calibration function 1 / (1 + exp(-slope * x + offset)", SortOrder = 1)]
```
yaeldekel: >-slope [](start = 141, length = 6)

Can you change this to `exp(slope * x + offset)`? (and in the `Offset` argument as well).

src/Microsoft.ML.Data/Prediction/Calibrator.cs
```diff
@@ -2086,7 +2086,7 @@ public sealed class NoArgumentsInput : CalibrateInputBase
...
-            public Double Slope = 1;
+            public Double Slope = -1;
```
yaeldekel: >Slope [](start = 26, length = 5)

`FixedPlattCalibratorTrainer.Arguments` also has a `Slope` field that defaults to 1.

yaeldekel:
>     /// P(x) = 1 / (1 + exp(-<see cref="PlattCalibrator.Slope"/> * x + <see cref="PlattCalibrator.Offset"/>)

This should also be fixed.

---
Refers to: src/Microsoft.ML.Data/Prediction/Calibrator.cs:1621 in f86faac. [](commit_id = f86faacd814ba63d0eee2a83c13699b0b3875462, deletion_comment = False)

frank-dong-ms:
>         protected void Done()

In this new implementation, we can get rid of this Done() method and remove all usage. These assertion is done automatically in Cleanup method

---
Refers to: test/Microsoft.ML.TestFramework/BaseTestBaseline.cs:145 in f86faac. [](commit_id = f86faacd814ba63d0eee2a83c13699b0b3875462, deletion_comment = False)

test/Microsoft.ML.Predictor.Tests/TestPredictors.cs
```diff
@@ -948,7 +949,7 @@ public void TestMulticlassEnsembleCombiner()
                             predGetters[i](ref preds[i]);
                         }
                         if (scores.All(s => !float.IsNaN(s)))
-                            CompareNumbersWithTolerance(score, scores.Sum() / predCount);
+                            CompareNumbersWithTolerance(score, scores.Sum() / predCount, digitsOfPrecision: 5);
```
frank-dong-ms: >digitsOfPrecision [](start = 89, length = 17)

why we use less precise here?

frank-dong-ms:
This one test: LdaWorkoutEstimatorCore is failing on all configurations and seems like real fail

yaeldekel:
LDA has some non determinism in it (see issue #4143), most of the LDA tests are disabled (or they don't do any comparisons of the outputs), I guess this one slipped through the cracks because it doesn't call `Done()`...

---
In reply to: [584859078](https://github.com/dotnet/machinelearning/pull/4823#issuecomment-584859078) [](ancestors = 584859078)

harishsk:
@yaeldekel Thanks for the info. I have added it to the list of skipped tests for now.

---
In reply to: [585197699](https://github.com/dotnet/machinelearning/pull/4823#issuecomment-585197699) [](ancestors = 585197699,584859078)

test/Microsoft.ML.Predictor.Tests/TestPredictors.cs
harishsk: The default is 7 digits. All tests are not equally precise. 

---
In reply to: [377899328](https://github.com/dotnet/machinelearning/pull/4823#discussion_r377899328) [](ancestors = 377899328)

harishsk:
>         protected void Done()

There is a problem with the current implementation. The assertion happens in Cleanup() but when Cleanup is called, the test context is lost. I am still debating whether to remove Done completely or find a different way to force Done to be called in the context of the original function. For now, I am leaving it as is. 

---
In reply to: [584851784](https://github.com/dotnet/machinelearning/pull/4823#issuecomment-584851784) [](ancestors = 584851784)

---
Refers to: test/Microsoft.ML.TestFramework/BaseTestBaseline.cs:145 in f86faac. [](commit_id = f86faacd814ba63d0eee2a83c13699b0b3875462, deletion_comment = False)

test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
```diff
@@ -485,10 +485,14 @@ public void OneClassMatrixFactorizationInMemoryDataZeroBaseIndex()
             var testPrediction = model.Transform(testDataView);
 
             var testResults = mlContext.Data.CreateEnumerable<OneClassMatrixElementZeroBasedForScore>(testPrediction, false).ToList();
+
+            // REVIEW: We are seeing lower precision on non-Windows platforms
```
frank-dong-ms: >REVIEW [](start = 15, length = 6)

maybe use below tags in comments so we can find this easily from task list: 
TODO TEST_STABILITY:

test/Microsoft.ML.Tests/Transformers/TextFeaturizerTests.cs
```diff
@@ -714,6 +714,7 @@ public void LdaWorkout()
         }
 
         [Fact]
+        [Trait("Category", "SkipInCI")]
         public void LdaWorkoutEstimatorCore()
```
frank-dong-ms: >LdaWorkoutEstimatorCore [](start = 20, length = 23)

I haven't see this fail in full build and CI build in last 14 days, this should be low fail rate tests and I think it is okay to left this one enabled, if we see this fails often later we can go back to this one

PR #4824, opened 2020-02-11
Enable OnnxTransformer to accept KeyDataViewTypes as if they were UInt32

antoniovs1029:
Addresses https://github.com/microsoft/NimbusML/issues/426

As discussed offline with @harishsk , @ganik and @pieths , the update in here is the best temporary solution to the problem.

Simply enables OnnxTransformer to accept a `KeyDataViewType` column as input of a model that actually expects an `UInt32` input column. This, because of the way that NimbusML works with Category columns in Pandas' DataFrames, by converting them into `KeyDataViewType` columns in ML.NET, without actually adding KeyToValueTransformers into the pipeline.

### DIFF
--- a/test/Microsoft.ML.Tests/OnnxConversionTest.cs
+++ b/test/Microsoft.ML.Tests/OnnxConversionTest.cs
@@ -1475,7 +1475,8 @@ namespace Microsoft.ML.Tests
             var originalData = loader.Load(dataPath);
             var pipeline1 = mlContext.Transforms.Conversion.MapValueToKey("Label");
             var mappedData = pipeline1.Fit(originalData).Transform(originalData);
-            string mappedDataPath = @"C:\Users\anvelazq\Desktop\is22\data-new.idv";
+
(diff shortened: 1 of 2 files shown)


test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -1450,6 +1450,74 @@ public void CopyColumnsOnnxTest()
...
+                    },
+                hasHeader: false
+            );
+
+            string dataPath = GetDataPath("iris.txt");
+            var originalData = loader.Load(dataPath);
+            var pipeline1 = mlContext.Transforms.Conversion.MapValueToKey("Label");
+            var mappedData = pipeline1.Fit(originalData).Transform(originalData);
+            string mappedDataPath = @"C:\Users\anvelazq\Desktop\is22\data-new.idv";
+            using (FileStream stream = new FileStream(mappedDataPath, FileMode.Create))
```
harishsk: Please fix hardcoded path

PR #4827, opened 2020-02-11
Fix off by 1 error with the cats_int64s attribute for the OneHotEncoder ONNX operator

pieths:


pieths:
I just ran those tests and added two more off by 1 fixes for the unit tests. The changes have been added to the pull request.

PR #4828, opened 2020-02-12
Add support for combining hashes in vector columns to HashingTransformer

yaeldekel:
I am making this change so that `CountTargetEncodingEstimator` (see PR #4514 ) can start using `HashingEstimator` instead of `HashJoiningTransform` (see this [comment](https://github.com/dotnet/machinelearning/pull/4514#issuecomment-584502932) in the above PR).

In addition to enabling `CountTargetEncodingEstimator` to use it, this will fix a bug when splitting datasets in the [CV command](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Commands/CrossValidationCommand.cs#L335) and in the [train-test/CV APIs](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs#L522). If a stratification column that is a vector type is specified, currently an exception will be thrown because the `RangeFilter` applied after hashing cannot handle vector columns.


yaeldekel:
Question: some of the input types for hashing have missing values (float, double and key type). The hashing transformer maps missing values in the input, to 0, the missing value of key type. What should we do in the combine case? 
I think the correct thing to do is to return 0 if any of the slots have missing values (that way, the behavior of a length-1 vector is identical whether `Combine` is true or false), but I would like to know what other people think.

src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
```diff
@@ -511,19 +511,20 @@ internal static void EnsureGroupPreservationColumn(IHostEnvironment env, ref IDa
...
+                    if (itemType is DateTimeDataViewType || itemType is DateTimeOffsetDataViewType || itemType is TimeSpanDataViewType)
+                        data = new TypeConvertingTransformer(env, origStratCol, DataKind.Int64, origStratCol).Transform(data);
+
+                    HashingEstimator.ColumnOptions columnOptions;
                     if (seed.HasValue)
-                        columnOptions = new HashingEstimator.ColumnOptionsInternal(samplingKeyColumn, origStratCol, 30, (uint)seed.Value);
+                        columnOptions = new HashingEstimator.ColumnOptions(samplingKeyColumn, origStratCol, 30, (uint)seed.Value, combine: true);
                     else if (((ISeededEnvironment)env).Seed.HasValue)
-                        columnOptions = new HashingEstimator.ColumnOptionsInternal(samplingKeyColumn, origStratCol, 30, (uint)((ISeededEnvironment)env).Seed.Value);
+                        columnOptions = new HashingEstimator.ColumnOptions(samplingKeyColumn, origStratCol, 30, (uint)((ISeededEnvironment)env).Seed.Value, combine: true);
```
codemzs: You can merge these conditions in one by defining var localSeed = seed.HasValue ? seed.Value : (ISeededEnvironment)env).Seed.HasValue ? (ISeededEnvironment)env).Seed.Value : null;

if(localSeed.HasValue)
    ...
else
    columnOptions = new HashingEstimator.ColumnOptions(samplingKeyColumn, origStratCol, 30, combine: true);

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -122,14 +128,15 @@ private static VersionInfo GetVersionInfo()
             return new VersionInfo(
                 modelSignature: "HASHTRNS",
                 // verWrittenCur: 0x00010001, // Initial
-                verWrittenCur: 0x00010002, // Invert hash key values, hash fix
+                //verWrittenCur: 0x00010002, // Invert hash key values, hash fix
```
codemzs: >v [](start = 18, length = 1)

space

src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
justinormont: I don't think we want `HashingEstimator` listening to the global seed. The hashing is a different meaning of "seed"; this isn't a PRNG seed. 

See https://github.com/dotnet/machinelearning/issues/4752#issuecomment-580686290

/cc @najeeb-kazmi

src/Microsoft.ML.Data/Commands/CrossValidationCommand.cs
```diff
@@ -328,11 +328,14 @@ private string GetSplitColumn(IChannel ch, IDataView input, ref IDataView output
...
-                    output = new HashingEstimator(Host, origStratCol, stratificationColumn, 30).Fit(input).Transform(input);
+                    stratificationColumn = input.Schema.GetTempColumnName("strat");
+
+                    // HashingEstimator currently handles all primitive types except for DateTime, DateTimeOffset and TimeSpan.
+                    var itemType = type.GetItemType();
+                    if (itemType is DateTimeDataViewType || itemType is DateTimeOffsetDataViewType || itemType is TimeSpanDataViewType)
+                        input = new TypeConvertingTransformer(Host, origStratCol, DataKind.Int64, origStratCol).Transform(input);
+
+                    output = new HashingEstimator(Host, stratificationColumn, origStratCol, 30).Fit(input).Transform(input);
                 }
```
harishsk: I realize the previous code too used 30 bits. But the default seems to be 31 bits. Any reason why we have 30 here? If there is a specific reason for 30 bits, can you please add a note to that in the code?

src/Microsoft.ML.Data/DataLoadSave/DataOperationsCatalog.cs
```diff
@@ -511,19 +511,18 @@ internal static void EnsureGroupPreservationColumn(IHostEnvironment env, ref IDa
...
+                    // HashingEstimator currently handles all primitive types except for DateTime, DateTimeOffset and TimeSpan.
+                    var itemType = type.GetItemType();
+                    if (itemType is DateTimeDataViewType || itemType is DateTimeOffsetDataViewType || itemType is TimeSpanDataViewType)
+                        data = new TypeConvertingTransformer(env, origStratCol, DataKind.Int64, origStratCol).Transform(data);
+
+                    var localSeed = seed.HasValue ? seed : ((ISeededEnvironment)env).Seed.HasValue ? ((ISeededEnvironment)env).Seed : null;
+                    var columnOptions =
+                        localSeed.HasValue ?
+                        new HashingEstimator.ColumnOptions(samplingKeyColumn, origStratCol, 30, (uint)localSeed.Value, combine: true) :
+                        new HashingEstimator.ColumnOptions(samplingKeyColumn, origStratCol, 30, combine: true);
```
harishsk: Similar comment as before. If the 30bits hashing is standard, can we define it as some global constant somewhere?

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -247,6 +262,8 @@ private Delegate GetGetterCore(DataViewRow input, int iinfo, out Action disposer
             var srcType = input.Schema[srcCol].Type;
             if (!(srcType is VectorDataViewType vectorType))
                 return ComposeGetterOne(input, iinfo, srcCol, srcType);
+            if (_columns[iinfo].Combine)
+                return ComposeGetterCombined(input, iinfo, srcCol, vectorType);
             return ComposeGetterVec(input, iinfo, srcCol, vectorType);
```
harishsk: We need to figure out the onnx export mechanism for this.

yaeldekel: For now I'm making this case non-exportable.

---
In reply to: [386741545](https://github.com/dotnet/machinelearning/pull/4828#discussion_r386741545) [](ancestors = 386741545)

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -124,17 +130,18 @@ private static VersionInfo GetVersionInfo()
                 modelSignature: "HASHTRNS",
                 // verWrittenCur: 0x00010001, // Initial
                 // verWrittenCur: 0x00010002, // Invert hash key values, hash fix
-                verWrittenCur: 0x00010003, // Uses MurmurHash3_x86_32
-                verReadableCur: 0x00010003,
+                //verWrittenCur: 0x00010003, // Uses MurmurHash3_x86_32
+                verWrittenCur: 0x00010004, // Add Combine option
+                verReadableCur: 0x00010004,
```
harishsk, line 135: We still haven't released a version with MurmurHash3. Do we need to increment this if we make sure to include it in the v1.5 release?

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -124,17 +130,18 @@ private static VersionInfo GetVersionInfo()
...
                 loaderSignature: LoaderSignature,
                 loaderAssemblyName: typeof(HashingTransformer).Assembly.FullName);
         }
 
-        private readonly HashingEstimator.ColumnOptionsInternal[] _columns;
+        private readonly HashingEstimator.ColumnOptions[] _columns;
         private readonly VBuffer<ReadOnlyMemory<char>>[] _keyValues;
         private readonly VectorDataViewType[] _kvTypes;
-        private readonly bool _isVersion1;
+        private readonly bool _useOldHashing;
```
harishsk, line 145: suggestion: Is it possible to come up with a better name? The only one I can think of is "_useOnnxCompatibleHashing" which would invert the meaning of the current name. Please use a a more descriptive name if possible.

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -1112,28 +1354,32 @@ private void AddMetaKeyValues(int i, DataViewSchema.Annotations.Builder builder)
...
 
-                opType = "MurmurHash3";
-                string murmurOutput = ctx.AddIntermediateVariable(_dstTypes[iinfo], "MurmurOutput");
-                var srcType = _srcTypes[iinfo].GetItemType().RawType;
-                if (_srcTypes[iinfo] is KeyDataViewType)
+                var srcType = _srcTypes[iinfo].GetItemType();
+                if (srcType is KeyDataViewType)
+                    return false;
+                if (_parent._columns[iinfo].Combine)
                     return false;
```
harishsk, line 1365: @lynx1820 Any thoughts why we are not supporting key types?

Lynx1820: Thanks for the remainder! I'll work on adding support for keytypes. Zero value keys don't get mapped to a hash but map to zero, a behavior not currently present.

yaeldekel: As @lynx1820 mentioned, in key types 0 is mapped to 0 instead of to its hash value, so it would be more complicated to export to ONNX.

---
In reply to: [424584542](https://github.com/dotnet/machinelearning/pull/4828#discussion_r424584542) [](ancestors = 424584542)

src/Microsoft.ML.Data/Transforms/Hashing.cs
yaeldekel: Technically users may create and save models containing this transformer once the PR that changed the version to 0x00010003 was merged. This most likely didn't happen, but to be on the safe side I think it's better to increment the version again. 

---
In reply to: [424565803](https://github.com/dotnet/machinelearning/pull/4828#discussion_r424565803) [](ancestors = 424565803)

test/Microsoft.ML.Tests/Transformers/HashTests.cs
```diff
@@ -124,7 +125,26 @@ public void TestOldSavingAndLoading()
             }
         }
 
-        private void HashTestCore<T>(T val, PrimitiveDataViewType type, uint expected, uint expectedOrdered, uint expectedOrdered3)
+        [Fact]
+        public void TestOldFormatModel()
+        {
+            string dataPath = GetDataPath("breast-cancer.txt");
```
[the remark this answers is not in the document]
yaeldekel: I think it would be better to increment the version again as I explained in the other comment, so I will leave this test.
(the existing test may be enough because the actual serialization format didn't change between versions 0x00010002 and 0x00010003, but I think it's better to have too much testing than too little testing :-). I did change he version of the old model from 0x00010002 to 0x00010003, so the tests are not identical). 

---
In reply to: [424674791](https://github.com/dotnet/machinelearning/pull/4828#discussion_r424674791) [](ancestors = 424674791)

src/Microsoft.ML.Data/Transforms/Hashing.cs
yaeldekel: I would like to keep it a variable that would have the value "false" by default, that's why I didn't want to change it to `_useOnnxCompatibleHashing`. Is this name that I changed it to acceptable?

---
In reply to: [424567216](https://github.com/dotnet/machinelearning/pull/4828#discussion_r424567216) [](ancestors = 424567216)

src/Microsoft.ML.Data/Commands/CrossValidationCommand.cs
yaeldekel: I am actually not sure why 30 was chosen. If you think it's important I can try changing it back to 31.

---
In reply to: [386729669](https://github.com/dotnet/machinelearning/pull/4828#discussion_r386729669) [](ancestors = 386729669)

harishsk: I have no reason to say it should change to 31 either :-). But since it is is used in a couple of places, can you please make it a named constant?

---
In reply to: [425102125](https://github.com/dotnet/machinelearning/pull/4828#discussion_r425102125) [](ancestors = 425102125,386729669)

harishsk:
That sounds good to me.

---
In reply to: [586994995](https://github.com/dotnet/machinelearning/pull/4828#issuecomment-586994995) [](ancestors = 586994995)

PR #4829, opened 2020-02-12
fix issue in WaiterWaiter caused by race condition

frank-dong-ms:
1. Fix issue in WaiterWaiter caused by race condition.
2. Re-enable affected tests as below:
EntryPointEvaluateRegression
EntryPointSdcaBinary
EntryPointSDCARegression
EntryPointChainedCrossValMacros
TestCrossValidationMacroWithStratification
TestOvaMacro
TrainAndPredictOnIris
CrossValidationIris
LinearClassifierTest
SavePipeTrainAndScoreFccTransformStr
BinaryClassifierLogisticRegressionNonNegativeTest
InpsectLinearModelParameters
TrainRegressionModel
CustomTransformer
TrainAndPredictIrisModelTest
LightGbmBinaryClassificationOnnxConversionTest

### DIFF
--- a/src/Microsoft.ML.Data/DataView/CacheDataView.cs
+++ b/src/Microsoft.ML.Data/DataView/CacheDataView.cs
@@ -683,12 +683,10 @@ namespace Microsoft.ML.Data
                 foreach (var w in _waiters)
                     w.Wait(pos);
 
-                // Back ground thread is filling data view and change the _rowCount from -1 to real row count.
-                // Original is: return pos < _parent._rowCount || _parent._rowCount == -1;
-                // In some case the change to _rowCount happens
-                // after pos < _parent._rowCount finish but before _parent._rowCount == -1
-                // thus cause flakyness
-                return _parent._rowCount == -1 || pos < _parent._rowCount;
+                // _parent._rowCount may or may not be initialized at this point. Only read the value once
+                // to avoid race conditions.
+                var rowCount = _parent._rowCount;
+                return rowCount == -1 || pos < rowCount;
             }
 
             public static Wrapper Create(CacheDataView parent, Func<int, bool> pred)
--- a/src/Microsoft.ML.Data/DataView/CacheDataView.cs
+++ b/src/Microsoft.ML.Data/DataView/CacheDataView.cs
@@ -682,7 +682,13 @@ namespace Microsoft.ML.Data
             {
                 foreach (var w in _waiters)
                     w.Wait(pos);
-                return pos < _parent._rowCount || _parent._rowCount == -1;
+
+                // Back ground thread is filling data view and change the _rowCount from -1 to real row count.
+                // Original is: return pos < _parent._rowCount || _parent._rowCount == -1;
+                // In some case the change to _rowCount happens
+                // after pos < _parent._rowCount finish but before _parent._rowCount == -1
+                // thus cause flakyness
+                return _parent._rowCount == -1 || pos < _parent._rowCount;
             }
 
             public static Wrapper Create(CacheDataView parent, Func<int, bool> pred)
--- a/src/Microsoft.ML.Data/DataView/CacheDataView.cs
+++ b/src/Microsoft.ML.Data/DataView/CacheDataView.cs
@@ -609,8 +609,9 @@ namespace Microsoft.ML.Data
 
             private TrivialWaiter(CacheDataView parent)
             {
-                Contracts.Assert(parent.GetRowCount().HasValue);
-                _lim = parent.GetRowCount().Value;
+                var rowCount = parent.GetRowCount();
+                Contracts.Assert(rowCount.HasValue);
+                _lim = rowCount.Value;
             }
 
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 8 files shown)


src/Microsoft.ML.Data/DataView/CacheDataView.cs
```diff
@@ -682,7 +682,11 @@ public bool Wait(long pos)
             {
                 foreach (var w in _waiters)
                     w.Wait(pos);
-                return pos < _parent._rowCount || _parent._rowCount == -1;
+
+                // _parent._rowCount may or may not be initialized at this point. Only read the value once
+                // to avoid race conditions.
+                var rowCount = _parent._rowCount;
+                return rowCount == -1 || pos < rowCount;
             }
```
harishsk: Great job figuring this out!! 
As discussed over chat, please move this fix to inside GetRowCount() and use GetRowCount everywhere throughout the file wherever _rowCount is used outside the CacheDataViewClass.

src/Microsoft.ML.Data/DataView/CacheDataView.cs
```diff
@@ -682,7 +686,7 @@ public bool Wait(long pos)
             {
                 foreach (var w in _waiters)
                     w.Wait(pos);
-                return pos < _parent._rowCount || _parent._rowCount == -1;
+                return !_parent.GetRowCount().HasValue || pos < _parent.GetRowCount().Value;
```
harishsk: I think we should call GetRowCount only once here too. Otherwise there may be an issue between the first and second calls.

src/Microsoft.ML.Data/DataView/CacheDataView.cs
```diff
@@ -1419,8 +1423,8 @@ public ImplOne(CacheDataView parent, DataViewRowCursor input, int srcCol, Ordere
                     : base(parent, input, srcCol, waiter)
                 {
                     _getter = input.GetGetter<T>(input.Schema[srcCol]);
-                    if (parent._rowCount >= 0)
-                        _values = new T[(int)parent._rowCount];
+                    if (parent.GetRowCount().HasValue)
```
sharwell:
```suggestion
                    if (parent.GetRowCount() is { } rowCount)
```

src/Microsoft.ML.Data/DataView/CacheDataView.cs
```diff
@@ -1419,8 +1423,8 @@ public ImplOne(CacheDataView parent, DataViewRowCursor input, int srcCol, Ordere
...
+                        _values = new T[(int)parent.GetRowCount().Value];
```
sharwell:
```suggestion
                        _values = new T[rowCount];
```

src/Microsoft.ML.Data/DataView/CacheDataView.cs
```diff
@@ -605,8 +609,8 @@ private sealed class TrivialWaiter : IWaiter
 
             private TrivialWaiter(CacheDataView parent)
             {
-                Contracts.Assert(parent._rowCount >= 0);
-                _lim = parent._rowCount;
+                Contracts.Assert(parent.GetRowCount().HasValue);
+                _lim = parent.GetRowCount().Value;
```
sharwell: The multiple reads here is not incorrect, but having them means it takes more time to think about these lines. I would recommend using a single call to `GetRowCount()` for this method.

src/Microsoft.ML.Data/DataView/CacheDataView.cs
frank-dong-ms: This fix works as we have adjust the order of these 2 conditions.
if GetRowCount() return null which means the row count has not been set, return true here, otherwise compare the value of RowCount with position. 

So no issue with this implementation and this read natural as we first check whether RowCount has value then compare value if necessary.

---
In reply to: [378696394](https://github.com/dotnet/machinelearning/pull/4829#discussion_r378696394) [](ancestors = 378696394)

src/Microsoft.ML.Data/DataView/CacheDataView.cs
frank-dong-ms: Thanks Sam, interesting gramma

---
In reply to: [378953044](https://github.com/dotnet/machinelearning/pull/4829#discussion_r378953044) [](ancestors = 378953044)

harishsk: I am not familiar with this syntax. Can one of you please explain?

---
In reply to: [379041295](https://github.com/dotnet/machinelearning/pull/4829#discussion_r379041295) [](ancestors = 379041295,378953044)

sharwell: It's equivalent to `parent.GetRowCount() is int rowCount`, but infers the type of `int` from the left hand side. It is _not_ the same as `is var rowCount`, because `is var` does not perform a null check.

src/Microsoft.ML.Data/DataView/CacheDataView.cs
sharwell: I'm in favor of single-call here for readability, even though both are correct.

harishsk: I agree. 

---
In reply to: [379060382](https://github.com/dotnet/machinelearning/pull/4829#discussion_r379060382) [](ancestors = 379060382)

frank-dong-ms: Ok, I will made the change as suggested

---
In reply to: [379061447](https://github.com/dotnet/machinelearning/pull/4829#discussion_r379061447) [](ancestors = 379061447,379060382)

PR #4831, opened 2020-02-13
Added exception for non-existent directory in ImageLoader

mstfbl:
Path.GetFullPath("C://non-Existant) was not throwing an exception. Hence there is a need to explicitly check for the existence of the imageFolder directory. Related to #4429 .


src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
```diff
@@ -103,9 +103,12 @@ internal ImageLoadingTransformer(IHostEnvironment env, string imageFolder = null
         internal ImageLoadingTransformer(IHostEnvironment env, string imageFolder = null, bool type = true, params (string outputColumnName, string inputColumnName)[] columns)
             : base(Contracts.CheckRef(env, nameof(env)).Register(nameof(ImageLoadingTransformer)), columns)
         {
-            // Throws ArgumentException if given imageFolder path is invalid or empty. Note: imageFolder may be null in this case.
+            // Throws System.InvalidOperationException if given imageFolder path is invalid or empty. Note: imageFolder may be null in this case.
```
harishsk: Isn't ArgumentException a better exception?

PR #4849, opened 2020-02-17
Fixed output schema of OnnxTransformer

### LINKED ISSUE #2980: ONNX outputs don't match model outputs after serialization

### LINKED ISSUE #2981: Cannot fit an Onnx Transform as part of a pipeline

harishsk:
This fixes a long standing issue about Onnx in ML.NET that the output variables of the onnx graph show up as columns with different names in the output schema of OnnxTransformer. After much investigation, it turns out that the output variables in the Onnx graph cannot be fixed because it is not possible to specify in onnx which particular node you want your output variable to be connected to. 

This fix, adds a known ".output" suffix to all Onnx models exported from ML.NET. And when reading a model, if the model was exported from ML.NET, it recognizes the suffix, strips it and maps it back to the input column. 

As a result of fixing this PR also contains updates to the tests and baseline files. 

Fixes #2980. Fixes #2981

### DIFF
--- a/test/Microsoft.ML.Functional.Tests/ONNX.cs
+++ b/test/Microsoft.ML.Functional.Tests/ONNX.cs
@@ -153,15 +153,10 @@ namespace Microsoft.ML.Functional.Tests
             var onnxEstimator = mlContext.Transforms.ApplyOnnxModel(modelPath);
             var onnxModel = onnxEstimator.Fit(data);
 
-            // TODO #2980: ONNX outputs don't match the outputs of the model, so we must hand-correct this for now.
-            // TODO #2981: ONNX models cannot be fit as part of a pipeline, so we must use a workaround like this.
-            var onnxWorkaroundPipeline = onnxModel.Append(
-                mlContext.Transforms.CopyColumns("Score", "Score").Fit(onnxModel.Transform(data)));
-
             // Create prediction engine and test predictions.
             var originalPredictionEngine = mlContext.Model.CreatePredictionEngine<HousingRegression, ScoreColumn>(model);
             // TODO #2982: ONNX produces vector types and not the original output type.
-            var onnxPredictionEngine = mlContext.Model.CreatePredictionEngine<HousingRegression, VectorScoreColumn>(onnxWorkaroundPipeline);
+            var onnxPredictionEngine = mlContext.Model.CreatePredictionEngine<HousingRegression, VectorScoreColumn>(onnxModel);
 
             // Take a handful of examples out of the dataset and compute predictions.
             var dataEnumerator = mlContext.Data.CreateEnumerable<HousingRegression>(mlContext.Data.TakeRows(data, 5), false);
--- a/test/Microsoft.ML.Functional.Tests/ONNX.cs
+++ b/test/Microsoft.ML.Functional.Tests/ONNX.cs
@@ -48,16 +48,11 @@ namespace Microsoft.ML.Functional.Tests
                 mlContext.Model.ConvertToOnnx(model, data, file);
 
             // Load the model as a transform.
-            // Note that when saving an ML.NET model as an ONNX model, the column types and column names will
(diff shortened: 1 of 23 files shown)


test/Microsoft.ML.Functional.Tests/ONNX.cs
```diff
@@ -109,7 +104,7 @@ public void SaveOnnxModelLoadAndScoreKMeans()
             // TODO #2980: ONNX outputs don't match the outputs of the model, so we must hand-correct this for now.
             // TODO #2981: ONNX models cannot be fit as part of a pipeline, so we must use a workaround like this.
```
ganik: Are these comments still valid? Do we do hand correction now?

test/Microsoft.ML.Functional.Tests/ONNX.cs
```diff
@@ -109,7 +104,7 @@ public void SaveOnnxModelLoadAndScoreKMeans()
...
             var onnxWorkaroundPipeline = onnxModel.Append(
-                mlContext.Transforms.CopyColumns("Score", "Score.onnx").Fit(onnxModel.Transform(data)));
+                mlContext.Transforms.CopyColumns("Score", "Score").Fit(onnxModel.Transform(data)));
```
ganik: >CopyColumns("Score", "Score") [](start = 37, length = 29)

Do we need CopyColumn now?

test/Microsoft.ML.Functional.Tests/ONNX.cs
harishsk: The comments are not valid. I have removed them in the next commit. 

---
In reply to: [381300021](https://github.com/dotnet/machinelearning/pull/4849#discussion_r381300021) [](ancestors = 381300021)

test/Microsoft.ML.Functional.Tests/ONNX.cs
harishsk: We don't need CopyColumns either. 

---
In reply to: [381300171](https://github.com/dotnet/machinelearning/pull/4849#discussion_r381300171) [](ancestors = 381300171)

PR #4855, opened 2020-02-19
Debugging for SSAForecast

mstfbl:


mstfbl:
For MatrixFactorizationSimpleTrainAndPredict(), I compare the results of the calculated leftMatrix and rightMatrix. The results of these matrices are different on working Linux builds vs. not working Linux builds.

mstfbl:
Update for MatrixFactorizationSimpleTrainAndPredict(): I tracked down the source of different results to [this line](https://github.com/dotnet/machinelearning/blob/8d1809eaa1c96de230c35445e9ac47835ebc5161/src/Microsoft.ML.Recommender/SafeTrainingAndModelBuffer.cs#L339). which leads to [this function](https://github.com/dotnet/machinelearning/blob/8d1809eaa1c96de230c35445e9ac47835ebc5161/src/Native/MatrixFactorizationNative/UnmanagedMemory.cpp#L79) (native C++ code with lots of memory operations). Two different values are the leftFactorMatrix and rightFactorMatrices which are set [here](https://github.com/dotnet/machinelearning/blob/8d1809eaa1c96de230c35445e9ac47835ebc5161/src/Microsoft.ML.Recommender/MatrixFactorizationPredictor.cs#L92). 

In addition, I noticed when we run this test 1000 times, on our builds it either always fails 1000 times or fails 1000 times. All of this makes me think that there may be other operations happening that may be affecting the memory addresses reserved for this MatrixFactorizationSimpleTrainAndPredict(), or some other kind of interference happening.

mstfbl:
Fix made for MatrixFactorizationSimpleTrainAndPredict() in MatrixFactorizationTrainer.PrepareBuffer() worked! This is because in the test we are hardcoding _threads = 1, we don't need to divide the training matrix into multiple blocks, which is what NrBins does.
This fix may also work for other similar out-of-range tests.

mstfbl:
MatrixFactorizationSimpleTrainAndPredict() test fixed with change in line 516 
of MatrixFactorizationTrainer.cs, where `NrBins` is set to 1 if the thread count is also 1. This works since in MatrixFactorizationSimpleTrainAndPredict() we set _threads = 1, and this way there is no need to divide the training matrix into more than 1 block.

mstfbl:
MulticlassTreeFeaturizedLRTest() is also passing tests with 1000 iterations.

mstfbl:
Now testing BinaryClassifierLogisticRegressionBinNormTest().

mstfbl:
TestOvaMacroWithUncalibratedLearner() is also passing!

mstfbl:
Unsure on EntryPointPipelineEnsembleGetSummary()

mstfbl:
The tests BinaryClassifierLogisticRegressionBinNormTest, BinaryClassifierSymSgdTest, MulticlassLRTest and SsaForecast are passing checks.

mstfbl:
Working on debugging BenchmarksProjectIsNotBroken on this draft PR.

mstfbl:
Now working on checking for NetCore 3.1 comparability & seeing if certain tests that are marked with "LessThanNetCore30OrNotNetCoreFact" would be functioning with NetCore3.1.

mstfbl:
Working on debugging SSAForecast tests

PR #4857, opened 2020-02-19
Added slot names support for OnnxTransformer

harishsk:
This PR adds support for persisting the SlotNames annotations of a column during onnx export and reading those back in OnnxTransformer and adding the annotations back to the column when the onnx model is read from disk.

Onnx natively does not have support for annotations. To work around this, we store some metadata in some unused portions of the graph. As an example, let us say we have an ML.NET model with an output column NGrams that outputs a vector of NGram counts. This column will have an Annotation in ML.NET named SlotNames. When this model is exported to onnx, we create an additional LabelEncoder node and store the SlotNames in the keys_strings attribute of the LabelEncoder.

The LabelEncoder is created with an input name of `$"mlnet.{column.Name}.unusedInput"`, an output name of $"mlnet.{column.Name}.unusedOutput" and a node name of `$"mlnet.{column.Name}.SlotNames"`. (All the actual output columns of the ML.NET model are suffixed with a `".output"` string)

Then when OnnxTransformer loads the graph it goes through the list of output nodes and creates output columns for each of them in its output schema. For each column it searches the graph for a node named `$"mlnet.{column.Name}.SlotNames"`. If it finds it, it reads the keys_strings attributes from that node and adds those strings as SlotNames annotation to that column. 

This SlotNames data should then be available as annotations on the column in both ML.NET and Nimbus.

### DIFF
--- a/src/Microsoft.ML.OnnxConverter/SaveOnnxCommand.cs
+++ b/src/Microsoft.ML.OnnxConverter/SaveOnnxCommand.cs
@@ -2,8 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using Google.Protobuf;
 using Microsoft.ML;
 using Microsoft.ML.Command;
@@ -188,7 +190,9 @@ namespace Microsoft.ML.Model.OnnxConverter
                 if (outputData.Schema[i].IsHidden)
                     continue;
 
-                var idataviewColumnName = outputData.Schema[i].Name;
+                var column = outputData.Schema[i];
+
+                var idataviewColumnName = column.Name;
 
                 // Since the last IDataView also contains columns of the initial IDataView, last IDataView's columns found in
                 // _inputToDrop should be removed too.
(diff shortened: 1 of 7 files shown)


test/BaselineOutput/Common/Onnx/BinaryClassification/BreastCancer/OneHotBagPipeline.txt
```diff
@@ -324,8 +324,8 @@
           {
             "name": "target_weights",
             "floats": [
-              0.50476193,
-              -0.97911227
+              0.504761934,
+              -0.979112267
```
yaeldekel: How come these changed?

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -1282,6 +1284,16 @@ public void WordTokenizerOnnxConversionTest()
...
+
+                    VBuffer<ReadOnlyMemory<char>> mlNetSlots = default;
+                    VBuffer<ReadOnlyMemory<char>> onnxSlots= default;
+                    transformedData.Schema[columnName].GetSlotNames(ref mlNetSlots);
+                    onnxResult.Schema[columnName].GetSlotNames(ref onnxSlots);
+                    Assert.Equal(mlNetSlots.Length, onnxSlots.Length);
+                    var mlNetSlotNames = mlNetSlots.DenseValues().ToList();
+                    var onnxSlotNames = onnxSlots.DenseValues().ToList();
+                    for (int j = 0; j < mlNetSlots.Length; j++)
+                        Assert.Equal(mlNetSlotNames[j].ToString(), onnxSlotNames[j].ToString());
```
yaeldekel: >Equal [](start = 31, length = 5)

nit: I think Assert.Equal also has an overload for IEnumerables.

src/Microsoft.ML.OnnxConverter/SaveOnnxCommand.cs
```diff
@@ -204,11 +208,39 @@ internal static void GetPipe(OnnxContextImpl ctx, IChannel ch, IDataView end, ou
...
+            string labelEncoderNodeName = $"mlnet.{column.Name}.SlotNames";
+
+            string[] oneVals = new string[] { "one" };
+            long[] dims = new long[] { 1, 1 };
+            var one = ctx.AddInitializer(oneVals, dims, labelEncoderNodeName);
+
+            var labelEncoderOutput = ctx.AddIntermediateVariable(NumberDataViewType.Int64, labelEncoderOutputName, true);
+            var node = ctx.CreateNode(opType, one, labelEncoderOutput, labelEncoderNodeName);
+            node.AddAttribute("keys_strings", slotNamesAsStrings);
+            node.AddAttribute("values_int64s", Enumerable.Range(0, slotNames.Length).Select(x => (long)x));
```
yaeldekel: >values_int64s [](start = 31, length = 13)

Why do we need this?
Does it even work when slot names are not numbers?

harishsk: These are unused. But are specified only to satisfy ORT. 

---
In reply to: [381867237](https://github.com/dotnet/machinelearning/pull/4857#discussion_r381867237) [](ancestors = 381867237)

test/BaselineOutput/Common/Onnx/BinaryClassification/BreastCancer/OneHotBagPipeline.txt
harishsk: Not sure. But I see this occurring off and on that the baselines numbers change when we run them locally. I am ignoring them because the change is in the 7th decimal place

---
In reply to: [381858622](https://github.com/dotnet/machinelearning/pull/4857#discussion_r381858622) [](ancestors = 381858622)

test/Microsoft.ML.Tests/OnnxConversionTest.cs
harishsk: I tried that already. But the Assert was firing even when all the strings were equal. Not sure why. 

---
In reply to: [381859560](https://github.com/dotnet/machinelearning/pull/4857#discussion_r381859560) [](ancestors = 381859560)

PR #4861, opened 2020-02-20
Fix for ColumnConcatenatingTransformer

Lynx1820:
Used the Concat operator instead of FeatureVectorizer to avoid conflicting output type errors, since FeatureVectorizer outputs strictly floats, but some trainers expect double outputs.


test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -571,9 +572,23 @@ public void KeyToVectorWithBagOnnxConversionTest()
...
+            {
+                // Evaluate the saved ONNX model using the data used to train the ML.NET pipeline.
+                var onnxEstimator = mlContext.Transforms.ApplyOnnxModel(onnxModelPath);
+                var onnxTransformer = onnxEstimator.Fit(data);
+                var onnxResult = onnxTransformer.Transform(data);
+                CompareSelectedColumns<float>("Score", "Score", transformedData, onnxResult);
+                CompareSelectedColumns<float>("Probability", "Probability", transformedData, onnxResult);
+                CompareSelectedColumns<bool>("PredictedLabel", "PredictedLabel", transformedData, onnxResult);
+            }
+            else
```
harishsk: Can you please remove the else clause here and in other tests too? (I mean retain the file comparison, but do it always, even if onnx runtime is supported) I meant to do it earlier, but I forgot.
What is happening is that the code in the else clause doesn't get run on our local builds and we forget to update the baselines which leads to build breaks. 
If we always to the file comparison, then the baseline comparison tests will fail in our local builds as well and we will remember to update them.

PR #4866, opened 2020-02-20
Fix for KeytoValue transformer

Lynx1820:
Fix a small bug where the source variable name of an ONNX graph was mistakenly always the input variable name, which is not always the case. Other ONNX transformers don't explicitly test for this functionality, so I didn't add a test case.

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/KeyToValue.cs
+++ b/src/Microsoft.ML.Data/Transforms/KeyToValue.cs
@@ -565,12 +565,11 @@ namespace Microsoft.ML.Transforms
                 {
                     var info = _parent.ColumnPairs[iinfo];
                     var inputColumnName = info.inputColumnName;
-
                     if (!ctx.ContainsColumn(inputColumnName))
                         continue;
-
+                    string srcVariableName = ctx.GetVariableName(inputColumnName);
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Data/Transforms/KeyToValue.cs
```diff
@@ -568,9 +568,9 @@ public void SaveAsOnnx(OnnxContext ctx)
 
                     if (!ctx.ContainsColumn(inputColumnName))
                         continue;
-
+                    string srcVariableName = ctx.GetVariableName(inputColumnName);
                     var dstVariableName = ctx.AddIntermediateVariable(_types[iinfo], info.outputColumnName, true);
-                    if (!_kvMaps[iinfo].SaveOnnx(ctx, inputColumnName, dstVariableName))
+                    if (!_kvMaps[iinfo].SaveOnnx(ctx, srcVariableName, dstVariableName))
```
harishsk: I think it would be a good idea to add a test case specifically because we don't have tests for this. Also please verify this fixes the Nimbus issue.

Lynx1820: Sounds good. I restructured the tests and got rid of a KeyToValue test that seems redundant and is not formatted to test different types of values.

PR #4869, opened 2020-02-21
Re-enabling disabled tests due to wrong nrBins value

mstfbl:
Fixed case with erroneous SafeTrainingAndModelBuffer.nrBins value, where if _threads = 1, nrBins should also equal 1. This way, when thread count is 1, there is no need to divide the training matrix into multiple blocks as 1 will do fine. As a result, we can re-enable the following disabled tests:

- MatrixFactorizationSimpleTrainAndPredict
- MulticlassTreeFeaturizedLRTest
- TestOvaMacroWithUncalibratedLearner
- EntryPointPipelineEnsembleGetSummary

### DIFF
--- a/test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
+++ b/test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
@@ -2124,7 +2124,6 @@ namespace Microsoft.ML.RunTests
 
         [LessThanNetCore30OrNotNetCoreFact("netcoreapp3.0 output differs from Baseline")]
         //Skipping test temporarily. This test will be re-enabled once the cause of failures has been determined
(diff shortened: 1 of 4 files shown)


test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -2124,7 +2124,6 @@ public void EntryPointMulticlassPipelineEnsemble()
 
         [LessThanNetCore30OrNotNetCoreFact("netcoreapp3.0 output differs from Baseline")]
         //Skipping test temporarily. This test will be re-enabled once the cause of failures has been determined
```
sharwell: 💡 This comment can also be deleted for each enabled test

sharwell:
Glad this worked even if it was only a partial fix!

PR #4875, opened 2020-02-21
Alternate solution for ColumnConcatenatingTransformer

Lynx1820:
- Added test isolating ColumnConcatenatingTransformer


test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -877,6 +877,36 @@ public void LoadingPredictorModelAndOnnxConversionTest()
             Done();
         }
 
+        [Fact]
```
harishsk: Looks like you will need to update baselines.

PR #4876, opened 2020-02-21
Incorporating varying tolerances through VaryingToleranceAttribute

mstfbl:
In this PR we are adding the ability to have varying tolerances for tests through VaryingToleranceAttribute.

The VaryingToleranceAttribute has been added to the following tests:

- MulticlassLRTest
- BinaryClassifierLogisticRegressionBinNormTest


test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
```diff
@@ -53,11 +54,14 @@ public void AutoFitMultiTest()
...
 
-        [TensorFlowFact]
+        // [TensorFlowFact]
+        [Theory, VaryingTolerance(3)]
         //Skipping test temporarily. This test will be re-enabled once the cause of failures has been determined
-        [Trait("Category", "SkipInCI")]
-        public void AutoFitImageClassificationTrainTest()
+        public void AutoFitImageClassificationTrainTest(int tolerance)
         {
+            // Replacing TensorFlowFact
```
harishsk: Can we add some log message here and in all instances where we are replacing a prior fact with a separate check in code?

test/Microsoft.ML.Predictor.Tests/TestPredictors.cs
```diff
@@ -146,14 +146,17 @@ public void EarlyStoppingTest()
...
         [TestCategory("Logistic Regression")]
         //Skipping test temporarily. This test will be re-enabled once the cause of failures has been determined
-        [Trait("Category", "SkipInCI")]
-        public void MulticlassLRTest()
+        public void MulticlassLRTest(int tolerance)
         {
-            RunOneAllTests(TestLearners.multiclassLogisticRegression, TestDatasets.iris, digitsOfPrecision: 4);
+            // Replacing LessThanNetCore30OrNotNetCoreFactAttribute
+            if (AppDomain.CurrentDomain.GetData("FX_PRODUCT_VERSION") != null)
+                return;
```
harishsk: Same here too. Please try to preserve the log message in the Fact that you replaced.

test/Microsoft.ML.TestFrameworkCommon/Attributes/VaryingToleranceAttribute.cs
```diff
@@ -0,0 +1,37 @@
...
+        public VaryingToleranceAttribute(int tolerance)
+        {
+            Tolerance = tolerance;
+        }
+
+        public int Tolerance { get; }
+        public override IEnumerable<object[]> GetData(MethodInfo testMethod)
+        {
+            Console.WriteLine(String.Format("Test \"{0}\" utilizes varying tolerances.", testMethod.Name));
+            yield return new object[] { Tolerance };
```
harishsk: Why yield? Why not directly return an enumerable?

test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
```diff
@@ -121,16 +120,13 @@ public void MatrixFactorizationSimpleTrainAndPredict()
 
             // Compute prediction errors
             var metrices = mlContext.Recommendation().Evaluate(prediction, labelColumnName: labelColumnName, scoreColumnName: scoreColumnName);
-
+            var toleranceDelta = Math.Pow(10, -1 * tolerance);
```
harishsk: Would it be better to return the toleranceDelta directly from the GetData method? The user can specify the tolerance as an integer of number of digits. And GetData can return the converted value directly.

harishsk: It appears you need integer values in some places. Can you find a way to standardize the comparison?

test/Microsoft.ML.TestFrameworkCommon/Attributes/VaryingToleranceAttribute.cs
mstfbl: Hi Harish. I utilized [this](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/yield)documentation in my original implementation and the pre-existing usage of [GetData in IterationDataAttribute](https://github.com/dotnet/machinelearning/blob/4493397fc35060d7d04c9333bf00164d320a6621/test/Microsoft.ML.TestFrameworkCommon/Attributes/IterationDataAttribute.cs#L29-L36), but now I realize with your review thatI'm not iterating when calling GetData, I can just return `(IEnumerable<object[]>)new object[] { Tolerance }`. Thank you!

mstfbl: Hi Harish. I utilized [this](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/yield) documentation in my original implementation and the pre-existing usage of [GetData in IterationDataAttribute](https://github.com/dotnet/machinelearning/blob/4493397fc35060d7d04c9333bf00164d320a6621/test/Microsoft.ML.TestFrameworkCommon/Attributes/IterationDataAttribute.cs#L29-L36). I attempted to return `(IEnumerable<object[]>)new object[] { Tolerance }` instead of using `yield`, but I get the following exception:

>     System.InvalidCastException : Unable to cast object of type 'System.Collections.Generic.List`1[System.Int32]' to type 'System.Collections.Generic.IEnumerable`1[System.Object[]]'.

By using `yield` I am able to bypass this restriction. Is there something I'm overlooking here?

test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
mstfbl: Hi Harish. Indeed I do need integer values in most places where `VaryingTolerance` is used, which is why I made this `Math.Pow(10, -1 * tolerance)` change in `MatrixFactorizationSimpleTrainAndPredict`. Using `Assert.Equal(double expected, double actual, int precision)` is a nice way to standardize this comparison, however the precision metric here doesn't work well when `double actual` is rounded up or down.

For example, let us have:
`double expected = 0.444445`
`double actual = 0.444444`
`int precision = 6`

Here, `expected` and `actual` are within 1e-6 of accuracy. However, with `Assert.Equal`, `expected` is rounded up to 0.444450, and `actual` is rounded down to 0.444440. As a result, `Assert.Equal` returns `false`, when actually it should return `true`. This is why I went with this approach here. What would you recommend in this case @harishsk ?

test/Microsoft.ML.Predictor.Tests/TestPredictors.cs
```diff
@@ -146,14 +146,19 @@ public void EarlyStoppingTest()
         /// <summary>
         /// Multiclass Logistic Regression test.
         /// </summary>
-        [LessThanNetCore30OrNotNetCoreFact("netcoreapp3.0 output differs from Baseline")]
+        //[LessThanNetCore30OrNotNetCoreFact("netcoreapp3.0 output differs from Baseline")]
+        [Theory, VaryingTolerance(4)]
```
harishsk: This is using the same tolerance as before. Maybe this test doesn't need the VaryingTolerance attribute. Please try enabling the test in CI builds without this attribute first.

test/Microsoft.ML.Predictor.Tests/TestPredictors.cs
```diff
@@ -274,10 +278,18 @@ public void BinaryClassifierLogisticRegressionTest()
             Done();
         }
 
-        [X64Fact("x86 output differs from Baseline")]
+        //[X64Fact("x86 output differs from Baseline")]
+        [Theory, VaryingTolerance(4)]
```
harishsk: Same comment as above for same precision value.

test/Microsoft.ML.Predictor.Tests/TestPredictors.cs
```diff
@@ -146,14 +146,19 @@ public void EarlyStoppingTest()
...
         /// Multiclass Logistic Regression test.
         /// </summary>
-        [LessThanNetCore30OrNotNetCoreFact("netcoreapp3.0 output differs from Baseline")]
+        // [LessThanNetCore30OrNotNetCoreFact("netcoreapp3.0 output differs from Baseline")]
         [TestCategory("Multiclass")]
         [TestCategory("Logistic Regression")]
-        //Skipping test temporarily. This test will be re-enabled once the cause of failures has been determined
-        [Trait("Category", "SkipInCI")]
-        public void MulticlassLRTest()
+        [Theory, VaryingTolerance(3)]
```
harishsk: Have you run these tests with IterationData?
For this test and the one below, how often do we see failures if we retain the tolerance at the original level? And if we reduce the tolerance by one as you have done, do the failures disappear completely?

mstfbl: These tests passed in this [build ](https://github.com/dotnet/machinelearning/pull/4855/commits/6062f629f49e1ff9b5fd54f373343c5ff452ef4f) with 1000 iterations. Therefore I'm removing the VartingTolerance Theory tag from these tests.

mstfbl:
Closing this PR, to open when `VaryingToleranceAttribute`is needed in the future.

PR #4881, opened 2020-02-24
Slightly simplified version of adding KeyToValue for onnx export

harishsk:
This is a variation of @antoniovs1029 's fix. This adds KeyToValueMappingTransformer in only one place. The results are identical between the two.

### DIFF
--- a/src/Microsoft.ML.OnnxConverter/SaveOnnxCommand.cs
+++ b/src/Microsoft.ML.OnnxConverter/SaveOnnxCommand.cs
@@ -15,6 +15,7 @@ using Microsoft.ML.EntryPoints;
 using Microsoft.ML.Internal.Utilities;
 using Microsoft.ML.Model.OnnxConverter;
 using Microsoft.ML.Runtime;
+using Microsoft.ML.Transforms;
 using Newtonsoft.Json;
 using static Microsoft.ML.Model.OnnxConverter.OnnxCSharpToProtoWrapper;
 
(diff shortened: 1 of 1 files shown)


antoniovs1029:
I think the only difference between this and my fix is that in here all of the non-hidden columns at the output schema, which are KeyDataViewTypes, will be mapped back to values.

In my other PR this doesn't happen for all columns, only for PredictedLabel and for columns that had the same name and where KeyDataViewTypes both in the input and output.

A scenario were my PR and this PR would be different, is one where the user actually wants to apply a `ToKey` in his NimbusML pipeline, and actually expects to see only keys in his output (from ONNX or otherwise). In this scenario I believe your PR here would automatically add a KeyToValueTransformer for that column, whereas my PR wouldn't, since the user actually wants to see keys. Gani was the one who mentioned this scenario to me some days ago, not sure if it's still something we would want to avoid.

ganik:
If ToKey applied on string and its output is among output columns of pipeline then Nimbus in python gives back categorical type, which is a tuple (string Value, int/uint Key). In ONNX we dont have categoricals so we can output either Value or a Key. Value is useful for presentation layer, like doing print etc. If there is a downstream learner that consumes ToKey output or we to stitch 2 ONNX models then its done off Key, while Value is ignored. I would sacrifice presentational purposes for functional ones, and would return Key instead of Value in ONNX models. PredictedLabel is the only exception as its usually final output and (ignoring stacking) used for presentation purposes only.

harishsk:
This PR is superseded by #4841 . Closing.

PR #4885, opened 2020-02-25
Typo corrections

mstfbl:
More typo corrections in this repository.
The typo corrects are almost all in comments, except for one variable which was misspelled.
We should really pay more attention to avoid misspellings, and perhaps do a periodical check to make sure no new mistakes are made.

### DIFF
--- a/test/Microsoft.ML.CodeAnalyzer.Tests/Code/ContractsCheckTest.cs
+++ b/test/Microsoft.ML.CodeAnalyzer.Tests/Code/ContractsCheckTest.cs
@@ -128,7 +128,7 @@ namespace TestNamespace
                         VerifyCS.Diagnostic(ContractsCheckAnalyzer.NameofDiagnostic.Rule).WithLocation(25, 53).WithArguments("CheckUserArg", "name", "\"sp\""),
                         new DiagnosticResult("CS0122", DiagnosticSeverity.Error).WithLocation("Test1.cs", 752, 24).WithMessage("'ICancelable' is inaccessible due to its protection level"),
                         new DiagnosticResult("CS0122", DiagnosticSeverity.Error).WithLocation("Test1.cs", 752, 67).WithMessage("'ICancelable.IsCanceled' is inaccessible due to its protection level"),
-                        new DiagnosticResult("CS1503", DiagnosticSeverity.Error).WithLocation("Test1.cs", 753, 91).WithMessage("Argument 2: cannot convert from 'Microsoft.ML.Runtime.IHostEnvironment' to 'Microsoft.ML.Runtime.IExceptionContext'"),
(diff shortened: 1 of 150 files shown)


test/Microsoft.ML.CodeAnalyzer.Tests/Code/ContractsCheckTest.cs
```diff
@@ -128,7 +128,7 @@ public async Task ContractsCheckFix()
                         VerifyCS.Diagnostic(ContractsCheckAnalyzer.NameofDiagnostic.Rule).WithLocation(25, 53).WithArguments("CheckUserArg", "name", "\"sp\""),
                         new DiagnosticResult("CS0122", DiagnosticSeverity.Error).WithLocation("Test1.cs", 752, 24).WithMessage("'ICancelable' is inaccessible due to its protection level"),
                         new DiagnosticResult("CS0122", DiagnosticSeverity.Error).WithLocation("Test1.cs", 752, 67).WithMessage("'ICancelable.IsCanceled' is inaccessible due to its protection level"),
-                        new DiagnosticResult("CS1503", DiagnosticSeverity.Error).WithLocation("Test1.cs", 753, 91).WithMessage("Argument 2: cannot convert from 'Microsoft.ML.Runtime.IHostEnvironment' to 'Microsoft.ML.Runtime.IExceptionContext'"),
+                        new DiagnosticResult("CS1503", DiagnosticSeverity.Error).WithLocation("Test1.cs", 753, 90).WithMessage("Argument 2: cannot convert from 'Microsoft.ML.Runtime.IHostEnvironment' to 'Microsoft.ML.Runtime.IExceptionContext'"),
```
antoniovs1029: Just curious, why is this changing?

tools-local/Microsoft.ML.InternalCodeAnalyzer/Utils.cs
```diff
@@ -19,7 +19,7 @@ public static bool NameIsGood(string name, int min, bool upper)
             // C# naming guidelines say, any initialism greater than two characters should not
             // be all upper cased. So: _readIOStream is good, and _readHttpStream is good. You
             // could imagine having two two-letter initialisms, like: _readIOUI, where you use
-            // two two character initialism, but I'm going to suppose that never happens since
+            // two character initialism, but I'm going to suppose that never happens since
```
antoniovs1029: I think this was correct the way it was before, after reading the whole comment

test/Microsoft.ML.CodeAnalyzer.Tests/Code/ContractsCheckTest.cs
mstfbl: I think it is because of the change here:

https://github.com/dotnet/machinelearning/blob/6dffa3280ef516c3607bc17c51844469848f2110/src/Microsoft.ML.Core/Utilities/Contracts.cs#L753

Here the misspelling *cancelled is corrected to "canceled," which has one fewer letter than its misspellings. Since this check is checking for the exact position of the error message (line 753, column 90), this correction brought this error message one character back.

PR #4888, opened 2020-02-25
Updates NetCoreApp 3.0 builds to NetCoreApp 3.1

### LINKED ISSUE #4865: move from net core 3.0 to 3.1 on CI test
as net core 3.0 will be out of support soon

mstfbl:
Fix #4865 . Updates our builds using NetCoreApp 3.0 to NetCoreApp 3.1.


test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -2123,7 +2123,7 @@ public void EntryPointMulticlassPipelineEnsemble()
             }
         }
 
-        [LessThanNetCore30OrNotNetCoreFact("netcoreapp3.0 output differs from Baseline")]
+        [LessThanNetCore30OrNotNetCoreFact("netcoreapp3.1 output differs from Baseline")]
```
frank-dong-ms: >netcoreapp3.1 output differs from Baseline" [](start = 44, length = 43)

Have you tried netcore3.1 also generates different result?

test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
```diff
@@ -4906,7 +4906,7 @@ public void TestSimpleTrainExperiment()
...
```
frank-dong-ms: >[LessThanNetCore30OrNotNetCoreFact("netcoreapp3.1 output differs from Baseline")] [](start = 8, length = 81)

same here

test/Microsoft.ML.Core.Tests/UnitTests/TestEntryPoints.cs
mstfbl: Hi @frank-dong-ms , good point, I'm now checking for this in one of my draft testing PRs.

mstfbl: In [this build](https://dev.azure.com/dnceng/public/_build/results?buildId=566312&view=logs&j=9d6f93fc-7103-540d-abb9-b79bad552b0d&t=ef9fb040-bca3-54d1-40bf-7394d88ead2f) where I ran these changes without the `LessThanNetCore30OrNotNetCoreFact`, we can see that the benchmarks are indeed different than the baselines.

.vsts-dotnet-ci.yml
```diff
@@ -13,20 +13,20 @@ resources:
 jobs:
 - template: /build/ci/job-template.yml
   parameters:
-    name: Centos_x64_NetCoreApp30
+    name: Centos_x64_NetCoreApp31
```
harishsk: I don't think we will support multiple versions of .netcoreapp 3.x. 
Can we simply make the whole thing NetCoreApp3?

harishsk:
@mstfbl It appears that upgrading to NetCoreApp31 will fix #4933 
Can you please address the comment above and verify if it really fixes that issue and close both?

.vsts-dotnet-ci.yml
mstfbl: That makes sense, though I do think that specifying which version of NetCoreApp3 (3.0 or 3.1) we're using would be helpful if in the future we see bugs that are specific to 3.0 or 3.1, as you pointed out with respect to #4933 .

I will update these references to version 3.0/3.1 to only have version "3", if you do believe that specifying the exact iteration of NetCoreApp3 won't be helpful as I mentioned above.

PR #4889, opened 2020-02-25
Added Onnx ValueToKey and KeyToValue support for more int types

Lynx1820:


src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs
```diff
@@ -789,38 +789,86 @@ private bool SaveAsOnnxCore(OnnxContext ctx, int iinfo, ColInfo info, string src
...
                     var t = InternalDataKindExtensions.ToInternalDataKind(DataKind.String).ToType();
                     castNode.AddAttribute("to", t);
                     node = ctx.CreateNode(opType, castOutput, labelEncoderOutput, ctx.GetNodeName(opType));
-                    var terms = GetTermsAndIds<long>(iinfo, out termIds);
+                    var terms = GetTermsAndIds<Int64>(iinfo, out termIds);
+                    node.AddAttribute("keys_strings", terms.Select(item => item.ToString()));
+                }
+                else if (type.Equals(NumberDataViewType.Int32))
+                {
+                    var castOutput = ctx.AddIntermediateVariable(TextDataViewType.Instance, "castOutput", true);
```
harishsk: I am approving the PR but I have a concern. 
Setting the last parameter to true skips adding type and shape information. That has been causing a lot of problems and also makes it harder to debug. If this PR fixes AveragedPerceptron then please go ahead and merge this. But please go back and use the default value of false to all AddIntermediateVariable calls here and figure out what errors we see and fix those.

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs
```diff
@@ -789,38 +789,86 @@ private bool SaveAsOnnxCore(OnnxContext ctx, int iinfo, ColInfo info, string src
...
                     var terms = GetTermsAndIds<double>(iinfo, out termIds);
                     node.AddAttribute("keys_floats", terms);
                 }
-                else if (info.TypeSrc.GetItemType().Equals(NumberDataViewType.Int64))
+                else if (type.Equals(NumberDataViewType.Int64))
                 {
                     // LabelEncoder doesn't support mapping int64 -> int64, so values are cast to strings
                     var castOutput = ctx.AddIntermediateVariable(TextDataViewType.Instance, "castOutput", true);
-                    castNode = ctx.CreateNode("Cast", srcVariableName, castOutput, ctx.GetNodeName(opType), "");
+                    castNode = ctx.CreateNode("Cast", srcVariableName, castOutput, ctx.GetNodeName("Cast"), "");
```
harishsk: Are the first five lines of each of these blocks the same? If so, can they be refactored better?
Even better, can you refactor all these into a single function that is a wrapper with a generic type around GetTermsAndIds?

src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs
Lynx1820: Done.

PR #4893, opened 2020-02-26
Debugging hanging AutoFitImageClassificationTrainTest

mstfbl:
Will be using this draft PR for general debugging purposes on CI

Notes:
Windows builds have 7,168 MBs of RAM


mstfbl:
Testing of `AutoFitImageClassificationTrainTest` is taking too long per test, so tesitng it with 1000 iterations isn't feasible.

mstfbl:
The tests AutoFitRecommendationTest and AutoFitRegressionTest are passing. AutoFitImageClassificationTrainTest is displaying errors every now and then.

mstfbl:
The reason why AutoFitImageClassificationTrainTest is crashing is after running `var result = context.Auto().CreateMulticlassClassificationExperiment(0).Execute(trainDataset, testDataset, columnInference.ColumnInformation)`, result.Best run might not always be updated. I saw this as I caught `result.BestRun` having a null value right when it is being called.

mstfbl:
The original bug with `AutoFitImageClassificationTrainTest ` is occuring due to the null returned value here:
https://github.com/Potapy4/dotnet-machinelearning/blob/acce35291a4fbd89b8bdacdcd12c4939ed6a88a6/src/Microsoft.ML.AutoML/Utils/BestResultUtil.cs#L61
For some reason, sometimes the validationMetrics of an IEnumerable<(RunDetail<T>) is null.

mstfbl:
There's an issue with this `Evaluation `function: 

https://github.com/dotnet/machinelearning/blob/f0a8a76a40348021b5a24f9b3027d53fff4a58cd/src/Microsoft.ML.Data/Evaluators/MulticlassClassificationEvaluator.cs#L506-L535

The returned `result `value can sometimes be null, which is what is causing `AutoFitImageClassificationTrainTest `to sometimes fail.

mstfbl:
I figured out the cause of the occasional crash of AutoFitImageClassificationTrainTest. When any exception occurs in RunnerUtil.TrainAndScorePipeline, instead of throwing the error, it is instead caught and ignored while a null metrics value (in line 49) is sent up through the call stack instead.

https://github.com/dotnet/machinelearning/blob/f0a8a76a40348021b5a24f9b3027d53fff4a58cd/src/Microsoft.ML.AutoML/Experiment/Runners/RunnerUtil.cs#L25-L49

mstfbl:
Update: AutoFitImageClassificationTrainTest with 100 iterations fail on Windows x64 builds with:

`System.Runtime.InteropServices.SEHException (0x80004005): External component has thrown an exception`
but also I get 

`System.FormatException: Tensorflow exception triggered while loading model. ---> System.OutOfMemoryException` 

in `Tensorflow.c_api.TF_SessionRun`, which is the C++-implementation of TensorFlow's training code. This seems related to Issue https://github.com/SciSharp/TensorFlow.NET/issues/485

mstfbl:
Will be using this draft PR for general debugging purposes on CI

mstfbl:
Will be using this PR to debug AutoFitImageClassificationTrainTest hanging occasionally on Windows builds.

mstfbl:
`AutoFitImageClassificationTrainTest ` is still occasionally hanging, mostly due to indisposed Tensorflow objects after the test is complete. I found this comment in Microsoft.ML.Vision/ImageClassificationTrainer.TrainModelCore to be of interest:

https://github.com/dotnet/machinelearning/blob/5d531d3660ed5972e3ab93ea5c8ff0b5bacae0a2/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs#L714-L721

mstfbl:
Adding the fix `(model as IDisposable)?.Dispose();` for freeing Tensor objects worked! This fix is necessary, as these Tensor objects made in the C TensorFlow libraries are not automatically cleaned up by C#'s Garbage Collector.

mstfbl:
Freeing Tensor objects in `model` in a finally statement in `TrainAndScorePipeline` works in fixing memory bug, and is safe to do when `model` is never saved in memory and written to disk always.

PR #4896, opened 2020-02-26
disable test parallelization for ML.Test assembly to avoid crash

frank-dong-ms:
In ML.Tests assembly, when some tests run in parallel, there are chance the host process will crash due to below exception: 
The thread tried to read from or write to a virtual address for which it does not have the appropriate access.
Unhandled exception at 0x00007FFA70E7B049 (ntdll.dll) in dotnet.exe.12884.dmp: 0xC0000374: A heap has been corrupted (parameters: 0x00007FFA70EE27F0).

Looked into this error in detail, they come from LightGBM/OnnxRuntime dll we are referencing, seems like null pointer error during object finalization. 
image

These crash issue can be mitigated if we disable test parallelization. At the meantime, I'm contacting LightGBM and OnnxRuntime team to take a deeper look, maybe they should do null pointer check at their end.

Below are combination of tests run in parallel likely to cause crash, there maybe more:
LightGBMBinaryEstimatorUnbalanced and BinaryClassificationTrainersOnnxConversionTest
LightGBMRegressorEstimator and BinaryClassificationTrainersOnnxConversionTest
LightGBMBinaryEstimatorUnbalanced and TestSGDBinary
LightGBMBinaryEstimatorUnbalanced and CommandLineOnnxConversionTest
LightGBMBinaryEstimatorCorrectSigmoid and MulticlassConfusionMatrixSlotNames
IrisVectorLightGbmWithLoadColumnName and PlattCalibratorOnnxConversionTest

### DIFF
--- a/test/Microsoft.ML.Tests/Properties/AssemblyInfo.cs
+++ b/test/Microsoft.ML.Tests/Properties/AssemblyInfo.cs
@@ -1,0 +1,9 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Xunit;
+
+// TODO: disable test parallelization for this assembly as running test in parallel sometimes cause test host process to crash
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
+
--- a/test/Microsoft.ML.Tests/Properties/AssemblyInfo.cs
+++ b/test/Microsoft.ML.Tests/Properties/AssemblyInfo.cs
@@ -4,6 +4,6 @@
 
 using Xunit;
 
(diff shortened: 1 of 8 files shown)


test/Microsoft.ML.Tests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Xunit;
+
+// TODO: disable test parallelization for this assembly as running test in parallel sometimes cause test host process to crash
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
+
```
harishsk: Can you please add an additional tag after the TODO to help us search for tags associated with this work item. Say something like // TODO: TEST_STABILITY

Issue #4119, opened 2019-08-15
Does this only apply to numeric columns?


cookt:
It would be good to call out which column types this applies to and for each column type what could be considered a 'missing' value (empty string, 0, null, etc.)

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 3725a2c2-9121-a125-a2c1-ffa5bb799e17
* Version Independent ID: d5e97e37-927e-5f3a-f219-78f972d6e371
* Content: [DataOperationsCatalog.FilterRowsByMissingValues(IDataView, String\[\]) Method (Microsoft.ML)](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.dataoperationscatalog.filterrowsbymissingvalues?view=ml-dotnet)
* Content Source: [dotnet/xml/Microsoft.ML/DataOperationsCatalog.xml](https://github.com/dotnet/ml-api-docs/blob/live/dotnet/xml/Microsoft.ML/DataOperationsCatalog.xml)
* Product: **dotnet-ml-api**
* GitHub Login: @sfilipi
* Microsoft Alias: **johalex**

Issue #3891, opened 2019-06-20
Examples need high-level explanation/summary


eerhardt:
The examples look well written, but there are a few things I would like to see:

1. There should be a high-level explanation of each one, what is it doing and why. There are 5 different examples here. The first 2 and the last 2 appear to be duplicates - I can't tell the difference between them by reading them.  Even between the first 2, it is subtle what the difference is (the first one you can't save and load the model, the second you can). It would be great to explicitly call this out before each example.
2. The 3rd example is fairly complicated, but with no explanation of what is happening. What defines a super alien? What is the overall scenario being achieved? What are the attributes for? It takes quite a bit to try to figure all these things out from the code without a high-level summary of the scenario.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 2af85725-9d23-0cff-c8ca-f915a94663bd
* Version Independent ID: 8afa2992-aa5b-4bb1-7841-6dcf26151288
* Content: [CustomMappingCatalog.CustomMapping(TransformsCatalog, Action&lt;TSrc,TDst&gt;, String, SchemaDefinition, SchemaDefinition) Method (Microsoft.ML)](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.custommappingcatalog.custommapping?view=ml-dotnet)
* Content Source: [dotnet/xml/Microsoft.ML/CustomMappingCatalog.xml](https://github.com/dotnet/ml-api-docs/blob/live/dotnet/xml/Microsoft.ML/CustomMappingCatalog.xml)
* Product: **dotnet-ml-api**
* GitHub Login: @sfilipi
* Microsoft Alias: **johalex**

najeeb-kazmi:
Duplicate examples are caused by a bug on the docs generation side. No ETA on that at the moment.

We should still provide high level explanations for the examples.

rosebyte:
#4899 added high-level explanations that distinguish each `CustomMapping` example and explain the custom in-memory scenario.

PR #4899, opened 2020-02-27
Misc doc/sample fixes, improvements, and typos

### LINKED ISSUE #2655: Add AppendCacheCheckpoint large dataset usage warning to documentation
https://github.com/dotnet/machinelearning-samples/blob/bfcd66167bb4518afb538661a97ad98c27685349/samples/csharp/getting-started/MulticlassClassification_Iris/IrisClassification/IrisClassificationConsoleApp/Program.cs#L50 has a warning: 
// Use in-memory cache for small/medium datasets to lower training time. 
// Do NOT use it (remove .AppendCacheCheckpoint()) when handling very large datasets. 

This should be added to the documentation. 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 2a03f5f8-1ce2-ae98-0f03-3a0cdfd2a208
* Version Independent ID: 58d012b4-4275-8c95-1634-08f7f66cbafb
* Content: [EstimatorChain&lt;TLastTransformer&gt;.AppendCacheCheckpoint(IHostEnvironment) Method (Microsoft.ML.Data)](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.estimatorchain-1.appendcachecheckpoint?view=ml-dotnet#Microsoft_ML_Data_EstimatorChain_1_AppendCacheCheckpoint_Microsoft_ML_IHostEnvironment_)
* Content Source: [dotnet/xml/Microsoft.ML.Data/EstimatorChain`1.xml](https://github.com/dotnet/ml-api-docs/blob/live/dotnet/xml/Microsoft.ML.Data/EstimatorChain`1.xml)
* Product: **dotnet-ml-api**
* GitHub Login: @sfilipi
* Microsoft Alias: **johalex**

### LINKED ISSUE #3407: Improve clarity of Reference Documentation for METRICS
I'm finding that most metrics have very little explanation in the reference.
Other times, it just tells what's the formula for that metric.

For example:

https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.regressionmetrics.rootmeansquarederror?view=ml-dotnet

https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.multiclassclassificationmetrics.logloss?view=ml-dotnet

image

From a developer point of view, what we need to explain is:

- Very short summary of what is that metric (not a formula. The formular is nice to have but not the fundamental thing to understand the metric)
- What to look for with that metric: "The closer to 1.00 the better?", viceversa?, basically, when the model is better or worse when looking at that metric.

Other samples where there's very little explanation about the metric:

https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.multiclassclassificationmetrics.microaccuracy?view=ml-dotnet

https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.regressionmetrics.rsquared?view=ml-dotnet
image

https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.regressionmetrics.meanabsoluteerror?view=ml-dotnet

### LINKED ISSUE #3627: Consider removing references to "needCol" and explaining what "rand" is in GetRowCursor
I realize this is still in preview, but I'm just putting a note here since I just spent an hour or so referring to this. Feel free to ignore if this is too early

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: cebd0074-01d3-d048-305a-cfea8b4083de
* Version Independent ID: 7f1aaf72-e345-e334-6237-d92abd3aa6d3
* Content: [IDataView Interface (Microsoft.ML)](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.idataview?view=ml-dotnet#feedback)
* Content Source: [dotnet/xml/Microsoft.ML/IDataView.xml](https://github.com/dotnet/ml-api-docs/blob/live/dotnet/xml/Microsoft.ML/IDataView.xml)
* Product: **dotnet-ml-api**
* GitHub Login: @sfilipi
* Microsoft Alias: **johalex**

### LINKED ISSUE #3841: CreateTextLoader documentation is missing details
The following parameters of CreateTextLoader API are missing detailed explanation and proper example:

* allowSparse: what is the sparse format?  + example
* allowQuoting: what does this exactly mean? does quoting cause the separator within quotes to be skipped? Is it single quotes or double quotes?
* hasHeader: seems like hasHeader only causes the header line to be skipped, but it doesn't help with automatic detection of column names. Is that the case?
 * dataSample: needs sample

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: a745f3b1-1356-0888-4804-f6141496da18
* Version Independent ID: 0a2d7c9e-9319-8393-bb4a-058bf72450ba
* Content: [TextLoaderSaverCatalog.CreateTextLoader Method (Microsoft.ML)](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.textloadersavercatalog.createtextloader?view=ml-dotnet#Microsoft_ML_TextLoaderSaverCatalog_CreateTextLoader_Microsoft_ML_DataOperationsCatalog_Microsoft_ML_Data_TextLoader_Column___System_Char_System_Boolean_Microsoft_ML_Data_IMultiStreamSource_System_Boolean_System_Boolean_System_Boolean_)
* Content Source: [dotnet/xml/Microsoft.ML/TextLoaderSaverCatalog.xml](https://github.com/dotnet/ml-api-docs/blob/live/dotnet/xml/Microsoft.ML/TextLoaderSaverCatalog.xml)

### LINKED ISSUE #3927: Distinguish between Calibrated and NonCalibrated methods
When would I use one versus the other?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b147624e-e01f-679a-900c-37f48f9d9e15
* Version Independent ID: 55d26134-c3a9-3a4f-5833-2f9b1dc8e0b1
* Content: [BinaryClassificationCatalog.CrossValidateNonCalibrated(IDataView, IEstimator&lt;ITransformer&gt;, Int32, String, String, Nullable&lt;Int32&gt;) Method (Microsoft.ML)](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.binaryclassificationcatalog.crossvalidatenoncalibrated?view=ml-dotnet#feedback)
* Content Source: [dotnet/xml/Microsoft.ML/BinaryClassificationCatalog.xml](https://github.com/dotnet/ml-api-docs/blob/live/dotnet/xml/Microsoft.ML/BinaryClassificationCatalog.xml)
* Product: **dotnet-ml-api**
* GitHub Login: @sfilipi
* Microsoft Alias: **johalex**

### LINKED ISSUE #4898: What is the exact behavior of CreateTextLoader<TInput> when dataSample is given?
We have this overload for `CreateTextLoader<TInput>`, where the schema is defined in `TInput`. 
https://github.com/dotnet/machinelearning/blob/c3d15927522ef9645dd335d9375fda15a2aaab34/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs#L90-L98

The `dataSample` argument is meant to be used to infer schema. Since `TInput` must contain at least one field, there is always at least one column in the schema. Then, this condition is never hit, and consequently, `dataSample` is never used to infer the schema with the `CreateTextLoader<TInput>` overload.
https://github.com/dotnet/machinelearning/blob/c3d15927522ef9645dd335d9375fda15a2aaab34/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L1118

~Presence of the `dataSample` argument is confusing here as it implies that a sample can be provided. In other places, this sample is used to infer schema, so the user would expect this to be the case here as well, but `dataSample` is ignored here.~

~I will update the documentation to reflect this, but this should be removed. Since this will be an API breaking change, this should be revisited for 2.0.~

najeeb-kazmi:
Misc doc/sample improvements, fixes, and typos.

Fixes #2655 
Fixes #3627 
Fixes #3927 
Fixes #3841 
Fixes #4898 
Fixes #3407 

Partially addresses #3891

Related #4119 
And a few minor fixes.

More to come.


src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1264,7 +1269,8 @@ private sealed class LoaderHolder
                 // Setting the defaults to the user provided values will avoid these in the output of the call CmdParser.GetSettings.
                 UseThreads = options.UseThreads,
                 HeaderFile = options.HeaderFile,
-                MaxRows = options.MaxRows
+                MaxRows = options.MaxRows,
+                HasHeader = options.HasHeader
```
yaeldekel: >HasHeader [](start = 36, length = 9)

Why does this need to change?

najeeb-kazmi: If `headerFile` is not provided, and we are trying to parse the schema from `dataSample` here:
https://github.com/dotnet/machinelearning/blob/f0a8a76a40348021b5a24f9b3027d53fff4a58cd/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L1118

And if `hasHeader` is also true, then `"header=+"` is added to `tmp` here:
https://github.com/dotnet/machinelearning/blob/f0a8a76a40348021b5a24f9b3027d53fff4a58cd/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L1261-L1268

Then the schema is not read from `dataSample` and this warning is thrown here:
https://github.com/dotnet/machinelearning/blob/f0a8a76a40348021b5a24f9b3027d53fff4a58cd/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs#L1280-L1285

yaeldekel: the call to to `CmdParser.GetSettings(...)` in line 1266, gets a string that includes all the arguments that are different between `options` (the user specified options) and
```
new Options() { ... }
```

Since what this method does is try to create an `Options` object from the schema in the file, we want to make sure that the user didn't specify any non-default arguments, because if they did - we use what the user wants instead of what the file has. However, there are a few arguments where we will still read the `Options` from the file even if they have non-default values, since they can differ each time you load the file:
* `UseThreads` - since if you load the file on a different machine you might or might not want to use threads
* `HeaderFile` - since if the header is in a separate file the user will need to specify it when the loader is created, depending on where it is located
* `MaxRows` - since at different times you might want to load just a certain number of rows 

The `HasHeader` argument is not one of the arguments that the user is allowed to change between runs, since the file either has a header or not, and if it does then the `Options` defined in the file already contains "header=+". So this code change should be reverted.


---
In reply to: [385412117](https://github.com/dotnet/machinelearning/pull/4899#discussion_r385412117) [](ancestors = 385412117)

najeeb-kazmi: So does `hasHeader` refer to the `dataSample` file and not the data file being loaded? If that is the case, I will update the documentation to reflect that.

As a user, I might think that `hasHeader` refers to the data I will load with `loader.Load` and want to skip the header there. But if I set it to true, and also provide a sample, and the sample   has header in the schema but not "header=+", the schema will not be read.

yaeldekel: `hasHeader` is used for getting slot names from `dataSample`. It is used for determining whether to skip the first line in the file specified in `loader.Load`. 
The `hasHeader` argument does not refer to the schema that is sometimes present in the file, it only refers to a first line in the file that contains slot names. 

---
In reply to: [385861537](https://github.com/dotnet/machinelearning/pull/4899#discussion_r385861537) [](ancestors = 385861537)

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/LoadingText.cs
```diff
@@ -55,11 +66,109 @@ public static void Example()
...
+
+            // Load a single file from path.
+            var singleFileCustomTypeData = loaderWithCustomType.Load(dataFiles[0]);
+            PrintRowCount(singleFileCustomTypeData);
+
+            // Expected Output:
+            //   10
+
+
+            // Create a TextLoader with unknown column length to illutrate
```
yaeldekel: >illutrate [](start = 65, length = 9)

Typo.

docs/samples/Microsoft.ML.Samples/Dynamic/DataOperations/LoadingText.cs
```diff
@@ -55,11 +66,109 @@ public static void Example()
...
+                {
+                    new TextLoader.Column("Features",
+                                            DataKind.Single,
+                                            new[] { new TextLoader.Range(0, null) })
+                },
+                dataSample: dataSample
+            );
+
+            var dataWithInferredLength = loaderWithUnknownLength.Load(dataFiles[0]);
+            PrintRowCount(dataWithInferredLength);
```
yaeldekel: >PrintRowCount [](start = 12, length = 13)

Should we print additional information here to "prove" that inferring the length works? For example, the type of the column of the schema of `dataWithInferredLength`.

src/Microsoft.ML.Data/TrainCatalog.cs
```diff
@@ -213,7 +213,8 @@ internal BinaryClassificationTrainers(BinaryClassificationCatalog catalog)
         /// <summary>
         /// Run cross-validation over <paramref name="numberOfFolds"/> folds of <paramref name="data"/>, by fitting <paramref name="estimator"/>,
         /// and respecting <paramref name="samplingKeyColumnName"/> if provided.
-        /// Then evaluate each sub-model against <paramref name="labelColumnName"/> and return metrics.
+        /// Then evaluate each sub-model against <paramref name="labelColumnName"/> and return <see cref="BinaryClassificationMetrics"/>, which
```
yaeldekel: >return [](start = 88, length = 6)

A list of BinaryClassificationMetrics. Or, "return a BinaryClassificationMetrics object for each sub-model".

yaeldekel: >evaluate each sub-model [](start = 17, length = 23)

Can you also mention that each sub-model is evaluated on the fold of data that it did not see during training?

src/Microsoft.ML.Data/DataLoadSave/EstimatorChain.cs
```diff
@@ -93,7 +93,10 @@ public EstimatorChain<TNewTrans> Append<TNewTrans>(IEstimator<TNewTrans> estimat
 
         /// <summary>
         /// Append a 'caching checkpoint' to the estimator chain. This will ensure that the downstream estimators will be trained against
-        /// cached data. It is helpful to have a caching checkpoint before trainers that take multiple data passes.
+        /// cached data. It is helpful to have a caching checkpoint before trainers or feature engineering that take multiple data passes.
+        /// It is also helpful to have after a slow operation, for example after dataset loading from a slow source or after feature
```
yaeldekel: >slow operation [](start = 47, length = 14)

This is also true only in case there are multiple passes over the data (since caching is done during the first pass).

yaeldekel: >It is also helpful [](start = 12, length = 18)

Can you also mention that adding a cache checkpoint at the end of the estimator chain is meaningless?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -427,9 +427,12 @@ internal bool TryUnparse(StringBuilder sb)
         public class Options
         {
             /// <summary>
-            /// Whether the input may include quoted values, which can contain separator characters, colons,
-            /// and distinguish empty values from missing values. When true, consecutive separators denote a
-            /// missing value and an empty value is denoted by \"\". When false, consecutive separators denote an empty value.
+            /// Whether the input may include double-quoted values. This parameter is used to distinguish separator characters
+            /// in an input value from actual separators. When true, separators within double quotes are treated as part of the
+            /// input value. When false, all separators, even those whitin quotes, are treated as delimiting a new column.
+            /// It is also used to distinguish empty values from missing values. When true, missing value are denoted by consecutive
```
yaeldekel: >It is also used to distinguish empty values from missing values [](start = 16, length = 63)

Is this correct? I know it used to be the case for text input, when the text type used to have a separate notion of "missing" and "empty", but not sure what the text loader does now in this case.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -427,9 +427,12 @@ internal bool TryUnparse(StringBuilder sb)
         public class Options
         {
             /// <summary>
-            /// Whether the input may include quoted values, which can contain separator characters, colons,
```
yaeldekel: >colons [](start = 101, length = 6)

Does this argument still work for colons? If yes, then this should be mentioned.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -478,11 +483,8 @@ public class Options
             public bool TrimWhitespace = Defaults.TrimWhitespace;
 
             /// <summary>
-            /// Whether the data file has a header with feature names.
-            /// Note: If a TextLoader is created with hasHeader = true but without a dataSample, then vector columns made by TextLoader will not contain slot name
-            /// annotations (slots being the elements of the given vector column), because the output schema is made when the TextLoader is made, and not when
-            /// TextLoader.Load(IMultiStreamSource source) is called. In addition, the case where dataSample = null and hasHeader = true indicates to the
-            /// loader that when it is given a file when <see cref="TextLoader.Load(IMultiStreamSource)"/> is called, it needs to skip the first line.
+            /// Whether the file has a header with feature names. When <see langword="true"/>, the loader will skip the first line when
```
yaeldekel: >skip the first line [](start = 111, length = 19)

Should we also mention the slot names?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -21,15 +21,31 @@ public static class TextLoaderSaverCatalog
...
-        /// <paramref name="dataSample"/>, then vector columns made by TextLoader will not contain slot name annotations (slots being the elements of the given vector column),
-        /// because the output schema is made when the TextLoader is made, and not when <see cref="TextLoader.Load(IMultiStreamSource)"/> is called.
-        /// In addition, the case where dataSample = null and hasHeader = true indicates to the loader that when it is given a file when Load()
-        /// is called, it needs to skip the first line.</param>
-        /// <param name="dataSample">The optional location of a data sample. The sample can be used to infer column names and number of slots in each column.</param>
-        /// <param name="allowQuoting">Whether the file can contain columns defined by a quoted string.</param>
-        /// <param name="trimWhitespace">Remove trailing whitespace from lines</param>
-        /// <param name="allowSparse">Whether the file can contain numerical vectors in sparse format.</param>
+        /// <param name="hasHeader">Whether the file has a header with feature names. When <see langword="true"/>, the loader will skip the first line when
```
yaeldekel: >param name="hasHeader" [](start = 13, length = 22)

When a dataSample is provided, then "true" indicates that the first line in the dataSample will be used for feature names, and that when Load() is called the first line will be skipped. When there is no dataSample provided, then "true" just indicates that the loader should skip the first line when Load() is called, but columns will not have slot names annotations. (here you can add the explanation about the slot names being defined in the constructor and not at Load() time.)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -21,15 +21,31 @@ public static class TextLoaderSaverCatalog
...
+        /// <see cref="TextLoader.Load(IMultiStreamSource)"/> is called. In addition, when <paramref name="hasHeader"/> is <see langword="true"/> but
+        /// <paramref name="dataSample"/> is not provided, vector columns created by the loader will not contain slot name annotations (slots being the elements of the given vector column),
+        /// because the output schema is made when the loader is created, and not when <see cref="TextLoader.Load(IMultiStreamSource)"/> is called.</param>
+        /// <param name="dataSample">The optional location of a data sample. The sample can be used to infer slot name annotations if present, and also the number
+        /// of slots in a column defined with <see cref="TextLoader.Range"/> with <see langword="null"/> maximum index.
+        /// If the sample has been saved with ML.NET's <see cref="SaveAsText(DataOperationsCatalog, IDataView, Stream, char, bool, bool, bool, bool)"/>,
+        /// it will also contain the schema information in the header that the loader can read even if <paramref name="columns"/> is <see langword="null"/>.</param>
```
yaeldekel: >even if <paramref name="columns"/> is <see langword="null"/> [](start = 95, length = 60)

In order to use the schema defined in the file, all other arguments should be left with their default values (including null for the columns).

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -21,15 +21,31 @@ public static class TextLoaderSaverCatalog
...
+        /// <param name="allowQuoting">Whether the input may include double-quoted values. This parameter is used to distinguish separator characters
+        /// in an input value from actual separators. When <see langword="true"/>, separators within double quotes are treated as part of the
+        /// input value. When <see langword="false"/>, all separators, even those whitin quotes, are treated as delimiting a new column.
+        /// It is also used to distinguish empty values from missing values. When <see langword="true"/>, missing value are denoted by consecutive
```
yaeldekel: >distinguish empty values from missing values [](start = 31, length = 44)

See my comment in TextLoader.cs.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -150,20 +170,18 @@ public static class TextLoaderSaverCatalog
...
-        /// <param name="allowQuoting">Whether the input may include quoted values,
-        /// which can contain separator characters, colons,
-        /// and distinguish empty values from missing values. When true, consecutive separators
-        /// denote a missing value and an empty value is denoted by \"\".
-        /// When false, consecutive separators denote an empty value.</param>
-        /// <param name="trimWhitespace">Remove trailing whitespace from lines</param>
-        /// <param name="allowSparse">Whether the input may include sparse representations for example,
-        /// if one of the row contains "5 2:6 4:3" that's mean there are 5 columns all zero
-        /// except for 3rd and 5th columns which have values 6 and 3</param>
+        /// <param name="hasHeader">Whether the file has a header with feature names. When <see langword="true"/>, the loader will skip the first line when
```
yaeldekel: >feature names [](start = 71, length = 13)

Since this overload won't create feature names (because it doesn't pass the data sample to the constructor), I would not mention it, and just leave the second sentence.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -105,13 +123,15 @@ public static class TextLoaderSaverCatalog
...
         /// <param name="separatorChar">The character used as separator between data points in a row. By default the tab character is used as separator.</param>
-        /// <param name="hasHeader">Whether the file has a header with feature names. Note: If a TextLoader is created with hasHeader = true but without a
-        /// dataSample, then vector columns made by TextLoader will not contain slot name annotations (slots being the elements of the given vector column),
-        /// because the output schema is made when the TextLoader is made, and not when <see cref="TextLoader.Load(IMultiStreamSource)"/> is called.
-        /// In addition, the case where dataSample = null and hasHeader = true indicates to the loader that when it is given a file when Load()
-        /// is called, it needs to skip the first line.</param>
-        /// <param name="allowQuoting">Whether the file can contain columns defined by a quoted string.</param>
-        /// <param name="trimWhitespace">Remove trailing whitespace from lines</param>
+        /// <param name="hasHeader">Whether the file has a header with feature names. When <see langword="true"/>, the loader will skip the first line when
```
yaeldekel: >feature names [](start = 71, length = 13)

Same comment as for the other overload.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
najeeb-kazmi: No, this is no longer the case. Verified with a small text dataset. Behavior is the same when we have "" or consecutive separators, whaterver the value of `allowQuoting`.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
najeeb-kazmi: @yaeldekel do you mean specifically colons in sparse datasets or generally? 

Sparse data with quotes, e.g. the following, is read as non-sparse regardless of the `allowQuoting` setting.
C1	C2	C3	C4	C5
1	2	3	4	5
5	"3:4"	4:5
5	1:2

The third line is read as
5	?	0	0	0

Generally speaking, any special characters within quotes are read correctly with `allowSparse = true`. e.g. the second line in this file: https://github.com/dotnet/machinelearning/blob/master/test/data/QuotingData.csv#L3
2,"This text contains extra punctuations and special characters.;*<>?!@#$%^&*()_+=-{}|[]:;'"

is read as:
2	"This text contains extra punctuations and special characters.;*<>?!@#$%^&*()_+=-{}|[]:;'"

Do you have an opinion on how this should be documented?

src/Microsoft.ML.Data/DataLoadSave/EstimatorChain.cs
najeeb-kazmi: @yaeldekel wouldn't caching help after applying an expensive transformation, even if that transformation takes only one pass? I'm thinking of a scenario where we transform once, and use the transformed data in multiple places.

yaeldekel: When you call `AppendCacheCheckpoint`, what happens is that the data gets cached right before calling `Fit()` on the next estimator on the chain. So the only thing that matters is how many passes over the data that `Fit()` call does. If it is 0 or 1, then there is no advantage in caching. If it is more than 1, then there is an advantage. 

Now that I think about it, there may be another case: Suppose you have a transformation that is expensive, and in the estimator chain there are two estimators operating on its output, each does one pass over the data. For example, suppose `pipeline` is a pipeline that creates column `C`, and then you have:
```
pipeline.Append(mlContext.Transforms.NormalizeMinMax("MinMax", "C")).Append(mlContext.Transforms.NormalizeMeanVar("MeanVar", "C"));
```
In that case, there **would** be an advantage to add a cache checkpoint before the `MinMax` estimator, since both of these estimators will need to compute `C`. (although, it doesn't seem that this is a very common case).

---
In reply to: [387409264](https://github.com/dotnet/machinelearning/pull/4899#discussion_r387409264) [](ancestors = 387409264)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
yaeldekel: What I was referring to was the case where there is a text column, and the text may contain a colon. What would be the combination of values for `AllowQuoting` and `AllowSparse` that would make TextLoader parse that correctly? I think that if `AllowSparse` is true, then you need `AllowQuoting` to be true, **and** the text containing the colon has to be in quotes (but don't take my word for it, I'm not sure if this is the case), and if `AllowSparse` is false, then the colon will be read as part of the text in the column whether there are quotes or not.

---
In reply to: [387406874](https://github.com/dotnet/machinelearning/pull/4899#discussion_r387406874) [](ancestors = 387406874)

yaeldekel: I don't think that these special characters need any quotes for TextLoader to be able to read them, the only characters are treated differently based on the value of `AllowQuoting` are the separator characters, and colon (since it has a special meaning in the sparse case).

---
In reply to: [387531140](https://github.com/dotnet/machinelearning/pull/4899#discussion_r387531140) [](ancestors = 387531140,387406874)

najeeb-kazmi: I'll verify this and update accordingly.

najeeb-kazmi: @yaeldekel the treatment of colons is independent of `allowQuoting` for both sparse and dense data.

This dense row containing a colon is read correctly regardless of `allowQuoting` and regardless of whether the second column is within quotes or not.
`2,This text contains extra punctuations and special characters.;*<>?!@#$%^&*()_+=-{}|[]:;'`

This sparse row is read correctly (when read as `DataKind.String`) with `allowSparse = true`, regardless of `allowQuoting` and regardless of whether the value `"4:5"` is within quotes or not. 
`5	3:"4:5"	4:5`

The only difference with `allowQuoting = true` is that the value is read as `4:5` instead of `"4:5"`, which is the same as the behavior for any other kind of text, not specific to colons.

I'm going to leave this as it is. #Resolved

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -441,7 +441,9 @@ public class Options
             public bool AllowQuoting = Defaults.AllowQuoting;
 
             /// <summary>
-            /// Whether the input may include sparse representations.
+            /// Whether the input may include sparse representations. For example, a row containing
+            /// "5 2:6 4:3" means that there are 5 columns, and the only non-zero are columns 2 and 4, which have values 6 and 3,
```
yaeldekel: >5 2:6 4:3 [](start = 17, length = 9)

The "5" can also appear in the middle of the line. So, for example, the file can contain this line:
1 2 10 2:3 7:4
this maps to a length 12 vector with the values 1, 2, 0, 0, 3, 0, 0, 0, 0, 4, 0, 0.
So the "10" in that line refers to the number of sparse elements **after** the dense values, indexed accordingly (in this case, index 0 would be the third element in the vector, index 1 the fourth and so on).

You should also mention the connection to the argument `InputSize`, which should be used when the file doesn't contain the number of sparse elements in every line like in "5 2:6 4:3" or in my example above. Note though, that when using `InputSize`, the specified size should be the **total** size, not just the size of the sparse part, but the indices are relative to where the sparse part begins. For example, if `InputSize=10`, then the line 
1 2 2:3 7:4 
would be mapped to 1,2,0,0,3,0,0,0,0,4. And if `InputSize=12`, then the same line would be mapped to 1,2,0,0,3,0,0,0,0,4,0,0.

PR #4927, opened 2020-03-10
fix LdaWorkoutEstimatorCore

frank-dong-ms:
fix LdaWorkoutEstimatorCore test.

resetRandomGenerator needs to be true here as multiple compare will be performed later. 
In lda_engine, a queue of samples with size of (num_of_threads - 2) will be created at first, each time a compare is performed the internal status of one sample (random number: rng_) is changed, so if size of queue is smaller the number of compare performed (in local workstation we have 12 cores thus the issue is not reproduced), dirty data will be used again for calculation and cause issue. set resetRandomGenerator to true will reset the random number rng_ every time before lda calculation thus fix the issue.

### DIFF
--- a/test/Microsoft.ML.Tests/Transformers/TextFeaturizerTests.cs
+++ b/test/Microsoft.ML.Tests/Transformers/TextFeaturizerTests.cs
@@ -728,6 +728,12 @@ namespace Microsoft.ML.Tests.Transformers
             builder.AddColumn("F1V", NumberDataViewType.Single, data);
             var srcView = builder.GetDataView();
 
+            //Attention: resetRandomGenerator needs to be true here as multiple compare will be performed later.
+            //In lda_engine, a queue of samples with size of (num_of_threads - 2) will be created at first,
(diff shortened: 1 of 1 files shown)


test/Microsoft.ML.Tests/Transformers/TextFeaturizerTests.cs
```diff
@@ -729,7 +728,7 @@ public void LdaWorkoutEstimatorCore()
             builder.AddColumn("F1V", NumberDataViewType.Single, data);
             var srcView = builder.GetDataView();
```
mstfbl: Perhaps a comment can be added here that explains the need to reset the random number generator for each received document due to way that ML.NET and LdaNative utilize multiple threads differently, for future readability.

frank-dong-ms: sure, thanks

---
In reply to: [390162926](https://github.com/dotnet/machinelearning/pull/4927#discussion_r390162926) [](ancestors = 390162926)

PR #4928, opened 2020-03-10
Fix for MulticlassNaiveBayesTrainer export to Onnx

Lynx1820:
- Adding support for a batch input dimension
- While ML.NET doesn't use this batch dimension, ORT does.

### DIFF
--- a/src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
+++ b/src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
@@ -440,7 +440,7 @@ namespace Microsoft.ML.Trainers
 
             //log1
             opType = "Sum";
-            var sumOutput = ctx.AddIntermediateVariable(_inputType, "SumOutput", true);
+            var sumOutput = ctx.AddIntermediateVariable(_inputType, "SumOutput");
             ctx.CreateNode(opType, new[] { featureHistogramName, one }, new[] { sumOutput }, ctx.GetNodeName(opType), "");
 
             var logOutput1 = ctx.AddIntermediateVariable(_inputType, "LogOutput", true);
@@ -491,7 +491,7 @@ namespace Microsoft.ML.Trainers
             node.AddAttribute("axes", list);
 
             opType = "Cast";
-            castOutput = ctx.AddIntermediateVariable(NumberDataViewType.Single, "CastOutput2", true);
+            castOutput = ctx.AddIntermediateVariable(NumberDataViewType.Single, "CastOutput2");
             node = ctx.CreateNode(opType, learnedAbsentFeatureLogProb, castOutput, ctx.GetNodeName(opType), "");
             t = InternalDataKindExtensions.ToInternalDataKind(DataKind.Single).ToType();
             node.AddAttribute("to", t);
@@ -505,7 +505,6 @@ namespace Microsoft.ML.Trainers
             ctx.CreateNode(opType, new[] { subOutput, logProbReduceSum, logOutput }, new[] { sumOutput }, ctx.GetNodeName(opType), "");
 
             opType = "Squeeze";
-            var squeezeOutput = ctx.AddIntermediateVariable(_inputType, "SqueezeOutput", true);
             var squeezeNode = ctx.CreateNode(opType, sumOutput, outputNames[1], ctx.GetNodeName(opType), "");
             squeezeNode.AddAttribute("axes", new long[] { 2 });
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
```diff
@@ -414,108 +415,119 @@ bool ISingleCanSaveOnnx.SaveAsOnnx(OnnxContext ctx, string[] outputNames, string
             var labelHistogramName = ctx.AddInitializer(labelHistogramExpanded, new long[] { _featureHistogram[0].Length, _labelHistogram.Length }, "labelHistogramExpanded");
             var learnedAbsentFeatureLogProb = ctx.AddInitializer(_absentFeaturesLogProb, new long[] { _absentFeaturesLogProb.Length, 1 }, "absentFeaturesLogProb");
 
-            var greaterOutput = ctx.AddIntermediateVariable(null, "greaterOutput", true);
+            var greaterOutput = ctx.AddIntermediateVariable(BooleanDataViewType.Instance, "greaterOutput", true);
             var opType = "Greater";
```
harishsk: I don't think type information gets added if the last parameter is true. Can you please confirm this? 
Adding the type and shape information makes it easier to read and debug the model. There are only a few cases I have seen where adding type information causes problems. Let us try to add type information as far as possible (in all the calls below) and skip adding type information only if it is causing problems (and let us document the problem)

harishsk: Also, do these fixes fix the tests in Nimbus?

---
In reply to: [390592978](https://github.com/dotnet/machinelearning/pull/4928#discussion_r390592978) [](ancestors = 390592978)

Lynx1820: Yeah, the info doesn't get added. I think adding the types would make it easier to read and debug especially for the casting nodes, even if the info doesn't get added to the graph (something similar is done in OneVersusAll transformer). I have removed the true from some variables, but I can't do that for most variables in this trainer, since the tensor shapes change from the original shape.

Lynx1820: Yes, this fixes Nimbus.

harishsk: If you look at OneVersusTrainer, shape information is added in most cases (except for a few calls in SaveAsOnnxPreProcess. If the tensor shape changes, is it possible to identify the shape and encode it?

---
In reply to: [390622962](https://github.com/dotnet/machinelearning/pull/4928#discussion_r390622962) [](ancestors = 390622962)

Lynx1820: Yup. Added the shape info.

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
```diff
@@ -414,108 +415,114 @@ bool ISingleCanSaveOnnx.SaveAsOnnx(OnnxContext ctx, string[] outputNames, string
...
 
-            var greaterOutput = ctx.AddIntermediateVariable(null, "greaterOutput", true);
+            var greaterOutput = ctx.AddIntermediateVariable(new VectorDataViewType(BooleanDataViewType.Instance, _featureHistogram[0].Length), "greaterOutput");
             var opType = "Greater";
             ctx.CreateNode(opType, new[] { featureColumn, zero }, new[] { greaterOutput }, ctx.GetNodeName(opType), "");
 
             opType = "Cast";
-            var isFeaturePresent = ctx.AddIntermediateVariable(null, "isFeaturePresent", true);
-            var node = ctx.CreateNode(opType, greaterOutput, isFeaturePresent, ctx.GetNodeName(opType), "");
+            var castOutput = ctx.AddIntermediateVariable(new VectorDataViewType(NumberDataViewType.Single, _featureHistogram[0].Length), "castOutput");
```
harishsk: >new VectorDataViewType(NumberDataViewType.Single, _featureHistogram[0].Length) [](start = 57, length = 78)

There seem to be about two or three shapes involved here. Can you please create the shapes in one place at the top and reuse the shapes? 
It would make the code more readable and make it easier to follow along?

PR #4934, opened 2020-03-11
Improved documentation for LdSvmTrainer

najeeb-kazmi:
Adds explanation of the algorithm, moves details to remarks, adds references to samples.

### DIFF
--- a/src/Microsoft.ML.StandardTrainers/LdSvm/LdSvmTrainer.cs
+++ b/src/Microsoft.ML.StandardTrainers/LdSvm/LdSvmTrainer.cs
@@ -26,11 +26,42 @@ using Microsoft.ML.Trainers;
 namespace Microsoft.ML.Trainers
 {
     /// <summary>
-    /// Non-Linear SVM that implements Local Deep SVM based on paper :
-    /// C. Jose, P. Goyal, P. Aggrwal, and M. Varma, Local deep
-    /// kernel learning for efficient non-linear svm prediction, in ICML, 2013.
-    /// http://research.microsoft.com/en-us/um/people/manik/code/LDKL/download.html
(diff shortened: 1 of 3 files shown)


src/Microsoft.ML.StandardTrainers/LdSvm/LdSvmTrainer.cs
```diff
@@ -26,11 +26,42 @@
...
+    /// |  |  |
+    /// | -- | -- |
+    /// | Machine learning task | Binary classification |
+    /// | Is normalization required? | Yes |
+    /// | Is caching required? | No |
+    /// | Required NuGet in addition to Microsoft.ML | None |
+    /// | Exportable to ONNX | No |
+    ///
+    /// ### Training Algorithm Details
+    /// Local Deep SVM (LD-SVM) is a generalization of Localized Multiple Kernel Learning for non-linear SVM, which aims to learn a different
```
natke: Suggest minor changes, along the lines of:

Local Deep SVM (LD-SVM) is an optimization of Localized Multiple Kernel Learning for non-linear SVM
 
Multiple kernel methods learn a different classifier for each point in the feature space.
 
The prediction cost of multiple kernel methods can be prohibitively expensive for large training sets because the cost is proportional to the number of support vectors, and these grow linearly with the size of the training set.
 
LD-SVM reduces the prediction cost by learning a tree-based local feature embedding that is sparse with high dimensionality, efficiently encoding non-linearities.
 
Using LD-SVM, the prediction cost grows logarithmically with the size of the training set, rather than linearly.

natke: Also when we say prediction cost, are we talking about speed, memory, CPU, or all of the above?

PR #4935, opened 2020-03-12
Nightlybuild fix

frank-dong-ms:
several issue here:
1. delete useless folder to avoid no disk space
2. add missing dependency for nightly build
3. fix LD_LIBRARY_PATH for CentOS to set proper native reference path
4. increase build time, seems it takes more time now for net core 3.0 to build

### DIFF
--- a/build/ci/job-template.yml
+++ b/build/ci/job-template.yml
@@ -53,8 +53,6 @@ jobs:
     - ${{ if and( eq(parameters.nightlyBuild, 'true'), eq(parameters.pool.name, 'Hosted Ubuntu 1604')) }}:
       - bash: echo "##vso[task.setvariable variable=LD_LIBRARY_PATH]$(nightlyBuildRunPath):$LD_LIBRARY_PATH"
         displayName: Set LD_LIBRARY_PATH for Ubuntu and CentOS to locate Native shared library in current running path
-      - bash: echo $LD_LIBRARY_PATH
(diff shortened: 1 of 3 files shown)


build/ci/job-template.yml
```diff
@@ -62,6 +62,15 @@ jobs:
         displayName: List latest package versions
       - script: $(dotnetPath) run --project $(packageUpdaterProjPath)
         displayName: Update package versions for nightly build
+      - ${{ if eq(parameters.buildScript, 'build.cmd') }}:
+        - powershell: |
```
harishsk: Can you please explain why this is useless? If this useless, why do we have it in our tree and have it build?
If the reason is different, can you please add the actual reason? (Also, small typo: useless)

frank-dong-ms: Yes, for NightlyBuild pipeline only NightlyBuild test project is necessary, other test projects are built from previous "Build" step. This "Build" are doing multiple things thatNightlyBuild pipeline also requires, like initialize tools, build native components. We are reusing the "Build" step from CI build pipeline but with side effect that we are also building too much test projects. We can separate the build step into smaller steps and only include necessary steps in nightly build pipeline but that can be quite large effort and I'm not seeing much value of this.

And thanks for pointing out the typo, I will fix the typo

---
In reply to: [391391849](https://github.com/dotnet/machinelearning/pull/4935#discussion_r391391849) [](ancestors = 391391849)

PR #4938, opened 2020-03-12
Added working version of checking whether file is available for access

mstfbl:
Recently AutoFitRegressionTest() and AutoFitBinaryTest() have been failing occasionally on our MachineLearning-Full CI builds with the error:

> System.IO.IOException : The process cannot access the file 'D:\a\1\s\bin\AnyCPU.Debug\Microsoft.ML.AutoML.Tests\netcoreapp2.1\...dataset' because it is being used by another process.
> 
> Stack trace
>    at System.IO.FileStream.ValidateFileHandle(SafeFileHandle fileHandle)
>    at System.IO.FileStream.CreateFileOpenHandle(FileMode mode, FileShare share, FileOptions options)
>    at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
>    at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
>    at Microsoft.ML.AutoML.TextFileSample.CreateFromFullFile(String path) in D:\a\1\s\src\Microsoft.ML.AutoML\ColumnInference\TextFileSample.cs:line 76
>  ...

(Specific errors for [AutoFitRegressionTest ](https://dev.azure.com/dnceng/public/_build/results?buildId=555337&view=logs&j=dd8eddb6-ecc6-5f65-73e6-df90e5693b94&t=5b4b90b5-382b-59ee-45ec-a64c47991239&l=202)and [AutoFitBinaryTest](https://dev.azure.com/dnceng/public/_build/results?buildId=545476&view=logs&j=41509eb4-74ce-5e57-61b4-bdf74b39e7c1&t=dbdc2969-5b98-5c39-1328-31d4a2fdc45e&l=286) from CI builds)

This indicates that the .dataset file that is trying to be reached still has a lock on it, and is not yet ready for accessing. The current method of checking whether this file is ready to be accessed (checking `new FileInfo(dataFile).Length > 0` is not sufficient. 

The fix below remedies this by obtaining these dataset files from the local `test/data` path, and thereby eliminating the possibility of download locks. This fix also does the same for other datasets that are available locally that were being downloaded from a commit before, so that similar file lock issues for those datasets can be avoided as well.

### DIFF
--- a/test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
+++ b/test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
@@ -8,10 +8,8 @@ using System.IO;
 using System.IO.Compression;
 using System.Linq;
 using System.Net;
-using System.Threading;
 using Microsoft.ML.Data;
-using Microsoft.ML.Internal.Utilities;
-using Microsoft.ML.Runtime;
+using Microsoft.ML.TestFrameworkCommon;
 
 namespace Microsoft.ML.AutoML.Test
 {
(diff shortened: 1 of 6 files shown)


test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs
```diff
@@ -49,7 +49,8 @@ public static IDataView GetUciAdultDataView()
 
         private static string DownloadIfNotExists(string baseGitPath, string dataFile)
         {
-            foreach (var nextIteration in Enumerable.Range(0, 10))
```
justinormont: I'd recommend replacing `DownloadIfNotExists()` w/ a call to [`ResourceManagerUtils.Instance.EnsureResourceAsync()`](https://github.com/dotnet/machinelearning/blob/d26b896da20a6de152a68e07a9b27969dbb63fe0/src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs#L87-L114) which at the time it was written, by @yaeldekel, I believed it to have no race conditions when the storage is local, but not for network shares due to lack of a remote atomic rename. It's designed as a thread safe downloader of large files.

Example of call to `EnsureResourceAsync()`:
https://github.com/dotnet/machinelearning/blob/d26b896da20a6de152a68e07a9b27969dbb63fe0/src/Microsoft.ML.TensorFlow/TensorflowUtils.cs#L179-L194

As a background I think this DownloadIfNotExists() function was added

mstfbl: Hi @justinormont , indeed there is a race condition here in DownloadIfNotExists(). If ResourceManagerUtils.Instance.EnsureResourceAsync() can indeed download these files in a thread-safe manner, I'll replace `client.DownloadFile` with this. Thank you!

mstfbl: Hey Justin, with `ResourceManagerUtils.Instance.EnsureResourceAsync()`, I'm seeing tests that are at some time using this function in the call stack are hanging. I've noticed that some of our TensorFlowImageClassification tests that are currently disabled due to timeout/hanging issues are also using  `ResourceManagerUtils.Instance.EnsureResourceAsync()`. Perhaps these two happenings are related, and that EnsureResourceAsync() is not actually thread-safe?

justinormont: I added an alternative method for checking for bad aka.ms links in your other PR:
https://github.com/dotnet/machinelearning/pull/4943/files#r392572981

I did notice we were retrying bad aka.ms links multiple times, causing a delay of 50 seconds (5 retries x 10 sec). This only affects bad URLs though.

**Testing:**
You could test `ResourceManagerUtils` locally by creating/running ~100 in parallel. This should simulate what's occurring in the CI.

**Singleton:**
`ResourceManagerUtils` is a singleton: 
https://github.com/dotnet/machinelearning/blob/d26b896da20a6de152a68e07a9b27969dbb63fe0/src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs#L23-L32

It's not obvious to me why it's a singleton; I think we could just make all functions static in the class.

**Deadlocks:**
I don't see any deadlocks from the mutex.

**Stall:**
In `return (await t).Message;`, we have potentially timed out via `timeoutTask`, but will still wait for a response from `DownloadResource()`:
https://github.com/dotnet/machinelearning/blob/d26b896da20a6de152a68e07a9b27969dbb63fe0/src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs#L151-L162

We do send the cancellation token to `DownloadResource()`, but it doesn't check the cancellation token until after it receives the next 4096 bytes:
https://github.com/dotnet/machinelearning/blob/d26b896da20a6de152a68e07a9b27969dbb63fe0/src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs#L285-L296

Hence if the server stops sending data, I think this will wait indefinitely. To fix, we could use `Stream.ReadAsync()`, which takes a cancellation token instead of the current `Read()`.

**Other Aync:**
I'm not an expert in C# async code. You may want to look closer at its usage in `ResourceManagerUtils`.

mstfbl:
This PR will be ready for merge once #4943 is merged into master.

mstfbl:
With [this build](https://dev.azure.com/dnceng/public/_build/results?buildId=564993&view=results), I have tested my changes with AutoFitRegressionTest() and AutoFitBinaryTest() running 100 iterations. These tests that used to hang are not not hanging.

PR #4939, opened 2020-03-12
Added IDisposable support for several classes

harishsk:
We have had a lot of instability in the Tensorflow tests. At least one of the issues has to do with memory leaks from ImageClassificationTrainer which holds references to the Tensorflow session and graphs. In order to fix this I have added IDisposable support for several of the classes involved in this scenario and fixed up the tests to call Dispose at the end of the tests.

### DIFF
--- a/src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using Microsoft.ML;
 using Microsoft.ML.Data;
 using Microsoft.ML.Runtime;
@@ -15,7 +16,7 @@ namespace Microsoft.ML.Data
     /// This class represents a data loader that applies a transformer chain after loading.
     /// It also has methods to save itself to a repository.
     /// </summary>
-    public sealed class CompositeDataLoader<TSource, TLastTransformer> : IDataLoader<TSource>
+    public sealed class CompositeDataLoader<TSource, TLastTransformer> : IDataLoader<TSource>, IDisposable
         where TLastTransformer : class, ITransformer
     {
         internal const string TransformerDirectory = TransformerChain.LoaderSignature;
@@ -110,5 +111,25 @@ namespace Microsoft.ML.Data
                 loaderSignature: LoaderSignature,
                 loaderAssemblyName: typeof(CompositeDataLoader<,>).Assembly.FullName);
         }
+
+        #region IDisposable Support
+        private bool _disposed = false;
+
+        private void Dispose(bool disposing)
+        {
+            if (_disposed)
+                return;
+
+            if (disposing)
+                Transformer.Dispose();
+
+            _disposed = true;
+        }
+
+        void IDisposable.Dispose()
+        {
+            Dispose(true);
+        }
+        #endregion
     }
 }
--- a/src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
@@ -39,6 +39,7 @@ namespace Microsoft.ML.Data
 
             Loader = loader;
             Transformer = transformerChain ?? new TransformerChain<TLastTransformer>();
+            _disposed = false;
         }
 
         private CompositeDataLoader(IHost host, ModelLoadContext ctx)
@@ -46,6 +47,7 @@ namespace Microsoft.ML.Data
             if (!ctx.LoadModelOrNull<IDataLoader<TSource>, SignatureLoadModel>(host, out Loader, LegacyLoaderDirectory))
                 ctx.LoadModel<IDataLoader<TSource>, SignatureLoadModel>(host, out Loader, LoaderDirectory);
             ctx.LoadModel<TransformerChain<TLastTransformer>, SignatureLoadModel>(host, out Transformer, TransformerDirectory);
+            _disposed = false;
         }
 
         private static CompositeDataLoader<TSource, TLastTransformer> Create(IHostEnvironment env, ModelLoadContext ctx)
@@ -113,7 +115,7 @@ namespace Microsoft.ML.Data
         }
 
         #region IDisposable Support
-        private bool _disposed = false;
+        private bool _disposed;
 
         private void Dispose(bool disposing)
         {
--- a/src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
@@ -39,7 +39,6 @@ namespace Microsoft.ML.Data
 
             Loader = loader;
             Transformer = transformerChain ?? new TransformerChain<TLastTransformer>();
-            _disposed = false;
         }
 
         private CompositeDataLoader(IHost host, ModelLoadContext ctx)
@@ -47,7 +46,6 @@ namespace Microsoft.ML.Data
             if (!ctx.LoadModelOrNull<IDataLoader<TSource>, SignatureLoadModel>(host, out Loader, LegacyLoaderDirectory))
                 ctx.LoadModel<IDataLoader<TSource>, SignatureLoadModel>(host, out Loader, LoaderDirectory);
             ctx.LoadModel<TransformerChain<TLastTransformer>, SignatureLoadModel>(host, out Transformer, TransformerDirectory);
-            _disposed = false;
         }
 
         private static CompositeDataLoader<TSource, TLastTransformer> Create(IHostEnvironment env, ModelLoadContext ctx)
@@ -117,20 +115,14 @@ namespace Microsoft.ML.Data
         #region IDisposable Support
         private bool _disposed;
 
-        private void Dispose(bool disposing)
+        public void Dispose()
         {
             if (_disposed)
                 return;
 
-            if (disposing)
                 Transformer.Dispose();
 
             _disposed = true;
-        }
-
-        void IDisposable.Dispose()
-        {
-            Dispose(true);
         }
         #endregion
     }
--- a/src/Microsoft.ML.Data/Scorers/MulticlassClassificationScorer.cs
+++ b/src/Microsoft.ML.Data/Scorers/MulticlassClassificationScorer.cs
@@ -386,21 +386,15 @@ namespace Microsoft.ML.Data
             #region IDisposable Support
             private bool _disposed;
 
-            private void Dispose(bool disposing)
+            void IDisposable.Dispose()
             {
                 // TODO: Is it necessary to call the base class Dispose()?
                 if (_disposed)
                     return;
 
-                if (disposing)
-                    (_bindable as IDisposable)?.Dispose();
+                (_bindable as IDisposable)?.Dispose();
 
                 _disposed = true;
-            }
-
-            void IDisposable.Dispose()
-            {
-                Dispose(true);
             }
             #endregion
         }
--- a/src/Microsoft.ML.Data/Scorers/MulticlassClassificationScorer.cs
+++ b/src/Microsoft.ML.Data/Scorers/MulticlassClassificationScorer.cs
@@ -126,7 +126,6 @@ namespace Microsoft.ML.Data
                 _getter = getter;
                 _metadataKind = metadataKind;
                 _canWrap = canWrap;
-                _disposed = false;
             }
 
             private LabelNameBindableMapper(IHost host, ModelLoadContext ctx)
@@ -146,7 +145,6 @@ namespace Microsoft.ML.Data
                 _getter = Utils.MarshalInvoke(_decodeInitMethodInfo, this, _type.ItemType.RawType, value);
                 _metadataKind = ctx.Header.ModelVerReadable >= VersionAddedMetadataKind ?
                     ctx.LoadNonEmptyString() : AnnotationUtils.Kinds.SlotNames;
-                _disposed = false;
             }
 
             private Delegate DecodeInit<T>(object value)
@@ -386,7 +384,7 @@ namespace Microsoft.ML.Data
             #region IDisposable Support
             private bool _disposed;
 
-            void IDisposable.Dispose()
+            public void Dispose()
             {
                 // TODO: Is it necessary to call the base class Dispose()?
                 if (_disposed)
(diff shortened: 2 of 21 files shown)


src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
```diff
@@ -110,5 +113,25 @@ private static VersionInfo GetVersionInfo()
                 loaderSignature: LoaderSignature,
                 loaderAssemblyName: typeof(CompositeDataLoader<,>).Assembly.FullName);
         }
+
+        #region IDisposable Support
+        private bool _disposed;
+
+        private void Dispose(bool disposing)
```
sharwell: 📝 This is the legacy `IDisposable` pattern. The new way to implement this is moving this implementation code directly into `IDisposable.Dispose()`.

src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
```diff
@@ -15,7 +16,7 @@ namespace Microsoft.ML.Data
     /// This class represents a data loader that applies a transformer chain after loading.
     /// It also has methods to save itself to a repository.
     /// </summary>
-    public sealed class CompositeDataLoader<TSource, TLastTransformer> : IDataLoader<TSource>
+    public sealed class CompositeDataLoader<TSource, TLastTransformer> : IDataLoader<TSource>, IDisposable
```
sharwell: 📝 By making the class disposable when `IDataLoader<TSource>` is not disposable, it will be easy for a consumer to fail to dispose of the object.

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -310,6 +316,7 @@ public void TensorFlowTransformInputShapeTest()
                 }
                 Assert.False(cursor.MoveNext());
             }
+            (tfModel as IDisposable).Dispose();
```
sharwell: ❗️ This doesn't handle null correctly. One of the following should be used:

```suggestion
            (tfModel as IDisposable)?.Dispose();
```

```suggestion
            ((IDisposable)tfModel).Dispose();
```

src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
harishsk: Weren't there two cases to be handled before? The one that calls Dispose(true) from from IDisposable.Dispose() and the other that calls Dispose(false) from the finalizer? (Or someplace else...I am not quite sure). How would that be handled?

---
In reply to: [396132625](https://github.com/dotnet/machinelearning/pull/4939#discussion_r396132625) [](ancestors = 396132625)

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
harishsk: In this case, tfModel is TensorFlowModel which implements IDisposable. So it should always be non-null. But I have added the ? check in case the pipeline above changes in the future. 

---
In reply to: [396132941](https://github.com/dotnet/machinelearning/pull/4939#discussion_r396132941) [](ancestors = 396132941)

src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
harishsk: We can't change interfaces now that they are shipped. Right now, when customers discover they have a memory leak, we don't have a solution. With this method at least we can tell them to add a Dispose call. 

Is there a better approach we can consider?


---
In reply to: [396132696](https://github.com/dotnet/machinelearning/pull/4939#discussion_r396132696) [](ancestors = 396132696)

src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
sharwell: >  the other that calls Dispose(false) from the finalizer

Finalizers are not allowed under the new dispose pattern.

src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
sharwell: Every case that I reviewed locally required breaking API changes. Default interface methods might be an option for .NET Core targets.

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -463,6 +471,7 @@ public void TensorFlowTransformInputOutputTypesTest()
...
```
frank-dong-ms: >(tfModel as IDisposable).Dispose(); [](start = 12, length = 35)

(tfModel as IDisposable)?.Dispose();

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1333,10 +1363,14 @@ internal bool ShouldReuse(string workspacePath, string trainSetBottleneckCachedV
         [InlineData(ImageClassificationTrainer.Architecture.MobilenetV2)]
         [InlineData(ImageClassificationTrainer.Architecture.ResnetV250)]
         [InlineData(ImageClassificationTrainer.Architecture.InceptionV3)]
-        //Skipping test temporarily. This test will be re-enabled once the cause of failures has been determined
-        [Trait("Category", "SkipInCI")]
         public void TensorFlowImageClassification(ImageClassificationTrainer.Architecture arch)
         {
+            if (NotCentOS7FactAttribute.IsCentOS7())
```
frank-dong-ms: >IsCentOS7 [](start = 40, length = 9)

as discussed pleas also skip in Ubuntu as this also hangs on Ubuntu like below:

[https://dev.azure.com/dnceng/public/_build/results?buildId=556713&view=logs&j=28859320-f5de-51e0-1fd2-7bea8c11cf7a&t=28348391-228f-5cbf-8fcd-3272a0755d32](https://dev.azure.com/dnceng/public/_build/results?buildId=556713&view=logs&j=28859320-f5de-51e0-1fd2-7bea8c11cf7a&t=28348391-228f-5cbf-8fcd-3272a0755d32)
[https://dev.azure.com/dnceng/public/_build/results?buildId=550567&view=logs&j=5aa5c7df-492a-5eaf-973a-62b7b0f0ee3b&t=ffdbd604-f3e2-5332-cf61-c8dd00799b47](https://dev.azure.com/dnceng/public/_build/results?buildId=550567&view=logs&j=5aa5c7df-492a-5eaf-973a-62b7b0f0ee3b&t=ffdbd604-f3e2-5332-cf61-c8dd00799b47)
[https://dev.azure.com/dnceng/public/_build/results?buildId=549005&view=logs&j=28859320-f5de-51e0-1fd2-7bea8c11cf7a&t=28348391-228f-5cbf-8fcd-3272a0755d32](https://dev.azure.com/dnceng/public/_build/results?buildId=549005&view=logs&j=28859320-f5de-51e0-1fd2-7bea8c11cf7a&t=28348391-228f-5cbf-8fcd-3272a0755d32)

src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
```diff
@@ -38,13 +39,15 @@ public CompositeDataLoader(IDataLoader<TSource> loader, TransformerChain<TLastTr
 
             Loader = loader;
             Transformer = transformerChain ?? new TransformerChain<TLastTransformer>();
+            _disposed = false;
```
eerhardt: `bool`s are initialized to `false` by the runtime. There is no need to re-initialize them to `false` here.

src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
```diff
@@ -110,5 +113,25 @@ private static VersionInfo GetVersionInfo()
...
+            if (_disposed)
+                return;
+
+            if (disposing)
+                Transformer.Dispose();
+
+            _disposed = true;
+        }
+
+        void IDisposable.Dispose()
```
eerhardt: Why are you explicitly implementing `IDisposable`? Instead, you can just have a `public void Dispose()` method here.

The reason to not explicitly implement an interface is so users of the class can just call `Dispose()` as necessary. They don't need to cast to `IDisposable` to call it.

src/Microsoft.ML.Data/Scorers/MulticlassClassificationScorer.cs
```diff
@@ -379,6 +382,21 @@ public RowImpl(DataViewRow row, DataViewSchema schema)
...
+            #region IDisposable Support
+            private bool _disposed;
+
+            void IDisposable.Dispose()
+            {
+                // TODO: Is it necessary to call the base class Dispose()?
+                if (_disposed)
+                    return;
+
+                (_bindable as IDisposable)?.Dispose();
```
eerhardt: `ISchemaBindableMapper` is internal, so if you want you can make it inherit from `IDisposable` without a breaking change.

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -154,6 +154,8 @@ public void TensorFlowTransforCifarEndToEndTest2()
             Assert.Equal(0, prediction.PredictedScores[0], 2);
             Assert.Equal(0, prediction.PredictedScores[1], 2);
             Assert.Equal(1, prediction.PredictedScores[2], 2);
+
+            transformer.Dispose();
```
eerhardt: You can use the `using` statement in all these tests. It will make this cleaner.

[https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement)

src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
justinormont: Is the component GA? Or can we still make a breaking change?

harishsk: All public interfaces are frozen. This would be one of those. 

---
In reply to: [396753025](https://github.com/dotnet/machinelearning/pull/4939#discussion_r396753025) [](ancestors = 396753025)

justinormont: > All public interfaces are frozen. This would be one of those.

There may be a difference between interfaces released as GA and the preview releases. I'm not sure which one this falls under.

/cc @eerhardt

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -389,7 +385,7 @@ public void TensorFlowTransformInputOutputTypesTest()
 
             var inputs = new string[] { "f64", "f32", "i64", "i32", "i16", "i8", "u64", "u32", "u16", "u8", "b" };
             var outputs = new string[] { "o_f64", "o_f32", "o_i64", "o_i32", "o_i16", "o_i8", "o_u64", "o_u32", "o_u16", "o_u8", "o_b" };
-            var tfModel = mlContext.Model.LoadTensorFlowModel(modelLocation);
+            using var tfModel = mlContext.Model.LoadTensorFlowModel(modelLocation);
```
justinormont: Consistency/user-expectations:
How will a user know which model types and prediction engines are disposable? Should we convert all models and prediction engines to be disposable?

Seems it should be an all-or-none for consistency.

src/Microsoft.ML.Data/DataLoadSave/CompositeDataLoader.cs
harishsk: IDataLoader is a core interface. It is part of the GA set. 

---
In reply to: [396760924](https://github.com/dotnet/machinelearning/pull/4939#discussion_r396760924) [](ancestors = 396760924)

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1185,10 +1189,10 @@ public void TensorFlowSentimentClassificationTest()
             // For explanation on how was the `sentiment_model` created 
             // c.f. https://github.com/dotnet/machinelearning-testdata/blob/master/Microsoft.ML.TensorFlow.TestModels/sentiment_model/README.md
             string modelLocation = @"sentiment_model";
-            var pipelineModel = mlContext.Model.LoadTensorFlowModel(modelLocation).ScoreTensorFlowModel(new[] { "Prediction/Softmax" }, new[] { "Features" })
+            using var pipelineModel = mlContext.Model.LoadTensorFlowModel(modelLocation).ScoreTensorFlowModel(new[] { "Prediction/Softmax" }, new[] { "Features" })
```
justinormont: Similar changes may need to be made in the AutoML code, and its use in Model Builder. Otherwise it will continue to leak there.

Perhaps: In a follow-up PR: add a using or call dispose() for TF models and TF prediction engines.

/cc @LittleLittleCloud

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
harishsk: I agree that it should be all or none for consistency. If we go for none, then we need to make the interfaces inherit from IDisposable and we can't do that now without a breaking change. If for "all" then the Dispose() calls for most would become no-ops and only Tensorflow (right now) needs a real Dispose() call. 
This seems like one of those cases that we will need to fix whenever we do a v2 release and can take breaking changes. Until then, I am okay with documenting for the tensorflow related transformers. 

---
In reply to: [396761653](https://github.com/dotnet/machinelearning/pull/4939#discussion_r396761653) [](ancestors = 396761653)

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
harishsk: I have updated the docs and samples. I could not figure out easily where the corresponding fixes need to be bade in AutoML, Model Builder and CodeGen. @LittleLittleCloud, can you please help with that?

---
In reply to: [396764702](https://github.com/dotnet/machinelearning/pull/4939#discussion_r396764702) [](ancestors = 396764702)

PR #4942, opened 2020-03-13
Fixes KMeans scoring differences between ORT and OnnxRunner

Lynx1820:
The KMeans ORT score predictions were off because batches were not being handled correctly. 
There is still an issue because Nimbus produces int32 predictions, when they should be uint32, since ML.NET returns key values, but I plan on working on that separately, since it's probably a Nimbus bug.

### DIFF
--- a/src/Microsoft.ML.KMeansClustering/KMeansModelParameters.cs
+++ b/src/Microsoft.ML.KMeansClustering/KMeansModelParameters.cs
@@ -337,14 +337,15 @@ namespace Microsoft.ML.Trainers
 
             // Compute -2XC^T. Note that Gemm always takes three inputs. Since we only have two here,
             // a dummy one, named zero, is created.
+            var dataViewType = new VectorDataViewType(NumberDataViewType.Single, _centroids.Length);
             var zeroName = ctx.AddInitializer(new float[] { 0f }, null, "zero");
-            var nameXC2 = ctx.AddIntermediateVariable(new VectorDataViewType(NumberDataViewType.Single, _centroids.Length), "XC2");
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.KMeansClustering/KMeansModelParameters.cs
```diff
@@ -331,19 +331,20 @@ bool ISingleCanSaveOnnx.SaveAsOnnx(OnnxContext ctx, string[] outputNames, string
...
-            var nameX2 = ctx.AddIntermediateVariable(null, "X2", true);
+            var nameX2 = ctx.AddIntermediateVariable(new VectorDataViewType(NumberDataViewType.Single, 1), "X2");
             var reduceNodeX2 = ctx.CreateNode("ReduceSumSquare", nameX, nameX2, ctx.GetNodeName("ReduceSumSquare"), "");
+            reduceNodeX2.AddAttribute("axes", new long[] { 1 });
 
             // Compute -2XC^T. Note that Gemm always takes three inputs. Since we only have two here,
             // a dummy one, named zero, is created.
             var zeroName = ctx.AddInitializer(new float[] { 0f }, null, "zero");
-            var nameXC2 = ctx.AddIntermediateVariable(null, "XC2", true);
+            var nameXC2 = ctx.AddIntermediateVariable(new VectorDataViewType(NumberDataViewType.Single, _centroids.Length), "XC2");
```
harishsk: Can we combine the usage of the VectorDataViewType in the two cases?

Lynx1820: done.

PR #4838, opened 2020-02-14
Optimize generic MethodInfo for Func<T1, T2, TResult>

sharwell:
Builds on #4836

PR #4943, opened 2020-03-13
Fixed and Added unit tests for EnsureResourceAsync hanging issue

mstfbl:
Fixed EnsureResourceAsync by removing the helper function ResourceManagerUtils.CheckValidDownload and by using WebClientResponseUri (an extension of WebClient). Also added unit test for EnsureResourceAsync.

ResourceManagerUtils.EnsureResourceAsync(...) appends the relative URL of a resource to the $MlNetResourcesUrl environment variable (equal to "https://aka.ms/mlnet-resources/" by default).

https://github.com/dotnet/machinelearning/blob/cdb1e4b38308d9256cbde9e740a14b3bc7d64c2f/src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs#L34

For example, this occurs in the unit test WordEmbeddingsTest.TestWordEmbeddings(), where `sentimend.emd` is downloaded from https://aka.ms/mlnet-resources/Text/Sswe/sentiment.emd .

There is a possibility of this combined URL not correctly pointing to a resource, which is why this CheckValidDownload function exists:

https://github.com/dotnet/machinelearning/blob/cdb1e4b38308d9256cbde9e740a14b3bc7d64c2f/src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs#L168-L184

This function was not correctly warning that the HTML of www.microsoft.com was downloaded (instead of an intended data file), as the size of microsoft.com is bigger than 4 kilobytes, and the site should have the sub-strings <head and <body (not <head> and <body>).


src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -161,22 +161,19 @@ private async Task<string> DownloadFromUrlAsync(IHostEnvironment env, IChannel c
...
         }
 
-        private static string CheckValidDownload(IChannel ch, string filePath, string url, ref bool deleteNeeded)
+        private static string CheckValidDownload(string filePath, string url, ref bool deleteNeeded)
         {
             // If the relative url does not exist, aka.ms redirects to www.microsoft.com. Make sure this did not happen.
-            // If the file is big then it is definitely not the redirect.
             var info = new FileInfo(filePath);
-            if (info.Length > 4096)
-                return null;
```
justinormont: I'm worried without a size check, a dataset consisting of HTML would not work. For instance when classifying news articles scraped from the news websites. The existing size check ensured that larger datasets were not blocked. 

We can subclass `WebClient` to track the final URI (after redirects): ([idea src](http://www.nullskull.com/a/1484/make-the-webclient-class-follow-redirects-and-get-target-url.aspx))
```c#

public class WebClientResponseUri : WebClient
{
	public Uri ResponseUri { get; private set; }

	protected override WebResponse GetWebResponse(WebRequest request)
	{
		WebResponse response = null;
		
		try
		{
			response = base.GetWebResponse(request);
			ResponseUri = response.ResponseUri;
		}
		catch
		{
		}
		
		return response;
	}
}
```

Then use to check the URI:
```c#
private Exception DownloadResource(IHostEnvironment env, IChannel ch, WebClientResponseUri /* Note new class name */ webClient, Uri uri, string path, string fileName, CancellationToken ct)
{
    ...
    using (var ws = fh.CreateWriteStream())
    {
        var headers = webClient.ResponseHeaders.GetValues("Content-Length");
        var requestUri = webClient.ResponseUri; /* -- start new part -- */
   
        if (requestUri.Host == "www.microsoft.com" && requestUri.ToString().Length < 60)
            return ch.Except($"Invalid aka.ms url: {uri}");  /* -- end new part --*/
    
        if (Utils.Size(headers) == 0 || !long.TryParse(headers[0], out var size))
            size = 10000000;
        ...
```

Have to modify `DownloadFromUrlWithRetryAsync()`:
```C#
        private async Task<string> DownloadFromUrlWithRetryAsync(IHostEnvironment env, IChannel ch, string url, string fileName,
            int timeout, string filePath, int retryTimes = 5)
		{
			var downloadResult = "";

			for (int i = 0; i < retryTimes; ++i)
			{
				var thisDownloadResult = await DownloadFromUrlAsync(env, ch, url, fileName, timeout, filePath);

				if (string.IsNullOrEmpty(thisDownloadResult))
					return thisDownloadResult;
				else
				{
					downloadResult += thisDownloadResult + @"\n";

					// Do not retry if the URL does not exist
					if (thisDownloadResult.Contains("does not exist"))
						return downloadResult;
				}

				await Task.Delay(10 * 1000);
			}

			return downloadResult;
		}
```

Then can remove `CheckValidDownload()`.

Tested on .NET Fiddle -- https://dotnetfiddle.net/5zs4c2 (can modify `url` in a correct value to see non-redirect case)

Perhaps could benefit from a unit test which tests a bad aka.ms link?

yaeldekel: These APIs cannot be used to download arbitrary datasets, they are only used to (internally) download resources from specific locations (namely, urls under [https://aka.ms/mlnet-resources/](https://aka.ms/mlnet-resources/)). So I don't think it is necessary to go to the trouble of making this relatively big change.

I do agree with the last sentence in your comment though, I remember that we had a bunch of tests for the `ResourceManagerUtils` class in TLC, not sure where they are.

---
In reply to: [392572981](https://github.com/dotnet/machinelearning/pull/4943#discussion_r392572981) [](ancestors = 392572981)

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -110,80 +108,64 @@ public async Task<ResourceDownloadResults> EnsureResourceAsync(IHostEnvironment
...
                     };
 
                 webClient.Disposed += disposed;
-                var t = Task.Run(() => DownloadResource(env, ch, webClient, new Uri(url), filePath, fileName, downloadCancel.Token));
+                var t = Task.Run(() => DownloadResource(webClient, new Uri(url), filePath, fileName, downloadCancel.Token));
 
                 UpdateTimeout(ref timeout);
                 var timeoutTask = Task.Delay(timeout).ContinueWith(task => default(Exception), TaskScheduler.Default);
-                ch.Info($"Downloading {fileName} from {url} to {filePath}");
+                Console.WriteLine($"Downloading {fileName} from {url} to {filePath}");
```
eerhardt: Libraries like ML.NET should never be writing to the Console directly.

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -328,4 +318,25 @@ public static ResourceDownloadResults GetErrorMessage(out string errorMessage, p
...
+        protected override WebResponse GetWebResponse(WebRequest request)
+        {
+            WebResponse response = null;
+
+            try
+            {
+                response = base.GetWebResponse(request);
+                ResponseUri = response.ResponseUri;
+            }
+            catch
```
eerhardt: Why are you catching and eating all exceptions here?

justinormont: I think the Try/Catch is not needed. 

Apologies. This is my fault for [recommending](https://github.com/dotnet/machinelearning/pull/4943#pullrequestreview-374711517) the [idea src](http://www.nullskull.com/a/1484/make-the-webclient-class-follow-redirects-and-get-target-url.aspx) a bit too closely. 

We're not calling the original `WebRequest.ResponseUri` ([src](https://github.com/microsoft/referencesource/blob/3b1eaf5203992df69de44c783a3eda37d3d4cd10/System/net/System/Net/WebResponse.cs#L193-L217)), so no exceptions to worry about from that line. Its docs say to override in a derived class to get the redirected URI ([docs](https://docs.microsoft.com/en-us/dotnet/api/system.net.webresponse.responseuri?view=netcore-3.1#System_Net_WebResponse_ResponseUri)).

I don't see direct documentation on which exceptions `WebRequest.GetWebResponse()` can throw, could dig deeper in its [src](https://github.com/dotnet/runtime/blob/036fb0779a5bc89026d2ec64172897ab86af5828/src/libraries/System.Net.WebClient/src/System/Net/WebClient.cs#L247-L252). Only reason I would see to catch for this is if we wanted more specific logging; anything else?

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
justinormont: Correct. 

This originated from my .NET fiddle (https://dotnetfiddle.net/5zs4c2) where I didn't have an `IChannel` implemented; hence wrote to Console. I'd recommend using just the updates in my original [comment](https://github.com/dotnet/machinelearning/pull/4943#discussion_r392572981). No need to remove `IChannel` and `IHostEnvironment`.

For the fiddle, I also had to use `using (var fh = env.CreateOutputFile(tempPath)) using (var ws = fh.CreateWriteStream())` instead of the existing `using (var fs = File.Create(tempPath)) using (var ws = new StreamWriter(fs))`, for the same reason. We should keep the older method.

mstfbl: Got it, I reverted these changes in my latest commit.

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -274,7 +258,15 @@ private Exception DownloadResource(IHostEnvironment env, IChannel ch, WebClient
                 using (var ws = fh.CreateWriteStream())
                 {
                     var headers = webClient.ResponseHeaders.GetValues("Content-Length");
-                    if (Utils.Size(headers) == 0 || !long.TryParse(headers[0], out var size))
+                    var requestUri = webClient.ResponseUri;
+                    Console.WriteLine("requestUri = " + requestUri);
+                    if (requestUri.Host == "www.microsoft.com" && requestUri.ToString().Length < 60)
+                    {
+                        Console.WriteLine($"The url '{uri}' does not exist. Url was redirected to '{requestUri}'.");
```
justinormont: Can remove:
```suggestion
```

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -274,7 +258,15 @@ private Exception DownloadResource(IHostEnvironment env, IChannel ch, WebClient
                 using (var ws = fh.CreateWriteStream())
                 {
                     var headers = webClient.ResponseHeaders.GetValues("Content-Length");
-                    if (Utils.Size(headers) == 0 || !long.TryParse(headers[0], out var size))
+                    var requestUri = webClient.ResponseUri;
+                    Console.WriteLine("requestUri = " + requestUri);
```
justinormont: Can remove:
```suggestion
```

test/Microsoft.ML.Core.Tests/UnitTests/TestUtilities.cs
```diff
@@ -299,5 +303,258 @@ public void CheckAreEqualDouble()
...
+
+                        if (t.Status != TaskStatus.RanToCompletion)
+                            Fail("Download did not complete succesfully");
+                        if (File.Exists(Path.Combine(saveToDir, fileName)))
+                            Fail($"File '{Path.Combine(saveToDir, fileName)}' should have been deleted.");
+                    }
+                }
+
+                // Good url, bad page.
+                if (!Uri.TryCreate("http://cnn.com/", UriKind.Absolute, out var cnn))
```
justinormont: Secure URLs:
```suggestion
                if (!Uri.TryCreate("https://cnn.com/", UriKind.Absolute, out var cnn))
```

src/Microsoft.ML.TensorFlow/TensorflowUtils.cs
```diff
@@ -181,7 +181,7 @@ internal static void DownloadIfNeeded(IHostEnvironment env, string url, string d
             using (var ch = env.Start("Ensuring meta files are present."))
             {
                 var ensureModel = ResourceManagerUtils.Instance.EnsureResourceAsync(env, ch, url, fileName, dir, timeout);
-                ensureModel.Wait();
+                ensureModel.Wait(5 * 60 * 1000); // 5 minute timeout period
```
justinormont: I think the TF models are around ~200MB. Is 5min too low for slower connections? For a 1Mb (mega-bit) connection, this would be 27min. 

How likely is a user to see the exception (`$@"Please copy the model file '{name}' from '{url}' to '{directory}'.");` sent to the channel? I think AutoML (API, CLI, & ModelBulder) will log them to file (not displayed), and would count them as a failed iteration and move on to the next iteration.

src/Microsoft.ML.Transforms/Text/WordEmbeddingsExtractor.cs
```diff
@@ -629,7 +629,7 @@ private string EnsureModelFile(IHostEnvironment env, out int linesToSkip, WordEm
                     string dir = kind == WordEmbeddingEstimator.PretrainedModelKind.SentimentSpecificWordEmbedding ? Path.Combine("Text", "Sswe") : "WordVectors";
                     var url = $"{dir}/{modelFileName}";
                     var ensureModel = ResourceManagerUtils.Instance.EnsureResourceAsync(env, ch, url, modelFileName, dir, Timeout);
-                    ensureModel.Wait();
+                    ensureModel.Wait(5 * 60 * 1000); // 5 minute timeout period
```
justinormont: Too low for the 6GB fastText word embedding.

This would require a 164 Mbps connection to finish in 5 min.

justinormont: As a baseline, this would timeout for me on my home connection: _(and for a connection 4x faster than mine)_
image

I expect I'll see similar results, but I'm checking the download time of the fastText 6GB model on my connection. https://aka.ms/tlc-resources/WordVectors/wiki.en.vec.

mstfbl: Thanks for bringing this up @justinormont . I made this edit to make sure these tests would not hang in our CI builds, and it seems they have made no difference.

test/Microsoft.ML.Core.Tests/UnitTests/TestResourceDownload.cs
```diff
@@ -0,0 +1,276 @@
...
+        {
+            var envVarOld = Environment.GetEnvironmentVariable(ResourceManagerUtils.CustomResourcesUrlEnvVariable);
+            var timeoutVarOld = Environment.GetEnvironmentVariable(ResourceManagerUtils.TimeoutEnvVariable);
+            var resourcePathVarOld = Environment.GetEnvironmentVariable(Utils.CustomSearchDirEnvVariable);
+            Environment.SetEnvironmentVariable(Utils.CustomSearchDirEnvVariable, null);
+
+            // Bad path.
+            try
+            {
+                if (!Uri.TryCreate($@"\\ct01\public\{Guid.NewGuid()}\", UriKind.Absolute, out var badUri))
```
justinormont: I'm not sure how the CI nodes will handle a Windows share that it, hopefully, can't access. Your internal machine may react differently.

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -328,4 +319,17 @@ public static ResourceDownloadResults GetErrorMessage(out string errorMessage, p
...
     }
+
+    public class WebClientResponseUri : WebClient
+    {
+        public Uri ResponseUri { get; private set; }
+
+        protected override WebResponse GetWebResponse(WebRequest request)
+        {
+            WebResponse response = null;
+            response = base.GetWebResponse(request);
```
justinormont: Nit -- can combine:
```suggestion
            WebResponse response = base.GetWebResponse(request);
```

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -269,12 +253,19 @@ private Exception DownloadResource(IHostEnvironment env, IChannel ch, WebClient
             string tempPath = Path.GetFullPath(Path.Combine(Path.GetDirectoryName(path), "temp-resource-" + guid.ToString()));
             try
             {
+                Console.WriteLine(String.Format("Trying to view file through the url: {0}", uri.ToString()));
```
justinormont: Unneeded; same info is printed in calling function on line: `ch.Info($"Downloading {fileName} from {url} to {filePath}");`
```suggestion
```

test/Microsoft.ML.Core.Tests/UnitTests/TestResourceDownload.cs
```diff
@@ -0,0 +1,276 @@
...
+                Done();
+            }
+            finally
+            {
+                // Set environment variable back to its old value.
+                Environment.SetEnvironmentVariable(ResourceManagerUtils.CustomResourcesUrlEnvVariable, envVarOld);
+                Environment.SetEnvironmentVariable(ResourceManagerUtils.TimeoutEnvVariable, timeoutVarOld);
+                Environment.SetEnvironmentVariable(Utils.CustomSearchDirEnvVariable, resourcePathVarOld);
+            }
+        }
```
justinormont: Specifically testing the code path which detects bad aka.ms would be helpful to ensure our detection logic remains working.

Potential URI:  `http://aka.ms/mlnet/badurltest/{Guid.NewGuid()}`. The added "mlnet/badurltest" token is to be nice to the aka.ms folks who would otherwise receive unexplainable 404s.

mstfbl: Good point, I've replaced the earlier URL test with this to avoid unnecessary extra test cases and longer testing times.

test/Microsoft.ML.Core.Tests/UnitTests/TestResourceDownload.cs
mstfbl: This works on both local and CI nodes in my experience, but I have changed it so that there are no Windows share directories involved.

justinormont: Did you push the commit w/ the change? I still see `$@"\\ct01\public\{Guid.NewGuid()}\"`.

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -113,8 +113,7 @@ public async Task<ResourceDownloadResults> EnsureResourceAsync(IHostEnvironment
                 await DownloadFromUrlWithRetryAsync(env, ch, absoluteUrl.AbsoluteUri, fileName, timeout, filePath), absoluteUrl.AbsoluteUri);
         }
 
-        private async Task<string> DownloadFromUrlWithRetryAsync(IHostEnvironment env, IChannel ch, string url, string fileName,
-            int timeout, string filePath, int retryTimes = 5)
+        private async Task<string> DownloadFromUrlWithRetryAsync(IHostEnvironment env, IChannel ch, string url, string fileName, int timeout, string filePath, int retryTimes = 5)
```
harishsk: Can you please revert this? This is not good for a few reasons:
1. Your change does not change behavior in any way. It just puts the whole function on a different line.
2. It does not improve formatting to make it more readable either because the previous formatting is more readable. The deliberate word wrap exists to prevent horizontal scrolling while reading code.
3. It makes reviewers pause and read more carefully as to whether you changed something significant.

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -125,8 +124,12 @@ public async Task<ResourceDownloadResults> EnsureResourceAsync(IHostEnvironment
                 if (string.IsNullOrEmpty(thisDownloadResult))
                     return thisDownloadResult;
                 else
+                {
                     downloadResult += thisDownloadResult + @"\n";
-
+                    // Do not retry if the URL does not exist
+                    if (thisDownloadResult.Contains("does not exist"))
```
harishsk: Can you please check codes instead of strings?

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -274,6 +258,8 @@ private Exception DownloadResource(IHostEnvironment env, IChannel ch, WebClient
                 using (var ws = fh.CreateWriteStream())
                 {
                     var headers = webClient.ResponseHeaders.GetValues("Content-Length");
+                    if (IsRedirectToDefaultPage(uri))
+                        return ch.Except($"Invalid aka.ms url: {uri}");
```
harishsk: Are we guaranteed that this is an aka.ms url here?

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -311,6 +297,28 @@ private Exception DownloadResource(IHostEnvironment env, IChannel ch, WebClient
...
+            try
+            {
+                HttpWebRequest myHttpWebRequest = (HttpWebRequest)WebRequest.Create(uri.AbsoluteUri);
+                myHttpWebRequest.AllowAutoRedirect = false;
+                HttpWebResponse myHttpWebResponse = (HttpWebResponse)myHttpWebRequest.GetResponse();
+            }
+            catch (WebException e)
+            {
+                // Redirects to https://www.microsoft.com/en-us/?ref=aka
+                if (e.Message == "The remote server returned an error: (302) Moved Temporarily.")
```
harishsk: The message from web servers is not guaranteed to remain the same. Like we discussed earlier, you should be using http status codes. 
See [https://docs.microsoft.com/en-us/dotnet/api/system.net.httpwebresponse.statuscode?view=netframework-4.8#System_Net_HttpWebResponse_StatusCode](https://docs.microsoft.com/en-us/dotnet/api/system.net.httpwebresponse.statuscode?view=netframework-4.8#System_Net_HttpWebResponse_StatusCode)

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -328,4 +336,15 @@ public static ResourceDownloadResults GetErrorMessage(out string errorMessage, p
         private static extern int chmod(string pathname, int mode);
 #pragma warning restore IDE1006
     }
+
+    public class WebClientResponse : WebClient
```
harishsk: This class doesn't seem to be providing a whole lot of functionality. Is this strictly necessary? Is it possible to read the response directly from WebClient?

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -274,6 +258,8 @@ private Exception DownloadResource(IHostEnvironment env, IChannel ch, WebClient
                 using (var ws = fh.CreateWriteStream())
                 {
                     var headers = webClient.ResponseHeaders.GetValues("Content-Length");
+                    if (IsRedirectToDefaultPage(uri))
```
harishsk: This still does not answer the question as to why we have invalid aka.ms urls. Which urls are those?

mstfbl: These URLs are those that are outputted as absoluteUrl here:

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -121,12 +121,12 @@ public async Task<ResourceDownloadResults> EnsureResourceAsync(IHostEnvironment
             for (int i = 0; i < retryTimes; ++i)
             {
                 var thisDownloadResult = await DownloadFromUrlAsync(env, ch, url, fileName, timeout, filePath);
-
                 if (string.IsNullOrEmpty(thisDownloadResult))
                     return thisDownloadResult;
                 else
+                {
```
harishsk: Please follow convention. (It is non-standard...but) In the ML.NET codebase, when an if/else statement has just a single line, it is not enclosed in braces. Just like the previous if statement.

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
harishsk: That means they are not aka.ms urls. Is that correct? If so, why would they be subject to redirection?

---
In reply to: [394132286](https://github.com/dotnet/machinelearning/pull/4943#discussion_r394132286) [](ancestors = 394132286)

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
harishsk: You resolved this comment. And removed the reference to WebClientResponse from the function above. Do you still need this class?

---
In reply to: [394102879](https://github.com/dotnet/machinelearning/pull/4943#discussion_r394102879) [](ancestors = 394102879)

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
mstfbl: Not always. In PR #4838, I am trying to use `EnsureResourceAsync` instead of `DownloadIfNotExists` here, to use the async nature of `EnsureResourceAsync` and fix hanging issues with some of our hanging tests :

https://github.com/dotnet/machinelearning/blob/d849ba4c7831586018d5b2cca6bdd2b1fa228ddf/test/Microsoft.ML.AutoML.Tests/DatasetUtil.cs#L40-L48

I found that when I use `EnsureResourceAsync` here, and ran the test `AutoFitRegressionTest`, it attempted to download from the url `https://aka.ms/mlnet-resources/https://raw.githubusercontent.com/dotnet/machinelearning/e78971ea6fd736038b4c355b840e5cbabae8cb55/test/data/generated_regression_dataset.csv%22`, which is obviously false. This link was subject to redirection.

justinormont: **Why have an aka.ms test:**
When creating a new component that uses ResourceManagerUtils' `EnsureResourceAsync()`, or when modifying a component's download path, it's easy to get the path wrong. Detection difficulty is exacerbated as incorrect aka.ms URLs silently return a webpage instead of a 404.

The previous check for this was broken. The method I recommended was checking the final URI (for uri.host=="www.microsoft.com" && uri.ToString().Length < somelimit). 

Having a 301/302 is an ok response for URIs. So may not be strict enough.

Perhaps check for: return code, & final host. Perhaps restricting the check to only aka.ms links (in case someone want to use `ResourceManagerUtils` for non-aka.ms links).

If we want more strict checking we could verify the filesize and/or file hash. We would also need versioned/immutable resource files in our stores (or old ML․NET versions will fail validation w/ if a file is updated).

**Where to store resource files:**
For large files, all of our download URLs should be aka.ms links pointing to a CDN (and CDN retrieves from the actual location). The aka.ms link give us forward-compatibility as we can point these to new locations should we modify our file hosting locations. The CDN gives us good worldwide download speeds.

Storage locations:
* Tiny unit test sized files -- in repo, use as a local file when possible
* Medium -- alternative repo w/ pinned commit (not master), downloaded by `ResourceManagerUtils`
* Large and above -- aka.ms link to CDN, downloaded by `ResourceManagerUtils`

The URLs [[1](https://raw.githubusercontent.com/dotnet/machinelearning/e78971ea6fd736038b4c355b840e5cbabae8cb55/test/data/generated_regression_dataset.csv), [2](https://raw.githubusercontent.com/dotnet/machinelearning/e78971ea6fd736038b4c355b840e5cbabae8cb55/test/data/generated_regression_dataset.csv")] are linking into this repo. Generally I don't think there's much reason for grabbing from the same repo using a web call; this was likely created when the AutoML code was in a different repo.

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
mstfbl: Apologies for the confusion, I marked comments as resolved as I fixed them locally but before I pushed my changes to keep track of what I worked on.

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -121,12 +121,10 @@ public async Task<ResourceDownloadResults> EnsureResourceAsync(IHostEnvironment
             for (int i = 0; i < retryTimes; ++i)
             {
                 var thisDownloadResult = await DownloadFromUrlAsync(env, ch, url, fileName, timeout, filePath);
-
                 if (string.IsNullOrEmpty(thisDownloadResult))
```
harishsk: What was the point of deleting the previous line? And the line below? Please try to keep your PRs clean.

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
harishsk: I think you mean PR #4938. 

So, the first decision to make is whether EnsureResourceAsync is exclusively for downloading from aka.ms urls or not. 

If it is exclusively for aka.ms urls, then it will always prepend aka.ms, and your usage of this function in PR #4938 is wrong.

If however we wish to treat this as a generic download function, then you cannot assume that it will always redirect back to microsoft.com. In this case checking for http codes is the only reliable thing to do. 

Justin is right that you shouldn't be using web requests to download files that are already in the repo. If we also adopt his suggestion that all resources ML.NET uses should be abstracted behind aka.ms, then it is okay to check for final redirection going back to microsoft. And if we do detect such a condition that means our usage is wrong or the aka.ms entry needs to be updated and in this case we should be throwing an exception. 

So please figure out what else this function is needed for and whether we are willing to adopt the policy that we are going to use aka.ms urls for everything. 

---
In reply to: [394139313](https://github.com/dotnet/machinelearning/pull/4943#discussion_r394139313) [](ancestors = 394139313)

justinormont: We could allow the calling function to verify the file, and we would need to return the needed info to the calling function to let it make a decision. For instance, the final URI.

Though it's a bit more complicated as we'd then be outside of the mutex. We'll be causing a race condition if we allow the calling function to then delete the file.
  
thread 1: calls `EnsureResourceAsync()`, gets mutex
thread 2: calls `EnsureResourceAsync()`, waiting on mutex

thread 1: downloads file, unlocks mutex, returns to calling function
thread 2: grabs mutex, sees that file exists, unlocks mutex, returns to its calling function

thread 1: checks and deletes file
thread 2: opens file, ...{{dies}}...

Perhaps we could do dependency injection style of sending an optional lambda function to `EnsureResourceAsync()` which verifies the file while within the mutex?

harishsk: Yeah, I think this would be a bit of overkill. I would rather let the calling function do the higher level verification. 

---
In reply to: [394608927](https://github.com/dotnet/machinelearning/pull/4943#discussion_r394608927) [](ancestors = 394608927)

justinormont: Verification done properly (no race conditions) from the calling function would involve another named mutex, and much of the steps provided by `EnsureResourceAsync()`.

I do agree this is bit of overkill. I'd recommend the simple route of having `EnsureResourceAsync()` check for bad aka.ms links and leave the verification at that.

harishsk: @mstfbl Can you please verify if this code path always will be used only for aka.ms links? Only in that case I think we can explicitly check for microsoft.com as the redirected uri for failure. 

---
In reply to: [394646902](https://github.com/dotnet/machinelearning/pull/4943#discussion_r394646902) [](ancestors = 394646902)

mstfbl: @harishsk `EnsureResourceAsync()` will be used only for aka.ms links. The AutoML test datasets that are currently downloaded with `DownloadIfNotExists` from earlier commits exist in the ML.NET repo, so I see now there is no need to use `EnsureResourceAsync()` for them.
In this case, I'll add back the explicit check for https://microsoft.com/ as the redirected url for failure.

harishsk: Are you sure? Can you please take a look at line 39 in ResourcemanagerUtils.cs?
It appears that MlNetResourcesUrl can be overridden by an environment variable. How would it work if somebody ran this code with a different url set in the environment?
What is the use case for that?

mstfbl: Noted. The purpose of IsRedirectToDefaultPage(string url) is to distinguish the redirects that can happen from faulty aka.ms URLs. I added a check where this redirect check would only apply to aka.ms URLs.

mstfbl:
I have added a check to ensure only resources downloaded from the "aka.ms" host can use EnsureResourceAsync(). I have also edited the unit test for EnsureResourceAsync() to reflect this.

src/Microsoft.ML.Transforms/Text/WordEmbeddingsExtractor.cs
```diff
@@ -628,7 +628,7 @@ private string EnsureModelFile(IHostEnvironment env, out int linesToSkip, WordEm
                 {
                     string dir = kind == WordEmbeddingEstimator.PretrainedModelKind.SentimentSpecificWordEmbedding ? Path.Combine("Text", "Sswe") : "WordVectors";
                     var url = $"{dir}/{modelFileName}";
-                    var ensureModel = ResourceManagerUtils.Instance.EnsureResourceAsync(Host, ch, url, modelFileName, dir, Timeout);
```
harishsk: Sorry I missed this in the earlier reviews. Why is it necessary to change Host to env?

mstfbl: No worries! In all the other cases where `EnsureResourceAsync` is used, the provided `IHostEnvironment env` is used, not `IHost Host`. This is so that the function can get the proper environment. This particular usage of `EnsureResourceAsync` was the only case where `Host` was being provided, so I fixed it.

PR #4949, opened 2020-03-18
Add allowBinary Flag to internal columnInference 

LittleLittleCloud:
Add allowBinary flag to internal columnInference API. to stop parsing label as `Boolean` type when it's not binary experiment.

### DIFF
--- a/src/Microsoft.ML.AutoML/ColumnInference/ColumnTypeInference.cs
+++ b/src/Microsoft.ML.AutoML/ColumnInference/ColumnTypeInference.cs
@@ -224,6 +224,10 @@ namespace Microsoft.ML.AutoML
             }
         }
 
+        /// <summary>
+        /// Get Experts for parsing raw values.
+        /// </summary>
+        /// <returns></returns>
         private static IEnumerable<ITypeInferenceExpert> GetExperts()
         {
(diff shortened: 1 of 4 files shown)


src/Microsoft.ML.AutoML/ColumnInference/ColumnTypeInference.cs
```diff
@@ -357,12 +361,18 @@ private static InferenceResult InferTextFileColumnTypesCore(MLContext context, I
...
+            if (labelColumn.HasAllBooleanValues() && allowBinary)
             {
                 labelColumn.SuggestedType = BooleanDataViewType.Instance;
             }
 
+            // if labelColumn has Boolean values and allowBianry is false, set its type to Text because it's raw type must be string (or it will be parse as single).
+            if(labelColumn.SuggestedType == BooleanDataViewType.Instance && !allowBinary)
+            {
+                labelColumn.SuggestedType = TextDataViewType.Instance;
+            }
```
justinormont: I don't think the right answer is passing down a single use "allowBinary". This is meant to be more generic.

Is there a cleaner way? Perhaps could add a `ToString` transform to the pipeline to modify? Or perhaps modify the type in the returned `InferenceResult `?

PR #4950, opened 2020-03-18
fix race condition for test MulticlassTreeFeaturizedLRTest

frank-dong-ms:
method in TreeEnsembleFeaturizerTransform will be called from multi-threading, make variable "temp" as local variable to avoid race condition.

### DIFF
--- a/src/Microsoft.ML.FastTree/TreeEnsembleFeaturizer.cs
+++ b/src/Microsoft.ML.FastTree/TreeEnsembleFeaturizer.cs
@@ -742,8 +742,9 @@ namespace Microsoft.ML.Data
                 mapper =
                     (in TInput src, ref Single dst) =>
                     {
-                        // Attention: this method will be used in multipe threading,
-                        // don't put temp variable outside of this method to avoid race condition
+                        //Attention: This method is called from multiple threads.
+                        //Do not move the temp variable outside this method.
+                        //If you do, the variable is shared between the threads and results in a race condition.
                         ulong temp = 0;
                         if (isNa(in src))
                         {
@@ -761,8 +762,9 @@ namespace Microsoft.ML.Data
                 mapper =
                     (in TInput src, ref Single dst) =>
                     {
-                        // Attention: this method will be used in multipe threading,
-                        // don't put temp variable outside of this method to avoid race condition
+                        //Attention: This method is called from multiple threads.
+                        //Do not move the temp variable outside this method.
+                        //If you do, the variable is shared between the threads and results in a race condition.
                         ulong temp = 0;
                         if (isNa(in src))
                         {
--- a/src/Microsoft.ML.FastTree/TreeEnsembleFeaturizer.cs
+++ b/src/Microsoft.ML.FastTree/TreeEnsembleFeaturizer.cs
@@ -742,6 +742,8 @@ namespace Microsoft.ML.Data
                 mapper =
                     (in TInput src, ref Single dst) =>
                     {
+                        // Attention: this method will be used in multipe threading,
+                        // don't put temp variable outside of this method to avoid race condition
                         ulong temp = 0;
                         if (isNa(in src))
                         {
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.FastTree/TreeEnsembleFeaturizer.cs
```diff
@@ -735,14 +735,14 @@ public static IDataTransform CreateForEntryPoint(IHostEnvironment env, Arguments
...
             var isNa = Conversions.Instance.GetIsNAPredicate<TInput>(type);
-            ulong temp = 0;
 
             ValueMapper<TInput, Single> mapper;
             if (seed == 0)
             {
                 mapper =
                     (in TInput src, ref Single dst) =>
                     {
+                        ulong temp = 0;
```
harishsk: Very nice find! :-) Can you please add some comments about this change? Otherwise some future developer might want to "optimize" this and move it back. 

Also, this pattern deserves more investigation. Do we have other mappers that are using variables set outside the mapper function definition?

frank-dong-ms: Yes, sure, I will add comments on this.

---
In reply to: [394130751](https://github.com/dotnet/machinelearning/pull/4950#discussion_r394130751) [](ancestors = 394130751)

frank-dong-ms: @sharwell do you have any tool recommendation to detect similar race condition issue either static code analyze or run time detection?

---
In reply to: [394131983](https://github.com/dotnet/machinelearning/pull/4950#discussion_r394131983) [](ancestors = 394131983,394130751)

frank-dong-ms: No similar issue found from other mappers but mapper are not only function used in multi-threading condition and it is impossible to view for all similar issue.

---
In reply to: [394130751](https://github.com/dotnet/machinelearning/pull/4950#discussion_r394130751) [](ancestors = 394130751)

sharwell: I'm not aware of such a tool. It would be easier to write a more targeted tool that identified variables captured by a lambda so they could be reviewed for cases like this.

src/Microsoft.ML.FastTree/TreeEnsembleFeaturizer.cs
```diff
@@ -742,6 +742,8 @@ public static IDataTransform CreateForEntryPoint(IHostEnvironment env, Arguments
                 mapper =
                     (in TInput src, ref Single dst) =>
                     {
+                        // Attention: this method will be used in multipe threading,
+                        // don't put temp variable outside of this method to avoid race condition
```
harishsk: Nitpick: The comment is slightly confusing. Can you please rephrase it? Say, something like:

This method is called from multiple threads. Do not move the temp variable outside this method. If you do, the variable is shared between the threads and results in a race condition.

harishsk: Isnt there a test that can be enabled back because of this fix?

frank-dong-ms: No, the test (MulticlassTreeFeaturizedLRTest) is not disabled but I see several fail on machine learning full test set.
As test machine only has 2 cores so the failure rate is not that high but if we run the test locally the failure rate is more than 10 percent.

---
In reply to: [394607002](https://github.com/dotnet/machinelearning/pull/4950#discussion_r394607002) [](ancestors = 394607002)

frank-dong-ms: Sure, thanks

---
In reply to: [394606675](https://github.com/dotnet/machinelearning/pull/4950#discussion_r394606675) [](ancestors = 394606675)

src/Microsoft.ML.FastTree/TreeEnsembleFeaturizer.cs
frank-dong-ms: Got it, thanks

---
In reply to: [394334398](https://github.com/dotnet/machinelearning/pull/4950#discussion_r394334398) [](ancestors = 394334398)

PR #4963, opened 2020-03-23
Fixes multiclass logistic regression

Lynx1820:
Fixes dimension bug for MulticlassLogisticRegression and SdcaMaximumEntropy in Nimbus (https://github.com/microsoft/NimbusML/issues/473)
Unsqueeze was adding dimension 0 to the label tensor - to get shape (1,#samples), instead of the expected shape (#samples, 1).


src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/MulticlassLogisticRegression.cs
```diff
@@ -979,40 +979,26 @@ private bool SaveAsOnnxCore(OnnxContext ctx, string[] outputs, string featureCol
...
-                if (outputs[i] != DefaultColumnNames.PredictedLabel)
-                    continue;
-                predictedLabelUint32 = DefaultColumnNames.PredictedLabel;
-                predictedLabelInt64 = ctx.AddIntermediateVariable(NumberDataViewType.Int64, "PredictedLabelInt64", true);
-                outputs[i] = predictedLabelInt64;
-                break;
-            }
-
-            Host.CheckValue(predictedLabelInt64, nameof(predictedLabelInt64));
+            string labels = ctx.AddIntermediateVariable(NumberDataViewType.Int64, "Labels", true);
```
harishsk: Naming Nit: If this is not a vector, should this be in singluar as "Label" and "label"?

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/MulticlassLogisticRegression.cs
```diff
@@ -979,40 +979,26 @@ private bool SaveAsOnnxCore(OnnxContext ctx, string[] outputs, string featureCol
...
 
             string opType = "LinearClassifier";
-            var node = ctx.CreateNode(opType, new[] { featureColumn }, outputs, ctx.GetNodeName(opType));
+            var node = ctx.CreateNode(opType, new[] { featureColumn }, new[] { labels, outputs[1] }, ctx.GetNodeName(opType));
```
harishsk: Is the assumption here that the outputs array contain strings in a fixed order? If so, can you please add an assert about what you are expecting in outputs[1] ?(Maybe upon function entry along with the checks)

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/MulticlassLogisticRegression.cs
```diff
@@ -979,40 +979,26 @@ private bool SaveAsOnnxCore(OnnxContext ctx, string[] outputs, string featureCol
...
+            opType = "Unsqueeze";
+            var unsqueezeOutput = ctx.AddIntermediateVariable(NumberDataViewType.Int64, "CastNodeOutput");
+            var unsqueezeNode = ctx.CreateNode(opType, labels, unsqueezeOutput, ctx.GetNodeName(opType), "");
+            unsqueezeNode.AddAttribute("axes", new long[] { 1 });
+
             // Onnx outputs an Int64, but ML.NET outputs UInt32. So cast the Onnx output here
             opType = "Cast";
-            var castNodeOutput = ctx.AddIntermediateVariable(NumberDataViewType.UInt32, "CastNodeOutput", true);
-            var castNode = ctx.CreateNode(opType, predictedLabelInt64, castNodeOutput, ctx.GetNodeName(opType), "");
+            var castNode = ctx.CreateNode(opType, unsqueezeOutput, outputs[0], ctx.GetNodeName(opType), "");
```
harishsk: Same here for outputs[0]. Please add an assert on function entry.

src/Microsoft.ML.StandardTrainers/Standard/LogisticRegression/MulticlassLogisticRegression.cs
Lynx1820: yup, that's the assumption that many other trainers have

PR #4966, opened 2020-03-24
Updated OnnxScoringEstimator's documentation

### LINKED ISSUE #4872: Update docs to remove mentions of GPU support in ApplyOnnxModel
When Onnxruntime released 1.0, they changed the API and broke the interface from 0.5 so that now it is not possible to ever exercise the GPU code path unless someone compiles ML.NET privately and links it to Microsoft.ML.OnnxRuntime.Gpu specifically.

Related to #4478 and #4656

antoniovs1029:
PR #4919 changed the way users should work with the Onnxruntime's nugets, now they should include either Onnxruntime or Onnxruntime.Gpu nuget along the OnnxTransformer one.

I also updated a couple of other things that weren't right anymore.

Please, let me know it I should update the docs anywhere else.

Fixes #4872

### DIFF
--- a/src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
+++ b/src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
@@ -702,17 +702,17 @@ namespace Microsoft.ML.Transforms.Onnx
     /// | Does this estimator need to look at the data to train its parameters? | No |
     /// | Input column data type | Known-sized vector of <xref:System.Single> or <xref:System.Double> types |
     /// | Output column data type | As specified by the ONNX model |
-    /// | Required NuGet in addition to Microsoft.ML | Microsoft.ML.OnnxTransformer (always),  Microsoft.ML.OnnxRuntime.Gpu (only if GPU processing is used) |
+    /// | Required NuGet in addition to Microsoft.ML | Microsoft.ML.OnnxTransformer (always),  either Microsoft.ML.OnnxRuntime 1.2.0 (for CPU processing) or Microsoft.ML.OnnxRuntime.Gpu 1.2.0 (only if GPU processing is desired) |
     /// | Exportable to ONNX | No |
     ///
-    /// Supports inferencing of models in ONNX 1.2 and 1.3 format (opset 7, 8 and 9), using the
-    /// [Microsoft.ML.OnnxRuntime](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime/) library.
+    /// Supports inferencing of models in ONNX 1.6 format (opset 11), using the
+    /// [Microsoft.ML.OnnxRuntime](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime/) library (version 1.2.0).
     /// Models are scored on CPU by default. If GPU execution is needed (optional), use the
-    /// NuGet package available at [Microsoft.ML.OnnxRuntime.Gpu](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu/)
-    /// and download [CUDA 9.1 Toolkit](https://developer.nvidia.com/cuda-downloads) and [cuDNN](https://developer.nvidia.com/cudnn).
+    /// NuGet package available at [Microsoft.ML.OnnxRuntime.Gpu](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu/) 1.2.0 instead
+    /// and download [CUDA Toolkit](https://developer.nvidia.com/cuda-downloads) and [cuDNN](https://developer.nvidia.com/cudnn) (use the versions indicated on [Onnxruntime's documentation](https://github.com/Microsoft/onnxruntime#default-gpu-cuda)).
     /// Set parameter 'gpuDeviceId' to a valid non-negative integer. Typical device ID values are 0 or 1.
     /// The inputs and outputs of the ONNX models must be Tensor type. Sequence and Maps are not yet supported.
-    /// OnnxRuntime currently works on Windows and Ubuntu 16.04 Linux 64-bit platforms. Mac OS to be supported soon.
+    /// OnnxRuntime works on Windows, MacOS and Ubuntu 16.04 Linux 64-bit platforms.
     /// Visit [ONNX Models](https://github.com/onnx/models) to see a list of readily available models to get started with.
     /// Refer to [ONNX](http://onnx.ai) for more information.
     ///
--- a/src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
+++ b/src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
@@ -707,11 +707,15 @@ namespace Microsoft.ML.Transforms.Onnx
     ///
     /// Supports inferencing of models in ONNX 1.6 format (opset 11), using the
     /// [Microsoft.ML.OnnxRuntime](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime/) library (version 1.2.0).
-    /// Models are scored on CPU by default. If GPU execution is needed (optional), use the
-    /// NuGet package available at [Microsoft.ML.OnnxRuntime.Gpu](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu/) (version 1.2.0) instead of the Microsoft.ML.OnnxRuntime nuget (which is for CPU processing)
-    /// and download [CUDA 10.1 Toolkit](https://developer.nvidia.com/cuda-downloads) and [cuDNN 7.6.5](https://developer.nvidia.com/cudnn) (as indicated on [Onnxruntime's documentation](https://github.com/Microsoft/onnxruntime#default-gpu-cuda)).
+    /// Models are scored on CPU by default.
+    ///
+    /// If GPU execution is needed (optional), use the
+    /// NuGet package available at [Microsoft.ML.OnnxRuntime.Gpu](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu/) (version 1.2.0) instead of the Microsoft.ML.OnnxRuntime nuget (which is for CPU processing). To use Microsoft.ML.OnnxRuntime.Gpu
+    /// you need to have a CUDA supported GPU (for reference see this list of [supported GPUs](https://developer.nvidia.com/cuda-gpus#compute)), then download [CUDA 10.1 Toolkit](https://developer.nvidia.com/cuda-downloads) and [cuDNN 7.6.5](https://developer.nvidia.com/cudnn) (as indicated on [Onnxruntime's documentation](https://github.com/Microsoft/onnxruntime#default-gpu-cuda)).
     /// Set parameter 'gpuDeviceId' to a valid non-negative integer. Typical device ID values are 0 or 1.
+    ///
     /// The inputs and outputs of the ONNX models must be Tensor type. Sequence and Maps are not yet supported.
+    ///
     /// OnnxRuntime works on Windows, MacOS and Ubuntu 16.04 Linux 64-bit platforms.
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
```diff
@@ -702,17 +702,17 @@ public NamedOnnxValue GetNamedOnnxValue()
...
     /// | Exportable to ONNX | No |
     ///
-    /// Supports inferencing of models in ONNX 1.2 and 1.3 format (opset 7, 8 and 9), using the
-    /// [Microsoft.ML.OnnxRuntime](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime/) library.
+    /// Supports inferencing of models in ONNX 1.6 format (opset 11), using the
+    /// [Microsoft.ML.OnnxRuntime](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime/) library (version 1.2.0).
     /// Models are scored on CPU by default. If GPU execution is needed (optional), use the
-    /// NuGet package available at [Microsoft.ML.OnnxRuntime.Gpu](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu/)
-    /// and download [CUDA 9.1 Toolkit](https://developer.nvidia.com/cuda-downloads) and [cuDNN](https://developer.nvidia.com/cudnn).
+    /// NuGet package available at [Microsoft.ML.OnnxRuntime.Gpu](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu/) 1.2.0 instead
```
Lynx1820: What is the "1.2.0 instead" indicating? It's unclear to me.

src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
```diff
@@ -702,17 +702,17 @@ public NamedOnnxValue GetNamedOnnxValue()
     /// | Does this estimator need to look at the data to train its parameters? | No |
     /// | Input column data type | Known-sized vector of <xref:System.Single> or <xref:System.Double> types |
     /// | Output column data type | As specified by the ONNX model |
-    /// | Required NuGet in addition to Microsoft.ML | Microsoft.ML.OnnxTransformer (always),  Microsoft.ML.OnnxRuntime.Gpu (only if GPU processing is used) |
+    /// | Required NuGet in addition to Microsoft.ML | Microsoft.ML.OnnxTransformer (always),  either Microsoft.ML.OnnxRuntime 1.2.0 (for CPU processing) or Microsoft.ML.OnnxRuntime.Gpu 1.2.0 (only if GPU processing is desired) |
```
Lynx1820: What's the reason for changing from "used" to "desired"?

src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
antoniovs1029: Instead of the Microsoft.ML.OnnxRuntime package that was mentioned previously in that paragraph. I will explicitly mention that name to make it clearer.

src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
antoniovs1029: I just found it confusing because you can't use GPU processing without the nuget, so you should install it before it is "used".

Do you think it's clearer with 'used'?

Lynx1820: hmm would "available" be a better word? Doesn't ml.net fail if there isn't a GPU, but you try to use the nuget?

luisquintanilla:
```suggestion
    /// | Required NuGet in addition to Microsoft.ML | Microsoft.ML.OnnxTransformer (always),  either Microsoft.ML.OnnxRuntime 1.2.0 (for CPU processing) or Microsoft.ML.OnnxRuntime.Gpu 1.2.0 (for GPU processing if GPU is available) |
```

src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
```diff
@@ -702,17 +702,17 @@ public NamedOnnxValue GetNamedOnnxValue()
...
     /// | Exportable to ONNX | No |
     ///
-    /// Supports inferencing of models in ONNX 1.2 and 1.3 format (opset 7, 8 and 9), using the
-    /// [Microsoft.ML.OnnxRuntime](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime/) library.
+    /// Supports inferencing of models in ONNX 1.6 format (opset 11), using the
+    /// [Microsoft.ML.OnnxRuntime](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime/) library (version 1.2.0).
     /// Models are scored on CPU by default. If GPU execution is needed (optional), use the
-    /// NuGet package available at [Microsoft.ML.OnnxRuntime.Gpu](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu/)
-    /// and download [CUDA 9.1 Toolkit](https://developer.nvidia.com/cuda-downloads) and [cuDNN](https://developer.nvidia.com/cudnn).
+    /// NuGet package available at [Microsoft.ML.OnnxRuntime.Gpu](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu/) (version 1.2.0) instead of the Microsoft.ML.OnnxRuntime nuget (which is for CPU processing)
```
natke: Should we say somewhere that the GPU needs to be CUDA supported?

antoniovs1029: Well, I don't think it's necessary since the instructions already mention that CUDA is necessary. Do you think I should mention explicitly that the GPU needs to be CUDA supported?

src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
antoniovs1029: So I've accepted Luis' suggestion 😄 

Regarding your questions, @natke 
a) I hadn't tested the case of installing both, because I was told by @harishsk that users should install only one. I have just made a quick test installing both on my laptop (without GPU), and it doesn't throw any errors when running, so I guess it is simply using the non-GPU nuget. I am having some troubles using my remote GPU computer, so I haven't had the chance to test it there.
b) If I install the GPU nuget on a computer without GPU, the following error is thrown when _running_ the code:

```
    System.TypeInitializationException : The type initializer for 'Microsoft.ML.OnnxRuntime.NativeMethods' threw an exception.
    ---- System.DllNotFoundException : Unable to load DLL 'onnxruntime' or one of its dependencies: The specified module could not be found. (Exception from HRESULT: 0x8007007E)
```

Reason for this is that the onnxruntime DLL has a dependency on CUDA and CuDNN, so if those aren't installed, it can't load onnxruntime either.

antoniovs1029: I have tested this now on my remote GPU. And when installing both nugets, the GPU nuget gets ignored when running a sample code, it throws the exception that is expected when trying to use a GPU without the GPU nuget:
`Unable to find an entry point named 'OrtSessionOptionsAppendExecutionProvider_CUDA' in DLL 'onnxruntime'`

So in general it seems that when installing both nugets, the GPU one gets completely ignored.

src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
natke: If I am working on my current mac machine (which has an Intel GPU, not compatible with CUDA as far as I know) I might read these instructions and not realize the compatibility issues (until I try and install CUDA), since we advertise ML.NET as cross platform.

natke: Perhaps we could add a link to https://developer.nvidia.com/cuda-gpus#compute? @antoniovs1029?

antoniovs1029: I see. So should I include a link to the list of CUDA's supported GPUs (https://www.geforce.com/hardware/technology/cuda/supported-gpus?field_gpu_type_value=All&page=1 ) Or would it be enough to mention that it will only work for GPUs that support CUDA?

antoniovs1029: Thanks, I've added that link. Please let me know if you think there's more I need to change.

src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
```diff
@@ -707,11 +707,15 @@ public NamedOnnxValue GetNamedOnnxValue()
     ///
     /// Supports inferencing of models in ONNX 1.6 format (opset 11), using the
     /// [Microsoft.ML.OnnxRuntime](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime/) library (version 1.2.0).
-    /// Models are scored on CPU by default. If GPU execution is needed (optional), use the
-    /// NuGet package available at [Microsoft.ML.OnnxRuntime.Gpu](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu/) (version 1.2.0) instead of the Microsoft.ML.OnnxRuntime nuget (which is for CPU processing)
-    /// and download [CUDA 10.1 Toolkit](https://developer.nvidia.com/cuda-downloads) and [cuDNN 7.6.5](https://developer.nvidia.com/cudnn) (as indicated on [Onnxruntime's documentation](https://github.com/Microsoft/onnxruntime#default-gpu-cuda)).
+    /// Models are scored on CPU by default.
+    ///
+    /// If GPU execution is needed (optional), use the
```
natke:
```suggestion
    /// To run on a GPU, use the
```

src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
```diff
@@ -707,11 +707,15 @@ public NamedOnnxValue GetNamedOnnxValue()
...
+    /// NuGet package available at [Microsoft.ML.OnnxRuntime.Gpu](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu/) (version 1.2.0) instead of the Microsoft.ML.OnnxRuntime nuget (which is for CPU processing). To use Microsoft.ML.OnnxRuntime.Gpu
```
natke:
```suggestion
    /// NuGet package [Microsoft.ML.OnnxRuntime.Gpu](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu/) (version 1.2.0) instead of the Microsoft.ML.OnnxRuntime nuget (which is for CPU processing). Microsoft.ML.OnnxRuntime.Gpu
```

src/Microsoft.ML.OnnxTransformer/OnnxTransform.cs
```diff
@@ -707,11 +707,15 @@ public NamedOnnxValue GetNamedOnnxValue()
...
+    /// you need to have a CUDA supported GPU (for reference see this list of [supported GPUs](https://developer.nvidia.com/cuda-gpus#compute)), then download [CUDA 10.1 Toolkit](https://developer.nvidia.com/cuda-downloads) and [cuDNN 7.6.5](https://developer.nvidia.com/cudnn) (as indicated on [Onnxruntime's documentation](https://github.com/Microsoft/onnxruntime#default-gpu-cuda)).
```
natke:
```suggestion
    /// requires a [CUDA supported GPU](https://developer.nvidia.com/cuda-gpus#compute), the [CUDA 10.1 Toolkit](https://developer.nvidia.com/cuda-downloads), and [cuDNN 7.6.5](https://developer.nvidia.com/cudnn) (as indicated on [Onnxruntime's documentation](https://github.com/Microsoft/onnxruntime#default-gpu-cuda)).
```

PR #4967, opened 2020-03-24
Updated build docs for .NET Core 3.1

mstfbl:
Updated Windows and Unix build docs for .NET Core 3.1.

Some specific things to note:

- In addition to other platforms, .NET Core 3.1 supports Ubuntu **16.04**+ and macOS 10.13 (High Sierra)
- [Visual Studio 2019 Version **16.4 or higher** is needed for .NET Core 3.1.
- /src files always build on both `netstandard2.0 `and `netcoreapp3.1`, and /tests only build on one target framework at a time: `netcoreapp2.1`, `net461 `or `netcoreapp3.1`

### DIFF
--- a/docs/building/netcoreapp3.1-instructions.md
+++ b/docs/building/netcoreapp3.1-instructions.md
@@ -1,4 +1,9 @@
-ML.NET now builds for .NET Core 3.1 by default. However to run tests on .NET Core 3.1, you need to do a few manual steps.
+ML.NET source code files build for .NET Core 3.1 and .NET Core 2.0. However, ML.NET test files only build on one target framework below at a time:
+- .NET Core 2.1
+- .NET Core 3.1
+- .NET Framework 4.6.1
+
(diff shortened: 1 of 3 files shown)


docs/building/netcoreapp3.1-instructions.md
```diff
@@ -1,4 +1,4 @@
-ML.NET now builds for .NET Core 3.0 by default. However to run tests on .NET Core 3.0, you need to do a few manual steps.
+ML.NET now builds for .NET Core 3.1 by default. However to run tests on .NET Core 3.1, you need to do a few manual steps.
```
harishsk: Is this accurate? I think we build for netcoreapp21 when you run build.

mstfbl: The src files always build on both netstandard2.0 and netcoreapp3.1. Meanwhile the tests only build on one target framework at a time: netcoreapp2.1, net461 or netcoreapp3.1 (Thanks @eerhardt !) I updated this document to reflect this.

PR #4968, opened 2020-03-24
upgrade benchmark dotnet package to latest version

frank-dong-ms:
As we have upgrade CI run from netcore 3.0 to netcore 3.1, we need also to upgrade benchmark dotnet to latest version which has netcore 3.1 support

### DIFF
--- a/test/Microsoft.ML.Benchmarks/TextPredictionEngineCreation.cs
+++ b/test/Microsoft.ML.Benchmarks/TextPredictionEngineCreation.cs
@@ -11,7 +11,7 @@ using Microsoft.ML.Trainers;
 
 namespace micro
 {
-    [CoreJob]
+    [SimpleJob]
     public class TextPredictionEngineCreationBenchmark
     {
         private MLContext _context;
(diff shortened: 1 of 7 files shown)


test/Microsoft.ML.Benchmarks/TextPredictionEngineCreation.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace micro
 {
-    [CoreJob]
+    [SimpleJob]
```
adamsitnik: what is the purpose of this attribute? So far it was set to run only on `.NET Core`. If we don't want to run it only for .NET Core, we can just remove the attribute.

frank-dong-ms: yes, as far as I know benchmark only runs on dotnet core

---
In reply to: [397732628](https://github.com/dotnet/machinelearning/pull/4968#discussion_r397732628) [](ancestors = 397732628)

PR #4973, opened 2020-03-25
Disabled AutoFitImageClassificationTrainTest for test stability

mstfbl:
Disabled `AutoFitImageClassificationTrainTest `from runs on CI, as it continues to hang on builds.


harishsk:
@justinormont We are finding that this test is hanging on Windows builds and are disabling it again. It appears that it has the same issues we discussed about adding Dispose calls to AutoML. 
@mstfbl is looking into it and if you have any pointers on where he can start looking, that would be very useful. Thanks.

PR #4974, opened 2020-03-25
Fixes OneHotEncoding Issue

Lynx1820:
OneHotEncoding always increases the dimension by one, which is causing issues with Nimbus.
See: https://github.com/microsoft/NimbusML/issues/472

- Added shape info to make models more easy to read
- Replaced ReduceSum with Squeeze (makes more sense imo)
- Isolated KeyToVector test to verify several outputKind modes - Skipping OneHotEncodingEstimator.OutputKind.Binary for now, since this mode is not working and it'll require implementation of KeyToBinaryVector


test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -569,23 +567,21 @@ public void KeyToVectorWithBagOnnxConversionTest()
...
                 // Evaluate the saved ONNX model using the data used to train the ML.NET pipeline.
                 var onnxEstimator = mlContext.Transforms.ApplyOnnxModel(onnxModelPath);
                 var onnxTransformer = onnxEstimator.Fit(data);
                 var onnxResult = onnxTransformer.Transform(data);
-                CompareSelectedColumns<float>("Score", "Score", transformedData, onnxResult);
-                CompareSelectedColumns<float>("Probability", "Probability", transformedData, onnxResult);
-                CompareSelectedColumns<bool>("PredictedLabel", "PredictedLabel", transformedData, onnxResult);
+                CompareResults("F2", "F2", transformedData, onnxResult);
             }
-            CheckEquality(subDir, onnxTextName);
```
Lynx1820: Removing the pipeline to isolate transformer. OneHotEncoding is already used in a pipeline within RemoveVariablesInPipelineTest.

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -1770,7 +1766,6 @@ private void CompareResults(string leftColumnName, string rightColumnName, IData
             var rightColumn = right.Schema[rightColumnName];
             var leftType = leftColumn.Type.GetItemType();
             var rightType = rightColumn.Type.GetItemType();
-            Assert.Equal(leftType, rightType);
```
Lynx1820: Review: Types don't necessarily have to be equal. For example, an ML.NET type can be Key{uint32} and uint32 in onnx.

PR #4978, opened 2020-03-27
Fixed & Reactivated AutoFitImageClassificationTrainTest hanging by freeing old Tensor objects

mstfbl:
`AutoFitImageClassificationTrainTest` is occasionally hanging, even after PR #4939 . The issue here is that Tensor objects saved in `ITransformer model` (now renamed as 'estimatorModel`) are not being automatically freed by C#'s Garbage Collector, as these Tensor objects are made in TensorFlow's C libraries. 

**Edited on 4/9/2020**: This PR makes ExperimentResult and CrossValidationExperimentResult implements IDisposable to free remaining C Tensor objects in memory in a deterministic manner. This method of freeing these objects ensures the user will not face null-reference/use-after-free errors when trying to access the model, as this clean up is done after GC runs.

**Edited on 4/9/2020**: I confirmed that this fixes the hanging "out of memory" and/or "long running test" issues by running `AutoFitImageClassificationTrainTest` for 100 iterations, in addition to running all the other unit tests in [this build](https://dev.azure.com/dnceng/public/_build/results?buildId=595447&view=logs&j=9dffbc46-9322-5a58-fb37-6d66c044e90d&t=11098bf6-eb78-583b-8eab-f14f48444a6b). In all of these builds, none of the issues described occur. These builds all time-out because running 100 iterations of `AutoFitImageClassificationTrainTest` takes more than 1 hour.

I have also reactivated the `AutoFitImageClassificationTrainTest` unit test with this fix.

**Edit: **


src/Microsoft.ML.AutoML/Experiment/Runners/RunnerUtil.cs
```diff
@@ -10,6 +10,8 @@ namespace Microsoft.ML.AutoML
 {
     internal static class RunnerUtil
     {
+        private static ITransformer _estimatorModel;
```
harishsk: Why does this need to be a class variable? Previously the estimator and model were local to the function. Can it not remain local to the function?

src/Microsoft.ML.AutoML/Experiment/Runners/RunnerUtil.cs
```diff
@@ -25,21 +27,21 @@ public static (ModelContainer model, TMetrics metrics, Exception exception, doub
             try
             {
                 var estimator = pipeline.ToEstimator(trainData, validData);
-                var model = estimator.Fit(trainData);
+                _estimatorModel = estimator.Fit(trainData);
```
harishsk: Why is it necessary to rename the model? It is causing needless changes elsewhere in the code.

mstfbl: Here I renamed `model` to `estimatorModel` as I think it does a better job of explaining the model that is returned by a estimator.Fit() call. In addition `TrainAndScorePipeline `returned `ModelContainer` variable is named `model`, and I thought that was confusing to have 2 variables named model (one an `ITransformer`, the other is a `ModelContainer`).

Here the options are renaming the returned ModelContainer (which results in the need to modify this name in other places where  `TrainAndScorePipeline ` is called), or renaming the model that is returned by the estimator.Fit() function (which results in the obvious need to rename it in other parts of this code). @justinormont what do you think?

justinormont: No strong opinions on the naming. Is there a convention in ML․NET? 

Generally I think of a fit "pipeline" as a "model". Though aligning on the ML․NET terminology would be good too.

Terms you'll find in the AutoML code:
* `Recipe`: General layout for how the AutoML will solve the problem at hand
* `Pipeline`: A `recipe` which is fit to the current dataset; for instance handling the N categorical columns in the specific dataset
* `Model`: A trained `pipeline` which can now give predictions

See more: https://patents.google.com/patent/US20190018821A1/en

src/Microsoft.ML.AutoML/Experiment/Runners/RunnerUtil.cs
```diff
@@ -48,6 +49,12 @@ public static (ModelContainer model, TMetrics metrics, Exception exception, doub
                 logger.Error($"Pipeline crashed: {pipeline.ToString()} . Exception: {ex}");
                 return (null, null, ex, double.NaN);
             }
+            finally
+            {
+                // Free Tensor objects in model. Tensor objects made in TensorFlow's C
+                // libraries are not automatically cleaned up by C#'s Garbage Collector.
+                (estimatorModel as IDisposable)?.Dispose();
```
justinormont: What occurs when the calling function tries to use the wrapped-and-returned-model? Unless the TF components are recreated behind the scenes, I would expect a null reference or use-after-free.

src/Microsoft.ML.AutoML/Experiment/Runners/RunnerUtil.cs
mstfbl: I've removed the `estimatorModel` name, and corrected ModelContainer's name to reflect that it is a `modelContainer `and not just a `model`. However now ModelContainer.Dispose() needs to be called to free these Tensor objects, and I don't know where this should occur. Any ideas?

src/Microsoft.ML.Vision/ImageClassificationTrainer.cs
```diff
@@ -1509,5 +1509,9 @@ public void Dispose()
 
             _isDisposed = true;
         }
+
+        // Finalizer to clean-up remaining Tensor objects generated in TensorFlow C libraries
+        // not cleaned up by GC
+        ~ImageClassificationModelParameters() => Dispose();
```
justinormont, line 1515: Congrats on this working out! The usability will be much better, plus cleaner code-wise. Otherwise we have to make many classes in ML․NET `IDisposible` plus teach users the cases where they need to call `Dispose()`.

Might want to run locally and watch the memory usage over the run to see that it's stable after the 1st model is created (in a run of 100 image models). I'd expect it will oscillate but the peaks will remain stable; though the GC can delay the call to the finalizer.

After you're sure the leak is properly handled, I'd remove the `IDisposible` and `Dispose()` in this class and have the finalizer directly do the work inside of the current `Dispose()`:

```c#
~ImageClassificationModelParameters() {
  if (_session?.graph != IntPtr.Zero)
  {
    _session.graph.Dispose();
  }
  if (_session != null && _session != IntPtr.Zero)
  {
    _session.close();
  }
}
```

Then start removing ML․NET's recently added `IDisposible` which were added to support the disposing of the TF models.

mstfbl:
I am now disposing of Tensor models and their C-library objects in `ExperimentResult.cs` and `CrossValidateExperimentResult.cs`, and have added manual `result.Dispose` commands to the AutoML test cases. I have tested this approach in my debugging PR, and saw that no out-of-memory errors occured. Here's that test build of running `AutoFitImageClassificationTrainTest` for 100 iterations: https://dev.azure.com/dnceng/public/_build/results?buildId=595447&view=results

test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
```diff
@@ -119,6 +121,7 @@ public void AutoFitRegressionTest()
                     new ColumnInformation() { LabelColumnName = DatasetUtil.MlNetGeneratedRegressionLabel });
 
             Assert.True(result.RunDetails.Max(i => i.ValidationMetrics.RSquared > 0.9));
+            result.Dispose();
```
justinormont, line 124: A user will need to return their model from a function like this, and the `result.Dispose()` will dispose of it (and all the other models) before it can be returned. 

Can you modify this unit test run within a function which returns the one best model to the parent function, and cleans up the rest of the models trained?

mstfbl: I believe the `bestRun `model is stored as a copy of the model of the best run in `runDetails`. So would it be enough to only Dispose of the models in `runDetails `and not dispose of `bestRun`'s model?
  
Here are the hash codes of the models of the two models, one is in `bestRun `and one is the only model in `runDetails`. As the hash codes are different, they should be distinct (but the same) models. `Object.ReferenceEquals` also confirms this.
diff

src/Microsoft.ML.AutoML/Experiment/ModelContainer.cs
```diff
@@ -30,8 +31,10 @@ internal ModelContainer(MLContext mlContext, FileInfo fileInfo, ITransformer mod
             }
         }
 
-        public ITransformer GetModel()
+        public ITransformer GetModel(bool isModelDisposed)
         {
```
harishsk, line 35: The model is owned by the ModelContainer.   Shouldn't it know whether it was disposed or not?

src/Microsoft.ML.AutoML/Experiment/ModelContainer.cs
```diff
@@ -30,8 +31,10 @@ internal ModelContainer(MLContext mlContext, FileInfo fileInfo, ITransformer mod
...
+            if (isModelDisposed)
+                throw new ObjectDisposedException("Trained model was disposed before");
             // If model stored in memory, return it
```
harishsk, line 38: Why is it necessary to throw this exception specifically? If the user tried to access a disposed object, won't the framework automatically throw this exception?

src/Microsoft.ML.AutoML/API/ExperimentResults/CrossValidationExperimentResult.cs
```diff
@@ -36,5 +37,31 @@ public class CrossValidationExperimentResult<TMetrics>
...
+        {
+            if (_disposed)
+                return;
+            foreach (var runDetail in RunDetails)
+            {
+                foreach (var result in runDetail.Results)
+                {
+                    (result.Model as IDisposable)?.Dispose();
+                    result.IsModelDisposed = true;
+                }
```
harishsk, line 61: This really doesn't resolve the problem of one of the elements of the enumerable needing to have a different lifetime than the rest of the items.

src/Microsoft.ML.AutoML/Experiment/ModelContainer.cs
mstfbl: There is no warning given when one attempts to use a disposed object, so throwing this exception is necessary.

src/Microsoft.ML.AutoML/API/ExperimentResults/CrossValidationExperimentResult.cs
mstfbl: In that case, it makes sense to dispose of the entire `RunDetail runDetails` object instead of just the model, as I did just before [this commit](https://github.com/dotnet/machinelearning/pull/4978/commits/45681b424d09c03b2ed5c9764c8ee54eee8ce580). This is a bit inefficient as its disposing more than just the Model, but it ensures all the elements in the enumerable have the same lifetime.

harishsk: This class holds both an enumerable of IDisposable objects and a separate reference to one of the elements in that enumerable. This makes situation tricky because the requirement now is that the two elements have separate lifetimes. There isn't much difference between this implementation and the previous one because the BestRun model is a reference to an element in the enumerable and when that gets disposed, BestRun will too. 

This still doesn't meet @justinormont's main concern that the user might want to save BestRun and use it after the ExperimentResult object has been disposed. That use case is valid but it goes to counter to the common expectation that when an object is disposed, all its containing objects are disposed. The correct way to resolve this would require the following:

* BestRun is not a reference to an element in the enumerable but a separate copy. 
* When ExperimentResult / CrossValidationExperimentResult are diposed, they should dispose all the elements in the enumerable and BestRun
* But the user might want to manage memory more carefully. So they must be given an option to dispose the enumerable independently if they wish (without disposing the whole ExperimentResult / CrossValidationExperimentResult object)

cc: @justinormont 
Please let me know what you think. 




---
In reply to: [408451825](https://github.com/dotnet/machinelearning/pull/4978#discussion_r408451825) [](ancestors = 408451825)

mstfbl: I believe the `BestRun` model is actually a separate copy and not a reference, as I explained why [here in my comment above](https://github.com/dotnet/machinelearning/pull/4978#discussion_r406985346). Please let me know if this isn't correct.

src/Microsoft.ML.AutoML/API/ExperimentResults/CrossValidationExperimentResult.cs
```diff
@@ -35,6 +36,46 @@ public class CrossValidationExperimentResult<TMetrics>
         {
             RunDetails = runDetails;
             BestRun = bestRun;
+            _disposed = false;
+            _disposedRunDetails = false;
```
harishsk, line 40: These are not necessary. Boolean values are initialized to false by default.

src/Microsoft.ML.AutoML/API/ExperimentResults/CrossValidationExperimentResult.cs
```diff
@@ -35,6 +36,46 @@ public class CrossValidationExperimentResult<TMetrics>
...
+            if (!_disposedRunDetails)
+                (RunDetails as IDisposable)?.Dispose();
+            (BestRun as IDisposable)?.Dispose();
+            _disposed = true;
+            _disposedRunDetails = true;
+        }
+
+        /// <summary>
+        /// Releases unmanaged Tensor objects in models stored in RunDetail instances
+        /// </summary>
```
harishsk, line 67: Please add a more detailed description here that this is provided to support use cases where the user might want different lifetimes for the best model and the rest of the experimental runs.

harishsk:
>             DataViewSchema modelInputSchema,

If you are not going to support modelFileInfo being null, you should throw an exception if it is null

---
Refers to: src/Microsoft.ML.AutoML/Experiment/Runners/RunnerUtil.cs:22 in 087c0d5. [](commit_id = 087c0d5e386c4c2d0c321ce90ac94f0df29f115b, deletion_comment = False)

test/Microsoft.ML.AutoML.Tests/AutoFitTests.cs
```diff
@@ -36,6 +36,7 @@ public void AutoFitBinaryTest()
             Assert.NotNull(result.BestRun.Estimator);
             Assert.NotNull(result.BestRun.Model);
             Assert.NotNull(result.BestRun.TrainerName);
+            result.Dispose();
         }
```
harishsk, line 40: You can also use "using"

mstfbl:
>             DataViewSchema modelInputSchema,

With the current implementation, modelFileInfo is never null as in the case where the given root directory is null, a temp file in a temp path is returned.

---
In reply to: [614421392](https://github.com/dotnet/machinelearning/pull/4978#issuecomment-614421392) [](ancestors = 614421392)

---
Refers to: src/Microsoft.ML.AutoML/Experiment/Runners/RunnerUtil.cs:22 in 087c0d5. [](commit_id = 087c0d5e386c4c2d0c321ce90ac94f0df29f115b, deletion_comment = False)

src/Microsoft.ML.AutoML/API/ExperimentResults/ExperimentResult.cs
```diff
@@ -36,5 +37,46 @@ public class ExperimentResult<TMetrics>
...
+        /// Releases unmanaged Tensor objects in models stored in RunDetail instances
+        /// </summary>
+        /// <remarks>
+        /// Invocation of DisposeRunDetails() is necessary to clean up remaining C library
+        /// Tensor objects and avoid a memory leak. Compared to Dispose(), DisposeRunDetails()
+        /// only disposes of RunDetails so that the best determined model in BestRun can have
+        /// a different lifetime than models in other experimental runs.
+        /// </remarks>
+        public void DisposeRunDetails()
+        {
```
harishsk, line 74: Is there any code that exercises this code path?

src/Microsoft.ML.AutoML/API/ExperimentResults/ExperimentResult.cs
```diff
@@ -36,5 +37,46 @@ public class ExperimentResult<TMetrics>
...
+        /// Releases unmanaged Tensor objects in models stored in RunDetail and BestRun instances
+        /// </summary>
+        /// <remarks>
+        /// Invocation of Dispose() is necessary to clean up remaining C library Tensor objects and
+        /// avoid a memory leak. Dispose() guarantees that models in BestRun and models in RunDetails
+        /// have the same lifetime.
+        /// </remarks>
+        public void Dispose()
+        {
+            if (_disposed)
```
harishsk, line 55: Readability nit: Dispose can call DisposeRunDetails.

harishsk:
>             DataViewSchema modelInputSchema,

Yes, but that knowledge does not exist here in this function. modelFileInfo gets passed without any validation to ModelContainer.  Tomorrow, someone else may make a code change without realizing that. 
It is good practice to validate your function parameters. Please add either a debug assert explicitly throw if parameter is null. 

---
In reply to: [614783589](https://github.com/dotnet/machinelearning/pull/4978#issuecomment-614783589) [](ancestors = 614783589,614421392)

---
Refers to: src/Microsoft.ML.AutoML/Experiment/Runners/RunnerUtil.cs:22 in 087c0d5. [](commit_id = 087c0d5e386c4c2d0c321ce90ac94f0df29f115b, deletion_comment = False)

PR #4993, opened 2020-04-01
Update Featurizers to latest nuget.

michaelgsharp:
This code brings the Featurizers project up to date with the latest code from the native featurizers.
It also adds support for CentOS7 that was not there originally.


src/Microsoft.ML.Featurizers/Common.cs
```diff
@@ -8,6 +8,16 @@
 
 namespace Microsoft.ML.Featurizers
 {
+    // Training state of the native featurizers
+    internal enum TrainingState
+    {
```
harishsk, line 13: Is it possible to have better names than Finished and Completed which are almost synonymous? How about enum TrainingState { NotStarted, InProgress, Completed, TransformerCreated}

Other suggestions for the final enum: ModelCreated, Finalized, ModelFinalized, etc.

src/Microsoft.ML.Featurizers/Common.cs
```diff
@@ -89,40 +99,36 @@ protected override bool ReleaseHandle()
     internal class ErrorInfoStringSafeHandle : SafeHandleZeroOrMinusOneIsInvalid
     {
         [DllImport("Featurizers", EntryPoint = "DestroyErrorInfoString", CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity]
-        private static extern bool DestroyErrorInfoString(IntPtr errorString, IntPtr errorStringSize);
+        private static extern bool DestroyErrorInfoString(IntPtr errorString);
```
harishsk, line 103: Can you please explain why the size is no longer necessary?

src/Microsoft.ML.Featurizers/Common.cs
```diff
@@ -221,30 +224,25 @@ internal static TypeId GetNativeTypeIdFromType(this Type type)
             throw new InvalidOperationException($"Unsupported type {type}");
         }
 
-        // The Native Featurizers do not currently support CentOS7, this method checks the OS and returns true if it is CentOS7.
-        internal static bool OsIsCentOS7()
+        internal static string PointerToString(IntPtr data)
         {
```
harishsk, line 228: It appears we are walking the native C string to find the null pointer and the length. Is this a perf optimization? If so have we measured it?

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -346,46 +341,17 @@ public TransformedDataSafeHandle(IntPtr handle, DestroyTransformedDataNative des
...
             }
         }
 
         #endregion
 
-        #region TimePoint
-
-        // Exact native representation
-        [StructLayout(LayoutKind.Sequential, Pack = 1)]
-        internal struct NativeTimePoint
```
harishsk, line 359: Why is packing no longer necessary? Has the layout changed on the native side?

harishsk:
>                     data.MoveNext();

Please check return value of MoveNext() and take appropriate action


---
Refers to: src/Microsoft.ML.Featurizers/RobustScaler.cs:457 in 35d33a5. [](commit_id = 35d33a5c8cb1d587ff73de25cb7e73ce8c32c7b2, deletion_comment = False)

harishsk:
>                             OnDataCompletedHelper(estimatorHandle, out errorHandle);

Assert(fitResult != FitResult.ResetAndContinue) would be helpful

Also a check that the user hasn't accidentally sent in an empty dataview

---
Refers to: src/Microsoft.ML.Featurizers/RobustScaler.cs:484 in 35d33a5. [](commit_id = 35d33a5c8cb1d587ff73de25cb7e73ce8c32c7b2, deletion_comment = False)

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -570,8 +572,7 @@ internal byte[] CreateTransformerSaveData()
                 if (!success)
                     throw new Exception(GetErrorDetailsAndFreeNativeMemory(errorHandle));
 
-                using (var savedDataHandle = new SaveDataSafeHandle(buffer, bufferSize))
-                {
+                using (var savedDataHandle = new SaveDataSafeHandle(buffer, bufferSize)) {
```
harishsk, line 575: I don't think this is the C# convention. Can you please use the previous format of braces on a separate line?

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -606,10 +606,15 @@ internal abstract class TypedColumn<T> : TypedColumn
...
-
+            private readonly int _intPtrSize;
+            private readonly int _structSize;
             internal LongTypedColumn(string source, string prefix) :
                 base(source, prefix)
             {
+                _intPtrSize = IntPtr.Size;
+
+                // The native struct is 25 bytes + 4 size_t.
+                _structSize = 25 + (_intPtrSize * 4);
```
harishsk, line 617: Please fix as discussed to use sizeof.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
```diff
@@ -451,38 +417,88 @@ internal TimePoint(ReadOnlySpan<byte> rawData, int intPtrSize)
...
-#endif
+                return result;
             }
 
         };
 
-        #endregion
+        #region Union
+        // The folowing structs/enums are to enable us to simulate Unions on the Native side
+
```
harishsk, line 442: Why are these unions instead of being just separate data types like all other transformers?

src/Microsoft.ML.Featurizers/Common.cs
michaelgsharp: We have changed all strings to be c strings with the final character as null. This made interop of other types easier (like vectors of string). So now instead of getting the size back we just scan until the null character.

src/Microsoft.ML.Featurizers/Common.cs
michaelgsharp: This is tied to the previous comment. Since we are now not returning the size we need to walk the c string to find the end. Not a perf optimization, just needed based on interop changes.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Per offline discussion this whole section needs to be re-added in. It was an accidental removal due to a rebase error.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Good catch. Fixed that here and everywhere else it occurs.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: This change was done on the native side several months ago, so I am not 100% positive on all the reasons. One idea though was to reduce the amount of interop functions. This also only has 2 possible input types on the Native side. Though aren't very good reasons, but again it was a while ago and we can't remember the exact reasons we went that route. We may, at some point in the future, go to variant types in the Native side, and if we do than this how we would do it.

src/Microsoft.ML.Featurizers/DateTimeTransformer.cs
michaelgsharp: Fixed.

src/Microsoft.ML.Featurizers/Common.cs
michaelgsharp: These are the same as the native side, so I think its best to leave them like this. I do agree though that the names are super close, so I will bring this up with my team.

PR #4994, opened 2020-04-01
Bitmap(file) locks the file.

### LINKED ISSUE #4585: Cannot delete image file passsed to PredictionEngine for Image Classification after disposing the engine
### System information

- **OS version/distro**: Windows 10 Pro 10.0.18363
- **.NET Version (eg., dotnet --info)**:   3.1.0

### Issue

- **What did you do?**: I've tried to delete an image file used by PredictionEngine after disposing the engine
- **What happened?**: An exception was thorwn: "The process cannot access the file  because it is being used by another process."
- **What did you expect?**: Image file deletion

### Source code / logs
```
private ImagePrediction ClassifySingleImage(MLContext mLContext, ITransformer model, string imagePath)
{
    var imageData = new ImageData()
    {
        ImagePath = imagePath
    };
    var predictor = mLContext.Model.CreatePredictionEngine<ImageData, ImagePrediction>(model);
    var prediction = predictor.Predict(imageData);
    predictor.Dispose();
    System.IO.File.Delete(imagePath);  // throws a FileIO exception
    return prediction;
}
```

ganik:
Fixes #4585 

Construct "new Bitmap(file)" locks the underlying file for the lifetime of Bitmap. Use workaround suggested https://stackoverflow.com/questions/6576341/open-image-from-file-then-release-lock

### DIFF
--- a/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
+++ b/src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
@@ -238,8 +238,10 @@ namespace Microsoft.ML.Data
                                 path = Path.Combine(_parent.ImageFolder, path);
 
                             // to avoid locking file, use the construct below to load bitmap
-                            using(var bmpTemp = new Bitmap(path))
-                                dst = new Bitmap(bmpTemp) { Tag = path };
+                            var bytes = File.ReadAllBytes(path);
+                            var ms = new MemoryStream(bytes);
+                            dst = (Bitmap)Image.FromStream(ms);
+                            dst.Tag = path;
 
                             // Check for an incorrect pixel format which indicates the loading failed
                             if (dst.PixelFormat == System.Drawing.Imaging.PixelFormat.DontCare)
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.ImageAnalytics/ImageLoader.cs
```diff
@@ -238,8 +238,10 @@ private Delegate MakeGetterImageDataViewType(DataViewRow input, int iinfo, Func<
                                 path = Path.Combine(_parent.ImageFolder, path);
 
                             // to avoid locking file, use the construct below to load bitmap
-                            using(var bmpTemp = new Bitmap(path))
-                                dst = new Bitmap(bmpTemp) { Tag = path };
+                            var bytes = File.ReadAllBytes(path);
```
harishsk, line 241: This is interesting. Can you please help me understand why the change was necessary?

ganik: The first change was not preserving the original pixel encoding (format) from the file. It looks like the copy constructor was modifying the format.  The second iteration preserves original encoding by moving all the bytes into the byte stream, so only underlying media is changed (memory vs file). See https://stackoverflow.com/questions/4803935/free-file-locked-by-new-bitmapfilepath for more details

PR #4997, opened 2020-04-03
fix tensorflow test hanging issue

frank-dong-ms:
fixed 2 issue of tensorflow tests:
1. fix issue in Resource Manager, use ReadAsync instead of Read to let thread timeout corrently
2. move download ImageSet to constructor of test class to avoid duplicate download for different tests

### DIFF
--- a/src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
+++ b/src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
@@ -127,14 +127,22 @@ namespace Microsoft.ML.Internal.Utilities
 
             for (int i = 0; i < retryTimes; ++i)
             {
-                var thisDownloadResult = await DownloadFromUrlAsync(env, ch, url, fileName, timeout, filePath);
+                try
+                {
+                    var thisDownloadResult = await DownloadFromUrlAsync(env, ch, url, fileName, timeout, filePath);
 
-                if (string.IsNullOrEmpty(thisDownloadResult))
-                    return thisDownloadResult;
-                else
-                    downloadResult += thisDownloadResult + @"\n";
+                    if (string.IsNullOrEmpty(thisDownloadResult))
+                        return thisDownloadResult;
+                    else
+                        downloadResult += thisDownloadResult + @"\n";
 
-                await Task.Delay(10 * 1000);
+                    await Task.Delay(10 * 1000);
+                }
+                catch (Exception ex)
+                {
+                    // ignore any Exception and retrying download
+                    ch.Warning($"{i+1} - th try: Dowload {fileName} from {url} fail with exception {ex.Message}");
+                }
             }
 
             return downloadResult;
--- a/src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
+++ b/src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
@@ -281,12 +281,18 @@ namespace Microsoft.ML.Internal.Utilities
                     var buffer = new byte[blockSize];
                     long total = 0;
 
-                    var task = s.ReadAsync(buffer, 0, blockSize);
-                    task.Wait(ct);
-                    int count = task.Result;
                     // REVIEW: use a progress channel instead.
-                    while (count > 0)
+                    while (true)
                     {
+                        var task = s.ReadAsync(buffer, 0, blockSize, ct);
+                        task.Wait();
+                        int count = task.Result;
+
+                        if(count <= 0)
+                        {
+                            break;
+                        }
+
                         ws.Write(buffer, 0, count);
                         total += count;
                         if ((total - (total / printFreq) * printFreq) <= blockSize)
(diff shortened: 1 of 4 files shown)


test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1250,25 +1263,17 @@ public void TensorFlowStringTest()
...
-                "images");
-
-            //Download the image set and unzip
-            string finalImagesFolderName = DownloadImageSet(
-                imagesDownloadFolderPath);
-
-            string fullImagesetFolderPath = Path.Combine(
-                imagesDownloadFolderPath, finalImagesFolderName);
+            // set timeout to 3 minutes, download sometimes will stuck so set smaller timeout to retry download
+            string timoOutOldValue = Environment.GetEnvironmentVariable(ResourceManagerUtils.TimeoutEnvVariable);
```
harishsk, line 1269: Why do we want do this only for these tests and not for all tests? We can choose a smaller timeout and higher retry count for all downloads.

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
```diff
@@ -1338,25 +1346,17 @@ internal bool ShouldReuse(string workspacePath, string trainSetBottleneckCachedV
...
+            Environment.SetEnvironmentVariable(ResourceManagerUtils.TimeoutEnvVariable, (3 * 60 * 1000).ToString());
```
harishsk, line 1354: Will setting the environment variables here and resetting it at the end of the test have any impact?
It appears that you have moved the download code to the constructor. But the timeout is not being set there. So by the time the code gets here wouldn't the download already have completed with a different timeout value?

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -127,14 +127,22 @@ public async Task<ResourceDownloadResults> EnsureResourceAsync(IHostEnvironment
...
+                        downloadResult += thisDownloadResult + @"\n";
 
-                await Task.Delay(10 * 1000);
+                    await Task.Delay(10 * 1000);
+                }
+                catch (Exception ex)
+                {
+                    // ignore any Exception and retrying download
+                    ch.Warning($"{i+1} - th try: Dowload {fileName} from {url} fail with exception {ex.Message}");
+                }
```
harishsk, line 145: We have lots of trouble with failed downloads. Can we make the exceptions more visible? 

That is, we should log all the failures as you have done. But if in the end the download still fails after the retries, how about throwing an exception that download failed even after retries?

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
frank-dong-ms: this setting (timeout) is hardcoded in our source code as 10 minutes but not a option that user can set at https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Vision/ImageClassificationTrainer.cs#L1321

this setting can be override by set the env variable I'm doing this as if everything is fine 3 minutes is enough if something go wrong then we hope this to fail fast as we are retrying 5 times.

we are not seeing other tests have similar issue so I only do this in affected tests.

test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/TensorflowTests.cs
frank-dong-ms: They are downloading different things.

In constructor we are downloading some image set and all tests use same images so do need to download duplicate images and I move this to constructor. this is refine the code performance and remove unnecessary downloads.

this env variable is to set timeout for some metadata for test and metadata download is causing the hanging.

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
frank-dong-ms: yes, we indeed throw exception if retry still fails at line: https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.TensorFlow/TensorflowUtils.cs#L190

src/Microsoft.ML.Core/Utilities/ResourceManagerUtils.cs
```diff
@@ -268,21 +278,28 @@ private Exception DownloadResource(IHostEnvironment env, IChannel ch, WebClient
                         size = 10000000;
 
                     long printFreq = (long)(size / 10.0);
-                    var buffer = new byte[4096];
+                    var buffer = new byte[blockSize];
                     long total = 0;
-                    int count;
+
+                    var task = s.ReadAsync(buffer, 0, blockSize);
+                    task.Wait(ct);
```
harishsk, line 285: ReadAsync allows you to pass a CancellationToken directly. You can choose to use that.

PR #4998, opened 2020-04-03
Add SrCnnEntireAnomalyEstimator

mengaims:

See Issue #5053 for description of this change.


src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,198 @@
...
+using Microsoft.ML.CommandLine;
+using Microsoft.ML.Data;
+using Microsoft.ML.Runtime;
+
+namespace Microsoft.ML.Transforms.TimeSeries
+{
+
+    public sealed class SrCnnEntireTransformer : OneToOneTransformerBase
+    {
+        internal const string Summary = "This transformer detct anomalies for input timeseries using SRCNN";
```
ganik, line 13: >detct [](start = 58, length = 5)

typo

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,198 @@
...
+            public string Source;
+
+            [Argument(ArgumentType.Required, HelpText = "The name of the new column.", ShortName = "tgt", SortOrder = 2)]
+            public string Target;
+
+            [Argument(ArgumentType.Required, HelpText = "The threshold to determine anomaly, score larger than the threshold is considered as anomaly.",
+                ShortName = "thre", SortOrder = 106)]
+            public double Threshold = 0.3;
+
+            [Argument(ArgumentType.AtMostOnce, HelpText = "The detection mode, affect output vector length",
```
ganik, line 30: >affect [](start = 79, length = 6)

affects

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: Fixed.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: Fixed.

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
```diff
@@ -187,5 +187,33 @@ public static class TimeSeriesCatalog
...
                 confidenceUpperBoundColumn, confidenceLevel, variableHorizon);
+
+        /// <summary>
+        /// Create <see cref="SrCnnEntireAnomalyEstimator"/>, which detects timeseries anomalies for entire input using SRCNN algorithm.
+        /// </summary>
+        /// <param name="catalog">The transform's catalog.</param>
+        /// <param name="outputColumnName">Name of the column resulting from the transformation of <paramref name="inputColumnName"/>.
+        /// The column data is a vector of <see cref="System.Double"/>. The length of this vector varies depending on <paramref name="srCnnDetectMode"/>.</param>
+        /// <param name="inputColumnName">Name of column to transform. The column data must be <see cref="SrCnnTsPoint"/>.</param>
+        /// <param name="threshold">The threshold to determine anomaly, score larger than the threshold is considered as anomaly. Must be in [0,1]. Default value is 0.3.</param>
```
harishsk, line 198: Why the restriction that the inputColumnName must be of type SrCnnTsPoint? The anomaly detectors we already ship do not have this restriction. 

Is it because of the need for a single structure that has both timestamp and the associated value to detect the anomaly on? 

If we have this restriction, how will this affect usage when data is loaded from file through a TextLoader? 

Would it make more sense to have a separate parameter in the API that identifies what the timestamp column is?

src/Microsoft.ML.TimeSeries/SrCnnDataType.cs
```diff
@@ -0,0 +1,100 @@
...
+            Timestamp = timestamp;
+            Value = value;
+        }
+    }
+
+    /// <summary>
+    /// Dataview type of <see cref="SrCnnTsPoint"/>.
+    /// </summary>
+    public sealed class SrCnnTsPointDataViewType : StructuredDataViewType
+    {
```
harishsk, line 50: Why do we need this? And what is it being used for?

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
```diff
@@ -187,5 +187,33 @@ public static class TimeSeriesCatalog
...
+        /// <param name="batchSize">Divide the input data into batches to fit srcnn model.
+        /// When set to -1, use the whole input to fit model instead of batch by batch, when set to a positive integer, use this number as batch size.
```
harishsk, line 200: It appears that it is the batch size that is driving this new estimator. 

Would it be better to call this `SrCnnBatchedAnomalyEstimator` and the function as `DetectAnomalySrCnnInBatchedMode` ?

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+                {
+                    _slotNames = new VBuffer<ReadOnlyMemory<char>>(_parent.OutputLength, new[] { "Is Anomaly".AsMemory(), "Anomaly Score".AsMemory(), "Mag".AsMemory(),
+                        "Expected Value".AsMemory(), "Boundary Unit".AsMemory(), "Upper Boundary".AsMemory(), "Lower Boundary".AsMemory() });
+                }
+                else if (_parent.OutputLength == 4)
+                {
+                    _slotNames = new VBuffer<ReadOnlyMemory<char>>(_parent.OutputLength, new[] { "Is Anomaly".AsMemory(), "Raw Score".AsMemory(), "Mag".AsMemory(),
+                        "Expected Value".AsMemory() });
+                }
+            }
```
harishsk, line 228: Can you please restructure this to initialize array of const strings just once and use references to subsets of the array for each specific case?

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+
+            Host.CheckUserArg(options.BatchSize == -1 || options.BatchSize >= 12, nameof(Options.BatchSize), "BatchSize must be -1 or no less than 12.");
+            BatchSize = options.BatchSize;
+
+            SrCnnDetectMode = options.SrCnnDetectMode;
+            if (SrCnnDetectMode.Equals(SrCnnDetectMode.AnomalyOnly))
+            {
+                Sensitivity = options.Sensitivity;
+                OutputLength = 3;
+            }
```
harishsk, line 102: Can we please create names for these constants of 3, 4, 7 and use those throughout the code?. E.g. AnomalyOnlyVectorLength, AnomalyAndEVVectorLength, and AnomalyAndMarginVectorLength?
(Or something shorter if possible)

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+            private static readonly int _minTrainingPoint = 12;
+            private static readonly int _lookaheadWindowSize = 5;
+            private static readonly int _backAddWindowSize = 5;
+            private static readonly int _avergingWindowSize = 3;
+            private static readonly int _judgementWindowSize = 21;
+            private static readonly double _eps = 1e-8;
+            private static readonly double _deanomalyThreshold = 0.35;
+
+            private static readonly Double[] _factors = new Double[]{
+                    184331.62871148242, 141902.71648305038, 109324.12672037778, 84289.9974713784, 65038.57829581667, 50222.84038287002,
```
harishsk, line 289: Can you please add some comments on how these numbers came about? What do they mean? What would need to be done if there was a bug and if this table needs to be updated....etc?

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+                            trainingData.Add(val.Value);
+                            prevTimestamp = val.Timestamp;
+                        }
+                    }
+                }
+
+                _host.CheckIO(trainingTimestamp.Count >= _minTrainingPoint, "Input must have no less than 12 valid points to fit a model.");
+
+                if (trainingTimestamp.Count >= _minTrainingPoint)
+                {
```
harishsk, line 415: You are throwing an exception above if `trainingTimeStamp.Count` is less than `_minTrainingPoint`.

Either you don't need this if check here or if you keep the check here, you will need an else block and can move the exception to the else block

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
```diff
@@ -187,5 +187,33 @@ public static class TimeSeriesCatalog
             new SsaForecastingEstimator(CatalogUtils.GetEnvironment(catalog), outputColumnName, inputColumnName, windowSize, seriesLength, trainSize,
                 horizon, isAdaptive, discountFactor, rankSelectionMethod, rank, maxRank, shouldStabilize, shouldMaintainInfo, maxGrowth, confidenceLowerBoundColumn,
                 confidenceUpperBoundColumn, confidenceLevel, variableHorizon);
+
+        /// <summary>
+        /// Create <see cref="SrCnnEntireAnomalyEstimator"/>, which detects timeseries anomalies for entire input using SRCNN algorithm.
+        /// </summary>
```
harishsk, line 193: Can we add some references to the algorithm? Any published papers or wikipedia articles?

test/Microsoft.ML.TimeSeries.Tests/TimeSeriesDirectApi.cs
```diff
@@ -543,12 +567,241 @@ public void AnomalyDetectionWithSrCnn()
...
+
+        [Fact]
+        public void AnomalyDetectionWithSrCnnEntireAnomalyOnly()
+        {
+            var ml = new MLContext(1);
+
+            // Generate sample series data with an anomaly
+            DateTime currentTime = new DateTime(2020, 01, 01);
+            var data = new List<TimeSeriesDataWithTimestamp>();
+            for (int index = 0; index < 20; index++)
```
harishsk, line 587: Can we add a test that demonstrates loading data from a file and computing anomalies?

test/Microsoft.ML.TimeSeries.Tests/TimeSeriesDirectApi.cs
```diff
@@ -543,12 +567,241 @@ public void AnomalyDetectionWithSrCnn()
...
+                    Assert.Equal(1, prediction.Prediction[0]);
+                else
+                    Assert.Equal(0, prediction.Prediction[0]);
+                k += 1;
+            }
+        }
+
+        [Fact]
+        public void AnomalyDetectionWithSrCnnEntireAnomalyOnly()
+        {
```
harishsk, line 581: Are the three tests below nearly identical except for the mode?
If so, can you please refactor them using [Theory] instead of [Fact]?

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+
+        internal int BatchSize { get; }
+
+        internal SrCnnDetectMode SrCnnDetectMode { get; }
+
+        internal double Sensitivity { get; }
+
+        internal int OutputLength { get; }
+
+        internal SrCnnEntireModeler AnomalyModel;
```
yaeldekel, line 72: I think all of these can be private readonly fields.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+            public SrCnnDetectMode SrCnnDetectMode = SrCnnDetectMode.AnomalyOnly;
+
+            [Argument(ArgumentType.AtMostOnce, HelpText = "The sensitivity of boundaries.",
+                ShortName = "sens", SortOrder = 404)]
+            public double Sensitivity = 99;
+        }
+
+        internal readonly string InputColumnName;
+
+        internal readonly string OutputColumnName;
```
yaeldekel, line 60: `OneToOneTransformerBase` class has a field called `ColumnPairs` which contains this information.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+            }
+            else
+            {
+                OutputLength = 0;
+            }
+
+            AnomalyModel = new SrCnnEntireModeler(Host, this, input);
+        }
+
+        private SrCnnEntireTransformer(IHost host, ModelLoadContext ctx)
```
yaeldekel, line 122: >SrCnnEntireTransformer [](start = 16, length = 22)

Please add a comment inside this method and in the `SaveModel` method detailing the binary format (for an example of what I mean, see here: [https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.ImageAnalytics/ImageResizer.cs#L231-L238](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.ImageAnalytics/ImageResizer.cs#L231-L238)).

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+            Host.Assert(Threshold >= 0 && Threshold <= 1);
+            Host.Assert(BatchSize == -1 || BatchSize >= 12);
+            Host.Assert(Sensitivity >= 0 && Sensitivity <= 100);
+            Host.Assert(OutputLength == 3 || OutputLength == 4 || OutputLength == 7);
+
+            base.SaveColumns(ctx);
+            ctx.Writer.Write(InputColumnName);
+            ctx.Writer.Write(OutputColumnName);
+            ctx.Writer.Write(Threshold);
+            ctx.Writer.Write((Int32)BatchSize);
```
yaeldekel, line 187: >Int32 [](start = 30, length = 5)

`BatchSize` is an int, no need to cast.

yaeldekel: Same for `OutputLength`.

---
In reply to: [415779054](https://github.com/dotnet/machinelearning/pull/4998#discussion_r415779054) [](ancestors = 415779054)

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+            ctx.Writer.Write((Int32)SrCnnDetectMode);
```
yaeldekel, line 188: >Int32 [](start = 30, length = 5)

Please use `int` instead of `Int32`.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+
+            ctx.SaveModel(AnomalyModel, "SrCnnEntire");
+        }
+
+        private protected override IRowMapper MakeRowMapper(DataViewSchema schema) => new Mapper(this, schema);
+
+        private sealed class Mapper : OneToOneMapperBase
+        {
+            private readonly SrCnnEntireTransformer _parent;
+            private readonly VBuffer<ReadOnlyMemory<Char>> _slotNames;
```
yaeldekel, line 200: >Char [](start = 52, length = 4)

nit: use lower case `char`.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+            private readonly int _inputColumnIndex;
+
+            public Mapper(SrCnnEntireTransformer parent, DataViewSchema inputSchema)
+                : base(parent.Host.Register(nameof(Mapper)), parent, inputSchema)
+            {
+                if (!inputSchema.TryGetColumnIndex(parent.InputColumnName, out _inputColumnIndex))
```
yaeldekel, line 206: >inputSchema [](start = 21, length = 11)

The `base` constructor checks this, no need to check again.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+                    throw Host.ExceptSchemaMismatch(nameof(inputSchema), "input", parent.InputColumnName);
+
+                var colType = inputSchema[_inputColumnIndex].Type;
+                if (!(colType is SrCnnTsPointDataViewType))
+                    throw Host.ExceptSchemaMismatch(nameof(inputSchema), "input", parent.InputColumnName, "SrCnnTsPoint", colType.ToString());
+
```
yaeldekel, line 212: There is a virtual method `CheckInputColumn` in the base class that is called in the `base` constructor, this should be moved there.

yaeldekel: Sorry, I should have said that `CheckInputColumn` is in the base class of the transformer, not the mapper.

---
In reply to: [415788241](https://github.com/dotnet/machinelearning/pull/4998#discussion_r415788241) [](ancestors = 415788241)

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+                var getSrc = input.GetGetter<SrCnnTsPoint>(input.Schema[ColMapNewToOld[iinfo]]);
+
+                disposer = null;
+
+                ValueGetter<VBuffer<double>> del =
+                    (ref VBuffer<double> dst) =>
+                    {
+                        getSrc(ref src);
+                        if (src == null)
+                            return;
```
yaeldekel, line 256: >return [](start = 28, length = 6)

What does `src == null` indicate? Don't you want to assign something to `dst` so that users would know that `src` was `null`? (maybe all zeros? or all `NaN`s?)

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+                        var result = VBufferEditor.Create(ref dst, _parent.OutputLength);
+                        result.Values.Fill(Double.NaN);
+                        Double[] values = new Double[_parent.OutputLength];
```
yaeldekel, line 259: >Double [](start = 24, length = 6)

nit: Please use lower case `double` (throughout the file).

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+                        if (_parent.AnomalyModel.Inference(src, out values))
```
yaeldekel, line 260: >Inference [](start = 49, length = 9)

Please change this method to fill in the values directly into `result.Values` instead of allocating a new array for each row.

yaeldekel: Ok, now that I see that `Inference` looks up the array in a dictionary, I see that it's not possible. But still, you don't need to allocate a new array for values, you can write instead
```
_parent.AnomalyModel.Inference(src, out var values)
```

---
In reply to: [415798844](https://github.com/dotnet/machinelearning/pull/4998#discussion_r415798844) [](ancestors = 415798844)

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+                disposer = null;
+
+                ValueGetter<VBuffer<double>> del =
+                    (ref VBuffer<double> dst) =>
+                    {
+                        getSrc(ref src);
+                        if (src == null)
+                            return;
+                        var result = VBufferEditor.Create(ref dst, _parent.OutputLength);
+                        result.Values.Fill(Double.NaN);
```
yaeldekel, line 258: >Fill [](start = 38, length = 4)

I think it would be more efficient to do this only in the `else` to the `if` below.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+                            }
+                        }
+                        dst = result.Commit();
+                    };
+
+                return del;
+            }
+        }
+
+        internal class SrCnnEntireModeler : ICanSaveModel
```
yaeldekel, line 274: >internal [](start = 8, length = 8)

This class can be private.

Also, the way I understand it, the only method in this class that is used externally is `Inference`, and that method only calls `TryGetValue` into a dictionary. I would consider changing this to a static class that has a static method `Train`, that returns the dictionary itself, and having the transformer use that dictionary directly (and be in charge of saving/loading it).

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,903 @@
...
+                    keyList.Add(pair.Key);
+                    valList.AddRange(pair.Value);
+                }
+                ctx.Writer.WriteLongStream(keyList);
+                ctx.Writer.WriteDoubleStream(valList);
+            }
+
+            public void Train(IDataView data) => Train(new RoleMappedData(data, null, _parent.InputColumnName));
+
+            public void Train(RoleMappedData data)
```
yaeldekel, line 385: >RoleMappedData [](start = 30, length = 14)

There is no need for this overload, you can train directly on the `IDataView` since you know which column you are using as input (`_parent.InputColumnName`).

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
yaeldekel: I think it would be easier and more user friendly to make the input for this transformer be two columns (one of type `DateTime` and one of type `double`. I would even consider allowing the `Timestamp` input to be of type `long`, since `ML.NET`'s support for `DateTime` is very basic (it can only be loaded by `TextLoader` if the values are written in a very specific format, and there are currently no transformers that output `DateTime` type).

---
In reply to: [415369363](https://github.com/dotnet/machinelearning/pull/4998#discussion_r415369363) [](ancestors = 415369363)

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,198 @@
...
+                {
+                    _slotNames = new VBuffer<ReadOnlyMemory<char>>(_parent.OutputLength, new[] { "Alert".AsMemory(), "Raw Score".AsMemory(), "Mag".AsMemory() });
+                }
+                else if (_parent.OutputLength == 7)
+                {
+                    _slotNames = new VBuffer<ReadOnlyMemory<char>>(_parent.OutputLength, new[] { "Is Anomaly".AsMemory(), "Anomaly Score".AsMemory(), "Mag".AsMemory(),
+                        "Expected Value".AsMemory(), "Boundary Unit".AsMemory(), "Upper Boundary".AsMemory(), "Lower Boundary".AsMemory() });
+                }
+                //TODO
+            }
```
harishsk, line 228: Can you please add more details on exactly what the TODO work is?  Or delete them if they are already done.

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
harishsk: I agree with @yaeldekel's suggestion above. 

---
In reply to: [415894364](https://github.com/dotnet/machinelearning/pull/4998#discussion_r415894364) [](ancestors = 415894364,415369363)

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
mengaims: The key difference of this new estimator with the old one is that it implements entire anomaly detection for timeseries, instead of streaming anomaly detection in the old one. The batch size here is an optional parameter. When the timeseries is too long, it's better to divide the timeseries into batches to do entire anomaly detection to get better performance and higher quality.

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
mengaims: Thanks for your suggestions. It's no problem for us to split the SrCnnTsPoint into two columns, since this is not a very complex data structure. But I have some concerns here, to my understanding, TextLoader is something to make user easier to read input data from a file, in yahoo dataset and NAB dataset that PowerBI uses, the timestamps are represented in formatted datetime strings, if we make it long, the user still need to do some conversion and could not get dataview directly. Besides, the long timestamp also have different interpretings, is it Unix timestamp in seconds or in milliseconds, or in microseconds, etc? If TextLoader is a must supported way, could you consider to add some support for DateTime, since if ML.NET decides to support timeseries related models, the needs for DateTime type may be something common in the future.

mengaims: > I agree with @yaeldekel's suggestion above.
> 
> In reply to: [415894364](https://github.com/dotnet/machinelearning/pull/4998#discussion_r415894364) [](ancestors = 415894364,415369363)

Thanks for your suggestions. It's no problem for us to split the SrCnnTsPoint into two columns, since this is not a very complex data structure. But I have some concerns here, to my understanding, TextLoader is something to make user easier to read input data from a file, in yahoo dataset and NAB dataset that PowerBI uses, the timestamps are represented in formatted datetime strings, if we make it long, the user still need to do some conversion and could not get dataview directly. Besides, the long timestamp also have different interpretings, is it Unix timestamp in seconds or in milliseconds, or in microseconds, etc? If TextLoader is a must supported way, could you consider to add some support for DateTime, since if ML.NET decides to support timeseries related models, the needs for DateTime type may be something common in the future.

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
mengaims: > It appears that it is the batch size that is driving this new estimator.
> 
> Would it be better to call this `SrCnnBatchedAnomalyEstimator` and the function as `DetectAnomalySrCnnInBatchedMode` ?

The key difference of this new estimator with the old one is that it implements entire anomaly detection for timeseries, instead of streaming anomaly detection in the old one. The batch size here is an optional parameter. When the timeseries is too long, it's better to divide the timeseries into batches to do entire anomaly detection to get better performance and higher quality.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
yaeldekel, line 260: >Inference [](start = 49, length = 9)

I am a bit confused by what this mapper does. The `Inference` method looks up an array of values in a dictionary where the keys are the timestamps seen during training. That means that for new data, it will return a vector of `NaN`s for any timestamp that was not seen during training (which is likely to be almost always). Is that the intention of your algorithm?

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
antoniovs1029: TextLoader does support loading DateTime objects from text files, and in case there's ambiguity on how to load the format in the file, then users can use a CustomMappingTransformer, as shown in the tests I've recently added to DateTimeTransformerTests in this PR:

https://github.com/dotnet/machinelearning/pull/5063/files

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
harishsk: If a batch value is provided then the detection does not happen on the entire series at once, but happens in batches. But with a batch size of -1, the detection happens on the whole series. 
Even though batch size is optional with a default value, it is this parameter that controls the behavior.  So I would like to suggest a name change as above (`SrCnnBatchedAnomalyEstimator` and the function as `DetectAnomalySrCnnInBatchedMode`)

---
In reply to: [416644152](https://github.com/dotnet/machinelearning/pull/4998#discussion_r416644152) [](ancestors = 416644152)

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
mengaims: > I agree with @yaeldekel's suggestion above.
> 
> In reply to: [415894364](https://github.com/dotnet/machinelearning/pull/4998#discussion_r415894364) [](ancestors = 415894364,415369363)
Fixed.

src/Microsoft.ML.TimeSeries/SrCnnDataType.cs
mengaims: > Why do we need this? And what is it being used for?

Removed since no need for SrCnnTsPoint now.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > Can you please restructure this to initialize array of const strings just once and use references to subsets of the array for each specific case?

Updated as you suggested.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > Can we please create names for these constants of 3, 4, 7 and use those throughout the code?. E.g. AnomalyOnlyVectorLength, AnomalyAndEVVectorLength, and AnomalyAndMarginVectorLength?
> (Or something shorter if possible)

Updated as you suggested.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > Can you please add some comments on how these numbers came about? What do they mean? What would need to be done if there was a bug and if this table needs to be updated....etc?

Added.

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
mengaims: > Can we add some references to the algorithm? Any published papers or wikipedia articles?

Added.

test/Microsoft.ML.TimeSeries.Tests/TimeSeriesDirectApi.cs
mengaims: > Are the three tests below nearly identical except for the mode?
> If so, can you please refactor them using [Theory] instead of [Fact]?

Updated as you suggested.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > You are throwing an exception above if `trainingTimeStamp.Count` is less than `_minTrainingPoint`.
> 
> Either you don't need this if check here or if you keep the check here, you will need an else block and can move the exception to the else block

Updated as you suggested.

test/Microsoft.ML.TimeSeries.Tests/TimeSeriesDirectApi.cs
mengaims: > Can we add a test that demonstrates loading data from a file and computing anomalies?

Added.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > I think all of these can be private readonly fields.

Updated as you suggested.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > `OneToOneTransformerBase` class has a field called `ColumnPairs` which contains this information.

Removed and refer to ColumnPairs for column name.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > > SrCnnEntireTransformer [](start = 16, length = 22)
> 
> Please add a comment inside this method and in the `SaveModel` method detailing the binary format (for an example of what I mean, see here: https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.ImageAnalytics/ImageResizer.cs#L231-L238).

Added.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: Updated.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > > Int32 [](start = 30, length = 5)
> 
> Please use `int` instead of `Int32`.

Updated.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > > inputSchema [](start = 21, length = 11)
> 
> The `base` constructor checks this, no need to check again.

Refactored Mapper to inherit directly from MapperBase to deal with two input columns.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > Sorry, I should have said that `CheckInputColumn` is in the base class of the transformer, not the mapper.
> 
> In reply to: [415788241](https://github.com/dotnet/machinelearning/pull/4998#discussion_r415788241) [](ancestors = 415788241)

Refactored Mapper to inherit directly from MapperBase to deal with two input columns.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > > return [](start = 28, length = 6)
> 
> What does `src == null` indicate? Don't you want to assign something to `dst` so that users would know that `src` was `null`? (maybe all zeros? or all `NaN`s?)

Updated. Assign NaN now.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > > Double [](start = 24, length = 6)
> 
> nit: Please use lower case `double` (throughout the file).

Updated.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > Ok, now that I see that `Inference` looks up the array in a dictionary, I see that it's not possible. But still, you don't need to allocate a new array for values, you can write instead
> 
> ```
> _parent.AnomalyModel.Inference(src, out var values)
> ```
> 
> In reply to: [415798844](https://github.com/dotnet/machinelearning/pull/4998#discussion_r415798844) [](ancestors = 415798844)

Updated.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > > internal [](start = 8, length = 8)
> 
> This class can be private.
> 
> Also, the way I understand it, the only method in this class that is used externally is `Inference`, and that method only calls `TryGetValue` into a dictionary. I would consider changing this to a static class that has a static method `Train`, that returns the dictionary itself, and having the transformer use that dictionary directly (and be in charge of saving/loading it).

Updated as you suggested.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > > RoleMappedData [](start = 30, length = 14)
> 
> There is no need for this overload, you can train directly on the `IDataView` since you know which column you are using as input (`_parent.InputColumnName`).

Removed in the new static class.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > Can you please add more details on exactly what the TODO work is? Or delete them if they are already done.

Deleted.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > > Inference [](start = 49, length = 9)
> 
> I am a bit confused by what this mapper does. The `Inference` method looks up an array of values in a dictionary where the keys are the timestamps seen during training. That means that for new data, it will return a vector of `NaN`s for any timestamp that was not seen during training (which is likely to be almost always). Is that the intention of your algorithm?

Yes, your understanding is right.

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
mengaims: > If a batch value is provided then the detection does not happen on the entire series at once, but happens in batches. But with a batch size of -1, the detection happens on the whole series.
> Even though batch size is optional with a default value, it is this parameter that controls the behavior. So I would like to suggest a name change as above (`SrCnnBatchedAnomalyEstimator` and the function as `DetectAnomalySrCnnInBatchedMode`)
> 
> In reply to: [416644152](https://github.com/dotnet/machinelearning/pull/4998#discussion_r416644152) [](ancestors = 416644152)

I have some concerns to use "batch", since firstly, there is already a sample called DetectAnomalyBySrCnnBatchPrediction(https://github.com/dotnet/machinelearning/blob/master/docs/samples/Microsoft.ML.Samples/Dynamic/Transforms/TimeSeries/DetectAnomalyBySrCnnBatchPrediction.cs), this sample is for the old streaming estimator, but call it in a "batch" way. This sample was added following other models of TimeSeries. I'm not sure if it will confuse users if we call the new estimator "batch". And besides, in Kensho Anomaly Detector API, we call the two APIs "Last API" and "Entire API", so it would be better if we could continue use the name "Entire" to make it easier to switch among different platforms (also we are relasing SR to AML and uses the same word).

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,905 @@
...
+                _parent = parent;
+
+                SourceColumn inputCol = SourceColumn.Parse(_parent.ColumnPairs[0].inputColumnName);
+
+                if (!inputSchema.TryGetColumnIndex(inputCol.Source[0], out _timestampColumnIndex))
+                    throw Host.ExceptSchemaMismatch(nameof(inputSchema), "inputTimestamp", inputCol.Source[0]);
+                if (!(inputSchema[_timestampColumnIndex].Type is DateTimeDataViewType))
+                    throw Host.ExceptSchemaMismatch(nameof(inputSchema), "inputTimestamp", inputCol.Source[0], "DateTime", inputSchema[_timestampColumnIndex].Type.ToString());
+
+                _slotNames = new VBuffer<ReadOnlyMemory<char>>(_parent.OutputLength, _outputFields.ToList().GetRange(0, _parent.OutputLength).Select(x => x.AsMemory()).ToArray());
```
harishsk, line 286: >ToList().GetRange(0, _parent.OutputLength) [](start = 99, length = 42)

ToList().GetRange() may not be necessary. You can call Take() on the array

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
harishsk: Is there a formula for this? If so, can you please add the formula with a description of the variables in the formula?

---
In reply to: [418089567](https://github.com/dotnet/machinelearning/pull/4998#discussion_r418089567) [](ancestors = 418089567)

test/Microsoft.ML.TimeSeries.Tests/TimeSeriesDirectApi.cs
```diff
@@ -84,12 +84,30 @@ public TimeSeriesData(float value)
...
+        //    public double[] Prediction { get; set; }
+        //}
+
+        //private sealed class SrCnnAnomalyDetectionWithMargin
+        //{
+        //    [VectorType(7)]
+        //    public double[] Prediction { get; set; }
+        //}
+
         [Fact]
```
harishsk, line 111: Please delete this

test/Microsoft.ML.TimeSeries.Tests/TimeSeriesDirectApi.cs
```diff
@@ -557,12 +575,159 @@ public void AnomalyDetectionWithSrCnn(bool loadDataFromFile)
...
         }
+
+        [Theory]
+        [InlineData(SrCnnDetectMode.AnomalyOnly)]
+        [InlineData(SrCnnDetectMode.AnomalyAndExpectedValue)]
+        [InlineData(SrCnnDetectMode.AnomalyAndMargin)]
+        [InlineData(SrCnnDetectMode.AnomalyOnly, true)]
+        [InlineData(SrCnnDetectMode.AnomalyAndExpectedValue, true)]
+        [InlineData(SrCnnDetectMode.AnomalyAndMargin, true)]
+        public void AnomalyDetectionWithSrCnnEntire(SrCnnDetectMode mode, bool loadDataFromFile=false)
```
harishsk, line 594: You can replace all this with just [CombinatorialData]
See [https://github.com/AArnott/Xunit.Combinatorial](https://github.com/AArnott/Xunit.Combinatorial)

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
harishsk: Sorry, I don't see the references. Can you please point them out?

---
In reply to: [418089729](https://github.com/dotnet/machinelearning/pull/4998#discussion_r418089729) [](ancestors = 418089729)

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
harishsk: Sounds good. Thanks for the explanation.

---
In reply to: [418108643](https://github.com/dotnet/machinelearning/pull/4998#discussion_r418108643) [](ancestors = 418108643)

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
mengaims: Sorry to confuse you, the algorithm description is added in SrCnnEntireAnomalyEstimator.cs ln28 to ln 56.

---
In reply to: [418259971](https://github.com/dotnet/machinelearning/pull/4998#discussion_r418259971) [](ancestors = 418259971,418089729)

test/Microsoft.ML.TimeSeries.Tests/TimeSeriesDirectApi.cs
mengaims: deleted

---
In reply to: [418241868](https://github.com/dotnet/machinelearning/pull/4998#discussion_r418241868) [](ancestors = 418241868)

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: Fixed.

mengaims: > > ToList().GetRange(0, _parent.OutputLength) [](start = 99, length = 42)
> 
> ToList().GetRange() may not be necessary. You can call Take() on the array

Resolved.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
mengaims: > Is there a formula for this? If so, can you please add the formula with a description of the variables in the formula?
> 
> In reply to: [418089567](https://github.com/dotnet/machinelearning/pull/4998#discussion_r418089567) [](ancestors = 418089567)

Formula added.

test/Microsoft.ML.TimeSeries.Tests/TimeSeriesDirectApi.cs
mengaims: > Please delete this

Deleted.

src/Microsoft.ML.TimeSeries/SrCnnDataType.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System;
+using Microsoft.ML.Data;
+
+namespace Microsoft.ML.Transforms.TimeSeries
+{
+    /// <summary>
+    /// The detect modes of SrCnn models.
+    /// </summary>
+    public enum SrCnnDetectMode
```
yaeldekel, line 13: >SrCnnDetectMode [](start = 16, length = 15)

This can be part of one of the other files.

src/Microsoft.ML.TimeSeries/SrCnnEntireAnomalyEstimator.cs
```diff
@@ -0,0 +1,119 @@
...
+            string valueColumnName = null,
+            double threshold = 0.3,
+            int batchSize = 1024,
+            SrCnnDetectMode srCnnDetectMode = SrCnnDetectMode.AnomalyOnly,
+            double sensitivity = 99.0)
+        {
+            Contracts.CheckValue(env, nameof(env));
+            _host = env.Register(nameof(SrCnnEntireAnomalyEstimator));
+            _timestampColumnName = timestampColumnName ?? outputColumnName;
+            _valueColumnName = valueColumnName ?? outputColumnName;
```
yaeldekel, line 80: This can't work. You can pick at most one of the input columns to "automatically" have the same name as the output column, but not both. In this case, I would require all of them to be specified explicitly, to avoid confusion.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,910 @@
...
+                if (TryUnparseCore(sb))
+                    return sb.ToString();
+                return string.Empty;
+            }
+        }
+
+        internal sealed class Options : TransformInputBase
+        {
+            [Argument(ArgumentType.Required, HelpText = "The source columns defined in ManyToOneColumn type.", ShortName = "src", SortOrder = 1, Purpose = SpecialPurpose.ColumnName)]
+            public SourceColumn Source;
```
yaeldekel, line 58: >SourceColumn [](start = 19, length = 12)

I would simply have two string arguments for the two input columns.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,910 @@
...
+using Microsoft.ML.Runtime;
+using Microsoft.ML.Transforms.TimeSeries;
+
+[assembly: LoadableClass(typeof(SrCnnEntireTransformer),
+        typeof(SrCnnEntireTransformer), null, typeof(SignatureLoadModel),
+        "SrCnnEntire Transformer",
+        SrCnnEntireTransformer.LoaderSignature)]
+namespace Microsoft.ML.Transforms.TimeSeries
+{
+    public sealed class SrCnnEntireTransformer : OneToOneTransformerBase
```
yaeldekel, line 22: >OneToOneTransformerBase [](start = 49, length = 23)

Now that this transformer has two input columns and one output column, it cannot derive from `OneToOneTransformerBase`.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,910 @@
...
+            public SrCnnDetectMode SrCnnDetectMode = SrCnnDetectMode.AnomalyOnly;
+
+            [Argument(ArgumentType.AtMostOnce, HelpText = "The sensitivity of boundaries.",
+                ShortName = "sens", SortOrder = 404)]
+            public double Sensitivity = 99;
+        }
+
+        private int OutputLength { get; }
+
+        private Options _options;
```
yaeldekel, line 82: >Options [](start = 16, length = 7)

readonly.

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,910 @@
...
+
+        private Dictionary<long, double[]> _anomalyDict;
```
yaeldekel, line 84: >Dictionary [](start = 16, length = 10)

readonly.

src/Microsoft.ML.TimeSeries/ExtensionsCatalog.cs
```diff
@@ -187,5 +187,34 @@ public static class TimeSeriesCatalog
...
+        /// <param name="batchSize">Divide the input data into batches to fit srcnn model.
+        /// When set to -1, use the whole input to fit model instead of batch by batch, when set to a positive integer, use this number as batch size.
+        /// Must be -1 or a positive integer no less than 12. Default value is 1024.</param>
+        /// <param name="srCnnDetectMode">An enum type of <see cref="SrCnnDetectMode"/>.
+        /// When set to AnomalyOnly, the output vector would be a 3-element Double vector of (IsAnomaly, RawScore, Mag).
+        /// When set to AnomalyAndExpectedValue, the output vector would be a 4-element Double vector of (IsAnomaly, RawScore, Mag, ExpectedValue).
+        /// When set to AnomalyAndMargin, the output vector would be a 7-element Double vector of (IsAnomaly, AnomalyScore, Mag, ExpectedValue, BoundaryUnit, UpperBoundary, LowerBoundary).
+        /// Default value is AnomalyOnly.</param>
+        /// <param name="sensitivity">Sensitivity of boundaries, only useful when srCnnDetectMode is AnomalyAndMargin. Must be in [0,100]. Default value is 99.</param>
```
klausmh, line 208: Can you add comments that details what the effect of the parameter is?

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,910 @@
...
+                    cumSumList[i] = (cumSumList[i] - cumSumShift[i - n]) / n;
+                }
+                for (int i = 1; i < n; ++i)
+                {
+                    cumSumList[i] /= (i + 1);
+                }
+                return cumSumList;
+            }
+
+            private static double CalculateSocre(double mag, double avgMag)
```
klausmh, line 575: nit: CalculateScore

src/Microsoft.ML.TimeSeries/SrCnnEntireTransformer.cs
```diff
@@ -0,0 +1,910 @@
...
+                        batchResult.AddRange(TrainCore(trainingData.GetRange(i * batchSize - batchSize + len, batchSize), options, length));
+                    }
+                }
+
+                anomalyDict = trainingTimestamp.Zip(batchResult, (k, v) => new { k, v }).ToDictionary(x => x.k, x => x.v);
+            }
+
+            private static List<double[]> TrainCore(List<double> values, Options option, int length)
+            {
+                List<double[]> results = values.Select(x => new double[length]).ToList();
```
klausmh, line 445: There is a lot of LINQ and ToList() here. Is that really necessary? I am worried about the potential number of allocatons

src/Microsoft.ML.TimeSeries/SrCnnEntireAnomalyEstimator.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// </summary>
+    /// <remarks>
+    /// <format type="text/markdown"><![CDATA[
+    /// To create this estimator, use
+    /// [DetectEntireAnomalyBySrCnn](xref:Microsoft.ML.TimeSeriesCatalog.DetectEntireAnomalyBySrCnn(Microsoft.ML.TransformsCatalog,System.String,System.String,System.String,System.Double,System.Int32,SrCnnDetectMode,System.Double))
+    /// ###  Estimator Characteristics
+    /// |  |  |
+    /// | -- | -- |
+    /// | Does this estimator need to look at the data to train its parameters? | Yes |
+    /// | Input timestamp column data type | <xref:System.DateTime> |
```
klausmh, line 23: For other time series functionality in ML.NET, there is no DateTime parameter (only a sequence of doubles)? I think there are assumptions about data points being equidistant still right?

PR #5013, opened 2020-04-10
MurmurHash Onnx Export

Lynx1820:
The following data input types are currently not supported by onnxruntime's murmurHash operator: float, double, ulong, long and ordered hashing(vectors). Once added, ml.net will be able to support them as well.


src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -834,6 +1089,77 @@ private void AddMetaKeyValues(int i, DataViewSchema.Annotations.Builder builder)
...
+                string opType;
+                OnnxNode murmurNode;
+
+                opType = "MurmurHash3";
+                string murmurOutput = (_srcTypes[iinfo].RawType == typeof(ReadOnlyMemory<char>)) ? dstVariable: ctx.AddIntermediateVariable(_dstTypes[iinfo], "MurmurOutput");
+                var srcType = _srcTypes[iinfo].GetItemType().RawType;
+
+                // Numeric input types are limited to those supported by the Onnxruntime MurmurHash operator, which currently only supports
+                // uints and ints. Thus, ulongs, longs, doubles, floats, and booleans are not supported.
+                if (srcType == typeof(ushort) || srcType == typeof(short) || srcType == typeof(sbyte) || srcType == typeof(byte))
```
KsenijaS, line 1104: there is no support for Key types (if (_types[iinfo].RawType == typeof(KeyDataViewType))), key types need to have +1 at the end of the hash value.

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -834,6 +1089,77 @@ private void AddMetaKeyValues(int i, DataViewSchema.Annotations.Builder builder)
...
+                else if (srcType == typeof(uint) || srcType == typeof(int) || srcType == typeof(ReadOnlyMemory<char>))
+                {
+                    murmurNode = ctx.CreateNode(opType, srcVariable, murmurOutput, ctx.GetNodeName(opType), "com.microsoft");
+                }
+                else
+                {
+                    return false;
+                }
+
+                murmurNode.AddAttribute("positive", 1);
```
KsenijaS, line 1120: also you have to note that there is no support for useOrderedHashing yet.

src/Microsoft.ML.Core/Utilities/Hashing.cs
```diff
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Buffers.Binary;
```
yaeldekel, line 6: >System.Buffers.Binary; [](start = 6, length = 22)

Is this used?

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -122,8 +123,9 @@ private static VersionInfo GetVersionInfo()
             return new VersionInfo(
                 modelSignature: "HASHTRNS",
                 // verWrittenCur: 0x00010001, // Initial
-                verWrittenCur: 0x00010002, // Invert hash key values, hash fix
-                verReadableCur: 0x00010002,
+                //verWrittenCur: 0x00010002, // Invert hash key values, hash fix
+                verWrittenCur: 0x00010003,
```
yaeldekel, line 127: >0x00010003 [](start = 31, length = 10)

Please add a comment explaining the version bump (similar to the line above).

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -245,9 +247,15 @@ private Delegate GetGetterCore(DataViewRow input, int iinfo, out Action disposer
             disposer = null;
             input.Schema.TryGetColumnIndex(_columns[iinfo].InputColumnName, out int srcCol);
             var srcType = input.Schema[srcCol].Type;
-            if (!(srcType is VectorDataViewType vectorType))
-                return ComposeGetterOne(input, iinfo, srcCol, srcType);
-            return ComposeGetterVec(input, iinfo, srcCol, vectorType);
+            if (GetVersionInfo().VerWrittenCur == 0x00010002)
```
yaeldekel, line 250: >GetVersionInfo() [](start = 16, length = 16)

This calls the method `GetVersionInfo`, which always returns the same thing (currently 0x00010002, after your change 0x00010003).
The correct way to do what you are trying to achieve, is to add a `_version` field to the transformer. This field would be initialized according to `ctx.Header.ModelVerWritten` in the constructor from file, and to 0x00010003 in the other constructors.

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -796,16 +1065,22 @@ public ColInfo(string outputColumnName, string inputColumnName, DataViewType typ
...
-                _types = new DataViewType[_parent._columns.Length];
-                for (int i = 0; i < _types.Length; i++)
-                    _types[i] = _parent.GetOutputType(inputSchema, _parent._columns[i]);
+                _srcTypes = new DataViewType[_parent._columns.Length];
+                _dstTypes = new DataViewType[_parent._columns.Length];
+                for (int i = 0; i < _dstTypes.Length; i++)
+                {
+                    _dstTypes[i] = _parent.GetOutputType(inputSchema, _parent._columns[i]);
+                    inputSchema.TryGetColumnIndex(_parent.ColumnPairs[i].inputColumnName, out var colIndex);
+                    _srcTypes[i] = inputSchema[colIndex].Type;
```
yaeldekel, line 1082: Instead of these two lines, you can instead write
```
_srcTypes[i] = inputSchema[_parent.ColumnPairs[i].inputColumnName].Type;
```

src/Microsoft.ML.Data/Transforms/Hashing.cs
yaeldekel: For the unsigned integer types, you need to look at `srcTypes[iinfo].GetItemType()` itself, to see if it is a key type or not.

---
In reply to: [407106422](https://github.com/dotnet/machinelearning/pull/5013#discussion_r407106422) [](ancestors = 407106422)

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -1764,32 +1855,33 @@ private void CompareResults(string leftColumnName, string rightColumnName, IData
         {
             var leftColumn = left.Schema[leftColumnName];
             var rightColumn = right.Schema[rightColumnName];
-            var leftType = leftColumn.Type.GetItemType();
-            var rightType = rightColumn.Type.GetItemType();
+            var leftType = leftColumn.Type.GetItemType().RawType;
+            var rightType = rightColumn.Type.GetItemType().RawType;
+            Assert.Equal(leftType, rightType);
```
yaeldekel, line 1860: Why is this changing? I am assuming that it is because ONNX models don't have the concept of key types, but it would make all the other tests less strict as well. Perhaps add a boolean argument to this method, such as `compareItemType`?

Is this the first transformer that outputs key types that can be exported to ONNX? If not, then how is this handled in other transformers? 
If yes, then we should decide how to handle exporting key types, because of the `+1` difference in the outputs.

test/Microsoft.ML.Tests/RangeFilterTests.cs
```diff
@@ -29,7 +29,7 @@ public void RangeFilterTest()
             Assert.Equal(2L, cnt);
 
             data = ML.Transforms.Conversion.Hash("Key", "Strings", numberOfBits: 20).Fit(data).Transform(data);
-            var data2 = ML.Data.FilterRowsByKeyColumnFraction(data, "Key", upperBound: 0.5);
+            var data2 = ML.Data.FilterRowsByKeyColumnFraction(data, "Key", upperBound: 0.15);
```
yaeldekel, line 32: >15 [](start = 89, length = 2)

Why is this changing?

yaeldekel:
>         public void TestHashTransformFloat()

Please add a backwards compatibility test for version 0x00010002 models.

---
Refers to: test/Microsoft.ML.TestFramework/DataPipe/TestDataPipe.cs:1054 in 49bdad3. [](commit_id = 49bdad310c99d521962b39fd43df91c47aa33612, deletion_comment = False)

test/Microsoft.ML.Tests/RangeFilterTests.cs
Lynx1820: It's changing because the hash values are different, so the fractional values will also change.

src/Microsoft.ML.Data/Transforms/Hashing.cs
Lynx1820: Added versioning + test

src/Microsoft.ML.Data/Transforms/Hashing.cs
Lynx1820: I did support OrderHashing - take a look at  line 1143

src/Microsoft.ML.Core/Utilities/Hashing.cs
Lynx1820: Nope. Got rid of it.

src/Microsoft.ML.Core/Utilities/Hashing.cs
```diff
@@ -174,6 +175,87 @@ public static uint MurmurHash(uint hash, ReadOnlySpan<char> span, bool toUpper =
             return hash;
         }
 
+        public static uint MurmurHashV2(uint hash, ReadOnlySpan<char> span, bool toUpper = false)
+        {
```
harishsk, line 179: There is a lot of context missing here. Can you please ask Ksenija for the docs she wrote and add a detailed explanation here as to why MurmurHasV2 was necessary and how it is different from the previous version?
Please add one section for the overall hashing differences and one specifically for strings.

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -1146,6 +1146,97 @@ public void PcaOnnxConversionTest()
...
+                CompareSelectedColumns<uint>("Education", "Education", transformedData, onnxResult);
+            }
+            Done();
+        }
+
+        [Theory]
+        [CombinatorialData]
+        public void MurmurHashUIntTest(
+            [CombinatorialValues(DataKind.SByte, DataKind.Int16, DataKind.Int32, DataKind.Byte, DataKind.UInt16, DataKind.UInt32)] DataKind type,
+            bool useOrderedHashing,
```
harishsk, line 1199: I am guessing float and double are not present here because ORT doesn't support it yet? If so, can you please add a note about it?

src/Microsoft.ML.Data/Transforms/Hashing.cs
Lynx1820: Added support for keytype back.

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -167,6 +170,7 @@ private DataViewType GetOutputType(DataViewSchema inputSchema, HashingEstimator.
               base(Contracts.CheckRef(env, nameof(env)).Register(RegistrationName), GetColumnPairs(columns))
         {
             _columns = columns.ToArray();
+            _version = 0x00010003;
```
yaeldekel, line 173: >0x00010003 [](start = 23, length = 10)

I was going to suggest having a static variable containing the current version, or calling `GetVersionInfo()` here, but now that I think about it, this field can actually be of bool type, `_isVersion2` or something similar. That way it is `false` by default, and only needs to be initialized in the ctor with `ModelLoadContext`.

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -1146,6 +1146,100 @@ public void PcaOnnxConversionTest()
             Done();
         }
 
+        private class HashData
+        {
+            public uint Data { get; set; }
+        }
+
+        [Fact]
+        public void MurmurHashKeyTest()
```
yaeldekel, line 1155: >MurmurHashKeyTest [](start = 20, length = 17)

Shouldn't we have both `MurmurHashStringTest` and `MurmurHashKeyTest`?

src/Microsoft.ML.Data/Transforms/Hashing.cs
yaeldekel: This takes care of the scalar column case, but not of vector columns.
(see how ordered hashing is computed in lines 1008-1020 in this file). I'm not sure how this can be handled in ONNX, does the `MurmurHash3` operator handle vectors at all (ordered or unordered)?

---
In reply to: [413203701](https://github.com/dotnet/machinelearning/pull/5013#discussion_r413203701) [](ancestors = 413203701)

test/Microsoft.ML.Tests/OnnxConversionTest.cs
yaeldekel, line 1155: >MurmurHashKeyTest [](start = 20, length = 17)

If ONNX can handle vector columns, please also add a test for that.

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -834,6 +1112,96 @@ private void AddMetaKeyValues(int i, DataViewSchema.Annotations.Builder builder)
...
+                opType = "BitShift";
+                string bitShiftOutput = ctx.AddIntermediateVariable(_dstTypes[iinfo], "BitShiftOutput");
+                var shiftValue = ctx.AddInitializer((ulong)(32 - _parent._columns[iinfo].NumberOfBits), false);
+                var shiftNode = ctx.CreateNode(opType, new[] { murmurOutput, shiftValue }, new[] { bitShiftOutput }, ctx.GetNodeName(opType), "");
+                shiftNode.AddAttribute("direction", "LEFT");
+
+                opType = "BitShift";
+                var outputVariable = dstVariable;
+                if(_srcTypes[iinfo].GetItemType() is KeyDataViewType)
+                    outputVariable = ctx.AddIntermediateVariable(_dstTypes[iinfo], "BitShiftOutput2");
```
yaeldekel, line 1159: >outputVariable [](start = 20, length = 14)

Why do we need this intermediate variable for key types?

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -834,6 +1112,96 @@ private void AddMetaKeyValues(int i, DataViewSchema.Annotations.Builder builder)
...
+
+                if (_srcTypes[iinfo].GetItemType() is KeyDataViewType)
+                {
+                    opType = "Cast";
+                    string castOutput = ctx.AddIntermediateVariable(NumberDataViewType.Int64, "CastOutput2");
+                    var castNode = ctx.CreateNode(opType, outputVariable, castOutput, ctx.GetNodeName(opType), "");
+                    var t = NumberDataViewType.Int64.RawType;
+                    castNode.AddAttribute("to", t);
+
+                    opType = "Add";
```
yaeldekel, line 1171: >Add [](start = 30, length = 3)

Does ONNX not know how to apply `Add` to unsigned integer types?

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -834,6 +1112,96 @@ private void AddMetaKeyValues(int i, DataViewSchema.Annotations.Builder builder)
...
+                    string addOutput = ctx.AddIntermediateVariable(NumberDataViewType.Int64, "AddOutput");
+                    string one = ctx.AddInitializer(1);
+                    ctx.CreateNode(opType, new[] { castOutput, one }, new[] { addOutput }, ctx.GetNodeName(opType), "");
+
+                    opType = "Cast";
+                    var castNodeFinal = ctx.CreateNode(opType, addOutput, dstVariable, ctx.GetNodeName(opType), "");
+                    var tFinal = NumberDataViewType.UInt32.RawType;
```
yaeldekel, line 1178: >tFinal [](start = 24, length = 6)

If I understand correctly, these three nodes are supposed to add 1 to the output, and then convert back to the original type. But key types are not always uint, they can also be other unsigned integer types.

src/Microsoft.ML.Data/Transforms/Hashing.cs
Lynx1820: Yeah, it's not currently supported. Planning to request support and remove additional operators once that is done.

src/Microsoft.ML.Data/Transforms/Hashing.cs
Lynx1820: +1 is only for key types, so the output changes

test/Microsoft.ML.Tests/OnnxConversionTest.cs
Lynx1820: I added the String test to MurmurHashTest to make more concise. Key needs its own test, since ValueToKey needs to be added to the pipeline.

src/Microsoft.ML.Data/Transforms/Hashing.cs
Lynx1820: updated!

src/Microsoft.ML.Data/Transforms/Hashing.cs
Lynx1820: Agreed. I'll make a note about the lack of support for vectors.

test/Microsoft.ML.Tests/OnnxConversionTest.cs
Lynx1820: Yes, I added this because ONNX doesn't have a key concept,so I thought it would make more sense to compare raw types. I actually reverted this change, as now I think this could be a standalone PR.

test/Microsoft.ML.Tests/OnnxConversionTest.cs
Lynx1820: Added!

test/Microsoft.ML.Tests/OnnxConversionTest.cs
Lynx1820: Added.

src/Microsoft.ML.Core/Utilities/Hashing.cs
Lynx1820: Added a brief description. @KsenijaS can you take a look and add anything you think is important to mention?

Lynx1820: Added a brief description. @KsenijaS can you take a look and add anything you think is important to mention?

src/Microsoft.ML.Core/Utilities/Hashing.cs
```diff
@@ -174,6 +174,94 @@ public static uint MurmurHash(uint hash, ReadOnlySpan<char> span, bool toUpper =
             return hash;
         }
 
+        // MurmurHashV2 is an implementation of MurmurHash3_x86_32 (https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.cpp#L94) used by Onnxruntime's
+        // MurmurHash operator, implemented to have matching hashing results between ORT and ML.NET.
+        // One key difference between the two implementations is that strings use a different hashing algorithm in the previous version, but
+        // every type uses the same implementation on V2.
+        // The V1 String Hashing Algorithm had the following properities:
+        //     - Case Conversion: used inside the hashing algorithm in ML.Net.
+        //     - Mock UTF8 encoding: strings in C# are UTF16 and need to be converted to UTF8 before hashing
```
harishsk, line 183: Please elaborate on what Mock UTF8 encoding is. If I recall correctly, we are omitting certain code pages. 
@KsenijaS Didn't you have an implementation that supported the full UTF8 encoding? Can we use that?

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -834,6 +1110,96 @@ private void AddMetaKeyValues(int i, DataViewSchema.Annotations.Builder builder)
...
+                string opType;
+                OnnxNode murmurNode;
+
+                opType = "MurmurHash3";
+                string murmurOutput = ctx.AddIntermediateVariable(_dstTypes[iinfo], "MurmurOutput");
+                var srcType = _srcTypes[iinfo].GetItemType().RawType;
+
+                // Numeric input types are limited to those supported by the Onnxruntime MurmurHash operator, which currently only supports
+                // uints and ints. Thus, ulongs, longs, doubles, floats, and booleans are not supported.
+                if (srcType == typeof(ushort) || srcType == typeof(short) || srcType == typeof(sbyte) ||
```
yaeldekel, line 1125: >short [](start = 67, length = 5)

Would it make sense to cast the signed types to int and the unsigned types to uint?

harishsk: I don't think it makes a difference. MurmurHash in ORT converts everything back to a uint before computing the hash.

src/Microsoft.ML.Data/Transforms/Hashing.cs
yaeldekel: So after looking at it again, my previous comment was inaccurate. The `HashingTransformer` **always** outputs a key type, so these additional nodes that you are adding (Cast=>Add 1=>Cast back) actually have to be added in **all** cases, not only when the input is a key type. On the other hand, the raw type of the output of `HashingTransformer` is always `uint` (see the [`GetOutputType`](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Transforms/Hashing.cs#L149) ), so the previous way you did the Cast operator was correct.

The difference between a key type input and a non-key unsigned integer type input, is that for keys a value of 0 is mapped to 0 (0 is the `NA` value of key types, and for regular unsigned integers it is hashed just like any other value. For example, the uint hashing function is

```
            public uint HashCore(uint seed, uint mask, in uint value)
                => (Hashing.MixHash(Hashing.MurmurRound(seed, value)) & mask) + 1;
```

and the uint key type hashing function is 

```
            public uint HashCore(uint seed, uint mask, in uint value)
                => value == 0 ? 0 : (Hashing.MixHash(Hashing.MurmurRound(seed, value)) & mask) + 1;
```

Mapping 0 to 0 is currently not handled at all in the ONNX export, not sure if it can be handled... we might need to return `false` for key type inputs if this cannot be handled, similarly to what you do for floats and doubles.


---
In reply to: [414083637](https://github.com/dotnet/machinelearning/pull/5013#discussion_r414083637) [](ancestors = 414083637)

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -472,6 +580,13 @@ public uint HashCore(uint seed, uint mask, in float value)
                 => float.IsNaN(value) ? 0 : (Hashing.MixHash(Hashing.MurmurRound(seed, FloatUtils.GetBits(value == 0 ? 0 : value))) & mask) + 1;
         }
 
+        private readonly struct HashFloatV2 : IHasher<float>
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public uint HashCore(uint seed, uint mask, in float value)
```
yaeldekel, line 586: >HashCore [](start = 24, length = 8)

The `1` still has to be added for all the new `V2` implementations. It is added because the output of `HashingTransformer` is a key type, and 0 is the `NA` value of key types.

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -490,34 +605,79 @@ public uint HashCore(uint seed, uint mask, in double value)
...
+            {
+                if (double.IsNaN(value))
+                    return 0;
+
+                ulong v = FloatUtils.GetBits(value == 0 ? 0 : value);
+                var hash = Hashing.MurmurRound(seed, Utils.GetLo(v));
+                var hi = Utils.GetHi(v);
+                if (hi != 0)
+                    hash = Hashing.MurmurRound(hash, hi);
+                return (Hashing.MixHashV2(hash, 4) & mask);
```
yaeldekel, line 621: >4 [](start = 48, length = 1)

What does this number signify? Is it the number of bytes of the input?
I see that it is always called with the value 4, should we consider making the public signature of this method just `MixHashV2(hash)`, and then have it call a private method that takes two parameters?

test/Microsoft.ML.Tests/OnnxConversionTest.cs
```diff
@@ -1797,6 +1906,7 @@ private void CompareResults(string leftColumnName, string rightColumnName, IData
             else if (leftType == TextDataViewType.Instance)
                 CompareSelectedColumns<ReadOnlyMemory<char>>(leftColumnName, rightColumnName, left, right);
 
+
         }
```
yaeldekel, line 1910: nit: remove empty lines.

src/Microsoft.ML.Data/Transforms/Hashing.cs
yaeldekel: Replace

```
                    inputSchema.TryGetColumnIndex(_parent.ColumnPairs[i].inputColumnName, out var colIndex);
                    _srcTypes[i] = inputSchema[colIndex].Type;
```

with the above.


---
In reply to: [410913977](https://github.com/dotnet/machinelearning/pull/5013#discussion_r410913977) [](ancestors = 410913977)

test/Microsoft.ML.Tests/RangeFilterTests.cs
yaeldekel: I see. Another option would be to keep the `0.5` fraction, and change the number in the `Assert` to match the new value of `cnt`. I'm fine with either :).

---
In reply to: [413200283](https://github.com/dotnet/machinelearning/pull/5013#discussion_r413200283) [](ancestors = 413200283)

src/Microsoft.ML.Data/Transforms/Hashing.cs
yaeldekel: Are vectors not supported at all? Or did you mean just for the ordered case?
If vectors are supported, just not ordered, then the change you made to return `false` in the ordered case (line 1144), should be made only if the input is a vector - for scalars it can still do what you did before:

```
seed = Hashing.MurmurRound(seed, 0);
```
If you meant that vectors are not supported at all, then that would be the first thing you need to check, and return false in case of vector inputs.


---
In reply to: [414090190](https://github.com/dotnet/machinelearning/pull/5013#discussion_r414090190) [](ancestors = 414090190)

src/Microsoft.ML.Data/Transforms/Hashing.cs
yaeldekel: See my other comment regarding this - the output of `HashingTransformer` is always a key type.

---
In reply to: [413957695](https://github.com/dotnet/machinelearning/pull/5013#discussion_r413957695) [](ancestors = 413957695)

test/Microsoft.ML.Tests/OnnxConversionTest.cs
yaeldekel: Is it possible to have a scalar and a vector test for each type? Also, if you bring back support for ordered hashing in the non-vector case, to test both ordered and non ordered for scalars?

---
In reply to: [414106161](https://github.com/dotnet/machinelearning/pull/5013#discussion_r414106161) [](ancestors = 414106161)

src/Microsoft.ML.Data/Transforms/Hashing.cs
Lynx1820: gotcha, but what if the hash value is uint.max? Wouldn't that end up mapping to 0?

src/Microsoft.ML.Data/Transforms/Hashing.cs
harishsk: MixHashV2 is called with a different length when called from MurmurHashV2 for strings. So I think it should remain the way it is. 

---
In reply to: [415252298](https://github.com/dotnet/machinelearning/pull/5013#discussion_r415252298) [](ancestors = 415252298)

test/Microsoft.ML.Tests/OnnxConversionTest.cs
Lynx1820: done!

src/Microsoft.ML.Data/Transforms/Hashing.cs
Lynx1820: done!

src/Microsoft.ML.Data/Transforms/Hashing.cs
Lynx1820: yeah i meant only for vectors, 

---
In reply to: [415255169](https://github.com/dotnet/machinelearning/pull/5013#discussion_r415255169) [](ancestors = 415255169,414090190)

src/Microsoft.ML.Data/Transforms/Hashing.cs
yaeldekel: This is why we only allow hashing into at most 31 bits, to prevent this situation.

---
In reply to: [416030419](https://github.com/dotnet/machinelearning/pull/5013#discussion_r416030419) [](ancestors = 416030419)

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -834,6 +1109,99 @@ private void AddMetaKeyValues(int i, DataViewSchema.Annotations.Builder builder)
...
+                else
+                {
+                    return false;
+                }
+
+                murmurNode.AddAttribute("positive", 1);
+                var seed = _parent._columns[iinfo].Seed;
+                if (_parent._columns[iinfo].UseOrderedHashing)
+                {
+                    if (srcShape[1] > 1) return false;
```
yaeldekel, line 1148: >1 [](start = 33, length = 1)

Doesn't this need to be `srcShape[0]`? 

I think it would be better to look at `_srcTypes[iinfo]` to check whether the input is a vector or not.

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -834,6 +1109,99 @@ private void AddMetaKeyValues(int i, DataViewSchema.Annotations.Builder builder)
             }
 
             protected override Delegate MakeGetter(DataViewRow input, int iinfo, Func<int, bool> activeOutput, out Action disposer) => _parent.GetGetterCore(input, iinfo, out disposer);
+
+            private bool SaveAsOnnxCore(OnnxContext ctx, int iinfo, string srcVariable, string dstVariable)
```
yaeldekel, line 1113: >SaveAsOnnxCore [](start = 25, length = 14)

The case of key type inputs is still not handled here: if the input is a key type, then 0 is mapped to 0. If we would like to support this case then it would need to involve an "If" node in the ONNX graph. Otherwise, this method should check whether the input type is key and return false in this case.

src/Microsoft.ML.Data/Transforms/Hashing.cs
yaeldekel: The difference between `MixHash` and `MixHashV2` is only this line:
```
hash ^= (uint)len;
```
and it is called with the constant value `4` everywhere except for the `MurmurHashV2` for strings. I understand the purpose of somehow mixing in the length of the string when hashing string, but what value does it add to always mix in `4`? Is this how ONNX runtime hashes uints? 

---
In reply to: [416079308](https://github.com/dotnet/machinelearning/pull/5013#discussion_r416079308) [](ancestors = 416079308,415252298)

src/Microsoft.ML.Data/Transforms/Hashing.cs
Lynx1820: srcShape[0] is the dimension reserved for batches, so it's always -1.

src/Microsoft.ML.Data/Transforms/Hashing.cs
harishsk: Yes, this is how ORT hashes uints. 
Please [see here](https://github.com/microsoft/onnxruntime/blob/78fde2c4cbeaaaf46fa97e1ffaea0fa01cbbcb33/onnxruntime/contrib_ops/cpu/murmur_hash3.cc#L155).

This is part of the reason why we had to redo our hashing so that we can be aligned with ORT.

src/Microsoft.ML.Data/Transforms/Hashing.cs
Lynx1820: I'll make another PR for key datatypes, as I think it'll be more involved.

src/Microsoft.ML.Data/Transforms/Hashing.cs
Lynx1820: Changed it tho!

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -834,6 +1109,101 @@ private void AddMetaKeyValues(int i, DataViewSchema.Annotations.Builder builder)
...
+                var shiftValue = ctx.AddInitializer((ulong)(32 - _parent._columns[iinfo].NumberOfBits), false);
+                var shiftNode = ctx.CreateNode(opType, new[] { murmurOutput, shiftValue }, new[] { bitShiftOutput }, ctx.GetNodeName(opType), "");
+                shiftNode.AddAttribute("direction", "LEFT");
+
+                opType = "BitShift";
+                string bitShiftOutput2 = ctx.AddIntermediateVariable(_dstTypes[iinfo], "BitShiftOutput2");
+                shiftNode = ctx.CreateNode(opType, new[] { bitShiftOutput, shiftValue }, new[] { bitShiftOutput2 }, ctx.GetNodeName(opType), "");
+                shiftNode.AddAttribute("direction", "RIGHT");
+
+                var vectorShape = new VectorDataViewType(NumberDataViewType.Int64, (int)srcShape[1]);
```
yaeldekel, line 1167: >srcShape [](start = 88, length = 8)

This can change to `_srcTypes[iinfo].VectorSize` as well (or `ValueCount`? Or is it `GetValueCount()`? I don't remember the exact syntax). Then you don't need `srcShape` at all.

src/Microsoft.ML.Data/Transforms/Hashing.cs
yaeldekel: Thanks for providing the link to the ORT implementation. `MurmurHash3_x86_32` is called in two places: when the input is a string (in which case, the length of the string is passed as `len`), and when the input is a number (in which case the number of bytes is passed as `len`). In the numeric case, ORT has a check that verifies that the number has 4 bytes, but the value passed to `MurmurHash3_x83_32` is still the element size, and not the constant `4`.

If we want to truly be consistent with ORT hashing (and actually, with the `MurmurHash3_x86_32` algorithm), I think we should also use `sizeof` instead of the constant `4` when calling `MixHash`. We should definitely do that for `long`, `ulong` and `double` (which cannot be exported to ONNX anyway). However, if we do that for types with less than 4 bytes, the values computed by the exported model as it is written in this PR will not be consistent with the ML.NET values (because of the additional "Cast" operator added to convert the original input type to `uint`). This means that we would only be able to export to ONNX models with inputs that are either ints, uints or strings. To me that is a reasonable price to pay, since I don't think it is common to have pipelines where people are trying to hash bytes/shorts. What do you think @harishsk, @Lynx1820 ?



---
In reply to: [416808389](https://github.com/dotnet/machinelearning/pull/5013#discussion_r416808389) [](ancestors = 416808389)

src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -834,6 +1109,101 @@ private void AddMetaKeyValues(int i, DataViewSchema.Annotations.Builder builder)
...
+                string opType;
+                string castOutput;
+                OnnxNode castNode;
+                OnnxNode murmurNode;
+
+                opType = "MurmurHash3";
+                string murmurOutput = ctx.AddIntermediateVariable(_dstTypes[iinfo], "MurmurOutput");
+                var srcShape = ctx.RetrieveShapeOrNull(srcVariable);
+                var srcType = _srcTypes[iinfo].GetItemType().RawType;
+                if (_srcTypes[iinfo] is KeyDataViewType) return false;
```
yaeldekel, line 1124: >if (_srcTypes[iinfo] is KeyDataViewType) return false; [](start = 16, length = 54)

nit: Please put the `return false` in a separate line.

src/Microsoft.ML.Data/Transforms/Hashing.cs
harishsk: The current implementation is consistent with ORT.  @Lynx1820 has tests in OnnxConversionTest.cs for all data types (except long, ulong, double and float). 

You are correct that when `MurmurHash3_x86_32` is called the `len` parameter is the size in bytes. This in turn gets translated to a `nblocks` value of `0`. This causes the code in lines 122 to 131 to be executed only once. But on line 120, the data `data` gets cast to a pointer of blocks of type `uint32_t` and it hashes each block as a `uint32_t` value.

The MurmurHash3 algorithm is defined in terms of hashing everything as `uint32_t` blocks. So to be consistent, it is the right thing to do always hash all types of size <= 4 bytes as `uint32_t` types.  We are already doing that and the tests are passing with consistent hashing values between ML.NET and ORT.

I do agree with you about the constant 4 and that it should be expressed in more generic terms. I think it should be changed to `sizeof(uint)` to make it explicit that everything is being hashed as a `uint`. 

@Lynx1820 can you please make that change?

@yaeldekel would that fix be sufficient to resolve this comment?




---
In reply to: [417988309](https://github.com/dotnet/machinelearning/pull/5013#discussion_r417988309) [](ancestors = 417988309,416808389)

yaeldekel: I think for now this is sufficient. There is still a question about the 8 byte types in case ORT decides to support that in the future - but since we don't know how they would choose to do it, the approach we take is fine.

---
In reply to: [418289134](https://github.com/dotnet/machinelearning/pull/5013#discussion_r418289134) [](ancestors = 418289134,417988309,416808389)

src/Microsoft.ML.Core/Utilities/Hashing.cs
KsenijaS: Yes, but also partial UTF8 support is sufficient. Mock UTF8 doesn't cover emojis and some special math characters

PR #5021, opened 2020-04-13
LoadColumnName support in TextLoader and auto map class members 

### LINKED ISSUE #4480: LoadColumn should not be required when loading from a file with headers
### System information

- **OS version/distro**: Windows 10 (10.0.18362)
- **.NET Version (eg., dotnet --info)**:  3.0.100

### Issue

- **What did you do?**
Created a type anottated with `LoadColumnName`
Created a `TextLoader` from this type and `useHeaders` set to `true` and called the `Load` method

- **What happened?**
System.InvalidOperationException: 'Property 'IESCode' is missing the LoadColumnAttribute attribute'

- **What did you expect?**
The Load method should map the properties from the file headers when available, not forcing the use of column index.

### Source code / logs

```fsharp
System.InvalidOperationException: 'Property 'IESCode' is missing the LoadColumnAttribute attribute'
```

```fsharp
type EnadeMicrodata =
    {
        [<LoadColumnName("CO_IES")>]
        IESCode: int

        [<LoadColumnName("CO_CATEGAD")>]
        AdministrativeCategory: int
    }

let mlContext = MLContext ()
let _textLoader = mlContext.Data.CreateTextLoader<EnadeMicrodata>(hasHeader = true).Load("MICRODADOS_ENADE_2017.txt")
```

kevencarneiro:
This pull request resolves #4480.

These features were add in this pull request:
* Support of `LoadColumnName` attribute in `TextLoader` class (when loaded file has headers).
* Possibility to ignore a member when using `CreateTextLoader` using the `LoadColumnIgnoreAttribute`
* Feature to map the `TInput` properties/fields name to column of TextLoader (when loaded file has headers)

### DIFF
--- a/Microsoft.ML.sln
+++ b/Microsoft.ML.sln
@@ -94,7 +94,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ML.CpuMath.Perfor
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ML.CpuMath.UnitTests", "test\Microsoft.ML.CpuMath.UnitTests\Microsoft.ML.CpuMath.UnitTests.csproj", "{E97D8B5A-3035-4D41-9B0D-77FF8FB8D132}"
 EndProject
-Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Microsoft.ML.FSharp.Tests", "test\Microsoft.ML.FSharp.Tests\Microsoft.ML.FSharp.Tests.fsproj", "{802233D6-8CC0-46AD-9F23-FEE1E9AED9B3}"
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Microsoft.ML.FSharp.Tests", "test\Microsoft.ML.FSharp.Tests\Microsoft.ML.FSharp.Tests.fsproj", "{802233D6-8CC0-46AD-9F23-FEE1E9AED9B3}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ML.ImageAnalytics", "src\Microsoft.ML.ImageAnalytics\Microsoft.ML.ImageAnalytics.csproj", "{00E38F77-1E61-4CDF-8F97-1417D4E85053}"
 EndProject
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -1513,6 +1513,16 @@ namespace Microsoft.ML.Data
         /// </example>
         public IDataView Load(IMultiStreamSource source) => new BoundLoader(this, source);
 
+        // 1.4.0 BACKCOMPAT OVERLOAD -- DO NOT TOUCH
+        internal static TextLoader CreateTextLoader<TInput>(IHostEnvironment host,
+           bool hasHeader,
+           char separator,
+           bool allowQuoting,
+           bool supportSparse,
+           bool trimWhitespace,
+           IMultiStreamSource dataSample = null)
+           => CreateTextLoader<TInput>(host, hasHeader, separator, allowQuoting, supportSparse, trimWhitespace, Defaults.AutoMapLoadColumns, dataSample);
+
         internal static TextLoader CreateTextLoader<TInput>(IHostEnvironment host,
            bool hasHeader = Defaults.HasHeader,
            char separator = Defaults.Separator,
--- a/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
+++ b/src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
@@ -277,6 +277,19 @@ namespace Microsoft.ML.Data
                 Contracts.CheckParam(index >= 0, nameof(index), "Must be non-negative");
                 Min = index;
                 Max = index;
+                Name = null;
+            }
+
+            /// <summary>
+            /// A range representing a single value. Will result in a scalar column.
+            /// </summary>
+            /// <param name="name">The name of the field of the table to read.</param>
+            public Range(string name)
+            {
+                Contracts.CheckValue(name, nameof(name));
+                Min = -1;
+                Max = -1;
+                Name = name;
             }
 
             /// <summary>
@@ -293,6 +306,7 @@ namespace Microsoft.ML.Data
 
                 Min = min;
                 Max = max;
+                Name = null;
                 // Note that without the following being set, in the case where there is a single range
                 // where Min == Max, the result will not be a vector valued but a scalar column.
                 ForceVector = true;
@@ -313,6 +327,15 @@ namespace Microsoft.ML.Data
             /// </summary>
             [Argument(ArgumentType.AtMostOnce, HelpText = "Last index in the range")]
             public int? Max;
+
+            /// <summary>
+            /// The name of the input column.
+            /// </summary>
+            /// <remarks>
+            /// This value, if non-<c>null</c>, overrides <see cref="Min" /> and <see cref="Max" />.
+            /// </remarks>
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Name of the column")]
+            public string Name;
 
             /// <summary>
             /// Whether this range extends to the end of the line, but should be a fixed number of items.
@@ -496,6 +519,13 @@ namespace Microsoft.ML.Data
             public bool HasHeader;
 
             /// <summary>
+            /// Whether the members name of the class passed as type argument should be mapped to <see cref="LoadColumnNameAttribute"/>.
+            /// When <see langword="true"/> requires the <see cref="HasHeader"/> option to be <see langword="true"/>.
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Use class properties as LoadColumnNameAttribute to read from text file", ShortName = "automap")]
+            public bool AutoMapLoadColumns;
+
+            /// <summary>
             /// Whether to use separate parsing threads.
             /// </summary>
             [Argument(ArgumentType.AtMostOnce, HelpText = "Use separate parsing threads?", ShortName = "threads", Hide = true)]
@@ -530,6 +560,7 @@ namespace Microsoft.ML.Data
             internal const char Separator = '\t';
             internal const bool HasHeader = false;
             internal const bool TrimWhitespace = false;
+            internal const bool AutoMapLoadColumns = false;
         }
 
         /// <summary>
@@ -682,10 +713,13 @@ namespace Microsoft.ML.Data
                     // Also determine if any columns have a range that extends to the end. If so, then we need
                     // to look at some data to determine the number of source columns.
                     bool needInputSize = false;
+                    bool needColumnLocalization = false;
                     foreach (var col in cols)
                     {
                         if (Utils.Size(col.Source) == 0)
                             throw ch.ExceptUserArg(nameof(Column.Source), "Must specify some source column indices");
+                        if (col.Source.Any(r => r.Name != null))
+                            needColumnLocalization = true;
                         if (!needInputSize && col.Source.Any(r => r.AutoEnd && r.Max == null))
                             needInputSize = true;
                     }
@@ -699,6 +733,33 @@ namespace Microsoft.ML.Data
                         Cursor.GetSomeLines(dataSample, 100, ref lines);
                     else if (headerFile == null && parent.HasHeader)
                         Cursor.GetSomeLines(dataSample, 1, ref lines);
+
+                    if (needColumnLocalization)
+                    {
+                        if (lines == null)
+                            throw ch.ExceptUserArg(nameof(dataSample), "Can't determinate column index without sample data");
+
+                        var header = Parser.GetHeaderFields(parent, lines[0]);
+                        var headerMapping = Enumerable.Zip(header.Indices, header.Spans,
+                            (index, span) => new
+                            {
+                                Index = index,
+                                Span = span
+                            })
+                            .Where(x => !x.Span.IsEmpty)
+                            .GroupBy(x => x.Span.ToString().ToLowerInvariant());
+
+                        foreach (var source in cols.SelectMany(x => x.Source).Where(x => x.Name != null))
+                        {
+                            var mapping = headerMapping.FirstOrDefault(x => x.Key == source.Name.ToLowerInvariant());
+                            if (mapping == null)
+                                throw ch.ExceptUserArg(nameof(dataSample),
+                                    $"Column {source.Name} not found in sample data. To ignore this column, use the {nameof(LoadColumnIgnoreAttribute)}.");
+                            var mappingIndex = mapping.Min(x => x.Index);
+                            source.Min = mappingIndex;
+                            source.Max = mappingIndex;
+                        }
+                    }
 
                     if (needInputSize && inputSize == 0)
                     {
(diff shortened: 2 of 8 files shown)


kevencarneiro:
Some tests are failing due to changes in `LoadFromTextFile` method that now has the input passed as sample data and including SlotNames in files. Shouldn't this verification looks if every existing property is present, but allow new properties to be inserted?
The `EntryPointCatalog` is also failing, but the verification shouldn't follow the same logic? Verify if all existing EntryPoints are present, but allow new EntryPoints?

Or in this case is OK to replace the baseline test file?

harishsk:
@kevencarneiro Thank you for contribution. 
I will review your proposed changes and get back to you soon.

Microsoft.ML.sln
```diff
@@ -94,7 +94,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ML.CpuMath.Perfor
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ML.CpuMath.UnitTests", "test\Microsoft.ML.CpuMath.UnitTests\Microsoft.ML.CpuMath.UnitTests.csproj", "{E97D8B5A-3035-4D41-9B0D-77FF8FB8D132}"
 EndProject
-Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Microsoft.ML.FSharp.Tests", "test\Microsoft.ML.FSharp.Tests\Microsoft.ML.FSharp.Tests.fsproj", "{802233D6-8CC0-46AD-9F23-FEE1E9AED9B3}"
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Microsoft.ML.FSharp.Tests", "test\Microsoft.ML.FSharp.Tests\Microsoft.ML.FSharp.Tests.fsproj", "{802233D6-8CC0-46AD-9F23-FEE1E9AED9B3}"
 EndProject
```
harishsk, line 98: Any idea why this change is necessary?

kevencarneiro: This change was made automatically when I opened the project in Visual Studio for Mac.

harishsk:
The failures you are seeing with respect to slotnames is odd. I am looking into it and will post an update soon.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -194,25 +217,27 @@ public static class TextLoaderSaverCatalog
...
                 throw Contracts.ExceptParam(nameof(path), "File does not exist at path: {0}", path);
             }
 
-            // REVIEW: it is almost always a mistake to have a 'trainable' text loader here.
-            // Therefore, we are going to disallow data sample.
+            var source = new MultiFileSource(path);
             return TextLoader.CreateTextLoader<TInput>(CatalogUtils.GetEnvironment(catalog), hasHeader, separatorChar,
-                allowQuoting, allowSparse, trimWhitespace).Load(new MultiFileSource(path));
+                allowQuoting, allowSparse, trimWhitespace, autoMapLoadColumns, dataSample: source).Load(source);
         }
```
harishsk, line 241: The onnx related test failures you are seeing are a result of this line. 
When you supply a data sample and the file has a header, the header fields are added as annotations to columns which get saved in the onnx model when the pipeline is exported. Since the earlier models were created without the data sample, they don't have annotations and you are seeing failures when comparing against the baseline. 

It is debatable whether we should always add the column annotations when a header is present. There is a comment in the older version that explicitly disallows a data sample. This was prior to my time, so I am not sure what the background is on this. 

@eerhardt / @yaeldekel : Do either of you know why it is considered a mistake to have a 'trainable' text loader here and using a data sample?

@kevencarneiro : It is possible to take your PR and retain the old behavior if you change the lines above to the following:
```
            var source = new MultiFileSource(path);
            var dataSample = autoMapLoadColumns ? source : null;
            return TextLoader.CreateTextLoader<TInput>(CatalogUtils.GetEnvironment(catalog), hasHeader, separatorChar,
                allowQuoting, allowSparse, trimWhitespace, autoMapLoadColumns, dataSample: dataSample).Load(source);
```

harishsk:
@eerhardt / @yaeldekel 
Can at least one of you please review this PR in addition to myself?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1479,16 +1551,38 @@ void ICanSaveModel.Save(ModelSaveContext ctx)
...
+
+                var mappingLoadColumnAttr = memberInfo.GetCustomAttribute<LoadColumnAttribute>();
+                var mappingLoadColumnNameAttr = memberInfo.GetCustomAttribute<LoadColumnNameAttribute>();
+                var memberInfoType = memberInfo is FieldInfo ? "Field" : "Property";
+
+                if (!hasHeader && mappingLoadColumnAttr == null)
+                    throw host.Except(
+                        $"You should set {nameof(hasHeader)} to true in order to use {nameof(LoadColumnNameAttribute)}. " +
+                        $"To automatically map the name of {memberInfoType} to {nameof(LoadColumnNameAttribute)}, set {nameof(autoMapLoadColumns)} to true. " +
+                        $"To specify a fixed column index, use {nameof(LoadColumnAttribute)}. " +
```
harishsk, line 1567: If autoMalLoadColumns is true, then the hasHeader must be true. 
Can we factor this out of the loop as an explicit parameter check at the beginning of the function?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1479,16 +1551,38 @@ void ICanSaveModel.Save(ModelSaveContext ctx)
...
+                        $"Invalid {memberInfoType} name: {memberInfo.Name}.");
```
harishsk, line 1568: This should go in the documentation. Can you please shorten the exception messages (here and below) to be focused on the specific exception?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1479,16 +1551,38 @@ void ICanSaveModel.Save(ModelSaveContext ctx)
...
+                        $"To ignore this member, use {nameof(LoadColumnIgnoreAttribute)}. " +
+                        $"Invalid {memberInfoType} name: {memberInfo.Name}.");
 
                 var mappingAttrName = memberInfo.GetCustomAttribute<ColumnNameAttribute>();
 
                 var column = new Column();
                 column.Name = mappingAttrName?.Name ?? memberInfo.Name;
-                column.Source = mappingAttr.Sources.ToArray();
+                column.Source =  (mappingLoadColumnNameAttr != null)
+                    ? mappingLoadColumnNameAttr.Sources.Select(x => new Range(x)).ToArray()
```
harishsk, line 1582: What happens if two or more of LoadColumn, LoadColumnName and ColumnName are specified and there is a conflict? Can you please identify that and throw an exception.

harishsk:
The failure in EntryPointCatalog is because a new parameter has been added to the public API and it is failing a comparison against the current baseline. The current baseline is stored in test\BaselineOutput\Common\EntryPoints\core_manifest.json
I think it is okay to update the baseline in this case. @yaeldekel : Can you please verify?

---
In reply to: [614830681](https://github.com/dotnet/machinelearning/pull/5021#issuecomment-614830681) [](ancestors = 614830681)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
yaeldekel: Regarding the reason not to have a trainable text loader in the `LoadFromTextFile` API: When you create a trainable text loader, some parameters of the text loader are learned from the data sample. Specifically, slot names, if they exist, and in some cases the cardinality of some columns in the data (the range can be specified with only a minimum and no maximum, indicating the text loader needs to figure out the end of the range on its own). Lets imagine that a user loads a data view using `LoadFromTextFile`, and fits a pipeline using the data view to create a model. They then load new data using `LoadFromTextFile`, and try to apply the model to it. If the data view returned from `LoadFromTextFile` uses a trainable text loader, it could end up having different schemas depending on the actual data in the text file, which might cause the model to fail when applied to the second dataset. 



---
In reply to: [409767689](https://github.com/dotnet/machinelearning/pull/5021#discussion_r409767689) [](ancestors = 409767689)

yaeldekel: Given this, I think that the new `autoMapLoadColumns` option should not be present in `LoadFromTextFile`, only in the `CreateTextLoader` APIs.

---
In reply to: [411977600](https://github.com/dotnet/machinelearning/pull/5021#discussion_r411977600) [](ancestors = 411977600,409767689)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoaderSaverCatalog.cs
```diff
@@ -85,6 +85,16 @@ public static class TextLoaderSaverCatalog
             IMultiStreamSource dataSample = null)
             => new TextLoader(CatalogUtils.GetEnvironment(catalog), options, dataSample);
 
+        // 1.4.0 BACKCOMPAT OVERLOAD -- DO NOT TOUCH
```
yaeldekel, line 88: >1.4.0 BACKCOMPAT OVERLOAD -- DO NOT TOUCH [](start = 11, length = 41)

It seems to me that these overloads are not needed, for two reasons:
First of all, since none of the arguments here are optional, it doesn't actually provide backwards compatibility (if a user has code that only uses some of the arguments, it will not compile with this overload). Second of all, since the arguments in the current signature **are** optional, adding another optional argument at the end of the argument list, wouldn't break existing code, so it is backwards compatible. 
Is there a different reason to have these overloads that I am not aware of?

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1479,16 +1551,38 @@ void ICanSaveModel.Save(ModelSaveContext ctx)
...
-                    throw host.Except($"{(memberInfo is FieldInfo ? "Field" : "Property")} '{memberInfo.Name}' is missing the {nameof(LoadColumnAttribute)} attribute");
+                var ignoreLoadColumnAttr = memberInfo.GetCustomAttribute<LoadColumnIgnoreAttribute>();
+                if (ignoreLoadColumnAttr != null && ignoreLoadColumnAttr.Ignore)
+                    continue;
+
+                var mappingLoadColumnAttr = memberInfo.GetCustomAttribute<LoadColumnAttribute>();
+                var mappingLoadColumnNameAttr = memberInfo.GetCustomAttribute<LoadColumnNameAttribute>();
+                var memberInfoType = memberInfo is FieldInfo ? "Field" : "Property";
+
+                if (!hasHeader && mappingLoadColumnAttr == null)
```
yaeldekel, line 1563: >mappingLoadColumnAttr [](start = 34, length = 21)

It seems that you are assuming here that if `mappingLoadColumnAttr == null`, then necessarily `mappingLoadColumnNameAttr != null`, but you are not checking to make sure this is true.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
yaeldekel: I think that other than to be able to factor this check out of the loop, the `autoMapLoadColumns` argument is not needed, so I would prefer not to have it in the signature of the method at all.

---
In reply to: [409848508](https://github.com/dotnet/machinelearning/pull/5021#discussion_r409848508) [](ancestors = 409848508)

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -1479,16 +1551,38 @@ void ICanSaveModel.Save(ModelSaveContext ctx)
             for (int index = 0; index < memberInfos.Length; index++)
             {
                 var memberInfo = memberInfos[index];
-                var mappingAttr = memberInfo.GetCustomAttribute<LoadColumnAttribute>();
 
-                if (mappingAttr == null)
-                    throw host.Except($"{(memberInfo is FieldInfo ? "Field" : "Property")} '{memberInfo.Name}' is missing the {nameof(LoadColumnAttribute)} attribute");
+                var ignoreLoadColumnAttr = memberInfo.GetCustomAttribute<LoadColumnIgnoreAttribute>();
```
yaeldekel, line 1555: >LoadColumnIgnoreAttribute [](start = 73, length = 25)

Why is this being added? What is the scenario where you would need to have a public field/property in the class but not have it loaded by the text loader?
If this is really necessary, then I would instead of introducing this new class, have the use the `LoadColumnNameAttribute`, and have an empty or null `Name` indicate that this field should be ignored. There is also a `NoColumnAttribute` class that can alternatively be used.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -700,6 +734,33 @@ public Bindings(TextLoader parent, Column[] cols, IMultiStreamSource headerFile,
...
+
+                        var header = Parser.GetHeaderFields(parent, lines[0]);
+                        var headerMapping = Enumerable.Zip(header.Indices, header.Spans,
+                            (index, span) => new
+                            {
+                                Index = index,
+                                Span = span
+                            })
+                            .Where(x => !x.Span.IsEmpty)
+                            .GroupBy(x => x.Span.ToString().ToLowerInvariant());
```
yaeldekel, line 750: >ToLowerInvariant [](start = 60, length = 16)

Column names are usually case sensitive, so there is no need to call `ToLowerInvariant`.

src/Microsoft.ML.Data/DataLoadSave/Text/TextLoader.cs
```diff
@@ -495,6 +518,13 @@ public class Options
                 HelpText = "Data file has header with feature names. Header is read only if options 'hs' and 'hf' are not specified.")]
             public bool HasHeader;
 
+            /// <summary>
+            /// Whether the members name of the class passed as type argument should be mapped to <see cref="LoadColumnNameAttribute"/>.
+            /// When <see langword="true"/> requires the <see cref="HasHeader"/> option to be <see langword="true"/>.
+            /// </summary>
+            [Argument(ArgumentType.AtMostOnce, HelpText = "Use class properties as LoadColumnNameAttribute to read from text file", ShortName = "automap")]
+            public bool AutoMapLoadColumns;
```
yaeldekel, line 526: >AutoMapLoadColumns [](start = 24, length = 18)

This is only used in the generic overload, that creates a `TextLoader` from a given type, and not in the overload that takes an `Options` object, so it doesn't need to be a part of the `Options` class.

yaeldekel:
Hi @kevencarneiro, thanks for tackling this issue!

As @harishsk mentioned, the failure in the `EntryPointCatalog` test is being caused because the `Options` class has changed, and it is used as the input to the `TextLoader`'s entry point. But as I commented in the TextLoader.cs file, I think the `Options` class does not need to change, so it will not be necessary to change the entrypoint catalog baseline. 

The other failure will also be solved if you revert the change that adds the data sample to the constructor of `TextLoader` in the `LoadFromTextFile` API.

---
In reply to: [613231161](https://github.com/dotnet/machinelearning/pull/5021#issuecomment-613231161) [](ancestors = 613231161)

Microsoft.ML.sln
harishsk: Can you please revert this change and verify if everything still works?

michaelgsharp:
Closing for now due to inactivity. If you would like to resume working on this please feel free to either reopen this PR or create a new one.

PR #5031, opened 2020-04-16
Handle NaN optimization metric in AutoML

### LINKED ISSUE #4663: NaN metric value handling in AutoML
AutoML API code is not handing `NaN` values for metrics. During the sweep, when a model returns a `NaN` value for the metric being optimized, AutoML crashes.

See background: https://github.com/dotnet/machinelearning/issues/4648#issuecomment-574886743

### LINKED ISSUE #5042: Return average metrics in AutoML CrossValSummaryRunner
Related to #4663 
See also https://github.com/dotnet/machinelearning/pull/5031#discussion_r409233020

`CrossValSummaryRunner` in AutoML gets invoked when the dataset size is less than 15k rows. It runs 10-fold cross validation and computes the average optimization metric across the folds. It then reports all the metrics from the fold that has the optimization metric closest to this average.
https://github.com/dotnet/machinelearning/blob/214926fcf5753ad62acf32c0759bdaf8fcd13b73/src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs#L76-L78

A better way to do this would be to calculate the average of all the metrics, instantiate a new metrics class with these averages, and return that in the run details. This could reuse the code for calculating the average of non_NaN metrics from #5031. The following two things will need to be figured out and need more discussion:

1. `PerClassLogLoss`: For multiclass classification, the ordering of the class labels may be different across the 10 folds. So, the `PerClassLogLoss` metric from each fold will have different indices for the class labels. In this situation, a standardized ordering would need to be figured out and the averages calculated for each class accordingly.

2. `ConfusionMatrix`: For multiclass classification, the same problem as above needs solving. In addition, for both binary and multiclass classification, what confusion matrix is returned needs discussion. The distribution of class labels will be different across the folds, so what exactly is the "average" of a confusion matrix? Do we return a confusion matrix at all? Do we just return the confusion matrix from the fold with optimization metric closest to the average (current behavior)? If we are going this route, the confusion matrix property in the metrics classes will need to be made internally settable, as there is no constructor that takes the confusion matrix as an argument.

najeeb-kazmi:
Fixes #4663 
Fixes #5042

In AutoML, `CrossValSummaryRunner` is invoked if the dataset contains less than 15000 rows. It runs 10-fold cross validation on it, and then returns the model from the fold with the best optimization metric. It does this by looking for the index of the model with the best metric in the list of run results, and returning the element at this index.

If the metric in all 10 folds is `NaN`, then this index is `-1`, resulting in an `IndexOutOfRangeException`.

### DIFF
--- a/src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
+++ b/src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
@@ -6,7 +6,9 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using Microsoft.ML.Data;
 using Microsoft.ML.Runtime;
+using Microsoft.ML.Trainers.FastTree;
 
 namespace Microsoft.ML.AutoML
 {
@@ -90,7 +92,7 @@ namespace Microsoft.ML.AutoML
         {
             var newResults = results.Where(r => !double.IsNaN(r.score));
             // Return NaN iff all scores are NaN
-            if (newResults.Count() ==0)
+            if (newResults.Count() == 0)
                 return double.NaN;
             // Return average of non-NaN scores otherwise
             return newResults.Average(r => r.score);
--- a/src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
+++ b/src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
@@ -6,9 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using Microsoft.ML.Data;
 using Microsoft.ML.Runtime;
-using Microsoft.ML.Trainers.FastTree;
 
 namespace Microsoft.ML.AutoML
 {
--- a/src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
+++ b/src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
@@ -6,7 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using Microsoft.ML.Data;
+using Google.Protobuf.WellKnownTypes;
 using Microsoft.ML.Runtime;
 
 namespace Microsoft.ML.AutoML
@@ -77,7 +77,7 @@ namespace Microsoft.ML.AutoML
             var bestModel = trainResults.ElementAt(bestFoldIndex).model;
 
             // Get the metrics from the fold whose score is closest to avg of all fold scores
-            var avgScore = trainResults.Average(r => r.score);
+            var avgScore = GetAverageOfNonNaNScores(trainResults);
             var indexClosestToAvg = GetIndexClosestToAverage(trainResults.Select(r => r.score), avgScore);
             var metricsClosestToAvg = trainResults[indexClosestToAvg].metrics;
 
@@ -87,14 +87,32 @@ namespace Microsoft.ML.AutoML
             return (suggestedPipelineRunDetail, runDetail);
         }
 
+        private static double GetAverageOfNonNaNScores(List<(ModelContainer model, TMetrics metrics, Exception exception, double score)> results)
+        {
+            var newResults = results.Where(r => !double.IsNaN(r.score));
+            // Return NaN iff all scores are NaN
+            if (newResults.Count() ==0)
+                return double.NaN;
+            // Return average of non-NaN scores otherwise
+            return newResults.Average(r => r.score);
+        }
+
         private static int GetIndexClosestToAverage(IEnumerable<double> values, double average)
         {
+            // Average will be NaN iff all values are NaN.
+            // Return the first index in this case.
+            if (double.IsNaN(average))
+                return 0;
+
             int avgFoldIndex = -1;
             var smallestDistFromAvg = double.PositiveInfinity;
             for (var i = 0; i < values.Count(); i++)
             {
-                var distFromAvg = Math.Abs(values.ElementAt(i) - average);
-                if (distFromAvg < smallestDistFromAvg || smallestDistFromAvg == double.PositiveInfinity)
+                var value = values.ElementAt(i);
+                if (double.IsNaN(value))
+                    continue;
+                var distFromAvg = Math.Abs(value - average);
+                if (distFromAvg < smallestDistFromAvg)
                 {
                     smallestDistFromAvg = distFromAvg;
                     avgFoldIndex = i;
--- a/src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
+++ b/src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
@@ -6,7 +6,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using Google.Protobuf.WellKnownTypes;
 using Microsoft.ML.Runtime;
 
 namespace Microsoft.ML.AutoML
--- a/src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
+++ b/src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using Microsoft.ML.Data;
 using Microsoft.ML.Runtime;
 
 namespace Microsoft.ML.AutoML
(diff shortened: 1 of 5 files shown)


src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
```diff
@@ -70,6 +71,9 @@ internal class CrossValSummaryRunner<TMetrics> : IRunner<RunDetail<TMetrics>>
 
             // Get the model from the best fold
             var bestFoldIndex = BestResultUtil.GetIndexOfBestScore(trainResults.Select(r => r.score), _optimizingMetricInfo.IsMaximizing);
+            // bestFoldIndex will be -1 if the optimization metric for all folds is NaN.
+            // In this case, return model from the first fold.
+            bestFoldIndex = bestFoldIndex != -1 ? bestFoldIndex : 0;
```
justinormont, line 76: I'd also look into places where the metric gets compared. 

For instance below in `GetIndexClosestToAverage()`:
https://github.com/dotnet/machinelearning/blob/9726b4103e066d410a2b0728fb3e2058c081e15c/src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs#L86-L100

While `GetIndexClosestToAverage()` could be adjusted to handle CV folds returning `NaN`, it would be better to remove the function and instead return a new instance of the metric class w/ the actual averages. The current function was created before the AutoML code had access to create a new instance of the metric classes, so it just returned the closest to the average of the folds.

I would also look at a bit further to where AutoML compares the return metrics from each model in the sweep, where it chooses which is the best. The comparison to `NaN` may also be there.

najeeb-kazmi: I have changed the calculation of average to exclude any folds with NaN metrics. In finding the best run, there are not any comparisons to NaN, but an index of -1 is returned if *all* the runs have NaN metric. I have added a warning there checking for that case.

As for returning a Metrics object containing the averages of the metrics across the folds, I think this would be out of scope for this particular issue. Also, since we only have the type `TMetrics` here, creating the right metric would be fairly involved.

src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
```diff
@@ -70,27 +71,102 @@ internal class CrossValSummaryRunner<TMetrics> : IRunner<RunDetail<TMetrics>>
...
+                    accuracyMicro: GetAverageOfNonNaNScores(newMetrics.Select(x => x.MicroAccuracy)),
+                    accuracyMacro: GetAverageOfNonNaNScores(newMetrics.Select(x => x.MacroAccuracy)),
+                    logLoss: GetAverageOfNonNaNScores(newMetrics.Select(x => x.LogLoss)),
+                    logLossReduction: GetAverageOfNonNaNScores(newMetrics.Select(x => x.LogLossReduction)),
+                    topKPredictionCount: newMetrics.ElementAt(0).TopKPredictionCount,
+                    topKAccuracy: GetAverageOfNonNaNScores(newMetrics.Select(x => x.TopKAccuracy)),
+                    // TODO:
+                    //   Figure out whether class label ordering can be different across different folds.
+                    //   If yes, whether it is possible to get the information from the objects available.
+                    perClassLogLoss: null);
```
justinormont, line 123: The classes can be different in ordering and some can be missing. The missing arises from small datasets or skewed labels where all given classes may not be represented in the training dataset. 

I'm not sure what our log-loss code does for labels which are not in the test dataset (perhaps `0.0` or `NaN`).

justinormont: Would it be possible to return the closest to average fold's `perClassLogLoss`?

src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
```diff
@@ -70,27 +71,102 @@ internal class CrossValSummaryRunner<TMetrics> : IRunner<RunDetail<TMetrics>>
...
+
+                var result = new BinaryClassificationMetrics(
+                    auc: GetAverageOfNonNaNScores(newMetrics.Select(x => x.AreaUnderRocCurve)),
+                    accuracy: GetAverageOfNonNaNScores(newMetrics.Select(x => x.Accuracy)),
+                    positivePrecision: GetAverageOfNonNaNScores(newMetrics.Select(x => x.PositivePrecision)),
+                    positiveRecall: GetAverageOfNonNaNScores(newMetrics.Select(x => x.PositiveRecall)),
+                    negativePrecision: GetAverageOfNonNaNScores(newMetrics.Select(x => x.NegativePrecision)),
+                    negativeRecall: GetAverageOfNonNaNScores(newMetrics.Select(x => x.NegativeRecall)),
+                    f1Score: GetAverageOfNonNaNScores(newMetrics.Select(x => x.F1Score)),
+                    auprc: GetAverageOfNonNaNScores(newMetrics.Select(x => x.AreaUnderPrecisionRecallCurve)));
```
justinormont, line 104: Can you add in the `ConfusionMatrix`? Current users of AutoML get back a confusion matrix for `BinaryClassificationMetrics` and `MulticlassClassificationMetrics`.

Returning the `ConfusionMatrix` from the most average fold is suitable for the moment.

I think this would entail adding the confusion matrix to the constructor:
https://github.com/dotnet/machinelearning/blob/919bc8b886a6bc20c5022e8dad00f84bc1248236/src/Microsoft.ML.Data/Evaluators/Metrics/BinaryClassificationMetrics.cs#L113-L114

Or allowing it to be set after creation: ([src](https://github.com/dotnet/machinelearning/blob/919bc8b886a6bc20c5022e8dad00f84bc1248236/src/Microsoft.ML.Data/Evaluators/Metrics/BinaryClassificationMetrics.cs#L86))
```diff
-public ConfusionMatrix ConfusionMatrix { get; } 
+public ConfusionMatrix ConfusionMatrix { get; internal set; }
```

src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
```diff
@@ -70,27 +71,102 @@ internal class CrossValSummaryRunner<TMetrics> : IRunner<RunDetail<TMetrics>>
...
+                    rSquared: GetAverageOfNonNaNScores(newMetrics.Select(x => x.RSquared)));
+                return result as TMetrics;
+            }
+
+            return null;
+        }
+
+        private static double GetAverageOfNonNaNScores(IEnumerable<double> results)
+        {
+            var newResults = results.Where(r => !double.IsNaN(r));
```
justinormont, line 146: What's your thoughts on metrics that can be +/-Infinity? Should they included in the average or not? Any trade-offs?

Behavior currently:
* Input includes only `double.PositiveInfinity` _xor_ `double.NegativeInfinity`, `GetAverageOfNonNaNScores()` would return average=+/-Inf
* Input has both `double.PositiveInfinity` _and_ `double.NegativeInfinity`, `GetAverageOfNonNaNScores()` would return average=`NaN`

src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
justinormont: Thanks. Current method looks good for the moment. 

For averaging the `ConfusionMatrix` in the future, it stores the class names within itself. The class names can be used to align the classes and allow for proper averaging.

https://github.com/dotnet/machinelearning/blob/770327a2074d8ab39a023bc9252a713a3afa8293/src/Microsoft.ML.Data/Evaluators/Metrics/ConfusionMatrix.cs#L62

It mentions the class index may also be stored, though I'm unsure how that could arise. Perhaps for binary classification there's no class names to store?

src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
justinormont: Thanks. Looks good. 

In the future, we can use the `MetricsStatistics` class to do the averaging so that we're not duplicating functionality. It has the benefit of handling the `PerClassLogLoss`, though I suspect it's not properly handing the class ordering differences.

https://github.com/dotnet/machinelearning/blob/d849ba4c7831586018d5b2cca6bdd2b1fa228ddf/src/Microsoft.ML.Transforms/MetricStatistics.cs#L84-L88

src/Microsoft.ML.AutoML/Experiment/Runners/CrossValSummaryRunner.cs
najeeb-kazmi: So I've looked at the calculation of `LogLoss`, which is the main metric I would be concerned about being infinity. It looks to me that it can only be `Double.PositiveInfinity` if the aggregation over the all the rows in the evaluation set overflows  `Double.MaxValue`.

Here's the code for binary classification:
https://github.com/dotnet/machinelearning/blob/062be280a9d6de23838d4db5ad93fea9d7d0c1f6/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs#L465-L466

This is only `Double.PositiveInfinity` if `_logLoss` is, which will only be the case if it overflows here

https://github.com/dotnet/machinelearning/blob/062be280a9d6de23838d4db5ad93fea9d7d0c1f6/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs#L527

because `logloss` cannot be `Double.PositiveInfinity`. `prob` will never be 0 if  label is positive, and `prob` will never be 1 if label is negative.

https://github.com/dotnet/machinelearning/blob/062be280a9d6de23838d4db5ad93fea9d7d0c1f6/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs#L663-L677

This does mean that if `prob` is `NaN` for any row, then `logloss` will be `NaN` for that row, and `_logLoss` for the entire evaluation set will be `NaN`.

najeeb-kazmi: Similarly for multiclass classification:

`LogLoss` will only be `Double.PositiveInfinity` if `_totalLogLoss` is
https://github.com/dotnet/machinelearning/blob/062be280a9d6de23838d4db5ad93fea9d7d0c1f6/src/Microsoft.ML.Data/Evaluators/MulticlassClassificationEvaluator.cs#L276

and that will only be the case if it overflows here
https://github.com/dotnet/machinelearning/blob/062be280a9d6de23838d4db5ad93fea9d7d0c1f6/src/Microsoft.ML.Data/Evaluators/MulticlassClassificationEvaluator.cs#L332

because `loglossCurr` itself will never be `Double.PositiveInfinity`, as it can only be as large as `-Math.Log(Epsilon)`, which is approximately 34.54
https://github.com/dotnet/machinelearning/blob/062be280a9d6de23838d4db5ad93fea9d7d0c1f6/src/Microsoft.ML.Data/Evaluators/MulticlassClassificationEvaluator.cs#L451-L464

najeeb-kazmi: I agree with the principle of reporting `NaN` log loss if a single row has `NaN` probability and log loss. Because if there is a problem, the user needs to know via the metric. While it *may* be fine to suppress this for a handful of rows, this leads to questions on what to do when a lot or majority of rows have this problem. So, I think it is better to not suppress `NaN`s in the calculation of total log loss.

By the same principle, we should not be suppressing infinities in the metrics for all the folds. If one fold has an infinity, the average returned should be infinity.

justinormont: For log-loss, the `Infinity` occurs when the model is perfectly confident (p=0 or p=1) and wrong, for any of the predicted labels in the scoring dataset.

I created a dotnet fiddle to show this:
https://dotnetfiddle.net/DPIn5Y

It's demonstrating the code for binary log-loss:
https://github.com/dotnet/machinelearning/blob/062be280a9d6de23838d4db5ad93fea9d7d0c1f6/src/Microsoft.ML.Data/Evaluators/BinaryClassifierEvaluator.cs#L663-L677

najeeb-kazmi: Ah yes, that's the true label, not the predicted label.

justinormont: I filed an issue to investigate if we should threshold the probability for log-loss in binary classification. This would cause the AutoML code (discussed above) to not receive a `Infinity` value.

Issue: https://github.com/dotnet/machinelearning/issues/5055 **Investigate thresholding binary log-loss**

PR #5036, opened 2020-04-16
Update package dependency in CodeGenerator

LittleLittleCloud:
This PR 
- make CodeGenerator depends on a release AutoML package instead of project, to avoid the latest change in model load/unload mechanism, which will break ModelBuilder.


justinormont:
Is there an issue filed for this? What's new in the latest model load/unload mechanism? Why does this break ModelBuilder?

LittleLittleCloud:
> Is there an issue filed for this? What's new in the latest model load/unload mechanism? What breaks in ModelBuilder?

The latest change in loading model is it will check the namespace of customMapping, so we can't just copy the source code of customMapping from Model generating program (ModelBuilder) to Model consuming program since ModelBuilder is still using the old ML.Net package. This change will be fixed after we move customMapping classes into CodeGen Helper package. But before that, we need to target the dependency of CodeGenerator to the old ML.Net package if we want to make any changes in CodeGen.

src/Microsoft.ML.CodeGenerator/Microsoft.ML.CodeGenerator.csproj
```diff
@@ -17,7 +17,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\Microsoft.ML.AutoML\Microsoft.ML.AutoML.csproj" />
+    <PackageReference Include="Microsoft.ML.AutoML" Version="0.17.0-preview2-28706-1" />
```
justinormont, line 20: _(tagged on a line of code to allow for a threaded conversation)_

> > Is there an issue filed for this? What's new in the latest model load/unload mechanism? What breaks in ModelBuilder?
> 
> The latest change in loading model is it will check the namespace of customMapping, so we can't just copy the source code of customMapping from Model generating program (ModelBuilder) to Model consuming program since ModelBuilder is still using the old ML.Net package. This change will be fixed after we move customMapping classes into CodeGen Helper package. But before that, we need to target the dependency of CodeGenerator to the old ML.Net package if we want to make any changes in CodeGen.
> 

Is there a timeline for when we can revert this PR?

LittleLittleCloud: we can remove that after preview3 is release and ModelBuilder upgrade to it.

justinormont: @eerhardteerhardt, @harishsk, @antoniovs1029 : Was there a breaking change for how models are loaded? Issue seems to be the use of namespaces in customMapping was changed.

harishsk: There was a change to customMapping but you don't need to pin the version number. If you change the generated code to not supply a lambda and actually supply the delegate of the mapper from the customMapper it should work correctly.

LittleLittleCloud: Thanks for the suggestion but we don't want to update CustomMapping class in CodeGen since we are going to move them into a separate package. To us pin to version number is the quickest way to work around

harishsk: Can you please elaborate what you mean? How would moving CodeGen to a different class affect the fix?

LittleLittleCloud: We are not moving CodeGen but customMapping class to a separate package. And by that time CodeGen will also refer to that separate package, and those customMappings will be able to loaded because they will be from the same dll of where the model is generated.

JakeRadMSFT: @harishsk does the old method of registering the custom mapping still work? or was the custom mapping change breaking?

LittleLittleCloud: @harishsk another reason is that the last released(1.5.0-preview2) ML.Net can't load model generated by the latest ML.Net(1.5.0-preview3). So we need CodeGenerator and ModelBuilder pin to the old ML.Net version, as the latest ML.Net is not released yet.

PR #5038, opened 2020-04-16
clarifying parameters on time series

gvashishtha:

Some context:

Also note that, here we are talking about three different parameters:
1.	seriesLength is the length of the whole timeseries.
2.	trainSize is the number of points from the beginning of the series that are used for training.
3.	windowSize is the size of rolling window during the training of SSA.

The first parameter is dictated by the series and even though it’s not used in SSA, it’s not meaningful for the user to “set” it arbitrarily. The third parameter is *recommended* to be at least twice as the biggest seasonality in the data that the user care about. The second parameter must be more twice the third parameter. 

So to summarize we should have:
1.	2*Seasonality length < windowSize
2.	2*windowSize < trainSize
3.	trainSize < seriesLength

The first constraint is recommended while the other two are required.

### DIFF
--- a/src/Microsoft.ML.TimeSeries/AdaptiveSingularSpectrumSequenceModeler.cs
+++ b/src/Microsoft.ML.TimeSeries/AdaptiveSingularSpectrumSequenceModeler.cs
@@ -209,9 +209,13 @@ namespace Microsoft.ML.Transforms.TimeSeries
         /// The constructor for Adaptive SSA model.
         /// </summary>
         /// <param name="env">The exception context.</param>
-        /// <param name="trainSize">The length of series from the beginning used for training. (parameter N)</param>
+        /// <param name="trainSize">The length of series from the beginning used for training (parameter N).
+        ///  Must be at least twice the windowSize.!-- </param>
         /// <param name="seriesLength">This parameter must be greater than windowSize</param>
-        /// <param name="windowSize">The length of the window on the series for building the trajectory matrix (parameter L). Should be more than twice the maximum seasonality in the data.</param>
+        /// <param name="windowSize">The length of the window on the series for building the trajectory matrix (parameter L).
+        ///  We recommend you set this to be more than twice the maximum seasonality in the data. For example, 
+        /// if your data can exhibit both monthly and yearly seasonality, and you have data points from each day, 
+        /// set this to be twice the number of days in a year.</param>
         /// <param name="discountFactor">The discount factor in [0,1] used for online updates (default = 1).</param>
         /// <param name="rankSelectionMethod">The rank selection method (default = Exact).</param>
         /// <param name="rank">The desired rank of the subspace used for SSA projection (parameter r). This parameter should be in the range in [1, windowSize].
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.TimeSeries/AdaptiveSingularSpectrumSequenceModeler.cs
```diff
@@ -209,9 +209,13 @@ private static VersionInfo GetVersionInfo()
         /// The constructor for Adaptive SSA model.
         /// </summary>
         /// <param name="env">The exception context.</param>
-        /// <param name="trainSize">The length of series from the beginning used for training.</param>
-        /// <param name="seriesLength">The length of series that is kept in buffer for modeling (parameter N).</param>
-        /// <param name="windowSize">The length of the window on the series for building the trajectory matrix (parameter L).</param>
+        /// <param name="trainSize">The length of series from the beginning used for training (parameter N).
+        ///  Must be at least twice the windowSize.!-- </param>
```
mstfbl, line 213:
```suggestion
        /// Must be at least twice the windowSize.</param>
```

src/Microsoft.ML.TimeSeries/AdaptiveSingularSpectrumSequenceModeler.cs
```diff
@@ -209,9 +209,13 @@ private static VersionInfo GetVersionInfo()
...
+        /// <param name="seriesLength">This parameter must be greater than windowSize</param>
```
mstfbl, line 214:
```suggestion
        /// <param name="seriesLength">This parameter must be greater than windowSize.</param>
```

src/Microsoft.ML.TimeSeries/AdaptiveSingularSpectrumSequenceModeler.cs
```diff
@@ -209,9 +209,13 @@ private static VersionInfo GetVersionInfo()
...
+        /// <param name="windowSize">The length of the window on the series for building the trajectory matrix (parameter L).
+        ///  We recommend you set this to be more than twice the maximum seasonality in the data. For example,
```
mstfbl, line 216:
```suggestion
        /// We recommend you set this to be more than twice the maximum seasonality in the data. For example,
```

src/Microsoft.ML.TimeSeries/AdaptiveSingularSpectrumSequenceModeler.cs
```diff
@@ -209,9 +209,13 @@ private static VersionInfo GetVersionInfo()
...
         /// <param name="env">The exception context.</param>
-        /// <param name="trainSize">The length of series from the beginning used for training.</param>
-        /// <param name="seriesLength">The length of series that is kept in buffer for modeling (parameter N).</param>
-        /// <param name="windowSize">The length of the window on the series for building the trajectory matrix (parameter L).</param>
+        /// <param name="trainSize">The length of series from the beginning used for training (parameter N).
+        ///  Must be at least twice the windowSize.!-- </param>
+        /// <param name="seriesLength">This parameter must be greater than windowSize</param>
+        /// <param name="windowSize">The length of the window on the series for building the trajectory matrix (parameter L).
+        ///  We recommend you set this to be more than twice the maximum seasonality in the data. For example, 
+        /// if your data can exhibit both monthly and yearly seasonality, and you have data points from each day,
```
mstfbl, line 217:
```suggestion
        /// if your data can exhibit both monthly and yearly seasonality, and you have data points from each day,
```

PR #5041, opened 2020-04-20
Converted potentially large variables to type long

### LINKED ISSUE #3228: Overflow in MultiClassNaiveBayes
We're storing the count of rows in an `int`. This causes an overflow for large datasets. In my case, Criteo 1TB w/ 4.4B rows. Recommend changing to a `long`.

https://github.com/dotnet/machinelearning/blob/d2bf3e72fa935b12644b0a695ad0d044a8077a83/src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs#L242

Error:
```
Unexpected exception: Arithmetic operation resulted in an overflow., 'System.OverflowException'
   at System.Linq.Enumerable.Sum(IEnumerable`1 source)
```

mstfbl:
Fixes #3228

As explained in Issue #3228, very large datasets more than 2.14 billion rows of data can cause overflow when, say, the sum of these labels are obtained, **and** if these are stored as ints. This PR converts arrays and matrices for storing labels and features in their respective histograms from type `int` to type `long`. In addition, this PR updates the version of NaiveBayesMulticlassTrainer's Loader to preserve backwards compatibility.

### DIFF
--- a/src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
+++ b/src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
@@ -259,12 +259,26 @@ namespace Microsoft.ML.Trainers
         /// <summary>
         /// Get the label histogram.
         /// </summary>
-        public IReadOnlyList<long> GetLabelHistogram() => _labelHistogram;
+        [Obsolete("This API is deprecated, please use GetLabelHistogramLong() which returns _labelHistogram " +
+            "with type IReadOnlyList<long> to avoid overflow errors with large datasets.", true)]
+        public IReadOnlyList<int> GetLabelHistogram() => Array.ConvertAll(_labelHistogram, x => (int)x);
+
+        /// <summary>
+        /// Get the label histogram with generic type long.
+        /// </summary>
+        public IReadOnlyList<long> GetLabelHistogramLong() => _labelHistogram;
 
         /// <summary>
         /// Get the feature histogram.
         /// </summary>
-        public IReadOnlyList<IReadOnlyList<long>> GetFeatureHistogram() => _featureHistogram;
+        [Obsolete("This API is deprecated, please use GetFeatureHistogramLong() which returns _featureHistogram " +
+            "with type IReadOnlyList<long> to avoid overflow errors with large datasets.", true)]
+        public IReadOnlyList<IReadOnlyList<int>> GetFeatureHistogram() => Array.ConvertAll(_featureHistogram, x => Array.ConvertAll(x, y=> (int)y));
+
+        /// <summary>
+        /// Get the feature histogram with generic type long.
+        /// </summary>
+        public IReadOnlyList<IReadOnlyList<long>> GetFeatureHistogramLong() => _featureHistogram;
 
         /// <summary>
         /// Instantiates new model parameters from trained model.
--- a/src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
+++ b/src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
@@ -231,7 +231,8 @@ namespace Microsoft.ML.Trainers
         {
             return new VersionInfo(
                 modelSignature: "MNABYPRD",
-                verWrittenCur: 0x00010001, // Initial
+                //verWrittenCur: 0x00010001, // Initial
+                verWrittenCur: 0x00010002, // Histograms are of type long
                 verReadableCur: 0x00010001,
                 verWeCanReadBack: 0x00010001,
                 loaderSignature: LoaderSignature,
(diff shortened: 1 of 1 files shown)


src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
```diff
@@ -294,25 +294,25 @@ private NaiveBayesMulticlassModelParameters(IHostEnvironment env, ModelLoadConte
             : base(env, LoaderSignature, ctx)
         {
             // *** Binary format ***
-            // int: _labelCount
-            // int[_labelCount]: _labelHistogram
+            // int: _labelCount (read during reading of _labelHistogram in ReadLongArray())
```
mstfbl, line 297: I added this comment as we are not explicitly reading _labelCount here, but in `ReadLongArray()` as shown below:

https://github.com/dotnet/machinelearning/blob/214926fcf5753ad62acf32c0759bdaf8fcd13b73/src/Microsoft.ML.Core/Utilities/Stream.cs#L681-L688

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
```diff
@@ -339,22 +339,23 @@ private protected override void SaveCore(ModelSaveContext ctx)
...
             // int: _labelCount
-            // int[_labelCount]: _labelHistogram
+            // long[_labelCount]: _labelHistogram
             // int: _featureCount
-            // int[_labelCount][_featureCount]: _featureHistogram
+            // long[_labelCount][_featureCount]: _featureHistogram
             // int[_labelCount]: _absentFeaturesLogProb
-            ctx.Writer.WriteIntArray(_labelHistogram.AsSpan(0, _labelCount));
+            ctx.Writer.Write(_labelCount);
+            ctx.Writer.WriteLongStream(_labelHistogram);
```
mstfbl, line 347: `.AsSpan()` is not implemented for converting from 'System.Span<long>' to 'System.Collections.Generic.IEnumerable<long>. The only other way around this is using `Array.Copy()`, but that introduces needless array copying. I also don't see a case where not all of `_labelHistogram` is not serialized. As such, `_labelHistogram` is always serialized whole here.

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
```diff
@@ -339,22 +339,23 @@ private protected override void SaveCore(ModelSaveContext ctx)
...
             ctx.Writer.Write(_featureCount);
             for (int i = 0; i < _labelCount; i += 1)
             {
                 if (_labelHistogram[i] > 0)
-                    ctx.Writer.WriteIntsNoCount(_featureHistogram[i].AsSpan(0, _featureCount));
+                    ctx.Writer.WriteLongStream(_featureHistogram[i]);
```
mstfbl, line 352: Sames as above.

`.AsSpan()` is not implemented for converting from 'System.Span<long>' to 'System.Collections.Generic.IEnumerable<long>. The only other way around this is using `Array.Copy()`, but that introduces needless array copying. I also don't see a case where not all of `_labelHistogram` is not serialized. As such, `_labelHistogram` is always serialized whole here.

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
```diff
@@ -294,25 +309,33 @@ private NaiveBayesMulticlassModelParameters(IHostEnvironment env, ModelLoadConte
...
-            // int[_labelCount][_featureCount]: _featureHistogram
+            // long[_labelCount][_featureCount]: _featureHistogram
             // int[_labelCount]: _absentFeaturesLogProb
-            _labelHistogram = ctx.Reader.ReadIntArray() ?? new int[0];
+            if (ctx.Header.ModelVerWritten >= 0x00010002)
+                _labelHistogram = ctx.Reader.ReadLongArray() ?? new long[0];
+            else
+            {
+                _labelHistogram = Array.ConvertAll(ctx.Reader.ReadIntArray() ?? new int[0], x => (long)x);
+            }
```
harishsk, line 322: If ReadIntArray returns null, it likely means the file is bad. Should you be throwing an error in this case?

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
```diff
@@ -294,25 +309,33 @@ private NaiveBayesMulticlassModelParameters(IHostEnvironment env, ModelLoadConte
...
             for (int iLabel = 0; iLabel < _labelCount; iLabel += 1)
             {
                 if (_labelHistogram[iLabel] > 0)
                 {
-                    _featureHistogram[iLabel] = ctx.Reader.ReadIntArray(_featureCount);
+                    if (ctx.Header.ModelVerWritten >= 0x00010002)
+                        _featureHistogram[iLabel] = ctx.Reader.ReadLongArray(_featureCount);
+                    else
+                        _featureHistogram[iLabel] = Array.ConvertAll(ctx.Reader.ReadIntArray(_featureCount) ?? new int[0], x => (long)x);
                     for (int iFeature = 0; iFeature < _featureCount; iFeature += 1)
```
harishsk, line 339: Same comment as above

src/Microsoft.ML.StandardTrainers/Standard/MulticlassClassification/MulticlassNaiveBayesTrainer.cs
mstfbl: Hey Harish, the array being read from `ctx.Reader.ReadIntArray(int size)` can return null if the size of the array being loaded is 0. Here's the source code:

https://github.com/dotnet/machinelearning/blob/7628d6cd06a0cf8294ddeebbb9613aaec72ab4e7/src/Microsoft.ML.Core/Utilities/Stream.cs#L605-L641

Issue #5044, opened 2020-04-21
support AVX instruction set at CPUMath native


frank-dong-ms:
In CPUMath native we only support SSE.
In dotnet framework and dotnet core 2.1 we are using CPUMath native directly which means we only have SSE.
In dotnet core 2.2 or later version, dotnet core supports hardware Intrinsics so have support for both AVX and SSE.

Support AVX in CPUMath native will have performance benefit as well as generates identical result from different dotnet framework version using.

Issue #3856, opened 2019-06-12
Multi class LR behaves differently on .NetCore 3.0


yaeldekel:
Four tests are failing on the .NetCore 3.0 builds:
`EnsemblesMultiAveragerTest`
`EnsemblesMultiClassBootstrapSelectorTest`
`EnsemblesMultiVotingCombinerTest`
`EnsemblesMultiStackCombinerTest`

The difference in the baselines is something like this:
`L1 regularization selected 13 of 15 weights.` vs.
`L1 regularization selected 11 of 15 weights.`

frank-dong-ms:
Investigating now, caused by different version of CPUMath gives slightly different result

frank-dong-ms:
The difference comes from CPUMath using different instruction set:
  1. net framework and net core 2.1 uses CpuMathUtils.netstandard that uses SSE instruction set;
   2. net core 3.1 uses CpuMathUtils.netcoreapp that uses AVX, SSE or direct floating point calculation depending on hardward avaibility.
AVX and SSE generates slightly different result due to nature of floating point math.
Use below issue to track: https://github.com/dotnet/machinelearning/issues/5044

frank-dong-ms:
created separate baseline for netcore app 31

Issue #1096, opened 2018-09-28
BaseLine Mismatching for various tests when run against Release-Intrinsics


Anipik:
The List of the tests that fail are

- [x] TestCrossValidationMacro
- [x] BinaryClassifierLogisticRegressionBinNormTest
- [x] BinaryClassifierLogisticRegressionGaussianNormTest
- [x] BinaryClassifierLogisticRegressionNonNegativeTest
- [x] BinaryClassifierLogisticRegressionNormTest
- [x] BinaryClassifierTesterThresholdingTest
- [x] BinaryClassifierSymSgdTest
- [x] BinaryClassifierLogisticRegressionTest
- [x] BinaryClassifierPerceptronTest
- [ ] MulticlassLRNonNegativeTest
- [ ] DefaultCalibratorPerceptronTest
- [ ] PAVCalibratorPerceptronTest
- [ ] FastTreeBinaryClassificationCategoricalSplitTest
- [x] LinearClassifierTest
- [x] NoCalibratorLinearSvmTest
- [x] PAVCalibratorLinearSvmTest
- [x] PcaAnomalyTest
- [x] RandomCalibratorPerceptronTest
- [x] RegressorOlsTestOne
- [x] LpGcNormAndWhiteningWorkout
- [x] MultiClassificationLRSaveModelToOnnxTest
- [x] BinaryClassificationLRSaveModelToOnnxTest
- [x] EntryPointLinearPredictorSummary
- [x] EntryPointPcaPredictorSummary
- [x] BinaryClassifierFieldAwareFactorizationMachineTest
- [x] PcaWorkout
- [x] RffWorkout
- [x] SavePipeSsaSpikeNoData // unhandled exception
- [x] ChangePointDetectionWithSeasonality
- [ ] EntryPointPipelineEnsemble
All of these tests fail because the baseline numbers are different on netcoreapp3.0

cc @danmosemsft @eerhardt @shauheen @Ivanidzo4ka

Anipik:
Some of the tests that don't complete or  fail
- [ ]
- [ ]
- [ ]

Anipik:
@shauheen can you assign some one to look into following failures 
 RffWorkout
 SavePipeSsaSpikeNoData // unhandled exception
 ChangePointDetectionWithSeasonality
 EntryPointPipelineEnsemble

These are not caused by mismatching in baseline files

Anipik:
Tests Disabled For Netcoreapp3.0 

[...]

PR #5047, opened 2020-04-22
NetCore3.1 generates different test result issue

frank-dong-ms:
Address issues that NetCore 3.1 generates different test result:
https://github.com/dotnet/machinelearning/issues/1096
https://github.com/dotnet/machinelearning/issues/3856

The difference comes from CPUMath using different instruction set:
1. dotnet framework and dotnet core 2.1 uses CpuMathUtils.netstandard that uses SSE instruction set through CPUMath native library;
2. Since dotnet core 2.2, dotnet core has direct support for AVX and SSE hardware Intrinsics so dotnet core 3.1 use AVX or SSE instruction set directly. Dotnet core 3.1 uses CpuMathUtils.netcoreapp that uses AVX, SSE or direct floating point calculation depending on hardward avaibility. 

AVX and SSE generates slightly different result due to nature of floating point math.
Use below issue to track: #5044

So in short term we are separating test baseline for netcore 3.1 until we add AVX support at CPUMath native library.


test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -111,22 +111,41 @@ protected override void Initialize()
...
+            // Use netcore 3.1 result file if necessary.
+            // The small difference comes from CPUMath using different instruction set:
+            // 1. net framework and net core 2.1 uses CpuMathUtils.netstandard that uses SSE instruction set;
+            // 2. net core 3.1 uses CpuMathUtils.netcoreapp that uses AVX, SSE or direct floating point calculation
+            // depending on hardward avaibility.
+            // AVX and SSE generates slightly different result due to nature of floating point math.
+            // So Ideally we should adding AVX support at CPUMath native library, 
+            // use below issue to track: https://github.com/dotnet/machinelearning/issues/5044
+            if (AppDomain.CurrentDomain.GetData("FX_PRODUCT_VERSION") != null)
+                configurationDirs.Add("NetCore3.1Result");
```
harishsk, line 147: This is a good change overall to move to multiple configuration dirs. 

What happens if we have stored different results for win-x64-netcoreapp21, win-x64-netcoreapp31, win-x86-netcoreapp31, winx-x86-netcoreapp21?

Even if we store netcoreapp21 results by default into win-x64 and win-x86, we would still 2 separate directories for win-x64-netcoreapp31 and win winx86-netcoreapp31.

Does that sound correct? Or am I missing something?

Also can you please change the directory name to be just reflect the configuration (e.g. netcoreapp31 as opposed to NetCore3.1Result)?

frank-dong-ms: Yes, I can change the folder name to netcoreapp31.

What happens if we have stored different results for win-x64-netcoreapp21, win-x64-netcoreapp31, win-x86-netcoreapp31, winx-x86-netcoreapp21?

 - in that case, we have to return win-x64-netcoreapp21, win-x64-netcoreapp31, win-x86-netcoreapp31, winx-x86-netcoreapp21 as configuration directory and make sure we don't have any "expected" result file at folder win-64 or netcoreapp31. And during get file stage we will check one by one if file exists. Let me make the logic complete and send out new iteration.


---
In reply to: [413162621](https://github.com/dotnet/machinelearning/pull/5047#discussion_r413162621) [](ancestors = 413162621)

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -111,22 +111,61 @@ protected override void Initialize()
...
+            if (AppDomain.CurrentDomain.GetData("FX_PRODUCT_VERSION") != null)
+                configurationDirs.Add("netcoreapp31");
+
+            // don't need x64 as this is the default case
+            if (!Environment.Is64BitProcess)
+                configurationDirs.Add("x86");
+
             if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
-                return "osx-x64";
+                configurationDirs.Add("osx-x64");
```
harishsk, line 145: I think the logic below should apply to these as well. We may have osx-x64, osx-x64-netcoreapp21 and osx-netcoreapp31 (and similarly for linux)

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -111,22 +111,61 @@ protected override void Initialize()
...
-                return "linux-x64";
+                configurationDirs.Add("linux-x64");
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                 if (Environment.Is64BitProcess)
-                    return "win-x64";
+                {
+                    configurationDirs.Add("win-x64");
+
+                    if (AppDomain.CurrentDomain.GetData("FX_PRODUCT_VERSION") != null)
+                        configurationDirs.Add($"win-x64-netcoreapp31");
```
harishsk, line 154: Since we will be applying this for all the OSes, this can be factored out and checked once outside this if block

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -111,22 +111,61 @@ protected override void Initialize()
             ML.AddStandardComponents();
         }
 
-        private string GetConfigurationDir()
-        {
+        private IEnumerable<string> GetConfigurationDirs()
+        {
+            // Sometimes different configuration will have combination.
+            // for example: one test can run on windows, have different result both
```
harishsk, line 117: Thanks for explaining the rationale in detail.

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -111,22 +111,61 @@ protected override void Initialize()
...
+
+                    if (AppDomain.CurrentDomain.GetData("FX_PRODUCT_VERSION") != null)
+                        configurationDirs.Add($"win-x64-netcoreapp31");
+                    else
+                        configurationDirs.Add($"win-x64-netcoreapp21");
+                }
                 else
-                    return "win-x86";
+                {
+                    configurationDirs.Add("win-x86");
```
harishsk, line 161: The order of these should be reversed. 
We should check from the most specific to the least specific. The structure in the folder names we have is <os>-<processor>-<framework>. So we should check win-x86-netcoreapp31 first, followed by win-x86 followed by 
 common.

frank-dong-ms: yes, that make sense, thanks

---
In reply to: [413234549](https://github.com/dotnet/machinelearning/pull/5047#discussion_r413234549) [](ancestors = 413234549)

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
frank-dong-ms: yes, in theory we should apply this to all OSes but since we haven't investigate osx and linux case yet I don't want to make all the combination now because that seems little bit overkill and make all the combination will make the file exists check every time we try to reference baseline file that will likely impact our performance. 
I'm considering add osx and linux if this is necessary after investigation, make sense to you?

---
In reply to: [413231007](https://github.com/dotnet/machinelearning/pull/5047#discussion_r413231007) [](ancestors = 413231007)

harishsk: Yes, there might be a small delay because we are checking for an extra directory. But this delay happens in CI verification tests and doesn't represent the true performance of the algorithms. Those are covered by the benchmark tests.
I think it is okay to add it right now. But if you wish to add it later after investigation, I am okay with that too. 

---
In reply to: [413245167](https://github.com/dotnet/machinelearning/pull/5047#discussion_r413245167) [](ancestors = 413245167,413231007)

frank-dong-ms: Yeah, I think we better add any specific combination in future if necessary to keep the logic simple

---
In reply to: [413247756](https://github.com/dotnet/machinelearning/pull/5047#discussion_r413247756) [](ancestors = 413247756,413245167,413231007)

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
frank-dong-ms: yeah, let's add these combination if necessary

---
In reply to: [413230475](https://github.com/dotnet/machinelearning/pull/5047#discussion_r413230475) [](ancestors = 413230475)

PR #5057, opened 2020-04-23
Fixed X64Fact tests

mstfbl:
Fixed X64Fact tests with x86 specific baselines added

Currently obtaining baselines from x86 builds from CI


harishsk:
Yes, I noticed that too when I was debugging earlier. 

Can you please do me a favor? 

Now that we have gone towards different baselines, I think it would be good to be able to easily track which tests are using configuration specific baselines. Can you please add some code in BaseTestBaseline.cs to log whenever we use a configuration specific baseline? I just realized that this is something we should do. 


---
In reply to: [399621483](https://github.com/dotnet/machinelearning/pull/5057#issuecomment-399621483) [](ancestors = 399621483)

harishsk:
You can do it as a separate PR if you wish.

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -238,6 +239,15 @@ protected void Log(string fmt, params object[] args)
             Output.WriteLine(fmt, args);
         }
 
+        protected void LogConfigurationSpecificBaseline(string baselineConfigDir)
```
harishsk, line 242: So this is going to log for every file comparison? Can this be done only once per test?

test/Microsoft.ML.Predictor.Tests/TestPredictors.cs
```diff
@@ -187,7 +187,7 @@ public void MulticlassSdcaTest()
         /// <summary>
         /// Multiclass Logistic Regression test with a tree featurizer.
         /// </summary>
-        [X64Fact("x86 output differs from Baseline")]
```
frank-dong-ms, line 190: >X64Fact [](start = 9, length = 7)

can we remove X64Fact now?

harishsk: As @mstfbl pointed out above, I think there are still around 4 tests that are causing test process crashes in x86. They need investigation.

frank-dong-ms: I see, thanks

---
In reply to: [414314991](https://github.com/dotnet/machinelearning/pull/5057#discussion_r414314991) [](ancestors = 414314991)

mstfbl:
I will make a separate PR for logging when a configuration specific benchmark is utilized.

PR #5060, opened 2020-04-24
Added logging for when specific configuration baseline is used

mstfbl:
Certain unit tests require configuration specific baselines, for when they differ from "normal" baselines. A unit test may also use more than one configuration, such as `linux-x64` and `netcoreapp31`. 

This PR adds logging for each unit test that utilizes a configuration specific baseline. This PR also corrects a small typo.

### DIFF
--- a/test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
+++ b/test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
@@ -97,6 +97,7 @@ namespace Microsoft.ML.RunTests
             _baselineCommonDir = Path.Combine(baselineRootDir, "Common");
             _baselineBuildStringDir = Path.Combine(baselineRootDir, BuildString);
             _baselineConfigDirs = GetConfigurationDirs();
+            _usedSpecificBaselineConfigs = new HashSet<string>();
 
             string logDir = Path.Combine(OutDir, _logRootRelPath);
             Directory.CreateDirectory(logDir);
--- a/test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
+++ b/test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
@@ -75,7 +75,7 @@ namespace Microsoft.ML.RunTests
         private string _baselineCommonDir;
         private string _baselineBuildStringDir;
         private IEnumerable<string> _baselineConfigDirs;
-        private HashSet<string> _usedSpecificBaselineConfigs;
+        private string _usedSpecificBaselineConfig;
 
         // The writer to write to test log files.
         protected TestLogger TestLogger;
@@ -97,7 +97,6 @@ namespace Microsoft.ML.RunTests
             _baselineCommonDir = Path.Combine(baselineRootDir, "Common");
             _baselineBuildStringDir = Path.Combine(baselineRootDir, BuildString);
             _baselineConfigDirs = GetConfigurationDirs();
-            _usedSpecificBaselineConfigs = new HashSet<string>();
 
             string logDir = Path.Combine(OutDir, _logRootRelPath);
             Directory.CreateDirectory(logDir);
@@ -170,12 +169,11 @@ namespace Microsoft.ML.RunTests
                 _normal ? "completed normally" : "aborted",
                 IsPassing ? "passed" : "failed");
 
-            if (_usedSpecificBaselineConfigs.Count > 0)
+            if (_usedSpecificBaselineConfig != null)
             {
-                Log(String.Format("Test {0} is using {1} configuration specific baselines.", TestName,
-                    String.Join(",", _usedSpecificBaselineConfigs)));
+                Log(String.Format("Test {0} is using {1} configuration specific baselines.",
+                    TestName, _usedSpecificBaselineConfig));
             }
-            _usedSpecificBaselineConfigs.Clear();
 
             if (!_normal)
                 Assert.Equal(0, _failures);
@@ -269,7 +267,7 @@ namespace Microsoft.ML.RunTests
                 baselinePath = Path.GetFullPath(Path.Combine(_baselineCommonDir, subDir, baselineConfigDir, name));
                 if (File.Exists(baselinePath))
                 {
-                    _usedSpecificBaselineConfigs.Add(baselinePath);
+                    _usedSpecificBaselineConfig = baselineConfigDir;
                     return baselinePath;
                 }
             }
@@ -285,7 +283,7 @@ namespace Microsoft.ML.RunTests
                 baselinePath = Path.GetFullPath(Path.Combine(_baselineBuildStringDir, subDir, baselineConfigDir, name));
                 if (File.Exists(baselinePath))
                 {
-                    _usedSpecificBaselineConfigs.Add(baselinePath);
+                    _usedSpecificBaselineConfig = baselineConfigDir;
                     return baselinePath;
                 }
             }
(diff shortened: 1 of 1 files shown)


test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -75,6 +75,7 @@ protected BaseTestBaseline(ITestOutputHelper output) : base(output)
         private string _baselineCommonDir;
         private string _baselineBuildStringDir;
         private IEnumerable<string> _baselineConfigDirs;
+        private HashSet<string> _usedSpecificBaselineConfigs;
```
harishsk, line 79: Why do you need a HashSet?

mstfbl: Because more than one specific configuration can be used. For example, using a netcoreapp3.1 and linux-x64 baseline.

harishsk: Sorry I am not following. Can you please walk me through the scenario end to end when the tests are running where you would see the length of `_usedSpecificBaselineConfigs' being greater than 1?

---
In reply to: [415395501](https://github.com/dotnet/machinelearning/pull/5060#discussion_r415395501) [](ancestors = 415395501)

mstfbl: Sure. For example, for `BinaryPriorTest`, if a baseline is available at, say `test\BaselineOutput\Common\PriorPredictor\linux-x64\netcoreapp31\BinaryPrior-TrainTest-breast-cancer-out.txt`, there would be 2 specific configurations for which a baseline is available: `linux-x64` and `netcoreapp31`. In this case, it would be correct to say that the baseline being used is for Linux x64 **and** NET Core 3.1, and hence `_usedSpecificBaselineConfigs` would have length of 2.

Of course, this assumes that there may be baselines available for a test that differs on Linux x64 and NET Core 3.1 together, that are different than the standard baselines. Currently, no such baselines exist. And the current implementation here does not support the case where a baseline would be in the folder `\linux-x64\netcoreapp31\` or `\linux-x64-netcoreapp31\`.

https://github.com/dotnet/machinelearning/blob/bfd2773dd59ac302bf659ed6959738aa68dd5756/test/Microsoft.ML.TestFramework/BaseTestBaseline.cs#L250-L279

https://github.com/dotnet/machinelearning/blob/bfd2773dd59ac302bf659ed6959738aa68dd5756/test/Microsoft.ML.TestFramework/BaseTestBaseline.cs#L114-L151

However, I believe this scenario is possible and can occur in the future, hence this PR's implementation. Of course, it doesn't make sense to use a HashSet if we assume that no two specific configurations can occur at the same time, or that if it does, it will be in a folder named, say `\linux-x64-netcoreapp31\`. But we should decide on how to handle baselines in this case, so that I can keep or change this PR's implementation.

harishsk: > > Test BinaryPriorTest is using linux-x64, netcoreapp31 configuration specific baselines.

Let us assume that different baselines existed for a particular test. Under what conditions will you see this statement in the log? Can you please attempt a repro by simply copying the current baselines into a new directory of the desired configuration?

mstfbl: As I said above, in the current implementation of the code I posted above, you won't ever see this exact statement in the log. This is due to the way in which these configuration-specific paths are checked: it only checks 1 possible configuration at a time.

As you asked, I copied the baseline lines into the directory `test\BaselineOutput\Common\PriorPredictor\win-x64\netcoreapp31\`, but, again, this path was never looked at. The only paths that were checked in `BaseTestBaseline.cs` were:

- test\BaselineOutput\Common\PriorPredictor\win-x64\
- test\BaselineOutput\Common\PriorPredictor\netcoreapp31\
This behavior is expected because again, only 1 specific configuration possibility is being checked at a time.

This is why I said above we should decide whether or not there can be baselines specific to 2 or more specific-configurations (such as a linux-x64-netcoreapp31 baseline). If not, there's no need of a HashSet and a simple string can be used to store a used specific configuration for logging. If there can be 2 or more specific configurations that result in brand-new baselines, then changes to `GetBaselinePath(string subDir, string name)` is necessary to reflect the possibility of there being differing baselines when 2 or more specific configurations are used. @harishsk @frank-dong-ms Do you expect a situation in which 2 specific baselines can result in brand-new baselines?

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -96,6 +97,7 @@ protected override void Initialize()
             _baselineCommonDir = Path.Combine(baselineRootDir, "Common");
             _baselineBuildStringDir = Path.Combine(baselineRootDir, BuildString);
             _baselineConfigDirs = GetConfigurationDirs();
+            _usedSpecificBaselineConfigs = new HashSet<string>();
```
frank-dong-ms, line 100: >_usedSpecificBaselineConfigs [](start = 12, length = 28)

I don't think this is necessary, already covered in _baselineConfigDirs.

You can add any combination to GetConfigurationDirs method and to _baselineConfigDirs.
Say Linux x64 netcoreapp31 then you can add Linux-x64-netcoreapp31 configuration and add baseline file in that configuration folder for specific test if necessary

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -168,6 +170,13 @@ protected override void Cleanup()
                 _normal ? "completed normally" : "aborted",
                 IsPassing ? "passed" : "failed");
 
+            if (_usedSpecificBaselineConfigs.Count > 0)
+            {
+                Log(String.Format("Test {0} is using {1} configuration specific baselines.", TestName,
+                    String.Join(",", _usedSpecificBaselineConfigs)));
```
frank-dong-ms, line 176: >_usedSpecificBaselineConfigs [](start = 37, length = 28)

If this _usedSpecificBaselineConfigs is for better logging then I'm fine with the change

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
frank-dong-ms, line 100: >_usedSpecificBaselineConfigs [](start = 12, length = 28)

Do you have the ability to log configuration SingleDebug and SingleRelease?

harishsk: Unfortunately, SingleDebug and SingleRelease are treated differently right now. 

@mstfbl : After this PR is fixed and merged, can you please create a new PR to merge the logic for handing Debug and Release with the same logic as GetConfigurationDirs()? (Not high priority)

---
In reply to: [416078739](https://github.com/dotnet/machinelearning/pull/5060#discussion_r416078739) [](ancestors = 416078739)

test/Microsoft.ML.TestFramework/BaseTestBaseline.cs
```diff
@@ -268,11 +280,14 @@ protected string GetBaselinePath(string subDir, string name)
                 return baselinePath;
 
             // check again for a platform specific dir
-            foreach (var baselineCOnfigDir in _baselineConfigDirs)
+            foreach (var baselineConfigDir in _baselineConfigDirs)
```
harishsk, line 283: As I mentioned above, this logic is almost a repetition of the above to handle Debug and Release cases. This can be merged with the logic above if we treat Debug and Release as part of the overall configuration. (But in a new PR. I think might complicate the selection order and we should discuss separately)

Issue #5067, opened 2020-04-28
Update NumberOfThreads default value for MatrixFactorization trainer 


LittleLittleCloud:
The default value for MatrixFactorization will use all threads, which might trigger a warning in [libmf](https://github.com/cjlin1/libmf/blob/e70b9a32f7df32cce961bbbb997da074759a16fe/mf.cpp#L4024) when it exceeds two times of nr_bins.

LittleLittleCloud:
https://github.com/dotnet/machinelearning/blob/5a30444f3c85dfafa1f6032813904fbc8e69eae7/src/Microsoft.ML.Recommender/MatrixFactorizationTrainer.cs#L516

Just update `Math.Max(20, 2 * _threads)` to `Math.Max(20, 2 * _threads)-1` should work

harishsk:
@LittleLittleCloud Can you please submit a PR?

Lynx1820:
Thanks for the PR @LittleLittleCloud. Closing this issue.

PR #5071, opened 2020-04-29
Fix MatrixFactorization trainer's warning

LittleLittleCloud:
### Problem to fix
MatrixFactorization's default options will throw a warning msg `Warning: insufficient blocks may slow down the trainingprocess (4*nr_threads^2+1 blocks is suggested)` when running on PC with more than 10 threads.

### Cause
The warning msg is from libmf, which will check nr_bin and nr_thread, if nr_bin <= 2* nr_thread, it will show that warning. 

MatrixFactorization uses `max(20, 2*nr_thread)` to initialize nr_bin, when nr_thread < 10, it's fine because nr_bin will be 20, which is greater than 2\*nr_thread, but if nr_thread is >= 10, nr_bin's value will be 2\*nr_thread, which is less than 2*nr_thread + 1 and trigger warning.

### Fix
initialize nr_bin with 2 * nr_threads + 1. 

### Related issue
#5067

### DIFF
--- a/test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
+++ b/test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
@@ -129,7 +129,7 @@ namespace Microsoft.ML.Tests.TrainerEstimators
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
             {
                 // Linux case
-                var expectedUnixL2Error = 0.612974867782832; // Linux baseline
+                var expectedUnixL2Error = 0.610332110253861; // Linux baseline
                 Assert.InRange(metrices.MeanSquaredError, expectedUnixL2Error - linuxTolerance, expectedUnixL2Error + linuxTolerance);
             }
             else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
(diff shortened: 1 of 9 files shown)


test/Microsoft.ML.Tests/TrainerEstimators/MatrixFactorizationTests.cs
```diff
@@ -96,10 +96,10 @@ public void MatrixFactorizationSimpleTrainAndPredict()
             // MF produce different matrices on different platforms, so check their content on Windows.
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
-                Assert.Equal(0.301269173622131, leftMatrix[0], 5);
-                Assert.Equal(0.558746933937073, leftMatrix[leftMatrix.Count - 1], 5);
-                Assert.Equal(0.27028301358223, rightMatrix[0], 5);
-                Assert.Equal(0.390790820121765, rightMatrix[rightMatrix.Count - 1], 5);
+                Assert.Equal(0.290507137775421, leftMatrix[0], 5);
+                Assert.Equal(0.558072924613953, leftMatrix[leftMatrix.Count - 1], 5);
```
Lynx1820, line 100: Do we know why these baselines are changing?

LittleLittleCloud: Yes, because MF result is corresponding to nr_bins, we change how we initialize nr_bins, so the baseline also changed

PR #5079, opened 2020-05-01
ColumnSelectingTransformer now infers ONNX shape

michaelgsharp:
ColumnSelectingTransformer wasn't inferring the ONNX shape during ONNX export. This was causing issues when the shape needed to be known. This PR adds in the shape inference step by removing the skip flag to skip it.

### DIFF
--- a/src/Microsoft.ML.Data/Transforms/ColumnSelecting.cs
+++ b/src/Microsoft.ML.Data/Transforms/ColumnSelecting.cs
@@ -741,7 +741,7 @@ namespace Microsoft.ML.Transforms
                         continue;
 
                     var srcVariable = ctx.GetVariableName(srcCol.Name);
-                    var dstVariable = ctx.AddIntermediateVariable(dstCol.Type, dstCol.Name, true);
+                    var dstVariable = ctx.AddIntermediateVariable(dstCol.Type, dstCol.Name);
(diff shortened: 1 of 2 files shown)


src/Microsoft.ML.Data/Transforms/ColumnSelecting.cs
```diff
@@ -741,7 +741,7 @@ public void SaveAsOnnx(OnnxContext ctx)
                         continue;
 
                     var srcVariable = ctx.GetVariableName(srcCol.Name);
-                    var dstVariable = ctx.AddIntermediateVariable(dstCol.Type, dstCol.Name, true);
+                    var dstVariable = ctx.AddIntermediateVariable(dstCol.Type, dstCol.Name);
```
Lynx1820, line 744: I think in general, we're trying to move away from shape inferencing, since it makes it harder to debug . Is possible to declare the right shape instead?

michaelgsharp: When true is specified in that spot, the output shape in the graph is null and we can't get its shape. When we remove true, then the correct output shape is listed. I thought that we were moving more towards having the shape listed in the graph instead of not listed, is that not true?

Issue #1216, opened 2018-10-10
Investigate why two predictor tests fail with "Unknown command: 'train'" on x86


tannergooding:
There are currently two `TestPredictors` tests disabled on x86 due to: `[ConditionalFact(typeof(Environment), nameof(Environment.Is64BitProcess))] // x86 fails with "Unknown command: 'train'; Format error at (83,3)-(83,4011): Illegal quoting"`.

This issue needs further investigation.

artidoro:
@tannergooding are you looking into this?

frank-dong-ms:
Following test are marking with this issue but they appear to have different reason of fail:
1. TestMulticlassEnsembleCombiner: this is lightgbm test and should be marked as LightGBMFact and only runs on x64
2. TestTreeEnsembleCombiner: Heap corruption
3. TestTreeEnsembleCombinerWithCategoricalSplits: Heap corruption
4. TestEnsembleCombiner: Heap corruption

PR #5081, opened 2020-05-02
fix x86 crash

frank-dong-ms:
fixes https://github.com/dotnet/machinelearning/issues/1216.

TreeEnsembleCombiner has a bug that causing byte array out of range and corrupts heap

### DIFF
--- a/src/Microsoft.ML.FastTree/TreeEnsemble/TreeEnsembleCombiner.cs
+++ b/src/Microsoft.ML.FastTree/TreeEnsemble/TreeEnsembleCombiner.cs
@@ -65,9 +65,9 @@ namespace Microsoft.ML.Trainers.FastTree
                 foreach (var t in tree.TrainedEnsemble.Trees)
                 {
                     var bytes = new byte[t.SizeInBytes()];
-                    int position = -1;
+                    int position = 0;
                     t.ToByteArray(bytes, ref position);
-                    position = -1;
+                    position = 0;
                     var tNew = new InternalRegressionTree(bytes, ref position);
                     if (paramA != 1)
                     {
--- a/test/Microsoft.ML.Predictor.Tests/TestPredictors.cs
+++ b/test/Microsoft.ML.Predictor.Tests/TestPredictors.cs
@@ -625,7 +625,7 @@ namespace Microsoft.ML.RunTests
             Done();
         }
 
-        [X64Fact("x86 fails. Associated GitHubIssue: https://github.com/dotnet/machinelearning/issues/1216")]
+        [Fact]
         public void TestTreeEnsembleCombiner()
         {
             var dataPath = GetDataPath("breast-cancer.txt");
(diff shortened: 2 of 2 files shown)


src/Microsoft.ML.FastTree/TreeEnsemble/TreeEnsembleCombiner.cs
```diff
@@ -65,9 +65,9 @@ IPredictor IModelCombiner.CombineModels(IEnumerable<IPredictor> models)
                 foreach (var t in tree.TrainedEnsemble.Trees)
                 {
                     var bytes = new byte[t.SizeInBytes()];
-                    int position = -1;
```
harishsk, line 68: Interesting. Can you please explain this issue a bit more? Why does this happen in x64 but not in x86? This is managed memory. Why is it corrupting the unamanaged heap?

frank-dong-ms: Yes, this is index out of range issue that corrupts memory but this one is index was out if range in former (start to use byte array from index -1)...
This memory corrupted (byte array) is allocated in managed but used as unmanaged (from fixed section in C# and pointer) like below:
[https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.FastTree/TreeEnsemble/InternalRegressionTree.cs#L101](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.FastTree/TreeEnsemble/InternalRegressionTree.cs#L101)
[https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.FastTree/Utils/ToByteArrayExtensions.cs#L113](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.FastTree/Utils/ToByteArrayExtensions.cs#L113)
when position is -1, the pointer ((int*)(pBuffer + position)) is accessing memory it should not.


I'm still not sure why this issue not repro in x64. In theory this can also corrupt x64 memory.

---
In reply to: [419011127](https://github.com/dotnet/machinelearning/pull/5081#discussion_r419011127) [](ancestors = 419011127)

sharwell: SZArrays on x64 have a 4-byte padding between the length and the start of the data.

https://github.com/dotnet/runtime/blob/200b197559fc49f91b8db3f20fd6df6b5660d7bf/src/coreclr/src/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.CoreCLR.cs#L306-L321

sharwell: I suspected this was the issue from reading above, but special thanks to @stephentoub for finding the actual code that proves it 👍

harishsk: @eerhardt also confirmed this in email. 
@sharwell, @stephentoub and @eerhardt Thanks for shedding more light on the issue.

PR #5082, opened 2020-05-02
Use KeyTypeAttribute from Schema in CreateTextLoader

### LINKED ISSUE #2642: ReadFromTextFile cannot load data as KeyType
`CreateTextLoader` can read in data as KeyTypes, but `ReadFromTextFile` cannot.

`CreateTextLoader` works like this:
```cs
new TextLoader.Column(matrixColumnIndexColumnName, DataKind.U4, new [] { new TextLoader.Range(1) }, new KeyCount(20)),
```

Now, there is a `KeyType(Count = 20)` attributes that you can put on a class property, but this only seems to be used by scoring. I'd like to be able to use that, or something like that, to read in data directly as KeyType.

aslotte:
Fixes #2642 

This PR reads the KeyTypeAttribute from the schema of the incoming type and sets that on the corresponding column when loading a text file using the CreateTextLoader.

### DIFF
--- a/test/Microsoft.ML.Tests/TextLoaderTests.cs
+++ b/test/Microsoft.ML.Tests/TextLoaderTests.cs
@@ -877,5 +877,45 @@ namespace Microsoft.ML.EntryPoints.Tests
                 Assert.StartsWith("Should define at least one public, readable field or property in TInput.", ex.Message);
             }
         }
+
+        public class BreastCancerInputModelWithKeyType
+        {
+            [LoadColumn(0)]
+            public bool IsMalignant { get; set; }
+
+            [LoadColumn(1), KeyType(10)]
+            public uint Thickness { get; set; }
+        }
+
+        public class BreastCancerInputModelWithoutKeyType
+        {
+            [LoadColumn(0)]
+            public bool IsMalignant { get; set; }
+
+            [LoadColumn(1)]
+            public uint Thickness { get; set; }
+        }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


test/Microsoft.ML.Tests/TextLoaderTests.cs
```diff
@@ -877,5 +877,45 @@ public void TestTextLoaderNoFields()
...
+
+        [Fact]
+        public void TestLoadTextWithKeyTypeAttribute()
+        {
+            var mlContext = new MLContext(seed: 1);
+            string breastCancerPath = GetDataPath(TestDatasets.breastCancer.trainFilename);
+
+            var data = mlContext.Data.CreateTextLoader<BreastCancerInputModelWithKeyType>(separatorChar: ',').Load(breastCancerPath);
+
+            Assert.True(data.Schema[1].Type.GetKeyCount() == 10);
```
eerhardt, line 907: Instead of `Assert.True(something == expected)`, it is better to use `Assert.Equal(expected, actual)`. The reason is when it fails, xunit will print what the expected and actual values are, which helps figure out the issue, without having to re-run the test and debug.

aslotte: Great point @eerhardt, thanks for the review. Pushed the recommended change.

harishsk:
@aslotte Thank you for the contribution. I have merged your PR.

aslotte:
Thanks all!

PR #5090, opened 2020-05-05
Adding vector tests for KeyToValue and ValueToKey

Lynx1820:


harishsk:
>             {

nit: Please add a newline above.

But more importantly, ValueToKeyMapping (and related) transforms have caused a bunch of issues in Nimbus. Can you please run the Nimbus suite of tests and make sure that nothing is broken there as a result of these changes?

---
Refers to: src/Microsoft.ML.Data/Transforms/ValueToKeyMappingTransformer.cs:813 in ebea3a0. [](commit_id = ebea3a04ad963a91be9f284473c428a9b7da40da, deletion_comment = False)

PR #5104, opened 2020-05-07
Support more types for HashEstimator

wangyems:
-Verified compatibility(change required in ML.NET in this PR) between ORT(https://github.com/microsoft/onnxruntime/pull/3827) and ML.NET
-Some changes made in a couple of 64-bit hash function because the existed implementation is not exactly the same as MurmurHash3_x86_32
-bump to ort1.3 pre-release and fix some test cases


src/Microsoft.ML.Data/Transforms/Hashing.cs
```diff
@@ -599,8 +599,7 @@ public uint HashCore(uint seed, uint mask, in double value)
                 ulong v = FloatUtils.GetBits(value == 0 ? 0 : value);
                 var hash = Hashing.MurmurRound(seed, Utils.GetLo(v));
                 var hi = Utils.GetHi(v);
-                if (hi != 0)
-                    hash = Hashing.MurmurRound(hash, hi);
+                hash = Hashing.MurmurRound(hash, hi);
```
yaeldekel, line 602: >hash = Hashing.MurmurRound(hash, hi); [](start = 16, length = 37)

If you change the hashing function, you need to bump the version of the model, while maintaining backwards compatibility.

justinormont: If there isn't a unit test for backwards compatibility: could add one which loads an old model containing a `CategoricalHash` transform, runs a row of data through it, and checks for equality on its output (from the `CateoricalHash`).

wangyems: Agree. Since this is a draft PR, I just modified the existing test cases and wait for ort 1.3 release. I'll add a separate test specifically for testing new types once it's time to bring this PR official

wangyems:
A weird test, haven't figure out which hash function it calls, but can pass 32bits test cases
HashTests.cs L253:
HashTestCore((ulong)value, new KeyDataViewType(typeof(ulong), ulong.MaxValue - 1), eKey, eoKey, e3Key);

src/Microsoft.ML.Data/Transforms/Hashing.cs
wangyems: synced with Harish that the change will be made just on V2 hash function for 64bit(not released for V2 so backward compatibility is not required this time)

yaeldekel: I think it's better to be on the safe side, since users can theoretically use nightly versions (or build their own version of ML.NET from the source code), and create models using a version from between releases. 

Either way, the original implementations of `IHasher` (those that don't have "V2" in their name) should not change.

---
In reply to: [425509011](https://github.com/dotnet/machinelearning/pull/5104#discussion_r425509011) [](ancestors = 425509011)

harishsk: The nightly versions are the "bleeding edge". We do not make any supportability guarantees for the nightly versions. 

---
In reply to: [425607553](https://github.com/dotnet/machinelearning/pull/5104#discussion_r425607553) [](ancestors = 425607553,425509011)

wangyems: > I think it's better to be on the safe side, since users can theoretically use nightly versions (or build their own version of ML.NET from the source code), and create models using a version from between releases.
> 
> Either way, the original implementations of `IHasher` (those that don't have "V2" in their name) should not change.
> 
> In reply to: [425509011](https://github.com/dotnet/machinelearning/pull/5104#discussion_r425509011) [](ancestors = 425509011)

Those non-V2s have been reverted to original

harishsk:
>   

Is this file still part of the change list? If so, can you please revert this?

---
Refers to: Directory.Build.props:2 in 792703d. [](commit_id = 792703df4eb4868883ea06079ae92fc1137d706c, deletion_comment = False)

test/Microsoft.ML.Tests/Transformers/HashTests.cs
```diff
@@ -261,9 +278,20 @@ private void HashTestPositiveIntegerCore(ulong value, uint expected, uint expect
         [Fact]
         public void TestHashIntegerNumbers()
         {
+            //32bit
+            //HashTestPositiveInteger32BitCore(0, 842, 358, 20);
+            //HashTestPositiveInteger32BitCore(1, 502, 537, 746);
+            //HashTestPositiveInteger32BitCore(2, 407, 801, 652);
+
```
harishsk, line 285: Please remove the commented lines here if they are not meant to be part of the code.

Issue #673, opened 2018-08-13
.NET data type system instead of DvTypes


codemzs:
# .NET data type system instead of DvTypes
## Motivation
Machine Learning datasets often have missing values and to accommodate them along with C# native
types without increasing the memory footprint DvType system was created. If we were to use
`Nullable<T>` then we are looking at additional memory for `HasValue` boolean field plus another 3
bytes for 4 byte alignment. The C# native types that are replaced using DvTypes are bool as DvBool,
sbyte as DvInt1, int16 as DvInt2, int32 as DvInt4, int64 as DvInt8, DvDateTime as System.DateTime,
DvDateTimeZone as combination of DvDateTime and DvInt2 offset, DvTimeSpan as SysTimeSpan and string
as DvText. Float and Double types already have a special value called NaN that can be used for
missing value. DvType system achieves a smaller memory footprint by denoting special value for
missing value which is usually the smallest number that can be represented by the native type that
is encapsulated by DvType, example, DvInt1's missing value indicator would be SByte.MinValue and in
the case of types that represent date/time types it is a value that represent maximum ticks. 

We plan to remove DvTypes to make IDataView a general commodity that can be used in other products
and for this to happen it would be nice if it did not having a dependency on a special type system.
If in future we find having DvTypes was useful then we can consider exposing it natively from .NET
platform. Once we remove DvTypes then ML.NET platform will be using native non-nullable C# types.
Float or double types can be used to represent missing value.

## Column Types 
Columns in ML.NET make up the dataset and `ColumnType` defines a column. At high level there are two
kinds of column, first is `PrimitiveType` and that comprises of types such as `NumberType`,
`BoolType`, `TextType`, `DateTimeType`, `DateTimeZoneType`, `KeyType`, second is `Structured type`
and it comparises of `VectorType`. `ColumnType` is primarily made up of `Type` and `DataKind`.
`Type` could refer to any type but it is instantiated with a type referred by `DataKind` which is an
identifer for data types that comprises of DvTypes, native C# types such as float, double and custom
big integer UInt128.  

## Type conversion
DvTypes have implicit and explicit override for assignment operator that handles type conversion.
Lets consider DvInt1 for example:

| To  | From | Current behavior
|:-:|:-:|:-:
| DvInt1 | sbyte | Copy the value as it is 
| DvInt1 | sbyte? | Assign missing value if null otherwise copy the value as it is 
| sbyte | DvInt1 | Copy if not a missing value otherwise throw exception 
[...]

PR #5106, opened 2020-05-07
added in standard conversions from types to ReadOnlyMemory<char>

michaelgsharp:
Currently in the standard transformations we don't support converting types to a string representation. Since all C# types support `ToString()` functionality, this PR adds in standard transformations to type `ReadOnlyMemory<char>` into the `TypeConvertingTransformer`.

### DIFF
--- a/src/Microsoft.ML.Data/Data/Conversion.cs
+++ b/src/Microsoft.ML.Data/Data/Conversion.cs
@@ -937,8 +937,8 @@ namespace Microsoft.ML.Data.Conversion
         public void Convert(in U4 src, ref TX dst) => dst = src.ToString().AsMemory();
         public void Convert(in U8 src, ref TX dst) => dst = src.ToString().AsMemory();
         public void Convert(in UG src, ref TX dst) => dst = string.Format("0x{0:x16}{1:x16}", src.High, src.Low).AsMemory();
-        public void Convert(in R4 src, ref TX dst) => dst = src.ToString().AsMemory();
-        public void Convert(in R8 src, ref TX dst) => dst = src.ToString().AsMemory();
+        public void Convert(in R4 src, ref TX dst) => dst = src.ToString("G7", CultureInfo.InvariantCulture).AsMemory();
+        public void Convert(in R8 src, ref TX dst) => dst = src.ToString("G17", CultureInfo.InvariantCulture).AsMemory();
         public void Convert(in BL src, ref TX dst) => dst = src.ToString().AsMemory();
         public void Convert(in TS src, ref TX dst) => dst = string.Format("{0:c}", src).AsMemory();
         public void Convert(in DT src, ref TX dst) => string.Format("{0:o}", src).AsMemory();
--- a/src/Microsoft.ML.Data/Data/Conversion.cs
+++ b/src/Microsoft.ML.Data/Data/Conversion.cs
@@ -114,6 +114,7 @@ namespace Microsoft.ML.Data.Conversion
             AddStd<I1, R8>(Convert);
             AddAux<I1, SB>(Convert);
             AddStd<I1, BL>(Convert);
+            AddStd<I1, TX>(Convert);
 
             AddStd<I2, I1>(Convert);
             AddStd<I2, I2>(Convert);
@@ -123,6 +124,7 @@ namespace Microsoft.ML.Data.Conversion
             AddStd<I2, R8>(Convert);
             AddAux<I2, SB>(Convert);
             AddStd<I2, BL>(Convert);
+            AddStd<I2, TX>(Convert);
 
             AddStd<I4, I1>(Convert);
             AddStd<I4, I2>(Convert);
@@ -132,6 +134,7 @@ namespace Microsoft.ML.Data.Conversion
             AddStd<I4, R8>(Convert);
             AddAux<I4, SB>(Convert);
             AddStd<I4, BL>(Convert);
+            AddStd<I4, TX>(Convert);
 
             AddStd<I8, I1>(Convert);
             AddStd<I8, I2>(Convert);
@@ -141,6 +144,7 @@ namespace Microsoft.ML.Data.Conversion
             AddStd<I8, R8>(Convert);
             AddAux<I8, SB>(Convert);
             AddStd<I8, BL>(Convert);
+            AddStd<I8, TX>(Convert);
 
             AddStd<U1, U1>(Convert);
             AddStd<U1, U2>(Convert);
@@ -151,6 +155,7 @@ namespace Microsoft.ML.Data.Conversion
             AddStd<U1, R8>(Convert);
             AddAux<U1, SB>(Convert);
             AddStd<U1, BL>(Convert);
+            AddStd<U1, TX>(Convert);
 
             AddStd<U2, U1>(Convert);
             AddStd<U2, U2>(Convert);
@@ -161,6 +166,7 @@ namespace Microsoft.ML.Data.Conversion
             AddStd<U2, R8>(Convert);
             AddAux<U2, SB>(Convert);
             AddStd<U2, BL>(Convert);
+            AddStd<U2, TX>(Convert);
 
             AddStd<U4, U1>(Convert);
             AddStd<U4, U2>(Convert);
@@ -171,6 +177,7 @@ namespace Microsoft.ML.Data.Conversion
             AddStd<U4, R8>(Convert);
             AddAux<U4, SB>(Convert);
             AddStd<U4, BL>(Convert);
+            AddStd<U4, TX>(Convert);
 
             AddStd<U8, U1>(Convert);
             AddStd<U8, U2>(Convert);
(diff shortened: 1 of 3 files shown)


michaelgsharp:
After syncing with @harishsk, we thought that since all C# types natively support a `ToString()`, that just adding it in to the standard conversions made sense. @codemzs and @eerhardt we wanted your thoughts on that. If you think its ok we will leave it as is. If not, we can move it into the aux conversions and add extra helper methods as needed to match what we have for the standard conversions.

harishsk:
@michaelgsharp Please look at the test failures. You may need to update the baselines as a result of this change.

michaelgsharp:
You are correct Harish, once this goes in we can remove the ToStringTransformer.

test/Microsoft.ML.Tests/Transformers/ConvertTests.cs
```diff
@@ -121,7 +121,7 @@ private sealed class TestStringClass
             public string A;
         }
 
-        [Fact]
+        [Fact, TestCategory("RunSpecificTest")]
```
harishsk, line 124: Should this be reverted?

michaelgsharp: Yes, it should. Reverting now.

harishsk:
Can you also please add a note in docs\code\IDataViewTypeSystem.md that we conversions to string are now supported?

src/Microsoft.ML.Data/Data/Conversion.cs
```diff
@@ -912,6 +927,24 @@ public void Convert(in BL src, ref SB dst)
...
+        public void Convert(in U2 src, ref TX dst) => dst = src.ToString().AsMemory();
+        public void Convert(in U4 src, ref TX dst) => dst = src.ToString().AsMemory();
+        public void Convert(in U8 src, ref TX dst) => dst = src.ToString().AsMemory();
+        public void Convert(in UG src, ref TX dst) => dst = string.Format("0x{0:x16}{1:x16}", src.High, src.Low).AsMemory();
+        public void Convert(in R4 src, ref TX dst) => dst = src.ToString("G7", CultureInfo.InvariantCulture).AsMemory();
+        public void Convert(in R8 src, ref TX dst) => dst = src.ToString("G17", CultureInfo.InvariantCulture).AsMemory();
+        public void Convert(in BL src, ref TX dst) => dst = src.ToString().AsMemory();
+        public void Convert(in TS src, ref TX dst) => dst = string.Format("{0:c}", src).AsMemory();
+        public void Convert(in DT src, ref TX dst) => string.Format("{0:o}", src).AsMemory();
+        public void Convert(in DZ src, ref TX dst) => string.Format("{0:o}", src).AsMemory();
```
codemzs, line 945: What is the motivation behind this change? 

My concern here is this could potentially increase garbage collector pressure, however, in the case of a transformer you could control that pressure by re-using the buffer.

harishsk: Microsoft.ML.Featurizers has a few transformers that have overlapping functionality with transformers that are already present in ML.NET. The motivation for this change is to bring in the those featurizers and merge their functionality with existing transformers. The ToStringTransformer only implements conversion from basic data types to string. So we are looking to merge it with the more general TypeConvertingTransformer by adding string support for it.

In a transformer, can we truly reuse the buffer? Wouldn't we hand over the reference to the string back to the caller after the conversion? 

Do you see this conversion function being in the hot path to cause  garbage collector pressure?

Or do you think this garbage collection concern is serious enough that we should have string conversion being done in a separate transformer with a separate API?

---
In reply to: [423448756](https://github.com/dotnet/machinelearning/pull/5106#discussion_r423448756) [](ancestors = 423448756)

codemzs: The caller hands over a reference to the transfomer which is why the getters of the transformers take output as `ref` type, for example look at the [getter ](https://github.com/dotnet/machinelearning/blob/a2406f64c96bb2e5bd7e7c9ba91487e31d673d6b/src/Microsoft.ML.ImageAnalytics/ImageResizer.cs#L292)for ImageLoader Transform. We reuse buffer and keep the GC pressure low, this is a fundamental principle in ML .NET that gives all the performance benefits you have been outline in [KDD'2019 publication](https://www.kdd.org/kdd2019/accepted-papers/view/machine-learning-at-microsoft-with-ml.net).

I have a reason to believe this can be a hot path, may be if you can share benchmarks to show me this isn't by actually using several string conversions using PR vs using transformer where you can reuse the buffer then it would be great. The last type we overhauled the type system we think extensive benchmark and testing that you can see [here](https://github.com/dotnet/machinelearning/issues/673). I expect the same anytime type system is changed.

michaelgsharp: Hey Zeeshan,

In this case the caller hands over a `ReadOnlyMemory<char>` reference. From my understanding, since its readonly, we cannot modify it or the underlying buffer, and must allocate a new one. I also looked at the documentation for [ReadOnlyMemory](https://docs.microsoft.com/en-us/dotnet/api/system.readonlymemory-1?view=netcore-3.1) and couldn't see any way to modify it at all. Do you know of a way to modify it? Or do you know if we have any examples in code I can use as a reference? I was unable to find anything.

Assuming we can't modify it and need to return a new instance, I did some testing converting an int to a string (not in ML.Net) and then returning a `ReadOnlyMemory<char>` pointing to it. 

The first way took a `ref ReadOnlyMemory<char>` and assigned it to `value.ToString().AsMemory()`. 
```
for (int x = 0; x < NumberOfIterations; x++)
{
    // readOnlyMemory is passed in by ref
    readOnlyMemory = x.ToString().AsMemory();
}
```

The second way was to also take a `ref ReadOnlyMemory<char>` but also a ref `StringBuilder` and use the StringBuilder to create the string and memory. 
```
for (int x = 0; x < NumberOfIterations; x++)
{
    // stringBuilder and readOnlyMemory are passed in by ref.
    // the stringBuilder is using Clear, which should keep the underlying buffer
    stringBuilder.Clear();
    stringBuilder.Append(x);
    readOnlyMemory = stringBuilder.ToString().AsMemory();
}
```

In Release mode with 214 million allocations, the first approach of `ToString().AsMemory()` took about 6 seconds on average where the second approach took about 8 seconds on average. Unless there is another faster way, which I couldn't find, the fastest way to convert an int to `ReadOnlyMemory<char>` is the first approach. Each approach also used about 13 MB of memory total.

I have also created a benchmark in ML.Net that uses the first approach to create the `ReadOnlyMemory<char>`.  It ran 10 million rows in 1.25 seconds, and the entire test only used 30 MB of memory (I am assuming the majority of this memory is from ML.Net itself as the first tests I did had more rows and used less memory). I have not done any other benchmarks tests as of yet.

@codemzs does this address your concern? Or is there something else I can do to address it? Or do you know a better way to deal with the `ReadOnlyMemory<char>`?

@eerhardt do you know of a better way to handle the `ReadOnlyMemory<char>?

eerhardt: With ML.NET's current infrastructure, this is probably as good as you can make it. But in later versions of .NET Core (if we wanted to start targeting them), you could reshare the buffers in some cases by using the following methods:

https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal.trygetarray
and
https://docs.microsoft.com/en-us/dotnet/api/system.int32.tryformat

The idea being something like this:

```C#

public void Convert(in I4 src, ref TX dst)
{
    if (MemoryMarshal.TryGetArray(dst, out ArraySegment<char> segment))
    {
        if (src.TryFormat(segment.Array, out int charsWritten))
        {
             dst = segment.Array.AsMemory(0, charsWritten);
             return;
        }
    }

    char[] newBuffer = new char[11]; // 11 is the most characters needed to represent all 32-bit ints
    src.TryFormat(newBuffer, , out int charsWritten);
    dst = newBuffer.AsMemory(0, charsWritten);
}

```

What's dangerous about this approach is that `dst` might be someone else's array that we are clobbering over in order to reuse it. To prevent that, we could write a little sentinel value at the beginning of the array to mark it as ours. Say something like 2 `char.MaxValue` elements. And then we could check if the array we got from `MemoryMarshal.TryGetArray` was "ours" by checking the sentinel.

eerhardt: Why don't we want to use the StringBuilder overloads directly above this block?

harishsk: In this particular case the conversion being sought is from some data type like `int` to `string` and not to `StringBuilder`. If we wanted to use the above conversions, we would have to create a local `StringBuilder` convert the given data type to `StringBuilder` and then convert it back to `string`.  @michaelgsharp has discovered this method is slower. 

---
In reply to: [428265503](https://github.com/dotnet/machinelearning/pull/5106#discussion_r428265503) [](ancestors = 428265503)

harishsk: It appears that those APIs are supported on .NET Core 3.1 and 2.1 which ML.NET supports but not supported on netfx461 which ML.NET needs. 
In that case, is it okay to go ahead with this PR until those APIs are fully supported on all ML.NET configs?


---
In reply to: [428262800](https://github.com/dotnet/machinelearning/pull/5106#discussion_r428262800) [](ancestors = 428262800)

